PR #314, opened 2018-04-23
SPEC-1064: Fix insertedIds index in expected bulk write result

jmikola:
https://jira.mongodb.org/browse/SPEC-1064

### DIFF
--- a/source/retryable-writes/tests/bulkWrite.yml
+++ b/source/retryable-writes/tests/bulkWrite.yml
@@ -294,7 +294,7 @@ tests:
             error: true
             result:
                 deletedCount: 0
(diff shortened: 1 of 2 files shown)


source/retryable-writes/tests/bulkWrite.yml
```diff
@@ -294,7 +294,7 @@ tests:
             error: true
             result:
                 deletedCount: 0
-                insertedIds: { 0: 2 }
+                insertedIds: { 1: 2 }
```
jmikola: Drivers that don't report `insertedIds` in their bulk write result (e.g. libmongoc) would not have caught this mistake.

PR #318, opened 2018-04-25
Adding mixed case hostname tests in SDAM

rathisekaran:
SPEC-915


ajdavis:
The YML looks good, and PyMongo passes the new test. Please generate a JSON file according to instructions in the repo's top-level README and add the JSON file to the repo..

PR #319, opened 2018-04-27
SPEC-1063 writeConcernErrors are retryable

kevinAlbs:
Props to Shane for hashing out the details.

### DIFF
--- a/source/retryable-writes/retryable-writes.rst
+++ b/source/retryable-writes/retryable-writes.rst
@@ -106,6 +106,8 @@ Retryable Error
      listed above, but with an error message containing the substring "not
      master" or "node is recovering"
 
+   - a server response with a write concern error meeting the above criteria
+
    The criteria for retryable errors is similar to the discussion in the SDAM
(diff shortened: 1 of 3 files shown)


source/retryable-writes/retryable-writes.rst
```diff
@@ -106,6 +106,23 @@ Retryable Error
      listed above, but with an error message containing the substring "not
      master" or "node is recovering"
 
+   - a server response with a write concern error with any of the following
```
jyemin: This is a subset of the list of error code for normal server errors.  Would it be incorrect to use the same list of error codes for both?  Are we explicitly saying that none of the others from above would be considered retryable if they happened to appear in a write concern error, or just that currently none of the others would ever be sent by MongoDB as a write concern error?  If the latter, consider just merging the two lists into one and saying that it applied to both normal server errors and write concern errors.

kevinAlbs:
Added a test description to server-discovery-and-monitoring-tests.rst. @jyemin PTAL

PR #320, opened 2018-04-29
Add clientOptions to retryable writes tests

ajdavis:
Bring retryable writes tests closer to transactions tests: specify retryWrites in the clientOptions object instead of in the test's README. Also, update Command Monitoring bulk format to match the Retryable Writes format. A step toward a unified CRUD/APM test syntax we can use for most spec tests.

### DIFF
--- a/source/command-monitoring/tests/unacknowledgedBulkWrite.yml
+++ b/source/command-monitoring/tests/unacknowledgedBulkWrite.yml
@@ -12,9 +12,10 @@ tests:
       name: "bulkWrite"
       arguments:
         requests:
-          - insertOne: 
+          - name: "insertOne"
+            arguments:
               document: { _id: "unorderedBulkWriteInsertW0", x: 44 }
-        ordered: false
+        options: { ordered: false }
         writeConcern: { w: 0 }
     expectations:
       -
--- a/source/retryable-writes/tests/README.rst
+++ b/source/retryable-writes/tests/README.rst
@@ -15,10 +15,11 @@ that drivers can use to prove their conformance to the Retryable Writes spec.
 Several prose tests, which are not easily expressed in YAML, are also presented
 in this file. Those tests will need to be manually implemented by each driver.
 
-Tests will require a MongoClient with ``retryWrites`` enabled. Integration tests
-will require a running MongoDB cluster with server versions 3.6.0 or later. The
-``{setFeatureCompatibilityVersion: 3.6}`` admin command will also need to have
-been executed to enable support for retryable writes on the cluster.
+Tests will require a MongoClient created with options defined in the tests.
+Integration tests will require a running MongoDB cluster with server versions
+3.6.0 or later. The ``{setFeatureCompatibilityVersion: 3.6}`` admin command
+will also need to have been executed to enable support for retryable writes on
+the cluster.
 
 Server Fail Point
 =================
@@ -137,6 +138,8 @@ Each YAML file has the following keys:
 
   - ``description``: The name of the test.
 
+  - ``clientOptions``: Parameters to pass to MongoClient().
+
   - ``failPoint``: Document describing options for configuring the
     ``onPrimaryTransactionalWrite`` fail point on the primary server. This
     document should be merged with the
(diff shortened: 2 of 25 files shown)


ajdavis:
Merged in https://github.com/mongodb/specifications/commit/c0ce96daac5661f2603bfa52c9b3f683c176a292 and https://github.com/mongodb/specifications/commit/bebf9b688f64afdec46a860f102bad0841b04c02

source/command-monitoring/tests/unacknowledgedBulkWrite.yml
```diff
@@ -12,9 +12,10 @@ tests:
       name: "bulkWrite"
       arguments:
         requests:
-          - insertOne: 
+          - name: "insertOne"
+            arguments:
               document: { _id: "unorderedBulkWriteInsertW0", x: 44 }
-        ordered: false
+        options: { ordered: false }
```
mbroadst: why use a different format in the YAML here? Shouldn't this just be:
```
options:
  ordered: false
```
?

source/command-monitoring/tests/unacknowledgedBulkWrite.yml
```diff
@@ -12,9 +12,10 @@ tests:
...
         writeConcern: { w: 0 }
```
mbroadst: I find this confusing about the other test runner as well, but I suspect its because of python kwargs. If we are introducing an `options` field, it stands to reason that we would put all of our options there. Why do we special case some fields like `writeConcern`?

source/retryable-writes/tests/README.rst
```diff
@@ -137,15 +138,17 @@ Each YAML file has the following keys:
...
   - ``description``: The name of the test.
 
+  - ``clientOptions``: Parameters to pass to MongoClient().
+
   - ``failPoint``: Document describing options for configuring the
     ``onPrimaryTransactionalWrite`` fail point on the primary server. This
     document should be merged with the
     ``{ configureFailPoint: "onPrimaryTransactionalWrite" }`` command document.
 
   - ``operation``: Document describing the operation to be executed. The
```
mbroadst: please document the newly added `options` field

source/command-monitoring/tests/unacknowledgedBulkWrite.yml
jmikola: This is kosher YAML, and was already used elsewhere where the sub-object could easily be expressed on one line. It could also have been done for the `clientOptions` keys, since those are only defining a `retryWrites` option.

source/command-monitoring/tests/unacknowledgedBulkWrite.yml
jmikola: I believe this is because `ordered` is always an option when creating a bulk write object, and write concern is either inherited from the parent collection or specified per-option (in some drivers) at the time the bulk write is executed. Placing these together in `options` would require the test harness to pull out certain options.

source/retryable-writes/tests/README.rst
jmikola: The `options` field above was only added to the command monitoring spec tests, which doesn't discuss the test format in its [README](https://github.com/mongodb/specifications/blob/master/source/command-monitoring/tests/README.rst). That said, the retryable write [README](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/README.rst) also doesn't discuss this format, as it only explains as far down as `tests.operation.arguments` and says that those arguments should correspond to names in the CRUD spec.

source/command-monitoring/tests/unacknowledgedBulkWrite.yml
ajdavis: Right: at least in PyMongo, this gets implemented like:

```
wc = ... get writeConcern from test ...
options = ... get options like "ordered" from test ...
collection = collection.with_options(write_concern=wc)
collection.bulk_write(requests, options)
```
This is consistent with the CRUD Spec, which puts "ordered" in the options passed to bulk_write(), but doesn't put writeConcern there.

Anyway, my goal here is a small step towards consistency among test runners. If you want, I can cancel this in favor of one big change later this summer that unifies them all.

PR #321, opened 2018-05-03
 SPEC-1071 Test more retryable write errors

ShaneHarvey:
Test that writes are retried after Interrupted, InterruptedAtShutdown, and InterruptedDueToReplStateChange.

Unfortunately all the other error codes (NotMaster, ShutdownInProgress, etc..) are reported in the "writeErrors" array with this failpoint. For example:
```
{
	"n" : 1,
	"writeErrors" : [
		{
			"index" : 0,
			"code" : 9001,
			"errmsg" : "Failing write for 9c6992a7-12a8-4dc5-ab61-c950879c45d0 - 47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=:11 due to failpoint. The write must not be reflected."
		}
	],
	"ok" : 1
}
```

Only Interrupted, InterruptedAtShutdown, and InterruptedDueToReplStateChange are reported as command failures:
```
{
	"ok" : 0,
        "code": 11601,
        "errmsg": "Failing write for 8858ab46-1aba-4f81-a803-87ae6b38ca76 - 47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=:2 due to failpoint. The write must not be reflected."
}
```


ShaneHarvey:
I think we can use the newly added "failCommand" failpoint (SERVER-34551) for all the other error codes. It would require >=3.7.10.

ajdavis:
Interesting that the error code you pass to the server in enableFailPoint affects where the error appears in the reply structure. Let's use the new failpoint. We need to use YAML to restrict the new tests to new servers, and introduce a new bit of YAML syntax to tell test runners *which* failpoint to enable. The latter is something we need to do for Transactions tests anyway.

ShaneHarvey:
> Interesting that the error code you pass to the server in enableFailPoint affects where the error appears in the reply structure. Let's use the new failpoint. We need to use YAML to restrict the new tests to new servers, and introduce a new bit of YAML syntax to tell test runners which failpoint to enable. The latter is something we need to do for Transactions tests anyway.

Yeah I agree. Interesting to note the new failCommand failpont totally superscedes onPrimaryTransactionalWrite. I think failCommand is what drivers really wanted in 3.6.

jmikola:
> Interesting to note the new failCommand failpont totally superscedes onPrimaryTransactionalWrite. I think failCommand is what drivers really wanted in 3.6.

Does this avoid the oddity with `onPrimaryTransactionalWrite` never kicking in with commands that omit a transaction ID (came up in recent UpdateMany and DeleteMany tests)?

source/retryable-writes/tests/insertOne.yml
```diff
@@ -53,3 +53,58 @@ tests:
                 data:
                     - { _id: 1, x: 11 }
                     - { _id: 2, x: 22 }
+    -
+        description: "InsertOne succeeds after InterruptedDueToReplStateChange"
+        failPoint:
```
jmikola: I might suggest leaving this `failPoint` key as-is and just adding `configureFailPoint: NAME` as the first key. This will effectively allow drivers to use the `failPoint` document in the test case as the command document directly.

jmikola:
> I think we can use the newly added "failCommand" failpoint (SERVER-34551) for all the other error codes. It would require >=3.7.10.

Would it make sense to leave the current tests as-is and move these new tests using `failCommand` into their own file is an appropriate `minServerVersion`?

Either way, I think all spec tests would get updated to support the new `failPoint` syntax (so each test case can specify which fail point to use).

source/retryable-writes/tests/insertOne.yml
ShaneHarvey: Thanks SGTM!

ShaneHarvey:
Added `configureFailPoint` to all 'failPonts' and moved the new tests into `retryableErrors.yml` requiring 3.7.10.

PR #322, opened 2018-05-15
reorder transactions bulk test to avoid specifying how mixed operatio…

saghm:
…ns should combine

The Ruby driver unfortunately does not currently combine `deleteOne` and `deleteMany` operations in a bulk write to execute as a single command. While this needs to be fixed, I think it would make sense for the transactions tests to be agnostic to this given that the CRUD spec currently doesn't define the behavior of bulk operations.

### DIFF
--- a/source/transactions/tests/bulk.yml
+++ b/source/transactions/tests/bulk.yml
@@ -70,19 +70,21 @@ tests:
             - name: deleteOne
               arguments:
                 filter: {_id: 4}
-            - name: deleteMany
-              arguments:
-                filter: {_id: {$gte: 6}}
             - name: updateMany
               arguments:
                 filter: {_id: {$gte: 2}}
                 update: {$set: {z: 1}}
+            # Keep deleteMany segregated from deleteOne, so that drivers that aren't able to coalesce
+            # adjacent mixed deletes into a single delete command will still pass this test
+            - name: deleteMany
+              arguments:
+                filter: {_id: {$gte: 6}}
           session: session0
         result:
           deletedCount: 4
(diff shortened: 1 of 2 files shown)


source/transactions/tests/bulk.yml
```diff
@@ -70,19 +70,19 @@ tests:
...
               arguments:
                 filter: {_id: 4}
-            - name: deleteMany
-              arguments:
-                filter: {_id: {$gte: 6}}
             - name: updateMany
               arguments:
                 filter: {_id: {$gte: 2}}
                 update: {$set: {z: 1}}
+            - name: deleteMany
```
ShaneHarvey: Can you add a comment explaining why deleteMany needs to separated from deleteOne (similar to the replaceOne comment above)?

ShaneHarvey:
Also make sure you run the make file to convert your update yml file to json.

source/transactions/tests/bulk.yml
```diff
@@ -70,19 +70,21 @@ tests:
...
                 filter: {_id: 4}
-            - name: deleteMany
-              arguments:
-                filter: {_id: {$gte: 6}}
             - name: updateMany
               arguments:
                 filter: {_id: {$gte: 2}}
                 update: {$set: {z: 1}}
+            # Keep deleteMany segregated from deleteOne, so that drivers that aren't able to coalesce
+            # adjacent updates and replaces into a single delete command will still pass this test
```
ShaneHarvey: "updates and replaces" -> "deletes and deleteManys"

saghm: Shoot, I could have sworn I updated all of that

PR #323, opened 2018-05-16
SPEC-1066: Support nameOnly option in listCollections command

rathisekaran:


source/enumerate-collections.rst
```diff
@@ -13,7 +13,7 @@ Enumerating Collections
 :Type: Standards
 :Server Versions: 1.8-2.7.5, 2.8.0-rc3 and later
 :Last Modified: September 25, 2017
-:Version: 0.3.1
+:Version: 0.4
```
jmikola: Last Modified should be bumped as well, before merging.

source/enumerate-collections.rst
```diff
@@ -265,7 +266,17 @@ Example return::
         "system.replset"
     ]
 
-Drivers MUST strip the database name from the returned collection names.
+
+
+Drivers MAY implement a MongoClient method that returns an Iterable of strings,
```
jmikola: `listCollections` would be a method on the Database class. https://github.com/mongodb/specifications/blob/master/source/enumerate-collections.rst#driver-methods already mentions that all of the methods in this spec "should be on the database class".

If you're going to use "Iterable" here, you should probably copy over the term definition from https://github.com/mongodb/specifications/blob/master/source/enumerate-databases.rst#terms (and any other relevant terms, such as "MongoDatabase"). You may also want to add "MongoCollection" if a later section of this spec is going to talk about a method that returns an iterable of those objects.

source/enumerate-collections.rst
```diff
@@ -303,6 +314,8 @@ Example return (a cursor which returns documents, not a simple array)::
 When returning this information as a cursor, a driver SHOULD use the
 method name ``listCollections`` or an idiomatic variant.
 
+Drivers SHOULD specify the nameOnly option when executing the ``listCollections`` command for this method.
```
ShaneHarvey: I don't think this method should specify nameOnly since it's about getting the full collection information. Perhaps drivers SHOULD/MAY allow the nameOnly option to be passed?

source/enumerate-collections.rst
```diff
@@ -320,6 +333,8 @@ Example return (in PHP, but abbreviated)::
       [5] => class MongoCollection#11 { }
     }
 
+Drivers SHOULD specify the nameOnly option when executing the ``listCollections`` command for this method.
```
ShaneHarvey: MUST instead of SHOULD? Also can you format ``nameOnly``?

source/enumerate-collections.rst
```diff
@@ -265,7 +266,17 @@ Example return::
         "system.replset"
     ]
 
-Drivers MUST strip the database name from the returned collection names.
+
+
+Drivers MAY implement a MongoClient method that returns an Iterable of strings,
+each of which corresponds to a name in the collections array of the
```
jmikola: Before the `listCollections` command, drivers simply queried `system.namespaces`, so there was no results array to speak of. There also technically isn't one when the command returns a cursor, since drivers construct a cursor from that response and iterate on _that_.

I think the most straightforward language here is to simply say that the result is an iterable of strings, where each string corresponds to a collection name (without discussing whether or not we're using a command). You can certainly keep the sentence about the method name below.

source/enumerate-collections.rst
```diff
@@ -251,8 +251,9 @@ All methods:
 Getting Collection Names
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-Drivers MAY implement a method to enumerate all collections, and return only
-the collection names.
+MongoDB 4.0 introduced a ``nameOnly`` boolean option to the ``listCollections``
```
jmikola: I propose moving this to the last paragraph of this section.

source/enumerate-collections.rst
```diff
@@ -265,7 +266,17 @@ Example return::
...
 
-Drivers MUST strip the database name from the returned collection names.
+
+
+Drivers MAY implement a MongoClient method that returns an Iterable of strings,
+each of which corresponds to a name in the collections array of the
+``listCollections`` command result. This method SHOULD be named
+``listCollections``.
+
+Older versions of the server that do not support the ``nameOnly`` option for the
```
jmikola: If we move the "MongoDB 4.0 introduces..." paragraph to prefix this text, you can then say that "versions of the server between 2.7.6 (inclusive) and 4.0 (exclusive) do not support the `nameOnly` option...". This will purposefully omit earlier server versions, which rely on queries to `system.namespaces`.

source/enumerate-collections.rst
```diff
@@ -265,7 +266,17 @@ Example return::
...
+``listCollections`` command will ignore it without raising an error. Therefore,
+drivers SHOULD always specify the ``nameOnly`` option when they only intend to
+access collection names from the ``listCollections`` command result.
```
jmikola: This sentence can remain as-is, since it only applies to `listCollections` command. :+1:

source/enumerate-collections.rst
```diff
@@ -250,6 +250,9 @@ All methods:
 
 Getting Collection Names
 ~~~~~~~~~~~~~~~~~~~~~~~~
+Drivers MAY implement a MongoClient method that returns an iterable of strings,
```
jmikola: Other parts of the document have a blank line after the heading underline. Should we have that here as well (if only for consistency)?

source/enumerate-collections.rst
```diff
@@ -333,7 +331,7 @@ Example return (in PHP, but abbreviated)::
       [5] => class MongoCollection#11 { }
     }
 
-Drivers SHOULD specify the nameOnly option when executing the ``listCollections`` command for this method.
+Drivers MUST specify the ``nameOnly`` option when executing the ``listCollections`` command for this method.
```
jmikola: This seems to conflict with "SHOULD always specify the `nameOnly` option" from the "Getting Collection Names" section. Should we also use `MUST` there to ensure these are consistent?

That would leave "Getting Full Collection Information" as the lone differentiator, for which you already say that drivers MAY expose the option.

PR #326, opened 2018-05-18
SPEC-1068: Adding change-stream YAML tests

daprahamian:
Adds automated tests for as many change-stream tests as we can. We cannot automate all tests as we need some failpoints created. Those tests are still scaffolded out, with the `skip` field added to denote that they are not complete.


source/change-streams/tests/README.rst
```diff
@@ -0,0 +1,100 @@
...
+  Each test will have some of the following fields:
+
+  - ``description``: The name of the test.
+  - ``skip``: Notes that this test is incomplete and should not be run
+  - ``minServerVersion``: The minimum server version to run this test against. If not present, assume there is no minimum server version.
+  - ``maxServerVersion``: Reserved for later use
+  - ``failPoint``: Reserved for later use
+  - ``topology``: The server topology/topologies to run the test against.
+    Valid topologies are ``single``, ``replicaset``, and ``sharded``.
+    Can be either a single topology, or a list of topologies.
```
jyemin: Easier to consume if this is always a list.

source/change-streams/tests/README.rst
```diff
@@ -0,0 +1,100 @@
...
+    - ``commandName``: Name of the command to run
+    - ``arguments``: Array of arguments for the command.
+
+  - ``expectations``: Optional list of command-started events
+  - ``result``: Document with ONE of the following fields:
+
+    - ``error``: Describes an error received during the test
+    - ``success``: An array of documents expected to be received from the changeStream
+
+For every test, if a value does not exist, a value from ``defaults`` should be used in its place.
```
jyemin: This makes it less straightforward to implement a test runner.  Since we have to write ~10 of those, and only write the spec tests once, just repeat in each test, using YAML references where possible to avoid duplication.

Also, none of our other test runners require this so it's out of character to do it here.

source/change-streams/tests/README.rst
```diff
@@ -0,0 +1,100 @@
...
+  - Assert that the error MATCHES ``results.error``
+
+- Else
+
+  - Assert that no error was expected for the test
+  - Assert that the changes received from ``changeStream`` MATCH the results in ``results.success``
+
+- If there are any ``expectations``
+
+  - assert that they MATCH the actual APM events
```
jyemin: Specify whether that includes the events generated by operations, or just the ones from the change stream (and if the latter, the rules for filtering them out).

source/change-streams/tests/change-streams.yml
```diff
@@ -0,0 +1,197 @@
...
+  -
+    description: The watch helper must not throw a custom exception when executed against a single server topology, but instead depend on a server error
+    topology: single
+    operations: []
+    result:
+      error:
+        errmsg: The $changeStream stage is only supported on replica sets
+        code: 40573
+  -
+    description: ChangeStream must continuously track the last seen resumeToken
```
jyemin: Does this require a failpoint?

source/change-streams/tests/change-streams.yml
```diff
@@ -0,0 +1,197 @@
...
+    skip: true
+  -
+    description: The killCursors command sent during the “Resume Process” must not be allowed to throw an exception.
+    skip: true
+  -
+    description: The server returns change stream responses in the specified ``ServerResponse`` format
+    result:
+      success:
+        -
+          _id: {}
```
jyemin: How is this supposed to be compared?

source/change-streams/tests/change-streams.yml
```diff
@@ -0,0 +1,197 @@
...
+          operationType: insert
+          ns:
+            db: *database_name
+            coll: *collection_name
+          documentKey: {}
```
jyemin: Is this actually the empty document?  How is it supposed to be compared?

source/change-streams/tests/change-streams.yml
```diff
@@ -0,0 +1,197 @@
...
+          command:
+            aggregate: *collection_name
+            pipeline:
+              - {$changeStream: {}}
+          command_name: aggregate
+          database_name: *database_name
+    result:
+      -
+        success:
+          operationType: insert
```
jyemin: This is different than "The server returns change stream responses in the specified ``ServerResponse`` format" test.  That one has a result that's a document with a success field that's an array.  This one has a result field that's an array with a document containing a success field whose value is a document.  This one doesn't conform to the README while the first one does.,

source/change-streams/tests/change-streams.yml
```diff
@@ -0,0 +1,197 @@
...
+            db: *database_name
+            coll: *collection_name
+          documentKey: {}
+          fullDocument:
+            z: 3
+  -
+    description: Executing a `watch` helper on a Collection results in notifications for changes to the specified collection
+    expectations:
+      - 
+        command_started_event:
```
jyemin: I'm surprised we don't have a getMore event as well, since according to the README we start the change stream (which sends this aggregate), before inserting the documents.  In order to be notified of the documents we'd have to issue a getMore

source/change-streams/tests/change-streams.yml
```diff
@@ -0,0 +1,197 @@
...
+    minServerVersion: "3.7.3"
+    expectations:
+      - 
+        command_started_event:
+          command:
+            aggregate: *collection_name
+            pipeline:
+              - 
+                $changeStream:
+                  startAtClusterTime: {}
```
jyemin: Have to document how to compare the cluster time (basically to ignore it)

source/change-streams/tests/README.rst
```diff
@@ -0,0 +1,100 @@
...
+  - Drop the database ``database_name``
+  - Create the database ``database_name`` and the collection ``database_name.collection_name``
+
+- Create a new MongoClient ``client``, with APM enabled
+- Create a changeStream ``changeStream`` against the specified ``target``
+- Run every operation in ``operations`` in serial against the server
+- Wait until either
+
+  - An error occurres
+  - All operations have been successful AND the changeStream has received ``numChanges`` changes
```
jyemin: Is numChanges redundant with the size of the success array?

source/change-streams/tests/change-streams.yml
```diff
@@ -0,0 +1,197 @@
...
+    description: ChangeStream will automatically resume one time on a resumable error (including not master) with the initial pipeline and options, except for the addition/update of a resumeToken.
+    skip: true
+  -
+    description: ChangeStream will not attempt to resume on a server error
+    skip: true
+  -
+    description: ChangeStream will perform server selection before attempting to resume, using initial readPreference
+    skip: true
+  -
+    description: Ensure that a cursor returned from an aggregate command with a cursor id, and an initial empty batch, is not closed on the driver side.
```
jyemin: Does this require a failpoint?

source/change-streams/change-streams.rst
```diff
@@ -536,45 +536,23 @@ so resuming is impossible.
 Test Plan
 =========
```
jyemin: Should we move these into the README?  I think that's what our other specs do for prose tests.

source/change-streams/tests/README.rst
daprahamian: I think so in simple cases. My thinking was that, in the future, we might need to test cases where we get a success, then an error, so you couldn't simply calculate it based on the successes length.

source/change-streams/tests/change-streams.yml
daprahamian: This one I wasn't sure of. I mostly copy-pasted the current prose tests in here. The issue with this one is that we'd have to introduce an entirely new type of assertion to cover this use case, and I was not sure of the best way to go about that. How does each driver track the `resumeToken`, and what is the best way to express an assertion based on `resumeToken`?

source/change-streams/tests/README.rst
jyemin: The error wouldn't be in the success array in that case.  I advise to keep it simple until you need to make it more complex.

source/change-streams/tests/change-streams.yml
daprahamian: No, this was another one that I was not sure how to express as a test.

source/change-streams/tests/change-streams.yml
jyemin: Thinking about it a bit more I think it will require some sort of failpoint.  Because the point is that on a failure the driver resume with the proper resumeToken, and we need a failpoint to force a failure that requires resuming

source/change-streams/tests/change-streams.yml
jyemin: Seems more confusing to have these placeholders in here that we don't know how to implement. Consider taking them out and opening a SPEC ticket instead to track the desire.

source/change-streams/tests/change-streams.yml
daprahamian: I was typing up a huge response on the rationale for this, but in the process I decided that it might make more sense to have all the results be expressed as extended JSON, as it will be more language-agnostic. Thoughts?

source/change-streams/tests/change-streams.yml
daprahamian: 1. From what I can tell, a `getMore` is not necessarily guaranteed, since results can be returned in the initial `find`
2. The test is more about asserting that the change stream is against a specific collection, which is why we are checking the initial aggregate.

jyemin: 1. It's guaranteed in this case because the document insertion happens after the change stream has been established with the initial `aggregate` command
2. OK, but are drivers expected to ignore any getMore commands when comparing?  If so, say that in the README.  But better if we can actually assert on the getMore commands, if the behavior is predicable.

source/change-streams/tests/change-streams.yml
daprahamian: Turns out that extended json doesn't actually help here. Basically, the idea is for `{}` to be an `any` object that matches any non-primitive. I don't have a way of knowing the `_id`, but I still want to assert that it is there. Since `MATCH` is basically subset matching, it's meant to ensure that there is some sort of object at `_id`.

I was considering an alternative to be some sort of magic string, like `##ANY##` that would work in a similar role.

source/change-streams/tests/change-streams.yml
daprahamian: Same as for `_id`

source/change-streams/tests/change-streams.yml
daprahamian: I wrote in the README:

> - If there are any ``expectations``
>   - assert that they MATCH the actual APM events

This is meant to equate to the following pseudocode:

```
for (expected, idx) in expectations:
  Assert that actual[idx] MATCHES expected
```

If that is not clear, I can make it more elaborate.

I'd also rather not have a test for the getMore, since it is not actually relevant to the assertion

source/change-streams/change-streams.rst
daprahamian: done

source/change-streams/tests/README.rst
daprahamian: done

source/change-streams/tests/README.rst
daprahamian: done

source/change-streams/tests/change-streams.yml
daprahamian: done: removed all skippable tests

source/change-streams/tests/README.rst
daprahamian: I'm actually going to modify this to include using the "global" mongo client for all non-change-stream operations.

daprahamian: I've updated the instructions to separate out the operations from the changeStream, such that only the changeStream events are recorded

source/change-streams/tests/change-streams.yml
daprahamian: Done

source/change-streams/tests/README.rst
daprahamian: done

source/change-streams/tests/change-streams.yml
jyemin: I think this is ok so longs as it's documented in the README

source/change-streams/tests/change-streams.yml
jyemin: Acknowledged

source/change-streams/tests/change-streams.json
```diff
@@ -0,0 +1,402 @@
...
+                  "$changeStream": {}
+                }
+              ]
+            },
+            "command_name": "aggregate",
+            "database_name": "change-stream-tests"
+          }
+        }
+      ],
+      "result": [
```
jyemin: This is backwards.  `result` should be a document and `success` should be an array.

source/change-streams/tests/README.rst
```diff
@@ -0,0 +1,151 @@
...
+
+Each YAML files has the following keys:
+
+- ``database_name``: The default database
+- ``collection_name``: The default collection
+- ``tests``: An array of tests that are to be run independently of each other.
+  Each test will have some of the following fields:
+
+  - ``description``: The name of the test.
+  - ``minServerVersion``: The minimum server version to run this test against. If not present, assume there is no minimum server version.
```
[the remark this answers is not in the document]
daprahamian: 👍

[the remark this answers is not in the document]
daprahamian: Would it be okay to use camelCase instead of snake_case? Or should I try to make it conform as much as possible?

source/change-streams/tests/README.rst
```diff
@@ -0,0 +1,151 @@
...
+  - ``failPoint``: Reserved for later use
+  - ``target``: The entity to run a changeStream on. Valid values are:
+  
+    - ``collection``: Watch changes on collection ``database_name.collection_name``
+    - ``db``: Watch changes on database ``database_name``
+    - ``client``: Watch changes on entire clusters
+  - ``numChanges``: The number of changes to expect on a changeStream
+  - ``topology``: An array of server topologies to run the test against
+    Valid topologies are ``single``, ``replicaset``, and ``sharded``.
+  - ``operations``: Array of documents, each describing an operation. Each document has the following fields:
```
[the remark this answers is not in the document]
daprahamian: As long as we are okay with expanding that specification to have `database` and `collection` fields.

source/change-streams/tests/README.rst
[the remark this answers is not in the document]
jyemin: I'd rather it be consistent so that we can share as much testing code as possible.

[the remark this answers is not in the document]
daprahamian: [Actually, retryableWrites appears to use `minServerVersion`](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/insertOne.yml#L5). Which one should we be using?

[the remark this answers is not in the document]
jyemin: Looks like it's Java driver's out of date test runner.  Copied from the command monitoring tests, which do use the ones I suggested.  Go with consistency with retryable writes

source/change-streams/tests/README.rst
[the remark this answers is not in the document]
jyemin: Yeah, I'm ok with that.

source/change-streams/tests/change-streams.yml
```diff
@@ -0,0 +1,255 @@
...
+        commandName: insert
+        arguments:
+          - { z: 3 }
+    expectations:
+      - 
+        command_started_event:
+          command:
+            aggregate: *collection_name
+            pipeline:
+              - {$changeStream: {}}
```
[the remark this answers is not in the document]
daprahamian: Does the spec actually require that we send `fullDocument: "default"`? The JS impl only sends it if it is specified, and otherwise leaves it empty.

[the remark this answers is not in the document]
jyemin: That's how I read it, though either would be correct.  Regardless, to write spec tests we have to choose one or the other interpretation.

source/change-streams/tests/change-streams.json
daprahamian: done

source/change-streams/tests/change-streams.yml
```diff
@@ -0,0 +1,255 @@
...
+          command:
+            aggregate:
+              $numberInt: "1"
+            pipeline:
+              - 
+                $changeStream:
+                  allChangesForCluster: true
+                  startAtClusterTime:
+                    ts:
+                      $timestamp: {}
```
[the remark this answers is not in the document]
daprahamian: Implemented with magic `42`

source/change-streams/tests/README.rst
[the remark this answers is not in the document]
daprahamian: done

source/change-streams/tests/README.rst
```diff
@@ -0,0 +1,156 @@
...
+
+    - Assert that ``expected`` equals ``actual``
+
+  - Else if ``expected`` is an ``array``
+
+    - for every ``idx``/``value`` in ``expected``
+
+      - Assert that ``value`` MATCHES ``actual[idx]``
+
+  - Else
```
jmikola: Considering the earlier conditions, this assumes that all possible types are limited to:

 * number
 * string
 * boolean
 * null
 * array
 * object (this else case)

Would it be safer to explicitly check for an object here and either omit a final `else` or use it to report some error for an unsupported type?

source/change-streams/tests/README.rst
```diff
@@ -0,0 +1,156 @@
...
+
+    - for every ``key``/``value`` in ``expected``
+
+      - Assert that ``value`` MATCHES ``actual``
```
jmikola: This is a bit confusing. Are you asserting that all properties within the `expeceted` object have the value of `actual`? Is there a reason this isn't consistent with the array type case, where you respect the index/key in the assertion?

source/change-streams/tests/README.rst
```diff
@@ -0,0 +1,156 @@
...
+
+For each YAML file, for each element in ``tests``:
+
+- If ``topology`` does not include the topology of the server instance(s), skip this test.
+- Use ``globalClient`` to
+
+  - Drop the database ``database_name``
+  - Create the database ``database_name`` and the collection ``database_name.collection_name``
+
+- Create a new MongoClient ``client``, with APM enabled
```
jmikola: Is APM actually something that is enabled when a client is constructed? IIRC, the [APM spec](https://github.com/mongodb/specifications/blob/master/source/command-monitoring/command-monitoring.rst) doesn't talk about registering APM individual clients but it does have an example for registering a global event subscriber. I suppose this may vary by driver.

Would it make sense to simply have "Register APM listeners for `client`" as its own bullet?

source/change-streams/tests/README.rst
```diff
@@ -0,0 +1,156 @@
...
+- Using ``client``, create a changeStream ``changeStream`` against the specified ``target``. Use ``changeStreamPipeline`` and ``changeStreamOptions`` if they are non-empty
+- Using ``globalClient``, run every operation in ``operations`` in serial against the server
+- Wait until either
```
jmikola: A colon after "either" would work nicely.

source/change-streams/tests/README.rst
```diff
@@ -0,0 +1,156 @@
...
+
+  - An error occurres
```
jmikola: Typo: "occurs"

source/change-streams/tests/README.rst
```diff
@@ -0,0 +1,156 @@
...
+  - All operations have been successful AND the changeStream has received as many changes as there are ``result.success``
```
jmikola: "there are in"

source/change-streams/tests/README.rst
```diff
@@ -0,0 +1,156 @@
...
+- Else
+
+  - Assert that no error was expected for the test
+  - Assert that the changes received from ``changeStream`` MATCH the results in ``results.success``
+
+- If there are any ``expectations``
+
+  - For each APM ``commandStarted`` events on ``client``
+
+    - Filter out any ``ismaster`` events
```
jmikola: `isMaster`

source/change-streams/tests/README.rst
```diff
@@ -0,0 +1,156 @@
...
+  - For each (``expected``, ``idx``) in ``expectations``
+
+    - Assert that ``actual[idx]`` MATCHES ``expected``
+
+- Close the MongoClient ``client``
+
+After running all tests
+
+- Close the MongoClient ``globalClient``
+- Shut down the server topology
```
jmikola: This reads as if drivers should terminate the entire cluster (e.g. via Mongo Orchestration). Is that really necessary? Wouldn't it be sufficient to drop the relevant databases?

source/change-streams/tests/README.rst
```diff
@@ -0,0 +1,156 @@
...
+===========
+
+The following tests have not yet been automated, but MUST still be tested
+
+1. ``ChangeStream`` must continuously track the last seen ``resumeToken``
+2. ``ChangeStream`` will throw an exception if the server response is missing the resume token
+3. ``ChangeStream`` will automatically resume one time on a resumable error (including `not master`) with the initial pipeline and options, except for the addition/update of a ``resumeToken``.
+4. ``ChangeStream`` will not attempt to resume on a server error
+5. ``ChangeStream`` will perform server selection before attempting to resume, using initial ``readPreference``
+6. Ensure that a cursor returned from an aggregate command with a cursor id, and an initial empty batch, is not closed on the driver side.
```
jmikola: I don't see any need to wrap "and an initial empty batch" in a comma-delimited clause.

source/change-streams/tests/README.rst
```diff
@@ -0,0 +1,156 @@
...
+7. The ``killCursors`` command sent during the “Resume Process” must not be allowed to throw an exception.
```
jmikola: Curly double quotes indicate this might have been copied out of Google Docs. Should these be `"` for consistency?

source/change-streams/tests/README.rst
```diff
@@ -0,0 +1,156 @@
...
+8. ``$changeStream`` stage for ``ChangeStream`` against a server ``>=4.0`` that has not received any results yet MUST include a ``startAtClusterTime`` when resuming a changestream.
```
jmikola: `startAtClusterTime` looks outdated. Should this be `startAtOperationTime`?

Additional grammar fix: "a `startAtOperationTime` option when..."

source/change-streams/tests/change-streams.yml
```diff
@@ -0,0 +1,391 @@
...
+database_name: &database_name "change-stream-tests"
+collection_name_2: &collection_name_2 "test2"
+database_name_2: &database_name_2 "change-stream-tests-2"
+tests:
+  -
+    description: The watch helper must not throw a custom exception when executed against a single server topology, but instead depend on a server error
+    minServerVersion: "3.6.0"
+    target: collection
+    topology:
+      - single
```
jmikola: I assume this would also fail if executed against a `mongos` cluster where individual shards are not replica sets. FWIW, I believe this is currently supported but in the future sharded clusters may require replica sets, so we may not care about it.

source/change-streams/tests/change-streams.yml
```diff
@@ -0,0 +1,391 @@
...
+    changeStreamPipeline: []
+    changeStreamOptions: {}
+    operations: []
+    expectations: []
+    result:
+      error:
+        errmsg: The $changeStream stage is only supported on replica sets
```
jmikola: Does it make sense to assert server-side error messages, which are prone to change? A more robust approach may be to rely on the error code alone, which should remain constant.

source/change-streams/tests/change-streams.yml
```diff
@@ -0,0 +1,391 @@
...
+          command:
+            aggregate: *collection_name
+            cursor: {}
+            pipeline:
+              - 
+                $changeStream:
+                  fullDocument: default
+          command_name: aggregate
+          database_name: *database_name
+    result: {}
```
jmikola: Shouldn't there be an error for this test? According to the test README, the flowchart indicated that we should either expect `result.success` or an error to compare with `result.error`. I don't believe it addressed a case where `result` was empty.

> Document with ONE of the following fields: `success` / `error`

source/change-streams/change-streams.rst
```diff
@@ -536,45 +536,7 @@ so resuming is impossible.
...
-1. The server returns change stream responses in the specified ``ServerResponse`` format
-
-2. Executing a ``watch`` helper on a Collection results in notifications for changes to the specified collection
-
-3. Executing a ``watch`` helper on a Database results in notifications for changes to all collections in the specified database.
-
-4. Executing a ``watch`` helper on a MongoClient results in notifications for changes to all collections in all databases in the cluster, excluding the ``admin``, ``local``, and ``config`` databases.
-
-5. ``ChangeStream`` will resume after a ``killCursors`` command is issued for its child cursor.
+See `here <tests/README.rst>`_
```
jmikola: It would be preferable to use `tests/README.rst` as the link name, as it's more contextual. We're also not pressed for space here.

source/change-streams/tests/change-streams.yml
```diff
@@ -0,0 +1,391 @@
...
+        -
+          _id: "42"
+          documentKey: "42"
+          operationType: insert
+          ns:
+            db: *database_name
+            coll: *collection_name
+          fullDocument:
+            x:
+              $numberInt: "1"
```
jmikola: Is Extended JSON notation necessary here? I'm not familiar with other spec tests doing this.

source/change-streams/tests/change-streams.yml
```diff
@@ -0,0 +1,391 @@
...
+            cursor: {}
+            pipeline:
+              - 
+                $changeStream:
+                  fullDocument: default
+          command_name: aggregate
+          database_name: *database_name
+    result: {}
+  -
+    description: The server returns change stream responses in the specified ``ServerResponse`` format
```
jmikola: Can we remove the backtick-quoting and/or write this as "server response"? I don't see any reason to use RST syntax in a YAML test description.

source/change-streams/tests/change-streams.yml
```diff
@@ -0,0 +1,391 @@
...
+      -
+        database: *database_name
+        collection: *collection_name
+        name: insertOne
+        arguments:
+          document:
+            z: 3
+    expectations:
+      - 
+        command_started_event:
```
jmikola: Should APM not expect to see the inserts from the global client? For drivers that allow registering listeners globally, perhaps we need some language to mention that the APM listeners should only observe activity on the client under test (i.e. client executing the change stream).

source/change-streams/tests/change-streams.yml
```diff
@@ -0,0 +1,391 @@
...
+              $numberInt: "3"
+  -
+    description: Change Stream should error when an invalid aggregation stage is passed in
+    minServerVersion: "3.6.0"
+    target: collection
+    topology:
+      - replicaset
+    changeStreamPipeline:
+      -
+        $darmokAndJaladAtTanagra: SHAKA WHEN THEN WALLS FELL
```
jmikola: Typo: `SHAKA WHEN THE WALLS FELL`

I'm all for fun, but I don't think it belongs in specs. I'd propose `$unsupported: foo`.

source/change-streams/tests/change-streams.yml
```diff
@@ -0,0 +1,391 @@
...
+        database: *database_name
+        collection: *collection_name
+        name: insertOne
+        arguments:
+          document:
+            z: 3
+    expectations:
+      -
+        command_started_event:
+
```
jmikola: Are the line breaks after `command_started_event:` necessary? I see this done in a few places, but not everywhere. If it's not needed I'd suggest removing it in all instances for consistency.

source/change-streams/tests/change-streams.yml
```diff
@@ -0,0 +1,391 @@
...
+          command_name: aggregate
+          database_name: *database_name
+    result:
+      error:
+        errmsg: "Unrecognized pipeline stage name: '$darmokAndJaladAtTanagra'"
+        code: 40324
+
+  -
+    description:  A fresh ChangeStream against a server >=4.0 will always include `startAtOperationTime` in the `$changeStream` stage.
+    minServerVersion: "4.0.0-rc0"
```
jmikola: AFAIK, `minServerVersion` comparisons only use X.Y.Z and don't support labels after a hyphen. I'd suggest that this either use the corresponding 3.8.x patch version (if one applies) or simply say "4.0.0".

source/change-streams/tests/change-streams.yml
```diff
@@ -0,0 +1,391 @@
...
+              - 
+                $changeStream:
+                  fullDocument: default
+                  startAtOperationTime:
+                    $timestamp:
+                      i: 42
+                      t: 42
+          command_name: aggregate
+          database_name: *database_name
+    result: {}
```
jmikola: Another instance of not asserting a result. If this is a valid test case where an APM check is needed then the README should be updated to discuss this.

source/change-streams/tests/change-streams.yml
```diff
@@ -0,0 +1,391 @@
+collection_name: &collection_name "test"
+database_name: &database_name "change-stream-tests"
+collection_name_2: &collection_name_2 "test2"
+database_name_2: &database_name_2 "change-stream-tests-2"
```
jmikola: I'd propose `collection2_name` and `database2_name` instead. When reading these identifiers, it makes it more immediately clear that this is the second subject.

source/change-streams/change-streams.rst
```diff
@@ -388,26 +388,26 @@ Single Server Topologies
...
 
-``startAtClusterTime`` and ``resumeAfter`` are mutually exclusive; if both ``startAtClusterTime`` and ``resumeAfter`` are set, the server will return an error. Drivers MUST NOT throw a custom error, and MUST defer to the server error.
+``startAtOperationTime`` and ``resumeAfter`` are mutually exclusive; if both ``startAtOperationTime`` and ``resumeAfter`` are set, the server will return an error. Drivers MUST NOT throw a custom error, and MUST defer to the server error.
 
-If neither ``startAtClusterTime`` or ``resumeAfter`` are specified, and the server version is >= ``4.0`` drivers MUST set a default ``startAtClusterTime`` value using a ``$clusterTime`` from any server response (ex: ``ismaster``). This allows change streams to be resumed before the first notification is received.
+If neither ``startAtOperationTime`` or ``resumeAfter`` are specified, and the server version is >= ``4.0`` drivers MUST set a default ``startAtOperationTime`` value using an ``operationTime`` from any server response (ex: ``ismaster``). This allows change streams to be resumed before the first notification is received.
```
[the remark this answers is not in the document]
daprahamian: AFAIK all server responses return an `operationTime`. The server spec for `startAtOperationTime` actually states that the operation time must be obtained as early as possible, rather than on the fly.

I will reword it to explain this.

source/change-streams/tests/README.rst
```diff
@@ -0,0 +1,156 @@
...
+    - ``success``: An array of documents expected to be received from the changeStream
+
+Spec Test Match Function
+========================
+
+The definition of MATCH or MATCHES in the Spec Test Runner is as follows:
+
+- MATCH takes two values, ``expected`` and ``actual``
+- Notation is "Assert [actual] MATCHES [expected]
+- Assertion passes if ``expected`` is a subset of ``actual``, with the values ``42`` and ``"42"`` acting as placeholders for "any value"
```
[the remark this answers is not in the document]
daprahamian: `42` and `"42"` are the actual magic values

source/change-streams/tests/README.rst
daprahamian: Those are all of the valid JSON types. We could also rework it to be:

```
If expected is a JSON object:
  ...
Else if expected is a JSON array:
  ...
Else:
  Assert that expected equals actual
```

source/change-streams/tests/README.rst
daprahamian: I wanted to explicitly make clear that we should only be recording APM events from `client`, and not from the `globalClient`

source/change-streams/tests/README.rst
daprahamian: This is wrong, it should be the other way around.

source/change-streams/tests/README.rst
daprahamian: I only state this b/c I said `Create a MongoDB server topology.` before running the tests. I could remove both references and instead say to drop the databases after the tests.

source/change-streams/tests/change-streams.yml
daprahamian: It seems like extended JSON would be the one uniform standard we can expect across all drivers.

jyemin: True, but in this case `x : 1` would probably suffice.

source/change-streams/tests/change-streams.yml
jyemin: Good point.  I didn't realize some drivers had global listeners.  I'm not clear how a driver with a global listener could filter the inserts.  Perhaps just say that if a driver has a global listener, insert events should be ignored.

source/change-streams/tests/change-streams.yml
daprahamian: It still seems better to err on the side of strict notation, rather than using the relaxed format.

source/change-streams/tests/change-streams.yml
daprahamian: Do any drivers _exclusively_ use global listeners? The language of the test runner says: "For each APM ``commandStarted`` events on ``client``", meaning only APM events on `client`. If that is not possible, then yeah, we should add a filter for specific APM events.

source/change-streams/change-streams.rst
```diff
@@ -388,26 +388,26 @@ Single Server Topologies
...
 
-``startAtClusterTime`` and ``resumeAfter`` are mutually exclusive; if both ``startAtClusterTime`` and ``resumeAfter`` are set, the server will return an error. Drivers MUST NOT throw a custom error, and MUST defer to the server error.
+``startAtOperationTime`` and ``resumeAfter`` are mutually exclusive; if both ``startAtOperationTime`` and ``resumeAfter`` are set, the server will return an error. Drivers MUST NOT throw a custom error, and MUST defer to the server error.
 
-If neither ``startAtClusterTime`` or ``resumeAfter`` are specified, and the server version is >= ``4.0`` drivers MUST set a default ``startAtClusterTime`` value using a ``$clusterTime`` from any server response (ex: ``ismaster``). This allows change streams to be resumed before the first notification is received.
+If neither ``startAtOperationTime`` nor ``resumeAfter`` are specified, and the server version is >= ``4.0`` drivers MUST set a default ``startAtOperationTime`` (a default timestamp can be obtained from the ``operationTime`` field on any server response, like an ``isMaster``. This allows change streams to be resumed before the first notification is received.
```
jmikola: The parenthetical is missing a closing parenthesis. I would suggest writing it out as a normal sentence, though (perhaps with "(e.g. `isMaster`)" at the end).

This still seems a bit ambiguous. In https://github.com/mongodb/specifications/pull/326#discussion_r190962952 you wrote:

> The server spec for startAtOperationTime actually states that the operation time must be obtained as early as possible, rather than on the fly.

Wouldn't we want to use the most recently observed operation time? If we merely use the earliest collected operation time, there's no telling how old that may be if the topology has been persisted. Tracking of operation time may fall outside the scope of this spec, so this may be a question for @ajdavis and/or @rstam.

source/change-streams/tests/README.rst
jmikola: Noted. I see you reworked it to make the non-object/array case the `else` condition with a basic equality check. I think that works nicely.

source/change-streams/tests/change-streams.yml
jmikola: My concern was that some drivers may convert the JSON object to a native type instead of BSON (e.g. result document's may be compared as PHP objects or Python dictionaries). In that case, `$numberInt` isn't going to be resolved to a 32-bit BSON integer type.

If we're going to use Extended JSON here, I think the test README needs to note that drivers must route these values through "Extended JSON -> BSON" conversion first. From there, they can decide whether to compare the command results in BSON form or their native types. For drivers that don't represent raw BSON, command results are typically accessed as native types (e.g. PHP objects, Python dictionaries), be it from the command helper or the reply accessor on APM event objects.

source/change-streams/tests/change-streams.yml
jmikola: PHP uses global listeners. This is a design necessity because our client objects and event listeners are both request-scoped, while the libmongoc clients are process-scoped and persisted across requests. Furthermore, since multiple PHP client objects could share the same libmongoc client within the same request (an edge case, but still possible), there isn't a good way to associate event listeners with individual PHP clients.

Simply stating that the listener should ignore events on the global client should be sufficient, though.

source/change-streams/tests/README.rst
```diff
@@ -101,23 +94,25 @@ For each YAML file, for each element in ``tests``:
...
-- Create a new MongoClient ``client``, with APM enabled
+- Create a new MongoClient ``client``
 - Using ``client``, create a changeStream ``changeStream`` against the specified ``target``. Use ``changeStreamPipeline`` and ``changeStreamOptions`` if they are non-empty
 - Using ``globalClient``, run every operation in ``operations`` in serial against the server
-- Wait until either
+- Wait until either:
 
-  - An error occurres
-  - All operations have been successful AND the changeStream has received as many changes as there are ``result.success``
+  - An error occurrs
```
jmikola: Typo: "occurs"

source/change-streams/tests/change-streams.yml
```diff
@@ -0,0 +1,339 @@
...
+    result:
+      success: []
+  -
+    description: The server returns change stream responses in the specified server response format
+    minServerVersion: "3.6.0"
+    target: collection
+    topology:
+      - replicaset
+    changeStreamPipeline: []
+    changeStreamOptions: {}
```
jmikola: The previous test ("$changeStream must be the first stage in a change stream pipeline sent to the server") doesn't specify these options. Do they need to be included here if these are essentially default values?

source/change-streams/tests/change-streams.yml
```diff
@@ -0,0 +1,339 @@
...
+            aggregate: *collection_name
+            cursor: {}
+            pipeline:
+              - 
+                $changeStream:
+                  fullDocument: default
+          command_name: aggregate
+          database_name: *database_name
+    result:
+      success: []
```
jmikola: Shouldn't the `insertOne` operation be observed here? Or is omitted because the test only concerns itself with the `aggregate` command structure in the APM event. I suppose that would explain why the subsequent "The server returns change stream responses in the specified server response format" test ignores the APM format.

source/change-streams/change-streams.rst
daprahamian: I actually misspoke. The operation time should be the most recent operation time when the changeStream is created. That operationTime should be tracked by the changeStream until it receives its first resume token.

source/change-streams/tests/README.rst
```diff
@@ -0,0 +1,152 @@
...
+
+    - Assert that ``actual[idx]`` MATCHES ``expected``
+
+- Close the MongoClient ``client``
+
+After running all tests
+
+- Close the MongoClient ``globalClient``
+- Drop database ``database_name``
+- Drop database ``database2_name``
```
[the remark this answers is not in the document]
daprahamian: This is supposed to be run after all tests have completed, so it should not impact their change streams.

As of right now, the last write operation that we run before creating the local client is creating the collection `database2_name.collection2_name`, which is not showing up in the changeStream.

source/change-streams/tests/change-streams.yml
daprahamian: Yes, since the test is only meant to test the `aggregate` command structure, that is the only event listed. IMO it is best to test only as much as necessary to assert the condition.

source/change-streams/tests/change-streams.yml
daprahamian: That is a mistake. I will add them in.

source/change-streams/tests/change-streams.yml
daprahamian: My current JS impl actually does the following:

1. Read expected value in as Native JSON
2. Convert actual value from BSON --> Extended JSON string --> Native JSON
3. Compare the two values.

Would it make sense to clarify in the object format that results are written in extended json? Then, a driver could decide how it wants to process the values?

source/change-streams/change-streams.rst
jmikola: If I understand correctly, the initial `aggregate` command should omit `startAtOperationTime` if not provided by the user. It should then capture `startAtOperationTime` from the `aggregate` response and use that for any resume behavior (if the change stream must resume before an actual change comes back with a proper resume token). If so, let's just make sure this is clear in the spec (if it isn't already).

daprahamian: No, the only reason to omit a `startAtOperationTime` against 4.0 is if there is a resume token. Any changeStream without a resume token should include `startAtOperationTime`, either one provided by the user, or one based on a previous command like an `isMaster`.

source/change-streams/tests/change-streams.yml
jmikola: I just saw the change to the test README re: extended JSON and allowing drivers to choose one of three styles of comparison. Please disregard my previous comment, which was written prematurely. I still think extended JSON is unnecessary here, but I won't fight for it.

source/change-streams/tests/README.rst
```diff
@@ -77,10 +77,16 @@ Pseudocode implementation of ``actual`` MATCHES ``expected``:
         Assert that actual[idx] MATCHES value
     Else if expected is a JSON object:
       For every key/value in expected
-        Assert that actual MATCHES value
+        Assert that actual[key] MATCHES value
     Else:
       Assert that expected equals actual
 
+The expected values for ``result.success`` and ``expectations`` are written in Extended JSON. Drivers may adopt any of the following approaches to comparisons, as long as they are consistent:
```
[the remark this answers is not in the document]
jyemin: I have no issues with pedantry.  It's a spec!

We don't need to note it because an extended JSON parser must be able to parse either without being told which it is:

> By default, a parser MUST accept values in either Canonical Extended JSON format or Relaxed Extended JSON format as described in this specification. A parser MAY allow users to restrict parsing to only Canonical Extended JSON format or only Relaxed Extended JSON format.

jmikola:
LGTM. Please note https://github.com/mongodb/specifications/pull/326#discussion_r191530050 and my earlier suggestion about consolidating the JSON array/object comparison logic.

PR #328, opened 2018-05-22
SPEC-1092 Test reset session state to "no transaction"

ShaneHarvey:
Should I expand this and test every operation that can be performed (delete, update, find, agg, etc..) or do you think a single case is enough? My assumption is that there is a single place the driver should reset the sate and that a single test is enough. I want to avoid creating a combinatorial explosion of tests.

### DIFF
--- a/source/transactions/tests/commit.yml
+++ b/source/transactions/tests/commit.yml
@@ -456,7 +456,7 @@ tests:
       collection:
         data: []
 
-  - description: reset session state
+  - description: reset session state commit
 
     operations:
       - name: startTransaction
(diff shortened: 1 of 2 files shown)


source/transactions/tests/commit.yml
```diff
@@ -455,3 +455,88 @@ tests:
...
+        arguments:
+          session: session0
+        result:
+          errorContains: no transaction started
+      # Calling abort should also error with "no transaction started".
+      - name: abortTransaction
+        arguments:
+          session: session0
+        result:
+          errorContains: no transaction started
```
ajdavis: Let's split this test in two. One test does start, insert, commit, insert, commit, and the other does start, insert, commit, insert, **abort**.

I'm concerned in this test that calling commit, then abort, muddies the water. We know that insert should reset the state, and then calling "commit" and getting an error proves that the state is reset. But then we call abort, and the state of the transaction at that point could be the result of the insert call **or** the commit call. So let's split up the test to be more precise.

ShaneHarvey: Update with two tests almost as you've described. The first does commit, insert, commit and the second does abort, insert, abort.

PR #329, opened 2018-05-23
SPEC-1093 omit session ID with parallelCollectionScan

xdg:


ShaneHarvey:
I think this needs more clarification. How do drivers deal with this change? PyMongo has a `parallel_scan` helper that accepts a `session` argument. We cannot just silently ignore the session as that is not compatible with transactions or causal consistency. I suggest that drivers MUST raise an error if a parallel_scan helper method is called with a session. 

Actually it would be even better if the server told us that it does not support `parallelCollectionScan` with and `lsid` by returning an error. Then drivers could rely on the server for an error when an explicit session is passed in.

xdg:
I think treating it like unack'd writes is fine.  If you want to amend the spec to state that it fails if given an explicit session, please go ahead.

behackett:
I _think_ we can use the session for the initial parallelCollectionScan call, and not use the session for the cursors.

behackett:
That would avoid backward breaking changes in drivers that already take a session param for the helper, and avoid breaking existing applications.

ShaneHarvey:
The server requires that the same session is used for the initial command and all the cursors.

PR #330, opened 2018-05-23
SPEC-1082: test transactional runCommand

jyemin:
Test that runCommand works as specified.  Two aspects of the assertions

* That all the transaction-related fields are added to the command
* That the transaction's read preference is obeyed, regardless of the runCommand read preference.  This assertion is implicit for the secondary read preference test, in that the command would fail if it was actually executed against a secondary (I think it would at any rate).

### DIFF
--- a/source/transactions/tests/README.rst
+++ b/source/transactions/tests/README.rst
@@ -98,7 +98,14 @@ For each YAML file, for each element in ``tests``:
    - Enter a "try" block or your programming language's closest equivalent.
    - If ``name`` is "startTransaction", "commitTransaction", or
      "abortTransaction", call the named method on ``session0`` or
-     ``session1``, depending on the "session" argument.
+     ``session1``, depending on the "session" argument. 
+   - If ``name`` is "runCommand", call the runCommand method on the database
+     specified in the test. Pass the argument named "command" to the runCommand 
+     method. Pass ``session0`` or ``session1`` to the runCommand method, depending 
+     on which session's name is in the arguments list. If ``arguments`` 
+     contains no "session", pass no explicit session to the method. If ``arguments`` 
+     includes "readPreference", also pass the read preference to the runCommand 
+     method.    
    - Otherwise, ``name`` refers to a CRUD method, such as ``insertOne``.
      Execute the named method on the "transactions-tests" database on the "test"
      collection, passing the arguments listed. Pass ``session0`` or ``session1``
--- a/source/transactions/tests/run-command.json
+++ b/source/transactions/tests/run-command.json
@@ -68,6 +68,78 @@
       ]
     },
     {
+      "description": "run command with secondary read preference in client option and primary read preference in transaction options",
+      "clientOptions": {
+        "readPreference": "secondary"
+      },
+      "operations": [
+        {
+          "name": "startTransaction",
+          "arguments": {
+            "session": "session0",
+            "options": {
+              "readPreference": {
+                "mode": "Primary"
+              }
+            }
+          }
+        },
+        {
+          "name": "runCommand",
+          "arguments": {
+            "command": {
+              "count": "test"
+            },
+            "session": "session0"
+          },
+          "result": {
+            "n": 0
+          }
+        },
+        {
+          "name": "commitTransaction",
+          "arguments": {
+            "session": "session0"
+          }
+        }
+      ],
+      "expectations": [
+        {
+          "command_started_event": {
+            "command": {
+              "count": "test",
+              "readConcern": null,
+              "lsid": "session0",
+              "txnNumber": {
+                "$numberLong": "1"
+              },
+              "startTransaction": true,
+              "autocommit": false,
+              "writeConcern": null
+            },
+            "command_name": "count",
+            "database_name": "transaction-tests"
+          }
+        },
+        {
+          "command_started_event": {
+            "command": {
+              "commitTransaction": 1,
+              "lsid": "session0",
+              "txnNumber": {
+                "$numberLong": "1"
+              },
+              "startTransaction": null,
+              "autocommit": false,
+              "writeConcern": null
+            },
+            "command_name": "commitTransaction",
+            "database_name": "admin"
+          }
+        }
+      ]
+    },
+    {
       "description": "run command with explicit primary read preference",
       "operations": [
         {
@@ -135,7 +207,7 @@
       ]
     },
     {
-      "description": "run command with explicit secondary read preference",
+      "description": "run command fails with explicit secondary read preference",
       "operations": [
         {
           "name": "startTransaction",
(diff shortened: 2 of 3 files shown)


ShaneHarvey:
"This assertion is implicit for the secondary read preference test, in that the command would fail if it was actually executed against a secondary (I think it would at any rate)."

What if we run an insert command instead?

source/transactions/tests/run-command.json
```diff
@@ -0,0 +1,205 @@
...
+        },
+        {
+          "name": "runCommand",
+          "arguments": {
+            "command": {
+              "count": "test"
+            },
+            "session": "session0",
+            "readPreference": {
+              "mode": "Secondary"
```
rstam: The spec says that IF a generic runCommand is called with an EXPLICIT read preference, that the driver should use the explicit read preference overriding the transaction read preference.

This command should therefore result in an error (because for now transactions require read preference Primary).

The relevant quote from the spec is:

> For drivers that allow read concern to be passed to a particular operation, this specification says, "If the user supplies an explicit readConcern via a method option, however, drivers MUST apply the readConcern, which will result in a server error." The same principle applies to write concern and read preference. The alternative is to silently ignore the user's read concern (or other options) if it is prohibited with transactions, but this would be a surprising and undetectable deviation from the user's intent.

jyemin: @ShaneHarvey I'll leave this one for you to ponder

ajdavis: Yeah, this test should fail, not succeed. It should use read preference secondary and then throw "read preference in a transaction must be primary". Let's update this test to check that the command indeed fails.

Let's add a second test that puts read preference secondary in clientOptions, read preference primary in the startTransaction arguments and then executes runCommand with no readPreference. That runCommand should succeed.

jyemin: The spec fragment that I based this on is:

> The RunCommand method is considered a read operation and MUST use the transaction’s read preference.

I'll wait for Shane to weigh in since he already approved this.

source/transactions/tests/run-command.json
```diff
@@ -0,0 +1,205 @@
...
+          }
+        },
+        {
+          "name": "runCommand",
+          "arguments": {
+            "command": {
+              "count": "test"
+            },
+            "session": "session0",
+            "readPreference": {
```
rstam: This is ambiguous...

1. In all other tests a readPreference means create a new database/collection object with the readPreference (I think? because most methods *don'* take an explicit readPreference)
2. In this test it *could* (and probably does) mean pass an explicit readPreference to runCommand

I think it's a mistake for readPreference to be buried inside the arguments (unless it *really* is an argument to the operation as could be the case in an explicit readPreference passed to runCommand).

source/transactions/tests/run-command.json
ShaneHarvey: @rstam I think you are confusing readConcern and readPreference. According to the spec runcommand should always use the transaction's readPreference.

source/transactions/tests/run-command.json
ShaneHarvey: Opened https://jira.mongodb.org/browse/SPEC-1097

source/transactions/tests/run-command.json
ShaneHarvey: Oh, you're right Robert, I'm the one who is confused. I didn't realize that section of the spec was added.

jyemin: Done

rstam:
@jyemin Shane opened https://jira.mongodb.org/browse/SPEC-1097

Is this pull request dependent on that?

source/transactions/tests/README.rst
```diff
@@ -98,7 +98,14 @@ For each YAML file, for each element in ``tests``:
...
      "abortTransaction", call the named method on ``session0`` or
-     ``session1``, depending on the "session" argument.
+     ``session1``, depending on the "session" argument. 
+   - If ``name`` is "runCommand", call the runCommand method on the database
+     specified in the test. Pass the argument named "command" to the runCommand 
+     method. Pass ``session0`` or ``session1`` to the runCommand method, depending 
+     on which session's name is in the arguments list. If ``arguments`` 
+     contains no "session", pass no explicit session to the method. If ``arguments`` 
+     includes "readPreference", also pass the read preference to the runCommand 
+     method.
```
rstam: This text explaining how to resolve the ambiguity of readPreference would not be necessary if the ambiguity didn't exist in the first place.

Shane opened: https://jira.mongodb.org/browse/SPEC-1097

source/transactions/tests/run-command.json
```diff
@@ -0,0 +1,236 @@
...
+          }
+        },
+        {
+          "name": "commitTransaction",
+          "arguments": {
+            "session": "session0"
+          }
+        }
+      ],
+      "expectations": [
```
rstam: How is this testing that the correct read preference was used?

source/transactions/tests/run-command.json
```diff
@@ -0,0 +1,236 @@
...
```
rstam: How is this testing that the correct read preference was used?

source/transactions/tests/run-command.json
```diff
@@ -0,0 +1,236 @@
...
+        },
+        {
+          "name": "runCommand",
+          "arguments": {
+            "command": {
+              "count": "test"
+            },
+            "session": "session0",
+            "readPreference": {
+              "mode": "Primary"
```
rstam: What's being tested? The default read preference is already primary. Even though we're specifying an explicit read preference of primary, if the explicit value is the same as the default value we have no way of knowing which value was actually used.

source/transactions/tests/run-command.json
```diff
@@ -0,0 +1,236 @@
...
+          }
+        },
+        {
+          "name": "commitTransaction",
+          "arguments": {
+            "session": "session0"
+          }
+        }
+      ],
+      "expectations": [
```
rstam: How is this testing that the correct read preference was used?

jyemin:
No, it's not dependent on SPEC-1097.   SPEC-1097 would just refactor this test a bit but it doesn't invalidate it.

source/transactions/tests/run-command.json
jyemin: I thought it was testing it indirectly in that the count and the commitTransaction commands succeed, and that could only happen if run on a primary (since secondary read-only transactions are not supported).  But I just tried it and confirmed that a secondary will happily execute a read-only transaction, so this is not a good test.  @ShaneHarvey does that surprise you?
I suppose the test could be improved by making the command one that has to run on a primary in order to succeed, e.g. insert

rstam:
OK. I agree that these tests could be adopted as is and refactored after the fact for/by SPEC-1097. Either order is fine.

source/transactions/tests/run-command.json
jyemin: SPEC-1097 will be done separately, at which time this test can be updated with all the others.

source/transactions/tests/run-command.json
ShaneHarvey: > But I just tried it and confirmed that a secondary will happily execute a read-only transaction, so this is not a good test. does that surprise you?

Very surprising. Txns on secondaries are not supposed to work AFAIK https://jira.mongodb.org/browse/SERVER-34557.

Can you open a server ticket about it?

source/transactions/tests/run-command.json
ShaneHarvey: The insert would not succeed if it used Secondary readPreference.

source/transactions/tests/run-command.json
ShaneHarvey: I suppose it doesn't? But it doesn't hurt to test this combination can be done without error.

source/transactions/tests/run-command.yml
```diff
@@ -0,0 +1,167 @@
...
+            lsid: session0
+            txnNumber:
+              $numberLong: "1"
+            startTransaction:
+            autocommit: false
+            writeConcern:
+          command_name: commitTransaction
+          database_name: admin
+
+  - description: run command fails with explicit secondary read preference
```
ShaneHarvey: Can you add another test like this but put the secondary read preference on startTransaction?

source/transactions/tests/run-command.json
jyemin: Note: When Robert originally commented, it was a count (which should also fail, but didn't when i tried it)

source/transactions/tests/run-command.yml
jyemin: Done

jyemin:
Pushed new commit with test requested by @ShaneHarvey

PR #331, opened 2018-05-24
fix unchanged command monitoring options to follow previous changes

saghm:
The command monitoring bulk tests were updated to use a different syntax for options, but a few places in the insertMany file were missed.

### DIFF
--- a/source/command-monitoring/tests/insertMany.json
+++ b/source/command-monitoring/tests/insertMany.json
@@ -32,7 +32,9 @@
                   "x": 22
                 }
               ],
-              "ordered": true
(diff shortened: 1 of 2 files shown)


ajdavis:
@saghm there's a mistake in here that I didn't notice while reviewing; I discovered it once I updated the C Driver's copy of insertMany.json. We shouldn't change the "expectations" arrays, only the "operations" arrays. Can you change the expectations back please?

saghm:
Oops, sorry about that! https://github.com/mongodb/specifications/pull/347

PR #335, opened 2018-06-01
SPEC-1071 Specify which commands to fail in failCommand

ShaneHarvey:
Update the failCommand failpoint with "failCommands" so that the failpoint is reliable on a multi-member replica set (https://jira.mongodb.org/browse/SERVER-34943).

The "failCommands" argument was added in SERVER-35004.

### DIFF
--- a/source/retryable-writes/tests/retryableErrors.json
+++ b/source/retryable-writes/tests/retryableErrors.json
@@ -9,7 +9,7 @@
       "x": 22
     }
   ],
-  "minServerVersion": "4.0.0",
+  "minServerVersion": "4.1.0",
   "tests": [
     {
       "description": "InsertOne succeeds after connection failure",
--- a/source/retryable-writes/tests/retryableErrors.json
+++ b/source/retryable-writes/tests/retryableErrors.json
@@ -9,16 +9,22 @@
       "x": 22
     }
   ],
-  "minServerVersion": "3.7.9",
+  "minServerVersion": "4.0.0",
   "tests": [
     {
       "description": "InsertOne succeeds after connection failure",
+      "clientOptions": {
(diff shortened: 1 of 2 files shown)


source/retryable-writes/tests/retryableErrors.json
```diff
@@ -9,16 +9,22 @@
       "x": 22
     }
   ],
-  "minServerVersion": "3.7.9",
+  "minServerVersion": "4.1.0",
```
rstam: Why 4.1.0? Shouldn't it be 4.0.0?

rstam: Or maybe 4.0.0-rc1?

ShaneHarvey:
Tested in Python locally against a single member replica and a 3 member with and without auth.

source/retryable-writes/tests/retryableErrors.json
ShaneHarvey: There has not been a 4.0 release with the "failCommands" feature, see the comment in YAML
```
# TODO: this should change to 4.0.0 once a 4.0 version is released that has
# SERVER-35004.
```

rstam: Sorry. I only looked at the JSON (assuming it was the same as the YAML, which it is except that the comments are lost).

Apparent 4.0.0 latest already supports this, as the tests are failing on Evergreen.

ShaneHarvey: We can change it to 4.0.0-rc2 (or just 4.0.0) once rc2 is released.

ShaneHarvey:
Thanks for the quick review!

PR #337, opened 2018-06-01
Add collectionOptions to command monitoring tests

jyemin:
This commit aligns the command monitoring tests with the format we agreed on for transaction spec tests for specifying the collection options to apply before executing an operation


source/command-monitoring/tests/unacknowledgedBulkWrite.json
```diff
@@ -13,6 +13,11 @@
       "comment": "On a 2.4 server, no GLE is sent and requires a client-side manufactured reply",
       "operation": {
         "name": "bulkWrite",
```
rstam: Where's the "object" field?

Or do we not have consistent formats across tests yet?

PR #338, opened 2018-06-04
Add connection string test for subdelimiters in userinfo

xdg:
https://tools.ietf.org/html/rfc3986#section-3.2.1 allows subdelimiters in user info sections without percent escaping.  This commit adds a test to verify that drivers allow them.

### DIFF
--- a/source/connection-string/connection-string-spec.rst
+++ b/source/connection-string/connection-string-spec.rst
@@ -11,8 +11,8 @@ Connection String Spec
 :Advisors: \A. Jesse Jiryu Davis, Jeremy Mikola, Anna Herlihy
 :Status: Approved
 :Type: Standards
-:Last Modified: June 10, 2017
-:Version: 1.2
+:Last Modified: June 04, 2018
+:Version: 1.3
 
 .. contents::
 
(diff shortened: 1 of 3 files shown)


rozza:
The spec explicitly states:
> Both username and password MUST be encoded according to RFC 3986 Section 2.1, "Percent-Encoding".

I think although they may survive url unescaping, it is not a good example of a well formed connection string according to the Spec.

> For historical reasons, this spec does not require drivers to throw an exception if other characters listed in RFC 3986 as delimiters appear in the username or password.

I'm unclear if some implementations would throw an error in this scenario. So I think if you want to mandate explicit support for this test, it would requires a change in the verbage of the spec.

See: https://github.com/mongodb/specifications/commit/4631ccd4f825fb1a3aba204510023f9b4d193a05

xdg:
Why this is relevant: I noticed that our Atlas test connection URIs use "=", which is a subdelimiter, and the Perl driver choked on that because it was an unescaped reserved character.  If Atlasis generating URIs that are not escaping subdelimiters, then we ought to encourage drivers to follow the URI RFC and allow subdelimiters and this test is designed to do so.

Our spec is still too vague on the topic of encoding.  RFC 3986 Section 2.1 only describes the mechanism, it doesn't describe when/what to encode.  The controlling sections of the RFC are 2.2 and 3.2.1.

Here is the relevant bit of 2.2:

```
   URI producing applications should percent-encode data octets that
   correspond to characters in the reserved set unless these characters
   are specifically allowed by the URI scheme to represent data in that
   component.
```

And 3.2.1:

```
   userinfo    = *( unreserved / pct-encoded / sub-delims / ":" )
```

While the spec does not require the driver to throw errors if delimiters are used, we should provide a test for the desired behavior and let drivers that need back-compatibility skip that test (as we tend to do elsewhere).  The spec language about errors doesn't asserting that the URI follows the RFC rules -- instead it specifies a scheme-specific additional constraint on interpreting the userinfo section.  (E.g. according to the RFC, one could have ":" characters in the password, which we disallow.)

Actually, read literally, our spec says that the username and password both must be percent-encoded, even for ASCII characters, as there is no mention of the allowed characters set.

How about I update the spec to clarify the allowed/disallowed characters per the descriptions above (2.2 and 3.2.1)?  (While keeping the scheme-specific behaviors and optionality we already have.)

rozza:
> How about I update the spec to clarify the allowed/disallowed characters per the descriptions above (2.2 and 3.2.1)? (While keeping the scheme-specific behaviors and optionality we already have.)

Sounds good to me @xdg

xdg:
Rebased and merged.

PR #339, opened 2018-06-05
 SPEC-1085 Test that writes are not retried inside a transaction

ShaneHarvey:
Adds a test to ensure that drivers do not retry insert operations within a transaction even when retryWrites=true. Also adds errorLabelsContain and ErrorLabelsOmit to verify that the insert operation raises an exception with the "TransientTransactionError" error label.

I've only added a single test (of insertOne) for now. We can expand this to test more write operations when we agree on the format.

### DIFF
--- a/source/transactions/tests/README.rst
+++ b/source/transactions/tests/README.rst
@@ -31,7 +31,7 @@ commands listed in the "failCommands" field. See `SERVER-35004`_ and
 
 The ``failCommand`` fail point may be configured like so::
 
-    db.runCommand({
+    db.adminCommand({
         configureFailPoint: "failCommand",
         mode: <string|document>,
         data: {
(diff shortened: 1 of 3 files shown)


source/transactions/tests/README.rst
```diff
@@ -35,41 +85,54 @@ Each YAML file has the following keys:
...
     MongoClient.startSession().
 
   - ``operations``: Array of documents, each describing an operation to be
     executed. Each document has the following fields:
 
-      - ``name``: The name of the operation on ``object``.
+    - ``name``: The name of the operation on ``object``.
+
+    - ``object``: The name of the object to perform the operation on. Can be
+      "database", collection", "session0", or "session1".
```
ajdavis: missing "

jyemin:
I'm going to try to implement this before approving, but looks good.

PR #343, opened 2018-06-07
SPEC-1086 Test errorLabels in transactions

ShaneHarvey:
No changes to the test format. Adds test cases for transient, non-transient, and unknown commit result error labels.


ShaneHarvey:
Yep I added errorLabelsContain/Omit descriptions in https://github.com/mongodb/specifications/pull/339.

PR #345, opened 2018-06-07
SPEC-1090 Test retryable writeConcernErrors

ShaneHarvey:
WriteConcernFailed is not a retryable writes error, see [SPEC-1063](https://jira.mongodb.org/browse/SPEC-1063).

Adds tests for retryable and non-retryable writeConcernErrors:
- All retryable server errors are tested in `insertOne-serverErrors.yml`. 
- For all other operations, test PrimarySteppedDown and writeConcernError ShutdownInProgress.


source/retryable-writes/tests/retryableErrors.yml
```diff
@@ -2,9 +2,8 @@ data:
     - { _id: 1, x: 11 }
     - { _id: 2, x: 22 }
 
-# TODO: this should change to 4.0.0 once a 4.0 version is released that has
-# SERVER-35004.
-minServerVersion: '4.1.0'
+# TODO: this should change to 4.0 once 4.0.0 is released.
+minServerVersion: '3.99'
```
ShaneHarvey: I think 3.99 is appropriate now that the 4.0.0 rc's have the fail point.

ShaneHarvey:
> I'm also confused as to why we have a retryableErrors.yml/json file - surely all the test files test retryable errors.
I think I meant to name this file allRetryableErrors because it tests all of them.

> Can these tests be merged into their respective command test files eg: insertOne.yml, insertMany.yml, findAndDelete.yml

They need to be separate because they require 4.0.0-rc3 but I can split them and add bulkWrite-4.0.yml, deleteOne-4.0.yml, etc..

ShaneHarvey:
Split into insertOne-serverErrors.yml, bulkWrite-serverErrors.yml, etc...

All of the new retryable server errors are tested in insertOne-serverErrors.yml. For all other operations, only PrimarySteppedDown and writeConcernError ShutdownInProgress are tested.

source/retryable-writes/tests/insertOne-serverErrors.yml
```diff
@@ -303,3 +302,170 @@ tests:
...
+        clientOptions:
+            retryWrites: true
+        failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+                failCommands: ["insert"]
+                writeConcernError:
+                    code: 11602
+                    errmsg: Replication is being shut down
```
kevinAlbs: errmsg shouldn't be checked in this case. Should we leave it as an empty string?

ShaneHarvey: Agreed the errmsg shouldn't be checked but I think it's good to include a message since the server always(?) does. Also it's easier to debug:
```
>>> _raise_write_concern_error({"code":11602, "errmsg": ""})
  File "<stdin>", line 1, in <module>
  File "pymongo/helpers.py", line 208, in _raise_write_concern_error
    error.get("errmsg"), error.get("code"), error)
pymongo.errors.WriteConcernError
```
Vs:
```
>>> _raise_write_concern_error({"code":11602, "errmsg": "Replication is being shut down"})
  File "<stdin>", line 1, in <module>
  File "pymongo/helpers.py", line 208, in _raise_write_concern_error
    error.get("errmsg"), error.get("code"), error)
pymongo.errors.WriteConcernError: Replication is being shut down
```

ShaneHarvey:
Thanks @rozza and @kevinAlbs!

PR #352, opened 2018-06-13
Specify that startAtOperationTime is a changeStream option

prashantmital:
The spec clearly states for each change stream option whether it is (a) an option of the `$changeStream` aggregation stage or (b) an aggregation command option. This had not been the case for the recently added `startAtOperationTime` option and this change fixes that.


prashantmital:
@daprahamian not sure who would need to approve this. I tagged you since you were the one who added the changes related to `startAtOperationTime` to the spec in the first place...

mbroadst:
Thanks!

PR #353, opened 2018-06-13
change stream startAtOperationTime is no longer a document

daprahamian:
When originally writing the spec, `startAtClusterTime` was a document with a single member `ts` that was a `Timestamp`. Now, we instead have `startAtOperationTime`, which is just a raw timestamp.


mbroadst:
@rathisekaran this will require other drivers to update if they haven't already completed the change stream followup work

daprahamian:
Closing, as this was included in #355

jmikola:
@daprahamian: Can the branch be deleted?

PR #355, opened 2018-06-14
Updates spec to detail acquisition of startAtOperationTime for ChangeStreams

daprahamian:
Fixes SPEC-1114

### DIFF
--- a/source/causal-consistency/causal-consistency.rst
+++ b/source/causal-consistency/causal-consistency.rst
@@ -136,12 +136,9 @@ implied. See the ``causalConsistency`` section.
 MongoClient changes
 ===================
 
-There are no public API changes to ``MongoClient`` to support causal consistency.
+There are no API changes to ``MongoClient`` to support causal consistency.
 Applications indicate whether they want causal consistency by setting the
 ``causalConsistency`` field in the options passed to the ``startSession`` method.
-
-While not public, the ``MongoClient`` MUST maintain a private ``lastOperationTime`` member
-containing the most recently observed ``operationTime`` received.
 
 SessionOptions changes
 ======================
@@ -283,9 +280,7 @@ The ``operationTime`` MUST be stored in the ``ClientSession`` to later be passed
 stored in either case.
 
 Drivers MUST examine all responses from the server for the presence of an
-``operationTime`` field and store the value in the ``ClientSession``. Additionally,
-the most recent ``operationTime`` seen MUST be stored in the ``MongoClient`` field
-``lastOperationTime``.
+``operationTime`` field and store the value in the ``ClientSession``.
 
 When connected to a standalone node command replies do not include an
 ``operationTime`` field. All operations against a standalone node are causally
--- a/source/causal-consistency/causal-consistency.rst
+++ b/source/causal-consistency/causal-consistency.rst
@@ -136,9 +136,12 @@ implied. See the ``causalConsistency`` section.
 MongoClient changes
 ===================
 
-There are no API changes to ``MongoClient`` to support causal consistency.
+There are no public API changes to ``MongoClient`` to support causal consistency.
 Applications indicate whether they want causal consistency by setting the
 ``causalConsistency`` field in the options passed to the ``startSession`` method.
+
+While not public, the ``MongoClient`` MUST maintain a private ``lastOperationTime`` member
+containing the most recently observed ``operationTime`` received.
 
 SessionOptions changes
 ======================
@@ -280,7 +283,9 @@ The ``operationTime`` MUST be stored in the ``ClientSession`` to later be passed
 stored in either case.
 
 Drivers MUST examine all responses from the server for the presence of an
-``operationTime`` field and store the value in the ``ClientSession``.
+``operationTime`` field and store the value in the ``ClientSession``. Additionally,
+the most recent ``operationTime`` seen MUST be stored in the ``MongoClient`` field
+``lastOperationTime``.
 
 When connected to a standalone node command replies do not include an
 ``operationTime`` field. All operations against a standalone node are causally
--- a/source/change-streams/change-streams.rst
+++ b/source/change-streams/change-streams.rst
@@ -173,7 +173,9 @@ Driver API
     private resumeToken: Document;
 
     /**
-     * The most recent operationTime received in a server response at the
+     * An Timestamp that specifies the earliest time at which changes can be received.
+     * Can be specified by the user.
+     * Defaults tothe most recent operationTime received in a server response at the
      * time of ChangeStream creation.
      */
     private startAtOperationTime: Timestamp;
@@ -547,9 +549,7 @@ Why do we need to send a default ``startAtOperationTime`` in the ``$changeStream
 -----------------------------------------------------------------------------------------
 
 ``startAtOperationTime`` allows a user to create a resumable change stream even when a result
-(and corresponding resultToken) is not available until a later point in time. However, 
-it has the added benefit of allowing a change stream to resume and receive all changes that
-it may have missed since it's first creation.
+(and corresponding resultToken) is not available until a later point in time.
 
 For example:
 
--- a/source/change-streams/change-streams.rst
+++ b/source/change-streams/change-streams.rst
@@ -273,7 +273,7 @@ Driver API
     collation: Optional<Document>;
 
     /**
-     * The change stream will only provides changes that occurred after the
+     * The change stream will only provide changes that occurred at or after the
      * specified timestamp. Any command run against the server will return
      * an operation time that can be used here.
      * @since 4.0
@@ -420,7 +420,7 @@ Once a ``ChangeStream`` has encountered a resumable error, it MUST attempt to re
 - If the ``ChangeStream`` has not received any changes, and ``resumeAfter`` is not specified, and the max wire version is >= ``7``:
 
     - The driver MUST execute the known aggregation command.
-    - IThe driver MUST specify the ``startAtOperationTime`` key set to the ``startAtOperationTime`` provided by the user or saved from the original aggregation.
+    - The driver MUST specify the ``startAtOperationTime`` key set to the ``startAtOperationTime`` provided by the user or saved from the original aggregation.
     - The driver MUST NOT set a ``resumeAfter`` key.
     - In this case, the ``ChangeStream`` will return all changes that occurred after the specified ``startAtOperationTime``.
 - Else:
--- a/source/change-streams/change-streams.rst
+++ b/source/change-streams/change-streams.rst
@@ -173,14 +173,6 @@ Driver API
     private resumeToken: Document;
 
     /**
-     * An Timestamp that specifies the earliest time at which changes can be received.
-     * Can be specified by the user.
-     * Defaults tothe most recent operationTime received in a server response at the
-     * time of ChangeStream creation.
-     */
-    private startAtOperationTime: Timestamp;
-
-    /**
      * The pipeline of stages to append to an initial ``$changeStream`` stage
      */
     private pipeline: Array<Document>;
@@ -283,8 +275,7 @@ Driver API
     /**
      * The change stream will only provides changes that occurred after the
      * specified timestamp. Any command run against the server will return
-     * an operation time that can be used here. The default value is an
-     * operation time obtained from the server before the change stream was created.
+     * an operation time that can be used here.
      * @since 4.0
      * @see https://docs.mongodb.com/manual/reference/method/db.runCommand/
      * @note this is an option of the `$changeStream` pipeline stage.
@@ -362,6 +353,8 @@ Command syntax:
       ...
     }
 
+Drivers MUST use the ``ns`` returned in the ``aggregate`` command to set the ``collection`` option in subsequent ``getMore`` commands.
+
 ``MongoClient.watch`` helper
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
@@ -380,6 +373,8 @@ Command syntax:
     }
 
 The helper MUST run the command against the `admin` database
+
+Drivers MUST use the ``ns`` returned in the ``aggregate`` command to set the ``collection`` option in subsequent ``getMore`` commands.
 
 ChangeStream
 ------------
@@ -404,21 +399,11 @@ startAtOperationTime
 
 ``startAtOperationTime`` specifies that a change stream will only return changes that occurred at or after the specified ``Timestamp``.
 
-The server expects ``startAtOperationTime`` as a BSON Timestamp
-
-.. code:: typescript
-
-    {
-      ts: Timestamp;
-    }
-
-Drivers MUST allow users to specify a ``startAtOperationTime`` option in the ``Collection.watch`` and ``Database.watch`` helpers. They MUST allow users to specify this value as a raw ``Timestamp``.
+The server expects ``startAtOperationTime`` as a BSON Timestamp. Drivers MUST allow users to specify a ``startAtOperationTime`` option in the ``Collection.watch`` and ``Database.watch`` helpers. They MUST allow users to specify this value as a raw ``Timestamp``.
 
 ``startAtOperationTime`` and ``resumeAfter`` are mutually exclusive; if both ``startAtOperationTime`` and ``resumeAfter`` are set, the server will return an error. Drivers MUST NOT throw a custom error, and MUST defer to the server error.
 
-If neither ``startAtOperationTime`` nor ``resumeAfter`` are specified, and the max wire version is >= ``7`` drivers SHOULD set a default ``startAtOperationTime`` using the ``MongoClient`` field ``lastOperationTime`` (see `Causal Consistency <../causal-consistency/causal-consistency.rst#mongoclient-changes>`_.
-
-If the ``lastOperationTime`` is not available, ``startAtOperationTime`` MAY be omitted. If this is the case, then the ``ChangeStream`` MUST save the ``lastOperationTime`` when the initial ``aggregate`` command returns.
+If neither ``startAtOperationTime`` nor ``resumeAfter`` are specified, and the max wire version is >= ``7`` the ``ChangeStream`` MUST save the ``operationTime`` from the initial ``aggregate`` command when it returns.
 
 resumeAfter
 ^^^^^^^^^^^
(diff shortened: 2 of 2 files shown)


source/change-streams/change-streams.rst
```diff
@@ -408,7 +408,9 @@ Drivers MUST allow users to specify a ``startAtOperationTime`` option in the ``C
 
 ``startAtOperationTime`` and ``resumeAfter`` are mutually exclusive; if both ``startAtOperationTime`` and ``resumeAfter`` are set, the server will return an error. Drivers MUST NOT throw a custom error, and MUST defer to the server error.
 
-If neither ``startAtOperationTime`` nor ``resumeAfter`` are specified, and the max wire version is >= ``7`` drivers MUST set a default ``startAtOperationTime``. A default timestamp can be obtained from the ``operationTime`` field on any server response (e.g. ``isMaster``). This allows change streams to be resumed before the first notification is received.
+If neither ``startAtOperationTime`` nor ``resumeAfter`` are specified, and the max wire version is >= ``7`` drivers SHOULD set a default ``startAtOperationTime`` using the ``MongoClient`` field ``lastOperationTime`` (see `Causal Consistency <../causal-consistency/causal-consistency.rst#mongoclient-changes>`_.
+
+If the ``lastOperationTime`` is not available, ``startAtOperationTime`` MAY be omitted. If this is the case, then the ``ChangeStream`` MUST save the ``operationTime`` value returned from the server in response to the initial aggregation call issued by the change stream.
```
xdg: Here, I think I'd be more specific and say that in this case it must save the `lastOperationTime` from MongoClient (which is guaranteed to be populated at that point).  That's more prescriptive than saying "value returned from the server in response..." but amounts to the same thing.

source/change-streams/change-streams.rst
```diff
@@ -425,7 +427,7 @@ Once a ``ChangeStream`` has encountered a resumable error, it MUST attempt to re
 - If the ``ChangeStream`` has not received any changes, and ``resumeAfter`` is not specified, and the max wire version is >= ``7``:
 
     - The driver MUST execute the known aggregation command.
-    - The driver MUST specify the ``startAtOperationTime`` key set to the original timestamp from when the changestream was first created.
+    - The driver MUST specify the ``startAtOperationTime`` key set to the original timestamp used in the initial aggregation command.
```
xdg: I think you should add `startAtOperationTime` as a private member (like we do with `resumeToken`) and reference that here.  Then it's clearer we mean whatever startAtOperationTime was either passed in originally or was set earlier.

source/change-streams/change-streams.rst
```diff
@@ -534,6 +536,29 @@ The `CursorKilled` or `Interrupted` error implies implies some other actor kille
 The `CappedPositionLost` error implies falling off of the back of the oplog,
 so resuming is impossible.
 
+-----------------------------------------------------------------------------------------
+Why do we need to send a default ``startAtOperationTime`` in the ``$changeStream`` stage?
+-----------------------------------------------------------------------------------------
+
+``startAtOperationTime`` allows a user to create a change stream that does not return
```
xdg: how about: "allows a user to create a resumable change stream even when a result (and corresponding resultToken) is not available until a later point in time."

source/change-streams/change-streams.rst
```diff
@@ -534,6 +536,29 @@ The `CursorKilled` or `Interrupted` error implies implies some other actor kille
...
+results until a later point in time. However, it has the added benefit of allowing
+a change stream to resume and receive all changes that it may have missed since it's first
+creation.
+
+For example:
+
+- A client creates a ``ChangeStream``, and calls ``watch``
+- The ``ChangeStream`` sends out the initial ``aggregate`` call, and receives a response
+with no initial values.
```
xdg: Add "Because there are no initial values, there is no latest resumeToken."

source/change-streams/change-streams.rst
```diff
@@ -546,16 +546,16 @@ so resuming is impossible.
 Why do we need to send a default ``startAtOperationTime`` in the ``$changeStream`` stage?
 -----------------------------------------------------------------------------------------
 
-``startAtOperationTime`` allows a user to create a change stream that does not return
-results until a later point in time. However, it has the added benefit of allowing
-a change stream to resume and receive all changes that it may have missed since it's first
-creation.
+``startAtOperationTime`` allows a user to create a resumable change stream even when a result
+(and corresponding resultToken) is not available until a later point in time. However,
```
xdg: I think the second sentence "However, ..." is either unnecessary or can be streamlined.

source/change-streams/change-streams.rst
```diff
@@ -172,6 +172,12 @@ Driver API
      */
     private resumeToken: Document;
 
+    /**
+     * The most recent operationTime received in a server response at the
+     * time of ChangeStream creation.
```
xdg: Maybe note that this can be provided by the user in the options.

source/change-streams/change-streams.rst
```diff
@@ -534,6 +542,29 @@ The `CursorKilled` or `Interrupted` error implies implies some other actor kille
 The `CappedPositionLost` error implies falling off of the back of the oplog,
 so resuming is impossible.
 
+-----------------------------------------------------------------------------------------
+Why do we need to send a default ``startAtOperationTime`` in the ``$changeStream`` stage?
+-----------------------------------------------------------------------------------------
+
+``startAtOperationTime`` allows a user to create a resumable change stream even when a result
+(and corresponding resultToken) is not available until a later point in time. However,
```
xdg: The second sentence either needs to be removed or streamlined, now.

source/causal-consistency/causal-consistency.rst
```diff
@@ -280,7 +283,9 @@ The ``operationTime`` MUST be stored in the ``ClientSession`` to later be passed
 stored in either case.
 
 Drivers MUST examine all responses from the server for the presence of an
-``operationTime`` field and store the value in the ``ClientSession``.
+``operationTime`` field and store the value in the ``ClientSession``. Additionally,
```
jyemin: Drivers are already required to store the most recent cluster time in order to gossip it.  See https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst#receiving-the-current-cluster-time.  The clusterTime field in $clusterTime is guaranteed to be at least as recent as the operationTime.  Why shouldn't drivers just use that?  

The Java driver PR currently does that and it looks fine to me: https://github.com/rozza/mongo-java-driver/pull/278/files#diff-62448c5972eb4eea359195432e9003d0R91, and simpler than requiring another field to be tunneled through multiple layers of API.

source/change-streams/change-streams.rst
```diff
@@ -534,6 +542,29 @@ The `CursorKilled` or `Interrupted` error implies implies some other actor kille
...
 so resuming is impossible.
 
+-----------------------------------------------------------------------------------------
+Why do we need to send a default ``startAtOperationTime`` in the ``$changeStream`` stage?
+-----------------------------------------------------------------------------------------
+
+``startAtOperationTime`` allows a user to create a resumable change stream even when a result
+(and corresponding resultToken) is not available until a later point in time. However, 
+it has the added benefit of allowing a change stream to resume and receive all changes that
+it may have missed since it's first creation.
```
jyemin: its, not it's

source/change-streams/change-streams.rst
jyemin: resumeToken, not resultToken

source/change-streams/change-streams.rst
```diff
@@ -408,7 +416,9 @@ Drivers MUST allow users to specify a ``startAtOperationTime`` option in the ``C
 
 ``startAtOperationTime`` and ``resumeAfter`` are mutually exclusive; if both ``startAtOperationTime`` and ``resumeAfter`` are set, the server will return an error. Drivers MUST NOT throw a custom error, and MUST defer to the server error.
 
-If neither ``startAtOperationTime`` nor ``resumeAfter`` are specified, and the max wire version is >= ``7`` drivers MUST set a default ``startAtOperationTime``. A default timestamp can be obtained from the ``operationTime`` field on any server response (e.g. ``isMaster``). This allows change streams to be resumed before the first notification is received.
+If neither ``startAtOperationTime`` nor ``resumeAfter`` are specified, and the max wire version is >= ``7`` drivers SHOULD set a default ``startAtOperationTime`` using the ``MongoClient`` field ``lastOperationTime`` (see `Causal Consistency <../causal-consistency/causal-consistency.rst#mongoclient-changes>`_.
+
+If the ``lastOperationTime`` is not available, ``startAtOperationTime`` MAY be omitted. If this is the case, then the ``ChangeStream`` MUST save the ``lastOperationTime`` when the initial ``aggregate`` command returns.
```
rozza: > If the ``lastOperationTime`` is not available, ``startAtOperationTime`` MAY be omitted. If this is the case, then the ``ChangeStream`` MUST save the ``lastOperationTime`` when the initial ``aggregate`` command returns.

Doesn't this change preclude the need to have a `lastOperationTime` in the MongoClient at all? 

Could it not work like: Get last operation time from the ClientSession, if null (i.e. no previous operations on the session) save the initial operation time from the aggregate command in the ChangeStream cursor.  

This allows empty initial batches to be retried with the need to plumb back to the MongoClient.

source/change-streams/change-streams.rst
```diff
@@ -172,6 +172,14 @@ Driver API
      */
     private resumeToken: Document;
 
+    /**
+     * An Timestamp that specifies the earliest time at which changes can be received.
```
jyemin: "A" not "An"

source/change-streams/change-streams.rst
```diff
@@ -172,6 +172,14 @@ Driver API
...
+     * Can be specified by the user.
+     * Defaults tothe most recent operationTime received in a server response at the
```
jyemin: Add space in "tothe"

source/change-streams/change-streams.rst
jyemin: I agree, though is there even any benefit of getting the operation time from the client session?  I mean, 99% of the time it will be an implicit session with no operation time anyway.  I suggest simplifying to just the fallback case.

daprahamian: The server design doc specifically said "Start all change streams with a 'startAtOperationTime', and remember that time. ", which is why we do the initial tracking. This change is meant to be a backup in the case that the operationTime is not known.

source/causal-consistency/causal-consistency.rst
rozza: I don't think we should mandate the design decision of  how / where the `lastOperationTime` is stored. Just that its made accessible for commands, so they require.  

For example the Java drivers MongoCollection does not have access to the MongoClient but does now have access to the lastOperationTime.

source/change-streams/change-streams.rst
cswanson310: Ok this is actually an interesting question. I think using the operationTime from the aggregate should be sufficient, see below.

When the server receives a change stream without any `resumeAfter` or `startAtOperationTime`, it starts watching at the most-recent *majority committed* oplog entry. In contrast, the operation time returned from each command represents the time of the most recent write, *not necessarily majority-committed*. This means the following thing can happen:
1. A client starts watching a change stream. Suppose the driver does not attach a 'startAtOperationTime', so it starts watching at time _T<sub>majority</sub>_
2. This operation returns with operationTime _T<sub>recent_op</sub>_ > _T<sub>majority</sub>_

There are potentially many oplog entries between _T<sub>recent_op</sub>_ and _T<sub>majority</sub>_. Normally this change stream would return all of them. However, if there's a network error on the next getMore then we will start watching the second time at _T<sub>recent_op</sub>_, and miss everything in between.

Contrast this scenario to if the driver always specified `startAtOperationTIme` when opening a change stream. In that scenario, a network error during the initial aggregate would always return the same results, namely everything after _T<sub>recent_op</sub>_ (or some other operation time the driver has stored, which is probably more recent than _T<sub>majority</sub>_).

Now, it's not clear to me the difference in semantics here matters at all. When you open a new change stream, we don't guarantee much about what you will or will not see. In either case above, if the client opened a change stream, then that client performed a write, it would see that write.

A related case which we were attempting to solve is this: Suppose a client opens a change stream and watches/iterates it for several hours with no results (possibly because of a $match that filtered everything out). After several hours there's a network error during one of the getMores, around the same time that an event happens which the change stream was waiting for. We should not resume that change stream from the most-recent majority-committed oplog entry, because that might be skipping that interesting event. In either proposal here, the driver would correctly restart from the original operationTime (scan through those hours of history 😵) then find the interesting event ✅🙌🎉

The extra hours of scanning is an interesting downside that I don't think we've considered well. In this scenario your oplog probably still has enough history and it'll be fine except maybe bringing a bunch of irrelevant oplog into memory. However, the server will actually error if the time you specified is no longer in the oplog (since we'd potentially miss entries in that case), thus preventing you from resuming in the above scenario if the hours changed to something more like days which was larger than your oplog.

Perhaps the better thing to do here would be for the server to return the information we use internally [in SERVER-29929](https://jira.mongodb.org/browse/SERVER-29929). This information is not returned from normal getMores on change streams; it's only passed between mongos and mongod. We could surface this information though. If we did so, drivers could use this to keep track of the most recent time that the stream is guaranteed not to return anything before. This would avoid scanning hours of oplog and potentially raising an error upon resuming if there was a network error.

cswanson310: cc @asya999 and @stbrody

stbrody: You seem to have laid out the behavior and options well, I'll defer to @asya999 and @storch as to what to do from here.

jyemin: Thanks Charlie.  Based on this feedback, we can simplify the spec by requiring use of the operationTime from the aggregate command. Any objections?

daprahamian: 👍 I'll remove the tracking in the MongoClient and incorporate these changes.

ShaneHarvey: @cswanson310 
> In either proposal here, the driver would correctly restart from the original operationTime (scan through those hours of history 😵) 

I think drivers could avoid that by updating the default startAtOperationTime with the operationTime from each (empty) getMore response. Then when the resume happens it resumes from the previous getMore operationTime.

source/change-streams/change-streams.rst
```diff
@@ -354,6 +353,8 @@ Command syntax:
       ...
     }
 
+Drivers MUST use the ``ns`` returned in the ``aggregate`` command to set the ``collection`` option in subsequent ``getMore`` commands.
```
xdg: Under what circumstances would this not be the same as the collection used to send the original aggregate command? Is this some special database/client watch implementation thing?

source/change-streams/change-streams.rst
```diff
@@ -390,25 +393,17 @@ Single Server Topologies
...
-The server expects ``startAtOperationTime`` as a BSON Timestamp
-
-.. code:: typescript
-
-    {
-      ts: Timestamp;
-    }
-
-Drivers MUST allow users to specify a ``startAtOperationTime`` option in the ``Collection.watch`` and ``Database.watch`` helpers. They MUST allow users to specify this value as a raw ``Timestamp``.
+The server expects ``startAtOperationTime`` as a BSON Timestamp. Drivers MUST allow users to specify a ``startAtOperationTime`` option in the ``Collection.watch`` and ``Database.watch`` helpers. They MUST allow users to specify this value as a raw ``Timestamp``.
```
xdg: Not allowed for MongoClient.watch?

source/change-streams/change-streams.rst
```diff
@@ -390,25 +393,17 @@ Single Server Topologies
...
 
 ``startAtOperationTime`` and ``resumeAfter`` are mutually exclusive; if both ``startAtOperationTime`` and ``resumeAfter`` are set, the server will return an error. Drivers MUST NOT throw a custom error, and MUST defer to the server error.
 
-If neither ``startAtOperationTime`` nor ``resumeAfter`` are specified, and the max wire version is >= ``7`` drivers MUST set a default ``startAtOperationTime``. A default timestamp can be obtained from the ``operationTime`` field on any server response (e.g. ``isMaster``). This allows change streams to be resumed before the first notification is received.
+If neither ``startAtOperationTime`` nor ``resumeAfter`` are specified, and the max wire version is >= ``7`` the ``ChangeStream`` MUST save the ``operationTime`` from the initial ``aggregate`` command when it returns.
```
xdg: Only if there are no results?  Or always?

source/change-streams/change-streams.rst
```diff
@@ -425,7 +420,10 @@ Once a ``ChangeStream`` has encountered a resumable error, it MUST attempt to re
 - If the ``ChangeStream`` has not received any changes, and ``resumeAfter`` is not specified, and the max wire version is >= ``7``:
 
     - The driver MUST execute the known aggregation command.
-    - The driver MUST specify the ``startAtOperationTime`` key set to the original timestamp from when the changestream was first created.
+    - If the original aggregation command did not include a user-provided ``startAtOperationTime``:
+
+        - The driver MUST specify the ``startAtOperationTime`` key set to the ``operationTime`` saved from the response to the original ``aggregate`` command.
+
```
xdg: I would collapse this to a single bullet point: "The driver MUST specify the ``startAtOperationTime`` key set to the ``startAtOperationTime`` provided by the user or saved from the original aggregation."

source/change-streams/change-streams.rst
xdg: OK.  Didn't realize but I agree we ought to pass back what we get.

source/change-streams/change-streams.rst
cswanson310: @ShaneHarvey I also thought about that - it _seems_ right, but I don't think it is. The operationTime returned on each getMore has no substantial correlation with what the underlying scan on the oplog may or may not have consumed yet. This is why we had to invent a new thing for SERVER-29929.

In particular, the collection scan on the oplog is going to stop at the most-recent majority-committed entry, but the operationTime could still be ahead of that. Anything in between those two could be lost.

@jyemin that sounds reasonable to me!

ShaneHarvey: Ah thanks for explaining that again Charlie! I missed it in your original comment.

source/change-streams/change-streams.rst
jyemin: This is similar to what drivers must do for listCollections.

source/change-streams/change-streams.rst
```diff
@@ -425,7 +420,7 @@ Once a ``ChangeStream`` has encountered a resumable error, it MUST attempt to re
 - If the ``ChangeStream`` has not received any changes, and ``resumeAfter`` is not specified, and the max wire version is >= ``7``:
 
     - The driver MUST execute the known aggregation command.
-    - The driver MUST specify the ``startAtOperationTime`` key set to the original timestamp from when the changestream was first created.
+    - IThe driver MUST specify the ``startAtOperationTime`` key set to the ``startAtOperationTime`` provided by the user or saved from the original aggregation.
```
jyemin: Ithe => the

source/change-streams/change-streams.rst
```diff
@@ -275,8 +275,7 @@ Driver API
     /**
      * The change stream will only provides changes that occurred after the
```
jyemin: While we're here: make provides singular

PR #358, opened 2018-06-18
Prohibit explicit readConcern in transaction

ajdavis:


ShaneHarvey:
Why is this changing? I thought we agreed that if a readConcern/writeConcern/readPreference is passed as an argument it overrides the transaction setting.  I don't understand why readConcern should be treated different from writeConcern/readPreference.

ajdavis:
I reached this portion of the spec in the C Driver, where we need to test the behavior of per-operation read concern, write concern, and read preference. I realized while writing tests that the readConcern rule doesn't work: see my edits to the Justifications section in this PR. SPEC-1118, but all the info is in the PR.

source/transactions/transactions.rst
```diff
@@ -226,14 +226,11 @@ readConcern to the first command in a transaction if and only if the
 readConcern is supplied and not the default. Drivers MUST NOT add to
 subsequent commands the readConcern from the transaction or any
 readConcern inherited from the collection, database, or client. If the
-user supplies an explicit readConcern via a method option, however,
-drivers MUST apply the readConcern, which will result in a server error.
```
ajdavis: Note: this is false for the first command in a transaction. The server does *not* raise an error, it doesn't know that the driver was provided with a readConcern in two ways.

ajdavis:
Reviewers plus @xdg and @jmikola, do drivers that you know of allow readConcern to be explicitly passed as a parameter to particular operations, such as "find"?

mbroadst:
@ajdavis node does

source/transactions/transactions.rst
```diff
@@ -226,14 +226,11 @@ readConcern to the first command in a transaction if and only if the
...
-See `Users can pass prohibited options to operations in transactions`_.
+user supplies an explicit readConcern via a method option, the driver MUST
+raise an error with the message "Cannot set read concern after starting a
```
kevinAlbs: In this case, should the transaction also be aborted? Suppose the user starts a transaction and executes the first command correctly (without specifying a read concern). But then a later command includes a read concern, causing a driver error. The transaction isn't aborted, and the user could still commit the transaction if they ignored the error. Is that okay? Would the transaction have been aborted before, when we passed read concern to the server?

xdg:
FWIW, I don't like having different rules for read preference, read concern and write concern on operations in a transaction.  We ignore one, error on another, and pass through a third for the server to error on.  That's not great to explain to users.

source/transactions/transactions.rst
ajdavis: A quick test shows that with the current spec (before this PR) you could do:
```
start transaction with readConcern majority
do a find_one in the txn, successfully
do a find_one in the txn with readConcern snapshot
  server replies "Only the first command in a transaction may specify a readConcern"
do a find_one in the txn, successfully
```
The server error does *not* abort the transaction. So if we say that the new client-side error doesn't abort the transaction either then this is a smaller change. I also think it's consistent with a principle that applies to other scenarios: incorrectly specified commands cause parse errors that don't abort the transaction, whereas errors caused by operations on the data (like a write conflict or oversized transaction) do abort it.

ajdavis:
After discussion with Shane, we agree that:

- We think the current readConcern behavior is a spec bug and it must be changed for those drivers that allow per-operation readConcern. The fix might lag our 4.0 releases a little if necessary; consider the currently implemented behavior a bug.
- For consistency with the change to readConcern, let's specify the same client-side error for writeConcern, even though the currently specified behavior correctly causes a server error if users make a mistake.

I've uploaded a new commit with the writeConcern change.

source/transactions/transactions.rst
```diff
@@ -1060,40 +1061,61 @@ we specify now that the readPreference must be checked per-operation.
...
+.. code:: python
 
-For drivers that allow read concern to be passed to a particular
+   client = MongoClient("mongodb://localhost/?readConcernLevel=majority")
+   with client.start_session() as s:
+       # Transaction uses readConcern majority.
+       with s.start_transaction():
+           # The first command in a transaction. Which readConcern?
+           client.db.collection.distinct(
+               readConcern={'level': 'snapshot'})
```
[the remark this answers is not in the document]
ajdavis: D'oh! Fixed.

source/transactions/transactions.rst
kevinAlbs: Ok, that makes sense.

ajdavis:
@behackett and @ShaneHarvey will you review or will you abstain? Thanks.

p-mongo:
The spec ticket for this change is SPEC-1118.

PR #360, opened 2018-06-19
Updated change stream tests

rozza:
startAtOperation time is no longer automatically added to the command - so examples can be removed from the tests.

SPEC-1114


jyemin:
Sorry Ross, I saw this pop up and thought it was the Java driver changes!

PR #362, opened 2018-06-19
Add "arrayFilters" to wire version features

ajdavis:


jmikola:
@ajdavis: This should be good to merge.

PR #365, opened 2018-06-25
SPEC-1124: Checking error messages for resumable errors is redundant

jmikola:
https://jira.mongodb.org/browse/SPEC-1124

Per the SDAM spec, the numeric codes for errors that may contain "not master" or "node is recovering" do not overlap with the three error codes specifically excluded by the change stream spec. Therefore, such errors should already be considered resumable based on the second criterion.

### DIFF
--- a/source/change-streams/change-streams.rst
+++ b/source/change-streams/change-streams.rst
@@ -9,7 +9,7 @@ Change Streams
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: 3.6
-:Last Modified: July 27, 2018
+:Last Modified: July 30, 2018
 :Version: 1.5.1
 
 .. contents::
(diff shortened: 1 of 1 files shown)


source/change-streams/change-streams.rst
```diff
@@ -61,9 +61,6 @@ An error is considered resumable if it meets any of the following criteria:
     * - CursorKilled
       - 237
 
-- a server error response with an error message containing the substring "not
-  master" or "node is recovering"
```
jmikola: @daprahamian: I do have one more question before a lead signs off on this. A few lines below, we have:

> An error on an aggregate command is not a resumable error. Only errors on a getMore command may be considered resumable errors.

If I understand correctly, we _never_ resume under _any_ circumstance for an error encountered during the execution of an `aggregate` command. This includes the first `aggregate`. Any subsequent `aggregate` commands are naturally included because those would only be running as part of a resume action (and we only resume once).

As such, I'm quite confused by the following prose test:

6. ``ChangeStream`` will not attempt to resume on a server error

That seems incorrect, so I've opened https://github.com/mongodb/specifications/pull/366 to revise it.

daprahamian: I agree. That should probably be changed to "`ChangeStream` will not attempt to resume if the initial aggregate request fails"

PR #366, opened 2018-06-26
Correct prose test for not resuming during an aggregate

jmikola:
This test appears to date back to the original spec commit in 9784d5d2fd52f36b15ff41d262ab171d3968ab0a; however, it does not jive with the spec's current language, which states that errors on aggregate commands are never to be retried. Only errors from getMore commands may be considered resumable.

Adding @mbroadst since this was his original commit and @kevinAlbs since he recently worked on error logic in https://github.com/mongodb/specifications/commit/7021f759c4c2a918808660a2e4e55ba1f4e4acb5.


kevinAlbs:
In the the original spec commit a resumable error was considered:

> Any error encountered which is not a server error, with the exception of a “not master” server response. An example might be a timeout error, or network error.

I think the intent behind this prose test was to ensure we don't resume on non-resumable errors.

I think we should test both:
- we don't resume non-resumable errors (the server errors that are blacklisted)
- we never resume on aggregate

PR #368, opened 2018-07-02
Clarify use of nameOnly in collection enumeration

xdg:
nameOnly doesn't allow filters on anything other than `name`.


xdg:
@behackett, see Jeremy's question above.  I'm indifferent; I just wanted to document the problem I hit.

PR #371, opened 2018-07-10
Clarify difference between Iterables returned by read methods

jmikola:
@rathisekaran: This is a small documentation change, so I don't think it warrants a version bump or changelog entry.

### DIFF
--- a/source/crud/crud.rst
+++ b/source/crud/crud.rst
@@ -137,6 +137,8 @@ Read
      * MAY setup a cursor to be executed upon iteration against the $out collection such
      * that if a user were to iterate a pipeline including $out, results would be returned.
      *
+     * Note: implementations should return a cursor for Iterable<Document>.
+     *
      * @see https://docs.mongodb.com/manual/reference/command/aggregate/
      */
     aggregate(pipeline: Document[], options: Optional<AggregateOptions>): Iterable<Document>;
@@ -168,6 +170,10 @@ Read
     /**
      * Finds the distinct values for a specified field across a single collection.
      *
+     * Note: Unlike other read methods (e.g. aggregate, find), this method will not return
+     * a cursor. The Iterable<any> return type corresponds to the "values" array in the
+     * distinct command's result document.
+     *
      * @see https://docs.mongodb.com/manual/reference/command/distinct/
      */
     distinct(fieldName: string, filter: Document, options: Optional<DistinctOptions>): Iterable<any>;
@@ -182,6 +188,8 @@ Read
      *
      * Note: If $explain is specified in the modifiers, the return value is a single
      * document. This could cause problems for static languages using strongly typed entities.
+     *
+     * Note: implementations should return a cursor for Iterable<Document>.
      *
      * @see https://docs.mongodb.com/manual/core/read-operations-introduction/
      */


mbroadst:
I'm okay with this, but it feels like there was some effort made to avoid using the term "Cursor" in this spec. I'd like to defer to someone else who has more feeling about what's correct here. For instance, why don't we simply say that `distinct` returns `Array<any>` vs `Iterable<any>`. Alternatively, it seems that `Iterable<Document>` == `Cursor`, whereas `Iterable<any>` == some collection of any types.

jyemin:
Do we need to point out that distinct doesn't return a cursor, that it's different from find/aggregate?  From an API perspective, a driver should be able to handle a server change that made distinct return a cursor, similar to what happened with aggregate back in 2.6?  Until then, it's just a cursor that happens to only ever return a single batch.

jmikola:
> Do we need to point out that distinct doesn't return a cursor, that it's different from find/aggregate?

I tried to avoid mention of whether the `distinct` command itself returned a command cursor, although I couldn't help but refer to the `values` array in its result document. Frankly, I wouldn't expect the server to upgrade the `distinct` command in the future as we can already implement a more robust solution using the aggregation pipeline and a `$group` stage (using the examined values as our bucket IDs).

> For instance, why don't we simply say that distinct returns `Array<any>` vs `Iterable<any>`. Alternatively, it seems that `Iterable<Document>` == `Cursor`, whereas `Iterable<any>` == some collection of any types.

For all the lengths the spec goes to not to mention a cursor, it is an unspoken rule that `Iterable<Document>` is a cursor, or at least some iterable object that composes a cursor. I was hesitant to use `Array<any>` for the same reason as not wanting to get into the specifics about `distinct`'s cursor behavior, since that implies we have the entire result set available in one go and we're suggesting a specific type. An iterable is already defined in the spec's terms and also allows for flexibility if results need to be streamed (e.g. cursors issuing `getMore` requests).

> I'm okay with this, but it feels like there was some effort made to avoid using the term "Cursor" in this spec.

Other doc blocks referred to cursor when discussing particular options, so I'd like to think the notes on `aggregate()` and `find()` are reasonable and can stay. It's pretty lightly worded but clarifies something that should be obvious to anyone implementing those methods.

That said, if everyone is in favor of removing the "note" I added to `distinct()`'s doc block and just changing its return type to `Array<any>`, we can do that. If so, I'd prefer we not add a new term for "Array".

jmikola:
@mbroadst, @jyemin: Can you follow up on my last comment?

ajdavis:
@jyemin it seems that we're awaiting your input here.

source/crud/crud.rst
```diff
@@ -137,6 +137,8 @@ Read
      * MAY setup a cursor to be executed upon iteration against the $out collection such
      * that if a user were to iterate a pipeline including $out, results would be returned.
      *
+     * Note: implementations should return a cursor for Iterable<Document>.
```
jyemin: If we're talking about something like Java's Iterable, which is defined as

    public interface Iterable<T> {
        Iterator<T> iterator();
    }

then an Iterable can be iterated multiple times, so it can't also be a cursor.  Rather, it's an object that can produce an Iterator, and it's each Iterator that is actually backed by a cursor.  Java and .NET drivers override this iterator() method to return, effectively, a CloseableIterator, so that applications can ensure that the cursor is closed even when the Iterator is not exhausted.

We have two options here:

1. Change the note to say something like "Iterable<Document> instances returned by this method should produce iterators that are backed by a cursor"
2. Change the return type to Iterator<Document>.  Whether a driver produces Iterables that in turn produce Iterators can be treated as an acceptable deviation based on language idioms.

In both cases the Iterator must have some way to be closed (e.g. in Java by implementing the AutoCloseable interface)

I also think that distinct should say the exact same thing.  The fact that distinct currently does not return a cursor is an implementation detail that may change.  We got burned by this a while ago when aggregate started returning a cursor, and I can see distinct going the same way, say by being re-implemented as an aggregation pipeline.  But even if that doesn't happen there is no reason an application need be aware of the difference.

jmikola: I prefer the language stating that the iterator should be backed by a cursor.

I think it's more flexible to avoid referring to Iterator, since that's going to be a type in most languages and the CRUD spec already defines "Iterable" in its terms for this purpose.

I would change this to:

```
Note: the Iterable<Document> return value should be backed by a cursor.
```

source/crud/crud.rst
```diff
@@ -168,6 +170,10 @@ Read
     /**
      * Finds the distinct values for a specified field across a single collection.
      *
+     * Note: Unlike other read methods (e.g. aggregate, find), this method will not return
+     * a cursor. The Iterable<any> return type corresponds to the "values" array in the
+     * distinct command's result document.
```
jmikola: @jyemin: Here, would you prefer I add something to the effect of:

```
Note: As of MongoDB 4.0, the Iterable<any> return type corresponds to the "values"
array in the distinct command's result document. This may change in the future if the
result becomes backed by a cursor (similar to aggregate and find).
```

source/crud/crud.rst
```diff
@@ -183,6 +189,8 @@ Read
      * Note: If $explain is specified in the modifiers, the return value is a single
      * document. This could cause problems for static languages using strongly typed entities.
      *
+     * Note: implementations should return a cursor for Iterable<Document>.
```
jmikola: I would change this to:

```
Note: the Iterable<Document> return value should be backed by a cursor.
```

jmikola:
@jyemin: Please take another look.

PR #374, opened 2018-07-24
Document that with an empty collection, the count A/F pipeline result is empty

derickr:
We ran into this in the PHP Library [PHPLIB-376|https://jira.mongodb.org/browse/PHPLIB-376] and it was reported that the Node driver had a similar issue.


source/crud/crud.rst
```diff
@@ -619,6 +619,10 @@ The count of documents is returned in the 'n' field, similar to the `count`
 command. countDocuments options other than filter, skip, and limit are added as
 options to the `aggregate` command.
 
+If the collection is empty, the pipeline returns is an empty array, and hence
```
mbroadst: there is an extra `is` here. I would suggest rewriting as:
```
In the event this aggregation is run against an empty collection, an empty array will be returned with no ``n`` field. Drivers MUST interpret this result as a ``0`` count.
```

daprahamian:
Can we add a spec test for this?

derickr:
@daprahamian — certainly, I'll go add one.

derickr:
@daprahamian https://github.com/mongodb/specifications/pull/375

PR #375, opened 2018-07-25
Add tests for count methods with empty collection

derickr:


derickr:
I should have added this commit to #374 — my bad.

PR #376, opened 2018-07-25
SPEC-1135: Add upsertedCount to UpdateResult

jmikola:
https://jira.mongodb.org/browse/SPEC-1135


jmikola:
It looks like `upsertedCount` was already in most spec tests but just missing from the UpdateOne and UpdateMany arrayFilters tests.

PR #377, opened 2018-07-26
SPEC-1136: Add insertedCount to BulkWriteResult

jmikola:
https://jira.mongodb.org/browse/SPEC-1136

`insertedCount` is a required BulkWriteResult field, so there's no reason we shouldn't include it in the expectation.

### DIFF
--- a/source/crud/tests/write/bulkWrite-arrayFilters.json
+++ b/source/crud/tests/write/bulkWrite-arrayFilters.json
@@ -28,48 +28,57 @@
     {
       "description": "BulkWrite with arrayFilters",
       "operation": {
+        "name": "bulkWrite",
         "arguments": {
-          "options": {
-            "ordered": true
-          },
           "requests": [
             {
+              "name": "updateOne",
               "arguments": {
+                "filter": {},
+                "update": {
+                  "$set": {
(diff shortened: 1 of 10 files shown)


ShaneHarvey:
Looks like I screwed up the JSON formatting when I added this file... LGTM. Should we also add insertedCount to all the other uses of bulkWrite?

jmikola:
> all the other uses of bulkWrite

AFAIK, this is the only test for `bulkWrite()` pending [SPEC-870](https://jira.mongodb.org/browse/SPEC-870).

ShaneHarvey:
See:
- https://github.com/mongodb/specifications/blob/master/source/transactions/tests/bulk.yml#L87
- https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/bulkWrite-serverErrors.yml
- https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/bulkWrite.yml
- https://github.com/mongodb/specifications/blob/master/source/transactions/tests/transaction-options.yml#L630

jmikola:
Updated and tested against PHPLIB (with minor changes to our test harness). Please review when you get a chance (and feel free to test Python as well).

PR #382, opened 2018-08-02
SPEC-1145 Update auth spec for ismaster not erroring

xdg:
When the auth spec was developed, the server would give an ismaster
error for unknown users.  That was fixed in SERVER-34421 before the 4.0
GA.  This commit removes language that implies the server will error.

### DIFF
--- a/source/auth/auth.rst
+++ b/source/auth/auth.rst
@@ -6,7 +6,7 @@ Driver Authentication
 =====================
 
 :Spec: 100
-:Spec Version: 1.8
+:Spec Version: 1.7.1
 :Title: Driver Authentication
 :Author: Craig Wilson, David Golden
 :Advisors: Andy Schwerin, Bernie Hacket, Jeff Yemin, David Golden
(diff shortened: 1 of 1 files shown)


source/auth/auth.rst
```diff
@@ -6,7 +6,7 @@ Driver Authentication
 =====================
 
 :Spec: 100
-:Spec Version: 1.7
+:Spec Version: 1.8
```
jmikola: Since [SPEC-1145](https://jira.mongodb.org/browse/SPEC-1145) is marked as not needing driver changes, I think this should either remain 1.7 or be bumped to 1.7.1. That jives with the discussion we had in [SPEC-1139](https://jira.mongodb.org/projects/SPEC/issues/SPEC-1139) about proposals for spec versioning. To repeat your quote of the current guidelines on our internal wiki:

> * Major change: First digit: API Change, New Feature (i.e. two parts of the spec fight with each other and one of them is resolved). These require a ‘Major’ DRIVERS ticket.
 > * Minor change: Second digit: Test Change, Bug fix (i.e. pseudo-code written incorrectly). These require a ‘Minor’ DRIVERS ticket.
> * Trivial refinements and clarifications with no action required by any driver author do not warrant a version bump or DRIVERS ticket.

Without needing driver changes, this seems like the last case.

xdg:
Rebased and merged as 4f212acb5154a3c3d124525ad1f0f308efa93ef8

PR #385, opened 2018-08-06
Connection string: test comma separated value to array

rweinberger:
Add a test to check that comma-separated option values in the connection string are correctly parsed to an array ([example use case: compressors](https://github.com/mongodb/specifications/blob/master/source/compression/OP_COMPRESSED.rst#compressors))


mbroadst:
@xdg ping

xdg:
I don't think that is mandated by the spec generally, even if it happens to be true for compression as a specific key.  IIUC, the spec only talks about lists for repeated keys: `foo=bar&foo=baz`.

I don't think we should merge this change without also addressing the spec.  I haven't really thought through how well it plays with key-value pairs, which do define the use of a comma as a key-value pair delimiter.

What is this: `foo=bar:baz,fizz`?  Is that a list of `bar:baz` and `fizz`, or is that a bad key-value pair map because `fizz` is missing a trailing colon/value?

What is `foo=bar,,baz`?  Is that a 3-element list with an empty middle value?  Or is the missing value elided from the list?

The connection spec is generally context free and I fear that adding another meaning to comma requires parsing to be context aware in ways that haven't been fully explored.

mbroadst:
I see your rationale here, but if we are going to be this restrictive then I propose we apply it to all cases.  Shall we remove the section for "Key value pairs"? We similarly do not have tests for these value types at all, what does `foo=bar:baz,,fiz:buzz` translate to? What about `foo=bar:,baz`?

It often feels like we go to great lengths to make this specification context-free at potentially great cost to future implementors. The node driver implemented the new connection string parser to spec, and ran into this very real error.

I propose one of the following:
 - change the compression ([OP_COMPRESSED](https://github.com/mongodb/specifications/blob/master/source/compression/OP_COMPRESSED.rst)) spec to use the documented form: `mongodb://localhost/?compressors=snappy&compressors=zlib`
 - add some limited text to the connection string spec indicating that this form is supported, and accept this test

xdg:
I think compression defines rules for what to do with its string (e.g. split on commas).  All connection spec needs to do is ensure that we get properly delimited strings.

xdg:
Sort of.  It has integration tests for `mongodb://localhost:27017/?compressors=snappy,zlib` and `mongodb://localhost:27017/?compressors=zlib,snappy`.

I propose we close this issue and open a spec ticket to add connection string unit tests to the compression spec.  (E.g. addressing how things like `zlib, snappy` (with embedded space) is parsed -- which this PR doesn't even do.)

PR #387, opened 2018-08-20
feat(sdam): add test for single connection to a sharded cluster

mbroadst:
/cc @ShaneHarvey I believe the python driver doesn't support this presently.


mbroadst:
Closing due to enough interest to the contrary 😄

ShaneHarvey:
For others following along, we actually test the opposite of this in https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/tests/single/direct_connection_mongos.yml

PR #388, opened 2018-08-21
Do not require unknown topology type when user specified connection t…

p-mongo:
…o replica set.

https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#initial-topologytype
specifies the following for initial topology type:

> The user MAY be able to initialize it to ReplicaSetNoPrimary.
> This provides the user a way to tell the client it can only connect
> to replica set members. Similarly the user MAY be able to
> initialize it to Sharded, to connect only to mongoses.

Ruby driver does offer users the option to force a topology type via
:connect and :replica_set options, and this test gives an explicit
replica set name which puts the driver into forced topology mode
rather than discovering the topology from the cluster. In this mode
the Ruby driver starts out in Replica Set topology rather than Unknown.

### DIFF
--- a/source/server-discovery-and-monitoring/tests/monitoring/required_replica_set.json
+++ b/source/server-discovery-and-monitoring/tests/monitoring/required_replica_set.json
@@ -32,7 +32,7 @@
             "topology_description_changed_event": {
               "topologyId": "42",
               "previousDescription": {
-                "topologyType": "Unknown",
+                "topologyType": null,
                 "servers": []
               },
               "newDescription": {
--- a/source/server-discovery-and-monitoring/tests/monitoring/required_replica_set.yml
+++ b/source/server-discovery-and-monitoring/tests/monitoring/required_replica_set.yml
@@ -24,7 +24,8 @@ phases:
           topology_description_changed_event:
             topologyId: "42"
             previousDescription:
-              topologyType: "Unknown"
+              # Can be "Unknown" or "ReplicaSetNoPrimary"
+              topologyType:
               servers: []
             newDescription:
               topologyType: "ReplicaSetNoPrimary"


p-mongo:
Suggested by @mbroadst

p-mongo:
Matt said `null` means do not assert anything and is already in the specs, but this does not appear to be the case. What are your thoughts on making topology type be either a string or a list of strings?

ShaneHarvey:
I see think we have three options:
1. `topologyType: null` which means the actual topologyType may be anything.
1. `topologyType: ["Unknown", "ReplicaSetNoPrimary"]` (as you suggest) which means the actual topologyType may be either Unknown or ReplicaSetNoPrimary.
1. `topologyType: "DefaultForReplicaSet"` which means the actual topologyType MUST be the driver's default replica set topologyType.

I think option 3 is the best way to go since it ensures that each driver does what it is expected to do. Options 1 and 2 leave room for bugs like  topologyType being ReplicaSetNoPrimary when it should be Unknown or vice versa.

p-mongo:
"DefaultForReplicaSet" isn't really a thing in Ruby driver. If a user specifies a list of hosts only, we start out in Unknown. If a user specifies a RS name we start out in ReplicaSetNoPrimary. I'm not sure if you meant a driver-wide default or the initial value for this particular test.

ShaneHarvey:
By "DefaultForReplicaSet" I mean the default for this specific scenario (user gives an replica set name).  Maybe "DefaultWhenProvidedReplicaSetName" is clearer. The important part is that it corresponds to a single topology type for each driver.

mbroadst:
While it's not explicitly defined in the test format for SDAM tests, I was alluding to general rules we have been gravitating towards in the [command monitoring](https://github.com/mongodb/specifications/tree/master/source/command-monitoring/tests#fake-placeholder-values) and [transactions](https://github.com/mongodb/specifications/tree/master/source/transactions/tests#null-values) tests, as well as others. A value of `null` is generally understood to mean "omit this from inspection" during your assertion phase, and I think that properly conveys the intent of this test and the spec in general. The spec claims you MAY do whatever you want, so let's not check it.

ajdavis:
Hi, sorry for the delayed review. Let's not make this change. Most or all drivers have implemented SDAM Monitoring and they pass this test, which means they have logic something like this in the MongoClient initialization:

- Create an empty topology with topology type Unknown.
- Publish topology_opening_event.
- Parse the URI. If it includes "replicaSet=", set topology type to ReplicaSetNoPrimary.
- Publish topology_description_changed_event, with previous and current topologies.

Regardless of how each driver passes this test, they all do. That's a very good thing because it means that all drivers publish a consistent sequence of SDAM Monitoring events, and consistency is the main purpose of specs and spec tests.

Had the Ruby Driver ever passed this test before?

mbroadst:
@ajdavis, if we're going to take that path then I propose that we add language to the SDAM spec indicating that all TopologyDescription's MUST begin in an `Unknown` state, even when initialized with a particular state.

ajdavis:
Thanks, I agree.

p-mongo:
Ruby driver has two features/differences from other drivers (at least Python/shell):

1. By default, Ruby driver performs topology discovery, whereas python/shell perform a direct connection.
2. Ruby driver allows its user to force a topology (to any one of the 3).

The test in question would pass IF there was no replica set name given in the URI. In Ruby the replica set name is not needed there as the driver will discover topology from a single seed by default unless the user requested a specific connection type and thus forced topology. Specifying replica set name in the URI in other drivers may put them in auto discovery mode, however since Ruby driver is in auto discovery mode by default having a replica set name in the URI puts it into forced topology mode.

Thus the easiest way for the Ruby driver to pass this test is to modify the URI, which we already do in other tests because the semantics of a URI including an address and no connection options are different for Ruby driver vs others.

From my conversation with Matt I gather than Node driver does not offer users the option of forcing topology, thus I wonder which other drivers have this functionality and are thus closer to Ruby driver in their operation.

Starting with an unknown topology always: how is this going to work with the single initial topology?

> The user MUST be able to set the initial TopologyType to Single.

My previous attempts at making the test in question pass involved making the initial topology as exposed in SDAM events unknown, and this is awkward and convoluted.

The difference between unknown and RS as initial topology is an unknown topology can transition to anything whereas RS cannot. If a Ruby driver user specified RS name in URI and tries to connect to a sharded cluster, the driver will ignore and drop every single cluster member because none of them match the forced topology. When you are talking about starting in unknown topology, do you mean for the driver to maintain this forced topology behavior and have two sets of topology descriptions (one internal that is the "real" one and one fake one that is exposed in SDAM events), or are you intending a behavior change that essentially removes forcing topologies as a feature?

ajdavis:
Yeah, you can think of my request as an initial "fake" topology description that is sent as topology_description_changed_event.previousDescription. The initial topology description's type is always Unknown, no matter what is in the URI or what options are passed to MongoClient.

Another way to think of it is this: create the topology description with type Unknown before parsing the URI and before handling any options passed to MongoClient. This is the previousDescription. Then, parse the URI and handle options, the same way the Ruby Driver behaves now. That creates newDescription. Publish the topology_description_changed_event.

ajdavis:
Thanks!

jmikola:
> While it's not explicitly defined in the test format for SDAM tests, I was alluding to general rules we have been gravitating towards in the command monitoring and transactions tests, as well as others. A value of null is generally understood to mean "omit this from inspection" during your assertion phase, and I think that properly conveys the intent of this test and the spec in general. The spec claims you MAY do whatever you want, so let's not check it.

Note that "generally understood" may change with [SPEC-1143](https://jira.mongodb.org/browse/SPEC-1143), at least as far as the CRUD spec is concerned, so that we can assert that `null` is a respected value for inserted/upserted `_id` values.

PR #391, opened 2018-09-06
Update tests for read concern levels

thejonathanma:
Specification tests were missing a few read concern levels. This commit updates document tests to include linearizable, snapshot, and available. It also updates connection-string tests to include linearizable and available.


jmikola:
I think this warrants a [SPEC ticket](http://jira.mongodb.org/browse/SPEC), so @rathisekaran can later track the change with a DRIVERS ticket and request all drivers to resync their tests. Since all of the changes pertain to one spec (read/write concern), it should be easy to track in a single SPEC ticket using that component.

PR #392, opened 2018-09-06
Add a test for SDAM when secondary with different RS name responds first

p-mongo:


p-mongo:
@ajdavis Ruby already changes connection strings in spec tests for this reason. I can modify the connection strings in this PR to be generally suitable for the majority of drivers(?) and have Ruby be different.

ajdavis:
I was thinking about that. The suitable connection string for all drivers would be "mongodb://a,b/?replicaSet=set-b". If we use that connection string then all drivers would pass the test.

However, you're no longer testing the thing you intended to test: the Ruby driver's rules for discovering the replica set name from an isMaster response. Instead, you'd test all drivers' rules for rejecting a replica set member whose setName doesn't match the "replicaSet=" option in the URI. That's already tested in wrong_set_name.yml, secondary_wrong_set_name.yml, and secondary_wrong_set_name_with_primary.yml. None of those tests is exactly the same as this scenario, but they're pretty close.

I think that the test you wrote is valuable as it is. But it's valuable *only* for drivers that behave like the Ruby Driver: they discover the replica set name from the first isMaster response if they're provided multiple addresses in the URI and no "replicaSet=" option. That's an important test and it only works for some drivers, so let's not put it in the specifications repository.

p-mongo:
Alright then.

PR #399, opened 2018-10-02
SPEC-1142: Test continue-on-error behavior with unordered insertMany() and bulkWrite()

jmikola:
https://jira.mongodb.org/browse/SPEC-1142

These are the first CRUD tests to trigger an error (and use ordered=false, if that matters). Some test harnesses may need revisions to detect the error and possibly extract a write result from the error for additional assertions.

I had to update PHPLIB's test runner and am currently in the process of updating Swift's.

### DIFF
--- a/source/crud/tests/write/bulkWrite.yml
+++ b/source/crud/tests/write/bulkWrite.yml
@@ -332,3 +332,79 @@ tests:
                 data:
                     - {_id: 2, x: 23 }
                     - {_id: 3, x: 33 }
+    -
+        description: "BulkWrite continue-on-error behavior with unordered (preexisting duplicate key)"
+        operation:
+            name: "bulkWrite"
+            arguments:
+                requests:
+                    -
+                        name: "insertOne"
+                        arguments:
+                            document: { _id: 2, x: 22 }
+                    -
+                        name: "insertOne"
+                        arguments:
+                            document: { _id: 3, x: 33 }
+                    -
+                        name: "insertOne"
+                        arguments:
+                            document: { _id: 4, x: 44 }
+                options: { ordered: false }
+        outcome:
+            error: true
+            result:
+                deletedCount: 0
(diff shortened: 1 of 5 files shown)


source/crud/tests/write/bulkWrite.yml
```diff
@@ -332,3 +332,41 @@ tests:
...
+                        name: "insertOne"
+                        arguments:
+                            document: { _id: 3, x: 33 }
+                    -
+                        name: "insertOne"
+                        arguments:
+                            document: { _id: 4, x: 44 }
+                options: { ordered: false }
+        outcome:
+            error: true
```
jmikola: I just realized that [tests/README.rst](https://github.com/mongodb/specifications/blob/master/source/crud/tests/README.rst) would benefit from having this defined. I'll make that change and incorporate it into this PR soon.

jmikola:
@saghm: I needed to use different inserts in each test because the bulkWrite and insertMany tests each have different fixture data (insertMany only has one document), but the variation you pointed out was not purposeful. I personally don't see any difference between the tests, but I'm happy to split them up so that bulkWrite and insertMany each test both scenarios (i.e. 4 new tests total).

source/crud/tests/write/bulkWrite.yml
jmikola: I've updated the test format descriptions in a later commit within this PR.

jmikola:
I split the tests to cover duplicate IDs in the existing collection data or just within the `requests` argument. I also updated the test format description (borrowing language from retryable writes). The `error` field is not documented.

These tests pass both pass on PHPLIB, although I'm still working on amending Swift's test runner to support `error`.

jmikola:
The Swift driver now passes these tests with https://github.com/mongodb/mongo-swift-driver/pull/122. I think this is ready for a final look.

PR #400, opened 2018-10-02
SPEC-1160 Test delete/updateMany with retryWrites

ajdavis:
These CRUD operations are not retryable, so they should ignore
retryWrites=true in the URI and not set the txnId.

### DIFF
--- a/source/retryable-writes/tests/README.rst
+++ b/source/retryable-writes/tests/README.rst
@@ -24,7 +24,7 @@ the cluster.
 Server Fail Point
 =================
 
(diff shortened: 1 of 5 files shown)


ajdavis:
I've updated the README, @ShaneHarvey .

@daprahamian , bulkWrite.yml includes some tests with deleteMany and updateMany. Those tests aren't explicitly written to ensure that deleteMany and updateMany work with retryWrites=true, but they'd fail if those two WriteModels didn't work with retryWrites.

PR #401, opened 2018-10-04
SPEC-1161 Test that txn reads don't inherit readConcern

ajdavis:
The Transactions Spec says, "Drivers MUST override all other collection,
database, or client readConcerns with the transaction’s readConcern.
Drivers MUST add this readConcern to the first command in a transaction
if and only if the readConcern is supplied and not the default."

I believe this applies when a transaction has *no* readConcern as well:
drivers ignore collection, database, or client readConcern and use no
readConcern in this case.

When I synced the tests for SPEC-1121, I found a new libmongoc bug, CDRIVER-2841. We *do* obey the rule, "Only send readConcern in the first transaction command." But it's the wrong readConcern: it should be the transaction's readConcern, not the collection's. This test further proves that drivers follow the readConcern rules in transactions.

### DIFF
--- a/source/transactions/transactions.rst
+++ b/source/transactions/transactions.rst
@@ -220,14 +220,7 @@ is NULL or the readConcern in defaultTransactionOptions is NULL, the
 readConcern is inherited from the MongoClient associated with this
 session.
 
-Drivers MUST override all other collection, database, or client
-readConcerns with the transaction’s readConcern. Drivers MUST add this
-readConcern to the first command in a transaction if and only if the
-readConcern is supplied and not the default. Drivers MUST NOT add to
-subsequent commands the readConcern from the transaction or any
-readConcern inherited from the collection, database, or client.
-
-If the user supplies an explicit readConcern via a method option, the driver 
+If the user supplies an explicit readConcern via a method option, the driver
 MUST raise an error with the message "Cannot set read concern after starting a
 transaction."
 See `Users cannot pass readConcern or writeConcern to operations in transactions`_.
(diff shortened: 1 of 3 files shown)


mbroadst:
I propose we change the language of the spec in that case, its very easy to read that statement as: "Drivers MUST override all other collection, database, or client readConcerns with the transaction’s readConcern **if present**" (my emphasis), which is how I think we got here, correct?

ajdavis:
@mbroadst I moved the readConcern prose to a different section and made it clearer.

mbroadst:
@ajdavis I think that's much clearer, thank you!

source/transactions/transactions.rst
```diff
@@ -220,14 +220,7 @@ is NULL or the readConcern in defaultTransactionOptions is NULL, the
 readConcern is inherited from the MongoClient associated with this
```
kevinAlbs: Based on the above, we might copy the readConcern from the client, but only when starting the transaction.

When we run the first read operation, we'll use whatever read concern that was copied into the TransactionOptions.

This is the behavior we expect, right?

source/transactions/transactions.rst
```diff
@@ -580,9 +573,19 @@ Constructing the first command within a transaction
...
 
 When constructing the first command within a transaction, drivers MUST
-add the ``lsid``, ``txnNumber``, ``readConcern``, ``startTransaction`` and
-``autocommit`` fields. This is an example of an insert command that
-begins a server transaction:
+add the ``lsid``, ``txnNumber``, ``startTransaction``, and ``autocommit``
+fields.
+
+Drivers MUST include the transaction's readConcern in the first command in a
+transaction if and only if the readConcern is supplied and not the default. The
```
ShaneHarvey: Can this say "not the server's default" so it's clear what "default" we're referencing?

source/transactions/transactions.rst
ajdavis: That's right. I'll make that super-clear in " Constructing the first command within a transaction"

ajdavis:
New patch, a few words to address Shane's and Kevin's comments about transactions.rst

ajdavis:
@mbroadst look good to you now?

PR #402, opened 2018-10-11
SPEC-1163 require session pool clearing after fork

xdg:


source/sessions/driver-sessions.rst
```diff
@@ -830,6 +830,13 @@ of the cursor.  For language runtimes that provide the ability to attach finaliz
 that are run prior to garbage collection, the cursor class SHOULD return an implicit session
 to the pool in the finalizer if the cursor has not already been exhausted.
 
+If a driver supports process forking, the session pool needs to be cleared on
+one side of the forked processes (just like sockets need to reconnect).
+Drivers MUST provide a way to clear the session pool without sending
+``endSessions``.  Drivers MAY make this automatic when the process ID changes.
+If they do not, they MUST document how to clear the session pool wherever they
+document fork support.
```
ajdavis: Let's name this API so drivers that provide it use a consistent name, and let's add a prose test to tests/README.

xdg: We don't define any API for session pools and many (most?) drivers may not expose them to end-users.  E.g. the Perl driver provides a `reconnect` method on MongoClient that is documented as the thing to call after forking.  Inside that method, the driver reconnects sockets and now clears the pool.

source/sessions/driver-sessions.rst
```diff
@@ -1069,6 +1069,8 @@ ensure that they close any explicit client sessions and any unexhausted cursors.
     * Iterate through enough documents (3) to force a ``getMore``
     * Assert that the server receives a non-zero lsid equal to the lsid that ``find`` sent.
 
+11. For drivers that support forking, test that the session pool can be cleared after a fork
+    without calling ``endSession``.
```
ajdavis: E.g., create ClientSession, record its lsid, delete it (so the lsid is pushed into the pool). Fork. In the parent, create a ClientSession and assert its lsid is the same. In the child, create a ClientSession and assert its lsid is different.

ShaneHarvey:
What should we say about ClientSessions that are checked out of the pool at the time of the fork?

Perhaps it's worth it for the child process to discard the ServerSession once the ClientSession is closed:
```
s = client.start_session()
fork()
s.end_session() # In parent process the server session goes back into the pool. In the child, it gets discarded.
```

In terms of using the session in both processes I think drivers should do no special detection and provide a rationale similar to "Why do we say drivers MUST NOT attempt to detect unsafe multi-threaded use of ClientSession?"

xdg:
Squashed and merged as 6a31988385f54e4ec9daddfbf21344a2a96e2656

PR #403, opened 2018-10-11
Clarify that Timestamp and Counter are big endian fields

derickr:


mbroadst:
ping @jmikola, we just had a PR for this from the community in #404 (Joannis from MongoKitten)

PR #404, opened 2018-10-18
Explicitly add endianness to the timestamp definition

Joannis:
This added some confusion in MongoKitten and wasn't found out until diving into MongoDB source code


mbroadst:
Thanks @Joannis, I'm closing this as it duplicates the previous PR #403

PR #405, opened 2018-10-19
Add tests for RS primary becoming ghost and mongos

p-mongo:
This is in addition to the existing test for primary becoming standaone.


p-mongo:
This was implemented in Ruby driver as part of https://jira.mongodb.org/browse/RUBY-1492.

ajdavis:
Looks good, and I bet it covers some untested code paths in a few drivers.

PR #406, opened 2018-10-29
Handling for mismatched RS name should be the same as me mismatch

p-mongo:
Which is to stop processing & possibly transition to no primary if affected server was a primary.

### DIFF
--- a/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
+++ b/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
@@ -1108,6 +1108,8 @@ updateRSWithPrimaryFromMember
     # SetName is never null here.
     if topologyDescription.setName != description.setName:
         remove this server from topologyDescription and stop monitoring it
+        checkIfHasPrimary()
+        return
 
     if description.address != description.me:
         remove this server from topologyDescription and stop monitoring it


ShaneHarvey:
Is it possible to add a test for this or do you know if one already exists?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1108,6 +1108,8 @@ updateRSWithPrimaryFromMember
     # SetName is never null here.
     if topologyDescription.setName != description.setName:
         remove this server from topologyDescription and stop monitoring it
+        checkIfHasPrimary()
+        return
```
ajdavis: Good, we definitely need to return here. I think we don't need checkIfHasPrimary, however - we are processing a non-primary ServerDescription, so there must still be a primary server in the TopologyDescription, right? For the same reason, my checkIfHasPrimary() call below on line 1116 is unnecessary (but harmless) as well.

p-mongo: How would checkIfHasPrimary get invoked in this scenario?

p-mongo:
@ShaneHarvey I believe I ran into test failures without both changes in this PR (check primary & return) but I will double check the spec tests.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
ajdavis: Good point! Yes we need checkIfHasPrimary here. The pseudocode change looks good to me. So, does this already have a YAML test?

p-mongo: If neither checkIfHasPrimary nor return are here, all spec tests tests pass. This is because the logic at the botom of update_rs_with_primary_from_member is, essentially, checkIfHasPrimary. However that flow is hard to understand because if there is a set name mismatch, we fall through to me mismatch check after having supposedly removed the server from the topology and stopped monitoring it.

Removing checkIfHasPrimary but keeping return fails rs/stepdown_change_set_name.yml.

Removing return invokes the fall-through behavior which makes all tests pass but again is difficult to understand.

What would you like me to do?

PR #407, opened 2018-10-29
Specify when topology type unknown becomes RS with primary

p-mongo:
I couldn't find where the spec said to change topology type to RS with primary. This point is explicitly defined for other topology type transitions.

There are two ways of doing this - either conditionally in updateRSFromPrimary based on whether the topology is already ReplicaSetWithPrimary, or unconditionally in two places where we go from another topology to ReplicaSetWithPrimary. This PR proposes the second route.

### DIFF
--- a/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
+++ b/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
@@ -957,11 +957,11 @@ with the new one.
   ,TopologyType Unknown,TopologyType Sharded,TopologyType ReplicaSetNoPrimary,TopologyType ReplicaSetWithPrimary
   ServerType Unknown,no-op,no-op,no-op,`checkIfHasPrimary`_
   ServerType Standalone,`updateUnknownWithStandalone`_,`remove`_,`remove`_,`remove`_ and `checkIfHasPrimary`_
-  ServerType Mongos,Set type to TopologyType Sharded,no-op,`remove`_,`remove`_ and `checkIfHasPrimary`_
-  ServerType RSPrimary,`updateRSFromPrimary`_,`remove`_, `updateRSFromPrimary`_,`updateRSFromPrimary`_
-  ServerType RSSecondary,Set type to ReplicaSetNoPrimary then `updateRSWithoutPrimary`_,`remove`_,`updateRSWithoutPrimary`_,`updateRSWithPrimaryFromMember`_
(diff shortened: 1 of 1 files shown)


ajdavis:
I had "factored" this pseudocode by defining a subroutine called "checkIfHasPrimary". The existing pseudocode ensures that checkIfHasPrimary is only run once other checks have complete: for example, the new primary has the correct setName and maxWireVersion, etc. So let's not make this change, it's redundant with the checkIfHasPrimary definition.

Should we make some other change so that checkIfHasPrimary is more clearly explained here?

p-mongo:
Okay I see. `checkIfHasPrimary` to me meant "in a RS topology, see if there's a primary". The code I wrote required topology be of an RS type when checkIfHasPrimary is invoked. I did not interpret checkIfHasPrimary to mean "transition topology from any type to RS with primary or RS without primary based on whether there is a primary".

p-mongo:
Consider going from unknown topology to a primary one. The way the spec is currently written, a subset of what is supposed to happen is:

<pre>
(on unknown topology:)
if topologyDescription.setName is null:
    topologyDescription.setName = description.setName

... some time later ...

(starting with unknown topology:)
Set TopologyType to ReplicaSetWithPrimary if there is an RSPrimary in TopologyDescription.servers, otherwise set it to ReplicaSetNoPrimary.
</pre>

I added a check to Unknown topology in Ruby driver to fail construction if it is given an RS name (or other RS options). With this check, an unknown topology cannot have an RS name and it is impossible to implement the flow of the spec as stated above.

Ruby driver as currently implemented first creates a RSWithoutPrimary topology copying servers in unknown but adding RS options, then updates servers, then transitions to primary if necessary. Conceptually this to me is much easier to, in particular, invariant check.

mbroadst:
@ajdavis not [exactly true](https://github.com/mongodb/mongo-python-driver/blob/master/pymongo/topology.py#L94), but agreed that from initial transition no driver I've seen ever transitions _into_ an `Unknown` state. Does that happen for you Oleg?

p-mongo:
No, Ruby does not transition into Unknown from something other than Unknown.

p-mongo:
However, per the table in https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#topologytype-table Unknown topology can be "organically" instantiated during sdam flow (as opposed to fixed instantiations during cluster construction) when already in Unknown and discovering a ghost. I am considering adding a spec test for this case.

ajdavis:
Let's not add a spec test for that case, since PyMongo, the C Driver, and others like them wouldn't pass through those states in real life. A Ruby-specific test for your driver would be better for the scenario of discovering an RSGhost while the topology type is Unknown.

PR #409, opened 2018-11-02
Verify max set version and max election id on topologies

p-mongo:
Existing specs verify set version & election id on individual servers. This PR adds verification of max of both of those on topologies.

Implemented in Ruby in https://github.com/mongodb/mongo-ruby-driver/pull/1117.

### DIFF
--- a/source/server-discovery-and-monitoring/tests/rs/equal_electionids.json
+++ b/source/server-discovery-and-monitoring/tests/rs/equal_electionids.json
@@ -60,7 +60,11 @@
         },
         "topologyType": "ReplicaSetWithPrimary",
         "logicalSessionTimeoutMinutes": null,
-        "setName": "rs"
+        "setName": "rs",
+        "maxSetVersion": 1,
+        "maxElectionId": {
+          "$oid": "000000000000000000000001"
+        }
       }
     }
   ]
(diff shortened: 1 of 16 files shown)


ajdavis:
Sorry for the delayed review. When I first wrote these tests, I included some tests that validate the behavior of maxSetVersion and maxElectionId implicitly. For example, a driver passes primary_disconnect_electionid.yml only if it tracks maxElectionId. Your change tests this part of the spec explicitly, which I think is worthwhile, and I tested that the C Driver can implement these tests easily and they succeed.

Matt and David, do you agree that this is a worthwhile change?

xdg:
If the existing tests *cover* the use of maxSetVersion and maxElectionId, then I think this is not a worthwhile change.  Making 13 drivers update their test harnesses and specs for something that doesn't add coverage isn't the best use of engineering time. 

If others disagree and this change is made, the README for tests also needs to be updated.

p-mongo:
This change adds coverage for max set version/max election id as exposed on topology descriptions, which is part of the specification. Without this change a driver may internally track max set version/max election id, never expose it and pass spec tests.

Additionally, assertions on max set version/max election id are helpful during driver *development* in that they highlight potential problems earlier. A relatively frequent occurrence  I noticed with spec tests was the failures would be detached from the time and place that actually caused them, making debugging (and subsequently development) take longer than it would had the failures been more related to the underlying problem.

mbroadst:
I tend to agree with David, but also am convinced these are worthy changes. @xdg, at the same time we're batching up quite a few SDAM changes here, I imagine we'll be doing the work in all drivers either way.

mbroadst:
@p-mongo sorry for the delayed review on all of these, I was finishing up code review and hope to test them all in the node driver this afternoon

mbroadst:
@p-mongo these pass locally, and you have two approvals - please merge.

p-mongo:
https://jira.mongodb.org/browse/SPEC-1171

PR #410, opened 2018-11-02
SPEC-1280 Extend test coverage for topology events

p-mongo:
This PR adds:

1. A test for SDAM events involving server removal. This is the first spec test exercising server closed event.
2. A test for events when both primary and secondary respond. I use this to verify that secondary response's events are correct. Existing tests stop at the primary response.

### DIFF
--- a/source/server-discovery-and-monitoring/tests/monitoring/replica_set_me_mismatch.yml
+++ b/source/server-discovery-and-monitoring/tests/monitoring/replica_set_me_mismatch.yml
@@ -1,0 +1,201 @@
+description: "Monitoring a topology that is a replica set with a me mismatch in seeds"
+uri: "mongodb://a,b"
+phases: 
+  # phase 1
+  - 
+    responses: 
+      - 
+        - "a:27017"
+        - 
+          ok: 1
+          ismaster: true
+          setName: "rs"
+          setVersion: 1
+          primary: "c:27017"
+          hosts: 
+            - "c:27017"
+            - "d:27017"
+          minWireVersion: 0
+          maxWireVersion: 4
+    outcome: 
+      events: 
+        - 
+          topology_opening_event: 
+            topologyId: "42"
+        - 
+          topology_description_changed_event: 
+            topologyId: "42"
+            previousDescription: 
+              topologyType: "Unknown"
+              servers: []
+            newDescription: 
+              topologyType: "Unknown"
+              servers: 
+                - 
+                  address: "a:27017"
+                  arbiters: []
+                  hosts: []
+                  passives: []
+                  type: "Unknown"
+                - 
+                  address: "b:27017"
+                  arbiters: []
+                  hosts: []
+                  passives: []
+                  type: "Unknown"
+        - 
+          server_opening_event: 
+            topologyId: "42"
+            address: "a:27017"
+        - 
+          server_opening_event: 
+            topologyId: "42"
+            address: "b:27017"
+        - 
+          server_description_changed_event: 
+            topologyId: "42"
+            address: "a:27017"
+            previousDescription: 
+              address: "a:27017"
+              arbiters: []
+              hosts: []
+              passives: []
+              type: "Unknown"
+            newDescription: 
+              address: "a:27017"
+              arbiters: []
+              hosts: 
+                - "c:27017"
+                - "d:27017"
+              passives: []
+              primary: "c:27017"
+              setName: "rs"
+              type: "RSOther"
+        - 
+          topology_description_changed_event: 
+            topologyId: "42"
+            previousDescription: 
+              topologyType: "Unknown"
+              servers: 
+                - 
+                  address: "a:27017"
+                  arbiters: []
+                  hosts: []
+                  passives: []
+                  type: "Unknown"
+                - 
+                  address: "b:27017"
+                  arbiters: []
+                  hosts: []
+                  passives: []
+                  type: "Unknown"
+            newDescription: 
+              topologyType: "ReplicaSetNoPrimary"
+              setName: rs
+              servers: 
+                - 
+                  address: "c:27017"
+                  arbiters: []
+                  hosts: []
+                  passives: []
+                  type: "Unknown"
+                - 
+                  address: "d:27017"
+                  arbiters: []
+                  hosts: []
+                  passives: []
+                  type: "Unknown"
+        - 
+          server_opening_event: 
+            topologyId: "42"
+            address: "c:27017"
+        - 
+          server_opening_event: 
+            topologyId: "42"
+            address: "d:27017"
+
+  # phase 2
+  - 
+    responses: 
+      - 
+        - "c:27017"
+        - 
+          ok: 1
+          ismaster: true
+          setName: "rs"
+          setVersion: 1
+          primary: "c:27017"
+          hosts: 
+            - "c:27017"
+            - "d:27017"
+          minWireVersion: 0
+          maxWireVersion: 4
+    outcome: 
+      events: 
+        - 
+          server_description_changed_event: 
+            topologyId: "42"
+            address: "c:27017"
+            previousDescription: 
+              address: "c:27017"
+              arbiters: []
+              hosts: []
+              passives: []
+              type: "Unknown"
+            newDescription: 
+              address: "c:27017"
+              arbiters: []
+              hosts: 
+                - "c:27017"
+                - "d:27017"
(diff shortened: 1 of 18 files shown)


source/server-discovery-and-monitoring/tests/monitoring/replica_set_me_mismatch.yml
```diff
@@ -0,0 +1,201 @@
...
+            newDescription: 
+              address: "a:27017"
+              arbiters: []
+              hosts: 
+                - "c:27017"
+                - "d:27017"
+              passives: []
+              primary: "c:27017"
+              setName: "rs"
+              type: "RSOther"
```
ajdavis: Why is this RSOther? I'd expect RSPrimary. RSOther is defined:

"setName" in response, "hidden: true" or not primary,  
secondary, nor arbiter.

source/server-discovery-and-monitoring/tests/monitoring/replica_set_with_second_seed_removal.yml
```diff
@@ -0,0 +1,100 @@
...
+              servers: []
+            newDescription: 
+              topologyType: "Unknown"
+              servers: 
+                - 
+                  address: "a:27017"
+                  arbiters: []
+                  hosts: []
+                  passives: []
+                  type: "Unknown"
```
ajdavis: I expect the second server to still be in the description here - no isMaster response has been processed yet.

source/server-discovery-and-monitoring/tests/monitoring/replica_set_me_mismatch.yml
```diff
@@ -0,0 +1,201 @@
+description: "Monitoring a topology that is a replica set with a me mismatch in seeds"
```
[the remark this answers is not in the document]
p-mongo: I'll rename to replica_set_other_seed. The goal of this test was to obtain the primary response from a server not in the seed list.

source/server-discovery-and-monitoring/tests/monitoring/replica_set_me_mismatch.yml
```diff
@@ -0,0 +1,201 @@
...
+uri: "mongodb://a,b"
```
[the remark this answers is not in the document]
p-mongo: This test uses the same URI as for example https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/tests/monitoring/replica_set_with_primary.yml, which doesn't give replicaSet.

Ruby driver has different behavior when RS topology is forced or discovered; see for example https://github.com/mongodb/mongo-ruby-driver/blob/master/spec/spec_tests/data/sdam_monitoring/standalone.yml vs https://github.com/mongodb/mongo-ruby-driver/blob/master/spec/spec_tests/data/sdam_monitoring/discovered_standalone.yml. If RS topology is forced, the sequence of events will be different - along the lines of https://github.com/mongodb/mongo-ruby-driver/blob/master/spec/spec_tests/data/sdam_monitoring/required_replica_set.yml, topology will transition from unknown/no servers to RSWithoutPrimary directly, skipping Unknown with servers.

source/server-discovery-and-monitoring/tests/monitoring/replica_set_with_primary_removal.yml
```diff
@@ -0,0 +1,159 @@
...
+              servers: []
+            newDescription: 
+              topologyType: "Unknown"
+              servers: 
+                - 
+                  address: "a:27017"
+                  arbiters: []
+                  hosts: []
+                  passives: []
+                  type: "Unknown"
```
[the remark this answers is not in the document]
p-mongo: This was an oversight in the test runner, it did not verify that all of the "actual" servers were present in the "expected" server list.

source/server-discovery-and-monitoring/tests/monitoring/replica_set_me_mismatch.yml
p-mongo: Ruby driver additionally requires that the primary's idea of the primary matches the address which the driver has for it. This was added for https://jira.mongodb.org/browse/RUBY-1241 in https://github.com/mongodb/mongo-ruby-driver/commit/b35b667123710fc4c9a0550cdb105b09cc2be065. A comment in the ticket indicated this may be relevant to other drivers. Thoughts?

p-mongo:
I made the changes in https://github.com/mongodb/mongo-ruby-driver/pull/1150, will update this PR once we decide what to do about the primary/other situation.

source/server-discovery-and-monitoring/tests/monitoring/replica_set_me_mismatch.yml
ajdavis: Huh ... sorry, can you explain some more to me? What does this rule accomplish that is not already accomplished by this part of updateRSFromPrimary?:
```
    for each server in topologyDescription.servers:
        if server.address not in description's "hosts", "passives", or "arbiters":
            remove the server and stop monitoring it
```

source/server-discovery-and-monitoring/tests/monitoring/replica_set_me_mismatch.yml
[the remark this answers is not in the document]
ajdavis: Thanks, turns out PyMongo can discover an RS when given two hosts like this. But let's add "replicaSet=rs", which many of the other spec tests have.

source/server-discovery-and-monitoring/tests/monitoring/replica_set_other_chain.yml
```diff
@@ -0,0 +1,222 @@
+description: "Multiple RSOther responses with different servers in each"
+uri: "mongodb://a,x/"
```
ajdavis: Let's add replicaset=rs here too.

source/server-discovery-and-monitoring/tests/monitoring/replica_set_other_chain.yml
```diff
@@ -0,0 +1,222 @@
...
+                  passives: []
+                  primary: "c:27017"
+                  setName: "rs"
+                  type: "RSOther"
+                -
+                  address: "c:27017"
+                  arbiters: []
+                  hosts: []
+                  passives: []
+                  type: "Unknown"
```
ajdavis: "PossiblePrimary" type

source/server-discovery-and-monitoring/tests/monitoring/replica_set_other_change.yml
```diff
@@ -0,0 +1,225 @@
...
+                  passives: []
+                  primary: "b:27017"
+                  setName: "rs"
+                  type: "RSOther"
+                -
+                  address: "b:27017"
+                  arbiters: []
+                  hosts: []
+                  passives: []
+                  type: "Unknown"
```
ajdavis: PossiblePrimary

source/server-discovery-and-monitoring/tests/monitoring/replica_set_other_change.yml
```diff
@@ -0,0 +1,225 @@
...
+                -
+                  address: "c:27017"
+                  arbiters: []
+                  hosts: []
+                  passives: []
+                  type: "Unknown"
```
ajdavis: PossiblePrimary

mbroadst:
Since deciding on the test format has been deferred to this PR, please include an update to describe how test runners should behave given multiple phases: are event collected in a single batch, or compared at the end of each phase? If we choose the latter, please propose some format to deal with the initial events not triggered by a response.

p-mongo:
@mbroadst Added prose to explain how to handle events and phases.

@ajdavis I added https://github.com/mongodb/specifications/pull/410/commits/dccadda39e3d9c5436a918560215278b5406fb1a  to cover the change made in https://github.com/mongodb/mongo-ruby-driver/commit/b35b667123710fc4c9a0550cdb105b09cc2be065. Without this change, the driver removes a:27017 and adds b:27017 as unknown upon hearing from a, but it has not succeeded in establishing a connection to b yet (hasn't even attempted, in fact). It is possible that the driver will never be able to connect to b, and by removing a it has lost its last functioning link to the cluster.

IanWhalen:
@p please file a SPEC ticket.

jyemin:
@p-mongo please either revive this PR or abandon it.

p-mongo:
This PR is in my queue to address.

PR #411, opened 2018-11-05
Define SDAM server description equality and specify that equal descriptions do not run SDAM flow

p-mongo:
Without this test an implementation can publish server description & topology change events on each ismaster response, even when there are no changes between the ismaster response being processed and the previous one.

https://jira.mongodb.org/browse/RUBY-1557?

### DIFF
--- a/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
+++ b/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
@@ -306,30 +306,33 @@ Fields:
   The last opTime reported by the server; an ObjectId or null.
   (Only mongos and shard servers record this field when monitoring
   config servers as replica sets.)
-* type: a `ServerType`_ enum value. Default Unknown.
-* minWireVersion, maxWireVersion:
+* (=) type: a `ServerType`_ enum value. Default Unknown.
+* (=) minWireVersion, maxWireVersion:
   the wire protocol version range supported by the server.
   Both default to 0.
   `Use min and maxWireVersion only to determine compatibility`_.
-* me: The hostname or IP, and the port number, that this server was configured with in the replica set. Default null.
-* hosts, passives, arbiters: Sets of addresses.
+* (=) me: The hostname or IP, and the port number, that this server was
+  configured with in the replica set. Default null.
+* (=) hosts, passives, arbiters: Sets of addresses.
   This server's opinion of the replica set's members, if any.
   These `hostnames are normalized to lower-case`_.
   Default empty.
   The client `monitors all three types of servers`_ in a replica set.
-* tags: map from string to string. Default empty.
-* setName: string or null. Default null.
-* setVersion: integer or null. Default null.
-* electionId: an ObjectId, if this is a MongoDB 2.6+ replica set member that
+* (=) tags: map from string to string. Default empty.
+* (=) setName: string or null. Default null.
+* (=) setVersion: integer or null. Default null.
+* (=) electionId: an ObjectId, if this is a MongoDB 2.6+ replica set member that
   believes it is primary. See `using setVersion and electionId to detect stale primaries`_.
   Default null.
-* primary: an address. This server's opinion of who the primary is.
+* (=) primary: an address. This server's opinion of who the primary is.
   Default null.
 * lastUpdateTime: when this server was last checked. Default "infinity ago".
-* logicalSessionTimeoutMinutes: integer or null. Default null.
+* (=) logicalSessionTimeoutMinutes: integer or null. Default null.
 
 "Passives" are priority-zero replica set members that cannot become primary.
 The client treats them precisely the same as other members.
+
+Fields marked (=) are used for `Server Description Equality`_ comparison.
 
 .. _configured:
 
@@ -677,7 +680,10 @@ Parsing an ismaster response
 
 The client represents its view of each server with a `ServerDescription`_.
 Each time the client `checks`_ a server,
-it replaces its description of that server with a new one.
+it MUST replace its description of that server with a new one.
+This replacement MUST happen even if the new server description compares
+equal to the previous one, in order to keep client-tracked attributes
+like last update time and round trip time up to date.
 
 ServerDescriptions are created from ismaster outcomes as follows:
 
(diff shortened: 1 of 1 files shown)


ajdavis:
Context: the SDAM Monitoring spec says ServerDescriptionChanged event fires "when the old server description is not equal to the new server description". The term "equal" is not defined, but later the spec says ServerDescriptionChangedEvent is "Published when server description changes, but does NOT include changes to the RTT."

I think this is a useful test that matches the spec.

mbroadst:
@oleg do you have a spec amendment for what "equal" means in this case? I support this change, it expressed the bug in the node implementation, but I think we could make it easier for others to fix by very clearly defining equality (and not necessarily by all defined properties).

p-mongo:
Yes I can add that.

mbroadst:
@ajdavis I occurs to me that I can't find anywhere that we actually specify what the `events` field in `outcome` is! We all make an assumption that we should save off SDAM events, and compare them to this list, but the Monitoring readme says "this is the same as SDAM, use that test runner", and the primary README for SDAM doesn't mention `events` anywhere. Should `events` be cleared out between phases? The way the node test runner is implemented right now that's not the case, so this added test case ends up with _all_ events.

mbroadst:
I see what's happening here:
- this is the first monitoring test with multiple `phase`s
- we definitely need definition about how we treat `events`, and capturing events for comparison
  - right now it appears that `events` are aggregated from both setup + running the phase(s)

So this test breaks this expectation by assuming that a second phase will be run and the `events` should be empty. I think the most "correct" approach would be to move `events` to a new top-level property, and have it mean "all the events, in order, that ever occur during the following phases", but that also means we'd have to change all of the monitoring tests.

Another approach, that would likely require no changes, would be simply to fill out the `events` inside an `outcome` _only_ for the last phase. In this entry, all events for all phases should be listed in order. Either way, let's pick something and actually document it for our future selves.

p-mongo:
Chronologically the tests currently in #410 was the first ones I wrote which involved multiple phases. I started with what is currently called "me mismatch" there though it's really an "other seed", and then added "primary and secondary" as a simpler case of just having multiple phases with events in them.

I had to modify the Ruby runner to assign each event to a phase.

Putting all events into a single global list seems unnecessarily weak.

mbroadst:
Why do you consider putting all events in a global list "weak"? Either way this is how the current tests are written.  For example, the first event in your `outcome` is:
```
{
  "topology_opening_event": {
    "topologyId": "42"
  }
}
```

This event is not the result of the response in that phase, but rather of the initial opening of the topology.  So we will need to update the test format and document the changes. I've provided a few alternatives, maybe you have a better idea?

p-mongo:
Having a single list of events without breaking them down by phases means an event can in theory be published in the wrong phase without failing the tests. Even if this does not happen, I found it much easier to separate events by phases for debugging purposes as it made very clear that, for instance, the driver did the right thing when processing a response from primary but then did not do the right thing when processing a response from secondary. If all events are bundled together then it's not clear where the issue may be, if the actual events don't match the expected ones.

Regarding topology/server opening events being bundled with the first phase, this can be repaired by giving them their own phase which I have done in https://github.com/mongodb/mongo-ruby-driver/pull/1149/commits/319d71702a2b5176260f5ebbade234a9523fb195. This test change required no runner changes to pass in Ruby.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -838,6 +841,19 @@ are parsed from the ismaster response in the obvious way.
 
 .. _updates its view of the topology:
 
+Server Description Equality
+```````````````````````````
+
+For the purposes of updating topology description, two server descriptions
+MUST be considered equal if the data they contain which is parsed from
+the respective ismaster response is identical. If the parsed ismaster data
+is different, the server descriptions MUST be considered not equal.
```
ajdavis: Let's be clearer that this excludes RTT and lastUpdateTime.

ajdavis:
I have no opinion right now about whether to merge the responses and events into one list each, or keep them as-is.

mbroadst:
I think providing an empty initial phase, with expected events, and then keep events related to each phase in the phase definition is fine. Please define this somewhere before we go any further though. 
The proposed approach of leaving them all in a single list was meant to require the least amount of work from all drivers to implement, but I think you have a solid argument and it's worth proposing to the team at large.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
mbroadst: Agreed. I was looking for something more explicit than "they have the same ismaster", which fields? Can we ignore some ismaster fields? This equality check will be happening every 10s in some cases, can we reduce the number of fields we need to check? Can we take a subset of fields and hash them for comparison?

ajdavis: Hold up =) Because the equality check happens once every 10 seconds that's an argument for keeping it simple, not for optimizing it! Let's leave it up to maintainers how they want to check that fields are equal.

mbroadst: Ok, well then let's at least enumerate which fields should not be considered.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -306,31 +306,34 @@ Fields:
...
+* (=) setName: string or null. Default null.
+* (=) setVersion: integer or null. Default null.
+* (=) electionId: an ObjectId, if this is a MongoDB 2.6+ replica set member that
   believes it is primary. See `using setVersion and electionId to detect stale primaries`_.
   Default null.
-* primary: an address. This server's opinion of who the primary is.
+* (=) primary: an address. This server's opinion of who the primary is.
   Default null.
 * lastUpdateTime: when this server was last checked. Default "infinity ago".
 * logicalSessionTimeoutMinutes: integer or null. Default null.
```
ajdavis: Let's include logicalSessionTimeoutMinutes among things marked (=)

p-mongo:
https://jira.mongodb.org/browse/SPEC-1176

PR #412, opened 2018-11-06
These two topology descriptions were missing set names, add those

p-mongo:


p-mongo:
Yes, the idea was to catch incorrect tests in addition to bugs in implementation.

mbroadst:
@p-mongo please convert the updated yaml tests to json, and merge

PR #413, opened 2018-11-06
SPEC-1122 Handle postBatchResumeToken in change streams.

kevinAlbs:
This solves the problem of resuming a change stream with a potentially very old resume token. This can happen if the change stream has not received any events in a long time. And consequences of this are: (1) the oplog may have already rolled over and (2) loading a very old oplog entry can be a huge performance hit.

In 4.2, the query team is planning to have responses of aggregate and getMore include a minPromisedResumeToken. This will represent a resume token greater or equal to the last document returned. When not in the middle of a returning documents in a batch, a change stream should resume using the most recent minPromisedResumeToken.

This commit does not include tests. Tests will be added after the corresponding server work is done.

### DIFF
--- a/source/change-streams/change-streams.rst
+++ b/source/change-streams/change-streams.rst
@@ -9,8 +9,8 @@ Change Streams
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: 3.6
-:Last Modified: December 4, 2018
-:Version: 1.5.2
+:Last Modified: January 3, 2019
+:Version: 1.6.0
 
 .. contents::
 
@@ -84,11 +84,11 @@ Server Specification
 Response Format
 ---------------
 
-If the aggregation with ``$changeStream`` specified completes successfully, the resultant stream will deliver documents in the following format:
+If an aggregate command with a ``$changeStream`` stage completes successfully, the response contains documents with the following structure:
 
 .. code:: typescript
 
-  {
+  class ChangeStreamDocument {
     /**
      * The id functions as an opaque token for use when resuming an interrupted
      * change stream.
@@ -98,7 +98,7 @@ If the aggregation with ``$changeStream`` specified completes successfully, the 
     /**
      * Describes the type of operation represented in this change notification.
      */
-    operationType: "insert" | "update" | "replace" | "delete" | "invalidate" | "drop" | "dropDatabase";
+    operationType: "insert" | "update" | "replace" | "delete" | "invalidate" | "drop" | "dropDatabase" | "rename";
 
     /**
      * Contains two fields: “db” and “coll” containing the database and
@@ -155,6 +155,46 @@ If the aggregation with ``$changeStream`` specified completes successfully, the 
     removedFields: Array<String>;
   }
 
+The responses to a change stream aggregate or getMore have the following structures:
+
+.. code:: typescript
+
+  /**
+   * Response to a successful aggregate.
+   */
+  {
+      ok: 1,
+      cursor: {
+         ns: String,
+         id: Int64,
+         firstBatch: Array<ChangeStreamDocument>,
+         /**
+          * postBatchResumeToken is returned in MongoDB 4.2 and later.
+          */
+         postBatchResumeToken: Document
+      },
+      operationTime: Timestamp,
+      $clusterTime: Document,
+  }
+
+  /**
+   * Response to a successful getMore.
+   */
+  {
+      ok: 1,
+      cursor: {
+         ns: String,
+         id: Int64,
+         nextBatch: Array<ChangeStreamDocument>
+         /**
+          * postBatchResumeToken is returned in MongoDB 4.2 and later.
+          */
+         postBatchResumeToken: Document
+      },
+      operationTime: Timestamp,
+      $clusterTime: Document,
+  }
+
 **NOTE:** The above format is provided for illustrative purposes, and is subject to change without warning.
 
 ----------
@@ -167,7 +207,13 @@ Driver API
     /**
      * The resume token (_id) of the document the iterator last returned
      */
-    private resumeToken: Document;
+    private documentResumeToken: Document;
+
+    /**
+     * The most recent postBatchResumeToken returned in an aggregate or
+     * getMore response. For pre-4.2 versions of MongoDB, this remains unset.
+     */
+    private postBatchResumeToken: Document;
 
     /**
      * The pipeline of stages to append to an initial ``$changeStream`` stage
@@ -393,6 +439,8 @@ A ``ChangeStream`` is an abstraction of a `TAILABLE_AWAIT <https://github.com/mo
 
 A change stream MUST track the last resume token returned by the iterator to the user, caching it locally for use in future attempts to resume.  A driver MUST raise an error on the first response received without a resume token (e.g. the user has removed it with a pipeline stage), and close the change stream.  The error message SHOULD resemble “Cannot provide resume functionality when the resume token is missing”.
 
+A change stream MUST track the latest ``postBatchResumeToken`` included in an aggregate or getMore response if available.
+
 A change stream MUST attempt to resume a single time if it encounters any resumable error.  A change stream MUST NOT attempt to resume on any other type of error, with the exception of a “not master” server error.  If a driver receives a “not master” error (for instance, because the primary it was connected to is stepping down), it will treat the error as a resumable error and attempt to resume.
 
 In addition to tracking the most recently delivered resume token, change streams MUST also track the read preference specified when the change stream was created. In the event of a resumable error, a change stream MUST perform server selection with the original read preference before attempting to resume.
@@ -411,14 +459,19 @@ startAtOperationTime
 
 The server expects ``startAtOperationTime`` as a BSON Timestamp. Drivers MUST allow users to specify a ``startAtOperationTime`` option in the ``watch`` helpers. They MUST allow users to specify this value as a raw ``Timestamp``.
 
-``startAtOperationTime`` and ``resumeAfter`` are mutually exclusive; if both ``startAtOperationTime`` and ``resumeAfter`` are set, the server will return an error. Drivers MUST NOT throw a custom error, and MUST defer to the server error.
+``startAtOperationTime``, ``resumeAfter``, and ``startAfter`` are all mutually exclusive; if any two are set, the server will return an error. Drivers MUST NOT throw a custom error, and MUST defer to the server error.
 
-If neither ``startAtOperationTime`` nor ``resumeAfter`` are specified, and the max wire version is >= ``7``, and the initial ``aggregate`` command does not return a resumeToken (indicating no results), the ``ChangeStream`` MUST save the ``operationTime`` from the initial ``aggregate`` command when it returns.
+The ``ChangeStream`` MUST save the ``operationTime`` from the initial ``aggregate`` response when the following critera are met:
+
+- None of ``startAtOperationTime``,  ``resumeAfter``, ``startAfter`` were specified in the ``ChangeStreamOptions``.
+- The max wire version is >= ``7``.
+- The initial ``aggregate`` response did not include a ``documentResumeToken`` (indicating no results).
+- The initial ``aggregate`` response did not include a ``postBatchResumeToken``.
 
 resumeAfter
 ^^^^^^^^^^^
 
-``resumeAfter`` is used to resume a changeStream that has been stopped to ensure that only changes starting with the log entry immediately *after* the provided token will be returned. If the resume token specified does not exist, the server will return an error. 
+``resumeAfter`` is used to resume a ``ChangeStream`` that has been stopped to ensure that only changes starting with the log entry immediately *after* the provided token will be returned. If the resume token specified does not exist, the server will return an error.
 
 Resume Process
 ^^^^^^^^^^^^^^
@@ -427,24 +480,104 @@ Once a ``ChangeStream`` has encountered a resumable error, it MUST attempt to re
 
 - Perform server selection.
 - Connect to selected server.
-- If the ``ChangeStream`` has not received any changes, and ``resumeAfter`` is not specified, and the max wire version is >= ``7``:
+- If the ``ChangeStream`` has a cached ``postBatchResumeToken`` and has returned all documents in the most recent batch (or the most recent batch was empty):
 
-    - The driver MUST execute the known aggregation command.
-    - The driver MUST specify the ``startAtOperationTime`` key set to the ``startAtOperationTime`` provided by the user or saved from the original aggregation.
-    - The driver MUST NOT set a ``resumeAfter`` key.
-    - In this case, the ``ChangeStream`` will return all changes that occurred after the specified ``startAtOperationTime``.
+  - The driver MUST set ``resumeAfter`` to the cached ``postBatchResumeToken``.
+  - The driver MUST NOT set ``startAfter``. If ``startAfter`` was in the original aggregation command, the driver MUST remove it.
+  - The driver MUST NOT set ``startAtOperationTime``. If ``startAtOperationTime`` was in the original aggregation command, the driver MUST remove it.
+- Else if the ``ChangeStream`` has a cached ``documentResumeToken``:
+
+  - The driver MUST set ``resumeAfter`` to the cached ``documentResumeToken``.
+  - The driver MUST NOT set ``startAfter``. If ``startAfter`` was in the original aggregation command, the driver MUST remove it.
+  - The driver MUST NOT set ``startAtOperationTime``. If ``startAtOperationTime`` was in the original aggregation command, the driver MUST remove it.
+- Else if the ``ChangeStream`` was created with ``startAfter``:
+
+  - The driver MUST set ``resumeAfter`` to the value of the originally used ``startAfter``.
+  - The driver MUST NOT set ``startAfter``. The driver MUST remove it from the original aggregation command.
+  - The driver MUST NOT set ``startAtOperationTime``.
+- Else if the ``ChangeStream`` was created with ``resumeAfter``:
+
+  - The driver MUST set ``resumeAfter`` to the value of the originally used ``resumeAfter``.
+  - The driver MUST NOT set ``startAfter``.
+  - The driver MUST NOT set ``startAtOperationTime``.
+- Else if the ``ChangeStream`` has a saved operation time (either from an originally specified ``startAtOperationTime`` or saved from the original aggregation) and the max wire version is >= ``7``:
+
+  - The driver MUST not set ``resumeAfter``.
+  - The driver MUST NOT set ``startAfter``.
+  - The driver MUST set ``startAtOperationTime`` to the value of the originally used ``startAtOperationTime`` or the one saved from the original aggregation.
+
 - Else:
+  - The driver MUST not set ``resumeAfter``, ``startAfter``, or ``startAtOperationTime``.
+  - The driver MUST use the original aggregation command to resume.
 
-    - The driver MUST execute the known aggregation command.
-    - The driver MUST specify a ``resumeAfter`` with the last known ``resumeToken``.
-    - The driver MUST NOT set a ``startAtOperationTime``.
-    - If a ``startAtOperationTime`` key was part of the original aggregation command, the driver MUST remove it.
-    - In this case, the ``ChangeStream`` will return notifications starting with the oplog entry immediately *after* the provided token.
+When ``resumeAfter`` is specified the ``ChangeStream`` will return notifications starting with the oplog entry immediately *after* the provided token.
 
 If the server supports sessions, the resume attempt MUST use the same session as the previous attempt's command.
 
 A driver SHOULD attempt to kill the cursor on the server on which the cursor is opened during the resume process, and MUST NOT attempt to kill the cursor on any other server.
 
+
+Exposing All Resume Tokens
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+:since: 4.2
+
+Users can retrieve the ``documentResumeToken`` by inspecting the _id on each ``ChangeDocument``. But since MongoDB 4.2, aggregate and getMore responses also include a ``postBatchResumeToken``. Drivers use one or the other when automatically resuming, as described in `Resume Process`_.
+
+Drivers MUST expose a mechanism to retrieve the same resume token that would be used to automatically resume. It MUST be possible to use this mechanism after iterating every document. It MUST be possible for users to use this mechanism periodically even when no documents are getting returned (i.e. ``getMore`` has returned empty batches). Drivers have two options to implement this.
+
+Option 1: ChangeStream::getResumeToken()
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code:: typescript
+
+  interface ChangeStream extends Iterable<Document> {
+    /**
+     * Returns a resume token that should be used to resume after the most
+     * recently returned change.
+     */
+    public getResumeToken() Optional<Document>;
+  }
+
+This method returns the cached ``documentResumeToken`` or ``postBatchResumeToken`` following rules from `Resume Process`_:
+
+- If the ``ChangeStream`` has no cached ``documentResumeToken`` or ``postBatchResumeToken`` this returns an unset optional.
+- If the ``ChangeStream`` has a cached ``postBatchResumeToken`` and has returned all documents in the most recent batch (or the most recent batch was empty) this returns the ``postBatchResumeToken``.
+- Otherwise, this returns the cached ``documentResumeToken``.
+
+This MUST be implemented in synchronous drivers. This MAY be implemented in asynchronous drivers.
+
+Option 2: Event Emitted for Resume Token
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Allow users to set a callback to listen for new resume tokens. The exact interface is up to the driver, but it MUST meet the following criteria:
+
+- The callback is set in the same manner as a callback used for receiving change documents.
+- The callback accepts a resume token as an argument.
+- The callback (or event) MAY include an optional ChangeDocument, which is unset when called with resume tokens sourced from ``postBatchResumeToken``.
+
+A possible interface for this callback MAY look like:
+
+.. code:: typescript
+
+  interface ChangeStream extends Iterable<Document> {
+    /**
+     * Returns a resume token that should be used to resume after the most
+     * recently returned change.
+     */
+    public onResumeTokenChanged(ResumeTokenCallback:(Document resumeToken) => void);
+  }
+
+This MUST not be implemented in synchronous drivers. This MAY be implemented in asynchronous drivers.
+
+Not Blocking on Iteration
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Synchronous drivers MUST provide a way to iterate a change stream without blocking until a change document is returned. This MUST give the user an opportunity to get the most up-to-date resume token, even when the change stream continues to receive empty batches in getMore responses. This allows users to call ``ChangeStream::getResumeToken()`` after iterating every document and periodically when no documents are getting returned.
+
+Although the implementation of tailable awaitData cursors is not specified, this MAY be implemented with a ``tryNext`` method on the change stream cursor.
+
+All drivers MUST document how users can iterate a change stream and receive *all* resume token updates. `Why do we allow access to the resume token to users`_ shows an example. The documentation MUST state that users intending to store the resume token should use this method to get the most up to date resume token.
 
 Notes and Restrictions
 ^^^^^^^^^^^^^^^^^^^^^^
@@ -494,16 +627,20 @@ Imagine a scenario in which a user wants to process each change to a collection 
 
 .. code:: python
 
-  resumeToken = None
   localChange = getChangeFromLocalStorage()
+  resumeToken = getResumeTokenFromLocalStorage()
+
   if localChange:
     processChange(localChange)
-    resumeToken = localChange['_id']
 
   try:
-      for change in db.collection.watch([...], resumeAfter=resumeToken):
-          persistToLocalStorage(change)
-          processChange(change)
+      change_stream = db.collection.watch([...], resumeAfter=resumeToken)
+      while True:
+          change = change_stream.try_next()
+          persistResumeTokenToLocalStorage(change_stream.get_resume_token())
+          if change:
+            persistChangeToLocalStorage(change)
+            processChange(change)
   except Exception:
       log.error("...")
 
@@ -560,6 +697,16 @@ In the above example, not sending ``startAtOperationTime`` will result in the ch
 the changes that occurred while the server and client are partitioned. By sending ``startAtOperationTime``,
 the server will know to include changes from that previous point in time.
 
+--------------------------------------------------
+Why do we need to expose the postBatchResumeToken?
+--------------------------------------------------
+
+Resume tokens refer to an oplog entry. The resume token from the ``_id`` of a document corresponds the oplog entry of the change. The ``postBatchResumeToken`` represents the oplog entry the change stream has scanned up to on the server (not necessarily a matching change). This can be a much more recent oplog entry, and should be used to resume when possible.
+
+Attempting to resume with an old resume token may degrade server performance since the server needs to scan through more oplog entries. Worse, if the resume token is older than the last oplog entry stored on the server, then resuming is impossible.
+
+Imagine the change stream matches a very small percentage of events. On a ``getMore`` the server scans the oplog for the duration of ``maxAwaitTimeMS`` but finds no matching entries and returns an empty response (still containing a ``postBatchResumeToken``). There may be a long sequence of empty responses. Then due to a network error, the change stream tries resuming. If we tried resuming with the cached ``documentResumeToken`` this throws out the oplog scanning the server had done for the long sequence of getMores with empty responses. But resuming with the last ``postBatchResumeToken`` skips the unnecessary scanning of unmatched oplog entries.
+
 Test Plan
 =========
 
@@ -574,9 +721,9 @@ There should be no backwards compatibility concerns.
 Reference Implementations
 =========================
 
- - NODE (NODE-1055)
- - PYTHON (PYTHON-1338)
- - RUBY (RUBY-1228)
+- NODE (NODE-1055)
+- PYTHON (PYTHON-1338)
+- RUBY (RUBY-1228)
 
 Changelog
 =========
@@ -618,4 +765,6 @@ Changelog
 | 2018-09-09 | Added dropDatabase to change stream operationType          |
 +------------+------------------------------------------------------------+
 | 2018-12-14 | Added ``startAfter`` to change stream options              |
++------------+------------------------------------------------------------+
+| 2018-11-06 | Added handling of ``postBatchResumeToken``.                |
 +------------+------------------------------------------------------------+


source/change-streams/change-streams.rst
```diff
@@ -169,6 +208,12 @@ Driver API
      */
     private resumeToken: Document;
```
kevinAlbs: Question for drivers reviewers: How do people feel about adding a public method to retrieve the current resume token? As it stands, users who are implementing their own resume logic would have no way to retrieve the minPromisedResumeToken.

source/change-streams/change-streams.rst
```diff
@@ -417,13 +464,17 @@ Once a ``ChangeStream`` has encountered a resumable error, it MUST attempt to re
 - If the ``ChangeStream`` has not received any changes, and ``resumeAfter`` is not specified, and the max wire version is >= ``7``:
 
     - The driver MUST execute the known aggregation command.
-    - The driver MUST specify the ``startAtOperationTime`` key set to the ``startAtOperationTime`` provided by the user or saved from the original aggregation.
+    - If the ``ChangeStream`` has a saved ``minPromisedResumeToken``:
+
+         - The driver MUST specify ``resumeAfter`` with the cached ``minPromisedResumeToken``.
+    - Else:
+
+         - The driver MUST specify the ``startAtOperationTime`` key set to the ``startAtOperationTime`` provided by the user or saved from the original aggregation.
```
kevinAlbs: If we didn't have to support pre 4.2 versions we could simplify the resume logic to omit ever using startAtOperationTime.

p-mongo:
I think I figured it out, but can the difference between resume token and min promised resume token be explicitly stated? I.e. the former refers to a document which was returned by the change stream, and could be old relative to the time at which said document was returned, with the latter indicating there are no changes between the most recent returned document and a certain point in time/oplog.

source/change-streams/change-streams.rst
p-mongo: Since our guidance for users is to keep track of resume tokens (e.g. https://docs.mongodb.com/ruby-driver/master/tutorials/ruby-driver-change-streams/#resuming-a-change-stream), this seems required to me.

skriptble: If we add this method, would we suggest that users always use this method?

source/change-streams/change-streams.rst
```diff
@@ -169,6 +208,12 @@ Driver API
      */
     private resumeToken: Document;
 
+    /**
+     * The most recent promisedMinResumeToken returned in an aggregate or
+     * getMore response. For pre-4.2 versions of MongoDB, this remains unset.
+     */
+    private minPromisedResumeToken: Document;
```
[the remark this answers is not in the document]
jmikola: Reading the description of [SPEC-1122](https://jira.mongodb.org/browse/SPEC-1122) again, I agree that drivers need only track a single resume token.

I would think we could use the existing `resumeToken` property and simply change its documentation. So instead of:

```
  /**
   * The resume token (_id) of the document the iterator last returned
   */

  /**
   * The most recent promisedMinResumeToken returned in an aggregate or
   * getMore response. For pre-4.2 versions of MongoDB, this remains unset.
   */
  private minPromisedResumeToken: Document;
```

We have:

```
  /**
   * The resume token. This will either be the (_id) of the document most recently
   * returned by the iterator, or the most recent promisedMinResumeToken returned
   * in an aggregate or getMore response (if available).
   */
```

This would also eliminate other language in the current PR:

> When the change stream has returned all documents in a (possibly empty) batch (i.e. it needs to send a ``getMore``), it MUST overwrite the cached ``resumeToken`` with the value of the cached ``minPromisedResumeToken``.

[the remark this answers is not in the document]
kevinAlbs: Yes, only one resume token needs to be tracked at a given time.

I'm thinking something closer to:
```
/**
   * The resume token. This will either be the (_id) of the document most recently
   * returned by the iterator, or the most recent promisedMinResumeToken returned
   * in an aggregate or getMore response (if available, and all documents in the most
   * recent batch have been returned).
   */
```

I want to clearly convey that you should only set resumeToken to minPromisedResumeToken after all documents in the current batch have been returned.

source/change-streams/change-streams.rst
daprahamian: 👍 in favor of adding a method to get resume token

source/change-streams/change-streams.rst
kevinAlbs: I don't have a strong opinion on whether we check wire version. But if we don't check wire version, we could foreseeably backport the server change to 4.0 and have new drivers use it correctly.

kevinAlbs:
Hi SPEC-1122 reviewers,

@skriptble brought to my attention (and I believe @cswanson310 agrees) that we could solve another problem here, described in SERVER-37669: it's easy to project out the _id field of documents returned in a change stream with $replaceRoot. I'm very much in favor of doing this.

The proposed solution is to change the response of an agg/getMore to include an array of resume tokens in the body of the response.
```
{
    ok: 1,
    cursor: {
       ns: String,
       id: Int64,
       firstBatch: [
        <doc1>,
        <doc2>,
        <doc3>
      ]
    },
    operationTime: Timestamp,
    $clusterTime: Document,
    resumeTokens: [
      <resumeToken for doc1>,
      <resumeToken for doc2>,
      <minPromisedResumeToken>
    ]
}
```

The `resumeTokens` array corresponds to documents in the batch. However, the last entry might secretly be what the current PR calls the `minPromisedResumeToken`. This might be the same value as the resume token for the last document, but drivers won't have to care.

Before I update the review, I want to ask if anyone has outright objections (reply or thumbs up this comment). We can argue minor points (naming, etc.) in the review.

Best,
Kevin

source/change-streams/change-streams.rst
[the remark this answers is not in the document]
p-mongo: > Yes, only one resume token needs to be tracked at a given time.

This works when the application is guaranteed to process all documents returned in a batch, but does not work for the case when an application is processing documents individually and may stop at any time. In the latter case resume token for each document still must be tracked to resume from that document onward.

p-mongo:
How about the server putting min promised resume token into the last change stream document as its id? Then no changes at all would be needed in the drivers.

p-mongo:
Or, if id on a change stream document means something to the server, the array of *resume tokens* on the top level should have the same length as the number of documents returned, with the last token either being id of the last document or min promised resume token.

kevinAlbs:
> How about the server putting min promised resume token into the last change stream document as its id? Then no changes at all would be needed in the drivers.

That wouldn't solve the problem of https://jira.mongodb.org/browse/SERVER-37669

kevinAlbs:
> Or, if id on a change stream document means something to the server, the array of _resume tokens_ on the top level should have the same length as the number of documents returned, with the last token either being id of the last document or min promised resume token.

That's what I'm proposing. `resumeTokens` has the same length as `firstBatch`. The last resume token is the `minPromisedResumeToken`

p-mongo:
My bad, I misread the code. Looks good now that I read it correctly.

skriptble:
> > Or, if id on a change stream document means something to the server, the array of _resume tokens_ on the top level should have the same length as the number of documents returned, with the last token either being id of the last document or min promised resume token.
> 
> That's what I'm proposing. `resumeTokens` has the same length as `firstBatch`. The last resume token is the `minPromisedResumeToken`

What @cswanson310 and I discussed was slightly different in that the array of resume tokens would be exactly 1 larger than the number of documents, but I don't see an issue with wholesale replacing the last resume token with the `minPromisedResumeToken`. Drivers would still have to account for a `resumeTokens` array of length 1 when the batch is empty though.

source/change-streams/change-streams.rst
jmikola: I don't see the need to check wire version, as I expect any driver implementations could simply check the response document for `aggregate` or `getMore` for a `promisedMinResumeToken` key (if one exists).

source/change-streams/change-streams.rst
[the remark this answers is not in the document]
jmikola: > To clarify: are we guaranteed to receive back a minPromisedResumeToken on an aggregate call, and will that be useable? Would this eliminate the need for startAtOperationTime?

@daprahamian: In https://github.com/mongodb/specifications/pull/413#discussion_r231524556, @cswanson310 is proposing have 4.2+ servers always return `minPromisedResumeToken` and potentially back-porting it to earlier versions if needed. If `minPromisedResumeToken` shows up in an `aggregate` or `getMore` response, that makes `startAtOperationTime` obsolete because we now have an actual resume token; however, drivers should maintain functionality for `startAtOperationTime` in the event we're talking to an older server (we really won't know until that first `aggregate` response comes in).

> all documents in the most recent batch have been returned

@kevinAlbs: I'm curious how this would actually come up. If we're iterating through a local batch of documents (from the initial `aggregate` or a previous `getMore`), the driver is capturing the `minPromisedResumeToken` on that response and then overwriting it as we observe `_id` values of documents within that batch (if any). If at any point down the line we issue a new `getMore` or `aggregate` command, it's either because we've reached the end of a batch and need more documents or we encountered a resumable error on that `getMore` and need to issue a new `aggregate` to reconstruct the cursor. In either event, there would be no more documents left to iterate in the last local batch.

@p-mongo: On the same token (no pun intended), I don't follow how your scenario is any different. Even if documents in the change stream are being processed individually, they are obtained from the change stream cursor's local batch in a particular order. Assuming there are documents in the batch, the resume token should always be the `_id` of the most recently returned document from the change stream (based on the iterator's current position).

jmikola:
> The resumeTokens array corresponds to documents in the batch. However, the last entry might secretly be what the current PR calls the minPromisedResumeToken. This might be the same value as the resume token for the last document, but drivers won't have to care.

@kevinAlbs, @skriptble, @cswanson310: In a `getMore` response with actual documents (i.e. not an empty batch), `minPromisedResumeToken` points to an oplog position around the time of that `getMore` and fills in for the lack of any document-derived resume token. But if the batch _did_ contain documents, would the `minPromisedResumeToken` point to an oplog position before or after those changes? I would expect it to point to the change _before_ the first change document in that batch, and would therefore expect it to be first in this out-of-band array of resume tokens (instead of last). Am I mistaken?

source/change-streams/change-streams.rst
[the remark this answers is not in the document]
kevinAlbs: Good point. A change stream can't get a resumable error while returning documents in the middle of a batch and not doing any IO. However, if we provide a public change stream function like `getResumeToken`, then we need to keep track of what resume token to return.

kevinAlbs:
> > The resumeTokens array corresponds to documents in the batch. However, the last entry might secretly be what the current PR calls the minPromisedResumeToken. This might be the same value as the resume token for the last document, but drivers won't have to care.
> 
> @kevinAlbs, @skriptble, @cswanson310: In a `getMore` response with actual documents (i.e. not an empty batch), `minPromisedResumeToken` points to an oplog position around the time of that `getMore` and fills in for the lack of any document-derived resume token. But if the batch _did_ contain documents, would the `minPromisedResumeToken` point to an oplog position before or after those changes? I would expect it to point to the change _before_ the first change document in that batch, and would therefore expect it to be first in this out-of-band array of resume tokens (instead of last). Am I mistaken?

`minPromisedResumeToken` points to an oplog position after (or at) the last change (at least from our perspective - it may not have a corresponding oplog entry). It is a promise from the server that documents won't be returned before this resume token.

jmikola:
@kevinAlbs: @skriptble just explained to me why `minPromisedResumeToken` should come _after_ any change documents, rather than before. If we consider the worst case that the oplog only spans 4 hours and our change stream criteria matches only a single change near the oldest point of the oplog, the `aggregate` or `getMore` response can return that document (whose resume token will soon drop off from the oplog entirely) and a `minPromisedResumeToken` ideally pointing to the newest part of the oplog (since it knows at the time it executed `aggregate` or `getMore` that there is nothing of interest behind that point in the oplog other than the one change it's returning). If the entire oplog had no change matching our criteria, `minPromisedResumeToken` could also still point to somewhere near the latest point in the oplog.

Assuming that logic makes sense, I think my previous misunderstanding is resolved.


if an `aggregate` or `getMore` included one

source/change-streams/change-streams.rst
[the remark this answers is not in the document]
gormanb: Agreed that this is necessary to support the getResumeToken() interface. Regarding the language, I'd propose the following:

`The resume token. In the event that the iterator has returned all documents in the most recent batch, this MUST be the 'minPromisedResumeToken' for that batch. In the case where we are still iterating the batch, this MUST be the (_id) of the document most recently returned by the iterator.`

In particular, the driver's resume token should be set to 'minPromisedResumeToken' as soon as the final document in the current batch is returned, rather than the resume token of that final document.

kevinAlbs:
I have heard no objections to the new approach, so there's a new commit in this review:

- includes a public getter for the current resume token
- the resume token now tracks tokens from a "resumeTokens" array in the aggregate/getMore response.

source/change-streams/change-streams.rst
```diff
@@ -405,28 +454,45 @@ If neither ``startAtOperationTime`` nor ``resumeAfter`` are specified, and the m
...
+ChangeStream::resumeToken
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The locally cached ``ChangeStream::resumeToken`` MUST be updated by the following rules.
+
+Upon creating a ``ChangeStream``, if ``resumeAfter`` is set in ``ChangeStreamOptions``, then ``ChangeStream::resumeToken`` MUST be set to the value of ``resumeAfter``.
+
+Upon sending the initial ``aggregate`` or a ``getMore`` command, if the response includes ``resumeTokens``, the following rules apply:
+- If the response has an empty batch (no documents) then ``ChangeStream::resumeToken`` MUST be set to ``resumeTokens[0]`` (the only element in ``resumeTokens``).
+- Otherwise, when document ``i`` of the current batch is returned, ``ChangeStream::resumeToken`` MUST be set to ``resumeTokens[i]`` (the only element in ``resumeTokens``).
```
p-mongo: Wouldn't resumeTokens have more than one element in this case?

p-mongo:
Looks good otherwise. Here is the new user guidance as I understand it:

<pre>
token = nil
while true
  begin
    stream = collection.watch([], resume_after: token)
    enum = stream.to_enum
    while doc = enum.next
      process(doc)
      token = doc['_id']
    end
    token = stream.resume_token
  rescue Mongo::Error
    sleep 1
  end
end
</pre>

source/change-streams/change-streams.rst
kevinAlbs: Doh, right. Fixed.

source/change-streams/change-streams.rst
mbroadst: Agreed users should be guided to always use this accessor if its now present.

skriptble:
> Looks good otherwise. Here is the new user guidance as I understand it:
> 
> token = nil
> while true
>   begin
>     stream = collection.watch([], resume_after: token)
>     enum = stream.to_enum
>     while doc = enum.next
>       process(doc)
>       token = doc['_id']
>     end
>     token = stream.resume_token
>   rescue Mongo::Error
>     sleep 1
>   end
> end

I think we are going the route of always encouraging users to get the resume token from the change stream, so the example would change to:
```
 token = nil
 while true
   begin
     stream = collection.watch([], resume_after: token)
     enum = stream.to_enum
     while doc = enum.next
       process(doc)
       token = stream.resume_token
     end
   rescue Mongo::Error
     sleep 1
   end
 end
```

source/change-streams/change-streams.rst
jmikola: @dstorch: I agree that `startAtOperationTime` should become obsolete with `promisedMinResumeToken`, but please make sure drivers are informed when/if the server drops support for `startAtOperationTime`!

source/change-streams/change-streams.rst
```diff
@@ -165,10 +208,16 @@ Driver API
...
   interface ChangeStream extends Iterable<Document> {
     /**
-     * The resume token (_id) of the document the iterator last returned
+     * The token to be used during the resume process.
      */
     private resumeToken: Document;
 
+    /**
+     * A public getter for resumeToken. If resumeToken has not been set yet,
+     * returns a disengaged optional.
```
jmikola: According to a Google search, "disengaged optional" seems like a C++ term that implies that optional cannot be dereferenced. Is that correct?

source/change-streams/change-streams.rst
```diff
@@ -405,28 +454,45 @@ If neither ``startAtOperationTime`` nor ``resumeAfter`` are specified, and the m
...
+ChangeStream::resumeToken
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The locally cached ``ChangeStream::resumeToken`` MUST be updated by the following rules.
+
+Upon creating a ``ChangeStream``, if ``resumeAfter`` is set in ``ChangeStreamOptions``, then ``ChangeStream::resumeToken`` MUST be set to the value of ``resumeAfter``.
+
+Upon sending the initial ``aggregate`` or a ``getMore`` command, if the response includes ``resumeTokens``, the following rules apply:
+- If the response has an empty batch (no documents) then ``ChangeStream::resumeToken`` MUST be set to ``resumeTokens[0]`` (the only element in ``resumeTokens``).
+- Otherwise, when document ``i`` of the current batch is returned, ``ChangeStream::resumeToken`` MUST be set to ``resumeTokens[i]``.
```
jmikola: In https://github.com/mongodb/specifications/pull/413#discussion_r231948022, @gormanb proposed skipping the resume token of the last document in the batch in favor of the final token (i.e. `minPromisedResumeToken`).

Alternatively, that last `minPromisedResumeToken` can get picked up when the user iterates beyond that last document in the local batch, immediately before a `getMore` would be issued. In that case, there is likely no point where `getResumeToken()` would return the `minPromisedResumeToken`, as the next time the application has control would be after the `getMore` response is processed.

These two bullets under "response includes `resumeTokens`" do seem a bit inconsistent in that:

 * When the batch is empty, we immediately pick up on `resumeTokens[0]`.
 * When the batch isn't empty, we only update `ChangeStream::resumeToken` when advance the iterator.

In the empty batch case, `getResumeToken()` would immediately have a new value. That's not necessarily the case with the non-empty batch case, although perhaps that depends on the implementation details for the iterators. For instance, does the initial `watch()` method return a change stream iterator already positioned at the first document or must users advance it to the first position?

source/change-streams/change-streams.rst
mbroadst: agreed, I think you can let the TypeScript stand for itself here, `A public getter for resumeToken` is sufficient

p-mongo:
@skriptble That pseudocode probably won't work because resume token on the stream is the one after all documents. A further issue is that `enum` is currently a standard library Enumerable and probably won't lend itself easily to having an additional field tacked on for the resume token. Thus in practice we may need:

<pre>
 token = nil
 while true
   begin
     stream = collection.watch([], resume_after: token)
     stream.each_document do |doc, resume_token|
       process(doc)
       token = resume_token
     end
   rescue Mongo::Error
     sleep 1
   end
 end
</pre>

The existing `stream.to_enum` API probably will not be getting this new resume token functionality.

source/change-streams/change-streams.rst
kevinAlbs: "disengaged optional" sounds cool but I concede.

source/change-streams/change-streams.rst
kevinAlbs: > In the empty batch case, getResumeToken() would immediately have a new value. That's not necessarily the case with the non-empty batch case, although perhaps that depends on the implementation details for the iterators. For instance, does the initial watch() method return a change stream iterator already positioned at the first document or must users advance it to the first position?

Maybe I'm missing your point. But in the non-empty case, we can only update the cached resume token until after we've returned the document. This shouldn't depend on the implementation details of the iterator.

p-mongo:
Reading another comment I remembered that the driver must track resume tokens for resuming on errors, and therefore the change stream can only be iterated once. The above updated Ruby code then should indeed be the new user guidance.

source/change-streams/change-streams.rst
gormanb: The proposed usage pattern seems a bit cumbersome to me, since it effectively requires the application to consume two parallel "streams" of data from the cursor in order to remain consistent. I'm also quite skeptical that users will follow this advice, or even understand why it's necessary; it's more likely that they'll think "why do I need to call this? I already have the most recent resume token in the document I just read."

Since the `minPromisedResumeToken` is only relevant to the user in cases where the driver cannot resume internally, would it be possible to provide the `minPromisedResumeToken` to the app as part of the exception thrown by the driver on resume failure? Consuming the stream would then look like this:

```
next()
next()
next()
[batch exhausted]
[getMore request fails]
[driver tries to resume with minPromisedResumeToken, fails]
[driver throws exception, bundles minPromisedResumeToken]
[application resumes using minPromisedResumeToken]
```

This seems to me to be a much more natural usage pattern. Our guidance could then be "on failure of a changestream, the resume token returned with the exception is guaranteed to be the most recent point from which it is safe to resume. It is strongly recommended to always use this token."

kevinAlbs:
Note, the server changes are still TBD. This pull request will be updated after.

source/change-streams/change-streams.rst
jmikola: Sorry, I wasn't suggesting its removal -- just asking a clarifying question.

source/change-streams/change-streams.rst
jmikola: This might explain what I was trying to get across:

#### Example with empty batches

1. `Collection::watch()` is invoked and `aggregate` returns an empty batch with a single `minPromisedResumeToken`. The ChangeStream iterator is returned.
1. `ChangeStream::getResumeToken()` returns that `minPromisedResumeToken`, since it was picked up immediately from the `aggregate` response.
1. Accessing the current position of the ChangeStream iterator returns a disengaged optional (i.e. no document).
1. Advancing the ChangeStream iterator issues a `getMore` command, which still returns an empty batch with a single `minPromisedResumeToken`.
1. `ChangeStream::getResumeToken()` returns the `minPromisedResumeToken` from the last `getMore`.
1. Accessing the current position of the ChangeStream iterator still returns a disengaged optional (i.e. no document).

In this case of empty batches, `ChangeStream::getResumeToken()` will return the same `minPromisedResumeToken` no matter whether it is called before or after access of the ChangeStream iterator's current document. Advancing the iterator and invoking a `getMore` will certainly see the resume token updated.

#### Example with non-empty batches

1. `Collection::watch()` is invoked and `aggregate` returns a batch with two documents. The `resumeTokens` array contains the first document's resume token and a trailing `minPromisedResumeToken`. The ChangeStream iterator is returned.
1. `ChangeStream::getResumeToken()` returns nothing, since we've yet to access/return a document. (note: this seems like a subtle behavioral change)
1. Accessing the current position of the ChangeStream iterator returns the _first_ document in the batch. When this document is accessed, the internal `ChangeStream.resumeToken` property is updated.
1. `ChangeStream::getResumeToken()` now returns the resume token for the _first_ document.
1. The ChangeStream iterator is advanced.
1. Accessing the current position of the ChangeStream iterator returns the _second_ document in the batch. When this document is accessed, the internal `ChangeStream.resumeToken` property is updated. (note: I believe  @gormanb's proposal would see the `resumeToken` updated to `minPromisedResumeToken` at this point)
1. `ChangeStream::getResumeToken()` now returns the resume token for the _second_ document.
1. The ChangeStream iterator is advanced. Before the `getMore` is issued, I believe we'd want to update `resumeToken` to `minPromisedResumeToken`; however, this doesn't appear to be outlined in the spec as written (unless I'm missing something). `getMore` returns a new batch containing a few documents.
1. `ChangeStream::getResumeToken()` returns the `minPromisedResumeToken` from the previous batch, since we've yet to access/return a document from the second batch returned by `getMore`.

Is this sequence of events (for both cases) accurate?

kevinAlbs: > Example with empty batches

This is accurate.

> Example with non-empty batches
> [...]
> 2. ChangeStream::getResumeToken() returns nothing [...]

Yes, if this is the very first response. If there was a previous response (empty or non-empty) it would return the last cached resume token.

> 6. Accessing the current position of the ChangeStream iterator returns the second document in the batch. When this document is accessed, the internal ChangeStream.resumeToken property is updated. (note: I believe @gormanb's proposal would see the resumeToken updated to minPromisedResumeToken at this point)
> 7. ChangeStream::getResumeToken() now returns the resume token for the second document.

The current proposal written here (which will likely change) assumes that `resumeTokens` would have length 2 in this case. `resumeTokens[0]` is the resume token for the first document, and `resumeTokens[1]` is the `minPromisedResumeToken`. As soon as the second document is iterated, the cached resume token becomes the value of `minPromisedResumeToken`.

source/change-streams/change-streams.rst
skriptble: I think there are two use cases for users concerning change stream resume tokens: a) the user wants to resume the stream manually, and b) the user does not want to resume the stream manually.

For the second case, I think we don't need to expose resume tokens in the document, and those users might benefit from being able to remove the resume token all together (as in the case of [SERVER-37669](https://jira.mongodb.org/browse/SERVER-37669).

For the first case, users want to have the best chance of the resume token working when they use it manually. Only being able to retrieve the minPromisedResumeToken on network error might mean that the resume token the user wants to manually restart from is too old to actually use. Here's an example: Say the user has tens of thousands of change streams but changes to each happen less than once before the oplog rolls over. To help ensure that the system administrator can do rolling upgrade, the user takes each resume token and places it in a cache. With the `minPromisedResumeToken` only being accessible via exception, the user cannot retrieve it during normal operation. Therefore, when the system administrator upgrades the application servers, the resume tokens for older change streams won't be valid and will cause the application to create new change streams. If the oplog is large, this means that the server will now have to search through the oplog, even though we already know that there are no entries that will meet the criteria. This may cause performance issue.

For both use cases, the resume token is control information, it allows manipulation of the change stream itself, whereas the rest of the information is the data and metadata about the change. I think it would be advantageous to have a separate path for users to retrieve the control information, and for that path to be consistent.

There is an additional problem with all of the proposed methods, though. In some drivers, if an empty batch is returned from the server, the driver does not return control to the user. This means that even if we do expose a `getResumeToken` method, it will still be difficult for users to retrieve and use that token. I think we need another approach to allowing users to retrieve this information, but that is a driver API change. When we add this new `minPromisedResumeToken`, we should also tell users to not use the resume token in the document and instead retrieve the resume token via that new driver API.

source/change-streams/change-streams.rst
jmikola: Per our offline conversation, I think this can be resolved.

@kevinAlbs explained that, for non-empty batches, the server will likely replace the last resume token in the `resumeTokens` array with the `minPromisedResumeToken` instead of the last document's actual resume token, since it is both functionally equivalent and more efficient. Therefore, the size of `resumeTokens` is either one (in the case of an empty batch) or exactly the same size as the number of change documents. We agreed that the spec would do well to have a FAQ entry added to document this, but we don't need to describe this behavior in detail within this section of the spec.

source/change-streams/change-streams.rst
```diff
@@ -547,6 +611,14 @@ In the above example, not sending ``startAtOperationTime`` will result in the ch
...
 the server will know to include changes from that previous point in time.
 
+---------------------------------------------------------------
+Why do we provide a public getter for the current resume token?
+---------------------------------------------------------------
+
+Users should be able to implement their own custom resume logic (e.g. save the resume token in a file and restart later). `ChangeStream::resumeToken`_ explains how we set the current ``resumeToken`` to use on resume. When possible, resume tokens are taken from the ``resumeTokens`` of an aggregate/getMore response instead of the ``_id`` of a ``ChangeDocument``. The resume tokens in ``resumeTokens`` may point to more recent entries in the oplog. This makes it much less likely that a change stream cursor will roll off of the oplog and helps performance of resumes.
+
+The ``resumeTokens`` array may not be conveniently accessible to clients (outside of using APM), so the public getter provides an easy way to implement custom resume logic.
+
```
jmikola: TODO: Add FAQ entry explaining how `minPromisedResumeToken` appears in the `resumeTokens` array for both empty and non-empty batches. In particular, it will replace the resume token of the final document for non-empty batches.

See https://github.com/mongodb/specifications/pull/413#discussion_r233237081 for more context.

source/change-streams/change-streams.rst
asya999: resuming with a token explicitly by the application is the only way to resume if the application crashes.  it seems required that there be a way to get real most recent resume time.

source/change-streams/change-streams.rst
```diff
@@ -98,7 +98,7 @@ If the aggregation with ``$changeStream`` specified completes successfully, the
     /**
      * Describes the type of operation represented in this change notification.
      */
-    operationType: "insert" | "update" | "replace" | "delete" | "invalidate" | "drop" | "dropDatabase";
+    operationType: "insert" | "update" | "replace" | "delete" | "invalidate" | "drop" | "dropDatabase" | "rename";
```
kevinAlbs: Drive-by fix, "rename" is another operationType that can be returned by a change stream.

source/change-streams/change-streams.rst
```diff
@@ -167,7 +207,13 @@ Driver API
...
      * The resume token (_id) of the document the iterator last returned
      */
-    private resumeToken: Document;
+    private documentResumeToken: Document;
+
+    /**
+     * The most recent postBatchResumeToken returned in an aggregate or
+     * getMore response. For pre-4.2 versions of MongoDB, this remains unset.
+     */
+    private postBatchResumeToken: Document;
```
kevinAlbs: I'm contradicting the decision made in [these comments](https://github.com/mongodb/specifications/pull/413#discussion_r231523141) to better distinguish the resume token sourced from the document _id and the resume token sourced from postBatchResumeToken.

source/change-streams/change-streams.rst
kevinAlbs: Added FAQ entry explaining why we need to expose postBatchResumeToken.

kevinAlbs:
This is ready for review again.

- The server decision is to add a `postBatchResumeToken` to aggregate/getMore responses in a change stream. (Note, older in this PR that was referred to as `minPromisedResumeToken`). We've decided not to resolve SERVER-37669 now.
- We discovered it wasn't sufficient for all drivers to provide a `ChangeStream::getResumeToken()` to allow users to retrieve the latest resume token. This change clarifies that synchronous drivers must provide some kind of `tryNext` change stream cursor API. This allows iterating a tailable awaitData cursor without blocking on repeated getMore attempts, so users can retrieve a resume token in between getMores, even if they return empty batches.

source/change-streams/change-streams.rst
```diff
@@ -414,25 +462,96 @@ Once a ``ChangeStream`` has encountered a resumable error, it MUST attempt to re
...
 
 
+Exposing All Resume Tokens
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+:since: 4.2
+
+Users can retrieve the ``documentResumeToken`` by inspecting the _id on each ``ChangeDocument``. But since MongoDB 4.2, aggregate and getMore responses also include a ``postBatchResumeToken``. Drivers use one or the other when automatically resuming, as described in `Resume Process`_.
+
+Drivers MUST expose a mechanism to retrieve the same resume token that would be used to automatically resume. It MUST be possible to use this mechanism after iterating every document *and* after every response to a getMore (including empty responses). Drivers have two options to implement this.
```
skriptble: `It MUST be possible to use this mechanism ... after every response to a getMore (including empty responses).` I think this is too strict. The way we are planning on making it possible for users to retrieve the tokens in the Go driver is by them providing a timeout via `context.Context`, which means there might be a number of `getMore` commands that are sent before control is returned to the user.

source/change-streams/change-streams.rst
```diff
@@ -414,25 +462,96 @@ Once a ``ChangeStream`` has encountered a resumable error, it MUST attempt to re
...
+A possible interface for this callback MAY look like:
+
+.. code:: typescript
+
+  interface ChangeStream extends Iterable<Document> {
+    /**
+     * Returns a resume token that should be used to resume after the most
+     * recently returned change.
+     */
+    public onResumeTokenChanged(ResumeTokenCallback:(Document resumeToken) => void);
```
p-mongo: I thought we decided for this option to return change documents and resume tokens together?

source/change-streams/change-streams.rst
```diff
@@ -481,16 +600,20 @@ Imagine a scenario in which a user wants to process each change to a collection
...
-    resumeToken = localChange['_id']
 
   try:
-      for change in db.collection.watch([...], resumeAfter=resumeToken):
-          persistToLocalStorage(change)
-          processChange(change)
+      change_stream = db.collection.watch([...], resumeAfter=resumeToken)
+      while True:
+          change = change_stream.try_next()
+          persistResumeTokenToLocalStorage(change_stream.get_resume_token())
```
p-mongo: I think this needs to be called twice, before and after iterating the changes. Use case which I believe the current code does not cover: there is a single change made 1 hour ago, the document's resume token in this case should be 1 hour old but post-batch resume token could be minutes old.

source/change-streams/change-streams.rst
kevinAlbs: Ah I see. I've change this to require drivers provide a way to return control back to the user when iterating a change stream, even when no documents are getting returned.

source/change-streams/change-streams.rst
kevinAlbs: This is an example and not the required interface. I chose to leave it less strictly specified because I think it might be preferable in some drivers to have a separate event just for resume token updates. For example, in NodeJS, I've checked with @mbroadst that it would not be necessary to have both the document and resume token as part of the callback arguments.

source/change-streams/change-streams.rst
kevinAlbs: Perhaps I'm misunderstanding. In this example I'm using `try_next`. That should not block until a document is iterated.

source/change-streams/change-streams.rst
```diff
@@ -155,6 +155,46 @@ If the aggregation with ``$changeStream`` specified completes successfully, the
...
+         ns: String,
+         id: Int64,
+         nextBatch: Array<ChangeStreamDocument>
+      },
+      operationTime: Timestamp,
+      $clusterTime: Document,
+      /**
+       * postBatchResumeToken is returned in MongoDB 4.2 and later.
+       */
+      postBatchResumeToken: Document
```
cswanson310: @gormanb didn't we move this within the 'cursor' subobject?

source/change-streams/change-streams.rst
```diff
@@ -155,6 +155,46 @@ If the aggregation with ``$changeStream`` specified completes successfully, the
...
+      ok: 1,
+      cursor: {
+         ns: String,
+         id: Int64,
+         nextBatch: Array<ChangeStreamDocument>
+      },
+      operationTime: Timestamp,
+      $clusterTime: Document,
+      /**
+       * postBatchResumeToken is returned in MongoDB 4.2 and later.
```
cswanson310: Fine with me to merge this, but we may have to ammend this to MongoDB 4.2 and later, MongoDB 4.0.x and later, etc. depending on how backports go.

source/change-streams/change-streams.rst
cswanson310: Seems like this can be resolved?

kevinAlbs: Oh right. TIL you can resolve conversations in GitHub PRs.

source/change-streams/change-streams.rst
kevinAlbs: Good catch! That was the most recent decision. Updated.

source/change-streams/change-streams.rst
```diff
@@ -414,25 +462,96 @@ Once a ``ChangeStream`` has encountered a resumable error, it MUST attempt to re
...
     - The driver MUST NOT set a ``resumeAfter`` key.
-    - In this case, the ``ChangeStream`` will return all changes that occurred after the specified ``startAtOperationTime``.
 - Else:
 
-    - The driver MUST execute the known aggregation command.
-    - The driver MUST specify a ``resumeAfter`` with the last known ``resumeToken``.
-    - The driver MUST NOT set a ``startAtOperationTime``.
-    - If a ``startAtOperationTime`` key was part of the original aggregation command, the driver MUST remove it.
-    - In this case, the ``ChangeStream`` will return notifications starting with the oplog entry immediately *after* the provided token.
+  - The driver MUST execute the known aggregation command.
```
jmikola: If I'm understanding this list correctly, the `Else` just above this line is for a case where `resumeAfter` was specified by the user. In the lines that follow, it looks like `resumeAfter` is only ever populated with either the cached `postBatchResumeToken` or `documentResumeToken`. Am I missing something, or do the lines below only take effect _after_ the initial aggregate is executed with the user-specified options?

If this is simply the resume process, then does it matter at all if `resumeAfter` was originally specified by the user ("and ``resumeAfter`` is not specified" in the third list item ~12 lines above).

source/change-streams/change-streams.rst
```diff
@@ -414,25 +462,96 @@ Once a ``ChangeStream`` has encountered a resumable error, it MUST attempt to re
...
 
 
+Exposing All Resume Tokens
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+:since: 4.2
+
+Users can retrieve the ``documentResumeToken`` by inspecting the _id on each ``ChangeDocument``. But since MongoDB 4.2, aggregate and getMore responses also include a ``postBatchResumeToken``. Drivers use one or the other when automatically resuming, as described in `Resume Process`_.
+
+Drivers MUST expose a mechanism to retrieve the same resume token that would be used to automatically resume. It MUST be possible to use this mechanism after iterating every document. It MUST be possible for users to use this mechanism periodically even when no documents are getting returned. Drivers have two options to implement this.
```
jmikola: Suggest adding "(i.e. `getMore` has returned empty batches)" after "are getting returned", unless you think it's clear as-is.

source/change-streams/change-streams.rst
```diff
@@ -414,25 +462,96 @@ Once a ``ChangeStream`` has encountered a resumable error, it MUST attempt to re
...
+- If the ``ChangeStream`` has no cached ``documentResumeToken`` or ``postBatchResumeToken`` this returns an unset optional.
+- If the ``ChangeStream`` has a cached ``postBatchResumeToken`` and has returned all documents in the most recent batch (or the most recent batch was empty) this returns the ``postBatchResumeToken``.
+- Otherwise, this returns the cached ``documentResumeToken``.
+
+This MUST be implemented in synchronous drivers. This MAY be implemented in asynchronous drivers.
+
+Option 2: Event Emitted for Resume Token
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Allow users to set a callback to listen for new resume tokens. The exact interface is up to the driver, but it MUST meet these criteria:
```
jmikola: "MUST meet the following criteria:"

source/change-streams/change-streams.rst
```diff
@@ -414,25 +462,96 @@ Once a ``ChangeStream`` has encountered a resumable error, it MUST attempt to re
...
+     */
+    public onResumeTokenChanged(ResumeTokenCallback:(Document resumeToken) => void);
+  }
+
+This MUST not be implemented in synchronous drivers. This MAY be implemented in asynchronous drivers.
+
+Not Blocking on Iteration
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Synchronous drivers MUST provide a way to iterate a change stream without blocking until a change document is returned. This MUST give the user an opportunity to get the most up-to-date resume token, even when change stream continues to receive empty batches in getMore responses. This allows users to call ``ChangeStream::getResumeToken()`` after iterating every document and periodically when no documents are getting returned.
```
jmikola: "even when *the* change stream"

source/change-streams/change-streams.rst
```diff
@@ -414,25 +462,96 @@ Once a ``ChangeStream`` has encountered a resumable error, it MUST attempt to re
...
+
+Although the implementation of tailable awaitData cursors is not specified, this MAY be implemented with a ``tryNext`` method on the change stream cursor.
```
jmikola: perhaps add "or separate methods for advancing the iterator and accessing the current element (if any)" -- I know that's what we do in PHP and I believe libmongoc is similar.

kevinAlbs: libmongoc does have a `mongoc_cursor_next` (which iterates and returns the document or NULL) and `mongoc_cursor_current` (which returns the current document). But the problem is that advancing the iterator should not block indefinitely waiting for a document to be returned, not that we need separate methods for iteration and access.

`mongoc_cursor_next` only blocks for a maximum of waiting on one getMore response. If the most recent getMore returns an empty batch, then `mongoc_cursor_next` just returns NULL. Therefore, it's possible to get the updated resume tokens during change stream iteration with `mongoc_cursor_next` without blocking on documents getting returned.

source/change-streams/change-streams.rst
kevinAlbs: Doh, fixed.

source/change-streams/change-streams.rst
kevinAlbs: Done.

source/change-streams/change-streams.rst
kevinAlbs: Agreed. Done.

source/change-streams/change-streams.rst
kevinAlbs: Good catch!

The else is for:
change stream has cached documentResumeToken
OR
change stream was specified with resumeAfter
OR
max wire version < 7

But yes, there should be another elseif to account for a resumeAfter specified when no postBatchResumeToken or documentResumeToken have been cached.

kevinAlbs: Actually... I feel like there might be a way to better organize these conditions. I'll take another look in a bit.

kevinAlbs: Ok, hopefully this is clearer (albeit verbose).

source/change-streams/change-streams.rst
```diff
@@ -10,7 +10,7 @@ Change Streams
 :Type: Standards
 :Minimum Server Version: 3.6
 :Last Modified: July 30, 2018
-:Version: 1.5.1
+:Version: 1.6.1
```
jmikola: Last modified date should probably also be bumped.

source/change-streams/change-streams.rst
```diff
@@ -462,28 +462,37 @@ Once a ``ChangeStream`` has encountered a resumable error, it MUST attempt to re
...
-    - The driver MUST NOT set a ``resumeAfter`` key.
-- Else:
+  - The driver MUST set ``resumeAfter`` to the value of the originally used ``startAfter``.
+  - The driver MUST NOT set ``startAfter``. The driver MUST remove it from the original aggregation command.
+  - The driver MUST NOT set ``startAtOperationTime``.
+- Else if the ``ChangeStream`` was created with ``resumeAfter``:
+
+  - The driver MUST set ``resumeAfter`` to the value of the originally used ``resumeAfter``.
+  - The driver MUST NOT set ``startAfter``.
+  - The driver MUST NOT set ``startAtOperationTime``.
```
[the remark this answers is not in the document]
kevinAlbs: Right, if you specify any two of: {`startAtOperationTime`, `resumeAfter`, `startAfter`} in the options used to start a change stream, we'll return back a server error.

This only applies to the aggregate used in resume.

source/change-streams/change-streams.rst
```diff
@@ -462,28 +462,37 @@ Once a ``ChangeStream`` has encountered a resumable error, it MUST attempt to re
...
+- Else if the ``ChangeStream`` has a saved operation time (either from an originally specified ``startAtOperationTime`` or saved from the original aggregation) and the max wire version is >= ``7``:
 
-  - The driver MUST execute the known aggregation command.
-  - If the ``ChangeStream`` has returned all documents in the most recent batch (or the most recent batch was empty) and has a cached ``postBatchResumeToken``:
+  - The driver MUST not set ``resumeAfter``.
+  - The driver MUST NOT set ``startAfter``.
+  - The driver MUST set ``startAtOperationTime`` to the value of the originally used ``startAtOperationTime`` or the one saved from the original aggregation.
```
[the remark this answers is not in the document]
kevinAlbs: Yes, we'll save the operation time from the aggregate response when:
- None of ``startAtOperationTime``,  ``resumeAfter``, ``startAfter`` were specified in the ``ChangeStreamOptions``.
- The max wire version is >= ``7``.
- The initial ``aggregate`` response did not include a ``documentResumeToken`` (indicating no results)
- The initial ``aggregate`` response did not include a ``postBatchResumeToken``.

But this brings up a good point. The ``startAtOperationTime`` section was out of date. I updated it.

source/change-streams/change-streams.rst
kevinAlbs: Done.

source/change-streams/change-streams.rst
```diff
@@ -398,14 +446,19 @@ startAtOperationTime
...
 
-If neither ``startAtOperationTime`` nor ``resumeAfter`` are specified, and the max wire version is >= ``7``, and the initial ``aggregate`` command does not return a resumeToken (indicating no results), the ``ChangeStream`` MUST save the ``operationTime`` from the initial ``aggregate`` command when it returns.
+The ``ChangeStream`` MUST save the ``operationTime`` from the initial ``aggregate`` response when the following critera are met:
+
+- None of ``startAtOperationTime``,  ``resumeAfter``, ``startAfter`` were specified in the ``ChangeStreamOptions``.
+- The max wire version is >= ``7``.
+- The initial ``aggregate`` response did not include a ``documentResumeToken`` (indicating no results).
+- The initial ``aggregate`` response did not include a ``postBatchResumeToken``.
 
 resumeAfter
```
kevinAlbs: As a side-note, I think we should add a `startAfter` section (since we have separate sections for `resumeAfter` and `startAtOperationTime`), which could possibly capture the info in [this comment](https://jira.mongodb.org/browse/SERVER-35030?focusedCommentId=1962023&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-1962023). If you agree, I'll create a follow-up spec ticket after this is merged.

PR #414, opened 2018-11-13
Add uri options tests

saghm:


source/uri-options/tests/README.rst
```diff
@@ -0,0 +1,55 @@
...
+  - ``host``: Included for compatibility with the Connection String spec tests. This will always be ``~``.
+  - ``port``: Included for compatibility with the Connection String spec tests. This will always be ``~``.
+- ``auth``: An object containing the following keys:
+  - ``username``: Included for compatibility with the Connection String spec tests. This will always be ``~``.
+  - ``password``: Included for compatibility with the Connection String spec tests. This will always be ``~``.
+  - ``db``: Included for compatibility with the Connection String spec tests. This will always be ``~``.
+- ``options``: An object containing key/value pairs for each parsed query string
+  option.
+
+If a test case includes a null value for one of these keys (e.g. ``auth: ~``,
```
prashantmital: @saghm can comment as to why we have the keys within `hosts`/`auth` present when all of them are `null`? In the connection string spec, the top-level key is itself set to `null` when it doesn't contain any information. To re-use the test-runner we have in python, I would need to do the same here. That said, I am not averse to modifying the runner to be able to handle this pattern provided we have a good rationale for stroing `null`-valued keys.

source/uri-options/tests/single-threaded-options.json
```diff
@@ -0,0 +1,25 @@
+{
+  "tests": [
+    {
+      "description": "Valid options specific to single-threaded drivers are parsed correctly",
```
prashantmital: Can we add a note to the `README.rst` about what to do about this test in a driver that is not multi-threaded? I'm guessing we just don't add this `.json` file to the test directory, but it might save someone a bit of pain down the line to be explicit about it.

source/uri-options/tests/README.rst
saghm: Interesting, I must have missed the specs that had the hosts field as null; I'll update this to do that

saghm: Just pushed the changes

source/uri-options/tests/read-preference-options.json
```diff
@@ -0,0 +1,65 @@
...
+        "readPreferenceTags": [
+          "dc:ny,rack",
+          "dc:ny"
+        ],
+        "maxStalenessSeconds": "120"
+      }
+    },
+    {
+      "description": "Arbitrary string read preference does not cause a warning",
+      "uri": "mongodb://example.com/?readPreference=arbitraryButStillValid",
```
prashantmital: Can you clarify why this does not raise a warning? It seems odd that we do validate the `readPreferenceTags` (since an invalid tag raises a warning) but don't do the same for the `readPreference` itself.

source/uri-options/tests/tls-options.json
```diff
@@ -0,0 +1,82 @@
+{
+  "tests": [
+    {
+      "description": "Valid required tls options are parsed correctly",
+      "uri": "mongodb://example.com/?tls=true&tlsCAFilePath=ca.pem&tlsClientCertFilePath=cert.pem&tlsClientKeyPassword=hunter2",
+      "valid": true,
+      "warning": false,
```
prashantmital: Should this raise a warning as well since these files don't exist/aren't readable? Or maybe we can add dummy files to the test artifacts?

source/uri-options/tests/read-preference-options.json
saghm: This was more of a judgment call than anything else. In other places (like read/write concerns), we explicitly don't throw errors on unknown levels to make things more forward compatible. Read preferences are a bit different given that they're only sent to mongos's (and not mongod's), so some level of validation will need to be done in the driver, but I wasn't sure if this was something that should be done in the connection string parsing or not. The distinction between this and the tags is that the warning for tags is about the _format_ of the value, not the value itself. While we might someday add an additional string that is a valid read preference, we aren't likely to allow tag sets that are anything besides `string -> string` maps.

That being said, given that this is a judgment call, maybe the best solution is to just remove this test entirely and not mandate either a warning or a lack thereof. Assuming that's acceptable to you, I can push this change tomorrow morning.

source/uri-options/tests/tls-options.json
prashantmital: Fair enough. Looks like there needs to be better separation of URI parsing and URI validation in the Python driver.

source/uri-options/tests/read-preference-options.json
prashantmital: I had thought about the point you make about the warning in tags stemming from the formatting of content and not the content itself. I think that's a reasonable rule to live by as far as this spec is concerned. 

To avoid confusion, can we change the 'Accepted values' for `readPreference` to say something like 

> any string; currently supported read preferences are 'primary', 'primaryPreferred', ...

The current value does seem to imply that the `readPreference` value _needs_ to be one of the strings mentioned there or it will be invalid. Conversely, we can also simply point the reader to the read preference spec to see for themselves what read preferences are supported.

prashantmital: > Conversely, we can also simply point the reader to the read preference spec to see for themselves what read preferences are supported.

Actually, this doesn't seem possible since the spec makes no mention of the camel case names of read preferences: https://github.com/mongodb/specifications/blob/master/source/driver-read-preferences.rst#mode

saghm: Changing it to a link sounds good to me; I'll update the doc accordingly

source/uri-options/tests/tls-options.json
saghm: I think don't think it's necessarily an issue for the Python driver to eagerly check the files (although it will make the testing a bit more difficult); I mostly just don't want to mandate that all drivers have to do it eagerly

source/uri-options/tests/read-preference-options.json
saghm: Hmm, good catch. It does seem to mention them below in the "communication with mongos" section. I'll update the language to reflect that camelCase is expected.

source/uri-options/tests/single-threaded-options.json
p-mongo: If multi-threaded clients do not recognize single-threaded options, shouldn't multi-threaded clients warn when such options are encountered? And therefore there should be multi-threaded client tests that expect these warnings?

source/uri-options/tests/tls-options.json
p-mongo: I agree that parsing a URI should be separate from verifying whether a file that is referenced exists (though as a user I would certainly like the driver to inform me with a warning or an error when a file I specified does not exist).

source/uri-options/tests/single-threaded-options.json
saghm: The connection string spec already specified that drivers warn on unrecognized options; for multithreaded options, this will just be like any other unrecognized option.

The README mentions that drivers need to skip optional tests for options they don't implement. In the case of this test, I think the cleanest solution is just for multi-threaded drivers not to include the file in the repo, although if a driver author prefers to skip it programmatically, that's fine too.

source/uri-options/tests/tls-options.json
saghm: Yep! The test isn't saying that the driver can't warn at all, just that it doesn't need to be done as part of URI parsing. I'm guessing that most drivers already raise an error in the code path where the file is actually read (i.e. when the TLS connection is initiated), so forcing them to add code to check for the file and raise a warning before this seems redundant at best.

source/uri-options/tests/connection-options.yml
```diff
@@ -0,0 +1,87 @@
+tests:
+    -
+        description: "Valid connection and timeout options are parsed correctly"
+        uri: "mongodb://example.com/?appname=URI-OPTIONS-SPEC-TEST&connectTimeoutMS=20000&heartbeatFrequencyMS=5000&localThresholdMS=3000&maxIdleTimeMS=50000&replicaSet=uri-options-spec&retryWrites=true&serverSelectionTimeoutMS=15000&socketTimeoutMS=7500"
+        valid: true
+        warning: false
+        hosts: ~
```
p-mongo: I also do not understand the logic behind specifying a null host here. If the idea is to reuse connection string spec runner with the expectation that that runner parses the host, then the host here should be "example.com", and this test should fail as written. If the idea is to ignore hosts when running URI options tests then they should not be specified in the spec tests.

source/uri-options/tests/single-threaded-options.json
saghm: I think this spec is explicitly not about how each driver should handle an option that it doesn't support, but just how it should handle certain ones that it does support. I'm not fond of the idea of introducing changes to the format of the connection string spec tests (which currently have no mechanism for differentiating between single- and multi-threaded drivers) and therefore requiring spec test runner changes; because the only tests that would require those sort of changes are not in the scope of the spec in the first place, I'd much rather give driver authors the ability to fix the problem non-programmatically.

source/uri-options/tests/connection-options.json
```diff
@@ -0,0 +1,147 @@
+{
+  "tests": [
+    {
+      "description": "Valid connection and timeout options are parsed correctly",
+      "uri": "mongodb://example.com/?appname=URI-OPTIONS-SPEC-TEST&connectTimeoutMS=20000&heartbeatFrequencyMS=5000&localThresholdMS=3000&maxIdleTimeMS=50000&replicaSet=uri-options-spec&retryWrites=true&serverSelectionTimeoutMS=15000&socketTimeoutMS=7500",
```
behackett: This one is a challenge for drivers without a connection pool because of maxIdleTimeMS. Maybe put tests for things like that in a connection-pool-options.json and let PHP and Perl skip the whole file?

source/uri-options/tests/tls-options.json
behackett: I think the problem @prashantmital might be hitting is the Python driver validates that the files exist and are readable. Another tricky part of this is the file paths are obviously different on Unix and Windows.

Adding empty files to the test suite seems reasonable, since we won't need an absolute file path (avoids the Windows problem) and the contents don't matter (we're testing parsing the options, not validating their contents). That would solve the Python problem and any other drivers that do the same sort of validation. There doesn't need to be a mandate that all drivers do that extra validation PyMongo does.

prashantmital: I would be in favor of adding in the empty files like @behackett suggested.

prashantmital:
@saghm can we add another test to test the 'last entry wins' parsing logic? For instance, `zlibcompression=3&zlibCompressionLevel=4&zlibCompressionLevel=8` should result in `zlibCompressionLevel=8`. This would also nail down the warning behavior (should drivers raise a warning?) for this case.

source/uri-options/tests/tls-options.json
saghm: I'll add them

source/uri-options/tests/connection-options.json
saghm: Will do!

source/uri-options/tests/read-preference-options.json
jmikola: Sorry for being a bit late to the thread here, but I would argue that drivers should enforce unknown read preference modes. I don't think this prior comment from @saghm was entirely accurate:

> Read preferences are a bit different given that they're only sent to mongos's (and not mongod's), so some level of validation will need to be done in the driver, but I wasn't sure if this was something that should be done in the connection string parsing or not.

Drivers do use read preferences locally to perform [server selection](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#read-preference), and that spec outlines the five supported modes.

For prior art, I can tell you that both libmongoc and PHP report errors for unsupported modes:

 * https://github.com/mongodb/mongo-c-driver/blob/1.13.0/src/libmongoc/src/mongoc/mongoc-uri.c#L1318
 * https://github.com/mongodb/mongo-php-driver/blob/1.5.3/src/MongoDB/ReadPreference.c#L80

IMO, validating that the mode string is one of five possible values (after canonicalization; note that libmongoc and PHP use `strcasecmp()`) is sensible to avoid errors down the line.

saghm: > I don't think this prior comment from @saghm was entirely accurate

I think we're saying the same thing? I was saying that since we don't just forward read preferences to all server types (like we do with read/write concerns) and actually use them internally, the driver will need to validate them. I'm not arguing that drivers shouldn't validate them, which would be impossible; I'm just saying that I'm not convinced that the spec should enforce that the drivers have to validate them _here_.

jmikola: Sounds good. I likely focused too much on the "only sent to mongos" bit of your last comment and didn't fully process the "validation will need to be done in the driver" bit.

If this spec does not intend to dictate read preference validation, would it be preferable to simply omit any tests for invalid mode strings? As-is, the "Arbitrary string read preference does not cause a warning" test may be requiring a driver to _not_ validate read preferences during URI parsing.

saghm: Yes, that's what we decided; check the current diff, the test is already gone

source/uri-options/tests/tls-options.json
```diff
@@ -2,7 +2,7 @@
   "tests": [
     {
       "description": "Valid required tls options are parsed correctly",
-      "uri": "mongodb://example.com/?tls=true&tlsCAFilePath=ca.pem&tlsClientCertFilePath=cert.pem&tlsClientKeyPassword=hunter2",
+      "uri": "mongodb://example.com/?tls=true&tlsCAFile=ca.pem&tlsClientCertFilePath=cert.pem&tlsClientKeyPassword=hunter2",
```
prashantmital: You missed a path:`"uri": "mongodb://example.com/?tls=true&tlsCAFile=ca.pem&tlsClientCertFile=cert.pem&tlsClientKeyPassword=hunter2"`

saghm: Good catch; thanks!

prashantmital:
Python POC has been updated with the new TLS option names.

source/uri-options/tests/single-threaded-options.json
jmikola: I assume this conversation was resolved, but I don't think it's a problem. Single-threaded drivers will likewise need to skip the connection pool test file.

source/uri-options/tests/tls-options.yml
```diff
@@ -0,0 +1,64 @@
...
+            username: ~
+            password: ~
+            db: ~
+        options:
+            tls: "true"
+            tlsCAFile: "ca.pem"
+            tlsClientCertFile: "cert.pem"
+            tlsClientKeyPassword: "hunter2"
+    -
+        description: "tlsAllowInvalidCertificates is parsed correctly"
```
jmikola: Should there be error tests for these options as well? Looking back, most other test files asserted valid parsing for all related options in the first test and then included several tests that testing individual parse failures.

source/uri-options/tests/tls-options.yml
```diff
@@ -0,0 +1,64 @@
+tests:
+    -
+        description: "Valid required tls options are parsed correctly"
```
jmikola: I assume testing that SRV schemes default tls to true may be outside the scope of this spec? I imagine that might also be a bit complicated, as you'd have to ensure the SRV record itself was valid just to run such a test.

source/uri-options/tests/concern-options.json
```diff
@@ -0,0 +1,78 @@
...
+        "username": null,
+        "password": null,
+        "db": null
+      },
+      "options": {
+        "w": "arbitraryButStillValid"
+      }
+    },
+    {
+      "description": "Invalid wTimeoutMS causes a warning",
```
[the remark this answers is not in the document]
saghm: For wTimeoutMS, I think it's fine to leave it off to allow the C driver (and the drivers that wrap it, if they also implement this suite) to more easily run the tests.

Do you think it would be better to explicitly test for all of those things? I'm not sure if there are any other existing deviations like the C driver one for wTimeoutMS that we would want to accommodate or not.

source/uri-options/tests/connection-options.yml
```diff
@@ -0,0 +1,76 @@
...
+        auth:
+            username: ~
+            password: ~
+            db: ~
+        options:
+            appname: "URI-OPTIONS-SPEC-TEST"
+            connectTimeoutMS: "20000"
```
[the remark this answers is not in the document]
saghm: This was an intentional decision, although it could have gone either way. Basically, my view is that it's better to assume the minimum amount of work drivers might do here rather than the maximum; from the standpoint of the spec, all that matters is that correct values are accepted and incorrect values are rejected, not that the conversions are done immediately.

If the consensus is that it would be better to put in the actual types instead of strings, I'm fine with updating all the tests to do that.

source/uri-options/tests/connection-pool-options.yml
```diff
@@ -0,0 +1,32 @@
...
+        uri: "mongodb://example.com/?maxIdleTimeMS=50000"
+        valid: true
+        warning: false
+        hosts: ~
+        auth:
+            username: ~
+            password: ~
+            db: ~
+        options:
+            appname: "URI-OPTIONS-SPEC-TEST"
```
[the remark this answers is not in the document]
saghm: This was an oversight from when I factored out maxIdleTimeMS into its own file; sorry for the confusion.

source/uri-options/tests/read-preference-options.yml
```diff
@@ -0,0 +1,40 @@
+tests:
+    -
+        description: "Valid connection and timeout options are parsed correctly"
```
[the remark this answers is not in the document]
saghm: Fixed; sorry for the oversight

source/uri-options/tests/tls-options.yml
saghm: Given that these all accept arbitrary strings, what would an example of an invalid option be? Due to an earlier request, PEM files for the valid names were added to accommodate drivers that immediately check for the existence of the files, but I'd prefer not to mandate that drivers have to check the filesystem during URI parsing.

source/uri-options/tests/tls-options.yml
saghm: I'm not quite sure what that test would even look like in the current format. I think that marking that `tls` is expected to be true based on `mongodb+srv` assumes that drivers implement the decision about whether to enable TLS implicitly in the URI parsing, which I don't think is something we want to mandate in the spec.

source/uri-options/tests/README.rst
```diff
@@ -0,0 +1,58 @@
...
+array of test case objects, each of which have the following keys:
+
+- ``description``: A string describing the test.
+- ``uri``: A string containing the URI to be parsed.
+- ``valid``: A boolean indicating if the URI should be considered valid. 
+  This will always be true, as the Connection String spec tests the validity of the structure, but 
+  it's still included to make it easier to reuse the connection string spec test runners that 
+  drivers already have.
+- ``warning``: A boolean indicating whether URI parsing should emit a warning.
+- ``hosts``: An array of host objects, each of which have the following keys:
```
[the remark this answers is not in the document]
saghm: Fixed

source/uri-options/tests/connection-pool-options.yml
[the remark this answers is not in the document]
saghm: Fixed

source/uri-options/tests/concern-options.json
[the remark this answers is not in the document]
jmikola: Testing out-of-range (at least for smaller values; not trying to test 32-bit or 64-bit limits) for numeric values seems sensible and not too difficult. I'm only aware of `wTimeoutMS` and `-1` for libmongoc, so I think the `wTimeoutMS` test could just use `-2` and document with a YAML comment that we're only doing so to account for libmongoc-based drivers supporting a legacy value.

source/uri-options/tests/connection-options.yml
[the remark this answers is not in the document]
jmikola: I vote for using correct types in the tests, since we already have prior art for doing so. I assume it will not cause problems for other drivers, and some could even take advantage of correct types in the test format to assert that they convert types correctly (I imagine some are already doing that in the connection string spec runner).

source/uri-options/tests/tls-options.yml
jmikola: Does `tlsAllowInvalidCertificates` accept an arbitrary string? The tls path options certainly do, but this is a boolean option. I expect drivers could check for errors if a value such as "maybe" were passed for a boolean option (e.g. `tlsAllowInvalidCertificates`, `retryWrites`).

source/uri-options/tests/tls-options.yml
jmikola: Fair enough. Drivers could very well enforce that default when resolving SRV records and incorporate it into the URI options returned from the records.

Also, per discussion in #drivers Slack last week, I think a few drivers actually try to resolve SRV records earlier than when we'd logically expect MongoClient to perform _any_ IO (and initialize SDAM), so best to leave this untested here for maximum compatibility. Otherwise, we might require a valid SRV record exist just to run what should be unit tests.

source/uri-options/tests/tls-options.yml
saghm: I'll add tests to assert that non-boolean values for `tlsAllowInvalidCertificates`, ``tlsAllowInvalidHostnames`, and `tlsInsecure` cause warnings.

source/uri-options/tests/tls-options.yml
saghm: Given the discussion, I'm going to keep things as they are now in the tests with regards to SRV (i.e. not testing it in this spec).

source/uri-options/tests/concern-options.json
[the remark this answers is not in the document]
saghm: Testing that `-2` gives a warning seems reasonable to me, so I'll add that for each of the tests it makes sense for

source/uri-options/tests/connection-options.yml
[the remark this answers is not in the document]
saghm: Okay, I'll update them to use proper types

source/uri-options/tests/tls-options.yml
saghm: Implemented

source/uri-options/tests/connection-options.yml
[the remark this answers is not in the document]
saghm: Implemented

source/uri-options/tests/connection-options.yml
```diff
@@ -0,0 +1,105 @@
...
+    -
+        description: "Non-numeric socketTimeoutMS causes a warning"
+        uri: "mongodb://example.com/?socketTimeoutMS=invalid"
+        valid: true
+        warning: true
+        hosts: ~
+        auth: ~
+        options: {}
+    -
+        description: "Too low socketTimeoutMS causes a warning"
```
prashantmital: @saghm any reason for this being the only test to be missing the `options` key? It isn't hard to handle but the inconsistency doesn't make sense to me.

saghm: Looks like an oversight; I'll add it now

source/uri-options/tests/tls-options.yml
```diff
@@ -0,0 +1,89 @@
...
+        uri: "mongodb://example.com/?tls=true&tlsCAFile=ca.pem&tlsCertificateKeyFile=cert.pem&tlsCertificateKeyPassword=hunter2"
```
prashantmital: This should be`tlsCertificateKeyFilePassword=hunter2`

source/uri-options/tests/auth-options.json
```diff
@@ -0,0 +1,20 @@
...
+      "uri": "mongodb://example.com/?authMechanism=GSSAPI&authMechanismProperties=SERVICE_NAME:other,CANONICALIZE_HOST_NAME:true&authSource=$external",
+      "valid": true,
+      "warning": false,
+      "hosts": null,
+      "auth": null,
+      "options": {
+        "authmechanism": "GSSAPI",
+        "authMechanismProperties": {
+          "SERVICE_NAME": "other",
+          "CANONICALIZE_HOST_NAME": "true"
```
prashantmital: Why is this string `'true'` instead of boolean `true`?

source/uri-options/tests/compression-options.yml
```diff
@@ -0,0 +1,44 @@
+tests:
+    -
+        description: "Valid compression options are parsed correctly"
+        uri: "mongodb://example.com/?compressors=zlib&zlibCompressionLevel=9"
+        valid: true
+        warning: false
+        hosts: ~
+        auth: ~
+        options:
+            compressors: "zlib"
```
prashantmital: Since `compressors` is a 

> comma separated list of strings, e.g. “snappy,zlib”

I think this should be 1-member list with `"zlib"` as its only member.

source/uri-options/tests/tls-options.json
```diff
@@ -0,0 +1,105 @@
...
+      "uri": "mongodb://example.com/?tls=true&tlsCAFile=ca.pem&tlsCertificateKeyFile=cert.pem&tlsCertificateKeyFilePassword=hunter2",
+      "valid": true,
+      "warning": false,
+      "hosts": null,
+      "auth": null,
+      "options": {
+        "tls": true,
+        "tlsCAFile": "ca.pem",
+        "tlsCertificateKeyFile": "cert.pem",
+        "tlsCertificateKeyPassword": "hunter2"
```
prashantmital: This should be `"tlsCertificateKeyFilePassword"`.

source/uri-options/tests/read-preference-options.json
```diff
@@ -0,0 +1,47 @@
...
+      "description": "Valid read preference options are parsed correctly",
+      "uri": "mongodb://example.com/?readPreference=primaryPreferred&readPreferenceTags=dc:ny,rack:1&maxStalenessSeconds=120&readPreferenceTags=dc:ny",
+      "valid": true,
+      "warning": false,
+      "hosts": null,
+      "auth": null,
+      "options": {
+        "readPreference": "primaryPreferred",
+        "readPreferenceTags": [
+          "dc:ny,rack",
```
prashantmital: This does not reflect the `rack:1` from the YAML file. You might need to regenerate the JSONs.

source/uri-options/tests/connection-options.yml
```diff
@@ -0,0 +1,106 @@
...
+        valid: true
+        warning: false
+        hosts: ~
+        auth: ~
+        options:
+            appname: "URI-OPTIONS-SPEC-TEST"
+            connectTimeoutMS: 20000
+            heartbeatFrequencyMS: 5000
+            localThresholdMS: 3000
+            maxIdleTimeMS: 50000
```
prashantmital: This is missing from the connection string.

source/uri-options/tests/compression-options.yml
prashantmital: Nevermind, I suppose this isn't necessary (since it could just be a string with comma-separated values).

source/uri-options/tests/auth-options.json
saghm: I wasn't sure if we should be prescriptive about how/when drivers parse the authMechanismProperties from strings. Since it's a lot easier to go from the boolean to the string than from the string to the boolean though, I'll update this to use the boolean.

source/uri-options/tests/read-preference-options.json
saghm: Forgot to update the JSON files before pushing; I'll do that now

saghm:
@prashantmital are these working with the Python driver now?

source/uri-options/tests/single-threaded-options.json
```diff
@@ -0,0 +1,24 @@
...
+      "auth": null,
+      "options": {
+        "serverSelectionTryOnce": false
+      }
+    },
+    {
+      "description": "Invalid serverSelectionTryOnce causes a warning",
+      "uri": "mongodb://example.com/?serverSelectionTryOnce=invalid",
+      "valid": true,
+      "warning": false,
```
jyemin: Should be `warning: true`

source/uri-options/tests/compression-options.yml
```diff
@@ -0,0 +1,44 @@
...
+            zlibCompressionLevel: 9
+    -
+        description: "Invalid compressors cause a warning"
```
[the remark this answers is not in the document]
saghm: The Connection String spec specifically mandates that drivers should warn rather than error on invalid options (see the last part of [this section](https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.rst#values) for the part about values and the third question [here](https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.rst#q-a) for the rationale. This spec is specifically designed to be compatible with the Connection String spec, so we followed those guidelines.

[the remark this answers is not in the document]
jyemin: Thanks.  It looks like the Java driver is inconsistent about this.  It warns for certain things like booleans with odd values, but errors for others, like integers.  We'll fix it up.

PR #415, opened 2018-11-13
Only examine non-unknown server descriptions to determine incompatibi…

p-mongo:
…lity

Without this, a replica set topology where the client received a response from primary and is waiting for a response from secondary is considered incompatible, since the secondary's wire version is taken to be 0.

https://jira.mongodb.org/browse/RUBY-1583

### DIFF
--- a/source/server-discovery-and-monitoring/tests/rs/incompatible_ghost.yml
+++ b/source/server-discovery-and-monitoring/tests/rs/incompatible_ghost.yml
@@ -1,4 +1,4 @@
-description: "Compatible topology with incompatible ghost"
+description: "Incompatible ghost"
 uri: "mongodb://a,b/?replicaSet=rs"
 phases:
   - responses:
--- a/source/server-discovery-and-monitoring/tests/rs/incompatible_ghost.yml
+++ b/source/server-discovery-and-monitoring/tests/rs/incompatible_ghost.yml
@@ -1,0 +1,32 @@
+description: "Compatible topology with incompatible ghost"
+uri: "mongodb://a,b/?replicaSet=rs"
+phases:
+  - responses:
+      -
+        - "a:27017"
+        - ok: 1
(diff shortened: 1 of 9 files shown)


p-mongo:
I am wondering if only data bearing servers should be examined, analogously with logical session timeout. The current implementation in Ruby driver considers all non-unknown servers as specified in this PR.

ajdavis:
I think that Unknown servers should *not* be considered but arbiters *should* be considered when the driver is determining whether it is compatible with the topology. This change LGTM. Is there a prose change we should make in the SDAM spec?

ajdavis:
RSOther and RSGhost members might be real members of the replica set, or they might be on their way to become members. So let's consider their wire versions when determining driver compatibility. Considering all server types but Unknown is simple and consistent with some of the other spec rules.

So, the existing test and prose change LGTM. You told me in a meeting that there are associated Ruby new spec tests that are incoming, let's add them to this PR>

p-mongo:
Added tests for arbiters, ghosts and others.

p-mongo:
Spec ticket: https://jira.mongodb.org/browse/SPEC-1172

source/server-discovery-and-monitoring/tests/rs/incompatible_ghost.yml
```diff
@@ -0,0 +1,32 @@
+description: "Compatible topology with incompatible ghost"
```
ajdavis: Remember to update: "Incompatible"

source/server-discovery-and-monitoring/tests/rs/incompatible_other.json
```diff
@@ -0,0 +1,54 @@
+{
+  "description": "Compatible topology with incompatible other",
```
ajdavis: Incompatible

PR #416, opened 2018-11-14
SPEC-1140 Support mongos pinning for sharded transactions

ShaneHarvey:
Jira: https://jira.mongodb.org/browse/SPEC-1140

This change adds the "Mongos Pinning for Sharded Transactions" section to the transactions spec and updates the tests for 4.2 sharded transactions.

I have run the tests against a 4.1.6 sharded cluster with two mongoses, see [PYTHON-1673](https://jira.mongodb.org/browse/PYTHON-1673).

Some open questions:
- [X] Done added two prose tests for unpinning: A buggy driver could still satisfy these spec tests by always selecting a single mongos for all operations on a session -- regardless if those operations are part of a transaction or not. Is it possible to test that drivers un-pinn a session properly? Eg:
  ```python
  with client.start_session() as s:
    # All operations in this block use the same mongos.
    with s.start_transaction():
      client.test.test.insert_one({}, session=s) 
      client.test.test.find_one({}, session=s)
    # Can we test that this next operation is not pinned to the previous mongos?
    doc = client.test.test.find_one({}, session=s)
  ```
- [X] Done updated the server selection spec: Should we update the server selection spec to mention that sharded transactions do not go through the normal server selection process? Similar to cursors: https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#cursors
- [X] Done: Should we specify what happens the pinned mongos becomes unavailable? Does any spec cover the same for cursors?

### DIFF
--- a/source/transactions/tests/README.rst
+++ b/source/transactions/tests/README.rst
@@ -157,6 +157,7 @@ Load each YAML (or JSON) file using a Canonical Extended JSON parser.
 
 Then for each element in ``tests``:
 
+#. If the``skipReason`` field is present, skip this test completely.
 #. Create a MongoClient and call
    ``client.admin.runCommand({killAllSessions: []})`` to clean up any open
    transactions from previous test failures.
--- a/source/transactions/tests/README.rst
+++ b/source/transactions/tests/README.rst
@@ -291,3 +291,59 @@ afterClusterTime
 A ``readConcern.afterClusterTime`` value of ``42`` in a command-started event
 is a fake cluster time. Drivers MUST assert that the actual command includes an
 afterClusterTime.
+
+Mongos Pinning Prose Tests
+==========================
+
+The following tests ensure that a ClientSession is properly unpinned after
+a sharded transaction. Initialize these tests with a MongoClient connected
+to multiple mongoses.
+
+These tests use a cursor's address field to track which server an operation
+was run on. If this is not possible in your driver, use command monitoring
+instead.
+
+#. Test that starting a new transaction on a pinned ClientSession unpins the
+   session and normal server selection is performed for the next operation.
+
+   .. code:: python
+
+      @require_server_version(4, 1, 5)
+      @require_mongos_count_at_least(2)
+      def test_unpin_for_next_transaction(self):
+        client = MongoClient(mongos_hosts)
+        with client.start_session() as s:
+          # Session is pinned to Mongos.
+          with s.start_transaction():
+              client.test.test.insert_one({}, session=s)
+
+          addresses = set()
+          for _ in range(20):
+            with s.start_transaction():
+              cursor = client.test.test.find({}, session=s)
+              assert next(cursor)
+              addresses.add(cursor.address)
+
+          assert len(addresses) > 1
+
+#. Test non-transaction operations using a pinned ClientSession unpins the
+   session and normal server selection is performed.
+
+   .. code:: python
+
+      @require_server_version(4, 1, 5)
+      @require_mongos_count_at_least(2)
+      def test_unpin_for_non_transaction_operation(self):
+        client = MongoClient(mongos_hosts)
+        with client.start_session() as s:
+          # Session is pinned to Mongos.
+          with s.start_transaction():
+              client.test.test.insert_one({}, session=s)
+
+          addresses = set()
+          for _ in range(20):
+            cursor = client.test.test.find({}, session=s)
+            assert next(cursor)
+            addresses.add(cursor.address)
+
+          assert len(addresses) > 1
--- a/source/transactions/tests/README.rst
+++ b/source/transactions/tests/README.rst
@@ -159,13 +159,36 @@ Then for each element in ``tests``:
 
 #. Create a MongoClient and call
    ``client.admin.runCommand({killAllSessions: []})`` to clean up any open
-   transactions from previous test failures. The command will fail with message
-   "operation was interrupted", because it kills its own implicit session. Catch
-   the exception and continue.
+   transactions from previous test failures.
+
+   To workaround `SERVER-38335`_, ensure this command does not send
+   an implicit session, otherwise the command will fail with an
+   "operation was interrupted" error because it kills itself and (on a sharded
+   cluster) future commands may fail with:
+   "Encountered error from localhost:27217 during a transaction :: caused by :: operation was interrupted".
+
+   If your driver cannot run this command without an implicit session, then
+   either skip this step and live with the fact that previous test failures
+   may cause later tests to fail or use the `killAllSessionsByPattern` command
+   instead. During each test record all session ids sent to the server and at
+   the end of each test kill all the sessions ids (using a different session):
+
+   .. code:: python
+
+      # Be sure to use a distinct session to avoid "operation was interrupted".
+      session_for_cleanup = client.start_session()
+      recorded_session_uuids = []
+      # Run test case and record session uuids...
+      client.admin.runCommand({
+          'killAllSessionsByPattern': [
+              {'lsid': {'id': uuid}} for uuid in recorded_session_uuids]},
+          session=session_for_cleanup)
 
    - When testing against a sharded cluster, create a list of MongoClients that
-     are directly connected to each mongos. Run the killAllSessions command on
-     ALL mongoses.
+     are directly connected to each mongos. Run the `killAllSessions`
+     (or `killAllSessionsByPattern`) command on ALL mongoses.
+
+   .. _SERVER-38335: https://jira.mongodb.org/browse/SERVER-38335
 
 #. Create a collection object from the MongoClient, using the ``database_name``
    and ``collection_name`` fields of the YAML file.
@@ -315,7 +338,7 @@ instead.
         with client.start_session() as s:
           # Session is pinned to Mongos.
           with s.start_transaction():
-              client.test.test.insert_one({}, session=s)
+            client.test.test.insert_one({}, session=s)
 
           addresses = set()
           for _ in range(20):
@@ -338,7 +361,7 @@ instead.
         with client.start_session() as s:
           # Session is pinned to Mongos.
           with s.start_transaction():
-              client.test.test.insert_one({}, session=s)
+            client.test.test.insert_one({}, session=s)
 
           addresses = set()
           for _ in range(20):
--- a/source/transactions/tests/README.rst
+++ b/source/transactions/tests/README.rst
@@ -149,7 +149,7 @@ ensure that no new bugs have been introduced related to arbiters.)
 
 A driver that implements support for sharded transactions MUST also run these
 tests against a MongoDB sharded cluster with multiple mongoses and
-**server version 4.1.5 or later**. Including multiple mongoses (and
+**server version 4.1.6 or later**. Including multiple mongoses (and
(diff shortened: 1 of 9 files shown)


source/transactions/transactions.rst
```diff
@@ -344,9 +344,9 @@ progress" state, then Drivers MUST raise an error containing the message
 startTransaction SHOULD report an error if the driver can detect that
 transactions are not supported by the deployment. A deployment does not
 support transactions when the deployment does not support sessions, or
-maxWireVersion < 7, or the topology type is Sharded, see `How to Check
-Whether a Deployment Supports
-Sessions <https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst#how-to-check-whether-a-deployment-supports-sessions>`__.
+maxWireVersion < 7, or the maxWireVersion < 8 and the topology type is Sharded,
+see `How to Check Whether a Deployment Supports Sessions
+ <https://github.com/mongodb/specifications/blob/master/source/sessions driver-sessions.rst#how-to-check-whether-a-deployment-supports-sessions>`_.
```
p-mongo: The slash to space change here seems bogus.

ajdavis:
LGTM.

Response to open questions:

1. Since the cursor must remember its server, then in PyMongo and probably most drivers there should be a way to compare cursor addresses. Perhaps a prose test (for what they're worth....) like?:
```
with client.start_session() as s:
  client.test.test.insert_one({}, session=s)
  addresses = set()
  for _ in range(10000):
    cursor = client.test.test.find({}, session=s)
    assert next(cursor)
    addresses.add(cursor.address)
  assert len(addresses) > 1
```

2. Yes, let's update the server selection spec.

3. No, I don't think anything specifies cursor behavior when the cursor's server is type Unknown. Yes, let's specify what happens when the mongos becomes unavailable: each operation in the transaction continues to try using the mongos anyway. I propose that you specify that whatever error type the driver throws in the cursor scenario (server selection error, connection failure, or whatever) SHOULD also be the error type in the transaction scenario, but let's leave this vague.

source/transactions/tests/README.rst
```diff
@@ -153,6 +162,7 @@ Then for each element in ``tests``:
    transactions from previous test failures. The command will fail with message
    "operation was interrupted", because it kills its own implicit session. Catch
    the exception and continue.
+##. When testing against a sharded cluster run this command on ALL mongoses.
```
jyemin: Using which MongoClient?  Seems like we'd have to create one MongoClient that's directly connected to each mongos server in the seed list.

source/transactions/transactions.rst
```diff
@@ -733,6 +733,19 @@ format:
...
+
+The server supports sharded transactions starting in MongoDB 4.2 (
+maxWireVersion 8). The server requires that drivers MUST send all commands for
+a single transaction to the same mongos.
+
+After the driver selects a mongos for the first command within a transaction,
+the driver MUST pin the ClientSession to the selected mongos. Drivers MUST
+send all commands that are part of the same transaction, including
+commitTransaction and abortTransaction and any retries thereof, to the
+same mongos.
```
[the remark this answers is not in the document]
ajdavis: Interesting. Let me review the state diagram:

https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#clientsession-changes

After "abort" we will never abort or commit again, so we could unpin. After "commit", we can keep committing, so we can't unpin, as you say.

I think commit and abort should have the same effect on pinning, therefore I think *neither* should unpin.

Instead, any non-transaction operation in a session MUST unpin.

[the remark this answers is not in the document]
jyemin: +1

source/transactions/tests/README.rst
```diff
@@ -268,3 +291,59 @@ afterClusterTime
...
+
+Mongos Pinning Prose Tests
+==========================
+
+The following tests ensure that a ClientSession is properly unpinned after
+a sharded transaction. Initialize these tests with a MongoClient connected
+to multiple mongoses.
+
+These tests use a cursor's address field to track which server an operation
+was run on. If this is not possible in your driver, use command monitoring
```
mbroadst: Out of curiosity, if it's possible to write these using command monitoring why not write them as YAML tests?

mbroadst: Oh, missed Jesse's comment above. I'm guessing it's not prose because we don't have an easy way to specify a loop in the test runner format. Resolving the comment.

ShaneHarvey:
@EshaMaharishi can you give this a review?

source/transactions/tests/README.rst
```diff
@@ -163,11 +200,18 @@ Then for each element in ``tests``:
    into the test collection, using writeConcern "majority".
 #. If ``failPoint`` is specified, its value is a configureFailPoint command.
    Run the command on the admin database to enable the fail point.
+
+   - When testing against a sharded cluster run this command on ALL mongoses.
```
jmikola: I assume this requires that drivers either:

 * provide some API to return all nodes within the topology (i.e. all known mongos nodes in the case of a shard cluster) and allow commands to explicitly use a selected mongos
 * have the test suite parse the connection string and extract each mongos host so that individual connections can be made for the purpose of configuring fail points

source/transactions/tests/commit.json
```diff
@@ -350,6 +350,7 @@
     },
     {
       "description": "write concern error on commit",
+      "skipReason": "SERVER-37458 Mongos does not yet support writeConcern on commit",
```
jmikola: Is this merely a comment or are drivers expected to actually skip any and all tests that include a `skipReason` field? If the latter, that may be worth specifying in the test readme.

source/transactions/transactions.rst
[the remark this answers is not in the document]
jmikola: @jyemin Looks like language for unpinning has been added in the paragraph below. Can this comment thread be resolved?

source/transactions/tests/README.rst
```diff
@@ -140,19 +143,53 @@ Use as integration tests
...
 
 Run a MongoDB replica set with a primary, a secondary, and an arbiter,
-**server version 4.0.0-rc4 or later**. (Including a secondary ensures that
+**server version 4.0.0 or later**. (Including a secondary ensures that
 server selection in a transaction works properly. Including an arbiter helps
 ensure that no new bugs have been introduced related to arbiters.)
 
+A driver that implements support for sharded transactions MUST also run these
+tests against a MongoDB sharded cluster with multiple mongoses and
+**server version 4.1.5 or later**. Including multiple mongoses (and
```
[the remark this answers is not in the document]
ShaneHarvey: 4.1.5 was the first server version that could pass these tests.

source/transactions/tests/README.rst
ShaneHarvey: Yes I think the second one makes the most sense. AFAIK there's no way to discover new mongoses from a cluster expect through the seed addresses in a connection string.

source/transactions/transactions.rst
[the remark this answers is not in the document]
ShaneHarvey: @EshaMaharishi recovery token will be added in https://jira.mongodb.org/browse/SPEC-1168.

source/transactions/tests/transaction-options.json
```diff
@@ -147,7 +147,7 @@
       "description": "transaction options inherited from client",
       "clientOptions": {
         "w": 1,
-        "readConcernLevel": "local"
```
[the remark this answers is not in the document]
ShaneHarvey: Thanks, reverted these changes.

source/transactions/tests/README.rst
```diff
@@ -140,19 +143,53 @@ Use as integration tests
...
-   transactions from previous test failures. The command will fail with message
-   "operation was interrupted", because it kills its own implicit session. Catch
-   the exception and continue.
+   transactions from previous test failures.
+
+   To workaround `SERVER-38335`_, ensure this command does not send
+   an implicit session, otherwise the command will fail with an
+   "operation was interrupted" error because it kills itself and (on a sharded
+   cluster) future commands may fail with:
+   "Encountered error from localhost:27217 during a transaction :: caused by :: operation was interrupted".
```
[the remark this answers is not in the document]
ShaneHarvey: Added "with and error similar too:"

source/transactions/tests/commit.json
ShaneHarvey: It is already described:
```
  - ``skipReason``: Optional, string describing why this test should be
    skipped.
```

source/transactions/tests/pin-mongos.yml
```diff
@@ -0,0 +1,290 @@
+# Test that all the operations go to the same mongos.
+#
+# The assertion is implicit: that all the read operations succeed,
+# as if the driver does not properly pin to a mongos then one of them
+# will eventually be sent to a different server, which is unaware of the
+# transaction, and it will report an error.
```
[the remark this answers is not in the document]
ShaneHarvey: Hopefully it's more clear now. Can you take another look?

source/transactions/tests/commit.json
ShaneHarvey: Also added:
```
#. If the``skipReason`` field is present, skip this test completely.
```

ShaneHarvey:
Updated in respond to Esha's and Jeremy's reviews. I believe I've resolved all the outstanding comments. For anyone interested in the python POC see [PYTHON-1673](https://jira.mongodb.org/browse/PYTHON-1673).

source/transactions/tests/pin-mongos.yml
```diff
@@ -0,0 +1,298 @@
...
+# operations in a transaction will eventually be sent to a different mongos,
+# which is unaware of the transaction, and the mongos will return a command
+# error. An example of such an error is:
+# {
+#   'ok': 0.0,
+#   'errmsg': 'cannot continue txnId -1 for session 28938f50-9d29-4ca5-8de5-ddaf261267c4 - 47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU= with txnId 1',
+#   'code': 251,
+#   'codeName': 'NoSuchTransaction',
+#   'errorLabels': ['TransientTransactionError']
+# }
```
EshaMaharishi: In this example, the "different" mongos has seen a higher transaction (txnId 1) than what's in the request (txnId -1), which I wouldn't expect to happen normally.

If you want, you could give an example of the typical case, where the request's transaction number is higher than what the "different" mongos has seen (which is why I'd expect to see the 'missing startTransaction' error).

ShaneHarvey: This is the error I get back from the server when attempting to continue a transaction on a different Mongos. I think the problem here is just that the server's error message is unclear.  txnId -1  is not txnNumber that the driver sent to the server, it's the server's txnId value for this session.

EshaMaharishi: Hm, you're right, the source code on mongos is a bit odd. I'll file a server ticket for it.

EshaMaharishi: Filed https://jira.mongodb.org/browse/SERVER-38420.

ShaneHarvey: Thanks!

PR #418, opened 2018-11-15
SPEC-1169: Aggregate with $out should not specify batchSize

jmikola:
https://jira.mongodb.org/browse/SPEC-1169

### DIFF
--- a/source/crud/crud.rst
+++ b/source/crud/crud.rst
@@ -12,7 +12,7 @@ Driver CRUD API
 :Status: Approved
 :Type: Standards
 :Minimum Server Version: 2.6
-:Last Modified: October 25, 2018
+:Last Modified: November 15, 2018
 
 .. contents::
 
@@ -217,6 +217,10 @@ Read
      * This option is sent only if the caller explicitly provides a value. The default is to not send a value.
      * If specified, drivers SHOULD apply this option to both the original aggregate command and subsequent
      * getMore operations on the cursor.
+     *
+     * Drivers MUST NOT specify a batchSize of zero in an aggregate command that includes an $out stage, as
+     * that will prevent the pipeline from executing. Drivers SHOULD leave the cursor.batchSize command option
+     * unset in an aggregate command that includes an $out stage.
      *
      * @see https://docs.mongodb.com/manual/reference/command/aggregate/
      */
@@ -1722,6 +1726,7 @@ Q: Where is ``singleBatch`` in FindOptions?
 Changes
 =======
 
+* 2018-11-15: Aggregate commands with an $out stage should not specify batchSize
 * 2018-10-25: Note how results are backed for aggregate, distinct, and find operations
 * 2018-07-25: Added upsertedCount to UpdateResult.
 * 2018-06-07: Deprecated the count helper. Added the estimatedDocumentCount and countDocuments helpers.


source/crud/crud.rst
```diff
@@ -218,6 +218,10 @@ Read
      * If specified, drivers SHOULD apply this option to both the original aggregate command and subsequent
      * getMore operations on the cursor.
      *
+     * Drivers MUST NOT specify a batchSize of zero in an aggregate command that includes an $out stage, as
+     * that will prevent the pipeline from executing. Drivers SHOULD leave the cursor.batchSize command option
```
saghm: Out of curiosity, what's the rationale for using "SHOULD" here instead of "MUST"?

source/crud/crud.rst
```diff
@@ -218,6 +218,10 @@ Read
...
+     * unset in an aggregate command that includes an $out stage.
+     *
```
jmikola: I considered adding another paragraph with:

> Drivers that setup a cursor to be executed upon iteration against the $out collection MAY apply the batchSize option to that cursor, as that has no impact on the execution of the pipeline.

But that's outside the scope of the SPEC ticket. I also don't have any experience with drivers that do return cursors for `$out` pipelines and have no idea if they use batchSize in that additional capacity. I was careful to only refer to the initial `aggregate` command in the PR text (without any impact on other users of batchSize), so hopefully we can leave it at that.

saghm: Ah, right, I forgot about how the spec gives drivers leeway on returning a cursor after a `$out` aggregation

behackett: @jyemin - I believe the Java and/or .NET drivers return cursors that work with $out?

source/crud/crud.rst
jmikola: There is technically no harm in specifying batchSize >= 1, as that ensures the initial `aggregate` command does start pipeline execution to generate at least one result document. Of course, with `$out`, we'll expect `aggregate` to return an empty batch of results and a cursor ID of zero (since there really are no result documents to return).

When using `batchSize` of zero, `aggregate` returns an empty batch (of course) with a non-zero cursor ID. That would require iteration via `getMore` (presumably with batchSize >= 1) to ensure at least one document is requested from its result, which will trigger the pipeline to execute.

jmikola: If that didn't answer the question, I was open to the fact that drivers could have passed [SPEC-826](https://jira.mongodb.org/browse/SPEC-826) simply by including a check for zero but still allowing larger batch sizes. So if drivers are currently allowing a batchSize >= 1 and that doesn't conflict with the spec test you added, they can continue to do so. Otherwise, SHOULD might as well read as MUST per our existing RFC conventions.

source/crud/crud.rst
jyemin: In Java, Collection.aggregate returns an AggregateIterable.  AggregateIterable has a toCollection method that returns void, which throws if the pipeline doesn't terminate with $out.  It also has an iterator method that returns a MongoCursor, which for a $out aggregation as a cursor over the collection that you $out-ed to.

jyemin: I think in the latter case the batchSize is ignored for the aggregate, and used for establishing the cursor.

jmikola: > I think in the latter case the batchSize is ignored for the aggregate, and used for establishing the cursor.

@jyemin: Are you comfortable with the language in the PR? It doesn't sound as if it conflicts with Java's current behavior.

jyemin: Yeah, seems fine.

PR #419, opened 2018-11-16
test(mongodb+srv): add tests for encoded userinfo with srv uris

mbroadst:


mbroadst:
Updated to just have a single test that covers all cases

PR #420, opened 2018-11-16
SPEC-884 add database-level aggregation

kvwalker:
[SPEC-884](https://jira.mongodb.org/browse/SPEC-884)

### DIFF
--- a/source/crud/crud.rst
+++ b/source/crud/crud.rst
@@ -207,7 +207,7 @@ Read
      * Note: result iteration should be backed by a cursor. Depending on the implementation,
      * the cursor may back the returned Iterable instance or an iterator that it produces.
      *
-     * @see https://docs.mongodb.com/manual/reference/method/db.aggregate/
+     * @see https://docs.mongodb.com/manual/reference/command/aggregate/#dbcmd.aggregate
      */
     aggregate(pipeline: Document[], options: Optional<AggregateOptions>): Iterable<Document>;
 
@@ -689,7 +689,7 @@ Because of this anomaly in the wire protocol, it is up to the driver to enforce 
 Database-level aggregation
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The server supports several collection-less aggregation source stages like ``$currentOp`` and ``$listLocalSessions``. The shell allows users to run ``db.aggregate``, which uses a value of 1 as the collection name and executes the operation against the entire database. Drivers support for database-level aggregation will allow users to receive a cursor from these collection-less aggregation source stages.
+The server supports several collection-less aggregation source stages like ``$currentOp`` and ``$listLocalSessions``. The database aggregate command requires a collection name of 1 for collection-less source stages. Drivers support for database-level aggregation will allow users to receive a cursor from these collection-less aggregation source stages.
 
 Write
 -----
--- a/source/crud/crud.rst
+++ b/source/crud/crud.rst
@@ -198,6 +198,21 @@ Read
 
   }
 
+  interface Database {
+
+    /**
+     * Runs an aggregation framework pipeline on the database for pipeline stages
+     * that do not require an underlying collection, such as $currentOp and $listLocalSessions.
+     *
+     * Note: result iteration should be backed by a cursor. Depending on the implementation,
+     * the cursor may back the returned Iterable instance or an iterator that it produces.
+     *
+     * @see https://docs.mongodb.com/manual/reference/method/db.aggregate/
+     */
+    aggregate(pipeline: Document[], options: Optional<AggregateOptions>): Iterable<Document>;
+
+  }
+
   class AggregateOptions {
 
     /**
@@ -674,20 +689,7 @@ Because of this anomaly in the wire protocol, it is up to the driver to enforce 
 Database-level aggregation
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-  interface Db {
-
-    /**
-     * Runs an aggregation framework pipeline on the database for pipeline stages
-     * that do not require an underlying collection, such as $currentOp and $listLocalSessions.
-     *
-     * Note: result iteration should be backed by a cursor. Depending on the implementation,
-     * the cursor may back the returned Iterable instance or an iterator that it produces.
-     *
-     * @see https://docs.mongodb.com/manual/reference/method/db.aggregate/
-     */
-    aggregate(pipeline: Document[], options: Optional<AggregateOptions>): Iterable<Document>;
-
-  }
+The server supports several collection-less aggregation source stages like ``$currentOp`` and ``$listLocalSessions``. The shell allows users to run ``db.aggregate``, which uses a value of 1 as the collection name and executes the operation against the entire database. Drivers support for database-level aggregation will allow users to receive a cursor from these collection-less aggregation source stages.
 
 Write
 -----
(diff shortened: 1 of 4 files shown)


source/crud/crud.rst
```diff
@@ -670,6 +670,25 @@ The OP_QUERY wire protocol only contains a numberToReturn value which drivers mu
...
+  interface Db {
+
+    /**
+     * Runs an aggregation framework pipeline on the database for pipeline stages
+     * that do not require an underlying collection, such as $currentOp and $listLocalSessions.
+     *
+     * Note: result iteration should be backed by a cursor. Depending on the implementation,
+     * the cursor may back the returned Iterable instance or an iterator that it produces.
+     *
+     * @see https://docs.mongodb.com/manual/reference/method/db.aggregate/
```
mbroadst: This should continue to point to the [command documentation](https://docs.mongodb.com/manual/reference/command/aggregate/), rather than this link which is to the mongo shell documentation (let's _definitely_ not use the shell as reference 😄). The command documentation gives guidance around passing `1` in as the value for `aggregate` for "collection agnostic commands"

source/crud/crud.rst
```diff
@@ -670,6 +670,25 @@ The OP_QUERY wire protocol only contains a numberToReturn value which drivers mu
 
 Because of this anomaly in the wire protocol, it is up to the driver to enforce the user-specified limit. Each driver MUST keep track of how many documents have been iterated and stop iterating once the limit has been reached. When the limit has been reached, if the cursor is still open, a driver MUST kill the cursor.
 
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+Database-level aggregation
```
mbroadst: Are there YAML tests on the way?

xdg:
Everyone chiming in -- thanks for your enthusiasm, but this is not ready for team review; this is just for Katherine and me to iterate on for now.

source/crud/crud.rst
```diff
@@ -670,6 +685,12 @@ The OP_QUERY wire protocol only contains a numberToReturn value which drivers mu
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The server supports several collection-less aggregation source stages like ``$currentOp`` and ``$listLocalSessions``. The shell allows users to run ``db.aggregate``, which uses a value of 1 as the collection name and executes the operation against the entire database. Drivers support for database-level aggregation will allow users to receive a cursor from these collection-less aggregation source stages.
```
xdg: Instead of defining this in terms of the shell, I would say that the database aggregate command requires a collection name of 1 for collection-less source stages.

source/crud/crud.rst
```diff
@@ -670,6 +670,25 @@ The OP_QUERY wire protocol only contains a numberToReturn value which drivers mu
...
 
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+Database-level aggregation
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+  interface Db {
+
+    /**
+     * Runs an aggregation framework pipeline on the database for pipeline stages
+     * that do not require an underlying collection, such as $currentOp and $listLocalSessions.
```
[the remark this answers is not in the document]
xdg: I think "such as" is a useful pointer for a fresh reader to find out what this means.  It may go stale, but slowly.

xdg:
I'm happy with the text, so I think you can move on to the tests.

source/crud/tests/README.rst
```diff
@@ -57,6 +57,8 @@ Each YAML file has the following keys:
 
     - ``name``: The name of the operation as defined in the specification.
```
xdg: I think you should mention that 'db-aggregate' means to run the database-level aggregate command.

source/crud/tests/db/db-aggregate.json
```diff
@@ -0,0 +1,147 @@
...
+                "command.lsid": false
+              }
+            }
+          ]
+        }
+      },
+      "outcome": {
+        "result": [
+          {
+            "command": {
```
jmikola: This looks like an APM expectation. Is this really an aggregation result due to the database-level aggregation being the only in-progress operation at the time it is executed and therefore it only reports itself?

source/crud/tests/db/db-aggregate.yml
```diff
@@ -0,0 +1,58 @@
...
+
+tests:
+    -
+        description: "Aggregate with $currentOp"
+        operation:
+            name: db-aggregate
+            database_name: admin
+            arguments:
+                pipeline:
+                  - $currentOp: {allUsers: true, idleConnections: true}
```
jmikola: Would `false` for both of these be more robust on the off chance other operations are running on the server? Given that the expected output is only this database-level aggregation, it seems reasonable to ignore other things if possible.

That might also be more portable, since `allUsers: true` requires an `inprog` privilege. I assume most drivers would have that permission during testing, but it's still an extra loose end.

source/crud/tests/db/db-aggregate.yml
```diff
@@ -0,0 +1,58 @@
...
+                  - $match:
+                      command.aggregate: {$eq: 1}
+                  - $project: {command: 1}
+                  - $project: {command.lsid: false}
```
jmikola: Are two `$project` stages necessary so that `{command: 1}` can implicitly exclude everything _except_ `_id` and `command`, and then this stage excludes the `command` document's `lsid` field alone?

Small nit, but I think it'd be consistent to use either integers or booleans for all [projection syntax](https://docs.mongodb.com/manual/reference/operator/aggregation/project/index.html). I'd suggest `{command.lsid: 0}` here since we're already using `1` above (and it's more concise).

source/crud/tests/db/db-aggregate.json
kvwalker: Yes, it is an aggregation result that is only reporting itself.

source/crud/tests/db/db-aggregate.yml
kvwalker: Yes, the two `$project` stages are necessary for the reason you identified.

source/crud/tests/db/db-aggregate.yml
```diff
@@ -0,0 +1,58 @@
+data: []
+
+tests:
+    -
+        description: "Aggregate with $currentOp"
+        operation:
+            name: db-aggregate
+            database_name: admin
```
ShaneHarvey: The transaction spec test format allows running methods on a database with a slightly different format. https://github.com/mongodb/specifications/blob/master/source/transactions/tests/run-command.yml#L12-L22

I think it would be beneficial to standardize on the existing format:
```
database_name: &database_name "transaction-tests"
...
tests:
...
        operation:
            name: aggregate   # method name to run
            object: database  # object on which method is defined
            arguments: ...
```

This way drivers can share more code between test parsers/runners.

jmikola: @kvwalker: Are you up for opening a new spec ticket to change this before drivers start POC-ing this change? A BC-friendly way to introduce `object` for CRUD tests may be to have it default to "collection" if undefined, and have these tests specify `object: database`.

kvwalker: I made [SPEC-1184](https://jira.mongodb.org/browse/SPEC-1184) to track this.

PR #422, opened 2018-11-30
feat(crud): document support for `maxAwaitTimeMs` in aggregate

mbroadst:
If a `$changeStream` stage is run in an aggregation, a tailable
awaitdata cursor is returned and therefore `maxAwaitTimeMS` should
apply.

SPEC-1034

### DIFF
--- a/source/crud/crud.rst
+++ b/source/crud/crud.rst
@@ -255,7 +255,20 @@ Read
      * @see https://docs.mongodb.com/manual/reference/command/aggregate/
      */
     maxTimeMS: Optional<Int64>;
-    
+
+    /**
+     * The maximum amount of time for the server to wait on new documents to satisfy a tailable cursor
+     * query.
+     *
+     * This options only applies to aggregations which return a TAILABLE_AWAIT cursor. Drivers
+     * SHOULD always send this value, if the cursor is not a TAILABLE_AWAIT cursor the server will
+     * ignore it.
+     *
+     * @note this option is an alias for `maxTimeMS`, used on `getMore` commands
+     * @note this option is not set on the `aggregate` command
+     */
+    maxAwaitTimeMS: Optional<Int64>;
+
     /**
      * Enables users to specify an arbitrary string to help trace the operation through
      * the database profiler, currentOp and logs. The default is to not send a value.
@@ -269,7 +282,7 @@ Read
      *
      * This option is sent only if the caller explicitly provides a value. The default is to not send a value.
      *
-     * @see http://docs.mongodb.com/manual/reference/command/aggregate/ 
+     * @see http://docs.mongodb.com/manual/reference/command/aggregate/
      */
     hint: Optional<(String | Document)>;
   }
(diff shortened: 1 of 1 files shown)


source/crud/crud.rst
```diff
@@ -255,7 +255,14 @@ Read
...
      */
     maxTimeMS: Optional<Int64>;
-    
+
+    /**
+     * The maximum amount of time for the server to wait on new documents to satisfy a tailable cursor
+     * query. This only applies to stages which return a TAILABLE_AWAIT cursor. When the cursor is
+     * not a TAILABLE_AWAIT cursor, this option is ignored.
+     */
+    maxAwaitTimeMS: Optional<Int64>;
```
ShaneHarvey: To clarify, are drivers supposed to always send the maxAwaitTimeMS if it's set? And the server applies or ignores the option based on the cursor type?

jmikola: I realized that my previous note from the [change stream spec](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst) ("this is an aggregation command option") was unclear. This is only an alias for setting `maxTimeMS` on a subsequent `getMore` command and never goes on the `aggregate` command or `$changeStream` stage. I opened https://github.com/mongodb/specifications/pull/425 to address that.

That said, I assume you were asking if a `getMore` for a non-TAILABLE_AWAIT cursor would complain about using this option. I'm not sure, but libmongoc [will ignore the option](https://github.com/mongodb/mongo-c-driver/blob/1.13.0/src/libmongoc/src/mongoc/mongoc-cursor.c#L1649) unless the cursor is TAILABLE_AWAIT. I think it's reasonable for other drivers to do the same

source/crud/crud.rst
```diff
@@ -255,7 +255,14 @@ Read
      * @see https://docs.mongodb.com/manual/reference/command/aggregate/
      */
     maxTimeMS: Optional<Int64>;
-    
+
+    /**
+     * The maximum amount of time for the server to wait on new documents to satisfy a tailable cursor
```
jmikola: I think this can borrow some language from the Change Stream spec and state:

```
* @note this option is an alias for `maxTimeMS`, used on `getMore` commands
* @note this option is not set on the `aggregate` command
```

mbroadst: 👍

source/crud/crud.rst
mbroadst: @ShaneHarvey how about this updated text

ShaneHarvey: I think drivers should always send the option because how are we supposed to know if the server cursor is tailable or not?

jmikola: As for deciding whether to send this option or not, we can always use the existing boilerplate language:

> This option is sent only if the caller explicitly provides a value. The default is to not send a value.

But @ShaneHarvey makes a good point about the driver even knowing if the cursor is tailable or not. IIRC, drivers only know to create a TAILABLE_AWAIT cursor through `watch()` because they assume that `$changeStream` will yield such a cursor. I'm not sure how that will work in this case if the aggregate command helper doesn't solicit a `cursorType` option (as `find()` does).

mbroadst: The only stage that returns this is `$changeStream`, so I think it's safe to assume that you know you're getting a tailable cursor when specifying that stage. Either way the language allows for all behavior, and only indicates you SHOULD NOT send in that case, as it will be ignore. This doesn't preclude drivers from sending it all the time

source/crud/crud.rst
```diff
@@ -255,7 +255,18 @@ Read
...
+     * query. This only applies to aggregations which return a TAILABLE_AWAIT cursor. Drivers
+     * SHOULD NOT send this value when the cursor is not a TAILABLE_AWAIT cursor, but if they do
+     * the server will ignore it.
```
ShaneHarvey: > but if they do the server will ignore it."

Is this true? I would not be surprised if the server considers this undefined (untested) behavior that they are free to change between releases.

source/crud/crud.rst
ShaneHarvey: My objection is that the "SHOULD NOT send this value" wording encourages drivers to inspect the aggregation pipeline (which goes against the general sentiment that drivers shouldn't change behavior by inspecting pipelines and queries and such*) and that it encourages different behavior between drivers. 

*I know $out also goes against this philosophy too but that ship has sailed.

mbroadst: okay, check the text now. Let's get this thing merged in

jmikola: LGTM.

mbroadst: ping @ShaneHarvey

ShaneHarvey: Thanks, LGTM!

PR #427, opened 2018-12-07
Adding unit tests for connection monitoring and pooling

daprahamian:
Work in progress for adding unit tests for Connection Monitoring and Pooling. Test runner format is mostly set, but I will be adding more tests in tomorrow.


daprahamian:
@mbroadst Yes, that was my plan regardless of whether or not we added integration tests. Right now, the tests themselves are still in flux, but I'd appreciate review of the README and the test runner format.

source/connection-monitoring-and-pooling/tests/README.rst
```diff
@@ -0,0 +1,165 @@
...
+
+.. contents::
+
+--------
+
+Introduction
+============
+
+The YAML and JSON files in this directory are platform-independent tests that
+drivers can use to prove their conformance to the Connection Monitoring and Pooling (CPAM) Spec.
```
saghm: This probably should be changed to "CMAP" for clarity

source/connection-monitoring-and-pooling/tests/README.rst
```diff
@@ -0,0 +1,165 @@
...
+- ``description``: A text description of what the test is meant to assert
+- ``operations``: A list of operations to perform. Each operation has some of the following fields:
+
+  - ``command``: A string describing which command to issue.
+  - ``args``: An array of arguments to pass to the command
+  - ``thread``: The name of the thread in which to run this command. If not specified, runs in the default thread
+  - ``object``: For some command, specifies the object on which to run the command
+  - ``returnTo``: For some commands, specifies what to name the return value of the command
+
+- ``error``: Indicates that the main thread is expected to error during this test. Subfields (like ``message``) indicate what error is expected
```
saghm: Rather than saying "subfields like `message`", I think it would be better to actually document the error subfield in the same way that subfields for each `operation` are. It's not clear to me from the current wording whether `message` is required on all `error` fields of if there are any other possible subfields besides `message`, and I think it would be more convenient for implementers not to have to search through all of the tests to determine this.

source/connection-monitoring-and-pooling/tests/README.rst
```diff
@@ -0,0 +1,165 @@
...
+- ``events``: An array of all connection monitoring events expected to occur while running ``operations``
```
saghm: I think documentation of the subfields of each `event` would be useful so that implementers don't have to read through all of the tests to determine what the format of an `event` is.

source/connection-monitoring-and-pooling/tests/README.rst
```diff
@@ -0,0 +1,181 @@
...
+
+Unit Test Format:
+=================
+
+All Unit Tests have some of the following fields:
+
+- ``description``: A text description of what the test is meant to assert
+- ``operations``: A list of operations to perform. Each operation has some of the following fields:
+
+  - ``command``: A string describing which command to issue.
```
saghm: Naming this something other than "command" might be a good idea given that "command" usually refers to a server command; maybe "name" (or "operationName", if specificity is preferred)?

saghm:
The test format that the README describes feels a lot more complex than it needs to be. Having a mechanism for arbitrary variable lookup feels like overkill. I think what the [transactions tests](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst#test-format) did is preferable, namely specifying the minimum set of objects that suffice for all tests and then just creating them at the start of each test. There's no need to allow the test format to encode more complexity than the tests require.

source/connection-monitoring-and-pooling/tests/README.rst
daprahamian: oops

source/connection-monitoring-and-pooling/tests/README.rst
daprahamian: I think I'll go with `name` since that is what the transactions tests use

source/connection-monitoring-and-pooling/tests/README.rst
daprahamian: The `event` here maps to whatever fields are specified for an event in the specification. I can, in the final version, link to the `event` section of the spec, but I'd prefer to not duplicate that information in multiple places. It would make adding fields to any single event require a change here. And I am hoping that implementers actually read the spec itself to learn the format of the events rather than just the test runner description.

source/connection-monitoring-and-pooling/tests/README.rst
daprahamian: Like with my comment about `event` below, this is specifically pinned to how events are defined in the spec, and I'd like to avoid duplicating information.

source/connection-monitoring-and-pooling/tests/pool-creation.yml
```diff
@@ -0,0 +1,108 @@
...
+        object: pool
+        returnTo: conn3
+      - command: release
+        object: pool
+        args: [ { $$ref: [ "conn3" ] } ]
+      - command: acquire
+        object: pool
+        returnTo: conn4
+      - command: start
+        args: [ "t1" ]
```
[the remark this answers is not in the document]
daprahamian: On `t` variables: I can definitely rename these to `thread1`, `thread2`, etc, to be a little more self-documenting. Would that make sense?

On `args`: are you suggesting that we change the `args` array to an `args` object, and have each driver provide their own mapping from that object to whatever input is necessary?

mbroadst:
@daprahamian point two above is explicit not a goal of this project, so no extra effort should be made here to reach that goal.

I think Sam is correct that we have effectively specified a programming language in this spec runner (or at least approaching a turing machine), and it seems we have a bit of scope explosion here. Let's take the transactions spec approach, and then leave it to follow-on work to research whether it makes sense to come up with something this flexible

daprahamian:
@mbroadst Which parts of the test runner should I focus on removing? We still have to support the ability to:

- Have multiple pools (currently 1-2)
- Have multiple connections acquired and released in arbitrary order (from what I can tell at least 4 arbitrary connections)
- Be able to execute on multiple threads (definitely need at least 2 threads).

It seems to me that removing the flexible nature of these tests will require us to either cut scope on what we can automate tests for, or write an overly-complex (with respect to verbosity and specificity) test runner.

mbroadst:
- The two pools can be defined at start of the test run
- Two threads can be created at start of the test run
- In terms of the connections acquired and released in arbitrary order, can't we agree upon reasonable naming schemes? e.g. `pool1-connection-1`, `pool2-connection-4`, so these are easy to identify in the event checking section? 

I agree that it's more verbose, but such is the nature of the format we've chosen to write our tests in. We should make this as easy as possible for everyone to implement in a reasonable timeframe. I am very open to discussion about unified test runners, but we will have to scope that work and include others who have opinions as well.

source/connection-monitoring-and-pooling/tests/README.rst
saghm: I think the issue is that the spec doesn't define what the errors look like in YAML, just what they look like as an abstract driver interface. For someone implementing the tests, they're going to have to go through every instance of `error` in the YAML files to figure out what they need to read in (as most languages cannot convert YAML/JSON to their native object type without knowing the structure of the object first), so I feel like it would be much more convenient to just list what the field names and value types are here.

saghm:
From looking at the tests some more, I think the need for identifiers for threads/connections can also be eliminated. Instead of using a context object with arbitrary names, drivers can just append each new thread/connection to a list and refer to them by their index.

source/connection-monitoring-and-pooling/tests/README.rst
```diff
@@ -0,0 +1,181 @@
...
+- RESOLVE takes two value, ``context`` and ``unresolved``
+- Notation is ``RESOLVE(context, unresolved)``
+
+Pseudocode implementation of ``RESOLVE(context, unresolved)``:
+
+::
+
+  if unresolved is a JSON array:
+    resolved = []
+    for every idk/value in unresolved:
```
p-mongo: `idx`?

source/connection-monitoring-and-pooling/tests/README.rst
```diff
@@ -0,0 +1,181 @@
...
+      resolved[idx] = RESOLVE(context, value)
+    return resolved
+  else if unresolved is a JSON object:
+    if unresolved["$$ref""] is an array:
```
p-mongo: Unmatched quotes.

source/connection-monitoring-and-pooling/tests/README.rst
```diff
@@ -0,0 +1,181 @@
...
+
+  - Assert that no errors were thrown by the main thread
+
+- ``expectedEvents = []``
+- for every ``idx/value`` in ``events``: 
+
+  - ``expectedEvents[idx] = RESOLVE(context, value)``
+
+- calculate ``actualEvents`` as every Connection Event emitted whose ``type`` is not in ``ignore``
+- if ``expectedEvents`` is not empty, then for every (``expectedEvent``, ``i``) in ``expectedEvents``
```
p-mongo: s/i/idx/ for consistency with the other index reference.

p-mongo: Also please use the same order of value and index throughout.

source/connection-monitoring-and-pooling/tests/README.rst
```diff
@@ -0,0 +1,181 @@
...
+
+Prose Tests
+===========
+
+The following tests have not yet been automated, but MUST still be tested
+
+#. All ConnectionPoolOptions MUST be specified at the MongoClient level
+#. All ConnectionPoolOptions MUST be the same for all pools created by a MongoClient
+#. A user MUST be able to specify all ConnectionPoolOptions via a URI string
+#. A user MUST be able to subscribve to Connection Monitoring Events in a manner idiomatic to their language and driver when ``enableConnectionMonitoring`` is true
```
p-mongo: Typo in subscribe

source/connection-monitoring-and-pooling/tests/README.rst
```diff
@@ -0,0 +1,181 @@
...
+============================
+
+If an any point a value in a test is of the form ``{ $$ref: string[] }``, that value is considered a Context Reference. Context References are resolved by replacing the value with the current value of ``context[...$$ref]``
+
+The definition of RESOLVE in the Spec Test Runner is as follows:
+
+- RESOLVE takes two value, ``context`` and ``unresolved``
+- Notation is ``RESOLVE(context, unresolved)``
+
+Pseudocode implementation of ``RESOLVE(context, unresolved)``:
```
p-mongo: What is a context exactly? Because reading the spec tests it seems that connection pool and connection objects are used as "contexts".

source/connection-monitoring-and-pooling/tests/README.rst
```diff
@@ -0,0 +1,181 @@
...
+
+- ``returnTo = object.acquire()``: call ``acquire`` on Pool ``object``, returning the acquired connection
+- ``object.release(connection, force?)``: call ``release`` on Pool ``object``, passing in connection and optional force flag
+- ``object.clear()``: call ``clear`` on Pool ``object``
+- ``object.close()``: call ``close`` on Pool ``object``
+
+Spec Test Context References
+============================
+
+If an any point a value in a test is of the form ``{ $$ref: string[] }``, that value is considered a Context Reference. Context References are resolved by replacing the value with the current value of ``context[...$$ref]``
```
p-mongo: What is the reason for doubling dollar signs in $$ref?

p-mongo: And what does `context[...$$ref]` mean?

source/connection-monitoring-and-pooling/tests/README.rst
```diff
@@ -0,0 +1,181 @@
...
+- ``expectedEvents = []``
+- for every ``idx/value`` in ``events``: 
+
+  - ``expectedEvents[idx] = RESOLVE(context, value)``
+
+- calculate ``actualEvents`` as every Connection Event emitted whose ``type`` is not in ``ignore``
+- if ``expectedEvents`` is not empty, then for every (``expectedEvent``, ``i``) in ``expectedEvents``
+
+  - Assert that ``actualEvents[i]`` exists
+  - Assert that ``actualEvents[i]`` MATCHES ``expectedEvent``
```
p-mongo: I see that event order is not mandated, was this also intentional for e.g. connection creation event vs wait queue entering event?

daprahamian: Event order should be mandated. Every `actualEvent` should line up with a corresponding `expectedEvent`

source/connection-monitoring-and-pooling/tests/pool-acquire.yml
```diff
@@ -0,0 +1,147 @@
...
+      - type: connectionPoolAcquire
+        pool:
+          id: {$$ref: [ pool, id ]}
+        connection:
+          id: 1
+      - type: connectionPoolAcquire
+        pool:
+          id: {$$ref: [ pool, id ]}
+        connection:
+          id: 2
```
p-mongo: Since connection creation (which defines the id) and connection return from the pool are separate operations, there is no requirement that the events for connection checkout will have the ids in ascending order.

saghm:
> Starting ids from 0 is not good, 1 is a much better choice.

Is this in reference to my earlier proposal about IDs for threads/connections? The reason I proposed "0" as the initial ID is that the most logical way to store the threads in the order they were created is by using a list, and I think that `threads[id]` is more clear than `threads[id - 1]` or putting a sentinel value as the first element of the list.

source/connection-monitoring-and-pooling/tests/pool-acquire.yml
daprahamian: The pool is created with no minPoolSize, meaning there are no pre-generated connections. So every call to acquire will result in the creation of a connection for that call. The calls to acquire are in parallel, but the waitQueue mandates that they run acquisition one at a time, ensuring that each one will get a connection in order.

p-mongo: The spec does not say that connection creation must be performed under the lock of a wait queue, only that the wait queue must be threadsafe  and fair. If connections are (always) created by a background thread, creating them under the wait queue's lock is at a minimum very difficult. If connection timeout exceeds wait queue timeout, a connection won't ever be created by a client so configured. Even if the tests don't perform any network i/o there is still a lot that happens between when an id is allocated and when a connection can be checked out, and if these operations are done in a background thread then the sequence of ids as seen by the application may be non-monotonic even in the absence of network i/o.

daprahamian:
I'm skeptical of the idea that implementing a pool list, and then a 2d list of pools --> connections, and then parsing the corresponding indexes out of string names like `pool2` and `pool1-connection3` is simpler than maintaining a key/value store of name to object. I could understand not wanting to have arbitrary references with a funky `$$ref` syntax, but it seems like eliminating the name/object mapping elects to make a more brittle test runner, and can incur more work in the long run should we ever expand the scope of what we are testing.

saghm:
I don't think I've suggested anything about two-dimensional lists, and my latest suggestion does away with the need for parsing numbers; we'd just need three single-dimensional lists (one for pools, one for connections, and one for threads) and then we can just use the numbers themselves as the identifiers rather than strings. I think implementing lookup for arbitrary identifiers is a lot more complex than necessary since we're already defining names with sequential numbers.

daprahamian:
@saghm I was referring to @mbroadst's suggestion of using a naming convention like `pool1-connection1`. w/ regards to arrays, that works decently well for pools and threads, but connections need a specific way to identify themselves for later tests. Making the returned connections append to a list is complicated by the nondeterminism of running multiple threads. We would have to set the return key of an acquired connection instead of pushing it on to the end, at which point we are basically using a key/value store, just with numbers instead of indexes.

saghm:
Do the tests ever start up threads outside the main thread? I was under the impression that the threads spun up by the tests don't spin up additional threads of their own

daprahamian:
To clarify: there are no determinism issues with the threads themselves, but rather how the threads execute their acquire calls:

```yml
- name: start
- name: start
- name: start
- name: acquire
  thread: 1
- name: acquire
  thread: 2
- name: acquire
  thread: 3
```

Here, we don't know which thread acquired which connection. There is at least one test that relies on being able to say "release the connection that thread x acquired"

source/connection-monitoring-and-pooling/tests/pool-acquire.yml
p-mongo: If there is a single background thread that creates all connections, then, in the absence of network i/o, I imagine the sequence of connection ids as exposed to the application through events should in fact be monotonic. But once network i/o gets added to the picture, even async i/o on the same thread, the sequence would no longer need to be monotonic (e.g. due to handshake taking varying amounts of time and thus connections becoming ready in a different order from their creation/id allocation).

saghm:
I see. Do threads ever need to acquire more than one connection, or just a single one per thread?

daprahamian:
@saghm AFAIK only one per thread. I've been using the threads to test

- Simultaneous acquisition of connections, and the waitQueue respecting acquire order
- Ensuring that waitQueueSize is properly asserted

saghm:
In that case, I think that we can just make an array the size of the number of threads and put the connections in the same index as the thread they correspond to (with languages that have resizable arrays using `null` or a similar language primitive for the initial elements). If we want to check the connection for thread N, we can just look at connection N.

source/connection-monitoring-and-pooling/tests/README.rst
```diff
@@ -0,0 +1,152 @@
...
+
+- ``wait(ms)``: Sleep the current thread for ``ms`` milliseconds
+
+  - ``ms``: The number of milliseconds to sleep the current thread for
+
+- ``waitFor(target)``: wait for thread ``target`` to finish executing. Propagate any errors to the main thread.
+
+  - ``target``: The name of the thread to wait for.
+
+- ``returnTo = createPool(options)``: creates and returns a new Connection Pool with the specified options.
```
saghm: This can be removed now, right?

daprahamian: oops! yes, that can be removed!

source/connection-monitoring-and-pooling/tests/pool-checkout-error-closed.yml
```diff
@@ -0,0 +1,23 @@
...
+  - name: checkIn
+    connection: conn1
+  - name: close
+  - name: checkOut
+error:
+  type: PoolClosedError
+  message: Attempted to check out a connection from closed connection pool
+events:
+  - type: ConnectionPoolCreated
+  - type: ConnectionCheckedOut
```
saghm: `connectionId: 42` should be added to this `ConnectionCheckedOut` event (and the `ConnectionCheckedIn` event below it)

source/connection-monitoring-and-pooling/tests/pool-checkout-no-stale.yml
```diff
@@ -0,0 +1,35 @@
...
+  - type: ConnectionCreated
+    connectionId: 3
+  - type: ConnectionPoolCleared
+  - type: ConnectionCheckOutStarted
+
+  # Note: we cannot deterministically assert which connections get destroyed in which order,
+  # just that they will all be destroyed before connection check out.
+  - type: ConnectionClosed
+    reason: stale
+  - type: ConnectionClosed
```
saghm: `connectionId=42` should be added to these `ConnectionClosed` events as well

source/connection-monitoring-and-pooling/tests/pool-close.yml
```diff
@@ -0,0 +1,10 @@
+version: 1
+style: unit
+description: A Pool MUST be able to be manually closed.
+operations:
+  - name: close
+events:
+  - type: ConnectionPoolCreated
+    address: "42"
+  - type: ConnectionPoolClosed
+    address: "42"
```
saghm: It might be worth having an empty `ignore` array here to make things easier for statically-typed languages

source/connection-monitoring-and-pooling/tests/pool-close.yml
```diff
@@ -0,0 +1,10 @@
+version: 1
+style: unit
+description: A Pool MUST be able to be manually closed.
+operations:
+  - name: close
+events:
+  - type: ConnectionPoolCreated
+    address: "42"
```
saghm: I think it would be easier for most languages if we stuck to `42` as an integer rather than also using as a string (like here)

saghm:
The YAML files are somewhat inconsistent in regards to when a `42` value is specified in an event; in some tests, it's present for fields like `connectionId` and `address`, and in others, it's missing. I think for the sake of consistency, we should be specifying these fields in every event type where the spec defines the field and put `42` as the value when we don't care what it is.

source/connection-monitoring-and-pooling/tests/pool-create-max-size.yml
```diff
@@ -0,0 +1,42 @@
...
+    thread: thread1
+  - name: checkIn
+    connection: conn1
+  - name: waitFor
+    target: thread1
+events:
+  - type: ConnectionPoolCreated
+  - type: ConnectionCheckOutStarted
+  - type: ConnectionCreated
+  - type: ConnectionReady
```
saghm: Having `ConnectionReady` here complicates the tests a bit; because it isn't supposed to be emitted unless the connection has successfully completed the handshake and authenticated to the server, I don't think it's possible to emit this event without doing I/O. My understanding is that these tests are intended to be run without doing any I/O, so is it worth testing `ConnectionReady` at all? I'm not sure how we could test this in Ruby using that actual code path where the event is emitted, which means the test wouldn't be that meaningful.

source/connection-monitoring-and-pooling/tests/pool-create-min-size.yml
```diff
@@ -0,0 +1,22 @@
...
+    ms: 50
+  - name: checkOut
+events:
+  - type: ConnectionPoolCreated
+  - type: ConnectionCreated
+  - type: ConnectionCreated
+  - type: ConnectionCreated
+  - type: ConnectionCreated
+  - type: ConnectionCreated
+  # Ensaures that by the time pool is closed, there are at least 5 connections
```
saghm: "Ensures"

saghm:
Here's my proposed set of revisions for the YAML tests: https://github.com/saghm/specifications/commit/ad4e089005b793be4cf9da8dd6e8006a0ea59b7d

I changed all of the events so that each instance of an event type has the exact same set of keys present (adding in `42` as the value for each new key added). I also changed instances of "42" to `42`, since most languages will not implicitly treat them the same, and I think it's better to use the same exact value in each case as our wildcard.

I have all the tests passing in my branch of the Ruby driver with these revisions checked into the repo. I need to do a little cleanup of the code, but I hope to have it pushed up for review by the end of day today.

source/connection-monitoring-and-pooling/tests/wait-queue-fairness.yml
```diff
@@ -0,0 +1,80 @@
+version: 1
+style: unit
+description: must issue Connections to threads in the order that the threads eneterd the queue
```
saghm: "entered"

PR #428, opened 2018-12-10
SPEC-1128: Add tests for all change stream event types

kvwalker:


source/change-streams/tests/change-streams.yml
```diff
@@ -2,6 +2,7 @@ collection_name: &collection_name "test"
 database_name: &database_name "change-stream-tests"
 collection2_name: &collection2_name "test2"
 database2_name: &database2_name "change-stream-tests-2"
+collection_rename: &collection_rename "test-rename"
```
mbroadst: Hm this requires changes to the spec runner itself, right? I don't see mention of it in the `README`. Is there a way to reuse the `database2`/`collection2` for this purpose (this would require minimal changes in everyones spec runner)? Otherwise, either we need to add a `database3`/`collection3`, or some sort of extensible format (maybe `namespaces: ['test.change-stream-tests', 'test2.change-stream-tests-2'..]`)

source/change-streams/tests/change-streams.yml
```diff
@@ -297,3 +298,117 @@ tests:
           fullDocument:
             z:
               $numberInt: "3"
+  -
+    description: Test all change event types
+    minServerVersion: "3.6.0"
```
mbroadst: I think you'll need to bump the minimum server version for some of these, which means you might need to break this test apart into a few individual tests.

source/change-streams/tests/change-streams.yml
mbroadst: looks like the actual `collection_rename` is still here, but no longer used anywhere. Safe to remove this now?

source/change-streams/tests/change-streams.yml
```diff
@@ -297,3 +297,140 @@ tests:
...
+          command:
+            aggregate: *collection_name
+            cursor: {}
+            pipeline:
+              -
+                $changeStream:
+                  fullDocument: default
+          command_name: aggregate
+          database_name: *database_name
+    result:
```
daprahamian: If this test is testing "drop, rename, and invalidate event types", should we not be seeing a "drop" event somewhere in the result?

mbroadst: whoops, good catch

mbroadst: Thanks @kvwalker, LGTM

source/change-streams/tests/change-streams.yml
```diff
@@ -297,3 +297,163 @@ tests:
...
+    description: Test insert, update, replace, and delete event types
+    minServerVersion: "3.6.0"
+    target: collection
+    topology:
+      - replicaset
+    changeStreamPipeline: []
+    changeStreamOptions: {}
+    operations:
+      -
+        database: *database_name
```
jmikola: @kvwalker: I think it's worth pointing out that this deviates from `database_name` being used in the CRUD tests (https://github.com/mongodb/specifications/pull/437#discussion_r243255739). If you decide to keep the operation-level DB name in those tests, it may be worth standardizing with this format.

source/change-streams/tests/change-streams.yml
```diff
@@ -297,3 +297,163 @@ tests:
...
+    result:
+      success:
+        -
+          operationType: insert
+          ns:
+            db: *database_name
+            coll: *collection_name
+          fullDocument:
+            x:
+              $numberInt: "1"
```
jmikola: I'm not familiar with the other change stream tests, but is the `_id` field purposefully omitted here? Would it make sense to use numeric `_id` values in the example data (as is done in most CRUD tests IIRC) so that we can assert it as well?

source/change-streams/tests/change-streams.yml
```diff
@@ -297,3 +297,163 @@ tests:
...
+            db: *database_name
+            coll: *collection_name
+          to:
+            db: *database_name
+            coll: *collection2_name
+        -
+          operationType: invalidate
+  -
+    description: Test drop and invalidate event types
+    minServerVersion: "4.0.1"
```
jmikola: Unrelated to this PR, but I didn't realize the change stream tests have this at the test-level, instead of file-level as is done in other specs :\

source/change-streams/tests/change-streams.yml
```diff
@@ -297,3 +297,163 @@ tests:
...
+    topology:
+      - replicaset
+    changeStreamPipeline: []
+    changeStreamOptions: {}
+    operations:
+      -
+        database: *database_name
+        collection: *collection_name
+        name: drop
+        arguments:
```
jmikola: I saw blank YAML values being used in some transaction tests, which evaluates to `null`; however, I'd propose using explicit `~` or `null` here if the key is needed for clarity.

That said, can we just make `arguments` optional? It really has no purpose for `drop`, but I assume this may be the first operation in the test suite where we're calling a method with no arguments.

source/change-streams/tests/change-streams.yml
kvwalker: Most of the other change stream tests don't include `_id`. I could add it here, but what benefit would asserting it provide over what we currently have?

source/change-streams/tests/change-streams.yml
jmikola: Per https://github.com/mongodb/specifications/pull/437, the CRUD tests no longer have a `database` field within an `operation` element, so that's no longer a concern.

It looks like this field is still necessary because the `target` field only specifies the type of object (i.e. client, database, collection) and we need this to denote the name of the corresponding object to invoke `watch()`. In `changes-streams.yml`, this juggles between `*database_name` and `*database2_name`. Do you concur that this is still necessary? If so, we can resolve.

source/change-streams/tests/change-streams.yml
jmikola: @mbroadst: GitHub says this thread is outdated. OK to resolve?

source/change-streams/tests/change-streams.yml
jmikola: I may be misunderstanding the test runner behavior, but with `_id` omitted here, is the test runner supposed to completely ignore the field when evaluating the assertion? I assume the response from the server _is_ going to include `_id` and that the spec runner would assert that `fullDocument` is equal to what we have here.

source/change-streams/tests/change-streams.yml
jmikola: @daprahamian: Is this something we can change in the interest of cross-spec consistency? If so, I'll open a spec ticket. It looks like most all of the tests are using `minServerVersion: 3.6.0` and a few use `minServerVersion: 3.8.0`. IMO, those could be split into their own YAML files (the entire spec currently only has two files for error and non-error tests).

source/change-streams/tests/change-streams.yml
kvwalker: Yes, it's still necessary.

source/change-streams/tests/change-streams.yml
kvwalker: According to the README, "Assertion passes if ``expected`` is a subset of ``actual``" so the spec runner does not need to assert that the `_id` is correct, although the server response will include the `_id`.

source/change-streams/tests/change-streams.yml
jmikola: Opened https://jira.mongodb.org/browse/SPEC-1193 to track this.

PR #429, opened 2018-12-10
SPEC-1141: Add startAfter to ChangeStreamOptions

kvwalker:


source/change-streams/change-streams.rst
```diff
@@ -278,6 +278,16 @@ Driver API
...
+
+    /**
+     * Similar to 'resumeAfter', this option takes a resume token and starts a
+     * new change stream returning the first notification after the token.
+     * This will allow users to watch collections that have been dropped and recreated
+     * or newly renamed collections without missing any notifications.
+     * @since 4.2
+     * @see https://docs.mongodb.com/master/changeStreams/#change-stream-start-after
+     */
+     startAfter: Optional<Timestamp>
```
jmikola: I think type should be `Optional<Document>`, since it's a resume token (like `resumeAfter`, not `startAtOperationTime`). Also trailing `;` for consistency.

source/change-streams/change-streams.rst
```diff
@@ -278,6 +278,16 @@ Driver API
      * @note this is an option of the `$changeStream` pipeline stage.
      */
     startAtOperationTime: Optional<Timestamp>;
+
+    /**
+     * Similar to 'resumeAfter', this option takes a resume token and starts a
+     * new change stream returning the first notification after the token.
+     * This will allow users to watch collections that have been dropped and recreated
+     * or newly renamed collections without missing any notifications.
```
jmikola: Should we mention that this option is mutually exclusive with `resumeAfter`? If you look at `document_source_change_stream.cpp` in the code review for [SERVER-35030](https://jira.mongodb.org/browse/SERVER-35030) you can see that the server enforces that. I don't think drivers need to add client-side validation, but it may be helpful to document this with a line like:

> The server will report an error if ``startAfter`` and ``resumeAfter`` are both specified.

source/change-streams/change-streams.rst
```diff
@@ -278,6 +278,19 @@ Driver API
...
+     *
+     * The server will report an error if startAfter and resumeAfter are both specified.
```
mbroadst: nit: wrap `startAfter` and `resumeAfter`

jmikola: Do you mean wrap in backticks? If so, perhaps that should be done for resumeAfter a few lines above as well. It currently uses single quotes.

mbroadst: yeah let's wrap them all in backticks to indicate that they are known property names

ShaneHarvey:
I am a little confused by this change. Have we considered any alternative APIs for this feature? For example do we even need to expose 'startAfter' at all? What if resumeAfter always behaves like "startAfter" instead? ie when starting a changestream we always pass the resumeAfter token as "startAfter"?

Why is it desirable that using resumeAfter after a collection drop results in an error?

mbroadst:
@ShaneHarvey See [SERVER-35030](https://jira.mongodb.org/browse/SERVER-35030) for explanation of the subtle difference between these keys. I'm not sure we can hide it on the driver side, you will need contextual information about whether the resume is happening after an invalidate change

PR #430, opened 2018-12-10
SPEC-1183 Increase attempts in unpinning tests to reduce test flakes

ShaneHarvey:
For motivation see https://jira.mongodb.org/browse/SPEC-1183.

### DIFF
--- a/source/transactions/tests/README.rst
+++ b/source/transactions/tests/README.rst
@@ -337,13 +337,15 @@ instead.
       @require_mongos_count_at_least(2)
       def test_unpin_for_next_transaction(self):
         client = MongoClient(mongos_hosts)
+        # Create the collection.
+        client.test.test.insert_one({})
         with client.start_session() as s:
           # Session is pinned to Mongos.
           with s.start_transaction():
             client.test.test.insert_one({}, session=s)
 
           addresses = set()
-          for _ in range(20):
+          for _ in range(50):
             with s.start_transaction():
               cursor = client.test.test.find({}, session=s)
               assert next(cursor)
@@ -360,13 +362,15 @@ instead.
       @require_mongos_count_at_least(2)
       def test_unpin_for_non_transaction_operation(self):
         client = MongoClient(mongos_hosts)
+        # Create the collection.
+        client.test.test.insert_one({})
         with client.start_session() as s:
           # Session is pinned to Mongos.
           with s.start_transaction():
             client.test.test.insert_one({}, session=s)
 
           addresses = set()
-          for _ in range(20):
+          for _ in range(50):
             cursor = client.test.test.find({}, session=s)
             assert next(cursor)
             addresses.add(cursor.address)


ShaneHarvey:
> That you hit this makes me doubt the randomness in PyMongo's server selector.

I thought about this as well but cannot find any evidence of a faulty server selection. We use `random.choice` from [Python's standard library](https://docs.python.org/3.6/library/random.html#random.choice) which is as (peusdo) random as it gets.

Perhaps a more likely theory is that the default 15 milliseconds localThresholdMS is effectively removing one of the Mongoses from server selection before we get to the random selection. How do you feel about increasing this test's localThresholdMS to 1000 ms for good measure?

jyemin:
If the MongoClient is created relatively fresh for each test, another possibility is that the second mongos hasn't even been discovered before the test completes.

mbroadst:
Agree with Jeff, this smells like something's up elsewhere in the test or in PyMongo's implementation. Bumping to 50 is harmless, but might you still run into the flakiness in the future given the right circumstances?

source/transactions/tests/README.rst
```diff
@@ -337,13 +337,17 @@ instead.
...
+        # Create the collection.
+        client.test.test.insert_one({})
         with client.start_session() as s:
           # Session is pinned to Mongos.
           with s.start_transaction():
             client.test.test.insert_one({}, session=s)
 
           addresses = set()
-          for _ in range(20):
+          for _ in range(50):
```
jyemin: You want to keep this at 50 still?

source/transactions/tests/README.rst
```diff
@@ -360,13 +364,17 @@ instead.
       @require_mongos_count_at_least(2)
       def test_unpin_for_non_transaction_operation(self):
         client = MongoClient(mongos_hosts)
+        # Wait until more than one mongos is discovered.
+        wait_until(lambda: len(client.nodes) > 1)
```
jyemin: You don't think we also need to bump the latency threshold?

source/transactions/tests/README.rst
ShaneHarvey: Yep. Is there any downside?

source/transactions/tests/README.rst
ShaneHarvey: Done.

source/transactions/tests/README.rst
jyemin: Only that it's unnecessary, so why change it.

ShaneHarvey: So with 20 attempts it's about a 1 in a million chance for each test to fail. Let's calculate the chance of a PyMongo evergreen patch build of failing. There are 2 tests and 211 tasks that run this test:
(1/(.5**20))/(2*211) = 2484

So it’s a 1/2484 chance for each patch build to fail and that's only counting pymongo. Raising it to 50:
(1/(.5**50))/(2*211) = 2668009257920

Much more wiggle room!

mbroadst: :D

PR #431, opened 2018-12-11
Interim review for retryable reads test plan and tests 

vincentkam:
Requesting interim feedback on the test plan and count tests while I finish the tests for the rest of the read operations :)


vincentkam:
Added find, distinct, and aggregate tests for review as well: they follow the format of the count tests, with the exception of "aggregate", which has an additional test to ensure that aggregates with {{$out}} are not retried.

source/retryable-reads/tests/README.rst
```diff
@@ -0,0 +1,269 @@
...
+that drivers can use to prove their conformance to the Retryable Reads spec.
+
+Prose tests, which are not easily expressed in YAML, are also presented
+in this file. Those tests will need to be manually implemented by each driver.
+
+Tests will require a MongoClient created with options defined in the tests.
+Integration tests will require a running MongoDB cluster with server versions
+4.2.0 or later. N.B. Strictly speaking, the server should implement the
+"Early Failure on Socket Disconnect" slated for 4.2.0, but this is not
+neccessary for initial testing.
```
jmikola: typo: necessary

source/retryable-reads/tests/README.rst
```diff
@@ -0,0 +1,269 @@
...
+    db.adminCommand({
+        configureFailPoint: "failCommand",
+        mode: <string|document>,
+        data: {
+          failCommands: ["commandName", "commandName2"],
+          closeConnection: <true|false>,
+          errorCode: <Number>,
+        }
+    });
+Some tests depend on a server fail point, ``failCommand``, which
```
jmikola: Should there be a newline here between the code sample and paragraph text?

source/retryable-reads/tests/README.rst
```diff
@@ -0,0 +1,269 @@
...
+
+.. _SERVER-29606: https://jira.mongodb.org/browse/SERVER-35004
+
+The fail point may be configured like so::
+
+    db.runCommand({
+        configureFailPoint: "failCommand",
+        mode: <string|document>,
+        data: <document>
+    });
```
jmikola: Isn't this redundant with the previous example, which was also for `failCommand` but illustrated the format of the `data` field?

source/retryable-reads/tests/README.rst
```diff
@@ -0,0 +1,269 @@
...
+  ``true``, the connection on which the command is executed will be closed
+  and the client will see a network error.
+- ``errorCode``: Integer option, which is unset by default. If set, the
+  specified command error code will be returned as a command error.
+
+Disabling Fail Point after Test Execution
+-----------------------------------------
+
+After each test that configures a fail point, drivers should disable the
+``failCommands`` fail point to avoid spurious failures in
```
jmikola: The fail point is `failCommand`. `failCommands` is one of the `data` fields.

source/retryable-reads/tests/README.rst
```diff
@@ -0,0 +1,269 @@
...
+subsequent tests. The fail point may be disabled like so::
+
+    db.runCommand({
+        configureFailPoint: "failCommands",
```
jmikola: I think this should be `failCommand`

source/retryable-reads/tests/README.rst
```diff
@@ -0,0 +1,269 @@
...
+    });
+
+Network Error Tests
+===================
+
+Network error tests are expressed in YAML and should be run against a standalone,
+shard cluster, or single-node replica set. Until `SERVER-34943`_ is resolved,
+these tests cannot be run against a multi-node replicaset.
+
+
```
jmikola: Extra newline.

source/retryable-reads/tests/README.rst
```diff
@@ -0,0 +1,269 @@
...
+.. _SERVER-34943: https://jira.mongodb.org/browse/SERVER-34943
+
+Test Format
+-----------
+
+Test Format
+===========
```
jmikola: Any reason for duplicate headings here? "Test Format" is already above.

source/retryable-reads/tests/README.rst
```diff
@@ -0,0 +1,269 @@
...
+
+    - ``name``: The name of the operation on ``object``.
+
+    - ``object``: The name of the object to perform the operation on. Can be
+      "database", "collection", or "client."
+
+    - ``collectionOptions``: Optional, parameters to pass to the Collection()
+      used for this operation.
+
+    - ``command_name``: Present only when ``name`` is "runCommand". The name
```
jmikola: Is this actually relevant to these spec tests? I only see `command_name` referenced in APM expectations. Also, the spec isn't intended to pertain to `runCommand` so I'm not sure why you'd even test for that.

source/retryable-reads/tests/README.rst
```diff
@@ -0,0 +1,269 @@
...
+        exception. Implicitly true if ``result`` contains ``errorContains``,
+        ``errorCodeName``, ``errorLabelsContain``, and/or
+        ``errorLabelsOmit``. If ``false``, the test should expect no error or
+        exception, even if ``result`` contains ``errorContains``,
+        ``errorCodeName``, ``errorLabelsContain``, and/or ``errorLabelsOmit``.
+
+      - ``errorContains``: Optional. A substring of the expected error message.
+
+      - ``errorCodeName``: Optional. The expected "codeName" field in the server
+        error response.
```
jmikola: Is this intended to be checked after the driver translates the numeric error code? I may be behind on other spec tests, but I don't recall seeing this field before, and I don't recall servers returning a "codeName" field in an `{ok:0}` error response.

source/retryable-reads/tests/README.rst
```diff
@@ -0,0 +1,269 @@
...
+
+Drivers MUST ensure that ``mapReduce`` is retried only when ``out: { inline:
+1}``.  The tests for ``mapReduce`` should test against the same conditions as
+described in ``aggregate.yml`` and ``aggregate-serverErrors.yml``, with the
+exception that ``mapReduce`` MUST not retry when the `out` stage is not
+``{inline: 1}``.
+
+A recommended data set would be three documents that contain values for x such
+that ``x:0``, ``x:1`` and ``x:2``. The map function should increment each of
+these values by one, and the reduce function should add all x's together.
```
jmikola: If this is a prose test, I don't know if it's helpful to refer to YAML tests. Since those can be synced after the fact, it makes it all the more likely that drivers' mapReduce prose tests may become inconsistent (even within the driver itself). I think a single `mapReduce` command document, which drivers can translate for their mapReduce helper (if one exists) would be sufficient. You can advise them how to setup the fail point here and assert that the command is successful. Repeating the same test without `{inline: 1}` could then expect a failure.

source/retryable-reads/tests/README.rst
```diff
@@ -0,0 +1,269 @@
...
+functions should be tested against the same conditions described in ``find.yml``
+and ``find-ServerErrors.yml``. The tests' setup and expectations can be adapted
+from the "Download by Name when revision is 0" test at
+https://github.com/mongodb/specifications/blob/master/source/gridfs/tests/download_by_name.yml
+and the "Download when there is one chunk" test at
+https://github.com/mongodb/specifications/blob/master/source/gridfs/tests/download.yml.
+
+.. _File Download: https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst#file-download-by-filename
+
+.. _Generic Find on File Collection:  https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst#generic-find-on-files-collection
```
jmikola: On second thought, I dunno if it's worth even having this as a prose test. If drivers know that their GridFS implementation uses the same code path as `collection.find()`, there shouldn't be anything to test. Also, file downloads may or may not entail `getMore` commands, which makes it likely that only small file downloads (i.e. single batch of results) will survive a retryable error when reading. Drivers cannot reliably guarantee complete retryability for GridFS read operations and users may not even be aware of whether a file read entails `getMore` commands.

source/retryable-reads/tests/README.rst
jmikola: After reading "Optionally Retryable Commands" below, I understand why you referred to YAML tests here. Those likely contain the varieties of failures to test. If this is common to both the mapReduce and "other optional retryable commands", then perhaps you can just summarize those in this test README. I'm not sure if that makes it any more or less likely that these prose tests could get out of sync. I don't feel strongly about this, so if you believe referring to YAML is best I trust your judgement.

source/retryable-reads/tests/README.rst
```diff
@@ -0,0 +1,269 @@
...
+=============================
+
+If a driver chooses to implement retryability for the optionally retryable
+enumeration commands (e.g. ``client.listDatabases()``,
+``db.listCollections()``), the tests for the enumeration commands should test
+against the same conditions as described in ``find.yml`` and
+``find-serverErrors.yml``.
+
+    
+Replica Set Failover Test
```
jmikola: Do I assume correctly that this is pending SERVER-34943, per the note above about not being able to test with multi-node replicasets?

source/retryable-reads/tests/README.rst
```diff
@@ -0,0 +1,269 @@
...
+shard cluster, or single-node replica set. Until `SERVER-34943`_ is resolved,
+these tests cannot be run against a multi-node replicaset.
+
+
+.. _SERVER-34943: https://jira.mongodb.org/browse/SERVER-34943
+
+Test Format
+-----------
+
+Test Format
```
behackett: It looks like the actual test files use a different format with more fields than what is documented here. For example, failCommand_failPoint, retryable_command_started_event and retryable_operation. This format seems different from the retryable writes format. Can you document these fields and how they are used?

source/retryable-reads/tests/README.rst
vincentkam: You're right: command_name is not needed here, nor is it needed in the APM expectations. Removing them.

source/retryable-reads/tests/README.rst
vincentkam: I beleive this field was first introduced in the transactions tests: https://github.com/mongodb/specifications/tree/master/source/transactions/tests
Admittedly, we're not using any of these fields, so I could drop them from the test plan, leaving only "error" defined.

source/retryable-reads/tests/README.rst
vincentkam: Correct! I'll add another note about the need for that ticket to be resolved.

source/retryable-reads/tests/README.rst
vincentkam: Thanks Jeremy: I think I'll continue referring to the YAML :)

source/retryable-reads/tests/README.rst
vincentkam: If all drivers know that their GridFS implementation uses the same code path as `collection.find(),` I would be comfortable simply noting this in the test plan/spec as why there are no explicit tests for GridFS. However, if we were committed to testing the GridFS API for retryability, it would only be the initial find that would be on the files collection that would be retried (and we could add this note to the test plan).

source/retryable-reads/tests/README.rst
vincentkam: As discussed offline, I'll refactor the anchors I used so that these fields no longer exist as top-level fields.

vincentkam:
Okay, I think the test plan and tests should be ready for review again!

source/retryable-reads/tests/README.rst
jmikola: If `error` is defined, would it be a boolean field? IIRC, that's what the CRUD spec uses and I'm contemplating using the same in my transaction helper spec tests.

source/retryable-reads/tests/README.rst
jmikola: > We would only be retrying if the find on the files collection fails, I think.

Initial `find` on the chunks collection could also be retried. Anything that isn't a `getMore` is retryable.

In any event, GitHub says the diff for this thread is outdated so I'll resolve this. I don't feel strongly either way, but my preference is not to bother testing GridFS (we didn't do so for retryable writes, either).

source/retryable-reads/tests/README.rst
vincentkam: Yes, error is now a boolean field in the latest version! https://github.com/mongodb/specifications/pull/431/files#diff-e6269564f9bd464986ef56193a952c88R131

source/retryable-reads/tests/aggregate.yml
```diff
@@ -0,0 +1,99 @@
...
+        clientOptions:
+            retryReads: false
+        failPoint: *failCommand_failPoint
+        operation:
+            <<: *retryable_operation
+            arguments:
+                pipeline:
+                    - $sort: {x: 1}
+                    - $match: 
+                        _id: {$gt: 1}
```
jmikola: Similar to my earlier comment, I think this can be inlined (`- { $match: {_id: {$gt: 1}}}`) and moved above `$sort`.

source/retryable-reads/tests/aggregate.yml
```diff
@@ -0,0 +1,99 @@
...
+                    - $out: "output-collection"
```
jmikola: AFAICT, there's no need to clean up this collection since the `aggregate` command will fail. Correct?

source/retryable-reads/tests/changeStreams-client.watch-serverErrors.yml
```diff
@@ -0,0 +1,144 @@
...
+        clientOptions:
+            retryReads: true
+        failPoint: &failCommand_failPoint
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data: { failCommands: [aggregate], errorCode: 11600 }
+        operation: &retryable_operation
+            name: watch
+            object: client
+            result: []
```
jmikola: Since the change stream is tailable, how do you intend to assert the contents of the returned iterator?

I believe the [Change Stream test runner](https://github.com/mongodb/specifications/tree/master/source/change-streams/tests#spec-test-runner) instructs drivers to only assert up to the contents of the array in the test's `operation.result` field. I assume drivers can then abandon iteration on the tailable iterator after that point and pass the test. In this case, there is really nothing to assert so I might suggest omitting `result` entirely.

source/retryable-reads/tests/changeStreams-client.watch-serverErrors.yml
```diff
@@ -0,0 +1,144 @@
...
+data:
+    - {_id: 1, x: 11}
+    - {_id: 2, x: 22}
+    - {_id: 3, x: 33}
+    
+minServerVersion: '4.1.6' # 4.2
+
+tests:
+    -
+        description: "client.watch succeeds after InterruptedAtShutdown"
```
jmikola: Similar comment regarding tests for expecting errors on `retryReads: false` and after two successive server errors. I'll omit repeating this comment in future test files, as I imagine it applies to all read operations (not just `aggregate` and change streams).

source/retryable-reads/tests/changeStreams-client.watch-serverErrors.yml
```diff
@@ -0,0 +1,144 @@
...
+            result: []
+        expectations:
+            - &retryable_command_started_event
+                command_started_event:
+                    command:
+                        aggregate: *collection_name
+                        cursor: {}
+                        pipeline:
+                          - $changeStream:
+                                fullDocument: default
```
jmikola: This seems like a reasonable place to inline the expected pipeline if it wouldn't make the line too long. I think a good rule of thumb is to inline things like documents and lengthy command arguments, which aren't terribly important to the spec itself.

source/retryable-reads/tests/count-serverErrors.yml
```diff
@@ -0,0 +1,141 @@
...
+    -
+        description: "Count succeeds after InterruptedAtShutdown"
+        clientOptions:
+            retryReads: true
+        failPoint: &failCommand_failPoint
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data: { failCommands: [count], errorCode: 11600 }
+        operation: &retryable_operation
+            name: count
```
jmikola: Should this be tested even though the helper is deprecated? Most drivers still have the method so I suppose there's good reason to do so.

I'm curious if drivers like Go and Swift, which will have their first public release after the deprecation, will need special logic to skip this. I suppose it's not a huge problem provided all of the `count()` helper tests are in their own file (they can just omit the test files in their runners).

source/retryable-reads/tests/gridfs-download-serverErrors.yml
```diff
@@ -0,0 +1,152 @@
...
+    files:
+        -
+            _id: { "$oid" : "000000000000000000000003" }
+            length: 2
+            chunkSize: 4
+            uploadDate: { "$date" : "1970-01-01T00:00:00.000Z" }
+            md5: "c700ed4fdb1d27055aa3faa2c2432283"
+            filename: "length-2"
+            contentType: "application/octet-stream"
+            aliases: []
```
jmikola: According to the [GridFS spec](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst), `md5`, `contentType`, and `aliases` are all deprecated. I think it makes sense to omit them. AFAIK, no driver is required to use these fields and I'm not aware of any that relied on `md5` to validate files during read operations.

source/retryable-reads/tests/gridfs-download-serverErrors.yml
```diff
@@ -0,0 +1,152 @@
...
+            metadata: {}
+    chunks:
+        - { _id : { "$oid" : "000000000000000000000003" }, files_id : { "$oid" : "000000000000000000000004" }, n : 0, data : { $hex : "11223344" } }
```
jmikola: Should your test README refer to the GridFS spec's test README for information about handling `$hex` when loading data fixtures?

source/retryable-reads/tests/gridfs-download-serverErrors.yml
```diff
@@ -0,0 +1,152 @@
...
+        clientOptions:
+            retryReads: true
+        failPoint: &failCommand_failPoint
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data: { failCommands: [find], errorCode: 11600 }
+        operation: &retryable_operation
+            operation: download
+            arguments:
+                id: { "$oid" : "000000000000000000000003" }
```
jmikola: Is this intended to assert that we read the file's contents properly, or are we happy with an error not having occurred? I'm OK with either, just curious what the intention was here since `result` is omitted.

source/retryable-reads/tests/gridfs-download-serverErrors.yml
```diff
@@ -0,0 +1,152 @@
+database_name: &database_name "retryable-reads-tests"
+collection_name: &collection_name "files"
+
+data:
+    files:
```
jmikola: Shouldn't this be `fs.files` and `fs.chunks`? I'm pretty sure the GridFS spec tests assume the prefixes are added when fixtures are loaded, but this is a different test runner.

As for the collection name above, I wonder if you should introduce the concept of `bucket_name` and use `object: gridfsbucket` below. Otherwise, these GridFS test files seem quite inconsistent with other files in this spec.

source/retryable-reads/tests/gridfs-download-serverErrors.yml
jmikola: Second thought: since this spec has its own test runner I wonder if we can just use actual extended JSON for binary data instead of the test-specific `$hex` format. IIRC, the GridFS spec/tests were written before we had a standardized extended JSON format.

source/retryable-reads/tests/gridfs-downloadByName.yml
```diff
@@ -0,0 +1,76 @@
...
+    -
+        description: "Download_by_name succeeds on second attempt"
+        clientOptions:
+            retryReads: true
+        failPoint: &failCommand_failPoint
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+                failCommands:
+                        - find
```
jmikola: Seems like this can be `failCommands: [find]` for conciseness.

source/retryable-reads/tests/listCollectionNames-serverErrors.yml
```diff
@@ -0,0 +1,136 @@
...
+            retryReads: true
+        failPoint: &failCommand_failPoint
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data: { failCommands: [listCollections], errorCode: 11600 }
+        operation:  &retryable_operation
+            name: "listCollectionNames"
+            object: database
+            result:
+                error: false
```
jmikola: If we're not asserting anything, then simply omitting both `error` (assuming you end up moving it to `operation.error`) _and_ `result` should be fine. This goes for other tests where you may not want to assert anything other than the error condition as well (e.g. GridFS, change streams).

source/retryable-reads/tests/listCollectionNames.yml
```diff
@@ -0,0 +1,61 @@
+database_name: &database_name "retryable-reads-tests"
+collection_name: &collection_name "coll"
+
+data: []
+minServerVersion: '4.1.6' # 4.2
+
+tests:
+    -
+        description: "ListCollections succeeds on first attempt"
```
jmikola: Should the descriptions refer to `ListCollectionNames`?

source/retryable-reads/tests/gridfs-downloadByName-serverErrors.yml
```diff
@@ -0,0 +1,149 @@
...
+            filename: "abc"
+            contentType: "application/octet-stream"
+            aliases: []
+            metadata: {}
+    chunks:
+        - { _id : { "$oid" : "000000000000000000000001" }, files_id : { "$oid" : "000000000000000000000001" }, n : 0, data : { $hex : "11" } }
+
+tests:
+    -
+        description: "Download_by_name succeeds after InterruptedAtShutdown"
```
jmikola: `DownloadByName`? Likewise for other YAML files where you used underscore naming instead of camel case.

source/retryable-reads/tests/listCollectionObjects.yml
```diff
@@ -0,0 +1,61 @@
...
+data: []
+minServerVersion: '4.1.6' # 4.2
+
+tests:
+    -
+        description: "ListCollectionObjects succeeds on first attempt"
+        clientOptions:
+            retryReads: true
+        operation: &retryable_operation
+            name: "listCollectionObjects"
```
jmikola: Does this need to be quoted? Just a small inconsistency when looking at all tests together.

source/retryable-reads/tests/mapReduce-serverErrors.yml
```diff
@@ -0,0 +1,146 @@
...
+        failPoint: &failCommand_failPoint
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data: { failCommands: [mapReduce], errorCode: 11600 }
+        operation: &retryable_operation
+            name: mapReduce
+            object: collection
+            arguments:
+                map: "function inc() { return emit(0, this.x + 1) }"
+                reduce: "function sum(key, values) { return values.reduce((acc, x) => acc +x); }"
```
jmikola: These would be the first tests we have for `mapReduce` operations, but I think you should use Extended JSON for JavascriptWithoutScope BSON types here. I'm not sure if all drivers allow strings for these arguments, but all should accept Javascript BSON types.

jmikola: Also, spacing looks odd on `acc +x`. Should there be another space between `+` and `x`?

I also didn't realize the server supported `=>` syntax for its Javascript (ECMAscript 6 syntax?).

source/retryable-reads/tests/README.rst
```diff
@@ -0,0 +1,197 @@
...
+      "database", "collection", or "client."
+
+    - ``arguments``: Optional, the names and values of arguments.
+
+    - ``result``: Optional. The return value from the operation, if any. This
+      field may be a scalar (e.g. in the case of a count), a single document, or
+      an array of documents in the case of a multi-document read. If the
+      operation is expected to return an error, the ``result`` is a single
+      document that has one or more of the following fields:
+      - ``error``: Optional. If ``true``, the test should expect an error or
```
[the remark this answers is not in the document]
ShaneHarvey: I'm fine with `result.error`. I like the similarity to `result.errorCode` et al.
> As a minor plus for leaving error as a subefield of result, as it currently stands, the transactions test runner can be used to run these tests with very few modifications and (thus far) no conflicting changes

That's what I've done in Python. The major change is that these tests have a single `"operation"` while the transaction tests use a list of `"operations"`. The crud tests also use a single `"operation"` so I think that's fine too as long as we never need to run multiple operations in a single test.

source/retryable-reads/tests/listIndexes.yml
```diff
@@ -0,0 +1,60 @@
...
+        failPoint: &failCommand_failPoint
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+                failCommands:
+                        - listIndexes
+                closeConnection: true
+        operation: *retryable_operation                
+        expectations:
+             - *retryable_command_started_event
```
ShaneHarvey: This test is missing a `*retryable_command_started_event`.

source/retryable-reads/tests/listIndexNames.yml
```diff
@@ -0,0 +1,60 @@
...
+        failPoint: &failCommand_failPoint
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+                failCommands:
+                        - listIndexes
+                closeConnection: true
+        operation: *retryable_operation                
+        expectations:
+             - *retryable_command_started_event
```
ShaneHarvey: This test is missing a `*retryable_command_started_event`.

source/retryable-reads/tests/listDatabaseNames.yml
```diff
@@ -0,0 +1,59 @@
...
+        failPoint: &failCommand_failPoint
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+                failCommands:
+                        - listDatabases
+                closeConnection: true
+        operation: *retryable_operation                
+        expectations:
+             - *retryable_command_started_event
```
ShaneHarvey: This test is missing a `*retryable_command_started_event`.

source/retryable-reads/tests/listDatabases.yml
```diff
@@ -0,0 +1,59 @@
...
+        failPoint: &failCommand_failPoint
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+                failCommands:
+                        - listDatabases
+                closeConnection: true
+        operation: *retryable_operation                
+        expectations:
+             - *retryable_command_started_event
```
ShaneHarvey: This test is missing a `*retryable_command_started_event`.

source/retryable-reads/tests/mapReduce-serverErrors.yml
```diff
@@ -0,0 +1,147 @@
+database_name: &database_name "retryable-reads-tests"
+collection_name: &collection_name "coll"
+
+data:
+    - {_id: 1, x: 11}
+    - {_id: 2, x: 22}
+    - {_id: 3, x: 33}
```
ShaneHarvey: This should be:
```
    - {_id: 1, x: 0}
    - {_id: 2, x: 1}
    - {_id: 3, x: 2}
```
To match the mapReduce.yml tests. Also this fixes the expected results in this file. Right now my tests are failing with: ```AssertionError: Lists differ: [{'_id': 0.0, 'value': 69.0}] != [{'_id': 0, 'value': 6}]```

source/retryable-reads/tests/README.rst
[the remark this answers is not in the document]
jmikola: I disagree with nesting `error` under `result`. In the case of CRUD tests, an operation may raise an exception (`error: true`) but we still want to assert the BulkWriteResult included in the exception. In that case, `result` is a separate document with `insertedCount`, etc.

I imagine this doesn't come up at all in the retryable reads spec tests, since you don't deal with bulk write results, but it should explain why CRUD placed `error` as its own field parallel to `result`.

source/retryable-reads/tests/changeStreams-db.watch.yml
```diff
@@ -0,0 +1,64 @@
...
+            retryReads: true
+        operation: &retryable_operation
+            name: watch
+            object: database
+        expectations:
+            - &retryable_command_started_event
+                command_started_event:
+                    command:
+                        aggregate:
+                        $numberInt: "1"
```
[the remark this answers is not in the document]
jmikola: I know some drivers parse the YAML files (Ruby, IIRC), where `1` would parse as an integer, and there may be variation among drivers' JSON decoders. We know that the mongo shell considers `1` to be a float (which is why NumberInt and NumberLong wrappers exist), but if drivers are all parsing `1` in a JSON spec test as an integer and the assertion is strict about types, I think we can safely use `1` here. As you noted, `1` is used in other tests.

On the topic of strict type checking, I do think these APM assertions should rightfully fail if the spec test expected `1.0` and the actual command used `1` (or vice versa).

source/retryable-reads/tests/count-serverErrors.yml
jmikola: I concur that it's easy to skip if needed, since all `count()` tests are in their own file(s). LGTM.

source/retryable-reads/tests/aggregate-serverErrors.yml
```diff
@@ -0,0 +1,150 @@
...
+data:
+    - {_id: 1, x: 11}
+    - {_id: 2, x: 22}
+    - {_id: 3, x: 33}
+    
+minServerVersion: '4.1.6' # 4.2
+
+tests:
+    -
+        description: "Aggregate succeeds after InterruptedAtShutdown"
```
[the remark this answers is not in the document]
jmikola: "Count fails after two NotMaster errors" and "Count fails after NotMaster when retryReads is false" LGTM. Feel free to replicate across other methods.

source/retryable-reads/tests/gridfs-download-serverErrors.yml
jmikola: On the off chance this is a sensible improvement for the GridFS spec tickets (along with `object: gridfsbucket` and `bucket_name` at the top-level), feel free to open a SPEC ticket for GridFS so we can follow up with those improvements down the line.

source/retryable-reads/tests/gridfs-download-serverErrors.json
```diff
@@ -0,0 +1,755 @@
...
+          "$oid": "000000000000000000000003"
+        },
+        "files_id": {
+          "$oid": "000000000000000000000004"
+        },
+        "n": 0,
+        "data": {
+          "$binary": {
+            "base64": "ESIzRA==",
+            "subType": 0
```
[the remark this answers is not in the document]
vincentkam: Fixed!

source/retryable-reads/tests/listIndexes.yml
vincentkam: Noted, will fix!

source/retryable-reads/tests/listIndexNames.yml
vincentkam: Noted, will fix!

source/retryable-reads/tests/listDatabases.yml
vincentkam: Thanks, will fix!

source/retryable-reads/tests/listDatabaseNames.yml
vincentkam: Thanks, will fix!

source/retryable-reads/tests/mapReduce-serverErrors.yml
vincentkam: Apologies, will fix!

source/retryable-reads/tests/count-serverErrors.yml
```diff
@@ -0,0 +1,169 @@
...
+        description: "Count fails after NotMaster when retryReads is false"
+        clientOptions:
+            retryReads: false
+        failPoint:
+            <<: *failCommand_failPoint
+            data: { failCommands: [count], errorCode: 10107 }
+        operation:
+          <<: *retryable_operation
+          result: 
+                error: true
```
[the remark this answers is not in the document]
vincentkam: Thanks, will fix!

vincentkam:
"Will fix"es are now "fixed", with the exception of `minServerVersion`. The pair of discussed tests has been added to all "serverError" tests.

source/retryable-reads/tests/distinct.yml
```diff
@@ -0,0 +1,71 @@
...
+            retryReads: true
+        failPoint:  &failCommand_failPoint
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+                failCommands: [distinct]
+                closeConnection: true
+        operation: *retryable_operation                
+        expectations:
+             - *retryable_command_started_event
```
ShaneHarvey: Missing the second command:
```suggestion
             - *retryable_command_started_event
             - *retryable_command_started_event
```

source/retryable-reads/tests/listCollectionNames.yml
```diff
@@ -0,0 +1,59 @@
...
+        failPoint: &failCommand_failPoint
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+                failCommands:
+                        - listCollections
+                closeConnection: true
+        operation: *retryable_operation                
+        expectations:
+             - *retryable_command_started_event
```
ShaneHarvey: Missing the second command:
```suggestion
             - *retryable_command_started_event
             - *retryable_command_started_event
```

source/retryable-reads/tests/listCollections.yml
```diff
@@ -0,0 +1,59 @@
...
+        failPoint: &failCommand_failPoint
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+                failCommands:
+                        - listCollections
+                closeConnection: true
+        operation: *retryable_operation                
+        expectations:
+             - *retryable_command_started_event
```
ShaneHarvey: Missing the second command:
```suggestion
             - *retryable_command_started_event
             - *retryable_command_started_event
```

source/retryable-reads/tests/listDatabaseObjects.yml
```diff
@@ -0,0 +1,59 @@
...
+        failPoint: &failCommand_failPoint
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+                failCommands:
+                        - listDatabases
+                closeConnection: true
+        operation: *retryable_operation                
+        expectations:
+             - *retryable_command_started_event
```
ShaneHarvey: Missing the second command:
```suggestion
             - *retryable_command_started_event
             - *retryable_command_started_event
```

source/retryable-reads/tests/listCollectionObjects.yml
```diff
@@ -0,0 +1,59 @@
...
+        failPoint: &failCommand_failPoint
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+                failCommands:
+                        - listCollections
+                closeConnection: true
+        operation: *retryable_operation                
+        expectations:
+             - *retryable_command_started_event
```
ShaneHarvey: Missing the second command:
```suggestion
             - *retryable_command_started_event
             - *retryable_command_started_event
```

source/retryable-reads/tests/listIndexNames.yml
ShaneHarvey: Still missing:
```suggestion
             - *retryable_command_started_event
             - *retryable_command_started_event
```

source/retryable-reads/tests/aggregate.yml
```diff
@@ -0,0 +1,96 @@
...
+        clientOptions:
+            retryReads: true
+        operation: &retryable_operation
+            name: aggregate
+            object: collection
+            arguments:
+                pipeline:
+                    - $match: {_id: {$gt: 1}}                  
+                    - $sort: {x: 1}                        
+                batchSize: 2
```
ShaneHarvey: Setting a batchSize here causes an extraneous getMore command so I think we should remove the `batchSize` for these tests. *Note* the same problem exists in the *find* tests. 

We should also have a test that an error on a getMore is not retried for both aggregate and find.
```suggestion
```

source/retryable-reads/tests/aggregate-serverErrors.yml
```diff
@@ -0,0 +1,178 @@
...
+            data: { failCommands: [aggregate], errorCode: 11600 }
+        operation: &retryable_operation
+            name: aggregate
+            object: collection
+            arguments:
+                pipeline:
+                    - $match: 
+                        _id: {$gt: 1}
+                    - $sort: {x: 1}                        
+                batchSize: 2
```
ShaneHarvey: Remove `batchSize` because it causes extraneous getMores.
```suggestion
```

source/retryable-reads/tests/find-serverErrors.yml
```diff
@@ -0,0 +1,180 @@
...
+        clientOptions:
+            retryReads: true
+        failPoint: &failCommand_failPoint
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data: { failCommands: [find], errorCode: 11600 }
+        operation: &retryable_operation
+            name: find
+            object: collection
+            arguments: { filter: {}, sort: { _id: 1 }, limit: 4, batchSize: 2 }
```
ShaneHarvey: Remove `batchSize` because it causes extraneous getMores.
```suggestion
            arguments: { filter: {}, sort: { _id: 1 }, limit: 4 }
```

source/retryable-reads/tests/find.yml
```diff
@@ -0,0 +1,80 @@
...
+        clientOptions:
+            retryReads: true
+        operation: &retryable_operation
+            name: find
+            object: collection
+            arguments:
+                filter: {}
+                sort: {_id: 1}
+                limit: 4
+                batchSize: 2
```
ShaneHarvey: Remove `batchSize` because it causes extraneous getMores.
```suggestion
```

source/retryable-reads/tests/findOne-serverErrors.yml
```diff
@@ -0,0 +1,173 @@
...
+        failPoint: &failCommand_failPoint
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data: { failCommands: [find], errorCode: 11600 }
+        operation: &retryable_operation
+            name: findOne
+            object: collection
+            arguments:
+                filter: {_id: 1}
+                result: {_id: 1, x: 11}
```
ShaneHarvey: `result` is indented too far:
```suggestion
            result: {_id: 1, x: 11}
```

source/retryable-reads/tests/findOne.yml
```diff
@@ -0,0 +1,68 @@
...
+            retryReads: true
+        failPoint: &failCommand_failPoint
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+              failCommands: [find]
+              closeConnection: true
+        operation: *retryable_operation                
+        expectations:
+             - *retryable_command_started_event
```
ShaneHarvey: Add missing event
```suggestion
             - *retryable_command_started_event
             - *retryable_command_started_event
```

source/retryable-reads/tests/changeStreams-db.watch.yml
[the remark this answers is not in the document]
ShaneHarvey: Still waiting on this one.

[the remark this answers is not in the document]
vincentkam: Apologies, fixed!

source/retryable-reads/tests/README.rst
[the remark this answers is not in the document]
vincentkam: I'm okay with `error` appearing at the same level as `result` if it helps future tests, even if retryable-reads admittedly doesn't need it.

source/retryable-reads/tests/changeStreams-client.watch-serverErrors.yml
```diff
@@ -0,0 +1,169 @@
...
+        operation: &retryable_operation
+            name: watch
+            object: client
+        expectations:
+            - &retryable_command_started_event
+                command_started_event:
+                    command:
+                        aggregate: 1
+                        cursor: {}
+                        pipeline: [ { $changeStream: { fullDocument: "default" } } ]
```
ShaneHarvey: This needs the allChangesForCluster option too:
```suggestion
                        pipeline: [ { $changeStream: { fullDocument: "default", allChangesForCluster: true } } ]
```

source/retryable-reads/tests/changeStreams-client.watch-serverErrors.yml
```diff
@@ -0,0 +1,169 @@
...
+                    database_name: *database_name
```
ShaneHarvey: Db should be admin for cluster level changes.
```suggestion
                    database_name: admin
```

source/retryable-reads/tests/README.rst
```diff
@@ -0,0 +1,158 @@
...
+
+Test Format
+-----------
+
+Each YAML file has the following keys:
+
+- ``database_name`` and ``collection_name``: Optional. The database and
+  collection to use for testing.
+  
+- ``bucket_name``: Optional. The GridFS bucket name to use for testing.
```
ShaneHarvey: A few questions about the gridfs tests:
1. Can we provide more guidance on how to run the GridFS tests? We should mention how to create the GridFSBucket and that "data" will (optionally?) have "fs.files" and "fs.chunks".
1. What's the goal of "bucket_name"? Is it just to inform the runner that this is a gridfs test?
1. On that point, can you mention what should be the canonical way to tell that this is a gridfs test? Do we look for "gridfs" in the filename, the presence of "bucket_name", the presence of "fs.files" in "data"?

source/retryable-reads/tests/listDatabaseObjects-serverErrors.yml
```diff
@@ -0,0 +1,163 @@
...
+    -
+        description: "ListDatabaseObjects succeeds after InterruptedAtShutdown"
+        clientOptions:
+            retryReads: true
+        failPoint: &failCommand_failPoint
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data: { failCommands: [listDatabases], errorCode: 11600 }
+        operation:  &retryable_operation
+            name: listDtabaseObjects
```
ShaneHarvey: There's a typo here, should be `listDatabaseObjects`.

source/retryable-reads/tests/README.rst
```diff
@@ -0,0 +1,158 @@
...
+        configureFailPoint: "failCommand",
+        mode: "off"
+    });
+
+Network Error Tests
+===================
+
+Network error tests are expressed in YAML and should be run against a standalone,
+shard cluster, or single-node replica set. Until `SERVER-34943`_ is resolved,
+these tests cannot be run against a multi-node replicaset.
```
ShaneHarvey: The change streams tests cannot be run on a standalone. I think we can add a tag to only run those tests on sharded clusters and replica sets. Maybe we can adopt the `topology: ["replicaset", "sharded"]` format from [the change stream tests themselves](https://github.com/mongodb/specifications/tree/master/source/change-streams/tests#spec-test-format)?

vincentkam:
I really like the "speeding up tests" section idea and have added it!

source/retryable-reads/tests/README.rst
vincentkam: I like that: I've taken your suggestion!

source/retryable-reads/tests/aggregate-serverErrors.yml
```diff
@@ -0,0 +1,173 @@
...
+    -
+        description: "Aggregate succeeds after InterruptedAtShutdown"
+        clientOptions:
+            retryReads: true
+        failPoint: &failCommand_failPoint
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data: { failCommands: [aggregate], errorCode: 11600 }
+        operation: &retryable_operation_succeeds
+            <<: &retryable_operation
```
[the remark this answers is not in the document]
vincentkam: 😺 😺 😺

vincentkam:
Manually rebased, squashed, and pushed!
https://github.com/mongodb/specifications/commit/973c7ffff796e426512a7e9f74b8c64c0b880e30

PR #435, opened 2018-12-14
[WIP] Convenient API for Transactions spec tests

jmikola:
https://jira.mongodb.org/browse/DRIVERS-556

This is a WIP PR intended for early review. Please ignore if you're not advising the spec or implementing a POC.


source/convenient-api-for-transactions/tests/README.rst
```diff
@@ -0,0 +1,107 @@
...
+    - ``name``: The name of the operation on ``object``.
+
+    - ``object``: The name of the object on which to perform the operation. The
+      value will be either "collection" or "session0".
+
+    - ``arguments`` (optional): Names and values of arguments to pass to the
+      operation.
+
+    - ``error`` (optional): If ``true``, the test should expect the operation
+      to emit an an error or exception. If ``false`` or omitted, drivers MUST
```
p-mongo: `an` duplicated

p-mongo:
Can the top level directory be renamed to something starting with `transactions` to be next to the other transactions directory?

source/transactions-convenient-api/tests/README.rst
```diff
@@ -0,0 +1,107 @@
...
+- ``tests``: An array of tests that are to be run independently of each other.
+  Each test will have some or all of the following fields:
+
+  - ``description``: The name of the test.
+
+  - ``failPoint`` (optional): The ``configureFailPoint`` command document to run
+    to configure a fail point on the primary server.
+
+  - ``sessionOptions`` (optional): Names and values of options to pass to
+    ``MongoClient.startSession()``.
```
jmikola: Add `clientOptions`, since we now have some tests for that.

source/transactions-convenient-api/tests/README.rst
```diff
@@ -0,0 +1,107 @@
...
+
+    - ``name``: The name of the operation on ``object``.
+
+    - ``object``: The name of the object on which to perform the operation. The
+      value will be either "collection" or "session0".
+
+    - ``arguments`` (optional): Names and values of arguments to pass to the
+      operation.
+
+    - ``error`` (optional): If ``true``, the test should expect the operation
```
jmikola: Let's replace this with `result` syntax from Transactions spec runner for consistency. This spec won't assert much about errors -- likely only that errors from `withTransactions` _don't_ have certain labels.

Additionally, we can add a test for the callback yielding a duplicate key error, which should propagate up through `withTransaction`.

source/transactions-convenient-api/tests/callback-aborts.json
```diff
@@ -0,0 +1,223 @@
...
+            "command": {
+              "insert": "test",
+              "documents": [
+                {
+                  "_id": 2
+                }
+              ],
+              "ordered": true,
+              "lsid": "session0",
+              "txnNumber": {
```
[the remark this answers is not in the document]
jmikola: Fixed in last commit. I now assert that `autocommit` is not present and ignore `txnNumber` entirely. Asserting that both `autocommit` and `startTransaction` are not present effectively asserts that the command is not part of a transaction.

source/transactions-convenient-api/tests/callback-commits.json
```diff
@@ -0,0 +1,285 @@
...
+            "command": {
+              "insert": "test",
+              "documents": [
+                {
+                  "_id": 3
+                }
+              ],
+              "ordered": true,
+              "lsid": "session0",
+              "txnNumber": {
```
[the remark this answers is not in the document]
jmikola: Same solution as `callback-aborts.json`.

source/transactions-convenient-api/tests/callback-retry.json
```diff
@@ -0,0 +1,279 @@
...
+              "autocommit": false,
+              "readConcern": null,
+              "startTransaction": null,
+              "writeConcern": null
+            },
+            "command_name": "insert",
+            "database_name": "withTransaction-tests"
+          }
+        }
+      ],
```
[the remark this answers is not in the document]
jmikola: This relates to two transaction spec tests. In ["abort ignores TransactionAborted" from abort.yml](https://github.com/mongodb/specifications/blob/b093e4d9d8eaca421a55308a10222907e8934c8e/source/transactions/tests/abort.yml#L279), the test performs an additional insert after the DuplicateKey error to assert that the server has aborted the transaction on its own, and then has the client send its own abort, which tests that we can abort and already-aborted transaction. I don't think we need to repeat that test here.

In ["DuplicateKey errors do not contain transient label" from error-labels.yml](https://github.com/mongodb/specifications/blob/b093e4d9d8eaca421a55308a10222907e8934c8e/source/transactions/tests/error-labels.yml#L7), the test simply asserts that the failed insert has the appropriate error label and then executes an abort on its own to ensure things are cleaned up. That seems like the more appropriate behavior here, so I'll just add an expected abort command to this test.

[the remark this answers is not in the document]
jmikola: Fixed in last commit. Please resolve if that works.

source/transactions-convenient-api/tests/commit-transienttransactionerror.yml
```diff
@@ -0,0 +1,202 @@
...
+            txnNumber: { $numberLong: "1" }
+            autocommit: false
+            # omitted fields
+            readConcern: ~
+            startTransaction: ~
+            writeConcern: ~
+          command_name: commitTransaction
+          database_name: admin
+      -
+        command_started_event:
```
jyemin: According to the transaction spec, a NotMaster error is an ordinary retryable write error, and therefore the existing commitTransaction implementation will already retry it without ever having to loop around in withTransaction.  So I'd expect there to be three events here: insert, commitTransaction, and an identical commitTransaction as the last one.

I think most of the tests in this file (and maybe a few others) have this problem.

source/transactions-convenient-api/tests/commit-transienttransactionerror.yml
```diff
@@ -0,0 +1,202 @@
...
+          operations:
+            -
+              name: insertOne
+              object: collection
+              arguments:
+                session: session0
+                document: { _id: 1 }
+              result:
+                insertedId: 1
+        result:
```
jyemin: There shouldn't be a result here, as the transaction ultimately succeeds and therefore any transient errors are swallowed.  Other tests in this file have the same problem.

source/transactions-convenient-api/tests/callback-retry.yml
```diff
@@ -0,0 +1,199 @@
...
+            documents:
+              - { _id: 1 }
+            ordered: true
+            lsid: session0
+            # txnNumber is incremented when retrying the transaction
+            txnNumber: { $numberLong: "2" }
+            startTransaction: true
+            autocommit: false
+            # omitted fields
+            readConcern: ~
```
jyemin: This insert and the next require a readConcern with afterClusterTime, as the session is causally consistent and abortTransaction sets the operationTime.

source/transactions-convenient-api/tests/commit-transienttransactionerror.yml
jmikola: This makes sense. I think the correct fix is to bump the fail point's `mode.times` up to 2 so we fail on both the original attempt and the internal retry attempt. Control would then be returned to `withTransaction()` and we can repeat the entire transaction and succeed with the insert and commitTransaction commands (no retries needed).

Also, since `withTransaction()` ultimately succeeds after the retry, I don't think I should be expecting an error at all. That jives with what you mentioned in https://github.com/mongodb/specifications/pull/435#discussion_r245122170.

While we'll no longer be asserting details about the error, I do want to confirm that we rightfully expect a [TransientTransactionError](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#transienttransactionerror) label on the `commitTransaction` command result, since it's added by the server (we don't request that the fail point close the connection). [UnknownTransactionCommitResult](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#unknowntransactioncommitresult) is only ever added by the driver in certain cases (e.g. network, server selection errors), none of which apply here since we do receive a server response.

source/transactions-convenient-api/tests/commit-transienttransactionerror.yml
jmikola: Acknowledged in https://github.com/mongodb/specifications/pull/435#discussion_r245136611. Will fix.

source/transactions-convenient-api/tests/callback-retry.yml
jyemin: I concur

source/transactions-convenient-api/tests/commit-writeconcernerror.yml
```diff
@@ -0,0 +1,196 @@
+database_name: &database_name "withTransaction-tests"
+collection_name: &collection_name "test"
+
+data: []
+
+tests:
+  -
+    description: commitTransaction is not retried after write concern error (WriteConcernFailed)
```
jyemin: My reading of the [transactions spec](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#unknowntransactioncommitresult) is that this should be labelled as "UnknownTransactionCommitResult" and so be retried.

source/transactions-convenient-api/tests/commit-writeconcernerror.yml
```diff
@@ -0,0 +1,196 @@
...
+            autocommit: false
+            # omitted fields
+            readConcern: ~
+            startTransaction: ~
+            writeConcern: ~
+          command_name: commitTransaction
+          database_name: admin
+    outcome:
+      collection:
+        data: []
```
jyemin: This expectation is not being satisfied.  It seems that the server commits the transaction even though it responds with a writeConcernError.

source/transactions-convenient-api/tests/commit-writeconcernerror.yml
```diff
@@ -0,0 +1,196 @@
...
```
jyemin: This expectation is not being satisfied. It seems that the server commits the transaction even though it responds with a writeConcernError.

Not sure if there's a way to indicate in the failpoint that the commit shouldn't succeed.

source/transactions-convenient-api/tests/commit-transienttransactionerror.yml
```diff
@@ -0,0 +1,191 @@
...
+          closeConnection: false
+    operations:
+      -
+        name: withTransaction
+        object: session0
+        callback: *callback
+    expectations: *expectations
+    outcome: *outcome
+  -
+    description: transaction is retried after commitTransaction TransientTransactionError (PreparedTransactionInProgress)
```
jyemin: What makes this one retryable?

jyemin: For `251 # NoSuchTransaction` above, the server is adding the transient TransientTransactionError label in the response, but that's not happening for this one.  Perhaps this requires a 4.2 server?  Is does seem like a distributing transaction-related error code.

source/transactions-convenient-api/tests/callback-retry.yml
p-mongo: I have the same thoughts as Jeff on adding cluster time to the two inserts in this file.

source/transactions-convenient-api/tests/commit-transienttransactionerror.yml
```diff
@@ -0,0 +1,202 @@
...
+            # omitted fields
+            readConcern: ~
+            startTransaction: ~
+            writeConcern: ~
+          command_name: commitTransaction
+          database_name: admin
+      -
+        command_started_event:
+          command:
+            insert: *collection_name
```
p-mongo: I do not have this insert because the fail point produces a UNKNOWN_TRANSACTION_COMMIT_RESULT_LABEL and I retry the commit in with_transaction rather than from the top.

source/transactions-convenient-api/tests/commit-writeconcernerror.yml
jmikola: Fixed in https://github.com/mongodb/specifications/pull/435/commits/e747584bd473ae78533968b2abe96cc0e91c7d45. Please acknowledge and resolve if that works.

source/transactions-convenient-api/tests/commit-writeconcernerror.yml
jmikola: This seems like the same issue you raised in https://github.com/mongodb/specifications/pull/435#discussion_r245158138. See my responses there and acknowlege if the commit resolves things.

Based on the original transaction spec tests, write ops are still applied after write concern error returned by a fail point, so I think it's correct for all of the `commit-writeconcernerror.yml` tests to expect a single `{ _id: 1 }` document in their outcome data.

source/transactions-convenient-api/tests/commit-writeconcernerror.yml
jyemin: Looks good

source/transactions-convenient-api/tests/callback-retry.yml
jyemin: Not quite.  There's another insert below this one that also requires a read concern.

source/transactions-convenient-api/tests/commit-writeconcernerror.yml
```diff
@@ -0,0 +1,176 @@
...
+
+tests:
+  -
+    description: commitTransaction is not retried after write concern error (WriteConcernFailed)
+    failPoint:
+      configureFailPoint: failCommand
+      mode: { times: 1 }
+      data:
+          failCommands: ["commitTransaction"]
+          closeConnection: false
```
jyemin: Sorry, thought I had already commented on this, but I think I only mentioned it in Slack: 

None of the failpoints in this file work unless the `closeConnection : false` is removed.  With it, the writeConcernError is not included in the response.

Please remove all the closeConnection fields (as false is the default anyway).

source/transactions-convenient-api/tests/commit-writeconcernerror.yml
jmikola: If I understand correctly, you're saying that the driver's internal retry-once behavior (documented under [UnknownTransactionCommitResult](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#unknowntransactioncommitresult)) will result in `commitTransaction` being retried and succeeding on the second attempt. I just added some more content to the original spec document to talk about this, so please give that a read.

I think the right thing to do here is to configure the fail point to occur twice, to allow control to return to `withTransaction` so we can rightfully decide _not_ to retry.

source/transactions-convenient-api/tests/commit-transienttransactionerror.yml
jyemin: Not working the way you expected.  "UnknownTransactionCommitResult" is indeed being added, and so WithTransaction just retries the commit rather than the whole transaction.  Here's the Java driver's implementation that is adding that label: https://github.com/mongodb/mongo-java-driver/blob/757c3279bb08738e2209d38285b18a9ef3d0ab35/driver-core/src/main/com/mongodb/operation/CommitTransactionOperation.java#L75-L100

See https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#unknowntransactioncommitresult for the transactions spec section that this corresponds to.

jyemin: The tests below that use error codes 79 and 100 should exercise the code path that retries the whole transaction after a server error response to commitTransaction.

source/transactions-convenient-api/tests/commit-writeconcernerror.yml
jyemin: Yes, that's what I'm saying.  Where should I look in the original spec document exactly?

source/transactions-convenient-api/tests/commit-writeconcernerror.yml
jmikola: Removed in https://github.com/mongodb/specifications/pull/435/commits/cb72ce2b81408ba345f2f4006948ff697c32aad9

source/transactions-convenient-api/tests/commit-transienttransactionerror.yml
jmikola: Should be fixed by https://github.com/mongodb/specifications/pull/435/commits/c09a0c5c81e970b8539a05d1cb3a55a31e1b3be5.

source/transactions-convenient-api/tests/commit-transienttransactionerror.yml
jmikola: Indeed, that was introduced in 4.1.6. See https://github.com/mongodb/specifications/pull/435/commits/48d54031d7468880b65edefce9bd37746a8c3dd1

source/transactions-convenient-api/tests/commit-transienttransactionerror.yml
jmikola: This should also be fixed now.

jyemin:
LGTM!

source/transactions-convenient-api/tests/transaction-options.yml
```diff
@@ -0,0 +1,258 @@
...
+            autocommit: false
+            writeConcern: { w: 1 }
+            # omitted fields
+            readConcern: ~
+            startTransaction: ~
+          command_name: commitTransaction
+          database_name: admin
+    outcome: *outcome
+  -
+    description: withTransaction explicit transaction options override client options
```
samantharitter: I think you may have mixed up the descriptions for this test and the one that begins on L223

jmikola: Fixed in https://github.com/mongodb/specifications/pull/435/commits/6e91e113d22e512706a05dc3d65d54ec06e1ae6e. Thanks!

source/transactions-convenient-api/tests/commit-writeconcernerror.yml
```diff
@@ -0,0 +1,269 @@
...
+  -
+    # This test configures the fail point to return a contrived error that
+    # is WriteConcernFailed but omits the errInfo field that would identify it
+    # as a wtimeout error. This tests that drivers do not assume that all
+    # WriteConcernFailed errors are due to a replication timeout.
+    description: commitTransaction is retried after WriteConcernFailed non-timeout error
+    skipReason: SPEC-1185 Drivers should use majority write concern when retrying commitTransaction
+    failPoint:
+      configureFailPoint: failCommand
+      mode: { times: 2 }
```
p-mongo: I get half as many commitTransactions failed as specified here (2 produces 1 failure, 4 produces 2 failures).

p-mongo: On 4.0 and 4.1 development versions.

source/transactions-convenient-api/tests/commit-writeconcernerror.yml
```diff
@@ -0,0 +1,269 @@
...
+    # The write operation is still applied despite the write concern error
+    outcome: &outcome
+      collection:
+        data:
+          - { _id: 1 }
+  -
+    # This test configures the fail point to return a contrived error that
+    # is WriteConcernFailed but omits the errInfo field that would identify it
+    # as a wtimeout error. This tests that drivers do not assume that all
+    # WriteConcernFailed errors are due to a replication timeout.
```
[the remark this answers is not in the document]
jmikola: Done: https://github.com/mongodb/specifications/pull/457. Thanks!

PR #436, opened 2018-12-17
SPEC-1170: SDAM monitoring multi phase testing

p-mongo:
Background: currently, there are tests for SDAM state transitions ("sdam") and some of the events generated by SDAM flow ("sdam monitoring"), but there are areas not covered by the existing monitoring tests, specifically:

- server removal/server close events
- some situations when the first response is not from primary
- multiple servers responding

The existing "sdam" tests cover topology outcomes in most of these cases, but those tests do not specify the order in which events should be published (or event contents, but this is a minor point).

Therefore, #410 proposes to add a number of "sdam monitoring" tests to formalize the order of sdam events being published in various situations.

In order for the new tests to be possible, a change in spec runner is necessary which is to perform assertions on events generated during cluster/topology construction prior to any responses being received from any of the servers. This PR is extracted from #410 and contains the readme change explaining how to handle the first phase consisting of assertions only without any responses, as well as a sample test broken up into two phases. The sample test illustrates having a dedicated phase for events published during client construction prior to processing any sdam responses.

### DIFF
--- a/source/server-discovery-and-monitoring/tests/README.rst
+++ b/source/server-discovery-and-monitoring/tests/README.rst
@@ -25,7 +25,9 @@ Each YAML file has the following keys:
 
 Each phase object has two keys:
 
-- responses: An array of "response" objects.
+- responses: (optional) An array of "response" objects. If not provided,
+  the test runner should construct the client and perform assertions specified
+  in the outcome object without processing any responses.
 - outcome: An "outcome" object representing the TopologyDescription.
 
 A response is a pair of values:
--- a/source/server-discovery-and-monitoring/tests/README.rst
+++ b/source/server-discovery-and-monitoring/tests/README.rst
@@ -17,12 +17,13 @@ Format
 
 Each YAML file has the following keys:
 
-- description: Some text.
+- description: A textual description of the test.
 - uri: A connection string.
 - phases: An array of "phase" objects.
+  A phase of the test optionally sends inputs to the client,
+  then tests the client's resulting TopologyDescription.
 
-A "phase" of the test sends inputs to the client, then tests the client's
-resulting TopologyDescription. Each phase object has two keys:
+Each phase object has two keys:
 
 - responses: An array of "response" objects.
 - outcome: An "outcome" object representing the TopologyDescription.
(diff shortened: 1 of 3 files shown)


source/server-discovery-and-monitoring/tests/README.rst
```diff
@@ -93,19 +105,29 @@ Set the client's initial TopologyType to ReplicaSetNoPrimary.
...
+
 Test Phases
 ~~~~~~~~~~~
 
 For each phase in the file, parse the "responses" array.
 Pass in the responses in order to the driver code.
 If a response is the empty object `{}`, simulate a network error.
 
-Once all responses are processed, assert that the phase's "outcome" object
+For non-monitoring tests,
```
jmikola: Is the early line break intentional?

source/server-discovery-and-monitoring/tests/monitoring/replica_set_with_removal.yml
```diff
@@ -52,6 +37,25 @@ phases:
           server_opening_event:
             topologyId: "42"
             address: "b:27017"
+  -
```
jmikola: It looks like this change moves `responses` from the first object in the `phases` array to the second. Previously, the first object has both `responses` and `outcome` fields and the second object had only an `outcome` field.

This doesn't jive with the test README, which says that each object in the `phases` array has two keys. Could we add some language explaining that `responses` is optional?

source/server-discovery-and-monitoring/tests/README.rst
```diff
@@ -17,12 +17,13 @@ Format
...
-- description: Some text.
+- description: A textual description of the test.
 - uri: A connection string.
 - phases: An array of "phase" objects.
+  A phase of the test optionally sends inputs to the client,
+  then tests the client's resulting TopologyDescription.
 
-A "phase" of the test sends inputs to the client, then tests the client's
-resulting TopologyDescription. Each phase object has two keys:
+Each phase object has two keys:
```
jmikola: Per my comment below, I think this should read "has the following keys" and then "(optional)" should be added after the "responses" key in a manner consistent with how we document optional fields in other specs (see: https://github.com/mongodb/specifications/tree/master/source/crud/tests).

p-mongo:
The spec ticket which this PR falls under is https://jira.mongodb.org/browse/SPEC-1170.

source/server-discovery-and-monitoring/tests/README.rst
p-mongo: Yes, it makes the file readable in a terminal as well as makes the diffs lower in size and more readable.

source/server-discovery-and-monitoring/tests/monitoring/replica_set_with_removal.yml
p-mongo: Added.

IanWhalen:
@jmikola are you cool with merging this?

jmikola:
>  are you cool with merging this?

My previous LGTM still stands. I only requested that the commit/PR be revised to include the SPEC ticket number.

PR #437, opened 2018-12-18
SPEC-1184 add `object` field to CRUD spec tests

kvwalker:


source/crud/tests/db/db-aggregate.yml
```diff
@@ -1,11 +1,13 @@
+database_name: &database_name "admin"
```
jmikola: I assume this is only used to denote the database-under-test? I just realized that we don't use `collection_name` in any other CRUD tests -- perhaps because they never cared about what collection they used.

source/crud/tests/db/db-aggregate.yml
```diff
@@ -1,11 +1,13 @@
...
+
 data: []
 
 tests:
     -
         description: "Aggregate with $currentOp"
         operation:
             name: db-aggregate
-            database_name: admin
+            object: database
```
jmikola: I wonder if `name` should be changed to `aggregate` since we can now use `object` to differentiate. Either way, drivers that don't yet support database-level aggregation would fail on this spec test due to `db-aggregate` not being recognized or attempting to run `$currentOp` on a collection-level aggregation.

jmikola: @ShaneHarvey: thoughts?

ShaneHarvey: Yeah I like the idea of changing name to simply `aggregate`.

source/crud/tests/db/db-aggregate.yml
kvwalker: Yes, these tests need to use the `admin` database so I suppose the other CRUD tests don't care about which collection they're using.

source/crud/tests/README.rst
```diff
@@ -58,6 +60,9 @@ Each YAML file has the following keys:
     - ``name``: The name of the operation as defined in the specification. The
       name `db-aggregate` refers to database-level aggregation.
 
+    - ``object``: The name of the object to perform the operation on. Can be
+      "database" or "collection". Defaults to "collection" if undefined.
+
     - ``database_name`` (optional): The name of the database against which to execute the operation.
```
rozza: I think this can be removed as `database_name` has moved to a top level field.

jmikola: The [tests](https://github.com/mongodb/specifications/blob/master/source/crud/tests/db/db-aggregate.yml) don't appear to be using the top-level `database_name` field yet. These tests are a bit unique in that they're running `$currentOp`, which is outside the scope of a single database. While "admin" is the target database on which the aggregation is executed, it's not technically the database under test (in the same way a collection would be in other tests).

That said, I do agree we should use one of the other. A top-level `database_name` key would be consistent with other spec tests.

jmikola:
> LGTM, but I propose we lump making DRIVERS tickets for this into Jeremy's CRUD spec updates using command monitoring. This will require every driver to make test runner changes, so let's reduce that pain.

SGTM. @kvwalker: Feel free to add this issue to [SPEC-1144](https://jira.mongodb.org/browse/SPEC-1144), which is an epic.

source/crud/tests/README.rst
jmikola: @rozza: Good to resolve?

rozza: All good - thanks :+1:

PR #438, opened 2019-01-03
Test leading zeroes in datetime ms

kevinAlbs:
This tests the bug discovered in [CDRIVER-2910](https://jira.mongodb.org/browse/CDRIVER-2910). I see an argument being made that the bug is pretty specific and likely wouldn't catch bugs in other drivers. But since the fix should be tested in the C driver, it seemed wasteful not to add it here too.


mbroadst:
The more tests the better!

PR #439, opened 2019-01-03
Spec 1153: add read concern support to aggregation $out

prashantmital:


source/read-write-concern/tests/README.rst
```diff
@@ -52,6 +52,37 @@ array of test case objects, each of which have the following keys:
...
+  - ``description``: The name of the test.
+  - ``operation``: Document describing the operation to be executed. This will
+    have the following fields:
+    - ``name``: The name of the operation as defined in the specification. The
+      name `db-aggregate` refers to database-level aggregation.
+    - ``arguments``: The names and values of arguments from the specification.
+  - ``error``: If ``true``, the test should expect an error or exception. Note
+      that some drivers may report server-side errors as a write error within a
+      write result object.
+  - ``readConcernDocument:`` A document indicating the read concern to be sent to the server.
```
jmikola: I have a few concerns with these tests:

1. This is adding CRUD tests to the read/write concern spec, which already has their own test format
2. We're introducing a new `readConcernDocument` field, which has no prior art in any other tests and comes off as a a stopgap until we have APM assertions on CRUD tests (see: [SPEC-1133](https://jira.mongodb.org/browse/SPEC-1133)).
3. The test cases themselves specify `readConcern` under the `arguments` field of the "aggregate" operation. AFAIK, none of our spec tests specify per-operation read/write concern options, since that is not a required feature in any of the tests. Instead, there is syntax in some spec test runners for providing such options at the client, database, or collection level.

To address point 2, I would propose that we add this ticket to the [SPEC-1144](https://jira.mongodb.org/browse/SPEC-1144) epic, or at least have its tests depend on [SPEC-1133](https://jira.mongodb.org/browse/SPEC-1133).

To address point 3, I would propose we utilize `operations.collectionOptions` as is done in the transactions spec test runner. AFAIK, the CRUD spec tests also don't use that syntax yet, but it makes sense to introduce that as we standardize on a common format.

source/read-write-concern/read-write-concern.rst
```diff
@@ -94,7 +94,11 @@ The read concern option is available for the following operations:
 - ``geoSearch`` command
 - ``mapReduce`` command with {out : inline} output option
 
-``aggregate`` command with $out and ``mapReduce`` command with "out" set to anything other than "inline" do not support readConcern.
+
+Starting in MongoDB 4.2, the ``aggregate`` command with $out does not support the "linearizable" readConcern.
+Versions older than 4.2 do not support readConcern for the ``aggregate`` command with $out.
+
```
jmikola: The first sentence is not entirely clear that levels other than "linearizable" are supported. I wonder if we can improve it.

> Starting in MongoDB 4.2, an `aggregate` command with `$out` supports a read concern; however, it does not support the "linearizable" level (attempting to do so will result in a server error). Server versions before 4.2 do not support a read concern at all for `aggregate` commands with `$out`.

If the collection's read concern level is "linearizable", what do we want drivers to do?

 * Avoid inheriting the read concern from the collection and send the `aggregate` command without it
 * Blindly inherit the read concern and allow the server to error
 * Preemptively raise an error to the user (probably a worse choice than allowing the server to error, as this is not very forward-compatible if things change in future server versions)

I'm leaning towards drivers being aware that "linearizable" is permitted when they decide to inherit a collection-level read concern for an `aggregate()` method. This may be consistent with other behavior where drivers choose to inherit a write concern only if the selected server supports it for the given command (e.g. `findAndModify`).

Note: I may have missed https://github.com/mongodb/specifications/commit/4341779d81b9cd451303800a152c0c808a994035 originally adding that no `aggregate` commands with `$out` support a read concern at all. Previously, PHPLIB would only avoid inheriting the collection's read concern if its level was "majority" (since "local" was OK) or if we were in a transaction (see: [`MongoDB\Collection::aggregate()`](https://github.com/mongodb/mongo-php-library/blob/1.4.2/src/Collection.php#L204) and similar for [`mapReduce()`](https://github.com/mongodb/mongo-php-library/blob/1.4.2/src/Collection.php#L941)). I might have missed this if [SPEC-970](https://jira.mongodb.org/browse/SPEC-970) predated the "driver changes needed" attribute (I don't see any drivers issues linked to the spec change).

source/read-write-concern/tests/README.rst
prashantmital: @jmikola I didn't know about SPEC-1133, but it seems like a great idea to finish that work before adding tests for this. I decided to write these tests the way I because CRUD tests didn't have command monitoring support, and write/read concern tests didn't have support for specifying commands - which resulted in the current hodgepodge.

source/read-write-concern/read-write-concern.rst
prashantmital: I'm in favor of the second option: 

> Blindly inherit the read concern and allow the server to error

When an aggregate command is encountered with `$out`, drivers should run a simple check to determine the server version. If the version is <4.2, the readConcern is removed and a warning message logged. If it is deemed to be >4.2  then we blindly pass along the readConcern. Rationale: if we decide to make the driver readConcern 'aware' for the case of linearizable, then it stands to reason that we would need to perform some kind of validation on the user-provided readConcern document which is explicitly forbidden by our spec: 

https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#unknown-levels-and-additional-options-for-string-based-readconcerns

jmikola: I don't think the behavior I'm describing violates the [read concern spec's prohibition on level string validation](https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#unknown-levels-and-additional-options-for-string-based-readconcerns). We're not validating the value provided by the user (be it through the connection string or our [ReadConcern constructor](http://php.net/manual/en/mongodb-driver-readconcern.construct.php)). Instead, we're _inspecting_ the value when the `aggregate()` helper is invoked to decide if it's safe to inherit for the given server version. If the user provided an "!invalid" string for their read concern, PHPLIB would happily send that along to the server and allow it to yield an error, which I believe is the intention of the section you quoted in the read/write concern spec.

prashantmital: Makes sense. In that case, I'd support having drivers inspect readConcern and not passing it along to `db.aggregate()` if it is linearizable (and the pipeline contains a `$out`). The main reason I'd be in favor of this pattern is because in its absence, running aggregation pipelines with an `$out` stage against collections initialized with the "linearizable" readConcern would not be possible without having to create a new Collection object with a `$out`-compatible readConcern. 

On the other hand, this might give users the false impression that their linearizable readConcern is being honored by the aggregation pipeline when in fact it is not (sure we would log a warning but I'm not sure how many users actually check the warning messages as long as their code executes). In that case, it might be better to error out to make it clear to the user that they are using a readConcern that is incompatible with their aggregation pipeline. AFAICT most drivers are equipped with helper methods that make it easy to instantiate new Collection objects with modified readConcern (e.g. `pymongo` has `Collection.with_options` and Java has `MongoCollection.withReadConcern`). These should make it simple enough for users to create a new collection/db with a compatible readConcern to run their pipeline.

jmikola: Whichever route we go with, I think the CRUD tests will ultimately be able to borrow concepts from the transactions spec test format, which allows us to apply specific options to the client, database, or collection object under test. Combined with APM, we could then assert that even invalid read concerns (e.g. "linearizable" for an `$out` pipeline) are assembled into the outgoing command.

source/read-write-concern/tests/README.rst
jmikola: See: https://github.com/mongodb/specifications/pull/453

source/read-write-concern/read-write-concern.rst
behackett: I think we should just pass linearizable to the server and let it error. For two reasons:
- Someday linearlizable might actually be supported. It would be nice to not have to change drivers at that point.
- It seems weird to me that we would error driver side on linearlizable, but not on read concerns that don't exist. Let's just leave it to the serve to decide.

behackett: Otherwise, I agree with @jmikola on the text.

prashantmital:
Superseded by #475

PR #444, opened 2019-01-14
SPEC-1165: Clarify use of random numbers

derickr:
State that they don't need to be cryptographically secure, and add a test to
test that the unique value is different in forked processes.

### DIFF
--- a/source/objectid.rst
+++ b/source/objectid.rst
@@ -80,14 +80,16 @@ timestamp value.
 Random Value
 ------------
 
-A 5-byte field consisting of a random value generated once per process.
+A 5-byte field consisting of a random value generated once per process. This
+random value is unique to the machine and process.
 
 Drivers MUST NOT have an accessor method on an ObjectID class for obtaining
 this value.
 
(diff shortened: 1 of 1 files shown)


source/objectid.rst
```diff
@@ -85,6 +85,10 @@ A 5-byte field consisting of a random value generated once per process.
 Drivers MUST NOT have an accessor method on an ObjectID class for obtaining
 this value.
```
kevinAlbs: Can we also reiterate "The random value is unique to the machine and process"?

Maybe we decided against adding recommendations in a previous review, but I think a recommendation on how to generate this random number might be helpful. E.g.
"If possible, use a PRNG with OS supplied entropy that SHOULD NOT block to wait for more entropy to become available. An example is Python's random.SystemRandom. Otherwise, seed a deterministic PRNG to ensure uniqueness of process and machine by combining time, process ID, and hostname"

derickr: I've updated the text to your suggestion, but left out the Python example.

kevinAlbs: LGTM!

PR #447, opened 2019-01-22
WRITING-3780 Client side encryption spec tests

kevinAlbs:
A draft for the client side encryption spec tests.

- As the limitations of mongocryptd become known, we'll add tests to exercise those limitations.
- After mockupCryptd and the mock KMS server support failpoints, we'll test different error scenarios.
- The tests don't have data for encrypted cipher text, key UUID's, or key material. However, the helper script included, `generate.py`, substitutes those values to make it easy to update.


source/client-side-encryption/tests/README.rst
```diff
@@ -0,0 +1,223 @@
...
+    - ``collectionOptions``: Optional, parameters to pass to the Collection()
+      used for this operation.
+
+    - ``command_name``: Present only when ``name`` is "runCommand". The name
+      of the command to run. Required for languages that are unable preserve
+      the order keys in the "command" argument when parsing JSON/YAML.
+
+    - ``arguments``: Optional, the names and values of arguments.
+
+    - ``result``: The return value from the operation, if any. This field may
```
jmikola: Since `operations` is plural and contains a list of operation documents, I think this field belongs within each operation. AFAIK, that is what the transactions spec tests already do and it's what I've implemented in my PR to revise the CRUD spec test format in https://github.com/mongodb/specifications/pull/453.

Note that Transactions and CRUD still diverge a bit because CRUD does not represent errors under `result` but instead has a separate `error` field. This is mainly because a BulkWriteException may have a BulkWriteResult in it and we want to assert (a) that error occurred and (b) some state about the result object. I don't think that's relevant in your case, so no objections if you want to continue using the Transactions format here (with `errorContains` and `errorCodeName`) for simplicity.

kevinAlbs: It is within each operation, unless I indented it wrong. Example: https://github.com/mongodb/specifications/pull/447/files#diff-d25b8fce66348003fdf8d126eda5c8ccR76

jmikola: Ah, my mistake. I've been looking at too much RST and YAML recently :P

source/client-side-encryption/tests/README.rst
```diff
@@ -0,0 +1,220 @@
...
+
+  - ``mockupcryptdPostCmd``: Optional, a command to send to the mock
+    mongocryptd after the test ends.
+
+  - ``operations``: Array of documents, each describing an operation to be
+    executed. Each document has the following fields:
+
+    - ``name``: The name of the operation on ``object``.
+
+    - ``object``: If present, indicates the target collection.
```
jmikola: This is a deviation from other spec tests where `object` is a string (e.g. "collection", "session0"). I assume that the FLE tests require the ability to use different collection objects for tests within the same YAML/JSON file.

Can we come up with an alternative syntax that won't conflict with future goals for a unified format? `operation.collectionNamespace` alongside `operation.collectionOptions` might do the trick. Then, `object` can remain consistent with other specs and you can likely omit the field entirely in your YAML/JSON files since it should default to "collection".

source/client-side-encryption/tests/README.rst
```diff
@@ -0,0 +1,220 @@
...
+
+  inserts in the ``db.other`` collection.
+
+- The addition of the ``mongocryptd_expectations``, for APM events coming from mongocryptd.
+
+
+Each YAML file has the following keys:
+
+- ``database_name`` and ``collection_name``: The default database and collection
+  to use in operations. This can be overridden by a per operation ``object``.
```
jmikola: I'd propose breaking these into separate bullets for consistency with other spec formats.

source/client-side-encryption/tests/README.rst
```diff
@@ -0,0 +1,220 @@
...
+          "object": { "database": "db", "collection": "other" }
+          "documents": [ { "_id": 1 }]
+      }
+
+  inserts in the ``db.other`` collection.
+
+- The addition of the ``mongocryptd_expectations``, for APM events coming from mongocryptd.
+
+
+Each YAML file has the following keys:
```
jmikola: Is there any need to express server version/topology requirements? If so, can we pull in `runOn` syntax from the CRUD/transaction specs? If not, should the README above specify some criteria for deciding to run these tests (e.g. driver is built with FLE support enabled)?

source/client-side-encryption/tests/README.rst
```diff
@@ -0,0 +1,220 @@
...
+
+  - ``outcome``: Array describing the return value and/or expected state of
+    the collection after the operation is executed. Contains the following
+    fields:
+
+    - ``database``: The database name for the target collection.
+
+    - ``collection``: The name of the target collection.
+
+    - ``data``: Expected data in the target collection.
```
jmikola: I think this syntax also deviates from other spec formats (see: [CRUD](https://github.com/mongodb/specifications/tree/master/source/crud/tests/README.rst) and [Transactions](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst)).

`data` would normally be nested under `collection`, which also supports a `name` property. Will FLE require a custom database name to be specified? If so, perhaps we can add `database_name` under `collection` as well. That would introduce new syntax, but not conflict with any prior art.

source/client-side-encryption/tests/basic.template.yml
```diff
@@ -0,0 +1,42 @@
...
+            documents:
+              - &doc0_encrypted { _id: 1, ssn: {{get_ciphertext("457-55-5462", schema=0, field="ssn")}} }
+            ordered: true
+          command_name: "insert"
+      - command_started_event:
+          command:
+            find: *collection_name
+            filter: { _id: 1 }
+          command_name: find
+      - command_succeeded_event:
```
jyemin: Consider leaving out the succeeded events, for a few reasons

1. None of our other integration spec tests include them
2. If we include them for one of the commands, we have to include them for all.  Otherwise we'll have to specify complex matching rules in terms of what actual events to skip.
3. I think the intent here is to prove that the event is generated before decryption takes place.  If so, we can prove that in a single specialized test that only specifies succeeded events, or perhaps in a prose test.

source/client-side-encryption/tests/basic.template.yml
```diff
@@ -0,0 +1,42 @@
...
+    operations:
+      - name: insertOne
+        arguments:
+          document: &doc0 { _id: 1, ssn: "457-55-5462" }
+      - name: findOne
+        arguments:
+          filter: { _id: 1 }
+        result: *doc0
+    expectations:
+      - command_started_event:
```
jyemin: Because we're using the same MongoClient for listing collections and finding data keys, the events will have to reflect that, or else we'll have to develop a scheme for filtering them out.

Here's a gist with the actual events I'm getting when executing this test: https://gist.github.com/jyemin/eb14c657bac6f5934ec8a43344026957

I believe the extra events are due to `libmongocrypt` currently not doing any caching of schemas or data keys.  If so, these are the events I would actually expect: https://gist.github.com/jyemin/6b70fbdc375e768520726e81790349f7

One possible surprise here: because we're using the same MongoClient to look up the data keys, the driver naively tries to encrypt the find command sent to the key vault collection.  We may want to figure out a way to prevent that, though in practice it's probably not a big deal.

source/client-side-encryption/tests/basic.template.yml
jyemin: @jmikola also interested in what think about this one.

source/client-side-encryption/tests/basic.template.yml
```diff
@@ -0,0 +1,42 @@
...
+          command:
+            insert: *collection_name
+            documents:
+              - &doc0_encrypted { _id: 1, ssn: {{get_ciphertext("457-55-5462", schema=0, field="ssn")}} }
+            ordered: true
+          command_name: "insert"
+      - command_started_event:
+          command:
+            find: *collection_name
+            filter: { _id: 1 }
```
jmikola: I realize you're probably going to change the operation to use `find` instead of `findOne` (per Jeff's earlier comment about the CRUD spec). If not, though, I wonder if it would have made sense to assert `limit` here.

source/client-side-encryption/tests/basic.template.yml
jmikola: It looks like the intention here is to assert that the server responds with encrypted data, while the operation's result assertion above expects unencrypted data. I wonder if we could accomplish this by using outcome assertions on the collection under test, if that was directed to use another MongoClient without FLE enabled.

I second Jeff's concerns about complications with deciding what commandSucceededEvents to skip. As-is, I think we'd need an assertion for the `insert` response to appear before this.

If my suggestion above about using outcome assertions isn't feasible, then I think a prose test would be warranted. That said, I wonder if the outcome assertions _not_ using FLE across the board would also help in other tests to assert that FLE properly encrypts application data before it hits server storage.

source/client-side-encryption/tests/README.rst
kevinAlbs: I think we can get away with just a string value. If we need to do operations on other collections, I could add a "collection0", "collection1". I tentatively changed it.

source/client-side-encryption/tests/README.rst
jmikola: Sorry, I forgot that the CRUD spec wasn't updated yet (pending [SPEC-1255](https://jira.mongodb.org/browse/SPEC-1255)). I'm referring to the syntax in https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst#test-format.

source/client-side-encryption/tests/README.rst
```diff
@@ -0,0 +1,208 @@
...
+
+  - ``kmsPostCmd``: Optional, a command to send to the mock KMS server
+    after the test ends.
+
+  - ``operations``: Array of documents, each describing an operation to be
+    executed. Each document has the following fields:
+
+    - ``name``: The name of the operation on ``object``.
+
+    - ``object``: Optional. If present, indicates the target collection.
```
jmikola: I think this should note specific strings to expect, such as "collection1" and "collection2", as is done in https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst#test-format re: multiple session values. That would also signal to spec runners that they need not expect strings like "database" and "collection" here.

source/client-side-encryption/tests/README.rst
```diff
@@ -0,0 +1,208 @@
...
+         "documents": [ { "_id": 1 } ]
+     }
+
+  inserts on the default collection, but
+
+  .. code:: json
+
+      {
+          "name": "insertOne",
+          "object": { "database": "db", "collection": "other" }
```
jmikola: I think this section is still outdated given the recent changes to `object`.

source/client-side-encryption/tests/basic.template.yml
jyemin: I support the idea of an outcome assertion.  It's in line with other similar test runners, though in this case it is complicated by the need to use a different MongoClient.

jmikola: Based on @kevinAlbs last comment, it doesn't seem like the outcome assertion or commandSucceededEvent are necessary here. If commandStartedEvent for `insert` contains encrypted data, we can trust that is what is being stored server-side. Do we still want a prose test to verify a server response with encrypted data by observing commandSucceededEvent?

source/client-side-encryption/tests/README.rst
kevinAlbs: Fixed.

source/client-side-encryption/tests/basic.template.yml
kevinAlbs: I'm adding back "outcome". It's actually a bit easier to verify that data is stored as ciphertext now that auto encryption is set on the collection level. Operations that require auto encryption pass that as a collection option. The "outcome" data will be run on a collection without auto encryption/decryption, so it should see the ciphertext.

kevinAlbs:
This is now just an extension of the transaction tests (only adds fields):
- `json_schema` to create a server side JSON schema
- `key_vault` To create a KeyVault object for every test.
- `tests.operations.collectionOptions.auto_encrypt` to enable auto encryption on a collection per operation

There may be more additions to come, but I removed tentative things like kmsFailpoint. We don't even have a mock KMS server yet, so let's not assume we'll need it.

@jyemin if this new format looks good to you, I'll proceed to write tests for operations supported by mongocryptd ("insert", "find", "delete", "distinct").

kevinAlbs:
Tweaked for the latest API changes: KeyVault is no longer passed to `db.getCollection().`

source/client-side-encryption/tests/basic.yml
```diff
@@ -0,0 +1,53 @@
...
+      - name: insertOne
+        collectionOptions: { auto_encrypt: { enabled: True } }
+        arguments:
+          document: &doc0 { _id: 1, ssn: "457-55-5462" }
+      - name: find
+        collectionOptions: { auto_encrypt: { enabled: True } }
+        arguments:
+          filter: { _id: 1 }
+        result: [*doc0]
+    expectations:
```
jyemin: Add the key vault query as well. This may require that we specify the database_name in the expectations, unless we put the key vault in the same database as all the other collections.

kevinAlbs: Done in new version.

kevinAlbs:
Moving this out of WIP. Now includes:

- YAML tests for operations that mongocryptd supports that pass using Java PoC
- YAML tests based on mongocryptd limitations in the "Query feature support" doc.
- Prose tests

Some tests skip, with the skipReason indicating the blocking ticket.

source/client-side-encryption/tests/aggregate.template.yml
```diff
@@ -0,0 +1,61 @@
+runOn:
```
jyemin: Add a test for db/client level aggregate.  It should fail, iirc

source/client-side-encryption/tests/aggregate.template.yml
```diff
@@ -0,0 +1,61 @@
...
+collection_name: &collection_name "default"
+
+data:
+  - &doc0_encrypted { _id: 1, ssn: {{ciphertext("457-55-5462", field="ssn")}} }
+json_schema: {{schema()}}
+key_vault_data: [{{key()}}]
+
+tests:
+  - description: "Aggregate with deterministic encryption"
+    skipReason: "SERVER-39395"
```
jyemin: Are there no aggregate tests that can pass yet?

source/client-side-encryption/tests/bulk.template.yml
```diff
@@ -0,0 +1,63 @@
...
+  - description: "Bulk write with encryption"
+    clientOptions:
+      auto_encrypt_opts:
+        kms_providers:
+          aws: {} # Credentials filled in from environment.
+    operations:
+      - name: bulkWrite
+        arguments:
+          requests:
+            - name: insertOne
```
jyemin: Consider testing the other request types as well.

source/client-side-encryption/tests/delete.template.yml
```diff
@@ -0,0 +1,58 @@
+runOn:
```
jyemin: Test both deleteOne and deleteMany (as you already do with update)

source/client-side-encryption/tests/find.template.yml
```diff
@@ -0,0 +1,121 @@
+runOn:
```
jyemin: There needs to be a test of getMore.  You can do it easily with a find of four encrypted documents and a batchSize of 2.

source/client-side-encryption/tests/insert.template.yml
```diff
@@ -0,0 +1,55 @@
+runOn:
```
jyemin: Test both insertOne and insertMany

source/client-side-encryption/tests/aggregate.template.yml
jyemin: Even an empty pipeline would buy us some code coverage.

source/client-side-encryption/tests/find.template.yml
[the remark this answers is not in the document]
kevinAlbs: Added a test.

source/client-side-encryption/tests/README.rst
```diff
@@ -0,0 +1,251 @@
...
+
+    - ``result``: |txn|
+
+  - ``expectations``: |txn|
+
+  - ``outcome``: |txn|
+
+
+
+Use as integration tests
```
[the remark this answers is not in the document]
kevinAlbs: Added one for mapReduce with a `skipReason` since Java PoC currently does not error on unsupported commands.

source/client-side-encryption/tests/README.rst
```diff
@@ -0,0 +1,251 @@
...
+   #. Drop the ``admin.datakeys`` collection using writeConcern "majority".
+   #. Insert the data specified into the ``admin.datakeys`` with write concern "majority".
+#. Create a MongoClient using ``clientOptions``.
+   #. If ``client_side_encryption_opts`` includes ``aws`` as a KMS provider, pass in AWS credentials from the environment.
+#. Create a collection object from the MongoClient, using the ``database_name``
+   and ``collection_name`` fields from the YAML file.
+#. Drop the test collection, using writeConcern "majority".
+#. If the YAML file contains a ``data`` array, insert the documents in ``data``
+   into the test collection, using writeConcern "majority".
+#. If ``failPoint`` is specified, its value is a configureFailPoint command.
```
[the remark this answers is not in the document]
kevinAlbs: Good point, done.

source/client-side-encryption/tests/README.rst
```diff
@@ -0,0 +1,251 @@
...
+     using the ``hasErrorLabel`` method.
+   - If the operation returns a raw command response, eg from ``runCommand``,
+     then compare only the fields present in the expected result document.
+     Otherwise, compare the method's return value to ``result`` using the same
+     logic as the CRUD Spec Tests runner.
+
+#. If the test includes a list of command-started events in ``expectations``,
+   compare them to the actual command-started events using the
+   same logic as the Command Monitoring Spec Tests runner.
+#. If ``failPoint`` is specified, disable the fail point to avoid spurious
```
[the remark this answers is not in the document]
kevinAlbs: Done.

source/client-side-encryption/tests/aggregate.template.yml
kevinAlbs: Added a failing db aggregate test.

source/client-side-encryption/tests/aggregate.template.yml
kevinAlbs: Added a passing empty pipeline test. But until SERVER-39395 is in, I don't think there's much else I can test.

source/client-side-encryption/tests/bulk.template.yml
kevinAlbs: Added updateOne and deleteOne

source/client-side-encryption/tests/delete.template.yml
kevinAlbs: Done.

source/client-side-encryption/tests/find.template.yml
kevinAlbs: Added a passing getMore.template.yml test.

source/client-side-encryption/tests/insert.template.yml
kevinAlbs: Done.

source/transactions/tests/explain.template.yml
```diff
@@ -0,0 +1,63 @@
+runOn:
```
jyemin: I think this was added to transactions folder by mistake

source/client-side-encryption/tests/aggregate.template.yml
jyemin: What's the name of that test? I'm not finding it

source/transactions/tests/count.template.yml
```diff
@@ -0,0 +1,44 @@
+runOn:
```
jyemin: Added in error?

source/transactions/tests/explain.template.yml
kevinAlbs: Doh, fixed.

source/transactions/tests/count.template.yml
kevinAlbs: Yup, removed.

source/client-side-encryption/tests/aggregate.template.yml
kevinAlbs: Last test in this file, "Database aggregate should fail"

jyemin: Ack

source/client-side-encryption/tests/README.rst
```diff
@@ -0,0 +1,257 @@
...
+
+- A ``json_schema`` to set on the collection used for operations.
+
+- A ``key_vault_data`` of data that should be inserted in the key vault collection before each test.
+
+- Introduction ``client_side_encryption_opts`` to `clientOptions`
+
+- Addition of `$db` to command in `command_started_event`
+
+- Addition of `$type` to command_started_event and outcome.
```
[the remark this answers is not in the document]
jyemin: The semantics of `$$type` is that any actual value matching the BSON type indicated by the BSON type string is considered a match.  So 

                            "updates": [
                                {
                                    "u": {
                                        "$set": {
                                           "random": {
                                                    "$$type": "binData"
                                            }
                                     }
                                }
                            ]

means that the value of the `updates.u.$set.random` field in the actual command must be of type binData.

@kevinAlbs minor nit:  Not sure what the string "binData" comes from.  I don't see it in bsonspec.org or in extended JSON spec.  Maybe change to "binary"?

source/client-side-encryption/tests/getMore.template.yml
```diff
@@ -0,0 +1,70 @@
...
+            find: datakeys
+            filter: { _id: { $in: [ {{key()["_id"]}} ] } }
+            $db: admin
+          command_name: find
+      - command_started_event:
+          command:
+            getMore: { $$type: "long" }
+            collection: *collection_name
+            batchSize: 2
+            $readPreference: { mode: "primaryPreferred" }
```
jyemin: Why did you include $readPreference here?  I suspect it will fail when running on anything but a standalone.  I suggest to remove it.

source/client-side-encryption/tests/README.rst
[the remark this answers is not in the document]
kevinAlbs: Nice, stealing your description.

The value "binData" comes from: https://docs.mongodb.com/manual/reference/bson-types/

Which is also supported by the $type query operator: https://docs.mongodb.com/manual/reference/operator/query/type/#op._S_type

source/client-side-encryption/tests/getMore.template.yml
kevinAlbs: Removed.

kevinAlbs:
New commits, added a deduplication prose test. Test data ciphertext will be updated once crypto updates are in libmongocrypt.

source/client-side-encryption/tests/mapReduce.template.yml
```diff
@@ -0,0 +1,26 @@
...
+
+data:
+  - &doc0_encrypted { _id: 1, x: 1, ssn: {{ciphertext("457-55-5462", field="ssn")}} }
+  - &doc1_encrypted { _id: 2, x: 2, ssn: {{ciphertext("123-45-6789", field="ssn")}} }
+json_schema: {{schema()}}
+key_vault_data: [{{key()}}]
+
+tests:
+  - description: "mapReduce deterministic encryption"
+    skipReason: "TODO: throw on unsupported commands in Java PoC"
```
jyemin: Let's remove this and i'll fix the Java PoC

source/client-side-encryption/tests/aggregate.template.yml
```diff
@@ -0,0 +1,135 @@
...
+      - name: aggregate
+        object: database
+        arguments:
+          pipeline:
+            - $currentOp: { allUsers: false, idleConnections: false, localOps: true }
+            - $match: { command.aggregate: { $eq: 1 } }
+            - $project: { command: 1 }
+            - $project: { command.lsid: 0 }
+        result:
+          errorContains: "Expected collection"
```
jyemin: This should be:

     "automatic encryption is not supported"

according to spec.

source/client-side-encryption/tests/mapReduce.template.yml
kevinAlbs: Done.

source/client-side-encryption/tests/aggregate.template.yml
kevinAlbs: Done.

kevinAlbs:
New commit, but those two tests now fail with Java PoC.

source/client-side-encryption/tests/distinct.template.yml
```diff
@@ -0,0 +1,87 @@
...
+            query: { ssn: {$eq: {{ciphertext("457-55-5462", field="ssn")}} } }
+          command_name: distinct
+    outcome:
+      collection:
+        # Outcome is checked using a separate MongoClient without auto encryption.
+        data:
+          - *doc0_encrypted
+          - *doc1_encrypted
+          - *doc2_encrypted
+  - description: "Distinct fails when filtering on a random encrypted field"
```
jyemin: Change to "a deterministic encrypted field".

kevinAlbs: This is actually a duplicate test. Removed.

kevinAlbs:
New changes:
- Updated ciphertexts per the crypto changes (no IV, longer key).
- Swap order for distinct fields for consistency with other tests, Node test runner hit an issue with this.
- Fix outcome assertions.
- Added a test for all BSON types (types.template.yml)
- Added a prose test for failing for < max wire version 8 per CDRIVER-3153 (I thought about adding a spec test, but this should probably fail on MongoClient construction with encryption options, not the first operation).
- Updated README to say tests should be run with and without auth
- Audited and validated existing skipped tests. keyAltName test has been unskipped, as has findAndModify.
- Added a skipped spec test for bypassAutoEncryption per CDRIVER-3139
- Tested a bypassed command (ping) and renamed mapReduce test to unsupportedCommand.template.yml)
- Standardized the values (no more arbitrary ssns, just example strings like "string0", "string1")

I verified all tests pass with latest mongocryptd + [libmongocrypt master](https://github.com/10gen/libmongocrypt/commit/48aa7d4e7b4f9d0679563c1e63adb3629c601968) on the Java PoC.

source/client-side-encryption/tests/unsupportedCommand.json
```diff
@@ -120,7 +120,7 @@
     ],
     "tests": [
         {
-            "description": "mapReduce deterministic encryption",
+            "description": "mapReduce deterministic encryptio (unsupported)",
```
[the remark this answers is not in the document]
jyemin: Just a nit.  You can resolve this comment when you fix it.  I don't need to re-review.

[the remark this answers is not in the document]
kevinAlbs: Fixed.

source/client-side-encryption/tests/README.rst
```diff
@@ -0,0 +1,296 @@
...
+   - Expect that document to have "masterKey.provider" set to "aws"
+
+#. Test explicit encrypt and decrypt with the "local" KMS provider.
+
+   - Create a `KeyVault` with a "local" KMS provider.
+   - Insert the a key document in to the key vault.
+   - Use `KeyVault.encrypt` to encrypt the value "hello" with the following:
+
+     - the algorithm "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic"
+     - a fixed 16 byte initialization vector
```
jyemin: Remove this line

source/client-side-encryption/tests/README.rst
```diff
@@ -0,0 +1,296 @@
...
+   - Expect the value is equal to the string "hello"
+
+#. Test explicit encrypt and decrypt with the "aws" KMS provider.
+
+   - Create a `KeyVault` with a "aws" KMS provider.
+   - Insert the a key document in to the key vault.
+   - Use `KeyVault.encrypt` to encrypt the value "hello" with the following:
+
+     - the algorithm "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic"
+     - a fixed 16 byte initialization vector
```
jyemin: Remove this line

source/client-side-encryption/tests/README.rst
```diff
@@ -0,0 +1,296 @@
...
+     - Expect a `string` to succeed. An exception MUST be thrown.
+     - Expect a `document` to fail. An exception MUST be thrown.
+     - Expect a `null` to fail. An exception MUST be thrown.
+     - Expect a BSON binary subtype 6 to fail. An exception MUST be thrown.
+
+   - Use `KeyVault.encrypt` to attempt to encrypt a document using randomized encryption.
+
+     - Expect a `document` to succeed.
+     - Expect a BSON binary subtype 6 to fail. An exception MUST be thrown.
+     - Expect a `null` to fail. An exception MUST be thrown.
```
jyemin: This concerns me.  Since null is treated almost the same as the absence of the key, I could see situations where applications inadvertently include a null value for an encrypted field.  Perhaps drivers should just silently not encrypt null.  Consider a case like this:

    MongoDB Enterprise > db.test.insert({_id : 1})
    WriteResult({ "nInserted" : 1 })
    MongoDB Enterprise > db.test.find({y : null})
    { "_id" : 1 }

This is a natural way that applications search for absence of a field, but if `y` is an encrypted field, this query will fail because the driver will throw trying to encrypt the query filter.

Feel free to resolve this if you'd prefer to deal with it at the spec level

source/client-side-encryption/tests/README.rst
```diff
@@ -0,0 +1,296 @@
...
+
+#. Test explicit encrypting an auto encrypted field.
+
+   - Create a `KeyVault` with either a "local" or "aws" KMS provider
+   - Create a collection with a JSONSchema specifying an encrypted field.
+   - Use `KeyVault.encrypt` to encrypt a value.
+   - Create a document, setting the auto-encrypted field to the value.
+   - Insert the document. Verify an exception is thrown.
+
+#. Test explicit encrypting an auto encrypted field.
```
jyemin: This test is identical to the one above it.

source/client-side-encryption/tests/README.rst
kevinAlbs: Done.

source/client-side-encryption/tests/README.rst
kevinAlbs: Done.

source/client-side-encryption/tests/README.rst
kevinAlbs: Oh, interesting. I'm not sure what the best decision is, but this is probably either a spec question, and perhaps one the query team should weigh in on. This would require a change in mongocryptd to ignore `null` on encrypted fields instead of erroring.

Regardless of that decision, perhaps explicit encryption should still error, because if the user is attempting to explicitly encrypt `null`, they could just pass `null` directly. What do you think?

source/client-side-encryption/tests/README.rst
kevinAlbs: Removed.

source/client-side-encryption/tests/README.rst
jyemin: Reflecting more about this, perhaps it's not worth worrying about.  Applications can use `$exists` in filters, and $ifNull in aggregation expressions, and those will do the right thing without requiring the use of null

source/client-side-encryption/tests/maxWireVersion.template.yml
```diff
@@ -0,0 +1,20 @@
+runOn:
+  - maxServerVersion: "4.0.0"
```
jyemin: This should be just `4.0`.  Otherwise I can see implementations interpreting that it shouldn't run on a `4.0.10` server, for example.  Alternatively it could be `4.0.<maxInteger>`, but that just seems weird.

kevinAlbs: Done.

kevinAlbs:
@jyemin I added (failing) spec and prose tests for mongocryptd options. See spawningMongocryptd.json

source/client-side-encryption/tests/spawningMongocryptd.template.yml
```diff
@@ -0,0 +1,95 @@
+runOn:
```
jyemin: Unfortunately I think all these tests are suspect.  

1. The first one (Bypassing spawning mongocryptd fails to connect) fails when run non-in-isolation, because the previous test has already spawned mongocryptd and so this test is able to connect to the previously spawned one.
2. The second one (Spawning mongocryptd on a different port fails to connect) succeed for the wrong reason: all but the first spawned mongocrypd will fail to spawn (duplicate pid)
3. The third one (Insert on mongocryptd spawned with different URI) fails because the mongocryptd fails for the same reason as the second succeeds: the mongocryptd on port 27030 doesn't actually spawn, and so we fail to connect on that port.

source/client-side-encryption/tests/README.rst
```diff
@@ -0,0 +1,289 @@
...
+
+   - Create a `KeyVault` with either a "local" or "aws" KMS provider
+   - Create a collection with a JSON Schema specifying an encrypted field.
+   - Use `KeyVault.encrypt` to encrypt a value.
+   - Create a document, setting the auto-encrypted field to the value.
+   - Insert the document. Verify an exception is thrown.
+
+#. Test bypassing automatic spawning of mongocryptd.
```
jyemin: Does this duplicate the first YAML/JSON test in spawnMongocrypt?

source/client-side-encryption/tests/README.rst
```diff
@@ -0,0 +1,289 @@
...
+
+   - Create a collection with a JSON Schema specifying an encrypted field.
+   - Spawn a mongocryptd process with: `mongocryptd --port 27030`
+   - Create a `MongoClient` configured with `clientSideEncryptionOpts` of the following:
+      - `kmsProviders = { "aws": { "access_key_id": <AWS access key id> "secret_access_key": <AWS secret access key>`
+      - `extraOptions = { "mongocryptdBypassSpawn": true, "mongocryptdURI": "mongodb://localhost:27030" }`
+   - Insert a document into the collection with a value on the encrypted field. Expect this to succeed.
+   - Find the document. Verify the value on the encrypted field matches the originally set value.
+
+#. Test spawning mongocryptd from a different path.
```
jyemin: This is going to be a headache to automate.

source/client-side-encryption/tests/types.yml
```diff
@@ -0,0 +1,622 @@
...
+          command:
+            find: *collection_name
+            filter: {_id: 1}
+          command_name: find
+    outcome:
+      collection:
+        # Outcome is checked using a separate MongoClient without auto encryption.
+        data:
+          - *doc11_encrypted
+  - description: "type=dbPointer"
```
mbroadst: dbPointer is deprecated, can we remove this test or somehow make it optional to run?

kevinAlbs: I don't know of any existing way to make this optional. This (and the other types tests) should really be tested by tests of mongocryptd. I still like having having them, but it seems safe to remove dbPointer.

source/client-side-encryption/tests/README.rst
kevinAlbs: It tests the opposite. Since the YAML/JSON test has no way to specify "spawn this process outside of the test", the YAML/JSON test verifies that if you bypass the spawn, the client fails to connect. This tests that starting the mongocryptd process beforehand, and bypassing spawn, succeeds.

kevinAlbs:
- Fixed the document validation errors.
- Removed db pointer + numberlong test from types.json.

I have not updated spawningMongocryptd.json yet. I added to our weekly meeting agenda to consider removing this requirement. If we do need it, I need to rethink those tests.

kevinAlbs:
Removed spawning tests and commented in WRITING-3793 to do this as follow-up work if still applicable.

source/client-side-encryption/tests/keyAltName.yml
```diff
@@ -0,0 +1,72 @@
...
+  - description: "Replace with key alt name fails"
+    clientOptions:
+      autoEncryptOpts:
+        kmsProviders:
+          aws: {} # Credentials filled in from environment.
+    operations:
+      - name: updateOne
+        arguments:
+          filter: {}
+          update: { encrypted_w_altname: "string0" }
```
mbroadst: `updateOne` requires at least one atomic operation. This should be:
```
update: { $set: { encrypted_w_altname: "string0" } }
```

kevinAlbs: oh right, fixed!

source/client-side-encryption/tests/types.template.yml
```diff
@@ -0,0 +1,527 @@
...
+    clientOptions:
+      autoEncryptOpts:
+        schemaMap:
+          "default.default": {{schema_w_type("double")}}
+        kmsProviders:
+          aws: {} # Credentials filled in from environment.
+    operations:
+      - name: insertOne
+        arguments:
+          document: &doc4 { _id: 1, encrypted_long: {"$double": "1.23"} }
```
jyemin: This should be just:

    encrypted_long: 1.23

or if it has to be canonical:

    encrypted_long: {"$numberDouble": "1.23"}

jyemin: CC @mbroadst

kevinAlbs: Fixed.

kevinAlbs:
- Fixed errors per new mongocryptd (db version v4.3.0-7-g5e998f0)
- Added a test for a local schema without encryption to localSchema.template.yml
- Added back the test with an invalid update to updateOne.template.yml. Even though it's a little unrelated, it could prevent another implementation bug.

source/client-side-encryption/tests/localSchema.json
```diff
@@ -0,0 +1,276 @@
...
+                                "test"
+                            ]
+                        }
+                    },
+                    "kmsProviders": {
+                        "aws": {}
+                    }
+                }
+            },
+            "operations": [
```
jyemin: This definitely doesn't pass on j3071 branch. It should be just:

           "operations": [
                {
                    "name": "insertOne",
                    "arguments": {
                        "document": {
                            "_id": 1,
                            "encrypted_string": "string0"
                        }
                    },
                    "result": {
                        "errorContains": "JSON schema keyword 'required' is only allowed with a remote schema"
                    }
                }

kevinAlbs: gah I updated the template file but forgot to regenerate the .json/.yml files. Fixed.

kevinAlbs:
stand back everyone

PR #453, opened 2019-01-28
SPEC-1210: Document multi-operation CRUD test format with APM

jmikola:
https://jira.mongodb.org/browse/SPEC-1210


source/crud/tests/README.rst
```diff
@@ -54,24 +42,19 @@ Each YAML file has the following keys:
 
   - ``description``: The name of the test.
 
-  - ``operation``: Document describing the operation to be executed. This will
-    have the following fields:
+  - ``operations``: Array of documents, each describing an operation to be
+    executed. Each document has the following fields:
 
-    - ``name``: The name of the operation as defined in the specification. The
-      name `db-aggregate` refers to database-level aggregation.
```
jmikola: @kvwalker: Tagged you on review for this bit. I think this was a leftover and can be safely removed.

source/crud/tests/README.rst
```diff
@@ -90,10 +78,41 @@ Each YAML file has the following keys:
...
 
+Optional Fields in Results
+--------------------------
+
 Expected results for some tests may include optional fields, such as
 ``insertedId`` (for InsertOneResult), ``insertedIds`` (for InsertManyResult),
 and ``upsertedCount`` (for UpdateResult). Drivers that do not implement these
 fields can ignore them.
+
+Asserting Nonexistent Fields
```
jmikola: This is something I've had in the back of my mind for a while. On a related note, I think we can use `{ $exists: true }` in place of magic number `42` in other spec tests (e.g. checking for the presence of values we don't particular care about, like cursor IDs).

Thoughts? If this is controversial, I can revert to the [Null Values](https://github.com/mongodb/specifications/tree/master/source/transactions/tests#null-values) language from the Transactions spec.

source/crud/tests/README.rst
kvwalker: Yep, this can be removed.

source/crud/tests/README.rst
prashantmital: I do prefer this to `null`/`42` (even though I am a big H2G2 fan) as it makes the test writer's intentions explicit.

prashantmital: I do prefer this to `null`/`42` (even though I am a big H2G2 fan) as it makes the test writer's intentions explicit.

source/crud/tests/README.rst
```diff
@@ -82,6 +65,11 @@ Each YAML file has the following keys:
       result object if their BulkWriteException (or equivalent) provides access
       to a write result object.
 
+  - ``expectations`` (optional): List of command-started events.
```
prashantmital: This suggests that there is a standard way to describe `command-started events`. Is the expected format of these events that same as that mentioned [in the transactions spec](https://github.com/mongodb/specifications/tree/master/source/transactions/tests#command-started-events)? If so, we should probably make 'command-started events' a link to the same.

source/crud/tests/README.rst
```diff
@@ -12,29 +12,17 @@ Introduction
...
-tied to a MongoDB server version.
-
-Format
-======
+Test Format
+===========
 
 Each YAML file has the following keys:
 
+- ``collection_name`` (optional): The collection to use for testing.
```
prashantmital: I will need to add `collectionOptions` to this in order to write tests for https://jira.mongodb.org/browse/SPEC-1153. Would you be in favor of adding optional fields for `collectionOptions` and `databaseOptions`  (both would be arrays of documents)? This would be added under `tests` of course.

ShaneHarvey: The transaction tests have collectionOptions placed on the same level as 'object' like this:
```
  - description: collection writeConcern without transaction
    operations:
      - name: insertOne
        object: collection
        collectionOptions:
          writeConcern:
            w: majority
        arguments:
```
https://github.com/mongodb/specifications/tree/master/source/transactions/tests#test-format

source/crud/tests/README.rst
jmikola: The original format for these expectations comes from the Command Monitoring spec test. Unfortunately, that spec's [README.md](https://github.com/mongodb/specifications/blob/master/source/command-monitoring/tests/README.rst) also doesn't have a real format defined (like other spec test READMEs) and assumes that the document fields in the YAML/JSON file correspond to the event models from the spec itself.

I'm not sure it makes sense to link to the Transaction spec's README here, as the only thing it says for this field is "The event listener used for these tests MUST ignore the security commands listed in the Command Monitoring Spec." I don't think that's particularly helpful, as the [Security](https://github.com/mongodb/specifications/blob/master/source/command-monitoring/command-monitoring.rst#security) section in the command monitoring spec already discusses such commands being filtered out. AFAIK, libmongoc doesn't even raise events for these commands (vs. using an empty document) -- I'm not sure if other drivers do the same or actually report events w/ empty documents.

prashantmital: Yes, I think that will suffice.

source/crud/tests/README.rst
jmikola: The Transaction spec defines a `clientOptions` key under the `tests` field, which I omitted here because the CRUD specs had no use for it. I think it'd make sense to add it for completeness' sake, though.

As for `collectionOptions`, Transactions already defines that as a field within the operation document (e.g. `tests.operations[i]`). Would that work for you?

The only other relevant spec I can think of with its own deviating test format is Change Streams, and that does not define `clientOptions` or `collectionOptions` anywhere (yet), so we don't have to worry about potentially introducing a new conflict.

vincentkam:
As per offline request to review the CRUD test format changes, LGTM!

source/crud/tests/README.rst
prashantmital: I missed `clientOptions`, but I think we should definitely add it here as well.

Having a `collectionOptions` field at the same level as `object` would work perfectly. It also makes sense to put it there so that it can be changed for different test cases.

source/crud/tests/README.rst
```diff
@@ -90,10 +78,41 @@ Each YAML file has the following keys:
...
+- Instead of a `tests[i].operations` array, a single operation was defined as a
+  document in `tests[i].operation`. That document consisted of only the `name`,
+  `arguments`, and an optional `object` field.
+
+- Instead of `error` and `result` fields within each element in the
+  `tests[i].operations` array, the single operation's error and result were
+  defined under the `tests[i].outcome.error` and `tests[i].outcome.result`
+  fields.
+
+The legacy format should not conflict with the newer, multi-operation format
```
prashantmital: I don't understand what is meant by this. @jmikola is this trying to say that test runners _should_ support both the updated and legacy formats? Do you intend to update all CRUD spec-tests to the new format or leave them as-is? IIUC whatever format we leave the CRUD tests in after this PR is the one drivers will _have_ to support (after they sync tests the next time).

jmikola: This is permitting drivers to support both. @xdg brought up a concern with this PR requiring more changes than necessary for drivers, which might distract from necessary 4.2 work. I do not intend to update all existing CRUD tests to this format.

My priority is to get the better, unified format documented so that you can use it for [SPEC-1153](https://jira.mongodb.org/browse/SPEC-1153), which requires APM. @kvwalker also needed `object:` syntax documented for [SPEC-1184](https://jira.mongodb.org/browse/SPEC-1184). At this point, I don't think @vincentkam really needed anything from this PR, as his retryable reads tests are consistent with the Transactions spec's format; however, the format in this PR should now be consistent with that as well.

source/crud/tests/README.rst
jmikola: Added both to the format. Feel free to resolve.

source/crud/tests/README.rst
jmikola: Updated the format documentation to refer to Command Monitoring spec. Feel free to resolve.

source/crud/tests/README.rst
jmikola: Added sections to explain how `{ $exists: true }` and `{ $exists: false }` work. That said, these still aren't necessary for any existing CRUD tests. My goal here is to get this accepted so we can start using it going forward (and ultimately replace the less explicit behavior in other spec runners).

jmikola:
Note: after the README changes are approved, I will plan to revert changes made to the `bulkWrite-arrayFilters` test (https://github.com/mongodb/specifications/pull/453/commits/d98379813d8f855f731cd5453965d890b6589d1f); however, as @ShaneHarvey mentioned in https://github.com/mongodb/specifications/pull/453#pullrequestreview-197820100, the modifications in that file may be helpful to understand what the new format will look like in practice.

source/crud/tests/README.rst
daprahamian: I am a big fan of this idea in general, but since [`$exists` is a valid query operator](https://docs.mongodb.com/manual/reference/operator/query/exists/), could we use something that would never be used in MQL?

source/crud/tests/README.rst
```diff
@@ -90,10 +100,155 @@ Each YAML file has the following keys:
...
+elements in ``outcome.data`` are written in
+`Extended JSON <../../extended-json.rst>`_. Drivers may adopt any of the
+following approaches to comparisons, as long as they are consistent:
+
+- Convert ``actual`` to Extended JSON and compare to ``expected``
+- Convert ``expected`` and ``actual`` to BSON, and compare them
+- Convert ``expected`` and ``actual`` to native representations, and compare
+  them
+
+Extra Fields in Actual Results
```
daprahamian: Does this apply for arrays as well? For example, would:

- ``expected`` is `[1, 4, 5]`
- ``actual`` is `[1, 4, 5, 6]`

be considered a match?

jmikola: No sir. I'll clarify that this is only for extra fields in documents (e.g. `command_started_event.command`, `outcome.data[i]`, and documents within result objects).

jmikola: See https://github.com/mongodb/specifications/pull/453/commits/517d8510340fc225b4e61a790e098869aef528a8

source/crud/tests/README.rst
prashantmital: Just out of curiosity, why is it problematic to have a construct that _may_ also be a part of MQL in an object containing *expected* key, value pairs? AFAICT, the expectations will only ever be parsed by the language implementing the spec and never by the server...

daprahamian: @prashantmital I am worried about introspection of outgoing queries. For example, with the following APM event:

```
{
  "command": {
    "find": "bar",
    "filter": {
      "elbow": {
        "$exists": false
      }
    }
    ...
  }
  ...
}
```

How would we write the `expected` version of that event?

jmikola: Ah, very good point. I really did not expect `$exists` to be used anywhere other than the top-level of a document -- although I didn't explicitly mention that in the README. Re-reading the [match documentation](https://github.com/mongodb/specifications/tree/master/source/change-streams/tests#spec-test-match-function) from the Change Stream spec, it seems like you intended assertions of `42` to occur at any level of the document, but I wonder if that's actually ever needed in practice. In the Transactions spec, I believe assertions for nonexistent fields via `null` is only done at the top-level (e.g. absence of a `writeConcern` field).

If we apply that restriction, I don't think we'd have a risk of conflict with actual query operators. A real `$exists` operator might appear under a `filter` key for a `find` command (or similar, such as `query` for `findAndModify`) or deeper withing an aggregation pipeline.

If you do think it's important to allow this comparison at any level, we can create a more contrived name. In my experience with PHP, at some point in our test runner I just convert the document to extended JSON and do a straight up comparison rather than crawl the document recursively (as suggested in the Change Stream spec test algorithm). I'm not sure what other drivers are doing.

source/crud/tests/README.rst
```diff
@@ -54,34 +42,56 @@ Each YAML file has the following keys:
...
+    - ``error`` (optional): If ``true``, the test should expect the operation
+      to emit an error or exception. If ``false`` or omitted, drivers MUST
+      assert that no error occurred.
 
-    - ``result``: The return value from the operation. This will correspond to
-      an operation's result object as defined in the CRUD specification. This
-      field may be omitted if ``error`` is ``true``. If this field is present
-      and ``error`` is ``true`` (generally for multi-statement tests), the
-      result reports information about operations that succeeded before an
+    - ``result`` (optional): The result of executing the operation. This will
```
prashantmital: @jmikola since `operations` is an array potentially containing multiple operations, we might need to clarify which 'result' the `result` key corresponds to? AFAICT we have 2 options:

1. Make this a list (say, `results`) with `N` entries where `N` is the number of elements in the `operations` list. Then the `n`-th entry in `results` is the expected result of the `n`-th operation in `operations`.
2. Restrict `result` to always refer to the return value of the *last* operation in the `operations` list.

ShaneHarvey: Each item in the "operations" list is its own "operation" which contains its own "result" field. For example: https://github.com/mongodb/specifications/blob/master/source/transactions/tests/delete.yml#L17-L32

jmikola: @prashantmital: I don't think there's a problem with the format as written. Does Shane's response clear this up?

The new format places `error` and `result` fields under each document element in the `operations` array. Previously in the legacy format, `error` and `result` appeared under `outcome` (parallel to the single `operation` field).

source/crud/tests/write/bulkWrite-arrayFilters.yml
```diff
@@ -4,42 +4,65 @@ data:
...
+              -
+                q: {}
+                u: { $set: { "y.$[i].b": 2 } }
+                multi: false
+                upsert: false
+                arrayFilters: [ { "i.b": 3 } ]
+              -
+                q: {}
+                u: { $set: { "y.$[i].b": 2 } }
+                multi: false
```
prashantmital: Not that it matters since you'll be reverting this commit, but I think since this comes from a `updateMany`, it should be `multi: true`.

prashantmital: Not that it matters since you'll be reverting this commit, but I think since this comes from a `updateMany`, it should be `multi: true`.

source/crud/tests/write/bulkWrite-arrayFilters.yml
```diff
@@ -4,42 +4,65 @@ data:
...
+                upsert: false
+                arrayFilters: [ { "i.b": 1 } ]
+            ordered: true
+            writeConcern: { $exists: false }
+            bypassDocumentValidation: { $exists: false }
+            command_name: update
```
prashantmital: Can you clarify how this and the next `database_name` key are supposed to be used?

source/crud/tests/write/bulkWrite-arrayFilters.yml
```diff
@@ -4,42 +4,65 @@ data:
...
+            database_name: *database_name
```
prashantmital: Can you clarify how `database_name` and `command_name` are intended to be used? I wasn't quite able to figure it out.

source/crud/tests/README.rst
prashantmital: Yes, my bad. Thanks @ShaneHarvey !

source/crud/tests/write/bulkWrite-arrayFilters.yml
jmikola: These were indented too far and should be children of `command_started_event`. They are properties on the [CommandStartedEvent](https://github.com/mongodb/specifications/blob/master/source/command-monitoring/command-monitoring.rst#api) object. Will fix.

source/crud/tests/write/bulkWrite-arrayFilters.yml
jmikola: Good catch! I'm also going to remove `upsert` and `multi` fields with a `false` value (see: https://github.com/mongodb/specifications/pull/467).

source/crud/tests/README.rst
ShaneHarvey: How about `{"$assertExists": true}`?

jmikola: Created [SPEC-1215](https://jira.mongodb.org/browse/SPEC-1215) to track this idea. I'll remove both of these sections from the PR, as the original text is now quoted in that ticket.

source/crud/tests/README.rst
jmikola: @prashantmital: I revised this paragraph a bit. Feel free to resolve if your original questions are answered.

jmikola:
Tests are now segregated by `v1` and `v2` directories to avoid having tests with different formats in the same directory structure.

@kvwalker: I've modified your original `db-aggregate` test to use `operations` if you want to take another look. The test file has been moved into the `v2` folder. @rozza: Since you also modified those tests (adding `minServerVersion`), you might want to take a look as well.

PR #454, opened 2019-01-29
SPEC-1161 do not inherit read concern in transaction options test

p-mongo:
cf7248b2574b1a5397a6e18362acfe66f2b55a43 prohibited inheriting read
concern from client when starting a transaction, however this test
attempts to inherit the read concern in this way.

### DIFF
--- a/source/transactions/tests/transaction-options.json
+++ b/source/transactions/tests/transaction-options.json
@@ -210,9 +210,6 @@
               },
               "startTransaction": true,
               "autocommit": false,
-              "readConcern": {
-                "level": "local"
-              },
               "writeConcern": null
             },
             "command_name": "insert",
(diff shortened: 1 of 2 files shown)


jmikola:
@p-mongo: I believe one of the tests for the `withTransaction()` spec violates SPEC-1161. Would you mind updating that as part of this commit? In `transaction-options.yml`:

 * "withTransaction inherits transaction options from client" attempts to inherit `readConcernLevel: local` from the client. I'd suggest adding a suffix of "(excluding read concern)" to the test. Addtionally, `readConcern: { level: local }` can be deleted from first expected command and replaced with `readConcern: ~` between `# omitted fields` and `writeConcern: ~`.
 * "withTransaction explicit transaction options override client options" defines `readConcernLevel: majority` on the client, but it is never inherited because we set a read concern with explicit transaction options. I don't think that this test needs to change at all, unless you want to remove `readConcernLevel` altogether (it serves no purpose).

ajdavis:
I'm declining to review this - please go ahead without me.

p-mongo:
The errors of my ways have been pointed out to me.

PR #455, opened 2019-01-29
SPEC-1185: Remove skips and add wtimeout in withTransaction tests

jmikola:
https://jira.mongodb.org/browse/SPEC-1185

### DIFF
--- a/source/transactions-convenient-api/tests/commit-retry.json
+++ b/source/transactions-convenient-api/tests/commit-retry.json
@@ -5,7 +5,6 @@
   "tests": [
     {
       "description": "commitTransaction succeeds after multiple connection errors",
-      "skipReason": "SPEC-1185 Drivers should use majority write concern when retrying commitTransaction",
       "failPoint": {
         "configureFailPoint": "failCommand",
         "mode": {
@@ -92,7 +91,8 @@
               },
               "autocommit": false,
               "writeConcern": {
-                "w": "majority"
+                "w": "majority",
+                "wtimeout": 10000
               },
               "readConcern": null,
               "startTransaction": null
@@ -111,7 +111,8 @@
               },
               "autocommit": false,
               "writeConcern": {
-                "w": "majority"
+                "w": "majority",
+                "wtimeout": 10000
               },
               "readConcern": null,
               "startTransaction": null
@@ -133,7 +134,6 @@
     },
     {
       "description": "commitTransaction retry only overwrites write concern w option",
-      "skipReason": "SPEC-1185 Drivers should use majority write concern when retrying commitTransaction",
       "failPoint": {
         "configureFailPoint": "failCommand",
         "mode": {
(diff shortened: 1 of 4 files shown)


p-mongo:
This PR means the tests as they exist in master are not passing, right?

jmikola:
@p-mongo: If you've not implemented [SPEC-1185](https://jira.mongodb.org/browse/SPEC-1185), these tests should currently fail for your driver. They previously would have been skipped (assuming your driver was honoring `skipReason`), and so shouldn't have prevented the driver from passing the overall test suite for `withTransaction()`. Does that answer the question?

p-mongo:
Test failures before this PR: https://gist.github.com/p-mongo/99e33374c8753c0aab9e697bb6ab0831

Test failures after this PR (still have 2): https://gist.github.com/p-mongo/f9b96a4d6b134196610507564a44f887

p-mongo:
Suggest adding https://gist.github.com/p-mongo/ec1fa044ab67f2c4b904f7c60948dfe3

p-mongo:
Are the transactions (non-api) tests passing for you all with the same driver tree that passes this PR? Because I have many failures there with missing wtimeouts.

p-mongo:
https://jira.mongodb.org/browse/DRIVERS-601 says:

> When commitTransaction is retried, either by the driver's internal retry-once logic *or explicitly by the user calling commitTransaction again*

commit.yml in transactions tests has:

<pre>
    operations:
      - name: startTransaction
        object: session0
      - name: commitTransaction
        object: session0
      # Rerun the commit (which does not increment the txnNumber).
      - name: commitTransaction
        object: session0
</pre>

Is the second commit sent with w:majority and wtimeout:10000?

jmikola:
> commit.yml in transactions tests has:

@p-mongo This is technically applies to the changes from https://github.com/mongodb/specifications/pull/442, but I think the test is correct as written. That test entails re-running commit after an empty transaction (i.e. start has been called, but no other operations are called before commit), and [commitTransaction](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#committransaction) states:

> It is valid to call commitTransaction when the session is in the "starting transaction" or "transaction in progress" state. When the session is in the "starting transaction" state, meaning no operations have been performed on this transaction, drivers MUST NOT run the commitTransaction command.

For the initial sequence of start, commit, commit, no commands should be sent to the server. Then we have a sequence of start, insertOne, and commit, which has the observed command sequence from the test.

jmikola:
> Test failures after this PR (still have 2): https://gist.github.com/p-mongo/f9b96a4d6b134196610507564a44f887
>
> Suggest adding https://gist.github.com/p-mongo/ec1fa044ab67f2c4b904f7c60948dfe3

I concur that the second and third `commitTransaction` commands in the "commit is retried after commitTransaction UnknownTransactionCommitResult (NotMaster)" test should have `{ w: majority, wtimeout: 10000 }` included. Will update that now.

jmikola:
> I will put the rest of my feedback in #442 then.

Since that PR is closed, I might suggest just opening a new PR with your proposed changes (assuming there is still a problem with the current transaction spec tests in `master`).

PR #459, opened 2019-01-29
Nest withTransaction's callback under operation.arguments

jmikola:
The previous spec tests weren't consistent with the test format. Thanks to @mbroadst for pointing that out.

Note: this may break your existing test harness.

### DIFF
--- a/source/transactions-convenient-api/tests/README.rst
+++ b/source/transactions-convenient-api/tests/README.rst
@@ -98,6 +98,24 @@ Each YAML file has the following keys:
       - ``data``: The data that should exist in the collection after the
         operations have run.
 
+``withTransaction`` Operation
+`````````````````````````````
+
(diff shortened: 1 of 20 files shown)


p-mongo:
Also, callback should be documented in the tests readme, it is currently not.

source/transactions-convenient-api/tests/README.rst
```diff
@@ -98,6 +98,24 @@ Each YAML file has the following keys:
       - ``data``: The data that should exist in the collection after the
         operations have run.
 
+``withTransaction`` Operation
+`````````````````````````````
```
jmikola: @p-mongo: Is this sufficient documentation?

jmikola: If not, I'll fix in a future PR.

PR #460, opened 2019-01-30
SPEC-1207 List all known examples of writeConcernError

ShaneHarvey:
Jira: https://jira.mongodb.org/browse/SPEC-1207

### DIFF
--- a/source/read-write-concern/read-write-concern.rst
+++ b/source/read-write-concern/read-write-concern.rst
@@ -13,7 +13,7 @@ Read and Write Concern
 :Type: Standards
 :Server Versions: 2.4+
 :Last Modified: January 29, 2019
-:Version: 1.5
+:Version: 1.5.1
 
 .. contents::
 
@@ -406,10 +406,12 @@ command failure. For example:
 
 - ``{ok:0, code: 251, codeName: "NoSuchTransaction", writeConcernError: {code: 91, codeName: "ShutdownInProgress"}}`` [#]_
 
-.. [#] This is only possible in a sharded cluster. When a write goes to
+.. [#] This is only possible in a sharded cluster. When a write is routed to
        multiple shards and more than one shard returns a writeConcernError,
-       then mongos will construct a "WriteConcernFailed" error where the errmsg
-       field contains the stringified writeConcernError from each shard.
+       then mongos will construct a new writeConcernError with the
+       "WriteConcernFailed" error code and an errmsg field contains the
+       stringified writeConcernError from each shard. Note that each shard may
+       return a different writeConcernError.
 
 .. [#] See https://jira.mongodb.org/browse/SERVER-38850
 
@@ -535,4 +537,4 @@ Version History
   - 2017-11-17 : Added list of commands that support readConcern 
   - 2017-12-18 : Added "available" to Readconcern level.
   - 2017-05-29 : Added user management commands to list of commands that write 
-  - 2017-05-29 : Added section listing all known examples of writeConcernError.
+  - 2019-01-29 : Added section listing all known examples of writeConcernError.


source/read-write-concern/read-write-concern.rst
```diff
@@ -380,6 +381,38 @@ Drivers SHOULD NOT parse server replies for "writeConcernError" in generic comma
...
+
+Note also that it is possible for a writeConcernError to be attached to a
+command failure. For example:
+
+- ``{ok:0, code: 251, codeName: "NoSuchTransaction", writeConcernError: {code: 91, codeName: "ShutdownInProgress"}}`` [#]_
+
+.. [#] This is only possible in a sharded cluster. When a write goes to
+       multiple shards and more than one shard returns a writeConcernError,
+       then mongos will construct a "WriteConcernFailed" error where the errmsg
+       field contains the stringified writeConcernError from each shard.
```
behackett: Can you be explicit here that different shards can return different write concern errors so this error message can refer to multiple different errors?

source/read-write-concern/read-write-concern.rst
```diff
@@ -502,3 +535,4 @@ Version History
   - 2017-11-17 : Added list of commands that support readConcern 
   - 2017-12-18 : Added "available" to Readconcern level.
   - 2017-05-29 : Added user management commands to list of commands that write 
+  - 2017-05-29 : Added section listing all known examples of writeConcernError.
```
jmikola: This should use the current date.

source/read-write-concern/read-write-concern.rst
```diff
@@ -12,7 +12,7 @@ Read and Write Concern
 :Status: Approved
 :Type: Standards
 :Server Versions: 2.4+
-:Last Modified: December 18, 2017
+:Last Modified: January 29, 2019
 :Version: 1.5
```
jmikola: Since this is an informational change (not new functionality), I'd consider it a patch and would suggest bumping to 1.5.1.

source/read-write-concern/read-write-concern.rst
```diff
@@ -380,6 +381,38 @@ Drivers SHOULD NOT parse server replies for "writeConcernError" in generic comma
...
+- ``{ok:1, writeConcernError: {code: 11601, codeName: "Interrupted"}}``
+- ``{ok:1, writeConcernError: {code: 11602, codeName: "InterruptedDueToReplStateChange"}}``
+- ``{ok:1, writeConcernError: {code: 64, codeName: "WriteConcernFailed", errmsg: "waiting for replication timed out", errInfo: {wtimeout: True}}}``
+- ``{ok:1, writeConcernError: {code: 64, codeName: "WriteConcernFailed", errmsg: "multiple errors reported : {...} at shardName1 :: and :: {...} at shardName2"}}`` [#]_
+- ``{ok:1, writeConcernError: {code: 50, codeName: "MaxTimeMSExpired"}}``
+- ``{ok:1, writeConcernError: {code: 100, codeName: "UnsatisfiableWriteConcern", errmsg: "Not enough data-bearing nodes"}}``
+- ``{ok:1, writeConcernError: {code: 79, codeName: "UnknownReplWriteConcern"}}``
+
+Note also that it is possible for a writeConcernError to be attached to a
+command failure. For example:
```
jmikola: Is a `writeConcernError` being attached to a command failure only relevant for mongos? You only present a single example here, which includes a footnote to the longer explanation. If so, I think we should just combine the footnote and this sentence into a single paragraph about mongos behavior.

source/read-write-concern/read-write-concern.rst
jmikola: The footnote says that this pertains to a sharded cluster where WC errors may be concatenated, but the example doesn't show that. Do we have a better example to demonstrate the concatenation behavior?

Specifically, I'm curious what the numeric `code` field will look like. I don't believe there's a code in [error_codes.err](https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.err) specific to a mongos merge scenario.

source/read-write-concern/read-write-concern.rst
jmikola: Oh, I didn't realize that there were multiple footnotes. Disregard.

PR #462, opened 2019-01-30
fix(convenient-txns): `transactionOptions` => `options`

mbroadst:
`callback` was moved into the `arguments` yesterday, but we forgot
to include `transactionOptions` here as well. Additionally, the
`transactionOptions` were renamed `options` to be consistent with
the transactions spec test format.

### DIFF
--- a/source/transactions-convenient-api/tests/transaction-options.yml
+++ b/source/transactions-convenient-api/tests/transaction-options.yml
@@ -148,9 +148,9 @@ tests:
                   document: { _id: 1 }
                 result:
                   insertedId: 1
-        transactionOptions:
-          readConcern: { level: snapshot }
-          writeConcern: { w: 1 }
+          options:
+            readConcern: { level: snapshot }
+            writeConcern: { w: 1 }
     expectations:
       -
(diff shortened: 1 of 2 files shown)


source/transactions-convenient-api/tests/transaction-options.yml
```diff
@@ -148,9 +148,9 @@ tests:
                   document: { _id: 1 }
                 result:
                   insertedId: 1
-        transactionOptions:
-          readConcern: { level: snapshot }
-          writeConcern: { w: 1 }
+          options:
```
jmikola: Indentation was definitely off here and needed correction, but I'm going to revert the rename before merging. This argument is definitely named `transactionOptions` in the `withTransaction()` spec, and I'd like to keep it that way as I expect we may need to add other options down the line (e.g. client-side operation timeouts).

jmikola: Reverted in https://github.com/mongodb/specifications/commit/a4a95b01c5dabb1edfdd7676408ce303193f1957.

PR #464, opened 2019-01-31
Update tests to use an "operations" array

vincentkam:
- Update README to reflect the use of an array named  "operations"
  instead single "operation"

### DIFF
--- a/source/retryable-reads/tests/README.rst
+++ b/source/retryable-reads/tests/README.rst
@@ -45,10 +45,8 @@ Network Error Tests
 ===================
 
 Network error tests are expressed in YAML and should be run against a standalone,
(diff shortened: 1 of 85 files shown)


source/retryable-reads/tests/README.rst
```diff
@@ -78,8 +78,8 @@ Each YAML file has the following keys:
   - ``failPoint``: Optional, a server failpoint to enable expressed as the
     configureFailPoint command to run on the admin database.
```
ShaneHarvey: Can we remove the SERVER-34943 comment now too? I think 4.1.7 is out.

vincentkam: Removed!

vincentkam:
Rebased, merged, and pushed! 😺

PR #468, opened 2019-02-04
SPEC-1204 Specify behavior of conflicting TLS options in the URI

saghm:


source/uri-options/uri-options.rst
```diff
@@ -36,6 +36,28 @@ document are to be interpreted as described in
...
+to the security implications of certain options, drivers MUST raise an
+error to the user during parsing if any of the following circumstances
+occur:
+
+1. Both ``tlsInsecure`` and ``tlsAllowInvalidCertificates`` appear in the 
+   URI options.
+2. Both ``tlsInsecure`` and ``tlsAllowInvalidHostnames`` appear in the 
+   URI options.
+3. All instances of ``tls`` and ``ssl`` in the URI options do not have the
+   same value. For ease of implementation, drivers MAY fulfill this
```
behackett: I think it would be weird an annoying if "tls=true&ssl=true" raised an exception.

behackett: Sorry, that comment wasn't super actionable. Let's require your original idea, the first sentence of bullet 3.

saghm: Sounds good; I'll update it to remove the second sentence

PR #470, opened 2019-02-06
add sleeps to make cmap tests more consistent

saghm:
When implementing the CMAP spec test runner, it was suggested that I have it sleep in between each operation being run in order to facilitate separate threads performing operations in a more deterministic order. Given that there are only two places in the tests where comments suggest that the operations may not be deterministic and that the spec runner description already provides a mechanism for specifying sleeps of a given length, it probably makes more sense to just add `wait` operations in the places where they're needed to save each implementation from having to do it manually. As an added bonus, this also makes the test run faster, as it greatly cuts down on the total amount of time the runner needs to sleep.

### DIFF
--- a/source/connection-monitoring-and-pooling/tests/wait-queue-fairness.yml
+++ b/source/connection-monitoring-and-pooling/tests/wait-queue-fairness.yml
@@ -26,14 +26,14 @@ operations:
     thread: thread2
     label: conn2
   - name: wait
-    ms: 20
+    ms: 10
   - name: start
     target: thread3
   - name: checkOut
     thread: thread3
     label: conn3
   - name: wait
-    ms: 30
+    ms: 10
   - name: start
     target: thread4
   - name: checkOut
--- a/source/connection-monitoring-and-pooling/tests/wait-queue-fairness.yml
+++ b/source/connection-monitoring-and-pooling/tests/wait-queue-fairness.yml
@@ -1,8 +1,9 @@
 version: 1
 style: unit
-description: must issue Connections to threads in the order that the threads eneterd the queue
+description: must issue Connections to threads in the order that the threads entered the queue
 poolOptions:
   maxPoolSize: 1
+  waitQueueTimeoutMS: 1000
 operations:
     # Check out sole connection in pool
   - name: checkOut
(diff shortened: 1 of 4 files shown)


source/connection-monitoring-and-pooling/tests/wait-queue-fairness.yml
```diff
@@ -1,31 +1,39 @@
...
     label: conn1
+  - name: wait
+    ms: 10
   - name: start
     target: thread2
   - name: checkOut
     thread: thread2
     label: conn2
+  - name: wait
+    ms: 20
```
daprahamian: Is it necessary for the `ms` count to increase for each one?

saghm: The idea is that if all of the threads are running concurrently (which will happen in languages without a GIL), there will be a race if they all sleep for the same amount of time, since they will wake up roughly the same time. By increasing each sleep, we can ensure that the threads wake up in the order we want.

daprahamian: Ah. In that case, you need to add the `thread` to each of these wait calls. Currently, all these `wait`s are on the main thread.

saghm: I think this still works; the later threads aren't started until after a sleep, meaning they won't attempt to check out a connection until after the previous one already has had a chance to enter the wait queue.

daprahamian: it will definitely still work, but if we are sleeping on the main thread instead of individual threads, they might as well only be 10ms per sleep

p-mongo:
The other aspect we discussed was having the threads be woken up/get connections in a different order from the order in which the threads are created.

source/connection-monitoring-and-pooling/tests/wait-queue-fairness.yml
saghm: Ah, I see what you're saying; I'll update it

saghm:
@p-mongo I think this still takes care of that; by only starting each thread after a sleep, we give the previous thread enough time to enter the wait queue, so there's no need for them to be "woken up" at all.

PR #471, opened 2019-02-07
 SPEC-1168 Update transaction spec for sharded transaction recoveryToken

ShaneHarvey:
Jira: https://jira.mongodb.org/browse/SPEC-1168

Updates transaction spec to support sharded transaction recoveryToken. This also augments the mongos session pinning rules added in [SPEC-1140](https://jira.mongodb.org/browse/SPEC-1140) because we can now send retries of commit/abort to a new mongos.

Note that to test these changes you must use the latest version of the server (at least v4.1.7-232-g473e96a) and you should decrease the `transactionLifetimeLimitSeconds` setting (problem described in the test README).


source/transactions/tests/commit.json
```diff
@@ -356,7 +356,6 @@
     },
     {
       "description": "write concern error on commit",
-      "skipReason": "SERVER-37458 Mongos does not yet support writeConcern on commit",
```
ShaneHarvey: SERVER-37458 was completed.

source/transactions/tests/README.rst
```diff
@@ -149,48 +159,19 @@ ensure that no new bugs have been introduced related to arbiters.)
...
-          'killAllSessionsByPattern': [
-              {'lsid': {'id': uuid}} for uuid in recorded_session_uuids]},
-          session=session_for_cleanup)
-
-   - When testing against a sharded cluster, create a list of MongoClients that
-     are directly connected to each mongos. Run the `killAllSessions`
-     (or `killAllSessionsByPattern`) command on ALL mongoses.
-
-   .. _SERVER-38335: https://jira.mongodb.org/browse/SERVER-38335
-
```
ShaneHarvey: SERVER-38335 was completed.

source/transactions/tests/pin-mongos.yml
```diff
@@ -296,3 +305,180 @@ tests:
...
+          database_name: admin
+
+    outcome:
+      collection:
+        data:
+          - _id: 1
+          - _id: 2
+          - _id: 3
+
+  - description: remain pinned after errors within a transaction and abort
```
ShaneHarvey: I would like to hear some thoughts one this behavior. As described in the spec (and tested here) a session is pinned to a mongos even after a network error while running a CRUD operation within the transaction. This has the interesting side effect that a subsequent `abort` attempt will block until the serverSelectionTimeout (assuming the operation failed because mongos is down/unreachable).

Since it will be common to retry the entire transaction in this case it seems somewhat pointless to remain pinned for the initial abort attempt. I did not attempt to workaround this case because I wanted to keep the session pin/unpin behavior simple. Any thoughts?

source/transactions/tests/README.rst
```diff
@@ -149,48 +159,19 @@ ensure that no new bugs have been introduced related to arbiters.)
 
 A driver that implements support for sharded transactions MUST also run these
 tests against a MongoDB sharded cluster with multiple mongoses and
-**server version 4.1.6 or later**. Including multiple mongoses (and
-initializing the MongoClient with multiple mongos seeds!) ensures that
-mongos transaction pinning works properly.
+**server version v4.1.7-232-g473e96a or later**. Some tests require
```
jyemin: Let's just say 4.1.8, or even 4.2.

source/transactions/tests/README.rst
```diff
@@ -210,15 +191,18 @@ Then for each element in ``tests``:
...
+     created with multiple (valid) mongos seed addreses.
 
 #. Call ``client.startSession`` twice to create ClientSession objects
    ``session0`` and ``session1``, using the test's "sessionOptions" if they
    are present. Save their lsids so they are available after calling
    ``endSession``, see `Logical Session Id`.
 #. For each element in ``operations``:
 
+   - If the operation ``name`` is a special test operation type, execute it and
+     go to the next operation, otherwise procede to the next step.
```
jyemin: procede -> proceed

source/transactions/tests/README.rst
```diff
@@ -86,7 +86,15 @@ Each YAML file has the following keys:
...
     skipped.
 
-  - ``clientOptions``: Optional, parameters to pass to MongoClient().
+  - ``topology``: Optional. An array of server topologies against which to run
+    the test. Valid topologies are "replicaset" and "sharded". The default is
+    ["replicaset", "sharded"].
+
+  - ``clientOptions``: Optional, parameters to pass to MongoClient():
+
+    - ``useMultipleMongoses``: Optional, boolean. If true and this test is
```
jyemin: This doesn't seem appropriate as a client option.   Consider making it top-level.

source/transactions/tests/README.rst
```diff
@@ -382,3 +401,31 @@ instead.
...
           assert len(addresses) > 1
+
+Q & A
+=====
+
+Why do some tests appear to hang for 60 seconds on a sharded cluster?
+
+`````````````````````````````````````````````````````````````````````
+There are two cases where this can happen. When the initial commitTransaction
+attempt fails on mongos A and is retryed on mongos B, mongos B will block
```
jyemin: retryed -> retried

source/transactions/tests/README.rst
```diff
@@ -382,3 +401,31 @@ instead.
...
+waiting for the transaction to complete. However because the initial commit
+attempt failed, the transaction will only completed after it is automatically
```
jyemin: Replace with "... the command will only complete after the transaction is automatically aborted ..."

source/transactions/tests/README.rst
```diff
@@ -382,3 +401,31 @@ instead.
...
+aborted for exceeding the shard's transactionLifetimeLimitSeconds setting.
+
+The second case is when a *single-shard* transaction is commited successfully
```
jyemin: commited -> committed

source/transactions/tests/README.rst
```diff
@@ -382,3 +401,31 @@ instead.
...
+on mongos A and then explicitly commited again on mongos B, mongos B will also
```
jyemin: commited -> committed

jyemin: Start a new sentence instead of using a comma

source/transactions/tests/README.rst
```diff
@@ -382,3 +401,31 @@ instead.
...
+block until the transactionLifetimeLimitSeconds timeout is hit at which point
+``{ok:1}`` will be returned. `SERVER-39349`_ requests that recovering the
+outcome of a completed single-shard transaction should not block.
+Note that this test suite only includes single shard transactions.
+
+To workaorund these issues, drivers SHOULD decrease the transaction timeout
```
jyemin: workaorund -> workaround

source/transactions/tests/README.rst
```diff
@@ -382,3 +401,31 @@ instead.
...
+setting by running setParameter **on each shard**. Setting the timeout to 3
```
jyemin: Will drivers-evergreen-tools be configured to do this?  It seems painful for a driver to discover eact shard, find the primary, send this command, and then reset it afterwards.

source/transactions/tests/README.rst
jyemin: I'm also not clear what the point is.  Since at least some of the sharded-only tests require this to be true, to pass any test run a driver will need to have multiple mongos available.  So can we just say that the sharded cluster tests require multiple mongos servers in the URI?

source/transactions/tests/README.rst
```diff
@@ -317,6 +329,13 @@ A ``readConcern.afterClusterTime`` value of ``42`` in a command-started event
 is a fake cluster time. Drivers MUST assert that the actual command includes an
 afterClusterTime.
 
+recoveryToken
+^^^^^^^^^^^^^
+
+A ``recoveryToken`` value of ``42`` in a command-started event is a fake
+recovery token. Drivers MUST assert that the actual command includes a
+"recoveryToken" field.
```
jyemin: Consider:

> Drivers MUST assert that the actual command includes a
"recoveryToken" field equal to the recovery token stored in the ClientSession

It seems a little loose otherwise.

source/transactions/tests/mongos-recovery-token.yml
```diff
@@ -0,0 +1,362 @@
+database_name: &database_name "transaction-tests"
+collection_name: &collection_name "test"
+
+data: []
+
+tests:
+  - description: commitTransaction explicit retries succeed on new mongos
```
jyemin: The test doesn't assert what the description says.  What are we really trying to test here?  That the recovery token is included on all commit attempts?  That it's the same for every attempt?

source/transactions/tests/mongos-recovery-token.yml
```diff
@@ -0,0 +1,362 @@
...
+        failPoint:
+          configureFailPoint: failCommand
+          mode: { times: 1 }
+          data:
+              failCommands: ["commitTransaction"]
+              closeConnection: true
+      # The first commitTransaction sees a retryable connection error due to
+      # the fail point and also fails on the server. The retry attempt on a
+      # new mongos will wait for the transaction to timeout and will fail
+      # because the transaction was aborted. Note that the retry attempt should
```
jyemin: I think this is racy.  If SDAM monitor for session0's pinned mongos happens to kick in at just the wrong time, the driver could select it.  

However, in that event I think the test will still pass.

source/transactions/tests/mongos-recovery-token.yml
```diff
@@ -0,0 +1,362 @@
...
+            startTransaction:
+            autocommit: false
+            writeConcern:
+            recoveryToken: 42
+          command_name: abortTransaction
+          database_name: admin
+
+    outcome:
+      collection:
+        data: []
```
jyemin: Add newline

source/transactions/tests/pin-mongos.yml
```diff
@@ -296,3 +305,180 @@ tests:
           - {_id: 2}
           - {_id: 3}
           - {_id: 4}
+
+  - description: remain pinned after errors within a transaction and commit
```
jyemin: Perhaps just

> remain pinned after errors within a transaction

source/transactions/tests/pin-mongos.yml
```diff
@@ -296,3 +305,180 @@ tests:
...
+      - name: insertOne
+        object: collection
+        arguments:
+          session: session0
+          document:
+            _id: 4
+        result:
+          errorLabelsContain: ["TransientTransactionError"]
+          errorLabelsOmit: ["UnknownTransactionCommitResult"]
+      # Session remains pinned to the first mongos and commitTransaction
```
jyemin: So the idea is that if a driver unpins, eventually this test is bound to fail?

source/transactions/tests/pin-mongos.yml
jyemin: Can you propose an alternative so we can evaluate?  I imagine it would involve some sort of storage of the last exception in the session and then basing the the abortTransaction behavior off of that?

Let's check with Aly on this as well.

source/transactions/tests/README.rst
jmikola: I agree that extra validation like this may be useful, but it doesn't seem appropriate for the `42` placeholder.

Perhaps we can borrow language from the [logical session ID](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst#logical-session-id) assertion, which uses a reference to a session object and discusses how it should be interpreted for the assertion.

source/transactions/tests/mongos-recovery-token.yml
```diff
@@ -0,0 +1,362 @@
+database_name: &database_name "transaction-tests"
+collection_name: &collection_name "test"
+
+data: []
+
+tests:
+  - description: commitTransaction explicit retries succeed on new mongos
+    topology: ["sharded"]
```
jmikola: It looks like this entire test pertains to mongos, so that might agree with making `topology` a top-level field.

source/transactions/tests/pin-mongos.yml
```diff
@@ -1,20 +1,17 @@
...
 #
-# The assertion is implicit: that all the read operations succeed.
-# If the driver does not properly pin to a single mongos then one of the
-# operations in a transaction will eventually be sent to a different mongos,
-# which is unaware of the transaction, and the mongos will return a command
-# error. An example of such an error is:
+# In tests that don't include command-started events the assertion is implicit:
+# that all the read operations succeed. If the driver does not properly pin to
+# a single mongos then one of the operations in a transaction will eventually
+# be sent to a different mongos, which is unaware of the transaction, and the
```
jmikola: Pedantic note: since [server selection uses random selection](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#why-change-from-mongos-high-availablity-ha-to-random-selection), is there a slight chance these tests will all hit the same mongos?

ShaneHarvey: Yes that's why it says "eventually". Do you want me to call this out explicitly? When I run these tests without pinning they always fail.

source/transactions/tests/mongos-recovery-token.yml
ShaneHarvey: Let's continue this conversation on the other comment thread: https://github.com/mongodb/specifications/pull/471/files/1562085a8001c04464a0efaf5fd134d6cf7c97df#r255563911

source/transactions/tests/mongos-recovery-token.yml
ShaneHarvey: Yeah it's just supposed to test that explicitly retrying commit succeeds. Renamed to simply "commitTransaction explicit retries succeed".

source/transactions/tests/README.rst
```diff
@@ -280,6 +264,34 @@ Then for each element in ``tests``:
...
+determined by the "session" argument (either "session0" or "session1").
+The session must already be pinned to a mongos server.
+
+Here is an example which instructs the test runner to enable the failCommand
+fail point on the mongos server which "session0" is pinned to::
+
+      # Enable the fail point only on the Mongos that session0 is pinned to.
+      - name: failPoint
+        session: session0
+        failPoint:
```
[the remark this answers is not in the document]
ShaneHarvey: Good point. I nested the arguments under 'arguments', added `object: testRunner`, and renamed failPoint to 'targetedFailPoint' to distinguish it from the higher level 'failPoint' directive.

source/transactions/transactions.rst
```diff
@@ -744,24 +753,59 @@ format:
...
+transaction but introduce two new driver concepts: mongos pinning and the
+``recoveryToken`` field.
 
-The server supports sharded transactions starting in MongoDB 4.2 (
-maxWireVersion 8). The server requires that drivers MUST send all commands for
-a single transaction to the same mongos.
+Mongos Pinning
+~~~~~~~~~~~~~~
+
+The server requires that drivers MUST send all commands for
```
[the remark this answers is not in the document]
ShaneHarvey: Sounds good, done.

source/transactions/transactions.rst
```diff
@@ -744,24 +753,59 @@ format:
...
 
+recoveryToken field
+~~~~~~~~~~~~~~~~~~~
+
+The ``recoveryToken`` field enables the driver (and the server) to recover
+the outcome of a sharded transaction on a new mongos.
+
+When a driver runs a command within a transaction, the mongos response
+includes a ``recoveryToken`` field. Drivers MUST append this ``recoveryToken``
+field to any subsequent commitTransaction or abortTransaction commands.
```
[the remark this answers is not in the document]
ShaneHarvey: Yes that was my intention. I updated the section to make that more clear.

source/transactions/tests/README.rst
jyemin: OK, still doesn't seem like a client option though.  I think of those as things like "retryReads=true"

jyemin: I'm going to resolve this conversation since there is a similar thread below. I'm fine with whatever you decide here.

source/transactions/tests/README.rst
```diff
@@ -97,7 +105,9 @@ Each YAML file has the following keys:
   - ``operations``: Array of documents, each describing an operation to be
     executed. Each document has the following fields:
 
-    - ``name``: The name of the operation on ``object``.
+    - ``name``: The name of the operation on ``object`` or the name of a
+      special test operation to run, eg. "failPoint". See
+      `Special Test Operations`_.
```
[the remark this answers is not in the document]
jyemin: We decided not to YAML-ify the live failover spec tests, so I don't think this is necessary.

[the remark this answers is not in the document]
ShaneHarvey: I already changed the format from:
```
      - name: failPoint
        session: session0
        failPoint:
          mode: { times: 1 }
          data:
            closeConnection: true
```
To:
```
      - name: targetedFailPoint
        object: testRunner
        arguments:
          session: session0
          failPoint:
            mode: { times: 1 }
            data:
              closeConnection: true
```

I prefer the newer format so I'm going to keep it unless there are objections.

source/transactions/tests/pin-mongos.yml
ShaneHarvey: > I imagine it would involve some sort of storage of the last exception in the session and then basing the the abortTransaction behavior off of that?

My alternative proposal would be to unpin after transient errors within a transaction. This way a subsequent abort attempt would already be unpinned and could immediately select any available mongos. In this proposal a commit attempt that selects a new mongos would error but I wouldn't expect an app to commit after a transient error within the transaction.

jyemin: I like that better if it works

source/transactions/tests/README.rst
```diff
@@ -86,7 +86,15 @@ Each YAML file has the following keys:
...
-  - ``clientOptions``: Optional, parameters to pass to MongoClient().
+  - ``topology``: Optional. An array of server topologies against which to run
+    the test. Valid topologies are "replicaset" and "sharded". The default is
+    ["replicaset", "sharded"].
+
+  - ``clientOptions``: Optional, parameters to pass to MongoClient():
+
+    - ``useMultipleMongoses``: Optional, boolean. If true and this test is
+      running against a sharded cluster, intialize the MongoClient for this
+      test with multiple mongos seed addresses.
```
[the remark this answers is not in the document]
jmikola: IMO, `clientOptions` are values that should either be easy to compose into a connection string or fed directly as-is for drivers that accept an options dictionary (or equivalent) in their MongoClient constructor.

AFAIK, `useMultipleMongoes` is neither of those and is instead telling the test runner that it should be using a connection string that has multiple mongos hosts in it. Either way, test runners will need to process this, but by moving it out, I think you'll save various drivers from having to manually pluck this value out before they pass `clientOptions` on to their MongoClient ctors.

source/transactions/tests/README.rst
```diff
@@ -86,7 +86,15 @@ Each YAML file has the following keys:
   - ``skipReason``: Optional, string describing why this test should be
     skipped.
 
-  - ``clientOptions``: Optional, parameters to pass to MongoClient().
+  - ``topology``: Optional. An array of server topologies against which to run
+    the test. Valid topologies are "replicaset" and "sharded". The default is
+    ["replicaset", "sharded"].
```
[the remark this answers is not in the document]
jmikola: Oy, I missed that retryable reads added `topology` at the per-operation level. I don't see anything in that README that talks about a default value, though, but I suppose that's a separate question for Vincent.

Is it feasible for both specs to keep this at the top-level? I was talking to @prashantmital about this only yesterday and explaining why I thought min/max server versions and topologies were suitable for top-level fields, as it cuts down on repeated information within the `tests[i]` document and encourages spec authors to split up their test files. Assuming it's possible to get retryable reads on board as well, while that is still being reviewed, I'd like to see this changed instead of going all-in on the change stream spec's format -- from what I saw, it was only used there because standalone and replica sets error tests were placed in the same file.

source/transactions/tests/README.rst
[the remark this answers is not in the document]
jmikola: I think borrowing existing `object:` syntax is sensible, so I'm :+1: on keeping this.

I don't follow where "targeted" comes from, unless that refers to "targeting" the pinned mongos for the session argument. I don't think it would be targetted outside of a mongos topology, though. Do you plan that operation name being used exclusively for mongos topologies?

I imagine down the line there may be cases where we'd want to target a specific replica set member -- could "targetedFailPoint" be re-used there as well? If so, I might suggest replacing the `session` argument here with `target` and still provide a session object. For a mongos topology, you could document that the target is inferred from the pinned mongos within the session, and we could define a different interpretation as needed down the line. LGTM if you prefer this as-is, though. If kept as-is, perhaps we could use something like `readPreference` to handle replica set targeting down the line (if the need arises).

source/transactions/tests/README.rst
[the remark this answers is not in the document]
vincentkam: I'm happy to move ``topology`` to the top-level for retryable reads. Good catch about the test plan README, I'll update it now. The default is currently ``["single", "replicaset", "sharded"]``

source/transactions/tests/pin-mongos.yml
jmikola: No worries. Was mainly asking for my own understanding :)

Fine as-is.

source/transactions/tests/README.rst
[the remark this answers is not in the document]
ShaneHarvey: > I don't follow where "targeted" comes from, unless that refers to "targeting" the pinned mongos for the session argument. 

The failpoint is targeted to the pinned session. I can rename to `mongosFailPoint` or `pinnedFailPoint` if you prefer, I just want to distinguish it from the higher level `failPoint` directive.

> Do you plan that operation name being used exclusively for mongos topologies?

I don't foresee using this outside of sharded transactions in the 4.2 timeframe so I don't think we should put too much time into thinking about how it should work on other topologies. A new spec can introduce a new test operation if/when they need it.

source/transactions/tests/README.rst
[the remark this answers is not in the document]
ShaneHarvey: > I think you'll save various drivers from having to manually pluck this value out before they pass `clientOptions` on to their MongoClient ctors.

I actually view this as one of the other benefits of including it in  clientOptions. If a driver forgets to implement the `useMultipleMongoes` then they'll see an error (or warning in the URI parser) like this:
```
ERROR: test_transactions_pin_mongos_countDocuments (test.test_transactions.TestTransactions)
  File "/Users/shane/git/mongo-python-driver/test/__init__.py", line 410, in wrap
    return f(*args, **kwargs)
  File "/Users/shane/git/mongo-python-driver/test/__init__.py", line 410, in wrap
    return f(*args, **kwargs)
  File "/Users/shane/git/mongo-python-driver/test/test_transactions.py", line 535, in run_scenario
    client = rs_client(event_listeners=[listener], **client_options)
  File "/Users/shane/git/mongo-python-driver/test/utils.py", line 166, in rs_client
    return _mongo_client(h, p, **kwargs)
  File "/Users/shane/git/mongo-python-driver/test/utils.py", line 144, in _mongo_client
    **client_options)
  File "/Users/shane/git/mongo-python-driver/pymongo/mongo_client.py", line 573, in __init__
    dict(common.validate(k, v) for k, v in keyword_opts.items()))
  File "/Users/shane/git/mongo-python-driver/pymongo/mongo_client.py", line 573, in <genexpr>
    dict(common.validate(k, v) for k, v in keyword_opts.items()))
  File "/Users/shane/git/mongo-python-driver/pymongo/common.py", line 681, in validate
    value = validator(option, value)
  File "/Users/shane/git/mongo-python-driver/pymongo/common.py", line 130, in raise_config_error
    raise ConfigurationError("Unknown option %s" % (key,))
pymongo.errors.ConfigurationError: Unknown option useMultipleMongoses
```

If it's not included in clientOptions then some of the tests just silently pass with a single mongos seed address.

source/transactions/tests/README.rst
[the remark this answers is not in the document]
ShaneHarvey: I also promoted this to the top-level and made the default `["single", "replicaset", "sharded"]`.

source/transactions/tests/README.rst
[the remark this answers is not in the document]
jmikola: Per the [connection string spec](https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.rst#keys), I think a logged warning is the best we can hope for (outright errors on unsupported keys aren't kosher -- not sure if that's a deviation for PyMongo here). I can't speak for other drivers, but we definitely wouldn't catch this at all in PHP and libmongoc would entirely depend on whether a log handler is configured for spec tests (AFAIK it's only done for things like URI unit tests, where the test wants to assert that a warning was logged).

A point I didn't raise before is that keeping this within `clientOptions` means the spec format docs aren't entirely accurate when they say these are options for constructing a MongoClient. We'd probably need to note any special handling (for plucking things out) in any spec READMEs that use this.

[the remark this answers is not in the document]
ShaneHarvey: Okay I relent. I moved it up one level.

source/transactions/transactions.rst
```diff
@@ -744,24 +753,59 @@ format:
...
 server selection.
 
+recoveryToken field
+~~~~~~~~~~~~~~~~~~~
+
+The ``recoveryToken`` field enables the driver (and the server) to recover
+the outcome of a sharded transaction on a new mongos.
+
+When a driver runs a command within a transaction, the mongos response
+includes a ``recoveryToken`` field. Drivers MUST append this ``recoveryToken``
```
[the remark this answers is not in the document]
ShaneHarvey: Yes, it is a safe assumption. Updated the spec.

source/transactions/tests/README.rst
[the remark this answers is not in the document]
ShaneHarvey: Updated now that drivers can assume recoveryToken is always a BSON document.

source/transactions/tests/README.rst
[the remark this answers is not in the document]
jmikola: I'm fine with keeping this as-is. Thanks for explaining the design rationale!

jyemin:
@ShaneHarvey, there are a few unresolved comment threads that I think are in your court.  Let me know if I'm mistaken.

Jeff

source/transactions/tests/pin-mongos.yml
ShaneHarvey: I updated the prose to say that drivers MUST unpin a session after any transient error within a transaction. This ensures that a subsequent abort does not block and so a subsequent transaction (like withTransaction) can proceed immediately.

Note this new behavior comes with its own downside though. Due to server limitations in 4.2, an abort on a new mongos will not actually abort the in-progress transaction on the shards. The transaction will remain open and potentially block other operations. This case is possible when a mongos crashes regardless of driver behavior, but if the connection error was simply a blimp and the mongos is healthy the entire time unpinning will cause more transactions to be left open on the server.

Personally, I prefer the updated prose. @jmikola, @EshaMaharishi, thoughts?

source/transactions/transactions.rst
```diff
@@ -744,24 +753,71 @@ format:
...
+Transient errors indicate that the transaction in question has already been
+aborted or that the pinned mongos is down/unavailable. Unpinning the session
+ensures that a subsequent abortTransaction does not block waiting on a server
+that is unreachable.
+
+Additionally, Drivers MUST unpin a ClientSession after running any
+commitTransaction or abortTransaction attempt regardless if that attempt
+succeeded or failed. After the initial commit or abort attempt, any mongos can
+satisfy a subsequent retry. Note, when the initial attempt fails with a
+retryable writes error, the automatic retry attempt MUST perform normal
```
jmikola: This should simply refer to "retryable error". I believe that term is already used elsewhere and refers back to the original definition in the retryable writes spec.

source/transactions/transactions.rst
```diff
@@ -744,24 +753,71 @@ format:
...
-commitTransaction and abortTransaction and any retries thereof, to the
-same mongos.
+send all subsequent commands that are part of the same transaction (excluding
+retries of commitTransaction and abortTransaction) to the same mongos.
+
+When to unpin
+^^^^^^^^^^^^^
+
+Drivers MUST unpin a ClientSession when a command within a transaction
+fails with a TransientTransactionError (including a server selection error).
```
jmikola: Isn't "(including a server selection error)" redundant in light of the description of a [TransientTransactionError](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#transienttransactionerror)? This includes network errors, server selection failures, and any other errors the server chooses to label.

Separate question: do we avoid discussing UnknownTransactionCommitResult here because the subsequent paragraph talks about unpinning as soon as a commit/abort is issued (i.e. before such a label would even be assigned)?

source/transactions/transactions.rst
ShaneHarvey: Yep that's right. Fixed.

source/transactions/transactions.rst
ShaneHarvey: Fixed the "server selection error" redundancy. 

Do we need to unpin after an UnknownTransactionCommitResult error?  I wrote this thinking that the unpinning rule would also apply to the commitTransaction commands run as part of the commit helper.

jmikola: > apply to the commitTransaction commands run as part of the commit helper.

Are you referring to `ClientSession.commitTransaction()` or the `withTransaction()` helper? I'm not sure it really matters, as the paragraph below this says:

> Additionally, Drivers MUST unpin a ClientSession after running any commitTransaction or abortTransaction attempt regardless if that attempt succeeded or failed.

That surely covers both cases.

source/transactions/transactions.rst
```diff
@@ -744,24 +753,59 @@ format:
...
 
-The server supports sharded transactions starting in MongoDB 4.2 (
-maxWireVersion 8). The server requires that drivers MUST send all commands for
-a single transaction to the same mongos.
+Mongos Pinning
+~~~~~~~~~~~~~~
+
+The server requires that drivers MUST send all commands for
+a single transaction to the same mongos (excluding retries of commitTransaction
+and abortTransaction) therefore this spec introduces the concept of pinning a
```
EshaMaharishi: I don't think it's supported to send abortTransaction to a different mongos.

It may appear to work because
- if the txnNumber does not match the number on the second mongos, the second mongos will return NoSuchTransaction from beginOrContinue: https://github.com/mongodb/mongo/blob/acddc7f35f0373ccb2e8fe9d45f42304b2b74f95/src/mongo/s/transaction_router.cpp#L514-L519
- if the txnNumber matches the number on the second mongos, the second mongos will return NoSuchTransaction on seeing that it does not have the participant list: https://github.com/mongodb/mongo/blob/acddc7f35f0373ccb2e8fe9d45f42304b2b74f95/src/mongo/s/transaction_router.cpp#L671-L673

source/transactions/transactions.rst
```diff
@@ -744,24 +753,59 @@ format:
...
+ClientSession to a mongos.
 
 After the driver selects a mongos for the first command within a transaction,
 the driver MUST pin the ClientSession to the selected mongos. Drivers MUST
-send all commands that are part of the same transaction, including
-commitTransaction and abortTransaction and any retries thereof, to the
-same mongos.
+send all subsequent commands that are part of the same transaction (excluding
+retries of commitTransaction and abortTransaction) to the same mongos.
```
EshaMaharishi: Ditto

source/transactions/transactions.rst
```diff
@@ -744,24 +753,59 @@ format:
...
+
+When to unpin
+^^^^^^^^^^^^^
+
+A pinned ClientSession MUST be unpinned after running the initial
+commitTransaction or abortTransaction attempt regardless if that attempt
```
EshaMaharishi: Ditto throughout this paragraph

source/transactions/transactions.rst
```diff
@@ -744,24 +753,59 @@ format:
...
+succeeded or failed. When the commitTransaction or abortTransaction attempt
+fails on the pinned mongos, the session MUST be unpinned and any
+subsequent (automatic or explicit) retry attempt MUST perform normal mongos
+server selection. When a commitTransaction or abortTransaction
+attempt succeeds on the pinned mongos, the session MUST also be unpinned.
```
EshaMaharishi: Why do we recommend unpinning after sending commitTransaction once?

It seems better to remain pinned, because only the original mongos can actually commit the transaction - sending commitTransaction to a recovery mongos can *only* be used to *recover* the decision.

source/transactions/transactions.rst
EshaMaharishi: Currently abortTransaction on mongos does not look at the recoveryToken.

source/transactions/transactions.rst
```diff
@@ -744,24 +753,59 @@ format:
...
+
+When a driver runs a command within a transaction, the mongos response
+includes a ``recoveryToken`` field. Drivers MUST append this ``recoveryToken``
+field to any subsequent commitTransaction or abortTransaction commands.
+Sending the ``recoveryToken`` along with commitTransaction and abortTransaction
+allows the mongos to recover the state of a transaction even if the mongos was
+restarted or is not the same mongos which started the transaction.
+
+Drivers MUST treat the ``recoveryToken`` field as an opaque BSON field and
+relay it back to the server unchanged.
```
EshaMaharishi: Should we document when the recoveryToken is returned by mongos, and that it may change (it currently does not, but we want to leave that option open for the future)?

source/transactions/transactions.rst
ShaneHarvey: Yes, I've noted this server behavior elsewhere in the spec but I thought that the server will improve in the future to make aborting on a new mongos possible. The alternative is to not send abort at all (which has the same behavior as sending abort to a new mongos) or remain pinned to a potentially unreachable server and block for 30 seconds in server selection.

Is there a server ticket for this already or should I create one?

EshaMaharishi: If the server started supporting aborting a transaction through a recovery mongos, there would probably be drivers changes needed (like passing the participant list back to the client, so the client can pass it to the recovery mongos).

I don't recommend drivers ever sending abortTransaction to a recovery mongos, since we have not tested that against the server and it is currently undefined in the same way using multiple mongos for transaction statements is undefined... But up to you.

source/transactions/transactions.rst
ShaneHarvey: If there's a retryable error we have to unpin to run a retry attempt. If the commit succeeds then an explicit retry can recover the previous outcome on any mongos using the recoveryToken. So optimistically unpinning the session after a commit regardless of the outcome seemed like a reasonable simplification.

However, I've now changed the spec to mandate unpinning after any transient error within a transaction so this "simplification" really no longer applies. I think I can remove this and instead say that a session must remain pinned after a successful commit (just like any other successful operation within a transaction). This would also avoid the [SERVER-39349](https://jira.mongodb.org/browse/SERVER-39349) problem that makes some of our spec tests run slower than they should.

source/transactions/transactions.rst
```diff
@@ -580,6 +580,13 @@ transaction, the server will return an error. The ``writeConcern`` argument
 of the commitTransaction and abortTransaction commands has
 semantics analogous to existing write commands.
 
+Behavior of the recoveryToken field
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Only included for sharded transactions and only when running a
+commitTransaction or abortTransaction command. See the
```
renctan: recoveryToken is only for commitTransaction. For that reason, drivers should only expect recoveryToken to exists in mongos response if ok: 1

source/transactions/transactions.rst
```diff
@@ -744,24 +753,70 @@ format:
...
 Starting a new transaction on a pinned ClientSession MUST unpin the
 session. Additionally, any non-transaction operation using a pinned
 ClientSession MUST unpin the session and the operation MUST perform normal
 server selection.
 
+recoveryToken field
+~~~~~~~~~~~~~~~~~~~
+
+The ``recoveryToken`` field enables the driver to recover the outcome of a
+sharded transaction on a new (or restarted) mongos. [#]_
```
renctan: not any ops, but on commitTransaction commands that had an ambiguous result: like network error.

source/transactions/transactions.rst
```diff
@@ -744,24 +753,70 @@ format:
...
+
+When a driver runs a command within a transaction, the mongos response
+includes a ``recoveryToken`` field. Drivers MUST track the most recently
+received ``recoveryToken`` field and MUST append this field to any subsequent
+commitTransaction or abortTransaction commands.
```
renctan: only commitTransaction

source/transactions/transactions.rst
ShaneHarvey: I opened https://jira.mongodb.org/browse/SERVER-39692 for the server to support (or define) this behavior. Please see this conversation for why we feel drivers must unpin and send abort to a new mongos: https://github.com/mongodb/specifications/pull/471#discussion_r254515917

source/transactions/transactions.rst
ShaneHarvey: Can you expand on this? I'm not sure if you're asking for changes or clarification.

source/transactions/transactions.rst
ShaneHarvey: Updated to only send recoveryToken with commitTransaction.

source/transactions/transactions.rst
ShaneHarvey: The section below says:
```
When a driver runs a command within a transaction, the mongos response
includes a ``recoveryToken`` field.
```
Do you want me to change it to say only (ok:1) responses include it? In my testing command errors also include a recoveryToken field.

renctan: It is not guaranteed to be there if the response is not ok: 1. In general, these are the cases when mongos caused the error (as opposed to errors from mongod bubbled up to mongos).

source/transactions/transactions.rst
renctan: The sentence sounds like recoveryToken can be used to recover the outcome of any ops in a sharded transaction. The token is only meant to recover the outcome of a commitTransaction command.

source/transactions/transactions.rst
ShaneHarvey: I removed the unpin after successful commit/abort rule and updated the tests.

source/transactions/transactions.rst
ShaneHarvey: Related question: should drivers attach the recoveryToken *only on retries*? This way when we run the original commit against a mongos that was restarted (or a new mongos) we'll get an immediate error instead of blocking until the transaction is automatically aborted.

renctan: Currently, mongos will only use the recoveryToken only if it doesn't know who is the coordinator. So it is harmless to attach it on the non-retry cases. Not sure if this will change in the future.

ShaneHarvey: I would argue that it is harmful to include it on the original attempt. When we include it on the original attempt and the mongos does not know the coordinator, the mongos will block for 60 seconds and then respond with a transient TransactionAborted error. 

If we do not include it on the original attempt and the mongos does not know the coordinator, the mongos will return a transient error *immediately*. 

I think the second case is much more desirable especially when we consider that the next operations on the session will very likely be retrying the entire transaction (as in the `withTransaction` API).

ShaneHarvey: Hmm, I just tested this and found that my assumption that "mongos will return a transient error _immediately_" is not true. Mongos does return an error immediately but the error does not include the `TransientTransactionError` label. The error is:
```
{'ok': 0.0, 'errmsg': 'cannot commit with no participants', 'code': 50940, 'codeName': 'Location50940'}
```

I suppose that the mongos cannot distinguish this case from the case where a client sends a *retry* without a recoveryToken so it cannot add the `TransientTransactionError` label. Is that correct?

source/transactions/transactions.rst
ShaneHarvey: Can you suggest a different phrasing? I'm not sure how to say "recover the outcome of a sharded transaction" more clearly.

Does the footnote below clarify this sufficiently? `In 4.2, a new mongos waits for the *outcome* of the transaction but...`

source/transactions/transactions.rst
ShaneHarvey: I updated to say: `Every successful (``ok:1``) command response in a sharded transaction includes a ``recoveryToken`` field. ...`

Should drivers also look for a `recoveryToken` within `ok:0` responses? I could add:
`
Every successful (``ok:1``) command response in a sharded transaction includes a ``recoveryToken`` field. Any unsuccessful (``ok:0``) command responses in a sharded transaction MAY include a ``recoveryToken`` field. ...`

source/transactions/transactions.rst
ShaneHarvey: Given the server behavior here I'm okay with leaving the spec as is (drivers send recoveryToken on all commit attempts). I don't think the behavior of either case is ideal but at least after blocking for 60 seconds the transaction can be retried automatically by `withTransaction`.

I'll open a new server ticket to discuss this problem because I'm not sure drivers can workaround it.

source/transactions/transactions.rst
renctan: For now, it is not needed for drivers to look for recoveryToken since it cannot commit the transaction anyway. The suggested text looks good to me.

source/transactions/transactions.rst
renctan: Perhaps just say "enables the driver to recover the outcome of a commitTransaction.

source/transactions/transactions.rst
renctan: Sounds good to me

source/transactions/transactions.rst
ShaneHarvey: Perfect, thanks.

source/transactions/transactions.rst
ShaneHarvey: I think the current wording is fine because the sharded txn design doc uses the same language, see "Best effort to **recover a transaction’s outcome** via secondary router"

I just reordered it slightly to match.

source/transactions/transactions.rst
ShaneHarvey: Great, I also think the driver behavior here is correct.  I think the issue is that this section from the server design doc may not have been implemented yet:
> Doing so will never initiate committing the transaction. **Instead, the recovery token in the request will be used to try to abort the transaction if a decision to commit has not already been made**, otherwise to recover the transaction's outcome.

The 60 second waiting period should be removed once the server implements this abort logic.

renctan: Hm... I think I understand the source of the confusion for about recoveryToken for abortTransasction. The recoveryToken was meant to be used only for commit from the start. What the sentence is saying is that recovery will try to abort the transaction if decision has not been made. The current implementation is it will always try to join the active coordinator. Only if the coordinator no longer exists (because the commit finished/or was aborted) will it try to call abort (just to find out the result, if the response returned error because "transaction already committed", then we know it was committed).

ShaneHarvey: Let's continue this conversation on https://jira.mongodb.org/browse/SERVER-39726?

ShaneHarvey:
I think the current version of the spec is complete. There are still a few cases to follow-up on which _might_ require driver changes ([SERVER-39726](https://jira.mongodb.org/browse/SERVER-39726), [SERVER-39349](https://jira.mongodb.org/browse/SERVER-39349), [SERVER-39692](https://jira.mongodb.org/browse/SERVER-39692)) but I don't think they should hold up this review

ShaneHarvey:
Update *now* all the spec tests are passing in my POC here (https://github.com/ShaneHarvey/mongo-python-driver/tree/PYTHON-1684).

source/transactions/transactions.rst
ShaneHarvey: In response to [this conversation with Esha](https://github.com/mongodb/specifications/pull/471#discussion_r258627327), I've updated this section so that drivers remain pinned after successful commits.

source/transactions/transactions.rst
```diff
@@ -770,23 +769,18 @@ Drivers MUST send all commands for a single transaction to the same mongos
...
 
 When to unpin
 ^^^^^^^^^^^^^
 
-Drivers MUST unpin a ClientSession when a command within a transaction
-fails with a TransientTransactionError. Transient errors indicate that the
-transaction in question has already been aborted or that the pinned mongos is
+Drivers MUST unpin a ClientSession when a command within a transaction,
+including commitTransaction and abortTransaction, fails with a
+TransientTransactionError. Transient errors indicate that the transaction
```
jmikola: I think I commented on this earlier, but `commitTransaction` would never fail with a TransientTransactionError. Do you need to specify UnknownTransactionCommitResult here?

ShaneHarvey: The commitTransaction helper never fails with TransientTransactionError but the individual commands run by the helper can. The spec is meant to mandate that when the initial commitTransaction command attempt fails with a TransientTransactionError, drivers must unpin before running any automatic retry attempt.

ShaneHarvey: The behavior is tested here: https://github.com/mongodb/specifications/pull/471/files#diff-55cfe16e56662a302c7ec014fbadadaaR173

ShaneHarvey: I updated the spec to say:
```
Additionally, drivers MUST unpin a ClientSession when any individual
commitTransaction command attempt fails with an UnknownTransactionCommitResult
error label. In cases where the UnknownTransactionCommitResult causes an
automatic retry attempt, drivers MUST unpin the ClientSession before performing
server selection for the retry.
```

source/transactions/tests/README.rst
ShaneHarvey: I'm making this change in https://github.com/10gen/mongo-orchestration/pull/267

source/transactions/tests/pin-mongos.yml
EshaMaharishi: > This ensures that a subsequent abort does not block

Why would the driver send abort after hearing TransientTransactionError?

ShaneHarvey: In the case of a network error (which is labelled a TransientTransactionError) the driver needs to send abort to release the transaction's resources. When the server returns a command error including the "TransientTransactionError" label, I suppose drivers could skip sending abortTransaction because the server has already aborted the transaction. This is an optimization that we considered in 4.0 but rejected for the sake of simplicity.

EshaMaharishi: I'm confused, does the driver label network errors as TransientTransactionError itself?

ShaneHarvey: Yes. The driver adds the TransientTransactionError error label to network errors: https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#transienttransactionerror

> Any command error that includes the "TransientTransactionError" error label in the "errorLabels" field. Any network error or server selection error encountered running any command besides commitTransaction in a transaction. In the case of command errors, the server adds the label; in the case of network errors or server selection errors where the client receives no server reply, the client adds the label.

EshaMaharishi: I see. So if the driver gets a "blip" network error and unpins from the original mongos, the driver may send abort to a different mongos, and that is the motivation for https://jira.mongodb.org/browse/SERVER-39692?

ShaneHarvey: Exactly. Other similar cases are when the original mongos is restarted and looses the in memory transaction state or the original mongos is healthy but there's a network partition preventing any connections from the driver.

EshaMaharishi: Ok, got it. I have no further comments.

PR #472, opened 2019-02-08
split WaitQueueTimeoutError into two separate errors

saghm:
This pull request splits `WaitQueueTimeoutError` into two errors based on whether the checkout
timed out while the thread was waiting to reach the front of the wait queue or due to the max pool size being reached and no connections being checked back in before the timeout.

I'm not sure what the best name for the new error is, so I put in something arbitrary to at least start the bikeshedding off of. I'm also not sure how best to test this, given that timing a `WaitQueueTimeoutError` to occur without also triggering a `MaxPoolSizeTimeoutError` (or whatever the final name is) requires some precise threading, so maybe this would best be tested with a prose test.

### DIFF
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -645,6 +645,15 @@ a manner idiomatic to the Driver and Language.
 .. code:: typescript
 
     /**
+     *  Thrown when a driver times out when attempting to check out
+     *  a Connection from a Pool
+     */
+    interface MaxPoolSizeTimeoutError {
+      message: 'Unable to check out a connection within the timeout due to the max pool size';
+      address: <pool address>;
+    }
+
+    /**
      *  Thrown when the driver attempts to check out a
      *  Connection from a closed Connection Pool
      */
@@ -658,7 +667,7 @@ a manner idiomatic to the Driver and Language.
      *  a Connection from a Pool
      */
     interface WaitQueueTimeoutError {
-      message: 'Timed out while checking out a connection from connection pool';
+      message: 'Timed out while waiting to check out a connection from the pool';
       address: <pool address>;
     }
 


ShaneHarvey:
Is it useful to make a distinction between these two cases? Consider two threads blocking in the wait queue with all sockets in use. After this change the first thread may raise a `MaxPoolSizeTimeoutError` while the second thread may raise either a `MaxPoolSizeTimeoutError` or a `WaitQueueTimeoutError` (depending on the timing of when the threads enter the queue). It seems like when facing either error the application has two choices: either increase `waitQueueTimeoutMS` or increase `maxPoolSize`.

p-mongo:
In this case the majority of the exceptions seen by the application should be max pool size timeout error, suggesting that the pool size needs to be increased.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -644,6 +644,15 @@ a manner idiomatic to the Driver and Language.
 
 .. code:: typescript
 
+    /**
+     *  Thrown when a driver times out when attempting to check out
+     *  a Connection from a Pool
+     */
+    interface MaxPoolSizeTimeoutError {
```
p-mongo: Maybe `PoolAtMaxSizeError` ?

daprahamian:
If we're going to do this, I'd prefer it to be an info item on the `WaitQueueTimeoutError`:

```ts
interface WaitQueueTimeoutError {
  message: 'Timed out while checking out a connection from connection pool';
  address: <pool address>;
  reason: 'maxPoolSizeReached'|'neverReachedFrontOfQueue';
}
```

p-mongo:
Why is this exception called a wait queue error when it may be produced by something other than the wait queue?

saghm:
@ShaneHarvey The reason for this distinction is that in some cases of the existing error, there could potentially be a really easy fix for the user by increasing the `maxPoolSize` (e.g. if the connection at the front of the wait queue is often blocked on other connections being checked), but in other cases, it might require restructuring of the user's application (e.g. if the maxPoolSize is extremely large but so many threads are attempting to acquire connections at once that even though the maxPoolSize is never reached, there are so many threads ahead in the wait queue that it takes longer than `waitQueueTimeoutMS` for a given thread to reach the front of the queue).

The complication with making this a reason on `WaitQueueTimeoutError` is that in the former case described above, the issue isn't really due to being stuck in the way queue but instead about `maxPoolsize`, so the error name would be a bit misleading. On the other hand, the name of the timeout is still `waitQueueTimeoutMS`, and I'd guess that there is little appetite to change this since a number of drivers already use this name.

ShaneHarvey:
My objection is that these two errors seem to represent the same case: a thread timed out in the wait queue. I don't think there is value in making a distinction based on where in the queue the thread timed out. If an app sees  `WaitQueueTimeoutError` it needs to either increase maxPoolSize, increase waitQueueTimeoutMS, or restructure the app. If an app sees `MaxPoolSizeTimeoutError` it also needs to either increase maxPoolSize, increase waitQueueTimeoutMS, or restructure the app.

My other objection is that it may be prohibitively difficult to implement this feature. For example, in python the "wait queue" is just a semaphore. There's no way to determine the position of the thread when a timeout happens.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -644,6 +644,15 @@ a manner idiomatic to the Driver and Language.
 
 .. code:: typescript
 
+    /**
+     *  Thrown when a driver times out when attempting to check out
```
jyemin: The comment is identical to the one for the existing error.

saghm:
It seems like the consensus is split here, but leaning towards not making this change. I'm fine with closing the PR given that existing spec behavior was already approved by all parties and this isn't widely accepted.

jyemin:
@saghm please either revive this PR or abandon it (i.e. close without merging)

PR #473, opened 2019-02-08
WRITING-3230: Add Retryable Reads spec and tests

vincentkam:
- Updated spec and tests for mapReduce per final review feedback

Updated spec sections:
 - Minimum server version (changed to 3.6 and added reasoning)
 - Unsupported Operations (added mapReduce information)
 - Documentation (added mapReduce information)
 - Q&A: I've added a new Q&A about why 4.2 isn't required.
 - Design Rationale:  I've updated the Rejected Designs section 
 - Default changed to true (and test added to test this)


source/retryable-reads/retryable-reads.rst
```diff
@@ -0,0 +1,644 @@
...
+
+3. Any future changes to the the applicable parts of `retryable writes
+   specification
+   <https://github.com/mongodb/specifications/blob/master/source/retryable-writes/>`__
+   may also need to be reflected in the retryable reads specification, and vice
+   versa.
+
+4. We may revisit the decision not retry ``Cursor.getMore()`` (see `Q&A`_).
+
+5. To improve performance on servers without “Early Failure on Socket
```
hanumantmk: it's not crazy, but probably not what you want.

1. you'll need to burn the session you've killed.  I'd have to think further, but I worry that after returning from kill, you may still have bits of it running (if there are any errors).
2. it will take a while.  a kill has to talk to every non-config mongod in the cluster (i.e. all primaries, all secondaries, etc, of every shard).  In addition, in order to protect the system against getting overloaded with these requests, every server allows no more than one killsession operation at a time.  Operations that attempt to killsessions while a killsession is running are batched together and run simultaneously after the current one finishes

If the point of retryable reads is partially to allow you to retarget faster, that may not get you what you want

vincentkam: Thanks Jason for chiming in :) Given the uncertainty, I can remove this text and the potential performance problem can be re-addressed in future only  if it's needed. @ShaneHarvey , are you comfortable with this?

source/retryable-reads/retryable-reads.rst
```diff
@@ -0,0 +1,644 @@
...
+name but MAY deviate to comply with their existing conventions.
+
+Requirements for Retryable Reads
+--------------------------------
+
+Supported Server Versions
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Drivers SHOULD verify server eligibility by ensuring that ``maxWireVersion`` is
+at least 6 because retryable reads require a MongoDB 36 standalone, replica set
```
ShaneHarvey: require a MongoDB 36 -> require at least a MongoDB 3.6

source/retryable-reads/retryable-reads.rst
```diff
@@ -0,0 +1,644 @@
...
+wire protocol message.
+
+Why isn't MongoDB 4.2 required?
+-------------------------------
+
+MongoDB 4.2 was initially considered as a requirement for retryable
+reads because MongoDB 4.2 implements support for "Early Failure on
+Socket Disconnect," changing the the semantics of socket disconnect to
+prevent ops from doing work that no client is interested in. This
+prevents applications from seeing degraded performance when a read is
```
ShaneHarvey: "seeing degraded performance when a read is" -> "seeing degraded performance when an expensive read is"

source/retryable-reads/retryable-reads.rst
```diff
@@ -0,0 +1,644 @@
...
+retried. Upon further discussion, we decided that should not be required
+to retry reads because any customers experiencing degraded performance
+can simply disable ``retryableReads``.
```
ShaneHarvey: Can we add that the resilience benefit of retryable reads far outweighs the minor risk of degraded performance?

source/retryable-reads/retryable-reads.rst
ShaneHarvey: I would like to see this idea be incorporated into the "Design Rationale" section. Maybe under a "rejected designs" section. This is useful information that we will want to refer back to when considering the client side operation timeout feature.

source/retryable-reads/retryable-reads.rst
```diff
@@ -0,0 +1,644 @@
...
+None.
+
+Future work
+===========
+
+1. A later specification may allow operations (including read) to be retried any
+   number of times during a singular timeout period.
+
+2. A later specification may have `drivers default to enabling retryable writes
+   <https://jira.mongodb.org/browse/WRITING-3115>`__, in which case, drivers
```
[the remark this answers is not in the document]
vincentkam: Thanks, fixed!

source/retryable-reads/retryable-reads.rst
```diff
@@ -0,0 +1,644 @@
+===============
+Retryable Reads
+===============
+
+:Spec Title: Retryable Reads
+:Spec Version: 1.0
+:Author: Vincent Kam 
+:Lead: \Bernie Hackett
```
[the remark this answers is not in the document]
vincentkam: Fixed

source/retryable-reads/retryable-reads.rst
```diff
@@ -0,0 +1,644 @@
...
+--------------------------
+
+This specification introduces the following client-level configuration option.
+
+retryReads
+~~~~~~~~~~
+
+This boolean option determines whether retryable behavior will be applied to all
+read operations executed within the MongoClient. This option MUST default to
+false [1]_, which implies no change in read behavior. `As with retryable writes
```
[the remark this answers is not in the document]
vincentkam: Ah sorry for the confusion: perhaps I'm misremembering, but I thought the intention was to have the retryable-writes and retryable-reads change at the same time (hence why the changing of the default is noted under "future work")?

source/retryable-reads/retryable-reads.rst
```diff
@@ -0,0 +1,644 @@
...
+``retryReads``. For any other names in the spec, drivers SHOULD use the defined
+name but MAY deviate to comply with their existing conventions.
+
+Requirements for Retryable Reads
+--------------------------------
+
+Supported Server Versions
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Drivers SHOULD verify server eligibility by ensuring that ``maxWireVersion`` is
```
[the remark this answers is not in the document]
vincentkam: The .NET driver hasn't migrated to using wire versions yet and still uses server version when checking features, although I suppose we could start using it for retry-reads. Changing it to MUST!

source/retryable-reads/retryable-reads.rst
vincentkam: Fixed!

source/retryable-reads/retryable-reads.rst
```diff
@@ -0,0 +1,644 @@
...
+at least 6 because retryable reads require a MongoDB 36 standalone, replica set
+or shard cluster, MongoDB 3.6 server wire version is 6 as defined in the `Server
+Wire version and Feature List specification
+<https://github.com/mongodb/specifications/blob/master/source/wireversion-featurelist.rst>`__.
+
+The minimum server version is 3.6 because
+
+1. It gives us version parity with retryable writes
```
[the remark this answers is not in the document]
vincentkam: Fixed!

source/retryable-reads/retryable-reads.rst
vincentkam: I like the clarification. Fixed!

source/retryable-reads/retryable-reads.rst
vincentkam: Added!

source/retryable-reads/retryable-reads.rst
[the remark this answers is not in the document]
jmikola: > I thought the intention was to have the retryable-writes and retryable-reads change at the same time

That was my understanding as well.

source/retryable-reads/retryable-reads.rst
```diff
@@ -0,0 +1,664 @@
...
+and MUST raise the original retryable error
+
+The above requirement can be fulfilled in one of two ways:
+
+1. During the retry attempt, the driver SHOULD recreate the command while
+   adhering to that operation's specification's server/wire version
+   requirements. If an error occurs while recreating the command, then the
+   driver MUST raise the original retryable error.
+
+   a. For example, if the wire version dips from *W*\ :sub:`0` to *W*\ :sub:`1`
```
jmikola: Is this really a sub-bullet of `1.`? As an example, it looks like it should just be a second paragraph.

source/retryable-reads/retryable-reads.rst
```diff
@@ -0,0 +1,664 @@
...
+      should be omitted. If the spec for operation *O* requires the driver to
+      error out if field *F* is defined when talking to a server with wire
+      version *W*\ :sub:`1`, then the driver must error out and raise the
+      original retryable error.
+
+2. Alternatively, if a driver chooses not to recreate the command as described
+   above, then a driver MUST NOT retry if the server/wire version dips after
+   server selection and MUST raise the original retryable error.
+
+   b. For example, if the wire version dips after server selection, the driver
```
jmikola: Same comment as above (this should probably just be a second paragraph). Beyond that, I don't think it'd make sense to start with `b` here because it's the first sub-bullet.

source/retryable-reads/retryable-reads.rst
vincentkam: Ah, I can see that. Fixed!

source/retryable-reads/retryable-reads.rst
vincentkam: Oops, good catch, some formatting weirdness happened there. I've converted the bullet into a second paragraph per your suggestion :)

source/retryable-reads/tests/README.rst
```diff
@@ -25,8 +25,8 @@ versions 4.0 or later.
 
 N.B. The spec specifies 3.6 as the minimum server version: however,
 ``failCommand`` is not supported on 3.6, so for now, testing requires MongoDB
-4.0. Once `DRIVERS-560`_ is resolved, tests will be added to allow testing
-Retryable Reads on MongoDB 3.6.
+4.0. Once `DRIVERS-560`_ is resolved, live failure integration tests will be
+added to test Retryable Reads on MongoDB 3.6.
```
jmikola: This was just cut from scope of [DRIVERS-560](https://jira.mongodb.org/browse/DRIVERS-560). The initial implementation of that spec will likely end up spamming a cluster with writes and ensuring that the sequence can survive one or more failovers.

If you need 3.6 testing for retryable reads, that may need to be added as a prose test.

ShaneHarvey: I would prefer not to add a prose test for 3.6 in this review. Instead I think we should publish the spec without testing 3.6 and implement 3.6 testing as follow on work in https://jira.mongodb.org/browse/SPEC-1218.

source/retryable-reads/retryable-reads.rst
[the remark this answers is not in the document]
ShaneHarvey: @behackett can you confirm or deny whether `retryReads=true` should be the default?

source/retryable-reads/retryable-reads.rst
```diff
@@ -0,0 +1,664 @@
...
+   Disconnect”, we may, we considered using ``killSessions`` to automatically
+   kill the previous attempt before running a retry.  We decided against this
+   because after killing the session, parts of it still may be running if there
+   are any errors.  Additionally, killing sessions takes time because a kill has
+   to talk to every non-config ``mongod`` in the cluster (i.e. all the primaries
+   and secondaries of each shard). In addition, in order to protect the system
+   against getting overloaded with these requests, every server allows no more
+   than one killsession operation at a time.  Operations that attempt to
+   killsessions while a killsession is running are batched together and run
+   simultaneously after the current one finishes
```
ShaneHarvey: Missing a period.

source/retryable-reads/retryable-reads.rst
```diff
@@ -0,0 +1,664 @@
...
+to 2n times.
+
+The note above will also apply if an application upgrades to a version of the
+driver where that defaults to enabling retryable reads.
+
+Rejected Designs
+----------------
+
+1. To improve performance on servers without “Early Failure on Socket
+   Disconnect”, we may, we considered using ``killSessions`` to automatically
```
ShaneHarvey: I think there's an extra `we may,` here.

source/retryable-reads/retryable-reads.rst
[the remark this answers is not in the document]
behackett: Just make it true now. We're making retryWrites true shortly. It doesn't have to happen at exactly the same time, just before 4.2.

source/retryable-reads/tests/README.rst
vincentkam: Ah okay, so no reads will be tested during DRIVERS-560, acknowledged. That being the case, I would also prefer to implement the 3.6 test as part of SPEC-1218,  adapting the tests that were developed as part of DRIVERS-560,

source/retryable-reads/retryable-reads.rst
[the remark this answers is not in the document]
vincentkam: Acknowledged, default changed!

source/retryable-reads/retryable-reads.rst
vincentkam: Thanks, fixed!

source/retryable-reads/retryable-reads.rst
vincentkam: Fixed!

source/retryable-reads/retryable-reads.rst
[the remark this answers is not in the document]
ShaneHarvey: Can you also add one test that a read is retried on a client without specifying the `retryReads` option? This ensures drivers set the default to true. Currently the test suite passes even with a driver that has retryReads default to false.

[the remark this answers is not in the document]
vincentkam: Ah good idea! Okay, I've added an additional test to find.yml/json!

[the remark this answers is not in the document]
vincentkam: Per offline discussion: we're going to make all the tests use the default when possible, except for one test which uses an explicit "true".

source/retryable-reads/tests/changeStreams-client.watch.yml
```diff
@@ -4,14 +4,11 @@ collection_name: &collection_name "coll"
 data:
     - {_id: 1, x: 11}
 
-minServerVersion: '4.1.7' # 4.2
+minServerVersion: '4.0'
```
ShaneHarvey: Missing topology (I think only in this file).

source/retryable-reads/tests/changeStreams-client.watch-serverErrors.yml
```diff
@@ -6,15 +6,12 @@ data:
...
+topology: [replicaset, sharded]
 
 tests:
     -
-        description: "client.watch succeeds after InterruptedAtShutdown"
-        topology:  [replicaset, sharded]
-        clientOptions:
-            retryReads: true
-        failPoint: &failCommand_failPoint
+        description: "client.watch succeeds after InterruptedAtShutdown"        failPoint: &failCommand_failPoint
```
ShaneHarvey: The format got messed up here. Notice how `description` and `failPoint` are on the same line. I think it's just in this file.

source/retryable-reads/tests/changeStreams-client.watch.yml
vincentkam: Sorry, fixed! (Some weird shenanigans with my editor)

source/retryable-reads/tests/changeStreams-client.watch-serverErrors.yml
vincentkam: Thanks, fixed!

ShaneHarvey: Can you fix the same issue with the rest of the tests in this file?

vincentkam: Gah, I'm really sorry! Fixed for real this time.

vincentkam:
Rebased, squashed and pushed! 😸

PR #475, opened 2019-02-13
Spec 1153: add read concern support to aggregation $out

prashantmital:
Supersedes #439.

### DIFF
--- a/source/crud/tests/v2/aggregate-out-readConcern.yml
+++ b/source/crud/tests/v2/aggregate-out-readConcern.yml
@@ -1,0 +1,41 @@
+data:
+    - {_id: 1, x: 11}
+    - {_id: 2, x: 22}
+    - {_id: 3, x: 33}
+
+minServerVersion: '4.1'
+
+collection_name: &collection_name 'testaggcollection'
+
+tests:
+  -
+    description: "readConcern Majority with out stage"
+    operations:
+      -
+        object: collection
+        name: aggregate
+        collectionOptions:
+          readConcern:
+            level: "majority"
+        arguments:
+          pipeline:
+            - $sort: {x : 1}
+            - $match:
+                _id: {$gt: 1}
+            - $out: &output_collection "other_test_collection"
+    expectations:
+      -
+        command_started_event:
+          command:
+            aggregate: *collection_name
+            pipeline:
+              - $sort: { x: 1 }
+              - $match: { _id: { $gt: 1 } }
+              - $out: *output_collection "other_test_collection"
+            readConcern: { level: "majority" }
+    outcome:
+      collection:
+        name: *output_collection
+        data:
+          - {_id: 2, x: 22}
+          - {_id: 3, x: 33}
--- a/source/crud/tests/v2/aggregate-out-readConcern.yml
+++ b/source/crud/tests/v2/aggregate-out-readConcern.yml
@@ -1,11 +1,11 @@
 data:
-    - {_id: 1, x: 11}
-    - {_id: 2, x: 22}
-    - {_id: 3, x: 33}
+    - { _id: 1, x: 11 }
+    - { _id: 2, x: 22 }
+    - { _id: 3, x: 33 }
 
 minServerVersion: '4.1'
 
-collection_name: &collection_name 'testaggcollection'
+collection_name: &collection_name 'test_aggregate_out_readconcern'
 
 tests:
   -
@@ -15,13 +15,11 @@ tests:
         object: collection
         name: aggregate
         collectionOptions:
-          readConcern:
-            level: "majority"
+          readConcern: { level: "majority" }
         arguments:
           pipeline:
-            - $sort: {x : 1}
-            - $match:
-                _id: {$gt: 1}
+            - $sort: { x : 1 }
+            - $match: { _id: { $gt: 1 } }
             - $out: &output_collection "other_test_collection"
     expectations:
       -
@@ -37,8 +35,8 @@ tests:
       collection:
         name: *output_collection
         data:
-          - {_id: 2, x: 22}
-          - {_id: 3, x: 33}
+          - { _id: 2, x: 22 }
+          - { _id: 3, x: 33 }
   -
     description: "readConcern local with out stage"
     operations:
@@ -46,14 +44,12 @@ tests:
         object: collection
         name: aggregate
         collectionOptions:
-          readConcern:
-            level: "local"
+          readConcern: { level: "local" }
         arguments:
           pipeline:
-            - $sort: {x : 1}
-            - $match:
-                _id: {$gt: 1}
-            - $out: &output_collection "other_test_collection"
+            - $sort: { x : 1 }
+            - $match: { _id: { $gt: 1 } }
+            - $out: *output_collection
     expectations:
       -
         command_started_event:
@@ -68,8 +64,8 @@ tests:
       collection:
         name: *output_collection
         data:
-          - {_id: 2, x: 22}
-          - {_id: 3, x: 33}
+          - { _id: 2, x: 22 }
+          - { _id: 3, x: 33 }
   -
     description: "readConcern available with out stage"
     operations:
@@ -77,14 +73,12 @@ tests:
         object: collection
         name: aggregate
         collectionOptions:
-          readConcern:
-            level: "available"
+          readConcern: { level: "available" }
         arguments:
           pipeline:
             - $sort: {x : 1}
-            - $match:
-                _id: {$gt: 1}
-            - $out: &output_collection "other_test_collection"
+            - $match: { _id: { $gt: 1 } }
+            - $out: *output_collection
     expectations:
       -
         command_started_event:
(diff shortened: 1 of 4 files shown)


source/crud/tests/v2/aggregate-out-readConcern.yml
```diff
@@ -0,0 +1,155 @@
...
+    expectations:
+      -
+        command_started_event:
+          command:
+            aggregate: *collection_name
+            pipeline:
+              - $sort: { x: 1 }
+              - $match: { _id: { $gt: 1 } }
+              - $out: *output_collection
+            readConcern: { level: "!invalid123" }
```
jmikola: Can we get a newline at end of file?

source/crud/tests/v2/aggregate-out-readConcern.yml
```diff
@@ -0,0 +1,155 @@
...
+        name: aggregate
+        collectionOptions:
+          readConcern:
+            level: "local"
+        arguments:
+          pipeline:
+            - $sort: {x : 1}
+            - $match:
+                _id: {$gt: 1}
+            - $out: &output_collection "other_test_collection"
```
jmikola: It shouldn't be necessary to redefine this anchor. I think you can just use `*output_collection` here.

source/crud/tests/v2/aggregate-out-readConcern.yml
```diff
@@ -0,0 +1,155 @@
...
+        object: collection
+        name: aggregate
+        collectionOptions:
+          readConcern:
+            level: "majority"
+        arguments:
+          pipeline:
+            - $sort: {x : 1}
+            - $match:
+                _id: {$gt: 1}
```
jmikola: Small suggestion to inline these as you did in the APM expectations below (`{ _id: { $gt: 1 } }`).

source/read-write-concern/read-write-concern.rst
```diff
@@ -96,6 +96,13 @@ The read concern option is available for the following operations:
 
 ``aggregate`` command with $out and ``mapReduce`` command with "out" set to anything other than "inline" do not support readConcern.
 
+Starting in MongoDB 4.2, an ``aggregate`` command with ``$out`` supports a ``readConcern``; however, it does not support
+the "linearizable" level (attempting to do so will result in a server error). Server versions before 4.2 do not support
+a ``readConcern`` at all for ``aggregate`` commands with ``$out``.
```
jmikola: I noticed that "snapshot" wasn't mentioned or tested. Do we know if that is supported if the `aggregate` is executed within a transaction? Since the transaction's read concern is specified on the initial transaction command, I assume a test would need to use multiple operations:

 * Start transaction with RC level "snapshot"
 * Execute `aggregate` with `$out` and _do not_ provide a RC option
 * Expect `aggregate` to experience an error
 * Observe that outgoing `aggregate` command included RC "snapshot", `startTransaction`, etc.

Note: I'm assuming that "snapshot" is not supported since the motivator for this spec change was [SERVER-20445](https://jira.mongodb.org/browse/SERVER-20445), which introduced "majority" read concern support. Before then, only "local" and "available" were supported. Therefore, I would still expect other RC levels to be unsupported.

ShaneHarvey: `$out` is not supported within a transaction so you'll get an error but it won't be caused by the read concern.

prashantmital: On MongoDB 4.1.8 here is the error generated: 
```
repl0:PRIMARY> db.testaggcollection.aggregate([{$sort: {x: 1}}, {$match: {_id: {$gt: 1}}}, {$out: "other_test_collection"}], {readConcern: {level: "snapshot"}})
2019-02-13T11:35:31.027-0800 E QUERY    [js] Error: command failed: {
	"operationTime" : Timestamp(1550086525, 1),
	"ok" : 0,
	"errmsg" : "readConcern level snapshot is only valid for the first transaction operation",
	"code" : 72,
	"codeName" : "InvalidOptions",
	"$clusterTime" : {
		"clusterTime" : Timestamp(1550086525, 1),
		"signature" : {
			"keyId" : NumberLong(0)
		}
	}
} : aggregate failed :
_getErrorWithCode@src/mongo/shell/utils.js:25:13
doassert@src/mongo/shell/assert.js:18:14
_assertCommandWorked@src/mongo/shell/assert.js:584:17
assert.commandWorked@src/mongo/shell/assert.js:674:16
DB.prototype._runAggregate@src/mongo/shell/db.js:260:9
DBCollection.prototype.aggregate@src/mongo/shell/collection.js:987:12
@(shell):1:1
```

@jmikola do you think I should explicitly mention that "snapshot" `readConcern` cannot be used with an aggregation pipeline containing the `$out` stage due to the limitation with transactions that Shane mentioned?

jmikola: > `$out` is not supported within a transaction so you'll get an error but it won't be caused by the read concern.

Ah, didn't realize that. Thanks!

> do you think I should explicitly mention that "snapshot" `readConcern` cannot be used with an aggregation pipeline containing the `$out` stage due to the limitation with transactions that Shane mentioned?

Rather than attempt to document what _isn't_ supported, I think it's easier to document what _is_ supported.

I think not originally doing so is how we ended up with the original spec language that this PR is changing. Previously, the spec said that `$out` did not support _any_ read concern. This might have been an over-simplification, because  "local" is essentially the default behavior and "available" (introduced a bit later in 3.6) behaves the same as "local" for unsharded collections. At the time, "majority" was probably the only read concern level with a functional difference and since it was not supported, the spec wrote off all read concerns.

Looking at your tests, you expect "local", "available", and "majority" to succeed, so I propose we explicitly list supported read concerns. This will help future-proof the spec, since we don't know what may be added down the line. We can always add a new supported level at a later time.

Also, saying that "snapshot" is unsupported might be accurate but a bit odd, since we know it simply cannot be used by virtue of `$out` not supporting transactions (and "snapshot" only relates to transactions). I think that's another reason why it'd preferable to just list the levels we know _are_ supported.

prashantmital:
The PyMongo implementation of this spec change is now ready to review: https://github.com/mongodb/mongo-python-driver/pull/400

source/crud/tests/v2/aggregate-out-readConcern.yml
```diff
@@ -0,0 +1,145 @@
...
+tests:
+  -
+    description: "readConcern majority with out stage"
+    operations:
+      -
+        object: collection
+        name: aggregate
+        collectionOptions:
+          readConcern: { level: "majority" }
+        arguments:
```
jmikola: Would it make sense to `&arguments` here instead of just `&output_collection`? I think the pipeline is the same for each test.

source/crud/tests/v2/aggregate-out-readConcern.yml
```diff
@@ -0,0 +1,145 @@
...
+          pipeline:
+            - $sort: { x : 1 }
+            - $match: { _id: { $gt: 1 } }
+            - $out: &output_collection "other_test_collection"
+    expectations:
+      -
+        command_started_event:
+          command:
+            aggregate: *collection_name
+            pipeline:
```
jmikola: Likewise, this can also use `&pipeline` to reduce duplication in future assertions (if you end up creating an anchor for `&arguments` above).

source/read-write-concern/read-write-concern.rst
```diff
@@ -94,7 +94,13 @@ The read concern option is available for the following operations:
 - ``geoNear`` command
 - ``geoSearch`` command
 
-``aggregate`` command with $out and ``mapReduce`` command with "out" set to anything other than "inline" do not support readConcern.
+Starting in MongoDB 4.2, an ``aggregate`` command with ``$out`` supports a ``readConcern``; however, it does not support
+the "linearizable" level (attempting to do so will result in a server error).
+
+Server versions before 4.2 do not support a ``readConcern`` at all for ``aggregate`` commands with ``$out``.
```
jmikola: Was this recently changed? IIRC from the other comments, "local" and "available" worked just fine in earlier versions. 4.2 just adds "majority" support.

source/read-write-concern/read-write-concern.rst
prashantmital: Thanks for laying that out so nicely! I'll modify the text here to explicitly mention the supported readConcerns.

source/read-write-concern/read-write-concern.rst
prashantmital: Indeed... But this was the language that had been previously used and I just moved where it occurred. I'll correct it.

source/crud/tests/v2/aggregate-out-readConcern.yml
```diff
@@ -26,15 +26,15 @@ tests:
...
               - $sort: { x: 1 }
               - $match: { _id: { $gt: 1 } }
-              - $out: *output_collection
+              - $out: "other_test_collection"
             readConcern: { level: "majority" }
     outcome:
       collection:
         name: *output_collection
-        data:
+        data: &data
```
jmikola: Would it make sense to define an `&outcome` anchor instead of this? I just noticed that the `outcome` blocks below are all the same.

PR #478, opened 2019-02-13
SPEC-1208: withTransaction() should retry commits after a wtimeout

jmikola:
https://jira.mongodb.org/browse/SPEC-1208

### DIFF
--- a/source/transactions-convenient-api/tests/commit-writeconcernerror.yml
+++ b/source/transactions-convenient-api/tests/commit-writeconcernerror.yml
@@ -19,10 +19,10 @@ tests:
             errmsg: "waiting for replication timed out"
             errInfo: { wtimeout: true }
     operations:
-      -
+      - &operation
         name: withTransaction
         object: session0
-        arguments: &arguments
+        arguments:
           callback:
             operations:
               -
@@ -33,10 +33,6 @@ tests:
                   document: { _id: 1 }
                 result:
                   insertedId: 1
-        result:
-          errorCodeName: WriteConcernFailed
-          errorLabelsContain: ["UnknownTransactionCommitResult"]
-          errorLabelsOmit: ["TransientTransactionError"]
     expectations: &expectations_with_retries
       -
         command_started_event:
@@ -96,7 +92,7 @@ tests:
           command_name: commitTransaction
           database_name: admin
     # The write operation is still applied despite the write concern error
-    outcome:
+    outcome: &outcome
       collection:
         data:
           - { _id: 1 }
@@ -118,19 +114,9 @@ tests:
             codeName: WriteConcernFailed
             errmsg: "multiple errors reported"
     operations:
-      -
-        name: withTransaction
-        object: session0
-        arguments: *arguments
-        result:
-          errorCodeName: WriteConcernFailed
-          errorLabelsContain: ["UnknownTransactionCommitResult"]
-          errorLabelsOmit: ["TransientTransactionError"]
+      - *operation
     expectations: *expectations_with_retries
-    outcome:
-      collection:
-        data:
-          - { _id: 1 }
+    outcome: *outcome
   -
     description: commitTransaction is not retried after UnknownReplWriteConcern error
     failPoint:
@@ -143,10 +129,7 @@ tests:
             codeName: UnknownReplWriteConcern
             errmsg: "No write concern mode named 'foo' found in replica set configuration"
     operations:
-      -
-        name: withTransaction
-        object: session0
-        arguments: *arguments
+      - <<: *operation
         result:
           errorCodeName: UnknownReplWriteConcern
           errorLabelsOmit: ["TransientTransactionError", "UnknownTransactionCommitResult"]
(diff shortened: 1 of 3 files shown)


source/transactions-convenient-api/tests/commit-writeconcernerror.yml
```diff
@@ -35,69 +35,9 @@ tests:
...
-            code: 64
-            codeName: WriteConcernFailed
-            errmsg: "multiple errors reported"
-    operations:
-      -
-        name: withTransaction
-        object: session0
-        arguments: *arguments
-    expectations:
+    expectations: &expectations_with_retries
```
jmikola: These expectations are the same as those for the first "WriteConcernFailed timeout error" test, so I created an anchor in the first test and used an alias here. The only difference between both tests is the fail point configuration.

source/transactions-convenient-api/tests/commit-writeconcernerror.yml
```diff
@@ -35,69 +35,9 @@ tests:
...
-            readConcern: ~
-            startTransaction: ~
-            writeConcern: ~
-          command_name: commitTransaction
-          database_name: admin
-    # The write operation is still applied despite the write concern error
-    outcome: &outcome
-      collection:
-        data:
-          - { _id: 1 }
```
jmikola: I'm not sure how this was passing, unless:

 * The server was actually committing the transaction despite our fail point configuration, and the data was present by the time drivers inspected the outcome collection.
 * Drivers that were running these tests had some error in their test runners, such as not asserting the outcome collection at all if the operation experienced an error

Something to look into either way. I suggest any drivers with working test runners verify that the `outcome` block is still processed regardless of whether the operation(s) succeeds or fails.

source/transactions-convenient-api/tests/commit-writeconcernerror.yml
```diff
@@ -208,7 +180,10 @@ tests:
             writeConcern: ~
           command_name: commitTransaction
           database_name: admin
-    outcome: *outcome
+    outcome:
```
jmikola: As with my comment above, this `outcome` assertion should never have passed.

The last two tests in this file for error codes UnknownReplWriteConcern and UnsatisfiableWriteConcern should _not_ retry `commitTransaction` after the first failure, and I would not expect the server to actually apply the data. Drivers should ensure that they are executing the `outcome` block and asserting that the collection is still empty after the operation(s) executes.

source/transactions-convenient-api/tests/commit-writeconcernerror.yml
```diff
@@ -228,34 +203,7 @@ tests:
...
-            txnNumber: { $numberLong: "1" }
-            autocommit: false
-            # omitted fields
-            readConcern: ~
-            startTransaction: ~
-            writeConcern: ~
-          command_name: commitTransaction
-          database_name: admin
-    outcome: *outcome
+    expectations: *expectations_without_retries
```
jmikola: Reusing APM expectations here as well, since the sequence of commands for the UnknownReplWriteConcern and UnsatisfiableWriteConcern tests are identical (much like the first two tests in this file).

source/transactions-convenient-api/transactions-convenient-api.rst
```diff
@@ -384,21 +376,36 @@ their users to design idempotent callbacks.
...
+*except* for write concern timeouts, so as not to exceed the user's original
+intention for ``wtimeout``. The current design of this specification no longer
+excludes write concern timeouts, and simply retries ``commitTransaction`` within
+its timeout period for any errors bearing the "UnknownTransactionCommitResult"
+label.
+
+This change was made in light of the forthcoming Client-side Operations Timeout
+specification (see: `Future Work`_), which we expect will allow the current
+120-second timeout for ``withTransaction`` to be customized and also obviate the
+need for users to specify ``wtimeout``.
```
jmikola: IIRC, one of the main concerns was that it'd be difficult to explain to how the `withTranasction` timeout (be it 120-second default now or client-side timeout later) interacted with `wtimeout` and just stick with a single timeout; however, I didn't want to dive too much into that. I think the bit about client-side operation timeout ultimately making `wtimeout` a thing of the past (for explicit application use at least) is sufficient.

source/transactions-convenient-api/tests/commit-writeconcernerror.yml
jyemin: CC @jstewart-mongo.  Can you check this please?

jstewart-mongo: I have confirmed in our tests that the `outcome` block is processed regardless of whether the operation(s) succeed or fail.

source/transactions-convenient-api/transactions-convenient-api.rst
ShaneHarvey: Thanks for explaining the rationale, makes sense to me.

jstewart-mongo:
In `commit-writeconcernerror.json`, the first two tests expect an exception to be thrown, but the tests succeed without an exception. In the last two tests, the write is actually applied on the server and then the `writeConcernError` is returned as described in SERVER-39346.

jmikola:
@jstewart-mongo: Please take a look at my latest commit, which removes error assertions for WriteConcernFailed(64) and expects all tests using `writeConcernError` with the fail point to actually apply writes to the collection.

jstewart-mongo:
@jmikola All updated tests in your latest commit pass in my branch.

jmikola:
@jyemin: Can we defer to @jstewart-mongo for an LGTM-to-merge here?

PR #481, opened 2019-02-14
Updating cmap spec tests with waitForEvent

daprahamian:
This is the solution that I arrived at to address the race conditions we were seeing: I added a `waitForEvent` command which operates as follows:

- Every thread keeps track of what events have occurred in a queue*
- When a `waitForEvent` command happens on a thread, the thread dequeues events until the necessary event is found
- If the event was not in the queue, the thread blocks until the event occurs.

Does this sound like a reasonable solution? If so, I'll update the README to include this command.

### DIFF
--- a/source/connection-monitoring-and-pooling/tests/README.rst
+++ b/source/connection-monitoring-and-pooling/tests/README.rst
@@ -64,9 +64,14 @@ Valid Unit Test Operations are the following:
 
   - ``ms``: The number of milliseconds to sleep the current thread for
 
-- ``waitFor(target)``: wait for thread ``target`` to finish executing. Propagate any errors to the main thread.
+- ``waitForThread(target)``: wait for thread ``target`` to finish executing. Propagate any errors to the main thread.
 
   - ``target``: The name of the thread to wait for.
+
+- ``waitForEvent(event, count)``: block the current thread until ``event`` has occurred ``count`` times
+
+  - ``event``: The name of the event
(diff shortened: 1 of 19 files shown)


source/connection-monitoring-and-pooling/tests/README.rst
```diff
@@ -64,10 +64,15 @@ Valid Unit Test Operations are the following:
 
   - ``ms``: The number of milliseconds to sleep the current thread for
 
-- ``waitFor(target)``: wait for thread ``target`` to finish executing. Propagate any errors to the main thread.
+- ``waitForThread(target)``: wait for thread ``target`` to finish executing. Propagate any errors to the main thread.
 
   - ``target``: The name of the thread to wait for.
 
+- ``waitForEvent(event, count)``: block the current thread until ``event`` has occurred ``count`` times
```
saghm: Should we specify that drivers should use some sort of timeout here and fail otherwise (to avoid infinite hangs e.g. due to an unfair wait queue being tested in `wait-queue-fairness.yml`)?

daprahamian: Are there any drivers that implement their test runners without some sort of timeout? I had taken that as a given.

saghm: Not sure; I mostly meant that we should add a note here that a timeout is generally useful, not actually adding a field for a timeout or anything.

PR #486, opened 2019-02-15
SPEC-1087 MUST raise error for unack'd writes with explicit session

ShaneHarvey:
https://jira.mongodb.org/browse/SPEC-1087

[SPEC-1019](https://jira.mongodb.org/browse/SPEC-1019) prohibited unack'd writes with explicit sessions but used the word `SHOULD`. This change makes it a `MUST`. I think is just a clarification of what all (or most) drivers already do.

I also fixed the order of the changelog.

### DIFF
--- a/source/sessions/driver-sessions.rst
+++ b/source/sessions/driver-sessions.rst
@@ -3,7 +3,7 @@ Driver Sessions Specification
 =============================
 
 :Spec Title: Driver Sessions Specification (See the registry of specs)
-:Spec Version: 1.5.0
+:Spec Version: 1.5.1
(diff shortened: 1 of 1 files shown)


ShaneHarvey:
Bernie just reminded me `SHOULD`'s  definition in https://www.ietf.org/rfc/rfc2119.txt:

> SHOULD   This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course.

I'm to close this because changing to MUST doesn't make sense in this case.

PR #488, opened 2019-02-16
Fix order of changelog in sessions spec

ShaneHarvey:


jmikola:
Side note: I'm not sure what purpose the trailing colons serve, as this is rendered in a table. Probably outside the scope of this PR, though.

ShaneHarvey:
The trailing colons mean this is a "[field list](http://docutils.sourceforge.net/docs/user/rst/quickref.html#field-lists)" so I think it's intentional. Github just happens to render field lists as a table which makes the trailing colon look awkward. The heading is also a a field list so I'm not too keen on changing it.

PR #489, opened 2019-02-19
SPEC-1212: Require drivers to document option precedence rules

jmikola:
https://jira.mongodb.org/browse/SPEC-1212

### DIFF
--- a/source/connection-string/connection-string-spec.rst
+++ b/source/connection-string/connection-string-spec.rst
@@ -11,7 +11,7 @@ Connection String Spec
 :Advisors: \A. Jesse Jiryu Davis, Jeremy Mikola, Anna Herlihy
 :Status: Approved
 :Type: Standards
-:Last Modified: February 19, 2018
+:Last Modified: February 19, 2019
 :Version: 1.4.1
 
 .. contents::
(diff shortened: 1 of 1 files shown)


xdg:
I think this is a product mgmt decision.  I have no objection to the wording if that's the decision.

source/connection-string/connection-string-spec.rst
```diff
@@ -11,8 +11,8 @@ Connection String Spec
 :Advisors: \A. Jesse Jiryu Davis, Jeremy Mikola, Anna Herlihy
 :Status: Approved
 :Type: Standards
-:Last Modified: June 04, 2018
-:Version: 1.4
+:Last Modified: February 19, 2018
+:Version: 1.4.1
```
slhommed: 2019?

jmikola:
@slhommed: Can you sign off on this?

slhommed:
I'm not convinced this is the right direction.  Can't the API itself dictate precedence?  Self documenting so to speak.

jmikola:
> Can't the API itself dictate precedence? Self documenting so to speak.

For drivers that have a MongoClient constructor with two arguments (e.g. a URI string and an options dictionary), I'm not sure how the API alone can self-document inheritance. This would require the driver to specify which argument takes priority over the other.

Even in the case of the Go driver, which is probably what instigated this spec ticket, the various methods for applying options to the "client options object" (or whatever it's called) will need to specify whether they will overwrite existing options or not.

slhommed:
> I'm not sure how the API alone can self-document inheritance.

I'd rather not have this cause any API changes so it sounds like docs are the best approach.

PR #493, opened 2019-02-20
test(crud): remove `result` expectation for $out agg test

mbroadst:
Drivers are not required to provide a cursor against the collection
indicated in a `$out`. The presence of a `result` as well as
`collection` fields in this test is inconsistent with the rest of
the crud tests, and makes it difficult for implementations to
share common code for checking expectations without writing very
specific code to check for this isolated case.


prashantmital:
Nice, I was hoping to change this as well! 🎉

PR #495, opened 2019-02-21
test(db-aggregate): only check local operations in $currentOp test

mbroadst:
When running these CRUD tests against a sharded cluster the
`$currentOp` stage will likely return many results, occasionally
too many to return the initiating aggregtion command we expect
to see in the results. Limiting the stage to return only local
operations corrects the race condition.


mbroadst:
💪

PR #497, opened 2019-02-22
SPEC-1168 mongo-orchestration reduces transaction timeout

ShaneHarvey:
This was changed in https://github.com/10gen/mongo-orchestration/pull/267.

### DIFF
--- a/source/transactions/tests/README.rst
+++ b/source/transactions/tests/README.rst
@@ -438,7 +438,9 @@ attempt fails on mongos A and is retried on mongos B, mongos B will block
 waiting for the transaction to complete. However because the initial commit
 attempt failed, the command will only complete after the transaction is
 automatically aborted for exceeding the shard's
-transactionLifetimeLimitSeconds setting.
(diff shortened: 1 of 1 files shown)


ShaneHarvey:
Are there any drivers *not* using mongo-orchestration these days?

mbroadst:
@ShaneHarvey 🖐 node is not

source/transactions/tests/README.rst
```diff
@@ -454,6 +454,9 @@ prematurely timing out any tests' transactions. To decrease the timeout, run::
 
   db.adminCommand( { setParameter: 1, transactionLifetimeLimitSeconds: 3 } )
 
+Note that mongo-orchestration automatically sets this timeout to 3 seconds so
+drivers using mongo-orchestration do not need to run these commands manually.
```
jmikola: Might be wise to specify exactly what version of MO sets this, as some drivers could be using older versions.

ShaneHarvey: Done.

PR #499, opened 2019-02-26
add URI options tests for conflicting TLS options

saghm:


source/uri-options/tests/tls-options.yml
```diff
@@ -71,19 +71,51 @@ tests:
         auth: ~
         options: {}
     -
-        description: "tlsInsecure=true and tlsAllowInvalidCertificates=false warns"
-        uri: "mongodb://example.com/?tlsInsecure=true&tlsAllowInvalidCertificates=false"
+        description: "tlsInsecure and tlsAllowInvalidCertificates both present raises an error"
```
jyemin: Let's be paranoid add two more tests, where the parameter values are false.

saghm: Will do

PR #502, opened 2019-02-26
Add more variants of the TLS-related URI options tests

jyemin:
I realized that some implementations would have order-dependent processing of the TLS options, so added some more variants that test the reverse order.


jyemin:
> Any reason you didn't add cases for `ssl=true&tls=true` not erroring? If we are testing all the possible cases, why leave out the last one?

No special reason.  I added them in the second patch.

PR #503, opened 2019-02-27
SPEC-1166 simplify language for server description list definition

p-mongo:


source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -267,8 +267,8 @@ Fields:
   a primary. Default null.
 * maxElectionId: an ObjectId or null. The largest electionId ever reported by
   a primary. Default null.
-* servers: a set of ServerDescription instances.
-  Default contains one server: "localhost:27017", ServerType Unknown.
+* servers: a set of ServerDescription instances, one for each of the
+  servers in the topology.
```
jmikola: Was the default (i.e. "localhost:27017", ServerType Unknown) in the original text redundant, since this field is not optional and would always be specified?

p-mongo: That default is driver dependent. Some drivers, when given no seeds, use localhost:27017 as the seed. Ruby when given no seeds creates an empty client that doesn't do anything.

behackett: Why does that create an inoperable client? I don't think any other driver works that way.

p-mongo: Since the client has no servers it will never discover any new servers. We make empty clients in the test suite to check options and such, but this is not a useful behavior as far as applications are concerned. If desired I will change the spec to require that an empty seed list results in the driver defaulting to ['localhost:27017'] as the seed list.

xdg: What problem is this PR trying to solve?

p-mongo: The current spec language states that the list of server descriptions has a default. In actuality, the seed list may have a default (and in case of Ruby, does not), but the list of server descriptions does not have a default - it is a function of whatever servers are in the topology.

This PR removes the language stating that the list of server descriptions has a default.

I can alternatively make a PR stating the seed list default should be `['localhost:27017']`.

xdg: I read the spec as meaning that a Topology must never be created empty -- that it must have localhost:27017/Unknown if not provided with something else.  Later the spec says this about initial configuration:

> The user MUST be able to set the initial servers list to a seed list of one or more addresses.

So, to me, the line you are proposing to change is normative, not informative, and should not be changed.

If your concern is that the spec line is vague about types, how about: `
Default contains one ServerDescription with address "localhost:27017", and type "Unknown"`

p-mongo: We are discussing this section: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#topologydescription

Topology description is not user input. A user does not specify topology descriptions, server descriptions, or which servers are in the topoogy - they can only specify seed lists. It does not make sense to say that the list of servers in a topology description has a default value, or is user configurable.

xdg: I'm discussing that section.  Like just about every other field, a default is specified for a new object of that type.  In this case, empty or arbitrary is not allowed.

IanWhalen:
Closing as per discussion in SPEC-1166.  Doesn't mean this is getting rejected, just that it's deferred for now and can be reopened later as we schedule that work.

PR #504, opened 2019-02-28
SPEC-1232: switch countDocuments to use _id: 1 for $group stage

daprahamian:
The $group stage of an aggregation pipeline can use any constant to
achieve the group-for-all effect. Switching to using 1 as a constant,
to avoid test runner issues associated with null values


ShaneHarvey:
Can you open and link to a spec ticket so we can do the normal SPEC->DRIVERS->language ticket jira foo?

daprahamian:
Spec ticket here: https://jira.mongodb.org/browse/SPEC-1232

PR #505, opened 2019-02-28
SPEC-1168 Update comments in transaction unpinning spec test

ShaneHarvey:
~~No functional changes here~~, just trying to improve these confusing and hard to read comments.

Minor change here to actually increase heartbeatFrequencyMS higher than the default 10 seconds.

### DIFF
--- a/source/transactions/tests/pin-mongos.yml
+++ b/source/transactions/tests/pin-mongos.yml
@@ -444,7 +444,7 @@ tests:
       # Increase heartbeatFrequencyMS to avoid the race condition where an in
       # flight heartbeat refreshes the first mongoes' SDAM state in between
       # the insert connection error and the single commit attempt.
-      heartbeatFrequencyMS: 5000
+      heartbeatFrequencyMS: 30000
     operations:
       - name: startTransaction
         object: session0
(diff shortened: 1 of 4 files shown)


source/transactions/tests/pin-mongos.yml
```diff
@@ -443,7 +443,7 @@ tests:
     clientOptions:
       # Increase heartbeatFrequencyMS to avoid the race condition where an in
       # flight heartbeat refreshes the first mongoes' SDAM state in between
-      # the initial commitTransaction and the retry attempt.
+      # the insert connection error and the single commit attempt.
       heartbeatFrequencyMS: 5000
```
jyemin: Isn't the default heartbeat frequency 10,000 ms, in which case this is actually decreasing it?  Maybe we should bump it up to 30,000 ms?

source/transactions/tests/pin-mongos.yml
```diff
@@ -483,10 +483,12 @@ tests:
           errorLabelsContain: ["TransientTransactionError"]
           errorLabelsOmit: ["UnknownTransactionCommitResult"]
       # Session unpins from the first mongos after the insert error and
-      # commitTransaction selects the second mongos and fails. If a driver
-      # mistakenly remains pinned then this commit succeed.
+      # commitTransaction selects the second mongos and fails. If this commit
```
jyemin: Instead of "and fails", say something like "which is unaware of the transaction and therefore fails with NoSuchTransaction error"

source/transactions/tests/pin-mongos.yml
ShaneHarvey: Yeah that's a very good point.

PR #507, opened 2019-03-01
add horizon option

saghm:


jmikola:
Can we get a spec ticket reference on this PR/commit?

behackett:
I'm holding off on reviewing this until the related server design document is approved.

xdg:
I'm involved in the server design review, so when that's done, we can revisit.  The option name is right, but the definition is incorrect.

behackett:
We're not actually implementing this anymore, so I'm going to close this PR. Thanks @saghm .

PR #510, opened 2019-03-01
SPEC-1203: Use runOn for retryable write test reqs

jmikola:
https://jira.mongodb.org/browse/SPEC-1203

This allows retryable writes spec tests to run on mongos.

### DIFF
--- a/source/retryable-writes/tests/updateMany.yml
+++ b/source/retryable-writes/tests/updateMany.yml
@@ -1,12 +1,16 @@
+runOn:
+    -
+        minServerVersion: "3.6"
+        topology: ["replicaset", "sharded"]
+
 data:
     - { _id: 1, x: 11 }
     - { _id: 2, x: 22 }
 
-minServerVersion: '3.6'
-
 tests:
     -
         description: "UpdateMany ignores retryWrites"
+        useMultipleMongoses: true
         operation:
             name: "updateMany"
             arguments:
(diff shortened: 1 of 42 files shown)


source/retryable-writes/tests/updateMany.yml
```diff
@@ -1,12 +1,16 @@
...
+
 data:
     - { _id: 1, x: 11 }
     - { _id: 2, x: 22 }
 
-minServerVersion: '3.6'
-
 tests:
     -
         description: "UpdateMany ignores retryWrites"
```
ShaneHarvey: This is unrelated to this review but shouldn't this file also have a test that enables the fail point and confirms that `UpdateMany` is not retried after a network error?

jmikola: Opened: https://jira.mongodb.org/browse/SPEC-1241

jmikola:
In https://github.com/mongodb/specifications/pull/511, @p-mongo had two suggestions:

 * Make `topology` plural (i.e. `topologies`), since it's value is an array. I'll note that change streams and transactions both use `topology` in their existing formats, but this is entirely new syntax and wouldn't conflict so it's an opportunity to correct that if we like.
 * Document `maxServerVersion` as inclusive for consistency with `minServerVersion`. This isn't very controversial, as we currently use non-existent versions like "2.4.99" for `maxServerVersion`.

I'm OK with both of these suggestions and would like to make the changes to both PRs before merging. Any objections?

jmikola:
Per Matt's comment in https://github.com/mongodb/specifications/pull/511#discussion_r266499578, I'll follow up with a documentation change for `maxServerVersion` and we can leave `topology` as-is in this PR.

PR #511, opened 2019-03-01
SPEC-1239: Use runOn for transaction and convenient API test reqs

jmikola:
https://jira.mongodb.org/browse/SPEC-1239

This applies the `runOn` syntax introduced in #510 to the transaction and convenient API spec tests. Additionally, it clears up some miscommunication about running the convenient API tests on sharded clusters.


ShaneHarvey:
LGTM, tested here https://evergreen.mongodb.com/version/5c7ec95ad6d80a0e9e38bd86 (failures are because of the countDocuments change in SPEC-1232.

source/transactions-convenient-api/tests/README.rst
```diff
@@ -28,16 +28,33 @@ Test Format
...
+  of these requirements, the tests may be executed; otherwise, this file should
+  be skipped. If this field is omitted, the tests can be assumed to have no
+  particular requirements and should be executed. Each element will have some or
+  all of the following fields:
+
+  - ``minServerVersion`` (optional): The minimum server version (inclusive)
+    required to successfully run the tests. If this field is omitted, it should
+    be assumed that there is no lower bound on the required server version.
+
+  - ``maxServerVersion`` (optional): The maximum server version (exclusive)
```
p-mongo: The maximum server version is inclusive in the existing tests, can it be also inclusive here?

source/transactions-convenient-api/tests/README.rst
```diff
@@ -28,16 +28,33 @@ Test Format
...
+    against which the tests can be run successfully. If this field is omitted,
+    it should be assumed that there is no upper bound on the required server
+    version.
+
+  - ``topology`` (optional): An array of server topologies against which the
```
p-mongo: I think `topology` should be plural, since its value is an array.

source/transactions-convenient-api/tests/README.rst
jmikola: Where did you see an inclusive maximum server version? AFAIK, only the retryable writes and CRUD tests use `maxServerVersion` (apart from change streams, which says it is "reserved for future use") and they document it as an exclusive upper bound.

p-mongo: I was referring to this: https://github.com/mongodb/specifications/tree/master/source/command-monitoring/tests#ignoring-tests-based-on-server-version

The current uses of maxserverversion are:

<pre>
spec/spec_tests/data/crud/write/findOneAndReplace-upsert_pre_2.6.yml:maxServerVersion: '2.4.99'
spec/spec_tests/data/crud/write/replaceOne-pre_2.6.yml:maxServerVersion: '2.4.99'
spec/spec_tests/data/crud/write/updateMany-pre_2.6.yml:maxServerVersion: '2.4.99'
spec/spec_tests/data/crud/write/updateOne-pre_2.6.yml:maxServerVersion: '2.4.99'
</pre>

These are all inclusive specifications.

Also, "max server version: 1.0" implies that 1.0 is an acceptable version just looking at it from English grammar standpoint.

source/transactions-convenient-api/tests/README.rst
mbroadst: Let's make a SPEC ticket to do this work later since `topology` is currently used in a number of other test formats. I agree with you it should be plural, but I think this should all happen at the same time.

jmikola:
@p-mongo: Per Matt's comment above, I'll follow up with a documentation change for `maxServerVersion` and we can leave `topology` as-is in this PR.

source/transactions-convenient-api/tests/README.rst
jmikola: https://jira.mongodb.org/browse/SPEC-1254

source/transactions-convenient-api/tests/README.rst
jmikola: I will make this change for this PR and #510. I also mentioned this in https://jira.mongodb.org/browse/SPEC-1255, as that will pertain to changing the CRUD spec tests mentioned above.

PR #517, opened 2019-03-07
SPEC-1046 prohibit CNAME with SRV/TXT

xdg:


source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
```diff
@@ -10,8 +10,8 @@ Initial DNS Seedlist Discovery
 :Authors: Derick Rethans
 :Status: Draft
 :Type: Standards
-:Last Modified: 2018-02-08
-:Version: 1.3.1
+:Last Modified: 2019-03-07
+:Version: 1.3.2
```
jmikola: I don't see a value for "driver changes needed" in [SPEC-1046](https://jira.mongodb.org/browse/SPEC-1046). If this is definitely just a clarification and won't require actual changes, I'm on board with a patch version bump. If driver changes were needed, then I think this would warrant a minor version bump.

Reading the spec ticket, I gather that there is no reason a driver would have supported CNAME in their original implementation. Is that correct?

LGTM if so.

xdg: Correct -- we're clarifying that drivers must not do something that probably none of them did in the first place.  Because there are security considerations, we're calling it out explicitly.

PR #518, opened 2019-03-08
 WRITING-3385 Connections survive primary step down

p-mongo:


source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,214 @@
...
+comparing socket handles used for the two operations.
+
+Acknowledged Write
+``````````````````
+
+This test requires server version 4.2 or higher.
+
+Perform the following operations:
+
+- Obtain a connection to the current primary.
```
xdg: Seems like this also needs the "may need lower level than public API" comment as I imagine many drivers don't easily expose specific connections for operations.

I wonder if these tests are going to need or could more easily be done some configuration specific to the test, e.g.:
* turn off retryable writes
* set connection pool to size 1

source/server-discovery-and-monitoring/server-discovery-and-monitoring-summary.rst
```diff
@@ -221,4 +221,6 @@ When the client sees such an error it knows its topology view is out of date.
 It MUST mark the server type "unknown."
 Multi-threaded and asynchronous clients MUST re-check the server soon,
 and single-threaded clients MUST request a scan before the next operation.
-The client SHOULD clear its connection pool for the server.
+The client SHOULD clear its connection pool for the server if the
```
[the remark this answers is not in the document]
p-mongo: This is existing language in the spec and I would think a SPEC ticket would be the appropriate venue to propose changes to it.

source/connections-survive-step-down/tests/README.rst
p-mongo: Setting connection pool size to 1 does not necessarily preclude a driver from using a different connection for each operation.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-summary.rst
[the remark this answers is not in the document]
xdg: OK.

source/connections-survive-step-down/tests/README.rst
xdg: It doesn't preclude it, but statistically, it's likely.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1514,7 +1514,11 @@ servers. For a "node is recovering" error, single-threaded clients MUST NOT
 mark the topology as "stale". If a node is recovering for some time, an
 immediate scan may not gain useful information.
 
-The client SHOULD clear its connection pool for the server.
+If the client is connected to server version 4.2 or higher,
+the client SHOULD keep any connections it has to the server open,
+and SHOULD NOT clear its connection pool for the server.
+If the client is connected to server version 4.0 or lower, the client
+SHOULD clear its connection pool to the server.
```
ShaneHarvey: Is this supposed to apply only when connected to replica sets or does it also include mongos?

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,214 @@
...
+   as the procedure is complete. Otherwise propagate the operation failure
+   error(*).
+6. Connect to the server E and execute the `replSetStepDown
+   <https://docs.mongodb.com/manual/reference/command/replSetStepDown/>`_
+   admin command as follows:
+   ``{replSetStepDown: 4, secondaryCatchUpPeriodSecs: 2}``.
+7. Connect directly to the server N and execute the following admin command:
+   ``{replSetStepUp: 1}``. If this command fails with an operation failure
+   error with code 125 ("Election failed"), repeat this step 7.
+8. Set the status of all servers in the topology to Unknown.
```
jmikola: For drivers that don't have direct access to SDAM internals (e.g. drivers built atop libmongoc), is there an alternative approach for this step? Perhaps by tweaking monitoring options for force a re-scan of the topology before server selection in the subsequent step?

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,214 @@
...
+
+This test requires server version 4.2 or higher.
+
+Perform the following operations:
+
+- Insert 5 documents into a collection.
+- Start a find operation on the collection with a batch size of 2, and
+  retrieve the first batch of results.
+- Step down the primary, and verify that another server has been elected as
+  a primary.
```
jmikola: Is it actually necessary to assert that another server has been elected? If we issue a find query on a primary and then step it down, I would think we could just continue iterating to ensure that the driver successfully issues a getMore against the former primary. Provided the 4.2 server isn't dropping its connections on a step down, I don't see what this test has to do with SDAM.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,214 @@
...
+- Using the same connection that was used for the initial find, continue
+  iterating the collection by issuing a getMore operation.
+- Verify that the next batch of results is returned successfully.
+
+Drivers that do not ordinarily allow applications to execute operations
+on specific connections may need to perform the getMore iteration test on
+a lower level than the public API exposed to applications.
```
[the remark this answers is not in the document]
jmikola: Is the concern that by not restricting the getMore to the same connection the driver could potentially open a new connection to the stepped down primary and that would not sufficiently test the "connections survive" behavior? If that's the case, I think restricting the connection pool size to 1 makes sense and also avoids needing to access low-level behavior in the driver.

I suppose for drivers without connection pools, this isn't a concern at all.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,214 @@
...
+- If the second unacknowledged insert succeeded, dispatch another unacknowledged
+  insert over the same connection[*].
+- Verify that the last unacknowledged insert produced a network error due
+  to the connection being closed by the server.
+
+[*] Depending on buffering performed by the network stack, the first write
+operation on a connection which was closed on the remote end may succeed.
+In the acknowledged write scenario, the error is detected upon (trying to)
+read the response. In the unacknowledged write scenario, an additional write
+may be needed to detect that the connection is closed.
```
jmikola: This talks about an additional write being needed to detect a closed connection; however, the sequence of operations suggests three possible inserts after the step down. Why can't we trust the second insert to detect the closed connection?

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,214 @@
...
+- Step down the primary, and verify that another server has been elected as
+  a primary.
+- Dispatch an acknowledged insert operation over the connection obtained in
+  the first step.
+- Verify that the result of the insert operation is a NotMaster error.
+- Verify that the connection used for the insert operation remains open.
+- Step up the server which was originally the primary.
+- Dispatch another acknowledged insert operation over the connection obtained
+  in the first step.
+- Verify that the insert succeeded and the data inserted is in the collection.
```
jmikola: Per the [SDAM spec](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#not-master-and-node-is-recovering), encountering a NotMaster error should update the server's state to Unknown. Is this test expecting the driver to bypass server selection and execute the insert against the server while it has an Unknown state?

If so, do I assume correctly that the server would maintain its Unknown state until the next monitoring interval?

source/connections-survive-step-down/tests/README.rst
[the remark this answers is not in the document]
ShaneHarvey: I don't think we need this restriction in the test. The goal of this project is for cursors to survive elections so this test should verify that cursors work regardless of if it used the same connection (or used maxPoolSize=1). For example, shouldn't the test pass when the pool has 2 connections and find is run on connection 1 and the getMore is run on connection 2?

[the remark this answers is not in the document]
p-mongo: @jmikola Ruby driver may return a different connection for each operation even with a connection pool size of 1. If this happens, the test would not be verifying that the original connection remained open.

@ShaneHarvey  The motivation behind these tests is that in a sharded cluster, sending a getMore over a new connection may send it  to a different shard, which wouldn't be able to use the cursor established on the first shard. Thus the requirement to test/use the exact same connection that was used for initial find.

source/connections-survive-step-down/tests/README.rst
p-mongo: Yes, the exact wording of this step can be modified. The goal is to refresh the driver's topology view and find out the current primary.

source/connections-survive-step-down/tests/README.rst
p-mongo: Consider the following scenario:

- Server A is the primary. We step it down.
- Server B is currently not a primary. The driver notices this and sends a getMore to the cluster.
- By the time the cluster receives the getMore, server A has gotten elected a primary again (instead of some other server).
- The getMore ended up getting routed to the current primary.

One solution to this would be to check server state after getMore and verify it is still not a primary, if you find this easier than ensuring that some other server is a primary before sending the getMore.

source/connections-survive-step-down/tests/README.rst
p-mongo: The second insert is actually the first insert after the close, and since the close is on the other end only (our end hasn't read or written anything yet and thus has no way of being aware of the close), the first write simply buffers the data and succeeds. I needed to write over 20k to the connection to have the first write fail.

I also ended up putting a half second sleep between the second and the third insert because sometimes even two consecutive inserts would be successfully buffered and not error.

source/connections-survive-step-down/tests/README.rst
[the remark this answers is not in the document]
ShaneHarvey: The readme at the top says that these test only apply to replica sets so I don't think we need to consider sharded clusters.

source/connections-survive-step-down/tests/README.rst
p-mongo: Good point, I believe I need to modify the procedure to stop sdam monitoring in this case, so that the driver thinks the no-longer-primary server is still the primary.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
p-mongo: Replica sets only per the top of the document.

source/connections-survive-step-down/tests/README.rst
jmikola: I wouldn't think a test would need to worry about A being reëlected after a step down, as [`replSetStepDown`](https://docs.mongodb.com/manual/reference/command/replSetStepDown/) applies a 60-second waiting period by default. That said, I don't see the harm in ensuring the SDAM picks up a new primary before we continue iteration on the cursor. IIRC, cursors remain pinned to the original server so we wouldn't have to worry about a getMore inadvertently targeting a different node. Also, retryable reads also don't come into play for getMore operations, so there's nothing to worry about there.

source/connections-survive-step-down/tests/README.rst
jmikola: That explanation makes sense. If the server closes the connection after the first unacknowledged write, the earliest a driver might detect a closed socket is when attempting to write the second insert. And you're explaining that we might need another attempt (second after the initial closure, but third overall) to overcome buffering or other subtleties.

It may be worth explaining this in a bit more detail for the benefit of others reading the final prose test. Also, a tip about using larger insert payloads to overcome network buffering might be helpful. IIRC, 8 KiB is a common network buffer size but performing an insert with 64+ KiB of data is probably safer.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
ShaneHarvey: This is the SDAM spec. My question is when we are connected to a 4.2 Mongos, should we reset the pool after a "not master" error?

source/connections-survive-step-down/tests/README.rst
p-mongo: The procedure described reduces that 60 second period to 4 seconds, and thus should not be relied on.

source/connections-survive-step-down/tests/README.rst
p-mongo: Would you prefer the language requiring the driver to send a sufficiently large payload (64kb+ as you said) to overcome network buffering over the current language suggesting multiple writes?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
p-mongo: I am not too familiar with sharded clusters, but my impression is a not master in a SC should be treated the same way it would be in a simple replica set. Are there any issues with keeping the pool in a sharded cluster?

ShaneHarvey: > but my impression is a not master in a SC should be treated the same way it would be in a simple replica set.

Yeah I think I agree. PyMongo treats these errors the same for both.

I do have a different concern though. This section says we should not reset the pool after "not master" or "node is recovering" errors. "node is recovering" includes errors caused during **shutdown** (not stepdown), like `InterruptedAtShutdown` and `ShutdownInProgress`. I think we want those shutdown errors to reset the pool even on 4.2.

p-mongo: If a node is shutting down, what is the point of creating a new connection to this node which will then return another InterruptedAtShutdown error? It's just as efficient (more efficient in fact, as I understand it) to receive that error over the existing connection.

ShaneHarvey: I don't understand what you mean by "creating a new connection". We're only talking about clearing (closing) the connections in the pool.

p-mongo: If no operations are sent to the server in question, then does it matter whether the pool is cleared or not?

xdg:
I wonder if we should reconsider the scope at a higher level.  What's really important to test:

* On receipt of "not master", don't close the pool (4.2+) – this can be confirmed via CMAP
* A cursor can continue to be iterated after server stepdown – this can be done purely via iteration, without regard to the underlying connections.  Yes, there might be various races that could cause this to succeed for unrelated reasons, but I don't think there are races that would cause it to fail.  As long as the test never fails, the odds are that it's working correctly
* The importance of the primary-stepdown-primary test is that connections remain open, i.e. they don't have to be reopened – this should show up in command monitoring and/or CMAP -- we should be able to execute writes without seeing a retry (from the EPIPE sending on server closed connection) or new connections spawning.

While the scope speaks to "on the same connection", I don't actually think that's what we need to verify -- we can verify the users' black-box experience.

xdg:
Something else I remembered -- I wonder if 'getMore' needs to be sent with secondary-ok read preference.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
ShaneHarvey: Yes it does matter. Let's say that the primary shuts down and that during the shutdown the application sees a ShutdownInProgress error. If we do not clear the pool then the Pool contains N dead sockets. When the node restarts we will still have N dead sockets and the next attempt to use this server will fail with a connection error. 

If instead we do clear the pool after ShutdownInProgress, then the next attempt to use this server will succeed on a new connection.

p-mongo: I see your point, however given that we do not currently have a definition of "an error caused by server shutdown" I would like to not add figuring this out to the step down tests. What would you suggest as changes to make in this PR?

p-mongo:
This project did not have a dependency on CMAP, however some if not all of the assertions appear to be simplifiable by leveraging CMAP. I will look into doing that.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
ShaneHarvey: I think we should introduce a new class of errors called "shutdown error" which includes `InterruptedAtShutdown` and `ShutdownInProgress` as the server does: https://github.com/mongodb/mongo/blob/r4.1.8/src/mongo/base/error_codes.err#L333

The we can say that drivers MUST clear the connection pool after a "shutdown error" regardless of the server version.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1514,7 +1514,11 @@ servers. For a "node is recovering" error, single-threaded clients MUST NOT
 mark the topology as "stale". If a node is recovering for some time, an
```
ShaneHarvey: In [updateRSFromPrimary](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#updatersfromprimary)  it says:

> The client SHOULD clear its connection pool for the old primary, too: the connections are all bad because the old primary has closed its sockets.

Does this also need to be updated to say that on >=4.2 we should not clear the pool?

ShaneHarvey:
@xdg the server scope doc has a goal that:

> Non-transactional reads on primaries that step down should be allowed to complete. This is true even if the specified ReadPreference is Primary

So it should work, but I agree we need to test it.

source/connections-survive-step-down/tests/README.rst
jmikola: Are two unacknowledged writes always guaranteed to overcome network buffering? If the buffer is 8 KiB, I'm curious how two very small writes might fill that up -- unless there is some other factor responsible for flushing the buffer.

p-mongo: They are not, I put a sleep between them to force the issue.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
p-mongo: @xdg What do you think about adding the above proposal to this step down project?

xdg: I think it's a good idea.

source/connections-survive-step-down/tests/README.rst
jmikola: Ah, in that case I'd vote for using larger writes in lieu of a sleep. Happy to defer to someone else's opinion on this, though.

source/connections-survive-step-down/tests/README.rst
jmikola: Ah, I just saw the `4` above. Is that value intended to allow the RS to restore itself to the original state before the subsequent test? I don't recall reading anything in the document about ensuring the RS is in an original state before running each test (perhaps I missed it), but even 4 seconds seems like it might be too long in some cases.

This reminds me of similar challenges in Emily's original [failover integration tests](https://github.com/estolfo/mo-tests) (and relevant to our upcoming failover testing spec). I believe the solution there was to use a freshly constructed replica set for each test, so there was no concern about the leftover state from the previous test.

Do you think that would be preferable here, or would it be too much work to expect drivers to coordinate with MO during the test suite?

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,214 @@
...
+This test requires server version 4.0 or lower.
+
+Perform the following operations:
+
+- Perform server selection for the primary server to obtain a server object
+  with an initialized connection pool.
+- Step down the primary, OR mock an operation returning NotMaster operation
+  failure error code.
+- Verify connection pool generation has been incremented by 1, AND/OR that
+  PoolCleared CMAP eent has been published.
```
jmikola: *event

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1514,11 +1514,31 @@ servers. For a "node is recovering" error, single-threaded clients MUST NOT
...
 immediate scan may not gain useful information.
 
-The client SHOULD clear its connection pool for the server.
+The following subset of "not master" errors is defined to be "server shutting
+down" errors:
```
xdg: According to the tables above, this is a subset of "node is recovering" errors.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1514,11 +1514,31 @@ servers. For a "node is recovering" error, single-threaded clients MUST NOT
...
+
+  * - Error Name
+    - Error Code
+  * - InterruptedAtShutdown
+    - 11600
+  * - ShutdownInProgress
+    - 91
+
+If the client is connected to server version 4.2 or higher, and the client
+receives a "not master" error which is not a "server shutting down" error,
```
xdg: `... receives a "not master" or "node is recovering" error which is not a "server shutting down" error, ...`

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1514,11 +1514,31 @@ servers. For a "node is recovering" error, single-threaded clients MUST NOT
...
+the client MUST keep any connections it has to the server open,
+and MUST NOT clear its connection pool for the server.
+If the client is connected to server version 4.2 or higher and receives a
+"server shutting down" error, or
+if the client is connected to server version 4.0 or lower and receives a
+"not master" error, the client MUST clear its connection pool to the server.
```
xdg: `... "not master" or "node is recovering" error ...`

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -2295,6 +2315,16 @@ See `"not master" and "node is recovering"`_,
...
 
+Why close connections when server is shutting down?
+'''''''''''''''''''''''''''''''''''''''''''''''''''
+
+When a server shuts down, it will return one of the "server shutting down"
+errors for each attempted operation and eventually will close all connections.
+Keeping a connection to a server which is shutting down open would only
+produce errors on this connection - such a connection will never be usable for
+any operations. In contrast, when a server returns "not master" error the
+connection may be usable for other operations (such as secondary reads).
```
xdg: But add a note that that is 4.2 specific.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,198 @@
...
+<https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst>`_.
+
+Perform the following operations:
+
+- Insert 5 documents into a collection.
+- Clear the connection pool to the current primary (see CMAP spec for
+  the definition of connection pool clear operation).
+- Start a find operation on the collection with a batch size of 2, and
+  retrieve the first batch of results.
+- Verify that a ConnectionCreated CMAP event has been published.
```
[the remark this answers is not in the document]
p-mongo: Since the following assertion is that a cmap event is NOT sent, it makes sense to assert that one is sent by the driver is some other circumstance, otherwise a driver which doesn't implement cmap events at all will pass the test. To guarantee that the event is sent, pool is cleared first. Would you like me to add this to the specification or remove the pool clear and publication verification entirely?

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,198 @@
...
+Not Master - Keep Connection Pool
+`````````````````````````````````
+
+This test requires server version 4.2 or higher.
+
+Perform the following operations:
+
+- Perform server selection for the primary server to obtain a server object
+  with an initialized connection pool.
+- Step down the primary, OR mock an operation returning NotMaster operation
```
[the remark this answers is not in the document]
p-mongo: My impression of the meeting last week was that as few tests as possible were desired, but if you want me to require both operations I can do that as well.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,198 @@
...
+  failure error code.
+- Verify that:
+  - Connection pool generation (per CMAP spec) has been incremented by 1
+  - PoolCleared CMAP event has not been published
```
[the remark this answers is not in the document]
p-mongo: Yes thank you.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,198 @@
...
+
+Perform the following operations:
+
+- Perform server selection for the primary server to obtain a server object
+  with an initialized connection pool.
+- Mock an operation returning ShutdownInProgress (91) operation
+  failure error code.
+- Verify that:
+  - Connection pool generation (per CMAP spec) has been incremented by 1
+  - PoolCleared CMAP event has not been published
```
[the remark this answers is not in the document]
p-mongo: Yes thank you.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
p-mongo: I see, thank you.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
p-mongo: Made this change, thank you.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
p-mongo: Made this change, thank you.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
p-mongo: Okay will add this.

source/connections-survive-step-down/tests/README.rst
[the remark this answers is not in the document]
xdg: I think we need both to still meet the "as possible" clause.  :-)  What I mean is that these are two very distinct cases and we need to cover both.

source/connections-survive-step-down/tests/README.rst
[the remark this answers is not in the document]
xdg: I think it's sufficient for you to add a comment explaining what you just said (inline or in a rationale section).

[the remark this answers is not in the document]
p-mongo: Added thank you.

source/connections-survive-step-down/tests/README.rst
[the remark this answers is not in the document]
p-mongo: Per offline discussion this paragraph has been changed to require mocking a not master response.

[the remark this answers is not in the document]
xdg: (side note: It's possible we might need a failpoint for drivers that can't mock a response)

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1514,11 +1514,32 @@ servers. For a "node is recovering" error, single-threaded clients MUST NOT
...
+    - 11600
+  * - ShutdownInProgress
+    - 91
+
+If the client is connected to server version 4.2 or higher, and the client
+receives a "not master" or "node is recovering" error which is not a
+"server shutting down" error, the client MUST keep any connections it has to
+the server open, and MUST NOT clear its connection pool for the server.
+If the client is connected to server version 4.2 or higher and receives a
+"server shutting down" error, or
```
jmikola: Premature line break here.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,209 @@
...
+
+This test requires server version 4.2 or higher.
+
+Perform the following operations:
+
+- Perform server selection for the primary server to obtain a server object
+  with an initialized connection pool.
+- Mock an operation returning NotMaster operation failure error code.
+- Verify that:
+  - Connection pool generation (per CMAP spec) has not changed
```
jmikola: If you view GitHub's rendering of the RST file, this does not render as a nested sub-list. Your alignment is correct, but per [Bullet Lists](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#bullet-lists) I think sub-lists require a line break above and below.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,209 @@
...
+Not Master - Keep Connection Pool
+`````````````````````````````````
+
+This test requires server version 4.2 or higher.
+
+Perform the following operations:
+
+- Perform server selection for the primary server to obtain a server object
+  with an initialized connection pool.
+- Mock an operation returning NotMaster operation failure error code.
```
jmikola: In some drivers, "mock" may refer to tests that don't actually involve a server. Assuming the intention here is to actually connect to a real replica set, would it be better to suggest using a fail point here? If so, perhaps you can just add "The failCommand fail point may be used for this purpose" after this sentence.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,209 @@
...
+Server Step Down Procedure
+--------------------------
+
+In principle, the current primary may be stepped down by running the following
+admin command on it: `{replSetStepDown: null}`. This will cause the current
+primary to step down and an election to be held to elect a new primary.
+A driver MAY use this simple procedure to step down the current primary.
+
+Note that MongoDB server 4.0 and lower will close all connections on step down,
+including the one which sent the `{replSetStepDown: 1}` command. Therefore
```
jmikola: Is `1` in `{replSetStepDown: 1}` intentional as the duration? I assume `null` was used in the paragraph above to intentionally _not_ specify a duration. These should probably be consistent, unless you'd like to explicitly suggest that test suites use a duration smaller than the 60-second default.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,209 @@
...
+configured in a specific way. This configuration should be done prior to
+running any of the below tests. When the tests are finished, the default
+cluster configuration should be restored.
+
+Adjust replica set configuration to have the following settings:
+
+- ``electionTimeoutMillis``: 1000
+
+Connect to each of the servers in the deployment directly and set the
+following server parameters:
```
jmikola: Does this entail using the [`setParameter`](https://docs.mongodb.com/manual/reference/command/setParameter/#dbcmd.setParameter) command? If so, can we link to it for clarity?

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,209 @@
...
+
+Cluster Configuration
+---------------------
+
+In order to have quick and repeatable elections, the deployment needs to be
+configured in a specific way. This configuration should be done prior to
+running any of the below tests. When the tests are finished, the default
+cluster configuration should be restored.
+
+Adjust replica set configuration to have the following settings:
```
jmikola: Suggest linking to [`replSetReconfig`](https://docs.mongodb.com/manual/reference/command/replSetReconfig/) here for clarity.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,209 @@
...
+Connect to each of the servers in the deployment directly and set the
+following server parameters:
+
+- ``enableElectionHandoff``: false
+
+Note: for the elections to work correctly, the values specified for
+replSetStepDown, secondaryCatchUpPeriodSecs and electionTimeoutMillis must
+satisfy the following relationship:
+
+``replSetStepDown`` >= ``secondaryCatchUpPeriodSecs`` + ``electionTimeoutMillis``/1000.0
```
jmikola: I think this is the only mention of `secondaryCatchUpPeriodSecs` in the document. Should we mention that its default value is 10 seconds (mentioned in [`replSetStepDown`](https://docs.mongodb.com/manual/reference/command/replSetStepDown/)? Alternatively, would it make sense to disregard `secondaryCatchUpPeriodSecs` entirely and suggest test suites use the `force` option?

I realize you go into more detail about this in the "Server Election Procedure" below, but it's not clear to me if that stands alone or is guidance for earlier tests.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,209 @@
...
+- Verify that:
+  - Connection pool generation (per CMAP spec) has been incremented by 1
+  - PoolCleared CMAP event has been published
+
+
+Server Step Down Procedure
+--------------------------
+
+In principle, the current primary may be stepped down by running the following
+admin command on it: `{replSetStepDown: null}`. This will cause the current
```
jmikola: RST needs double backticks for monospace formatting. Also in the following paragraph.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,209 @@
...
+following procedure:
+
+- Obtain a list of servers in the cluster.
+- Randomly or otherwise choose a server to be the new primary, other than the
+  current primary.
+- Follow the above server election procedure to elect the chosen server as
+  the new primary.
+
+
+Questions and Answers
```
jmikola: For drivers that do not have pools and do not implement CMAP, I think it'd make sense to discuss an alternative method of asserting connection behavior here. IIRC, we talked about checking [`connections.totalCreated`](https://docs.mongodb.com/manual/reference/command/serverStatus/#serverstatus.connections.totalCreated) from `serverStatus` in a previous meeting.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,209 @@
...
+Perform the following operations:
+
+- Insert 5 documents into a collection.
+- Clear the connection pool to the current primary (see CMAP spec for
+  the definition of connection pool clear operation, and
+  `Why clear the pool in getMore test?`_ for rationale).
+- Start a find operation on the collection with a batch size of 2, and
+  retrieve the first batch of results.
+- Verify that a ConnectionCreated CMAP event has been published.
+- Send a `replSetStepDown` command to the current primary and verify that
```
jmikola: Double backticks for monospace formatting in RST.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,209 @@
...
+- Perform server selection for the primary server to obtain a server object
+  with an initialized connection pool.
+- Mock an operation returning ShutdownInProgress (91) operation
+  failure error code.
+- Verify that:
+  - Connection pool generation (per CMAP spec) has been incremented by 1
+  - PoolCleared CMAP event has been published
+
+
+Server Step Down Procedure
```
jmikola: What is the purpose of this section and those that follow? Looking above, the only reference to actually stepping down a server is one of the later bullets in the "getMore Iteration" prose test. Are they purely for that?

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,209 @@
+===========================================
+Connections Survive Primary Step Down Tests
+===========================================
+
+These tests can be used to verify a driver's compliance with server discovery
+and monitoring requirements with respect to handling "not master" and
+"server shutting down" error responses from the server.
+
```
jmikola: Since there are a number of sections in this file, I think it'd make sense to insert a generated TOC here with `.. contents::`. That would make it easy to see all required tests at a glance.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
p-mongo: Repaired.

source/connections-survive-step-down/tests/README.rst
p-mongo: I see, thank you.

source/connections-survive-step-down/tests/README.rst
p-mongo: It is my impression that we have agreed not to test the server and only test the drivers, thus it should be fine for a driver to use a mock server for these tests. Also, @xdg  wanted to test both a mock and a real server initially and changed his mind to require testing a mock only.

source/connections-survive-step-down/tests/README.rst
p-mongo: This should be command name only, thank you.

source/connections-survive-step-down/tests/README.rst
p-mongo: Clarified that secondaryCatchUpPeriodSecs is part of replSetStepDown command, does this make it more clear?

source/connections-survive-step-down/tests/README.rst
p-mongo: Repaired thank you.

source/connections-survive-step-down/tests/README.rst
p-mongo: Added as a Q&A section, thank you.

source/connections-survive-step-down/tests/README.rst
p-mongo: Repaired, thank you.

source/connections-survive-step-down/tests/README.rst
p-mongo: Yes, linked here from that section.

source/connections-survive-step-down/tests/README.rst
p-mongo: Added, thank you.

source/connections-survive-step-down/tests/README.rst
xdg: Also, this is about "shutdown" not "stepdown".

source/connections-survive-step-down/tests/README.rst
jmikola: Works for me, thanks.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,224 @@
...
+sense to assert that one is sent by the driver is some other circumstance,
+otherwise a driver which doesn't implement CMAP events at all will pass the
+test. To guarantee that the event is sent, pool is cleared first.
+
+Alternative method of asserting connection behavior
+```````````````````````````````````````````````````
+
+Drivers which do not implement connection pools and CMAP specification may,
+instead of using CMAP events to assert that no new connections have been
+established, check `connections.totalCreated
```
jmikola: Grammar fix: "check the"

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,224 @@
...
+`````````````````
+
+This test requires server version 4.2 or higher, and makes use of events
+defined in the `CMAP specification
+<https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst>`_.
+
+Perform the following operations:
+
+- Insert 5 documents into a collection.
+- Clear the connection pool to the current primary (see CMAP spec for
```
jyemin: Clearing the pool is a low level operation in a driver, and not something generally exposed via a MongoClient.

Consider requiring that the insert just above use a different MongoClient than the find below.  That way the pool can be made to be empty as the find is the first operation.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,224 @@
...
+Not Master - Keep Connection Pool
+`````````````````````````````````
+
+This test requires server version 4.2 or higher.
+
+Perform the following operations:
+
+- Perform server selection for the primary server to obtain a server object
+  with an initialized connection pool.
+- Mock an operation returning NotMaster operation failure error code.
```
jyemin: I might have missed it, but is there an associated change to the SDAM spec for this?

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,224 @@
...
+  the definition of connection pool clear operation, and
+  `Why clear the pool in getMore test?`_ for rationale).
+- Start a find operation on the collection with a batch size of 2, and
+  retrieve the first batch of results.
+- Verify that a ConnectionCreated CMAP event has been published.
+- Send a ``replSetStepDown`` command to the current primary and verify that
+  the command succeeded, or follow `Server Step Down Procedure`_ below.
+- Retrieve the next batch of results from the cursor obtained in the find
+  operation, and verify that this operation succeeded.
+- Verify that no new ConnectionCreated CMAP events have been published.
```
behackett: For that specific pool, right?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
ShaneHarvey: I was not referring to this section where the comment is located. I was referring to the [updateRSFromPrimary](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#updatersfromprimary) section. (Github PRs don't let you put a comment on lines that haven't changed.)

So carrying on with my question about updateRSFromPrimary. According to the updated SDAM spec in this PR, drivers will *still* clear the pool when a primary steps down; when the new primary reports itself via a heartbeat, SDAM says that the connections to the old primary SHOULD be closed.

Personally I think that we should not change the SDAM spec in this PR and we should say that drivers still close connections when a primary steps down. Cursors will still survive elections which is the main motivation of this project.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,224 @@
...
+- Verify that:
+  - Connection pool generation (per CMAP spec) has been incremented by 1
+  - PoolCleared CMAP event has been published
+
+
+Server Step Down Procedure
+--------------------------
+
+In principle, the current primary may be stepped down by running the following
+admin command on it: ``{replSetStepDown: null}``. This will cause the current
```
behackett: Do you need force: true?

jmikola: That would probably speed things up and avoid the math formula pertaining to `secondaryCatchUpPeriodSecs`. I don't think this test needs to care about rollbacks, so we can probably just use `force: true` to run as quickly as possible.

p-mongo: `force: true` was not part of the guidance I received from the server team with respect to step downs.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
ShaneHarvey: One of the cases I'm considering in this thread is when a election occurs when an application is idle. The sequence of events would be:
1. Primary is P1
1. Primary steps down and new primary, P2, is elected.
1. SDAM heartbeat for P2 returns and reports that P2 is the primary.
1. SDAM invalidates the old primary P1 and clears its connection pool.
1. Connections have not survived primary step down.

source/connections-survive-step-down/tests/README.rst
p-mongo: If there is a rollback, since we are reading data that was just inserted, would it not be possible for the read to fail?

source/connections-survive-step-down/tests/README.rst
p-mongo: Given @jyemin comment above that a pool is not accessible for these tests to assert on, it is not clear to me how to specify that no events should be published for the pool belonging to the server on which getMore was executed.

jyemin: The _pool_ is not accessible (as in the ConnectionPool interface defined in CMAP), but the pool _events_ should be via event listeners registered on a MongoClient.

p-mongo: I already have the event assertion in the spec text, the suggestion was to also require that the event was for the pool related to the server which the getMore command was sent to. Can you suggest language which will accomplish this given Java driver's constraints on pool accessibility?

source/connections-survive-step-down/tests/README.rst
xdg: Not if the inserts are done with `w:majority`.

source/connections-survive-step-down/tests/README.rst
jyemin: We can just assert that _no_ new connections were created since the find command was initiated, regardless of which pool.  No other pool should have had a connection created in it either.

source/connections-survive-step-down/tests/README.rst
behackett: If it will speed things up, let's do it.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
ShaneHarvey: Any thoughts on this?

source/connections-survive-step-down/tests/README.rst
p-mongo: I removed the language related to pool clearing, as it was more of an implementation detail than an essential part of the specification.

source/connections-survive-step-down/tests/README.rst
p-mongo: This has been the original wording in this specification; @behackett  are you OK with the current language?

source/connections-survive-step-down/tests/README.rst
p-mongo: I added `force: true` to the simple command, however a forced step down still takes much longer (over 10 seconds in my testing) than one performed using the extended method.

Both seem to work, thus a driver may choose to implement the forced or the extended method depending on performance characteristics they would like their test suite to have (i.e. whether it is OK to wait 15 seconds for an election).

p-mongo:
@xdg @mbroadst This PR has been updated.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,146 @@
...
+If the driver implements CMAP specification, verify that no new
+ConnectionCreated CMAP events have been published. Otherwise verify that
+`connections.totalCreated` serverStatus value has increased by 1 using the same client that was used for test operation.
+
+
+How to verify the connection pool has not been cleared
+``````````````````````````````````````````````````````
+
+If the driver implements CMAP specification, verify that no new
+ConnectionCreated CMAP events have been published. Otherwise verify that
```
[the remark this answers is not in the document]
p-mongo: Repaired, thank you.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,146 @@
...
+`````````````````````````````````
+
+This test requires server version 4.2 or higher.
+
+- Set the following fail point: ``{configureFailPoint: "failCommand",
+  data: {failCommands: ["insert"], errorCode: 11600, mode: {times: 1}}``
+- Execute an insert into the test collection of a ``{test: 1}``
+  document.
+- Verify that the insert failed with an operation failure with 11600 code.
+- Verify that no new connections have been created following the instructions in section `How to verify the connection pool has not been cleared`_
```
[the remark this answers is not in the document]
p-mongo: I added another insert as you suggest, thank you.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,146 @@
...
+instead of using CMAP events to assert that no new connections have been
+established, check the `connections.totalCreated
+<https://docs.mongodb.com/manual/reference/command/serverStatus/#serverstatus.connections.totalCreated>`_
+value in serverStatus.
+
+
+How to verify the connection pool has been cleared
+``````````````````````````````````````````````````
+
+If the driver implements CMAP specification, verify that no new
```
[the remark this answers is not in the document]
p-mongo: OK I changed this to PoolCleared.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,146 @@
...
+ConnectionCreated CMAP events have been published. Otherwise verify that
+`connections.totalCreated` serverStatus value has increased by 1 using the same client that was used for test operation.
```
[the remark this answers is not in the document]
p-mongo: I think the idea was that server status check would be the operation that you are talking about.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -0,0 +1,146 @@
...
+
+
+
+Questions and Answers
+---------------------
+
+Do we need to wait for re-election after the first test?
+``````````````````````````````````````````````````````````
+
+Since test setup requires an insert into a collection, a primary must exist, so subsequent tests will block in server selection until a primary is available again.
```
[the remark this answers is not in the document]
p-mongo: Changed to say creation.

PR #521, opened 2019-03-13
Change Streams: Ensure that getMore right after a resume is retried

jyemin:
SPEC-1073

### DIFF
--- a/source/change-streams/tests/README.rst
+++ b/source/change-streams/tests/README.rst
@@ -33,7 +33,7 @@ Each YAML file has the following keys:
   - ``description``: The name of the test.
   - ``minServerVersion``: The minimum server version to run this test against. If not present, assume there is no minimum server version.
   - ``maxServerVersion``: Reserved for later use
-  - ``failPoint``: Reserved for later use
+  - ``failPoint``(optional): The configureFailPoint command document to run to configure a fail point on the primary server.
   - ``target``: The entity on which to run the change stream. Valid values are:
   
     - ``collection``: Watch changes on collection ``database_name.collection_name``
@@ -43,6 +43,9 @@ Each YAML file has the following keys:
     Valid topologies are ``single``, ``replicaset``, and ``sharded``.
   - ``changeStreamPipeline``: An array of additional aggregation pipeline stages to add to the change stream
   - ``changeStreamOptions``: Additional options to add to the changeStream
+
+    - ``batchSize``: The batch size to apply to the change stream
+
   - ``operations``: Array of documents, each describing an operation. Each document has the following fields:
 
     - ``database``: Database against which to run the operation
@@ -104,6 +107,9 @@ For each YAML file, for each element in ``tests``:
   - Drop the database ``database2_name``
   - Create the database ``database_name`` and the collection ``database_name.collection_name``
   - Create the database ``database2_name`` and the collection ``database2_name.collection2_name``
+  - If the the ``failPoint`` field is present, configure the fail point on the primary server. See
+    `Server Fail Point <../../transactions/tests#server-fail-point>`_ in the
+    Transactions spec test documentation for more information.
 
 - Create a new MongoClient ``client``
 - Begin monitoring all APM events for ``client``. (If the driver uses global listeners, filter out all events that do not originate with ``client``). Filter out any "internal" commands (e.g. ``isMaster``)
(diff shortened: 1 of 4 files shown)


source/change-streams/tests/README.rst
```diff
@@ -155,3 +155,8 @@ The following tests have not yet been automated, but MUST still be tested
 #. The ``killCursors`` command sent during the "Resume Process" must not be allowed to throw an exception.
 #. ``$changeStream`` stage for ``ChangeStream`` against a server ``>=4.0`` that has not received any results yet MUST include a ``startAtOperationTime`` option when resuming a changestream.
 #. ``ChangeStream`` will resume after a ``killCursors`` command is issued for its child cursor.
+#. ``ChangeStream`` will resume multiple times even when no intervening change stream documents are received by the cursor. Consider this scenario:
```
jyemin: Is it worth saying that this could be done as a unit test if possible, and manually otherwise?  I can't see a driver actually implementing this as an automated prose test, at least not until after the live failover testing infrastructure is in place.

ShaneHarvey: Can't we just test this with `failCommand` setting `getMore`  to fail >=2 times?

jmikola: Reading the steps in [SPEC-1073](https://jira.mongodb.org/browse/SPEC-1073) that lead to this change, I would think that multiple `getMore` failures could trigger this. I believe the current change stream tests don't allow for interleaving insert operations during change stream iteration, but perhaps we can simulate shorter `getMore` responses by using batchSize to only request a single change record at a time.

jyemin: So it would be 

# aggregate (succeeds)
# getMore (fails with retryable error)
# aggregate (succeed)
# getMore (fails with retryable error)
# aggregate (succeeds)
# getMore (succeeds)
# change stream document received (not sure how to do this un-racily)

jmikola: Is there a race condition if you insert all of the documents up front and start the initial change stream with the appropriate operation time indicating that you want to see all of those inserts? With a small batchSize, you could then successfully receive them despite the two `getMore` failures.

jyemin:
Added an automated test, and updated the README to reflect that we're now using batchSize and failpoint.

source/change-streams/tests/README.rst
jyemin: Done

source/change-streams/tests/change-streams.yml
```diff
@@ -456,3 +456,77 @@ tests:
             coll: *collection_name
         -
           operationType: invalidate
+  -
+    description: Test consecutive resume
+    minServerVersion: "3.6.0"
```
ShaneHarvey: I think we should make this 4.1.7 to prevent secondaries from triggering the fail point: https://jira.mongodb.org/browse/SERVER-34943

jyemin: Done.  I'll put up a patch release of the Java driver before merging this.  It's passing the test locally.

source/change-streams/tests/README.rst
```diff
@@ -43,6 +43,9 @@ Each YAML file has the following keys:
     Valid topologies are ``single``, ``replicaset``, and ``sharded``.
   - ``changeStreamPipeline``: An array of additional aggregation pipeline stages to add to the change stream
   - ``changeStreamOptions``: Additional options to add to the changeStream
+
+    - ``batchSize``: The batch size to apply to the change stream
```
jmikola: Is this needed? Typically, options do not get documented in the spec test format since they map to an Options class/struct defined in the spec. I realize this spec's format differs from transactions/CRUD, but I think this can be consistent with how we work with `options` elsewhere.

jyemin: I added it because I looked at the Java driver test runner and it didn't support any change stream options at all, and figured that would be the case for other test runners as well.  But will remove if you prefer it.  I can see how for CRUD tests we wouldn't want to enumerate them all.

source/change-streams/tests/change-streams.json
```diff
@@ -675,6 +675,123 @@
           }
         ]
       }
+    },
+    {
+      "description": "Test consecutive resume",
+      "minServerVersion": "3.6.0",
```
jmikola: I think the JSON file needs to be regenerated. This is out of sync with the YAML below.

jyemin: Yeah, it was sitting un-pushed in my local repo.  Done.

source/change-streams/tests/change-streams.yml
jyemin: Java driver tests are passing

source/change-streams/tests/README.rst
jyemin: Done.

PR #522, opened 2019-03-18
Transactions convenient API updates for retryable writes on by default

jyemin:
SPEC-1256: add retryWrites:false client option to tests that write outside of a transaction

### DIFF
--- a/source/transactions-convenient-api/tests/callback-aborts.yml
+++ b/source/transactions-convenient-api/tests/callback-aborts.yml
@@ -89,6 +89,8 @@ tests:
     # Session state will be NO_TXN when callback returns to withTransaction
     description: withTransaction still succeeds if callback aborts and runs extra op
     useMultipleMongoses: true
+    clientOptions:
+      retryWrites: false
     operations:
       -
         name: withTransaction
--- a/source/transactions-convenient-api/tests/callback-commits.yml
+++ b/source/transactions-convenient-api/tests/callback-commits.yml
@@ -97,6 +97,8 @@ tests:
     # Session state will be NO_TXN when callback returns to withTransaction
     description: withTransaction still succeeds if callback commits and runs extra op
     useMultipleMongoses: true
+    clientOptions:
+      retryWrites: false
     operations:
       -
         name: withTransaction
--- a/source/transactions-convenient-api/tests/callback-aborts.json
+++ b/source/transactions-convenient-api/tests/callback-aborts.json
@@ -126,6 +126,9 @@
     {
       "description": "withTransaction still succeeds if callback aborts and runs extra op",
       "useMultipleMongoses": true,
+      "clientOptions": {
+        "retryWrites": false
+      },
       "operations": [
         {
           "name": "withTransaction",
--- a/source/transactions-convenient-api/tests/callback-commits.json
+++ b/source/transactions-convenient-api/tests/callback-commits.json
@@ -143,6 +143,9 @@
     {
       "description": "withTransaction still succeeds if callback commits and runs extra op",
       "useMultipleMongoses": true,
+      "clientOptions": {
+        "retryWrites": false
+      },
       "operations": [
         {
           "name": "withTransaction",


source/transactions-convenient-api/tests/callback-aborts.yml
```diff
@@ -89,6 +89,8 @@ tests:
     # Session state will be NO_TXN when callback returns to withTransaction
     description: withTransaction still succeeds if callback aborts and runs extra op
     useMultipleMongoses: true
+    clientOptions:
+      retryWrites: false
```
jmikola: Is this necessary? The APM assertion below intentionally avoids asserting a `txnNumber` field in the final insert after the abort:

> This test is agnostic about retryWrites, so we do not assert the txnNumber. If retryWrites=true, the txnNumber will be incremented from the value used in the previous transaction; otherwise, the field will not be present at all.

source/transactions-convenient-api/tests/callback-commits.yml
```diff
@@ -97,6 +97,8 @@ tests:
     # Session state will be NO_TXN when callback returns to withTransaction
     description: withTransaction still succeeds if callback commits and runs extra op
     useMultipleMongoses: true
+    clientOptions:
+      retryWrites: false
```
jmikola: Likewise, the APM assertion for the final insert after the commit avoids asserting `txnNumber`.

jmikola:
Just saw [SPEC-1256](https://jira.mongodb.org/browse/SPEC-1256):

> A couple of convent txn API tests failed in the Java driver when implementing retryWrites on by default. It's because the Java driver does exact matching an command monitoring events, and the fact that the MongoClient started defaulting to true meant that the writes outside of a txn started to include txnNumber and lsid.
>
> The fix is simply to do with we did for the other tests when updating for retryable writes on by default: add clientOptions specifying retryWrites of false.

IMO, this highlights a bug with the Java driver's APM assertion logic. This spec points back to the [Transaction spec README](https://github.com/mongodb/specifications/tree/master/source/transactions/tests#command-started-events), which talks about special handling for `null` values and other placeholders. I'm curious how the Java driver respects those special assertions while still performing an exact match on the command document.

I think all APM assertions across the entire repository are written with the assumption that expected documents in APM assertions are subsets of actual command documents. I codified this in the [CRUD spec README](https://github.com/mongodb/specifications/blob/master/source/crud/tests/README.rst#evaluating-matches) but we have prior are in the [change stream](https://github.com/mongodb/specifications/blob/master/source/change-streams/tests/README.rst#spec-test-match-function) and [command monitoring](https://github.com/mongodb/specifications/blob/master/source/command-monitoring/tests/README.rst#additional-values) specs.

If we're on the same page that Java is deviating in its APM assertion logic, I'd argue against the proposed fix in this PR. Hiding the issue in the spec tests makes it likely that another driver with the same deviation (which has yet to sync with these tests) might not catch it and have a chance to correct itself.

jyemin:
> IMO, this highlights a bug with the Java driver's APM assertion logic. This spec points back to the Transaction spec README, which talks about special handling for null values and other placeholders. I'm curious how the Java driver respects those special assertions while still performing an exact match on the command document.

I don't think there's a bug.  It is a more restrictive assertion, but one that works with all current tests that assert on APM events.  

The way the Jave driver respects the special assertions is to first remove any of white-listed fields with null values that have been identified as special, and then proceed with the exact matching.

> I think all APM assertions across the entire repository are written with the assumption that expected documents in APM assertions are subsets of actual command documents. I codified this in the CRUD spec README but we have prior are in the change stream and command monitoring specs.

All the current tests that assert on APM pass, so in practice the existing spec tests have been specifying exact command matches.

> If we're on the same page that Java is deviating in its APM assertion logic, I'd argue against the proposed fix in this PR. 

Agreed that Java is deviating from the current CRUD spec.  The assertion logic was written back from when the original APM tests were developed, long before the more recent change to the CRUD spec test README.

 I have two requests:

1. That we accept this change, so that we don't have to rewrite this assertion logic in the Java driver as part of the retryable writes on by default ticket.  I think the risk you identify above about other drivers deviating is not so great.  And this goes along with Bernie's suggesting to not block 4.2 work on unrelated spec changed (i.e. CRUD v2).
2. That we re-consider that section of the CRUD spec test README.  I'm concerned that the logic described there makes it too easy to write an incorrect YAML test and not have anyone notice it in review (it's harder to spot missing things than incorrect things). You can end up with a driver that includes a field that it shouldn't, perhaps with the wrong value, and still pass all the test assertions.  I can tell you anecdotally that the exact matching algorithm in the Java driver assertions has caught more than one bug in the dev cycle.

jmikola:
> All the current tests that assert on APM pass, so in practice the existing spec tests have been specifying exact command matches.

Which specs is "all the current tests" referring to? AFAIK, the APM spec is the only one with assertions for command _replies_ (f.e. [bulkWrite.yml](https://github.com/mongodb/specifications/blob/6d9e2bbb9fda70bc9e5d6795e2891a1eb04ba8d8/source/command-monitoring/tests/bulkWrite.yml#L34)) and I believe those require logic to ignore extra fields in the server response (e.g. cluster time). For outgoing command assertions in other specs I suppose it's easier to anticipate exactly what fields might be present and set by the driver; however, I would think there's still the matter of ignoring a [gossiped cluster time](https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst#gossipping-the-cluster-time).

> Agreed that Java is deviating from the current CRUD spec. The assertion logic was written back from when the original APM tests were developed, long before the more recent change to the CRUD spec test README.

My previous comment explained that this assertion logic was not introduced in the CRUD spec. The allowance of extra fields in actual documents dates back to the change stream and APM specs and I pulled that language in when adding APM expectations into the current CRUD revision.

> 1. That we accept this change, so that we don't have to rewrite this assertion logic in the Java driver as part of the retryable writes on by default ticket. [...] And this goes along with Bernie's suggesting to not block 4.2 work on unrelated spec changed (i.e. CRUD v2).

Fair point. I'd appreciate it if you could add a YAML comment above `clientOptions` in this PR explaining why this is needed, though. Something to the effect of "Some drivers may do exact matches on command_started_event, so this is needed to ensure txnNumber is absent from non-transactional writes" would suffice. I would also propose removing the later comments in each file about "This test is agnostic about retryWrites..." since that will no longer be accurate.

> I think the risk you identify above about other drivers deviating is not so great.

I'm having trouble parsing this line. Does "not so great" mean the risk of drivers deviating in their APM assertion behavior:

- [ ] is a valid concern
- [ ] is not a problem that deserves attention

> 2. That we re-consider that section of the CRUD spec test README. I'm concerned that the logic described there makes it too easy to write an incorrect YAML test and not have anyone notice it in review (it's harder to spot missing things than incorrect things).

IMO, the current behavior of allowing/ignoring extra fields is useful to test optional behavior such as drivers leaving `upsert`/`multi` options unset or `false` by default. With regard to the APM spec, the behavior has allowed us to leave most of the tests as-is despite changes to outgoing commands and their replies (e.g. cluster time, transaction numbers). If you want to open a SPEC ticket to discuss this down the line, please make a note of its origins in the change stream and APM specs as it would be a change to make across the board (not just in CRUD).

On that note, I'm curious if you'll also need to add `retryWrites=false` (and introduce a `clientOptions` field) to the APM spec tests.

source/transactions-convenient-api/tests/callback-aborts.yml
jmikola: Please add a comment explaining why this is necessary and remove the "This test is agnostic about retryWrites" comment further down in this file.

Likewise for `callback-commits.yml`.

source/transactions-convenient-api/tests/callback-commits.yml
jmikola: Please add a comment explaining why this is necessary and remove the "This test is agnostic about retryWrites" comment further down in this file.

p-mongo:
Ruby has CI configurations with and without retryable writes. Specifying retryable writes to either setting as proposed in this PR would reduce test coverage in Ruby.

If the tests do (or are meant to) pass with the retryable writes set either way, I don't see why they should be limited to one of the two settings.

jyemin:
> IMO, the current behavior of allowing/ignoring extra fields is useful to test optional behavior such as drivers leaving upsert/multi options unset or false by default. With regard to the APM spec, the behavior has allowed us to leave most of the tests as-is despite changes to outgoing commands and their replies (e.g. cluster time, transaction numbers).

I agree that there are advantages to the current behavior, but I think they are outweighed by the potential for finding more bugs with exact matching.  But we can discuss that on a future SPEC ticket

In summary: I definitely did not realize this would generate so much discussion!  

Given that the Java driver deviates from spec, I'm going to withdraw the PR and instead handle it in the Java driver's test runner.

PR #523, opened 2019-03-20
SPEC-1253: Denote retryable read test reqs with runOn

jmikola:
https://jira.mongodb.org/browse/SPEC-1253

Adopts format introduced in https://github.com/mongodb/specifications/pull/510 and https://github.com/mongodb/specifications/pull/511.

### DIFF
--- a/source/retryable-reads/tests/aggregate-serverErrors.yml
+++ b/source/retryable-reads/tests/aggregate-serverErrors.yml
@@ -1,3 +1,6 @@
+runOn:
+    - { minServerVersion: "4.0" }
+
 database_name: &database_name "retryable-reads-tests"
 collection_name: &collection_name "coll"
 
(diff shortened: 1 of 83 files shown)


source/retryable-reads/tests/aggregate-serverErrors.yml
```diff
@@ -1,3 +1,6 @@
+runOn:
+    - { minServerVersion: "4.0" }
```
ShaneHarvey: Technically all the files that use `failCommand` (all of them I believe) should be skipped on 4.0 sharded clusters so this should be:
```
runOn:
    - { minServerVersion: "4.0", topology: ["single", "replicaset"]}
    - { minServerVersion: "4.1.8" }

jmikola: Just missed this while merging :man_facepalming: 

I'll fix it in a follow-up PR.

jmikola: https://github.com/mongodb/specifications/pull/525

PR #526, opened 2019-03-25
SPEC-1194: Add tests for `postBatchResumeToken` support.

DmitryLukyanov:
As it was proposed in the ticket and discussed in the slack with @kevinAlbs, I've added only `prose` tests.
In addition to the tests which were proposed/discussed with Kevin, I also added one more common test for tracking `postBatchResumeToken`.

### DIFF
--- a/source/change-streams/tests/README.rst
+++ b/source/change-streams/tests/README.rst
@@ -158,4 +158,46 @@ The following tests have not yet been automated, but MUST still be tested
 #. The ``killCursors`` command sent during the "Resume Process" must not be allowed to throw an exception.
 #. ``$changeStream`` stage for ``ChangeStream`` against a server ``>=4.0`` that has not received any results yet MUST include a ``startAtOperationTime`` option when resuming a changestream.
 #. ``ChangeStream`` will resume after a ``killCursors`` command is issued for its child cursor.
-
+#. - For a ``ChangeStream`` under these conditions:
+      - Running against a server ``>=4.0.7``.
+      - The batch is empty or has been iterated to the last document.
+   - Expected result: 
+       - ``getResumeToken`` must return the ``postBatchResumeToken`` from the current command response.
+#. - For a ``ChangeStream`` under these conditions:
+      - Running against a server ``<4.0.7``.
+      - The batch is empty or has been iterated to the last document.
+   - Expected result: 
+      - ``getResumeToken`` must return the ``_id`` of the last document returned if one exists.
+      - ``getResumeToken`` must return ``startAfter`` from the initial aggregate if the option was specified.
+      - ``getResumeToken`` must return ``resumeAfter`` from the initial aggregate if the option was specified.
+      - If neither the ``startAfter`` nor ``resumeAfter`` options were specified, the ``getResumeToken`` result must be empty.
+#. - For a ``ChangeStream`` under these conditions:
+      - The batch is not empty.
+      - The batch has been iterated up to but not including the last element.
+   - Expected result:
+      - ``getResumeToken`` must return the ``_id`` of the previous document returned.
+#. - For a ``ChangeStream`` under these conditions:
+      - The batch is not empty.
+      - The batch hasn’t been iterated at all.
+      - Only the initial ``aggregate`` command has been executed.
+   - Expected result:
+      - ``getResumeToken`` must return ``startAfter`` from the initial aggregate if the option was specified.
+      - ``getResumeToken`` must return ``resumeAfter`` from the initial aggregate if the option was specified.
+      - If neither the ``startAfter`` nor ``resumeAfter`` options were specified, the ``getResumeToken`` result must be empty.
+#. - For a ``ChangeStream`` under these conditions:
+      - Running against a server ``>=4.0.7``.
+      - The batch is not empty.
+      - The batch hasn’t been iterated at all.
+      - The stream has iterated beyond a previous batch and a ``getMore`` command has just been executed.
+   - Expected result:
+      - ``getResumeToken`` must return the ``postBatchResumeToken`` from the previous command response.
+#. - For a ``ChangeStream`` under these conditions:
+      - Running against a server ``<4.0.7``.
+      - The batch is not empty.
+      - The batch hasn’t been iterated at all.
+      - The stream has iterated beyond a previous batch and a ``getMore`` command has just been executed.
+   - Expected result:
+      - ``getResumeToken`` must return the ``_id`` of the previous document returned if one exists.
+      - ``getResumeToken`` must return ``startAfter`` from the initial aggregate if the option was specified.
+      - ``getResumeToken`` must return ``resumeAfter`` from the initial aggregate if the option was specified.
+      - If neither the ``startAfter`` nor ``resumeAfter`` options were specified, the ``getResumeToken`` result must be empty.
--- a/source/change-streams/change-streams.rst
+++ b/source/change-streams/change-streams.rst
@@ -9,7 +9,7 @@ Change Streams
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: 3.6
-:Last Modified: January 3, 2019
+:Last Modified: April 3, 2019
 :Version: 1.6.0
 
 .. contents::
@@ -169,7 +169,7 @@ The responses to a change stream aggregate or getMore have the following structu
          id: Int64,
          firstBatch: Array<ChangeStreamDocument>,
          /**
-          * postBatchResumeToken is returned in MongoDB 4.2 and later.
+          * postBatchResumeToken is returned in MongoDB 4.0.7 and later.
           */
          postBatchResumeToken: Document
       },
@@ -187,7 +187,7 @@ The responses to a change stream aggregate or getMore have the following structu
          id: Int64,
          nextBatch: Array<ChangeStreamDocument>
          /**
-          * postBatchResumeToken is returned in MongoDB 4.2 and later.
+          * postBatchResumeToken is returned in MongoDB 4.0.7 and later.
           */
          postBatchResumeToken: Document
       },
@@ -211,7 +211,7 @@ Driver API
 
     /**
      * The most recent postBatchResumeToken returned in an aggregate or
-     * getMore response. For pre-4.2 versions of MongoDB, this remains unset.
+     * getMore response. For pre-4.0.7 versions of MongoDB, this remains unset.
      */
     private postBatchResumeToken: Document;
 
@@ -335,7 +335,7 @@ Driver API
      *
      * The server will report an error if `startAfter` and `resumeAfter` are both specified.
      *
-     * @since 4.2
+     * @since 4.0.7
      * @see https://docs.mongodb.com/master/changeStreams/#change-stream-start-after
      * @note this is an option of the `$changeStream` pipeline stage.
      */
@@ -525,9 +525,9 @@ A driver SHOULD attempt to kill the cursor on the server on which the cursor is 
 Exposing All Resume Tokens
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-:since: 4.2
+:since: 4.0.7
 
-Users can retrieve the ``documentResumeToken`` by inspecting the _id on each ``ChangeDocument``. But since MongoDB 4.2, aggregate and getMore responses also include a ``postBatchResumeToken``. Drivers use one or the other when automatically resuming, as described in `Resume Process`_.
+Users can retrieve the ``documentResumeToken`` by inspecting the _id on each ``ChangeDocument``. But since MongoDB 4.0.7, aggregate and getMore responses also include a ``postBatchResumeToken``. Drivers use one or the other when automatically resuming, as described in `Resume Process`_.
 
 Drivers MUST expose a mechanism to retrieve the same resume token that would be used to automatically resume. It MUST be possible to use this mechanism after iterating every document. It MUST be possible for users to use this mechanism periodically even when no documents are getting returned (i.e. ``getMore`` has returned empty batches). Drivers have two options to implement this.
 
@@ -774,4 +774,7 @@ Changelog
 | 2018-11-06 | Added handling of ``postBatchResumeToken``.                |
 +------------+------------------------------------------------------------+
 | 2019-01-10 | Clarified error handling for killing the cursor.           |
++------------+------------------------------------------------------------+
+| 2019-04-03 | Updated the lowest server version that supports            |
+|            | ``postBatchResumeToken``.                                  |
 +------------+------------------------------------------------------------+


source/change-streams/tests/README.rst
```diff
@@ -158,4 +158,7 @@ The following tests have not yet been automated, but MUST still be tested
 #. The ``killCursors`` command sent during the "Resume Process" must not be allowed to throw an exception.
 #. ``$changeStream`` stage for ``ChangeStream`` against a server ``>=4.0`` that has not received any results yet MUST include a ``startAtOperationTime`` option when resuming a changestream.
 #. ``ChangeStream`` will resume after a ``killCursors`` command is issued for its child cursor.
-
+#. ``ChangeStream`` against a server ``>=4.2`` must track the latest ``postBatchResumeToken`` included in an ``aggregate`` or ``getMore`` response if available.
+#. ``ChangeStream`` against a server ``>=4.2`` must resume with the ``resumeAfter`` set to ``_id`` of the most recent document returned when the cursor is in the middle of a batch.
```
jmikola: This may also be a question for @kevinAlbs, but why would we ever resume before 

A batch consists of zero or more documents returned by the preceding `aggregate` or `getMore` command. Given that a resumable error occurs when a `getMore` operation fails, why or how would the driver ever resume in the _middle_ of a batch (i.e. while iterating locally cached documents)? Based on [Resume Process](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#resume-process), `postBatchResumeToken` always takes priority over other resume tokens or `startAtOperationTime`. [Response Format](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#response-format) also suggests that a 4.2+ response will always include `postBatchResumeToken`.

I think the intention here is that we should test that the [resume token accessor method](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#exposing-all-resume-tokens) returns the token we expect. Consider the following:

 * First batch from `aggregate` response
  * If the batch is empty or we've iterated to the last element, the accessor should return the `postBatchResumeToken` for a 4.2+ server and fall back to `startAfter` or `resumeAfter` options for pre-4.2 servers. In the absense of all of those options, there is nothing to return (while `startAtOperationTime` is useful for resuming, it's not applicable for this method).
  * If the batch is not empty but we haven't iterated at all, I don't think the spec is clear about what we should return. Do we peak at the first document in the batch and extract its document resume token (i.e. `_id`)?
  * If the batch is not empty and we've iterated up to but _not_ including the last element, return the cached document resume token
 * Second or later batch from `getMore` response
 * If the batch is empty, we've iterated to the last element, or it's not empty and we've iterated up to but _not_ including the last element, follow the corresponding rule as above
 *  If the batch is not empty but we haven't iterated at all, I think we're supposed to return the `postBatchResumeToken` from the previous command response (falling back to fall back to `startAfter` or `resumeAfter` for pre-4.2 servers), per the first bullet under `aggregate` above.

Does this make sense? /cc: @kevinAlbs

source/change-streams/tests/README.rst
```diff
@@ -158,4 +158,7 @@ The following tests have not yet been automated, but MUST still be tested
...
+#. ``ChangeStream`` against a server ``>=4.2`` must resume ``resumeAfter`` set to the ``postBatchResumeToken`` after returning all documents in a batch.
+#. ``ChangeStream`` against a server ``>=4.2`` must resume ``resumeAfter`` set to the ``postBatchResumeToken`` after returning an empty batch.
```
jmikola: These last two tests look OK, although I imagine a complete test of the resume token accessor might also accomplish this provided that drivers are also testing that the accessor's return value always agrees with what would be specified in the `resumeAfter` option for a resuming `aggregate` command's `$changeStream` stage.

source/change-streams/tests/README.rst
kevinAlbs: Agree with all of your points. We can't test resuming in the middle of iterating a batch because there's no way to cause the driver to resume then. We can only check the resume token returned by getResumeToken (or resume token event).

> Response Format also suggests that a 4.2+ response will always include postBatchResumeToken

Yes, 4.2+ change streams will always include a postBatchResumeToken. From an email sent by Bernard with subject "Change Stream High Water Mark Upgrade/Downgrade & Drivers Summary":

> Batches returned by change streams from 4.0.7+ or 4.2 will always include a postBatchResumeToken field within the cursor object, which is guaranteed to be equal to or later than the resume token of the final document in the batch.

> If the batch is not empty but we haven't iterated at all, I don't think the spec is clear about what we should return. Do we peak at the first document in the batch and extract its document resume token (i.e. _id)?

I created https://jira.mongodb.org/browse/SPEC-1263 to clarify this (though I think this can be done after tests). In that case, we should return the original `startAfter` or `resumeAfter` or an unset optional.

DmitryLukyanov: Done

source/change-streams/tests/README.rst
DmitryLukyanov: Done

DmitryLukyanov:
NOTE: The draft format for test cases was discussed with @kevinAlbs  in slack, but I wasn't able to implement exactly proposed markup, because of `rst` markup restrictions. Not sure that the implemented format is suitable here. If not so, I will change the format.

source/change-streams/tests/README.rst
```diff
@@ -158,4 +158,41 @@ The following tests have not yet been automated, but MUST still be tested
...
+#. - For a ``ChangeStream`` under these conditions:
+      - running against a server ``>=4.2``.
+      - the batch is empty or has been iterated to the last document.
+   - Expected result: 
+       - The ``getResumeToken`` must return the ``postBatchResumeToken`` from the current command response.
+#. - For a ``ChangeStream`` under these conditions:
+      - running against a server ``<4.2``.
+      - the batch is empty or has been iterated to the last document.
+   - Expected result: 
+       - The ``getResumeToken`` must return the ``startAfter`` or ``resumeAfter`` from the current command request.
```
kevinAlbs: Suggest changing "from the current command request" to  "from the initial aggregate"

source/change-streams/tests/README.rst
```diff
@@ -158,4 +158,41 @@ The following tests have not yet been automated, but MUST still be tested
...
+      - executing a ``getMore`` command.
+   - Expected result:
+      - The ``getResumeToken`` must return the ``postBatchResumeToken`` from the previous command response.
+#. - For a ``ChangeStream`` under these conditions:
+      - running against a server ``<4.2``.
+      - the batch is not empty.
+      - the batch hasn’t been iterated at all.
+      - executing a ``getMore`` command.
+   - Expected result:
+      - The ``getResumeToken`` must return the ``startAfter`` or ``resumeAfter`` from the previous command request.
```
kevinAlbs: In this case, we should use the _id of the last document returned (from the previous batch). If there are no previous documents returned, then we return ``startAfter``/``resumeAfter`` of the initial aggregate. And if that wasn't specified, we return nothing.

source/change-streams/tests/README.rst
```diff
@@ -158,4 +158,41 @@ The following tests have not yet been automated, but MUST still be tested
...
+      - The batch is not empty.
+      - The batch has been iterated up to but not including the last element.
+   - Expected result:
+      - The ``getResumeToken`` must return the last cached document resume token.
+#. - For a ``ChangeStream`` under these conditions:
+      - The batch is not empty.
+      - The batch hasn’t been iterated at all.
+      - Executing an ``aggregate`` command.
+   - Expected result:
+      - The ``getResumeToken`` must return the ``startAfter`` or ``resumeAfter`` from the initial ``aggregate`` request.
```
jmikola: Same suggestion as my previous suggestion about splitting this into three bullets and capturing the precedence of `startAfter`, then `resumeAfter`, and finally an empty return value.

source/change-streams/tests/README.rst
```diff
@@ -158,4 +158,41 @@ The following tests have not yet been automated, but MUST still be tested
...
+   - Expected result:
+      - The ``getResumeToken`` must return the ``postBatchResumeToken`` from the previous command response.
+#. - For a ``ChangeStream`` under these conditions:
+      - Running against a server ``<4.2``.
+      - The batch is not empty.
+      - The batch hasn’t been iterated at all.
+      - Executing a ``getMore`` command.
+   - Expected result:
+      - The ``getResumeToken`` must return the ``_id`` of the previous document returned.
+      - If there are no previous documents returned, then we return startAfter/resumeAfter of the initial aggregate.  if that wasn't specified, we return nothing.
```
jmikola: Same suggestion to use the three bullets here to capture the precedence of `startAfter`, `resumeAfter`, and an empty return value.

source/change-streams/tests/README.rst
DmitryLukyanov: Done

source/change-streams/tests/README.rst
```diff
@@ -158,4 +158,41 @@ The following tests have not yet been automated, but MUST still be tested
...
+      - The batch is not empty.
+      - The batch hasn’t been iterated at all.
+      - Executing an ``aggregate`` command.
+   - Expected result:
+      - The ``getResumeToken`` must return the ``startAfter`` or ``resumeAfter`` from the initial ``aggregate`` request.
+#. - For a ``ChangeStream`` under these conditions:
+      - Running against a server ``>=4.2``.
+      - The batch is not empty.
+      - The batch hasn’t been iterated at all.
+      - Executing a ``getMore`` command.
```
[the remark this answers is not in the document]
DmitryLukyanov: Done

source/change-streams/tests/README.rst
```diff
@@ -158,4 +158,41 @@ The following tests have not yet been automated, but MUST still be tested
...
+   - Expected result:
+      - The ``getResumeToken`` must return the ``postBatchResumeToken`` from the previous command response.
+#. - For a ``ChangeStream`` under these conditions:
+      - Running against a server ``<4.2``.
+      - The batch is not empty.
+      - The batch hasn’t been iterated at all.
+      - Executing a ``getMore`` command.
```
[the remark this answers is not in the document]
DmitryLukyanov: Done

source/change-streams/tests/README.rst
DmitryLukyanov: Done

source/change-streams/tests/README.rst
```diff
@@ -158,4 +158,41 @@ The following tests have not yet been automated, but MUST still be tested
...
+       - If ``startAfter`` and ``resumeAfter`` are absent, ``getResumeToken`` result must be empty.
+#. - For a ``ChangeStream`` under these conditions:
+      - The batch is not empty.
+      - The batch has been iterated up to but not including the last element.
+   - Expected result:
+      - The ``getResumeToken`` must return the last cached document resume token.
+#. - For a ``ChangeStream`` under these conditions:
+      - The batch is not empty.
+      - The batch hasn’t been iterated at all.
+      - Executing an ``aggregate`` command.
```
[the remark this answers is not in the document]
DmitryLukyanov: Yep, it sounds better.
Done

source/change-streams/tests/README.rst
```diff
@@ -158,4 +158,41 @@ The following tests have not yet been automated, but MUST still be tested
...
+#. - For a ``ChangeStream`` under these conditions:
+      - Running against a server ``>=4.2``.
+      - The batch is empty or has been iterated to the last document.
+   - Expected result: 
+       - The ``getResumeToken`` must return the ``postBatchResumeToken`` from the current command response.
+#. - For a ``ChangeStream`` under these conditions:
+      - Running against a server ``<4.2``.
+      - The batch is empty or has been iterated to the last document.
+   - Expected result: 
+       - The ``getResumeToken`` must return the ``startAfter`` or ``resumeAfter`` from the initial aggregate.
```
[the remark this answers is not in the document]
DmitryLukyanov: Done

source/change-streams/tests/README.rst
```diff
@@ -158,4 +158,46 @@ The following tests have not yet been automated, but MUST still be tested
...
+      - The batch is empty or has been iterated to the last document.
+   - Expected result: 
+      - ``getResumeToken`` must return ``startAfter`` from the initial aggregate if the option was specified.
+      - ``getResumeToken`` must return ``resumeAfter`` from the initial aggregate if the option was specified.
+      - If neither the ``startAfter`` nor ``resumeAfter`` options were specified, the ``getResumeToken`` result must be empty.
+#. - For a ``ChangeStream`` under these conditions:
+      - The batch is not empty.
+      - The batch has been iterated up to but not including the last element.
+   - Expected result:
+      - ``getResumeToken`` must return the last cached document resume token.
```
kevinAlbs: The test below phrases this like "must return the ``_id`` of the previous document returned". Make these consistent.

source/change-streams/tests/README.rst
```diff
@@ -158,4 +158,46 @@ The following tests have not yet been automated, but MUST still be tested
...
+#. ``ChangeStream`` against a server ``>=4.2`` must track the latest ``postBatchResumeToken`` included in an ``aggregate`` or ``getMore`` response.
+#. - For a ``ChangeStream`` under these conditions:
+      - Running against a server ``>=4.2``.
+      - The batch is empty or has been iterated to the last document.
+   - Expected result: 
+       - ``getResumeToken`` must return the ``postBatchResumeToken`` from the current command response.
+#. - For a ``ChangeStream`` under these conditions:
+      - Running against a server ``<4.2``.
+      - The batch is empty or has been iterated to the last document.
+   - Expected result:
```
kevinAlbs: The first expected result should be the _id of the last document returned (if one exists).

source/change-streams/tests/README.rst
```diff
@@ -158,4 +158,46 @@ The following tests have not yet been automated, but MUST still be tested
...
+      - The stream has iterated beyond a previous batch and a ``getMore`` command has just been executed.
+   - Expected result:
+      - ``getResumeToken`` must return the ``postBatchResumeToken`` from the previous command response.
+#. - For a ``ChangeStream`` under these conditions:
+      - Running against a server ``<4.2``.
+      - The batch is not empty.
+      - The batch hasn’t been iterated at all.
+      - The stream has iterated beyond a previous batch and a ``getMore`` command has just been executed.
+   - Expected result:
+      - ``getResumeToken`` must return the ``_id`` of the previous document returned.
```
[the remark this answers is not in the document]
DmitryLukyanov: Done

source/change-streams/tests/README.rst
```diff
@@ -158,4 +158,46 @@ The following tests have not yet been automated, but MUST still be tested
 #. The ``killCursors`` command sent during the "Resume Process" must not be allowed to throw an exception.
 #. ``$changeStream`` stage for ``ChangeStream`` against a server ``>=4.0`` that has not received any results yet MUST include a ``startAtOperationTime`` option when resuming a changestream.
 #. ``ChangeStream`` will resume after a ``killCursors`` command is issued for its child cursor.
-
+#. ``ChangeStream`` against a server ``>=4.2`` must track the latest ``postBatchResumeToken`` included in an ``aggregate`` or ``getMore`` response.
```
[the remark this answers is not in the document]
DmitryLukyanov: Done

source/change-streams/tests/README.rst
DmitryLukyanov: Done

source/change-streams/tests/README.rst
DmitryLukyanov: Done

IanWhalen:
@jyemin and @mbroadst looks like you're on this review but have no comments?  You OK to pull yourself off or LGTM?

Bumping this to a target of next week in the meantime.

source/change-streams/tests/README.rst
```diff
@@ -158,4 +158,46 @@ The following tests have not yet been automated, but MUST still be tested
...
-
+#. - For a ``ChangeStream`` under these conditions:
+      - Running against a server ``>=4.2``.
+      - The batch is empty or has been iterated to the last document.
+   - Expected result: 
+       - ``getResumeToken`` must return the ``postBatchResumeToken`` from the current command response.
+#. - For a ``ChangeStream`` under these conditions:
+      - Running against a server ``<4.2``.
+      - The batch is empty or has been iterated to the last document.
+   - Expected result:
```
jyemin: I suspect driver's will fail this test against a 4.0.7+ server.  From Kevin's quote of Bernard:

> Batches returned by change streams from 4.0.7+ or 4.2 will always include a postBatchResumeToken field

Since a driver is likely to just look for a postBatchResumeToken field in the response regardless of server version, this test will likely fail, as the driver will behave exactly as it would when connected to a 4.2 server.  

But this is exactly what we want: since the server team backported the change to 4.0, drivers might as take advantage of this feature with a 4.0.7+ server as well.

@kevinAlbs what do you think?  Am I missing something or should we change the change streams spec to say that postBatchResumeToken is 4.0.7+?

kevinAlbs: Good point. Since the server is backporting, I agree we should change the spec to say that postBatchResumeToken is 4.0.7+. For the sake of these tests, I think we can just change the conditions from ">=4.2" and "<4.2" to ">=4.0.7" and "<4.0.7". There are other references in the spec that reference postBatchResumeToken being introduced in 4.2. I can submit a PR to update those.

DmitryLukyanov: Done

jyemin: @DmitryLukyanov, go ahead and make the change in change-streams.rst to go along with this change?  One less PR to review separately, and this will keep the two changes together.

DmitryLukyanov: Done

jyemin:
@DmitryLukyanov, try to avoid force pushing on a PR. Makes it confusing for reviewers.

source/change-streams/tests/README.rst
jmikola: Changing 4.2+ to 4.07+ will only require us to check the server version in the test suite, correct? The implementation will still just pick up the field if it's available.

kevinAlbs: Yeah, the implementation shouldn't do version checks, just check if the `postBatchResumeToken` is part of the response or not.

PR #528, opened 2019-03-29
SPEC-1214 Add `NonRetryableChangeStreamError` label to resumable error definition

kvwalker:


ShaneHarvey:
Since there server now returns the label for [`ChangeStreamFatalError` and `ChangeStreamHistoryLost` errors](https://github.com/mongodb/mongo/blob/a5138072e1bb731bbbc8612c21d4bb9c4cc98270/src/mongo/base/error_codes.err#L354), is there any way we any add a spec test for this? Perhaps a test that projects out the `_id` on a >=4.1.10 sharded cluster in [tests/change-streams-errors.yml](https://github.com/mongodb/specifications/blob/master/source/change-streams/tests/change-streams-errors.yml)?

source/change-streams/change-streams.rst
```diff
@@ -47,6 +47,7 @@ An error is considered resumable if it meets any of the following criteria:
   network error)
 
 - *any* server error response from a getMore command excluding those
+  containing the error label `NonRetryableChangeStreamError` and those
```
jmikola: @ShaneHarvey: Is this something you intended to test? IIRC, we the transaction tests assert error labels after a `failCommand` fail point but I'm not sure that we can configure the fail point to return a specific error code. In the case of the txn tests, I think we're relying on existing driver or server behavior to apply an error label for an error code.

ShaneHarvey: The server added this new error label in https://jira.mongodb.org/browse/SERVER-39675 so we can write a spec test against >=4.1.10, see [my comment above](https://github.com/mongodb/specifications/pull/528#issuecomment-478098724).

source/change-streams/tests/change-streams-errors.yml
```diff
@@ -50,4 +50,27 @@ tests:
...
+    changeStreamOptions: {}
+    operations:
+      -
+        database: *database_name
+        collection: *collection_name
+        name: insertOne
+        arguments:
+          document:
+            z: 3
+    expectations: []
```
jmikola: https://github.com/mongodb/specifications/tree/master/source/change-streams/tests#spec-test-format suggests this field is optional. If we're not going to include any CommandStartedEvents here, I'd suggest just omitting the field entirely lest some test runner interpret this as "no events should be expected".

source/change-streams/tests/change-streams-errors.yml
```diff
@@ -50,4 +50,26 @@ tests:
...
       error:
-        code: 40324
\ No newline at end of file
+        code: 40324
+  -
+    description: Change Stream should error when _id is projected out
+    minServerVersion: "4.1.11"
+    target: collection
+    topology:
+      - replicaset
```
ShaneHarvey: Can you also add "sharded"?

PR #530, opened 2019-04-02
SPEC-1211: Modify change stream resume token test to only run if wire version is < 8

kvwalker:


samantharitter:
LGTM!  Will merge.

samantharitter:
Thank you for adding this @kvwalker  :)

PR #531, opened 2019-04-04
SPEC-1125 Test collection unack'd write concern is compatible with transactions

ShaneHarvey:
Jira: https://jira.mongodb.org/browse/SPEC-1125

These tests prove that a collection configured with WriteConcern(w=0) can still be used within a transaction. I added tests for all the supported CRUD methods for completeness.

Passing python driver patch: https://evergreen.mongodb.com/version/5ca66d46e3c33119ed332b5c

### DIFF
--- a/source/transactions/tests/write-concern.yml
+++ b/source/transactions/tests/write-concern.yml
@@ -43,13 +43,14 @@ tests:
             documents:
               - _id: 1
             ordered: true
-            readConcern:
-            lsid: session0
-            txnNumber:
-              $numberLong: "1"
-            startTransaction: true
-            autocommit: false
-            writeConcern:
+            <<: &transactionCommandArgs
+              lsid: session0
+              txnNumber:
+                $numberLong: "1"
+              startTransaction: true
+              autocommit: false
+              readConcern:
+              writeConcern:
           command_name: insert
           database_name: *database_name
       - command_started_event:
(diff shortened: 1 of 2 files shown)


source/transactions/tests/write-concern.yml
```diff
@@ -242,3 +247,401 @@ tests:
...
+
+  - description: unacknowledged write concern coll insertOne
+
+    operations:
+      - *startTransaction
+      - name: insertOne
+        object: collection
+        collectionOptions:
+          writeConcern:
+            w: 0
```
jmikola: Would it make sense to use [Extend Inline](https://blog.daemonl.com/2016/02/yaml.html) to reduce some duplication here?

```yaml
operations:
  - *startTransaction
  - name: insertOne
    <<: &collection_w0
      object: collection
      collectionOptions:
        writeConcern: { w: 0 }
    arguments: ...
```

And later you can just merge in `*collection_w0`:

```yaml
operations:
  - *startTransaction
  - name: insertOne
    <<: *collection_w0
    arguments: ...
```

ShaneHarvey: Perfect, done.

ShaneHarvey:
> Unless there's a concern that some drivers might be eagerly prohibiting it when startSession() is called, in which case we can punt on that to a future SPEC ticket.

Yeah the python driver does exactly that so we can't conform such a spec test.

source/transactions/tests/write-concern.yml
```diff
@@ -242,3 +230,327 @@ tests:
         result:
           # Client-side error.
           errorContains: transactions do not support unacknowledged write concern
+
+  - description: unacknowledged write concern coll insertOne
+
+    operations:
+      - *startTransaction
+      - name: insertOne
+        object: collection
```
jmikola: This is redundant. `<<` below already injects `object: collection`.

ShaneHarvey: Done.

PR #532, opened 2019-04-05
SPEC-1270 URI database does not configure authentication

xdg:


source/auth/auth.rst
```diff
@@ -87,6 +87,14 @@ mechanism (string)
...
 	* Includes additional properties for the given mechanism.
 
+Ambiguity between authentication source and URI database
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The presence of a database name in the URI connection string MUST NOT be
+interpreted as a user configuring authentication credentials.  The URI database
+name is only used as a default source for some mechanisms when authentication
+has been configured and a source is required but has not been specified.  See
+individual mechanism definitions for details.
```
jmikola: This is a new MUST NOT addition and not a simple grammar fix. I think it warrants a change log entry and bumping the last modified date. If the spec has a version number, I'd also suggest incrementing the patch version (not minor, since this isn't introducing new functionality/API).

source/connection-string/connection-string-spec.rst
```diff
@@ -131,6 +131,8 @@ The database to authenticate against. If provided it is everything after the Hos
 
 The following characters MUST NOT appear in the database name, once it has been decoded: slash ("/"), backslash ("\\"), space (" "), double-quote ("""), or dollar sign ("$"). The MongoDB Manual `says that <https://docs.mongodb.com/manual/reference/limits/#Restrictions-on-Field-Names>`_ period (".") is also prohibited, but drivers MAY allow periods in order to express a namespace (database and collection name, perhaps containing multiple periods) in this part of the URL.
 
+The presence of the auth database component without other credential data such as Userinfo or authentication parameters in connection options MUST NOT be interpreted as a request for authentication.
+
```
jmikola: Similar spec updates as noted for auth above.

source/auth/auth.rst
```diff
@@ -87,6 +87,14 @@ mechanism (string)
 mechanism_properties
 	* Includes additional properties for the given mechanism.
 
+Ambiguity between authentication source and URI database
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The presence of a database name in the URI connection string MUST NOT be
+interpreted as a user configuring authentication credentials.  The URI database
```
jmikola: Does this warrant a new spec test in https://github.com/mongodb/specifications/blob/master/source/auth/tests/connection-string.yml? 

I'll note that [`auth/tests`](https://github.com/mongodb/specifications/tree/master/source/auth/tests) doesn't have a README but I assume it uses the format in [`connection-string/tests/README.rst`](https://github.com/mongodb/specifications/blob/master/source/connection-string/tests/README.rst). The connection string's test README doesn't explicitly say that `auth` may be null (although it clearly is in many tests) and we also don't have any tests for soliciting a default database name without UserInfo specified.

I think a test like the following could accomplish this:

```yaml
    -
        description: "default database without credentials does not infer auth"
        uri: "mongodb://localhost/test"
        hosts: ~
        valid: true
        warning: false
        auth: ~
        options: ~
```

Not sure if this belongs in `auth`, `connection-string`, or both.

source/auth/auth.rst
xdg: Done in 3fffcae

source/connection-string/connection-string-spec.rst
xdg: Done in 3fffcae

source/auth/auth.rst
xdg: Done in 3fffcae

xdg: The connection string test README says that null value for a key like `auth` means no assertion is necessary, so that test wouldn't verify anything.  :-(

source/uri-options/uri-options.rst
```diff
@@ -88,8 +88,9 @@ pertaining to URI options apply here.
    * - authMechanism
      - any string; valid values are defined in the `auth spec 
        <https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#supported-authentication-methods>`_
-     - defined in `auth spec 
-       <https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#supported-authentication-methods>`_
+     - None; default values for authentication exist for constructing authentication credentials per the
+       `auth spec <https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#supported-authentication-methods>`_,
+       but must not be assigned here to distinguish between values specified by the user and defaulted by the implementation.
```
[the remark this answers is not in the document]
xdg: I think the tests are a bit vague about what the 'auth' field is supposed to mean.  If it's just "stuff I found in the string".  There's no test for `authSource` in valid-auth.yml and no where for it to go in the 'auth' field.

I'm hesitant to revise the test format and make work for everyone.  If the current format works and we're just clarifying, then maybe that's enough.  What do you think?

source/uri-options/uri-options.rst
```diff
@@ -102,9 +103,9 @@ pertaining to URI options apply here.
 
    * - authSource
      - any string
-     - If the auth mechanism is PLAIN, GSSAPI, or MONGODB-X509, the default value is "$external"
-
-       Otherwise, the value should fall back on the database specified in the URI, or "admin" if it is not specified
+     - None; default values for authentication exist for constructing authentication credentials per the
```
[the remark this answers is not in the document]
xdg: I'll see where it best fits.  We now have tests scattered across three specs, with similar format but (I fear) slightly different semantics.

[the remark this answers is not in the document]
xdg: Could you confirm what you mean by "valid"?   I think you mean that providing any 'auth*' option  counts as "configuring a Credential" and without username/password (assuming not X509), the Credential is invalid.

source/uri-options/uri-options.rst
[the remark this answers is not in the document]
jyemin: The problem I have with this clause of the sentence is that it introduce a distinction between two  undefined terms: "here" and "the implementation" with no guidance about what that actually means.  Then it uses a lower-case "must".  And there are no tests to clarify the distinction.

I suggest just striking everything after the last comma.

source/uri-options/uri-options.rst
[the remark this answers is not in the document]
jyemin: Valid according to the `valid` field in the tests.  So `valid: false` means the driver should raise an error.

source/uri-options/uri-options.rst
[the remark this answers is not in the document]
xdg: The more I've considered this, I don't think the *auth* tests are sufficient in their current structure, so I'm experimenting with making them more specific to translating from a URI to an auth credential to provide a stronger set of assertions.  I think the connection string and URI options tests can remain as is as tests of pure parsing output.  I'll post a new commit when I finish my experiment.

source/uri-options/uri-options.rst
```diff
@@ -88,8 +88,9 @@ pertaining to URI options apply here.
    * - authMechanism
      - any string; valid values are defined in the `auth spec 
        <https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#supported-authentication-methods>`_
-     - defined in `auth spec 
-       <https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#supported-authentication-methods>`_
+     - None; default values for authentication exist for constructing authentication credentials per the
```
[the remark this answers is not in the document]
xdg: This test already exists.

source/uri-options/uri-options.rst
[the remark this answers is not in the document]
xdg: Added in new commits.

source/uri-options/uri-options.rst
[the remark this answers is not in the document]
xdg: I reworded the text in the latest commit to avoid vague "must not".  Now it's just a statement to reiterate there is no default.  The URI options tests aren't expressive enough to test it, but it's effectively tested by the new auth tests that translate from URI to credential object.

xdg:
@jyemin @jmikola @behackett, PTAL

source/auth/tests/README.rst
```diff
@@ -0,0 +1,52 @@
...
+Format
+------
+
+Each YAML file contains an object with a single ``tests`` key. This key is an
+array of test case objects, each of which have the following keys:
+
+- ``description``: A string describing the test.
+- ``uri``: A string containing the URI to be parsed.
+- ``valid:`` A boolean indicating if the URI should be considered valid.
+- ``credential_configured:`` A boolean indicating if the URI is considered to have
```
jmikola: I suppose this addresses my original question in https://github.com/mongodb/specifications/pull/532/files#r273202088.

source/auth/tests/connection-string.yml
```diff
@@ -1,367 +1,333 @@
 tests:
     -
         description: "should use the default source and mechanism"
-        uri: "mongodb://user:password@localhost"
-        hosts: ~
+        uri: "mongosource://user:password@localhost"
```
jmikola: What is the reason for changing the scheme to `mongosource` in these tests? AFAIK only `mongodb` and `mongodb+srv` are supported so I would expect this to result in an invalid URI.

Was this a side effect of replacing `db:` with `source:` in the `credential` object?

xdg: 😂 Yes!  Will fix.

xdg:
As a sanity check, I'm also doing a POC for Perl. Will have updates tonight or tomorrow.

source/auth/tests/README.rst
```diff
@@ -0,0 +1,55 @@
...
+  configured a credential for the purposes of deciding if the driver should
+  authenticate to the topology.
+- ``credential``: An object containing the following keys:
+
+  - ``username``: A string containing the username. For auth mechanisms
+    that do not utilize a password, this may be the entire ``userinfo`` token
+    from the connection string.  If this is the empty string,
+    test harnesses should interpret it as unspecified.
+  - ``password``: A string containing the password.  If this is the empty string,
```
jyemin: I prefer null to the empty string for non-associating values.

source/auth/tests/README.rst
```diff
@@ -0,0 +1,55 @@
...
+- ``uri``: A string containing the URI to be parsed.
+- ``valid:`` A boolean indicating if the URI should be considered valid.
+- ``credential_configured:`` A boolean indicating if the URI is considered to have
+  configured a credential for the purposes of deciding if the driver should
+  authenticate to the topology.
+- ``credential``: An object containing the following keys:
+
+  - ``username``: A string containing the username. For auth mechanisms
+    that do not utilize a password, this may be the entire ``userinfo`` token
+    from the connection string.  If this is the empty string,
```
jyemin: I prefer null to the empty string for non-associating values.

source/auth/tests/README.rst
```diff
@@ -0,0 +1,55 @@
...
+    test harnesses should interpret it as unspecified.
+  - ``password``: A string containing the password.  If this is the empty string,
+    test harnesses should interpret it as unspecified.
+  - ``source``: A string containing the authentication database.
+  - ``mechanism``: A string containing the authentication mechanism.  The string "default"
```
jyemin: I prefer null to a magic string for non-associating values.

source/auth/tests/connection-string.yml
```diff
@@ -2,366 +2,336 @@ tests:
...
             username: "user@DOMAIN.COM"
-            password: ~
-            db: "$external"
-        options:
-            authmechanism: "GSSAPI"
+            password: ""
+            source: "$external"
+            mechanism: "GSSAPI"
+            mechanism_properties:
+                SERVICE_NAME: "mongodb"
```
jyemin: Same

source/auth/tests/connection-string.yml
```diff
@@ -2,366 +2,336 @@ tests:
...
```
jyemin: Same

source/auth/tests/connection-string.yml
```diff
@@ -2,366 +2,336 @@ tests:
...
```
jyemin: Remove, as this mechanism property is not configured in the connection string.

source/auth/tests/connection-string.yml
```diff
@@ -2,366 +2,336 @@ tests:
...
-            password: ~
-            db: "$external"
-        options:
-          authmechanism: "GSSAPI"
-          authmechanismproperties:
-            SERVICE_NAME: "other"
+            source: "$external"
+            mechanism: "GSSAPI"
+            mechanism_properties:
+                SERVICE_NAME: "mongodb"
```
jyemin: Same

source/auth/tests/README.rst
xdg: Yeah, it's clunky.  We already use null to mean "no assertion necessary" -- meaning any value or none at all is allowed -- so I was trying to succinctly say "required to be missing".   I'm open to other suggestions... false?  A separate "password_set" field?

jyemin: Where do we use `null` to mean any value or none at all?  In command monitoring expectations, we use an explicit `null` to mean that the value should not be present, which seems similar enough to this use case. In this context, it wouldn't make sense for `password: null` to mean "any value or none at all", so I'm fine with using `null` as the non-associating value here.  @jmikola any thoughts on the matter?

 @jmikola what do you think?

xdg: Connection string tests say `If a test case includes a null value for one of these keys (e.g. auth: ~, port: ~), no assertion is necessary. ` and this was modeled on that and has a similar line.  

Another option would be to remove keys that don't need assertions (e.g. if valid if false, no need for a credentials field), and use null to mean a value that must not be set.  I was trying to deviate less from existing test runners, but maybe that was the wrong goal versus test clarity.

source/auth/tests/README.rst
```diff
@@ -0,0 +1,55 @@
...
+Format
+------
+
+Each YAML file contains an object with a single ``tests`` key. This key is an
+array of test case objects, each of which have the following keys:
+
+- ``description``: A string describing the test.
+- ``uri``: A string containing the URI to be parsed.
+- ``valid:`` A boolean indicating if the URI should be considered valid.
+- ``credential_configured:`` A boolean indicating if the URI is considered to have
```
mbroadst: I had wondered about this given the previous format, but is there a reason we can't just set `auth` (or `credential` now) to `null` in the event that no credential should be generated? I'm not sure I see why we need another field?

xdg: Jeff seems to want to use null for "should not exist", so if we make that change for elements in credential, we can do it at the top level, too.

xdg:
Based on feedback, I'm going to revise the test cases as follows:
* "null" – field must not appear or must not be configured
* "missing" – we're making no assertion about whether this field exists or not or is configured or not

source/auth/tests/README.rst
jmikola: This is due to legacy decisions to have `null` imply "ignore this assertion". An alternative might be to amend the grammar to allow `credential` to be one of `null` (ignore), boolean `false`, or a document.

xdg:
@jmikola @mbroadst @jyemin Updated with use of null/missing as described.

source/auth/tests/README.rst
jyemin: Please resolve this one too.  Now that we're using null to indicate that a value is not set, this should be `mechanism: null` instead of `mechanism: "default"`

source/auth/tests/connection-string.yml
jyemin: Are you thinking that the connection string should be aware of the default values for mechanism properties?  If so, let's discuss.  If not, please remove from here and below.

xdg: This is not a test of connection string parsing alone.  It's a test of the auth spec itself, which indicates default values for attributes of a Credential object.  Each test is a series of assertions: given a URI, what auth *Credential object* is constructed.

This is why, for instance, not specifying a database name or authSource in the URI results in a "source" of "admin" up on line 10 (of this diff anyway) – that is the default value per the Auth spec.  This particular test asserts that a Credential has the correct default SERVICE_NAME property if GSSAPI is selected and no SERVICE_NAME is provided.

source/auth/tests/README.rst
xdg: Will do.

source/auth/tests/connection-string.yml
mbroadst: I think this is fine as is

jyemin: Yep, you're right.

source/auth/auth.rst
```diff
@@ -878,6 +886,13 @@ Q: Why does SCRAM sometimes SASLprep and sometimes not?
 Version History
 ===============
 
+Version 1.8.0 Changes
+    * Test format changed to improve specificity of behavior assertions.
+
+Version 1.7.2 Changes
```
jmikola: Previous version of the spec is 1.7.1. Was 1.7.2 briefly a version during the course of this PR?

Any reason not to just roll these changes into the list for 1.8.0?

source/auth/tests/connection-string.yml
jmikola: I think this can be resolved in light of https://github.com/mongodb/specifications/pull/532/files#r278946134

source/auth/tests/connection-string.yml
jmikola: I think this can be resolved in light of https://github.com/mongodb/specifications/pull/532/files#r278946134

PR #533, opened 2019-04-09
SPEC-1263: Clarify Exposing All Resume Tokens

kvwalker:
https://jira.mongodb.org/browse/SPEC-1263


source/change-streams/change-streams.rst
```diff
@@ -483,27 +483,8 @@ Once a ``ChangeStream`` has encountered a resumable error, it MUST attempt to re
...
-  - The driver MUST NOT set ``startAfter``. The driver MUST remove it from the original aggregation command.
-  - The driver MUST NOT set ``startAtOperationTime``.
-- Else if the ``ChangeStream`` was created with ``resumeAfter``:
-
-  - The driver MUST set ``resumeAfter`` to the value of the originally used ``resumeAfter``.
-  - The driver MUST NOT set ``startAfter``.
-  - The driver MUST NOT set ``startAtOperationTime``.
-- Else if the ``ChangeStream`` has a saved operation time (either from an originally specified ``startAtOperationTime`` or saved from the original aggregation) and the max wire version is >= ``7``:
+- Call ``getResumeToken()`` to receive the resume token.
+- If the ``ChangeStream`` has a saved operation time (either from an originally specified ``startAtOperationTime`` or saved from the original aggregation) and the max wire version is >= ``7``:
```
jmikola: With this change, it looks as if a saved operation time always takes precedence over the resume token. In the original text, this was an "else if" below all of the resume token conditions and before the final "else" condition that re-sends the original aggregation w/o any options.

I think you need an initial "if" condition above this case to resume with the `resumeAfter` option set to the return value of `getResumeToken()` and only fall over to this case if `getResumeToken()` could not return anything.

source/change-streams/change-streams.rst
```diff
@@ -544,8 +525,34 @@ Option 1: ChangeStream::getResumeToken()
...
   }
 
-This method returns the cached ``documentResumeToken`` or ``postBatchResumeToken`` following rules from `Resume Process`_:
+This method returns the cached ``documentResumeToken`` or ``postBatchResumeToken`` following these rules:
+
+- If the ``ChangeStream`` has a cached ``postBatchResumeToken`` and has returned all documents in the most recent batch (or the most recent batch was empty), return the cached ``postBatchResumeToken`` and:
+
+  - The driver MUST set ``resumeAfter`` to the cached ``postBatchResumeToken``.
+  - The driver MUST NOT set ``startAfter``. If ``startAfter`` was in the original aggregation command, the driver MUST remove it.
+  - The driver MUST NOT set ``startAtOperationTime``. If ``startAtOperationTime`` was in the original aggregation command, the driver MUST remove it.
```
jmikola: The "MUST set" and "MUST NOT set" lines here are no longer relevant, since we're only specifying what is returned. You can probably delete the trailing "and:" from each top-level bullet (e.g. "if" and "else if" conditions) and just remove the sub-bullets entirely.

You may want to preserve one set of the MUST and MUST NOT bullets to use above in the resume process docs. If `getResumeToken()` actually returns a value, I believe you'd always want to set it as `resumeAfter` and never set `startAfter` and `startAtOperationTime`.

source/change-streams/change-streams.rst
```diff
@@ -544,8 +525,34 @@ Option 1: ChangeStream::getResumeToken()
...
+  - The driver MUST set ``resumeAfter`` to the value of the originally used ``startAfter``.
+  - The driver MUST NOT set ``startAfter``. The driver MUST remove it from the original aggregation command.
+  - The driver MUST NOT set ``startAtOperationTime``.
+- Else if the ``ChangeStream`` was created with ``resumeAfter``, return ``resumeAfter`` and:
+
+  - The driver MUST set ``resumeAfter`` to the value of the originally used ``resumeAfter``.
+  - The driver MUST NOT set ``startAfter``.
+  - The driver MUST NOT set ``startAtOperationTime``.
+
+- If the ``ChangeStream`` has returned no documents in a non-empty batch from the original aggregate:
```
jmikola: This "If" bullet follows several "Else if" bullets above, which doesn't look correct.

There are also two more "If" bullets below, but they were there before your PR and I expect "else" was omitted because they each return.

source/change-streams/change-streams.rst
kevinAlbs: Also, I'd prefer not saying "Call getResumeToken()". [Exposing All Resume Tokens](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#exposing-all-resume-tokens) gives drivers the option of implementing `getResumeToken()` or adding a new event.

Considering Jeremy's other comment, I'm thinking something like:
- Follow the rules of `getResumeToken()` to check for a resume token. If the rules apply:
   - The driver MUST set ``resumeAfter`` to what the rules return.
   - The driver MUST NOT set ``startAfter``. If ``startAfter`` was in the original aggregation command, the driver MUST remove it.
   - The driver MUST NOT set ``startAtOperationTime``. If ``startAtOperationTime`` was in the original aggregation command, the driver MUST remove it.
- otherwise, if the change stream has a saved operation time ...

source/change-streams/change-streams.rst
```diff
@@ -544,8 +525,34 @@ Option 1: ChangeStream::getResumeToken()
     public getResumeToken() Optional<Document>;
   }
 
-This method returns the cached ``documentResumeToken`` or ``postBatchResumeToken`` following rules from `Resume Process`_:
+This method returns the cached ``documentResumeToken`` or ``postBatchResumeToken`` following these rules:
+
+- If the ``ChangeStream`` has a cached ``postBatchResumeToken`` and has returned all documents in the most recent batch (or the most recent batch was empty), return the cached ``postBatchResumeToken`` and:
```
jmikola: Noting the tests that @DmitryLukyanov created in https://github.com/mongodb/specifications/pull/526/files, I think there are some additional edge cases to mention here:

 * `aggregate` has just executed. The returned batch is not empty and has not been iterated at all. We ignore any document resume token or `postBatchResumeToken` and only return a user-specified option (`startAfter` or `resumeAfter`) if available.
 * `getMore` has just executed. The returned batch is not empty and has not been iterated at all. We return the previous batch's `postBatchResumeToken` if available. Otherwise, return the last document resume token we saw in the previous batch. Otherwise, fall back to a user-specified option (`startAfter` or `resumeAfter`) if available.

jmikola: Something else that occurred to me while reading this PR and implementing `postBatchResumeToken` in libmongoc is that the spec is not entirely clear on _when_ to update the cached `postBatchResumeToken`.

[Driver API](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#driver-api) states the following for `postBatchResumeToken`:

> The most recent postBatchResumeToken returned in an aggregate or getMore response.

Later on in the spec, [ChangeStream](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#changestream) states:

> A change stream MUST track the latest postBatchResumeToken included in an aggregate or getMore response if available.

I think both of these can easily be misinterpreted, which may lead to bugs when implementing resume logic or `getResumeToken()`. The `getMore` case in my previous comment may need to return `postBatchResumeToken` from the _previous_ command response (`aggregate` or `getMore`) if the most recently executed `getMore` batch is non-empty and has not been iterated. If drivers decide to update their cached `postBatchResumeToken` value immediately upon receiving an `aggregate` or `getMore` response, they may lose that _previous_ value that may need to be returned before iterating a non-empty batch.

We'd do well to clarify exactly when the cached `postBatchResumeToken` should be updated.

 * If `getMore` returns a non-empty batch, we need to wait until at least one document has been iterated before overwriting the last-cached `postBatchResumeToken`.
 * If `getMore` returns an empty batch, there is no harm in updating `postBatchResumeToken` immediately.

I think the above clarification can go into this PR since it's closely related to resume tokens, but if anyone objects feel free to open a SPEC ticket to handle this separately.

kevinAlbs: Good catch. I agree that's wrong or at least misleading.

I wonder if separating `postBatchResumeToken` and `documentResumeToken` in the [Driver API](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#driver-api) was a mistake. It makes the rules for resuming (well now `getResumeToken()`) simpler. Because, it lets us say "if x, return `postBatchResumeToken`, if y, return `documentResumeToken`". But if we need to add special rules of how we update the cached `postBatchResumeToken`, maybe we should go even further. Just have one `resumeToken` in `ChangeStream`. Something like:

```
interface ChangeStream extends Iterable<Document> {
  /**
   * A cached resume token.
   */
```

Then, we rework rules and put them in a new section ala:

**Updating the cached resume token**

- When the change stream is started:
    - If `resumeAfter` is specified, use that.
    - Otherwise, if `startAfter` is specified, use that.
    - Otherwise, `ChangeStream.resumeToken` remains unset.
- When `aggregate` or `getMore` returns:
    - If an empty batch was returned and a `postBatchResumeToken` is included, use that.
- When returning a document to the user:
    - If it's the last document in the batch and a `postBatchResumeToken` is included, use `postBatchResumeToken`.
    - Otherwise, use the `_id` of the document.

And then, `getResumeToken()` just returns `ChangeStream.resumeToken`.

Normally, I'd vote for creating a separate spec ticket instead of piling on more changes. But in this case we'll end up undoing the work here, so I vote for doing this as part of this PR.

@jmikola what do you think?

jmikola: "Updating the cached resume token" reads very clearly and would also greatly simplify logic for exposing the resume token (either getter or dispatching an event). I'd be in favor of this change.

jmikola: @kevinAlbs: One question. Would your initial case for "when the change stream is started" conflict with the following bullet in [Option 1: ChangeStream::getResumeToken()](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#option-1-changestream-getresumetoken)?

> If the ChangeStream has no cached documentResumeToken or postBatchResumeToken this returns an unset optional.

I assume we also wouldn't want to emit an event with the initial `resumeAfter` or `startAfter` option values, since users already have access to anything provided as a change stream option.

I do agree with consolidating the `documentResumeToken` and `postBatchResumeToken` tracking into a single field, though.

jmikola: @kevinAlbs: A correction to your last comment:

> **Updating the cached resume token**
> 
> - When the change stream is started:
>     - If `resumeAfter` is specified, use that.
>     - Otherwise, if `startAfter` is specified, use that.
>     - Otherwise, `ChangeStream.resumeToken` remains unset.

According to [Resume Process](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#resume-process), `startAfter` takes precedence over `resumeAfter`. I think those should be switched in your list.

kevinAlbs: > One question. Would your initial case for "when the change stream is started" conflict with the following bullet in Option 1: ChangeStream::getResumeToken()?

Hah, yeah. I think `getResumeToken()` should have returned the value of `startAfter`/`resumeAfter` in the first place. If an application is saving the value of `getResumeToken()`, it should be able to start a new change stream with `resumeAfter: getResumeToken()` wherever it left off, even if nothing has been iterated. What I have in mind:

```
savedResumeToken = load("resumeToken.json")
cs = db.coll.watch({resumeAfter: savedResumeToken})
save(cs.getResumeToken(), "resumeToken.json")
/* application crashes here, before iteration. */
while (doc = cs.next()) {
    process(doc);
    save (cs.getResumeToken(), "resumeToken.json")
}
```

> I assume we also wouldn't want to emit an event with the initial resumeAfter or startAfter option values, since users already have access to anything provided as a change stream option.

By the same reasoning, I think we should actually emit an event.
```
savedResumeToken = load("resumeToken.json")
cs = db.coll.watch({resumeAfter: savedResumeToken})
cs.on("resumeTokenChanged", (token) => save(token, "resumeToken.json"))
/* application crashes before first document returned. */
cs.on("change", (doc) => { ... })
```

That keeps the logic consolidated. We have one "Updating the cached resume token" section. `getResumeToken()` returns that cached token. And when that cached resume token changes async drivers emit an event with that token.

> According to Resume Process, startAfter takes precedence over resumeAfter. I think those should be switched in your list.

Order shouldn't matter. startAfter and resumeAfter are mutually exclusive.

jmikola: > I think we should actually emit an event.

In the event-emitting example you provided, I suppose the `resumeTokenChanged` handler isn't registered until _after_ the change stream is constructed with `watch()` so it may never need to receive a redundant event for the initial option as that would have occurred _during_ `watch()`.

> Order shouldn't matter. startAfter and resumeAfter are mutually exclusive.

Drivers rely on the server to raise an error if both are specified, but there are several libmongoc tests where we specify both at once and assert what happens with the outgoing command. I foresee doing the same for `getResumeToken()` tests, so this could come up if we're using a mock server.

kevinAlbs: > In the event-emitting example you provided, I suppose the resumeTokenChanged handler isn't registered until after the change stream is constructed with watch() so it may never need to receive a redundant event for the initial option as that would have occurred during watch().

I dunno, I guess that depends on implementation. Like, right now in Node attaching a "change" event listener [triggers the change stream iteration](http://mongodb.github.io/node-mongodb-native/3.2/api/ChangeStream.html):

> Attaching a change event listener to a Change Stream will switch the stream into flowing mode. Data will then be passed as soon as it is available.

Plus, it might not be redundant. Maybe a more compelling example is when you get a change stream already constructed, and don't have the initial options accessible:

```
cs = getSomeChangeStreamObject();
cs.on("resumeTokenChanged", (token) => save(token, "resumeToken.json"))
```

Foreseeably, a "resumeTokenChanged" event listener could be called immediately with the value of `startAfter`/`resumeAfter` if the change stream hasn't started iterating internally. But then, I guess there'd be weirdness of different not getting the first "resumeTokenChanged" event if you attach a "change" event first...

I vote we leave this up to the implementation and stay vague, like we have it now: "Allow users to set a callback to listen for new resume tokens.". If you think we should make a note of it, let's not continue piling on in this PR, and just create another spec ticket to clarify the first token that should be emitted in "resumeTokenChanged".

> Drivers rely on the server to raise an error if both are specified, but there are several libmongoc tests where we specify both at once and assert what happens with the outgoing command. I foresee doing the same for getResumeToken() tests, so this could come up if we're using a mock server.

Fair enough. I guess we need to stick to my original arbitrary ordering `startAfter` takes precedence over `resumeAfter`.

source/change-streams/change-streams.rst
```diff
@@ -205,15 +205,9 @@ Driver API
 
   interface ChangeStream extends Iterable<Document> {
     /**
-     * The resume token (_id) of the document the iterator last returned
+     * The resume token to be used for resuming the iterator
```
kevinAlbs: Hmm, now we use this for both resuming the change stream as part of the "Resume Process" and this is the thing we return in `getResumeToken()`. What do you think about making this more vague, like: "A cached resume token"? We describe how it's used later.

source/change-streams/change-streams.rst
```diff
@@ -575,6 +548,21 @@ A possible interface for this callback MAY look like:
 
 This MUST NOT be implemented in synchronous drivers. This MAY be implemented in asynchronous drivers.
 
+Updating the Cached Resume Token
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This method returns the cached ``resumeToken`` following these rules:
```
kevinAlbs: Reword this section. "This method returns" doesn't apply anymore. This describes how we update the cached `resumeToken`.

source/change-streams/change-streams.rst
```diff
@@ -483,27 +477,11 @@ Once a ``ChangeStream`` has encountered a resumable error, it MUST attempt to re
 
 - Perform server selection.
 - Connect to selected server.
-- If the ``ChangeStream`` has a cached ``postBatchResumeToken`` and has returned all documents in the most recent batch (or the most recent batch was empty):
-
-  - The driver MUST set ``resumeAfter`` to the cached ``postBatchResumeToken``.
+- If there is a cached ``resumeToken``:
```
DmitryLukyanov: this is how this comment looks like in my browser:
image

source/change-streams/change-streams.rst
```diff
@@ -575,6 +548,21 @@ A possible interface for this callback MAY look like:
 
 This MUST NOT be implemented in synchronous drivers. This MAY be implemented in asynchronous drivers.
 
+Updating the Cached Resume Token
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following rules describe how to update the cached ``resumeToken``:
```
DmitryLukyanov: Markup:
image

DmitryLukyanov:
Hi, a couple of small comments about this PR:
1. As I see, currently there are several incorrect page markups. See, the `File changed` comments.
2. Am I right that the only thing which is going to be done by `getResumeToken` is to return `resumeToken` which is currently `private` in `changeStream` class. if so, Maybe, it's better to make this field `public` and do not have `getResumeToken` at all?

kevinAlbs:
@DmitryLukyanov 

> Am I right that the only thing which is going to be done by getResumeToken is to return resumeToken which is currently private in changeStream class. if so, Maybe, it's better to make this field public and do not have getResumeToken at all?

Yes, `getResumeToken()` now just returns the cached `resumeToken` value.

We want to future-proof ourselves in case we change the logic for which resume token to return to the user (we've already changed it once). So the user should get the resume token by calling a getter method.

I think the cached resume token is comparable to [cluster time in ClientSession](https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst#clientsession). It's a read-only value set internally. To your point, it is defined as a public member:

```
interface ClientSession {
    MongoClient client;
    Optional<BsonDocument> clusterTime;
    SessionOptions options;
    BsonDocument sessionId;

    void advanceClusterTime(BsonDocument clusterTime);
    void endSession();
}
```

But the implementation is language dependant. Languages that support a way of defining properties on classes define it as a read only property like in [PyMongo](https://github.com/mongodb/mongo-python-driver/blob/4049b1493a970b7a15e1c171b178a55e4819e61f/pymongo/client_session.py#L358-L363).

But languages that with no notion of a property define a getter method, like [Java](http://mongodb.github.io/mongo-java-driver/3.10/javadoc/com/mongodb/session/ClientSession.html).

I consider a public read-only property like PyMongo's cluster_time equivalent to a getter method. Similarly, I think it's valid for PyMongo to implement `getResumeToken` as a public read-only property. Since ClientSession doesn't explicitly distinguish getter method vs public read-only property, I don't think we need to here either. But if you feel strongly, can you open a new spec ticket? This would be a discussion about the API, not the resume logic. It seems outside of the scope of this ticket.

DmitryLukyanov:
"But if you feel strongly, can you open a new spec ticket?" - `getResumeToken` is ok for me, it was just small suggestion.

source/change-streams/change-streams.rst
```diff
@@ -440,13 +434,13 @@ ChangeStream
...
 
-A change stream MUST track the last resume token returned by the iterator to the user, caching it locally for use in future attempts to resume.  A driver MUST raise an error on the first response received without a resume token (e.g. the user has removed it with a pipeline stage), and close the change stream.  The error message SHOULD resemble “Cannot provide resume functionality when the resume token is missing”.
+A change stream MUST track the last resume token, per `Updating the Cached Resume Token`_.
 
-A change stream MUST track the latest ``postBatchResumeToken`` included in an aggregate or getMore response if available.
+A driver MUST raise an error on the first document received without a resume token (e.g. the user has removed ``_id`` with a pipeline stage), and close the change stream.  The error message SHOULD resemble “Cannot provide resume functionality when the resume token is missing”.
```
kevinAlbs: "A driver" => "The driver"

source/change-streams/change-streams.rst
```diff
@@ -544,11 +524,6 @@ Option 1: ChangeStream::getResumeToken()
     public getResumeToken() Optional<Document>;
   }
```
kevinAlbs: How about adding:
"This method returns the cached `resumeToken`"

source/change-streams/change-streams.rst
jmikola: Looks like this has been fixed since this comment was posted.

source/change-streams/change-streams.rst
jmikola: Looks like this has been fixed since this comment was posted.

PR #534, opened 2019-04-11
SPEC-1274: add comment for unsupported commands

jstewart-mongo:
Added a paragraph in the retryable reads test README for handling unsupported commands in drivers.


source/retryable-reads/tests/README.rst
```diff
@@ -150,6 +150,14 @@ the SDAM's ``minHeartbeatFrequencyMS`` from 500ms to 50ms, thus decreasing the
 waiting time after a "not master" error:
 
 .. _minHeartbeatFrequencyMS: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#minheartbeatfrequencyms
+
+Unsupported commands
```
vincentkam: Hm, would this be effectively a duplicate of the "Optional Enumeration Commands" section below?

vincentkam: Perhaps the two sections could be combined under the heading "Unsupported optional enumeration commands" for greater clarity and minimal redundancy?

jstewart-mongo: You are correct. I missed that. I will close as the ticket as already fixed.

PR #536, opened 2019-04-24
SPEC-1279: Drivers send recoveryToken on abortTransaction

vincentkam:


source/transactions/tests/mongos-recovery-token.yml
```diff
@@ -269,7 +269,7 @@ tests:
       collection:
         data: []
 
-  - description: abortTransaction does not send recoveryToken
+  - description: abortTransaction sends recoveryToken
```
jyemin: Do we also need `abortTransaction retry succeeds/fails` on new mongos` tests, similar to the ones for commitTransaction?

vincentkam: I had considered the same possibility. I could be wrong, but I believe that since [drivers ignore `abortTransaction` errors](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#drivers-ignore-all-aborttransaction-errors), there's no need to have the `abortTransaction retry succeeds/fails` tests similar to the ones for `commitTransaction`

ShaneHarvey: I agree we don't need to add tests for `abortTransaction retry succeeds/fails on new mongos` because abortTransaction always succeeds.

vincentkam:
Merged!

PR #537, opened 2019-04-24
SPEC-1285: Add wildcardProjection to IndexOptions

divjotarora:


source/index-management.rst
```diff
@@ -682,6 +682,11 @@ Common API Components
      * server-side is used.
      */
     collation: Document;
+
+    /**
+     * Optionally specifies the wildcard projection of a wildcard index.
+     */
```
ShaneHarvey: Do you think it would be helpful to add an example `wildcardProjection` document to the doc string?

divjotarora: Maybe - I don't think it's consistent, though, because other options like `storageEngine` don't have an example. Also, unless drivers are going to be doing any validation on the document, I don't see the value of having an example here. It seems like something that would be useful for users of drivers, but not driver authors.

ShaneHarvey: I figured the other options _would_ have example but since they don't this SGTM.

jmikola: I concur that this is something to sort out in our user documentation, but the spec need only concern itself with enforcing that the option is a document type.

source/index-management.rst
```diff
@@ -682,6 +682,11 @@ Common API Components
      * server-side is used.
      */
     collation: Document;
+
+    /**
+     * Optionally specifies the wildcard projection of a wildcard index.
```
jmikola: This may be outside the scope of the PR, but I think it's be beneficial to remove "Optionally" from the prefix of each doc block description and instead wrap the types with `Optional<TYPE_NAME>` as we do in other specifications. Perhaps we can do this after merging this PR.

mbroadst:
Does it make sense to add some tests for these fields for the benefit of languages that are strict about what to include in the index options? Seems like a small amount of work for great potential benefit.

divjotarora:
Spec tests or prose tests?

PR #538, opened 2019-04-24
SPEC-1206 Support maxTimeMS on transaction commit

ShaneHarvey:
Jira: https://jira.mongodb.org/browse/SPEC-1206

Changes:
- TransactionOptions gains a new max_time_ms option. 
- MaxTimeMSExpired errors on commit (both `{ok:0, code: 50, codeName: "MaxTimeMSExpired"}` and `{ok:1, writeConcernError: {code: 50, codeName: "MaxTimeMSExpired"}}`) are labelled UnknownTransactionCommitResult.
- withTransaction does not retry commit after a MaxTimeMSExpired error. Aly requested that MaxTimeMSExpired halt any retry attempts.

I have a POC in pymongo implemented here: https://github.com/ShaneHarvey/mongo-python-driver/tree/SPEC-1206
Here is a passing patch build: https://evergreen.mongodb.com/version/5cc0e14230661576bcab0b82 (note the failures are due to [PYTHON-1796](https://jira.mongodb.org/browse/PYTHON-1796)/[SERVER-40685](https://jira.mongodb.org/browse/SERVER-40685))

### DIFF
--- a/source/transactions-convenient-api/tests/commit-retry.yml
+++ b/source/transactions-convenient-api/tests/commit-retry.yml
@@ -21,7 +21,7 @@ tests:
           failCommands: ["commitTransaction"]
           closeConnection: true
     operations:
-      -
+      - &withTransaction
         name: withTransaction
         object: session0
         arguments:
@@ -194,20 +194,7 @@ tests:
           errorCode: 10107 # NotMaster
           closeConnection: false
     operations:
-      -
-        name: withTransaction
-        object: session0
-        arguments:
-          callback:
-            operations:
-              -
-                name: insertOne
-                object: collection
-                arguments:
-                  session: session0
-                  document: { _id: 1 }
-                result:
-                  insertedId: 1
+      - *withTransaction
     expectations:
       -
         command_started_event:
@@ -270,3 +257,53 @@ tests:
       collection:
         data:
           - { _id: 1 }
+  -
+    description: commit is not retried after MaxTimeMSExpired error
+    failPoint:
+      configureFailPoint: failCommand
+      mode: { times: 1 }
+      data:
+          failCommands: ["commitTransaction"]
+          errorCode: 50 # MaxTimeMSExpired
+    operations:
+      - <<: *withTransaction
+        result:
+          errorCodeName: MaxTimeMSExpired
+          errorLabelsContain: ["UnknownTransactionCommitResult"]
+          errorLabelsOmit: ["TransientTransactionError"]
+    expectations:
+      -
+        command_started_event:
+          command:
+            insert: *collection_name
+            documents:
+              - { _id: 1 }
+            ordered: true
+            lsid: session0
+            txnNumber: { $numberLong: "1" }
+            startTransaction: true
+            autocommit: false
+            # omitted fields
+            readConcern: ~
+            writeConcern: ~
+          command_name: insert
+          database_name: *database_name
+      -
+        command_started_event:
+          command:
+            commitTransaction: 1
+            lsid: session0
+            txnNumber: { $numberLong: "1" }
+            autocommit: false
+            # omitted fields
+            readConcern: ~
+            startTransaction: ~
+            writeConcern: ~
+          command_name: commitTransaction
+          database_name: admin
+    outcome:
+      collection:
+        # In reality, the outcome of the commit is unknown but we fabricate
+        # the error with failCommand.errorCode which does not apply the commit
+        # operation.
+        data: []
--- a/source/transactions-convenient-api/tests/commit-retry.yml
+++ b/source/transactions-convenient-api/tests/commit-retry.yml
@@ -266,7 +266,21 @@ tests:
           failCommands: ["commitTransaction"]
           errorCode: 50 # MaxTimeMSExpired
     operations:
-      - <<: *withTransaction
+      - name: withTransaction
+        object: session0
+        arguments:
+          callback:
+            operations:
+              -
+                name: insertOne
+                object: collection
+                arguments:
+                  session: session0
+                  document: { _id: 1 }
+                result:
+                  insertedId: 1
+          options:
+            maxTimeMS: 60000
         result:
           errorCodeName: MaxTimeMSExpired
           errorLabelsContain: ["UnknownTransactionCommitResult"]
@@ -295,6 +309,7 @@ tests:
             lsid: session0
             txnNumber: { $numberLong: "1" }
             autocommit: false
+            maxTimeMS: 60000
             # omitted fields
             readConcern: ~
             startTransaction: ~
--- a/source/transactions-convenient-api/tests/commit-retry.yml
+++ b/source/transactions-convenient-api/tests/commit-retry.yml
@@ -280,7 +280,7 @@ tests:
                 result:
                   insertedId: 1
           options:
-            maxTimeMS: 60000
+            maxCommitTimeMS: 60000
         result:
           errorCodeName: MaxTimeMSExpired
           errorLabelsContain: ["UnknownTransactionCommitResult"]
--- a/source/transactions-convenient-api/tests/commit-writeconcernerror.yml
+++ b/source/transactions-convenient-api/tests/commit-writeconcernerror.yml
@@ -193,3 +193,24 @@ tests:
     expectations: *expectations_without_retries
     # failCommand with writeConcernError still applies the write operation(s)
     outcome: *outcome
+
+  -
+    description: commitTransaction is not retried after MaxTimeMSExpired error
+    failPoint:
+      configureFailPoint: failCommand
+      mode: { times: 1 }
+      data:
+          failCommands: ["commitTransaction"]
+          writeConcernError:
+            code: 50
+            codeName: MaxTimeMSExpired
+            errmsg: "operation exceeded time limit"
+    operations:
+      - <<: *operation
+        result:
+          errorCodeName: MaxTimeMSExpired
+          errorLabelsContain: ["UnknownTransactionCommitResult"]
+          errorLabelsOmit: ["TransientTransactionError"]
+    expectations: *expectations_without_retries
+    # failCommand with writeConcernError still applies the write operation(s)
+    outcome: *outcome
--- a/source/transactions-convenient-api/transactions-convenient-api.rst
+++ b/source/transactions-convenient-api/transactions-convenient-api.rst
@@ -388,8 +388,8 @@ their users to design idempotent callbacks.
 
 .. _Retry Transactions and Commit Operation: https://docs.mongodb.com/manual/core/transactions/#retry-transaction-and-commit-operation
 
-The commit is retried after a write concern wtimeout error
-----------------------------------------------------------
+The commit is retried after a write concern timeout (i.e. wtimeout) error
+-------------------------------------------------------------------------
 
 Per the Transactions specification, drivers internally retry
 ``commitTransaction`` once if it fails due to a retryable error (as defined in
--- a/source/transactions-convenient-api/transactions-convenient-api.rst
+++ b/source/transactions-convenient-api/transactions-convenient-api.rst
@@ -123,11 +123,11 @@ drivers MUST enforce a 120-second timeout to limit retry behavior and safeguard
 applications from long-running (or infinite) retry loops. Drivers SHOULD use a
 monotonic clock to determine elapsed time.
 
-If an UnknownTransactionCommitResult error is encountered for a commit and the
-error is not MaxTimeMSExpired and the retry timeout has not been exceeded, the
-driver MUST retry the commit. If the retry timeout has been exceeded, drivers
-MUST NOT retry the commit and allow ``withTransaction`` to propagate the
-error to its caller.
+If an UnknownTransactionCommitResult error is encountered for a commit, the
+driver MUST retry the commit if and only if the error is not MaxTimeMSExpired
+and the retry timeout has not been exceeded. Otherwise, the driver MUST NOT
+retry the commit and allow ``withTransaction`` to propagate the error to its
+caller.
 
 If a TransientTransactionError is encountered at any point, the entire
 transaction may be retried. If the retry timeout has not been exceeded, the
@@ -499,4 +499,5 @@ Changes
 
 2019-04-24: withTransaction does not retry when commit fails with
             MaxTimeMSExpired.
+
 2018-02-13: withTransaction should retry commits after a wtimeout
--- a/source/transactions-convenient-api/transactions-convenient-api.rst
+++ b/source/transactions-convenient-api/transactions-convenient-api.rst
@@ -3,14 +3,14 @@ Convenient API for Transactions
 ===============================
 
 :Spec Title: Convenient API for Transactions
-:Spec Version: 1.1
+:Spec Version: 1.2
 :Author: Jeremy Mikola
 :Lead: Jeff Yemin
 :Advisors: A\. Jesse Jiryu Davis, Kris Brandow, Oleg Pudeyev, Sam Ritter, Tess Avitabile
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: 4.0
-:Last Modified: 2019-02-13
+:Last Modified: 2019-04-24
 
 .. contents::
 
@@ -123,11 +123,10 @@ drivers MUST enforce a 120-second timeout to limit retry behavior and safeguard
 applications from long-running (or infinite) retry loops. Drivers SHOULD use a
 monotonic clock to determine elapsed time.
 
-If an UnknownTransactionCommitResult is encountered for a commit, the commit
-operation may be retried. If the retry timeout has not been exceeded, the driver
-MUST retry a commit that fails with an error bearing the
-"UnknownTransactionCommitResult" label. If the retry timeout has been exceeded,
-drivers MUST NOT retry the commit and allow ``withTransaction`` to propagate the
+If an UnknownTransactionCommitResult error is encountered for a commit and the
+error is not MaxTimeMSExpired and the retry timeout has not been exceeded, the
+driver MUST retry the commit. If the retry timeout has been exceeded, drivers
+MUST NOT retry the commit and allow ``withTransaction`` to propagate the
 error to its caller.
 
 If a TransientTransactionError is encountered at any point, the entire
@@ -204,9 +203,10 @@ This method should perform the following sequence of actions:
 9. If ``commitTransaction`` reported an error:
 
    a. If the ``commitTransaction`` error includes a
-      "UnknownTransactionCommitResult" label and the elapsed time of
-      ``withTransaction`` is less than 120 seconds, jump back to step eight. We
-      will trust ``commitTransaction`` to apply a majority write concern on
+      "UnknownTransactionCommitResult" label and the error is not
+      MaxTimeMSExpired and the elapsed time of ``withTransaction`` is less
+      than 120 seconds, jump back to step eight. We will trust
+      ``commitTransaction`` to apply a majority write concern on
       retry attempts (see:
       `Majority write concern is used when retrying commitTransaction`_).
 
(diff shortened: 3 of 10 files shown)


source/transactions-convenient-api/transactions-convenient-api.rst
```diff
@@ -123,11 +123,10 @@ drivers MUST enforce a 120-second timeout to limit retry behavior and safeguard
...
 monotonic clock to determine elapsed time.
 
-If an UnknownTransactionCommitResult is encountered for a commit, the commit
-operation may be retried. If the retry timeout has not been exceeded, the driver
-MUST retry a commit that fails with an error bearing the
-"UnknownTransactionCommitResult" label. If the retry timeout has been exceeded,
-drivers MUST NOT retry the commit and allow ``withTransaction`` to propagate the
+If an UnknownTransactionCommitResult error is encountered for a commit and the
+error is not MaxTimeMSExpired and the retry timeout has not been exceeded, the
+driver MUST retry the commit. If the retry timeout has been exceeded, drivers
```
jmikola: Would it make sense to replace "If the retry timeout has been exceeded" with "Otherwise,"? As-is, this paragraph doesn't technically state that drivers MUST NOT retry after a MaxTimeMSExpired error.

source/transactions-convenient-api/transactions-convenient-api.rst
```diff
@@ -268,7 +268,14 @@ This method can be expressed by the following pseudo-code:
                      * being retried (see: DRIVERS-601) */
                     this.commitTransaction();
                 } catch (error) {
-                    if (error.hasErrorLabel("UnknownTransactionCommitResult") &&
+                    /* Note: a maxTimeMS error will have the MaxTimeMSExpired
+                     * code (50) and can be reported as a top-level error or
+                     * inside writeConcernError, ie:
+                     * {ok:0, code: 50, codeName: "MaxTimeMSExpired"}
+                     * {ok:1, writeConcernError: {code: 50, codeName: "MaxTimeMSExpired"}}
```
jmikola: This may be off topic, but is there a particular reason we've resorted to checking error codes instead of relying on labels? Was there something about maxTimeMS that necessitated using an error code?

source/transactions-convenient-api/transactions-convenient-api.rst
```diff
@@ -381,14 +388,14 @@ their users to design idempotent callbacks.
 
 .. _Retry Transactions and Commit Operation: https://docs.mongodb.com/manual/core/transactions/#retry-transaction-and-commit-operation
 
-The commit is retried after a write concern timeout error
----------------------------------------------------------
+The commit is retried after a write concern wtimeout error
```
jmikola: Was this done for clarity's sake? IMO, a "write concern timeout" implies wtimeout. "write concern wtimeout" reads a bit strange, though. Perhaps "write concern timeout (i.e. wtimeout)"? Some text below still uses the phrase "write concern timeout" so I think the single "i.e." addition here would mesh nicely (similar to defining an acronym on its first use).

source/transactions/tests/error-labels.yml
```diff
@@ -959,3 +959,247 @@ tests:
...
+            _id: 1
+        result:
+          insertedId: 1
+      - name: aggregate
+        object: collection
+        arguments:
+          pipeline:
+            - $project:
+                _id: 1
+          maxTimeMS: 60000
```
jmikola: Should this also be included in the corresponding CommandStartedEvent? As-is, I expect the matcher will just ignore it in the actual outgoing command as an extra field not present in the assertion. I imagine the value doesn't really matter, since the fail point likely triggers immediately, but it'd be good to test that the driver sends what we expect.

source/transactions/tests/error-labels.yml
```diff
@@ -959,3 +959,247 @@ tests:
...
+
+  - description: add unknown commit label to writeConcernError MaxTimeMSExpired
+
+    failPoint:
+      configureFailPoint: failCommand
+      mode: { times: 1 }
+      data:
+          failCommands: ["commitTransaction"]
+          writeConcernError:
+            code: 50
```
jmikola: Various other tests omit `codeName` and only include `code` and `errmsg`. I don't recall if that was intentional, but if so would it make sense to do the same here to ensure that drivers are comparing against the numeric code instead of inspecting strings?

source/transactions/transactions.rst
```diff
@@ -150,6 +150,22 @@ This section is an overview of the public API for transactions:
          * The readPreference to use for this transaction.
          */
         Optional<ReadPreference> readPreference;
+
+        /**
+         * The readPreference to use for this transaction.
+         */
+        Optional<ReadPreference> readPreference;
```
jmikola: This is duplicated.

source/transactions/transactions.rst
```diff
@@ -150,6 +150,22 @@ This section is an overview of the public API for transactions:
...
+        /**
+         * The maximum amount of time to allow the *commit* to run.
+         *
+         * This option is only sent with the commitTransaction command and
+         * only if the caller explicitly provides a value.
+         * The default is to not send a value.
+         *
+         * @see https://docs.mongodb.com/manual/reference/command/commitTransaction/
+         */
+        Optional<Int64> maxTimeMS;
```
jmikola: Would it make sense to copy in some text elsewhere in the spec about permitted naming deviations, as we do for `maxTimeMS` in the [CRUD spec](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#id1)? That's assuming that some drivers might use an alternate type for this option and forgo the "MS" suffix.

source/transactions-convenient-api/tests/commit-retry.yml
```diff
@@ -270,3 +257,53 @@ tests:
       collection:
         data:
           - { _id: 1 }
+  -
+    description: commit is not retried after MaxTimeMSExpired error
```
jmikola: Does it make sense to configure a fail point for MaxTimeMSExpired without adding `maxTimeMS` to the outgoing commit command? I realize it probably doesn't matter since the driver's behavior when processing a response/error shouldn't be affected by the original outgoing command, but doing so might improve readability if someone comes across this down the line.

If this would be too much trouble (since you inherit `withTransaction` as one block), then perhaps a comment about that explaining that we're not bothering to specify `maxTimeMS` as a command option would work just as well for readability.

source/transactions-convenient-api/transactions-convenient-api.rst
ShaneHarvey: The server does not return the UnknownTransactionCommitResult error label. I can open a ticket for them to add it in future versions though.

source/transactions-convenient-api/transactions-convenient-api.rst
ShaneHarvey: Done.

source/transactions-convenient-api/transactions-convenient-api.rst
ShaneHarvey: Yeah, I changed it for clarity because now we are considering two different write concern timeouts. wtimeout:
`{ok:1, writeConcernError: {code: 64, codeName: "WriteConcernFailed", errmsg: "waiting for replication timed out", errInfo: {wtimeout: True}}}`
And maxTimeMS:
`{ok:1, writeConcernError: {code: 50, codeName: "MaxTimeMSExpired"}}`

Updated the title with "write concern timeout (i.e. wtimeout)".

source/transactions/transactions.rst
ShaneHarvey: Oops, how'd that get in here? Begone!

source/transactions/transactions.rst
ShaneHarvey: Good idea, done.

source/transactions/tests/error-labels.yml
ShaneHarvey: Fine by me.

source/transactions/tests/error-labels.yml
ShaneHarvey: Nice catch, done.

source/transactions-convenient-api/tests/commit-retry.yml
ShaneHarvey: Yeah that makes sense. I added maxTimeMS.

source/transactions/transactions.rst
```diff
@@ -150,6 +156,17 @@ This section is an overview of the public API for transactions:
          * The readPreference to use for this transaction.
          */
         Optional<ReadPreference> readPreference;
+
+        /**
+         * The maximum amount of time to allow the *commit* to run.
```
ShaneHarvey: Does anyone else think it may be misleading for this option to be named `maxTimeMS`. That seems to imply that it is a limit for the _entire transaction_ but it only applies to the _commit_.

I think we could improve this in one of two ways:
1. We can ask the server to support `maxTimeMS` on the initial `startTransaction: true` command. This way the server can apply the `maxTimeMS` limit the lifetime of the entire transaction similar to how `maxTimeMS` works on a cursor.
2. We can rename this option to make it more clear that it only applies to the commit, like `maxCommitTimeMS`.

Note I considered adding a `maxTimeMS` parameter to ClientSession.commitTransaction and rejected that idea because it would be inconsistent with writeConcern and it would also make it impossible to use maxTimeMS with the `withTransaction` API where the commit call is implicit.

Any thoughts?

ShaneHarvey:
Does anyone else think it may be misleading for this option to be named `maxTimeMS`. That seems to imply that it is a limit for the _entire transaction_ but it only applies to the _commit_.

I think we could improve this in one of two ways:
1. We can ask the server to support `maxTimeMS` on the initial `startTransaction: true` command. This way the server can apply the `maxTimeMS` limit the lifetime of the entire transaction similar to how `maxTimeMS` works on a cursor.
2. We can rename this option to emphasize that it only applies to the commit, like `maxCommitTimeMS`.

Note I considered adding a `maxTimeMS` parameter to ClientSession.commitTransaction and rejected that idea because it would be inconsistent with writeConcern and it would also make it impossible to use maxTimeMS with the `withTransaction` API where the commit call is implicit.

Any thoughts?

source/transactions-convenient-api/transactions-convenient-api.rst
jmikola: If they're not able to make the change for 4.2, I suppose it doesn't much matter since we'll need to add code to check the error code directly. I think this is worth raising with the server team, though, as I was under the impression this was the intended purpose for error labels (getting away from requiring drivers to inspect error codes).

jmikola:
> I think we could improve this in one of two ways:
> 1. We can ask the server to support `maxTimeMS` on the initial `startTransaction: true` command. This way the server can apply the `maxTimeMS` limit the lifetime of the entire transaction similar to how `maxTimeMS` works on a cursor.
> 2. We can rename this option to emphasize that it only applies to the commit, like `maxCommitTimeMS`.

Spreading a single `maxTimeMS` option over the entire transaction feels like it might overlap with client-side operation timeouts. Pursuing that now might result in creating a stop-gap solution (akin to pursuing a configurable timeout for `withTransaction`). I propose we table that idea until the client-side operation spec is resumed.

Renaming the option seems reasonable, since that would improve the self-documenting nature of the option. I do wonder if inventing a name like `maxCommitTimeMS` might result in the server wanting to follow a similar path as they did with `maxAwaitTimeMS`. We originally introduced `maxAwaitTimeMS` as an option for find operations to be used as the value for `maxTimeMS` on subsequent getMore commands. With [SERVER-34277](https://jira.mongodb.org/browse/SERVER-34277), the server renamed the getMore option to `maxAwaitTimeMS`. I assume `maxTimeMS` on commitTransaction commands is new for 4.2 -- if so, should we preemptively consult the server team and ask if they'd like to launch with `maxCommitTimeMS` out of the gate?

ShaneHarvey:
> Spreading a single maxTimeMS option over the entire transaction feels like it might overlap with client-side operation timeouts. Pursuing that now might result in creating a stop-gap solution (akin to pursuing a configurable timeout for withTransaction). I propose we table that idea until the client-side operation spec is resumed.

Allowing a user to configure `maxCommitTimeMS` also overlaps with client-side operation timeout, right? I think we should leave the idea on the table because, imo, spreading a single maxTimeMS option over the entire transaction seems more useful and easier to understand/document than `maxCommitTimeMS`. My biggest concern is that this approach would add extra work on the server side.

> With SERVER-34277, the server renamed the getMore option to maxAwaitTimeMS

FYI It doesn't look like that work has actually happened  but I see your point. If we settle on renaming the option to `maxCommitTimeMS` then I will open a server ticket about the command parameter name.

> I assume maxTimeMS on commitTransaction commands is new for 4.2 -- if so, should we preemptively consult the server team and ask if they'd like to launch with maxCommitTimeMS out of the gate?

`maxTimeMS` appears to be supported on 4.0 replica sets too. I'll have to sync with Aly and someone on the server to see if this is expected (and supported).

jmikola:
> Allowing a user to configure `maxCommitTimeMS` also overlaps with client-side operation timeout, right?

I don't think this overlaps with client-side operation timeouts any more than specifying `maxTimeMS` on any other individual operation.

In contrast, I think ehe idea of spreading a single `maxTimeMS` across multiple operations (or an entire transaction) overlaps _heavily_ with client-side operation timeouts. If the server were to implement that, I think it may make driver implementations of client-side operation timeouts less complex, as we wouldn't need to adjust the `maxTimeMS` in a series of operations on our own. For instance, if a user wants to run a complete transaction (or `withTransaction` callable) over ten seconds, drivers may send a 9.x-second `maxTimeMS` for the first operation, reduce it accordingly for the next operation, and continue doing so until the last operation -- as a means of allowing the server to abort earlier and allow us to catch a time overflow closer to the user's originally specified limit.

ShaneHarvey:
Yeah I agree Jeremy. Let's scrap option 1.

jmikola:
Repeating my comment from a Slack message earlier in the week:

> So it doesn't go overlooked:
>
>> if so, should we preemptively consult the server team and ask if they'd like to launch with maxCommitTimeMS out of the gate?
>
> If we're going to use maxCommitTimeMS, we should see if the server wants to rename their startTransaction option to that as well. This way we wouldn't have to handle the rename purely in the driver like we did with maxAwaitTimeMS

ShaneHarvey:
> If I understand this correctly, the server team already implemented this in 4.0.x, so perhaps there's no change of renaming it to maxCommitTimeMS. It would probably be more headache for a driver to need to determine whether to send maxTimeMS or maxCommitTimeMS depend on the minor version within 4.0.x (since we can't depend on a wire version bump if this was just for 4.2).

I'm fairly certain that `maxTimeMS` is support on _all_ commands. I believe it was implemented in 3.4 or 3.6 but I can't find the server ticket to link to at the moment.

ShaneHarvey:
Thanks Jeff! For others, this PR is ready to (re)review. I'm not expecting anymore changes.

source/transactions-convenient-api/transactions-convenient-api.rst
jmikola: @ShaneHarvey: Did you ever end up creating a SERVER ticket for this? If so, can you cross-reference it with [SPEC-1206](https://jira.mongodb.org/browse/SPEC-1206)?

source/transactions/transactions.rst
```diff
@@ -290,6 +302,17 @@ See `Why is readPreference part of TransactionOptions?`_.
...
 
+maxCommitTimeMS
+^^^^^^^^^^^^^^^
+
+The maximum amount of time to allow a single commitTransaction command to run.
+
+This option is only sent with the commitTransaction command(s) and only if the
+caller explicitly provides a value. The default is to not send a value.
+
+Note, this option is an alias for the `maxTimeMS` commitTransaction command
```
jmikola: Double backticks for code formatting on `maxTimeMS` here.

source/transactions/tests/error-labels.yml
```diff
@@ -959,3 +959,247 @@ tests:
       collection:
         data:
           - _id: 1
+
+  - description: do not add unknown commit label to MaxTimeMSExpired inside transactions
```
jmikola: In the transactions spec, you did add directions to add the UnknownTransactionCommitResult label if commitTransaction encounters a MaxTimeMSExpired error. Is the spec silent about errors encountered on other commands within a transaction, or am I just missing something that's not rendered within this PR diff?

Just want to confirm that we're not trying to test undefined behavior here.

I think the only other test in this file related to not adding labels is for DuplicateKeyException.

source/transactions/transactions.rst
```diff
@@ -12,7 +12,7 @@ Driver Transactions Specification
 :Status: Accepted (Could be Draft, Accepted, Rejected, Final, or Replaced)
 :Type: Standards
 :Minimum Server Version: 4.0 (The minimum server version this spec applies to)
-:Last Modified: 19-February-2019
+:Last Modified: 24-April-2019
```
jmikola: You might want to bump all of these dates before merging. No worries if not.

source/transactions-convenient-api/tests/commit-writeconcernerror.yml
```diff
@@ -193,3 +193,24 @@ tests:
...
+            codeName: MaxTimeMSExpired
+            errmsg: "operation exceeded time limit"
+    operations:
+      - <<: *operation
+        result:
+          errorCodeName: MaxTimeMSExpired
+          errorLabelsContain: ["UnknownTransactionCommitResult"]
+          errorLabelsOmit: ["TransientTransactionError"]
+    expectations: *expectations_without_retries
+    # failCommand with writeConcernError still applies the write operation(s)
```
jmikola: Under what circumstances would MaxTimeMSExpired get reported via a write concern error? I'm curious how that differs from a write concern timeout.

Is trusting that the write is still applied just something unique to using the fail point? I assume in a real world scenario, we wouldn't know whether (or how much of) the write might have been applied.

source/transactions-convenient-api/transactions-convenient-api.rst
ShaneHarvey: Thanks the reminder! I just opened [SERVER-41125](https://jira.mongodb.org/browse/SERVER-41125) for this and linked it to SPEC-1206.

source/transactions/transactions.rst
ShaneHarvey: Done.

source/transactions/transactions.rst
ShaneHarvey: Done.

source/transactions/tests/error-labels.yml
ShaneHarvey: This test ensures that drivers only add the "UnknownTransactionCommitResult" error label when `commitTransaction` fails with a maxTimeMS error and not when other commands fail with maxTimeMS errors.

>  Is the spec silent about errors encountered on other commands within a transaction, or am I just missing something that's not rendered within this PR diff?

Errors from other commands are discussed in the [TransientTransactionError section](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#transienttransactionerror). I don't think this PR needs to explicitly say that non-commitTransaction commands that fail with MaxTimeMSExpired must not be labelled.

source/transactions-convenient-api/tests/commit-writeconcernerror.yml
ShaneHarvey: > Under what circumstances would MaxTimeMSExpired get reported via a write concern error? I'm curious how that differs from a write concern timeout.

wtimeout only applies when the server is waiting for a writeConcern. maxTimeMS applies to the entire command execution (including the writeConcern wait). So when maxTimeMS is exceeded while waiting for writeConcern the server will return it as a writeConcernError:  ` {ok:1, writeConcernError: {code: 50, codeName: "MaxTimeMSExpired"}}`.

> Is trusting that the write is still applied just something unique to using the fail point? I assume in a real world scenario, we wouldn't know whether (or how much of) the write might have been applied.

It is not unique to the failpoint and this would also be the case in a real world scenario. MaxTimeMSExpired in a writeConcernError means that the write was applied on the primary but the given writeConcern could not be satisfied within the remaining maxTimeMS. We still label this  UnknownTransactionCommitResult because we don't know if the commit has satisfied the write concern; we may later find out that the primary stepped down and the write was rolled back or we may find out that the write succeeded on a majority.

jmikola: Thanks for the helpful explanation!

PR #541, opened 2019-04-26
Fix server version constraints for startAtOperationTime prose test

divjotarora:
This is necessary because a change stream on server versions >= 4.0.7 will cache the `postBatchResumeToken` from the initial aggregate response and use that for resuming rather than the cached operation time.

### DIFF
--- a/source/change-streams/tests/README.rst
+++ b/source/change-streams/tests/README.rst
@@ -156,7 +156,7 @@ The following tests have not yet been automated, but MUST still be tested
 #. ``ChangeStream`` will perform server selection before attempting to resume, using initial ``readPreference``
 #. Ensure that a cursor returned from an aggregate command with a cursor id and an initial empty batch is not closed on the driver side.
 #. The ``killCursors`` command sent during the "Resume Process" must not be allowed to throw an exception.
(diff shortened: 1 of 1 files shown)


source/change-streams/tests/README.rst
```diff
@@ -156,7 +156,7 @@ The following tests have not yet been automated, but MUST still be tested
 #. ``ChangeStream`` will perform server selection before attempting to resume, using initial ``readPreference``
 #. Ensure that a cursor returned from an aggregate command with a cursor id and an initial empty batch is not closed on the driver side.
 #. The ``killCursors`` command sent during the "Resume Process" must not be allowed to throw an exception.
-#. ``$changeStream`` stage for ``ChangeStream`` against a server ``>=4.0`` that has not received any results yet MUST include a ``startAtOperationTime`` option when resuming a changestream.
+#. ``$changeStream`` stage for ``ChangeStream`` against a server ``>=4.0`` and ``<4.0.7`` that has not received any results yet MUST include a ``startAtOperationTime`` option when resuming a changestream.
```
jmikola: LGTM.

FWIW, my implementation in the C driver forgoes any server version checking and instead decides what to assert based on whether or note the initial aggregate reply includes a `cursor.postBatchResumeToken` field.

PR #542, opened 2019-05-02
SPEC-1267: Add the ability to specify a pipeline to an update command

jstewart-mongo:


source/crud/crud.rst
```diff
@@ -1762,6 +1762,7 @@ Q: Where is ``singleBatch`` in FindOptions?
 Changes
 =======
 
+* 2019-05-01: Specify a document or pipeline for commands with updates.
```
behackett: Here, and possibly someplace else, it should be documented that the update can only be a pipeline with MongoDB 4.2+.

source/crud/tests/v2/bulkWrite-arrayFilters.yml
```diff
@@ -65,3 +65,60 @@ tests:
         data:
           - {_id: 1, y: [{b: 2}, {b: 2}]}
           - {_id: 2, y: [{b: 0}, {b: 2}]}
+  -
+    description: "BulkWrite with arrayFilters using update pipelines"
```
behackett: This test has to be skipped on MongoDB < 4.2.

jmikola: Please move this to a new test file and use `minServerVersion`. See https://github.com/mongodb/specifications/blob/master/source/crud/tests/v2/bulkWrite-arrayFilters.yml for prior art.

I'd also suggest using the exact server release in the corresponding SERVER ticket's fixVersion. Using a specific version (rather than 4.2.0) will make it easier for drivers to automatically start running this test when the feature first becomes available in the "latest" server binary.

source/crud/examples/java/src/main/java/examples/MongoCollectionUsageExample.java
```diff
@@ -126,11 +126,23 @@ public static void usageExample(MongoCollection<Document> col) {
         col.updateOne(new Document("x", 1), new Document("$set", new Document("x", 2)),
                       new UpdateOptions().upsert(true));
 
+        // using a pipeline for the update parameter
+        col.updateOne(new Document("x", 1), asList(new Document("$set", new Document("x", 2))));
```
jmikola: AFAIK, the `crud/examples` directory is obsolete. I'd propose just deleting the entire directory (and any references to it) in a subsequent PR.

Most drivers now coordinate with the docs team to maintain CRUD examples in our test suites, which are then published in the MongoDB manual. We should rely on that going forward. If the docs team has not already done so, I'd suggest reaching out to ensure they have a tracking ticket to request examples for update pipelines.

source/crud/tests/v2/bulkWrite-arrayFilters.yml
```diff
@@ -65,3 +65,60 @@ tests:
...
+        name: "bulkWrite"
+        arguments:
+          requests:
+            -
+              # UpdateOne when one document matches arrayFilters
+              name: "updateOne"
+              arguments:
+                filter: {}
+                update: [ { $set: { "y.$[i].b": 2 } } ]
+                arrayFilters: [ { "i.b": 3 } ]
```
jmikola: I'm not familiar with the feature, but does the set of `arrayFilters` simply apply to _all_ updates in the pipeline now (where it previously only applied to one)?

source/crud/tests/v2/bulkWrite-arrayFilters.yml
```diff
@@ -65,3 +65,60 @@ tests:
...
+      -
+        name: "bulkWrite"
+        arguments:
+          requests:
+            -
+              # UpdateOne when one document matches arrayFilters
+              name: "updateOne"
+              arguments:
+                filter: {}
+                update: [ { $set: { "y.$[i].b": 2 } } ]
```
jmikola: I think it'd make sense for these example to at least specify two updates in the array, instead of wrapping a single update document in a BSON array. If the server has already implemented this feature, there may be an example to use in their code review's JS tests.

source/crud/tests/v2/bulkWrite-arrayFilters.yml
```diff
@@ -65,3 +65,60 @@ tests:
         data:
           - {_id: 1, y: [{b: 2}, {b: 2}]}
           - {_id: 2, y: [{b: 0}, {b: 2}]}
+  -
+    description: "BulkWrite with arrayFilters using update pipelines"
+    operations:
+      -
+        name: "bulkWrite"
```
jmikola: The final PR should also add new tests files (with `minServerVersion`) for the `updateOne`, `updateMany`, and `findOneAndUpdate` CRUD methods. Those can be created once we're all set with this `bulkWrite` test, since I expect it will just entail some copying and slight modifications to APM.

source/crud/crud.rst
```diff
@@ -1655,7 +1655,7 @@ Find And Modify
 Update vs. Replace Validation
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The ``update`` family of operations require that the update document parameter MUST have only atomic modifiers. In practice, this means that introspection needs to happen on that document to enforce this. However, it is enough to only check the first element in the document. If it begins with a ``$`` sign and the rest of the document's elements do not, the server will throw an error. Note that it is required that an update document have at least one atomic modifier.
+The ``update`` family of operations require that the update parameter, which can be either a document or a pipeline, MUST have only atomic modifiers. In practice, this means that introspection needs to happen on the document or pipeline to enforce this. However, it is enough to only check the first element in the update documents. If the element begins with a ``$`` sign and the rest of the document's elements do not, the server will throw an error. Note that it is required that the documents in the update parameter have at least one atomic modifier.
```
jyemin: I don't see any necessity for drivers to inspect the documents in a pipeline in the way that they do for single document update.  There's nothing it could be besides an aggregation pipeline.

jmikola: If I understand your comment correctly, we can trust that the server will require all elements in this BSON array (i.e. pipeline) to consist of atomic operators so there is no need for the driver to inspect elements and reject replacement documents as we do for `updateOne` and `updateMany` operations (and that purely to satisfy CRUD API and force users to use `replaceOne`). If so, I agree that relying on the server to raise an error makes sense.

source/crud/crud.rst
jstewart-mongo: I have added this in the change comment.

source/crud/tests/v2/bulkWrite-arrayFilters.yml
jstewart-mongo: I have added a new test file - `updateWithPipelines.yml`.

source/crud/examples/java/src/main/java/examples/MongoCollectionUsageExample.java
jstewart-mongo: I have reverted all changes to the examples files.

source/crud/tests/v2/bulkWrite-arrayFilters.yml
jstewart-mongo: `arrayFilters` is not a supported option with pipelines. I have removed them from the update pipelines tests.

source/crud/tests/v2/bulkWrite-arrayFilters.yml
jstewart-mongo: The new update test file has contrived tests that have two update documents in the array. James Whalin commented that "Given we restrict users to the $addFields, $project and $replaceRoot stages currently, I am not sure if there are any good real-life multi-stage use cases."

source/crud/tests/v2/bulkWrite-arrayFilters.yml
jstewart-mongo: The new test file has test cases for these operations.

source/crud/crud.rst
jstewart-mongo: I have removed the comment.

source/crud/tests/v2/updateWithPipelines.yml
```diff
@@ -0,0 +1,82 @@
...
+          - { _id: 1, x: 1, foo: 1 }
+          - { _id: 2, x: 2, foo: 1 }
+  -
+    description: "FindOneAndUpdate using pipelines"
+    operations:
+      -
+        name: "findOneAndUpdate"
+        arguments:
+            filter: { y: 1 }
+            update: [ { $project: { x: 1 } }, { $addFields: { foo: 1 } } ]
```
jmikola: Would it make sense for this to include a `result` assertion like other `findOneAndUpdate` tests?

source/crud/tests/v2/updateWithPipelines.yml
```diff
@@ -0,0 +1,82 @@
...
+          - { _id: 1, x: 1, y: 1, u: {v: 1}, foo: 1 }
+          - { _id: 2, x: 2, y: 1 }
+  -
+    description: "UpdateMany using pipelines"
+    operations:
+      -
+        name: "updateMany"
+        arguments:
+            filter: {}
+            update: [ { $project: { x: 1 } }, { $addFields: { foo: 1 } } ]
```
jmikola: Would it make sense to include a `result` assertion for UpdateManyResult fields as other `updateMany` tests?

source/crud/tests/v2/updateWithPipelines.yml
```diff
@@ -0,0 +1,82 @@
...
+
+tests:
+  -
+    description: "UpdateOne using pipelines"
+    operations:
+      -
+        name: "updateOne"
+        arguments:
+            filter: { y: 1 }
+            update: [ { $replaceRoot: { newRoot: "$t" } }, { $addFields: { foo: 1 } } ]
```
jmikola: Would it make sense to include a `result` assertion for UpdateOneResult fields as other `updateOne` tests?

source/crud/tests/v2/updateWithPipelines.yml
jstewart-mongo: Added `result` assertion to the test.

source/crud/tests/v2/updateWithPipelines.yml
jstewart-mongo: Added `result` assertion to the test.

source/crud/tests/v2/updateWithPipelines.yml
jstewart-mongo: Added `result` assertion to the test.

source/crud/tests/v2/updateWithPipelines.yml
```diff
@@ -16,6 +16,14 @@ tests:
         arguments:
             filter: { y: 1 }
             update: [ { $replaceRoot: { newRoot: "$t" } }, { $addFields: { foo: 1 } } ]
+        result:
+          deletedCount: 0
+          insertedCount: 0
+          insertedIds: {}
```
jmikola: `deletedCount`, `insertedCount` and `insertedIds` are for a BulkWriteResult. Since this operation is `updateOne`, just the UpdateResult fields below will suffice.

source/crud/tests/v2/updateWithPipelines.yml
```diff
@@ -16,6 +16,14 @@ tests:
...
+          matchedCount: 1
+          modifiedCount: 1
+          upsertedCount: 0
+          upsertedIds: {}
```
jmikola: For UpdateResult, this would be `upsertedId` singular. The plural and object type is only for BulkWriteResult where we might need a map. That said, existing `updateOne` tests that don't actually upsert (i.e. `upsertedCount` is zero) simply omit this field so I'd suggest the same here.

IIRC, even though multiple documents could match the criteria, the `matchedCount` in the UpdateResult will stop at one for `updateOne` operations.

source/crud/tests/v2/updateWithPipelines.yml
```diff
@@ -40,6 +48,14 @@ tests:
...
             filter: {}
             update: [ { $project: { x: 1 } }, { $addFields: { foo: 1 } } ]
+        result:
+          deletedCount: 0
+          insertedCount: 0
+          insertedIds: {}
+          matchedCount: 2
+          modifiedCount: 2
+          upsertedCount: 0
+          upsertedIds: {}
```
jmikola: Same suggestions as above for `updateOne`.

source/crud/tests/v2/updateWithPipelines.yml
```diff
@@ -30,7 +38,7 @@ tests:
     outcome:
       collection:
         data:
-          - { _id: 1, x: 1, y: 1, u: {v: 1}, foo: 1 }
+          - { _id: 1, u: {v: 1}, foo: 1 }
```
jmikola: On second thought, the criteria for this `updateOne` may be ambiguous since it can match either of the two data fixture documents (both have `y: 1`). In practice, it's likely we'll match the first inserted document encountered by a scan but that's not technically guaranteed. I think it'd be safer to select by either `_id` or `x`. This has come up in previous CRUD tests and I recall we've made similar changes to avoid undefined behavior.

source/crud/tests/v2/updateWithPipelines.yml
```diff
@@ -64,6 +80,14 @@ tests:
...
             filter: { y: 1 }
             update: [ { $project: { x: 1 } }, { $addFields: { foo: 1 } } ]
+        result:
+          deletedCount: 0
+          insertedCount: 0
+          insertedIds: {}
+          matchedCount: 1
+          modifiedCount: 1
+          upsertedCount: 0
+          upsertedIds: {}
```
jmikola: The `result` for `findOneAndUpdate` will be the pre-modified document by default. In this case, the `result` should be one of the original data fixture documents instead of this UpdateResult. If you prefer to use the post-modified document for the `result` (and use the same value you'll have below in the outcome), you can add the `returnDocument: After` option to the `findOneAndUpdate` arguments to do so.

Since this also selects only one document, my previous comment about an ambiguous filter also applies here. I'd suggest selecting by `_id` or `x` to ensure we match only a single document.

source/crud/tests/v2/updateWithPipelines.yml
jstewart-mongo: I have removed `deletedCount`, `insertedCount` and `insertedIds`.

source/crud/tests/v2/updateWithPipelines.yml
jstewart-mongo: I have omitted the `upsertedIds` field.

source/crud/tests/v2/updateWithPipelines.yml
jstewart-mongo: Done

source/crud/tests/v2/updateWithPipelines.yml
jstewart-mongo: I have changed this to select `_id` instead of `y`.

source/crud/tests/v2/updateWithPipelines.yml
jstewart-mongo: Added the `returnDocument` option and select `_id`.

jmikola: Unless I'm misreading the diff, the `result` assertion still needs to be changed to the returned document instead of an UpdateResult.

jstewart-mongo: I missed this. Since the result would be the same as the outcome, I have removed `result`. No need to check it twice.

PR #544, opened 2019-05-08
SPEC-1159 ServerSessions involved in network errors are discarded

ShaneHarvey:
Jira: https://jira.mongodb.org/browse/SPEC-1159

POC implemented in PyMongo here: https://github.com/ShaneHarvey/mongo-python-driver/tree/SPEC-1159 and tested here https://evergreen.mongodb.com/version/5cdb32440305b96a2bbaccd1.

Changes:
- Introduces the concept of "dirty" ServerSessions. A session is marked as dirty once it has been involved in a network error. Dirty ServerSessions are discarded instead of being returned to the pool. 
- Adds spec tests that re-use the transaction spec test format and runner.
- Adds four new meta spec test operations to facilitate driver testing: "assertDifferentLsidOnLastTwoCommands", "assertSameLsidOnLastTwoCommands", "assertSessionDirty", and "assertSessionNotDirty".
- Adds `operation.error` field to the transaction spec test format. The field is borrowed from the retryable reads spec test format.

Done:
- [x] Discuss and decide whether we allow using an explicit session that is marked dirty. This approach still allows parallel operations to be running on the server when using explicit sessions which isn't great but at least they will be _related_ operations. Some (in my opinion bad) alternatives:
 - Drivers can raise an error when a dirty session is used for a later operation. I don't like this idea because how would an application recover from this kind of error?
 - A ClientSession can discard its dirty ServerSession and replace it with a new ServerSession from the pool. This seems wonky and I'm not sure it work be correct with respect to transactions.


ShaneHarvey:
The "Dirty session is discarded integration test" passes everywhere except on 3.6 sharded clusters where it fails pretty consistently. It is failing because the insert command takes too long and triggers the configured socket timeout which indicates that the insert is blocking behind the failed updateOne command. I'm not sure how that's possible because they are using different server sessions...  I'll either need to figure out why, or just delete the test.

ShaneHarvey:
I'm not sure why "Dirty session is discarded integration test" fails on 3.6 sharded clusters so I've just opted to remove it altogether. Even if we skip it on 3.6, I'm worried it will be racy elsewhere.

source/sessions/driver-sessions.rst
```diff
@@ -1197,6 +1212,25 @@ Instead, we require users to pass session as a parameter to each function::
...
+-----------------------------------------------------------------------------------
+
+When a network error is encountered when executing an operation with a
+``ClientSession``, the operation may be left running on the server. Re-using
+this ``ServerSession`` can lead to parallel operations which violates the
+rule that a session must be used sequentially. This results in multiple
+problems:
+
+#. killSessions to end an earlier operation would surprisingly also end a
+   later operation.
```
jmikola: What is the risk in `killSessions` ending a later operation? Is this only when dealing with implicit sessions? If you're using explicit sessions, regardless of whether the session is dirty or not, I would expect the user to realize that kill the session could end any operation that has not received an acknowledged response from the server. Some examples (with implicit or explicit):

**Implicit session without dirty tracking.** I execute a long-running updateMany and the socket dies. The operation is still running. I execute a second updateMany, which uses the same implicit session, and it succeeds. I then kill the implicit session's LSID and only the first operation is aborted. Even if multiple long-running operations were executed within that session, only those that encountered socket errors (or were otherwise unacknowledged and might have been executing asynchronously) will be aborted. I don't see how this relates to "ending a later operation".

**Explicit session without dirty tracking.** Similarly, I would expect killSessions to possibly abort operations that I've executed but have yet to receive a response (the could still be running). I'm not sure what a "later operation" would be in this context, either.

source/sessions/driver-sessions.rst
```diff
@@ -1197,6 +1212,25 @@ Instead, we require users to pass session as a parameter to each function::
...
+#. An otherwise unrelated operation that just happens to use that same server
+   session will potentially block waiting for the previous operation to
+   complete. For example, a transactional write will block a subsequent
+   transactional write.
```
jmikola: Am I correct in assuming that the server doesn't prohibit a session from being attached to two (or more) concurrent operations? Or does such a prohibition depend on whether we're operating within a transaction?

Outside of the context of a network error, is it possible that drivers could already spawn parallel operations on the server using the same session? One scenario that comes to mind would be if a driver actually executed write commands from an unordered bulk write in parallel (either in or outside of a transaction), and all ensuing write commands use the same LSID.

source/sessions/driver-sessions.rst
```diff
@@ -1197,6 +1212,25 @@ Instead, we require users to pass session as a parameter to each function::
...
+
+TODO: Justify why we allow applications to continue using an explicit session
+after a network error.
```
jmikola: I understand why we might want to never re-use a dirty session for implicit use, but I don't see the problem with allowing applications to continue using it for explicit use. Quoting your previous comment from the OP:

> This approach still allows parallel operations to be running on the server when using explicit sessions which isn't great but at least they will be related operations. Some (in my opinion bad) alternatives:
> * Drivers can raise an error when a dirty session is used for a later operation. I don't like this idea because how would an application recover from this kind of error?
> * A ClientSession can discard its dirty ServerSession and replace it with a new ServerSession from the pool. This seems wonky and I'm not sure it work be correct with respect to transactions.

Based on your earlier text, I might assume that the server prohibits concurrent execution of multiple operations using the same session (regardless of transactions).

I don't think we need to prohibit future use of dirty sessions, and I agree it would be disruptive for an application to deal with that. If the previous operation utilized retryable read/write behavior after a socket error and ultimately succeeded, the app shouldn't have to worry about anything blocking. If the previous operation failed completely, the app would have observed a socket error and is in a place to infer that future operations with the same session might block.

Silently replacing the ClientSession's underlying ServerSession does not seem kosher, as it violates the previous expectation that operations with the same ClientSession will all use the same LSID.

source/sessions/tests/README.rst
```diff
@@ -0,0 +1,71 @@
...
+Transactions spec tests.
+
+Several prose tests, which are not easily expressed in YAML, are also presented
+in the Driver Sessions Spec. Those tests will need to be manually implemented
+by each driver.
+
+Test Format
+===========
+
+The same as the Transactions spec tests format.
```
jmikola: I'd suggest a relative link to the `transactions/tests` directory here.

source/sessions/tests/README.rst
```diff
@@ -0,0 +1,71 @@
...
+
+Special Test Operations
+```````````````````````
+
+Certain operations that appear in the "operations" array do not correspond to
+API methods but instead represent special test operations. Such operations are
+defined on the "testRunner" object and are documented in the Transactions spec
```
jmikola: Just to confirm, is this the first use of "testRunner"? I seem to recall seeing it discussed before, but that might have been in the context of the FLE or Live Failover spec. I know Live Failover is no longer using this, but on the off chance FLE has some concept of it I'd suggest confirming with @kevinAlbs to make sure everything is consistent.

source/sessions/tests/README.rst
```diff
@@ -0,0 +1,71 @@
...
+test. Additional, session test specific operations are documented here:
+
+assertServerSessionPoolSize
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The "assertServerSessionPoolSize" operation instructs the test runner to
+assert that the test's MongoClient has a ServerSession pool containing the
+given number of ServerSessions::
+
+      - name: assertServerSessionPoolSize
```
jmikola: If session pools were configurable, I suppose we could run tests with a pool size of one and assert that subsequent commands didn't reuse the same LSID (proving that a dirty session was not reused). But session pools are not configurable, so this seemingly requires inspection of the driver's internal state.

I'm not a fan of introducing spec test operations that require internal APIs, as it makes this difficult if not impossible to implement in wrapping drivers (e.g. anything using libmongoc). Would this be the first case of a spec test requiring this (outside of perhaps the URI tests)?

That said, I wouldn't veto this change if it's necessary.

source/sessions/driver-sessions.rst
```diff
@@ -1221,3 +1255,4 @@ Change log
 :2018-06-07: Document that estimatedDocumentCount does not support explicit sessions
 :2018-07-19: Justify why session must be an explicit parameter to each function
 :2018-10-11: Session pools must be cleared in child process after fork
+:2019-05-01: v1.6.0 A ServerSession that is involved in a network error MUST be discarded
```
jmikola: I don't think "v1.6.0" is necessary here. First time I've seen a version included in the change log.

source/sessions/driver-sessions.rst
```diff
@@ -859,8 +871,11 @@ Algorithm to return a ServerSession instance to the server session pool
    from the end of the queue and discarded (or allowed to be garbage collected)
 
 2. Then examine the server session that is being returned to the pool and:
+    * If this session is marked dirty (ie it was involved in a network error)
```
jmikola: "i.e.", unless "ie" is commonly used elsewhere in specs.

source/sessions/driver-sessions.rst
ShaneHarvey: When both operations on the same session are part of a transaction (retryable writes/transactions), the server does not allow concurrent execution; the operations will be executed serially. 

When one (or both) operations on the same session are not transactional, the server allows concurrent execution.

> Outside of the context of a network error, is it possible that drivers could already spawn parallel operations on the server using the same session?

I am not aware of any other cases.

>  One scenario that comes to mind would be if a driver actually executed write commands from an unordered bulk write in parallel (either in or outside of a transaction), and all ensuing write commands use the same LSID.

This would be a bug in the driver and retryable writes would not work correctly.

source/sessions/driver-sessions.rst
ShaneHarvey: > but I don't see the problem with allowing applications to continue using it for explicit use.

I am also advocating that we allow applications to continue using explicit dirty sessions. I just wanted to highlight that this can still lead to parallel operations on the server and get everyone on the same page. Sorry if that wasn't clear.

> I don't think we need to prohibit future use of dirty sessions, and I agree it would be disruptive for an application to deal with that.
> Silently replacing the ClientSession's underlying ServerSession does not seem kosher, as it violates the previous expectation that operations with the same ClientSession will all use the same LSID.

Agreed on both points.

source/sessions/tests/README.rst
ShaneHarvey: The transaction spec already uses this concept for targetedFailPoint/assertSessionPinned/assertSessionUnpinned: https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst#special-test-operations

source/sessions/tests/README.rst
ShaneHarvey: Okay, I suspected that this would would hard or impossible to implement in various drivers. Do you have any suggestions for alternative tests?

Perhaps I should revive [this integration test](https://github.com/mongodb/specifications/pull/544/commits/cacd9a2438de793a92b1bf81c55411c2cb5c8d90#diff-1d90a5641b222162674af2ff11f1fea6L13) which indirectly tests that drivers discard dirty sessions. (I removed it because of a race condition in 3.6 sharded clusters)

source/sessions/driver-sessions.rst
ShaneHarvey: Done

source/sessions/driver-sessions.rst
ShaneHarvey: Done

source/sessions/tests/README.rst
ShaneHarvey: Done.

source/sessions/driver-sessions.rst
```diff
@@ -1197,6 +1212,25 @@ Instead, we require users to pass session as a parameter to each function::
...
+problems:
+
+#. killSessions to end an earlier operation would surprisingly also end a
+   later operation.
+#. An otherwise unrelated operation that just happens to use that same server
+   session will potentially block waiting for the previous operation to
+   complete. For example, a transactional write will block a subsequent
+   transactional write.
+
+TODO: Justify why we allow applications to continue using an explicit session
```
jyemin: Do you plan to do this before closing the PR?  A bit weird to leave a TODO in a spec.

source/sessions/tests/README.rst
jyemin: How about a test like this:

- set failpoint that closes the connection
- execute op 1
- execute op 2
- assert that the commandStartedEvent for op1 and op2 have different lsid field values

source/sessions/driver-sessions.rst
```diff
@@ -1228,8 +1228,26 @@ problems:
    complete. For example, a transactional write will block a subsequent
    transactional write.
 
-TODO: Justify why we allow applications to continue using an explicit session
-after a network error.
+Why do retry automatic retry attempts re-use a dirty implicit session?
```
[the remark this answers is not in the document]
ShaneHarvey: Done.

source/sessions/driver-sessions.rst
```diff
@@ -1228,8 +1228,26 @@ problems:
...
+----------------------------------------------------------------------
+
+The retryable writes spec requires that both the original and retry attempt
+use the same server session. The server will block the retry attempt until the
+initial attempt completes at which point the retry attempt will continue
+executing.
+
+For retryable reads that use an implicit session, drivers could choose to use a
+new server session for the retry attempt however this would loose the
```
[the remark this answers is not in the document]
ShaneHarvey: Done.

source/sessions/tests/README.rst
ShaneHarvey: Like this?:

```
    failPoint:
        mode: { times: 1 }
        data:
            failCommands: ["insert"]
            closeConnection: true
    operations:
      - name: insertOne
        object: collection
        arguments:
          document: {_id: 2}
        result:
          insertedId: 2
      - name: insertOne
        object: collection
        arguments:
          document: {_id: 3}
        result:
          insertedId: 3
      - name: assertDifferentLsidOnLastTwoCommands
        object: testRunner
```

jmikola: That test seems reasonable. Provided no other thread is using the client, I believe we can rely on the LIFO design of the implicit session pool.

If this is one test, I'd advocate for prose rather than introduce a `assertDifferentLsidOnLastTwoCommands` operation. A custom operation that depends on analyzing captured APM events for previous commands in the same `operations` array seems like it could be a headache to implement in some test runners that segregate the `operations` and `expectations` blocks.

ShaneHarvey: Why would it be a headache to allow the `operations` block to also have access to APM events?

Personally, I think very specialized testRunner assertions are much more maintainable than add prose tests. In my experience, prose tests take longer to implement and MUCH longer to code review because the reviewer also needs to read each prose test in the spec and make sure each test is properly translated into code. This continues to be a pain when new prose tests are added and old prose tests are updated.

For example, consider how easy it is to review my implementation of this spec change in python: https://github.com/ShaneHarvey/mongo-python-driver/blob/SPEC-1159/test/test_session.py#L1304-L1331
And compare that to the 1000+ lines of prose tests in the same file.

jmikola: > Why would it be a headache to allow the operations block to also have access to APM events?

I had libmongoc C's test runner in mind when I wrote that, but it looks like even though the APM monitoring and operation execution are defined in two different components, the array where observed events are appended is still accessible during operation execution.

Feel free to ignore my second paragraph in https://github.com/mongodb/specifications/pull/544#discussion_r283812069.

ShaneHarvey:
This is ready for another review. I've replace assertSessionPoolSize with assert(Same|Different)LsidOnLastTwoCommands and added a test that encounters a network error on a read operation.

source/sessions/tests/dirty-session-errors.yml
```diff
@@ -35,10 +31,14 @@ tests:
           session: session0
       - name: endSession
         object: session0
-      - name: assertServerSessionPoolSize
-        object: testRunner
+      - &find_with_implicit_session
```
jmikola: This diff removes the `insertOne` operation above, such that the sequence of operations is:

 * assertSessionNotDirty
 * endSession
 * find_with_implicit_session
 * assertSameLsidOnLastTwoCommands

The APM assertions below still include an `insert` command, though. I think the original intention was to use an explicit session for the insert and then return it to the pool to be used by the find command.

source/sessions/tests/dirty-session-errors.yml
```diff
@@ -135,10 +140,9 @@ tests:
           session: session0
       - name: endSession
         object: session0
-      - name: assertServerSessionPoolSize
+      - *find_with_implicit_session
```
jmikola: This diff also removes the `insertOne` operation.

Side question, if you're repeating two forms of inserts (w/ and w/o explicit sessions), would it make sense to create anchors for them as well as you did for `find_with_implicit_session`?

source/sessions/tests/dirty-session-errors.yml
```diff
@@ -182,7 +192,7 @@ tests:
           - {_id: 2}
           - {_id: 3}
 
-  - description: Dirty implicit session is discarded
+  - description: Dirty implicit session is discarded (write)
```
jmikola: In "Dirty implicit session is discarded (read)", you use `times: 2` so that the test can pass w/ and w/o retryable reads. Here, you have `retryWrites: true` enabled in client options. Would it make sense to just disable retryable reads and writes for both tests instead?

I realize this test expects the insert to succeed based on the outcome assertion. Is my question above a non-concern because the retried insert (which succeeds) will use the same implicit LSID and it will only be marked dirty and discarded after the operation ultimately completes or fails?

Separate question: for bulk writes, I assume all writes in the bulk sequence need to use the same session, so the same rules would apply: retries or later commands in the same `bulkWrite` operation will all use the same LSID regardless of any network error encountered along the way (this probably also assumes `ordered: false` to allow for continue-on-error behavior). Perhaps this is covered by existing language you added to the spec: "It is valid for a dirty session to be used for subsequent operations (e.g. an implicit retry attempt)" -- if so, I might suggest adding "later commands in a bulk write" to the e.g. parenthetical.

source/sessions/tests/dirty-session-errors.yml
```diff
@@ -231,9 +236,46 @@ tests:
...
     outcome:
       collection:
         data:
           - {_id: 1}
           - {_id: 2}
+
+  - description: Dirty implicit session is discarded (read)
+
+    # Enable the failpoint with times:2 so that this test can pass with or
+    # without retryable reads.
```
jmikola: Just to confirm, `times: 2` isn't a concern if retryable reads are not available because the test runner is instructed to clear any fail points after each test.

source/sessions/tests/dirty-session-errors.yml
ShaneHarvey: The insertOne operation is still there.

source/sessions/tests/dirty-session-errors.yml
```diff
@@ -0,0 +1,281 @@
...
+          session: session0
+      - name: endSession
+        object: session0
+      - &find_with_implicit_session
+        name: find
+        object: collection
+        arguments:
+          filter: {_id: -1}
+        result: []
+      - name: assertSameLsidOnLastTwoCommands
```
jyemin: Is this redundant? The command started events contain the actual lsid reference (session0) that is expected for each event, so that seems enough of a check

source/sessions/tests/dirty-session-errors.yml
ShaneHarvey: > Would it make sense to just disable retryable reads and writes for both tests instead?

Yes it would make sense but in order to disable retryable reads a driver has to implement retryable reads and I don't want to create that dependency for this spec change. Instead I choose to leave retryReads unspecified and make the read operation fail with/without retryable reads support by using `times:2`.

> Separate question: for bulk writes, I assume all writes in the bulk sequence need to use the same session

Yes that's also true, the latest commit expands the parenthetical.

source/sessions/tests/dirty-session-errors.yml
ShaneHarvey: Yes it's redundant but I included it for clarity.

source/sessions/tests/dirty-session-errors.yml
ShaneHarvey: The insertOne operations are still there.

The latest commit adds anchors for `insert_with_explicit_session`/`insert_with_implicit_session`.

source/sessions/tests/dirty-session-errors.yml
ShaneHarvey: Yep!

source/transactions/tests/README.rst
```diff
@@ -524,6 +529,7 @@ is the only command allowed in a sharded transaction that uses the
 Changelog
 =========
 
+:2019-05-14: Add operation level ``error`` field to assert any error.
```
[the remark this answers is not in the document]
ShaneHarvey: The test "Dirty implicit session is discarded (read)" uses it to assert that a network error is returned. Currently, the transaction spec tests have no way to assert this case so I borrowed the `operation.error` field concept from the retryable reads spec tests: https://github.com/mongodb/specifications/blob/master/source/retryable-reads/tests/README.rst#test-format

source/sessions/tests/dirty-session-errors.yml
jmikola: Ok, I must have misread the diff.

source/transactions/tests/README.rst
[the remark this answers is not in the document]
jmikola: Shane's comment explains why `error` was used in the session tests, but I think Jeff was also asking why this was the only change in the `transactions/` directory. That is because the session spec refers to the transactions spec's test format instead of repeating the section in its own README.

PR #546, opened 2019-05-10
Connection survive step down prose test tweaks

rozza:
Fix not master error code
Test interrupted at shutdown also clears the pool

DRIVERS-552


rozza:
Found a couple of minor typo issues when implementing for Java (rozza/mongo-java-driver#321)

PR #548, opened 2019-05-17
SPEC-1195 Specify effect of client-side errors on in-progress transactions


p-mongo:
http://jira.mongodb.org/browse/SPEC-1195

### DIFF
--- a/source/transactions/tests/README.rst
+++ b/source/transactions/tests/README.rst
@@ -154,6 +154,12 @@ Each YAML file has the following keys:
       multi-document read. If the operation is expected to return an error, the
       ``result`` is a single document that has one or more of the following
       fields:
+      
+      - ``clientError``: A boolean indicating whether the operation is
+        expected to fail due to a client side error (such as a document
+        being too large, or containing invalid keys). If the ``clientError``
+        key is not present but any of the other error keys are present,
+        the error is expected to have come from the server.
 
       - ``errorContains``: A substring of the expected error message.
 
@@ -165,6 +171,13 @@ Each YAML file has the following keys:
 
       - ``errorLabelsOmit``: A list of error label strings that the
         error is expected not to have.
+        
+      - ``state``: A map of expected session states. The keys can be
+        "session0" or "session1", and the values can be any of the states
+        defined in the transactions spec: "starting transaction",
+        "transaction in progress", "transaction comitted", "transaction aborted".
+        If a session is not specified in the session state map, its
+        state should not be asserted.
 
   - ``expectations``: Optional list of command-started events.
 
--- a/source/transactions/tests/README.rst
+++ b/source/transactions/tests/README.rst
@@ -147,19 +147,13 @@ Each YAML file has the following keys:
     - ``arguments``: Optional, the names and values of arguments.
 
     - ``error``: Optional. If true, the test should expect an error or
-      exception.
+      exception. This could be a server-generated or a driver-generated error.
 
     - ``result``: The return value from the operation, if any. This field may
       be a single document or an array of documents in the case of a
       multi-document read. If the operation is expected to return an error, the
       ``result`` is a single document that has one or more of the following
       fields:
-      
-      - ``clientError``: A boolean indicating whether the operation is
-        expected to fail due to a client side error (such as a document
-        being too large, or containing invalid keys). If the ``clientError``
-        key is not present but any of the other error keys are present,
-        the error is expected to have come from the server.
 
       - ``errorContains``: A substring of the expected error message.
 
@@ -171,13 +165,6 @@ Each YAML file has the following keys:
 
       - ``errorLabelsOmit``: A list of error label strings that the
         error is expected not to have.
-        
-      - ``state``: A map of expected session states. The keys can be
-        "session0" or "session1", and the values can be any of the states
-        defined in the transactions spec: "starting transaction",
-        "transaction in progress", "transaction comitted", "transaction aborted".
-        If a session is not specified in the session state map, its
-        state should not be asserted.
 
   - ``expectations``: Optional list of command-started events.
 
@@ -354,6 +341,20 @@ fail point on the mongos server which "session0" is pinned to::
             data:
               failCommands: ["commitTransaction"]
               closeConnection: true
+
+assertSessionTransactionState
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The "assertSessionTransactionState" operation instructs the test runner to
+assert that the transaction state of the given session is equal to the
+specified value. The possible values are as follows: ``no``, ``starting``,
+``in_progress``, ``committed``, ``aborted``.
+
+      - name: assertSessionTransactionState
+        object: testRunner
+        arguments:
+          session: session0
+          state: in_progress
 
 assertSessionPinned
 ~~~~~~~~~~~~~~~~~~~
--- a/source/transactions/tests/README.rst
+++ b/source/transactions/tests/README.rst
@@ -347,7 +347,7 @@ assertSessionTransactionState
 
 The "assertSessionTransactionState" operation instructs the test runner to
 assert that the transaction state of the given session is equal to the
-specified value. The possible values are as follows: ``no``, ``starting``,
+specified value. The possible values are as follows: ``none``, ``starting``,
 ``in_progress``, ``committed``, ``aborted``.
 
       - name: assertSessionTransactionState
--- a/source/transactions/transactions.rst
+++ b/source/transactions/transactions.rst
@@ -3,7 +3,7 @@ Driver Transactions Specification
 =================================
 
 :Spec Title: Driver Transactions Specification
-:Spec Version: 1.5.1
+:Spec Version: 1.5.2
 :Author: Shane Harvey
 :Spec Lead: A\. Jesse Jiryu Davis
 :Advisory Group: A\. Jesse Jiryu Davis, Matt Broadstone, Robert Stam, Jeff Yemin, Spencer Brody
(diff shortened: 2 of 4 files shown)


source/transactions/tests/README.rst
```diff
@@ -162,6 +168,13 @@ Each YAML file has the following keys:
 
       - ``errorLabelsOmit``: A list of error label strings that the
         error is expected not to have.
+        
+      - ``state``: A map of expected session states. The keys can be
```
jmikola: Can this be moved one level up (alongside `result`)? I realize it's being introduced to assert the state of a session _after_ an error occurs, but I don't think it needs to be limited to that purpose. I could see this being used in non-error tests down the line to generally assert the session's state after an operation.

If so, I might suggest `resultingTransactionState` as a field name, which would immediately clarify whether we're asserting the session's transaction state _after_ the operation is executed.

On a separate note, I'm curious whether all drivers can actually implement this. The [Transactions spec](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst) doesn't define any public API for exposing a Session's internal state. For PHP, we did end up adding an [`isInTransaction()`](https://jira.mongodb.org/browse/PHPC-1231) method, which will assist in implementing the [Convenient API for Transactions spec](https://github.com/mongodb/specifications/blob/master/source/transactions-convenient-api/transactions-convenient-api.rst); however, that doesn't report the exact state.

Perhaps this opens a wider discussion about whether or not spec tests should be limiting to using public API?

source/transactions/tests/README.rst
```diff
@@ -162,6 +168,13 @@ Each YAML file has the following keys:
...
+        "session0" or "session1", and the values can be any of the states
+        defined in the transactions spec: "starting transaction",
+        "transaction in progress", "transaction comitted", "transaction aborted".
```
jmikola: The Transactions spec doesn't define these states beyond enumerating them. If we end up using a more descriptive name than "state" for this map, then perhaps we can also use shorter strings (e.g. "starting", "in_progress").

source/transactions/tests/README.rst
```diff
@@ -151,6 +151,12 @@ Each YAML file has the following keys:
       multi-document read. If the operation is expected to return an error, the
       ``result`` is a single document that has one or more of the following
       fields:
+      
+      - ``clientError``: A boolean indicating whether the operation is
+        expected to fail due to a client side error (such as a document
+        being too large, or containing invalid keys). If the ``clientError``
+        key is not present but any of the other error keys are present,
+        the error is expected to have come from the server.
```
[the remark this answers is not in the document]
p-mongo: Happy to change this to whatever the consensus ends up being.

source/transactions/tests/README.rst
p-mongo: Good point regarding state being independent of the result, I can move it once the API is decided on per the other comments.

source/transactions/tests/README.rst
p-mongo: Happy to change to shorter strings. As you point out these states are not defined as constants, hence what to name them was not entirely clear. I went with the long phrases used in the spec as the starting point.

p-mongo:
A point which came up in a face to face discussion: Ruby driver has to have a code change to be compliant with the language proposed here for the transaction spec (not changing transaction state on client error) - specifically, if a large document is provided as the first insert in a transaction, the driver currently changes transaction state prior to (attempting to) serialize the document.

source/transactions/tests/README.rst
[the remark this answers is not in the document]
jmikola: Thinking about this some more, I wonder if a `clientError` assertion is even necessary. I would think that the combination of a malformed insert document _and_ the session state _not_ transitioning would be enough to guarantee compliance with [SPEC-1195](https://jira.mongodb.org/browse/SPEC-1195). We could even couple this with another test that asserts that the session state _is_ advanced after encountering a non-client error (e.g. socket error using a fail point).

This invites an extra spec test, but would save all drivers from having to implement more test runner logic. If so, I might propose renaming `errors-client.*` to `errors-session_state.*` or something to that effect that indicates all tests are related to session state transitions (or lack thereof).

jmikola:
I spoke with @behackett about this today and he agreed that this isn't a must-have for 4.2 but suggested finding some way to get this merged rather than invite bit rot by having the PR remain open for an extended period time. (@jyemin: this the convo you missed).

What do you think about using `skipReason` with a TODO item for relying on the Transactions spec providing some public API to expose a session's internal state? Alternatively, forget `skipReason` and drivers that don't provide an API for accessing the session's current state could just skip the test file manually. This relates to my earlier comment about renaming the test file to indicate that it concerns itself with session states rather than just "error-client".

IanWhalen:
@p please proceed with changing this spec as per the second paragraph of Jeremy's final comment.

p-mongo:
Okay, if I rename the file to `error-client-session-state`, would that be OK @jmikola ?

source/transactions/tests/README.rst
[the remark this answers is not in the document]
jmikola: Bumping this, as I'm still dubious that we need new spec test syntax to differentiate a client error from any other error. The non-negotiable dependency of these tests seems to be some facility to query the session for its transaction state and assert that it _hasn't_ changed.

To build upon this, I suppose the new test file could also include a test that asserts the state change for a non-client error. Again, we wouldn't need to inspect the error itself -- simply configure the fail point to trigger a runtime error and rely on the state assertion. 

Does anyone else see the value in differentiating the error type? My objective here is to limiting the amount of new syntax we introduce unless it's absolutely necessary.

source/transactions/tests/README.rst
jmikola: Shorter strings works for me. If we use "transaction" in the spec test field name (as I suggested in https://github.com/mongodb/specifications/pull/548/files#r285253695), we can omit the "transaction" prefix from each state string.

source/transactions/tests/README.rst
jmikola: Just realized this field is documented as a map and can support multiple assertions at a time. Since each operation can only be associated with a single session, I don't see a need for that. Moreover, I would have thought there no need to specify the session again since we could infer it from the operation's `arguments.session` field.

That said, I think there's a better approach beneath our noses here. @ShaneHarvey recently began adding [Special Test Operations](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst#special-test-operations) to the transaction spec tests. I think it'd be preferable to introduce a new special operation to allow assertion of a session state. Splitting this assertion into its own operation also avoids any potential misunderstanding about whether the state should be inspected before or after the driver operation is executed -- this means there'd be no need to add more instruction to the already lengthy list in [Use as Integration Tests](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst#use-as-integration-tests). 

Consider:

```
- name: assertSessionTransactionState
  object: testRunner
  arguments:
    session: session0
    state: in_progress
```

This is fairly self-documenting, and the fact that it'd a top-level operation means it'd be much easier for drivers to detect this unrecognized operation type (vs. the possibility of silently ignoring a new field in the `operations` document).

I went without the map syntax, as I still don't see a need for asserting multiple states at once; however, a test could easily perform multiple assertions by sequencing two of these operations. An additional benefit is that we could also assert the session's state before any operation is run (probably not needed for this PR, but potentially useful). Thoughts?

source/transactions/tests/README.rst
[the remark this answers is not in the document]
ShaneHarvey: I agree with Jeremy here although I think it would be more readable to add his proposed test in `errors-client.yml` alongside the other tests rather than in a new file.

source/transactions/tests/README.rst
ShaneHarvey: Agreed, I like the `assertSessionTransactionState` idea!

For inspecting a ClientSession's state, I think some drivers may just be required to add public API like `session.getState()`. Other drivers, like python and maybe Java, may be able to keep it private.

p-mongo:
Changes made per comments:

1. Removed `clientError` and replaced it with `error: true`. Added a note to the error description to state that it could be server or driver generated.

2. Changed to `assertSessionTransactionState` as proposed.

Open questions:

1. Is the `errors-client.yml` file name the still desired one?
2. Should I add the `skipReason` as mentioned earlier?

/cc @jmikola @ShaneHarvey

source/transactions/tests/README.rst
```diff
@@ -342,6 +342,20 @@ fail point on the mongos server which "session0" is pinned to::
               failCommands: ["commitTransaction"]
               closeConnection: true
 
+assertSessionTransactionState
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The "assertSessionTransactionState" operation instructs the test runner to
+assert that the transaction state of the given session is equal to the
+specified value. The possible values are as follows: ``no``, ``starting``,
+``in_progress``, ``committed``, ``aborted``.
```
jmikola: I realize "no" is the result of removing "transaction" from the list of states in the spec, but I think "none" would read better in this instance (as used in the [state diagram](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#clientsession-changes)).

source/transactions/transactions.rst
```diff
@@ -350,6 +350,13 @@ Note that "error" is not a state, it represents throwing an error due to
 an invalid operation. When such errors are thrown the session state is
 unchanged.
 
+Client side errors MUST NOT change transaction state. For example, if an
```
jmikola: "Client-side" is consistent with other usages.

source/transactions/transactions.rst
```diff
@@ -1373,6 +1380,7 @@ durable, which achieves the primary objective of avoiding duplicate commits.
 **Changelog**
 -------------
 
+:2019-06-10: Clarified client side error handling.
```
jmikola: Alternative suggestion: "Client side errors must not change transaction state". Not much longer, and it captures the entire change.

p-mongo:
Made the suggested changes, thank you @jmikola .

PR #549, opened 2019-05-21
Update transaction-related note in sessions spec 

p-mongo:
Since transactions have been implemented, language referring to transactions in the future tense can be updated.

### DIFF
--- a/source/sessions/driver-sessions.rst
+++ b/source/sessions/driver-sessions.rst
@@ -160,18 +160,8 @@ Applications end a session like this:
 
     session.endSession()
 
-While this specification does not deal with multi-document transactions (which
-don't even exist yet), it is expected that when they are implemented they will
-be based upon sessions. We can speculate that in the future we might have some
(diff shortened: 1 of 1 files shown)


source/sessions/driver-sessions.rst
```diff
@@ -160,18 +160,8 @@ Applications end a session like this:
...
-additional transaction-related methods for sessions such as:
-
-.. code:: typescript
-
-    transaction = session.beginTransaction()
-    transaction.commit()
-    transaction.abort()
-
-However, multi-document transactions are out of scope for this specification.
+While this specification does not deal with multi-document transactions, which
```
jmikola: I think we can drop "While", as the sentence doesn't read very well with it in place (both parts are clauses).

I might also suggest linking "specification" below to `../transactions/transactions.rst`.

p-mongo: Made both changes, thank you.

PR #551, opened 2019-05-28
Document databaseOptions field in spec tests

jmikola:


ShaneHarvey:
For others `databaseOptions` is already used in this test: https://github.com/mongodb/specifications/blob/master/source/transactions/tests/read-concern.yml#L585

PR #552, opened 2019-05-28
Don't assert results for basic CRUD ops with w:0

jmikola:
Per the CRUD spec, results fields are inaccessible for w:0 writes (including inserted count and ids).

Came across this while implementing [PHPLIB-365](https://jira.mongodb.org/browse/PHPLIB-365). See [Results](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#results) in the CRUD specification for additional context.

While we could replace these `result` field with a single `isAcknowledged: false` field, I think that would be even less portable for drivers that chose not to implement that API. My understanding from [Test Format](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst#test-format) is that `result` is optional (due to "if any") language, so I think the most correct thing to do here is to avoid any assertion at all.

### DIFF
--- a/source/transactions/tests/write-concern.yml
+++ b/source/transactions/tests/write-concern.yml
@@ -244,8 +244,7 @@ tests:
           session: session0
           document:
             _id: 1
-        result:
-          insertedId: 1
+        # avoid assertion since result fields are inaccessible for unacknowledged write concern
       - *commitTransaction
 
     expectations:
@@ -269,8 +268,7 @@ tests:
           documents:
             - _id: 1
             - _id: 2
-        result:
-          insertedIds: {0: 1, 1: 2}
+        # avoid assertion since result fields are inaccessible for unacknowledged write concern
       - *commitTransaction
 
     expectations:
@@ -305,14 +303,7 @@ tests:
             - name: insertOne
               arguments:
                 document: {_id: 1}
-        result:
-          deletedCount: 0
-          insertedCount: 1
-          insertedIds: {0: 1}
-          matchedCount: 0
-          modifiedCount: 0
-          upsertedCount: 0
-          upsertedIds: {}
+        # avoid assertion since result fields are inaccessible for unacknowledged write concern
       - *commitTransaction
 
     expectations:
@@ -336,8 +327,7 @@ tests:
           session: session0
           filter:
             _id: 0
-        result:
-          deletedCount: 1
+        # avoid assertion since result fields are inaccessible for unacknowledged write concern
       - *commitTransaction
 
     expectations:
@@ -367,8 +357,7 @@ tests:
           session: session0
           filter:
             _id: 0
-        result:
-          deletedCount: 1
+        # avoid assertion since result fields are inaccessible for unacknowledged write concern
       - *commitTransaction
 
     expectations:
@@ -400,10 +389,7 @@ tests:
           update:
             $inc: {x: 1}
           upsert: true
-        result:
-          matchedCount: 1
-          modifiedCount: 1
-          upsertedCount: 0
+        # avoid assertion since result fields are inaccessible for unacknowledged write concern
       - *commitTransaction
 
     expectations:
@@ -438,10 +424,7 @@ tests:
           update:
             $inc: {x: 1}
           upsert: true
-        result:
-          matchedCount: 1
-          modifiedCount: 1
-          upsertedCount: 0
+        # avoid assertion since result fields are inaccessible for unacknowledged write concern
       - *commitTransaction
 
     expectations:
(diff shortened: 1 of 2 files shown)


source/transactions/tests/write-concern.yml
```diff
@@ -244,8 +244,7 @@ tests:
           session: session0
           document:
             _id: 1
-        result:
-          insertedId: 1
+        # avoid assertion since result fields are inaccessible for unacknowledged write concern
```
ShaneHarvey: This command is part of a transaction so the driver ignores the w:0 write concern. A result *is* returned.

source/transactions/tests/write-concern.yml
```diff
@@ -269,8 +268,7 @@ tests:
           documents:
             - _id: 1
             - _id: 2
-        result:
-          insertedIds: {0: 1, 1: 2}
+        # avoid assertion since result fields are inaccessible for unacknowledged write concern
```
ShaneHarvey: This command is part of a transaction so the driver ignores the w:0 write concern. A result *is* returned.

source/transactions/tests/write-concern.yml
```diff
@@ -336,8 +327,7 @@ tests:
           session: session0
           filter:
             _id: 0
-        result:
-          deletedCount: 1
+        # avoid assertion since result fields are inaccessible for unacknowledged write concern
```
ShaneHarvey: This command is part of a transaction so the driver ignores the w:0 write concern. A result *is* returned.

source/transactions/tests/write-concern.yml
```diff
@@ -305,14 +303,7 @@ tests:
...
                 document: {_id: 1}
-        result:
-          deletedCount: 0
-          insertedCount: 1
-          insertedIds: {0: 1}
-          matchedCount: 0
-          modifiedCount: 0
-          upsertedCount: 0
-          upsertedIds: {}
+        # avoid assertion since result fields are inaccessible for unacknowledged write concern
```
ShaneHarvey: This command is part of a transaction so the driver ignores the w:0 write concern. A result *is* returned.

source/transactions/tests/write-concern.yml
```diff
@@ -367,8 +357,7 @@ tests:
           session: session0
           filter:
             _id: 0
-        result:
-          deletedCount: 1
+        # avoid assertion since result fields are inaccessible for unacknowledged write concern
```
ShaneHarvey: This command is part of a transaction so the driver ignores the w:0 write concern. A result *is* returned.

source/transactions/tests/write-concern.yml
```diff
@@ -400,10 +389,7 @@ tests:
           update:
             $inc: {x: 1}
           upsert: true
-        result:
-          matchedCount: 1
-          modifiedCount: 1
-          upsertedCount: 0
+        # avoid assertion since result fields are inaccessible for unacknowledged write concern
```
ShaneHarvey: This command is part of a transaction so the driver ignores the w:0 write concern. A result *is* returned.

source/transactions/tests/write-concern.yml
```diff
@@ -438,10 +424,7 @@ tests:
...
```
ShaneHarvey: This command is part of a transaction so the driver ignores the w:0 write concern. A result *is* returned.

source/transactions/tests/write-concern.yml
```diff
@@ -473,6 +456,7 @@ tests:
         arguments:
           session: session0
           filter: {_id: 0}
+        # findAndModify still returns result for unacknowledged write concern
```
ShaneHarvey: This isn't an unacknowledged write, no writeConcern is sent.

source/transactions/tests/write-concern.yml
```diff
@@ -502,6 +486,7 @@ tests:
           filter: {_id: 0}
           replacement: {x: 1}
           returnDocument: Before
+        # findAndModify still returns result for unacknowledged write concern
```
ShaneHarvey: This isn't an unacknowledged write, no writeConcern is sent.

source/transactions/tests/write-concern.yml
```diff
@@ -534,6 +519,7 @@ tests:
           update:
             $inc: {x: 1}
           returnDocument: Before
+        # findAndModify still returns result for unacknowledged write concern
```
ShaneHarvey: This isn't an unacknowledged write, no writeConcern is sent.

jmikola:
> As far as I can tell this change isn't needed.

Ah, that makes sense. Sorry for wasting your time on this, but thanks for the sanity check.

PR #553, opened 2019-05-29
SPEC-1302: Rename InterruptedDueToStepDown error code to InterruptedDueToReplStateChange

jmikola:
https://jira.mongodb.org/browse/SPEC-1302

See [DRIVERS-657](https://jira.mongodb.org/browse/DRIVERS-657) for additional context.

### DIFF
--- a/source/retryable-writes/retryable-writes.rst
+++ b/source/retryable-writes/retryable-writes.rst
@@ -3,14 +3,14 @@ Retryable Writes
 ================
 
 :Spec Title: Retryable Writes
-:Spec Version: 1.2
+:Spec Version: 1.4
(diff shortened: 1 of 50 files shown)


source/retryable-writes/retryable-writes.rst
```diff
@@ -3,14 +3,14 @@ Retryable Writes
 ================
 
 :Spec Title: Retryable Writes
-:Spec Version: 1.2
+:Spec Version: 1.4
```
ShaneHarvey: Should this be 1.3?

jmikola: Good catch!

PR #556, opened 2019-06-02
getNonce => getnonce

kevinAlbs:
Command names are case sensitive.
```
> db.runCommand({getnonce: 1})
{ "nonce" : "810478340848c1fe", "ok" : 1 }

> db.runCommand({getNonce: 1})
{
	"ok" : 0,
	"errmsg" : "no such command: 'getNonce'",
	"code" : 59,
	"codeName" : "CommandNotFound"
}
```

### DIFF
--- a/source/auth/auth.rst
+++ b/source/auth/auth.rst
@@ -287,13 +287,13 @@ MONGODB-CR
 :deprecated: 3.0
 :removed: 4.0
 
-MongoDB Challenge Response is a nonce and MD5 based system. The driver sends a `getNonce` command, encodes and hashes the password using the returned nonce, and then sends an `authenticate` command.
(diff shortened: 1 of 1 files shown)


Joannis:
Is that a more recent change? I still have `getNonce` in my implementation, not `getnonce`

jmikola:
@Joannis: The earliest evidence of `getnonce` I could dig up is [SERVER-469](https://jira.mongodb.org/browse/SERVER-469). I can't say whether older server versions might have allowed `getNonce`, but I can confirm that MongoDB 3.4.18 is case-sensitive. This appears to just be an outstanding typo that was missed when creating the spec.

Joannis:
Yeah, I did manage to use `getNonce` with older server versions, but I'm not sure which again. I'll make sure I adapt my driver to this.

PR #557, opened 2019-06-03
SPEC-1292: Add rule to ignore read preference for standalone servers

jyemin:
https://jira.mongodb.org/browse/SPEC-1292

I'm open to adding a prose test for this, but currently there is nothing covering the existing behavior.

### DIFF
--- a/source/server-selection/server-selection.rst
+++ b/source/server-selection/server-selection.rst
@@ -885,7 +885,7 @@ to the server differently:
 - Type Mongos: the read preference is sent to the server using the rules
   for `Passing read preference to mongos`_.
 
-- Type Standalone: the read preference is ignored
+- Type Standalone: clients MUST NOT send the read preference to the server
 
 - For all other types, using OP_QUERY: clients MUST always set the ``slaveOK`` wire
   protocol flag on reads to ensure that any server type can handle the
@@ -1686,7 +1686,7 @@ References
 Changes
 =======
 
-2019-05-20: Added rule to ignore read preference for standalone servers
+2019-05-20: Added rule to not send read preferene to standalone servers
 
 2017-11-10: Added application-configurated server selector.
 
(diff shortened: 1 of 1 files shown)


source/server-selection/server-selection.rst
```diff
@@ -885,6 +885,8 @@ to the server differently:
 - Type Mongos: the read preference is sent to the server using the rules
   for `Passing read preference to mongos`_.
 
+- Type Standalone: the read preference is ignored
```
xdg: Each time I've considered this spec change, I get worried about SlaveOK, and then remember that we're talking about Standalone and not Single.  But it makes me wonder if we should add something for clarity like: `(note: this does not apply to Topology Type Single for other server types than Standalone)`

jmikola: That note seems like a reasonable way to preemptively avoid a misunderstanding.

I would also suggest rewording "the read preference is ignored" as "the read preference is never sent to the server" or "drivers MUST NOT send the read preference to the server" if we want to be prescriptive.

In the context of this section ("the read preference is communicated to the server"), "is ignored" makes sense but a rewording would make this clearer both here and in the changelog.

source/server-selection/server-selection.rst
```diff
@@ -1684,6 +1686,8 @@ References
 Changes
 =======
 
+2019-05-20: Added rule to ignore read preference for standalone servers
```
jmikola: "to never send read preference for standalone servers". Optionally: "standalone server types" to avoid any ambiguity about topology/server types.

source/server-selection/server-selection.rst
jyemin: @xdg, if you look at this change in context I think the meaning is clear, as it appears in a section titled `Topology type: Single", and there are bullets that cover all other server types.  In that context, the suggested note sounds a bit weird.  

@jmikola, I agree with your suggestion and will make it in the next patch, but I'll wait to hear back from @xdg first.

xdg: Looking with more diff context, I agree that my additional language isn't needed, but I also agree with Jeremy about avoiding the term "ignored".

jyemin: Done

source/server-selection/server-selection.rst
jyemin: Done

PR #558, opened 2019-06-03
update changeStream test to be extended json

daprahamian:
replacing raw `{ "batchSize": 1 }` with `{ "batchSize": { "$numberInt": "1" } }` so the expectations conform to extended json.

### DIFF
--- a/source/change-streams/tests/change-streams.json
+++ b/source/change-streams/tests/change-streams.json
@@ -737,7 +737,9 @@
             "command": {
               "aggregate": "test",
               "cursor": {
-                "batchSize": 1
+                "batchSize": {
+                  "$numberInt": "1"
+                }
               },
               "pipeline": [
                 {
--- a/source/change-streams/tests/change-streams.yml
+++ b/source/change-streams/tests/change-streams.yml
@@ -499,7 +499,9 @@ tests:
         command_started_event:
           command:
             aggregate: *collection_name
-            cursor: {batchSize: 1}
+            cursor:
+              batchSize:
+                $numberInt: "1"
             pipeline:
               -
                 $changeStream:


jyemin:
`1` is valid extended json (conforms to relaxed mode). What was the impetus for this? Is it breaking some driver expectations?

daprahamian:
The node test runner expected the EJSON to be canonical, so this check broke the assertion. I can fix it on node's end by converting all the results to canonical, but it seems weird to mix the relaxed `1` in `batchSize` with the `{"$numberInt": "1"}` in `fullDocument.x`.

jyemin:
Adding Matt. We talked about this in the context of the FLE tests and decided not to change those.  What do you think we need to do @mbroadst?

source/change-streams/tests/change-streams.json
```diff
@@ -737,7 +737,9 @@
             "command": {
               "aggregate": "test",
               "cursor": {
-                "batchSize": 1
+                "batchSize": {
+                  "$numberInt": "1"
+                }
```
jmikola: Since `changeStreamOptions` for this test is defined as `{ batchSize: 1 }`, I think this command assertion may be making assumptions about how drivers serialize integers to BSON.

Let's assume for a moment that the `1` earlier in the JSON test could be parsed as either a 32-bit integer, 64-bit integer, or generic numeric type. This makes a few assumptions:

* If `1` is parsed as a generic numeric type, it will serialize as a BSON integer rather than a double because it is a whole number. I think this is a safe assumption to make, but I want to note it for the sake of completeness.
* When serializing an integer (or whole number), a 32-bit BSON integer type will be used if the number falls within the range of 32-bit signed integers. I think this may be an unsafe assumption, as the [BSON corpus spec](https://github.com/mongodb/specifications/tree/master/source/bson-corpus) doesn't dictate this behavior and could allow a driver to parse `1` as a 64-bit integer.
* This doesn't consider that a driver's change stream, aggregation, or cursor code might coerce a type for `batchSize` long after the JSON spec test is passed. A driver could theoretically always use a 64-bit integer. I'm not aware of any driver that does this (and we certainly have no cursor spec), but I'll point out that most of the existing `batchSize` assertions in the [APM spec tests](https://github.com/mongodb/specifications/tree/master/source/command-monitoring/tests) appear to use `$numberLong` for both the option argument and expected command document.

If we look to the APM tests as guidance, I think the best solution here would be to use the same extended JSON type in the option argument itself. And in that case I would suggest we keep `$numberLong` to remain consistent with the APM spec tests, which are presumably not a problem for any driver.

mbroadst:
We're now reading in these tests in relaxed EJSON, so maybe we can close this now? I do kind of like the idea that our tests are written in canonical for strictness though.

PR #561, opened 2019-06-06
SPEC-1298 Add "connectionError" as a valid reason for ConnectionCheckOutFailedEvent

ShaneHarvey:
Is there any way to add a spec test for this? If not I can add a prose test.

### DIFF
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -572,7 +572,7 @@ Events
        * Current valid values are:
        *   - "stale":           The pool was cleared, making the connection no longer valid
        *   - "idle":            The connection became stale by being available for too long
-       *   - "connectionError": The connection experienced an error, making it no longer valid
+       *   - "error":           The connection experienced an error, making it no longer valid
        *   - "poolClosed":      The pool was closed, making the connection no longer valid
        */
       reason: string|Enum;
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -572,7 +572,7 @@ Events
        * Current valid values are:
        *   - "stale":           The pool was cleared, making the connection no longer valid
        *   - "idle":            The connection became stale by being available for too long
-       *   - "error":           The connection experienced an error, making it no longer valid
+       *   - "connectionError": The connection experienced an error, making it no longer valid
        *   - "poolClosed":      The pool was closed, making the connection no longer valid
        */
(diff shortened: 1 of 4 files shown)


daprahamian:
As the test runner is currently set up, it is impossible to make a spec test for it. Right now, there is no way to make a connection fail to set up, b/c the spec test runner does not involve any I/O.

p-mongo:
I think it might be confusing to have a reason of "error" when some errors have their own reasons. But I am having a hard time proposing alternate wording ("setup error"?)

ShaneHarvey:
> I think it might be confusing to have a reason of "error" when some errors have their own reasons. But I am having a hard time proposing alternate wording ("setup error"?)

I agree it might be confusing. It l looks like the same problem is present with ConnectionClosedEvents. @daprahamian do you have any suggestions? For now I've kept it as "error".

daprahamian:
Maybe `"connectionError"`? Indicating that it is an error *from* the connection that caused the checkout to fail.

source/connection-monitoring-and-pooling/tests/pool-checkout-error-closed.yml
```diff
@@ -15,14 +15,22 @@ events:
...
   - type: ConnectionCheckedIn
+    address: 42
     connectionId: 42
   - type: ConnectionPoolClosed
     address: 42
+  - type: ConnectionCheckOutStarted
+    address: 42
+  - type: ConnectionCheckOutFailed
+    address: 42
+    reason: poolClosed
```
ShaneHarvey: This behavior is outlined in the spec (`ConnectionCheckOutFailed(reason="poolClosed"`)  but was not previously tested.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -568,7 +572,7 @@ Events
        * Current valid values are:
        *   - "stale":           The pool was cleared, making the connection no longer valid
        *   - "idle":            The connection became stale by being available for too long
-       *   - "error":           The connection experienced an error, making it no longer valid
+       *   - "connectionError": The connection experienced an error, making it no longer valid
```
ShaneHarvey: I renamed this constant as well to keep it consistent with `ConnectionCheckOutFailedEvent`.

daprahamian: I'd prefer to keep it as `"error"` b/c:

1. that would require more changes to existing implementations, and might wind up breaking some current impls
2. Since it is part of a `ConnectionClosed` event, it is a little redundant. All errors at the `Connection` level are a `connectionError`.

ShaneHarvey: Makes sense. Done.

PR #562, opened 2019-06-07
spec-1313 clarify username requirements for each auth mechanism

haris-sheikh:
https://jira.mongodb.org/browse/SPEC-1313

### DIFF
--- a/source/auth/auth.rst
+++ b/source/auth/auth.rst
@@ -87,10 +87,10 @@ mechanism (string)
 mechanism_properties
 	* Includes additional properties for the given mechanism.
 
-Specifying userinfo
-~~~~~~~~~~~~~~~~~~~
+Credential delimiter in URI implies authentication
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Providing a credential delimiter (i.e. @) means that the user has unambiguously specified userinfo, and will thus be treated as an authentication request.
+The presence of a credential delimiter (i.e. @) in the URI connection string is evidence that the user has unambiguously specified user information and MUST be interpreted as a user configuring authentication credentials (even if the username and/or password are empty strings).
 
 Ambiguity between authentication source and URI database
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/source/auth/auth.rst
+++ b/source/auth/auth.rst
@@ -87,6 +87,11 @@ mechanism (string)
 mechanism_properties
 	* Includes additional properties for the given mechanism.
 
+Specifying userinfo
+~~~~~~~~~~~~~~~~~~~
+
+Providing a credential delimiter (i.e. @) means that the user has unambiguously specified userinfo, and will thus be treated as an authentication request.
+
 Ambiguity between authentication source and URI database
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/source/auth/auth.rst
+++ b/source/auth/auth.rst
@@ -88,7 +88,7 @@ mechanism_properties
 	* Includes additional properties for the given mechanism.
 
 Credential delimiter in URI implies authentication
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The presence of a credential delimiter (i.e. @) in the URI connection string is evidence that the user has unambiguously specified user information and MUST be interpreted as a user configuring authentication credentials (even if the username and/or password are empty strings).
 
--- a/source/auth/auth.rst
+++ b/source/auth/auth.rst
@@ -312,7 +312,7 @@ As an example, given a username of "user" and a password of "pencil", the conver
 `````````````````````````````
 
 username
-	MUST be specified.
+	MUST be specified and non-zero length.
 
 source
 	MUST be specified. Defaults to the database name if supplied on the connection string or ``admin``.
@@ -365,7 +365,7 @@ Conversation
 
 username
 	SHOULD NOT be provided for MongoDB 3.4+
-	MUST be specified for MongoDB prior to 3.4
+	MUST be specified and non-zero length for MongoDB prior to 3.4
 
 source
 	MUST be "$external". Defaults to ``$external``.
@@ -418,7 +418,7 @@ GSSAPI is kerberos authentication as defined in `RFC 4752 <http://tools.ietf.org
 `MongoCredential`_ properties:
 
 username
-	MUST be specified.
+	MUST be specified and non-zero length.
 
 source
 	MUST be "$external". Defaults to ``$external``.
@@ -523,7 +523,7 @@ MongoDB supports either of these forms.
 `````````````````````````````
 
 username
-	MUST be specified.
+	MUST be specified and non-zero length.
 
 source
 	MUST be specified. Defaults to the database name if supplied on the connection string or ``$external``.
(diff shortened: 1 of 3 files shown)


xdg:
Hi, Haris.  There are a few reasons I don't think this is a good change:
* An empty username does not interfere with URI parsing; such an error should be caught during credential construction based on the authentication mechanism (as mandated by the auth spec).  Some authentication mechanisms do not require username or password.
* Even if we decided to prohibit it in URI parsing, the actual change you made is part of the paragraph on userinfo which refers to the entire "username:password" portion of the URI, so it could easily be misinterpreted to mean that an empty userinfo in invalid rather than an empty username is invalid.

p-mongo:
A prohibition on empty user name would make our URIs diverge (further) from the standard URIs. My interpretation of https://tools.ietf.org/html/rfc3986#section-3.2.1 and https://tools.ietf.org/html/rfc3986#section-3.2 is that user information can be empty, i.e. `http://@foo:123` is a valid URI.

Specifically in Ruby:

<pre>
irb(main):002:0> u=URI.parse('https://:test@bar.com/')
=> #<URI::HTTPS https://:test@bar.com/>
irb(main):004:0> u.user
=> ""
irb(main):005:0> u.password
=> "test"

irb(main):006:0> u=URI.parse('https://@bar.com/')
=> #<URI::HTTPS https://bar.com/>
irb(main):007:0> u.user
=> nil
irb(main):008:0> u.password
=> nil

irb(main):009:0> u=URI.parse('https://:@bar.com/')
=> #<URI::HTTPS https://:@bar.com/>
irb(main):010:0> u.user
=> ""
irb(main):011:0> u.password
=> ""
</pre>

jmikola:
@p-mongo: Thanks for raising the point about URI spec compliance. 

@xdg: I agree that this may be better addressed by an error in the auth component, or perhaps an amendment to the [Auth spec](https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst).

The original issue that lead to this proposed change is [CDRIVER-3012](https://jira.mongodb.org/browse/CDRIVER-3012), and the root problem there was an `@` character in a URI, which otherwise had no username or password info, was causing libmongoc to attempt (and predictably fail) authentication. An alternative change might be to just not infer that authentication should take place if the only indicator is `@` in a URI and there is neither a username nor password available (and no other authMechanism or authSource specified in the URI).

xdg:
@jmikola maybe the auth spec should be amended such that "username MUST be specified" means non-zero length?

jmikola:
> maybe the auth spec should be amended such that "username MUST be specified" means non-zero length?

That seems like a good chance. @behackett's testing in [SPEC-1313](https://jira.mongodb.org/browse/SPEC-1313?focusedCommentId=2276511&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2276511) demonstrates that the server rejects creation of a user with an empty string for a username.

xdg:
I like the changes to the auth spec.  Could you please also add some test cases to the [test file|https://github.com/mongodb/specifications/blob/master/source/auth/tests/connection-string.yml]?  Particularly, things like:
* `mongodb://@localhost.com/`
* `mongodb://:@localhost.com/`

jmikola:
@xdg: To clarify, should the two test cases above infer an auth mechanism based on presence of the credential delimiter (`@`)? The spec talks about not inferring an auth mechanism based on [`/database`](https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#ambiguity-between-authentication-source-and-uri-database), but I think it may be ambiguous about doing so due to `@` before the seed list.

xdg:
I believe that having a credential delimiter means the user has unambiguously (if oddly) specified userinfo and that should be treated as an authentication request. `mongodb://:@localhost.com/` should parse successfully and error when constructing a Credential object.

jmikola:
> I believe that having a credential delimiter means the user has unambiguously (if oddly) specified userinfo and that should be treated as an authentication request.

@haris-sheikh: I might suggest adding a new section to clarify this above or below the existing https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#ambiguity-between-authentication-source-and-uri-database section.

source/auth/tests/connection-string.yml
```diff
@@ -291,3 +291,11 @@ tests:
         description: "authSource without username is invalid (default mechanism)"
         uri: "mongodb://localhost/?authSource=foo"
         valid: false
+
+        description: "should throw an exception if not username provided"
```
jmikola: "should throw an exception if no username (userinfo implies default mechanism)"

source/auth/tests/connection-string.yml
```diff
@@ -291,3 +291,11 @@ tests:
...
+        uri: "mongodb://@localhost.com/"
+        valid: false
+
+        description: "should throw an exception if not username/password provided"
```
jmikola: "should throw an exception if no username/password (userinfo implies default mechanism)"

source/auth/tests/connection-string.yml
p-mongo: Suggest "if no username provided" or "if no username is provided" or "if username is not provided".

source/auth/auth.rst
```diff
@@ -87,6 +87,11 @@ mechanism (string)
 mechanism_properties
 	* Includes additional properties for the given mechanism.
 
+Specifying userinfo
```
jmikola: I wonder if this reads more clearly:

> *User information in URI implies authentication*
> 
> The presence of a credential delimiter (i.e. `@`) in the URI connection string is evidence that the user has unambiguously specified user information and MUST be interpreted as a user configuring authentication credentials (even if the username and/or password are empty strings).

This mostly keeps the first half of your original text but modifies the second half to borrow language from the subsequent section ("Ambiguity between authentication source and URI database"). Since we're talking about requiring driver behavior, it's helpful to use actionable terms (e.g. MUST, MUST NOT).

Alternative title: "Credential delimiter in URI implies authentication"

haris-sheikh: Yes that makes sense to me!

source/auth/auth.rst
```diff
@@ -87,10 +87,10 @@ mechanism (string)
 mechanism_properties
 	* Includes additional properties for the given mechanism.
 
-Specifying userinfo
-~~~~~~~~~~~~~~~~~~~
+Credential delimiter in URI implies authentication
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
jmikola: This should match the length of the title it underlines.

source/auth/tests/connection-string.yml
```diff
@@ -291,3 +291,11 @@ tests:
         description: "authSource without username is invalid (default mechanism)"
         uri: "mongodb://localhost/?authSource=foo"
         valid: false
+
+        description: "should throw an exception if no username provided (userinfo implies default mechanism)"
+        uri: "mongodb://@localhost.com/"
+        valid: false
+
```
prashantmital: @haris-sheikh you missed a `-` here which means running `make` in the top-level `source/` directory no longer works: 
```
$ make
tmpfile=$(mktemp);               \
	if js-yaml auth/tests/connection-string.yml > "$tmpfile"; then \
	    mv "$tmpfile" auth/tests/connection-string.json;            \
	else                              \
	    rm "$tmpfile";               \
	    exit 1;                       \
	fi
YAMLException: duplicated mapping key at line 295, column 9:
            description: "should throw an ex ...
            ^
make: *** [auth/tests/connection-string.json] Error 1
```

jmikola: @haris-sheikh: In the future, each PR with YAML changes should regenerate JSON files (see: https://github.com/mongodb/specifications#converting-to-json). That would have caught this mistake before merging, so please keep in mind for any future PRs.

Apart from that, all of the reviewers (myself included) missed that this PR was changing YAML without including a corresponding JSON change.

haris-sheikh: Oh okay, sorry for that. Will keep this in mind for any future PRs!

PR #563, opened 2019-06-07
SPEC-1289: Spec and test updates for $merge aggregation stage

jmikola:
https://jira.mongodb.org/browse/SPEC-1289

This revises language across several specs pertaining to aggregate commands that write. Some updates were also made for mapReduce while doing so.

New tests for $merge have been added alongside existing tests for $out. These tests are split off to denote the appropriate minServerVersion requirement.


source/crud/crud.rst
```diff
@@ -132,10 +132,11 @@ Read
...
      *
-     * Note: $out is a special pipeline stage that causes no results to be returned
-     * from the server. As such, the iterable here would never contain documents. Drivers
-     * MAY setup a cursor to be executed upon iteration against the $out collection such
-     * that if a user were to iterate a pipeline including $out, results would be returned.
+     * Note: $out and $merge are special pipeline stages that cause no results
+     * to be returned from the server. As such, the iterable here would never
+     * contain documents. Drivers MAY setup a cursor to be executed upon
+     * iteration against the output collection such that if a user were to
+     * iterate the return value, results would be returned.
```
jmikola: In [SPEC-1289](https://jira.mongodb.org/browse/SPEC-1289), I previously stated an intention to prohibit returning a cursor for `$merge` based on discussion in server scope documents. I later realized that could cause problems for strictly typed drivers that must return an iterable and decided against it.

source/crud/crud.rst
```diff
@@ -236,17 +237,17 @@ Read
      * If specified, drivers SHOULD apply this option to both the original aggregate command and subsequent
      * getMore operations on the cursor.
      *
-     * Drivers MUST NOT specify a batchSize of zero in an aggregate command that includes an $out stage, as
-     * that will prevent the pipeline from executing. Drivers SHOULD leave the cursor.batchSize command option
-     * unset in an aggregate command that includes an $out stage.
+     * Drivers MUST NOT specify a batchSize of zero in an aggregate command that includes an $out or $merge stage,
+     * as that will prevent the pipeline from executing. Drivers SHOULD leave the cursor.batchSize command option
+     * unset in an aggregate command that includes an $out or $merge stage.
```
jmikola: I was originally mistaken in thinking that [SERVER-23923](https://jira.mongodb.org/browse/SERVER-23923) would ensure that the pipeline is executed even if `batchSize` is zero; however, I realized that issue is only ensuring that the `aggregate` command's `writeConcern` is still applied on the `getMore` when iterating such a cursor. Therefore, it's not very relevant to drivers at all given our current design, and this language from [SPEC-826](https://jira.mongodb.org/browse/SPEC-826) must remain in place for 4.2+.

Drivers still need to prohibit sending `batchSize: 0` if they want to guarantee that a pipline with a write stage is executed after a single call to `aggregate()` with no iteration on the return value.

source/retryable-reads/tests/aggregate-merge.yml
```diff
@@ -0,0 +1,39 @@
+runOn:
+  -
+    minServerVersion: "4.1.12"
```
jmikola: FWIW, neither `$out` nor `$merge` can create a new output collection on a sharded cluster. That said, these tests expect an error and the fail point aborts the `aggregate` command so I don't think it's a concern.

source/crud/tests/v2/aggregate-merge.yml
```diff
@@ -0,0 +1,136 @@
+data:
+    - { _id: 1, x: 11 }
+    - { _id: 2, x: 22 }
+    - { _id: 3, x: 33 }
+
+minServerVersion: '4.1.12'
```
jmikola: I'm not sure exactly when `$merge` was introduced, but many of the SERVER tickets in its epic have this fixVersion so this seems safe to use. We're also just testing basic usage of `$merge`, so it's not a problem if its more complex options land in a later version.

source/crud/tests/v2/aggregate-merge.yml
```diff
@@ -0,0 +1,136 @@
...
+            aggregate: *collection_name
+            pipeline: *pipeline
+    outcome: &outcome
+      collection:
+        name: *output_collection
+        data:
+          - { _id: 2, x: 22 }
+          - { _id: 3, x: 33 }
+  -
+    description: "Aggregate with $merge and batch size of 0"
```
jmikola: This is the equivalent test for `$out` that was introduced in [SPEC-826](https://jira.mongodb.org/browse/SPEC-826).

source/crud/tests/v2/aggregate-merge.yml
```diff
@@ -0,0 +1,136 @@
...
+          <<: *arguments
+          batchSize: 0
+    expectations:
+      -
+        command_started_event:
+          command:
+            aggregate: *collection_name
+            pipeline: *pipeline
+            # TODO: check that cursor.batchSize does not exist once
+            # https://jira.mongodb.org/browse/SPEC-1215 has been resolved.
```
jmikola: It's not clear to me if the rules for comparing APM commands allows us to assert nonexistent fields of embedded documents. See https://github.com/mongodb/specifications/pull/560#pullrequestreview-246810295 for some earlier discussion on this. In any case, I think it's reasonable to wait on [SPEC-1215](https://jira.mongodb.org/browse/SPEC-1215) to sort this out. This comment is derived from some existing comments elsewhere in CRUD spec test files.

source/transactions/transactions.rst
```diff
@@ -1283,8 +1283,8 @@ The following commands are allowed inside transactions:
 
 8.  aggregate (including $lookup)
 
-    -  The $out stage is prohibited because it uses collection create
-       and rename operations.
+    -  The ``$out`` and ``$merge`` stages are prohibited because they use
+       collection create and rename operations.
```
jmikola: Based on a local test with MongoDB 4.1.13, a pipeline with `$merge` is causes `aggregate` to fail with code 263 ("OperationNotSupportedInTransaction") and message "$merge cannot be used in a transaction". This error occurs regardless of whether the collection exists prior to running the aggregation, so I assume the server is just preemptively rejecting the pipeline rather than waiting to see if it would attempt to create a new collection.

source/crud/tests/v2/aggregate-merge.yml
```diff
@@ -0,0 +1,136 @@
+data:
+    - { _id: 1, x: 11 }
+    - { _id: 2, x: 22 }
+    - { _id: 3, x: 33 }
+
+minServerVersion: '4.1.12'
+
+collection_name: &collection_name 'test_aggregate_merge'
+
+tests:
```
jyemin: Can we add tests that assert both $out and $merge are treated as writes, e.g. by specifying a non-default write concern and including the write concern in the command started event?

source/crud/tests/v2/aggregate-merge.yml
```diff
@@ -0,0 +1,136 @@
...
+    expectations:
+      -
+        command_started_event:
+          command:
+            aggregate: *collection_name
+            pipeline: *pipeline
+            readConcern: { level: "available" }
+    outcome: *outcome
+  -
+    description: "Aggregate with $merge and linearizable readConcern"
```
jyemin: Why we need a driver test to assert that the server doesn't support linearlizable read concern?  Does it test anything in the driver? Also, we'll just need to update this when the server starts to support it.

source/crud/tests/v2/aggregate-merge.yml
```diff
@@ -0,0 +1,136 @@
...
+        error: true
+    expectations:
+      -
+        command_started_event:
+          command:
+            aggregate: *collection_name
+            pipeline: *pipeline
+            readConcern: { level: "linearizable" }
+  -
+    description: "Aggregate with $merge and invalid readConcern"
```
jyemin: How strongly do you feel we need this test?  In Java the level is an enumerated value so this isn't even possible and we'd have to skip the test anyway.  More general argument against this: we're just testing the server here, not the driver.

source/crud/tests/v2/aggregate-merge.yml
ShaneHarvey: batchSize is an option specified inside "cursor" so I think you can do:
```
            aggregate: *collection_name
            pipeline: *pipeline
            cursor: {}
```

source/crud/crud.rst
ShaneHarvey: Is there any reason to send batchSize with $out/$merge at all? Can we just always ignore it for $out/$merge?

source/read-write-concern/read-write-concern.rst
```diff
@@ -89,17 +89,20 @@ The read concern option is available for the following operations:
 - ``count`` command
 - ``distinct`` command
 - ``find`` command
-- ``mapReduce`` command with {out : inline} output option
+- ``mapReduce`` command where ``out`` option is ``{ inline: 1 }``
```
ShaneHarvey: where _the_ ``out`` option is...

source/driver-read-preferences.rst
```diff
@@ -272,7 +272,7 @@ preference, but it only obeys the preference for these commands:
 
 * group
 * inline mapreduce
-* aggregate without $out specified
+* aggregate without write stage specified (e.g. $out)
```
ShaneHarvey: Suggestion: `(e.g. ``$out``, ``$merge``)`

source/retryable-reads/tests/aggregate-merge.yml
ShaneHarvey: Your comment suggests that this test will pass with and without the configureFailPoint. Can you could add an insert to create the output collection before running the `$merge`? That would avoid conflating the closeConnection error and the $merge error.

source/server-selection/server-selection.rst
```diff
@@ -711,17 +711,18 @@ the command and how it is invoked:
         - geoNear
         - geoSearch
         - group
-        - mapReduce, mapreduce (with out: {inline: 1})
+        - mapReduce where ``out`` option is ``{ inline: 1 }``
```
ShaneHarvey: where _the_ ``out`` option is...

source/transactions/transactions.rst
ShaneHarvey: I'm not sure if the $merge stage runs create/rename ops. Maybe we should just say they are prohibited and that's it:
```
-  The ``$out`` and ``$merge`` stages are prohibited.
```

source/crud/crud.rst
ShaneHarvey: @jyemin is this what you and Charlie decided?

Alternatively, we could mandate that drivers must return an empty cursor from $merge.

source/crud/crud.rst
jmikola: Per [SPEC-826](https://jira.mongodb.org/browse/SPEC-826), no driver should be setting `cursor.batchSize` on the initial `aggregate` command when the `batchSize` option is zero. That said, the `batchSize` option may still be relevant for drivers that choose to return a cursor on the output collection; however, we currently don't test for that behavior. If you like, I can add language here that suggests drivers _do_ apply the `batchSize` to any cursor on the output collection.

Testing this would probably require a separate test file, since we'd be covering optional behavior that various drivers would need to skip. If there's interest in that, I'd defer it to someone working on a driver that actually implements the behavior so the spec test isn't written blindly.

source/retryable-reads/tests/aggregate-merge.yml
jmikola: I think my previous comment was mistaken. [`$out`](https://docs.mongodb.com/v4.0/reference/operator/aggregation/out/) cannot output to a _sharded_ collection, but it _can_ create a new collection on a sharded cluster. Since there is no reason for the namespaces in these `$out` and `$merge` tests to be configured as a sharded collection, I don't think there's a problem here. Moreover, it looks like [`$merge`](https://docs.mongodb.com/master/reference/operator/aggregation/merge/) actually supports outputting to a sharded collection, so this concern certainly shouldn't apply to _this_ test.

The JSON for this `$merge` test is identical to the corresponding `$out` test in `aggregate.json` apart from the pipeline stage, description, and order of `name` and `object` fields.

source/crud/tests/v2/aggregate-merge.yml
jmikola: Ah, that makes complete sense. There's no reason for any other field to exist within `cursor` (based on the current command API), so it _must_ be an empty document.

source/crud/tests/v2/aggregate-merge.yml
jmikola: If anything, this is testing https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#unknown-levels-and-additional-options-for-string-based-readconcerns but I agree that there is little value to this test. Although Java may restrict the level to an enum value, isn't it possible to still specify a completely unsupported string using the [URI option](https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#via-connection-string)?

In any event, I agree with removing this and the linearizable test from this PR. I've opened [SPEC-1315](https://jira.mongodb.org/browse/SPEC-1315) to note your suggestion about removing the corresponding `$out` tests.

source/crud/tests/v2/aggregate-merge.yml
jmikola: Will remove.

See: https://github.com/mongodb/specifications/pull/563#discussion_r292032943

source/driver-read-preferences.rst
jmikola: I originally didn't want to mention `$merge` in this document, since it's quite old (2014) and deprecated, but there's no harm in this while I'm revising the text to be more future-proof.

source/crud/crud.rst
jmikola: Per conversation in https://github.com/mongodb/specifications/pull/563#discussion_r291723210, let me know if we should add language advising drivers to apply `batchSize` to any cursor created directly on the output cursor (for `$out` and/or `$merge`).

I don't think we need to require drivers to avoid creating such an output cursor for `$merge` if they already do so for `$out`. AFAIK, such cursors are still only created in a pre-query state so a user could simply ignore them and avoid running any query on a potentially much larger output collection.

source/crud/tests/v2/aggregate-merge.yml
jmikola: At present, I don't believe the CRUD spec has a reliable way to specify these options, and it seems odd to add such tests to the transactions spec (it's runner would support this). I've opened [SPEC-1316](https://jira.mongodb.org/browse/SPEC-1316) to track this request.

source/crud/crud.rst
jyemin: @ShaneHarvey we didn't discuss this.  I agree with Jeremy that it would be weird to return an Iterable for $out and null for $merge, and it's best to just return an Iterable in all cases.   Applications are free to ignore this iterable

source/crud/tests/v2/aggregate-merge.yml
jyemin: I would just set this to 4.2.0 since rc0 has been released

source/crud/crud.rst
ShaneHarvey: Thanks for clearing that up. SGTM!

source/crud/crud.rst
ShaneHarvey: I'll open a new spec ticket for this discussion. I don't want to hold up this review or expand its scope.

source/retryable-reads/tests/aggregate-merge.yml
ShaneHarvey: Oh great!

source/crud/crud.rst
jmikola: Sounds like no action required here. Resolving.

source/crud/tests/v2/aggregate-merge.yml
jmikola: Done and pushed. Resolving.

source/crud/crud.rst
jmikola: Feel free to link it to this issue and SPEC-826 for context.

PR #564, opened 2019-06-10
rename .template.yml to .yml.template

kevinAlbs:
The .template.yml files interfered with the Makefile that converted yml to json.


source/client-side-encryption/tests/generate.py
```diff
@@ -25,11 +25,11 @@
 filepath = sys.argv[1]
 filedir = os.path.dirname(filepath)
 (filename, ext) = os.path.splitext(os.path.basename(filepath))
-if ext != ".yml":
-    print("Input file must end with .yml")
+if ext != ".template":
+    print("Input file must end with .template.yml")
```
jmikola: Should this be `.yml.template`?

source/client-side-encryption/tests/local.yml
```diff
@@ -1,56 +0,0 @@
-runOn:
```
kevinAlbs: This slipped in originally, but ended up being renamed and hence no longer used.

source/client-side-encryption/tests/generate.py
```diff
@@ -25,11 +25,11 @@
...
     sys.exit(1)
 (filename, ext) = os.path.splitext(filename)
-if ext != ".template":
+if ext != ".yml":
```
jmikola: I'm not sure how this check differs from the previous check, but is this still accurate?

Is this attempting to validate the output filename (derived from the original template filename)? Perhaps it's just the error message below that's inaccurate.

kevinAlbs: Yeah the error message needed to be updated.

source/client-side-encryption/tests/generate.py
kevinAlbs: Fixed.

source/client-side-encryption/tests/generate.py
jmikola: Ah, I see what the logic is now. You're checking the order of extensions. :+1:

PR #565, opened 2019-06-12
SPEC-1319 Database-level aggregation should be tested on server versions 3.6+

prashantmital:
Closes [SPEC-1319](https://jira.mongodb.org/browse/SPEC-1319)

### DIFF
--- a/source/crud/tests/v2/db-aggregate.json
+++ b/source/crud/tests/v2/db-aggregate.json
@@ -3,7 +3,7 @@
   "minServerVersion": "3.6",
   "tests": [
     {
-      "description": "Aggregate with $currentOp",
+      "description": "Aggregate with $listLocalSessions",
       "operations": [
         {
           "name": "aggregate",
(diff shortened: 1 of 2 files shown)


ShaneHarvey:
Did you test this on bot replica sets and sharded clusters?

jmikola:
Obligatory @amschwerin:

20160923_andy_sharding

prashantmital:
Yeah, so the `localOps` option is needed to ensure these tests aren't flakey when running against a sharded cluster with multiple mongoses. That option was only added to the `$currentOp` pipeline stage in 4.0. @jmikola @ShaneHarvey do you have any ideas about how we can work around this or if it even matters?

prashantmital:
@mbroadst I see that you added the `localOps` option to the test to fix a race condition with the following message:
```
mbroadst committed on Feb 21
When running these CRUD tests against a sharded cluster the
`$currentOp` stage will likely return many results, occasionally
too many to return the initiating aggregtion command we expect
to see in the results. Limiting the stage to return only local
operations corrects the race condition.
```
I don't follow how we could end up having so many operations returned by the aggregation command that we miss the command we are looking for. Since we have a `$match` stage in the pipeline as well, I would assume that this aggregation query returns only 1 record. The only situation I can think of in which this might not be applicable is if the cluster runs aggregation commands for internal bookkeeping purposes.

jmikola:
@mbroadst added `localOps` and bumped the server version requirement to 4.0 _after_ the original tests were already matching on the command type (https://github.com/mongodb/specifications/commit/5cf02a5f11bf15a084dca0524f997b934a00d7d2#diff-bf3de05ff32d6f2d4da69c3394208833R15), so I assume that was done for a reason and the `$match` was not sufficient to filter out irrelevant commands in a sharded cluster.

ShaneHarvey:
What if we just do this? It should work on all clusters:
```
> db.aggregate([{$listLocalSessions:{}}, {$limit: 1}, {$addFields: {dummy: 1}}, {$project: {"_id":0, "dummy":1}}])
{ "dummy" : 1 }
```

prashantmital:
How about keeping the current test as they are, and adding in another file db-aggregate-3.6.yml that looks something like:
```
database_name: &database_name "admin"

minServerVersion: '3.6'

tests:
  -
    description: "Aggregate with $listLocalSessions"
    operations:
      -
        name: aggregate
        object: database
        arguments:
          pipeline:
            - $listLocalSessions: { }
            - $limit: 1
            - $addFields: { "dummy": "dummy field"}
            - $project: { "_id": 0, "dummy": 1}
        result:
          -
            command:
              aggregate: 1
              pipeline:
                - $listLocalSessions: { }
                - $limit: 1
                - $addFields: { "dummy": "dummy field"}
                - $project: { "_id": 0, "dummy": 1}
              cursor: {}
```

jmikola:
That would work for me. I don't think we care much about which stage we use, so long as we run an `aggregate: 1` command.

prashantmital:
I have updated the PR accordingly. Please re-review @jmikola @ShaneHarvey

prashantmital:
Passing patch build that uses the modified tests: https://evergreen.mongodb.com/version/5d017ce857e85a5a7a9baf6e

PR #566, opened 2019-06-13
SPEC-1321 test keyVaultNamespace

kevinAlbs:
Now explicitly says that "admin.datakeys" is the default. And changes the "missingKeys" test to validate that we pass the keyVaultNamespace option through and use it.

I validated on the old Java test runner (on j3071) with a change to add keyVaultNamespace. Though the Java test runner has changed a lot, so I don't have a PR for the newest test runner.


kevinAlbs:
@mbroadst would you like to review this or abstain from reviewing?

PR #567, opened 2019-06-13
SPEC-1322: Fix result assertion in change stream spec test

jmikola:
https://jira.mongodb.org/browse/SPEC-1322

### DIFF
--- a/source/change-streams/tests/change-streams.yml
+++ b/source/change-streams/tests/change-streams.yml
@@ -32,7 +32,17 @@ tests:
           command_name: aggregate
           database_name: *database_name
     result:
-      success: []
+      success:
+        -
+          _id: "42"
+          documentKey: "42"
+          operationType: insert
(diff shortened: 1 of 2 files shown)


source/change-streams/tests/change-streams.yml
```diff
@@ -32,7 +32,17 @@ tests:
...
+        -
+          _id: "42"
+          documentKey: "42"
+          operationType: insert
+          ns:
+            db: *database_name
+            coll: *collection_name
+          fullDocument:
+            x:
+              $numberInt: "1"
```
ShaneHarvey: I think this should be `x:1` instead of `$numberInt: "1"` but it looks like all the other tests already use `$numberInt` so lgtm.

jmikola: I'm inclined to agree, but didn't want to risk anything.

PR #570, opened 2019-06-15
WRITING-3189 add client side encryption

kevinAlbs:


kevinAlbs:
Per discussion, going to merge since formatting looks ok.

PR #572, opened 2019-06-18
feat: introduce `runOn` to the CRUDv2 tests

mbroadst:
SPEC-1255

### DIFF
--- a/source/crud/tests/README.rst
+++ b/source/crud/tests/README.rst
@@ -14,7 +14,7 @@ that drivers can use to prove their conformance to the CRUD spec.
 
 Running these integration tests will require a running MongoDB server or
 cluster with server versions 2.6.0 or later. Some tests have specific server
-version requirements as noted by ``minServerVersion`` and ``maxServerVersion``.
+version requirements as noted by the ``runOn`` section, if provided.
 
 Subdirectories for Test Formats
 -------------------------------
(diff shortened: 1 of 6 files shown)


source/crud/tests/v2/aggregate-out-readConcern.yml
```diff
@@ -1,10 +1,13 @@
+runOn:
+    -
+        minServerVersion: "4.1.0"
+        topology: ["replicaset", "sharded"]
```
jmikola: Correct me if I'm wrong, but this is really the only immediate need for adding `runOn`.

mbroadst: that is correct

source/crud/tests/README.rst
```diff
@@ -39,22 +39,34 @@ Test Format
...
+  - ``maxServerVersion`` (optional): The maximum server version (inclusive)
+    against which the tests can be run successfully. If this field is omitted,
+    it should be assumed that there is no upper bound on the required server
+    version.
+
+  - ``topology`` (optional): An array of server topologies against which the
+    tests can be run successfully. Valid topologies are "single", "replicaset",
+    and "sharded". If this field is omitted, the default is all topologies (i.e.
+    ``["single", "replicaset", "sharded"]``).
+
```
jmikola: Might have been missed from my previous comment, but can you add a note about `minServerVersion` and `maxServerVersion` under https://github.com/mongodb/specifications/blob/master/source/crud/tests/README.rst#legacy-test-format-for-single-operations? Feel free to copy-paste this:

> Instead of a top-level `runOn` field, server requirements are denoted by separate top-level `minServerVersion` and `maxServerVersion` fields. The minimum server version is an inclusive lower bound for running the test. The maximum server version is an exclusive upper bound for running the test. If a field is not present, it should be assumed that there is no corresponding bound on the required server version.

So long as the v1 directory exists, I'd like for that section to remain in sync and cover any test fields that would not be found in the v2 format.

p-mongo:
I think we have a precedent for fail points where they are described in one spec test readme and referenced in the other ones, would it be possible to pick a single place that is the authoritative description of runOn and reference it from other spec test readmes in a similar fashion?

jmikola:
> I think we have a precedent for fail points where they are described in one spec test readme and referenced in the other ones, would it be possible to pick a single place that is the authoritative description of runOn and reference it from other spec test readmes in a similar fashion?

Good point. I'd clarify that for fail points, we're linking to [this section](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst#server-fail-point) in the Transactions spec, which talks about the fail point itself (e.g. `failCommand`). The test format section that explains what the YAML/JSON key is tends to be copied verbatim between spec test READMEs.

We also have references to other specs when we talk about how to match documents or handle wildcards, but I don't believe any of the test _format_ sections (i.e. the bullets) refer to other specs.

I think a number of us would love to see this consolidated, but I'd propose holding off on that until after 4.2 and we can allocate some time to codifying Transactions as the "blessed" test format. That could then coincide with @mbroadst making a JSON schema to enforce the format and all that jazz.

mbroadst:
Yeah, I think we're about ready for a `test-runner.md` top level in this repo 💪

source/crud/tests/README.rst
```diff
@@ -137,6 +149,13 @@ single operation. Notable differences from the current format are as follows:
   defined under the ``tests[i].outcome.error`` and ``tests[i].outcome.result``
   fields.
 
+- Instead of a top-level ``runOn`` field, server requirements are denoted by
+  separate top-level ``minServerVersion`` and ``maxServerVersion`` fields. The
+  minimum server version is an inclusive lower bound for running the test. The
+  maximum server version is an exclusive upper bound for running the test. If a
+  field is not present, it should be assumed that there is no corresponding bound
+  on the required server version.
```
p-mongo: Is it possible for a test to use runOn and minSV/maxSV simultaneously? I think this is a difficult requirement to fulfill. If possible any given file should use one or the other.

jmikola: No test should use both simultaneously. This paragraph should make it clear that they are mutually exclusive:

> Instead of a top-level ``runOn`` field, server requirements are denoted by separate top-level ``minServerVersion`` and ``maxServerVersion`` fields.

This section is describing "Legacy Test Format for Single Operations" for the v1 directory and noting differences from the earlier "Test Format" section, which applies to the v2 directory.

p-mongo: I think I misread that as "server requirements *can be* denoted ...", thanks.

PR #573, opened 2019-06-18
Add command_name field to expectations for v2 CRUD tests

eppingere:
SPEC-1326

### DIFF
--- a/source/crud/tests/v2/aggregate-merge.yml
+++ b/source/crud/tests/v2/aggregate-merge.yml
@@ -25,6 +25,7 @@ tests:
           command:
             aggregate: *collection_name
             pipeline: *pipeline
+          command_name: "aggregate"
     outcome: &outcome
       collection:
         name: *output_collection
@@ -64,6 +65,7 @@ tests:
             aggregate: *collection_name
             pipeline: *pipeline
             readConcern: { level: "majority" }
+          command_name: "aggregate"
     outcome: *outcome
   -
     description: "Aggregate with $merge and local readConcern"
@@ -81,6 +83,7 @@ tests:
             aggregate: *collection_name
             pipeline: *pipeline
             readConcern: { level: "local" }
+          command_name: "aggregate"
     outcome: *outcome
   -
     description: "Aggregate with $merge and available readConcern"
@@ -98,4 +101,5 @@ tests:
             aggregate: *collection_name
             pipeline: *pipeline
             readConcern: { level: "available" }
+          command_name: "aggregate"
     outcome: *outcome
--- a/source/crud/tests/v2/aggregate-merge.json
+++ b/source/crud/tests/v2/aggregate-merge.json
@@ -69,7 +69,8 @@
                   }
                 }
               ]
-            }
+            },
+            "command_name" : "aggregate"
           }
         }
       ],
@@ -144,7 +145,8 @@
                 }
               ],
               "cursor": {}
-            }
+            },
+            "command_name" : "aggregate"
           }
         }
       ],
@@ -225,7 +227,8 @@
               "readConcern": {
                 "level": "majority"
               }
-            }
+            },
+            "command_name" : "aggregate"
           }
         }
       ],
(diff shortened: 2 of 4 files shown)


eppingere:
Let me know if there is anything else I need to do or if I did this completely wrong

source/crud/tests/v2/aggregate-merge.yml
```diff
@@ -25,6 +25,7 @@ tests:
           command:
             aggregate: *collection_name
             pipeline: *pipeline
+          command_name: "aggregate"
```
jmikola: These changes look fine, although I'm still not sure why this was necessary in the first place. 

In Slack, you mentioned:

> the spec for command monitors says that they will have a command_name field so that I can tell it's an aggregate command

The [Command Monitoring spec](https://github.com/mongodb/specifications/blob/master/source/command-monitoring/command-monitoring.rst) does refer to `commandName` as a property of all three command events (started, succeeded, and failed); however, the [spec's test README](https://github.com/mongodb/specifications/blob/master/source/command-monitoring/tests/README.rst) doesn't talk at all about the structure of `command_started_event` documents within an `expectation` array.

I don't think CRUD was alone in only asserting the `command` document. The [Retryable Reads spec tests](https://github.com/mongodb/specifications/tree/master/source/retryable-reads/tests) appears to do the same.

I'll also point out that other specs (excluding CRUD and Retryable Reads) assert the `database_name` property of the APM event. The general rule seems to be that if a field is missing, you simply avoid asserting it. This is surely the case for the APM spec itself, since it doesn't bother to assert fields such as `request_id` and `operation_id` in its own spec tests.

I previously stated in Slack:

> The omission of `command_name` was probably my oversight when drafting the CRUD tests. IIRC, `command_name` is mostly necessary for drivers that might not use an ordered dictionary for their command documents. Those were the document is ordered would infer an aggregate command by it being the first key in `command`.

In hindsight, that was incorrect. I was actually thinking of [`command_name`](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst#test-format) from the Transactions spec test format, which includes a `runCommand` operation. Drivers where a dictionary is unordered rely on _that_ field to properly construct the document for an outgoing command document. That has nothing to do with APM events, though.

I'm not sure what value `command_name` adds here. If your driver's spec test runner is properly matching the `command` document, then we can trust that the operation resulted in the appropriate command. `database_name` arguably provides more value, since that isn't discernible from the command document itself; however, that's only relevant for CRUD tests that expect a particular database name to be used (i.e. top-level `database_name` field).

It appears that `database_name` was introduced into a few other CRUD v2 tests by @prashantmital and I gather he only added it out of habit from working on other specs that already used the field in their APM assertions. Nothing about the operations in `bulkWrite-arrayFilters` or `updateWithPipelines` appears to actually care about the database used to execute the operations otherwise.

I won't suggest you introduce `database_name` to these tests, as I expect every test runner is capable of ignoring an assertion on the APM event's database name if the field is omitted in the test file. I expect that's also the case for `command_name`, too, which would make it equally safe to omit. Having said that, I won't veto adding either here.

eppingere: Makes sense. I will close the pull request?

jmikola: > It appears that database_name was introduced into a few other CRUD v2 tests by @prashantmital

For the record, I was mistaken there as well. Looks like `database_name` dates back to my own initial commits for v2 tests: https://github.com/mongodb/specifications/commit/5cf02a5f11bf15a084dca0524f997b934a00d7d2

> Makes sense. I will close the pull request?

If your driver's spec test runner is successfully asserting the `command` field and your team is happy with that, I'd be in favor of closing this and avoiding the need to request each driver sync new tests. Also happy to defer to @mbroadst.

eppingere: I have rewritten our spec test runner to not need the changes. Closing.

PR #575, opened 2019-06-19
SPEC-1331 use random AWS KMS key

kevinAlbs:
Updates generate.py to use a randomly generated data key (instead of a repeated value).


source/client-side-encryption/tests/find.json
```diff
@@ -11,7 +11,7 @@
             "_id": 1,
             "encrypted_string": {
                 "$binary": {
-                    "base64": "AQAAAAAAAAAAAAAAAAAAAAACtsdOjHK3CkpNjAAMznkYbeR6Z+yLzCRv4zOb7VfKnJdmU0W5MD8GVODx8K+KuBoCbE1SfGfPgX6NhS/RKpQ12w==",
```
shreyaskalyan: I'd also like to see some random BinData objects in here that are queried against (using something like the `_id` field) to see what they decrypt to.

kevinAlbs: We do test encrypting and decrypting random data with different values and different BSON types. Since we can't assert what the ciphertext is, we only currently check that a binary type is produced.

kevinAlbs:
Good catch. I updated localKMS to use a randomized local masterkey.

source/client-side-encryption/tests/find.json
kevinAlbs: I think I misinterpreted your comment. I added ciphertext that was encrypted with random encryption to the initial collection data. This validates we decrypt to the correct plaintext.

shreyaskalyan:
LGTM

PR #577, opened 2019-06-27
SPEC-1344 Add sleeps to make CMAP fairness test more reliable

ShaneHarvey:
See [SPEC-1344](https://jira.mongodb.org/browse/SPEC-1344) for an explanation.


daprahamian:
Can we add a prose test to replace the removed spec test?

mbroadst:
@ShaneHarvey I'm inclined to agree with @daprahamian on this one, is it difficult to implement a prose test for this?

ShaneHarvey:
I don't know how we could implement a proper prose test either. Wouldn't the race condition described in SPEC-1344 also happen in a prose test?

mbroadst:
the YAML test is confined by what we are able to do with the harness its working within. A prose test could simply require that you "prove" that the wait queue is "fair".

ShaneHarvey:
I've updated to keep the test in place but added some sleeps to make the test more reliable.

PR #580, opened 2019-06-28
SPEC-1335 Track option used to start changeStream for resume

prashantmital:
Adds a mandate for drivers to track the `resumeOption` (`startAfter`/`resumeAfter`) used to start a `changeStream` and ensure that the same option is re-used when resuming a `changeStream` with a `resumeToken`. 

This ensures that drivers that support the `postBatchResumeToken` in their `changeStream` APIs can safely resume a `changeStream` that returns an empty batch after an invalidate event.

If there are no glaring issues with this change, I will add a test.

### DIFF
--- a/source/change-streams/change-streams.rst
+++ b/source/change-streams/change-streams.rst
@@ -435,6 +435,8 @@ ChangeStream
 
 A ``ChangeStream`` is an abstraction of a `TAILABLE_AWAIT <https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#read>`_ cursor, with support for resumability.  Implementors MAY choose to implement a ``ChangeStream`` as an extension of an existing tailable cursor implementation.  If the ``ChangeStream`` is implemented as a type which owns a tailable cursor, then the implementor MUST provide a method to close the change stream, as well as satisfy the requirements of extending ``Iterable<Document>``.
 
+A change stream MUST track the name of the resume option used for creating the change stream as per `Tracking the Resume Option`_.
+
 A change stream MUST track the last resume token, per `Updating the Cached Resume Token`_.
 
 Drivers MUST raise an error on the first document received without a resume token (e.g. the user has removed ``_id`` with a pipeline stage), and close the change stream.  The error message SHOULD resemble “Cannot provide resume functionality when the resume token is missing”.
@@ -480,8 +482,7 @@ Once a ``ChangeStream`` has encountered a resumable error, it MUST attempt to re
 - Connect to selected server.
 - If there is a cached ``resumeToken``:
 
-  - The driver MUST set ``resumeAfter`` to the cached ``resumeToken``.
-  - The driver MUST NOT set ``startAfter``. If ``startAfter`` was in the original aggregation command, the driver MUST remove it.
+  - The driver MUST set the option named ``resumeOption`` (can be either ``resumeAfter`` or ``startAfter``) to the cached ``resumeToken``.
   - The driver MUST NOT set ``startAtOperationTime``. If ``startAtOperationTime`` was in the original aggregation command, the driver MUST remove it.
 - If there is no cached ``resumeToken`` and the ``ChangeStream`` has a saved operation time (either from an originally specified ``startAtOperationTime`` or saved from the original aggregation) and the max wire version is >= ``7``:
 
@@ -550,6 +551,18 @@ A possible interface for this callback MAY look like:
   }
 
 This MUST NOT be implemented in synchronous drivers. This MAY be implemented in asynchronous drivers.
+
+Tracking the Resume Option
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``ChangeStreamOptions`` member used to define the logical starting point for a change stream is referred to as the
+resume option. The following rule describes how to track the resume option:
+
+When the ``ChangeStream`` is started:
+
+  - If ``startAfter`` is set, the resume option is ``startAfter``.
+  - Else, the resume option is ``resumeAfter``.
+
 
 Updating the Cached Resume Token
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


source/change-streams/change-streams.rst
```diff
@@ -480,8 +482,7 @@ Once a ``ChangeStream`` has encountered a resumable error, it MUST attempt to re
 - Connect to selected server.
 - If there is a cached ``resumeToken``:
 
-  - The driver MUST set ``resumeAfter`` to the cached ``resumeToken``.
-  - The driver MUST NOT set ``startAfter``. If ``startAfter`` was in the original aggregation command, the driver MUST remove it.
+  - The driver MUST set the option named ``resumeOption`` (can be either ``resumeAfter`` or ``startAfter``) to the cached ``resumeToken``.
```
jmikola: Reading [SPEC-1335](https://jira.mongodb.org/browse/SPEC-1335) the problem this is trying to solve is:

 * One change stream ends with an `invalidate` event.
 * The `_id` from that `invalidate` event is captured and used as the `startAfter` option for a new change stream.
 * The new change stream executes successfully, and returns an empty initial batch with a `postBatchResumeToken`.
 * The change stream's cursor is killed.
 * The change stream is iterated, and attempting to use the `postBatchResumeToken` with the `resumeAfter` option yields a server-side error, because the token was associated with an `invalidate` event.

This seems like a server problem to me. Why is the server returning a value in `postBatchResumeToken` that cannot be used with `resumeAfter`?

Now, let's suppose we were connected to a version of the server that did _not_ return a `postBatchResumeToken`. In that case, the second change stream would still attempt to resume using the original token provided in the `startAfter` option to `watch()` (based on [Updating the Cached Resume Token](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#updating-the-cached-resume-token) and [Resume Process](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#resume-process), we'd never use the operation time). Since the driver has no way of knowing that the token originated with an `invalidate` event, I think we'd still hit this server error in this case. And since this scenario doesn't involve `postBatchResumeToken` at all, I'm curious if it's a better justification for the proposal in this PR.

cc: @cswanson310

prashantmital: If we are talking about server behavior, I would start by questioning the logic for having separate `startAfter`/`resumeAfter` options, especially when both take resume tokens that are identical. AFAICT, there would be no need for this ticket if there was just one resume option. In that situation, based on the server version `changeStream` would either terminate upon encountering an invalidate event, or would stay active if the server version being used supported `startAfter`. 

The scenario you describe without a `postBatchResumeToken` does indeed seem to be a better motivation for making the change proposed in the PR. If @cswanson310 gives this a green light, I can write a test for both situations (with and without `postBatchResumeToken`).

prashantmital: @jmikola see my updated comment above.

cswanson310: Oof this is a tricky problem and a nice find. Did this come up in testing? I like the proposal here. The conjecture that the server cannot return any other postBatchResumeToken unless there is another event in the oplog is correct. I suspect this wouldn't be a problem if the collection was dropped and then some other write happened. Still, this would remain a problem.

I do wonder whether having two mechanisms to resume was a mistake, but we debated that point at length when introducing 'startAfter' originally and I don't think this problem is enough motivation to open that can of worms again.

I have one concern with this approach: Suppose a user only wanted to use 'startAfter' to 'resume' a change stream after certain types of invalidating events. For example, if the stream wanted to handle drops and re-creates but was not equipped to handle renames. A driver automatically forwarding 'startAfter' might unintentionally keep the stream open after a rename. I acknowledge that a) this is probably unlikely or at least uncommon and b) if they aren't equipped to handle renames they'll probably close the cursor before even getting to the invalidate, but still an interesting case to think about.

I'm open to thinking about server changes that would alleviate this problem, but I can't think of anything that wouldn't involve changing the resume token somehow...

jmikola: > (a) the server changes its behavior to not return a postBatchResumeToken when returning batches from a changeStream in which the most recent event is an invalidate event
> (b) the server behavior remains the same and we ensure drivers have the requisite intelligence to always resume changeStreams

In either case, I think your suggested patch would address this. The point of my last paragraph was that `postBatchResumeToken` isn't actually relevant at all. It only gives the driver an alternative (preferred) value to use for the resume token.

I think we would need only two test for this PR, which would be:

1. If the `startAfter` option was used with `watch()`, we always resume with `startAfter`
2. If the `resumeAfter` option was used with `watch()`, we always resume with `resumeAfter`

If @cswanson310 doesn't get to this, you may want to follow up with @nzolnierzmdb. I believe he originally implemented the `startAfter` option. Also, you can see [SERVER-35030](https://jira.mongodb.org/browse/SERVER-35030) for some previous discussion about the option in comparison to `resumeAfter`.

I don't think the argument you're presenting here is new to `startAfter`. If anything, before `startAfter`, it would never have been possible to create a new change stream following an `invalidate` event. The [documentation for `startAfter`](https://docs.mongodb.com/master/changeStreams/#startafter-for-change-streams) actually highlights that as one of its benefits.

Perhaps the most conservative change here would be to only re-use the same option as was provided to `watch()` if we're resuming while never having received a change on the stream. I think we can trust that once the change stream emits an actual event document, we can trust its `_id` or `postBatchResumeToken` to be compatible with `resumeAfter`. That will also ensure that we don't introduce a bug whereby a driver actually attempts to continue beyond an `invalidate` event (were we to blindly continue to use `startAfter` for _all_ resume attempts).

ShaneHarvey: > This conservative approach should also ensure that we don't introduce a bug whereby a driver actually attempts to resume beyond an invalidate event (were we to blindly continue to use startAfter for all resume attempts).

I just want to point out that it makes no functional difference whether the driver uses `startAfter` or `resumeAfter` for the internal resume process. An `invalidate` event also sets the `cursor.id` which ends the cursor so the driver will never automatically resume after an `invalidate` event.

cswanson310: Hmm I don't have a really good argument for this but I do think the driver should continue to use resumeAfter. 'startAfter' was meant to be a distinction from 'resumeAfter' in the sense that 'startAfter' represents establishing a new stream after acknowledging the old one was invalidated, while 'resumeAfter' represents a resume after the client lost touch for some external reason. Also, for diagnosability in the logs and such it might be more confusing to see this less encouraged feature used everywhere? I have a slight preference for Jeremy's proposed conservative approach which I think will work, but I would accept using 'startAfter' if that's too tricky.

prashantmital: @cswanson310 while that might have been the intended usage for `startAfter` and `resumeAfter`, the fact is that both are currently exposed in the drivers as part of the ChangeStream public API. This means users are currently free to create change streams using commands like `Database.Collection.watch(resumeAfter=<change document _id>)` or `Database.Collection.watch(startAfter=<change document _id>)` and so, not all commands with `resumeAfter` in the logs will be the result of a driver resuming a change stream (a user might create a change stream with a non-null argument to `resumeAfter`), nor would all commands with `startAfter` be the result of a change stream being recreated by a user following an invalidate event (because users could conceivably write their applications to just always open change streams with `startAfter` since it is more permissive about the kinds of change document ids that can be used to start a stream). 

Furthermore, I would argue that from a log-parsing/diagnosability perspective it is better to go with using the same option as what was used to start the stream because that would ensure that aggregate commands corresponding to change stream creation and subsequent resume attempts all use the same set of change stream options. With the other design, someone parsing the logs looking for resume attempts corresponding to a change stream started with `startAfter` would need to factor in whether or not any results have been returned by the stream in order to figure out the structure of any subsequent calls to `aggregate` looking to resume the stream.

I'm sure I'm missing something, but using the same option seems like a better choice to me. If none of the arguments I've made in this response hold water, I will update the spec to only use `startAfter` on a resume attempt made before any results have been delivered.

cswanson310: Sure, granted. My point is not that none of the logs should contain startAfter, and also not that our users should not be able to use startAfter. My point was that we introduced the option to allow _users_ to open a _new_ stream when their old one was invalidated. We did not intend for drivers to use this instead of the original resumeAfter which was built for mostly for drivers. From this perspective, switching wholesale to startAfter feels like an abuse of the option. If we're going to switch all the way over we might as well consolidate the two options together and eliminate the distinction. In today's world, the value startAfter provides as an API (as opposed to just making resumeAfter work for invalidates) is a conceptual one: If I see "startAfter" in my application or in the logs, I know something different and maybe more interesting is going on.

prashantmital: Well, I am not proposing that we switching over wholesale to `startAfter`. I am proposing that we use the same option that the user used to create a `changeStream` to resume it. That said, I am happy to just implement this as you've recommended (only use `startAfter` when dealing with a cursor that has not yet returned any results) to finish up this spec and move on to the implementation work. 
On a side note, consolidating these options into one I think would be a good idea. The distinction between the options is hard to comprehend.

jmikola: I'm in favor of the conservative approach to only use `startAfter` for resuming when both of the following are true:

 * The `startAfter` option was also used for `watch()`
 * The change stream has yet to return a result document

prashantmital:
Superseded by #608

PR #581, opened 2019-06-28
Add note about other ChangeStreamDocument fields

ncarbon:


source/change-streams/change-streams.rst
```diff
@@ -197,6 +197,7 @@ The responses to a change stream aggregate or getMore have the following structu
   }
 
 **NOTE:** The above format is provided for illustrative purposes, and is subject to change without warning.
+There may be other fields existing in the ChangeStreamDocument that are left up to the driver to decide to include them in their API.
```
jmikola: Reading this in context of the outer section, it looks like this note should apply to the command response examples _and_ the structures defined above (especially since the command examples include a reference to ChangeStreamDocument).

Also, since the intention of your PR is to permit drivers to do something, I think it'd be a good candidate for MAY language.

I'd propose rewriting this whole paragraph and moving it to the very top (below the "Response Format" heading):

**NOTE:** The examples in this section are provided for illustrative purposes and are subject to change without warning. Drivers that provide a static type to represent ChangeStreamDocument MAY include additional fields in their API.

That said, I'm a bit torn with the spec is saying that ChangeStreamDocument can change on a whim between server versions while alluding to drivers attempting to map it as a static type (instead of exposing it as a generic document); however, that's beyond the scope of this PR and arguably goes back to [DRIVERS-471](https://jira.mongodb.org/browse/DRIVERS-471) and earlier.

PR #582, opened 2019-07-01
SPEC-1343 Clarify requirements for a "close" function

benjirewis:
Drivers that have idiomatic ways of disposing resources shouldn't have to implement the close function.

### DIFF
--- a/source/change-streams/change-streams.rst
+++ b/source/change-streams/change-streams.rst
@@ -433,7 +433,7 @@ Drivers MUST use the ``ns`` returned in the ``aggregate`` command to set the ``c
 ChangeStream
 ------------
 
-A ``ChangeStream`` is an abstraction of a `TAILABLE_AWAIT <https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#read>`_ cursor, with support for resumability.  Implementors MAY choose to implement a ``ChangeStream`` as an extension of an existing tailable cursor implementation.  If the ``ChangeStream`` is implemented as a type which owns a tailable cursor, then the implementor MUST provide a method to close the change stream, as well as satisfy the requirements of extending ``Iterable<Document>``.
+A ``ChangeStream`` is an abstraction of a `TAILABLE_AWAIT <https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#read>`_ cursor, with support for resumability.  Implementors MAY choose to implement a ``ChangeStream`` as an extension of an existing tailable cursor implementation.  If the ``ChangeStream`` is implemented as a type which owns a tailable cursor, then the implementor MUST provide a manner of closing the change stream, as well as satisfy the requirements of extending ``Iterable<Document>``. In languages with more idiomatic patterns of closing, such as those with RAII or destructors, drivers may use those patterns instead of a close method. 
 
 A change stream MUST track the last resume token, per `Updating the Cached Resume Token`_.
 
@@ -775,4 +775,7 @@ Changelog
 | 2019-04-12 | Clarified caching process for resume token.                |
 +------------+------------------------------------------------------------+
 | 2019-06-20 | Fix server version for addition of postBatchResumeToken    |
++------------+------------------------------------------------------------+
+| 2019-07-01 | Clarified that close may be implemented with more idiomatic|
+|            | patterns instead of a method.                              |
 +------------+------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


source/change-streams/change-streams.rst
```diff
@@ -433,7 +433,7 @@ Drivers MUST use the ``ns`` returned in the ``aggregate`` command to set the ``c
...
+A ``ChangeStream`` is an abstraction of a `TAILABLE_AWAIT <https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#read>`_ cursor, with support for resumability.  Implementors MAY choose to implement a ``ChangeStream`` as an extension of an existing tailable cursor implementation.  If the ``ChangeStream`` is implemented as a type which owns a tailable cursor, then the implementor MUST provide a manner of closing the change stream, as well as satisfy the requirements of extending ``Iterable<Document>``. In languages with more idiomatic patterns of closing, such as those with RAII or destructors, drivers may use those patterns instead of a close method.
```
saghm: "may" should be capitalized in the last line, i.e. "drivers MAY use those patterns".

source/change-streams/change-streams.rst
```diff
@@ -433,7 +433,7 @@ Drivers MUST use the ``ns`` returned in the ``aggregate`` command to set the ``c
...
+A ``ChangeStream`` is an abstraction of a `TAILABLE_AWAIT <https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#read>`_ cursor, with support for resumability.  Implementors MAY choose to implement a ``ChangeStream`` as an extension of an existing tailable cursor implementation.  If the ``ChangeStream`` is implemented as a type which owns a tailable cursor, then the implementor MUST provide a manner of closing the change stream, as well as satisfy the requirements of extending ``Iterable<Document>``. In languages with more idiomatic patterns of closing, such as those with RAII or destructors, drivers MAY use those patterns instead of a close method.
```
jmikola: Is the only change here that MAY is now capitalized (to use RFC terms) in "drivers may use"?

saghm: On the specific commit you commented on, yes. But the other commit in the PR adds the language about the `close` method requirement being fulfilled by RAII/destructors.

mbroadst: This reads a bit awkwardly to me, how about borrowing from the sessions spec, something like:
```
If your language has an idiomatic way of disposing resources you MAY choose to implement that in addition to, or instead of, an explicit close method, whichever is more idiomatic for your language.
```

benjirewis: I agree that this sounds more natural/explicit. Used this phrase (changed `disposing resources` to `disposing of resources`, though).

source/change-streams/change-streams.rst
```diff
@@ -776,3 +776,6 @@ Changelog
 +------------+------------------------------------------------------------+
 | 2019-06-20 | Fix server version for addition of postBatchResumeToken    |
 +------------+------------------------------------------------------------+
+| 2019-07-01 | Clarified that close may be implemented with more idiomatic|
```
jmikola: New change log entry seems like a good reason to bump the spec version. I believe it's currently 1.6.1. This is a minor clarification that neither changes or introduces functionality, so I think a patch bump to 1.6.2 is reasonable.

benjirewis: Sounds good!

source/change-streams/change-streams.rst
```diff
@@ -433,7 +433,7 @@ Drivers MUST use the ``ns`` returned in the ``aggregate`` command to set the ``c
...
+A ``ChangeStream`` is an abstraction of a `TAILABLE_AWAIT <https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#read>`_ cursor, with support for resumability.  Implementors MAY choose to implement a ``ChangeStream`` as an extension of an existing tailable cursor implementation.  If the ``ChangeStream`` is implemented as a type which owns a tailable cursor, then the implementor MUST provide a manner of closing the change stream, as well as satisfy the requirements of extending ``Iterable<Document>``. If your language has an idiomatic way of disposing of resources you MAY choose to implement that in addition to, or instead of, an explicit close method, whichever is more idiomatic for your language.
```
saghm: Sorry to continue the nitpicking! I think the final "whichever is more idiomatic for your language" clause isn't really necessary here; can we take it out?

benjirewis: I agree, as the last clause is implied by the earlier use of MAY. Changed.

PR #584, opened 2019-07-01
Fix yaml typo

saghm:
The YAML in the most recent changes to the auth spec tests was invalid, which caused the generation to JSON to fail when I attempted to run it for a different change. I've fixed it and generated the JSON again (which also picked up some unrelated changes to the CRUD spec tests, which were also updated without the makefile being run).


prashantmital:
👍 Yes, PLEASE. 👍

PR #585, opened 2019-07-01
Fix typo of "tlsAllowInsecure" to "tlsInsecure"

saghm:


saghm:
@skriptble actually gets the credit for catching this, not me

PR #587, opened 2019-07-02
SPEC-1351 NonResumableChangeStreamError error label test should have expectations

prashantmital:
Closes [SPEC-1351](https://jira.mongodb.org/browse/SPEC-1351).

This includes a bunch of unrelated changed because I based it off of #584 so that I could generate the JSON correctly. I will rebase off of master once #584 to get rid of that noise.

### DIFF
--- a/source/change-streams/tests/change-streams-errors.yml
+++ b/source/change-streams/tests/change-streams-errors.yml
@@ -70,6 +70,18 @@ tests:
         arguments:
           document:
             z: 3
+    expectations:
+      -
+        command_started_event:
+          command:
+            aggregate: *collection_name
+            cursor: {}
+            pipeline:
+              -
+                $changeStream:
+                  fullDocument: default
+              -
+                $project: { _id: 0 }
     result:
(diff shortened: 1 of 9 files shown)


source/change-streams/tests/change-streams-errors.yml
```diff
@@ -70,6 +70,18 @@ tests:
         arguments:
           document:
             z: 3
+    expectations:
```
jmikola: If I understand [SPEC-1351](https://jira.mongodb.org/browse/SPEC-1351) correctly, the concern is that a driver might resume and produce more than a single `command_started_event`? If so, I don't think adding this expectation achieves your goal, as `expectations` is permitted to be a subset of the total number of commands observed by APM. This is why I opened [SPEC-1341](https://jira.mongodb.org/browse/SPEC-1341) for another test.

prashantmital:
@jmikola indeed I realized that this doesn't test what I wanted it to. I'll close the PR.

PR #588, opened 2019-07-02
SPEC-1307 Fix server version for startAfter change stream option

divjotarora:


source/change-streams/tests/README.rst
```diff
@@ -168,7 +168,6 @@ The following tests have not yet been automated, but MUST still be tested
       - The batch is empty or has been iterated to the last document.
    - Expected result: 
       - ``getResumeToken`` must return the ``_id`` of the last document returned if one exists.
-      - ``getResumeToken`` must return ``startAfter`` from the initial aggregate if the option was specified.
```
jmikola: Was the removal of this line and the one below intentional? AFAIK, `startAfter` should still take precedence of the `resumeAfter` option.

divjotarora: I removed this line because the conditions for the test say "Running against a server <4.0.7" and `startAfter` isn't available until server version 4.1.1.

PR #590, opened 2019-07-03
[FLE] move test generation script into etc

kevinAlbs:
Since https://github.com/mongodb/specifications/pull/579 introduced another Python generation script for the corpus, I thought it'd be nice to move the test generation script into the `etc` directory of `client-side-encryption`. I went a little further, but there are no test changes or drivers changes required.

- renamed the test generation script `generate.py` to `generate-test.py` and moved it to `client-side-encryption/etc/` and now it takes an argument for the target directory instead of producing the tests in the current directory
- moved all `.yml.template` files from `client-side-encryption/tests` to `client-side-encryption/etc/test-templates`. Now the `tests` directory just contains the `yml` and `json` tests.
- modified `generate-corpus.py` to also take a target directory.


kevinAlbs:
@jyemin are you okay with this going in? I'd like to merge this soon to avoid rebase conflicts of other FLE spec test work in progress.

PR #591, opened 2019-07-03
SPEC-1353 and small test fix

samantharitter:
The "connections survive primary stepdown" tests mandate that connection pools be cleared, so the SDAM spec should mandate this, too.

Also, failCommand is only available in server versions 4.0 and later, so only run the "connections survive primary stepdown" tests on those versions.


mbroadst:
Hey @samantharitter, can you bump the spec minor version for this change? I think we can merge after that.

samantharitter:
@mbroadst ok, I bumped the SDAM minor version, and also changed the "last modified" fields to be today.

PR #595, opened 2019-07-03
SPEC-1296 : Adding ExceededTimeLimit, LockTimeout, and ClientDisconne…

rathisekaran:
…ct as retryable writes errors and corresponding error codes.

### DIFF
--- a/source/retryable-reads/tests/changeStreams-client.watch-serverErrors.yml
+++ b/source/retryable-reads/tests/changeStreams-client.watch-serverErrors.yml
@@ -147,3 +147,33 @@ tests:
         operations: [*retryable_operation_fails]
         expectations:
              - *retryable_command_started_event
+   -
+        description: "client.watch fails after ExceededTimeLimit"
+        clientOptions:
+            retryReads: false
+        failPoint:
+            <<: *failCommand_failPoint
+            data: { failCommands: [aggregate], errorCode: 262 }
+        operations: [*retryable_operation_fails]
+        expectations:
+             - *retryable_command_started_event
+   -
+        description: "client.watch fails after LockTimeout"
+        clientOptions:
+            retryReads: false
+        failPoint:
+            <<: *failCommand_failPoint
+            data: { failCommands: [aggregate], errorCode: 24 }
+        operations: [*retryable_operation_fails]
+        expectations:
+             - *retryable_command_started_event
+   -
+        description: "client.watch fails after ClientDisconnect"
+        clientOptions:
+            retryReads: false
+        failPoint:
+            <<: *failCommand_failPoint
+            data: { failCommands: [aggregate], errorCode: 279 }
+        operations: [*retryable_operation_fails]
+        expectations:
+             - *retryable_command_started_event
--- a/source/retryable-reads/tests/changeStreams-client.watch-serverErrors.yml
+++ b/source/retryable-reads/tests/changeStreams-client.watch-serverErrors.yml
@@ -147,33 +147,30 @@ tests:
         operations: [*retryable_operation_fails]
         expectations:
              - *retryable_command_started_event
-   -
-        description: "client.watch fails after ExceededTimeLimit"
-        clientOptions:
-            retryReads: false
+    -
+        description: "client.watch succeeds after ExceededTimeLimit"
         failPoint:
             <<: *failCommand_failPoint
             data: { failCommands: [aggregate], errorCode: 262 }
-        operations: [*retryable_operation_fails]
+        operations: [*retryable_operation]
         expectations:
              - *retryable_command_started_event
-   -
-        description: "client.watch fails after LockTimeout"
-        clientOptions:
-            retryReads: false
+             - *retryable_command_started_event
+    -
+        description: "client.watch succeeds after LockTimeout"
         failPoint:
             <<: *failCommand_failPoint
             data: { failCommands: [aggregate], errorCode: 24 }
-        operations: [*retryable_operation_fails]
+        operations: [*retryable_operation]
         expectations:
              - *retryable_command_started_event
-   -
-        description: "client.watch fails after ClientDisconnect"
-        clientOptions:
-            retryReads: false
+             - *retryable_command_started_event
+    -
+        description: "client.watch succeeds after ClientDisconnect"
         failPoint:
             <<: *failCommand_failPoint
             data: { failCommands: [aggregate], errorCode: 279 }
-        operations: [*retryable_operation_fails]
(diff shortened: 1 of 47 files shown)


source/retryable-reads/tests/changeStreams-db.watch-serverErrors.yml
```diff
@@ -147,7 +147,30 @@ tests:
...
-
+    -
+        description: "db.watch succeeds after ExceededTimeLimit"
+        failPoint:
+            <<: *failCommand_failPoint
+            data: { failCommands: [aggregate], errorCode: 262 }
+        operations: [*retryable_operation]
+        expectations:
+             - *retryable_command_started_event
+             - *retryable_command_started_event
```
jmikola: Why does `db.watch-serverErrors.yml` expect two events, while the analogous tests in `client.watch` and `db.coll.watch` only expect one event?

source/retryable-reads/tests/gridfs-download-serverErrors.yml
```diff
@@ -171,3 +171,33 @@ tests:
...
+    -
+        description: "Download succeeds after ExceededTimeLimit"
+        failPoint:
+            <<: *failCommand_failPoint
+            data: { failCommands: [find], errorCode: 262 }
+        operations: [*retryable_operation]
+        expectations:
+             - *retryable_command_started_event
+             - *retryable_command_started_event
+             - *find_chunks_command_started_event
```
jmikola: Confirming that this is consistent with other "succeed" tests earlier in the file, which include a trailing `*find_chunks_command_started_event`.

rathisekaran: Yes all of the 'Download succeeds after ...' tests include a trailing *find_chunks_command_started_event*

ShaneHarvey:
After making LockTimeout a transient error I've found a problem with this test from the `withTransaction` spec: [transaction is retried after commitTransaction TransientTransactionError (LockTimeout)
](https://github.com/mongodb/specifications/blob/master/source/transactions-convenient-api/tests/commit-transienttransactionerror.yml#L21)

The test attempts to prove that when a commit fails with LockTimeout the whole transaction is retried. However, once we make LockTimeout a `TransientTransactionError` then the behavior changes. Instead of retrying the whole transaction, pymongo considers LockTimeout to be an `UnknownTransactionCommitResult` error and only the commit is retried. PyMongo's behavior seems to follow the transaction spec which says:
> The driver MUST add the "UnknownTransactionCommitResult" error label when commitTransaction fails with a server selection error, network error, **retryable writes error**, MaxTimeMSExpired error, or write concern failed / timeout.

I think we need guidance from the server team on how to resolve this situation. We need to know what a driver should do after commitTransaction fails with each of these errors (ExceededTimeLimit, LockTimeout, and ClientDisconnect):
- always treat the error as "TransientTransactionError" so that `withTransaction` retries the entire transaction
- always treat the error as "UnknownTransactionCommitResult" so that `withTransaction` retries only the commit
- rely on the server to report the appropriate error label in the command response.

source/retryable-reads/tests/changeStreams-db.watch-serverErrors.yml
ShaneHarvey: I think the client.watch and db.coll.watch tests aren't right. They are using `retryReads: false` so they stop after the first failed command. Instead they should be testing the retry behavior like this file (db.watch-serverErrors.yml ) does .

jmikola: Ah, I didn't notice that `client.watch` and `db.coll.watch` were using `clientOptions`. I'll leave new comments to suggest that Rathi remove those.

source/retryable-reads/tests/changeStreams-client.watch-serverErrors.yml
```diff
@@ -147,3 +147,33 @@ tests:
...
+        failPoint:
+            <<: *failCommand_failPoint
+            data: { failCommands: [aggregate], errorCode: 24 }
+        operations: [*retryable_operation_fails]
+        expectations:
+             - *retryable_command_started_event
+   -
+        description: "client.watch fails after ClientDisconnect"
+        clientOptions:
+            retryReads: false
```
jmikola: Per https://github.com/mongodb/specifications/pull/595/files#r304042576, let's remove `retryReads: false` from these three tests and make the expectations consistent with `db.watch`.

source/retryable-reads/tests/changeStreams-db.coll.watch-serverErrors.yml
```diff
@@ -147,3 +147,33 @@ tests:
...
+        failPoint:
+            <<: *failCommand_failPoint
+            data: { failCommands: [aggregate], errorCode: 24 }
+        operations: [*retryable_operation_fails]
+        expectations:
+             - *retryable_command_started_event
+    -
+        description: "db.coll.watch fails after ClientDisconnect"
+        clientOptions:
+            retryReads: false
```
jmikola: Per https://github.com/mongodb/specifications/pull/595/files#r304042576, let's remove `retryReads: false` from these three tests and make the expectations consistent with `db.watch`.

source/retryable-reads/tests/changeStreams-client.watch-serverErrors.yml
rathisekaran: done.

source/retryable-reads/tests/changeStreams-db.coll.watch-serverErrors.yml
rathisekaran: Done.

source/retryable-reads/tests/changeStreams-db.coll.watch-serverErrors.yml
```diff
@@ -147,3 +147,30 @@ tests:
         operations: [*retryable_operation_fails]
         expectations:
              - *retryable_command_started_event
+    -
+        description: "db.coll.watch succeeds after ExceededTimeLimit"
+        failPoint:
+            <<: *failCommand_failPoint
+            data: { failCommands: [aggregate], errorCode: 262 }
+        operations: [*retryable_operation]
```
jmikola: I missed that these were also inconsistent with `db.watch-serverErrors.yml`. Good catch.

jyemin:
Abandoned.  There's a new PR up for this now.

jmikola:
Superseded by https://github.com/mongodb/specifications/pull/711

PR #598, opened 2019-07-07
[FLE] fix block quote formatting

kevinAlbs:


kevinAlbs:
Closing, folding into https://github.com/mongodb/specifications/pull/657

PR #600, opened 2019-07-08
SPEC-1350 Make fullDocument field in ChangeStreamOptions an optional string to handle the "default" case

ncarbon:


daprahamian:
I think this is going to require test runner changes, as there are multiple occurrences of asserting `fullDocument: default` [here](https://github.com/mongodb/specifications/blob/fa3faea156458c083222c035ec29880b43b4be23/source/change-streams/tests/change-streams.yml)

source/change-streams/change-streams.rst
```diff
@@ -260,16 +260,15 @@ Driver API
...
-     *
+     * Allowed values: ‘updateLookup’.  When set to ‘updateLookup’, the change notification
+     * for partial updates will include both a delta describing the changes to the document,
+     * as well as a copy of the entire document that was changed from some time after the 
+     * change occurred. The default is to not send a value. 
+     * For forward compatibility, a driver MUST NOT raise an error when a user provides an
+     * unknown value. The driver relies on the server to validate this option.
      * @note this is an option of the `$changeStream` pipeline stage.
      */
-    fullDocument: string = ‘default’;
```
jmikola: I left a comment in [SPEC-1350](https://jira.mongodb.org/browse/SPEC-1350) to ask if _all_ server versions since 3.6.0 allow this.

Beyond that, I think this warrants a bump to spec document's minor version (1.6.1 -> 1.7.0) in the document header, as well as a change log entry near the end of the document.

Also, this will warrant removing the `fullDocument` field (within `$changeStream`) from all APM assertions within the spec tests. AFAIK, that should instruct test runners to omit any check for the field (while still allowing its presence if drivers always send "default") per the [Spec Test Match Function](https://github.com/mongodb/specifications/tree/master/source/change-streams/tests#spec-test-match-function) rules.

source/change-streams/change-streams.rst
```diff
@@ -260,16 +260,15 @@ Driver API
...
-     * document that was changed from some time after the change occurred.  For forward
-     * compatibility, a driver MUST NOT raise an error when a user provides an unknown
-     * value. The driver relies on the server to validate this option.
-     *
+     * Allowed values: ‘updateLookup’.  When set to ‘updateLookup’, the change notification
+     * for partial updates will include both a delta describing the changes to the document,
+     * as well as a copy of the entire document that was changed from some time after the 
+     * change occurred. The default is to not send a value. 
+     * For forward compatibility, a driver MUST NOT raise an error when a user provides an
+     * unknown value. The driver relies on the server to validate this option.
```
jmikola: For the benefit of other reviewers, [`db.collection.watch()`](https://docs.mongodb.com/manual/reference/method/db.collection.watch/index.html#db.collection.watch) also does not mention the "default" string, so this is consistent with the server manual.

jmikola:
Is this still waiting to be merged? I think 2/3 reviewers may be sufficient.

I'm not sure why `master` was merged into the PR branch, but now that it has been I would suggest using "Squash and merge" to get this back into `master`. I'm not sure whether "Rebase and merge" would remove that commit, and if not it'd be preferable to omit the extra noise in the main repo branch.

PR #606, opened 2019-07-13
SPEC-1359 update tests with new prohibitions

kevinAlbs:
This updates the tests given the new prohibition of fields that are auto encrypted deterministically using keyAltName.

Validated by https://github.com/mongodb/mongo-java-driver/pull/517 with mongod 4.3.0-574-g6e02a4d

This is test removal, so it should be safe to sync these tests without updating mongod/mongocryptd.


source/client-side-encryption/corpus/corpus-encrypted.json
```diff
@@ -11,7 +11,7 @@
     "allowed": true,
     "value": {
       "$binary": {
-        "base64": "AgFkgAAAAAAAAAAAAAAAAAABM9KQ+Ph2lxyYUIV+Tu7gkrFndur280S5r43432bT+Q0WKKGSlu7s4anRPCvpwSyC7zP2VOD9ggBp29OvXsjm6A==",
```
kevinAlbs: Random values all changed during regeneration. It seems silly to include them the expectation data when there's really nothing to compare them against, I could remove them in a follow up commit if it matters.

source/client-side-encryption/corpus/corpus-encrypted.json
```diff
@@ -414,20 +400,6 @@
       }
     }
   },
-  "aws_binData=00_det_auto_altname": {
```
kevinAlbs: "auto" encrypted with "altname" is removed from the schema and data, since we cannot configure a schema with such a field. But "explicit" encrypted with "altname" still exists.

PR #608, opened 2019-07-15
SPEC-1335 clarify resume process for changeStreams started with startAfter

prashantmital:
Supersedes #580 

Prose tests had to be added instead of spec tests because the spec test format does not provide a way to initialize a change stream with a valid `startAfter` token.

### DIFF
--- a/source/change-streams/change-streams.rst
+++ b/source/change-streams/change-streams.rst
@@ -10,7 +10,7 @@ Change Streams
 :Type: Standards
 :Minimum Server Version: 3.6
 :Last Modified: April 3, 2019
-:Version: 1.6.2
+:Version: 1.6.3
 
 .. contents::
 
--- a/source/change-streams/change-streams.rst
+++ b/source/change-streams/change-streams.rst
@@ -481,10 +481,19 @@ Once a ``ChangeStream`` has encountered a resumable error, it MUST attempt to re
 - Connect to selected server.
 - If there is a cached ``resumeToken``:
 
-  - The driver MUST set ``resumeAfter`` to the cached ``resumeToken``.
-  - The driver MUST NOT set ``startAfter``. If ``startAfter`` was in the original aggregation command, the driver MUST remove it.
-  - The driver MUST NOT set ``startAtOperationTime``. If ``startAtOperationTime`` was in the original aggregation command, the driver MUST remove it.
-- If there is no cached ``resumeToken`` and the ``ChangeStream`` has a saved operation time (either from an originally specified ``startAtOperationTime`` or saved from the original aggregation) and the max wire version is >= ``7``:
+  - If the ``ChangeStream`` was started with ``startAfter`` and has yet to return a result document:
+
+    - The driver MUST set ``startAfter`` to the cached ``resumeToken``.
+    - The driver MUST NOT set ``resumeAfter``.
+    - The driver MUST NOT set ``startAtOperationTime``. If ``startAtOperationTime`` was in the original aggregation command, the driver MUST remove it.
+
+  - Else:
+
+    - The driver MUST set ``resumeAfter`` to the cached ``resumeToken``.
+    - The driver MUST NOT set ``startAfter``. If ``startAfter`` was in the original aggregation command, the driver MUST remove it.
+    - The driver MUST NOT set ``startAtOperationTime``. If ``startAtOperationTime`` was in the original aggregation command, the driver MUST remove it.
+
+- Else if there is no cached ``resumeToken`` and the ``ChangeStream`` has a saved operation time (either from an originally specified ``startAtOperationTime`` or saved from the original aggregation) and the max wire version is >= ``7``:
 
   - The driver MUST NOT set ``resumeAfter``.
   - The driver MUST NOT set ``startAfter``.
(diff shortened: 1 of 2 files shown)


source/change-streams/change-streams.rst
```diff
@@ -481,10 +481,19 @@ Once a ``ChangeStream`` has encountered a resumable error, it MUST attempt to re
...
 - If there is a cached ``resumeToken``:
 
-  - The driver MUST set ``resumeAfter`` to the cached ``resumeToken``.
-  - The driver MUST NOT set ``startAfter``. If ``startAfter`` was in the original aggregation command, the driver MUST remove it.
-  - The driver MUST NOT set ``startAtOperationTime``. If ``startAtOperationTime`` was in the original aggregation command, the driver MUST remove it.
-- If there is no cached ``resumeToken`` and the ``ChangeStream`` has a saved operation time (either from an originally specified ``startAtOperationTime`` or saved from the original aggregation) and the max wire version is >= ``7``:
+  - If the ``ChangeStream`` was started with ``startAfter`` and has yet to return a result document:
+
+    - The driver MUST set ``startAfter`` to the cached ``resumeToken``.
+    - The driver MUST NOT set ``resumeAfter``.
```
jmikola: I take it that there is no need for the "If ``resumeAfter`` was in the original aggregation command..." copypasta here because the original `aggregate` would have failed with a server error due to using both options at once and we would never get as far as needing to resume?

source/change-streams/tests/README.rst
```diff
@@ -160,6 +160,8 @@ The following tests have not yet been automated, but MUST still be tested
 #. Ensure that a cursor returned from an aggregate command with a cursor id and an initial empty batch is not closed on the driver side.
 #. The ``killCursors`` command sent during the "Resume Process" must not be allowed to throw an exception.
 #. ``$changeStream`` stage for ``ChangeStream`` against a server ``>=4.0`` and ``<4.0.7`` that has not received any results yet MUST include a ``startAtOperationTime`` option when resuming a changestream.
+#. ``$changeStream`` stage for ``ChangeStream`` started with ``startAfter`` against a server ``>=4.1.1`` that has not received any results yet MUST include a ``startAfter`` option and MUST NOT include a ``resumeAfter`` option when resuming a changestream.
+#. ``$changeStream`` stage for ``ChangeStream`` started with ``startAfter`` against a server ``>=4.1.1`` that has received at least one result MUST include a ``resumeAfter`` option and MUST NOT include a ``startAfter`` option when resuming a changestream.
```
jmikola: This looks like it's going to shift all of the numbers for later prose tests. I hope no drivers refer to them by number in their test suites.

Perhaps it'd be best not to use numbered bullets for these at all? If so, could we do that in a separate commit/PR?

source/change-streams/tests/README.rst
```diff
@@ -160,6 +160,8 @@ The following tests have not yet been automated, but MUST still be tested
 #. Ensure that a cursor returned from an aggregate command with a cursor id and an initial empty batch is not closed on the driver side.
 #. The ``killCursors`` command sent during the "Resume Process" must not be allowed to throw an exception.
 #. ``$changeStream`` stage for ``ChangeStream`` against a server ``>=4.0`` and ``<4.0.7`` that has not received any results yet MUST include a ``startAtOperationTime`` option when resuming a changestream.
+#. ``$changeStream`` stage for ``ChangeStream`` started with ``startAfter`` against a server ``>=4.1.1`` that has not received any results yet MUST include a ``startAfter`` option and MUST NOT include a ``resumeAfter`` option when resuming a changestream.
```
jmikola: I realize you're just being consistent with previous usage, but can we write "change stream" instead of "changestream"? "ChangeStream" is one thing, since it maps to the class name in the spec, but "changestream" does not read well. Feel free to make a global replacement in this file if that's easier.

source/change-streams/tests/README.rst
ShaneHarvey: Yeah I agree here. I've recently started referring to prose tests via their number in the spec. I think it much easier to track which test is which.

ShaneHarvey:
>  I suppose patch is preferred since this is purely internal behavior.

Agreed.

source/change-streams/change-streams.rst
prashantmital: That's right. Since an `aggregation` with both those options would fail anyway (with a non-resumable error), I didn't think it made sense to explicitly specify behavior for that case.

source/change-streams/tests/README.rst
prashantmital: I have been super confused by this usage myself. Why don't we simply use the class name `ChangeStream` everywhere in this document?

source/change-streams/tests/README.rst
prashantmital: Sounds like @jmikola is advocating for getting rid of numbered tests, whereas @ShaneHarvey is in favor of keeping the numbers. Can you please clarify which course of action I should take?

ShaneHarvey: Oh sorry I misread Jeremy's comment. I am pro-numbers and I would like for new tests to be added at the end of the list. I think it's easier for test implementors to refer to tests via the number.

jmikola: I was advocating for removing the number bullets because tests may be removed (see: [SPEC-1360](https://jira.mongodb.org/browse/SPEC-1360)).

prashantmital: For the time being, I will add the new tests to the end of the list. @jmikola I suppose we can pursue the question of whether to number the tests or not in a new JIRA ticket?

jmikola: Indeed. There's no need to do it here. I was just leaving the option open.

jmikola: https://jira.mongodb.org/browse/SPEC-1364

PR #611, opened 2019-07-18
Remove startAfter from changeStream test meant for server version <4.0.7

prashantmital:
Closes [SPEC-1367](https://jira.mongodb.org/browse/SPEC-1367)
Also renumbers changeStreams prose tests with fixed numbers in accordance with [SPEC-1364](https://jira.mongodb.org/browse/SPEC-1364).

PR #612, opened 2019-07-18
SPEC-1360 Remove unrealistic change stream prose tests

prashantmital:
Closes [SPEC-1360](https://jira.mongodb.org/browse/SPEC-1360).

### DIFF
--- a/source/change-streams/tests/README.rst
+++ b/source/change-streams/tests/README.rst
@@ -186,5 +186,6 @@ The following tests have not yet been automated, but MUST still be tested
       - ``getResumeToken`` must return ``startAfter`` from the initial aggregate if the option was specified.
       - ``getResumeToken`` must return ``resumeAfter`` from the initial aggregate if the option was specified.
       - If neither the ``startAfter`` nor ``resumeAfter`` options were specified, the ``getResumeToken`` result must be empty.
+    - Note that this test cannot be run against sharded topologies because in that case the initial ``aggregate`` command only establishes cursors on the shards and always returns an empty ``firstBatch``.
 17. ``$changeStream`` stage for ``ChangeStream`` started with ``startAfter`` against a server ``>=4.1.1`` that has not received any results yet MUST include a ``startAfter`` option and MUST NOT include a ``resumeAfter`` option when resuming a change stream.
 18. ``$changeStream`` stage for ``ChangeStream`` started with ``startAfter`` against a server ``>=4.1.1`` that has received at least one result MUST include a ``resumeAfter`` option and MUST NOT include a ``startAfter`` option when resuming a change stream.
--- a/source/change-streams/tests/README.rst
+++ b/source/change-streams/tests/README.rst
@@ -164,43 +164,27 @@ The following tests have not yet been automated, but MUST still be tested
 11. - For a ``ChangeStream`` under these conditions:
       - Running against a server ``>=4.0.7``.
       - The batch is empty or has been iterated to the last document.
-   - Expected result: 
+    - Expected result:
        - ``getResumeToken`` must return the ``postBatchResumeToken`` from the current command response.
 12. - For a ``ChangeStream`` under these conditions:
       - Running against a server ``<4.0.7``.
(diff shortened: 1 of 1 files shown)


source/change-streams/tests/README.rst
```diff
@@ -164,43 +164,28 @@ The following tests have not yet been automated, but MUST still be tested
 11. - For a ``ChangeStream`` under these conditions:
       - Running against a server ``>=4.0.7``.
       - The batch is empty or has been iterated to the last document.
-   - Expected result: 
+    - Expected result:
```
prashantmital: This was a formatting issue overlooked in #611. The Travis test wasn't smart enough to catch.

source/change-streams/tests/README.rst
```diff
@@ -164,43 +164,28 @@ The following tests have not yet been automated, but MUST still be tested
...
-      - If neither the ``startAfter`` nor ``resumeAfter`` options were specified, the ``getResumeToken`` result must be empty.
-#. ``$changeStream`` stage for ``ChangeStream`` started with ``startAfter`` against a server ``>=4.1.1`` that has not received any results yet MUST include a ``startAfter`` option and MUST NOT include a ``resumeAfter`` option when resuming a change stream.
-#. ``$changeStream`` stage for ``ChangeStream`` started with ``startAfter`` against a server ``>=4.1.1`` that has received at least one result MUST include a ``resumeAfter`` option and MUST NOT include a ``startAfter`` option when resuming a change stream.
+    - Note that this test cannot be run against sharded topologies because in that case the initial ``aggregate`` command only establishes cursors on the shards and always returns an empty ``firstBatch``.
+17. ``$changeStream`` stage for ``ChangeStream`` started with ``startAfter`` against a server ``>=4.1.1`` that has not received any results yet MUST include a ``startAfter`` option and MUST NOT include a ``resumeAfter`` option when resuming a change stream.
```
prashantmital: I forgot to hard-code these test numbers in #611

source/change-streams/tests/README.rst
```diff
@@ -164,43 +164,28 @@ The following tests have not yet been automated, but MUST still be tested
...
-      - The stream has iterated beyond a previous batch and a ``getMore`` command has just been executed.
-   - Expected result:
-      - ``getResumeToken`` must return the ``_id`` of the previous document returned if one exists.
-      - ``getResumeToken`` must return ``resumeAfter`` from the initial aggregate if the option was specified.
-      - If neither the ``startAfter`` nor ``resumeAfter`` options were specified, the ``getResumeToken`` result must be empty.
-#. ``$changeStream`` stage for ``ChangeStream`` started with ``startAfter`` against a server ``>=4.1.1`` that has not received any results yet MUST include a ``startAfter`` option and MUST NOT include a ``resumeAfter`` option when resuming a change stream.
-#. ``$changeStream`` stage for ``ChangeStream`` started with ``startAfter`` against a server ``>=4.1.1`` that has received at least one result MUST include a ``resumeAfter`` option and MUST NOT include a ``startAfter`` option when resuming a change stream.
+    - Note that this test cannot be run against sharded topologies because in that case the initial ``aggregate`` command only establishes cursors on the shards and always returns an empty ``firstBatch``.
```
jmikola: I think you also mentioned that all of the current spec tests are only expected to run against replica sets. I can also verify that by the `topology` field in each test.

Was this the only prose test that cannot be run against a sharded cluster? If so, perhaps we should mention earlier in the document that other prose tests _should_ be run against replica set and sharded clusters.

I also saw your discussion in `#server-query` and the answers explaining why sharded clusters are a unique case where we need to assert that a change eventually appears but cannot expect it in a particular batch. It sounds like there's an action item to revisit/rethink our driver testing for that, so would you mind creating a separate spec ticket to summarize the discussion/todos so we don't lose track of it?

prashantmital: Opened https://jira.mongodb.org/browse/SPEC-1374 and https://jira.mongodb.org/browse/SPEC-1375.
I will also update the prose tests to run the tests on sharded clusters.

PR #615, opened 2019-07-24
SPEC-1372 test batch splitting and BSON size limits

kevinAlbs:
Unfortunately, this proposed change does require updates to the Java tests, but I thought it would be worthwhile to test that the reduced BSON document size limit does not apply after automatic encryption.


source/client-side-encryption/tests/README.rst
```diff
@@ -293,6 +296,81 @@ Then, run the following final tests:
...
+
+#. Insert ``no_encryption_over_2mib``. Expect this to throw an exception due to exceeding the reduced maximum BSON document size.
+
+#. Insert ``encryption_equal_2mib``. Expect this to succeed since after encryption this still is below the normal maximum BSON document size.
+
+#. Bulk insert two copies of ``no_encryption_equal_2mib``. Expect the bulk write to succeed and split after first doc (i.e. two inserts occur). This may be verified using `command monitoring <https://github.com/mongodb/specifications/tree/master/source/command-monitoring/command-monitoring.rst>`_.
+
+#. Bulk insert two copies of ``encryption_equal_2mib``. Expect the bulk write to succeed and split after first doc (i.e. two inserts occur).
+
+Optionally, if it is possible to mock the maxWriteBatchSize (i.e. the maximum number of documents in a batch) test that setting maxWriteBatchSize=1 and inserting the two documents ``{ "_id": "a" }, { "_id": "b" }`` with ``client_encrypted`` splits the operation into two inserts.
```
kevinAlbs: I made this optional, since I'm not sure if it's reasonable to expect it to be easy for drivers to mock this for the test.

jyemin: acknowledged

source/client-side-encryption/tests/README.rst
```diff
@@ -293,6 +296,81 @@ Then, run the following final tests:
...
+
+Using ``client_encrypted`` perform the following operations:
+
+#. Insert ``no_encryption_equal_2mib``. Expect this to succeed.
+
+#. Insert ``no_encryption_over_2mib``. Expect this to throw an exception due to exceeding the reduced maximum BSON document size.
+
+#. Insert ``encryption_equal_2mib``. Expect this to succeed since after encryption this still is below the normal maximum BSON document size.
+
+#. Bulk insert two copies of ``no_encryption_equal_2mib``. Expect the bulk write to succeed and split after first doc (i.e. two inserts occur). This may be verified using `command monitoring <https://github.com/mongodb/specifications/tree/master/source/command-monitoring/command-monitoring.rst>`_.
```
jyemin: Can't insert two copies of this since the _id would be identical

source/client-side-encryption/tests/README.rst
```diff
@@ -293,6 +296,81 @@ Then, run the following final tests:
...
+
+#. Bulk insert two copies of ``encryption_equal_2mib``. Expect the bulk write to succeed and split after first doc (i.e. two inserts occur).
```
jyemin: Same

source/client-side-encryption/tests/README.rst
kevinAlbs: Doh, fixed.

source/client-side-encryption/tests/README.rst
```diff
@@ -293,6 +296,68 @@ Then, run the following final tests:
...
+
+      Mng0NCt4ZHVUYUJCa1kxNkVyNUR1QURhZ2h2UzR2d2RrZzh0cFBwM3R6NmdWMDFBMUN3YkQ5aXRRMkhGRGdQV09wOGVNYUMxT2k3NjZKelhaQmRCZGJkTXVyZG9uSjFk
+
+   Configure with the ``keyVaultNamespace`` set to ``admin.datakeys``.
+
+Using ``client_encrypted`` perform the following operations:
+
+#. Insert ``{ "_id": "no_encryption_equal_2mib", "unencrypted": <the string "a" repeated 2097095 times> }``.
+
+   Expect this to succeed.
```
jyemin: This doesn't succeed.  I think this is assuming that the test will succeed if the document to insert is equal to 2 mib.  But that's not the case.  What matters is the size of the whole insert command document, which is quite a bit larger.

jyemin: In my tests this succeeds when the string "a" is repeated a maximum of 2097006 times

kevinAlbs:
I've softened the boundaries a bit, so the documents that are meant to be close to but under 2 MiB are now sized to 2 MiB - 1000 bytes.

In order to validate that encryption resulting in documents over 2 MiB (but under 16 MiB) still works, I've generated a document that has 100 auto-encrypted fields (limits-doc.json), sized at 1005 bytes. After encryption this turns into about 10,000 bytes.

I've validated all tests except the optional one in a hacky addition to the groovy tests on my fork of the Java driver:
https://github.com/kevinAlbs/mongo-java-driver/tree/fle-limits

PR #617, opened 2019-07-29
fix default domain socket URI

kevinAlbs:


kevinAlbs:
Yeah, that's true. This will be a short lived fix.

PR #618, opened 2019-07-29
SPEC-1386 Test that auto encryption on views is prohibited

kevinAlbs:
Adds a spec test for a prohibited database operation for automatic encryption, and a prose test for a view operation.

I verified the new prose test with: https://github.com/kevinAlbs/JavaFLEProsePoC/blob/9ad5859b9aec03c8d4b4ecdef991c66a7de3d2a4/src/main/java/ViewTest.java

And verified the new spec test with the Java test runner.

### DIFF
--- a/source/client-side-encryption/tests/README.rst
+++ b/source/client-side-encryption/tests/README.rst
@@ -193,6 +193,12 @@ Prose Tests
 
 Tests for the ClientEncryption type are not included as part of the YAML tests.
 
+In the prose tests LOCAL_MASTERKEY refers to the following base64:
+
+.. code:: javascript
+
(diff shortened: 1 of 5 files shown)


source/client-side-encryption/tests/README.rst
```diff
@@ -216,12 +222,6 @@ First, perform the setup.
           "local": { "key": <base64 decoding of LOCAL_MASTERKEY> }
       }
 
-   Where LOCAL_MASTERKEY is the following base64:
```
kevinAlbs: Removed this repetition while here.

kevinAlbs:
Noting in the ticket that this should not be required for GA.

PR #619, opened 2019-07-29
fix change stream test formatting

kevinAlbs:


prashantmital:
Somehow I completely missed this PR. I'm guessing this is no longer an issue since Travis is passing on current master? If that's the case lets close it.

kevinAlbs:
I don't think Travis was catching this. The formatting is off, but the RST files still render. I rebased the change.

PR #620, opened 2019-07-30
SPEC-1369 + SPEC-1368 add all padding lengths to corpus

kevinAlbs:
And update the steps to explicitly decrypt the expected corpus fields.

Validated with Java driver:
https://github.com/mongodb/mongo-java-driver/pull/519

### DIFF
--- a/source/client-side-encryption/corpus/corpus-encrypted.json
+++ b/source/client-side-encryption/corpus/corpus-encrypted.json
@@ -166,7 +166,7 @@
     "algo": "det",
     "method": "explicit",
     "identifier": "altname",
-    "allowed": false,
+    "allowed": true,
     "value": {
       "$binary": {
         "base64": "AQFkgAAAAAAAAAAAAAAAAAACyvOW8NcqRkZYzujivwVmYptJkic27PWr3Nq3Yv5Njz8cJdoyesVaQan6mn+U3wdfGEH8zbUUISdCx5qgvXEpvw==",
@@ -420,7 +420,7 @@
     "algo": "det",
     "method": "explicit",
     "identifier": "altname",
-    "allowed": false,
+    "allowed": true,
     "value": {
       "$binary": {
         "base64": "AQFkgAAAAAAAAAAAAAAAAAAF6SJGmfD3hLVc4tLPm4v2zFuHoRxUDLumBR8Q0AlKK2nQPyvuHEPVBD3vQdDi+Q7PwFxmovJsHccr59VnzvpJeg==",
@@ -518,7 +518,7 @@
     "algo": "det",
     "method": "explicit",
     "identifier": "altname",
-    "allowed": false,
+    "allowed": true,
     "value": {
       "$binary": {
         "base64": "AQFkgAAAAAAAAAAAAAAAAAAFlg7ceq9w/JMhHcNzQks6UrKYAffpUyeWuBIpcuLoB7YbFO61Dphseh77pzZbk3OvmveUq6EtCP2pmsq7hA+QV4hkv6BTn4m6wnXw6ss/qfE=",
@@ -652,7 +652,7 @@
     "algo": "det",
     "method": "explicit",
     "identifier": "altname",
-    "allowed": false,
+    "allowed": true,
     "value": {
       "$binary": {
         "base64": "AQFkgAAAAAAAAAAAAAAAAAAHTMY2l+gY8glm4HeSsGfCSfOsTVTzYU8qnQV8iqEFHrO5SBJac59gv3N/jukMwAnt0j6vIIQrROkVetU24YY7sQ==",
@@ -824,7 +824,7 @@
     "algo": "det",
     "method": "explicit",
     "identifier": "altname",
-    "allowed": false,
+    "allowed": true,
     "value": {
       "$binary": {
         "base64": "AQFkgAAAAAAAAAAAAAAAAAAJEEpQNsiqMWPqD4lhMkiOJHGE8FxOeYrKPiiAp/bZTrLKyCSS0ZL1WT9H3cGzxWPm5veihCjKqWhjatC/pjtzbQ==",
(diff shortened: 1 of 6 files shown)


source/client-side-encryption/etc/update-corpus.py
```diff
@@ -0,0 +1,28 @@
+import json
+import sys
+
+description = """Update the existing corpus data from a file with only deterministically encrypted fields.
```
kevinAlbs: I added this script to make it easier to update the corpus data without updating the random fields to make a simpler diff.

source/client-side-encryption/etc/validate-corpus.py
```diff
@@ -0,0 +1,35 @@
+import json
+import sys
+
+description = """Validate that the same deterministically encrypted values in the corpus encrypt to the same ciphertext in the encrypted corpus.
```
kevinAlbs: Instead of having a more complicated O(n^2) check in drivers to check all deterministically encrypted values that  should be equal, I decided to opt for a python script to check `corpus-encrypted.json` once.

That way drivers check the one corresponding field, e.g. for `aws_double_det_explicit_id` just check that the ciphertext produced is an exact match for the `aws_double_det_explicit_id` field in `corpus-encrypted.json`. This makes it simpler to add the padding cases without adding more special cases to the check in the spec.

source/client-side-encryption/tests/README.rst
```diff
@@ -473,10 +476,10 @@ The corpus test exhaustively enumerates all ways to encrypt all BSON value types
 
    Iterate over each field of ``corpus_encrypted_expected`` and check the following:
 
-   - If the ``algo`` is ``det``, that the value exactly matches all fields in ``corpus_encrypted_actual`` with the same ``kms``, ``type``, and ``algo``.
```
kevinAlbs: Now, this rule is partially checked with `validate-corpus.py` above.

source/client-side-encryption/tests/README.rst
```diff
@@ -473,10 +476,10 @@ The corpus test exhaustively enumerates all ways to encrypt all BSON value types
...
-   - If the ``algo`` is ``rand`` and ``allowed`` is true, that the value does not match any fields in ``corpus_encrypted_actual`` with the same ``kms`` and ``type``.
-   - If the ``method`` is ``auto`` or ``explicit``, decrypt the value with ``client_encryption`` and validate the value exactly matches the corresponding field of ``corpus``.
-   - If the ``allowed`` is false, validate the value exactly matches the corresponding field of ``corpus``.
+   - If the ``algo`` is ``det``, that the value exactly the value of the corresponding field in ``corpus_encrypted_actual``.
+   - If the ``algo`` is ``rand`` and ``allowed`` is true, that the value does not the value of the corresponding field in ``corpus_encrypted_actual``.
+   - If ``allowed`` is true, decrypt the value with ``client_encryption``. Decrypt the corresponding field of ``corpus_encrypted`` and validate that they both match.
```
kevinAlbs: This satisfies SPEC-1369 by decrypting the corresponding value in `corpus_encrypted`.

source/client-side-encryption/corpus/corpus-encrypted.json
```diff
@@ -166,7 +166,7 @@
     "algo": "det",
     "method": "explicit",
     "identifier": "altname",
-    "allowed": false,
+    "allowed": true,
```
kevinAlbs: Oddly, these were set to the wrong value in `corpus-encrypted.json` but not `corpus.json`. Since this isn't used for comparison (and the Java driver was using the `allowed` from `corpus.json`) this didn't show up in tests, but showed up when I regenerated.

jyemin: Ack.

source/client-side-encryption/tests/README.rst
```diff
@@ -473,10 +476,10 @@ The corpus test exhaustively enumerates all ways to encrypt all BSON value types
 
    Iterate over each field of ``corpus_encrypted_expected`` and check the following:
 
-   - If the ``algo`` is ``det``, that the value exactly matches all fields in ``corpus_encrypted_actual`` with the same ``kms``, ``type``, and ``algo``.
-   - If the ``algo`` is ``rand`` and ``allowed`` is true, that the value does not match any fields in ``corpus_encrypted_actual`` with the same ``kms`` and ``type``.
-   - If the ``method`` is ``auto`` or ``explicit``, decrypt the value with ``client_encryption`` and validate the value exactly matches the corresponding field of ``corpus``.
-   - If the ``allowed`` is false, validate the value exactly matches the corresponding field of ``corpus``.
+   - If the ``algo`` is ``det``, that the value exactly the value of the corresponding field in ``corpus_encrypted_actual``.
```
jyemin: Ungrammatical sentence.  Did you mean: "that the value equals the value of ...."?

source/client-side-encryption/tests/README.rst
```diff
@@ -473,10 +476,10 @@ The corpus test exhaustively enumerates all ways to encrypt all BSON value types
...
+   - If the ``algo`` is ``rand`` and ``allowed`` is true, that the value does not the value of the corresponding field in ``corpus_encrypted_actual``.
```
jyemin: "does not equal"?

source/client-side-encryption/tests/README.rst
```diff
@@ -473,10 +476,10 @@ The corpus test exhaustively enumerates all ways to encrypt all BSON value types
...
+   - If ``allowed`` is true, decrypt the value with ``client_encryption``. Decrypt the corresponding field of ``corpus_encrypted`` and validate that they both match.
+   - If ``allowed`` is false, validate the value exactly matches the corresponding field of ``corpus`` (neither was encrypted).
```
jyemin: "matches/equals the value of the corresponding field"?

source/client-side-encryption/tests/README.rst
kevinAlbs: Oops, fixed.

source/client-side-encryption/tests/README.rst
kevinAlbs: Fixed.

source/client-side-encryption/tests/README.rst
kevinAlbs: Fixed.

kevinAlbs:
@markbenvenuto would you mind verifying I'm testing all such padding lengths by testing strings of sizes 0 to 16 bytes?

markbenvenuto:
Yeah, by testing from [1-16], you are testing all the padding cases. The 0 case is sort of redundant from a crypto case but it is a fine test case in general.

PR #621, opened 2019-07-30
Clarify when the cached recoveryToken should be cleared

ShaneHarvey:


source/transactions/transactions.rst
```diff
@@ -838,6 +838,9 @@ commitTransaction or abortTransaction commands. Tracking the most recently
 returned ``recoveryToken`` allows the server to update the ``recoveryToken``
 mid-transaction if needed.
 
+Drivers MUST clear a session's cached ``recoveryToken`` when transitioning to
```
jyemin: Can we test this, and if not, is it worth specifying?

ShaneHarvey: We can test it by adding a new special `testRunner` operation like `assertRecoveryToken` but I'm not sure it's worth it. I can't think of any cases where not clearing the recoveryToken would cause a bug. Maybe if the next transaction didn't return a recoveryToken in the response? 

AFAIK that could only happen if there was a server bug or the cluster changed from a (>=4.2) mongos to a (>=4.0) replica set.

p-mongo: In principle I think more tests won't hurt but whatever y'all decide is fine with me.

jyemin: I'll support this if we have a YAML or prose test to assert the behavior. If it's not worth testing, it's not worth specifying.

ShaneHarvey: I’ll add tests because I think it reasonably possible that mistakenly not clearing the token could cause other issues.

ShaneHarvey:
I opened https://jira.mongodb.org/browse/SPEC-1402 to add the requested tests.

PR #622, opened 2019-07-31
SPEC-1363 remove default upsert/multi

kevinAlbs:


source/client-side-encryption/etc/generate-test.py
```diff
@@ -452,21 +429,43 @@ def local_provider():
...
-injections = {
-    "schema": schema,
-    "ciphertext": ciphertext,
-    "key": key,
-    "local_provider": local_provider,
-    "schema_w_type": schema_w_type
-}
+if len(sys.argv) < 3:
+    print(description)
+    print("usage: python generate-test.py ./test-templates/<filename>.yml.template [...] <target directory>")
```
kevinAlbs: Some simplifications to make it easier to generate all spec tests at once.

PR #623, opened 2019-08-01
SPEC-1390 Use realistic $date values in encryption spec tests

ShaneHarvey:


ShaneHarvey:
@kevinAlbs I believe that travis is failing because the generate-test.py script writes json files in a different format than js-yaml. How do you feel about only generating the yaml files and letting the makefile create the json files instead?

kevinAlbs:
Huh, I don't recall how I produced that value. Maybe I mistook microseconds for milliseconds. I agree about updating generate-test.py, created https://github.com/mongodb/specifications/pull/624

PR #624, opened 2019-08-01
only generate YML tests from generate-test.py

kevinAlbs:


ShaneHarvey:
Can you also delete all the existing json files and regenerate they with the makefile? I think it will produce a large diff. Might as well do that in this PR too.

kevinAlbs:
Good idea, done.

PR #626, opened 2019-08-02
SPEC-1392 Make bson-corpus array test descriptions unique

### LINKED ISSUE #625: Two tests with same name in same suite.
These two tests in array.json in the bson corpus have the same name. When autogenerating tests, some test frameworks will fail because two tests in the same suite cannot have the same name. The names should be changed to reflect more precisely what is meant by "set incorrectly."

https://github.com/mongodb/specifications/blob/672a9f2a9dbc0e1dd2ca2a0d0b4599653d82a683/source/bson-corpus/tests/array.json#L17
https://github.com/mongodb/specifications/blob/672a9f2a9dbc0e1dd2ca2a0d0b4599653d82a683/source/bson-corpus/tests/array.json#L23

ShaneHarvey:
Fixes #625.

For reference this is the first test:
```
$ echo "130000000461000B00000010000A0000000000" | perl bsonview -x
 13000000 04 "a" 00 [ 0b000000 10 "" 00 0A000000 00 ] 00
```
And this is the second:
```
$ echo "150000000461000D000000106162000A0000000000" | perl bsonview -x
 15000000 04 "a" 00 [ 0d000000 10 "ab" 00 0A000000 00 ] 00
```

### DIFF
--- a/source/bson-corpus/tests/array.json
+++ b/source/bson-corpus/tests/array.json
@@ -14,13 +14,13 @@
             "canonical_extjson": "{\"a\" : [{\"$numberInt\": \"10\"}]}"
         },
         {
-            "description": "Single Element Array with index set incorrectly",
+            "description": "Single Element Array with index set incorrectly to empty string",
             "degenerate_bson": "130000000461000B00000010000A0000000000",
             "canonical_bson": "140000000461000C0000001030000A0000000000",
             "canonical_extjson": "{\"a\" : [{\"$numberInt\": \"10\"}]}"
         },
         {
-            "description": "Single Element Array with index set incorrectly",
+            "description": "Single Element Array with index set incorrectly to ab",
             "degenerate_bson": "150000000461000D000000106162000A0000000000",
             "canonical_bson": "140000000461000C0000001030000A0000000000",
             "canonical_extjson": "{\"a\" : [{\"$numberInt\": \"10\"}]}"
--- a/source/bson-corpus/tests/array.json
+++ b/source/bson-corpus/tests/array.json
@@ -24,6 +24,12 @@
             "degenerate_bson": "150000000461000D000000106162000A0000000000",
             "canonical_bson": "140000000461000C0000001030000A0000000000",
             "canonical_extjson": "{\"a\" : [{\"$numberInt\": \"10\"}]}"
+        },
+        {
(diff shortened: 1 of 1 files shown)


gatorcse:
Woah, so even though the spec says that the keys need to be numbers, we still need to handle the case when they aren't numbers as being valid bson? What are the rules for ordering then?

ShaneHarvey:
Unfortunately yes you still need to handle invalid keys when decoding bson. For example, the python driver [completely ignores the values of array keys](https://github.com/mongodb/mongo-python-driver/blob/3.8.0/bson/__init__.py#L238-L239). However, drivers _are_ required to properly generate the keys while encoding bson arrays.

This decision was reached because historically some drivers/applications have generated invalid 
 keys and unfortunated the server does (or did) not reject those "invalid" arrays. To maintain compatibility between languages we decided to simply ignore the keys while decoding. BSON documents are ordered so the order is implicitly know. Again feel free to examine the python driver as a reference implementation of the spec.

I you want to dive more deeply into the issues that motivated these tests, please see https://jira.mongodb.org/browse/SERVER-16814 and the linked tickets.

gatorcse:
Ok, so when _encoding_, we need to assign keys as ascending integers (and have the elements in order), but when _decoding_, we completely ignore the values of the keys? So if all of the keys were integers, but were out of order, we would order the array in the order that the elements were in the document, _not_ order by the integer values of the keys?

ShaneHarvey:
> Ok, so when encoding, we need to assign keys as ascending integers (and have the elements in order), but when decoding, we completely ignore the values of the keys? 

Yes.

> So if all of the keys were integers, but were out of order, we would order the array in the order that the elements were in the document, not order by the integer values of the keys?

Exactly.

gatorcse:
@ShaneHarvey Thanks for clarifying!

ShaneHarvey:
No problem. Your question made me realize that we don't have any tests for duplicate array indexes. I added this test where both keys are "0":
```
 $ echo 1b000000046100130000001030000a000000103000140000000000 | perl bsonview -x
 1b000000 04 "a" 00 [ 13000000 10 "0" 00 0A000000 10 "0" 00 14000000 00 ] 00
```

gatorcse:
Could probably also have a case for out of order integer keys ignoring the order.

PR #627, opened 2019-08-02
SPEC-1393 Connections survive primary step-down tests should always test for success

prashantmital:
Closes [SPEC-1393](https://jira.mongodb.org/browse/SPEC-1393)

### DIFF
--- a/source/connections-survive-step-down/tests/README.rst
+++ b/source/connections-survive-step-down/tests/README.rst
@@ -162,13 +162,13 @@ Questions and Answers
 ---------------------
 
 Do we need to wait for re-election after the first test?
-``````````````````````````````````````````````````````````
+````````````````````````````````````````````````````````
 
 Since test setup requires creation of a collection, a primary must exist, so subsequent tests will block in server selection until a primary is available again.
 
 
 Why do tests check for a successful insert operation in addition to checking that the pool was updated appropriately?
-```````````````````````````````````````````````````````````````````````````````````````````````````````````````````````
+`````````````````````````````````````````````````````````````````````````````````````````````````````````````````````
 
 Ensuring that we can run a successful insert after the primary steps down and without needing to recreate the
 ``MongoClient`` serves to test the resiliency of drivers in the event of a failover/election. Even though checking for
--- a/source/connections-survive-step-down/tests/README.rst
+++ b/source/connections-survive-step-down/tests/README.rst
@@ -118,6 +118,8 @@ This test requires a replica set with server version 4.0.
 - Execute an insert into the test collection of a ``{test: 1}``
   document.
 - Verify that the insert failed with an operation failure with 10107 code.
+- Execute an insert into the test collection of a ``{test: 1}``
+  document and verify that it succeeds.
 - Verify that the pool has been cleared following the instructions in section `How to verify the connection pool has been cleared`_
 
 
@@ -133,6 +135,8 @@ Perform the following operations on a client configured to NOT retry writes:
 - Execute an insert into the test collection of a ``{test: 1}``
   document.
 - Verify that the insert failed with an operation failure with 91 code.
+- Execute an insert into the test collection of a ``{test: 1}``
+  document and verify that it succeeds.
 - Verify that the pool has been cleared following the instructions in section `How to verify the connection pool has been cleared`_
 
 
@@ -148,6 +152,8 @@ Perform the following operations on a client configured to NOT retry writes:
 - Execute an insert into the test collection of a ``{test: 1}``
   document.
 - Verify that the insert failed with an operation failure with 11600 code.
+- Execute an insert into the test collection of a ``{test: 1}``
+  document and verify that it succeeds.
 - Verify that the pool has been cleared following the instructions in section `How to verify the connection pool has been cleared`_
 
 
--- a/source/connections-survive-step-down/tests/README.rst
+++ b/source/connections-survive-step-down/tests/README.rst
@@ -118,9 +118,9 @@ This test requires a replica set with server version 4.0.
 - Execute an insert into the test collection of a ``{test: 1}``
   document.
 - Verify that the insert failed with an operation failure with 10107 code.
+- Verify that the pool has been cleared following the instructions in section `How to verify the connection pool has been cleared`_
 - Execute an insert into the test collection of a ``{test: 1}``
   document and verify that it succeeds.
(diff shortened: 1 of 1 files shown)


p-mongo:
What is the purpose of this change exactly? The tests in question cover what happens when the respective operations fail, not what happens when other operations succeed.

p-mongo:
Also, now the pool could have been cleared as  a result of one of two operations, which makes the tests prone to false positives.

ShaneHarvey:
The purpose of this addition is to make sure that the next operation after a shutdown error will succeed. That makes the test better simulate real application behavior imo. The next operation should rediscover the primary and succeed.

I agree with Oleg about the false positives. To fix them I think the extra insert should happen _after_ verifying the pool has been cleared.

prashantmital:
@p-mongo the rationale is what @ShaneHarvey mentioned. I have also updated the tests to run the additional insert _after_ we check the CMAP Listener for pool cleared/not cleared events.

p-mongo:
OK, I think I understand what you are going for but it's not the purpose of the tests in question. Per the readme, the tests are for:

> These tests can be used to verify a driver's compliance with server discovery and monitoring requirements with respect to handling "not master" and "node is shutting down" error responses from the server.

ShaneHarvey:
I think it's these changes are exactly in line with the purpose of these tests; the extra insert verifies the expected SDAM behavior after a shutdown error. Besides, we already do the same "insert succeeds" check in the "Not Master - Keep Connection Pool" test. This change makes the tests more consistent and robust by performing the check for all the tests.

jmikola:
#### On adding post-clear inserts

It seems like everyone agrees a driver _should_ be able to successfully execute an insert after the pool is cleared, so I think it's just a question of how we can make this more relevant to this spec. What about adding an additional assertion after the successful insert to show that one new connection has been created?

Even though the [Live Failover Testing spec](https://jira.mongodb.org/browse/WRITING-3601) will ultimately test similar scenarios, it will likely do so using Atlas' automation agent instead of these fail points. This would be a unique opportunity to cover these cases in a different manner, given that the existing SDAM spec tests use [mocking with no actual server communication](https://github.com/mongodb/specifications/tree/master/source/server-discovery-and-monitoring/tests#mocking).

Side note: technically, "connections survive primary step-down" is only a 4.2+ feature, so could one also argue that all spec tests for pre-4.2 servers are out of scope? Or is this definitely in scope because of the following quote from "updateRSFromPrimary" in [SDAM: Actions](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#actions):

> If the old primary server version is 4.0 or earlier, the client MUST clear its connection pool for the old primary, too: the connections are all bad because the old primary has closed its sockets. If the old primary server version is 4.2 or newer, the client MUST NOT clear its connection pool for the old primary.

Other text around that paragraph (both in "updateRSFromPrimary" and "checkIfHasPrimary" that follows) seems to talk about how to handle the case of disconnecting and immediately re-checking a primary:

> A note on invalidating the old primary: when a new primary is discovered, the client finds the previous primary (there should be none or one) and replaces its description with a default ServerDescription of type "Unknown." A multi-threaded client MUST check that server as soon as possible. [JM: single-threaded will likely do so upon the next server selection attempt]

#### Question about an existing test

Reading the test in [Not Master - Keep Connection Pool](https://github.com/mongodb/specifications/tree/master/source/connections-survive-step-down/tests#not-master-keep-connection-pool), would it make more sense to verify that the pool hasn't been cleared after the failed insert (due to the fail point) but _before_ the successful insert that follows? This may make no difference if drivers are asserting based on CMAP events, since a PoolCleared event will either have been observed or not, but it could have an impact on drivers relying on `connections.totalCreated` from the `serverStatus` command, since a pool could be cleared and then repopulated for the successful insert. Additionally, I wonder if "using the same client that was used for test operation" for `serverStatus` might be incorrect, as merely executing the `serverStatus` command could trigger SDAM to repopulate a previously cleared pool.

p-mongo:
@jmikola 

> Side note: technically, "connections survive primary step-down" is only a 4.2+ feature, so could one also argue that all spec tests for pre-4.2 servers are out of scope?

Since the spec changes mandated both pre-4.2 and 4.2+ behavior, the tests cover both. On pre-4.2 servers the drivers still must close connections.

>  would it make more sense to verify that the pool hasn't been cleared after the failed insert (due to the fail point) but before the successful insert that follows

The idea is to send the insert over the same connection established prior to the failing insert. Conceivably the failing insert may leave the existing connection alone but inserting again would make a new connection, which would be contrary to the goal of the resiliency.

I suppose assertions can be made after the failing insert and after the successful insert, if this helps to clarify things.

p-mongo:
@prashantmital Could you add the following rationales to the Q&A section please:

> @oleg’s contention is (and rightly so) that the successful insert doesn’t make sense for the other scenarios since the spec test should be limited to verifying the implementation of the spec itself. Since the pool is cleared in other scenarios, verifying as much is sufficient to test that the spec has been satisfactorily implemented.
> On the other hand, @shane.harvey and I are of the opinion that these tests should also test the resiliency of the driver in the event of a primary stepping down. Simply ensuring that we are able to perform a successful write after the stepdown will help verify this without impacting what is already being tested.

> Even though the Live Failover Testing spec will ultimately test similar scenarios, it will likely do so using Atlas' automation agent instead of these fail points. This would be a unique opportunity to cover these cases in a different manner, given that the existing SDAM spec tests use mocking with no actual server communication.

prashantmital:
@p @jmikola please re-review. I have added a Q&A entry with the requested rationale.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -159,3 +165,14 @@ Do we need to wait for re-election after the first test?
 ``````````````````````````````````````````````````````````
 
 Since test setup requires creation of a collection, a primary must exist, so subsequent tests will block in server selection until a primary is available again.
+
+
+Why do tests check for a successful insert operation in addition to checking that the pool was updated appropriately?
+```````````````````````````````````````````````````````````````````````````````````````````````````````````````````````
```
jmikola: I think two backticks need to be trimmed here. GitHub may render it, but it's technically an RST warning if the text and header underline don't match up.

prashantmital: Done. Also did the same for the existing Q&A item (it too had 2 extra backticks).

PR #629, opened 2019-08-07
FLE prose test fixes

kevinAlbs:


source/client-side-encryption/external/external-schema.json
```diff
@@ -14,5 +14,6 @@
                "algorithm": "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic"
            }
        }
-    }
-}
\ No newline at end of file
+    },
+    "bsonType": "object"
```
jyemin: I don't think the Java driver implements the test that uses this.  Which one is it?

kevinAlbs: "External Key Vault Test"

I validated this with a PoC here: https://github.com/kevinAlbs/JavaFLEProsePoC/blob/master/src/main/java/ExternalKeyVaultTest.java

The PoC does the right thing. It uses this schema and encrypts "encrypted". I must not have copied it back over.

PR #630, opened 2019-08-07
SPEC-1383: add prose test description

jstewart-mongo:


source/retryable-writes/tests/README.rst
```diff
@@ -296,6 +296,41 @@ supported write operations:
...
+Prose Tests
+===========
+
+The following tests ensure that retryable writes work properly with replica sets
+and sharded clusters.
+
+#. Test that retryable writes raise an exception when using the MMAPv1 storage engine,
+   available in both MongoDB 3.6 and 4.0.
+
+   .. code:: java
```
jyemin: I think you should write this in prose rather than Java.  Especially the assertions are going to look very different in other languages (which don't all have exception chaining).

jstewart-mongo: Converted the Java code to prose.

source/retryable-writes/tests/README.rst
```diff
@@ -296,6 +296,23 @@ supported write operations:
...
+#. Test that retryable writes raise an exception when using the MMAPv1 storage engine,
+   available in both MongoDB 3.6 and 4.0. For this test, execute a write
+   operation, such as ``insertOne``, which should generate an
+   ``IllegalOperation`` exception. Assert that the error message is the
+   replacement error message::
+
+  This MongoDB deployment does not support retryable writes. Please add
+  retryWrites=false to your connection string.
+
+   and the error code is 20.
```
ShaneHarvey: This formatting is off here (https://github.com/mongodb/specifications/blob/7dcc969631d77d01a4a09530f2646748414595b0/source/retryable-writes/tests/README.rst#prose-tests).

jstewart-mongo: Fixed

source/retryable-writes/tests/README.rst
```diff
@@ -296,9 +296,28 @@ supported write operations:
   - ``bulkWrite()`` with ``ordered=true`` (no ``UpdateMany`` or ``DeleteMany``)
   - ``bulkWrite()`` with ``ordered=false`` (no ``UpdateMany`` or ``DeleteMany``)
 
+Prose Tests
+===========
+
+The following tests ensure that retryable writes work properly with replica sets
+and sharded clusters.
+
+#. Test that retryable writes raise an exception when using the MMAPv1 storage engine,
```
prashantmital: Let's use a hardcoded test number here instead of an autogenerated one. See https://jira.mongodb.org/browse/SPEC-1364 for the rationale.

prashantmital: Suggest rephrasing this for clarity:

```
Test that attempting a write operation with retryable writes enabled raises an exception when using the MMAPv1 storage engine (available on MongoDB <= 4.0). Start by initializing a client with ``retryWrites=True`` and attempt a write operation (e.g. ``insertOne``, ``insertMany``). This should generate an exception with the error code set to 20, and the following error message::

...
```

The current language makes it seem like the MMAPv1 storage engine is only available on 3.6 and 4.0 which is inaccurate. I am also not sure if it makes sense to specify the type of the exception (``IllegalOperation``) as not all drivers have a 1:1 correspondence between exceptions that they raise vs. those that are raised by the server (PyMongo, for instance, raises an `OperationFailure` exception in this case). However, since the pseudocode pertaining to this change does explicitly mention this exception, I would be okay if we also used it here.

jstewart-mongo: Done

jstewart-mongo: I have removed the `IllegalOperation` reference and reworded the prose.

PR #632, opened 2019-08-12
SPEC-1411 Avoid using readConcern snapshot in sharded txn tests

ShaneHarvey:
Passing patch (note the failures are due to an unrelated issue [PYTHON-1935](https://jira.mongodb.org/browse/PYTHON-1935)):  https://evergreen.mongodb.com/version/5d519ffa1e2d174dcb14bcef

### DIFF
--- a/source/transactions/tests/transaction-options-repl.yml
+++ b/source/transactions/tests/transaction-options-repl.yml
@@ -1,0 +1,117 @@
+runOn:
+  - minServerVersion: "4.0"
+    topology: ["replicaset"]
+
+database_name: &database_name "transaction-tests"
+collection_name: &collection_name "test"
+
+data: []
+
+tests:
+
+  - description: readConcern snapshot in startTransaction options
+
+    sessionOptions:
+      session0:
+        defaultTransactionOptions:
+          readConcern:
+            level: majority # Overridden.
+
+    operations:
+      - name: startTransaction
+        object: session0
+        arguments:
+          options:
+            readConcern:
+              level: snapshot
+      - name: insertOne
+        object: collection
+        arguments:
+          session: session0
+          document:
+            _id: 1
+        result:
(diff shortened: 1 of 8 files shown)


source/transactions/tests/transaction-options-repl.yml
```diff
@@ -0,0 +1,117 @@
...
+    topology: ["replicaset"]
+
+database_name: &database_name "transaction-tests"
+collection_name: &collection_name "test"
+
+data: []
+
+tests:
+
+  - description: readConcern snapshot in startTransaction options
```
ShaneHarvey: This is the only test that _needs_ to send 'snapshot' to the server so I've left it unchanged and simply skipped it on sharded clusters (`runOn` requires this test to only run on a replica set).

ShaneHarvey:
@jyemin are you able to review this change? If not, can you nominate someone else?

jyemin:
Is the sending of a distinct command (https://jira.mongodb.org/browse/SERVER-39704?focusedCommentId=2170328&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2170328) to all mongos servers an insufficient workaround?

ShaneHarvey:
Yeah, python sends distinct to all mongos' and still runs into this error on 4.2:
> pymongo.errors.OperationFailure: Transaction 35e8db55-11ae-41b6-9486-d2a1f3a440cb:1 was aborted on statement 0 due to: a non-retryable snapshot error :: caused by :: Encountered error from localhost:27217 during a transaction :: caused by :: Read timestamp Timestamp(1565400200, 35) is older than the oldest available timestamp.

In HELP-10518, Dianna mentions that SERVER-42205 might help avoid these errors so once 4.2.1 is released we can try reverting this change.

PR #634, opened 2019-08-20
Note topology requirements for step down tests

jmikola:
The getMore test requires a replica set because it invokes the `replSetStepDown` command.

I think it also makes sense for the other two "not master" tests to require a replica set, even if the fail point technically makes it possible to simulate the error in other topologies.

I've left the final two shutdown tests alone, as I don't think those codes are specific to replica sets and I imagine those tests could be executed against a sharded cluster or standalone. That said, perhaps we should note thjat sharded cluster URIs should only include a single mongos node since the test calls `configureFailPoint`.


samantharitter:
LGTM as well!

PR #636, opened 2019-08-22
Advise drivers to distinguish retry and resume attempts

jmikola:


jyemin:
What's the motivation behind this?  Is there a SPEC ticket that explains it?

jmikola:
> What's the motivation behind this? Is there a SPEC ticket that explains it?

In https://github.com/mongodb/mongo-php-library/pull/669/files#r316893393, I realized that PHPLIB's implementation of this prose test was using a fail point to trigger a "not master" error and that broke after we integrated retryable reads. I'm not sure how drivers were observing a resume attempt to satisfy this test, but I believe we were just counting `aggregate` commands, which obviously failed.

Presumably it wasn't a problem for anyone else, but I figured the advice might be helpful for a future driver down the line.

alcaeus:
> I'm not sure how drivers were observing a resume attempt to satisfy this test, but I believe we were just counting `aggregate` commands, which obviously failed.

Correct. We use APM to log all commands and check that the `aggregate` command is only run once. With retryable reads turned on, the server will automatically retry the `aggregate`, which our logic misinterpreted as resuming the change stream. There may be other ways to test this, but the note makes sense to help make the distinction between "retry" and 'resume".

PR #640, opened 2019-09-06
SPEC-1440 Make CMAP `PoolClosedEvent` optional

saghm:
This is the culmination of the discussions about CMAP in the Rust driver.

Kevin: Matt suggested I add you given that C++ is another language that fits into the same paradigm as Rust in terms of RAII (even though the C++ driver doesn't implement CMAP due to wrapping the C driver). No need to respond to this until after you're back from vacation.

### DIFF
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -307,6 +307,8 @@ When a pool is closed, it MUST first close all available connections in that poo
     mark pool as CLOSED
     for connection in availableConnections:
       close connection
+
+    # Skip if not implemented
     emit PoolClosedEvent
 
 Creating a Connection (Internal Implementation)
@@ -513,6 +515,11 @@ Events
 
     /**
      *  Emitted when a Connection Pool is closed
+     *
+     *  NOTE: Implementation of this event is OPTIONAL for drivers which implicitly
+     *  close the pool when the object is destroyed, e.g. through destructors or
+     *  RAII. See the "Why is the PoolClosedEvent optional?" section in "Design
+     *	Rationale" for more details.
      */
     interface PoolClosedEvent {
       /**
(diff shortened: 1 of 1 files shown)


source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -699,6 +706,10 @@ Why is waitQueueTimeoutMS optional for some drivers?
 
 We are anticipating eventually introducing a single client-side timeout mechanism, making us hesitant to introduce another granular timeout control. Therefore, if a driver/language already has an idiomatic way to implement their timeouts, they should leverage that mechanism over implementing waitQueueTimeoutMS.
 
+Why is the `PoolClosedEvent` optional?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In languages that make use of destructors/RAII to close the connection pool, it's difficult to emit the event in the same place as other drivers; assuming that all event handling is done directly in the pool rather than being spread across auxiliary types like the connection and wait queue, the connections will need a strong reference to the pool, making it impossible to destroy (and therefore close) the pool until all connections have been returned. Given that drivers emitting events in different orders when the same operations are performed is undesirable, the ``PoolClosedEvent`` should be made optional so that for all event types that any set of drivers all emit, those drivers will emit events of those types in the same order when the same sequence of operations is performed.
```
daprahamian: I would want the motivation to include the following basic points:

1. There is value in non-RAII languages (where an explicit `close` call is needed) to know that the Pool has transitioned to a `closed` state.
2. The value of such an event is low in RAII languages b/c the equivalent of `close` is just dereferencing the pool.
3. We want the connotations of the event to be uniform across all drivers. So if the event proves difficult to implement in an RAII language, it can be omitted.

p-mongo: I agree with @jyemin  in that consumers of events (i.e. applications/monitoring agents) will likely need to track the lifecycle of pools, thus requiring the closed event to release resources allocated for monitoring or to simply end the monitoring.

Reading the proposed language again, I am also wondering about this part:

> the connections will need a strong reference to the pool, making it impossible to destroy (and therefore close) the pool until all connections have been returned.

Since the issue being discussed is whether pool closed event is emitted or not, what is the link between the closed event being emitted and whether connections have strong or weak references to their pools? It seems to me that the changes being discussed should be local to pool only.

saghm:
Based on offline discussion, we won't be making this change. https://jira.mongodb.org/browse/SPEC-1485 will clarify the intentions behind the current spec behavior

PR #642, opened 2019-09-09
Fix SPEC-1362 Do not expect upsert/multi fields with default values in transaction tests


p-mongo:


source/crud/examples/php/usage_example.php
```diff
@@ -234,7 +234,7 @@
             // Required update param
             ['$inc' => ['x' => 2]],
             // Optional named params in an associative array
-            ['upsert' => false],
+            ['upsert' => true],
```
jmikola: It occurred to me that these example files are outdated and should probably be deleted: https://jira.mongodb.org/browse/SPEC-1444

In any event, I think you can omit this change from the PR.

p-mongo:
> I take it the test changes were done by searching for multi: false and upsert: false in YAML, deleting those lines, and then regenerating the JSON from YAML. Correct?

That is correct.

source/crud/examples/php/usage_example.php
p-mongo: Reverted this change.

p-mongo:
https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst says for `upsert`:

>    * This option is sent only if the caller explicitly provides a value. The default is to not send a value.

A driver which sends `null` is in violation of this sections as I understand it.

jmikola:
> My guess is whatever driver originally added these tests sends out these values, which is why they show up in the expectations. This will require that driver to stop sending the values now, causing side effect work for them.

@mbroadst: I may be misunderstanding your comment, but removing these fields from the expected `command_started_event` shouldn't break anything for drivers that currently _do_ send these fields. The APM spec's rules for [additional value assertions](https://github.com/mongodb/specifications/blob/master/source/command-monitoring/tests/README.rst#additional-values) are clear that actual commands may contain extra fields and that those should be ignored.

PR #644, opened 2019-09-13
SPEC-1434: Clarify ignoring configureFailPoint command in spec tests

alcaeus:
This clarifies that drivers should either run the `configureFailPoint` command used for the `targetedFailPoint` operation on a different client, or need to filter commands in the logger as the tests don't include `configureFailPoint` in the command expectations.

### DIFF
--- a/source/transactions/tests/README.rst
+++ b/source/transactions/tests/README.rst
@@ -342,6 +342,12 @@ fail point on the mongos server which "session0" is pinned to::
               failCommands: ["commitTransaction"]
               closeConnection: true
 
+Tests that use the "targetedFailPoint" operation do not include
+``configureFailPoint`` commands in their command expectations. Drivers MUST
+ensure that ``configureFailPoint`` commands do not appear in the list of logged
+commands, either by manually filtering it from the list of observed commands or
+by using a different MongoClient to execute ``configureFailPoint``.
+
 assertSessionTransactionState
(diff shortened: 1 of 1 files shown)


source/transactions/tests/README.rst
```diff
@@ -342,6 +342,11 @@ fail point on the mongos server which "session0" is pinned to::
               failCommands: ["commitTransaction"]
               closeConnection: true
 
+Drivers need to ensure to not log ``configureFailPoint`` commands in command
+loggers, either by filtering logged commands, or by using a different client to
+create fail points. Tests using the "targetedFailPoint" operation don't include
+the ``configureFailPoint`` in command expectations.
```
jmikola: I feel like the last sentence explains the justification here, and would propose flipping their order. Also, this might be a good place to use stronger spec language. Consider:

> Tests that use the "targetedFailPoint" operation do not include ``configureFailPoint`` commands in their command expectations. Drivers MUST ensure that ``configureFailPoint`` commands do not appear in the list of logged commands, either by manually filtering it from the list of observed commands or by using a different MongoClient to execute ``configureFailPoint``.

I also tweaked the sentence to say "do not appear in the list of logged commands", as ensuring "to not log" seemed contrary to the option of filtering it from the the list of logged commands.

alcaeus: Good call on the `MUST` to make this more explicit. I've updated the wording to the one you suggested. Thanks!

PR #645, opened 2019-09-13
SPEC-1332 Clarify that connection pools should only be created for da…

p-mongo:
…ta bearing servers

### DIFF
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -284,10 +284,9 @@ Creating a Connection Pool
 This specification does not define how a pool is to be created, leaving it
 up to the driver. Creation of a connection pool is generally an implementation
 detail of the driver, i.e., is not a part of the public API of the driver.
-
-The driver SHOULD only create connection pools for servers it has determined
-to be `data-bearing
-<https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#data-bearing-server-type>`_.
+The SDAM specification defines `when
+<https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#connection-pool-creation>`_
+the driver should create connection pools.
 
 Once a pool is created, if minPoolSize is set, the pool MUST immediately
 begin creating enough connections such that totalConnections >= minPoolSize.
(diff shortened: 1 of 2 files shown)


source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -281,7 +281,19 @@ Connection Pool Behaviors
 Creating a Connection Pool
 --------------------------
 
-Pool creation is mostly an implementation detail specific to the driver language. If minPoolSize is set, the pool MUST immediately create enough connections such that totalConnections >= minPoolSize. These connections MUST be created in a non-blocking manner. Drivers SHOULD additionally connect these connections if it is possible to do so in a non-blocking manner, either via the use of a Background Thread or via asynchronous I/O.
+This specification does not define how a pool is to be created, leaving it
+up to the driver. Creation of a connection pool is generally an implementation
+detail of the driver, i.e., is not a part of the public API of the driver.
+
+The driver SHOULD only create connection pools for servers it has determined
```
mbroadst: Does this belong in this spec? In my mind the CMAP spec describes a generic pool and its behaviors, not what types of server should have pools made for them. If we agree with this behavior, I think it should live in the SDAM spec

p-mongo: Moved this section to the sdam spec.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1292,6 +1292,16 @@ See the Driver Sessions Spec for the purpose of this value.
...
 
+
+Connection Pool Creation
+''''''''''''''''''''''''
+
+After a server check is complete, if the server is determined to be
+`data-bearing
+<https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#data-bearing-server-type>`_,
+if the server in question does not already have a connection pool,
+the driver SHOULD create a connection pool for the server.
```
daprahamian: Is this a SHOULD b/c not all drivers use connection pools?

I feel like the workflow we are going for is saying:

- If:
    - The driver supports connection pooling AND
    - The server is data-bearing
- Then:
    - By this point, the driver MUST have a connection pool for the server

p-mongo: No reason, I can change to your wording if you like.

daprahamian: Yeah, a wording change would be good.

mbroadst: This still reads awkwardly to me, how about: "For drivers that support connection pools: after a server check is complete, if the server is data-bearing and it does not already have a connection pool, the driver MUST create one for it"

p-mongo: Changed to suggested wording.

PR #647, opened 2019-09-13
SPEC-1365 Make CMAP spec tests consistently include addresses


p-mongo:


jmikola:
@p-mongo: Removing myself, as I have no working knowledge of the CMAP spec. @daprahamian seems like a preferable choice for second reviewer.

PR #648, opened 2019-09-20
Fix SPEC-1453 Change stream YAML spec tests incorrectly specify  key


p-mongo:


source/change-streams/tests/change-streams-errors.yml
```diff
@@ -42,7 +42,7 @@ tests:
             cursor: {}
             pipeline:
               - 
-                $changeStream:
+                $changeStream: {}
```
jmikola: IIRC, all `$changeStream` documents should at least have a `fullDocument` option, since the spec doesn't say that default behavior is to not send a value (as is often done in other specs for options). See: https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#driver-api

That said, I suppose an empty object here will just allow [additional values to be ignored](https://github.com/mongodb/specifications/blob/master/source/command-monitoring/tests/README.rst#additional-values)? Was `null` actually a problem because a test harness might expect a literal BSON null for `$changeStream`?

p-mongo: Per https://jira.mongodb.org/browse/SPEC-1350, as well as the current spec languages as far as I can see, fullDocument is optional. 31edce69d0fed2151a08c66204b7b60afcd6ab9b made this change but the yaml files in that commit were not changed correctly.

PR #649, opened 2019-09-23
SPEC-1255 convert crud v1 tests to use runOn format

p-mongo:


source/crud/tests/README.rst
```diff
@@ -149,13 +149,6 @@ single operation. Notable differences from the current format are as follows:
   defined under the ``tests[i].outcome.error`` and ``tests[i].outcome.result``
   fields.
 
-- Instead of a top-level ``runOn`` field, server requirements are denoted by
-  separate top-level ``minServerVersion`` and ``maxServerVersion`` fields. The
-  minimum server version is an inclusive lower bound for running the test. The
-  maximum server version is an exclusive upper bound for running the test. If a
-  field is not present, it should be assumed that there is no corresponding bound
-  on the required server version.
```
jmikola: @p-mongo: Considering [SPEC-1255](https://jira.mongodb.org/browse/SPEC-1255) has already been closed and [DRIVERS-679](https://jira.mongodb.org/browse/DRIVERS-679) and various language tickets are in progress, I don't think this PR should re-use the same issue. For better or worse, SPEC-1255 only addressed a subset of tests, which are now in the "v2" directory.

I think this warrants its own ticket; however, much like [SPEC-1193](https://jira.mongodb.org/browse/SPEC-1193) (to migrate change streams to use `runOn`), I think this task should ultimately wait on the unified test runner spec ([WRITING-4393](https://jira.mongodb.org/browse/WRITING-4393)) being done in the 4.4 timeframe. I expect that project will update all CRUD tests (among other specs) to a common format.

A separate concern with _only_ updating the CRUD's "legacy test format" to use `runOn` with this PR is that we're creating more work for drivers in advance of the unified test runner changes. This change would only apply to the "v1" CRUD test runner, and be made obsolete by any future effort to further modernize the "v1" tests.

If you'd like an issue to track this, I'd suggest creating a ticket in the same vein as [SPEC-1193](https://jira.mongodb.org/browse/SPEC-1193) and linking it against the WRITING ticket for the time being. That will make sure the folks working on the unified test format spec address this.

p-mongo:
OK I understand.

PR #650, opened 2019-09-26
SPEC-1398: Hint option for findAndModify update/replace operations

jmikola:
https://jira.mongodb.org/browse/SPEC-1398

This is related to #596 but I've split it out into its own PR since `findAndModify` support for `hint` is still waiting on [SERVER-42099](https://jira.mongodb.org/browse/SERVER-42099).

### DIFF
--- a/source/crud/crud.rst
+++ b/source/crud/crud.rst
@@ -12,7 +12,7 @@ Driver CRUD API
 :Status: Approved
 :Type: Standards
 :Minimum Server Version: 2.6
-:Last Modified: January 17, 2020
+:Last Modified: January 24, 2020
 
 .. contents::
 
@@ -1602,6 +1602,17 @@ Find And Modify
     collation: Optional<Document>;
 
     /**
+     * The index to use.
+     *
+     * This option is sent only if the caller explicitly provides a value. The default is to not send a value.
+     * This option is only supported by servers >= 4.4. Older servers >= 4.2 will report an error for using this option.
+     * For servers < 4.2, the driver MUST raise an error if the caller explicitly provides a value.
+     *
+     * @see https://docs.mongodb.com/manual/reference/command/findAndModify/
+     */
+    hint: Optional<(String | Document)>;
+
+    /**
      * The maximum amount of time to allow the query to run.
      *
      * This option is sent only if the caller explicitly provides a value. The default is to not send a value.
@@ -1681,6 +1692,17 @@ Find And Modify
      * @see https://docs.mongodb.com/manual/reference/command/findAndModify/
      */
     collation: Optional<Document>;
+
+    /**
+     * The index to use.
+     *
+     * This option is sent only if the caller explicitly provides a value. The default is to not send a value.
+     * This option is only supported by servers >= 4.4. Older servers >= 4.2 will report an error for using this option.
+     * For servers < 4.2, the driver MUST raise an error if the caller explicitly provides a value.
+     *
+     * @see https://docs.mongodb.com/manual/reference/command/findAndModify/
+     */
+    hint: Optional<(String | Document)>;
 
     /**
      * The maximum amount of time to allow the query to run.
@@ -1837,6 +1859,7 @@ Q: Why are client-side errors raised for some unsupported options?
 Changes
 =======
 
+* 2020-01-24: Added hint option for findAndModify update/replace operations.
 * 2020-01-17: Add allowDiskUse to FindOptions.
 * 2020-01-14: Deprecate oplogReplay option for find command
 * 2020-01-10: Clarify client-side error reporting for unsupported options
--- a/source/crud/tests/v2/findOneAndReplace-hint-clientError.json
+++ b/source/crud/tests/v2/findOneAndReplace-hint-clientError.json
@@ -1,0 +1,90 @@
+{
+  "runOn": [
+    {
+      "maxServerVersion": "4.0.99"
+    }
+  ],
+  "data": [
+    {
+      "_id": 1,
+      "x": 11
+    },
+    {
+      "_id": 2,
+      "x": 22
+    }
+  ],
+  "collection_name": "findOneAndReplace_hint",
+  "tests": [
+    {
+      "description": "FindOneAndReplace with hint string unsupported (client-side error)",
+      "operations": [
+        {
+          "object": "collection",
+          "name": "findOneAndReplace",
+          "arguments": {
+            "filter": {
+              "_id": 1
+            },
+            "replacement": {
+              "x": 33
+            },
+            "hint": "_id_"
+          },
+          "error": true
+        }
+      ],
+      "expectations": [],
+      "outcome": {
+        "collection": {
+          "data": [
+            {
+              "_id": 1,
+              "x": 11
+            },
+            {
+              "_id": 2,
+              "x": 22
+            }
+          ]
+        }
+      }
+    },
+    {
+      "description": "FindOneAndReplace with hint document unsupported (client-side error)",
+      "operations": [
+        {
+          "object": "collection",
+          "name": "findOneAndReplace",
+          "arguments": {
+            "filter": {
+              "_id": 1
+            },
+            "replacement": {
+              "x": 33
+            },
+            "hint": {
+              "_id": 1
+            }
+          },
+          "error": true
+        }
+      ],
+      "expectations": [],
+      "outcome": {
+        "collection": {
+          "data": [
+            {
+              "_id": 1,
+              "x": 11
+            },
+            {
+              "_id": 2,
+              "x": 22
+            }
+          ]
+        }
+      }
+    }
+  ]
+}
--- a/source/crud/tests/v2/findOneAndReplace-hint-clientError.yml
+++ b/source/crud/tests/v2/findOneAndReplace-hint-clientError.yml
@@ -1,0 +1,40 @@
+runOn:
+  - { maxServerVersion: "4.0.99" }
+
+data:
+  - { _id: 1, x: 11 }
+  - { _id: 2, x: 22 }
+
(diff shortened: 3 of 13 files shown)


source/crud/crud.rst
```diff
@@ -1475,6 +1475,17 @@ Find And Modify
      */
     collation: Optional<Document>;
 
+    /**
+     * The index to use.
+     *
+     * This option is sent only if the caller explicitly provides a value. The default is to not send a value.
+     * This option is only supported by servers >= 4.4. Older servers >= 4.2 will report an error for using this option.
+     * For servers < 4.2, the driver MUST raise an error if the caller explicitly provides a value.
```
jmikola: Note that this differs from #596. Strict checking of extra options in `findAndModify` was only introduced in 4.1.10 with [SERVER-40005](https://jira.mongodb.org/browse/SERVER-40005).

source/crud/tests/v2/findOneAndDelete-hint.yml
```diff
@@ -0,0 +1,51 @@
+runOn:
+  - { minServerVersion: "4.4.0" }
```
jmikola: The `runOn` requirements for these tests can be updated once [SERVER-42099](https://jira.mongodb.org/browse/SERVER-42099) is merged and released.

jmikola: Updated to 4.3.1.

jmikola:
Server functionality is available since the 4.3.1 release, so this is ready for review.

source/crud/tests/v2/findOneAndDelete-hint-error.yml
```diff
@@ -0,0 +1,52 @@
+runOn:
+  - { maxServerVersion: "4.1.10" }
```
jmikola: > Looks good, but can we add negative tests that prove drivers error when hint is specified when running on older server versions?

@jyemin: This is now done. The three new tests are copies of the original test with `result` replaced with `error: true` and `outcome` changed to assert the original collection data was unchanged. I've used 4.1.10 as a max server version here since that's when I know the server started raising its own error for unrecognized `findAndModify` options ([SERVER-40005](https://jira.mongodb.org/browse/SERVER-40005)), but there's no reason this couldn't be changed to 4.2.0 since drivers will likely use wire versions to determine whether a client-side error should be raised. AFAIK, that's why we don't mention specific patch versions in the spec text, but I like doing so in the tests (it definitely makes sense with min server version to allow testing before stable releases).

jmikola: While reviewing https://github.com/mongodb/specifications/pull/692 I realized that I should have actually used 4.1.9 here since `maxServerVersion` is _inclusive_. Since we're testing for client-side errors, this should allow execution on a server up to, but not including, the version when server-side support was introduced. I'll make the change.

jmikola:
Rebased on master to resolve conflicts. This is once again ready for review.

source/crud/crud.rst
ShaneHarvey: It looks like there are only tests for servers <4.2 or >=4.4. Should we add a test against 4.2 to make sure that the driver does not raise a client side error (via a command started event)?

source/crud/tests/v2/findOneAndDelete-hint-error.yml
```diff
@@ -0,0 +1,53 @@
...
+      -
+        object: collection
+        name: findOneAndDelete
+        arguments:
+          filter: &filter { _id: 1 }
+          hint: "_id_"
+        error: true
+    expectations:
+      -
+        command_started_event:
```
ShaneHarvey: According to the spec there should be no command started event: `For servers < 4.2, the driver MUST raise an error if the caller explicitly provides a value`

jmikola: Good catch. I'll strike these.

source/crud/crud.rst
jmikola: I didn't think to do this since it felt like testing server behavior, but you make a good point that the APM expectation would allow us to assert that the driver isn't raising a client-side error.

I should point out that the language here is quite similar to `hint` for `update` (#596), and in the C driver implementation we opted to go beyond "For servers < 3.4" and raise client-side errors for any wire version before the options introduction since it matched what was already done for the `collation` and `arrayFilters` options. I suppose having a similar test as you're proposing for `hint` on `update` would fail in that case, but I'm happy to sort that out when/if such a test is added.

As for this PR, I can add a new test.

source/crud/tests/v2/findOneAndDelete-hint-error.yml
jmikola: FWIW, having an empty or omitted `expectations` field here won't actually test that the driver does not send a command. The [test runner implementation](https://github.com/mongodb/specifications/blob/master/source/crud/tests/README.rst) considers this field optional and only asserts as many events as are found in the array (if any). We don't have a way to assert that **no** events are emitted.

ShaneHarvey: Interesting. Although, in this case the fact that the `findOneAnd*` raised an error _does_ mean the error was client-side because if the driver sent the command the server would have ignored the "hint" option and the operation would have succeeded.

source/crud/crud.rst
ShaneHarvey: I think one minor benefit for sending "hint" to 4.2 is that if the server backports "hint" support to 4.2 then drivers don't need to change.

jmikola: > I think one minor benefit for sending "hint" to 4.2 is that if the server backports "hint" support to 4.2 then drivers don't need to change.

Indeed. That did come up while discussing `update`.

source/crud/tests/v2/findOneAndDelete-hint-clientError.yml
```diff
@@ -0,0 +1,43 @@
+runOn:
+  # These tests assert that the driver raises client-side errors. Server
+  # versions >= 4.1.10 raise errors for unknown findAndModify options
+  # (SERVER-40005), but the spec requires client-side errors for < 4.2. The wire
+  # version for 4.2 was bumped in 4.1.1 (SERVER-35169), so we'll allow up to
+  # 4.1.0 (inclusive).
```
jmikola: @ShaneHarvey: I think this exposes a gap in our test coverage. Server versions >= 4.1.1 will be considered 4.2 based on their wire version, but we can't actually depend on the server to report errors for unsupported `findAndModify` options until 4.1.10. This means that the `hint` option could be silently ignored if the driver is connected to a server version between 4.1.1 and 4.1.9 (inclusive).

Since version checking is typically done by wire protocol version, I don't think there's an clear solution for this short of raising client-side errors for <4.4 servers (which I don't think we want to do). We may have to concede that we can't help users running pre-release versions of the database.

It's possible this applies to other commands as well, but `findAndModify` stands out as strict option validation was introduced fairly recently. As time goes on, there will be less need to worry about this.

jyemin: I don't think we need to care about the odd-numbered releases once the even-numbered release following it has been published.  No one should be testing against any odd-numbered releases except the latest one (currently 4.3.x).  Just use 4.0 and 4.2 as the release numbers.

source/crud/tests/v2/findOneAndDelete-hint-clientError.yml
```diff
@@ -0,0 +1,38 @@
+runOn:
+  - { maxServerVersion: "4.0.99" }
```
jmikola: Updated. There was no change to the serverError tests, but I replaced 4.1.0 here with 4.0.99 and removed the TMI comment. Since `maxServerVersion` is inclusive, specifying a high patch value seemed the most sensible way to signal <4.2.

jmikola:
> Do we need a POC to prove these are correct?

I can attempt a POC in PHPLIB and will follow up here to confirm whether the tests work.

source/crud/tests/v2/findOneAndDelete-hint-serverError.yml
```diff
@@ -0,0 +1,56 @@
+runOn:
+  # These tests assert that the driver does not raise client-side errors and
+  # instead relies on the server to raise an error. Support for findAndModify
+  # hint was added in 4.3.1 (SERVER-42099), so we'll allow up to 4.3.0
+  # (inclusive).
+  - { minServerVersion: "4.2.0", maxServerVersion: "4.3.0" }
```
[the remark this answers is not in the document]
jmikola: On second thought, this really doesn't depend at all on a wire version check in the driver (the spec only addresses wire version checks for < 4.2). In that case, I think it's correct to allow this to run on any server version prior to when we know the feature was actually implemented and the server will no longer raise an error.

jmikola:
I created https://github.com/mongodb/mongo-php-library/pull/711 to POC this in PHPLIB and discovered that `hint` is only supported for update/replace operations. [SERVER-42099](https://jira.mongodb.org/browse/SERVER-42099) has been revised accordingly to clarify this. It looks like [SERVER-44367](https://jira.mongodb.org/browse/SERVER-44367) will add support for both `delete` commands and `findAndModify` where `remove=true` and I've asked for a heads up if that's likely to make it in the server's 4.4 release. Either way, we'll surely address that in a separate SPEC ticket.

As for the POC and the tests in this PR, everything looks good. Of the six tests for each operation (findOneAndReplace and findOneAndUpdate), they all pass across 4.1.10, 4.2.1, and 4.3.2 replica set environments. Each environment passes two tests and skips the other four.

```
4.1.10: ..SSSS..SSSS
4.2.1:  SS..SSSS..SS
4.3.2:  SSSS..SSSS..
```

PR #656, opened 2019-10-12
SPEC-1431 add endpoint to AWS masterkey

kevinAlbs:
Note:
- Before drivers can start, MONGOCRYPT-181 and MONGOCRYPT-153 need to be merged (both of which are in review).
- Drivers will need to update bindings to wrap the new function `mongocrypt_ctx_setopt_masterkey_aws_endpoint`, and call that for data key creation if the optional "endpoint" is passed as part of the DataKeyOpts.masterkey
- Prototyped the tests this with Java here: https://github.com/kevinAlbs/JavaFLEProsePoC/blob/master/src/main/java/EndpointTest.java I had some trouble using my locally built bindings. I managed to get it passing as-is, then messed up my gradle configuration... which is why not all tests were implemented in that PoC.
- Shell needs to update as well

### DIFF
--- a/source/client-side-encryption/tests/README.rst
+++ b/source/client-side-encryption/tests/README.rst
@@ -128,15 +128,16 @@ Then for each element in ``tests``:
    #. Insert the data specified into the ``admin.datakeys`` with write concern "majority".
 
 #. Create a MongoClient.
-   
+
 #. Create a collection object from the MongoClient, using the ``database_name``
-   and ``collection_name`` fields from the YAML file. If a ``json_schema`` is defined in the test,
+   and ``collection_name`` fields from the YAML file. Drop the collection 
+   with writeConcern "majority". If a ``json_schema`` is defined in the test,
    use the ``createCollection`` command to explicitly create the collection:
+
    .. code:: typescript
 
       {"create": <collection>, "validator": {"$jsonSchema": <json_schema>}}
 
-#. Drop the test collection, using writeConcern "majority".
 #. If the YAML file contains a ``data`` array, insert the documents in ``data``
    into the test collection, using writeConcern "majority".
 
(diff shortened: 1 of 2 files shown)


source/client-side-encryption/tests/README.rst
```diff
@@ -128,15 +128,16 @@ Then for each element in ``tests``:
...
 
 #. Create a MongoClient.
-   
+
 #. Create a collection object from the MongoClient, using the ``database_name``
-   and ``collection_name`` fields from the YAML file. If a ``json_schema`` is defined in the test,
+   and ``collection_name`` fields from the YAML file. Drop the collection 
+   with writeConcern "majority". If a ``json_schema`` is defined in the test,
    use the ``createCollection`` command to explicitly create the collection:
+
```
kevinAlbs: Drive-by fix. These instructions were saying to drop the collection after creating it with a JSON schema.

source/client-side-encryption/tests/README.rst
```diff
@@ -497,3 +498,91 @@ The corpus test exhaustively enumerates all ways to encrypt all BSON value types
...
+   .. code:: javascript
+
+      {
+        region: "us-east-1",
+        key: "arn:aws:kms:us-east-1:579766882180:key/89fcc2c4-08b0-4bd9-9f25-e30687b580d0",
+        endpoint: "kms.us-east-2.amazonaws.com"
+      }
+
+   Expect this to fail with an exception containing the message: "Credential should be scoped to a valid region, not 'us-east-1'"
```
ShaneHarvey: Does this error message come from AWS or libmongocrypt?

kevinAlbs: AWS. I realize this unfortunately means AWS changing error messages will break our tests. But the point of this test is to verify we're using a real KMS endpoint correctly. The test that uses example.com could foreseeably fail with a parser error for other reasons.

ShaneHarvey: Is there an error code that we could look for instead of the error message? Alternatively, could drivers just look for "us-east-1" in the error message?

kevinAlbs: No error code unfortunately. Changed to just be "us-east-1".

PR #657, opened 2019-10-14
Remainder of FLE GA

kevinAlbs:
A batch of the remaining small SPEC changes.

JSON spec test changes were validated using C driver in-progress implementation.

Note for drivers implementers. The relevant changes that impact the FLE implementation are as follows:

- Update the return type (if necessary) of createDataKey to be a BSON binary (SPEC-1449 createDataKey returns UUID)
- Spawned mongocryptd should redirect stdout and stderr to /dev/null (SPEC-1469 silence 
mongocryptd by default)
- Resync the FLE JSON spec tests (SPEC-1466 test that fetching keys uses readConcern=majority)
- Update the prose test of "Data key and double encryption" to check command started events (SPEC-1452 test that created data keys insert with majority writeConcern)
- Check that bulk write batch splitting logic matches the section "Size limits for Write Commands" and update the prose test of "BSON size limits and batch splitting". (SPEC-1397 limit 2MiB limit to bulk writes)

cc @egiurleo

### DIFF
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -255,12 +255,12 @@ feature. and that auto encryption only occurs on collection level
 operations by including the following in the driver documentation for
 AutoEncryptionOpts:
 
-Automatic encryption is an enterprise only feature that only applies to
-operations on a collection. Automatic encryption is not supported for
-operations on a database or view, and operations that are not bypassed
-will result in error (see `libmongocrypt: Auto Encryption Whitelist`_).
-To bypass automatic encryption for all operations, set
-bypassAutoEncryption=true in AutoEncryptionOpts.
+   Automatic encryption is an enterprise only feature that only applies to
+   operations on a collection. Automatic encryption is not supported for
+   operations on a database or view, and operations that are not bypassed
+   will result in error (see `libmongocrypt: Auto Encryption Whitelist`_).
+   To bypass automatic encryption for all operations, set
+   bypassAutoEncryption=true in AutoEncryptionOpts.
 
 Explicit encryption/decryption and automatic decryption is a community
 feature. A MongoClient configured with bypassAutoEncryption=true will
@@ -270,9 +270,9 @@ Drivers MUST document that auto encryption requires the authenticated
 user to have the listCollections privilege action by including the
 following in the driver documentation for MongoClient.
 
-Automatic encryption requires the authenticated user to have the
-`listCollections privilege
-action <https://docs.mongodb.com/manual/reference/command/listCollections/#dbcmd.listCollections>`__.
+   Automatic encryption requires the authenticated user to have the
+   `listCollections privilege
+   action <https://docs.mongodb.com/manual/reference/command/listCollections/#dbcmd.listCollections>`__.
 
 See `Why is client side encryption configured on a MongoClient?`_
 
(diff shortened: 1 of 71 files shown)


source/client-side-encryption/etc/test-templates/aggregate.yml.template
```diff
@@ -27,14 +27,12 @@ tests:
       - command_started_event:
           command:
             listCollections: 1
-            cursor: {}
```
kevinAlbs: Note, C driver does not send a cursor document for listCollections if no batch size is specified. I did not see in the [enumerate collections spec](https://github.com/mongodb/specifications/blob/master/source/enumerate-collections.rst) any requirement to do so. So I removed this from expectations.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -699,40 +699,28 @@ Unencrypted data MUST NOT appear in the data of any command monitoring
...
+(analogous to constructing `OP_MSG payload type 0 <https://github.com/mongodb/specifications/blob/70628e30c96361346f7b6872571c0ec4d54846cb/source/message/OP_MSG.rst#sections>`_, not a document sequence).
+Automatic encryption returns a single (possibly modified) BSON document as the
+command to send.
 
--  maxBsonObjectSize = 2097152 bytes
--  maxMessageSizeBytes = 6000000 bytes
+Because automatic encryption increases the size of commands, the driver
+MUST split bulk writes at a reduced size limit before undergoing automatic
+encryption. The write payload MUST be split at 2MiB (2097152). Where batch
+splitting occurs relative to automatic encryption is implementation-dependant.
```
jyemin: "dependent"?

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -699,40 +699,28 @@ Unencrypted data MUST NOT appear in the data of any command monitoring
...
 
-(Note, these are 1/8th of the corresponding reported sizes of a MongoDB
-4.2 server).
+Drivers MUST not reduce the size limits for a single command before automatic
```
jyemin: Perhaps "for a single write" in a bulk write operation?

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: Ah, not the only occurrence. One day I will learn to spell this word.

source/client-side-encryption/client-side-encryption.rst

kevinAlbs:
Per discussion, some additional incoming tests for bulk write splitting before approval.

kevinAlbs:
Update bulk write prose tests – I will validate with libmongoc in a couple hours.

source/client-side-encryption/tests/README.rst
```diff
@@ -363,13 +367,9 @@ First, perform the setup.
...
 Using ``client_encrypted`` perform the following operations:
 
-#. Insert ``{ "_id": "no_encryption_under_2mib", "unencrypted": <the string "a" repeated (2097152 - 1000) times> }``. (Note 2097152 is 2^21 bytes, or 2 MiB).
-
-   Expect this to succeed.
+#. Insert ``{ "_id": "over_2mib_under_16mib", "unencrypted": <the string "a" repeated 2097152 times> }``.
 
-#. Insert ``{ "_id": "no_encryption_over_2mib", "unencrypted": <the string "a" repeated 2097152 times> }``.
-
-   Expect this to throw an exception due to exceeding the reduced maximum BSON document size.
```
ShaneHarvey: Do we have a test to ensure that a command <=16mib before encryption but > maxBsonObjectSize post encryption will fail?

kevinAlbs: Good call. Added two more tests.

kevinAlbs:
Validated all bulk write prose tests pass with the libmongoc in-progress implementation (after fixing a small bug uncovered in libmongoc's batch splitting logic).

PR #659, opened 2019-10-14
SPEC-1471: Reorder assertion for "keep connnection pool" test

jmikola:
https://jira.mongodb.org/browse/SPEC-1471

When verifying that the pool is not cleared, move the assertion immediately after the failed insert. This change should have no effect for drivers using CMAP for assertions, but may be relevant for those using connection counts from serverStatus.

See: https://github.com/mongodb/specifications/pull/627#issuecomment-540612468

### DIFF
--- a/source/connections-survive-step-down/tests/README.rst
+++ b/source/connections-survive-step-down/tests/README.rst
@@ -100,11 +100,10 @@ This test requires a replica set with server version 4.2 or higher.
 - Execute an insert into the test collection of a ``{test: 1}``
   document.
 - Verify that the insert failed with an operation failure with 10107 code.
+- Verify that the connection pool has not been cleared,
+  following the instructions in section `How to verify the connection pool has not been cleared`_ 
 - Execute an insert into the test collection of a ``{test: 1}``
   document and verify that it succeeds.
-- Verify that the connection pool has not been cleared,
-  following the instructions in section `How to verify the connection pool has not been cleared`_
-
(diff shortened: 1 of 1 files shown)


p-mongo:
Since the purpose of the test is to ensure the final insert is done over the existing connection, I believe the original assertion must remain in its place after the insert. I am OK with copying it to also be done before the insert.

prashantmital:
I am not sure I understand the need for this change? I think this is the only test in which the final insert is expected to reuse the existing connection and therefore the assertion is currently in the correct place. The other tests in which the connection is expected to be cleared should attempt the insert _after_ asserting the expected state of the pool. Am I missing something?

jmikola:
In hindsight, I think this change is unnecessary. Closing.

[`connections.totalCreated`](https://docs.mongodb.com/manual/reference/command/serverStatus/#serverstatus.connections.totalCreated) only increases over the lifetime of the server, so deferring the assertion to the end of the test does accomplish the same goal provided that the initial value of `connections.totalCreated` was captured before the first `insert` fails.

A "has not been cleared" test really doesn't care about order the same way a "has been cleared" test might. In this case, the value should never change. For a "has been cleared" test, we'd want to assert that it _did_ change but also examine _when_ it changed. I'm thinking of `connections.totalCreated` as I write this, but it's no different from asserting that an event was never published over the course of a test or _was_ published at a specific step.

PR #661, opened 2019-10-14
SPEC-1231 Explicitly state that SRV URI lookup is excluded from no I/O in constructor prohibition


p-mongo:


source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -442,6 +443,9 @@ as long as the error persists.
 
 See `clients do no I/O in the constructor`_ for the justification.
 
+If given an SRV URI, the constructor MUST perform I/O to resolve the URI
```
jyemin: Why must a driver do I/O in the constructor?  My experience with the Java driver was that it was less complex to do all SRV record lookup asynchronously once the polling requirement was added for mongos.

For TXT records, it made more sense to continue doing that lookup in the constructor because of aspects of the Java driver's design, but I see no reason to require it.  

Can you explain your thinking here?

p-mongo: This is exactly what I wrote in https://jira.mongodb.org/browse/SPEC-1231 and the decision there was to mandate the different behavior for initial lookup and srv polling. I think I personally prefer no i/o in constructor at all, if possible.

jyemin: After reading those comments, I still don't want support change.  The polling spec came later on, so it was reasonable for drivers to change behavior.  Do we really want to force drivers to change back?  What is the benefit to users?

p-mongo: What would you like to do?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -431,7 +431,8 @@ and `what's the point of periodic monitoring?`_)
 Client construction
 '''''''''''''''''''
 
-The client's constructor MUST NOT do any I/O.
+Except for the initial SRV lookup if given an SRV URI,
```
jyemin: Add TXT record lookup.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
jyemin: Remove this new paragraph entirely.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
jmikola: I suggest we rephrase:

> Except for [initial DNS seedlist discovery](https://github.com/mongodb/specifications/blob/master/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst) when given a connection string with a `mongodb+srv` scheme, the client's constructor MUST NOT perform any I/O."

The details about SRV/TXT records are not especially relevant to the SDAM spec. The important thing is that we're alluding to the DNS lookup behavior.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
jmikola: I concur that this paragraph can be removed. The earlier change in this PR allows exactly one exception to perform I/O in the constructor, but also doesn't require that drivers perform that I/O.

I quickly re-read the [Initial DNS Seedlist Discovery](https://github.com/mongodb/specifications/blob/master/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst) spec and I found no mention of requiring drivers to perform DNS lookup immediately during MongoClient construction. It seems like it's simply unspecified and we permit drivers to deviate, but any further clarification on that would better belong in the Initial DNS Seedlist Discovery spec, rather than SDAM.

p-mongo:
Made suggested changes.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
jyemin: I'll leave it to you to resolve this one @jmikola

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -431,7 +431,10 @@ and `what's the point of periodic monitoring?`_)
 Client construction
 '''''''''''''''''''
 
-The client's constructor MUST NOT do any I/O.
+Except for `initial DNS seed list discovery
```
jmikola: I think "seedlist" can be one word, since it's the title of the spec, but up to you.

p-mongo: Looking at this again, the phrase "seed list" occurs a number of times in the sdam spec, but "seedlist" does not occur at all.

PR #662, opened 2019-10-14
Fix SPEC-1378 mongos pinning test expects no unpinning on MaxTimeoutMSExpired


p-mongo:


source/transactions/tests/pin-mongos.yml
```diff
@@ -327,12 +327,12 @@ tests:
             mode: { times: 1 }
             data:
               failCommands: ["commitTransaction"]
-              errorCode: 50 # MaxTimeMSExpired
```
jmikola: The comment above this `targetedFailPoint` operation still states "Fail the commit with a non-transient error, eg MaxTimeMSExpired". I think that needs to be updated.

Also, would it make sense to have the comment on this line refer to "ManualInterventionRequired" (per: [error_codes.yml](https://github.com/mongodb/mongo/blob/d97cca2c6d5be0324da0dc4d56ca269714af8f34/src/mongo/base/error_codes.yml#L75))? Reading the description for [SPEC-1378](https://jira.mongodb.org/browse/SPEC-1378), I don't think there's any reason this tests still needs to refer to MaxTimeMSExpired at all (even with a negation). The important bit is that we're using an error code that does _not_ result in a transient error, which would become a UnknownTransactionCommitResult label for a `commitTransaction` command.

p-mongo: Made suggested changes, does this look OK?

PR #663, opened 2019-10-14
SPEC-1407 clarify that SRV polling tests require sharded cluster

kevinAlbs:


source/polling-srv-records-for-mongos-discovery/polling-srv-records-for-mongos-discovery.rst
```diff
@@ -150,9 +150,16 @@ specification. However, it might be possible to mock changes to DNS SRV
 records such that automated testing is doable. In any case, the following
 tests should be executed, either manually, or programmatically.
 
+To test, start a sharded cluster and mongos servers on ports 27017, 27018,
+27019, and 27020.
```
p-mongo: I think using the phrase "a sharded cluster *with* mongos servers on ports ..." will be more clear.

kevinAlbs: Agreed, done.

PR #664, opened 2019-10-14
SPEC-1408 add missing retryable reads criterion

kevinAlbs:


source/retryable-reads/retryable-reads.rst
```diff
@@ -75,6 +75,10 @@ NetworkTimeout                  89
 SocketException                 9001
 =============================== ==============
 
+- a server error response without an error code or one different from those
```
kevinAlbs: Given the above says:

> An error is considered retryable if it meets any of the criteria defined under Retryable Writes: Terms: Retryable Error, minus the final criterion about write concern errors. For convenience, the relevant criteria have been adapted to retryable reads and reproduced below.

I'm assuming this was intended to be included in the criteria.

PR #665, opened 2019-10-14
SPEC-1396 Drivers MUST clear connection pools when SDAM monitoring fails due to a network error


p-mongo:


source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1308,10 +1308,11 @@ Error handling
 Network error during server check
 `````````````````````````````````
 
-When a server `check`_ fails due to a network error,
-the client SHOULD clear its connection pool for the server:
+When a server `check`_ fails due to a network error (including a timeout),
+the client MUST clear its connection pool for the server:
```
ShaneHarvey: Any ideas of how we can test this? Maybe in the CMAP spec tests we can configure isMaster to fail once with a network error via configureFailPoint, wait for the monitor to run the doomed isMaster, and then assert that the pool has been cleared.

ShaneHarvey: I'm not saying we need tests in this PR. Just spitballing. The current changes LGTM.

p-mongo: I wrote several tests in https://github.com/mongodb/mongo-ruby-driver/pull/1475 but I would prefer to wait until the unified spec runner work is complete before proposing spec tests that touch both cmap and sdam.

jyemin: Do we need to wait for that?  It's a non-goal in the unified spec runner scope to define the format for non-”operations” tests, including CMAP and SDAM.

p-mongo: Our cmap spec runner does not do any i/o, thus as it is implemented it is unable to set fail points on the server.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1308,10 +1308,11 @@ Error handling
 Network error during server check
 `````````````````````````````````
 
-When a server `check`_ fails due to a network error,
-the client SHOULD clear its connection pool for the server:
+When a server `check`_ fails due to a network error (including a timeout),
```
mbroadst: The spec defines terms "network error" and "network timeout", should that second term be used here?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1308,10 +1308,11 @@ Error handling
...
+the client MUST clear its connection pool for the server:
 if the monitor's socket is bad it is likely that all are.
-(See `JAVA-1252 <https://jira.mongodb.org/browse/JAVA-1252>`_.)
+(See `JAVA-1252 <https://jira.mongodb.org/browse/JAVA-1252>`_,
+`SPEC-1396 <https://jira.mongodb.org/browse/SPEC-1396>`_.)
```
mbroadst: SPEC tickets are not public, I don't think we can include this in our public documentation

jyemin: +1

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
p-mongo: Changed.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
p-mongo: Removed.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
p-mongo: @ShaneHarvey It appears that failcommand fail point is ignored for ismaster commands.

p-mongo: Turns out it does work when ismaster is given in mixed case, https://jira.mongodb.org/browse/SERVER-44414.

PR #666, opened 2019-10-15
SPEC-1355 Explicitly define write concern option mappings


p-mongo:


jmikola:
> Are we going to open a DRIVERS ticket for this?

@jyemin: [SPEC-1355](https://jira.mongodb.org/browse/SPEC-1355) is marked "Driver Changes: Not Needed." I believe this ticket is just for clarification.

PR #667, opened 2019-10-15
FLE bypass killAllSessions

kevinAlbs:


source/client-side-encryption/client-side-encryption.rst
```diff
@@ -929,6 +929,7 @@ listIndexes            BYPASS
 renameCollection       BYPASS
 explain                AUTOENCRYPT
 ping                   BYPASS
+killAllSessions        BYPASS
```
ShaneHarvey: Do we also want to add the other sessions related commands: https://docs.mongodb.com/manual/reference/command/nav-sessions/
- killSessions
- killAllSessionsByPattern
- refreshSessions

kevinAlbs: Ooh, I was unaware of those. Seems reasonable to bypass them as well. Updated.

PR #668, opened 2019-10-16
SPEC-1301 Require auth on direct connections

xdg:
To address [GODRIVER-1344](https://jira.mongodb.org/browse/GODRIVER-1344).

### DIFF
--- a/source/auth/auth.rst
+++ b/source/auth/auth.rst
@@ -159,8 +159,8 @@ handshake:
    MongoDB Handshake fails with an error, drivers MUST treat this as
    an authentication error.
 
-#. If the server is not of type Standalone, RSPrimary, RSSecondary or
-   Mongos, no authentication is possible and the handshake is complete.
+#. If the server is of type RSArbiter, no authentication is possible and the
+   handshake is complete.
 
 #. If credentials exist:
 


source/auth/auth.rst
```diff
@@ -160,7 +160,8 @@ handshake:
    an authentication error.
 
 #. If the server is not of type Standalone, RSPrimary, RSSecondary or
-   Mongos, no authentication is possible and the handshake is complete.
+   Mongos and the connection is not a direct connection,
```
jyemin: Assuming we're going to add the arbiter exception:

Could we simply say that we authenticate unless it's an Arbiter?   For non-direct connections, is there any reason the driver would authenticate to any other type of server than the ones already listed?

xdg: IIRC, single threaded drivers use one socket for monitoring and operations and therefore will normally auth to anything they connect to.  Other than direct connection (which is user's responsibility), we don't want them to attempt auth to RSOther/RSGhost.

p-mongo:
For my curiosity, in what circumstances would a driver establish a connection to RSOther and use it without authentication? Other than for monitoring purposes.

source/auth/auth.rst
```diff
@@ -159,8 +159,11 @@ handshake:
    MongoDB Handshake fails with an error, drivers MUST treat this as
    an authentication error.
 
-#. If the server is not of type Standalone, RSPrimary, RSSecondary or
-   Mongos, no authentication is possible and the handshake is complete.
+#. If the server is of type Standalone, RSPrimary, RSSecondary or Mongos,
+   continue with authentication in Step 3. If the connection is a direct
+   connection and the server is not type RSArbiter, continue with
+   authentication in Step 3. Otherwise, no authentication is possible and the
+   handshake is complete.
```
jmikola: Would specifying a "single topology type" be preferable to "direct connection" in this text?

SDAM requires drivers to support [direct connections](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#general-requirements), but [Initial TopologyType](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#initial-topologytype) suggests that the behavior corresponds to forcing a single topology type:

> The user MUST be able to set the initial TopologyType to Single.

This came up recently in the scope for [WRITING-4391](https://jira.mongodb.org/browse/WRITING-4391) (unify replica set discovery behavior) and we agreed that to be strict about referring to direct connections as having a single topology type.

source/auth/auth.rst
jyemin: My practical question is: are multithreaded drivers like Java and C#, which just authenticate all pooled connections regardless of whether they are RSOther or RSSecondary (and whose handshakers are unaware of whether the connection is "direct" or not), in violation of this?  I would like the answer to be "no", and that the wording of the change makes that clear.

xdg: I'm not sure I follow.  You said, "which just authenticate all pooled connections regardless" -- by that do you mean it just happens they never actually auth in practice because they aren't selected?  What would minPoolSize do?  What does it mean to have a minPoolSize to a non-data-bearing node (outside direct connections)?

xdg:
@p-mongo -- a hidden secondary node is classified as RSOther because it's not selectable.  But a direct connection to it needs to auth.  TOOLS-2380 is an example of the problem this PR is trying to solve.

source/auth/auth.rst
xdg: Updated.

source/auth/auth.rst
jyemin: SDAM requires that, when is discovery mode, replica set primaries, secondaries, passives, and arbiters are included in the topology. But the built-in server selector (either for a writable server or to satisfy read preference) would only ever select a primary or secondary.  So yes, in normal practice passives and arbiters would never be selected and always have empty connection pools.

But the CMAP spec says nothing about treating passives and arbiters any differently.  In [this design rationale](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#why-do-we-set-minpoolsize-across-all-members-of-a-replicaset-when-most-traffic-will-be-against-a-primary), in fact, it specifically says not to treat different replica set member types any differently.  So you're absolutely right that if you set minPoolSize, drivers would end up creating connections to passives and arbiters in addition to primary and secondaries.  We could argue about whether that's the right behavior, but that's what it is now.

Given that specification, does it actually make sense to authenticate all server types (except arbiters, where auth would definitely fail) even when in discovery mode?  As an example, consider a replica set with a passive server that is, for some reason, not yet marked as a secondary; and an application with minPoolSize set:

* Driver creates minPoolSize connections to all servers, even the passive server
* Some time later the passive server becomes a secondary (perhaps after an initial sync completes?)
* Driver detects the change and updates the server type from RSOther to RSSecondary
* Server selection selects the new secondary for some read operation

Expected result: the pooled connections to the new secondary should be authenticated

p-mongo: As clarified in https://github.com/mongodb/specifications/pull/645, pools should only be created for servers the drivers knows to be data bearing, i.e., those it may legitimately connect to to perform an operation.

https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#data-bearing-server-type includes standalone in the data bearing server list.

xdg: Auth against a secondary doing initial sync might fail if the auth credentials haven't replicated.

jyemin: So be it.  Then the pool will be empty until authentication starts succeeding.  I think that's a reasonable behavior.

xdg:
I don't have a strong opinion between Jeff's opinion and Jeremy's LGTM.  Could leads please come to some agreement on how they'd like to go forward?

jyemin:
Jeremy and I discussed, and agreed that the following wording would be acceptable to both of us: 

> If the server is of type RSArbiter, no authentication is possible and the handshake is complete.

jmikola:
While talking this over with @jyemin, I realized that connection pool behavior for non-data-bearing nodes is unspecified. I opened [SPEC-1474](https://jira.mongodb.org/browse/SPEC-1474) to discuss that.

I think the rewording to simply exclude RSArbiter types from authentication is most equivalent to David's original text, with the minor difference that drivers may or may not end up authenticating against RSGhost and RSOther server types depending on whether they create pools (something [SPEC-1474](https://jira.mongodb.org/browse/SPEC-1474) should clarify). That said, I don't see the harm in authenticating those server types since they may very well become query-able nodes in due time if they are not otherwise removed from the topology.

xdg:
Thanks.  Amended and pushed.

PR #670, opened 2019-10-18
SPEC-1475: add SDAM monitoring test for debouncing update events

mbroadst:
The SDAM specification describes an algorithm for ServerDescription
equality, and mandates that this equality operator should be used
during topology updates to guard event emission. We should include
a test that expresses this part of the spec.

### DIFF
--- a/source/server-discovery-and-monitoring/tests/monitoring/standalone_suppress_equal_description_changes.json
+++ b/source/server-discovery-and-monitoring/tests/monitoring/standalone_suppress_equal_description_changes.json
@@ -1,0 +1,113 @@
+{
+  "description": "Monitoring a standalone connection - suppress update events for equal server descriptions",
+  "uri": "mongodb://a:27017",
+  "phases": [
+    {
+      "responses": [
+        [
+          "a:27017",
+          {
+            "ok": 1,
+            "ismaster": true,
+            "minWireVersion": 0,
+            "maxWireVersion": 4
+          }
+        ],
+        [
+          "a:27017",
+          {
+            "ok": 1,
+            "ismaster": true,
+            "minWireVersion": 0,
+            "maxWireVersion": 4
+          }
+        ]
+      ],
+      "outcome": {
+        "events": [
+          {
+            "topology_opening_event": {
+              "topologyId": "42"
+            }
+          },
+          {
+            "topology_description_changed_event": {
+              "topologyId": "42",
+              "previousDescription": {
+                "topologyType": "Unknown",
+                "servers": []
+              },
+              "newDescription": {
+                "topologyType": "Single",
+                "servers": [
+                  {
+                    "address": "a:27017",
+                    "arbiters": [],
+                    "hosts": [],
+                    "passives": [],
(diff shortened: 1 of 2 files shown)


source/server-discovery-and-monitoring/tests/monitoring/standalone_debounce_description_updates.yml
```diff
@@ -0,0 +1,73 @@
...
+              passives: []
+              type: "Unknown"
+            newDescription:
+              address: "a:27017"
+              arbiters: []
+              hosts: []
+              passives: []
+              type: "Standalone"
+        -
+          topology_description_changed_event:
```
jmikola: I haven't worked with SDAM spec tests before, but does the following quote from [Test Phases](https://github.com/mongodb/specifications/tree/master/source/server-discovery-and-monitoring/tests#test-phases) ensure that the observed events cannot exceed those included in `outcome.events`?

> For monitoring tests, once all responses are processed, assert that the events collected so far by the SDAM event listener are _equivalent_ to the events specified in the phase.

I wanted to ask, as our APM assertions in other operation-based tests would typically allow (and ignore) extra events to be emitted beyond what's in the spec tests.

If so, is the purpose of this test to assert that a second `server_description_changed_event` and/or `topology_description_changed_event` event is not emitted with identical data?

source/server-discovery-and-monitoring/tests/monitoring/standalone_debounce_description_updates.yml
```diff
@@ -0,0 +1,73 @@
+description: "Monitoring a standalone connection - multiple equal description updates are debounced"
```
jmikola: I hadn't heard of "debouncing" before this PR, but it seems to pertain to ensuring events don't fire too often (often in the context of UI). This is PR is about avoiding emission of a redundant event entirely because there is no change to report. Would it make more sense to say "suppress update events for unchanged server descriptions" or something to that effect?

p-mongo: I like Jeremy's wording.

source/server-discovery-and-monitoring/tests/monitoring/standalone_debounce_description_updates.yml
p-mongo: The specification calls for exact match between emitted events and the expected events.

In Ruby we sometimes explicitly exclude certain events from the checks (such as authenticate and sasl* commands), but in general we do perform equality comparison and not just a subset.

mbroadst: Yes, the purpose of this test is to show that only one set of events is emitted given the same input

source/server-discovery-and-monitoring/tests/monitoring/standalone_debounce_description_updates.yml
mbroadst: Updated file names and descriptions, PTAL

PR #671, opened 2019-10-21
separate tlsCertificateKeyFilePassword test to facilitate skippi…

saghm:
Currently, the test for `tlsCertificateKeyPassword` is bundled with the required TLS options, but the URI options spec specifies that `tlsCertificateKeyFilePassword` is only required if the language/runtime supports it. Similar to other non-required options, we should separate this test so that drivers that don't implement this option can skip it without skipping the required options.

### DIFF
--- a/source/uri-options/tests/tls-options.json
+++ b/source/uri-options/tests/tls-options.json
@@ -2,7 +2,7 @@
   "tests": [
     {
       "description": "Valid required tls options are parsed correctly",
-      "uri": "mongodb://example.com/?tls=true&tlsCAFile=ca.pem&tlsCertificateKeyFile=cert.pem&tlsCertificateKeyFilePassword=hunter2",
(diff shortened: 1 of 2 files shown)


source/uri-options/tests/tls-options.json
```diff
@@ -2,18 +2,28 @@
   "tests": [
     {
       "description": "Valid required tls options are parsed correctly",
-      "uri": "mongodb://example.com/?tls=true&tlsCAFile=ca.pem&tlsCertificateKeyFile=cert.pem&tlsCertificateKeyFilePassword=hunter2",
+      "uri": "mongodb://example.com/?tls=true&tlsCAFile=ca.pem&tlsCertificateKeyFile=cert.pem",
```
jmikola: I think this PR is missing the YAML change.

saghm:
I'm having some major issues with `js-yaml`, but I think this is ready to go now

PR #672, opened 2019-10-22
SPEC-1420: Expand use of error labels for retryable writes

egiurleo:
The server will soon be adding a `RetryableWriteError` label to retryable write errors. The drivers can use this label to determine whether to retry an error on a retryable write operation, rather than keeping track of a list of error codes.

See the [Retryable Writes Error Labels Proposal document](https://docs.google.com/document/d/1etljLCq1GLchti_7pVZta4JRaoCoc8OBnjVyYwj_slo) for more context.


source/retryable-writes/retryable-writes.rst
```diff
@@ -66,7 +66,23 @@ ClientSession
    name of this object MAY vary across drivers.
 
 Retryable Error
-   An error is considered retryable if it meets any of the following criteria:
+   An error is considered retryable if it has a RetryableWriteError label in
+   its top-level "errorLabels" field.
+
+   For server versions 4.3 and newer, MongoDB will add a RetryableWriteError label to
```
jmikola: I think we should either refer to the specific 4.3.x release (if possible), or stick to stable release versions and use "4.4". I believe the convention across most specs is to just use the stable version.

It's also not a problem if the spec ends up saying "4.4" and we use a specific 4.3.x version in the test file requirements.

source/retryable-writes/retryable-writes.rst
```diff
@@ -66,7 +66,23 @@ ClientSession
...
+   errors or server responses that it considers retryable before returning them to the
+   driver. As new server versions are released, the errors that are labeled with the
+   RetryableWriteError may change.
+
+   There are some cases where the driver MUST add the RetryableWriteError label to
+   an error or server response from a supported retryable write operation:
+   - any network exception
+   - any error with an error message containing the substrings "not master"
```
jmikola: Did we clarify whether this was still needed for 3.6+ servers?

source/retryable-writes/retryable-writes.rst
```diff
@@ -66,7 +66,23 @@ ClientSession
...
+   its top-level "errorLabels" field.
+
+   For server versions 4.3 and newer, MongoDB will add a RetryableWriteError label to
+   errors or server responses that it considers retryable before returning them to the
+   driver. As new server versions are released, the errors that are labeled with the
+   RetryableWriteError may change.
+
+   There are some cases where the driver MUST add the RetryableWriteError label to
+   an error or server response from a supported retryable write operation:
+   - any network exception
```
jmikola: Should we add "(e.g. socket timeout or error)" here?

source/transactions/transactions.rst
```diff
@@ -1383,6 +1388,7 @@ durable, which achieves the primary objective of avoiding duplicate commits.
 **Changelog**
 -------------
 
+:2019-10-21: Specify that a commit error can have two error labels
```
jmikola: I think this also warrants bumping the spec's patch version from 1.5.3 to 1.5.4.

source/retryable-writes/retryable-writes.rst
```diff
@@ -762,6 +781,9 @@ cluster (mongos just forwards the error from the shard's replica set).
 Changes
 =======
 
+2019-10-21: Change the definition of "retryable write" to be based on the
+RetryableWriteError label.
```
jmikola: Since this spec change entails changes to driver behavior and API, I think we should bump the spec version from 1.3.1 to 1.4.0.

source/retryable-writes/tests/insertOne-errorLabels.yml
```diff
@@ -0,0 +1,80 @@
...
+                arguments:
+                    document: { _id: 1, x: 11 }
+            outcome:
+                result:
+                    insertedId: 1
+                collection:
+                    data:
+                        - { _id: 1, x: 11 }
+    -
+        description: "InsertOne succeeds after a connection failure"
```
jmikola: This looks like it'd duplicated from `insertOne-serverErrors.yml`. Should this be deleted?

source/retryable-writes/tests/insertOne-errorLabels.yml
```diff
@@ -0,0 +1,80 @@
...
+                    - { _id: 1, x: 11 }
+    -
+        description: "InsertOne fails with a RetryableWriteError label after two connection failures"
+        failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 2 }
+            data:
+                failCommands: ["insert"]
+                closeConnection: true
+        operaion:
```
jmikola: "operation"

source/retryable-writes/tests/insertOne-errorLabels.yml
```diff
@@ -0,0 +1,80 @@
...
+            name: "insertOne"
+            arguments: { _id: 1, x: 11 }
+        outcome:
+            result:
+                insertedId: 1
+            collection:
+                data:
+                    - { _id: 1, x: 11 }
+    -
+        description: "InsertOne fails with a RetryableWriteError label after two connection failures"
```
jmikola: This looks unique, but it also shouldn't require a 4.3.0 server since drivers add the RetryableWriteError. I think this can be moved to `insertOne-serverErrors.yml`.

source/retryable-writes/tests/insertOne-errorLabels.yml
```diff
@@ -0,0 +1,80 @@
...
+
+    -
+        description: "InsertOne does not retry and fails on error without RetryableWriteError label"
+        failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+                failCommands: ["insert"]
+                errorLabels: []
+        operaion:
```
jmikola: "operation"

source/retryable-writes/tests/insertOne-errorLabels.yml
```diff
@@ -0,0 +1,80 @@
...
+                data: []
+
+    -
+        description: "InsertOne does not retry and fails on error without RetryableWriteError label"
+        failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+                failCommands: ["insert"]
+                errorLabels: []
```
jmikola: My first suggestion: Rather than use `errorLabels` here, can we simply use an error code that we know will not be labeled by either the driver or server (e.g. 112, WriteConflict) and move this to `insertOne-serverErrors.yml`?

If we do that, this might be redundant since we already have "InsertOne fails after Interrupted" in that file, which asserts that an insert fails after the first attempt receives a non-retryable error. In that case, I think we can delete this entirely and simply add an `errorLabelsOmit` assertion to that existing test.

If this PR is mainly going to focus on `insertOne` test files before we settle on changes to make across the board, I think you should take a look at all tests in `insertOne-serverErrors.yml` and `insertOne.yml` that have error assertions and see where you can add `errorLabelsContain` or `errorLabelsOmit` assertions. Those should be portable across all server versions, since the driver and server labeling should be consistent. Furthermore, any failures in such tests down the line should inform us of a possible deviation as soon as it pops up in a new server release.

source/retryable-writes/tests/insertOne-errorLabels.yml
```diff
@@ -0,0 +1,80 @@
+runOn:
+    -
+        minServerVersion: "4.3.0"
+        topology: ["replicaset", "sharded"]
+
+data: []
+
+tests:
+    -
+        description: "InsertOne succeeds after RetryableWriteError from server"
```
jmikola: After reviewing other tests in this file, I think this may be the only one that justifies keeping `insertOne-errorLabels.yml`.

source/retryable-writes/tests/README.rst
```diff
@@ -205,6 +205,9 @@ Each YAML file has the following keys:
       result object if their BulkWriteException (or equivalent) provides access
       to a write result object.
 
+      - ``errorLabelsContain``: A list of error label strings that the
+        error is expected to have.
```
jmikola: Per my other comments, I think we should also add `errorLabelsOmit` here if we introduce that syntax into the tests.

source/transactions/tests/retryable-commit-errorLabels.yml
```diff
@@ -0,0 +1,261 @@
...
+        arguments:
+        session: session0
+        document:
+            _id: 1
+        result:
+        insertedId: 1
+      - name: commitTransaction
+        object: session0
+        result:
+        errorLabelsContain: ["RetryableWriteError", "UnknownTransactionCommitResult"]
```
jmikola: If the driver adds "RetryableWriteError" for a network error, is there any reason this needs to be in a separate test file requiring a 4.3+ server? Related to my comments on the `insertOne` tests, I think we only need to require that server version when using the `errorLabels` fail point option.

If that's the case, then perhaps we can remove this test in favor of incorporating "RetryableWritesError" label assertions into existing commit tests. "add unknown commit label to connection errors" might be one such test.

source/transactions/tests/retryable-commit-errorLabels.yml
```diff
@@ -0,0 +1,261 @@
...
+        minServerVersion: "4.3.0"
+        topology: ["replicaset", "sharded"]
+
+database_name: &database_name "transaction-tests"
+collection_name: &collection_name "test"
+
+data: []
+
+tests:
+  - description: commitTransaction fails with RetryableWriteError label after two network exceptions
```
jmikola: This test might be redundant in light of "commitTransaction fails after two errors". Perhaps we can add a "RetryableWriteError" assertion to that test instead.

source/transactions/tests/retryable-commit-errorLabels.yml
```diff
@@ -0,0 +1,261 @@
...
+      - name: startTransaction
+        object: session0
+      - name: insertOne
+        object: collection
+        arguments:
+        session: session0
+        document:
+          _id: 1
+        result:
+        insertedId: 1
```
jmikola: Indentation is off here. Please take a pass through everything to see if this mistake was made elsewhere.

source/transactions/tests/retryable-commit-errorLabels.yml
```diff
@@ -0,0 +1,261 @@
...
+      - name: commitTransaction
```
jmikola: Since this fails, can we have an error assertion for the labels? See "commitTransaction fails after WriteConcernError Interrupted" for prior art.

source/transactions/tests/retryable-commit-errorLabels.yml
```diff
@@ -0,0 +1,261 @@
...
+            autocommit: false
+            writeConcern:
+          command_name: commitTransaction
+          database_name: admin
+    outcome:
+      collection:
+        data:
+          - _id: 1
+
+  - description: commitTransaction does not retry error without RetryableWriteError label
```
jmikola: I think this and the test that follow may be the only ones that justify placement in this new file, since you use `errorLabels` in the fail point.

source/transactions/tests/retryable-commit-errorLabels.yml
```diff
@@ -0,0 +1,261 @@
...
+            autocommit: false
+            writeConcern:
+          command_name: commitTransaction
+          database_name: admin
+    outcome:
+        collection:
+            data:
+                - _id: 1
+
+  - description: commitTransaction retries once after a network exception
```
jmikola: Does this add anything of value over the existing "commitTransaction succeeds after connection error" test?

source/transactions/tests/retryable-abort-errorLabels.yml
```diff
@@ -0,0 +1,190 @@
...
+            startTransaction:
+            autocommit: false
+            writeConcern:
+          command_name: abortTransaction
+          database_name: admin
+    outcome:
+      collection:
+          data: []
+
+  - description: abortTransaction retries once after network exception
```
jmikola: Is this any different than "abortTransaction succeeds after connection error" in `retryable-abort.yml`? Perhaps we can just add label assertions to existing abort tests, as those shouldn't require a particular server version.

source/transactions/tests/retryable-abort-errorLabels.yml
```diff
@@ -0,0 +1,190 @@
...
+        topology: ["replicaset, sharded"]
+
+database_name: &database_name "transaction-tests"
+collection_name: &collection_name "test"
+
+data: []
+tests:
+  - description: abortTransaction only retries once after RetryableWriteError from server
+    clientOptions:
+      retryWrites: false
```
jmikola: Is this relevant if the only insert we do is within a transaction? I realize it's in "abortTransaction only performs a single retry" but I don't understand why that test has it either. /cc @ShaneHarvey

source/transactions/tests/retryable-abort-errorLabels.yml
```diff
@@ -0,0 +1,190 @@
...
+            writeConcern:
+          command_name: abortTransaction
+          database_name: admin
+    outcome:
+      collection:
+        data: []
+
+  - description: abortTransaction does not retry without RetryableWriteError label
+    clientOptions:
+      retryWrites: false
```
jmikola: Likewise, not sure if this is relevant.

source/retryable-writes/retryable-writes.rst
egiurleo: Good to know, thanks.

source/retryable-writes/retryable-writes.rst
egiurleo: I asked whether we still needed to handle this case in the replication channel and got this response:
> siyuan.zhou 2:45 PM
> I’m afraid you’d have to audit the code to answer that.

Judah and Lingzhi don't know either. Should I make a server ticket to try and get this done? Should I try to figure it out myself? Is it worth spending time on right now?

source/retryable-writes/retryable-writes.rst
```diff
@@ -66,7 +66,23 @@ ClientSession
...
+   an error or server response from a supported retryable write operation:
+   - any network exception
+   - any error with an error message containing the substrings "not master"
+   or "node is recovering"
+
+   For versions older than 4.3, the MongoDB server does not add the RetryableWriteError
+   label to any errors; for these server versions, the driver MUST add
+   a RetryableWriteError label to errors that meet the following criteria:
 
    - any network exception (e.g. socket timeout or error)
```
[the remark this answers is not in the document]
egiurleo: 👍 Yeah that makes sense.

source/retryable-writes/tests/insertOne-errorLabels.yml
```diff
@@ -0,0 +1,80 @@
+runOn:
+    -
+        minServerVersion: "4.3.0"
```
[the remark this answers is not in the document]
egiurleo: Yes it'll definitely be a different release, but I don't think server team knows which yet. I'll change it to 4.3.x for easy grep-replace.

And yes this is just one example. I wanted your feedback before adding a bunch of new tests so that I didn't go too far off in the wrong direction.

source/retryable-writes/tests/insertOne-errorLabels.yml
egiurleo: I like that! Thank you.

egiurleo: I like that! Thank you!

source/retryable-writes/retryable-writes.rst
jmikola: We can keep this as-is in this PR. A new spec ticket to request investigation seems fine, too, even if we ultimately never budget time for it.

source/transactions/tests/retryable-commit-errorLabels.yml
egiurleo: No, it does not!

source/transactions/tests/retryable-abort-errorLabels.yml
egiurleo: My thinking is that this was testing the line: `abortTransaction is a retryable write command. Drivers MUST retry once after abortTransaction fails with a retryable error according to the Retryable Writes Specification., regardless of whether retryWrites is set on the MongoClient or not.` from the transactions spec. I agree thought that it's not obvious how this is relevant to the test.

egiurleo: /The test does not explicitly state that it is testing that abort still retries even if retryWrites is off.

ShaneHarvey: Yes the `abortTransaction only performs a single retry` test includes `retryWrites: false` to ensure that driver perform the retry regardless of the retryWrites setting. I don't think we need to set  `retryWrites: false` in these new tests as well.

source/retryable-writes/retryable-writes.rst
egiurleo: Here's the SPEC ticket: [SPEC-1477](https://jira.mongodb.org/browse/SPEC-1477). I've linked it to SPEC-1420 and DRIVERS-525 for visibility.

source/transactions/tests/retryable-commit-errorLabels.yml
```diff
@@ -0,0 +1,133 @@
+runOn:
+    -
+        minServerVersion: "4.3.x"
```
jmikola: We can't actually use ".x" here (or at least that isn't specified). It needs to be a real version number. Since we know 4.3.0 doesn't have the functionality we need, you may want to ask the server team if they expect to have the changes implemented for the next patch release. If so, you can preemptively put 4.3.1.

source/transactions/tests/retryable-commit-errorLabels.yml
```diff
@@ -0,0 +1,133 @@
...
+        arguments:
+        session: session0
+        document:
+          _id: 1
+        result:
+        insertedId: 1
+      - name: commitTransaction
+        object: session0
+        result:
+          errorLabelsContain: ["RetryableWriteError"]
```
jmikola: AFAIK, the driver only adds this label for pre-4.4 servers. This test also isn't expecting the commit to be retried, so shouldn't we assert that the error label is omitted?

source/transactions/tests/retryable-abort-errorLabels.yml
```diff
@@ -0,0 +1,126 @@
...
+        arguments:
+          session: session0
+          document:
+            _id: 1
+          result:
+            insertedId: 1
+      - name: abortTransaction
+        object: session0
+        result:
+          errorLabelsContain: ["RetryableWriteError"]
```
jmikola: As with the commit test, I wouldn't expect the driver to add the error label for a 4.4+ server. Also considering that the abort isn't retried in this case. Am I misunderstanding something?

source/retryable-writes/tests/insertOne-errorLabels.yml
```diff
@@ -0,0 +1,44 @@
...
+                errorLabels: []
+                errorCode: 11600 # InterruptedAtShutdown, normally a retryable error code
+            operation:
+                name: "insertOne"
+                arguments:
+                  document: { _id: 1, x: 11 }
+          outcome:
+            error: true
+            result:
+                errorLabelsOmit: ["RetryableWriteError"]
```
jmikola: Here we test that the label is omitted. I think that justifies also testing for omission in the similar abort/commit tests below.

source/transactions/tests/retryable-commit-errorLabels.yml
egiurleo: Yes! To be clear, I'm just putting that as a placeholder until I know the actual server number for easy search and replace.

source/retryable-writes/tests/bulkWrite-errorLabels.json
```diff
@@ -0,0 +1,182 @@
+{
+  "runOn": [
+    {
+      "minServerVersion": "4.3.x",
```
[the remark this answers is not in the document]
jmikola: In the meantime, perhaps we can add a comment for this in the YAML with a TODO item referencing the SERVER ticket we're waiting on. For example:

```
# TODO: Update once SERVER-??? is released
```

source/transactions/tests/retryable-abort-errorLabels.yml
```diff
@@ -93,7 +93,7 @@ tests:
       - name: abortTransaction
         object: session0
         result:
-          errorLabelsContain: ["RetryableWriteError"]
+          errorLabelsOmit: ["RetryableWriteError"]
```
jmikola: This relates to my earlier comment about `errorLabels: []` in the fail point suppressing the label for a code that is otherwise retryable. Perhaps we should add a comment after `[]` as well. Feel free to copy it around the various places where this applies.

source/retryable-writes/tests/insertOne-errorLabels.yml
```diff
@@ -13,32 +13,32 @@ tests:
...
-            result:
-                errorLabelsOmit: ["RetryableWriteError"]
-            collection:
-                data: []
+      failPoint:
+          configureFailPoint: failCommand
+          mode: { times: 1 }
+          data:
+              failCommands: ["insert"]
+              errorLabels: []
```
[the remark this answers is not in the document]
ldennis: Yes, if "errorLabels" field is specified, server will return exactly the same list of labels as specified regardless of error code (even if ok: 1). FailCommands takes absolute priority and completely overrides server normal behaviors.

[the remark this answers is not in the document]
egiurleo: Thanks, @ldennis! And thank you @jmikola for following up on this on the JIRA ticket. I'm going to add some comments to clarify the intention of those tests.

Jeremy, re: your thoughts -- I arrived at the same conclusion, so I appreciate the confirmation.

source/transactions/tests/retryable-abort-errorLabels.yml
egiurleo: Sorry to be clear -- are you asking me to add a comment after `errorLabels: []` or after the line that you've commented on here?

jmikola: A comment after `errorLabels: []` in the fail point would be helpful. This was referring to this bit of text from https://github.com/mongodb/specifications/pull/672#discussion_r338222564:

> Do you think it's worth having a comment after `[]` here to note that we're suppressing the label for an error code that would otherwise be considered retryable?

source/transactions/tests/error-labels.yml
```diff
@@ -408,7 +408,7 @@ tests:
       collection:
         data: []
 
-  - description: add unknown commit label to connection errors
+  - description: add retryable error and unknown commit label to connection errors
```
jmikola: Would it make sense to just use the actual label names in these descriptions? For example:

> Add RetryableWriteError and UnknownTransactionCommitResult labels to connection errors

If so, this might be a quick search/replace where "unknown commit label" and "transient label" exist in the `error-labels.yml` file's descriptions.

Either way, "label" should be pluralized in the description if we're now referring to two.

source/transactions/tests/retryable-abort-errorLabels.yml
egiurleo: I have added comments after all instances of `errorLabels: []` as well as comments clarifying the expectations of each test.

source/retryable-writes/retryable-writes.rst
```diff
@@ -66,9 +66,25 @@ ClientSession
...
+   An error is considered retryable if it has a RetryableWriteError label in
+   its top-level "errorLabels" field.
 
+   For server versions 4.4 and newer, MongoDB will add a RetryableWriteError label to
+   errors or server responses that it considers retryable before returning them to the
+   driver. As new server versions are released, the errors that are labeled with the
+   RetryableWriteError may change.
+
+   For all server versions that support retryable writes (MongoDB 3.6+), the driver
+   MUST add the RetryableWriteError label to an error or server response from a
```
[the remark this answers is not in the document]
egiurleo: Given our conversation in the meeting last week, I'm going to remove all instances of "not master" and "node is recovering" from the spec. I do like your suggestion on the structure of this section, though.

source/retryable-writes/retryable-writes.rst
```diff
@@ -66,9 +66,25 @@ ClientSession
...
 
 Retryable Error
-   An error is considered retryable if it meets any of the following criteria:
+   An error is considered retryable if it has a RetryableWriteError label in
+   its top-level "errorLabels" field.
 
+   For server versions 4.4 and newer, MongoDB will add a RetryableWriteError label to
+   errors or server responses that it considers retryable before returning them to the
+   driver. As new server versions are released, the errors that are labeled with the
+   RetryableWriteError may change.
```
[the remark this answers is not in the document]
egiurleo: Yes, good idea!

source/retryable-writes/retryable-writes.rst
```diff
@@ -66,9 +66,23 @@ ClientSession
...
+   a 4.4+ server that supports retryable writes, the driver MUST NOT add a
+   RetryableWriteError label to that error under any condition.
+
+   When the driver encounters a network error communicating with any server version
+   that supports retryable writes, it MUST add a RetryableWriteError label to that
+   error.
+
+   When receiving a command result with an error from a pre-4.4 server that supports
+   retryable writes, the driver MUST add a RetryableWriteError label to errors that meet
+   the following criteria:
```
egiurleo: @jmikola Let me know what you think of this section now. I removed all mentions of the "not master" and "node is recovering" messages because people mentioned in the meeting that those had been eliminated before 3.6, but let me know if you think I should add them back in the pre-4.4 section?

jmikola: I'm OK with this change, but I think other reviewers should sign off on it as well.

I might also suggest either calling this out in a new [Q&A entry](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/retryable-writes.rst#q-a) to explain why we're removing this. For example:

> Why did an earlier version of the spec parse errmsg to categorize retryable errors?

The answer could mention that the spec originally copied that behavior from SDAM. You can reference the [Use error messages to detect "not master" and "node is recovering"](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#use-error-messages-to-detect-not-master-and-node-is-recovering) section specifically and clarify that retryable writes is 3.6+ and we can safely rely on those servers to use consistent error codes. I believe that was the take-away from our meeting last week with server folks.

egiurleo: Okay, happy to do that.

source/retryable-writes/retryable-writes.rst
```diff
@@ -762,6 +775,9 @@ cluster (mongos just forwards the error from the shard's replica set).
 Changes
 =======
 
+2019-10-21: Change the definition of "retryable write" to be based on the
+RetryableWriteError label.
```
jmikola: You can also add a note that we're removing old behavior to determine retryable errors from parsing the errmsg. For example:

> Removed behavior for parsing errmsg to categorized retryable errors for pre-4.4 servers.

source/transactions/tests/error-labels.yml
```diff
@@ -698,7 +698,7 @@ tests:
         object: session0
         result:
           errorLabelsContain: ["UnknownTransactionCommitResult"]
-          errorLabelsOmit: ["TransientTransactionError"]
+          errorLabelsOmit: ["RetryableWriteError", "TransientTransactionError"]
```
jmikola: Just confirming, but this is a specific case where the commit result is unknown but we don't label it as retryable. Applications and the `withTransaction()` API are free to retry on their own per https://github.com/mongodb/specifications/blob/master/source/transactions-convenient-api/transactions-convenient-api.rst#the-commit-is-retried-after-a-write-concern-timeout-i-e-wtimeout-error. Is that correct?

jmikola: For the benefit of other reviewers, I'll point out that these tests correspond to https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#unknowntransactioncommitresult. Specifically:

> These errors codes mean that the provided write concern is not valid and therefore a retry attempt would fail with the same error.

That explains why we use the UnknownTransactionCommitResult label but _not_ RetryableWriteError.

source/transactions/tests/error-labels.yml
```diff
@@ -830,7 +830,7 @@ tests:
         data:
           - _id: 1
 
-  - description: omit unknown commit label to writeConcernError UnsatisfiableWriteConcern
+  - description: omit UnknownTransactionCommitResult label from writeConcernError UnsatisfiableWriteConcern
```
jmikola: For the benefit of other reviewers: I think the name is fine as-is, since this corresponds to https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#unknowntransactioncommitresult specifically instructing _not_ to add UnknownTransactionCommitResult to these errors. The fact that these tests also assert that RetryableWriteError and TransientTransactionError are not assigned is icing on the cake.

source/transactions/tests/retryable-abort-errorLabels.yml
```diff
@@ -0,0 +1,127 @@
...
+        # TODO: Update once SERVER-41245 is released
+        minServerVersion: "4.3.x"
+        topology: ["replicaset, sharded"]
+
+database_name: &database_name "transaction-tests"
+collection_name: &collection_name "test"
+
+data: []
+tests:
+  - description: abortTransaction only retries once after RetryableWriteError from server
```
jmikola: "with RetryableWriteError label from server" may read better. If you agree, feel free to make a similar change to other tests that used this original language. I think that's at least commitTransaction but I'm not sure if other tests copied as well.

source/transactions/tests/retryable-commit.yml
```diff
@@ -218,7 +218,7 @@ tests:
         object: session0
         result:
           errorCodeName: Interrupted
-          errorLabelsOmit: ["TransientTransactionError"]
+          errorLabelsOmit: ["RetryableWriteError", "TransientTransactionError"]
```
jmikola: Does it make sense to add UnknownTransactionCommitResult to the list of omitted labels? That may be similar to how you have RetryableWriteError and TransientTransactionError included in assertions for the UnknownTransactionCommitResult tests in `error-labels.yml`.

egiurleo: Yeah I think that makes sense! I added it in this test and one other.

source/retryable-writes/retryable-writes.rst
```diff
@@ -772,11 +772,30 @@ way to differentiate this error case from any other. The error code and errmsg
...
+used error codes to communicate error categories to the driver (see
+`Use error messages to detect "not master" and "node is recovering"`_).
+
+Retryable writes are only available on MongoDB versions 3.6 and newer. These
+server versions can be relied on to consistently return the proper error codes.
+Thus, it is sufficient that the driver check errors for the appropriate
+error codes or error labels to determine whether they are retryable without
+performing any error message parsing.
+
+_Use error messages to detect "not master" and "node is recovering": ..server-discovery-and-monitoring/server-discovery-and-monitoring.rst#use-error-messages-to-detect-not-master-and-node-is-recovering
```
jmikola: Should the path be "../server"? I think you may be missing a slash.

source/retryable-writes/retryable-writes.rst
```diff
@@ -66,9 +66,23 @@ ClientSession
    name of this object MAY vary across drivers.
 
 Retryable Error
-   An error is considered retryable if it meets any of the following criteria:
+   An error is considered retryable if it has a RetryableWriteError label in
+   its top-level "errorLabels" field.
 
-   - any network exception (e.g. socket timeout or error)
+   For server versions 4.4 and newer, MongoDB will add a RetryableWriteError label to
```
mbroadst: I have to say I'm finding it strange that behavior around what the driver should do wrt retryable errors is being conveyed in the "Terms" section. It seems fine to keep the first sentence here, as it falls under criteria for a retryable error, but I think the guidance around what drivers must do feels like it belongs in an "error handling" section. Thoughts?

jmikola: Perhaps we can add a new "Determining Retryable Errors" to section at the start of ["Implementing Retryable Writes"](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/retryable-writes.rst#implementing-retryable-writes). Then, the "Retryable Error" term can simply state:

> A network error or server response with a particular error code that warrants retrying a write operation. See _Determining Retryable Errors_ for more information.

mbroadst: SGTM

egiurleo: Matt, that's a good point. I'll add a new section. Thanks!

source/retryable-writes/tests/bulkWrite-errorLabels.yml
```diff
@@ -0,0 +1,78 @@
+runOn:
+    # TODO: Update once SERVER-41245 is released
```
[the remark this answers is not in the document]
egiurleo: Yup! Good catch.

source/retryable-writes/retryable-writes.rst
```diff
@@ -66,52 +66,9 @@ ClientSession
...
-
-   The criteria for retryable errors is similar to the discussion in the SDAM
-   spec's section on `Error Handling`_, but includes additional error codes. See
-   `What do the additional error codes mean?`_ for the reasoning behind these
-   additional errors.
-
-   .. _Error Handling: ../server-discovery-and-monitoring/server-discovery-and-monitoring.rst#error-handling
+   An error is considered retryable if it has a RetryableWriteError label in
+   its top-level "errorLabels" field. See `Determining Retryable Errors`_ for
+   more information.
```
egiurleo: @mbroadst Let me know what you think about this and the section below :)

mbroadst: LGTM!

PR #683, opened 2019-11-05
Update README.rst

lamontnelson:
Instructions on how to interpret PossiblePrimary for a multi-threaded implementation.

### DIFF
--- a/source/server-discovery-and-monitoring/tests/README.rst
+++ b/source/server-discovery-and-monitoring/tests/README.rst
@@ -56,7 +56,7 @@ so far. It has the following keys:
 A "server" object represents a correct ServerDescription within the client's
 current TopologyDescription. It has the following keys:
 
-- type: A ServerType name, like "RSSecondary". For multi-threaded implementations, the "PossiblePrimary" type should be interpreted as "RSUnknown".
+- type: A ServerType name, like "RSSecondary". For multi-threaded implementations, the "PossiblePrimary" type should be interpreted as "RSUnknown". See [ServerType](../server-discovery-and-monitoring.rst#servertype) for more details.
 - setName: A string with the expected replica set name, or null.
 - setVersion: absent or an integer.
 - electionId: absent, null, or an ObjectId.
--- a/source/server-discovery-and-monitoring/tests/README.rst
+++ b/source/server-discovery-and-monitoring/tests/README.rst
@@ -56,7 +56,7 @@ so far. It has the following keys:
 A "server" object represents a correct ServerDescription within the client's
 current TopologyDescription. It has the following keys:
 
-- type: A ServerType name, like "RSSecondary".
+- type: A ServerType name, like "RSSecondary". For multi-threaded implementations, the "PossiblePrimary" type should be interpreted as "RSUnknown".
 - setName: A string with the expected replica set name, or null.
 - setVersion: absent or an integer.
 - electionId: absent, null, or an ObjectId.
--- a/source/server-discovery-and-monitoring/tests/README.rst
+++ b/source/server-discovery-and-monitoring/tests/README.rst
@@ -56,7 +56,7 @@ so far. It has the following keys:
 A "server" object represents a correct ServerDescription within the client's
 current TopologyDescription. It has the following keys:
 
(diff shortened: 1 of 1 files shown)


p-mongo:
This is already defined in https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#servertype:

> Single-threaded clients use the PossiblePrimary type to maintain proper scanning order. Multi-threaded and asynchronous clients do not need this ServerType; it is synonymous with Unknown.

source/server-discovery-and-monitoring/tests/README.rst
```diff
@@ -56,7 +56,7 @@ so far. It has the following keys:
 A "server" object represents a correct ServerDescription within the client's
 current TopologyDescription. It has the following keys:
 
-- type: A ServerType name, like "RSSecondary".
+- type: A ServerType name, like "RSSecondary". For multi-threaded implementations, the "PossiblePrimary" type should be interpreted as "RSUnknown".
```
jmikola: I'll also point out that this section is mainly a description of the test format, and isn't the ideal place to include implementation details about SDAM itself.

ajdavis: I think this is a useful clarification of the test format itself.

lamontnelson: Yeah, when implementing it I was unclear whether I should ignore the cases with PossiblePrimary.

jmikola: I'd suggest adding the following to the end of the sentence:

```
(see: [ServerType](../server-discovery-and-monitoring.rst#servertype))
```

Cross-referencing it will at least point back to the canonical source of info for this. Also, it may be helpful to refer to "async and multi-threaded implementations".

p-mongo:
I think it is beneficial to have a single source of truth for each aspect of the specifications as much as is practical. To that end, I think hyperlinking to the definition of server type from the tests readme is a good idea, but duplicating the definition of server type is not something I am in agreement with.

jmikola:
> To that end, I think hyperlinking to the definition of server type from the tests readme is a good idea, but duplicating the definition of server type is not something I am in agreement with.

I'd be in favor of that. Perhaps:

> See _ServerType_ for details pertaining to async and multi-threaded drivers.

source/server-discovery-and-monitoring/tests/README.rst
mbroadst: @lamontnelson do you have an opinion on @jmikola s comment?

jmikola:
@lamontnelson Please either follow up on this PR or close it.

source/server-discovery-and-monitoring/tests/README.rst
lamontnelson: Sorry, this seems to have slipped through the cracks. I've added "async and multi-threaded implementations" and linked to the specification.

source/server-discovery-and-monitoring/tests/README.rst
```diff
@@ -56,7 +56,7 @@ so far. It has the following keys:
 A "server" object represents a correct ServerDescription within the client's
 current TopologyDescription. It has the following keys:
 
-- type: A ServerType name, like "RSSecondary".
+- type: A ServerType name, like "RSSecondary". For async and multi-threaded implementations, the "PossiblePrimary" type should be interpreted as "RSUnknown". See `ServerType <../server-discovery-and-monitoring.rst#servertype>`_ for more details.
```
jmikola, line 59: Suggestion per @p-mongo's comment in https://github.com/mongodb/specifications/pull/683#issuecomment-550047807.

```suggestion
- type: A ServerType name, like "RSSecondary". See `ServerType <../server-discovery-and-monitoring.rst#servertype>`_ for details pertaining to async and multi-threaded drivers.
```

I realize this PR originally used "RSUnknown", but I only see "Unknown" used in the spec itself. I'm not sure if that changed in the time since this PR was created, but if so it's probably all the more reason to refer back to the spec instead of duplicating spec language in the test README.

jmikola (changes requested):

source/server-discovery-and-monitoring/tests/README.rst
lamontnelson: Thanks, I like this change. This should be ready to go now. Better late than never!

nbbeeken (approved):

PR #686, opened 2019-11-12
Some small FLE grammar/typo fixes

egiurleo:
Let me know if this warrants a version bump/entry to the changelog (I don't think it does, but I'll defer to reviewers).

### DIFF
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -630,7 +630,7 @@ redirect to /dev/null). Users can pass the argument --logpath to
 extraOptions.mongocryptdSpawnArgs if they need to inspect mongocryptd
 logs.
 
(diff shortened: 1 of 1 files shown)


source/client-side-encryption/client-side-encryption.rst
```diff
@@ -630,7 +630,7 @@ redirect to /dev/null). Users can pass the argument --logpath to
 extraOptions.mongocryptdSpawnArgs if they need to inspect mongocryptd
 logs.
 
-Upon construction, the ClientEncryption MUST create a MongoClient to
+Upon construction, the MongoClient MUST create a MongoClient to
```
egiurleo: I believe this is a typo, but correct me if I'm wrong.

kevinAlbs: Oh good catch! ClientEncryption does not do automatic encryption, so it does not need to connect to or spawn mongocryptd.

PR #687, opened 2019-11-13
feat: support wrapping libraries in client metadata generation

mbroadst:
A section is added to support libraries which wrap MongoDB drivers
providing their own metadata to amend the metadata generated by
the driver for its handshake.

SPEC-1502


source/mongodb-handshake/handshake.rst
```diff
@@ -224,20 +224,57 @@ Example::
         - "Oracle JVM EE 9.1.1"
 
 
-Limitations
+Supporting Wrapping Libraries
```
jmikola: `===...` below needs to be extended for a longer title.

source/mongodb-handshake/handshake.rst
```diff
@@ -224,20 +224,57 @@ Example::
...
 ===========
 
-The entire metadata BSON document MUST NOT exceed 512 bytes. This includes all
-BSON overhead.  The ``client.application.name`` cannot exceed 128 bytes.  MongoDB
-will return an error if these limits are not adhered to, which will result in
-handshake failure. Drivers MUST validate these values and truncate driver
-provided values if necessary.
+Drivers MUST allow libraries which wrap the driver to amend the client metadata
```
jmikola: Quoting a basic definition of "amend":

> make minor changes in (a text) in order to make it fairer, more accurate, or more up-to-date.

I think "append to" would be more appropriate, as @behackett clarified in [SPEC-1502](https://jira.mongodb.org/browse/SPEC-1502) that we don't want any higher level libraries to overwrite anything that would normally be set by the driver itself.

source/mongodb-handshake/handshake.rst
```diff
@@ -224,20 +224,57 @@ Example::
...
 
+    /**
+     * Optional platform information for the wrapping driver.
+     */
+    platform: Optional<String>;
+  }
 
+Note that how these options are provided to a driver is left up to the implementor.
 
+If provided, these options MUST not replace the values used for metadata generation, but
```
behackett: All caps NOT

source/mongodb-handshake/handshake.rst
mbroadst: done

source/mongodb-handshake/handshake.rst
mbroadst: done

source/mongodb-handshake/handshake.rst
```diff
@@ -224,6 +224,49 @@ Example::
...
+
+If provided, these options MUST not replace the values used for metadata generation, but
+instead append to the respective fields, delimited by a `|` character. For example, when
+`Motor <https://docs.mongodb.com/ecosystem/drivers/motor/>`_ wraps PyMongo, the following
+fields are updated to include Motors "driver info":
+
+    {
+        client: {
+            driver: {
+                name: "nodejs|motor",
```
behackett: PyMongo|Motor is the actual value.

source/mongodb-handshake/handshake.rst
```diff
@@ -224,6 +224,49 @@ Example::
...
+                version: "2.6.1|2.0.0"
```
behackett: 3.6.0|2.0.0

source/mongodb-handshake/handshake.rst
```diff
@@ -224,6 +224,49 @@ Example::
...
+     */
+    platform: Optional<String>;
+  }
+
+Note that how these options are provided to a driver is left up to the implementor.
+
+If provided, these options MUST not replace the values used for metadata generation, but
+instead append to the respective fields, delimited by a `|` character. For example, when
+`Motor <https://docs.mongodb.com/ecosystem/drivers/motor/>`_ wraps PyMongo, the following
+fields are updated to include Motors "driver info":
```
jmikola: To clarify, would this entail changing existing behavior in any driver that already allows handshake fields to be appended and uses a different delimiter? For instance, libmongoc's `mongoc_handshake_data_append` uses ` / ` (with surrounding spaces).

I assume the MUST at the start of this sentence (prohibiting replacing existing data) still applies to this, but it may be clear if we split the sentence in two. For example:

> Additional options MUST be appended to the respective fields and be delimited by a `|` character.

Also, it may be prudent to discuss priorities should part of the handshake need to be truncated for space limitations. This probably belongs in its own section, but it seems relevant to this PR since we're introducing new API to append more data.

To use libmongoc's existing [`mongoc_handshake_append()`](https://github.com/mongodb/mongo-c-driver/blob/1.15.1/src/libmongoc/src/mongoc/mongoc-handshake.c#L609) function as an example, it allows appending to the driver name, version, and platform fields. The only truncation is applied to the "platform" field in [`_mongoc_handshake_build_doc_with_application()`](https://github.com/mongodb/mongo-c-driver/blob/1.15.1/src/libmongoc/src/mongoc/mongoc-handshake.c#L509), and we don't allow other fields be truncated. I assume the logic there is that platform is a nice-to-have field and is most likely to overflow (libmongoc tends to stuff CFLAGS in there), but if we had to truncate the driver name or other essential fields the handshake data itself wouldn't be too useful so we just fail completely rather than truncate.

source/mongodb-handshake/handshake.rst
```diff
@@ -233,12 +276,6 @@ will return an error if these limits are not adhered to, which will result in
 handshake failure. Drivers MUST validate these values and truncate driver
```
behackett: I think we should encourage driver devs to truncate the _platform_ field only. The driver name and version fields are too high value to truncate. Encourage implementers to only put high value information in the field (done't dump compiler and linker flags there) and place the highest value data at the beginning of the string.

source/mongodb-handshake/handshake.rst
jmikola: I assume Natalya is OK exploding these two strings on the delimiter character and matching them up? I suppose the only question on this would be what happens if another driver name is appended without a matching version (since everything in DriverInfoOptions is optional).

I'm not sure how this is any better a single field containing "foo-X.Y.Z|bar-X.Y.Z", but the original spec obviously didn't anticipate what we're trying to accomplish here and it's likely too late to change the schema.

source/mongodb-handshake/handshake.rst
jmikola: Relates to my comment in https://github.com/mongodb/specifications/pull/687#discussion_r345938767.

source/mongodb-handshake/handshake.rst
```diff
@@ -224,6 +224,49 @@ Example::
...
 
 
+Supporting Wrapping Libraries
+=============================
+
+Drivers MUST allow libraries which wrap the driver to append to the client
+metadata generated by the driver. The following class definition defines the
+options which MUST be supported:
+
+.. code:: typescript
```
ShaneHarvey: Not sure why but this code block does not render correctly (see the rendered diff).

source/mongodb-handshake/handshake.rst
```diff
@@ -224,6 +224,49 @@ Example::
...
+        client: {
+            driver: {
+                name: "nodejs|motor",
+                version: "2.6.1|2.0.0"
+            }
+        }
+    }
+
+
 Limitations
```
ShaneHarvey: Should drivers perform some validation on the DriverInfo fields to avoid hitting these size limits?

source/mongodb-handshake/handshake.rst
```diff
@@ -224,6 +224,49 @@ Example::
...
+metadata generated by the driver. The following class definition defines the
+options which MUST be supported:
+
+.. code:: typescript
+
+  class DriverInfoOptions {
+    /**
+     * The name of the library wrapping the driver.
+     */
+    name: Optional<String>;
```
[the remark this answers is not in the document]
jmikola: This will address my comment in https://github.com/mongodb/specifications/pull/687#discussion_r345940014 about possible issues matching up the exploded, delimited values if one was omitted. I might also suggest requiring them to be non-empty strings and also not contain the delimiter character.

source/mongodb-handshake/handshake.rst
```diff
@@ -224,6 +224,49 @@ Example::
...
+Supporting Wrapping Libraries
+=============================
+
+Drivers MUST allow libraries which wrap the driver to append to the client
+metadata generated by the driver. The following class definition defines the
+options which MUST be supported:
+
+.. code:: typescript
+
+  class DriverInfoOptions {
```
jyemin: Can we make the name of the class non-normative?  Java driver already defines it as MongoDriverInformation and it will be disruptive to change it now.

source/mongodb-handshake/handshake.rst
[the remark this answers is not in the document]
jyemin: I agree about name, but some wrappers may not be able to report their version.  I don't think Morphia, for example, has a concept of its own version that is accessible at runtime.

source/mongodb-handshake/handshake.rst
jyemin: She's already handling this in her scripts, including the lack of a matching version

source/mongodb-handshake/handshake.rst
jyemin: What if there are multiple levels of wrapping?  For example, in Java the Scala driver wraps the reactive streams driver, which wraps the core driver.  The Java driver handles that by making each of these fields by a list rather than a single value.  We can't realistically remove that from the Java driver, and it may be useful for others as well, e.g. maybe PyModm will wrap Motor some day, idk.

source/mongodb-handshake/handshake.rst
mbroadst: I think I fixed it.

source/mongodb-handshake/handshake.rst
mbroadst: fixed

source/mongodb-handshake/handshake.rst
mbroadst: fixed

source/mongodb-handshake/handshake.rst
mbroadst: fixed

source/mongodb-handshake/handshake.rst
mbroadst: I'm not sure we can, other than providing guidance around what fields to prioritize? Would we want to put size limits on actual fields?

source/mongodb-handshake/handshake.rst
mbroadst: added a section about this

source/mongodb-handshake/handshake.rst
mbroadst: I didn't assume it was, I thought most of our specs fall under the "acceptable deviations" clause of CRUD or SDAM. Do you want me to just omit the name completely? The language doesn't require you even to put it in a class, rather it says you need to support these options.

source/mongodb-handshake/handshake.rst
mbroadst: I'd lean towards "defy augury" here, but also because I'm not really sure I could imagine what that would look like. Maybe it makes sense if we owned all the component pieces, but out in the community you would have to plumb down support for this all the way through the stack (Wrapper A wraps Driver and provides its own API for things that wrap _it_, etc). Who ultimately appends to the list in these options? I would imagine a wrapping driver is not interested in exposing access to its internal `MongoClient`, but I might just be making a bad assumption here.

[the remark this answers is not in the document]
mbroadst: Made `name` non-optional

jyemin: Oh, augury...  Java driver may be special here in terms of the number of layers that are all controlled by us.  So long as you word it such that the Java driver is not violating the spec doing what it's currently doing, I'm ok with it.

source/mongodb-handshake/handshake.rst
jyemin: Whatever works.  I think most specs that define an API repeat the acceptable deviations section.

source/mongodb-handshake/handshake.rst
mbroadst: I updated the section on truncating below, and split the sentence here. What do you propose we do about existing deviations in choice of delimiter?

source/mongodb-handshake/handshake.rst
jmikola: > We can't realistically remove that from the Java driver, and it may be useful for others as well, e.g. maybe PyModm will wrap Motor some day, idk.

libmongoc exposes an API that allows whatever sits atop it to append a driver name/version. In turn, I would expect the PHP driver to provide an API to whatever sits atop it to append to the driver name/version. The PHP driver would just ensure that its own name/version ends up appearing before the next higher-level thing, and utilize string concatenation rather than tracking lists of things.

I imagine that's what Matt might have been hinting at with "plumb down support for this all the way through the stack", but I agree however Java currently handles this is fine provided its API accomplishes the same outcome.

Given that some drivers already have existing APIs to do this (e.g. PyMongo, libmongoc, Ruby), perhaps we can allow utmost flexibility and just propose DriverInfoOptions as one possible API. The MUST and MUST NOTs can then just refer to rules for assembling the driver name/version and platform strings.

behackett: I'd like to see Ruby switch to something closer to what we're proposing here. Dumping all wrapper information into the platform field is a bit of a mess.

source/mongodb-handshake/handshake.rst
mbroadst: I added a section called "Deviations" where I address this and other deviations brought up here, PTAL. /cc @jyemin

source/mongodb-handshake/handshake.rst
mbroadst: See my comment above, but I added a "Deviations" section to address this

source/mongodb-handshake/handshake.rst
mbroadst: I made a "Deviations" section that accommodates this, while still proposing an opinionated default for drivers who haven't done this yet.

jmikola: > I'd like to see Ruby switch to something closer to what we're proposing here. Dumping all wrapper information into the platform field is a bit of a mess.

@behackett: I think there's a misunderstanding here. I was only referring to Ruby possibly having an API to append data for a higher level library (e.g. Mongoid), which may be something to keep instead of moving over to DriverInfoOptions.

I was _not_ suggesting that Ruby continue to append Mongoid details in the platform field.

source/mongodb-handshake/handshake.rst
ShaneHarvey: I'm not for this idea anymore. It would be complicated to compute the actual size limit based on the 512 byte restriction.

source/mongodb-handshake/handshake.rst
```diff
@@ -224,20 +224,75 @@ Example::
...
+        platform: Optional<String>;
+    }
+
+
+Note that how these options are provided to a driver is left up to the implementor.
+
+If provided, these options MUST NOT replace the values used for metadata generation.
+The provided options MUST be appended to their respective fields, and be delimited by
+a ``|`` character. For example, when `Motor <https://docs.mongodb.com/ecosystem/drivers/motor/>`_
+wraps PyMongo, the following fields are updated to include Motors "driver info":
```
jmikola: `Motor's`, possessive

source/mongodb-handshake/handshake.rst
```diff
@@ -224,20 +224,75 @@ Example::
...
+        */
+        platform: Optional<String>;
+    }
+
+
+Note that how these options are provided to a driver is left up to the implementor.
+
+If provided, these options MUST NOT replace the values used for metadata generation.
+The provided options MUST be appended to their respective fields, and be delimited by
+a ``|`` character. For example, when `Motor <https://docs.mongodb.com/ecosystem/drivers/motor/>`_
```
jmikola: Might have been missed from a previous comment I made, but should we prohibit the three options from containing a delimiter character (noting below that you don't require a `|` and allow drivers to keep whatever they may already be using)? Alternatively, we can ignore this edge case.

I'm OK with either and just wanted to ensure that this question was considered.

mbroadst: added a note requiring validation of this

mbroadst: /cc @ShaneHarvey @jyemin, you both already approved but I added another `MUST NOT` here and `MUST throw an error`, which may be contentious

ShaneHarvey: Sounds good to me.

jyemin: Me too.

In the future don't force push.  Makes it hard to see diffs after initial review.  You can squash before merging.  :)

PR #689, opened 2019-11-19
SPEC-1481: Clarify integer sizes in URI options spec

alcaeus:
This clarifies that all integer options are 32-bit unless otherwise specified. The only exception is the wTimeoutMS option, which is a 64-bit value according to the [Read and Write Concern](https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#write-concern) spec.

### DIFF
--- a/source/uri-options/uri-options.rst
+++ b/source/uri-options/uri-options.rst
@@ -61,8 +61,9 @@ as optional. If an option is marked as optional, a driver MUST meet any
 conditions specified for leaving it out if it is not included. If a driver
 already provides the option under a different name, the driver MAY
 implement the old and new names as aliases. All keys and values MUST be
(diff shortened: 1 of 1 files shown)


alcaeus:
> It's honestly a bit strange to me that wTimeoutMS is required to be a 64-bit integer, since a 32-bit integer would allow up to nearly 25 _days_ of timeout, which seems plenty, but I'll defer to the read/write concern spec

FWIW, I agree with you on this. For the sake of consistency, it would be sensible to allow 64-bit integers to be passed through URI options instead of having to set these values through a different API. Changing `wTimeoutMS` in the Write and Read Concern spec to 32-bit could be an alternative, but I'm not sure of the implications.

PR #690, opened 2019-11-20
SPEC-1243: GridFS index checking should support indexes created in the shell.

rstam:


rstam:
@jyemin let me know who else should review this.

source/gridfs/gridfs-spec.rst
```diff
@@ -396,6 +396,18 @@ than checking the count in the case where the files collection is sharded.
 Drivers MUST check whether the indexes already exist before attempting to create them.
 This supports the scenario where an application is running with read-only authorizations.
 
+NOTE: when checking whether an index exists drivers MUST compare numeric values by value
```
jyemin: Nit:  The "NOTE" prefix is superfluous.  It reads just fine without it and makes it clear that there's a requirement in there and it's not just explanatory text.

rstam:
Squashed, rebased and merged.

PR #691, opened 2019-11-20
SPEC-1342: Amend database enumeration spec to include "authorizedDatabases" option.

rstam:


source/enumerate-databases.rst
```diff
@@ -81,6 +81,22 @@ For example, to list only databases whose names begin with "foo":
 
   > db.adminCommand({listDatabases: 1, filter: {name: /^foo/}});
 
+AuthorizedDatabases
+-------------------
+
+MongoDB 4.0.5 added an ``authorizedDatabases`` boolean option to the `listDatabases`_
+database command, which can be used to limit the command result to only include databases
+the user is authorized to use. Drivers SHOULD support the new ``authorizedDatabases``
```
jyemin: Generally SHOULD should be avoided, but it's consistent with the requirements for filter, so I'll go along with it.

source/enumerate-databases.rst
```diff
@@ -81,6 +81,22 @@ For example, to list only databases whose names begin with "foo":
 
   > db.adminCommand({listDatabases: 1, filter: {name: /^foo/}});
 
+AuthorizedDatabases
```
jmikola: I think this can be "Authorized Databases". I don't believe the section title maps to the option name. For instance, we have "Filters" for the previous section even though the option is `filter` (singular).

rstam:
Squashed, rebased and merged.

PR #692, opened 2019-11-20
SPEC-1418: Add support for allowDiskUse as a find() parameter.

rstam:


source/crud/tests/v2/find-allowdiskuse.yml
```diff
@@ -0,0 +1,50 @@
+runOn:
```
rstam: Until someone writes a POC we won't be sure these tests are correct.

source/crud/tests/v2/find-allowdiskuse.yml
```diff
@@ -0,0 +1,50 @@
...
+        object: collection
+        name: find
+        arguments:
+          filter: { }
+    expectations:
+      -
+        command_started_event:
+          command:
+            find: *collection_name
+            allowDiskUse:
```
rstam: The intended meaning of this line is that `allowDiskUse` is NOT present in the command.

source/crud/tests/v2/find-allowdiskuse.yml
```diff
@@ -0,0 +1,50 @@
...
+  -
+    description: "Find sends allowDiskuse false"
```
jyemin: Since we're not testing the "true" case, consider renaming to "Find sends allowDiskUse when specified"

source/crud/crud.rst
```diff
@@ -407,6 +407,17 @@ Read
 
   class FindOptions {
 
+    /**
+     * Enables writing to temporary files. When set to true, the find operation
```
jyemin: Consider making it clearer that it's the _server_ that's doing this, not the client.

source/crud/tests/v2/find-allowdiskuse.yml
jmikola: Sorry to tack on work, but I believe this is the first time we're asserting the absence of a field in a command body within the CRUD spec test suite. It may require some additional clarification in the CRUD test README.

The CRUD spec README currently only refers to the Command Monitoring spec when we talk about the structure of the `command_started_event` field. CRUD's section for [Evaluating Matches](https://github.com/mongodb/specifications/tree/master/source/crud/tests#evaluating-matches) doesn't discuss any special handling for `null` fields in a `result` or `command_started_event` field within the test. As-is, a CRUD test runner would likely assert that the command contains an `allowDiskUse` field with a value of `null`; rather than the desired outcome of asserting that the command document does not contain an `allowDiskUse` field.

I think this warrants adding a section to the CRUD test README that instructs test runners to use matching logic from the Transaction spec test runner for _events_ ([Null Values](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst#null-values) in particular). For prior art, you might just copy the section from the [Transactions Convenient API](https://github.com/mongodb/specifications/blob/master/source/transactions-convenient-api/tests/README.rst#command-started-events)) test runner.

----

Alternatively, we could remove this test case and punt on this until after [Unified test format](https://jira.mongodb.org/browse/DRIVERS-709) is implemented, since that should address the inconsistent handling of `command_started_event` assertions throughout various specs.

source/crud/tests/v2/find-allowdiskuse.yml
jyemin: The server work is done and available on "latest" so I think it makes sense for you to do it in the .NET driver.  It seems straightforward.

source/crud/tests/v2/find-allowdiskuse.yml
rstam: Since we've used `{ field : null }` before (e.g. in the transactions tests) to mean that "field" is not present, let's use that for now.

When we do the unified test format that effort should standardize how a JSON test specifies that a field is NOT present.

I think we should have some meta notation like:

    { field : { $exists : false } }
    { field : { $exists : true } }

but that will be decided as part of the Unified Test Format work.

jmikola: I'm fine with keeping `null` here, but I'd still request we add clarification to the test README so this is clearly communicated to implementing drivers. See my "I think this warrants adding a section..." paragraph in my last comment. There should be some existing text you can copy in.

jyemin: There is: https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst#null-values

source/crud/crud.rst
rstam: Done

source/crud/tests/v2/find-allowdiskuse.yml
rstam: Will do

source/crud/tests/v2/find-allowdiskuse.yml
rstam: I've added some text about this to `tests/README.RST`

source/crud/tests/v2/find-allowdiskuse.yml
rstam: This is testing the false case. The true case IS tested below.

I've reworded slightly to be more explicit.

rstam:
I'm beginning a POC using the C# driver. If that uncovers the need to make any changes (in particular to the tests) I will add a new commit here.

source/crud/crud.rst
```diff
@@ -11,8 +11,8 @@ Driver CRUD API
 :Advisors: Jeremy Mikola, Jeff Yemin
 :Status: Approved
 :Type: Standards
-:Minimum Server Version: 2.6
-:Last Modified: October 28, 2019
+:Minimum Server Version: 2.7
+:Last Modified: November 20, 2019
```
jyemin: Nit: update this to whatever date it happens to be when the PR is merged.

source/crud/tests/v2/find-allowdiskuse.yml
jyemin: From my perspective, this thread can now be resolved.

source/crud/crud.rst
```diff
@@ -407,6 +407,17 @@ Read
 
   class FindOptions {
 
+    /**
+     * Enables writing to temporary files on the server. When set to true, the server
+     * can write temporary data to disk while executing the find operation.
+     *
+     * This option is sent only if the caller explicitly provides a value. The default
+     * is to not send a value.
```
jmikola: I didn't consider this in my original review weeks ago, but in light of [SPEC-1542](https://jira.mongodb.org/browse/SPEC-1542) (see: [this change](https://github.com/mongodb/specifications/commit/43f595ab4bb1594629d2a4f7a9041777781d8b64#diff-5efd2e4ce61f4ccda0ce031099303996R1818)) I think we should specifying what happens with this option in an unsupported server version. I think we have two options:

 * Add "For servers < 3.6, the driver MUST raise an error if the caller explicitly provides a value." This is because 3.6+ servers report errors for unknown command options. Raising a client-side error covers us for earlier server versions using the `find` command as well as pre-3.2 servers using OP_QUERY that have no way to express this option. 
 * Add a note that this option will be ignored for server versions < 3.6. This may entail adapting some text from the `maxAwaitTimeMS` doc block: "For servers < 3.2, this option is ignored and not sent as allowDiskUse does not exist in the OP_QUERY wire protocol." However, we'd still need to clarify that server versions >=3.2 and <3.6 will silently ignore the command option.

@jyemin might additionally request a test case for the client-side error reporting, as in https://github.com/mongodb/specifications/pull/650#pullrequestreview-332539817. If so, that could be a new test file similar to `find-allowdiskuse.yml` but with `error: true` included in each test case's `operation` block. Support for `allowDiskUse` was added in 4.3.1 by [SERVER-42077](https://jira.mongodb.org/browse/SERVER-42077), so we could either use "4.3.0" or something in the 3.4-range (e.g. "3.4.99") depending on whether we want to assert any error (server- and client-side) or just client-side (before 3.6 introduced strict checks for unknown options).

source/crud/crud.rst
```diff
@@ -11,8 +11,8 @@ Driver CRUD API
 :Advisors: Jeremy Mikola, Jeff Yemin
 :Status: Approved
 :Type: Standards
-:Minimum Server Version: 2.6
-:Last Modified: October 28, 2019
+:Minimum Server Version: 2.7
```
jmikola: I don't think this line should be modified. If we do need to raise this value, such a change likely belongs in its own ticket.

source/crud/tests/v2/find-allowdiskuse.yml
```diff
@@ -0,0 +1,50 @@
+runOn:
+  - { minServerVersion: "4.4.0" }
```
jmikola: [SERVER-42077](https://jira.mongodb.org/browse/SERVER-42077) was released in 4.3.1, so we should use that version here. That will also make it easier for drivers to run these tests without any manual modifications (or worse, waiting until 4.4.0 is actually released). This would be consistent with many other recent spec tests.

source/crud/crud.rst
jyemin: allowDiskUse is similar to hint, isn't it?  Neither should change the result of the operation, but both can potentially have a large affect on the execution.  For allowDiskUse, it can mean that some executions will fail due to hitting the limit for in-memory sorts. For hint, it can mean that some executions may fail due to, say, a maxTimeExceeded error caused by the use of the wrong index.

So let's be consistent with hint and allowDiskUse.

source/crud/crud.rst
rstam: Good catch. I thought I was incrementing the document version!

source/crud/crud.rst
rstam: If I understand you correctly treating allowDiskUse consistently with how we treat hint means no further changes are needed. We only send allowDiskUse if the application specifies a value and we don't check server version.

source/crud/crud.rst
rstam: Done

source/crud/crud.rst
rstam: Done

source/crud/crud.rst
rstam: Done?

source/crud/tests/v2/find-allowdiskuse.yml
rstam: Done

source/crud/crud.rst
jyemin: @jmikola may need to correct me, but as I understand it, it means that a driver MUST throw a client side error if allowDiskUse is specified and server version is less than ... 4.4? or less than whatever server version started errors on unrecognized options.  At minimum it would mean throwing if server version is less than 3.2, since the old wire protocol certainly doesn't allow even trying to send allowDiskUse.

rstam: That's the exact opposite of what I thought you were saying.

I think AllowDiskUse can (and should) be safely ignored if a server doesn't support it.

Its presence or absence NEVER affects the correctness of the results. In other words, you're going to get the exact same result set whether you specify AllowDiskUse or not (assuming sufficient memory).

rstam: Here's the full matrix of scenarios as I see them:

1. Application does not set AllowDiskUse: nothing special to consider
2. Application sets AllowDiskUse and server ignores it: not a problem
3. Application sets AllowDiskUse and server returns an error because it doesn't recognize it: not a problem
4. Application sets AllowDiskUse and server ignores it: not a problem
5. Application sets AllowDiskUse and server supports it: intended behavior

jmikola: > allowDiskUse is similar to hint, isn't it? Neither should change the result of the operation if the operation succeeds, but both can potentially have a large affect on the execution.

@jyemin: I would not expect `allowDiskUse` to ever change the result of an operation, but I believe a `hint` may if it was used in conjunction with a sparse index since that could limit the pool of potential documents to match the query (an edge case I'll admit). 

----

> it means that a driver MUST throw a client side error if allowDiskUse is specified and server version is less than ... 4.4? or less than whatever server version started errors on unrecognized options. At minimum it would mean throwing if server version is less than 3.2, since the old wire protocol certainly doesn't allow even trying to send allowDiskUse.

I did some quick testing and it looks like server 3.2.21 raises an error for unrecognized options to the `find` command.

```
> db.runCommand({find:'foo',what:1})
{
	"waitedMS" : NumberLong(0),
	"ok" : 0,
	"errmsg" : "Failed to parse: { find: \"foo\", what: 1.0 }. Unrecognized field 'what'.",
	"code" : 9
}
```

Given that, I think we'd only need to throw client-side errors if the option would be used for a server version < 3.2 since that entails using OP_QUERY in the legacy manner without sending a `find` command.

Unfortunately, the server is inconsistent when it comes to reporting errors for unrecognized command options. It appears that validation for `find` was implemented when the command was introduced in 3.2, but as I discovered in [SPEC-1334](https://jira.mongodb.org/browse/SPEC-1334) validation for `update` didn't come until 3.4. Validation for `findAndModify` options was only introduced in 4.1.10 (see https://github.com/mongodb/specifications/pull/650#discussion_r328676881).

----

> Here's the full matrix of scenarios as I see them: ...

@rstam: Your list of scenarios above includes "Application sets AllowDiskUse and server ignores it: not a problem" twice. Was that a typo?

I agree that there's no need to test for the server returning an error if the option is unsupported (i.e. server versions >= [3.1.1](https://jira.mongodb.org/browse/SERVER-17282) and < [4.3.1](https://jira.mongodb.org/browse/SERVER-42077)). 

Per my original comment, I think raising an error is most consistent with other options we recently introduced, but I'm still OK with allowing it to be silently ignored if there's group preference for that. If we make a test, I'd propose using 3.1.0 as the `maxServerVersion`, since we know that's the last version of the server before the `find` command was introduced (alternatively, some suitably large 3.0.x version like 3.0.99 would suffice).

jmikola: > I'm still OK with allowing it to be silently ignored if there's group preference for that

To better clarify my last comment, if @jyemin agrees that `allowDiskUse` doesn't affect the result of a query (unlike `hint`, `collation`, and others) and is happy to ignore it when unsupported, I'll go along with that.

Going back to my first comment in this thread, "silently ignoring" is what we do for `maxAwaitTimeMS` before the `getMore` command, so one could argue this case is not much different -- both options affect _how_ the queries execute but not what documents they match.

jyemin: Seems like we're trading one error (client-side error that allowDiskUse not supported) for another that may happen only some times (hit 100 MB in-memory sort limit).  Especially as 3.0 is long since EOLed I'm fine with this outcome.

source/crud/tests/v2/find-allowdiskuse.yml
rstam: Done.

https://github.com/rstam/mongo-csharp-driver/pull/141

source/gridfs/gridfs-spec.rst
```diff
@@ -59,8 +59,8 @@ Chunk
   The default size for the data field in chunks is 255 KiB. Chunk documents have the following form:
 
   .. code:: javascript
-  
-    { 
+
+    {
```
rstam: Sorry for all the whitespace changes. The editor automatically removed all blanks at the end of a line (that's a good thing).

source/gridfs/gridfs-spec.rst
```diff
@@ -727,47 +727,58 @@ Generic Find on Files Collection
...
+    /**
+     * Enables writing to temporary files on the server. When set to true, the server
+     * can write temporary data to disk while executing the find operation on the files collection.
+     *
+     * This option is sent only if the caller explicitly provides a value. The default
+     * is to not send a value.
+     *
+     * @see https://docs.mongodb.com/manual/reference/command/find/
+     */
+    allowDiskUse: Optional<Boolean>;
```
jmikola: Is this change actually necessary?

IIRC, `Bucket::find()` is just a proxy for a `find()` operation on the `fs.files` collection and opens a cursor on the raw documents (nothing to do with opening download streams).

I don't recall why we defined GridFSFindOptions separately instead of supporting FindOptions in the CRUD spec, but the underlying operation should be the same. GridFSFindOptions was also never updated to support options such as `collation`. Was that intentional or just an oversight?

I might propose removing this and just creating a new SPEC ticket to consolidate and just reference FindOptions  and the CRUD spec unless there's good reason to keep them separate.

source/crud/crud.rst
jmikola: @rstam: Since it seems we've settled on not enforcing any client-side error, I would propose we added the following language (akin to what we have for `maxAwaitTimeMS`):

> For servers < 3.2, this option is ignored and not sent as allowDiskUse does not exist in the OP_QUERY wire protocol.

I don't think we need to mention that this is only supported in MongoDB 4.4+, as the doc block for `collation` doesn't talk about when support was introduced (only when we want to raise client-side errors).

rstam:
Squashed, rebased and pushed to master.

source/gridfs/gridfs-spec.rst
jmikola: Opened https://jira.mongodb.org/browse/SPEC-1549 to track this.

PR #694, opened 2019-12-03
Add final draft of UUID specification

jyemin:
This is the final version of the UUID specification.  It's copied from Google Docs, with a few small edits to fix the RST content hierarchy.

### DIFF
--- a/source/uuid.rst
+++ b/source/uuid.rst
@@ -1,0 +1,321 @@
+=============================
+Handling of Native UUID Types
+=============================
+
+:Spec Title: Handling of Native UUID Types
+:Spec Version: 1.0
+:Author: Jeff Yemin
+:Lead: Bernie Hackett
+:Advisory Group: Shane Harvey, Oleg Pudeyev, Robert Stam
+:Informed: drivers@
+:Status: Accepted
+:Type: Standards
+:Last Modified: 2019-11-19
+
+.. contents::
+
+--------
+
+Abstract
+========
+
+The Java, C#, and Python drivers natively support platform types for UUID, all of which by default encode them to and decode them from BSON binary subtype 3.  However, each encode the bytes in a different order from the others. To improve interoperability, BSON binary subtype 4 was introduced and defined the byte order according to `RFC 4122 <https://tools.ietf.org/html/rfc4122#section-4.1.2>`_, and a mechanism to configure each driver to encode UUIDs this way was added to each driver. The legacy representation remained as the default for each driver.
+ 
+This specification moves MongoDB drivers further towards the standard UUID representation by requiring an application relying on native UUID support to explicitly specify the representation it requires.
+
+Drivers that support native UUID types will additionally create helpers on their BsonBinary class that will aid in conversion to and from the platform native UUID type.
+
+META
+====
+
+The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this
+document are to be interpreted as described in
+`RFC 2119 <https://www.ietf.org/rfc/rfc2119.txt>`_.
+
+Specification
+=============
+
+Terms
+-----
+
+UUID
+    A Universally Unique IDentifier
+
+BsonBinary
+   An object that wraps an instance of a BSON binary value 
+
+Naming Deviations
+-----------------
+
+All drivers MUST name operations, objects, and parameters as defined in the following sections.
+
+The following deviations are permitted:
+
+* Drivers can use the platform's name for a UUID.  For instance, in C# the platform class is Guid, whereas in Java it is UUID.
+* Drivers can use a "to" prefix instead of an "as" prefix for the BsonBinary method names.
+
+Explicit encoding and decoding
+------------------------------
+
+Any driver with a native UUID type MUST add the following UuidRepresentation enumeration, and associated methods to its BsonBinary (or equivalent) class:
+
+.. code:: typescript
+   
+   /**
+           
+   enum UuidRepresentation {
+
+      /**
+       * An unspecified representation of UUID.  Essentially, this is the null 
+       * representation value. This value is not required for languages that      
+       * have better ways of indicating, or preventing use of, a null value.
+       */
+      UNSPECIFIED("unspecified"),
+
+      /**
+       * The canonical representation of UUID according to RFC 4122, 
+       * section 4.1.2 
+       * 
+       * It encodes as BSON binary subtype 4
+      */
+      STANDARD("standard"),
+
+      /**
+       * The legacy representation of UUID used by the C# driver.
+       *
+       * In this representation the order of bytes 0-3 are reversed, the 
+       * order of bytes 4-5 are reversed, and the order of bytes 6-7 are 
+       * reversed.
+       *
+       * It encodes as BSON binary subtype 3
+       */
+      C_SHARP_LEGACY("csharpLegacy"),
+      
+      /**
+       * The legacy representation of UUID used by the Java driver.
+       *
+       * In this representation the order of bytes 0-7 are reversed, and the 
+       * order of bytes 8-15 are reversed.
+       *
+       * It encodes as BSON binary subtype 3
+       */
+      JAVA_LEGACY("javaLegacy"),
+
+     /**
+      * The legacy representation of UUID used by the Python driver.
+      *
+      * As with STANDARD, this representation conforms with RFC 4122, section
+      * 4.1.2 
+      *
+      * It encodes as BSON binary subtype 3
+      */
+      PYTHON_LEGACY("pythonLegacy")
+   }
+
+   class BsonBinary {
+      /* 
+       * Construct from a UUID using the standard UUID representation
+       * [Specification] This constructor SHOULD be included but MAY be 
+       *                 omitted if it creates backwards compatibility issues
+       */
+      constructor(Uuid uuid) 
+
+      /*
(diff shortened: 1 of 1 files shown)


source/uuid.rst
```diff
@@ -0,0 +1,321 @@
...
+Implementation Notes
+--------------------
+
+Since changing the default UUID representation can reasonably be considered a backwards-breaking change, drivers that implement the full specification should stage implementation according to semantic versioning guidelines.  Specifically, support for this specification can be added to a minor release, but with several exceptions: 
+
+The default UUID representation should be left as is (e.g. JAVA_LEGACY for the Java driver) rather than be changed to UNSPECIFIED.  In a subsequent major release, the default UUID representation can be changed to UNSPECIFIED (along with appropriate documentation indicating the backwards-breaking change).
+Subtype 4 BSON Binary values should continue to be decoded to native UUID values regardless of UUID representation
+
+
+Test Plan
```
ShaneHarvey: Can you add dashes (or some other formatting) to the test descriptions? Eg:
- Create a BsonBinary instance with subtype equal to 4 and data equal to the hex-encoded string "00112233445566778899AABBCCDDEEFF"
- Assert that a call to BsonBinary.asUuid() returns the given UUID
- ...

jyemin: Whoops, thanks for catching that.  

I opted for nested enumerated lists.  Let me know how it looks.

source/uuid.rst
```diff
@@ -0,0 +1,329 @@
...
+Note also that none of the above applies when decoding to strictly typed maps, e.g. a Map<String, BsonValue> like Java or .NET's BsonDocument class.  In those cases the driver is always decoding to BsonBinary, and applications would use the asUuid methods to explicitly convert from BsonBinary to UUID.
+
+
+Implementation Notes
+--------------------
+
+Since changing the default UUID representation can reasonably be considered a backwards-breaking change, drivers that implement the full specification should stage implementation according to semantic versioning guidelines.  Specifically, support for this specification can be added to a minor release, but with several exceptions: 
+
+The default UUID representation should be left as is (e.g. JAVA_LEGACY for the Java driver) rather than be changed to UNSPECIFIED.  In a subsequent major release, the default UUID representation can be changed to UNSPECIFIED (along with appropriate documentation indicating the backwards-breaking change).
+Subtype 4 BSON Binary values should continue to be decoded to native UUID values regardless of UUID representation
```
rstam: This line seems to contradict what the table above (under "Implicit decoding and encoding") says.

jyemin: Which line of the table?  I don't see the conflict.

rstam: The table says to decode subtype 4 to either a BsonBinary or a UUID depending on which line of the table you look at.

This statement says that subtype 4 BsonBinary values should continue to be decoded to native UUID values regardless of the UUID representation.

Which is it?

jyemin: I see now.  I think this implementation note is incorrect and should be removed.  Any objections?

ShaneHarvey: Agreed. Should we add a line that says "Drivers should document that the default UUID representation will change to UNSPECIFIED in the next major release?"

jyemin: Done.  I don't love a MUST in an implementation note, but there isn't a good place to put it other than this.

source/uuid.rst
```diff
@@ -0,0 +1,329 @@
...
+
+1. Create a BsonBinary instance with the given UUID
+   
+   a. Assert that the BsonBinary instance's subtype is equal to 4 and data equal to the hex-encoded string "00112233445566778899AABBCCDDEEFF"
+
+2. Create a BsonBinary instance with the given UUID and UuidRepresentation equal to STANDARD
+
+   a. Assert that the BsonBinary instance's subtype is equal to 4 and data equal to the hex-encoded string "00112233445566778899AABBCCDDEEFF"
+
+3. Create a BsonBinary instance with the given UUID and UuidRepresentation equal to JAVA_LEGACY
```
p-mongo: If it's not too much trouble, can the java and csharp examples be swapped to maintain lexicographical order?

source/uuid.rst
```diff
@@ -0,0 +1,329 @@
...
+1. Create a BsonBinary instance with subtype equal to 4 and data equal to the hex-encoded string "00112233445566778899AABBCCDDEEFF"
+
+   a. Assert that a call to BsonBinary.asUuid() returns the given UUID
+   b. Assert that a call to BsonBinary.asUuid(STANDARD) returns the given UUID
+   c. Assert that a call to BsonBinary.asUuid(UNSPECIFIED) raises an error
+   d. Assert that a call to BsonBinary.asUuid(JAVA_LEGACY) raises an error
+   e. Assert that a call to BsonBinary.asUuid(CSHARP_LEGACY) raises an error
+   f. Assert that a call to BsonBinary.asUuid(PYTHON_LEGACY) raises an error
+
+2. Create a BsonBinary instance with subtype equal to 3 and data equal to the hex-encoded string "7766554433221100FFEEDDCCBBAA9988"
```
p-mongo: Also swap this java test with the csharp one below.

p-mongo:
Having implemented a POC in Ruby (https://github.com/mongodb/bson-ruby/pull/146), I have the following thoughts on the proposed specification:

1. To me the biggest point of confusion was whether the specification was applicable to the Ruby driver, which does not have a native UUID type, and if so what were its requirements. I've been told that it didn't apply at all; however, the Ruby driver may be called upon to interoperate with data generated by other drivers, and to this extent the explicit encoding and decoding provisions of the specifications do apply (I would say they apply to all drivers).

The specification as currently written seems to be written for the 3 drivers that are called out in it (Java, C# and Python). If you agree that this specification (at least the explicit encoding and decoding aspect of it) applies to all drivers, then I would propose making the following changes:

- In the abstract, begin by describing the current UUID situation: type 4 is RFC 4122 compliant and is the recommended one to use for new drivers, type 3 can have the bytes be stored in 3 different formats depending on which driver created the data and is currently the default in the 3 drivers. Then state that one goal of the specification is to define methods to convert between the stringified UUIDs and the byte sequences in both subtypes, and the second goal is to define a common interface for those drivers that have a native UUID type for conversion between BSON binary and the native UUID type. This can be interpreted as a slight scope expansion of the spec.

- In the explicit encoding & decoding section, separate the representation enum from the methods added to the binary class, and preface the methods added with a summary paragraph. In Ruby there are no enums thus the representation enum is not implementable.

- The Binary class is required to have two constructors, both taking UUID arguments. In Ruby the Binary objects are constructed from raw bytes and the type - not from string UUIDs. The POC I've done adds a separate (class) method to construct a Binary from a string UUID. Additionally, some languages do not allow overloading constructors  - I suspect Go may have difficulties with the proposed API arising from the language's typing conventions, and in Ruby since we don't have a UUID type it would be also quite challenging to have the constructor take a string UUID since we won't be able to easily disambiguate that from a binary string.

I suggest instead requiring that the language provides an idiomatic way to create a Binary from a string UUID (in the paragraph-to-be-added mentioned before), and giving the constructor API as an example.

If desired I can write the proposed prose.

2. All of the tests contain bytes that, in their hex representation, have two identical characters (00, 11, etc.). This means an implementation that erroneously flips the order of the characters in each byte passes all spec tests.

For example, given the following correct implementation:

<pre>
        when 'java_legacy'
          hex.sub(/\A(..)(..)(..)(..)(..)(..)(..)(..)(..)(..)(..)(..)(..)(..)(..)(..)\z/) do |m|
            "#{$8}#{$7}#{$6}#{$5}#{$4}#{$3}#{$2}#{$1}" +
            "#{$16}#{$15}#{$14}#{$13}#{$12}#{$11}#{$10}#{$9}"
          end
</pre>

... the following incorrect (I presume) implementation which also flips the order of the hex characters in each byte also passes all spec tests:

<pre>
        when 'java_legacy'
          hex.sub(/\A(..)(..)(..)(..)(..)(..)(..)(..)(..)(..)(..)(..)(..)(..)(..)(..)\z/) do |m|
            "#{$8.reverse}#{$7.reverse}#{$6.reverse}#{$5.reverse}#{$4.reverse}#{$3.reverse}#{$2.reverse}#{$1.reverse}" +
            "#{$16}#{$15}#{$14}#{$13}#{$12}#{$11}#{$10}#{$9}"
          end
</pre>

3. You may consider adding prose tests for languages that do not have enums mandating that invalid representation values when explicitly encoding and decoding produce errors.

source/uuid.rst
jyemin: STANDARD is still out of order though.  Do you want that moved to?

In general, though, how is lexicographical ordering helpful? I'd rather not change this now and have to go back and updated anyone's existing ordering in test files, if they referenced these numbers.

jyemin:
@p-mongo thanks for your note.  I have a few responses:

1. Yes, the spec was originally written with C#, Java, and Python drivers in mind
2. I don't _think_ any of the changes you suggest would require a change to any of those drivers that have already implemented this specification.
3. Given the above, I would prefer to put off these changes to a 1.1 release of this spec.  The Java driver is slated to release this tomorrow, and I would like some version of the spec published ASAP.

source/uuid.rst
p-mongo: The way it is is fine then.

p-mongo:
I agree that no changes to the 3 drivers should be necessary.

jyemin:
Thanks Oleg.  Please open a SPEC ticket with the changes you'd like to see in 1.1.

p-mongo:
Is there a spec ticket that this PR goes with?

jyemin:
No, it's a WRITING ticket.  You'll find it linked to DRIVERS-654

PR #695, opened 2019-12-04
SPEC-1476 add constraints for single-threaded connection to mongocryptd

kevinAlbs:
I believe this should only affect libmongoc's single-threaded client implementation currently. These were the changes I made in the implementation to prevent server selection from immediately failing, and from a delay of 5 seconds due to the cooldown.

### DIFF
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -639,7 +639,7 @@ MongoClient fails to connect after spawning, the server selection error
 is propagated to the user.
 
 Single-threaded drivers MUST connect with `serverSelectionTryOnce=false <https://github.com/mongodb/specifications/blob/6bc8afe3516d2274c3f646e8293af15cf8651e8c/source/server-selection/server-selection.rst#serverselectiontryonce>`_
-and MUST bypass `cooldownMS <https://github.com/mongodb/specifications/blob/6bc8afe3516d2274c3f646e8293af15cf8651e8c/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#cooldownms>`_ when connecting to mongocryptd. See `Why are serverSelectionTryOnce and cooldownMS disabled for single-threaded drivers?`_.
+and MUST bypass `cooldownMS <https://github.com/mongodb/specifications/blob/6bc8afe3516d2274c3f646e8293af15cf8651e8c/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#cooldownms>`_ when connecting to mongocryptd. See `Why are serverSelectionTryOnce and cooldownMS disabled for single-threaded drivers connecting to mongocryptd?`_.
 
 If the ClientEncryption is configured with mongocryptdBypassSpawn=true,
 then the driver is not responsible for spawning mongocryptd. If server
@@ -1239,8 +1239,8 @@ Although not technically necessary for client side encryption, it does
 provide a fallback against accidentally sending unencrypted data from
 misconfigured clients.
 
-Why are serverSelectionTryOnce and cooldownMS disabled for single-threaded drivers?
------------------------------------------------------------------------------------
+Why are serverSelectionTryOnce and cooldownMS disabled for single-threaded drivers connecting to mongocryptd?
+-------------------------------------------------------------------------------------------------------------
 
 By default, single threaded clients set serverSelectionTryOnce to true, which
 means server selection fails if a topology scan fails the first time (i.e. it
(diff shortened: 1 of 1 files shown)


source/client-side-encryption/client-side-encryption.rst
```diff
@@ -638,6 +638,9 @@ server selection on the MongoClient to mongocryptd fails. If the
 MongoClient fails to connect after spawning, the server selection error
 is propagated to the user.
 
+Single-threaded drivers MUST connect with `serverSelectionTryOnce=false <https://github.com/mongodb/specifications/blob/6bc8afe3516d2274c3f646e8293af15cf8651e8c/source/server-selection/server-selection.rst#serverselectiontryonce>`_
```
jmikola: Most cross-spec links use relative URIs, so I think this could be `../server-selection/server-selection.rst#serverselectiontryonce`. Likewise for the other links in this PR.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -638,6 +638,9 @@ server selection on the MongoClient to mongocryptd fails. If the
...
+and MUST bypass `cooldownMS <https://github.com/mongodb/specifications/blob/6bc8afe3516d2274c3f646e8293af15cf8651e8c/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#cooldownms>`_ when connecting to mongocryptd. See `Why are serverSelectionTryOnce and cooldownMS disabled for single-threaded drivers connecting to mongocryptd?`_.
```
jmikola: Do we need to consider `heartbeatFrequencyMS`, or would periodic monitoring not be relevant here? `mongocryptd` is basically a standalone server, so there's no topology to discover, and if there's a concern with it crashing (I dunno if there's logic to respawn it) then perhaps we can already rely on SDAM to reach out to the server after ["certain events"](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#monitoring) (e.g. connection errors).

Second question: should the default `connectTimeoutMS` value be lowered from its default? In some local testing on the PHP driver, I realized that a connection timeout may cause a server selection loop to block for more than the server selection timeout. Consider:

```php
$m = new MongoDB\Driver\Manager('mongodb://1.1.1.1:27017/', [
    'connectTimeoutMS' => 10000,
    'serverSelectionTimeoutMS' => 1000,
    'serverSelectionTryOnce' => false,
]);
$c = $m->executeCommand('admin', new MongoDB\Driver\Command(['ping' => 1]));
var_dump($c->toArray()[0]);
```

This blocked for 10 seconds instead of one second, which was a bit unexpected. I don't believe this edge case is discussed in the SDAM spec, and I've probably never bothered to test this before today.

If we plan to leave `serverSelectionTimeoutMS` as the 30-second default for this internal client, I suppose it makes no difference -- but I wanted to bring this up just in case it was relevant.

kevinAlbs: > Do we need to consider heartbeatFrequencyMS, or would periodic monitoring not be relevant here? mongocryptd is basically a standalone server, so there's no topology to discover, and if there's a concern with it crashing (I dunno if there's logic to respawn it) then perhaps we can already rely on SDAM to reach out to the server after "certain events" (e.g. connection errors).

IMO periodic monitoring is not necessary. There is logic [to attempt to respawn once](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#managing-mongocryptd) upon a server selection error.

> If spawning is necessary, the driver MUST spawn mongocryptd whenever server selection on the MongoClient to mongocryptd fails. If the MongoClient fails to connect after spawning, the server selection error is propagated to the user.

--

> This blocked for 10 seconds instead of one second, which was a bit unexpected. I don't believe this edge case is discussed in the SDAM spec, and I've probably never bothered to test this before today.

Whoa good catch. That is surprising! The [server selection spec says](https://github.com/mongodb/specifications/blob/single-threaded/source/server-selection/server-selection.rst/#single-threaded-server-selection) says:
> Because single-threaded selection can do a blocking immediate check, serverSelectionTimeoutMS is not a hard deadline. The actual maximum server selection time for any given request can vary from serverSelectionTimeoutMS minus minHeartbeatFrequencyMS to serverSelectionTimeoutMS plus the time required for a blocking scan.

So I guess it is ok for single threaded drivers to exceed serverSelectionTimeoutMS. Since serverSelectionTimeoutMS is lowered to 1 second, I've lowered connectTimeoutMS accordingly.

(Also side note, I think libmongoc could, and probably should, cap the connection attempt to serverSelectionTimeoutMS, since it does a non-blocking connect. I created https://jira.mongodb.org/browse/CDRIVER-3451)

source/client-side-encryption/client-side-encryption.rst

PR #696, opened 2019-12-04
update initial DNS seedlist discovery test to follow auth spec

saghm:
According to [one of the auth spec tests](https://github.com/mongodb/specifications/blob/0085c4b30d9e4e549b1f65d96375ce133aed52b4/source/auth/tests/connection-string.yml#L290-L293), specifying `authSource` without a username is invalid, but one of the initial DNS seedlist discovery tests does this. I'm not quite sure why no other driver has run into this issue yet, but this is causing the Rust implementation to fail because it raises an error when this test does not expect one.


mbroadst:
@saghm see SPEC-1511, I think we should remove the restriction from the _auth_ tests, not the initial seedlist tests. This is a breaking change for Atlas users potentially, so I'm in favor of changing it there.

PR #697, opened 2019-12-05
update version history

bazile-clyde:


source/auth/auth.rst
```diff
@@ -1125,6 +1125,9 @@ Q: Why does SCRAM sometimes SASLprep and sometimes not?
 Version History
 ===============
 
+Version 1.8.2 Changes
```
jyemin: So we should also bump the spec version on line 8 to 1.8.2, right?

bazile-clyde: Ah, yes. Done.

PR #699, opened 2019-12-12
SPEC-1423: Update auth URI options spec for Ruby Driver

egiurleo:
The Ruby driver implements GSSAPI authentication in another gem, mongo-ruby-kerberos, which is not included in the driver by default. To handle this, the Ruby driver tests skip GSSAPI-related tests unless this gem is installed by checking whether the word "GSSAPI" is present in the test description. To make the URI-option tests easier to run on the Ruby Driver, I have made two changes:

1. I added "(GSSAPI)" to the description of the first test so that the Ruby test runner will know to skip it if GSSAPI is not installed. This is consistent with the way descriptions are written in the [auth spec tests](https://github.com/mongodb/specifications/blob/master/source/auth/tests/connection-string.yml#L33).
2. I added a second test using a different mechanism that exists by default in the Ruby driver, so that the Ruby driver can verify that auth-related URI options are being correctly parsed even if the mongo-ruby-kerberos gem is not present.

### DIFF
--- a/source/uri-options/tests/auth-options.json
+++ b/source/uri-options/tests/auth-options.json
@@ -25,7 +25,6 @@
       "auth": null,
       "options": {
         "authMechanism": "SCRAM-SHA-1",
-        "authMechanismProperties": null,
         "authSource": "authSourceDB"
       }
     }
--- a/source/uri-options/tests/auth-options.json
+++ b/source/uri-options/tests/auth-options.json
@@ -1,7 +1,7 @@
 {
   "tests": [
     {
-      "description": "Valid auth options are parsed correctly",
+      "description": "Valid auth options are parsed correctly (GSSAPI)",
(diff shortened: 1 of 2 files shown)


source/uri-options/tests/auth-options.json
```diff
@@ -15,6 +15,19 @@
...
+    {
+      "description": "Valid auth options are parsed correctly (SCRAM-SHA-1)",
+      "uri": "mongodb://foo:bar@example.com/?authMechanism=SCRAM-SHA-1&authSource=authSourceDB",
+      "valid": true,
+      "warning": false,
+      "hosts": null,
+      "auth": null,
+      "options": {
+        "authMechanism": "SCRAM-SHA-1",
+        "authMechanismProperties": null,
```
saghm: I don't think we want to put `authMechanismProperties` here even with the value of `null`; the existing tests all just leave off fields that explicitly specified.

egiurleo: Good point. Fixed that.

PR #701, opened 2019-12-17
add prose test for bypassing mongocryptd spawning

kevinAlbs:
By passing a custom pidfilepath and port that differs from any possible existing mongocryptd running in the background, if the driver was mistakenly spawning mongocryptd when it should not have, the spawned mongocryptd will not self-terminate, so it can be detected by either the pidfile or by connecting with a MongoClient.

I've PoC'ed this test with libmongoc.

### DIFF
--- a/source/client-side-encryption/tests/README.rst
+++ b/source/client-side-encryption/tests/README.rst
@@ -596,3 +596,66 @@ Data keys created with AWS KMS may specify a custom endpoint to contact (instead
 
    Expect this to fail with an exception with a message containing the string: "parse error"
 
+Bypass spawning mongocryptd
+===========================
+
+The following tests that setting ``mongocryptdBypassSpawn`` really does bypass spawning mongocryptd.
+
+#. Create a MongoClient configured with auto encryption (referred to as ``client_encrypted``)
+
+   Configure the required options. Use the ``local`` KMS provider as follows:
+
+   .. code:: javascript
+
+      { "local": { "key": <base64 decoding of LOCAL_MASTERKEY> } }
+
+   Configure with the ``keyVaultNamespace`` set to ``admin.datakeys``.
+
+   Configure the following ``extraOptions``:
+
+   .. code:: javascript
+
+      {
+        "mongocryptdBypassSpawn": true
+        "mongocryptdSpawnArgs": [ "--pidfilepath=bypass-spawning-mongocryptd.pid", "--port=27021"]
+      }
+
+   Drivers MAY pass a different value to ``--port`` if they expect their testing infrastructure to be using port 27021. Pass a port that should be free.
+
+#. Use ``client_encrypted`` to run a "ping" command.
+
+#. Validate that mongocryptd was not spawned. This may be done in one of the following was:
+
+   - Check that the file ``bypass-spawning-mongocryptd.pid`` does not exist (preferred)
+   - Creating a MongoClient to localhost:27021 (or whatever was passed via ``--port``) with serverSelectionTimeoutMS=100
+
+The following tests that setting ``bypassAutoEncryption`` really does bypass spawning mongocryptd.
+
+#. Create a MongoClient configured with auto encryption (referred to as ``client_encrypted``)
+
+   Configure the required options. Use the ``local`` KMS provider as follows:
+
+   .. code:: javascript
+
+      { "local": { "key": <base64 decoding of LOCAL_MASTERKEY> } }
+
+   Configure with the ``keyVaultNamespace`` set to ``admin.datakeys``.
+
+   Configure with ``bypassAutoEncryption=true``.
+
+   Configure the following ``extraOptions``:
+
+   .. code:: javascript
+
+      {
+        "mongocryptdSpawnArgs": [ "--pidfilepath=bypass-spawning-mongocryptd.pid", "--port=27021"]
+      }
+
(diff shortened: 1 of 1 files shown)


source/client-side-encryption/tests/README.rst
```diff
@@ -596,3 +596,66 @@ Data keys created with AWS KMS may specify a custom endpoint to contact (instead
...
+      }
+
+   Drivers MAY pass a different value to ``--port`` if they expect their testing infrastructure to be using port 27021. Pass a port that should be free.
+
+#. Use ``client_encrypted`` to run a "ping" command.
+
+#. Validate that mongocryptd was not spawned. This may be done in one of the following was:
+
+   - Check that the file ``bypass-spawning-mongocryptd.pid`` does not exist (preferred)
+   - Creating a MongoClient to localhost:27021 (or whatever was passed via ``--port``) with serverSelectionTimeoutMS=100
```
kevinAlbs: I put two options for detecting an incorrectly spawned mongocryptd. In the successful case, nothing gets spawned and no files are created.

source/client-side-encryption/tests/README.rst
```diff
@@ -596,3 +596,66 @@ Data keys created with AWS KMS may specify a custom endpoint to contact (instead
...
+   .. code:: javascript
+
+      {
+        "mongocryptdBypassSpawn": true
+        "mongocryptdSpawnArgs": [ "--pidfilepath=bypass-spawning-mongocryptd.pid", "--port=27021"]
+      }
+
+   Drivers MAY pass a different value to ``--port`` if they expect their testing infrastructure to be using port 27021. Pass a port that should be free.
+
+#. Use ``client_encrypted`` to run a "ping" command.
```
ShaneHarvey: Currently, a driver could pass this test even if mongocryptdBypassSpawn=true was mistakenly implemented as bypassAutoEncryption=true. What do you think about validating that auto encryption still works in this case?

source/client-side-encryption/tests/README.rst
```diff
@@ -428,7 +428,7 @@ Views are prohibited
 
 
 Corpus Test
-===========
+~~~~~~~~~~~
```
kevinAlbs: Fixed the headings.

source/client-side-encryption/tests/README.rst
kevinAlbs: That does seem like a good idea. Updated the test to check that auto encryption still works (by spawning mongocryptd manually via a separate temporary MongoClient with auto encryption)

source/client-side-encryption/tests/README.rst
divjotarora: I tried implementing this test in Go and the file-checking method seems race-prone. I'm not sure if there's a big delay between mongocryptd being spawned and the file being flushed to disk, but but Go does not pick up the new file immediately, so the test passes even if I remove the `mongocryptdBypassSpawn` option. I think it's best to remove this option and only allow the `MongoClient` option.

Also, 100 ms wasn't enough for the server selection timeout. Even if mongocryptd was spawned, the client still got a server selection error. I think this is because the `Ping` with the encrypted client is very fast and does not go through mongocryptd, so the process hasn't started accepting connections by the time we check for existence. Given that the regular auto encryption code uses a 1 second server selection timeout, I think we should do the same here. 

@ShaneHarvey  thoughts?

source/client-side-encryption/tests/README.rst
```diff
@@ -596,3 +596,85 @@ Data keys created with AWS KMS may specify a custom endpoint to contact (instead
...
+   Drivers MAY pass a different value to ``--port`` if they expect their testing infrastructure to be using port 27021. Pass a port that should be free.
+
+#. Use ``client_encrypted`` to run a "ping" command.
+
+#. Validate that mongocryptd was not spawned. This may be done in one of the following ways:
+
+   - Check that the file ``bypass-spawning-mongocryptd.pid`` does not exist (preferred)
+   - Creating a MongoClient to localhost:27021 (or whatever was passed via ``--port``) with serverSelectionTimeoutMS=100
+
+#. Spawn mongocryptd manually. This may be done by creating a separate MongoClient with automatic encryption enabled and running a ``ping`` command.
```
divjotarora: Clarify that this process should not use the custom port or pidfile.

source/client-side-encryption/tests/README.rst
ShaneHarvey: Thanks @divjotarora. Instead of running "ping", what if the test runs a command which needs encryption? Then was can assert that the command fails (the mongocryptd mark command will raise a server selection timeout).

divjotarora: @kevinAlbs and I talked about this. I really like this idea. The spec already requires mongocryptd errors to be distinguished in some way, so we can probably do this by requiring that a mongocryptd error is thrown and is a server selection error. Just checking for a server selection error allows for false positives (e.g. mongocryptd is spawned but for some reason there's a server selection timeout when communicating with the regular Mongo cluster).

divjotarora: Note that checking for a mongocryptd server selection error will not work for the test that uses `bypassAutoEncryption` because that test doesn't communicate with mongocryptd.

kevinAlbs: Updated to do an insert requiring encryption and expect a server selection error from the MongoClient connected to mongocryptd.

kevinAlbs:
Updated to no longer have the file check, and bump the server selection timeout ms to 1000.

PR #702, opened 2019-12-20
SPEC-1525 Fix SDAM spec's description of opTime

patrickfreed:
[SPEC-1525](https://jira.mongodb.org/browse/SPEC-1525).

This PR fixes the description of opTime in the SDAM spec, and it also updates the section to clarify why only mongos records the value.

### DIFF
--- a/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
+++ b/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
@@ -305,7 +305,7 @@ Fields:
 * opTime: an opTime or null.
   An opaque value representing the position in the oplog of the most recent write. Default null.
   (Only mongos and shard servers record this field when monitoring
-  config servers as replica sets, at least until `drivers allow applications to use readConcern "afterOptime".<https://github.com/mongodb/specifications/blob/master/source/max-staleness/max-staleness.rst#future-feature-to-support-readconcern-afteroptime>`_)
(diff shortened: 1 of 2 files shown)


source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -302,10 +302,10 @@ Fields:
 * roundTripTime: the duration of the ismaster call. Default null.
 * lastWriteDate: a 64-bit BSON datetime or null.
   The "lastWriteDate" from the server's most recent ismaster response.
-* opTime: an ObjectId or null.
-  The last opTime reported by the server; an ObjectId or null.
+* opTime: an opTime or null.
+  An opaque value representing the position in the oplog of the most recent write. Default null.
```
ShaneHarvey: Can we use the same phrasing as [the max-staleness spec](https://github.com/mongodb/specifications/blob/master/source/max-staleness/max-staleness.rst#lastwrite) to keep the definitions in sync?

> opTime: an opaque value representing the most recent replicated write. ...

patrickfreed: Yeah, it would be best to keep them in line with each other. It looks like that one might need to be updated too, though, since I don't think `opTime` gives any guarantees on whether the write was replicated or not. For that, you'd need [`majorityOpTime`](https://docs.mongodb.com/manual/reference/command/isMaster/#isMaster.lastWrite.majorityOpTime) (I think).

patrickfreed: I've updated them both to a common definition, let me know if it seems accurate.

PR #703, opened 2020-01-07
SPEC-1533 primary mismatched me is not removed from server set

mbroadst:
This test validates the pseudocode provided in the SDAM specification, ensuring that a primary is always added to the working server set even if it has a `me` mismatch.

### DIFF
--- a/source/server-discovery-and-monitoring/tests/rs/primary_mismatched_me_not_removed.yml
+++ b/source/server-discovery-and-monitoring/tests/rs/primary_mismatched_me_not_removed.yml
@@ -1,0 +1,73 @@
+description: Primary mismatched me is not removed
+uri: mongodb://localhost:27017,localhost:27018/?replicaSet=rs
+
+phases: [
+  {
+    responses: [
+      ["localhost:27017", {
+        ok: 1,
+        hosts: [
+          "localhost:27017",
+          "localhost:27018"
+        ],
+        ismaster: true,
+        setName: "rs",
+        primary: "localhost:27017",
+        # me does not match the primary responder's address, but the server
+        # is still added because we don't me mismatch check the primary and all
+        # servers from a primary ismaster are added to the working server set
+        me: "a:27017",
+        minWireVersion: 0,
+        maxWireVersion: 7
+      }]
+    ],
+    outcome: {
+      servers: {
+        "localhost:27017": {
+          type: "RSPrimary",
+          setName: "rs"
+        },
+        "localhost:27018": {
+          type: "Unknown",
+          setName: null
+        }
+      },
+      topologyType: "ReplicaSetWithPrimary",
+      logicalSessionTimeoutMinutes: null,
+      setName: "rs"
+    }
+  },
+  {
+    responses: [
+      ["localhost:27018", {
+        ok: 1,
+        hosts: [
+          "localhost:27017",
+          "localhost:27018"
+        ],
+        ismaster: false,
+        secondary: true,
+        setName: "rs",
+        primary: "localhost:27017",
+        me: "localhost:27018",
+        minWireVersion: 0,
+        maxWireVersion: 7
+      }]
+    ],
+    outcome: {
+      servers: {
+        "localhost:27017": {
+          type: "RSPrimary",
+          setName: "rs"
+        },
+        "localhost:27018": {
+          type: "RSSecondary",
+          setName: "rs"
+        }
+      },
+      topologyType: "ReplicaSetWithPrimary",
+      logicalSessionTimeoutMinutes: null,
+      setName: "rs"
+    }
+  }
+]
--- a/source/server-discovery-and-monitoring/tests/rs/primary_mismatched_me_not_removed.json
+++ b/source/server-discovery-and-monitoring/tests/rs/primary_mismatched_me_not_removed.json
@@ -1,0 +1,77 @@
+{
+  "description": "Primary mismatched me is not removed",
+  "uri": "mongodb://localhost:27017,localhost:27018/?replicaSet=rs",
+  "phases": [
+    {
+      "responses": [
+        [
+          "localhost:27017",
+          {
+            "ok": 1,
+            "hosts": [
+              "localhost:27017",
+              "localhost:27018"
+            ],
+            "ismaster": true,
+            "setName": "rs",
+            "primary": "localhost:27017",
+            "me": "a:27017",
+            "minWireVersion": 0,
+            "maxWireVersion": 7
+          }
+        ]
+      ],
+      "outcome": {
+        "servers": {
+          "localhost:27017": {
+            "type": "RSPrimary",
+            "setName": "rs"
+          },
+          "localhost:27018": {
+            "type": "Unknown",
+            "setName": null
+          }
+        },
+        "topologyType": "ReplicaSetWithPrimary",
+        "logicalSessionTimeoutMinutes": null,
+        "setName": "rs"
+      }
+    },
+    {
+      "responses": [
+        [
+          "localhost:27018",
+          {
+            "ok": 1,
+            "hosts": [
+              "localhost:27017",
+              "localhost:27018"
+            ],
+            "ismaster": false,
+            "secondary": true,
+            "setName": "rs",
+            "primary": "localhost:27017",
+            "me": "localhost:27018",
+            "minWireVersion": 0,
+            "maxWireVersion": 7
+          }
+        ]
+      ],
+      "outcome": {
+        "servers": {
+          "localhost:27017": {
+            "type": "RSPrimary",
+            "setName": "rs"
+          },
+          "localhost:27018": {
+            "type": "RSSecondary",
+            "setName": "rs"
+          }
+        },
+        "topologyType": "ReplicaSetWithPrimary",
+        "logicalSessionTimeoutMinutes": null,
+        "setName": "rs"
+      }
+    }
+  ]
+}


source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1177,13 +1177,18 @@ updateRSFromPrimary
     if topologyDescription.setName is null:
         topologyDescription.setName = description.setName
 
-    else if topologyDescription.setName != description.setName:
+    if topologyDescription.setName != description.setName:
```
jyemin: Ah, good catch, though orthogonal to the rest of the PR.  Is it true that the spec tests already assume that there is no "else" here?

source/server-discovery-and-monitoring/tests/rs/split_horizon_me_mismatch.yml
```diff
@@ -0,0 +1,60 @@
+description: Split horizon me mismatch
```
jyemin: Since it's a bug that split horizons works this way, I think we should leave split horizons out of the description/file name

There are currently three test files about mismatched me:

* primary_mismatched_me.yml
* secondary_mismatched_me.yml
* primary_to_no_primary_mismatched_me.yml

All of them have a mismatch between the host name in the seed list and in the "me" field, but no mismatch between the "me" field and the "hosts" field.  So how about we add three new test files that are just like those, except with a mismatch between "me" and "hosts" (but a match between the seed list and "hosts")

For clarity, we could rename the existing three files/descriptions to indicate what mismatch we're talking about.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
p-mongo: As far as I can tell omitting "else" does not change the behavior in any way and just makes the case where topologydescription.setname was null less efficient.

p-mongo:
The added test passes on the Ruby driver.

source/server-discovery-and-monitoring/tests/rs/split_horizon_me_mismatch.yml
mbroadst: I'm not sure why we need three new test files, what does that cover that's not covered by this single one?

jyemin: For the same reason we have three test files for me mismatched with the seed list server name?  Are you implying that having three files for that is also unnecessary?  I'm just trying to keep things consistent.

mbroadst: I'm just saying I don't think we gain meaningful coverage by adding those three new tests, not that the existing test files aren't valuable. My question is: do you see an active path here where we are likely to have additional bugs, or are you suggesting we triplicate just for consistency?

jyemin: I may be thinking about this wrong, so let's start over.

Since the server team agrees that the "me" field value should always match one of values in the hosts/passives/arbiters arrays, let's assume the server will have its own tests that assert that.  In that case, I don't think we need to add a new test case where that invariant is broken.

If we can agree on that (and I can see that we might not), is there a simpler test that keeps that invariant (and therefore is more similar to the existing three tests for mismatched me)  but still catches the current difference in behavior between Java/.NET and PyMongo/Node?

mbroadst: Thanks. It turns out the position of the check in the pseudo code is wrong, and breaks the other me mismatch tests. I’ll fix that when I get in at noon. 

In the meantime, maybe I’m misunderstanding your initial ask. Would you be willing to provide an example test here? It would be interesting to see if it indeed expresses untested code paths for the SDAM update machinery

ShaneHarvey: > If we can agree on that (and I can see that we might not), is there a simpler test that keeps that invariant (and therefore is more similar to the existing three tests for mismatched me) but still catches the current difference in behavior between Java/.NET and PyMongo/Node?

Such a test (where the primary reports a "me" mismatch and the "me" is in the "hosts" field) already does exist. Actually two of them:

1. [primary_to_no_primary_mismatched_me.yml](https://github.com/mongodb/specifications/blob/0d7641c5560b11f195c0646d602d8c9e1f89f288/source/server-discovery-and-monitoring/tests/rs/primary_to_no_primary_mismatched_me.yml#L51)
2. [primary_mismatched_me.yml](https://github.com/mongodb/specifications/blob/0d7641c/source/server-discovery-and-monitoring/tests/rs/primary_mismatched_me.yml#L16)

Python actually does pass these tests even though we don't use the "me" mismatch rule on the primary's isMaster responses. We remove the primary because of another SDAM rule in [`updateRSFromPrimary`](https://github.com/mongodb/specifications/blob/0d7641c/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#updatersfromprimary):
```
for each server in topologyDescription.servers:
    if server.address not in description's "hosts", "passives", or "arbiters":
        remove the server and stop monitoring it
```

So there are two conclusions we can make:
- It's worthwhile to test the case where the hosts/me invariant is broken.
- Or the current test coverage is sufficient.

mbroadst: Jeff and I talked this through today, and we think the behavior of the Node and Python driver is actually correct. I've replaced the test here with one validating the pseudocode provided in the SDAM specification.

mbroadst:
@p-mongo you may wish to re-check that the Ruby driver passes the newly updated test.

source/server-discovery-and-monitoring/tests/rs/primary_mismatched_me_not_removed.yml
```diff
@@ -0,0 +1,69 @@
...
+          setName: null
+        }
+      },
+      topologyType: "ReplicaSetWithPrimary",
+      logicalSessionTimeoutMinutes: null,
+      setName: "rs"
+    }
+  },
+  {
+    responses: [
```
jyemin: Nit: this is proving that secondary mismatched me is removed, which is not what the description of the test says.  Don't we already test this somewhere else?

ShaneHarvey:
> Jeff and I talked this through today, and we think the behavior of the Node and Python driver is actually correct.

Ah I see. The existing [updateRSFromPrimary](https://github.com/mongodb/specifications/blob/37d0df8/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#updatersfromprimary) routine has this note:
> A note on checking "me": Unlike updateRSWithPrimaryFromMember, there is no need to remove the server if the address is not equal to "me": since the server address will not be a member of either "hosts", "passives", or "arbiters", the server will already have been removed.

p-mongo:
This test fails in Ruby: https://gist.github.com/p-mongo/1cb93cfa6d2bcb81798c445483f659dd

Note that the paths in sdam spec which update RS topology from non-primary have me mismatch provisions; the path which updates RS topology from primary doesn't, but has the following note also quoted by @ShaneHarvey :

>   A note on checking "me": Unlike `updateRSWithPrimaryFromMember`, there is no need to remove the server if the address is not equal to
  "me": since the server address will not be a member of either "hosts", "passives", or "arbiters", the server will already have been
  removed.

This paragraph of the spec makes the assumption that a response from primary will always have the primary's `me` in its hosts/passives/arbiters. The test being added violates this assumption as far as I can tell. If the intent is to ignore me for primary responses, I would think that this requires a corresponding change in the language of the spec.

However it also seems to me that the spec generally expects `me` to be equal to the server's address (the one the client connected to). There is a FAQ item lower in the spec which reads as follows:

> Using me field to detect seed list members that do not match host names in the replica set configuration
>
> Removal from the topology of seed list members where the "me" property does not
match the address used to connect prevents clients from being able to select
a server, only to fail to re-select that server once the primary has responded.

Again the spec makes the assumption that the address in `me` is the true address of the server, and this address should be matching the one applications are connecting to.

So, if this test is added as currently proposed I think a revision of the spec language with respect to how drivers should be treating the `me` field is necessary.

jyemin:
@p-mongo I initially agreed with you that the test violates an assumption made in the spec.  And in fact we do generally expect the `me` field to be equal to one of the values in hosts/passives/arbiters (we don't necessarily expect it to be equal to one of the seed list values, as people often end up using non-canonical addresses in their seed list).  The fact that the `me` field violated that invariant is a server bug, and I don't think we need to test driver responses to all possible server bugs.

The reason I think this test is still valid is because it's the only way we could think of to prove that the state machine is written correctly in all drivers.  If we just test cases where the "me" field equals a value in hosts, then the removal of a server that doesn't match `me` is a no-op in the case of a primary.  That's why the existing tests didn't catch the behavioral difference between Java/.NET/Ruby and Node/Python.

p-mongo:
I attempted to alter the Ruby driver to get it to pass this test, but there seems to be another issue with the description not being recognized as a primary the cause of which is not immediately apparent to me.

It is my understanding that the reasoning behind this PR can be summarized as follows, correct me if I am wrong:

> `me` address should be included in hosts/passives/arbiters. If `me` is not included it is a server bug. The spec assumes `me` is included. However, if `me` is not included we want all drivers to behave identically.

I would argue that the last sentence above means that the spec should explicitly call out that `me` can in fact be not included in hosts/passives/arbiters, and what the drivers are required to do in such situations. By "call out" I mean this situation and the corresponding requirement should be described in words, such as in an FAQ section, rather than only existing in a test case. Requirements that exist only in test cases essentially force implementers to guess as to what the spec author(s) *intended* (since the intention is never explicitly stated) and as such can miss other related cases that happen to not be covered by tests. Providing an explicit explanation and requirement allows implementers to incorporate the requirements conceptually into the core of their implementation, ultimately producing a higher quality implementation.

mbroadst:
@p-mongo sounds like there is a bug in the Ruby SDAM implementation.

The intent of this PR is to validate the behavior in the pseudocode for [updateRsFromPrimary](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#updatersfromprimary). Specifically, when a topology description is updated from a primary ismaster, and there is a primary `me` mismatch, the primary is **not** removed from the working set of servers. The reason there was discussion of `me` in hosts/passives/arbiters was because @ShaneHarvey expected that this behavior would have been covered by the [primary_mismatched_me](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/tests/rs/primary_mismatched_me.yml) test. It was not because of separately specified pseudocode to remove unknown servers from the working set.

You might disagree with some (or all) of the decisions here, but this PR introduces a test that validates the language in the spec *as it is today*. I propose we merge this, and then create SPEC tickets to further discuss changes you would like to see made to SDAM.

p-mongo:
I would say the only disagreement I have is in treating a spec test as equivalent to prose in specification. I do not object to the test spec in this PR being added.

source/server-discovery-and-monitoring/tests/rs/primary_mismatched_me_not_removed.yml
mbroadst: It does prove that, but I don't think you can write this test without two nodes. Do you feel strongly about this?

jyemin: I'm not objecting to the test.  Just that it's included in a test with the description: `Primary mismatched me is not removed`.  I think it should be a separate test with the description: `Secondary mismatched me is removed`.

My separate question was whether there's already a test for this elsewhere.  Is this testing something different than the `Secondary mismatched me` test that we already have?

PR #704, opened 2020-01-09
SPEC-1537 reduce precision required of extended bson implementations


p-mongo:
https://jira.mongodb.com/browse/SPEC-1537

### DIFF
--- a/source/bson-corpus/tests/double.json
+++ b/source/bson-corpus/tests/double.json
@@ -28,16 +28,16 @@
             "relaxed_extjson": "{\"d\" : -1.0001220703125}"
         },
         {
-            "description": "1.23456789012345677E+18",
+            "description": "1.2345678901234568e+18 - precision adjusted for Ruby",
             "canonical_bson": "1000000001640081E97DF41022B14300",
-            "canonical_extjson": "{\"d\" : {\"$numberDouble\": \"1.23456789012345677E+18\"}}",
-            "relaxed_extjson": "{\"d\" : 1.23456789012345677E+18}"
+            "canonical_extjson": "{\"d\" : {\"$numberDouble\": \"1.2345678901234568e+18\"}}",
+            "relaxed_extjson": "{\"d\" : 1.23456789012345678e+18}"
         },
         {
-            "description": "-1.23456789012345677E+18",
+            "description": "-1.2345678901234568e+18 - precision adjusted for Ruby",
             "canonical_bson": "1000000001640081E97DF41022B1C300",
-            "canonical_extjson": "{\"d\" : {\"$numberDouble\": \"-1.23456789012345677E+18\"}}",
-            "relaxed_extjson": "{\"d\" : -1.23456789012345677E+18}"
+            "canonical_extjson": "{\"d\" : {\"$numberDouble\": \"-1.2345678901234568e+18\"}}",
+            "relaxed_extjson": "{\"d\" : -1.2345678901234568e+18}"
         },
         {
             "description": "0.0",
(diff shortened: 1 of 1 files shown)


source/bson-corpus/tests/double.json
```diff
@@ -28,16 +28,16 @@
             "relaxed_extjson": "{\"d\" : -1.0001220703125}"
         },
         {
-            "description": "1.23456789012345677E+18",
+            "description": "1.2345678901234568e+18 - precision adjusted for Ruby",
             "canonical_bson": "1000000001640081E97DF41022B14300",
-            "canonical_extjson": "{\"d\" : {\"$numberDouble\": \"1.23456789012345677E+18\"}}",
-            "relaxed_extjson": "{\"d\" : 1.23456789012345677E+18}"
+            "canonical_extjson": "{\"d\" : {\"$numberDouble\": \"1.2345678901234568e+18\"}}",
+            "relaxed_extjson": "{\"d\" : 1.23456789012345678e+18}"
```
ShaneHarvey: I suspect that this test will now fail in other languages besides Python/Ruby because some actually do output ` 1.23456789012345677E+18`. The best option might be to change `canonical_bson` to a large double that is serialized to JSON the same across all drivers. `1.0001220703125` seems to work the same in all drivers (no one has complained about those tests yet). How about we try `1.0001220703125E+18`?

```python
>>> 1.0001220703125E+18
1.0001220703125e+18
```

p-mongo: This value passes tests in Ruby, I updated the PR.

source/bson-corpus/tests/double.json
```diff
@@ -28,16 +28,16 @@
             "relaxed_extjson": "{\"d\" : -1.0001220703125}"
         },
         {
-            "description": "1.23456789012345677E+18",
-            "canonical_bson": "1000000001640081E97DF41022B14300",
-            "canonical_extjson": "{\"d\" : {\"$numberDouble\": \"1.23456789012345677E+18\"}}",
-            "relaxed_extjson": "{\"d\" : 1.23456789012345677E+18}"
+            "description": "1.0001220703125e+18",
```
ShaneHarvey: Can we use capital `E` for consistency with other tests?

jyemin:
It does fail in the Java driver, though I haven't looked into why.  Here's the failure for the positive number:

    Failed to create expected canonical JSON for document with description '1.0001220703125e+18' 
    Expected :{"d":{"$numberDouble":"1.0001220703125e+18"}}
    Actual   :{"d":{"$numberDouble":"1.00012207031249997E18"}}

p-mongo:
I reproduced the discrepancy using MRI & JRuby. How about `1.2345678921232E+18`:

MRI:

<pre>
irb(main):011:0> 1.0001220703125
=> 1.0001220703125
irb(main):012:0> 1.0001220703125e+18
=> 1.0001220703125e+18
irb(main):013:0> 1.2345678921232E+18
=> 1.2345678921232e+18
</pre>

JRuby:

<pre>

irb(main):015:0> 1.0001220703125
=> 1.0001220703125
irb(main):016:0> 1.0001220703125e+18
=> 1000122070312499970.0
irb(main):017:0> 1.2345678921232E+18
=> 1234567892123200000.0
</pre>

jyemin:
It's better, but unfortunately Java's Double#toString method doesn't add the superfluous but harmless `+` before the exponent:

Expected :{"d":{"$numberDouble":"1.2345678921232E+18"}}
Actual   :{"d":{"$numberDouble":"1.2345678921232E18"}}

Not sure what to do about that...

jyemin:
Looks like I had changed these tests just in the Java driver a long time ago to remove the `+`.  So feel  free to proceed as we can just do that again with the updated file.

p-mongo:
@jyemin  Do I have your approval for this PR?

PR #705, opened 2020-01-09
SPEC-1538 Add a bson corpus test with invalid type for $code when $scope is also present


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/SPEC-1538

PR #706, opened 2020-01-09
SPEC-1539 Always include milliseconds when serializing datetime to extended json


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/SPEC-1539

PR #707, opened 2020-01-10
SPEC-1540: Error if hint specified for unacknowledged update using legacy opcode

jmikola:
https://jira.mongodb.org/browse/SPEC-1540
https://jira.mongodb.org/browse/SPEC-1542

I didn't realize the rationale behind this copypasta when originally implementing [SPEC-1334](https://jira.mongodb.org/browse/SPEC-1334), but I understand now why it's relevant for any option that may not have a place in a legacy opcode. This could technically apply to all legacy opcode usage, but I believe unacknowledged writes are the only place where modern drivers might still use them (if they aren't already using OP_MSG).

My second commit updates language for options other than `hint` to be more clear and also adds a Q&A entry to explain our current design. That commit does not entail driver changes.

### DIFF
--- a/source/crud/crud.rst
+++ b/source/crud/crud.rst
@@ -815,7 +815,7 @@ Basic
      *
      * This option is sent only if the caller explicitly provides a true value. The default is to not send a value.
      * For servers < 3.2, this option is ignored and not sent as document validation is not available.
-     * For unacknowledged writes using opcodes, the driver MUST raise an error if the caller explicitly provides a value.
+     * For unacknowledged writes using OP_INSERT, OP_UPDATE, or OP_DELETE, the driver MUST raise an error if the caller explicitly provides a value.
      */
     bypassDocumentValidation: Optional<Boolean>;
   }
@@ -827,7 +827,7 @@ Basic
      *
      * This option is sent only if the caller explicitly provides a true value. The default is to not send a value.
      * For servers < 3.2, this option is ignored and not sent as document validation is not available.
-     * For unacknowledged writes using opcodes, the driver MUST raise an error if the caller explicitly provides a value.
+     * For unacknowledged writes using OP_INSERT, the driver MUST raise an error if the caller explicitly provides a value.
      */
     bypassDocumentValidation: Optional<Boolean>;
   }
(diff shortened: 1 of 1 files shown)


source/crud/crud.rst
```diff
@@ -1813,6 +1818,7 @@ Q: Where is ``singleBatch`` in FindOptions?
 Changes
 =======
 
+* 2020-01-10: Error if hint specified for unacknowledged update using legacy opcode
```
kevinAlbs: Per discussion, this should also error for an unacknowledged OP_MSG. A user could be setting this in an unacknowledged write with OP_MSG in a 3.6-4.2 server. Since the option is unsupported, the server would always error, but no error would be returned. (Also made a [comment in SPEC-1540|https://jira.mongodb.org/browse/SPEC-1540?focusedCommentId=2718903&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2718903])

jmikola: I'm happy to make that change here, but I'll create a separate SPEC ticket to update the other, pre-existing options (e.g. `collation`) and let it be triaged accordingly at the next leads meeting.

source/crud/crud.rst
```diff
@@ -1810,9 +1815,14 @@ Q: Where is ``useCursor`` in AggregateOptions?
 Q: Where is ``singleBatch`` in FindOptions?
   Drivers have historically allowed users to request a single batch of results (after which the cursor is closed) by specifying a negative value for the ``limit`` option. For servers < 3.2, a single batch may be requested by specifying a negative value in the ``numberToReturn`` wire protocol field. For servers >= 3.2, the ``find`` command defines ``limit`` as a non-negative integer option but introduces a ``singleBatch`` boolean option. Rather than introduce a ``singleBatch`` option to FindOptions, the spec preserves the existing API for ``limit`` and instructs drivers to convert negative values accordingly for servers >= 3.2.
 
+Q: Why are client-side errors raised for some unsupported options?
```
kevinAlbs: Good thinking adding this rationale. Maybe it is also worth mentioning that unacknowledged writes using OP_QUERY commands really would give the client acknowledgement of error, hence why they do not require a client side error.

jmikola: Good idea. Updated.

PR #709, opened 2020-01-14
SPEC-1514: Specify behavior of empty URI authentication options

jstewart-mongo:
SPEC-1514

Added language to define how empty authentication-related options are to be handled. If an auth related option is specified but empty, the option should be assigned the default value.

### DIFF
--- a/source/auth/auth.rst
+++ b/source/auth/auth.rst
@@ -1161,7 +1161,7 @@ Version History
 ===============
 
 Version 1.9.0 Changes
-    * Clarify that drivers should raise and error when a connection string
+    * Clarify that drivers must raise an error when a connection string
       has an empty value for authSource.
 
 Version 1.8.3 Changes
--- a/source/auth/auth.rst
+++ b/source/auth/auth.rst
@@ -6,7 +6,7 @@ Driver Authentication
 =====================
 
 :Spec: 100
-:Spec Version: 1.8.3
+:Spec Version: 1.9.0
 :Title: Driver Authentication
 :Author: Craig Wilson, David Golden
 :Advisors: Andy Schwerin, Bernie Hacket, Jeff Yemin, David Golden
@@ -978,7 +978,7 @@ gssapiServiceName (deprecated)
 Errors
 ------
 
-Drivers SHOULD raise an error if the ``authSource`` option is specified in the connection string with an empty value, e.g. ``mongodb://localhost/admin?authSource=``.
+Drivers MUST raise an error if the ``authSource`` option is specified in the connection string with an empty value, e.g. ``mongodb://localhost/admin?authSource=``.
 
 
 Implementation
@@ -1160,14 +1160,16 @@ Q: Why does SCRAM sometimes SASLprep and sometimes not?
 Version History
 ===============
 
+Version 1.9.0 Changes
+    * Clarify that drivers should raise and error when a connection string
+      has an empty value for authSource.
+
 Version 1.8.3 Changes
     * Clarify that authSource in URI is not treated as a user configuring
       auth credentials.
 
 Version 1.8.2 Changes
     * Added MONGODB-IAM auth mechanism
-    * Clarify that drivers should raise and error when a connection string
-      has an empty value for authSource.
 
 Version 1.8.1 Changes
     * Clarify database to use for auth mechanism negotiation.
--- a/source/auth/auth.rst
+++ b/source/auth/auth.rst
@@ -974,6 +974,8 @@ authMechanismProperties=PROPERTY_NAME:PROPERTY_VALUE,PROPERTY_NAME2:PROPERTY_VAL
 gssapiServiceName (deprecated)
 	An alias for ``authMechanismProperties=SERVICE_NAME:mongodb``.
 
+If any of the auth related options are specified in the connection string without a value, e.g. ``mongodb://localhost/admin?authSource=``, the default value for the option should be used.
+
 
 Implementation
 --------------
--- a/source/auth/auth.rst
+++ b/source/auth/auth.rst
@@ -974,7 +974,11 @@ authMechanismProperties=PROPERTY_NAME:PROPERTY_VALUE,PROPERTY_NAME2:PROPERTY_VAL
 gssapiServiceName (deprecated)
 	An alias for ``authMechanismProperties=SERVICE_NAME:mongodb``.
 
-If any of the auth related options are specified in the connection string without a value, e.g. ``mongodb://localhost/admin?authSource=``, the default value for the option should be used.
+
+Errors
+------
+
+Drivers SHOULD raise an error if the ``authSource`` option is specified in the connection string with an empty value, e.g. ``mongodb://localhost/admin?authSource=``.
 
 
 Implementation
@@ -1162,6 +1166,8 @@ Version 1.8.3 Changes
 
 Version 1.8.2 Changes
     * Added MONGODB-IAM auth mechanism
+    * Clarify that drivers should raise and error when a connection string
+      has an empty value for authSource.
 
 Version 1.8.1 Changes
     * Clarify database to use for auth mechanism negotiation.
--- a/source/auth/tests/connection-string.json
+++ b/source/auth/tests/connection-string.json
@@ -113,6 +113,11 @@
       "valid": false
     },
     {
+      "description": "must raise an error when the authSource is empty without credentials",
+      "uri": "mongodb://localhost/admin?authSource=",
+      "valid": false
+    },
+    {
       "description": "should throw an exception if authSource is invalid (GSSAPI)",
       "uri": "mongodb://user%40DOMAIN.COM@localhost/?authMechanism=GSSAPI&authSource=foo",
       "valid": false
--- a/source/auth/tests/connection-string.json
+++ b/source/auth/tests/connection-string.json
@@ -25,6 +25,18 @@
       }
     },
     {
+      "description": "should use the database when the authSource is empty",
+      "uri": "mongodb://user:password@localhost/foo?authSource=",
+      "valid": true,
+      "credential": {
+        "username": "user",
+        "password": "password",
+        "source": "foo",
+        "mechanism": null,
+        "mechanism_properties": null
+      }
+    },
+    {
       "description": "should use the authSource when specified",
       "uri": "mongodb://user:password@localhost/foo?authSource=bar",
       "valid": true,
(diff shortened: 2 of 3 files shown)


source/auth/auth.rst
```diff
@@ -970,6 +970,8 @@ authMechanismProperties=PROPERTY_NAME:PROPERTY_VALUE,PROPERTY_NAME2:PROPERTY_VAL
 gssapiServiceName (deprecated)
 	An alias for ``authMechanismProperties=SERVICE_NAME:mongodb``.
 
+If any of the auth related options are specified in the connection string without a value, e.g. ``mongodb://localhost/admin?authSource=``, the default value for the option should be used.
```
rozza: Update the changes section and last modified.

source/auth/tests/connection-string.json
```diff
@@ -24,6 +24,18 @@
         "mechanism_properties": null
       }
     },
+    {
+      "description": "should use the database when the authSource is empty",
```
rozza: Also add a test with no database: `mongodb://user:password@localhost/?authSource=`

p-mongo:
I suggested the following in the spec ticket:

> It may be convenient for users to treat an empty authSource the same way as if authSource was not provided at all. This is also in line with common web development conventions.

However, I am having second thoughts about this approach. Currently the Ruby driver passes an empty auth source to the server during authentication, which subsequently fails:

<pre>
Mongo::Auth::Unauthorized (User a (mechanism: scram) is not authorized to access admin (auth source: , used mechanism: SCRAM-SHA-1, used server: localhost:14422 (PRIMARY)): Invalid database name: '' (on localhost:14422))
</pre>

Reading the auth spec again, I notice it has this language in https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#implementation:

> The source is determined by the following: 
>        if authSource is specified, it is used.

There is also precedent in the auth spec for treating empty strings as specified values (https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#credential-delimiter-in-uri-implies-authentication):

> The presence of a credential delimiter (i.e. @) in the URI connection string is evidence that the user has unambiguously specified user information and MUST be interpreted as a user configuring authentication credentials (even if the username and/or password are empty strings).

In light of this, I think performing a driver survey of how an empty string for authSource is handled is a good next step, and if there is consensus among existing implementations that this causes them to set authSource to the empty string, and subsequently allowing auth to fail, I currently am in favor of this approach.

jyemin:
I am not inclined to special-case the auth-related connection string options.  It seems confusing, and will complicate connection string parser implementations if we treat each option on a case-by-case basis.

So if we do anything, let's do it in the [connection string specification](https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.rst#defining-connection-options) and make it apply globally to all options.

Unfortunately, the one place I've seen that we've (sort of) defined the semantics of a missing value is for readPreferenceTags.  See https://github.com/mongodb/specifications/blob/d49015c0269354d16c13ff97698a7045d4354026/source/connection-string/connection-string-spec.rst#id22, the section about Lists.  This was probably a mistake, but it is necessary to be able to specify the empty tag set as the possible last set of tags.

source/auth/auth.rst
jmikola: Per discussion in #driver-devs, I think this should just specify that drivers should raise an error if `authSource` is an empty string. We also talked about opening separate tickets to (a) clarify how empty values in a connection string are parsed (in connection string spec) and (b) specify validation rules for other string options that might receive an empty value (in the URI options spec).

@jstewart-mongo: Neither the PR nor the commit reference [SPEC-1514](https://jira.mongodb.org/browse/SPEC-1514). Please make sure it's at least in the commit message so this change gets linked back to the JIRA issue (in both directions).

jstewart-mongo: @rozza I have updated the changes section and last modified.
@jmikola I have updated the commit message to reference the ticket.
I created a new subsection to describe what drivers should do when `authSource` is specified with an empty string.

source/auth/tests/connection-string.json
jstewart-mongo: I have removed this test and added a new one that makes it invalid when `authSource` is empty.

source/auth/auth.rst
```diff
@@ -13,7 +13,7 @@ Driver Authentication
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: 2.6
-:Last Modified: 2020-01-13
+:Last Modified: 2020-01-16
```
jmikola: This warrants a bump to the spec version. In light of https://github.com/mongodb/specifications/pull/713#discussion_r368152348, I suppose we could just go directly to 1.9.0 with this change since the diff here is more of a behavioral change.

source/auth/auth.rst
```diff
@@ -971,6 +971,12 @@ gssapiServiceName (deprecated)
 	An alias for ``authMechanismProperties=SERVICE_NAME:mongodb``.
 
 
+Errors
+------
+
+Drivers SHOULD raise an error if the ``authSource`` option is specified in the connection string with an empty value, e.g. ``mongodb://localhost/admin?authSource=``.
```
jmikola: Why SHOULD instead of MUST? Is there a concern that being prescriptive here would result in drivers raising errors for existing applications after they upgrade?

I noticed that https://github.com/mongodb/specifications/pull/714 is only addressing integer and boolean types. Any reason to omit string types there? It seems like the appropriate ticket to clarify that `foo=&...` be parsed as an empty string for the `foo` option (assuming it expects a string).

jmikola: In response to my second paragraph, I missed this comment from @jstewart-mongo in Slack:

> I glossed over this before, but the connection string spec contains the following in the section > Key Value Pair: Value (optional) -- the value if provided otherwise it defaults to an empty string.

Point noted on that, but my earlier question about MUST still stands.

jstewart-mongo: I did have the concern about the upgrade, but I do believe that an error should be raised, so I have changed the wording to use MUST.

source/auth/auth.rst
jstewart-mongo: I have bumped the version to 1.9.0.

source/auth/auth.rst
jmikola: > I have changed the wording to use MUST.

The PR still says "SHOULD". Has that change not been pushed?

source/auth/auth.rst
jmikola: Per https://github.com/mongodb/specifications/pull/713#discussion_r369150127, please update "Spec Version" in the header a few lines up.

source/auth/auth.rst
```diff
@@ -1152,6 +1158,8 @@ Version History
 
 Version 1.8.2 Changes
     * Added MONGODB-IAM auth mechanism
+    * Clarify that drivers should raise and error when a connection string
```
jmikola: I don't think this belongs under a previous version heading. Since you bumped the version to 1.9.0, there should probably be a new "Version 1.9.0 Changes" heading here.

Side note: I don't understand by introduction of MONGODB-IAM itself didn't warrant a minor version bump to 1.9.0 on its own, but that's a separate issue from https://github.com/mongodb/specifications/pull/697. I'm won't push that we correct that, but I did want to make a note of it for @bazile-clyde and @jyemin.

source/auth/auth.rst

source/auth/auth.rst
jstewart-mongo: I have revised the version history to include the new heading.

source/auth/auth.rst
```diff
@@ -1150,8 +1156,14 @@ Q: Why does SCRAM sometimes SASLprep and sometimes not?
 Version History
 ===============
 
+Version 1.9.0 Changes
+    * Clarify that drivers should raise and error when a connection string
```
p-mongo: "must raise an error" (must & an).

source/auth/auth.rst
p-mongo: I am not sure whether this hunk conflicts with changes elsewhere, but it looks like this is a copy of the 1.9.0 changelog and so thus hunk should be removed and just the paragraph under 1.9.0 left over.

jstewart-mongo: I have removed this text.

source/auth/auth.rst
jstewart-mongo: I have updated this text.

source/auth/auth.rst
p-mongo: It still appears to be in the diff. Also please merge/rebase this PR on master so that it is an accurate representation of what will ultimately be committed.

source/auth/tests/connection-string.yml
```diff
@@ -85,6 +85,10 @@ tests:
             mechanism: "GSSAPI"
             mechanism_properties:
                 SERVICE_NAME: "mongodb"
+    -
+        description: "should raise an error when the authSource is empty"
```
jmikola: Per the spec language change, this test should be renamed to "must raise an error..." (along with the JSON version).

source/auth/auth.rst
jstewart-mongo: I'm not sure how this change did not get pushed. The text is no longer in the diff.

jstewart-mongo:
@jyemin Please take a look.

source/auth/auth.rst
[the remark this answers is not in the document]
jmikola: I don't think it conflicts with SPEC-1511. https://github.com/mongodb/specifications/commit/af6b8b14606d43f1938a2a8da23801f05c6cf4ab doesn't care about the _value_ of `authSource` -- it just states that the presence of `authSource` alone is not sufficient to invoke authentication.

This issue is prohibiting an empty value for `authSource` in _any_ situation -- and if an error is raised (MUST per your other suggestion, which I support) then the driver would never even make it to the point where it might consider invoking auth.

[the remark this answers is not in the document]
jstewart-mongo: Done.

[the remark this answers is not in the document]
jyemin: Let's at least add a test that asserts that `mongodb://localhost/admin?authSource=` is also invalid.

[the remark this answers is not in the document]
jstewart-mongo: I have added a test with the suggested connection string.

jstewart-mongo:
@jyemin I have added the test you suggested. Please review.

PR #710, opened 2020-01-14
SPEC-1491: Deprecate oplogReplay find option

jmikola:
https://jira.mongodb.org/browse/SPEC-1491

### DIFF
--- a/source/crud/crud.rst
+++ b/source/crud/crud.rst
@@ -12,7 +12,7 @@ Driver CRUD API
 :Status: Approved
 :Type: Standards
 :Minimum Server Version: 2.6
-:Last Modified: January 10, 2020
+:Last Modified: January 14, 2020
 
 .. contents::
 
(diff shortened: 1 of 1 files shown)


source/crud/crud.rst
```diff
@@ -546,12 +546,16 @@ Read
...
 
     /**
-     * Internal replication use only - driver should not set
+     * Enables optimization when querying the oplog for a range of ts values
+     *
+     * Note: this option is intended for internal replication use only.
      *
      * This option is sent only if the caller explicitly provides a value. The default is to not send a value.
      * For servers < 3.2, the OplogReplay wire protocol flag is used and defaults to false.
+     * For servers >= 4.4, the server will ignore this option if set (see: SERVER-36186).
```
jmikola: [SERVER-36186](https://jira.mongodb.org/browse/SERVER-36186) notes that "the implementation can change to accept oplogReplay but entirely ignore its true/false value." Since this was released in 4.3.1, I will do a quick test to confirm that the server simply ignores this value. If it does raise a validation error, that's an unfortunate BC break on the server side and I expect we'll need to change our mild deprecation to actually force drivers to unset this option when sending a `find` command to a 4.3.1+ server.

jmikola: Confirmed with the PHP driver and APM that a 4.3.2 server doesn't mind if the driver sends an `oplogReplay` option when querying the `local.oplog.rs` collection. It seems to happily ignore it, so I don't think we need to mandate any functional driver changes here to pull it out of outgoing command docs.

PR #712, opened 2020-01-14
SPEC-1399 Handle absence of 'ns' field in index specifications

patrickfreed:
[SPEC-1399](https://jira.mongodb.org/browse/SPEC-1399)

Starting in MongoDB 4.4, the `ns` field will no longer be present in the index specifications returned from the `listIndexes` server command. This affects drivers that return statically defined models of index specifications as part of their own `listIndexes` helpers, as well as drivers that document the contents of any loose documents/dictionaries they return from that helper instead. 

This PR updates the index enumeration spec and test plan to describe how affected drivers should handle the `ns` field being missing.

After a brief survey, I _think_ the only affected drivers are PHP and legacy C#.

### DIFF
--- a/source/enumerate-indexes.rst
+++ b/source/enumerate-indexes.rst
@@ -365,8 +365,8 @@ Tests
   - verify all the indexes are represented in the result
   - verify the "unique" flags show up for the unique index
   - verify there are no duplicates in the returned list
-  - if the result consists of statically defined index models, verify that the ``ns`` fields are accurate if such a field is
-    defined as part of the model.
+  - if the result consists of statically defined index models that include an ``ns`` field, verify
+    that its value is accurate
 
 
 Backwards Compatibility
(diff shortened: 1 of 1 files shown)


source/enumerate-indexes.rst
```diff
@@ -353,13 +358,15 @@ Tests
...
 
 - Run the driver's method that returns a list of index information records, and:
 
-  - all the indexes are represented in the result
-  - the "unique" flags show up for the unique index
-  - there are no duplicates in the returned list
+  - verify all the indexes are represented in the result
+  - verify the "unique" flags show up for the unique index
+  - verify there are no duplicates in the returned list
+  - if the result consists of statically defined index models, verify that the ``ns`` fields are accurate if such a field is
```
jmikola: I think this could be rewritten as:

> if the result consists of statically defined index models that include an `ns` field, verify that its value is accurate

source/enumerate-indexes.rst
```diff
@@ -379,6 +386,9 @@ The shell implements the first algorithm for falling back if the
 
 Version History
 ===============
+0.5.0 - 2020-01-14
+    Update to mention handling of the ``ns`` field being removed from the
+    response to the ``listIndexes`` server command.
```
jmikola: I think it'd helpful to note the server version here. We can probably also make this more concise. Consider:

> MongoDB 4.4 no longer includes ``ns`` field in ``listIndexes`` responses.

source/enumerate-indexes.rst
patrickfreed: done

source/enumerate-indexes.rst
patrickfreed: done

PR #713, opened 2020-01-16
SPEC-1511: Lift restriction on authSource without credentials

jstewart-mongo:
Lift the restriction against authSource without credentials introduced in SPEC-1270.

### DIFF
--- a/source/auth/auth.rst
+++ b/source/auth/auth.rst
@@ -1154,10 +1154,12 @@ Q: Why does SCRAM sometimes SASLprep and sometimes not?
 Version History
 ===============
 
-Version 1.8.2 Changes
-    * Added MONGODB-IAM auth mechanism
+Version 1.8.3 Changes
     * Clarify that authSource in URI is not treated as a user configuring
       auth credentials.
+
+Version 1.8.2 Changes
+    * Added MONGODB-IAM auth mechanism
 
 Version 1.8.1 Changes
     * Clarify database to use for auth mechanism negotiation.
--- a/source/auth/auth.rst
+++ b/source/auth/auth.rst
@@ -6,7 +6,7 @@ Driver Authentication
 =====================
 
 :Spec: 100
-:Spec Version: 1.8.2
+:Spec Version: 1.8.3
 :Title: Driver Authentication
 :Author: Craig Wilson, David Golden
 :Advisors: Andy Schwerin, Bernie Hacket, Jeff Yemin, David Golden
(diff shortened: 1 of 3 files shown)


source/auth/auth.rst
```diff
@@ -101,6 +101,10 @@ name is only used as a default source for some mechanisms when authentication
 has been configured and a source is required but has not been specified.  See
 individual mechanism definitions for details.
 
+Similarly, The presence of the ``authSource`` option in the URI connection
```
jyemin: Nit, lowercase the `t` in `The`

source/auth/tests/connection-string.json
```diff
@@ -352,9 +352,10 @@
       "credential": null
     },
     {
-      "description": "authSource without username is invalid (default mechanism)",
+      "description": "authSource without username doesn't create credentials",
```
jyemin: Nit: `credentials` should be singular

source/auth/auth.rst
jstewart-mongo: Done.

source/auth/tests/connection-string.json
jstewart-mongo: Done

jstewart-mongo:
@xdg Thanks for the suggestion. I have renamed the section.

jstewart-mongo:
@jmikola Please take a look. Thanks.

source/auth/auth.rst
```diff
@@ -13,7 +13,7 @@ Driver Authentication
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: 2.6
-:Last Modified: 2020-01-13
+:Last Modified: 2020-01-16
```
jmikola: Spec version should be bumped for this. I think it's debatable if this warrants a bump to the patch or minor version. I'd be OK with a patch bump to 1.8.3 since it's more of a clarification than new functionality.

jstewart-mongo: I have bumped the version to 1.8.3.

jmikola: I see you added 1.8.3 to the "Version History" section below, but I was referring to the "Spec Version" a few lines up.

Side note: a "Version History" section for this spec seems inconsistent with the "Changelog" section in most of our other specs, which typically just has a date and a short description of the change. This isn't something to address now, though.

jstewart-mongo: I have updated the Spec Version.

PR #714, opened 2020-01-16
SPEC-1545: Clarify how empty values in a connection string are parsed

jstewart-mongo:
How an empty value is parsed for a string-based option is already covered in the section Key Value Pair in the connection string spec. For this spec, I am covering empty values for integer-based and boolean-based options.

### DIFF
--- a/source/connection-string/connection-string-spec.rst
+++ b/source/connection-string/connection-string-spec.rst
@@ -182,7 +182,7 @@ The values in connection options MUST be URL decoded by the parser. The values c
 - Integer:
   The value parsed as a integer. If the value is the empty string, then a WARN level logging message MUST be issued and the key MUST be ignored.
 - Boolean:
-  "true" and "false" strings MUST be supported. If the value is the empty string, then a WARN level logging message MUST be issued and the key is ignored. Drivers that cannot log warning messages or where warnings might be easily overlooked MUST raise an error.
+  "true" and "false" strings MUST be supported. If the value is the empty string, then a WARN level logging message MUST be issued and the key is ignored. Drivers that cannot log warning messages or where warnings might be easily overlooked MAY raise an error.
 
   - For legacy reasons it is RECOMMENDED that alternative values for true and false be supported:
 
--- a/source/connection-string/connection-string-spec.rst
+++ b/source/connection-string/connection-string-spec.rst
@@ -180,9 +180,9 @@ The values in connection options MUST be URL decoded by the parser. The values c
 - Strings:
     The value
 - Integer:
-    The value parsed as a integer
+    The value parsed as a integer. If the value is the empty string, then the value MUST be parsed as 0.
 - Boolean:
-    "true" and "false" strings MUST be supported.
+    "true" and "false" strings MUST be supported. If the value is the empty string, then the value MUST be parsed as false.
 
   - For legacy reasons it is RECOMMENDED that alternative values for true and false be supported:
 
--- a/source/connection-string/connection-string-spec.rst
+++ b/source/connection-string/connection-string-spec.rst
@@ -180,9 +180,9 @@ The values in connection options MUST be URL decoded by the parser. The values c
 - Strings:
     The value
 - Integer:
-  The value parsed as a integer. If the value is the empty string, then a WARN level logging message MUST be issued and the key MUST be ignored.
+  The value parsed as a integer. If the value is the empty string, then a WARN level logging message MUST be issued and the key MUST be ignored. Drivers that cannot log warning messages or where warnings might be easily overlooked MAY raise an error.
 - Boolean:
-  "true" and "false" strings MUST be supported. If the value is the empty string, then a WARN level logging message MUST be issued and the key is ignored. Drivers that cannot log warning messages or where warnings might be easily overlooked MAY raise an error.
+  "true" and "false" strings MUST be supported. If the value is the empty string, then a WARN level logging message MUST be issued and the key MUST be ignored. Drivers that cannot log warning messages or where warnings might be easily overlooked MAY raise an error.
 
   - For legacy reasons it is RECOMMENDED that alternative values for true and false be supported:
 
--- a/source/connection-string/connection-string-spec.rst
+++ b/source/connection-string/connection-string-spec.rst
@@ -180,9 +180,9 @@ The values in connection options MUST be URL decoded by the parser. The values c
 - Strings:
     The value
 - Integer:
-  The value parsed as a integer. If the value is the empty string, then a WARN level logging message MUST be issued and the key MUST be ignored. Drivers that cannot log warning messages or where warnings might be easily overlooked MAY raise an error.
+  The value parsed as a integer. If the value is the empty string, the key MUST be ignored.
 - Boolean:
-  "true" and "false" strings MUST be supported. If the value is the empty string, then a WARN level logging message MUST be issued and the key MUST be ignored. Drivers that cannot log warning messages or where warnings might be easily overlooked MAY raise an error.
+  "true" and "false" strings MUST be supported. If the value is the empty string, the key MUST be ignored.
 
   - For legacy reasons it is RECOMMENDED that alternative values for true and false be supported:
 
--- a/source/connection-string/connection-string-spec.rst
+++ b/source/connection-string/connection-string-spec.rst
@@ -180,9 +180,9 @@ The values in connection options MUST be URL decoded by the parser. The values c
 - Strings:
     The value
 - Integer:
-  The value parsed as a integer. If the value is the empty string, then a WARN level logging message must be issued and the key is ignored.
+  The value parsed as a integer. If the value is the empty string, then a WARN level logging message MUST be issued and the key MUST be ignored.
 - Boolean:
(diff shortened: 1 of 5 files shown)


ShaneHarvey:
Python's current behavior is to warn+ignore these options:
```
>>> client = MongoClient("mongodb://localhost/admin?maxIdleTimeMS=")
/Users/shane/git/mongo-python-driver/pymongo/common.py:756: UserWarning: maxidletimems must be an integer or float
  warnings.warn(str(exc))
>>> client = MongoClient("mongodb://localhost/admin?journal=")
/Users/shane/git/mongo-python-driver/pymongo/common.py:756: UserWarning: The value of journal must be 'true' or 'false'
  warnings.warn(str(exc))
```

I would prefer to keep this behavior instead of silently accepting empty strings. To me it seems like an empty string in the URI indicates a typo or bug. Raising a warning gives the user a clue that something is not configured correctly.

jstewart-mongo:
@ShaneHarvey I looked this over in the Java driver, and the Java driver also warns on empty values. I will update the spec to make that explicit and remove the changes for default values.

jstewart-mongo:
I have updated the spec to explicitly state that empty values for integer and boolean keys will generate a warning and will be ignored.

jstewart-mongo:
@jmikola @ShaneHarvey @jyemin @p-mongo Please take a look at the recent updates. Thanks.

source/connection-string/connection-string-spec.rst
```diff
@@ -180,9 +180,9 @@ The values in connection options MUST be URL decoded by the parser. The values c
 - Strings:
     The value
 - Integer:
-    The value parsed as a integer
+  The value parsed as a integer. If the value is the empty string, then a WARN level logging message must be issued and the key is ignored.
 - Boolean:
-    "true" and "false" strings MUST be supported.
+  "true" and "false" strings MUST be supported. If the value is the empty string, then a WARN level logging message must be issued and the key is ignored.
```
jmikola: I don't recall if this behavior being discussed in Slack, but I'm not a fan of logging a warning and ignoring the key, as this behavior differs significantly from a string option, where an empty string would be captured.

In this case, I'd much rather we raise an error and actually interrupt client construction. From the perspective of C and PHP, logging warnings are not a reliable way to inform our users, as they actually have to opt in to collecting log output.

Alternatively, I'd be OK with logging a warning and interpreting the value as the logical "empty" value, which would be zero or false for integers and booleans, respectively. In that case, we at least assign the option and ensure it will be picked up by some other driver component down the line. Ideally, it might even trigger a validation error and loudly inform the user.

My reasoning with preferring a conversion to zero/false over ignoring entirely is motivated by thinking that the user has already indicated they intend to _use_ this option and is more likely to just have forgotten to specify a value. If their intention was to not use the option at all I think they'd have been more likely to omit it entirely.

source/connection-string/connection-string-spec.rst
```diff
@@ -180,9 +180,9 @@ The values in connection options MUST be URL decoded by the parser. The values c
 - Strings:
     The value
 - Integer:
-    The value parsed as a integer
+  The value parsed as a integer. If the value is the empty string, then a WARN level logging message must be issued and the key is ignored.
```
p-mongo: Can MUST be capitalized in both sentences for clarity?

p-mongo: And I suggest also using the phrase "and the key MUST be ignored" for the same reason.

source/connection-string/connection-string-spec.rst
p-mongo: It is my impression that the warn-and-continue behavior was largely a part of the URI options specification, in the added spec tests (for example the numerous warn-and-continue tests in https://github.com/mongodb/specifications/blob/master/source/uri-options/tests/connection-options.yml). I am not seeing existing language in either the URI spec or the connection string spec stating how to handle unrecognized or invalid or missing values for URI options in general, however looking at the connection string spec I believe this test is prior art that ignores unrecognized option value:

https://github.com/mongodb/specifications/blob/master/source/connection-string/tests/valid-warnings.yml#L15

I generally support failing on invalid input (which includes missing value for boolean URI options in my mind) but I think the most consistent/compatible way of handling such values is to warn-and-continue.

> My reasoning with preferring a conversion to zero/false over ignoring entirely is motivated by thinking that the user has already indicated they intend to use this option and is more likely to just have forgotten to specify a value. 

I agree with this but I do not see how one can state that `false` was the intended value. For example, if a user gives `tls=&tlsInsecure=&retryReads=` did they mean to leave `tls` and `tlsInsecure` at their defaults (`false`) but change `retryReads` from `true` to `false`?

jmikola: > I think it makes more sense to interpret that as leaving all three options at their defaults (i.e. ignoring them in this case).

Fair point. I'd be OK with warning and ignoring in that case. Is there wiggle room to allow some drivers to be more strict and raise errors (for cases where warnings are basically useless)?

jstewart-mongo: Agree on this. Especially for options of integer type where the default value is non-zero, it would be safer to use the default value for the option.

source/connection-string/connection-string-spec.rst
jstewart-mongo: I have capitalized MUST and added the phrase you suggested.

source/connection-string/connection-string-spec.rst
```diff
@@ -180,9 +180,9 @@ The values in connection options MUST be URL decoded by the parser. The values c
 - Strings:
     The value
 - Integer:
-    The value parsed as a integer
+  The value parsed as a integer. If the value is the empty string, then a WARN level logging message MUST be issued and the key MUST be ignored.
```
jmikola: This warrants bumping the spec version and adding a changelog entry. Since this is a functional change an introducing new tests, I think a minor version bump is appropriate.

jstewart-mongo: I have bumped the minor version and added a changelog entry.

source/connection-string/connection-string-spec.rst
jstewart-mongo: We should allow drivers that cannot warn to raise errors. How about this wording:
_If the value is the empty string, then a WARN level logging message MUST be issued and the key is ignored. Drivers that cannot log warning messages or where warnings are meaningless MUST raise an error._

jmikola: Replace "are meaningless" with "might be easily overlooked" and that would work for me.

jstewart-mongo: I have updated the wording.

source/connection-string/connection-string-spec.rst
```diff
@@ -180,9 +180,9 @@ The values in connection options MUST be URL decoded by the parser. The values c
...
 - Boolean:
-    "true" and "false" strings MUST be supported.
+  "true" and "false" strings MUST be supported. If the value is the empty string, then a WARN level logging message MUST be issued and the key is ignored. Drivers that cannot log warning messages or where warnings might be easily overlooked MUST raise an error.
```
p-mongo: I think "MAY raise an error" is the appropriate language here since it refers to a permitted deviation.

jstewart-mongo: I have updated the language.

source/connection-string/connection-string-spec.rst
jmikola: I think this needs the same "MAY raise an error" sentence as we have for "Boolean".

source/connection-string/connection-string-spec.rst
```diff
@@ -180,9 +180,9 @@ The values in connection options MUST be URL decoded by the parser. The values c
 - Strings:
     The value
 - Integer:
-    The value parsed as a integer
+  The value parsed as a integer. If the value is the empty string, then a WARN level logging message MUST be issued and the key MUST be ignored. Drivers that cannot log warning messages or where warnings might be easily overlooked MAY raise an error.
```
jyemin: I'm having a hard time figuring how to apply the criteria of `where warnings might be easily overlooked`.  This doesn't seem specific to a driver, but rather to how logging is configured and what users do with their application log files, which is out of our control. Some applications upload logs to Splunk and alert on warnings, some log to /dev/null.  We just don't know.

I also think that leaving open this back door of allowing some drivers to error while most log goes against the goal of having portable connection strings.

jyemin: I would be fine just saying:

> If the value is the empty string, the key MUST be ignored

and leave the logging as an implementation detail.  I generally don't like having MUSTs about logging anyway, since we have no way of asserting the behavior.

jstewart-mongo: I have updated the language, removing the comments on logging.

jyemin: Well, I'm ok with it now, but I suspect other reviewers will not be! @jmikola, @p-mongo, would you like to weigh in?

jmikola: I'm OK with leaving it unspecified, but we could also consider "Drivers MAY log a warning to inform users that the key is being ignored." Will defer to you all on this, but spec LGTM as-is, too.

jstewart-mongo: I would prefer to leave it as-is.

PR #716, opened 2020-01-17
SPEC-1418: Note allowDiskUse find option is ignored for OP_QUERY

jmikola:
https://jira.mongodb.org/browse/SPEC-1418

Requested in https://github.com/mongodb/specifications/pull/692#discussion_r367667587 but this was missed before that PR was merged.

### DIFF
--- a/source/crud/crud.rst
+++ b/source/crud/crud.rst
@@ -412,7 +412,8 @@ Read
      * can write temporary data to disk while executing the find operation.
      *
      * This option is sent only if the caller explicitly provides a value. The default
-     * is to not send a value.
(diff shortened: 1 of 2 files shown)


rstam:
Note: do any drivers even send OP_QUERY message any more? (other than for the initial isMaster handshake).

It's only required when connected to server 2.4 or older, and we no longer support connecting to 2.4 servers and olders (min wire version was bumped to 2 a long time ago).

jmikola:
@rstam The [find](https://docs.mongodb.com/manual/reference/command/find/) command was introduced in 3.2. Drivers and applications that connect to a 2.6 or 3.0 server would be affected by this.

PR #718, opened 2020-01-24
SPEC-1534 Remove requirement that application-provided serverSelector…

prashantmital:
… NOT be called when no suitable servers are found

Closes https://jira.mongodb.org/browse/SPEC-1534

### DIFF
--- a/source/server-selection/server-selection.rst
+++ b/source/server-selection/server-selection.rst
@@ -1173,7 +1173,7 @@ Pseudocode for `multi-threaded or asynchronous server selection`_::
 
             servers = all servers in topologyDescription matching criteria
 
-            if serverSelector is not null and servers is not empty:
+            if serverSelector is not null:
                 servers = serverSelector(servers)
 
             if servers is not empty:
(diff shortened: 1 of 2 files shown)


source/server-selection/server-selection.rst
```diff
@@ -351,9 +351,9 @@ serverSelector
 ~~~~~~~~~~~~~~
 
 Implementations MAY allow configuration of an optional, application-provided function
-that augments the server selection rules.  The function takes as a parameter a list
+that augments the server selection rules.  The function takes as a parameter an iterable
```
jyemin: I don't think we need to change this from list to iterable.  It's not like we're requiring any existing implementations to change their type signatures.

rstam: So you are saying that in the spec "list" is to be interpreted loosely and means any platform specific type that is iterable?

It's certainly not going to be a `List<ServerDescrption>` in the C# driver.

It's an `IEnumerable<ServerDescription>` where `IEnumerable` is the C# name for `iterable`.

jyemin: Yes, list should be considered loosely.  Changing it to enumerable from list implies that we want drivers to make a change.  Also, list feels more generic to me and less like an an actual defined type (even though it is often a defined type).

PR #719, opened 2020-01-24
SPEC-1421: support shorter SCRAM conversation

jstewart-mongo:


source/auth/auth.rst
```diff
@@ -613,13 +613,7 @@ This same conversation over mongodb's sasl implementation would appear as follow
 | C: :javascript:`{saslStart: 1, mechanism: "SCRAM-SHA-1", payload: BinData(0, "biwsbj11c2VyLHI9ZnlrbytkMmxiYkZnT05Sdjlxa3hkYXdM")}`
```
jyemin: In order for it to work this way the driver has to send `options: { skipEmptyExchange: true }`

jyemin: Earlier versions of the server will ignore this, and the longer conversation will be used instead, and drivers have to be able to handle both.

source/auth/auth.rst
```diff
@@ -1093,6 +1087,17 @@ Drivers may need to remove support for association of more than one credential w
 	* Deprecation and removal of MongoClient constructors that take as an argument more than a single credential
 	* Deprecation and removal of methods that allow lazy authentication (i.e post-MongoClient construction)
 
+Drivers need to support both the shorter and longer SCRAM conversations over mongodb's sasl implementation. Earlier versions of the server required an extra round trip due to an implementation decision. This was accomplished by sending no bytes back to the server, as seen in the following conversation (extra round trip italicized):
+
+| C: :javascript:`{saslStart: 1, mechanism: "SCRAM-SHA-1", payload: BinData(0, "biwsbj11c2VyLHI9ZnlrbytkMmxiYkZnT05Sdjlxa3hkYXdM")}`
```
jyemin: This should also contain `options: { skipEmptyExchange: true }`.  Older server versions will just ignore it iirc.

source/auth/auth.rst
jyemin: See https://jira.mongodb.org/browse/SERVER-44857 for details.

sgolemon: ☝️ What @jyemin said.  I think if you add `options: { skipEmptyExchange: true }` here (and not below), then this diff should be fine.

We probably want to leave it out of the lower place since that spot is illustrating what a "long" exchange looks like.

source/auth/auth.rst
```diff
@@ -1093,6 +1087,17 @@ Drivers may need to remove support for association of more than one credential w
...
+| S: :javascript:`{conversationId : 1, payload: BinData(0,"cj1meWtvK2QybGJiRmdPTlJ2OXFreGRhd0xIbytWZ2s3cXZVT0tVd3VXTElXZzRsLzlTcmFHTUhFRSxzPXJROVpZM01udEJldVAzRTFURFZDNHc9PSxpPTEwMDAw"), done: false, ok: 1}`
+| C: :javascript:`{saslContinue: 1, conversationId: 1, payload: BinData(0, "Yz1iaXdzLHI9ZnlrbytkMmxiYkZnT05Sdjlxa3hkYXdMSG8rVmdrN3F2VU9LVXd1V0xJV2c0bC85U3JhR01IRUUscD1NQzJUOEJ2Ym1XUmNrRHc4b1dsNUlWZ2h3Q1k9")}`
+| S: :javascript:`{conversationId: 1, payload: BinData(0,"dj1VTVdlSTI1SkQxeU5ZWlJNcFo0Vkh2aFo5ZTA9"), done: false, ok: 1}`
+| *C: :javascript:`{saslContinue: 1, conversationId: 1, payload: BinData(0, "")}`*
+| *S: :javascript:`{conversationId: 1, payload: BinData(0,""), done: true, ok: 1}`*
+
+The extra round trip has been removed in server version 4.4.
```
sgolemon: > The extra round trip has been removed in server version 4.4...

...when `options: { skipEmptyExchange: true }` is specified during `saslStart`

jmikola:
FYI: @sgolemon also opened https://jira.mongodb.org/browse/CDRIVER-3506 with an implementation of this if we'd like to consider it as a second POC.

source/auth/auth.rst

source/auth/auth.rst
jstewart-mongo: I have added the `options` document.

source/auth/auth.rst
jstewart-mongo: I have updated the sentence with your suggested condition.

source/auth/auth.rst
jyemin: I would leave it in the "long" exchange because the driver is always going to add it

source/auth/auth.rst
```diff
@@ -6,14 +6,14 @@ Driver Authentication
 =====================
 
 :Spec: 100
-:Spec Version: 1.9.1
+:Spec Version: 1.9.2
```
jyemin: This is worth a bump of the minor version

source/auth/auth.rst
```diff
@@ -1158,6 +1163,9 @@ Q: Why does SCRAM sometimes SASLprep and sometimes not?
 Version History
 ===============
 
+Version 1.9.2 Changes
```
jyemin: As above.  It's worth a minor version bump

source/auth/auth.rst
```diff
@@ -610,16 +610,10 @@ follows:
 
 This same conversation over mongodb's sasl implementation would appear as follows:
 
-| C: :javascript:`{saslStart: 1, mechanism: "SCRAM-SHA-1", payload: BinData(0, "biwsbj11c2VyLHI9ZnlrbytkMmxiYkZnT05Sdjlxa3hkYXdM")}`
+| C: :javascript:`{saslStart: 1, mechanism: "SCRAM-SHA-1", payload: BinData(0, "biwsbj11c2VyLHI9ZnlrbytkMmxiYkZnT05Sdjlxa3hkYXdM"), options: { skipEmptyExchange: true }}`
```
jyemin: Let's copy this section (with appropriate changes to the payloads) to the SCRAM-SHA-256 section, so it's clear that drivers should implement skipEmptyExchange for that mechanism too.

source/auth/auth.rst
```diff
@@ -1093,6 +1087,17 @@ Drivers may need to remove support for association of more than one credential w
 	* Deprecation and removal of MongoClient constructors that take as an argument more than a single credential
 	* Deprecation and removal of methods that allow lazy authentication (i.e post-MongoClient construction)
 
+Drivers need to support both the shorter and longer SCRAM conversations over mongodb's sasl implementation. Earlier versions of the server required an extra round trip due to an implementation decision. This was accomplished by sending no bytes back to the server, as seen in the following conversation (extra round trip italicized):
```
jyemin: For clarity:    `SCRAM` => `SCRAM-SHA-1 and SCRAM-SHA-256`

source/auth/auth.rst

source/auth/auth.rst
jyemin: Nit: it should be 1.10.0

source/auth/auth.rst
jyemin: 1.10.0

source/auth/auth.rst
jyemin: @sgolemon can you re-review please?

source/auth/auth.rst
jstewart-mongo: Updated.

source/auth/auth.rst
jstewart-mongo: Updated.

source/auth/auth.rst
jyemin: Reading this again, I think that it's too easy to miss this new options thing, as it's in a section that's marked "as an example".  Let's add a new paragraph right above this Conversation section that says something like:

     Drivers MUST add a top-level options field to the saslStart command, whose value is a document containing a field named skipEmptyExchange whose value is true.  Older servers will ignore the options and continue with the longer exchange shown in the [Backwards Compatibility] section.  Newer servers will ... 

and so on.   A bit awkwardly phrased but maybe you can fix that.

I'm not sure whether to copy that paragraph to SCRAM-SHA-256.  It looks like some of the text from above is duplicated for SCRAM-SHA-256 and some is not.

source/auth/auth.rst
```diff
@@ -673,6 +667,13 @@ follows:
 | C: ``c=biws,r=rOprNGfwEbeRWgbNEkqO%hvYDpWUa2RaTCAfuxFIlj)hNlF$k0,p=dHzbZapWIk4jUhN+Ute9ytag9zjfMHgsqmmiz7AndVQ=``
 | S: ``v=6rriTRBi23WpRR/wtup+mMhUZUn/dB5nLTJRsjl95G4=``
 
+This same conversation over mongodb's sasl implementation would appear as follows:
```
jmikola: "MongoDB's SASL"

source/auth/auth.rst
```diff
@@ -1093,6 +1094,17 @@ Drivers may need to remove support for association of more than one credential w
 	* Deprecation and removal of MongoClient constructors that take as an argument more than a single credential
 	* Deprecation and removal of methods that allow lazy authentication (i.e post-MongoClient construction)
 
+Drivers need to support both the shorter and longer SCRAM-SHA-1 and SCRAM-SHA-256 conversations over mongodb's sasl implementation. Earlier versions of the server required an extra round trip due to an implementation decision. This was accomplished by sending no bytes back to the server, as seen in the following conversation (extra round trip italicized):
```
jmikola: "MongoDB's SASL"

source/auth/auth.rst
jstewart-mongo: I have added the paragraph to both the SCRAM-SHA-1 and SCRAM-SHA-256 sections.

source/auth/auth.rst
jstewart-mongo: Updated.

source/auth/auth.rst
jstewart-mongo: Updated.

PR #720, opened 2020-01-28
SPEC-1352 test omitting default write concern

kevinAlbs:
**Notes to reviewers**
I considered write commands that do not have drivers helpers to all be covered by one run command operation (for `create`). That includes:

- cleanupOrphaned
- cloneCollection
- cloneCollectionAsCapped
- create
- convertToCapped
- createRole
- createUser
- dropAllRolesFromDatabase
- dropAllUsersFromDatabase
- dropRole
- dropUser
- grantPrivilegesToRole
- grantRolesToRole
- grantRolesToUser
- moveChunk
- renameCollection
- revokePrivilegesFromRole
- revokeRolesFromRole
- revokeRolesFromUser
- updateRole
- updateUser
- dropDatabase

The following commands are already tested with default write concern (with the "null" expectation) in [write-concern.yml](https://github.com/mongodb/specifications/blob/0f803725487cc951877f6975061dd1e24056a045/source/transactions/tests/write-concern.yml#L75) transactions tests.
- abortTransaction
 - commitTransaction

The following server commands were tested:
- aggregate
    - with $out/$merge
- createIndexes
- dropIndexes
- delete
    - via deleteOne, deleteMany, bulkWrite
- findAndModify
    - via findOneAndUpdate, findOneAndDelete, findOneAndReplace
- insert
    - via insertOne, insertMany, bulkWrite
- mapReduce
- update
    - via updateOne, updateMany, replaceOnce, bulkWrite

The following are TBD:
- drop
    I think GridFS bucket drop is the only specified driver helper. Unsure if I should introduce new spec syntax around that.

`createIndex` and `dropIndex` are new spec test operations. The naming is based on the "Standard API" from the Index Management spec but should be implement-able from the Index View API.

PoC'ed with libmongoc (excluding mapReduce and create/drop index operations, which libmongoc does not have)

**Open questions**
1. If this approach seems ok, should I create corresponding tests for omitting the default read concern?
2. Should I test GridFS bucket operations? In this CRUD test runner (and introduce similar constructs from retryable reads)?


source/crud/tests/v2/default-write-concern.yml
```diff
@@ -0,0 +1,413 @@
...
+          command:
+            create: test_collection
+            writeConcern: null
+  - description: create and drop index
+    operations:
+      - object: collection
+        name: createIndex
+        arguments:
+          model:
+            x: 1
```
ShaneHarvey: Is this the first time that we are testing createIndex and createIndexes? What do you think about defining the argument format in the readme? I have no preference between the two argument forms presented here (`model: {x:1}` and  `{name: _x_1, key: {x:1}}`) but I do think we should standardize on one of them.

jmikola: Assuming this refers to the [Index Management](https://github.com/mongodb/specifications/blob/master/source/index-management.rst) spec, I'd suggest we stick to the `createIndex(keys: Document, options: IndexOptions)` API here. I think the `model` API is less common.

source/crud/tests/v2/default-write-concern.yml
```diff
@@ -0,0 +1,413 @@
+data:
+  - _id: 1
+    x: 11
+  - _id: 2
+    x: 22
+collection_name: default_write_concern_coll
+database_name: default_write_concern_db
+tests:
+  - description: Aggregate with $merge
```
jmikola: I think this will need to be moved to a separate test file, since [`$merge`](https://docs.mongodb.com/manual/reference/operator/aggregation/merge/) was introduced in 4.2. For simplicity, you may want to have a common `default-write-concern.yml` test file and then split certain tests with `minServerVersion` requirements into separate files (e.g. `default-write-concern-4.2.yml`).

The basic insert, update, and delete commands wouldn't need a `minServerVersion`, but you'll definitely want to split files findAndModify (3.2+) and other write commands (3.4+), per https://github.com/mongodb/specifications/blob/master/source/wireversion-featurelist.rst.

source/crud/tests/v2/default-write-concern.yml
```diff
@@ -0,0 +1,413 @@
+data:
+  - _id: 1
+    x: 11
+  - _id: 2
+    x: 22
+collection_name: default_write_concern_coll
```
jmikola: Consider the formatting used in other `v2` tests, for conciseness. Small documents can be expressed on one line (e.g. `- { _id: 1, x: 11 }`). Likewise for arguments and pipelines in actual tests.

Also, the collection and database names can be turned into references to reduce duplication of the original string. For instance, `aggregate-merge.yml` uses `collection_name: &collection_name 'test_aggregate_merge'`.

source/crud/tests/v2/default-write-concern.yml
```diff
@@ -0,0 +1,413 @@
...
+                _id:
+                  $gt: 1
+            - $merge:
+                into: other_test_collection
+    expectations:
+      - command_started_event:
+          command:
+            aggregate: default_write_concern_coll
+            pipeline: *ref_0
+            writeConcern: null
```
jmikola: Just to confirm, this is leveraging the recently added [Fields that must NOT be present in Actual Documents](https://github.com/mongodb/specifications/tree/master/source/crud/tests#fields-that-must-not-be-present-in-actual-documents) section of the test README, correct?

source/crud/tests/v2/default-write-concern.yml
```diff
@@ -0,0 +1,413 @@
...
+        name: runCommand
+        arguments:
+          command:
+            create: test_collection
+    expectations:
+      - command_started_event:
+          command:
+            create: test_collection
+            writeConcern: null
+  - description: create and drop index
```
jmikola: I would suggest adding "omits default write concern" as a common suffix to these tests, since some language test runners may report tests using the name in their descriptions.

source/crud/tests/v2/default-write-concern.yml
```diff
@@ -0,0 +1,413 @@
...
+              - name: _x_1
+                key:
+                  x: 1
+            writeConcern: null
+      - command_started_event:
+          command:
+            dropIndexes: default_write_concern_coll
+            index: _x_1
+            writeConcern: null
+  - description: delete one
```
jmikola: Existing tests refer to this the operation as a capitalized, camel-case term (DeleteOne in this case). So I'd suggest using "DeleteOne omits default write concern" here. FWIW, the existing Aggregate tests are already consistent with this, but I think the index ones could be "CreateIndex and DropIndex omit default write concern".

source/crud/tests/v2/default-write-concern.yml
```diff
@@ -0,0 +1,413 @@
...
+              x: 2
+            writeConcern: null
+      - command_started_event:
+          command:
+            findAndModify: default_write_concern_coll
+            query:
+              _id: 2
+            remove: true
+            writeConcern: null
+  - description: 'insert{one,many}'
```
jmikola: "InsertOne and InsertMany omit default write concern" may read better for drivers reporting descriptions.

source/crud/tests/v2/default-write-concern.yml
```diff
@@ -0,0 +1,413 @@
...
+            writeConcern: null
+      - command_started_event:
+          command:
+            delete: default_write_concern_coll
+            deletes:
+              - q:
+                  _id: 3
+                limit: 1
+            writeConcern: null
+  - description: 'findOneAnd{update,delete,replace}'
```
jmikola: "findAndModify operations omit default write concern".

Note that this test would require MongoDB 3.2+ (per https://github.com/mongodb/specifications/blob/master/source/wireversion-featurelist.rst).

source/crud/tests/v2/default-write-concern.yml
```diff
@@ -0,0 +1,413 @@
...
+              - _id: 3
+            writeConcern: null
+      - command_started_event:
+          command:
+            insert: default_write_concern_coll
+            documents:
+              - _id: 4
+              - _id: 5
+            writeConcern: null
+  - description: mapReduce
```
jmikola: Given that mapReduce, runCommand, createIndex, and dropIndex are not actually defined in the CRUD spec, I wonder if we should relocate the tests in this PR to a new sub-directory in https://github.com/mongodb/specifications/tree/master/source/read-write-concern/tests.

I think the intention of [SPEC-1352](https://jira.mongodb.org/browse/SPEC-1352) was not to test that drivers omit a `writeConcern` option by default, but that they specifically detect an empty write concern provided at the client/database/collection level and then omit it in outgoing commands. To actually test that, I think you'd want to make use of the `clientOptions` or `collectionOptions` keys defined in the [transaction spec test format](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst). Note that CRUD doesn't define those keys at all.

If you were to move these tests over to the write concern spec, then you could also address this by some small copypasta referencing the transaction spec test format, as is done in https://github.com/mongodb/specifications/blob/master/source/sessions/tests/README.rst#test-format. That would keep your new test README quite small.

source/crud/tests/v2/default-write-concern.yml
kevinAlbs: Correct. Noting in a comment.

source/crud/tests/v2/default-write-concern.yml
```diff
@@ -0,0 +1,413 @@
...
+            aggregate: default_write_concern_coll
+            pipeline: *ref_1
+            writeConcern: null
+  - description: runCommand with a write command
+    operations:
+      - object: database
+        name: runCommand
+        arguments:
+          command:
+            create: test_collection
```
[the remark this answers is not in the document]
kevinAlbs: Fixed.

source/crud/tests/v2/default-write-concern.yml
kevinAlbs: Done.

source/crud/tests/v2/default-write-concern.yml
kevinAlbs: Done.

source/crud/tests/v2/default-write-concern.yml
kevinAlbs: Done.

source/crud/tests/v2/default-write-concern.yml
kevinAlbs: Done.

source/crud/tests/v2/default-write-concern.yml
kevinAlbs: Changed to `keys`. I did not need to specify any index options for this test since it assumes the default `x_1` index name.

source/read-write-concern/tests/README.rst
```diff
@@ -52,6 +52,26 @@ array of test case objects, each of which have the following keys:
...
+- ``clientOptions`` may include a ``writeConcern``, distinct from the URI option ``w``, to indicate a client level write concern.
+- ``writeConcern`` in the ``clientOptions``, ``databaseOptions``, or ``collectionOptions`` may be an empty document to indicate a `server default write concern <https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#servers-default-writeconcern>`_. For example, in libmongoc:
+
+    .. code:: c
+
+       /* Create a default write concern, and set on a collection object. */
+       mongoc_write_concern_t *wc = mongoc_write_concern_new ();
+       mongoc_collection_set_write_concern (collection, wc);
+
+    If the driver has no way to explicitly set a default write concern on a client, database, or collection, then take no action.
```
kevinAlbs: PyMongo for example does not have a way to specify a default write concern on a MongoClient AFAICT, so I don't think I can require this. If this seems like overkill, I can remove optionally setting the default write concern in the ClientOptions.

source/crud/tests/v2/default-write-concern.yml
kevinAlbs: I like that idea. I added a section to read-write-concern/tests/README.rst describing the spec tests.

source/read-write-concern/tests/operation/default-write-concern-3.2.yml
```diff
@@ -0,0 +1,125 @@
...
+            findAndModify: *collection_name
+            query: {_id: 1}
+            update: {$set: {x: 1}}
+            writeConcern: null
+  - description: 'findOneAndUpdate with collection write concern unset omits default write concern'
+    clientOptions: {writeConcern: {}}
+    operations:
+      - name: findOneAndUpdate
+        object: collection
+        databaseOptions: {writeConcern: {}}
```
jmikola: Test title says "collection write concern" but this assigns a database-level WC. Is this backwards with the previous test, which says "database write concern" and sets collection-level?

source/read-write-concern/tests/operation/default-write-concern-3.2.yml
```diff
@@ -0,0 +1,125 @@
...
+    - minServerVersion: "3.2"
+
+tests:
+  - description: 'findAndModify operations omit default write concern'
+    clientOptions: {writeConcern: {}}
+    operations:
+      - name: findOneAndUpdate
+        object: collection
+        databaseOptions: {writeConcern: {}}
+        collectionOptions: {writeConcern: {}}
```
jmikola: What is the logic in applying a WC at all three levels? Presumably the collection option would overwrite any others, so isn't this redundant? I think this is definitely the case with database and collection options, although I imagine libmongoc might handle client-level WC separately.

source/read-write-concern/tests/operation/default-write-concern-3.4.yml
```diff
@@ -0,0 +1,197 @@
...
+    operations:
+      - name: mapReduce
+        object: collection
+        databaseOptions: {writeConcern: {}}
+        collectionOptions: {writeConcern: {}}
+        arguments:
+          map:
+            $code: 'function inc() { return emit(0, this.x + 1) }'
+          reduce:
+            $code: >-
```
jmikola: Is `>-` syntax for treating the value as a string and trimming any leading/trailing whitespace?

source/read-write-concern/tests/README.rst
```diff
@@ -52,6 +52,26 @@ array of test case objects, each of which have the following keys:
...
 - ``isAcknowledged:`` Indicates if the write concern should be considered acknowledged.
 
+Operation
+~~~~~~~~~
+
+These tests check that the default write concern is omitted in operations.
+
+The spec test format is an extension of `transactions spec tests <https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst>`_ with the following additions:
+
+- ``clientOptions`` may include a ``writeConcern``, distinct from the URI option ``w``, to indicate a client level write concern.
```
jmikola: I think this may cause more trouble than its worth, since most test runners currently expect `clientOptions` to contain known URI options. Related to my earlier comment about possible redundancy specifying client, database, _and_ collection options can we just use database _or_ collection options depending on the object used to execute the operation? I think `runCommand` would be the only thing requiring database options in that case.

source/read-write-concern/tests/README.rst
jmikola: I think `clientOptions` may be more trouble than its worth, since most test runners currently expect `clientOptions` to contain known URI options. Related to my earlier comment about possible redundancy specifying client, database, _and_ collection options can we just use database _or_ collection options depending on the object used to execute the operation? I think `runCommand` would be the only thing requiring database options in that case.

kevinAlbs: SGTM. Done.

source/read-write-concern/tests/operation/default-write-concern-3.2.yml
kevinAlbs: Yeah it is redundant. The collection write concern should overwrite. I suppose I was thinking a bug in the inheritance logic could result in differing behavior, so that it'd be safer to specify it everywhere. But I don't think it's a strong case. I've changed it to only apply at the collection or database level depending on the operation.

source/read-write-concern/tests/operation/default-write-concern-3.2.yml
kevinAlbs: It was intended, as the collectionOptions is the one that is unset. But this is no longer applicable now that it's changed to only set at one level.

source/read-write-concern/tests/operation/default-write-concern-3.4.yml
kevinAlbs: According to https://yaml-multiline.info/ says `>` folds lines with single spaces, and the `-` means to add no newline at the end.

Regardless, I changed it back to follow suit with how it is specified in the mapReduce.yml. I regret ever passing these files through a YML formatter...

source/read-write-concern/tests/README.rst
kevinAlbs: Removed, and simplified.

source/read-write-concern/tests/operation/default-write-concern-2.6.yml
```diff
@@ -0,0 +1,231 @@
...
+            updates:
+              - {q: {_id: 2}, u: {$set: {x: 2}}, multi: true}
+            writeConcern: null
+      - command_started_event:
+          command:
+            update: *collection_name
+            updates:
+              - {q: {_id: 2}, u: {x: 3}}
+            writeConcern: null
+  - description: DeleteOne with write concern unset omits default write concern
```
jmikola: Why does this "with write concern unset omits default write concern" test only exist for DeleteOne? Is it leftover from a previous PR revision and something that should be deleted, or did you just arbitrarily decide to test this for one operation? I noted that the findAndModify test only does this for findOneAndUpdate, so perhaps it's the latter.

I'm wondering if this might be a fragile test if a test runner decides to use some environment variable for its MongoClient (in the absence of `clientOptions` in the test itself), which could specify its own write concern. Perhaps that's not a concern since the [Transactions test runner](https://github.com/mongodb/specifications/tree/master/source/transactions/tests) instructs using a fresh MongoClient for each test.

source/read-write-concern/tests/operation/default-write-concern-3.4.yml
```diff
@@ -0,0 +1,113 @@
...
+  - description: Aggregate with $out omits default write concern
+    operations:
+      - object: collection
+        collectionOptions: {writeConcern: {}}
+        name: aggregate
+        arguments:
+          pipeline: &out_pipeline
+            - $match: {_id: {$gt: 1}}
+            - $out: *other_collection_name
+          batchSize: 2
```
jmikola: I don't think `batchSize` is relevant when using `$out`, so perhaps this can be deleted. It's also not mentioned in your APM assertion.

Likewise for the other aggregate tests later in this file. The `$merge` test in the other file doesn't use `batchSize`, so that's fine.

source/read-write-concern/tests/operation/default-write-concern-2.6.yml
kevinAlbs: It is the latter. Originally the tests were all like this, with no write concern specified on client/database/collection. In our later discussions, we decided to switch to testing with an explicit write concern set on a client/database/collection. In the slack discussion between the three of us, I proposed we still have one test of without an explicit default write concern set. Though, like the tests of a write concern unset on the client/database/collection, I think this test is not very valuable. Removed.

source/read-write-concern/tests/operation/default-write-concern-3.4.yml
kevinAlbs: Yeah this was copy-pasta from aggregate-out.yml in CRUD. Removed.

kevinAlbs:
I removed the "with write concern unset omits default write concern" tests, since they likely don't provide additional coverage. Let me know if you'd like to take another look.

jmikola:
I'm still LGTM. Over to @ShaneHarvey

source/read-write-concern/tests/operation/default-write-concern-3.4.yml
```diff
@@ -0,0 +1,92 @@
...
+        name: *other_collection_name
+        data:
+          - {_id: 2, x: 22}
+    expectations:
+      - command_started_event:
+          command:
+            aggregate: *collection_name
+            pipeline: *out_pipeline
+            writeConcern: null
+  - description: RunCommand with a write command omits default write concern
```
ShaneHarvey: This test, while correct, is a little misleading. The generic runCommand is never supposed to add writeConcern so it should omit non-default writeConcerns too: https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#id6
> If your driver offers a generic RunCommand method on your database object, WriteConcern MUST NOT be applied automatically to any command. A user wishing to use a WriteConcern in a generic command must manually include it in the command document passed to the method.

Also (assuming we keep this test) can we use a drop command instead of create? That will ensure that the test can be run multiple times without needing to drop the database.

kevinAlbs: Good point. Changed to a `drop` command and added to the description to note that runCommand should never inherit a write concern.

jmikola: Good catch, @ShaneHarvey!

ShaneHarvey: Whoops, I just learned that dropping a non-existent collection will fail with "ns not found". Can we change this to run a run-op delete command (or something else that can be run multiple times in a row)? Sorry for the churn.

kevinAlbs: Updated. Oddly, libmongoc's spec test runner actually ignored the ns not found error. I think that might be a bug on our end.

source/read-write-concern/tests/README.rst
```diff
@@ -52,6 +52,25 @@ array of test case objects, each of which have the following keys:
...
+
+- ``writeConcern`` in the ``databaseOptions`` or ``collectionOptions`` may be an empty document to indicate a `server default write concern <https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#servers-default-writeconcern>`_. For example, in libmongoc:
+
+    .. code:: c
+
+       /* Create a default write concern, and set on a collection object. */
+       mongoc_write_concern_t *wc = mongoc_write_concern_new ();
+       mongoc_collection_set_write_concern (collection, wc);
+
+    If the driver has no way to explicitly set a default write concern on a database or collection, then take no action.
```
ShaneHarvey: Can you clarify if "take no action" means skip the test or if it means ignore the writeConcern and continue with the test?

kevinAlbs: Good point, updated to say drivers should continue with the test.

PR #722, opened 2020-01-29
Fix SPEC-1517 Drivers must run SDAM flow even when server description is equal to the last one


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/SPEC-1517

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -850,18 +850,28 @@ are parsed from the ismaster response in the obvious way.
...
+are respectively equal as just defined.
 
 This specification does not prescribe how to compare server descriptions
 with different addresses for equality.
 
-Note: Server description for each server MUST be updated (replaced)
-every heartbeat. However, new description MUST NOT cause the SDAM flow
-to be executed if the new description is equal, as defined in this section,
-to the previous description.
+Server description for each server MUST be updated (replaced) on every
```
mbroadst: nit: "The server description"

mbroadst: I think we can remove this entire section because we already explicitly define what should be done when [parsing an ismaster response](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#updating-the-topologydescription) ("Each time the client checks a server, it processes the outcome (successful or not) to create a ServerDescription, and then it processes the ServerDescription to update its TopologyDescription."). 

The goal of this section should be to define when a driver can use `ServerDescription` equality to determine which events should be published.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -2373,6 +2383,33 @@ monitoring can also proactively detect state changes
 (primary stepdown, server becoming unavailable)
 that would otherwise cause future errors.
 
+Why is the driver required to run SDAM flow if server description has not changed?
```
mbroadst: I'm not sure we need this rational if we're already following the guidance in "Updating the TopologyDescription", the unit test should be enough to validate the behavior.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
p-mongo: I redistributed this language across existing sections to hopefully make it better align with existing language.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
p-mongo: I find it helpful, and the rationale as I understand it helps explain why the spec prescribes certain behavior or why a test exists which is what I believe is happening here. I do agree that the behavior is verified by the unit test but this seems orthogonal to the purpose of a rationale.

mbroadst: I don't think we should set precedent for adding a rationale section for every scenario we can conjure up. For example, we don't have a rationale section for why we remove servers that are `me` mismatched. 

If you think the prose is helpful to better understanding the test then I encourage you to add it inline in the YAML file as comments, helping the test reader to better understand why the test exists.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst
```diff
@@ -104,7 +104,9 @@ Guidance
 Publishing and Subscribing
 --------------------------
 
-The driver SHOULD publish events in a manner that is standard to the driver's language publish/subscribe patterns and is not strictly mandated in this specification.
+The driver SHOULD publish events in a manner that is standard to the driver's
```
mbroadst: There appear to be a few white space changes in this file. Can you please revert them so that we can focus on what's actually changing in the spec?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1558,8 +1555,8 @@ the client MUST clear its connection pool to the server.
 error messages to detect "not master" and "node is recovering"`_, and `other
 transient errors`_ and `Why close connections when a node is shutting down?`_.)
 
-Monitoring SDAM events
-''''''''''''''''''''''
+SDAM events
```
mbroadst: What motivated this change?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -850,33 +850,30 @@ are parsed from the ismaster response in the obvious way.
...
-to the previous description.
-
 Updating the TopologyDescription
 ''''''''''''''''''''''''''''''''
 
 Each time the client checks a server,
 it processes the outcome (successful or not)
 to create a `ServerDescription`_,
 and then it processes the ServerDescription to update its `TopologyDescription`_.
+The ServerDescription MUST be processed even if it is equal (as defined in
```
mbroadst: nit: I slightly prefer directly linking to sections, rather than referring to them in the paranthetical. I think it makes it much easier to click through (or hover over) while reading:
```
The ServerDescription MUST be processed even if it is considered [equal](/link/to/equality/description)...
```

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -850,33 +850,30 @@ are parsed from the ismaster response in the obvious way.
...
+`Server Description Equality`_) to the previous description.
 
-The TopologyDescription's `TopologyType`_ influences
-how the ServerDescription is processed.
-The following subsection
-specifies how the client updates its TopologyDescription
-when the TopologyType is Single.
-The next subsection treats the other types.
+The TopologyDescription's `TopologyType`_ influences how the ServerDescription
```
mbroadst: I believe this is a white space change as well

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -850,33 +850,30 @@ are parsed from the ismaster response in the obvious way.
...
 ```````````````````````````
 
-For the purposes of updating topology description and publishing SDAM events,
-two server descriptions having the same address
-MUST be considered equal if and only if the values of
-`ServerDescription`_ fields marked (=) are respectively equal.
+For the purpose of determining whether to publish SDAM events, two server
+descriptions having the same address MUST be considered equal if and only if
+the values of `ServerDescription`_ fields marked (=) are respectively equal.
+Two topology descriptions MUST be considered equal if and only if they have
```
mbroadst: I think this last sentence is redundant. I propose that we leave the previous language as-is and modify the equality description to better define what the comparison should be for the "set of server addresses" (I think you mean `hosts', 'passives', and 'arbiters' when you say "set of server addresses", if that's incorrect then we should fix that as well)

source/server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst
```diff
@@ -114,34 +116,60 @@ Event Order and Concurrency
...
 
-ServerDescriptions MUST be initialized with a default description in an “unknown” state, guaranteeing that the previous description in the events will never be null.
+ServerDescriptions MUST be initialized with a default description in an
+"unknown" state, guaranteeing that the previous description in the events will
+never be null.
+
+Meaningful Changes
+------------------
+
+If the driver processes a ServerDescription which is equal (as server
```
mbroadst: I find this paragraph incredibly difficult to parse, and I'm not sure what this prose buys us beyond the existing definition of `ServerDescription` equality in the top-level SDAM spec. Is any new information introduced here?

p-mongo:
Updated the PR.

p-mongo:
This PR had an old version of the test (the one @jyemin  proposed initially). It now has the most recent test described in the ticket.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst
```diff
@@ -13,8 +13,8 @@ SDAM Monitoring Specification
 :Advisory Group: Jeff Yemin, Craig Wilson, Jesse Davis
 :Status: Approved
 :Type: Standards
-:Minimum Server Version: 2.5
-:Last Modified: 29-Jan-2020
+:Minimum Server Version: 2.4
+:Last Modified: 12-Dec-2018
```
jmikola: This date doesn't look right.

Also, consider bumping the spec version to 1.2.0 since you're introducing new SHOULD/MUST lines.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst
```diff
@@ -434,4 +406,3 @@ Changelog
 - 12 DEC 2018: Clarified table of rules for readable/writable servers
 - 31 AUG 2016: Added table of rules for determining if topology has readable/writable servers.
 - 11 OCT 2016: TopologyDescription objects MAY have additional methods and properties.
-- 29 JAN 2020: Require that only meaningful changes produce events.
```
jmikola: I think there should still be a changelog entry.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -8,8 +8,8 @@ Server Discovery And Monitoring
 :Advisors: David Golden, Craig Wilson
 :Status: Accepted
 :Type: Standards
-:Version: 2.15
-:Last Modified: 2020-01-29
+:Version: 2.14
+:Last Modified: 2019-07-11
```
jmikola: Another date issue. I think the version should go up to 2.16, too.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -867,13 +867,13 @@ Each time the client checks a server,
...
-The TopologyDescription's `TopologyType`_ influences how the ServerDescription
-is processed. The following subsection specifies how the client updates its
-TopologyDescription when the TopologyType is Single. The next subsection
-treats the other types.
+The TopologyDescription's `TopologyType`_ influences
+how the ServerDescription is processed.
+The following subsection
+specifies how the client updates its TopologyDescription
+when the TopologyType is Single.
+The next subsection treats the other types.
```
jmikola: This word wrap looks overly aggressive.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -850,19 +850,16 @@ are parsed from the ismaster response in the obvious way.
...
-For the purposes of updating topology description and publishing SDAM events,
-two server descriptions having the same address
-MUST be considered equal if and only if the values of
-`ServerDescription`_ fields marked (=) are respectively equal.
+For the purpose of determining whether to publish SDAM events, two server
+descriptions having the same address MUST be considered equal if and only if
+the values of `ServerDescription`_ fields marked (=) are respectively equal.
+Two topology descriptions MUST be considered equal if and only if they have
+the same set of server addresses, and the server descriptions for each address
+are respectively equal as just defined.
```
jmikola: [SDAM Monitoring: Events](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst#events) states the following rules for dispatching events:

> ServerDescriptionChangedEvent: When the old server description is not equal to the new server description
> TopologyDescriptionChangedEvent: When the old topology description is not equal to the new topology description.

Given that, perhaps this text is sufficient to clarify that events would not be dispatched.

Beyond that, do we need to remind implementations that SDAM flow must be run even if server descriptions are considered "equal" under the above rules? I believe that's main point of SPEC-1517, and simply removing "purposes of updating topology description and..." from the first sentence may be too subtle a change to communicate that.

[Parsing an ismaster response](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#parsing-an-ismaster-response) already states:

> This replacement MUST happen even if the new server description compares equal to the previous one

I think that's clear for the server description case; however, [Updating the TopologyDescription](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#updating-the-topologydescription) does not have similar language requiring some action even if server/topology descriptions are equal. I think the new sentence talking about topology description equality would be relevant, since that previously wasn't defined/addressed.

@mbroadst: do you concur?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -850,19 +850,16 @@ are parsed from the ismaster response in the obvious way.
...
 
 This specification does not prescribe how to compare server descriptions
 with different addresses for equality.
 
-Note: Server description for each server MUST be updated (replaced)
```
jyemin: Do we also have to remove this sentence below:

> If the new server description is equal to the previous server description as defined in Server Description Equality, stop the processing.  

It seems like it's saying the same thing as this removed paragraph is.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
jyemin: Also, a lot of places in the spec refer to the definition of server description equality.  Let' not add a sentence to this section that is not relevant to those references.  If we need the sentence, let's add a new heading `Topology Description Equality`.  But let's only do that if there is some other section that needs to reference that definition.  Looking at the companion document titled `SDAM Monitoring Specification`, I don't see anything about when one would publish a TopologyDescriptionChangedEvent.  Should it be added?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
mbroadst: Yes, and if I'm reading this correctly then I think that's what @jyemin is alluding to above. Perhaps we should define `TopologyDescription` equality, and add clarifying language to `Updating the TopologyDescription`

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
p-mongo: I removed the language relating to topology description equality.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
p-mongo: Removed this also.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
p-mongo: I removed language pertaining to topology equality from this PR.

In the [sdam spec](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#topologydescription), topology description is not defined to contain any servers. This language matches Ruby implementation in that the topology object only contains topology-related attributes as they are described in the referenced section, and does not contain server descriptions.

[sdam events spec](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst#events) then prescribes that topology change event is published when:

> When the old topology description is not equal to the new topology description.

Based on this, my reading is when the set of servers in a topology changes, but the topology type remains the same, a topology description change event must NOT be published. However, I believe this reading contradicts the spec tests such as [this one](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/tests/monitoring/replica_set_with_primary.yml#L27) that appears to require a topology change event to be published when no topology attribute changed.

If there is agreement that the current definition of topology description and existing monitoring spec tests are contradictory, I propose opening a new spec ticket to resolve the contradiction. SPEC-1517 refers to always processing ismaster responses, which has much smaller scope than resolving discrepancy between topology description definition and existing spec tests (as well as, presumably, driver implementations).

jmikola: If I follow correctly, https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/tests/monitoring/replica_set_with_primary.yml#L27 is a problem because the test expects a `topology_description_changed_event` event when `topologyType` remains "Unknown" and only the list of servers has changed. I agree that seems odd and also agree that it seems independent of this PR and would warrant its own spec ticket.

p-mongo: I created https://jira.mongodb.org/browse/SPEC-1567.

p-mongo:
@mbroadst @jyemin do you have any comments on the current language in this PR?

jyemin:
It looks good to me but I'm going to wait to approve until all the open comment threads are resolved.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
p-mongo: This change is no longer part of this PR.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
p-mongo: This change is no longer part of this PR.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst
p-mongo: Done.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
p-mongo: This change is no longer part of this PR.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
p-mongo: This change is no longer part of this PR.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
p-mongo: This change is no longer part of this PR.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst
p-mongo: This change is no longer part of this PR.

p-mongo:
@jyemin PTAL

PR #723, opened 2020-01-31
RetryableWrites: findOneAndModify --> findAndModify

egiurleo:
I mistyped the command on which to execute the failPoint. It should read "findAndModify" rather than "findOneAndModify."


jmikola:
@egiurleo: I assume we don't need a DRIVERS ticket to sync this change since the test would have failed with the typo, correct?

egiurleo:
@jmikola Correct!

PR #724, opened 2020-01-31
OCSP spec + test plan changes

divjotarora:
A collection of spec/test fixes that I made while implementing the spec. Feel free to suggest changes I may have missed.

### DIFF
--- a/source/ocsp-support/ocsp-support.rst
+++ b/source/ocsp-support/ocsp-support.rst
@@ -3,7 +3,7 @@ OCSP Support
 ============
 
 :Spec Title: OCSP Support
-:Spec Version: 1.1.0
+:Spec Version: 1.1.1
 :Author: Vincent Kam
 :Lead: Jeremy Mikola
 :Advisory Group: Clyde Bazile *(POC author)*, Esha Bhargava *(Program Manager)*, Matt Broadstone, Bernie Hackett *(POC author)*, Shreyas Kaylan *(Server Project Lead)*, Jeremy Mikola *(Spec Lead)*
@@ -78,8 +78,9 @@ invalid, the driver SHOULD end the connection.
     stapled OCSP response, a driver SHOULD end the connection.
 
 2.  If a driver’s TLS library supports Stapled OCSP and the server
-    staples an OCSP response that does not cover the certificate it
-    presents, a driver SHOULD end the connection.
+    staples an OCSP response that does not cover the certificate it presents or
+    is invalid per `RFC 6960 Section 3.2 <https://tools.ietf.org/html/rfc6960#section-3.2>`_,
+    a driver SHOULD end the connection.
 
 3.  If a driver’s TLS library supports Stapled OCSP and the server
     staples an OCSP response that does cover the certificate it
@@ -102,9 +103,13 @@ invalid, the driver SHOULD end the connection.
     attempt to validate the status of the unvalidated certificates
     using the cached CRLs.
 
-7.  If the server’s certificate remains unvalidated and that certificate
-    has an OCSP endpoint, the driver SHOULD reach out to the OCSP
-    endpoint specified and attempt to validate that certificate.
+7.  If the server's certificate remains unvalidated and that certificate
+    has a list of OCSP responder endpoints, the driver SHOULD send HTTP
+    requests to the responders in parallel. The first valid response
+    that concretely marks the certificate status as good or revoked
+    should be used. A five-second timeout SHOULD be used for the requests.
+    The status for a response should only be checked if the response is
+    valid per `RFC 6960 Section 3.2 <https://tools.ietf.org/html/rfc6960#section-3.2>`_
 
 8.  If any unvalidated intermediate certificates remain and those
     certificates have OCSP endpoints, for each certificate, the
@@ -117,8 +122,8 @@ invalid, the driver SHOULD end the connection.
     the other certificates using those CRLs.\*
 
 10. Finally, the driver SHOULD continue the connection, even if the
-    status of all the unvalidated intermediate certificates has not
-    been confirmed yet. This means that the driver SHOULD default to
+    status of all the unvalidated certificates has not been
+    confirmed yet. This means that the driver SHOULD default to
     “soft fail” behavior, connecting as long as there are no
     explicitly invalid certificates—i.e. the driver will connect
     even if the status of all the unvalidated certificates has not
@@ -643,13 +648,14 @@ of checking this are:
 
 Changelog
 ==========
+**2020-2-19**: Clarify behavior for reaching out to OCSP responders.
+
 **2020-2-10**: 1.1.0: Add cache requirement.
 
 **2020-1-31**: 1.0.2: Add SNI requirement and clarify design rationale
 regarding minimizing round trips.
 
-**2020-1-28**: 1.0.1: Clarify behavior regarding nonces and tolerance
-periods.
+**2020-1-28**: Clarify behavior regarding nonces and tolerance periods.
 
 **2020-1-16**: 1.0.0: Initial commit.
 
--- a/source/ocsp-support/tests/README.rst
+++ b/source/ocsp-support/tests/README.rst
@@ -49,7 +49,7 @@ malicious server cases.
 
 +----------------------------------------+-----------------------------------------+-------------------------------------------+-------------------------------------------------+---------------------------------------------------+-----------------------------------------------------+-----------------------------------------------------------------------+--------------------------------------------------------------------+
 | **URI options**                        | **Test 1\:**                            | **Test 2\:**                              | **Test 3\:**                                    | **Test 4\:**                                      | **Soft Fail Test\:**                                | **Malicious Server Test 1\:**                                         | **Malicious Server Test 2\: No OCSP Responder + server w/ Must-**  |
-|                                        | **Valid cert + server that staples**    | **Invalid cert + server that staples**    | **Valid cert + server that does not staple**    | **Invalid cert + server that does not staple**    |                                                     | **Invalid cert + server w/ Must- Staple cert that does not staple**   | **Staple cert that does not staple**                               |
+|                                        | **Valid cert + server that staples**    | **Invalid cert + server that staples**    | **Valid cert + server that does not staple**    | **Invalid cert + server that does not staple**    | **No OCSP Responder + server that does not staple** | **Invalid cert + server w/ Must- Staple cert that does not staple**   | **Staple cert that does not staple**                               |
 +========================================+=========================================+===========================================+=================================================+===================================================+=====================================================+=======================================================================+====================================================================+
 | ``tls=true``                           | OK                                      | FAIL                                      | OK                                              | FAIL                                              | OK\*                                                | FAIL\*                                                                | FAIL\*                                                             |
 +----------------------------------------+-----------------------------------------+-------------------------------------------+-------------------------------------------------+---------------------------------------------------+-----------------------------------------------------+-----------------------------------------------------------------------+--------------------------------------------------------------------+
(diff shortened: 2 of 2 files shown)


source/ocsp-support/ocsp-support.rst
```diff
@@ -297,10 +298,9 @@ Suggested OCSP Behavior
 For drivers with finer-grain control over their OCSP behavior, the
 suggested OCSP behavior was chosen as a balance between security and
 availability, erring on availability while minimizing network round
-trips. Therefore, in the `Suggested Stapled OCSP
-Behavior <#suggested-ocsp-behavior>`__ and `Suggested Non-Stapled OCSP
-Behavior <#_win0ckq3thfv>`__ sections, drivers are advised not to reach
-out to OCSP endpoints and CRL distribution points to minimize network
+trips. Therefore, in the `Suggested OCSP Behavior <#suggested-ocsp-behavior>`__
+section, drivers are advised not to reach out to OCSP endpoints and CRL
```
behackett: I think we never want a driver to explicitly reach out to a CRL distribution point (though some TLS libraries may do that automagically). This is an OCSP spec, not a generalized rev checking spec.

divjotarora: Are you proposing a larger change to the `Suggested OCSP Behavior` checking to remove any mention of CRL distribution points?

behackett: No. This wording just makes it sounds like drivers must talk to CRL distribution points as part of point 7. That's not the intent. I think you need to treat these things separately. Drivers should not reach out to CRL distribution points at all. Drivers should not reach out to OCSP endpoints for intermediates.

divjotarora: Changed this to not mention CRL distribution points.

source/ocsp-support/ocsp-support.rst
```diff
@@ -100,9 +100,10 @@ invalid, the driver SHOULD end the connection.
     attempt to validate the status of the unvalidated certificates
     using the cached CRLs.
 
-7.  If the server’s certificate remains unvalidated and that certificate
-    has an OCSP endpoint, the driver SHOULD reach out to the OCSP
-    endpoint specified and attempt to validate that certificate.
+7.  If the server's certificate remains unvalidated and that certificate
+    has a list of OCSP responder endpoints, the driver SHOULD reach out to the
+    responders in a loop until a responder returns a response that concretely
+    marks the certificate status as good or revoked.
```
vincentkam: Hm, should we also clarify what happens when the OCSP responder is unavailable? Soft-fail and connect after a specified timeout period seems reasonable: what does Go/Python currently do in this particular case?

behackett: Right now I've got a 5 second timeout on the request to the OCSP URI. If the timeout expires I soft fail.

divjotarora: For Go, I'm using the user-supplied context, but that need not have a timeout associated with it (the user could pass in something like `context.TODO()` which never expires or gets cancelled). I think it's a reasonable idea for drivers to enforce a default timeout. The Go code would be similar to our server selection code, as we would check both the user-supplied context and our own timeout for expiration/cancellation.

divjotarora: Do we agree that 5 seconds is reasonable?

vincentkam: 5s sounds fine to me! I've also pinged Shreyas and if the shell is doing something drastically different we can circle back to this later.

divjotarora: Added a requirement to set a 5s timeout on OCSP requests in my latest commit.

source/ocsp-support/ocsp-support.rst
```diff
@@ -100,9 +100,12 @@ invalid, the driver SHOULD end the connection.
...
     using the cached CRLs.
 
-7.  If the server’s certificate remains unvalidated and that certificate
-    has an OCSP endpoint, the driver SHOULD reach out to the OCSP
-    endpoint specified and attempt to validate that certificate.
+7.  If the server's certificate remains unvalidated and that certificate
+    has a list of OCSP responder endpoints, the driver SHOULD send HTTP requests
+    to the responders in parallel. The first valid response that concretely marks
+    the certificate status as good or revoked should be used. A five-second timeout
+    MUST be used for the requests. The status for a response should only be checked
```
vincentkam: Perhaps "SHOULD" since we're using SHOULD like "MUST" in this section due to the inability to mandate MUST for drivers like C#/.NET (and possibly others)?

source/ocsp-support/ocsp-support.rst
```diff
@@ -100,9 +100,12 @@ invalid, the driver SHOULD end the connection.
...
+    if there are no errors parsing it and it is correctly signed.
```
vincentkam: As a side note, I think we should remove the word "intermediate" from point 10, line 121, since I believe even if the leaf cert's status is unknown, we'll still connect (i.e. soft fail)

source/ocsp-support/ocsp-support.rst
```diff
@@ -297,11 +300,9 @@ Suggested OCSP Behavior
 For drivers with finer-grain control over their OCSP behavior, the
 suggested OCSP behavior was chosen as a balance between security and
 availability, erring on availability while minimizing network round
-trips. Therefore, in the `Suggested Stapled OCSP
-Behavior <#suggested-ocsp-behavior>`__ and `Suggested Non-Stapled OCSP
-Behavior <#_win0ckq3thfv>`__ sections, drivers are advised not to reach
-out to OCSP endpoints and CRL distribution points to minimize network
-round trips.
+trips. Therefore, in the `Suggested OCSP Behavior <#suggested-ocsp-behavior>`__
```
vincentkam: Note: I think when rebasing you'll probably want to take the version in master as I've fixed the broken links using `:refs:` since we have two sections with the same name (albeit at different levels)

source/ocsp-support/tests/README.rst
```diff
@@ -67,8 +67,9 @@ to determine which versions of the server can be used for each column.
 Note: From the perspective of a driver that does not support OCSP
 stapling, the following sets of tests should be identical: {Test 1, Test
 3}, {Test 2, Test 4, Malicious Server Test 1}, and {Soft Fail Test,
-Malicious Server Test 2}. However, it does no harm to test these extra
-cases and may help reveal unexpected behavior.
+Malicious Server Test 2}. For drivers with full customization, both malicious
```
vincentkam: Nit: full customization => "full control over their OCSP behavior?"

source/ocsp-support/tests/README.rst
```diff
@@ -91,8 +92,9 @@ certificate is invalid.
 The mongo-orchestration configurations needed for testing can be found
 at ``.evergreen/orchestration/configs/servers/``. Tests that specify that a
 server should staple MUST use ``basic-tls-ocsp-mustStaple.json``. Tests that
-specify that a server should not staple MUST use ``basic-tls-ocsp.json``.
-The malicious server tests MUST use ``basic-tls-ocsp-mustStaple.json``.
+specify that a server should not staple MUST use
+``basic-tls-ocsp-disableStapling.json``. The malicious server tests MUST use
+``basic-tls-ocsp-mustStaple-disableStapling.json``.
```
vincentkam: Thanks for creating these 😸

source/ocsp-support/ocsp-support.rst
divjotarora: Agreed. I had it as SHOULD to match the previous version, but I've changed it to MUST.

source/ocsp-support/ocsp-support.rst
divjotarora: Done.

source/ocsp-support/ocsp-support.rst
divjotarora: Will do!

source/ocsp-support/tests/README.rst
divjotarora: Done.

source/ocsp-support/tests/README.rst
divjotarora: No problem!

source/ocsp-support/ocsp-support.rst
vincentkam: Oops, sorry I was unclear! I was advocating for the use of SHOULD over MUST for this part of the spec since we are unable to use the word "MUST" due to drivers that don't have fine grained control over their OCSP behavior. 

In other words:
"...has a list of OCSP responder endpoints, the driver SHOULD send HTTP requests...A five-second timeout SHOULD be used for the requests"

source/ocsp-support/ocsp-support.rst
vincentkam: Thanks!

source/ocsp-support/ocsp-support.rst
divjotarora: Ah, I misunderstood your previous comment. My reading was "if you're manually sending a request, it must use a 5-second timeout" but I can see why SHOULD is preferred here for drivers whose TLS libraries may not allow timeouts to be configured for OCSP requests. Changed to SHOULD in my latest commit.

source/ocsp-support/ocsp-support.rst
```diff
@@ -100,9 +100,12 @@ invalid, the driver SHOULD end the connection.
...
 
-7.  If the server’s certificate remains unvalidated and that certificate
-    has an OCSP endpoint, the driver SHOULD reach out to the OCSP
-    endpoint specified and attempt to validate that certificate.
+7.  If the server's certificate remains unvalidated and that certificate
+    has a list of OCSP responder endpoints, the driver SHOULD send HTTP requests
+    to the responders in parallel. The first valid response that concretely marks
+    the certificate status as good or revoked should be used. A five-second timeout
+    SHOULD be used for the requests. The status for a response should only be checked
+    if there are no errors parsing it and it is correctly signed.
```
behackett: I think what you're saying here is any response that fails https://tools.ietf.org/html/rfc6960#section-3.2 will be ignored? I've been treating failures in section 3.2 the same as REVOKED. This seems to be implying those failures should be treated as UNKNOWN? If so, that feels like a fundamental change to this spec, not just this verification step.

source/ocsp-support/ocsp-support.rst
```diff
@@ -76,8 +76,8 @@ invalid, the driver SHOULD end the connection.
     stapled OCSP response, a driver SHOULD end the connection.
 
 2.  If a driver’s TLS library supports Stapled OCSP and the server
-    staples an OCSP response that does not cover the certificate it
-    presents, a driver SHOULD end the connection.
+    staples an OCSP response that is invalid per `RFC 6960 Section 3.2 <https://tools.ietf.org/html/rfc6960#section-3.2>`_,
```
vincentkam: Should we keep the language regarding the response not covering the certificate it presents? The RFC notes "The certificate identified in a received response corresponds to the certificate that was identified in the corresponding request;". In the case of a stapled response, there is no explicit request, although I suppose you could argue that implicitly we are requesting the status of the server's cert. I feel like being a bit more explicit here probably wouldn't do any harm though?

divjotarora: I originally removed it because of the line you cited from the RFC but I can see how that might cause confusion. I agree that being explicit can't hurt. The new commit reverts that deletion.

source/ocsp-support/ocsp-support.rst
divjotarora: Resolving as we discussed this offline. The new commits explicitly state the verification constraints that we discussed during our meeting.

PR #726, opened 2020-02-02
SPEC-1562: retryable-writes test fixes

mbroadst:
We missed a few issues with the tests added for the error label work in retryable-writes. Each fix is isolated to a single commit in this PR, so you might want to just look at the commit message for each while reviewing. 

@egiurleo it concerns me that the tests as they are written today are passing in the Ruby driver, could it be that the driver is using the legacy retry behavior in your test runner or something similar?


mbroadst:
@egiurleo ah, that makes a lot more sense then! I would say generally when we make these types of more involved spec changes, there is an expectation that at least the author's driver is implementing the changes alongside the test/prose work. I would also expect in this case that the POCs do that work concurrently, but things get a little blurry when you're working on a spec amendment rather than a brand new spec. I think this a problem in process documentation, I'll make sure to bring it up at leads to discuss how to mitigate this in the future.

Great job on the changes, they definitely caught some bugs in Node's implementation! :v:

egiurleo:
@mbroadst Thanks for the clarification. I'll make sure to do that in the future.

source/retryable-writes/tests/insertOne-serverErrors.yml
```diff
@@ -481,4 +481,6 @@ tests:
             result:
                 errorLabelsContain: ["RetryableWriteError"]
             collection:
-                data: []
+                data:
+                    - { _id: 1, x: 11 }
```
jmikola: While you're changing this, can we use `{ _id: 3, x: 33 }` for the `insertOne` argument? That's more consistent with the other tests in this file.

Looking at the full test, I realized that it might fail with a duplicate key exception if `closeConnection` on the fail point wasn't hit for some reason. I'm being paranoid, but that could get weird if a driver's spec runner happened to incorrectly pass because it saw `outcome.error: true` and ignored the `outcome.result.errorLabelsContain` check.

mbroadst:
@jmikola @egiurleo I've pushed a new commit to this that explicitly adds the `RetryableWriteError` error label to relevant tests which are currently broken against mongos 4.3.x. I also created a spec ticket assigned to @egiurleo to revert these changes once [SERVER-45939](https://jira.mongodb.org/browse/SERVER-45939) is complete.

@iwysiu can you please sync your spec tests for the Go driver and verify that these changes fix your test runs?

source/retryable-writes/tests/insertOne-serverErrors.yml
mbroadst: Made that change. Regarding your second point I think we should assume that driver's have implemented their test runners appropriately. A duplicate key exception would also be enough of a hint I think that things were amiss, and the test runner should be evaluated.

ShaneHarvey:
I'll defer this review to @iwysiu, @jmikola, and @egiurleo.

PR #727, opened 2020-02-05
SPEC-1509: Specify hedge parameter for read preference

alcaeus:


source/server-selection/server-selection.rst
```diff
@@ -521,6 +526,15 @@ Eligibility MUST be determined from ``tag_sets`` as follows:
 - If the ``tag_sets`` list is not empty and no tag set in the list matches any
   candidate server, no servers are eligible servers.
 
+hedge
+`````
+
+The read preference ``hedge`` parameter is an object that configures how the
```
jmikola: Let's say "document" instead of "object" here, since that's more clear when discussing BSON.

source/server-selection/server-selection.rst
```diff
@@ -521,6 +526,15 @@ Eligibility MUST be determined from ``tag_sets`` as follows:
...
+server will perform hedged reads. It consists of the following keys:
+
+- ``enabled`` (default true): This enables hedging
```
jmikola: I think the earlier terminology about `hedge` defaulting to "empty" might lead one to believe that it's enabled by default if this is also defaulting to true.

These are the cases I'm wondering about:

 * Read preference omits `hedge` option. I think it's obvious that hedged reads won't be used at all.
 * Read preference specifies empty document for `hedge` option. This section suggests that `enabled` and `delay` will both default to true. This definitely seems like an edge case, but something to clarify in the spec.
 * Read preference specifies `hedge` document with at least one option (`enabled` OR `delay` in the current version). In that case, the other option would default accordingly.

Regarding defaults, I think we should also specify whether drivers should be filling in these defaults if ever a document is provided for `hedge` or if this is just reporting what the current server-side defaults are. I think it's more intuitive for drivers not to fill in defaults on their own. That agrees with existing conventions in the [CRUD spec](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst) where we say "the default is to not send a value".

source/server-selection/server-selection.rst
```diff
@@ -370,14 +374,15 @@ Components of a read preference
...
+``tag_sets``, ``maxStalenessSeconds``, and ``hedge``.  The ``mode`` prioritizes
+between primaries and secondaries to produce either a single suitable server or
+a list of candidate servers.  If ``tag_sets`` and ``maxStalenessSeconds`` are
+set, they determine which candidate servers are eligible for selection. If
+``hedge`` is set, it configures how server hedged reads are used.
 
 The default ``mode`` is 'primary'.  The default ``tag_sets``
 is a list with an empty tag set: ``[{}]``. The default ``maxStalenessSeconds``
-is -1 or null, depending on the language.
+is -1 or null, depending on the language. The default ``hedge`` is empty.
```
jmikola: Relates to my comment below, but should we differentiate between "empty" and "unset"? If the user provides an empty BSON document for `hedge`, should we send that to the server?

It may be more intuitive for users if we treat unset and empty the same; however, that would leave no way for an application to indicate that it wants to use `hedge` and fully rely on the server-side defaults for its options (`enabled` and `delay`). Perhaps this is something to ask one of the server engineers or product managers to determine what the desired user-facing behavior is.

source/server-selection/server-selection.rst
```diff
@@ -521,6 +526,15 @@ Eligibility MUST be determined from ``tag_sets`` as follows:
 - If the ``tag_sets`` list is not empty and no tag set in the list matches any
   candidate server, no servers are eligible servers.
 
+hedge
+`````
+
+The read preference ``hedge`` parameter is an object that configures how the
+server will perform hedged reads. It consists of the following keys:
```
jmikola: If we anticipate that the server might introduce additional options (i.e. keys) for `hedge` down the line, do we want to add some language similar to [Unknown Levels and Additional Options for String Based ReadConcerns](https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#unknown-levels-and-additional-options-for-string-based-readconcerns) in the RC/WC spec? I imagine some drivers will naturally want to model the `hedge` option instead of accepting an arbitrary BSON document.

I'll admit the cases aren't entirely analogous because the RC spec is merely prohibiting drivers from accepting new _values_ for a known option/key, while the argument here is whether we should allow drivers to pass new _keys_ to the server. I could be convinced that it's not a problem to allow strongly-typed drivers to model the `hedge` option with known keys and we could leave it unspecified if some drivers decide to just accept an untyped hash/document and ship that to the server. If we permit flexibility to ship arbitrary documents to the server, I would just confirm with the server engineer that we can rely on `mongod` and `mongos` to raise errors for unknown options -- as we'll want to be sure to avoid any case where we let the user specify an option (say, for 4.6+) and have it be silently ignored by a 4.4 server.

source/server-selection/server-selection.rst
alcaeus: Yes, the wording needs changing. I think we need to clarify omitting the `hedge` option entirely and submitting an empty document, in which case hedging is enabled with default values. Consider this paragraph:
> To explicitly enable hedging, an empty document must be sent to the server for the `hedge` option. This enables hedging with the default values specified above. To not enable hedging, the `hedge` option must be omitted entirely.

Regarding default values I would add the following sentence:
> Drivers MUST NOT fill in default values for any keys in the ``hedge`` document if they are omitted.

source/server-selection/server-selection.rst
alcaeus: I changed "empty" to "null" in the sentence above, but we may also choose a less language-specific "unset".

Submitting an empty `hedge` document would be the preferred way to turn on hedging with server defaults, although we may treat this as unset in favour of using `{ hedge: { enabled: true } }` to be more explicit.

source/server-selection/server-selection.rst
alcaeus: Tested manually to see the behaviour.

A 4.2 server completely ignores the `hedge` option and does not error when it is encountered:
```
MongoDB Enterprise mongos> db.runCommand({count: 'test', $readPreference: { mode: 'secondary', hedge: {asdf: true} }})
{
	"shards" : {
		
	},
	"n" : 0,
	"ok" : 1,
	"operationTime" : Timestamp(1581060684, 2),
	"$clusterTime" : {
		"clusterTime" : Timestamp(1581060684, 2),
		"signature" : {
			"keyId" : NumberLong(0)
		}
	}
}
```

So the server will not fail on unknown keys in the read preference document. When running against a 4.3 server, the server also doesn't throw an error when encountering unknown keys in the read preference document, but will fail when encountering an unknown key in the `hedge` document:
```
MongoDB Enterprise mongos> db.runCommand({count: 'test', $readPreference: { mode: 'secondary', hedge: {asdf: true} }})
{
	"ok" : 0,
	"errmsg" : "BSON field 'hedge.asdf' is an unknown field.",
	"code" : 40415,
	"codeName" : "Location40415",
	"operationTime" : Timestamp(1581060865, 1),
	"$clusterTime" : {
		"clusterTime" : Timestamp(1581060865, 1),
		"signature" : {
			"keyId" : NumberLong(0)
		}
	}
}
```

With this in mind, I would:
* Not require drivers to omit unknown fields in the read preference
* With the driver mantras in mind, we should not check the fields of the `hedge` document for now either. If the server adds additional hedging options in a future release, we will have to check wire versions then before submitting them. For now, relying on a server error seems the best course of action
* Given the impact of submitting a `hedge` option on a 4.2 server (meaning hedging was requested but the server doesn't hedge), I would also not raise an error at all, but I defer to your judgment in this case.

source/server-selection/server-selection.rst
jmikola: I would prefer "unset" over "null", as there's no chance it could be misinterpreted a present field with the BSON null value.

source/server-selection/server-selection.rst
jmikola: > So the server will not fail on unknown keys in the read preference document. When running against a 4.3 server, the server also doesn't throw an error when encountering unknown keys in the read preference document, but will fail when encountering an unknown key in the hedge document:

I'm glad we can rely on 4.4+ to raise errors for unsupported fields in the `hedge` document, as that does justify not having drivers enforce options (unless they need to model those options for strict typing, which they should be free to do).

The lack of validation for read preference fields (across both server versions) is unfortunate. I think we should err on the side of having the client (i.e. driver) raise an error for older wire versions. I believe this would be consistent with what we already do for [`maxStalenessSeconds`](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#maxstalenessseconds):

> During server selection, drivers (but not mongos) MUST raise an error if maxStalenessSeconds is a positive number, and any server's maxWireVersion is less than 5.

"(but not mongos)" suggests that the spec author was aware that mongos will just ignore the field if it is not supported, but a client-side error would pre-empty that from even happening.

Separately, you may want to open a SERVER ticket requesting mongos actually raise errors for unsupported read preference fields.

alcaeus: Added a section to require drivers to only send a `hedge` option if maxWireVersion >= 9. I've also opened [SERVER-46093](https://jira.mongodb.org/browse/SERVER-46093) to track checking for unknown read preference options.

source/server-selection/server-selection.rst
```diff
@@ -125,6 +125,10 @@ Terms
     Describes candidate servers that also meet the criteria specified by the
     ``tag_sets`` and ``maxStalenessSeconds`` read preference parameters.
 
+**Hedged Read**
+    A server mode in which the same query is dispatched in parallel to multiple
+    replica set members.
```
jmikola: I'll note that drivers will rely on a 4.4 mongos to perform dispatching, so this presently only applies to sharded clusters where shards are replica sets. I expect that will change down the line if drivers ultimately need to handle concurrent dispatch when connected to a replica set.

I don't think this text needs to be changed, but I wanted to point it out for the benefit of other reviewers.

source/server-selection/server-selection.rst
```diff
@@ -521,6 +527,24 @@ Eligibility MUST be determined from ``tag_sets`` as follows:
...
+`````
+
+The read preference ``hedge`` parameter is a document that configures how the
+server will perform hedged reads. It consists of the following keys:
+
+- ``enabled`` (default true): This enables hedging
+- ``delay`` (default true): This enables staggered reads
+
+To explicitly enable hedging, an empty document must be sent to the server for
+the `hedge` option. This enables hedging with the default values specified
```
jmikola: Double back-ticks for `hedge`.

source/server-selection/server-selection.rst
```diff
@@ -521,6 +527,24 @@ Eligibility MUST be determined from ``tag_sets`` as follows:
...
+above. To not enable hedging, the `hedge` option must be omitted entirely.
```
jmikola: Double back-ticks here, too.

Could we rephrase as "Hedging will be disabled if the `hedge` option is omitted"?

source/server-selection/server-selection.rst
```diff
@@ -521,6 +527,24 @@ Eligibility MUST be determined from ``tag_sets`` as follows:
...
+
+Drivers MUST NOT fill in default values for any keys in the ``hedge`` document
+if they are omitted. Drivers MUST raise an error if ``hedge`` is set, and the
+server's ``maxWireVersion`` is less than 9.
```
jmikola: As written, I'm not sure how this would play out when resolving a read preference as part of server selection. If the client is connected to a non-sharded topology, I suppose `hedge` would simply never get used and could be ignored. Unless you think server selection should raise a client-side error if it sees `hedge` and the topology is non-sharded -- in which case we could revisit that when/if drivers support `hedge` on their own.

Since we only send `readPreference` to mongos nodes, perhaps we should rephase this as:

> Drivers MUST NOT send a read preference where the `hedge` option is set to server version before 4.4 (i.e. `maxWireVersion` < 9) and MUST raise an error instead. This pertains to [passing a read preference to mongos](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#passing-read-preference-to-mongos).

source/server-selection/server-selection.rst
```diff
@@ -624,9 +652,11 @@ A valid ``$readPreference`` document for mongos has the following requirements:
     The ``maxStalenessSeconds`` field MUST be either be absent or be present
     exactly once with an integer value.
 
+    The ``hedge`` field MUST be either absent or be an object.
```
jmikola: "a document" instead of "an object".

source/server-selection/server-selection.rst
```diff
@@ -624,9 +652,11 @@ A valid ``$readPreference`` document for mongos has the following requirements:
...
+
 Mongos receiving a query with ``$readPreference`` SHOULD validate the
-``mode``, ``tags``, and ``maxStalenessSeconds`` fields according to rules 1 and 2 above,
-but SHOULD ignore unrecognized fields for
+``mode``, ``tags``, ``maxStalenessSeconds``, and ``hedge`` fields according to
+rules 1 and 2 above, but SHOULD ignore unrecognized fields for
 forward-compatibility rather than throwing an error.
```
jmikola: "SHOULD ignore unrecognized fields for forward-compatibility rather than throwing an error" makes me think that mongos' current behavior may be intentional. Perhaps this should be brought up in [SERVER-46093](https://jira.mongodb.org/browse/SERVER-46093) before someone from the server team follows up on it.

source/server-selection/server-selection.rst
alcaeus: I don't think drivers should go as far as checking for the topology in order to raise an error for an unsupported `hedge` option. Checking the wire version should suffice.

jmikola: I concur. Since the new text refers to sending for mongos, I'm good to resolve this.

source/server-selection/server-selection.rst
jmikola: I left a [comment in SERVER-46093](ttps://jira.mongodb.org/browse/SERVER-46093?focusedCommentId=2877497&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2877497).

source/server-selection/server-selection.rst
```diff
@@ -521,6 +527,25 @@ Eligibility MUST be determined from ``tag_sets`` as follows:
 - If the ``tag_sets`` list is not empty and no tag set in the list matches any
   candidate server, no servers are eligible servers.
 
+hedge
```
ShaneHarvey: All the other readPreference options are configurable through the connection string (`maxStalenessSeconds`, `readPreference`, `readPreferenceTags`). Should we add a new URI option for `hedge` maybe called `readPreferenceHedge`?

jmikola: Since there's already a conversation about this in [DRIVERS-723](https://jira.mongodb.org/browse/DRIVERS-723), let's continue it there rather than this comment thread.

source/server-selection/server-selection.rst
```diff
@@ -615,7 +643,8 @@ A valid ``$readPreference`` document for mongos has the following requirements:
     - 'secondaryPreferred'
     - 'nearest'
 
-2.  If the ``mode`` field is "primary", the ``tags`` and ``maxStalenessSeconds`` fields MUST be absent.
+2.  If the ``mode`` field is "primary", the ``tags``, ``maxStalenessSeconds``,
+    and ``hedge`` fields MUST be absent.
```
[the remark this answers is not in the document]
alcaeus: For now, I expanded the "Passing ReadPreference to Mongos" prose tests to cover the `hedge` option, as it would be covered by the `$readPreference` handling described there. Since we don't expect server selection to change based on the `hedge` option, I've avoided adding more tests there. Introducing a spec test runner for an option that is essentially passed through and not handled by the driver itself seems overkill.

alcaeus:
Spec updated to account for `delay` no longer being part of the `hedge` document. For now, I've held off on defining URI options for hedging, since it only affects mongos. We can revisit this at a later date.

source/server-selection/server-selection.rst
ShaneHarvey: I don't agree with the rationale for raising a client side error here because silently ignoring the `hedge` option doesn't break any guarantees about the query results. maxStalenessSeconds is different because ignoring maxStalenessSeconds would break the max staleness contract and possibly lead to stale data reads.

I think it's by design that servers ignore unrecognized readPreference options given the wording below:
```
Mongos receiving a query with ``$readPreference`` SHOULD validate the
``mode``, ``tags``, and ``maxStalenessSeconds`` fields according to rules 1 and 2 above,
but SHOULD ignore unrecognized fields for
forward-compatibility rather than throwing an error.
```

But either way is fine with me (client side error or not).

source/server-selection/server-selection.rst
```diff
@@ -521,6 +527,24 @@ Eligibility MUST be determined from ``tag_sets`` as follows:
...
 
+hedge
+`````
+
+The read preference ``hedge`` parameter is a document that configures how the
+server will perform hedged reads. It consists of the following keys:
+
+- ``enabled`` (default true): This enables hedging
+
+To explicitly enable hedging, an empty document must be sent to the server for
```
alcaeus: After discussing this with @jyemin, I've decided to change this and the following paragraph in commit b279fe4. When providing an API, allowing the user to build an empty hedge document makes no sense, as the user may have trouble understanding how the server behaves. This also removes API ambiguity where the `enabled` hedge option would effectively become a ternary bool (unset, `true`, and `false`).

If a driver allows the user to specify an arbitrary document for the `hedge` option, an empty document is still permitted.

source/server-selection/server-selection.rst
```diff
@@ -521,6 +527,30 @@ Eligibility MUST be determined from ``tag_sets`` as follows:
...
 
+hedge
+`````
+
+The read preference ``hedge`` parameter is a document that configures how the
+server will perform hedged reads. It consists of the following keys:
+
+- ``enabled``: Enables or disables hedging
+
+To enable hedging with server-side defaults, an empty ``hedge`` document must be
```
alcaeus: After discussing this with @jyemin, I've decided to change this and the following paragraphs in commit b279fe4. When providing an API, allowing the user to build an empty hedge document makes no sense, as the user may have trouble understanding how the server behaves. This also removes API ambiguity where the `enabled` hedge option would effectively become a ternary bool (unset, `true`, and `false`).

If a driver allows the user to specify an arbitrary document for the `hedge` option, an empty document is still permitted.

source/server-selection/server-selection.rst
```diff
@@ -521,6 +527,30 @@ Eligibility MUST be determined from ``tag_sets`` as follows:
...
+sent to the server. If no ``hedge`` document is passed or if the ``enabled``
+key is ``false``, the server will not perform hedged reads.
```
jyemin: My understanding from the server scope is that if no hedge document is included in the read preference, then it _will_ perform a hedged read.  Can you double-check this?

alcaeus: As per [SERVER-45440](https://jira.mongodb.org/browse/SERVER-45440), mongos will implicitly enable hedging when a read preference of "nearest" is specified, even when no hedge document is included. if the read preference is anything other than "nearest", no hedging will be performed. This is possible with [SERVER-45439](https://jira.mongodb.org/browse/SERVER-45439), which allows specifying a hedge document to explicitly enable hedging for other read preferences (except for "primary"). From my understanding, this also allows specifying a hedge document with `enabled: false` to disable hedging when using a "nearest" read preference.

Both tickets have already landed in 4.3.3. The `delay` option, which staggers hedged reads was removed from the current scope and will not be included in 4.4.

jyemin: In that case, this part of the "or" statement is inaccurate:

>  If no ``hedge`` document is passed or [...], the server will not perform hedged reads.

I don't think we have to specify the behavior when no hedge document is passed.  The server may change this from release to release.

jyemin: This also seems inaccurate:

> To enable hedging with server-side defaults, an empty ``hedge`` document must be
sent to the server.

source/server-selection/server-selection.rst
ShaneHarvey: Thinking about this again I propose that we do not raise a client-side error. Old servers will ignore the option and that's okay. The benefit of not raising a client-side error is that the server can backport hedged read support if needed and drivers will not need any changes. What do you think?

alcaeus: After considering, I've removed the last paragraph requiring drivers not to send a ``hedge`` document if the server does not support it and to throw a client-side error. If the server ticket linked above is resolved to throw errors on unknown read preference options, we can always revisit this.

source/server-selection/server-selection.rst
alcaeus: Reworded:

> Hedged reads are automatically enabled in MongoDB 4.4+ when using a ``nearest``
read preference. To explicitly enable hedging, the ``hedge`` document must be
passed. An empty document uses server defaults to control hedging, but the
``enabled`` key may be set to ``true`` or ``false`` to explicitly enable or
disable read hedging.

source/server-selection/server-selection-tests.rst
```diff
@@ -209,7 +209,7 @@ correctly passed to Mongos in the following scenarios:
 - SECONDARY_PREFERRED
 
   - the slaveOK wire protocol flag is set
-  - if tags are specified $readPreference is used, otherwise $readPreference is NOT used
+  - if tags or hedging are specified $readPreference is used, otherwise $readPreference is NOT used
```
mbroadst: nit: "If `tag_sets` or `hedge` are specified then `$readPreference` is used..."

alcaeus:
Merged in 4e570fc.

PR #728, opened 2020-02-05
SPEC-1552: support speculative authentication attempts in isMaster

jstewart-mongo:


source/auth/auth.rst
```diff
@@ -250,6 +250,48 @@ or "SCRAM-SHA-256").
...
+provided as a BSON object. Clients specifying this argument to ``isMaster`` will
+speculatively include the first ``saslStart`` command of an authentication handshake.
+This command may be provided to the server in parallel with any standard request for
+supported authentication mechanisms. This would permit clients to merge their
+``saslStart`` message with their ``isMaster`` request, and receive the ``saslStart``
+reply with the ``isMaster`` reply.
+
+If the speculatively issued command succeeds, the client should proceed with the next
+step of the exchange. If the command fails, the client should determine whether the
+``isMaster`` response contains enough information to select another mechanism.
```
jmikola: The next paragraph says older servers ignore `speculativeSaslStart`, so I assume the `isMaster` response won't include any bundled reply for auth. I wouldn't consider that a command failure, though, since `isMaster` itself will succeed.

Would "If the command fails" only apply on a 4.4+ server that processes `speculativeSaslStart` and reports a failure due to a typical authentication failure (as could happen today with the non-speculative conversation)? I think we should also clarify what happens if the speculative SASL conversation fails on 4.4+ for some reason and how that affects the wrapping `isMaster` response. Would the top-level `ok` field be `0`?

In the context of this paragraph, I think your second sentence is instructing drivers how to proceed on a pre-4.4 server. In that case, I'd suggest replacing "If the command fails" with "If the `isMaster` response does not include a `saslStart` reply..."

source/auth/auth.rst
```diff
@@ -250,6 +250,48 @@ or "SCRAM-SHA-256").
...
+Speculative Authentication via Handshake
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+:since: 4.4
+
+The ``isMaster`` handshake supports a new argument, ``speculativeSaslStart``,
+provided as a BSON object. Clients specifying this argument to ``isMaster`` will
+speculatively include the first ``saslStart`` command of an authentication handshake.
+This command may be provided to the server in parallel with any standard request for
+supported authentication mechanisms. This would permit clients to merge their
```
jmikola: Does "any standard request for supported authentication mechanisms" refer to anything else besides [mechanism negotiation via handshake](https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#mechanism-negotiation-via-handshake)? If not, I'd propose adding "(i.e. `saslSupportedMechs`)" or a reference to that section at the very end of this sentence for clarity.

source/auth/auth.rst
```diff
@@ -250,6 +250,48 @@ or "SCRAM-SHA-256").
...
+  
+
+Alternatively, the ``isMaster`` handshake supports another new argument,
+``speculativeAuthenticate``, provided as a BSON object. This argument provides an
+instance of the ``authenticate`` command with the ``MONGODB-X509`` mechanism.
+Servers accepting the request will respond with an ``isMaster`` reply containing the results
+in a field named ``speculativeAuthenticate``. Concurrent use of ``speculativeSaslStart``
+and ``speculativeAuthenticate`` will result in ``isMaster`` returning an error.  In this
+mode, authentication can occur in zero round trips and will be completed once ``isMaster``
+returns.
```
jmikola: The new paragraphs here don't actually instruct drivers to use these new `isMaster` options. Should we explicitly state that drivers MUST employ `speculativeAuthenticate` if the application has indicated it wishes to authenticate with `MONGODB-X509`? Likewise for requiring drivers to attempt `speculativeSaslStart` if the application indicates a SCRAM mechanism or no mechanism at all (in which case SCRAM-SHA-1 is used), per https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#defaults.

You may optionally want to note something to the effect of:

 *  drivers MUST NOT use `speculativeSaslStart` if an auth mechanism other than SCRAM-SHA-256 or SCRAM-SHA-1 has been indicated
 * drivers MUST NOT use `speculativeAuthenticate` if `MONGODB-X509` has not been indicated"

source/auth/auth.rst
```diff
@@ -250,6 +250,48 @@ or "SCRAM-SHA-256").
...
+
+An example conversation using ``speculativeAuthenticate`` is as follows:
+
+| C: :javascript:`{ isMaster: 1, speculativeAuthenticate: { mechanism: "MONGODB-X509", db: "$external" }, ... }`
+| S: :javascript:`{ ismaster: true, speculativeAuthenticate: { dbname: "$external", user : "CN=client,OU=KernelUser,O=MongoDB,L=New York City,ST=New York,C=US" }, ok: 1, ... }`
+
 --------------------------------
 Supported Authentication Methods
```
ShaneHarvey: (Commenting here because I can't comment below.) Can we add something to the "Test Plan" section about this feature?

source/mongodb-handshake/handshake.rst
```diff
@@ -224,6 +224,70 @@ Example::
         - "Oracle JVM EE 9.1.1"
 
 
+--------------------------
+Speculative Authentication
```
ShaneHarvey: Do we need to add a note to https://github.com/mongodb/specifications/blob/master/source/command-monitoring/command-monitoring.rst#security to ensure that drivers redact `speculativeAuthenticate` and `speculativeSaslStart`? Actually I believe that the initial isMaster handshake is not supposed to be published at all but I can't find that requirement anywhere.  Should we add a note to this spec to say that the initial handshake MUST NOT be published?

jmikola: Good point. [Command Monitoring](https://github.com/mongodb/specifications/blob/master/source/command-monitoring/command-monitoring.rst) does not say anything about not publishing `isMaster` commands and I don't see any mention of that in [SDAM](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst), either.

I agree that we can discuss security concerns here and would also propose that we update [Command Monitoring: Security](https://github.com/mongodb/specifications/blob/master/source/command-monitoring/command-monitoring.rst#security) to include `isMaster` with `speculativeAuthenticate` or `speculativeSaslStart` in its table of prohibited commands.

Additionally, I think we can make a separate SPEC ticket to update Command Monitoring and/or SDAM to say that `isMaster` commands shouldn't be reported with APM. I don't think any drivers actually do that, and it seems redundant given that SDAM has its own monitoring events.

jmikola:
> This seems like a more generic approach. (No need to hold up this review I'm just curious.)

Asked about this in [SERVER-44858](https://jira.mongodb.org/browse/SERVER-44858). Also asked about the schedule of that ticket, since the server implementation appears to have been in CR for nearly two months and the issue itself is currently scheduled in the backlog.

source/mongodb-handshake/handshake.rst
```diff
@@ -3,7 +3,7 @@ MongoDB Handshake
 =================
 
 :Spec Title: MongoDB Handshake
-:Spec Version: 1.1.0
+:Spec Version: 1.1.1
```
jyemin: As this is is a new feature, make it a minor version bump, not a patch bump.

source/auth/auth.rst
jyemin: >  the client should determine whether the `isMaster` response contains enough information to select another mechanism.

Can we make this specific?  Is this about `saslSupportedMechs` in the response?

source/auth/auth.rst
```diff
@@ -250,6 +250,48 @@ or "SCRAM-SHA-256").
...
 combination of the ``clientKey`` and ``serverKey`` parameters.
 
+Speculative Authentication via Handshake
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+:since: 4.4
+
+The ``isMaster`` handshake supports a new argument, ``speculativeSaslStart``,
+provided as a BSON object. Clients specifying this argument to ``isMaster`` will
+speculatively include the first ``saslStart`` command of an authentication handshake.
```
jyemin: If the application does not specify a mechanism (which is very likely), which mechanism should the driver default to in the speculative saslStart?  This seems like a problem because if we pick wrong the effort is wasted.

jstewart-mongo: I talked with @sgolemon about this. If a password is provided, the fallback mechanism should be SCRAM-SHA-256. I have added a statement addressing that in this section.

source/auth/auth.rst
jstewart-mongo: I have added your proposed text.

source/auth/auth.rst
jstewart-mongo: If speculative authentication fails, that will not affect `isMaster`. The response will contain `ok: 1` but not contain a `speculativeAuthenticate` reply. I have added wording here to describe this.
I have replaced "if the command fails" with your suggested wording.

jstewart-mongo: I have reworded this section based on the new `speculativeAuthenticate` argument.

source/auth/auth.rst
jstewart-mongo: I have added wording to this effect prior to the examples.

source/auth/auth.rst
jstewart-mongo: I have added a section to the Test Plan for speculative authentication.

source/mongodb-handshake/handshake.rst
jstewart-mongo: Done

source/mongodb-handshake/handshake.rst
jstewart-mongo: I have updated the Security section in Command Monitoring to include `isMaster` with `speculativeAuthenticate`.

jyemin: I feel like this is going to be painful to implement, since it's the only redacted command that requires looking at the content of the command rather than just the name.  I think we should go ahead and make @jmikola 's suggested change about not reporting ismaster commands at all.  Except that this is not only about SDAM.  It's about the handshake for all new connections.  So we could say that any command executed as part of the handshake MUST NOT be reported via command monitoring.

source/auth/auth.rst
```diff
@@ -250,6 +250,53 @@ or "SCRAM-SHA-256").
 The cache entry value MUST be either the ``saltedPassword`` parameter or the
 combination of the ``clientKey`` and ``serverKey`` parameters.
 
+Speculative Authentication via Handshake
```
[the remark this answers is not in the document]
jyemin: Note: we should not move the full conversation examples, as they are redundant to what's already in the auth spec.

source/mongodb-handshake/handshake.rst
jmikola: I don't think it's _technically_ difficult to implement, but I would agree that it might be odd to conditionally exclude commands from APM based on what they contain. If a user was expecting to see handshake or monitoring commands today, they might wonder why those suddenly disappear under certain circumstances.

I think excluding all handshake stuff from APM is a reasonable compromise, but that seems like an independent task that would be a dependency of this SPEC ticket. Tests for not logging handshakes with APM wouldn't necessarily involve `speculativeAuthenticate` at all, so I expect that would warrant its own set of tests (assuming we have tests for handshakes today -- I'm not sure).

Do we want to roll that change into this SPEC ticket, with tests, or offload to a separate ticket for visibility? If we consolidate the change into this ticket, we'll definitely want to revise the corresponding DRIVERS ticket to inform folks of this extra change once everything is merged.

jmikola:
@jstewart-mongo: I resolved my previous comment threads, but it looks like the question of whether to exclude all handshake commands from APM is outstanding so I'll hold off on an LGTM for now.

source/mongodb-handshake/handshake.rst
jmikola: @jyemin: In the interest of expediency, can we compromise by leaving the present text to conditionally exclude `isMaster` with `speculativeAuthenticate` from APM for now and then push "exclude all handshakes from APM" to a post-4.4 SPEC ticket?

jyemin: It is technically difficult in some drivers (Java, .NET) because of the way they do wire serialization.  That said, I think both of those drivers already exclude handshake-related commands from command monitoring, so this requirement becomes a no-op.

So I'm good with your suggestion.

source/auth/auth.rst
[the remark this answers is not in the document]
jstewart-mongo: I have removed this section in the last commit.

source/auth/auth.rst
jyemin: I no longer see the test plan in the latest commit.  So I'm not sure.

I think my point was that, given that we don't have any events that are published that prove that the driver is actually doing speculative authentication (unlike with CRUD, where we can use command monitoring to assert the internal behavior of the driver), the best we can do is to just integration test all the authentication methods against server versions which support speculative authentication and server versions that don't.  Drivers generally already have these tests (at least for SCRAM... There are at least some drivers that don't for X509 auth).

Some drivers are designed in a way where this behavior is unit-testable, but I'm not sure we should make that a requirement in the spec (like we do implicitly for SDAM, for example).

source/mongodb-handshake/handshake.rst
```diff
@@ -224,6 +224,99 @@ Example::
...
+            /* REQUIRED */
+            mechanism: "<string>"
+            /* REQUIRED */
+            db: "<string>"
+            payload: BinData(...)
+            options: {
+                skipEmptyExchange: true         /* REQUIRED */        
+            }
+        }
+    }
```
ShaneHarvey: I had trouble reading this because it merges the x509 and SCRAM speculativeAuthenticate documents. What do you think about splitting the description like this:

```
When the mechanism is ``MONGODB-X509``, `speculativeAuthenticate` has the following structure::
    {
        isMaster: 1,
        speculativeAuthenticate: {
            authenticate: 1,
            mechanism: "MONGODB-X509"
        }
    }

When the mechanism is ``SCRAM-SHA-1`` or ``SCRAM-SHA-256``, `speculativeAuthenticate` has the following structure::
    {
        isMaster: 1,
        speculativeAuthenticate: {
            saslStart: 1,
            mechanism: "<string>",   /* "SCRAM-SHA-1" or "SCRAM-SHA-256" */
            db: "<authentication database>",
            payload: BinData(...),
            options: {
                skipEmptyExchange: true         /* REQUIRED */
            }
        }
    }
```

source/auth/auth.rst
jstewart-mongo: Ok, I agree with this. I think it is best to remove the test plan section, since it would be redundant to state that drivers should run the integration tests that already exist.

source/mongodb-handshake/handshake.rst
jstewart-mongo: You are correct. Your suggestion reads much better. I have revised the text.

source/auth/auth.rst
jyemin: @ShaneHarvey what do you think?

source/mongodb-handshake/handshake.rst
jyemin: @ShaneHarvey ok to resolve this thread?

source/auth/auth.rst
```diff
@@ -1193,6 +1194,9 @@ Q: Why does SCRAM sometimes SASLprep and sometimes not?
 Version History
 ===============
 
+Version 1.11.0 Changes
+    * Support speculative authentication via handshake starting in version 4.4 of the server.
+
```
jmikola: It looks like the main content was moved to `mongodb-handshake/handshake.rst`. This file doesn't have anything more than a changelog entry and version bump in its header. Perhaps these should be removed?

jstewart-mongo: I have reverted the authentication spec.

source/mongodb-handshake/handshake.rst
```diff
@@ -224,6 +224,107 @@ Example::
...
+            db: "<authentication database>",
+            payload: BinData(...),
+            options: {
+                skipEmptyExchange: true         /* REQUIRED */
+            }
+        }
+    }  
+
+
+speculativeAuthenticate.mechanism
```
jyemin: As above, I think all this can go away.

source/mongodb-handshake/handshake.rst
```diff
@@ -224,6 +224,107 @@ Example::
...
+
+If an authentication mechanism is not provided either via connection string or code, but
+a credential is provided, drivers MUST use the SCRAM-SHA-256 mechanism for speculative
+authentication. If no authentication mechanism and no credential are provided, drivers
+MUST continue to authenticate using the standard authentication handshake.
+
+Older servers will ignore the ``speculativeAuthenticate`` argument. New servers will
+participate in the standard authentication conversation if this argument is missing.
+
+Drivers MUST set ``speculativeAuthenticate.saslStart`` if either of the authentication
```
jyemin: Given what I said above, this prose can be simplified a bit to just talk about use of `speculativeAuthenticate` in general rather than the specific fields within it.

source/mongodb-handshake/handshake.rst
```diff
@@ -224,6 +224,107 @@ Example::
...
+
+This value notifies newer servers to shorten the authentication conversation by
+one round trip. Older servers will ignore this field.
+
+If the ``isMaster`` command with a ``speculativeAuthenticate`` argument succeeds,
+the client should proceed with the next step of the exchange. If the ``isMaster``
+response does not include a ``speculativeAuthenticate`` reply and the ``ok`` field
+in the ``isMaster`` response is set to 1, drivers MUST authenticate using the standard
+authentication handshake.
+
```
jyemin: I think we should specify what the `speculativeAuthenticate` response looks like.  In particular, is it a full embedded command response with ok, errmsg, code, etc?

source/mongodb-handshake/handshake.rst
```diff
@@ -224,6 +224,107 @@ Example::
...
+This command may be provided to the server in parallel with any standard request for
+supported authentication mechanisms (i.e. ``saslSupportedMechs``). This would permit
+clients to merge their first message with their ``isMaster`` request, and
+receive the authentication reply with the ``isMaster`` reply.
+
+When the mechanism is ``MONGODB-X509``, `speculativeAuthenticate` has the following
+structure::
+
+    {
+        isMaster: 1,
```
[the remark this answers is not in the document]
ShaneHarvey: I'm not sure it's true that speculativeAuthenticate is exactly the saslStart command document for SCRAM since there is a "db" field. With OP_MSG we would actually send a "$db" field. Perhaps this difference should be explicitly called out?

source/mongodb-handshake/handshake.rst
ShaneHarvey: I think this would help.

source/mongodb-handshake/handshake.rst
```diff
@@ -224,6 +224,107 @@ Example::
...
+Speculative Authentication
+--------------------------
+
+:since: 4.4
+
+The ``isMaster`` handshake supports a new argument, ``speculativeAuthenticate``,
+provided as a BSON object. Clients specifying this argument to ``isMaster`` will
+speculatively include the first command of an authentication handshake.
+This command may be provided to the server in parallel with any standard request for
+supported authentication mechanisms (i.e. ``saslSupportedMechs``). This would permit
```
ShaneHarvey: It's not clear to me what the exact interaction with saslSupportedMechs is. This is pymongo's current code that discovers whether we should authenticate with SCRAM-SHA-256 or SCRAM-SHA-1 (only used when the user does not specify an authMechanism):
```python
        cmd = SON([
            ('ismaster', 1),
            ('saslSupportedMechs', source + '.' + credentials.username)])
        mechs = sock_info.command(
            source, cmd, publish_events=False).get('saslSupportedMechs', [])
        if 'SCRAM-SHA-256' in mechs:
            return _authenticate_scram(credentials, sock_info, 'SCRAM-SHA-256')
        else:
            return _authenticate_scram(credentials, sock_info, 'SCRAM-SHA-1')
```

Are we supposed to send speculativeAuthenticate? If so, do we choose SCRAM-SHA-1 or SCRAM-SHA-256? What happens if we choose the wrong mechanism?

source/auth/auth.rst
```diff
@@ -1102,44 +1101,6 @@ As a URI, those have to be UTF-8 encoded and URL-escaped, e.g.:
 - mongodb://%E2%85%A8:IV@mongodb.example.com/admin
 - mongodb://%E2%85%A8:I%C2%ADV@mongodb.example.com/admin
 
---------------------------
-Speculative Authentication
---------------------------
```
ShaneHarvey: I think it would be nice to keep this section and simply link to the handshake spec.

source/mongodb-handshake/handshake.rst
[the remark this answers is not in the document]
jyemin: Oh, quite true!  And I missed it because I assumed it was a fork and lift from the auth spec.

So yeah, let's just specify the differences, i.e. the additional fields that are required.

source/auth/auth.rst
jyemin: Yeah, that matches what you did when you split SDAM, which worked quite nicely.  SGTM.

source/auth/auth.rst
```diff
@@ -1193,6 +1194,9 @@ Q: Why does SCRAM sometimes SASLprep and sometimes not?
 Version History
 ===============
 
+Version 1.11.0 Changes
```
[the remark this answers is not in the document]

source/auth/auth.rst
jstewart-mongo: I have added the link to the handshake spec.

source/mongodb-handshake/handshake.rst
```diff
@@ -224,6 +224,107 @@ Example::
...
+clients to merge their first message with their ``isMaster`` request, and
```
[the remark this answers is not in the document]
jstewart-mongo: Updated

source/mongodb-handshake/handshake.rst
```diff
@@ -224,6 +224,107 @@ Example::
...
+receive the authentication reply with the ``isMaster`` reply.
```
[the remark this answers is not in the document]
jstewart-mongo: Updated

source/mongodb-handshake/handshake.rst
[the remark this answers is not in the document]
jstewart-mongo: I have added text to highlight the differences and added references to the auth spec.

source/mongodb-handshake/handshake.rst
jstewart-mongo: I have removed the descriptions of the fields.

source/mongodb-handshake/handshake.rst
jstewart-mongo: Sending `speculativeAutheticate` is optional. It reduces the number of round trips by 1. The paragraph starting on line 261 states that the SCRAM-SHA-256 mechanism should be used if no mechanism was specified. If the mechanism is not supported, the server will not send a `speculativeAuthenticate` reply in the `isMaster` response, and the standard authentication handshake must be performed.

source/mongodb-handshake/handshake.rst
jyemin: If it has no ok field, how do clients know if authentication has failed? Does the ismaster reply itself have ok: 0 at the top level? I wouldn't think so.

jstewart-mongo: If authentication fails when sending `speculativeAuthenticate`, the `isMaster` response will not contain a `speculativeAuthenticate` reply. Drivers should use the standard authentication handshake in that case. The top-level `ok` field is not affected by a speculative authentication failure.

source/mongodb-handshake/handshake.rst
```diff
@@ -224,6 +224,49 @@ Example::
...
+in the ``isMaster`` response is set to 1, drivers MUST authenticate using the standard
+authentication handshake.
+
+The ``speculativeAuthenticate`` reply has the same fields, except for the ``ok`` field,
+as seen in the conversation sections for MONGODB-X509, SCRAM-SHA-1 and SCRAM-SHA-256
+in the `Driver Authentication spec`_.
+
+If an authentication mechanism is not provided either via connection string or code, but
+a credential is provided, drivers MUST use the SCRAM-SHA-256 mechanism for speculative
+authentication. If no authentication mechanism and no credential are provided, drivers
```
ShaneHarvey: Can you add something about ''saslSupportedMechs" here? Maybe: ```...drivers MUST use the SCRAM-SHA-256 mechanism for speculative authentication and drivers MUST send ``saslSupportedMechs`` ```

source/mongodb-handshake/handshake.rst
ShaneHarvey: Thanks! I made a small suggestion in that paragraph.

source/mongodb-handshake/handshake.rst
jyemin: @sgolemon, this seems odd.  It means that every failed authentication attempt will be tried twice before the driver reports the failure, since the driver can't distinguish the failure of the speculative one from one where the server doesn't support speculation.

Can you comment?

source/mongodb-handshake/handshake.rst
jstewart-mongo: I have updated the paragraph with your suggestion.

source/mongodb-handshake/handshake.rst
sgolemon: Minor nit; `skipEmptyExchange` isn't actually required from a technical point of view.  It's highly encouraged (because why save one packet if you're just going to waste another), but omitting it will still function in a 3 round trip exchange.

source/mongodb-handshake/handshake.rst
sgolemon: @jstewart-mongo is correct.  The problem with reporting failure during speculation is *why* the failure would occur.

For SCRAM methods, a first step failure has nothing to do with incorrect password (since verification hash isn't sent till step two).  The most probable failure here (apart from unknown user because someone is trying to dictionary our auth db) is "incorrect mechanism (e.g. using SCRAM-SHA-256 when user is only configured for SCRAM-SHA-1".  For this, saslSupportedMechs is going to tell you the correct mechanism to use and a retry is appropriate.  If we returned an explicit failure, it would be easy (and understandable) for a client to (incorrectly) abort further authentication attempts without retrying using the saslSupportedMechs hint.

Note that this first attempt is "free" from a network point of view compared to existing authentication attempts, so the fallback to saslSupportedMechs is no worse than present.

For X509, we could have returned failure more eagerly, but that would have been inconsistent with SASL, so we kept the failure case the same.  Of course, with X509 the actual validation was performed during TLS establishment, asking for authentication here is just a formality.

jyemin: Thank you, Sara!  That makes a lot of sense now.

source/mongodb-handshake/handshake.rst
```diff
@@ -224,6 +224,50 @@ Example::
...
+
+The ``speculativeAuthenticate`` reply has the same fields, except for the ``ok`` field,
+as seen in the conversation sections for MONGODB-X509, SCRAM-SHA-1 and SCRAM-SHA-256
+in the `Driver Authentication spec`_.
+
+If an authentication mechanism is not provided either via connection string or code, but
+a credential is provided, drivers MUST use the SCRAM-SHA-256 mechanism for speculative
+authentication and drivers MUST send ``saslSupportedMechs``. If no authentication mechanism
+and no credential are provided, drivers MUST continue to authenticate using the standard
+authentication handshake.
```
ShaneHarvey: I'm not sure I follow this last sentence. If no mechanism and no credentials are provided how can a driver authenticate? Perhaps we can remove this sentence altogether?

jstewart-mongo: Agreed. I have removed the sentence.

source/mongodb-handshake/handshake.rst~
```diff
@@ -0,0 +1,426 @@
+=================
```
ShaneHarvey: I think this file was added by mistake.

jstewart-mongo: Yes it was. I have removed it.

PR #729, opened 2020-02-05
SPEC-1352 remove other_collection_name field

kevinAlbs:
Some follow-up of SPEC-1352. I accidentally introduced the top-level field "other_collection_name". Other spec tests of $out/$merge don't set that as a top-level field, and just reference it with YML anchors ([example](https://github.com/kevinAlbs/specifications/blob/wc-fix/source/crud/tests/v2/aggregate-out-readConcern.yml/#L26)). Credit to @divjotarora for catching this.

For a non-functional change like this, LMK if a separate SPEC ticket is warranted to resync the tests, or if I should add this commit to SPEC-1352.

### DIFF
--- a/source/read-write-concern/tests/operation/default-write-concern-3.2.yml
+++ b/source/read-write-concern/tests/operation/default-write-concern-3.2.yml
@@ -8,7 +8,6 @@ data:
   - {_id: 2, x: 22}
 collection_name: &collection_name default_write_concern_coll
 database_name: &database_name default_write_concern_db
-other_collection_name: &other_collection_name other_collection_name
 
 runOn:
     - minServerVersion: "3.2"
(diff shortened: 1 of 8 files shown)


source/read-write-concern/tests/operation/default-write-concern-3.2.yml
```diff
@@ -8,7 +8,6 @@ data:
   - {_id: 2, x: 22}
 collection_name: &collection_name default_write_concern_coll
 database_name: &database_name default_write_concern_db
-other_collection_name: &other_collection_name other_collection_name
```
jmikola: Just to confirm, this was never used within this file and just copypasted across all of the new tests in your last PR? Likewise for `default-write-concern-2.6.yml`.

kevinAlbs: Correct.

PR #730, opened 2020-02-05
Reword connections survive primary stepdown error handling rule

ShaneHarvey:
This change simplifies the wording of how drivers handle "node is shutting down" errors and updates the pseudocode to include this rule. This does not require any driver changes. The motivation is simply to make the spec easier to read.

### DIFF
--- a/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
+++ b/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
@@ -1455,6 +1455,7 @@ recovering" error::
 
     recovering_codes = [11600, 11602, 13436, 189, 91]
     notmaster_codes = [10107, 13435]
+    shutdown_codes = [11600, 91]
 
     def is_recovering(message, code):
         if code and code in recovering_codes:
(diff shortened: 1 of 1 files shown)


source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1545,14 +1552,9 @@ shutting down" errors:
...
-If the client is connected to server version 4.2 or higher, and the client
-receives a "not master" or "node is recovering" error which is not a
-"node is shutting down" error, the client MUST keep any connections it has to
-the server open, and MUST NOT clear its connection pool for the server.
-If the client is connected to server version 4.2 or higher and receives a
-"node is shutting down" error, or if the client is connected to server version
-4.0 or lower and receives a "not master" or "node is recovering" error,
-the client MUST clear its connection pool to the server.
+When handling a "not master" or "node is recovering" error, the client
+clears the server's connection pool if and only if the error is
```
p-mongo: Since this is a requirement, I am under the impression that the current preference is to express it as "MUST clear" instead of "clears".

ShaneHarvey: Done.

PR #731, opened 2020-02-06
SPEC-1562 transactions test fixes for RetryableWriteError label

iwysiu:
Cleaning up some of the files and adding the RetryableWriteError label to the failpoints that need it.

### DIFF
--- a/source/transactions/tests/retryable-abort-errorLabels.yml
+++ b/source/transactions/tests/retryable-abort-errorLabels.yml
@@ -13,18 +13,18 @@ tests:
       configureFailPoint: failCommand
       mode: { times: 2 }
       data:
-        failCommands: ["abortTransaction"]
-        errorCode: 112 # WriteConflict, not a retryable error code
-        errorLabels: ["RetryableWriteError"] # Override server behavior: send RetryableWriteError label with non-retryable error code
+          failCommands: ["abortTransaction"]
+          errorCode: 112 # WriteConflict, not a retryable error code
+          errorLabels: ["RetryableWriteError"] # Override server behavior: send RetryableWriteError label with non-retryable error code
     operations:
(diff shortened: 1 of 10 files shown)


source/transactions/tests/retryable-commit-errorLabels.yml
```diff
@@ -55,10 +55,10 @@ tests:
             commitTransaction: 1
             lsid: session0
             txnNumber:
-            $numberLong: "1"
+              $numberLong: "1"
             startTransaction:
             autocommit: false
-            writeConcern: { w: majority, wtimeout: 10000 }
+            writeConcern:
```
iwysiu: This is the only logic change I made, it doesn't get retried, so it shouldn't have a write concern.

source/transactions/tests/retryable-abort-errorLabels.yml
```diff
@@ -13,18 +13,18 @@ tests:
       configureFailPoint: failCommand
       mode: { times: 2 }
       data:
-        failCommands: ["abortTransaction"]
-        errorCode: 112 # WriteConflict, not a retryable error code
-        errorLabels: ["RetryableWriteError"] # Override server behavior: send RetryableWriteError label with non-retryable error code
+          failCommands: ["abortTransaction"]
+          errorCode: 112 # WriteConflict, not a retryable error code
+          errorLabels: ["RetryableWriteError"] # Override server behavior: send RetryableWriteError label with non-retryable error code
```
mbroadst: Can you include the SPEC ticket that reminds us to remove this?

iwysiu: That label is actually part of the test, and should still be there after the other ones get removed.

PR #732, opened 2020-02-10
SPEC-1460 Require inconsistent write concern validation


p-mongo:


p-mongo:
https://jira.mongodb.org/browse/SPEC-1460

source/read-write-concern/read-write-concern.rst
```diff
@@ -332,17 +332,13 @@ same as :javascript:`writeConcern: {journal: false}`.
 Inconsistent WriteConcern
 -------------------------
 
-Drivers SHOULD raise an error when an inconsistent ``WriteConcern`` is
```
jmikola: Perhaps you were waiting until merge time, but just in case: I think this warrants bumping the spec version to the next minor number, updating the last modified date, and adding a line to the changelog.

source/read-write-concern/read-write-concern.rst
```diff
@@ -332,17 +332,13 @@ same as :javascript:`writeConcern: {journal: false}`.
...
+Drivers MUST raise an error when an inconsistent ``WriteConcern`` is
 specified. The following is an exhaustive list of inconsistent ``WriteConcerns``:
 
 .. code:: typescript
 
    writeConcern = { w: 0, journal: true };
 
-If, for the sake of backwards compatibility, the driver allows inconsistent
-write concerns where ``w`` equals 0 but ``journal`` is set to ``true``, the
-driver MUST treat it as an ``Acknowledged WriteConcern``.
```
jmikola: What is the rationale for removing this paragraph? I think it's helpful given there are a number of drivers that may not be able to conform to the new MUST requirement for BC reasons (similar to your unified discovery spec).

p-mongo: It was my understanding that the requirement in [the spec ticket](https://jira.mongodb.org/browse/SPEC-1460) was to disallow inconsistent write concerns. This paragraph would have conflicted with such a requirement hence I removed it. What would you like me to do instead?

jmikola: @jyemin @mbroadst: I wonder, would it be a BC break if a driver were to start raising a client-side error for this write concern in a new minor release? I wouldn't consider it an API change.

If this paragraph does need to stay, I'd propose revising it as:

> If, for the sake of backwards compatibility, a driver allows the write concern `{ w: 0, journal: true }` to be constructed, that driver MUST treat it as an acknowledged write concern. This behavior MUST be changed to comply with this specification in that driver's next semver major version bump.

jyemin: I would be comfortable with this sort of break in a minor release, unless a driver's public reference or API docs specifically called it out as an allowed combination.

mbroadst: I'd prefer not to codify when a change should be made in this text, which is meant to be the "state of the art" and long-lived. Rather, if we agree with this change, then let's merge it in and create drivers tickets where each driver can figure out which version it most make sense to introduce the change in

source/read-write-concern/read-write-concern.rst
jmikola: LGTM-ed, but don't forget to bump this info above.

PR #734, opened 2020-02-10
Add OCSP cache requirement

vincentkam:


source/ocsp-support/ocsp-support.rst
```diff
@@ -3,14 +3,14 @@ OCSP Support
...
-:Spec Version: 1.0.2
+:Spec Version: 1.1.0
 :Author: Vincent Kam
 :Lead: Jeremy Mikola
 :Advisory Group: Clyde Bazile *(POC author)*, Esha Bhargava *(Program Manager)*, Matt Broadstone, Bernie Hackett *(POC author)*, Shreyas Kaylan *(Server Project Lead)*, Jeremy Mikola *(Spec Lead)*
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: 4.4
-:Last Modified: 2019-1-28
+:Last Modified: 2019-2-10
```
divjotarora: 2020

source/ocsp-support/ocsp-support.rst
```diff
@@ -134,6 +134,42 @@ Drivers SHOULD validate OCSP Responses in the manner specified in `RFC
 6960: 3.2 <https://tools.ietf.org/html/rfc6960#section-3.2>`__ to the
 extent that their TLS library allows.
 
+Suggested OCSP Caching Behavior
+-------------------------------
+Drivers with sufficient control over their TLS library's OCSP
+behavior SHOULD implement an OCSP cache. The key for this cache
+SHOULD be the certificate identifier (CertID) as specified in
```
divjotarora: nit: change to "certificate identifier (CertID) of the OCSP request"

source/ocsp-support/ocsp-support.rst
```diff
@@ -134,6 +134,42 @@ Drivers SHOULD validate OCSP Responses in the manner specified in `RFC
...
+       hashAlgorithm       AlgorithmIdentifier,
+       issuerNameHash      OCTET STRING, -- Hash of issuer's DN
+       issuerKeyHash       OCTET STRING, -- Hash of issuer's public key
+       serialNumber        CertificateSerialNumber }
+
+If a driver would accept a conclusive OCSP response (stapled or
+non-stapled), the driver SHOULD cache that response. We define a
+conclusive OCSP response as an OCSP response that indicates that a
+certificate is either valid or revoked. Thus, an unknown certificate
+status SHOULD BE considered conclusive and the corresponding OCSP
```
divjotarora: I think this is supposed to say "SHOULD NOT"

vincentkam: Already fixed! You started to review faster than I could push apparently :)

source/ocsp-support/ocsp-support.rst
vincentkam: Done!

source/ocsp-support/ocsp-support.rst
vincentkam: Erk, fixed! Right, it's 2020 now :)

source/ocsp-support/ocsp-support.rst
```diff
@@ -134,6 +134,41 @@ Drivers SHOULD validate OCSP Responses in the manner specified in `RFC
...
 extent that their TLS library allows.
 
+Suggested OCSP Caching Behavior
+-------------------------------
+Drivers with sufficient control over their TLS library's OCSP
+behavior SHOULD implement an OCSP cache. The key for this cache
+SHOULD be the certificate identifier (CertID) of the OCSP request
+as specified in `RFC 6960: 4.1.1
+<https://tools.ietf.org/html/rfc6960#section-4.1.1>`__.
+For convenience, the relevant section has been duplicated below\:
```
jmikola: Is is necessary to escape the colon? I thought RST only applies special formatting to the subsequent content if a line ends with `::`.

source/ocsp-support/ocsp-support.rst
```diff
@@ -134,6 +134,41 @@ Drivers SHOULD validate OCSP Responses in the manner specified in `RFC
...
+       issuerNameHash      OCTET STRING, -- Hash of issuer's DN
+       issuerKeyHash       OCTET STRING, -- Hash of issuer's public key
+       serialNumber        CertificateSerialNumber }
+
+If a driver would accept a conclusive OCSP response (stapled or
+non-stapled), the driver SHOULD cache that response. We define a
+conclusive OCSP response as an OCSP response that indicates that a
+certificate is either valid or revoked. Thus, an unknown certificate
+status SHOULD NOT be considered conclusive, and the corresponding OCSP
+response SHOULD NOT be cached.
```
jmikola: I think we can use "MUST NOT" for the two "SHOULD NOT" instances in this sentence. We're saying that drivers SHOULD cache results, but I think we can be more rigid when it comes to instructing drivers about whether or not to cache.

Alternatively, are you using SHOULD here because some drivers may have no control over the ability to (a) use a cache and (b) control what added to that cache? If so, then I suppose SHOULD makes sense.

source/ocsp-support/ocsp-support.rst
vincentkam: Ah you're right: my mistake! Fixed.

source/ocsp-support/ocsp-support.rst
vincentkam: Using MUST would be nice. The limiting factor, as you correctly suspected, is in fact due to potential limitations in drivers' control over their OCSP caching behavior. Dotnet for example uses a cache, but has zero control over what happens with that cache.

vincentkam:
> Do we want to think about only holding the cached responses until the midpoint between `thisUpdate` and `nextUpdate`? I think the server was considering that approach but it's not clear to me what the benefit is.

I'm fine with waiting till `nextUpdate`: I'm also not sure if there's any benefit for drivers to halve the validity period: I've pinged Shreyas offline in the channel and looped you in so we can confirm this :)

vincentkam:
Shreyas responded offline: he agrees that there shouldn't be any issues with drivers caching till `nextUpdate`. I've also clarified the wording regarding how the cache should interact with stapled responses, using slightly more precise language per Shreyas's description.
@divjotarora Ready for review again :)

vincentkam:
If there are no objections, I will push this once I receive Divjot's LGTM 😄

PR #735, opened 2020-02-10
Remove redundant content in Tests section

jmikola:


jmikola:
I would prefer to hold off on re-organizing the prose tests until the unified test format is implemented ([DRIVERS-709](https://jira.mongodb.org/browse/DRIVERS-709)). The reason for that is our existing retryable writes tests don't have syntax for APM expectations, but once we do, the entire [Command Construction Tests](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/README.rst#command-construction-tests) section can be ported to spec tests and we'd only need to move [Split Batch Tests](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/README.rst#split-batch-tests) to the prose section. This will avoid introducing numbered prose tests for a brief period and leaving a gap of unused numbers (since we don't want to re-use numbers for old prose tests once they're deleted).

I've created [SPEC-1572](https://jira.mongodb.org/browse/SPEC-1572) to track this and also consolidate the driver changes that will be needed by the conversion and relocation.

PR #736, opened 2020-02-11
SPEC-1505 Use a whitelist for classifying change stream errors

divjotarora:
## Overview of changes

- Describe new criteria for determining resumable error (new label for wire versions >= 9, manual whitelist checking for wire versions < 9)
- Change the "What do the additional error codes mean?" section to "Why is a whitelist of error codes preferable to a blacklist?"
- Change the test runner description to ignore `killCursors` events because sending `killCursors` during a resume attempt is a MAY in the spec
- Remove prose test for specific blacklisted error codes
- Remove prose test that asserted that a change stream resumes if the underlying cursor is killed. This was done because `CursorNotFound` is not a whitelisted error so it will no longer resumed.

## New Tests

- One per whitelisted error code (servers with new error label)
	- Min server version 4.3
	- Use `failGetMoreAfterCursorCheckout` to force whitelisted error
	- Assert that driver resumes on failed `getMore` and returns an event without errors
- One per whitelisted error code (servers without new error label)
	- Min server version 4.2, max server version 4.2.99
	- Use `failCommand` to force whitelisted error
	- Assert that driver resumes on failed `getMore` and returns an event without errors
- Driver does not resume
	- Min server version 4.2
	- Use `failCommand` to force an error not on the whitelist or previous blacklist
	- Assert that error is propagated to the user
- Driver resumes if error contains whitelist label on latest:
	- Min server version 4.3
	- Use `failCommand` to force an error not on the whitelist but include `ResumableChangeStreamError` label
	- - Assert that driver resumes on failed `getMore` and returns an event without errors
- Driver does not resume if error does not contain whitelist label on latest:
	- Min server version 4.3
	- Use `failCommand` to force an error on the whitelist
		- Because `failCommand` is used instead of the new fail point, the error will not contain the whitelist label
	- Assert that error propagated to the user

### Tests that were originally planned but were not added

- Driver doesn't resume if non-whitelisted error contains label on 4.2.x:
	- Min server 4.2, max 4.2.99
	- Use `failCommand` to force error not on whitelist but include whitelist label
	- Assert error is propagated to the user
	- **NOT POSSIBLE**: The ability to specify error labels on `failCommand` wasn't introduced until [SERVER-43941]( https://jira.mongodb.org/browse/SERVER-43941). This was released in 4.3.1, so it's not available when testing against 4.2.x.
- Driver resumes if whitelisted error does not contain label on 4.2.x:
	- Min server 4.2, max 4.2.99
	- Use `failCommand` to force error on whitelist but remove error label
	- Assert that driver resumes
	- Assert that driver returns an event without any errors
	- **NOT NEEDED**: the previous tests for whitelisted errors on 4.2 already accomplish this because `failCommand` never adds the label

## Additional Notes

- No test was possible for the `StaleConfig` error because it requires an `extraInfo` field. Support for this was added to the failpoint command in [SERVER-42239](https://jira.mongodb.org/browse/SERVER-42239) but that's in 4.3.3 so it's not usable for the 4.2 tests and the new failpoint used for the 4.4 tests does not support it
- Because `CursorNotFound` is no longer a resumable error, drivers that use it to force resume attempts in tests will need to mock a different error.
- Any drivers that disabled the failing test in `change-streams-errors.json` after SERVER-45505 was merged can disable it and expect it to pass as-is.
- Drivers will need to add failpoint support to their change stream test runners if they have not already implemented https://jira.mongodb.org/browse/SPEC-1073.
- Change stream spec test runners will need to pin to a single mongos for test execution. Otherwise, tests that set failpoints will not consistently pass.

The "Additional Notes" will likely be added as bullet points to the SPEC ticket in Jira so language authors can be aware of them.


source/change-streams/change-streams.rst
```diff
@@ -43,24 +43,56 @@ Resumable Error
...
 
     * - Error Name
       - Error Code
-    * - Interrupted
-      - 11601
-    * - CappedPositionLost
-      - 136
-    * - CursorKilled
-      - 237
+    * - HostUnreachable
```
divjotarora: This list was generated based on the server changes made in https://github.com/mongodb/mongo/commit/3d821c25e2944668b7359a0bf6e586cc8625b9a2#diff-cf7105bb60d0308203e7ef6ec12e07c3R74 (SERVER-45505).

source/change-streams/change-streams.rst
```diff
@@ -679,10 +711,9 @@ It was decided to remove this example from the specification for the following r
 What do the additional error codes mean?
 ----------------------------------------
 
-The `CursorKilled` or `Interrupted` error implies some other actor killed the cursor.
-
-The `CappedPositionLost` error implies falling off of the back of the oplog,
-so resuming is impossible.
+The additional codes were added to match the server changes made in
```
divjotarora: Not sure if this section is necessary anymore. I'm open to removing it completely if others agree.

source/change-streams/change-streams.rst
```diff
@@ -9,8 +9,8 @@ Change Streams
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: 3.6
-:Last Modified: April 3, 2019
-:Version: 1.7.0
+:Last Modified: Febrary 10, 2020
+:Version: 1.7.1
```
divjotarora: I wasn't sure what constitutes a minor vs patch version bump. Should this be 1.7.1 or 1.8.0?

source/change-streams/tests/change-streams-errors.json
```diff
@@ -102,10 +102,7 @@
       ],
       "result": {
         "error": {
-          "code": 280,
-          "errorLabels": [
-            "NonResumableChangeStreamError"
-          ]
+          "code": 20
```
divjotarora: As part of the changes for SERVER-45505, the query team changed this from generic error code 280 (`ChangeStreamFatalError`) to more specific error code 20 (`IllegalOperation`). I left this test in because it asserts that errors not in the whitelist are propagated to the user.

source/change-streams/change-streams.rst
kevinAlbs: Nice. Verified this list is correct by going through error_codes.yml:
https://gist.github.com/kevinAlbs/c5eeead90dca71097ee111530f5b1704

source/change-streams/change-streams.rst
kevinAlbs: IMO 1.8.0 seems warranted. This is a new behavior change. That being said I often forget to update the version...

source/change-streams/change-streams.rst
kevinAlbs: I like having it. It does provide convenient rationale for future readers. But also not opposed to removing it, as readers can go through the git history to see this conversation.

source/change-streams/change-streams.rst
kevinAlbs: Also, questions for other reviewers. As @divjotarora pointed out in the code review for SERVER-45505, the server adds ResumableChangeStreamError to replies of aggregate as well. Copying his comment here:

> The current drivers specification
> (https://github.com/mongodb/specifications/blob/master/source/change-streams/c...)
> explicitly states that all errors on aggregate are not resumeable errors. The
> operations layer of the driver will retry the aggregate once for some subset of
> errors (e.g. network failures, NotMaster errors) per the retryable reads
> specification, but if both the original and retry error, the error will be
> reported up to the change stream code, which won't look for this label to figure
> out if another aggregate should be sent. I believe this was done because it does
> not make sense to "resume" a change stream that hasn't been established in the
> first place.

Does it seem reasonable to resume on aggregate if it contains that label? If so, I can create a separate spec ticket. I don't think we should do it as part of this spec ticket, since it would be a non-trivial additional change, and outside of the scope of the problem this spec ticket solves.

source/change-streams/tests/change-streams-resume.yml
```diff
@@ -0,0 +1,140 @@
...
+        -
+          _id: "42"
+          documentKey: "42"
+          operationType: insert
+          ns:
+            db: *database_name
+            coll: *collection_name
+          fullDocument:
+            x:
+              $numberInt: "1"
```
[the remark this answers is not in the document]
divjotarora: Followed up offline that the first suggestion wouldn't catch the mistake of leaving in the blacklist logic. For the second suggestion, the min server version has to be 4.2 for the failpoint limitations, but I like this idea. I'll send out a full list of proposed test cases on Slack later today.

source/change-streams/tests/change-streams-resume.yml
```diff
@@ -0,0 +1,140 @@
+collection_name: &collection_name "test"
+database_name: &database_name "change-stream-tests"
+collection2_name: &collection2_name "test2"
+database2_name: &database2_name "change-stream-tests-2"
+tests:
+  -
+    description: "change stream resumes after a network error"
+    minServerVersion: "4.2"
```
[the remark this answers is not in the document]
kevinAlbs: Ah right, I forgot that these fail points were introduced later.

source/change-streams/tests/change-streams-resume.yml
```diff
@@ -0,0 +1,140 @@
...
+          database_name: *database_name
+      -
+        command_started_event:
+          command:
+            getMore: "42"
+            collection: *collection_name
+          command_name: getMore
+          database_name: *database_name
+      -
+        command_started_event:
```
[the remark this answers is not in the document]
divjotarora: Removed the `killCursors` expectations from the tests and updated the logic. I don't think the section you linked is the right place because it's talking about generically comparing BSON/JSON docs. I made the change in the part that talks about comparing `expectations` specifically.

divjotarora:
FYI: There may be a server issue whereby the `ResumableChangeStreamError` label will not be added to `getMore` commands that fail because of a fail point. I brought this up to the Query team and they're investigating. Making these tests work may involve another fail point.

source/change-streams/change-streams.rst
jyemin: It's going to require significant refactoring in some drivers to be able to resume aggregate.  I would prefer that the server, instead of labelling an aggregate response with a ResumableChangeStreamError, instead just returned a response with no documents in firstBatch, and then failed the first getMore.

But if that turns out to not be feasible, we'll have to take this on as new work.

source/change-streams/tests/change-streams-errors.json
jyemin: And cause problems for older drivers?  At the least, test expectations will be different.

source/change-streams/tests/change-streams-errors.yml
```diff
@@ -71,5 +71,4 @@ tests:
             z: 3
     result:
       error:
-        code: 280
-        errorLabels: [ "NonResumableChangeStreamError" ]
+        code: 20
```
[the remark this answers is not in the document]
jyemin: Let's also test for `280` error code on 4.2 servers.

Are we sure we don't want to push back on this error code change in the server?  Seems like this could break applications using older drivers against a 4.4 server.

source/change-streams/tests/change-streams-resume-errorLabels.yml
```diff
@@ -0,0 +1,1134 @@
+# Tests for resume behavior on server versions that support the ResumableChangeStreamError label
+collection_name: &collection_name "test"
+database_name: &database_name "change-stream-tests"
+collection2_name: &collection2_name "test2"
```
kevinAlbs: collection2_name and database2_name don't seem to be referenced in the rest of this file.

source/change-streams/tests/change-streams-resume-errorLabels.yml
```diff
@@ -0,0 +1,1134 @@
...
+          documentKey: "42"
+          operationType: insert
+          ns:
+            db: *database_name
+            coll: *collection_name
+          fullDocument:
+            x:
+              $numberInt: "1"
+  -
+    description: "change stream resumes after HostNotFound"
```
kevinAlbs: If the resuming tests on whitelisted errors are all the same except for the failpoint error code and description, you could use YAML aliases to deduplicate. Example: https://gist.github.com/kevinAlbs/5b520389a0509c2544fbb8633ba18cf3#file-anchors-yml-L66-L70
But I'm ok with this as-is too.

source/change-streams/tests/change-streams-resume-whitelist.yml
```diff
@@ -0,0 +1,1138 @@
+# Tests for resume behavior on server versions that do not support the ResumableChangeStreamError label
+collection_name: &collection_name "test"
+database_name: &database_name "change-stream-tests"
+collection2_name: &collection2_name "test2"
```
kevinAlbs: Same here, collection2/database2 seems unused.

source/change-streams/tests/README.rst
```diff
@@ -155,12 +155,12 @@ The following tests have not yet been automated, but MUST still be tested. All t
 #. ``ChangeStream`` will throw an exception if the server response is missing the resume token (if wire version is < 8, this is a driver-side error; for 8+, this is a server-side error)
 #. After receiving a ``resumeToken``, ``ChangeStream`` will automatically resume one time on a resumable error with the initial pipeline and options, except for the addition/update of a ``resumeToken``.
 #. ``ChangeStream`` will not attempt to resume on any error encountered while executing an ``aggregate`` command. Note that retryable reads may retry ``aggregate`` commands. Drivers should be careful to distinguish retries from resume attempts. Alternatively, drivers may specify `retryReads=false` or avoid using a [retryable error](../../retryable-reads/retryable-reads.rst#retryable-error) for this test.
-#. ``ChangeStream`` will not attempt to resume after encountering error code 11601 (Interrupted), 136 (CappedPositionLost), or 237 (CursorKilled) while executing a ``getMore`` command.
```
kevinAlbs: Is this being removed because it is redundant with the new "change stream errors on MaxTimeMSExpired" test? This test would still pass since there won't be a resumable label on any of these errors, right?

source/change-streams/change-streams.rst
```diff
@@ -43,35 +43,60 @@ Resumable Error
...
+      - 234
+    * - FailedToSatisfyReadPreference
+      - 133
+    * - ElectionInProgress
+      - 216
 
 An error on an aggregate command is not a resumable error. Only errors on a
 getMore command may be considered resumable errors.
 
-The criteria for resumable errors is similar to the discussion in the SDAM
```
kevinAlbs: This is here because the spec used to explicitly call out that error messages containing the substring "not master" or "node is recovering" as resumable. (libmongoc currently the check as well).

I looked through the history:
https://jira.mongodb.org/browse/SPEC-1051 introduced the "not master" and "node is recovering" criteria:
> - a server error response without an error code or one different from those
  listed above, but with an error message containing the substring "not
  master" or "node is recovering"

https://jira.mongodb.org/browse/SPEC-1124 removed the bullet

With this new change I think you can close https://jira.mongodb.org/browse/SPEC-1328 ("Change stream spec need not discuss "not master" errors") as "Gone Away"

If it seems out of scope we can defer, but I'd vote for adding another test that checks an error that is not on the whitelist but has the error message "not master" is not resumed.

source/change-streams/tests/change-streams-errors.yml
```diff
@@ -73,3 +73,32 @@ tests:
...
         errorLabels: [ "NonResumableChangeStreamError" ]
+  -
+    description: change stream errors on MaxTimeMSExpired
+    minServerVersion: "4.2"
+    failPoint:
+      configureFailPoint: failCommand
+      mode: { times: 1 }
+      data:
+          failCommands: ["getMore"]
+          errorCode: 50 # An error code that's not on the blacklist or whitelist
```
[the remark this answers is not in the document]
divjotarora: Done.

source/change-streams/tests/change-streams-resume-errorLabels.yml
divjotarora: Removed.

source/change-streams/tests/change-streams-resume-whitelist.yml
divjotarora: Removed.

source/change-streams/tests/README.rst
divjotarora: This test would still pass. I removed it because it's explicitly calling out the codes that were on the old blacklist. In my opinion, this doesn't add any coverage outside of the new spec test that you mentioned. Also, it makes these three errors seem special even though they're not, which could confuse someone reading the spec for the first time in the future if they don't have the all of the past context.

source/change-streams/change-streams.rst
divjotarora: I read through the spec tickets you linked and some of our specs do still mention checking for the error message substrings. The recent work for using error labels in retryable writes removed that check from the retryable writes work because nobody on the drivers or server team thought it was necessary to have checks for error codes not existing.

source/change-streams/tests/change-streams-resume-errorLabels.yml
divjotarora: Maybe I just don't have enough experience reading YAML, but I find the tests more readable this way. Open to changing if others agree though.

source/change-streams/tests/README.rst
kevinAlbs: That seems reasonable, just double checking.

PR #737, opened 2020-02-11
SPEC-1562 transaction test error labels updates

rozza:


source/transactions/tests/error-labels.yml
```diff
@@ -587,6 +587,7 @@ tests:
           writeConcernError:
             code: 91
             errmsg: Replication is being shut down
+            errorLabels: ["RetryableWriteError"]
```
rozza: I believe this is required for servers with a max wire version >= 9

source/transactions/tests/retryable-commit-errorLabels.yml
```diff
@@ -33,7 +33,7 @@ tests:
       - name: commitTransaction
         object: session0
         result:
-          errorLabelsOmit: ["RetryableWriteError", "TransientTransactionError", "UnknownTransactionCommitResult"]
+          errorLabelsOmit: ["RetryableWriteError", "TransientTransactionError"]
```
rozza: Although its not retryable I believe it still can be marked as `UnknownTransactionCommitResult`

source/transactions/tests/retryable-commit.yml
```diff
@@ -252,16 +252,16 @@ tests:
...
 
     failPoint:
       configureFailPoint: failCommand
       mode: { times: 1 }
       data:
           failCommands: ["commitTransaction"]
           writeConcernError:
-            code: 11601
-            errmsg: operation was interrupted
+            code: 100
```
rozza: I think 11601 should still produce a UnknownTransactionCommitResult - so opted to use 100 which explictly is not retryable and won't result in a UnknownTransactionCommitResult.

source/transactions/tests/retryable-commit-errorLabels.yml
ShaneHarvey: Yes that sounds correct to me. Interesting to note is that although session.commit_transaction() will not retry, withTransaction will.

source/transactions/tests/retryable-abort-errorLabels.yml
```diff
@@ -1,7 +1,7 @@
 runOn:
     -
         minServerVersion: "4.3.1"
-        topology: ["replicaset, sharded"]
+        topology: ["replicaset", "sharded"]
```
iwysiu: Now that these test run, the driver ignores all abortTransaction command errors, so it wouldn't get a server error to have labels on line 96: https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#drivers-ignore-all-aborttransaction-errors

rozza: @iwysiu I'm not sure I follow is the test redundant or worth keeping to ensure `RetryableWriteError` is omitted?

Or is it that `abortTransaction only retries once with RetryableWriteError from server` is actually wrong as it should be ignored?

iwysiu: sorry, I explained it badly. `abortTransaction does not retry without RetryableWriteError label` shouldn't have a result.errorLabelsOmit field because the driver shouldn't return the abortTransaction error to the user.

rozza: Got it - thanks :) I've updated that test also.

PR #738, opened 2020-02-11
Drivers may skip MMAPv1 test on mongos

jmikola:
https://jira.mongodb.org/browse/SPEC-1573

### DIFF
--- a/source/retryable-writes/tests/README.rst
+++ b/source/retryable-writes/tests/README.rst
@@ -317,6 +317,10 @@ and sharded clusters.
     retryWrites=false to your connection string.
 
    and the error code is 20.
+   
+   **Note**: Drivers that rely on ``serverStatus`` to determine the storage engine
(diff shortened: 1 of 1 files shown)


source/retryable-writes/tests/README.rst
```diff
@@ -317,6 +317,10 @@ and sharded clusters.
     retryWrites=false to your connection string.
 
    and the error code is 20.
+   
+   **Note**: Drivers that rely on ``serverStatus`` to determine the storage engine
+   in use MAY skip this test for sharded clusters, since ``mongos`` does not report
+   this information in its ``serverStatus`` response.
```
jmikola: Opting for a MAY here, since drivers would certainly detect MMAPv1 through other means. Perhaps there is an environment variable that indicates what storage engine is being used throughout the sharded cluster. Alternatively, a driver could poke through and talk directly to the shard's primary to determine what storage engine is running.

PR #739, opened 2020-02-12
SPEC-1248 Unify behavior around configuration for replica set discovery 

p-mongo:
https://jira.mongodb.org/browse/SPEC-1248

### DIFF
--- a/source/initial-dns-seedlist-discovery/tests/direct-connection-true.yml
+++ b/source/initial-dns-seedlist-discovery/tests/direct-connection-true.yml
@@ -2,4 +2,4 @@ uri: "mongodb+srv://test3.test.build.10gen.cc/?directConnection=true"
 seeds: []
 hosts: []
 error: true
-comment: Should fail because directConnection=false is incompatible with SRV URIs.
+comment: Should fail because directConnection=true is incompatible with SRV URIs.
--- a/source/initial-dns-seedlist-discovery/tests/direct-connection-true.yml
+++ b/source/initial-dns-seedlist-discovery/tests/direct-connection-true.yml
@@ -1,0 +1,5 @@
+uri: "mongodb+srv://test3.test.build.10gen.cc/?directConnection=true"
+seeds: []
+hosts: []
+error: true
+comment: Should fail because directConnection=false is incompatible with SRV URIs.
--- a/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
+++ b/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
@@ -365,18 +365,36 @@ The hostname portion of each address MUST be normalized to lower-case.
 Initial TopologyType
 ~~~~~~~~~~~~~~~~~~~~
 
-The user MUST be able to set the initial TopologyType to Single.
+If the ``directConnect`` URI option is specified when a MongoClient is
+constructed, the TopologyType must be initialized based on the value of
+the option according to the following table:
 
-The user MAY be able to initialize it to ReplicaSetNoPrimary.
-This provides the user a way to tell the client
-it can only connect to replica set members.
-Similarly the user MAY be able to initialize it to Sharded,
-to connect only to mongoses.
++-------------------------------+-----------------------+
+| directConnect option value    | Initial TopologyType  |
++===============================+=======================+
+| true                          | Single                |
++-------------------------------+-----------------------+
+| false                         | Unknown               |
++-------------------------------+-----------------------+
 
-The user MAY be able to initialize it to Unknown, to allow for discovery of any
-topology type based only on ismaster responses.
+If directConnect option is not specified, newly developed drivers MUST behave
+as if it was specified with the false value.
 
-The API for initializing TopologyType is not specified here.
+Since changing the starting topology can reasonably be considered a
+backwards-breaking change, existing drivers SHOULD stage implementation
+according to semantic versioning guidelines. Specifically, support for the
+``directConnection`` URI option can be added in a minor release.
+In a subsequent major release, the default starting topology can be changed
+to Unknown. Drivers MUST document this in a prior minor release.
+
+Existing drivers MUST deprecate other URI options, if any, for controlling
+topology discovery or specifying the deployment topology. If such a egacy
+option is specified and the ``directConnection`` option is also
+specified, and the values of the two options are semantically different,
+the driver MUST report an error during URI option parsing.
+
+The API for initializing TopologyType using language-specific native options
+is not specified here.
 Drivers might already have a convention, e.g. a single seed means Single,
 a setName means ReplicaSetNoPrimary,
 and a list of seeds means Unknown.
@@ -394,7 +412,13 @@ as long as all the required features are somehow supported.
 Initial setName
 ~~~~~~~~~~~~~~~
 
-The user MUST be able to set the client's initial replica set name.
+It is allowed to use ``directConnection=true`` in conjunction with the
+``replicaSet`` URI option. The driver must connect in Single topology and
+verify that setName matches the specified name, as per
+`verifying setName with TopologyType Single`_.
+
+When a MongoClient is initialized using language-specific native options,
+the user MUST be able to set the client's initial replica set name.
 A driver MAY require the set name in order to connect to a replica set,
 or it MAY be able to discover the replica set name as it connects.
 
@@ -404,10 +428,21 @@ Allowed configuration combinations
 Drivers MUST enforce:
 
 * TopologyType Single cannot be used with multiple seeds.
+* ``directConnection=true`` cannot be used with multiple seeds.
 * If setName is not null, only TopologyType ReplicaSetNoPrimary,
   and possibly Single,
   are allowed.
   (See `verifying setName with TopologyType Single`_.)
+
+Handling of SRV URIs resolving to single host
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When a driver is given an SRV URI, if the ``directConnection`` URI option
+is not specified, and the ``replicaSet`` URI option is not specified, the
+driver MUST start in Unknown topology, and follow the rules in the
+`TopologyType table`_ for transitioning to other topologies. In particular,
+the driver MUST NOT use the number of hosts from the initial SRV lookup
+to decide what topology to start in.
 
 heartbeatFrequencyMS
 ````````````````````
@@ -2365,6 +2400,12 @@ If the application uses some servers very infrequently,
 monitoring can also proactively detect state changes
 (primary stepdown, server becoming unavailable)
 that would otherwise cause future errors.
+
+Why is auto-discovery the preferred default?
+''''''''''''''''''''''''''''''''''''''''''''
+
+Auto-discovery is most resilient and is therefore preferred.
+
 
 Acknowledgments
 ---------------
--- a/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
+++ b/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
@@ -399,20 +399,15 @@ specified, and the values of the two options are semantically different,
 the driver MUST report an error during URI option parsing.
 
 The API for initializing TopologyType using language-specific native options
-is not specified here.
-Drivers might already have a convention, e.g. a single seed means Single,
-a setName means ReplicaSetNoPrimary,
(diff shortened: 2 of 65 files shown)


p-mongo:
https://jira.mongodb.org/browse/SPEC-1248

p-mongo:
Spec tests are in separate PR for easier review: https://github.com/mongodb/specifications/pull/743

p-mongo:
Some of the changes may have been cosmetic. I removed and regened all json files and there are no changes.

p-mongo:
For example this one:

<pre>

diff --git a/source/server-discovery-and-monitoring/tests/rs/compatible.yml b/source/server-discovery-and-monitoring/tests/rs/compatible.yml
index 3e041d6..6fade62 100644
--- a/source/server-discovery-and-monitoring/tests/rs/compatible.yml
+++ b/source/server-discovery-and-monitoring/tests/rs/compatible.yml
@@ -1,5 +1,7 @@
 description: "Replica set member with large maxWireVersion"
+
 uri: "mongodb://a,b/?replicaSet=rs"
+
 phases: [
     {
         responses: [
</pre>

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -365,18 +365,41 @@ The hostname portion of each address MUST be normalized to lower-case.
...
+| true             | no                    | Single                |
++------------------+-----------------------+-----------------------+
+| true             | yes                   | Single                |
++------------------+-----------------------+-----------------------+
+| false            | no                    | Unknown               |
++------------------+-----------------------+-----------------------+
+| false            | yes                   | ReplicaSetNoPrimary   |
++------------------+-----------------------+-----------------------+
+
+If directConnection option is not specified, newly developed drivers MUST behave
```
divjotarora: Nit: use a code block for directConnection for consistency.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -365,18 +365,41 @@ The hostname portion of each address MUST be normalized to lower-case.
...
+
+Since changing the starting topology can reasonably be considered a
+backwards-breaking change, existing drivers SHOULD stage implementation
+according to semantic versioning guidelines. Specifically, support for the
+``directConnection`` URI option can be added in a minor release.
+In a subsequent major release, the default starting topology can be changed
+to Unknown. Drivers MUST document this in a prior minor release.
+
+Existing drivers MUST deprecate other URI options, if any, for controlling
+topology discovery or specifying the deployment topology. If such a egacy
```
divjotarora: typo: egacy -> legacy

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -404,11 +433,22 @@ Allowed configuration combinations
 Drivers MUST enforce:
 
 * TopologyType Single cannot be used with multiple seeds.
+* ``directConnection=true`` cannot be used with multiple seeds.
```
divjotarora: Isn't this the same as the bullet above? My understanding is that the topology type cannot be Single unless `directConnection` is set to true.

source/uri-options/uri-options.rst
```diff
@@ -53,6 +53,14 @@ occur:
    same value. If all instances of ``tls`` and ``ssl`` have the same
    value, an error MUST NOT be raised.
 
+SRV URI with directConnection URI option
```
divjotarora: I don't recall why we decided to make SRV + `directConnection=true` an error. Is this because some drivers do initial seedlist discovery after the connection string has already been parsed and validated?

prashantmital: I would think this is because a single SRV record can inherently resolve to any number of resources. In the case it resolves to more than 1 resource, how does the driver determine which of these resoruces it should connect directly to?

source/server-discovery-and-monitoring/tests/sharded/discover_single_mongos.json
```diff
@@ -0,0 +1,28 @@
+{
+  "description": "Discover single mongos",
+  "uri": "mongodb://a/?directConnection=false",
+  "phases": [
+    {
+      "responses": [
+        [
+          "a:27017",
+          {
```
divjotarora: The driver fails this test because there's no min/max wire version in the response so it thinks the max wire version is 0, which is lower than the driver's lowest accepted (2). Looks like other tests you wrote specify 6. Can this be changed to match?

source/uri-options/uri-options.rst
divjotarora: I was thinking of this during the POC because Go does initial seedlist discovery very early on, so we know the exact number of addresses the SRV URI resolves to and could error appropriately. Given that drivers are designed differently and others may not have this information available to them at the right time, I'm fine leaving this as-is. Resolving.

source/server-discovery-and-monitoring/tests/rs/discover_arbiters.json
```diff
@@ -1,6 +1,6 @@
 {
-  "description": "Discover arbiters",
-  "uri": "mongodb://a/?replicaSet=rs",
```
divjotarora: What's the reasoning for changing this from `replicaSet=rs` to `directConnection=false`? They test two different things (replica set name -> start in RSNP topology type, `directConnection=false` -> start in Unknown topology type) so they should either be two separate tests or we can just reset this to the original test because the expectations are still correct.

mbroadst: I agree. The same question applies to a number of similar tests in the suite here.

source/server-discovery-and-monitoring/tests/single/direct_connection_replicaset.yml
```diff
@@ -0,0 +1,23 @@
...
+uri: "mongodb://a/?replicaSet=rs&directConnection=true"
+
+phases:
+  # We are connecting to a replica set member
+  - responses:
+    -
+      - "a:27017"
+      - ok: 1
+        ismaster: true
+        primary: true
```
patrickfreed: This field should actually contain the hostname of the primary, not a boolean. This caused some issues for the Rust driver's type-aware test file deserializer.

p-mongo: I removed this field because it was not used in the "Discover primary with directConnection URI option" test, is this OK?

patrickfreed: Yep lgtm

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -365,18 +365,41 @@ The hostname portion of each address MUST be normalized to lower-case.
...
+MUST behave as if it was specified with the false value.
+
+Since changing the starting topology can reasonably be considered a
+backwards-breaking change, existing drivers SHOULD stage implementation
+according to semantic versioning guidelines. Specifically, support for the
+``directConnection`` URI option can be added in a minor release.
+In a subsequent major release, the default starting topology can be changed
+to Unknown. Drivers MUST document this in a prior minor release.
+
+Existing drivers MUST deprecate other URI options, if any, for controlling
```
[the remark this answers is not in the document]
p-mongo: The language-native options are not deprecated by the changes in this PR, only driver-specific URI options pertaining to topology discovery are. A driver may choose to deprecate its language-native options or convert directconnection uri option to its existing language-native options.

I reworded the paragraph that you quoted to explicitly indicate that language-native options are not covered by the spec, do you agree with this change?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -2366,6 +2406,12 @@ monitoring can also proactively detect state changes
 (primary stepdown, server becoming unavailable)
 that would otherwise cause future errors.
 
+Why is auto-discovery the preferred default?
+''''''''''''''''''''''''''''''''''''''''''''
+
+Auto-discovery is most resilient and is therefore preferred.
```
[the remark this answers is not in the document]
p-mongo: I agree with you but this phrase was the language that was approved during design review, so unless there is in fact some data to back up this statement I prefer to leave existing language unchanged.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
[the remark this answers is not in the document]
p-mongo: Good catch, I added the SRV URI tests.

source/initial-dns-seedlist-discovery/tests/direct-connection-true.yml
```diff
@@ -0,0 +1,5 @@
+uri: "mongodb+srv://test3.test.build.10gen.cc/?directConnection=true"
+seeds: []
+hosts: []
+error: true
+comment: Should fail because directConnection=false is incompatible with SRV URIs.
```
patrickfreed: should say "direcConnection=true" here

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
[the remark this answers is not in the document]
patrickfreed: And what about a uri options test verifying `mongodb://host1,host2,host3/?directConnection=true` fails as well? Also note if we include that we also need to copy the "`directConnection=true` cannot be used with multiple seeds" guidance to that spec also.

source/initial-dns-seedlist-discovery/tests/direct-connection-true.yml
p-mongo: Fixed thanks.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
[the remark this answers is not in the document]
p-mongo: Added this test also, thanks.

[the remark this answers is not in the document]
p-mongo: Added language and a test for dc=false.

source/uri-options/tests/connection-options.yml
```diff
@@ -104,3 +104,46 @@ tests:
...
+      valid: false
+      warning: false
+      hosts: ~
+      auth: ~
+    -
+      description: directConnection=false
+      uri: "mongodb://example.com/?directConnection=false"
+      valid: true
+      warning: false
+      hosts: [example.com]
```
patrickfreed: According to the tests README, this field will [always be null](https://github.com/mongodb/specifications/blob/master/source/uri-options/tests/README.rst#format) and is just included so that the connection string test runner can be reused for these tests. If we do want to keep it around, it needs to match the format of the connection string tests (described [here](https://github.com/mongodb/specifications/tree/master/source/connection-string/tests#format)). I think rather than trying to do that, it's probably simpler to just leave this null, since we're not testing host parsing here anyways.

jyemin: @p can you respond please?

p-mongo: Removed the hosts.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -365,36 +365,60 @@ The hostname portion of each address MUST be normalized to lower-case.
...
+This spec does not cover language-specific native options that a driver may
+provide.
 
 Initial setName
 ~~~~~~~~~~~~~~~
 
-The user MUST be able to set the client's initial replica set name.
+It is allowed to use ``directConnection=true`` in conjunction with the
+``replicaSet`` URI option. The driver must connect in Single topology and
+verify that setName matches the specified name, as per
```
jyemin: Can you point me to the test that verifies that a non-matching replica set name with directConnection=true acts appropriately?  Something like

    mongodb://localhost/?directConnection=true&replicaSet=wrongName

If there isn't one, can it be added?

p-mongo: I added one but looking at the resulting topology, it is not clear that such a topology will fail all operations. We handle this case [during server selection](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/server_selector/selectable.rb#L218).

jyemin:
@p I think everyone planning to review has LGTM'd.

PR #740, opened 2020-02-13
SPEC-1248 SDAM spec test changes for readability

p-mongo:
This PR changes some of the SDAM spec tests to improve reliability. There are no content changes.


p-mongo:
The drivers will have to sync spec tests anyway for spec-1248, hence I believe now is a good time to make the change proposed in this PR.

The change in this PR was originally made in 2018 or 2019 because the tests in specs repo were hard to read.

mbroadst:
I'm inclined to agree with @jyemin here, let's please close this pull request unless there is something being fixed.

p-mongo:
This PR fixes:

- Inconsistent ordering of the fields in some of the spec tests, specifically URI being given after the responses and the outcomes. A person reading the test generally will start with the URI to understand the initial state of the driver.
- Excessive whitespace in some of the tests. This makes the files hard to read because they don't fit on one page.

I would also like to make the equivalent changes to all of the remaining sdam spec test files, for the same reasons.

alcaeus:
Closing since it's unfeasible to make these changes with DRIVERS-1293 changing a number of test files. While I see the benefit in having a (somewhat) consistent ordering of fields in tests, the git noise associated with such changes is difficult to argue against.

PR #741, opened 2020-02-13
Add OCSP URI option

vincentkam:


source/ocsp-support/ocsp-support.rst
```diff
@@ -3,14 +3,14 @@ OCSP Support
 ============
 
 :Spec Title: OCSP Support
-:Spec Version: 1.0.2
+:Spec Version: 1.3.0
```
vincentkam: Adding a URI option seeems worthy a  minor version bump. (1.2.0 was for the cache)

source/ocsp-support/ocsp-support.rst
```diff
@@ -102,9 +102,11 @@ invalid, the driver SHOULD end the connection.
     attempt to validate the status of the unvalidated certificates
     using the cached CRLs.
 
-7.  If the server’s certificate remains unvalidated and that certificate
-    has an OCSP endpoint, the driver SHOULD reach out to the OCSP
-    endpoint specified and attempt to validate that certificate.
+7.  If the server’s certificate remains unvalidated, that certificate
+    has an OCSP endpoint and ``tlsDisableOcspEndpointCheck`` is true (`if the
```
vincentkam: Thanks @saghm for brainstorming with me and helping me nail down this name!

source/ocsp-support/ocsp-support.rst
ShaneHarvey: Should this be 1.1 instead of 1.3? The previous version was 1.0.

source/ocsp-support/ocsp-support.rst
```diff
@@ -134,28 +136,69 @@ Drivers SHOULD validate OCSP Responses in the manner specified in `RFC
...
 
 The following requirements apply only to drivers that are able to
 enable/disable OCSP on a per MongoClient basis.
 
-1. If a connection string specifies `tlsInsecure=true` then the driver
-   MUST disable OCSP.
+1. If a connection string specifies ``tlsInsecure=true`` then the
+   driver MUST disable OCSP.
+
+2. If a connection string contains both ``tlsInsecure`` and
```
vincentkam: We have a similar requirement for `tlsInsecure` and `tlsAllowInvalidCertificates`.

source/ocsp-support/ocsp-support.rst
```diff
@@ -134,28 +136,69 @@ Drivers SHOULD validate OCSP Responses in the manner specified in `RFC
...
+   ``tlsDisableOcspEndpointCheck`` then the driver MUST throw an
+   error.
 
-2. If a driver supports ``tlsAllowInvalidCertificates``, and a
+3. If a driver supports ``tlsAllowInvalidCertificates``, and a
    connection string specifies ``tlsAllowInvalidCertificates=true``,
    then the driver MUST disable OCSP.
 
+4. If a driver supports ``tlsAllowInvalidCertificates``, and a
```
vincentkam: We have a similar requirement for `tlsInsecure` and `tlsAllowInvalidCertificates`.

source/ocsp-support/ocsp-support.rst
ShaneHarvey: "and ``tlsDisableOcspEndpointCheck`` is true" -> "and ``tlsDisableOcspEndpointCheck`` is false"?

source/ocsp-support/ocsp-support.rst
```diff
@@ -134,28 +136,69 @@ Drivers SHOULD validate OCSP Responses in the manner specified in `RFC
...
+   and ``tlsDisableOcspEndpointCheck``, then the driver MUST
+   throw an error.
+
 The remaining requirements in this section apply only to drivers that
 expose an option to enable/disable certificate revocation checking on a
 per MongoClient basis.
 
-1. Driver MUST enable OCSP support (with stapling if possible) when this
-   option is enabled.
+1. Driver MUST enable OCSP support (with stapling if possible) when
```
saghm: I think it's probably a bit more clear to say "when certificate revocation checking is enabled" rather than "when this option is enabled".

source/ocsp-support/ocsp-support.rst
```diff
@@ -134,28 +136,69 @@ Drivers SHOULD validate OCSP Responses in the manner specified in `RFC
...
+   this option is enabled.
 
 2. Drivers SHOULD throw an error if ``tlsInsecure=true`` or
-   ``tlsAllowInvalidCertificates=true`` are specified alongside the
+   ``tlsAllowInvalidCertificates=true`` or
```
saghm: Minor nit: it reads a bit easier to me as "Drivers SHOULD throw an error if any of `tlsInsecure=true`, `tlsAllowInvalidCertificates=true`, or `tlsDisableOcspEndpointCheck=true` is specifed...", but this is pretty subjective, so feel free it ignore if you disagree

source/uri-options/uri-options.rst
```diff
@@ -280,6 +284,11 @@ pertaining to URI options apply here.
      - no
      - Relax TLS constraints as much as possible (e.g. allowing invalid certificates or hostname mismatches); drivers must document the exact constraints which are relaxed by this option being true
 
+   * - tlsDisableOcspEndpointCheck
+     - "true" or "false"
+     - Non-stapled OCSP is enabled
```
saghm: In other places, we've linked directly to the other specs in this table; I think that makes sense here too

source/uri-options/uri-options.rst
```diff
@@ -280,6 +284,11 @@ pertaining to URI options apply here.
...
+     - Yes
+     - Disable non-stapled OCSP
```
saghm: For those not as familiar with the OCSP spec, it might be clearer to spell out exactly what happens here; maybe something like "Certificates will not be checked against OCSP endpoints even if they are not in the cache"

source/uri-options/uri-options.rst
```diff
@@ -3,25 +3,25 @@ URI Options Specification
...
+:Last Modified: 2020-2-13
 
 
 **Abstract**
 ------------
 
 Historically, URI options have been defined in individual specs, and
 drivers have defined any additional options independently of one another.
-Because of the frustration due to there not being a single place where 
+Because of the frustration due to there not being a single place where
```
saghm: In the future, submitting all the unrelated whitespace changes as a separate pull request would make this easier to review :)

source/ocsp-support/ocsp-support.rst
vincentkam: Oops, totally! Thanks for the catch! Fixed!

source/ocsp-support/ocsp-support.rst
vincentkam: Oops sorry, 1.2 may actually be the correct version, since this PR (https://github.com/mongodb/specifications/pull/734) seeks to bump the spec to 1.1.

source/uri-options/uri-options.rst
vincentkam: Ah I agree, sorry! I wanted to get feedback on this before I left for long weekend. 
I can revert the whitespace changes if you'd like

source/ocsp-support/ocsp-support.rst
```diff
@@ -607,6 +657,8 @@ of checking this are:
 
 Changelog
 ==========
+**2020-2-13**: 1.3.0: Add tlsOcsp URI option.
```
p-mongo: Old option name?

vincentkam: Ah thanks for the catch Oleg, fixed!

source/uri-options/uri-options.rst
vincentkam: I've added two links to this section :) One that explains what "if needed" means, and one that links to the more detailed explanation of the URI option

source/uri-options/uri-options.rst
vincentkam: True: I've updated the wording. I've left out mention of the cache, instead going for the more general "if needed", and added a link that explains what "if needed" means

source/uri-options/uri-options.rst
```diff
@@ -280,6 +284,11 @@ pertaining to URI options apply here.
      - no
      - Relax TLS constraints as much as possible (e.g. allowing invalid certificates or hostname mismatches); drivers must document the exact constraints which are relaxed by this option being true
 
+   * - tlsDisableOcspEndpointCheck
+     - "true" or "false"
+     - false i.e. driver will reach out to OCSP endpoints `if needed <../ocsp-support/ocsp-support.rst#id1>`__.
```
vincentkam: I admit this link is slightly fragile: I was unable to get a cross-document `:ref`` to work, and that section doesn't have a named id because there are two sections with the same name at different sublevels

source/ocsp-support/ocsp-support.rst
```diff
@@ -3,14 +3,14 @@ OCSP Support
 ============
 
 :Spec Title: OCSP Support
-:Spec Version: 1.0.2
+:Spec Version: 1.2.0
```
behackett: Either 1.0.3 or 1.1.0, not 1.2.0.

source/ocsp-support/ocsp-support.rst
```diff
@@ -102,8 +102,10 @@ invalid, the driver SHOULD end the connection.
     attempt to validate the status of the unvalidated certificates
     using the cached CRLs.
 
-7.  If the server’s certificate remains unvalidated and that certificate
-    has an OCSP endpoint, the driver SHOULD reach out to the OCSP
+7.  If the server’s certificate remains unvalidated, that certificate
+    has an OCSP endpoint and ``tlsDisableOcspEndpointCheck`` is false
```
behackett: I think Ocsp should be all caps (tlsDisableOCSPEndpointCheck) similar to tlsCAFile.

source/ocsp-support/ocsp-support.rst
```diff
@@ -134,28 +136,69 @@ Drivers SHOULD validate OCSP Responses in the manner specified in `RFC
 6960: 3.2 <https://tools.ietf.org/html/rfc6960#section-3.2>`__ to the
 extent that their TLS library allows.
 
+MongoClient Configuration
+--------------------------
+
+This specification introduces the client-level configuration option
+defined below. Drivers that can, on a per MongoClient basis, disable
+non-stapled OCSP while keeping stapled OCSP enabled MUST implement
```
behackett: Perfect. This is exactly the behavior we want.

source/ocsp-support/ocsp-support.rst
```diff
@@ -134,28 +136,69 @@ Drivers SHOULD validate OCSP Responses in the manner specified in `RFC
...
+this option.
+
+tlsDisableOcspEndpointCheck
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This boolean option determines whether MongoClient should refrain from
+reaching out to an OCSP endpoint i.e.  whether non-stapled OCSP should
+be disabled.  When set to true, a driver MUST NOT reach out to OCSP
+endpoints. When set to false, a driver MUST reach out to xOCSP
```
behackett: I'm guessing the x is a typo?

source/ocsp-support/ocsp-support.rst
behackett: These don't appear in conflict to me, since disabling OCSP endpoint checks also reduces security.

source/ocsp-support/ocsp-support.rst
behackett: Same comment as above, these don't seem in conflict to me.

source/ocsp-support/ocsp-support.rst
```diff
@@ -452,6 +491,17 @@ No. Although `RFC 5019, The Lightweight Online Certificate Status Protocol
...
 responses after ``nextUpdate``, this spec is not adhering to that RFC.
 
+Why was the decision made to allow OCSP to be enabled/disabled via a URI option?
+--------------------------------------------------------------------------------
+We initially hoped that we would be able to not expose any options
+specifically related to OCSP to the user, in accordance with the
+“\ `No Knobs” drivers mantra
+<https://github.com/mongodb/specifications#no-knobs>`__. However, we
+later decided that users may benefit from having the ability to toggle
+OCSP off when applications are deployed behind restrictive firewall
```
behackett: OCSP endpoint checking. Be explicit about that. We're not allowing users to disable all of OCSP.

source/uri-options/uri-options.rst
```diff
@@ -45,11 +45,15 @@ to the security implications of certain options, drivers MUST raise an
...
-1. Both ``tlsInsecure`` and ``tlsAllowInvalidCertificates`` appear in the 
+1. Both ``tlsInsecure`` and ``tlsAllowInvalidCertificates`` appear in the
    URI options.
-2. Both ``tlsInsecure`` and ``tlsAllowInvalidHostnames`` appear in the 
+2. Both ``tlsInsecure`` and ``tlsAllowInvalidHostnames`` appear in the
    URI options.
-3. All instances of ``tls`` and ``ssl`` in the URI options do not have the
+3. Both ``tlsInsecure`` and ``tlsDisableOcspEndpointCheck`` appear in
+   the URI options.
+4. Both ``tlsAllowInvalidCertificates`` and
```
behackett: This doesn't seem right to me. tlsDisableOcspEndpointCheck doesn't disable verifying stapled OCSP responses, but tlsAllowInvalidCertificates does.

source/ocsp-support/ocsp-support.rst
vincentkam: Ah sorry for the confusion: the current version of the spec is at 1.1.0, so this PR should bump it to 1.2. (I haven't rebased since I don't want to screw up force push during a review) https://github.com/vincentkam/specifications/blob/ocsp-support/source/ocsp-support/ocsp-support.rst

source/ocsp-support/ocsp-support.rst
vincentkam: Ah, sorry, I was instinctively following the camelCasing naming convention (which .NET/C# also uses) in which acronyms longer than two letters only have their initial letter capitalized, which can help with readability in the case where multiple acronyms occur together.  The only other multi letter acronym we have in the URI options appears to be "tls", but that occurs at the beginning of the word, although having it be lower case also fits with the aforementioned naming convention. 

Playing devil's advocate, the server team appears to use all caps for their mongod parameter options e.g. CRLFile (https://docs.mongodb.com/manual/tutorial/configure-ssl/)

See also: https://en.wikipedia.org/wiki/Camel_case#Programming_and_coding and https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions

Give the above I'd prefer to keep the name as is, but  if you feel strongly about naming capitalizing OCSP, I'm fine with changing it :)

source/ocsp-support/ocsp-support.rst
vincentkam: Ack, fixed!

source/ocsp-support/ocsp-support.rst
vincentkam: Hm, I don't think the options are conflict, but I was also going off the precedent set by tlsInsecure and tlsAllowInvalidCertificates:  wouldn't setting both tlsInsecure=true and tlsAllowInvalidCertificates=true reduce security? 

My reading of https://github.com/mongodb/specifications/blob/master/source/uri-options/uri-options.rst#conflicting-tls-options is that setting both these options isn't allowed despite the fact that tlsInsecure is a superset of tlsAllowInvalidCertificates (and thus setting them both to true shouldn't conflict). @saghm Any thoughts?

source/ocsp-support/ocsp-support.rst
vincentkam: Ah good call: fixed!

source/ocsp-support/ocsp-support.rst
saghm: I think the (very slight) precedent is for all caps from tlsCAFile like Bernie mentioned, with `tls` being lowercase because I don't think anybody would prefer something like `TLSCAFile` or `tLSCAFile`. That being said, it shouldn't actually matter much since the URI options are required to be case-insensitive. How about we use the case `OCSP` in the spec, but allow languages to use whatever casing their ecosystem prefers?

vincentkam: Ah, I totally forgot that URI options are case insensitive. In which case, I'll just change it to `OCSP`

source/ocsp-support/ocsp-support.rst
saghm: I think it would be consistent with the existing rules for `tlsAllowInvalidCertificates` and `tlsAllowInvalidHostnames` to forbid `tlsDisableOcspEndpointCheck` and `tlsInsecure` from both being present. The consensus we came to when we were writing the URI options spec was that there isn't any utility in specifying `tlsInsecure` along with something like `tlsAllowInvalidCertificates` due to the fact that the latter will be redundant if the two are consistent and confusing regardless of the semantics we pick if they aren't consistent; I think that applies here as well.

source/uri-options/uri-options.rst
saghm: It doesn't matter now; it just would have saved me a minute or to figuring out if there were any changes here I actually had to review :)

source/ocsp-support/ocsp-support.rst
behackett: I relent.

PR #742, opened 2020-02-14
SPEC-1575 Rename MONGODB-IAM to MONGODB-AWS

bazile-clyde:


source/auth/auth.rst
```diff
@@ -709,17 +709,17 @@ mechanism
 mechanism_properties
 	MUST NOT be specified.
 
-MONGODB-IAM
+MONGODB-AWS
```
kevinAlbs: If @jyemin agrees, can we also move this section from underneath "SASL Mechanisms" to underneath the heading "MongoDB Custom Mechanisms"? I believe this is our own custom mechanism (i.e. not part of the SASL mechanisms listed https://www.iana.org/assignments/sasl-mechanisms/sasl-mechanisms.xhtml).

jyemin: In this context though, isn't it a SASL mechanism even if we haven't registered the name?  Since we're using saslStart for it.

kevinAlbs: Hmm, I suppose it is our own custom SASL mechanism, but still a SASL mechanism. Apologies for the noise.

jyemin: No worries.  It's a fair question!

PR #743, opened 2020-02-17
SPEC-1248 Unify behavior around configuration for replica set discovery (spec tests)

p-mongo:
Summary of changes:

- All tests that had a single seed were duplicated into two versions:
  one with directConnection=true and one with directConnection=false
- Added single mongos discovery test
- Added a standalone discovery test
- Renamed foo_discovered.* tests to discover_foo.* for consistency
- All RS tests were duplicated to use directConnection=false option
  and the replicaSet=foo option to discover the RS
- Test descriptions updated to match consistent file naming

### DIFF
--- a/source/server-discovery-and-monitoring/tests/rs/discover_arbiters.yml
+++ b/source/server-discovery-and-monitoring/tests/rs/discover_arbiters.yml
@@ -1,6 +1,6 @@
 description: "Discover arbiters"
 
-uri: "mongodb://a/?replicaSet=rs"
+uri: "mongodb://a/?directConnection=false"
 
 phases: [
 
--- a/source/server-discovery-and-monitoring/tests/rs/discover_hidden.yml
+++ b/source/server-discovery-and-monitoring/tests/rs/discover_hidden.yml
@@ -1,4 +1,4 @@
-description: "Replica set discovery from primary"
+description: "Hidden discovered"
 
 uri: "mongodb://a/?directConnection=false"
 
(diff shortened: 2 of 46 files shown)


p-mongo:
https://jira.mongodb.org/browse/SPEC-1248

jyemin:
Added Divjot and Prashant, as they were RLGTMs on the scope.

source/server-discovery-and-monitoring/tests/rs/discover_arbiters.yml
```diff
@@ -1,6 +1,6 @@
 description: "Discover arbiters"
```
prashantmital: I think this should be renamed to "Discover arbiters with directConnection URI option" just so that the difference between this and the duplicated test is more explicit. I don't think renaming the test file is necessary, but since the description is used to identify the tests in evergreen (at least this is true for PyMongo), I think there is value in making it explicit w.r.t. what is being tested.

source/server-discovery-and-monitoring/tests/rs/discover_hidden.yml
```diff
@@ -0,0 +1,50 @@
+description: "Hidden discovered"
```
prashantmital: Is this a newly added test? What is it testing and why can it not be tested with the replicaSet option as well?

prashantmital: On a closer look, this seems to duplicate the discover RSOther test. It should probably be removed?

source/server-discovery-and-monitoring/tests/rs/discover_arbiters.yml
prashantmital: If folks agree, this would apply to descriptions from all tests that now use the `directConnection` option.

p-mongo: Done.

source/server-discovery-and-monitoring/tests/rs/discover_hidden.yml
p-mongo: The outcomes are the same but server responses are different, because several server responses are treated by the driver as mapping to RSOther state. I added the test with replicaSet option.

p-mongo:
Per the feedback from @jyemin and @divjotarora I added the contents of this PR to https://github.com/mongodb/specifications/pull/739, please continue the review there.

PR #744, opened 2020-02-18
SPEC-1544 AWS Auth support for AWS Lambda

bazile-clyde:


source/auth/auth.rst
```diff
@@ -914,7 +923,9 @@ request. If so, then in addition to a username and password, users MAY also prov
 
    "mongodb://<access_key>:<secret_key>@mongodb.example.com/?authMechanism=MONGODB-AWS&authMechanismProperties=AWS_SESSION_TOKEN:<security_token>"
 |
-If a username and password are not provided, drivers MUST query a link-local AWS address for temporary credentials.
+AWS Lambda runtimes set several `environment variables <https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime>`_ during initialization. To support AWS Lambda runtimes Drivers MUST check a subset of these variables, i.e., ``AWS_ACCESS_KEY_ID``, ``AWS_SECRET_ACCESS_KEY``, and ``AWS_SESSION_TOKEN``, for the access key ID, secret access key and session token, respectively if AWS credentials are not explicitly provided in the URI. The ``AWS_SESSION_TOKEN`` may or may not be set. However, if ``AWS_SESSION_TOKEN`` is set Drivers MUST use its value as the session token.
```
markbenvenuto: We need to declare what happens if only some of the environment variables are defined. I did not specify anywhere. The server follows what I described below.

-----------
If the environment variables ``AWS_ACCESS_KEY_ID``, ``AWS_SECRET_ACCESS_KEY``, and ``AWS_SESSION_TOKEN` are all set, proceed with temporary AWS auth.

If the environment variables ``AWS_ACCESS_KEY_ID``, and ``AWS_SECRET_ACCESS_KEY`` are all set, proceed with regular AWS auth.

Else the driver will ignore the environment variables. For instance, if only one of the environment variables are set, the driver will simply ignore it and not error.

jyemin: If `authMechanism=MONGODB-AWS` is specified, the driver has to error if you can't find credentials.  Are you saying that drivers should proceed to checking ECS endpoint, then EC2 endpoint, in this situation (if one of the env variables is not set)?

kevinAlbs: In addition, if a username/password is omitted but the session token is passed through the URI, should drivers use it or ignore it when using environment variables. E.g. should passing a URI like
`mongodb://localhost?authMechanismProperties=AWS_SESSION_TOKEN:<security_token>`
but only having `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` set use the session token from the URI?

bazile-clyde: I just made a new commit clarifying what do in these cases.

markbenvenuto: Thanks

PR #745, opened 2020-02-18
regenerate uri options tests json files


p-mongo:


source/uri-options/tests/concern-options.json
```diff
@@ -36,15 +36,6 @@
         "w": "arbitraryButStillValid"
       }
     },
-    {
```
mbroadst: It looks to me like this test [exists in YML](https://github.com/mongodb/specifications/blob/master/source/uri-options/tests/concern-options.yml#L41), why is it being deleted here?

source/uri-options/tests/tls-options.json
```diff
@@ -13,17 +13,17 @@
...
-        "uri": "mongodb://example.com/?tlsCertificateKeyFilePassword=hunter2",
-      	"valid": true,
-		"warning": false,
-		"hosts": null,
-		"auth": null,
-		"options": {
-		  "tlsCertificateKeyFilePassword": "hunter2"
-		}
-	},
+    {
```
mbroadst: is this just a white space change?

source/uri-options/tests/concern-options.json
p-mongo: The test you are linking to is a different test which exists in the json.

source/uri-options/tests/tls-options.json
p-mongo: Appears to be this way.

source/uri-options/tests/concern-options.json
mbroadst: I'm commenting on removed JSON from `source/uri-options/tests/concern-options.json`, and I'm linking to `source/uri-options/tests/concern-options.yml`. How is that a different test? And where is that test rendered to JSON?

mbroadst: oof, sorry the tests are so similarly named.

PR #749, opened 2020-02-24
SPEC-1419 Support for collection and index creation in Transactions

alcaeus:
https://jira.mongodb.org/browse/SPEC-1419

This PR adds spec changes to allow for collection and index creation within a transaction. The tests are added to ensure that drivers don't inspect commands or prevent creating collections or indexes in transaction on their own.

The test include four new test runner operations to check if collections/indexes exist/do not exist. Note that these currently can't be used along with a `session` argument as the server does not allow running `listCollections` and `listIndexes` commands in a transaction. The tests thus only assert against the state outside of the transaction.

Along with test runner operations, the new `createCollection` and `dropCollection` operations have been added on the `database` object and `createIndex` to the `collection` object.

A POC for the PHP library can be found in https://github.com/mongodb/mongo-php-library/pull/722.

### DIFF
--- a/source/transactions/tests/README.rst
+++ b/source/transactions/tests/README.rst
@@ -384,6 +384,70 @@ the given session is not pinned to a mongos::
         arguments:
           session: session0
 
+assertCollectionExists
+~~~~~~~~~~~~~~~~~~~~~~
+
+The "assertCollectionExists" operation instructs the test runner to assert that
+the given collection exists in the database::
+
+      - name: assertCollectionExists
+        object: testRunner
+        arguments:
+          database: db
+          collection: test
+
+Use a ``listCollections`` command to check whether the collection exists. Note
+that it is currently not possible to run ``listCollections`` from within a
+transaction.
+
+assertCollectionNotExists
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The "assertCollectionNotExists" operation instructs the test runner to assert
+that the given collection does not exist in the database::
+
+      - name: assertCollectionNotExists
+        object: testRunner
+        arguments:
+          database: db
+          collection: test
+
+Use a ``listCollections`` command to check whether the collection exists. Note
+that it is currently not possible to run ``listCollections`` from within a
+transaction.
+
+assertIndexExists
+~~~~~~~~~~~~~~~~~
+
+The "assertIndexExists" operation instructs the test runner to assert that the
+index with the given name exists on the collection::
+
+      - name: assertIndexExists
+        object: testRunner
+        arguments:
+          database: db
+          collection: test
+          index: t_1
+
+Use a ``listIndexes`` command to check whether the index exists. Note that it is
+currently not possible to run ``listIndexes`` from within a transaction.
+
+assertIndexNotExists
+~~~~~~~~~~~~~~~~~~~~
+
+The "assertIndexNotExists" operation instructs the test runner to assert that
+the index with the given name does not exist on the collection::
+
+      - name: assertIndexNotExists
+        object: testRunner
+        arguments:
+          database: db
+          collection: test
+          index: t_1
+
+Use a ``listIndexes`` command to check whether the index exists. Note that it is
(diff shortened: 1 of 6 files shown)


source/transactions/tests/README.rst
```diff
@@ -384,6 +384,70 @@ the given session is not pinned to a mongos::
...
+          collection: test
+
+Use a ``listCollections`` command to check whether the collection exists. Note
+that it is currently not possible to run ``listCollections`` from within a
+transaction.
+
+assertCollectionNotExists
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The "assertCollectionExists" operation instructs the test runner to assert that
```
jyemin: assertCollectionExists  => assertCollectionNotExists

source/transactions/tests/create-collection.yml
```diff
@@ -0,0 +1,155 @@
...
+            create: *collection_name
+            lsid: session0
+            txnNumber:
+              $numberLong: "1"
+            startTransaction: true
+            autocommit: false
+            writeConcern:
+          command_name: create
+          database_name: *database_name
+      - command_started_event:
```
jyemin: It seems odd to be asserting an expectation due to a command that is run solely to assert something in the test harness.  Perhaps better to use a separate MongoClient to run listCollections and listIndexes, and that way they won't show up in command monitoring events.  I think there is already such a MongoClient lying around that is used for asserting the "outcome" of the operations

source/transactions/transactions.rst
```diff
@@ -1288,7 +1288,7 @@ The following commands are allowed inside transactions:
 
 3.  killCursors
 
-4.  insert
+4.  insert, including into a non-existing collection to implicitly create it
```
jyemin: Nit: "to implicitly create it" => "that implicitly creates it"

source/transactions/transactions.rst
```diff
@@ -1298,12 +1298,18 @@ The following commands are allowed inside transactions:
 
 8.  aggregate (including $lookup)
 
-    -  The ``$out`` and ``$merge`` stages are prohibited.
+    -  The ``$out`` stage is only allowed if the target does not exist. The
```
jyemin: Did you consider a test for this?  It seems wise since some drivers have different code paths for aggregate operations that write via $out/$merge

source/transactions/tests/create-collection.yml
alcaeus: Good point. I've removed the assertions.

source/transactions/transactions.rst
alcaeus: This was originally part of the scope but has since been removed. I removed the tests but forgot this change, so I removed it now.

jyemin:
Doesn't look like you pushed a new commit yet.

alcaeus:
Force of habit: I amended the original commits. Sorry about that.

jyemin:
Yeah, I prefer that the rebases happen after the PR is closed.  Makes it easier to see what changed as a result of the review.

source/transactions/tests/create-collection.yml
```diff
@@ -0,0 +1,131 @@
...
+        arguments:
+          database: *database_name
+          collection: *collection_name
+      - name: commitTransaction
+        object: session0
+      - name: assertCollectionExists
+        object: testRunner
+        arguments:
+          database: *database_name
+          collection: *collection_name
```
jmikola: > Note that these currently can't be used along with a session argument as the server does not allow running listCollections and listIndexes commands in a transaction.

I'm not suggesting you change this, but couldn't we technically use `object: session0` _here_ since the session would no longer be in an active transaction?

alcaeus: Yes, once the collection is done we could use a `session` argument for the check. I'm not sure if there's an advantage to checking within the session, but I can add it if it makes sense to do so.

ShaneHarvey:
Please link the spec ticket in the commit when you merge.

PR #751, opened 2020-02-25
SPEC-1565 Some Retryable Writes specification improvements

egiurleo:
Three small changes:
* Remove the SPEC-1565 comments from the test files; we were originally going to remove the error labels from the failPoint command and have mongos always propagate RetryableWriteError labels, but it looks like the current behavior is here to stay (see discussion in [SERVER-45939](https://jira.mongodb.org/browse/SERVER-45939)).
* Specify that the driver should only add the error label if retryWrite is true.
* Make it clear that mongos will sometimes choose to retry internally, in which case it won't propagate the error label to the driver.

### DIFF
--- a/source/retryable-writes/retryable-writes.rst
+++ b/source/retryable-writes/retryable-writes.rst
@@ -3,14 +3,14 @@ Retryable Writes
 ================
 
 :Spec Title: Retryable Writes
-:Spec Version: 1.5.1
+:Spec Version: 1.5.2
 :Author: Jeremy Mikola
 :Lead: \A. Jesse Jiryu Davis
 :Advisors: Robert Stam, Esha Maharishi, Samantha Ritter, and Kaloian Manassiev
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: 3.6
-:Last Modified: 2020-02-10
+:Last Modified: 2020-02-25
 
 .. contents::
 
@@ -808,6 +808,10 @@ to developers.
 
 Changes
 =======
+
+2020-02-25: State that the driver should only add the RetryableWriteError label
+when retryWrites is on, and make it clear that mongos will sometimes perform
+internal retries and not return the RetryableWriteError label.
 
 2020-02-10: Remove redundant content in Tests section.
 
--- a/source/retryable-writes/retryable-writes.rst
+++ b/source/retryable-writes/retryable-writes.rst
@@ -216,20 +216,22 @@ This label might be added to an error in a variety of ways:
 
 When the driver encounters a network error communicating with any server version
 that supports retryable writes, it MUST add a RetryableWriteError label to that
-error.
+error if the MongoClient performing the operation has the retryWrites
+configuration option set to true.
 
 For server versions 4.4 and newer, MongoDB will add a RetryableWriteError label to
 errors or server responses that it considers retryable before returning them to the
 driver. As new server versions are released, the errors that are labeled with the
 RetryableWriteError label may change. When receiving a command result
 with an error from a 4.4+ server that supports retryable writes, the driver
-MUST NOT add a RetryableWriteError label to that error under any condition.   
-Note: With a WriteConcernError response the top level document or the 
+MUST NOT add a RetryableWriteError label to that error under any condition.
+Note: With a WriteConcernError response the top level document or the
 WriteConcernError document may contain the RetryableWriteError error label.
 
 When receiving a command result with an error from a pre-4.4 server that supports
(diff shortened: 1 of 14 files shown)


source/retryable-writes/retryable-writes.rst
```diff
@@ -212,24 +212,34 @@ Determining Retryable Errors
...
+configuration option set to true.
 
 For server versions 4.4 and newer, MongoDB will add a RetryableWriteError label to
 errors or server responses that it considers retryable before returning them to the
 driver. As new server versions are released, the errors that are labeled with the
 RetryableWriteError label may change. When receiving a command result
 with an error from a 4.4+ server that supports retryable writes, the driver
-MUST NOT add a RetryableWriteError label to that error under any condition.   
-Note: With a WriteConcernError response the top level document or the 
-WriteConcernError document may contain the RetryableWriteError error label.
```
egiurleo: I moved these two lines to the top of this section because I thought they were random/easily lost here.

source/retryable-writes/retryable-writes.rst
```diff
@@ -212,24 +212,34 @@ Determining Retryable Errors
...
+MUST NOT add a RetryableWriteError label to that error under any condition.
+
+During a retryable write operation on a sharded cluster, mongos may retry the
```
rozza: Is there a test case for this scenario?

source/retryable-writes/retryable-writes.rst
```diff
@@ -265,6 +275,11 @@ the following criteria:
 
 - a server response with a write concern error response containing any of the previously listed codes
 
+To understand why the driver should only add the RetryableWriteError label
```
rozza: Is there a test case for this scenario?

source/retryable-writes/retryable-writes.rst
egiurleo: I believe the [errorLabel test cases](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/insertOne-errorLabels.yml) cover this scenario -- they run on sharded clusters and test the case where the server returns an error label and when it doesn't, which is essentially the behavior described here. Does that seem reasonable?

source/retryable-writes/retryable-writes.rst
egiurleo: Added!

source/retryable-writes/retryable-writes.rst
```diff
@@ -212,24 +212,34 @@ Determining Retryable Errors
 
 When connected to a MongoDB instance that supports retryable writes (versions 3.6+),
 the driver MUST treat all errors with the RetryableWriteError label as retryable.
-This label might be added to an error in a variety of ways:
+This error label can be found in the top-level "errorLabels" field of the error.
+With a WriteConcernError response, the top level document or the
```
mbroadst: Are `WriteConcernError` Ruby specific? You might need to loosen the language a bit here, maybe something like "In a server error response with a `writeConcernError` field..."

source/retryable-writes/retryable-writes.rst
mbroadst: I agree, the behavior seems covered by this test.

source/retryable-writes/retryable-writes.rst
```diff
@@ -212,24 +212,34 @@ Determining Retryable Errors
...
-MUST NOT add a RetryableWriteError label to that error under any condition.   
-Note: With a WriteConcernError response the top level document or the 
-WriteConcernError document may contain the RetryableWriteError error label.
+MUST NOT add a RetryableWriteError label to that error under any condition.
+
+During a retryable write operation on a sharded cluster, mongos may retry the
+operation internally, in which case it will not add a RetryableWriteError label to
+any error that occurs after those internal retries to prevent excessive retrying.
+The driver MUST NOT add a RetryableWriteError label to errors it receives from
+a version 4.4+ mongos.
```
mbroadst: I wonder if _this_ behavior isn't covered by our tests though?

source/retryable-writes/retryable-writes.rst
egiurleo: I think you're right about it being Ruby specific. Thanks for flagging that.

source/retryable-writes/retryable-writes.rst
egiurleo: I believe it is. Every file that tests error label behavior is run on a sharded cluster, and all of those tests include a case where the server does not return an error label and the driver is expected to not retry. Do you think there's something more we need to do to test this case?

PR #753, opened 2020-02-26
add spec test for incompatible server becoming compatible

saghm:
https://jira.mongodb.org/browse/SPEC-1586


saghm:
I've updated the PR description. I'm not sure why travis is failing; the issues seem to be files I didn't touch, but I've rebased and pushed again, so hopefully it works this time. And yep, the Rust driver failed the test before the fix and passed after.

PR #755, opened 2020-03-03
SPEC-1584 Clarify absence of 'ns' field in idIndex for listCollections

alcaeus:
https://jira.mongodb.org/browse/SPEC-1584

This adds the note added to the "Enumerate indexes" in #712 to the "Enumerate collections" spec. The `idIndex` option also no longer contains a `ns` field, which wasn't included in the original PR.

### DIFF
--- a/source/enumerate-collections.rst
+++ b/source/enumerate-collections.rst
@@ -12,7 +12,7 @@ Enumerating Collections
 :Status: Draft
 :Type: Standards
 :Server Versions: 1.8-2.7.5, 2.8.0-rc3 and later
-:Last Modified: March 3, 2020
+:Last Modified: March 17, 2020
 :Version: 0.6.0
 
 .. contents::
@@ -302,9 +302,9 @@ result for ``system.namespaces``.
 The returned result for each variant MUST be equivalent, and each collection
 that is returned MUST use the field names ``name`` and ``options``.
 
-In MongoDB 4.4, the ``ns`` field was removed from the index specifications,
-including in the ``idIndex`` option returned from the ``listCollections``
-command.
+In MongoDB 4.4, the ``ns`` field was removed from the index specifications, so
+the index specification included in the ``idIndex`` field of the collection
+information will no longer contain an ``ns`` field.
 
 - For drivers that report those index specifications in the form of documents or
   dictionaries, no special handling is necessary, but any documentation of the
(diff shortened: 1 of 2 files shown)


source/enumerate-collections.rst
```diff
@@ -302,6 +302,22 @@ result for ``system.namespaces``.
...
+- For drivers that report those index specifications in the form of documents or
+  dictionaries, no special handling is necessary, but any documentation of the
+  contents of the documents/dictionaries MUST indicate that the ``ns`` field
+  will no longer be present in MongoDB 4.4+. If the contents of the
+  documents/dictionaries are undocumented, then no special mention of the ``ns``
+  field is necessary.
+- For drivers that report those index specifications in the form of statically
+  defined models, the driver MUST manually populate the ``ns`` field of the
+  models with the appropriate namespace if the server does not report it in the
+  ``listIndexes`` command response. The ``ns`` field is not required to be a
```
jyemin: listIndexes -> listCollections

source/enumerate-collections.rst
```diff
@@ -302,6 +302,22 @@ result for ``system.namespaces``.
 The returned result for each variant MUST be equivalent, and each collection
 that is returned MUST use the field names ``name`` and ``options``.
 
+In MongoDB 4.4, the ``ns`` field was removed from the index specifications,
+including in the ``idIndex`` option returned from the ``listCollections``
```
patrickfreed: `idIndex` isn't a collection option / it isn't part of the returned `options` subdocument, so this should probably be reworded. Maybe replace "including" onwards with something like: "so the index specification included in the `idIndex` field of the collection information will no longer contain an `ns` field". 

This distinction made me realize that this spec is a little out of date, since it seems to indicate that the only things returned from `listCollections` are "name/option" pairs, which of course isn't true given that something like `idIndex` exists. This is just a holdover from when listing collections involved querying the `<db>.system.namespaces` collection. I filed [SPEC-1596](https://jira.mongodb.org/browse/SPEC-1596) to discuss these issues and work towards updating this spec. 

One thing to note is that until that ticket is addressed, the following line _technically_ means that `idIndex` shouldn't be returned from this method at all (though I have a strong feeling most drivers _do_ in fact report these extra fields):
>The returned result for each variant MUST be equivalent, and each collection that is returned MUST use the field names name and options.

jyemin: Can we get this resolved and the PR approved?

alcaeus: I've reworded the sentence as suggested. I can work on rewording the other sections to bring the spec up-to-date in a separate pull request pending outcome of discussion in SPEC-1596.

PR #762, opened 2020-03-06
Change index key name from "key" to "keys"

divjotarora:
This change was made because the read/write concern tests added for the 4.4 cycle also have a `createIndex` operation and they use "keys" as the YAML/JSON key name instead of "key". See https://github.com/mongodb/specifications/blob/master/source/read-write-concern/tests/operation/default-write-concern-3.4.json#L120 for an example.


divjotarora:
Merging despite the CI failure because the failure is actually on the master branch, not my commit.

PR #763, opened 2020-03-09
SPEC-1552 Ban ismaster from being published in APM events

divjotarora:
When the driver sends an isMaster command, the server replies with
ismaster in all lowercase. This commit updates the command monitoring
spec to account for this when redacting isMaster commands with
speculative authentication fields.

### DIFF
--- a/source/command-monitoring/command-monitoring.rst
+++ b/source/command-monitoring/command-monitoring.rst
@@ -250,7 +250,7 @@ value MUST be replaced with an empty BSON document. The list is as follows:
    * - ``copydbgetnonce``
    * - ``copydbsaslstart``
    * - ``copydb``
-   * - ``isMaster`` when ``speculativeAuthenticate`` is present
+   * - ``isMaster`` or ``ismaster`` when ``speculativeAuthenticate`` is present
 
 ---
 API


ShaneHarvey:
> When the driver sends an isMaster command, the server replies with ismaster in all lowercase.

I'm fine with this change since drivers can send `{isMaster:1}` or `{ismaster:1}` but can you explain why  the "ismaster" response field matters?

divjotarora:
@ShaneHarvey We're redacting the command we send to the server because the speculative auth attempt includes the SCRAM payload. I think this means we should redact the server response as well because the server will include the response to the `saslStart` in the `isMaster` response if the speculative attempt is successful.

ShaneHarvey:
I get that we need to redact the isMaster command response but I don't see how the "ismaster" field in the response is related. Is Go looking for the "ismaster" field to determine if the command response originates from an isMaster command?

divjotarora:
No, that was just a gap in my understanding of all of this. It should just be "if the command sent to the server is redacted, the reply should be redacted as well".

divjotarora:
@ShaneHarvey I was thinking about this a little more and realized there's probably an easier way to deal with security-sensitive commands. I've filed https://jira.mongodb.org/browse/SPEC-1552 for this discussion.

ShaneHarvey:
The spec already has this requirement:
 > Some commands and replies will contain sensitive data and in order to not risk the leaking of this data to external sources or logs **their commands AND replies** MUST be redacted from the events

PR #764, opened 2020-03-09
SPEC-1592 Allow hinting the delete command

egiurleo:
[SPEC-1592](https://jira.mongodb.org/browse/SPEC-1592): Allow hinting the delete command

Server version 4.3.4 introduces support for the `hint` option on delete commands as well as findAndModify commands with `remove: true` (i.e. FindOneAndDelete commands). This adds information to the CRUD spec about these options and adds tests for the following cases: 
- hint supported (>= 4.3.4)
- server returns an error if the hint option is provided (>= 3.4 for delete, >= 4.2 for findAndModify)
- client returns an error if the hint option is provided (< 3.4 for delete, < 4.2 for findAndModify).

You can find a working [Ruby POC here](https://github.com/mongodb/mongo-ruby-driver/pull/1775). (It looks like there are a lot of evergreen failures, but we're having some trouble with jruby right now, so they're unrelated to the POC.)


egiurleo:
@jyemin I don't think it was intentional, as it's not mentioned in the PR that added the update tests (#596). I can add clientError/serverError tests for updateOne and updateMany. I also realized that I should probably add a test for deletes within a bulkWrite, so I'll do that as well.

source/crud/crud.rst
```diff
@@ -978,6 +978,17 @@ Basic
      * @see https://docs.mongodb.com/manual/reference/command/delete/
      */
     collation: Optional<Document>;
+
+    /**
+     * The index to use.
```
mbroadst: Considering these are probably copy-pasted from the spec to drivers, I wonder if we can do a little better in our doc string here. For the [find command](https://docs.mongodb.com/manual/reference/command/find/#dbcmd.find), our docs say:

> Optional. Index specification. Specify either the index name as a string or the index key
> pattern. If specified, then the query system will only consider plans using the hinted index.

egiurleo: Sure thing! I'll update all the instances of this option.

egiurleo:
@jyemin I added the update tests, if you wouldn't mind taking another look.

egiurleo:
Right! I'll just do that right now.

egiurleo:
@jyemin I added bulkWrite tests and updated my POC to make sure they work. Let me know what you think.

egiurleo:
@jyemin Good point! That's done now.

PR #766, opened 2020-03-10
SPEC-1576 A monitor's connection handshake satisfies a check and updates the topology

ShaneHarvey:
Jira: https://jira.mongodb.org/browse/SPEC-1576

The motivation for this ticket is faster initial SDAM discovery and faster rediscovery. If a client does not use the initial handshake then it require 2 ismasters to update a server's description.

### DIFF
--- a/source/server-discovery-and-monitoring/server-monitoring.rst
+++ b/source/server-discovery-and-monitoring/server-monitoring.rst
@@ -7,7 +7,7 @@ Server Monitoring
 :Status: Accepted
 :Type: Standards
 :Version: Same as the `Server Discovery And Monitoring`_ spec
-:Last Modified: 2020-02-20
+:Last Modified: 2020-03-09
 
 .. contents::
 
@@ -85,10 +85,6 @@ The client begins monitoring a server when:
 * ... `updateRSWithoutPrimary`_ or `updateRSFromPrimary`_
   discovers new replica set members.
 
-When checking a server, clients MUST NOT include SCRAM mechanism
-negotiation requests with the ``isMaster`` command, as doing so would
-make monitoring checks more expensive for the server.
-
 The following subsections specify how monitoring works,
 first in multi-threaded or asynchronous clients,
 and second in single-threaded clients.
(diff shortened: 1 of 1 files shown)


ShaneHarvey:
Should I add a prose test section to the SDAM tests and add a test for this? Although, I'm not sure what the test would be. Maybe
> Create a client directly connected to one server with `heartbeatFrequencyMS=10000` and immediately run a 'ping' command. Assert that the client issued a single isMaster from the monitor.

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -158,6 +156,10 @@ be able to proceed anyway.
 Clients update the topology from each handshake
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+When a monitor check creates a new connection, it MUST perform the
+`connection handshake`_ and MUST use the handshake's ismaster response to
```
mbroadst: nit: we can just say "MUST use the handshake's response to...", we've already begun the process of removing the term "master" from the server, and since we already call this thing a handshake there's no need to refer to `ismaster` here.

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -158,6 +156,10 @@ be able to proceed anyway.
...
+create the next ServerDescription and update the topology.
```
mbroadst: I wonder if this is somewhat prescriptive, can we just say "use the handshake's response to update the ServerDescription"

ShaneHarvey: I understand and I'd like to rephrase this. For more background I think we do want to mandate that the Monitor's check only runs a single isMaster command in this case. We want to prevent drivers from running an extra isMaster directly after the handshake. In other words, the handshake MUST satisfy the initial check (or any subsequent check that creates a new connection). Perhaps that wording is better?

> When a monitor check creates a new connection, the `connection handshake`_ response MUST be used to satisfy the check and update the topology.

source/server-discovery-and-monitoring/server-monitoring.rst

source/server-discovery-and-monitoring/server-monitoring.rst
mbroadst: great, I like it

PR #771, opened 2020-03-19
Fix invalid JSON in top.json parse error cases

xdg:
Drive by fix.

### DIFF
--- a/source/bson-corpus/tests/top.json
+++ b/source/bson-corpus/tests/top.json
@@ -69,11 +69,11 @@
     "parseErrors": [
         {
             "description" : "Bad $regularExpression (extra field)",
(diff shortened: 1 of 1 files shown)


source/bson-corpus/tests/top.json
```diff
@@ -69,11 +69,11 @@
     "parseErrors": [
         {
             "description" : "Bad $regularExpression (extra field)",
-            "string" : "{\"a\" : \"$regularExpression\": {\"pattern\": \"abc\", \"options\": \"\", \"unrelated\": true}}}"
+            "string" : "{\"a\" : {\"$regularExpression\": {\"pattern\": \"abc\", \"options\": \"\", \"unrelated\": true}}}"
```
mbroadst: _wow_ how did we miss this! I guess test runners are just verifying that _some_ error is thrown?

xdg: That's my guess.  I happened to have a test looking for a more specific error and found this bug.

PR #778, opened 2020-03-25
SPEC-1654 Add limit fields to command started events for delete commands

egiurleo:
There was a bug in the Ruby spec runner, so I didn't catch this omission in the proof of concept.


mbroadst:
@egiurleo was this needed because the ruby driver does strict checking on the command results? I would expect an absence of a field here not to cause a match failure

egiurleo:
The Ruby driver does strict checking, as does Python, I believe (@prashantmital is the one who originally brought this to my attention).

jyemin:
Removed myself

PR #780, opened 2020-03-30
SPEC-1648 Add support for commitQuorum option to index creation

patrickfreed:
[SPEC-1648](https://jira.mongodb.org/browse/SPEC-1648)

This PR adds support for the `commitQuorum` option to the various index creation helpers. Because this option is required, this PR also standardizes the way that options are handled in index management helpers.


source/index-management.rst
```diff
@@ -185,6 +204,44 @@ Standard API
...
+     * This option is sent only if the caller explicitly provides a value. The default is to not send a value.
+     */
+    commitQuorum: Optional<Int32 | String>;
+
+    /**
+     * The maximum amount of time to allow the index build to take before returning an error.
+     *
+     * This option is sent only if the caller explicitly provides a value. The default is to not send a value.
+     */
+    maxTimeMS: Optional<Int64>;
```
patrickfreed: I included this option here since we had a [DRIVERS ticket](https://jira.mongodb.org/browse/DRIVERS-374) ensuring all drivers already support it on the creation / dropping helpers. The ticket didn't ensure the listing helpers had it though, so I didn't include an options definition for those helpers. 

The Rust and Ruby drivers appear to be the only ones that don't actively support this as of yet (we just gotten to it yet in Rust, and Ruby didn't [properly test it](https://jira.mongodb.org/browse/RUBY-1218?focusedCommentId=1595722&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-1595722) with the driver's API when they closed this ticket).

source/index-management.rst
```diff
@@ -572,6 +648,10 @@ Common API Components
     /**
      * Optionally tells the server to build the index in the background and not block
      * other tasks.
+     *
+     * @note Starting in MongoDB 4.2, this option is ignored by the server.
```
patrickfreed: as part of the work that led to `commitQuorum`, `background` became a no-op.

source/index-management.rst
```diff
@@ -689,25 +769,33 @@ Common API Components
...
+
+  However, it might be that a driver needs to expose write concern to a user per operation for various reasons. It is permitted to allow a write concern option, but since writeConcern is a top-level command option, it MUST NOT be specified as part of an ``IndexModel`` passed into the helper. It SHOULD be specified via the options parameter of the helper, if there is one. For example, it would be ambiguous to specify write concern for one or more models passed to ``createIndexes()``, but it would not be to specify it via the ``CreateIndexesOptions``. If a helper does not have options defined for it in this specification, an appropriate options type SHOULD be defined for it, and write concern SHOULD be specified through those options. This allows for compatibility with any future options that may be required to be accepted by those helpers.
+
```
patrickfreed: This option was a little confusing to me at first, so I added a few questions that I wanted the answers to when I was thinking about how to document this option.

source/index-management.rst
```diff
@@ -689,25 +785,36 @@ Common API Components
...
+  Starting in MongoDB 4.4, secondaries build indexes simultaneously with the primary, and after starting an index build, the primary will wait for a certain number of data-bearing nodes, including itself, to have completed the build before it commits the index. ``commitQuorum`` configures this node requirement. Once the index is committed, all the secondaries replicate the commit too. If a secondary had already completed the index build, the commit will be quick, and no new replication lag would be introduced. If a secondary had not finished building the index before the primary committed it (e.g. if ``commitQuorum: "all"`` was not used), then that secondary may lag behind the primary while it finishes building and committing the index.
+
+  The server-default value for ``commitQuorum`` is "all", which means the primary will wait for all voting data-bearing nodes to complete building the index before it commits it.
```
patrickfreed: This isn't actually true yet, but it will be true when 4.4 GA's. (see [SERVER-46823](https://jira.mongodb.org/browse/SERVER-46823))

source/index-management.rst
```diff
@@ -71,7 +71,18 @@ All drivers MUST offer at least one of the sections of operations, the Standard
 Operation Parameters
 --------------------
 
-All drivers MUST include the specified parameters in each operation, with the exception of the options parameter which is OPTIONAL. As of 3.4 (see https://jira.mongodb.org/browse/SERVER-769) the server validates options passed to the createIndexes command -- drivers should be aware when testing that passing arbitrary options when the driver does not validate them could fail on the server.
+All drivers MUST include the specified parameters in each operation. This does not preclude a driver from offering more. A driver SHOULD NOT require a user to specify the options parameter if they wish to use the server defaults.
```
mbroadst, line 74: Let's work really hard to avoid `SHOULD` ever sneaking into our specs again. Do we have to include this caveat if the API notes that a given property is `Optional`? I'm guessing this came up due to [SERVER-47193](https://jira.mongodb.org/browse/SERVER-47193)?

source/index-management.rst
```diff
@@ -71,7 +71,18 @@ All drivers MUST offer at least one of the sections of operations, the Standard
...
+
+As of 3.4 (see https://jira.mongodb.org/browse/SERVER-769) the server validates options passed to the createIndexes command -- drivers should be aware when testing that passing arbitrary options when the driver does not validate them could fail on the server.
```
mbroadst, line 76: nit: add backticks around `createIndexes` since its a constant (I have a mild preference for making these specifications deeply interlinked, or at least making things like this visually distinct)

source/index-management.rst
```diff
@@ -71,7 +71,18 @@ All drivers MUST offer at least one of the sections of operations, the Standard
...
+
+Deviations
+**********
+
+A non-exhaustive list of acceptable deviations are as follows:
+
+* Using named parameters in place of an options hash or class. For instance, ``collection.createIndex({x: 1}, commitQuorum: "all")``.
```
mbroadst, line 83: I think you might be better off using python as an example here, this looks a little like a screwed up javascript call to me (even though I think its swift?). How about: `collection.create_index({ x: 1 }, commit_quorum="all")`?

source/index-management.rst
```diff
@@ -71,7 +71,18 @@ All drivers MUST offer at least one of the sections of operations, the Standard
...
+
+* When using an ``Options`` class, if multiple ``Options`` classes are structurally equatable, it is permissible to consolidate them into one with a clear name. For instance, it would be permissible to use the name ``CreateIndexOptions`` as the options for ``createIndex`` and ``createIndexes``.
```
mbroadst, line 85: I feel like we're getting overly prescriptive here, why do we need to clarify this?

source/index-management.rst
```diff
@@ -185,6 +204,44 @@ Standard API
...
 
+  interface CreateIndexOptions {
+    /**
+     * Specifies how many data-bearing members of a replica set, including the primary, must
+     * complete the index builds successfully before the primary marks the indexes as ready.
+     *
+     * This option accepts the same values for the "w" field in a write concern plus "all",
+     * which indicates all voting data-bearing nodes.
+     *
+     * @note This option is only available in MongoDB 4.4+.
```
mbroadst, line 223: nit: tags like this are usually put at the bottom of the docblock

source/index-management.rst
```diff
@@ -185,6 +204,44 @@ Standard API
...
+     *
+     * Drivers MUST NOT throw an error if this option is specified when creating an index on
+     * an earlier version of the server. Instead, drivers MUST rely on the server for returning
+     * an appropriate error.
+     *
+     * This option is sent only if the caller explicitly provides a value. The default is to not send a value.
```
mbroadst, line 229: feels like a `@note` as well

source/index-management.rst
mbroadst: also I think a better tag to use would be [@since](https://jsdoc.app/tags-since.html)

source/index-management.rst
```diff
@@ -185,6 +204,44 @@ Standard API
...
+
+    /**
+     * The maximum amount of time to allow the index build to take before returning an error.
+     *
+     * This option is sent only if the caller explicitly provides a value. The default is to not send a value.
+     */
+    maxTimeMS: Optional<Int64>;
+  }
+
+  interface CreateIndexesOptions {
```
mbroadst, line 241: If there's no distinction at all, lets only document one type `CreateIndexOptions` and use it in both places, rather than describe combining them as an acceptable deviation.

source/index-management.rst
```diff
@@ -185,6 +204,44 @@ Standard API
...
+     * an appropriate error.
+     *
+     * This option is sent only if the caller explicitly provides a value. The default is to not send a value.
+     */
+    commitQuorum: Optional<Int32 | String>;
+
+    /**
+     * The maximum amount of time to allow the index build to take before returning an error.
+     *
+     * This option is sent only if the caller explicitly provides a value. The default is to not send a value.
```
mbroadst, line 236: sounds like a `@note`

source/index-management.rst
```diff
@@ -185,6 +204,44 @@ Standard API
...
+
+  interface CreateIndexesOptions {
+    // same as CreateIndexOptions
+  }
+
+  interface DropIndexOptions {
+   /**
+     * The maximum amount of time to allow the index drop to take before returning an error.
+     *
+     * This option is sent only if the caller explicitly provides a value. The default is to not send a value.
```
mbroadst, line 249: sounds like a `@note`

source/index-management.rst
```diff
@@ -405,20 +466,35 @@ Index View API
...
-    dropOne(model: IndexModel): Result;
+    dropOne(model: IndexModel, options: Optional<DropOneIndexOptions>): Result;
 
     /**
      * Drops all indexes in the collection.
      */
     dropAll(): Result;
   }
 
+  interface CreateOneIndexOptions {
```
mbroadst, line 499: I'd say we can reuse the common options type here as well

source/index-management.rst
```diff
@@ -689,25 +769,33 @@ Common API Components
...
+
+Q: What does the commitQuorum option do?
+  Prior to MongoDB 4.4, secondaries would simply replicate index builds once they were completed on the primary. Building indexes requires an exclusive lock on the collection being indexed, so the secondaries would be blocked from replicating all other operations while the index build took place. This would introduce replication lag correlated to however long the index build took.
+
+  Starting in MongoDB 4.4, secondaries build indexes simultaneously with the primary, and after starting an index build, the primary will wait for a certain number of data-bearing nodes, including itself, to have completed the build before it commits the index. ``commitQuorum`` configures this node requirement. Once the index is committed, all the secondaries replicate the commit too. If a secondary had already completed the index build, the commit will be quick, and no new replication lag would be introduced. If a secondary had not finished building the index before the primary committed it (e.g. if ``commitQuorum: "all"`` was not used), then that secondary may lag behind the primary while it finishes building and committing the index.
```
mbroadst, line 803: Nice description, very helpful.

source/index-management.rst
patrickfreed: done and done

source/index-management.rst
patrickfreed: fixed

source/index-management.rst
patrickfreed: I couldn't both backtick and link due to how links are specified in rst, so I just backticked it. 

As an aside, would there be any support behind a proposal to allow new specifications to be written in markdown? It seems like the only place rst is common is in the python ecosystem and the mongodb driver specifications.

source/index-management.rst
patrickfreed: See discussion above about this

source/index-management.rst
mbroadst: :+1:

I think we'd need a DRIVERS ticket to discuss that. IIRC the preference for RST here was for the table of contents, which Markdown doesn't make for you automatically iirc.

source/index-management.rst
mbroadst: why not? Maybe it should be a `@note` over there too

patrickfreed: fixed

source/index-management.rst
patrickfreed: fixed

source/index-management.rst
patrickfreed: fixed

source/index-management.rst
```diff
@@ -185,6 +212,48 @@ Standard API
...
 
+  interface CreateIndexOptions {
+    /**
+     * Specifies how many data-bearing members of a replica set, including the primary, must
+     * complete the index builds successfully before the primary marks the indexes as ready.
+     *
+     * This option accepts the same values for the "w" field in a write concern plus "all",
+     * which indicates all voting data-bearing nodes.
+     *
+     * Drivers MUST NOT throw an error if this option is specified when creating an index on
```
divjotarora, line 223: If we're making an effort to keep things consistent with the CRUD spec, can this be changed to use wording from there as well? For example, for the `hint` option on `Delete` helpers, the CRUD spec says:
```
This option is only supported by servers >= 4.4. Older servers >= 3.4 will report an error for using this option.
```
If you prefer the wording here instead, let's file a follow-on DRIVERS ticket to make CRUD consistent?

source/index-management.rst
```diff
@@ -405,20 +478,39 @@ Index View API
      * @note Drivers MAY opt to implement this method signature, the signature that
      *   takes an IndexModel as a parameter, or for those languages with method
      *   overloading MAY decide to implement both.
+     *
+     * @note Drivers MAY combine the two options types into a single one. If the options are
+     *   explicitly typed, the combined options type MUST be named DropOneIndexOptions or an acceptable
+     *   variation.
      */
-    dropOne(keys: Document, options: IndexOptions): Result;
+    dropOne(keys: Document, indexOptions: IndexOptions, options: Optional<DropOneIndexOptions>): Result;
```
divjotarora, line 486: Do we know how many drivers implement this method or the next one? Go currently implements `dropOne(name, *DropIndexesOptions)` and it would be tough to add either of these two given no overloading.

source/index-management.rst
```diff
@@ -689,25 +785,36 @@ Common API Components
...
+
+  The server-default value for ``commitQuorum`` is "all", which means the primary will wait for all voting data-bearing nodes to complete building the index before it commits it.
+
+Q: Why would a user want to specify a non-default ``commitQuorum``?
+  Like ``w: "majority"``, ``commitQuorum: "all"`` doesn't include non-voting data-bearing nodes such as analytics nodes. If a user wanted to ensure these nodes didn't lag behind, then they would specify ``commitQuorum: <number of voting nodes + n>``.
```
divjotarora, line 808: What is `n`? I'm assuming number of analytics nodes? If so, I think it's clearer to say `commitQuorum: <total number of nodes, including non-voting nodes>`.

source/index-management.rst
divjotarora: For context, I'm asking because the spec says drivers can implement the version that takes an `IndexModel`, which Go does, but also says "This is OPTIONAL until partial indexes are implemented", which already exist (https://docs.mongodb.com/manual/core/index-partial/) so it seems like something is out of date here.

source/index-management.rst
patrickfreed: So I included the language about not manually throwing an error in accordance with the [mantra](https://github.com/mongodb/specifications#where-possible-depend-on-server-to-return-errors), but maybe I should relax that to only apply to 3.4+? Ideally, we could give the mantras some teeth by actually enforcing them with "MUST", but I'm not sure if it would be frustrating for users to get different types of errors for the same thing on different server versions. Given this is somewhat of an obscure option, I don't think it'll matter in this case, but it may set a precedent that could be used on a more highly used option.

Alternatively, it could just be exactly consistent with the CRUD spec.

divjotarora: What does a 3.2 server do?

patrickfreed: Doesn't look like we have 3.2 binaries for Ubuntu 18.04, so I can't verify locally, but I imagine it just silently ignores it.

divjotarora: Tested on 3.2.22 and 3.4.24 locally. 3.2 does silently ignore the option. This is unfortunate, but I do think we should stay consistent with the CRUD spec here and say something like:

```
   * This option is only supported by servers >= 4.4. Older servers >= 3.4 will report an error for using this option.
   * For servers < 3.4, the driver MUST raise an error if the caller explicitly provides a value.
```

It will cause different types of errors depending on the operation, but I think that's better than specifying `commitQuorum` and having it silently do nothing.

patrickfreed: Since we're doing server version checks, I think we might want to manually through errors for the buggy servers (versions 4.2-4.2.5). Apparently, setting this option could [break things silently](https://jira.mongodb.org/browse/SERVER-47193?focusedCommentId=3019605&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-3019605). The only problem with that is I think drivers only verify server version via maxWireVersion, so that would be we'd have to manually throw errors for all of 4.2. At that point, it might just be easiest to say that a driver should throw an error if server version is < 4.4...

patrickfreed: Given that there's an actual danger to users, I would say I'm now leaning towards requiring manual validation for this option all the time, and that we can rely on the server for validation of all other options.

mbroadst: @patrickfreed let's not do any client side validation for this, the danger to users is relatively low and will be documented in the FCV documentation. Also, now that the server is patching the issue for 4.2, it will likely be ported to 4.4's FCV as well. I'm going to approve this now, let's move on to implementation.

patrickfreed: After offline discussion, we decided to require manual driver validation, since specifying the option could result in problems for users on 4.2.0-4.2.5. Updated the spec to reflect that.

source/index-management.rst
```diff
@@ -185,6 +212,47 @@ Standard API
...
 
+  interface CreateIndexOptions {
+    /**
+     * Specifies how many data-bearing members of a replica set, including the primary, must
+     * complete the index builds successfully before the primary marks the indexes as ready.
+     *
+     * This option accepts the same values for the "w" field in a write concern plus "votingMembers",
+     * which indicates all voting data-bearing nodes.
+     *
+     * This option is only supported by servers >= 4.4. Drivers MUST manually raise an error if this option
```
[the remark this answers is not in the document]
patrickfreed: There's a question about it in the Q&A, is that sufficient or should I add a note here too?

PR #781, opened 2020-04-01
SPEC-1663 Reduce race conditions in SDAM error handling

ShaneHarvey:
This PR updates the SDAM spec to reduce race conditions in SDAM error handling as part of [SPEC-1579](https://jira.mongodb.org/browse/SPEC-1579). I spun this off as a separate PR because these changes are self contained and do not depend on the streaming protocol.  In summary:

- Add topologyVersion field to ServerDescription.
- Add SDAM spec tests for topologyVersion comparison.
- Clients ignore stale errors based on generation number and topologyVersion.

### DIFF
--- a/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
+++ b/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
@@ -9,7 +9,7 @@ Server Discovery And Monitoring
 :Status: Accepted
 :Type: Standards
 :Version: 2.16
-:Last Modified: 2020-02-20
+:Last Modified: 2020-03-31
 
 .. contents::
 
@@ -214,6 +214,54 @@ minHeartbeatFrequencyMS
 Defined in the `Server Monitoring spec`_. This value MUST be 500 ms, and
 it MUST NOT be configurable.
 
+pool generation number
+``````````````````````
+
+The pool's generation number which starts at 0 and is incremented each time
+the pool is cleared. Defined in the `Connection Monitoring and Pooling spec`_.
+
+connection generation number
+````````````````````````````
+
+The pool's generation number at the time this connection was created.
+Defined in the `Connection Monitoring and Pooling spec`_.
+
+error generation number
+```````````````````````
+
+The error's generation number is the generation of the connection on which the
+application error occured. Note that when a network error occurs before the
+handshake completes then the error's generation number is the generation of
+the pool at the time the connection attempt was started.
+
+State Change Error
+``````````````````
+
+A server reply document indicating a "not master" or "node is recovering"
+error. One of the following errors:
+
+.. list-table::
+  :header-rows: 1
+
+  * - Error Name
+    - Error Code
+  * - NotMaster
+    - 10107
+  * - NotMasterNoSlaveOk
+    - 13435
+  * - InterruptedAtShutdown
+    - 11600
+  * - InterruptedDueToReplStateChange
+    - 11602
+  * - NotMasterOrSecondary
+    - 13436
+  * - PrimarySteppedDown
+    - 189
+  * - ShutdownInProgress
+    - 91
+
+Starting in MongoDB 4.4 these errors may also include a toplogyVersion field.
+
 Data structures
 '''''''''''''''
 
@@ -332,6 +380,9 @@ Fields:
   Default null.
 * lastUpdateTime: when this server was last checked. Default "infinity ago".
 * (=) logicalSessionTimeoutMinutes: integer or null. Default null.
+* (=) topologyVersion: A topologyVersion or null. Default null.
+  The "topologyVersion" from the server's most recent ismaster or State Change
+  Error response.
 
 "Passives" are priority-zero replica set members that cannot become primary.
 The client treats them precisely the same as other members.
@@ -514,11 +565,20 @@ Parsing an ismaster response
 ''''''''''''''''''''''''''''
 
 The client represents its view of each server with a `ServerDescription`_.
-Each time the client `checks`_ a server,
-it MUST replace its description of that server with a new one.
-This replacement MUST happen even if the new server description compares
-equal to the previous one, in order to keep client-tracked attributes
-like last update time and round trip time up to date.
+Each time the client `checks`_ a server, it MUST replace its description of
+that server with a new one if and only if one or more of the following
+conditions are true:
+
+#. The current ServerDescription's topologyVersion is null.
+#. The new ServerDescription's topologyVersion is null.
+#. The current ServerDescription's topologyVersion is less or equal to the
+   new ServerDescription's topologyVersion.
+
+(See `Replacing the TopologyDescription`_ for an example implementation.)
+
+This replacement MUST happen even if the new server description compares equal
+to the previous one, in order to keep client-tracked attributes like last
+update time and round trip time up to date.
 
 ServerDescriptions are created from ismaster outcomes as follows:
 
@@ -669,6 +729,43 @@ MongoDB 3.6 and later include a ``logicalSessionTimeoutMinutes`` field if
 logical sessions are enabled in the deployment. Clients MUST check for this
 field and set the ServerDescription's logicalSessionTimeoutMinutes field to this
 value, or to null otherwise.
+
+topologyVersion
+```````````````
+
+MongoDB 4.4 and later include a ``topologyVersion`` field in all isMaster
+and State Change Error responses. Clients MUST check for this field in both
+"ok: 1" and "ok: 0" replies and set the ServerDescription's topologyVersion
+field to this value, or to null otherwise. The topologyVersion helps the
+client and server determine the relative freshness of topology information
+in concurrent messages. (See `What is the purpose of topologyVersion?`_)
+
+The topologyVersion is a subdocument with two fields, "processId" and
+"counter":
+
+.. code:: typescript
+
+    {
+        topologyVersion: {processId: <ObjectId>, counter: <int64>},
+        ( ... other fields ...)
+    }
+
+topologyVersion Comparison
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To compare a topologyVersion from an isMaster or State Change Error
+response to the current ServerDescription's topologyVersion:
+
+#. If the response topologyVersion is unset or the ServerDescription's
+   topologyVersion is null, the client MUST assume the response is more recent.
+#. If the response's topologyVersion.processId is not equal to the
+   ServerDescription's, the client MUST assume the response is more recent.
+#. If the response's topologyVersion.processId is equal to the
+   ServerDescription's, the client MUST use the counter field to determine
+   which topologyVersion is more recent.
+
+See `Replacing the TopologyDescription`_ for an example implementation of
+topologyVersion comparison.
 
 Other ServerDescription fields
 ``````````````````````````````
@@ -1149,7 +1246,7 @@ we distinguish two phases of connecting to a server and using it for application
 
 If there is a network error or timeout on the connection before the handshake completes,
 the client MUST replace the server's description
-with a default ServerDescription of type Unknown,
+with a default ServerDescription of type Unknown, with null topologyVersion,
 and fill the ServerDescription's error field with useful information.
 
 If there is a network timeout on the connection after the handshake completes,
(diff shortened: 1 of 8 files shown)


source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1149,7 +1246,7 @@ we distinguish two phases of connecting to a server and using it for application
 
 If there is a network error or timeout on the connection before the handshake completes,
 the client MUST replace the server's description
-with a default ServerDescription of type Unknown,
+with a default ServerDescription of type Unknown, with null topologyVersion,
```
mbroadst, line 1249: I think the "null topologyVersion" is implied by replacing the server description with a default `ServerDescription`?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1305,6 +1407,20 @@ recovering" error::
         if is_shutdown(code) or (error was from <4.2):
             clear connection pool for server
 
+    def shouldHandleError(td, response):
```
mbroadst, line 1410: nit: the pseudocode is inconsistent about snake_case. Also, I would expand the variables below, for example: `td` => `topology_description`, I think it pays to be verbose in this specification.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1305,6 +1407,20 @@ recovering" error::
...
+        generation = get connection generation from response
+        if generation < currentGeneration:
+            # Stale generation number.
+            return False
+
+        currentTv = currentServer.topologyVersion
+        if currentTv is null or "topologyVersion" not in response:
+            return True
+        # True if the current server's topologyDescription is < error’s
+        return compareTopologyVersion(currentTv, response["topologyVersion"]) < 0
```
mbroadst, line 1422: `compareTopologyVersion` is no longer defined anywhere, should we reintroduce the pseudo code for this?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1272,26 +1369,31 @@ recovering" error::
...
-                handle_notmaster_or_recovering(wce["errmsg"], wce["code"])
+                handle_notmaster_or_recovering(response, wce["errmsg"], wce["code"])
 
     def parse_query_response(response):
         if the "QueryFailure" bit is set in response flags:
             if is_notmaster_or_recovering(response["$err"], response["code"]):
-                handle_notmaster_or_recovering(response["$err"], response["code"])
+                handle_notmaster_or_recovering(response, response["$err"], response["code"])
+
+    def handle_notmaster_or_recovering(response, message, code):
```
mbroadst, line 1389: is there an extra helper here now called `handle_state_change_error`?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1921,6 +2089,73 @@ shard:
...
+   marks the server Unknown, and clears the server's pool.
+#. The client re-checks the server and marks it Primary.
+#. Thread B receives the second non-timeout network error and the client
+   marks the server Unknown again.
+
+The core issue is that the client processes errors in arbitrary order
+and may overwrite fresh information about the server's status with stale
+information. Using CMAP's generation number avoids the race condition because
+the duplicate (or stale) network error can be identified (changes in
+**bold**). This change applies to both the old protocol and new streaming
```
mbroadst, line 2110: will we continue to have an "old protocol" and a "streaming protocol", or will we just have a "protocol"?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1921,6 +2089,73 @@ shard:
...
+#. The client re-checks the server and marks it Primary.
+#. Thread B receives the second non-timeout network error, **and the
+   client ignores the error because the error originated from a
+   connection with generation 1.**
+
+What is the purpose of topologyVersion?
+'''''''''''''''''''''''''''''''''''''''
+
+topologyVersion solves the following race condition among application threads
+and the monitor when handling State Change Errors:
```
mbroadst, line 2128: nit: link to State Change Error description

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -332,6 +380,9 @@ Fields:
   Default null.
 * lastUpdateTime: when this server was last checked. Default "infinity ago".
 * (=) logicalSessionTimeoutMinutes: integer or null. Default null.
+* (=) topologyVersion: A topologyVersion or null. Default null.
+  The "topologyVersion" from the server's most recent ismaster or State Change
```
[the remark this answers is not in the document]
jyemin: In the definition above `state change error` is defined as `A server reply document`.  So maybe this should just be `state change error` or `state change error reply`

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -514,11 +565,20 @@ Parsing an ismaster response
...
-This replacement MUST happen even if the new server description compares
-equal to the previous one, in order to keep client-tracked attributes
-like last update time and round trip time up to date.
+Each time the client `checks`_ a server, it MUST replace its description of
+that server with a new one if and only if one or more of the following
+conditions are true:
+
+#. The current ServerDescription's topologyVersion is null.
+#. The new ServerDescription's topologyVersion is null.
+#. The current ServerDescription's topologyVersion is less or equal to the
```
jyemin, line 574: Nit: `less than`?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -670,6 +730,43 @@ logical sessions are enabled in the deployment. Clients MUST check for this
...
+
+    {
+        topologyVersion: {processId: <ObjectId>, counter: <int64>},
+        ( ... other fields ...)
+    }
+
+topologyVersion Comparison
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To compare a topologyVersion from an isMaster or State Change Error
```
jyemin, line 756: I know we discussed this in the google doc...

Comparing this algorithm with the one in the `topologyVersion Comparison` section above, they seem identical, except this one defines how to compare two non-null topology versions in more detail.  Can we DRY this up?  As an implementor I'm not clear how to proceed and where I would use the preceding definition of the comparison logic and where I'd use this one.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -670,6 +730,43 @@ logical sessions are enabled in the deployment. Clients MUST check for this
...
+
+#. If the response topologyVersion is unset or the ServerDescription's
+   topologyVersion is null, the client MUST assume the response is more recent.
+#. If the response's topologyVersion.processId is not equal to the
+   ServerDescription's, the client MUST assume the response is more recent.
+#. If the response's topologyVersion.processId is equal to the
+   ServerDescription's, the client MUST use the counter field to determine
+   which topologyVersion is more recent.
+
+See `Replacing the TopologyDescription`_ for an example implementation of
```
jyemin, line 767: And they both refer to the same example implementation.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1349,6 +1465,26 @@ clear the server's connection pool if and only if the error is
...
+Clients MUST NOT handle errors that have stale generation numbers. That is,
+when handling a network or command error, if the error's generation number
+is not equal to the current pool's generation number then the error MUST be
+ignored.
+
+Clients MUST NOT handle errors that have stale toplogyVersions. That is,
+a client MUST NOT handle a command error unless one or more of the following
+conditions are true:
+
+#. The current ServerDescription's topologyVersion is null.
```
jyemin, line 1480: Here's a third place where the comparison algorithm is defined.  I think this needs to be DRY'd up

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1349,6 +1465,26 @@ clear the server's connection pool if and only if the error is
 error messages to detect "not master" and "node is recovering"`_, and `other
 transient errors`_ and `Why close connections when a node is shutting down?`_.)
 
+Reducing race conditions in error handling
```
jyemin, line 1468: I think this section needs to be defined above the pseudocode.  Maybe it should just be rolled into the https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#not-master-and-node-is-recovering section, in the prose before the pseudocode?

jyemin: Also, it doesn't seem like we have any way to test this.  Is that correct?

ShaneHarvey: I moved the generation number rule (which is required on both "network errors" and State Change Errors) to a new section above called "Application Errors".  I moved the topologyVersion rule to the not-master-and-node-is-recovering section.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1490,6 +1599,12 @@ Once the client has taken the lock it must do no I/O::
 
         newTopologyDescription = client.topologyDescription.copy()
 
+        # Ignore this update if the current topologyVersion is greater than
+        # the new ServerDescription's.
+        if not shouldHandleServerDescription(td, server):
```
mbroadst, line 1604: nit: it's easy to overlook the `not` here which is a python keyword, perhaps we should change the method name to something like `shouldDiscardServerUpdate` which better matches what we use the method for.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1502,6 +1617,33 @@ Once the client has taken the lock it must do no I/O::
...
+                return 0
+            elif counter1 < counter2:
+                return -1
+            else:
+                return 1
+        else:
+            # Assume greater.
+            return -1
+
+    def shouldHandleServerDescription(topologyDescription, server):
```
mbroadst, line 1640: I ran into some minor issues implementing this that I think could be corrected by being more explicit in our specification. Specifically, if `topologyDescription` or `server` in this case are null, then you are accessing invalid properties on those types below. In JavaScript we can be a little clever:
```
function shouldHandleServerDescription(previousServerDescription, incomingServerDescription) {
  const previousTopologyVersion =
    previousServerDescription && previousServerDescription.topologyVersion;
  const incomingTopologyVersion =
    incomingServerDescription && incomingServerDescription.topologyVersion;

  return compareTopologyVersion(previousTopologyVersion, incomingTopologyVersion) <= 0;
}
```
But I almost prefer being insanely verbose in documentation that all drivers will have to follow:
 - If (topologyDescription == null || server == null) return true
 - If you can't assume that a server is valid before you pass it in, then check that `topologyDescription.hasServer(server.address)` before you use `currentServer`
 - check that `currentServer` has a `topologyVersion` before using that

Feels like a nitpick, but I think it pays to be verbose in these situations.

ShaneHarvey: How could topologyDescription or `incomingServerDescription` ever be null? Can you point me to where the spec says that could be the case?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
ShaneHarvey: I renamed shouldHandleServerDescription to isStaleServerDescription and shouldHandleError to isStaleError.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
mbroadst: oops, I had placed my logic before the pseudocode to verify that the incoming description was a known server. Disregard.

ShaneHarvey:
@jyemin, are you still planning to implement the tests?

jyemin:
Yes, but you shouldn't wait for me if PyMongo and Node are both passing them.

PR #787, opened 2020-04-16
SPEC-1663 Add SDAM tests for detecting stale application errors

ShaneHarvey:
Jira: [SPEC-1663](https://jira.mongodb.org/browse/SPEC-1663)

This change adds test coverage for detecting stale application errors via SDAM spec tests. Notably, this change:
- Requires drivers to mock application errors configured through the new `applicationErrors` field.
- Adds `pool` to the `outcome` assertion so that drivers can assert a pool's generation.
- Adds tests for SDAM handling of stale application errors based on `generation` and `topologyVersion`.


source/server-discovery-and-monitoring/tests/README.rst
```diff
@@ -40,6 +42,13 @@ A response is a pair of values:
   The empty response `{}` indicates a network error
   when attempting to call "ismaster".
 
+An "applicationError" object has the following keys:
+
+- address: The source address, for example "a:27017".
+- generation: The error's generation number, for example `1`.
+- error: An error response, for example `{ok: 0, errmsg: "not master"}`.
+  The empty error response `{}` indicates a network error.
```
ShaneHarvey, line 50: I cargo culted the "`{}` indicates a network error" concept from "responses" above. Let me know if you have a better idea.

source/server-discovery-and-monitoring/tests/README.rst
```diff
@@ -23,11 +23,13 @@ Each YAML file has the following keys:
   A phase of the test optionally sends inputs to the client,
   then tests the client's resulting TopologyDescription.
 
-Each phase object has two keys:
+Each phase object has the following keys:
 
+- description: (optional) A textual description of this phase.
```
ShaneHarvey, line 28: I added the `description` field to make the json files easier to read and make the test phases easier to debug.

source/server-discovery-and-monitoring/tests/README.rst
```diff
@@ -114,9 +129,14 @@ events published during client construction.
...
-For each phase in the file, parse the "responses" array.
-Pass in the responses in order to the driver code.
-If a response is the empty object `{}`, simulate a network error.
+For each phase in the file:
+
+#. Parse the "responses" array. Pass in the responses in order to the driver
+   code. If a response is the empty object `{}`, simulate a network error.
+
+#. Parse the "applicationErrors" array. For each element, simulate the error
+   on an application connection. If a response is the empty object `{}`,
```
p-mongo, line 138: I do not see "application connection" being defined in https://github.com/mongodb/specifications/tree/master/source/server-discovery-and-monitoring/tests, is this a new concept and if yes can it be defined?

jyemin: Perhaps "pooled connection"?

source/server-discovery-and-monitoring/tests/rs/application_error_handling.yml
```diff
@@ -0,0 +1,351 @@
...
+    setName: rs
+
+- description: Ignore stale application error (topologyVersion less)
+  applicationErrors:
+  - address: a:27017
+    generation: 0
+    error:
+      ok: 0
+      errmsg: ShutdownInProgress
+      code: 91
```
jyemin, line 42: Please add cases for all 7 of the recovering/non-primary error codes/messages, as well as one other error code for a negative test.

source/server-discovery-and-monitoring/tests/README.rst
jyemin: I'd like to see something that allows us to represent a network timeout error.  So maybe we need a few extra sibling fields instead, e.g.

    networkError: true
    isTimeout: true

source/server-discovery-and-monitoring/tests/README.rst
```diff
@@ -40,6 +42,13 @@ A response is a pair of values:
   The empty response `{}` indicates a network error
   when attempting to call "ismaster".
 
+An "applicationError" object has the following keys:
```
jyemin, line 45: Did you think about adding `handshakeError`?

source/server-discovery-and-monitoring/tests/README.rst
```diff
@@ -64,6 +73,12 @@ current TopologyDescription. It has the following keys:
 - minWireVersion: absent or an integer.
 - maxWireVersion: absent or an integer.
 - topologyVersion: absent, null, or a topologyVersion document.
+- pool: (optional) A "pool" object.
```
mbroadst, line 76: I'm not sure we should take this approach. The seemingly small changes here would require large refactoring of an existing test runner, and feels a bit shoehorned.

Right now the tests are wholly relegated to `TopologyDescription` and `ServerDescription`, and generally run from the perspective of some kind of `Monitor`. This PR asks us to conceptually move our testing to a higher level where we are now coordinating a connection pool with our topology description - effectively from a `Topology` perspective (or `Cluster`, whatever your driver calls it). 

I don't fundamentally disagree with the approach, but I think it's a pretty big change that's subtly introduced by the addition of a single field here. I think if we make this change then we would want to be very explicit about how to implement this runner. For instance: `simulate a network error on an application connection` leaves a lot up to the implementer.

jyemin: I agree that it's a pretty big change, and could require refactoring not only of a driver's existing test runner but also of the topology/monitor/pool implementation itself in order to allow for the mocking required by these tests.

But I can't think of a better alternative.  All these tests are required in some form or another to ensure correct implementation, and I'd rather we take a shot at spec tests rather than write them all as prose tests.
 
What did you have in mind with advice on implementing a runner?  I'm not sure how prescriptive we can be given the design variations that likely exist between drivers. Maybe we should all POC these tests and see what we can say that is common for all of the implementations?

mbroadst: I was thinking maybe we could take an approach that looks more like the transaction test runner format (precursor to the unified test runner format). In that case we could at least write very explicit helpers like we have for the existing assertion helpers. I'm thinking like `triggerNetworkError` on object `connection` or something. This approach is similarly prone to misinterpretation during implementation, but at least the intent is codified and well documented closer to the actual code.

jyemin: How would we test all the interleavings with that approach? e.g. a non-primary error on an application connection arriving with a stale topology version vs a non-stale primary version?

source/server-discovery-and-monitoring/tests/rs/application_error_handling.yml
ShaneHarvey: I will add this once we agree on the format. (I want to avoid churn.)

source/server-discovery-and-monitoring/tests/README.rst
jyemin: I think we discussed _not_ doing this, but now i think we should keep it.

ShaneHarvey:
I’ve updated the tests to:

- Add back `when` such that we can test the before/after handshake completes section of the spec. In particular, network timeouts after the handshake MUST NOT mark the server unknown while network timeouts before the handshake completes MUST mark the server unknown. See [tests/rs/error_handling_handshake.yml](https://github.com/mongodb/specifications/pull/787/files#diff-e7a4398e861015b753286ca9a31d2298).
- Add tests for all the state change error codes. 
- Add `maxWireVersion` so the tests can assume that "shutdown" errors will clear the connection pool but other command errors will not.

ShaneHarvey:
Per offline discussion, I've made the (hopefully) final update here to:
- split these tests into smaller files in the new `'errors'` directory.
- auto-generate all combinations of errors.
- test command errors pre/post 4.2 to assert that only shutdown errors (and network errors) will clear the connection pool.

PR #788, opened 2020-04-17
SPEC-1669 admin => keyvault for FLE tests/examples

kevinAlbs:
Do not use the admin database for the key vault in FLE tests
because it cannot be dropped in sharded clusters in 4.4.

The change is entirely find-replace. I validated by syncing and running libmongoc's client side encryption tests.

### DIFF
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -506,7 +506,7 @@ collection is unsupported (mongocryptd returns an error):
 .. code:: python
 
    opts = AutoEncryptionOpts (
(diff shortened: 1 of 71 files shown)


source/client-side-encryption/client-side-encryption.rst
```diff
@@ -506,7 +506,7 @@ collection is unsupported (mongocryptd returns an error):
 .. code:: python
 
    opts = AutoEncryptionOpts (
-      key_vault_namespace="admin.keyvault",
+      key_vault_namespace="datakeys.keyvault",
```
kevinAlbs, line 509: Ah, this is flipped now. Updating.

PR #789, opened 2020-04-17
SPEC-1664 Do not support the hint option on any unacknowledged writes

egiurleo:
In this PR I:
- Updated the CRUD spec to reflect that unacknowledged hint option should never be supported
- Deleted tests where hint option is used successfully in unacknowledged writes
- Updated tests where hint option causes a client failure to not be limited to any server versions


divjotarora:
Forgot to ask earlier: why is it incorrect to do an unacknowledged delete/update/etc with a hint on 4.4+? I know the `arrayFilters` option already works this way for updates, but I don't see any rationale written down for it.

egiurleo:
@divjotarora The rationale is that the provided hint option could still be incorrectly formatted or the specified index could no longer exist, and the user would never know.

divjotarora:
@egiurleo That makes sense. Can we add something to https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#q-a to make sure this is written down for future reference?

egiurleo:
@divjotarora Sure thing.

egiurleo:
@divjotarora added a new Q&A for your question.

PR #790, opened 2020-04-17
SPEC-1666 Clarify the conditions for adding the retryable writes error label during transactions

egiurleo:
Clarify that the RetryableWriteError label should not be added to errors that occur during writes within a transaction. @divjotarora did a POC and confirmed that these changes work in Go.

### DIFF
--- a/source/transactions/tests/error-labels.json
+++ b/source/transactions/tests/error-labels.json
@@ -1144,11 +1144,10 @@
             }
           },
           "result": {
-            "errorLabelsContain": [
-              "TransientTransactionError"
-            ],
+            "errorLabelsContain": [],
             "errorLabelsOmit": [
               "RetryableWriteError",
+              "TransientTransactionError",
               "UnknownTransactionCommitResult"
             ]
           }
--- a/source/transactions/tests/error-labels.json
+++ b/source/transactions/tests/error-labels.json
@@ -1106,6 +1106,109 @@
       }
     },
     {
+      "description": "do not add RetryableWriteError label to writeConcernError ShutdownInProgress that occurs within transaction",
+      "failPoint": {
+        "configureFailPoint": "failCommand",
+        "mode": {
+          "times": 1
+        },
+        "data": {
+          "failCommands": [
+            "insert"
+          ],
+          "writeConcernError": {
+            "code": 91,
+            "errmsg": "Replication is being shut down",
+            "errorLabels": [
+              "RetryableWriteError"
+            ]
+          }
+        }
+      },
+      "operations": [
+        {
+          "name": "startTransaction",
+          "object": "session0",
+          "arguments": {
+            "options": {
+              "writeConcern": {
+                "w": "majority"
+              }
+            }
+          }
+        },
+        {
+          "name": "insertOne",
+          "object": "collection",
+          "arguments": {
+            "session": "session0",
+            "document": {
+              "_id": 1
+            }
+          },
+          "result": {
+            "errorLabelsContain": [
+              "TransientTransactionError"
+            ],
+            "errorLabelsOmit": [
+              "RetryableWriteError",
+              "UnknownTransactionCommitResult"
+            ]
+          }
+        },
+        {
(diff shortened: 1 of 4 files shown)


source/transactions/tests/error-labels.yml
```diff
@@ -86,7 +86,7 @@ tests:
         result:
           # Note, the server will return the errorLabel in this case.
           errorLabelsContain: ["TransientTransactionError"]
-          errorLabelsOmit: ["UnknownTransactionCommitResult"]
+          errorLabelsOmit: ["RetryableWriteError", "UnknownTransactionCommitResult"]
```
egiurleo, line 89: This test should already pass, just including the RetryableWriteError label to make it consistent with the rest of the tests in this file.

source/transactions/tests/error-labels.yml
```diff
@@ -143,7 +143,7 @@ tests:
...
```
egiurleo, line 146: This test should already pass, just including the RetryableWriteError label to make it consistent with the rest of the tests in this file.

source/transactions/tests/error-labels.yml
```diff
@@ -200,7 +200,7 @@ tests:
...
```
egiurleo, line 203: This test should already pass, just including the RetryableWriteError label to make it consistent with the rest of the tests in this file.

divjotarora:
Discussed offline and @egiurleo is going to see if we can add tests to cause write concern errors during transactions. In the Go POC, the current tests passed even if I only update the code to check txn requirements in the command error case, not the write concern error case.

egiurleo:
Will work on this shortly and ping both reviewers.

DmitryLukyanov:
I can confirm that the currently provided tests pass on c# driver

source/transactions/tests/error-labels.json
```diff
@@ -1098,6 +1105,106 @@
...
+          "object": "collection",
+          "arguments": {
+            "session": "session0",
+            "document": {
+              "_id": 1
+            }
+          },
+          "result": {
+            "errorLabelsContain": [
+              "TransientTransactionError"
```
DmitryLukyanov, line 1148: The error 91 doesn't look like a network error. Should this label really be added?

egiurleo: You're right, this is a typo! Thanks for catching. Fixing this now.

source/transactions/tests/error-labels.yml
```diff
@@ -670,6 +673,68 @@ tests:
...
+          command:
+            abortTransaction: 1
+            lsid: session0
+            txnNumber:
+              $numberLong: "1"
+            startTransaction:
+            autocommit: false
+            writeConcern:
+          command_name: abortTransaction
+          database_name: *database_name
```
divjotarora, line 732: This expects the command to go to the `transactions-tests` database but per https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#aborttransaction-1, `abortTransaction` must be sent to `admin`.

source/transactions/tests/error-labels.yml
```diff
@@ -670,6 +673,68 @@ tests:
...
+          writeConcernError:
+            code: 91
+            errmsg: Replication is being shut down
+
+    operations:
+      - name: startTransaction
+        object: session0
+        arguments:
+          options:
+            writeConcern:
```
divjotarora, line 692: This configures the transaction to use write concern majority, but the `abortTransaction` expectation below expects `writeConcern: `. Per https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#writeconcern, I think the expectation should include majority. Alternatively, we could just remove the `writeconcern` field from the expectation. That seems like something that should be tested by the options tests, not this.

egiurleo: Yup, agreed.

source/transactions/tests/error-labels.yml
egiurleo: 👍

PR #792, opened 2020-04-28
SPEC-1677 add wait step to 'Parse lastWriteDate' test

kevinAlbs:
Validated in libmongoc: https://jira.mongodb.org/browse/CDRIVER-2985

### DIFF
--- a/source/max-staleness/max-staleness-tests.rst
+++ b/source/max-staleness/max-staleness-tests.rst
@@ -89,8 +89,8 @@ while still testing that the driver parses lastWriteDate
 and converts it to an appropriate unit.)
 
 .. [1] This ensures the next insert occurs at least one second after server
-selection (another write may have occurred from the noop writer just before
-server selection if mongod was configured with periodicNoopIntervalSecs=1).
+   selection (another write may have occurred from the noop writer just before
+   server selection if mongod was configured with periodicNoopIntervalSecs=1).
 
 Absent lastWriteDate
 --------------------
--- a/source/max-staleness/max-staleness-tests.rst
+++ b/source/max-staleness/max-staleness-tests.rst
@@ -88,7 +88,9 @@ but less than 10 seconds greater.
 while still testing that the driver parses lastWriteDate
 and converts it to an appropriate unit.)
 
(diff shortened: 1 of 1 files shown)


source/max-staleness/max-staleness-tests.rst
```diff
@@ -87,6 +88,8 @@ but less than 10 seconds greater.
 while still testing that the driver parses lastWriteDate
 and converts it to an appropriate unit.)
 
+.. [1] This ensures the next insert occurs at least one second after server selection (another write may have occurred from the noop writer just before server selection if mongod was configured with periodicNoopIntervalSecs=1).
+
```
ShaneHarvey, line 92: Can you wrap this line at 80 chars?

kevinAlbs: Done.

kevinAlbs:
I was also curious, because we had seen this fail, though much less frequently, before periodicNoopIntervalSecs was set to 1 in mongo orchestration. I suspect that something else write could still interleave (e.g. maybe an expired session being removed).

source/max-staleness/max-staleness-tests.rst
ShaneHarvey: Sorry, the newlines need to be indented: https://docutils.sourceforge.io/docs/user/rst/quickref.html#footnotes

```
.. [1] This...
   selection (another...
   server selection...
```

kevinAlbs: One day I will learn how to speak rst...

ShaneHarvey:
>  I suspect that something else write could still interleave (e.g. maybe an expired session being removed).

Does it need to be something else? I think the no-op writer could still have been the cause even with the 10 second interval. Reducing the interval to 1 second just made the race much more likely.

kevinAlbs:
I think the noop writer only writes if the server is idle. At least, that's what https://jira.mongodb.org/browse/SERVER-23892 describes.

https://github.com/mongodb/mongo/blob/master/src/mongo/db/repl/noop_writer.cpp#L167
IIUC the noop writer only writes if the lastAppliedOptime has changed since it last woke up.

ShaneHarvey:
I implemented this in [PYTHON-2218](https://jira.mongodb.org/browse/PYTHON-2218) and can confirm the extra sleep fixes the race.

PR #793, opened 2020-04-29
SPEC-1676 Add support for hidden option to index creation

divjotarora:
A few things I wasn't sure of:

1. The `hidden` property for an index can be changed via the `collMod` command. I don't see any references to it in our existing specs, so I didn't mention this anywhere. Is a note in the resulting DRIVERS ticket sufficient so that any drivers with `collMod` helpers can update accordingly?

2. The `hidden` property is now returned in the `listIndexes` results documents. An example is `{v: 2, key: {x: 1}, hidden: true}`. This might affect drivers that have helper types for `listIndexes` results, but I don't see any other index options mentioned in https://github.com/mongodb/specifications/blob/master/source/enumerate-indexes.rst#getting-full-index-information, so I didn't make any edits to that spec.

### DIFF
--- a/source/index-management.rst
+++ b/source/index-management.rst
@@ -782,6 +782,15 @@ Common API Components
      * Optionally specifies the wildcard projection of a wildcard index.
      */
     wildcardProjection: Document;
+
+    /**
+     * Optionally specifies that the index should exist on the target collection but should not be used by the query
+     * planner when executing operations.
+     *
+     * This option is only supported by servers >= 4.4. For servers < 3.4, drivers MUST raise an error if the caller
+     * explicitly provides a value.
+     */
+    hidden: Boolean;
   }
 
 ---------
--- a/source/index-management.rst
+++ b/source/index-management.rst
@@ -787,8 +787,7 @@ Common API Components
      * Optionally specifies that the index should exist on the target collection but should not be used by the query
      * planner when executing operations.
      *
-     * This option is only supported by servers >= 4.4. For servers < 3.4, drivers MUST raise an error if the caller
-     * explicitly provides a value.
+     * This option is only supported by servers >= 4.4.
      */
     hidden: Boolean;
   }


source/index-management.rst
```diff
@@ -782,6 +782,15 @@ Common API Components
      * Optionally specifies the wildcard projection of a wildcard index.
      */
     wildcardProjection: Document;
+
+    /**
+     * Optionally specifies that the index should exist on the target collection but should not be used by the query
+     * planner when executing operations.
+     *
+     * This option is only supported by servers >= 4.4. For servers < 3.4, drivers MUST raise an error if the caller
```
jyemin, line 790: Why must drivers raise for this one, when they don't for any of the other options above?

divjotarora: I think it's consistent to validate these options like we do for other command options in the CRUD spec. I filed https://jira.mongodb.org/browse/DRIVERS-1220 to address this for the others too. I'm also fine with removing this clause right now and deferring to the decision made for DRIVERS-1220 if you think that'd be better.

jyemin: I'd prefer to be consistent with the other options for this command.

divjotarora: Removed the constraint for raising an error.

patrickfreed:
The option addition lgtm, but I wonder if this indicates we should introduce `hideIndex` / `unhideIndex` like the shell did. The "hidden" field isn't really useful if it can't be toggled, and currently the only way to do that is through `collMod` in a `runCommand` for most drivers.

divjotarora:
@patrickfreed I was thinking about this too and I agree that just exposing an option to create a hidden index isn't very helpful, especially given that server scope calls out hiding an existing index before dropping it as the main motivation. Like you said, we could add `Collection.HideIndex(name)` to the standard API and `IndexView.Hide(name)` to the view API. One possible complication: the shell helpers described in the scope can take either a keys document or the index name as a string, so this could be an issue for drivers that don't have overloading (e.g. Go), but we can work around this if we need to.

patrickfreed:
Yeah, I think that sounds good. I'd say also add overloads for ones for models as well like we do for `dropIndex` with the same guidance about picking which ones to implement. 

As we talked about in the commitQuorum PR, the language around acceptable deviations for overloading in this spec could be improved to explicitly allow languages like go to do `dropIndexByX` or something, since I think as written that is currently implied anyways. Did we ever file a DRIVERS ticket for that?

divjotarora:
@jyemin Thoughts on helpers to hide/unhide indexes?

jyemin:
It makes me sad that we now have to add partial support for `collMod` to the public API, but I see the argument that without it the feature is kind of useless.

@behackett what do you think?

divjotarora:
@patrickfreed I filed https://jira.mongodb.org/browse/DRIVERS-1221.

behackett:
I don't think we should add these helpers, for a few reasons. For one, collmod can be called through the driver command helper trivially using the name of the index. Helpers seem like an unnecessary expansion of the driver API. Also, it's not clear if collmod will make it, in part or whole, into the versioned API.

divjotarora:
@patrickfreed Given Bernie's comment, are you OK merging this change as-is?

@jyemin I don't know much about the versioned API project. Is it worth filing a follow-on DRIVERS ticket to consider adding helpers to hide/unhide indexes once we know if `collMod` will be in the versioned API?

jyemin:
Yes, please open a follow-on ticket for additional helpers.

divjotarora:
Filed https://jira.mongodb.org/browse/DRIVERS-1226.

Issue #258, opened 2018-01-09
Improper link in Decimal128 page


shanehsu:
The page currently links a specification of [this][1], where [this method][2] is explained.

However, after implementing that, I found out the BSON C source actually implements [this][3]

And [the comment in the driver][4] seems to suggest this also. I only found this out after my implementation failed the test cases in `test-decimal128.c`. 

Or am I mistaken?

[1]: https://github.com/mongodb/specifications/blob/master/source/bson-decimal128/decimal128.rst#bson-decimal128-implementation-details
[2]: https://en.wikipedia.org/wiki/Decimal128_floating-point_format#Densely_packed_decimal_significand_field
[3]: https://en.wikipedia.org/wiki/Decimal128_floating-point_format#Binary_integer_significand_field
[4]: https://github.com/mongodb/mongo-c-driver/blob/28f5a2fb109a405ac068f16337b40f1073ad2d2d/src/libbson/src/bson/bson-decimal128.c#L207

behackett:
I don't think the link is wrong. The spec does three things:
- Provide a brief introduction to IEEE Decimal128, with the link you mention providing more detail
- Specify how to serialize it to BSON using binary integer decimal format (not DPD)
- Provides a large set of test cases

JohnAD:
I'm in the middle finishing a bson library for the new Nim programming language and this one *really* threw me for a loop. I had eventually deduced the answer and while the docs don't mention that DPD is not being used, the regular reference to  IEEE 754  2008 leads to docs that largely imply that DPD is the only correct way to do this. I actually wrote fully functioning sliding 10-bit DBD code and was frustrated by how the the mongo test cases never worked.

Just like the previous author, I confirmed the distinction by reading the C algo. Unfortunately "binary integer decimal format" can sound like descriptive wording rather than a spec.

So, may I suggest that, in just a few places, it should explicitly say "BSON does not use Densely Packed Binary for the significand, instead it is stored as a simple 110bit unsigned integer." If would have saved me many hours of work and much frustration. I suspect it will help others also.

Side question: can I assume that the "leading decimal" that is stored in the Combination field is never used (always zero)? I've got code to support it :), but I can't think of a circumstance that it would be needed if not using DPD.

I'm happy to do a PR if this sounds acceptable to you all.

behackett:
The first sentence of the specification links to the homepage for the specification for the encoding:

[...]

PR #795, opened 2020-05-02
Add coefficient encoding detail to Decimal128 specification

JohnAD:
The current decimal128 spec mentions that it is based on IEEE 754-2008. That standard allows the coefficient to be encoded in one of two ways:

1. "decimal encoding" using 10-bit declets for groups of 3 digits. This method is also called Densely Packed Decimal (DPD).
2. "binary encoding" using a simple unsigned integer number (BID).

The specification sheet later formally declares that the BSON encoding is based on "Decimal Arithmetic Encodings" (http://speleotrove.com/decimal/decbits.html) which only allows for the coefficient to be encoded as DPD (method 1). To quote from that document:

> coefficient continuation
>
> The remaining, less significant, digits of the coefficient. The coefficient continuation is a multiple of 10 bits (the multiple depending on the format), and the most significant group is on the left (is placed first).
> 
> Each 10-bit group represents three decimal digits, using Densely Packed Decimal encoding.[1]

However, the drivers, and the MongoDB server itself don't do this. They actually use the BID method. For example in the `libbson` code:

https://github.com/mongodb/libbson/blob/master/src/bson/bson-decimal128.c#L215

So, the changes in this PR explicitly let the reader know of which encoding is used and that it differs from the referenced "Decimal Arithmetic Encodings" document.

Also see ref #258

Side note: unfortunately, the IEEE document costs $104 USD and is legally restricted from public distribution; so the reference is of limited value to some.

### DIFF
--- a/source/bson-decimal128/decimal128.rst
+++ b/source/bson-decimal128/decimal128.rst
@@ -76,7 +76,9 @@ Clamping:
 
 Binary Integer Decimal (BID):
    MongoDB uses this binary encoding for the coefficient as specified in ``IEEE
-   754-2008``. The byte order is little-endian, like the rest of the BSON types.
+   754-2008`` section 3.5.2 using method 2 "binary encoding" rather than method 1
+   "decimal encoding". The byte order is little-endian, like the rest of the BSON
+   types.
 
 
 Value Object:
@@ -95,9 +97,12 @@ BSON Decimal128 implementation details
 
 The ``BSON Decimal128`` data type implements the `Decimal Arithmetic Encodings
 <http://speleotrove.com/decimal/decbits.html>`_ specification, with certain
-exceptions around value integrity.  When a value cannot be represented exactly,
-the value will be rejected.
+exceptions around value integrity and the coefficient encoding.  When a value
+cannot be represented exactly, the value will be rejected.
 
+The coefficient MUST be stored as an unsigned binary integer (BID) rather than
+the densely-packed decimal (DPD) shown in the specification. See either the 
+`IEEE Std 754-2008` spec or the driver examples for further detail.
 
 The specification defines several statuses which are meant to signal
 exceptional `circumstances <http://speleotrove.com/decimal/daexcep.html>`_,


JohnAD:
BTW, for context, I'm writing a BSON library for Nim, a relatively new programming language. My `bson` is missing support for Decimal128, so I'm also writing a Decimal128 library. I'm about 1/3rd of the way finished on the Decimal128 library, but this not being documented really threw me for a loop.

Though, as a bonus, my Decimal128 library will also support the DPD method as an option. :grin: 

https://github.com/JohnAD/decimal128

https://github.com/JohnAD/bson

source/bson-decimal128/decimal128.rst
```diff
@@ -76,7 +76,9 @@ Clamping:
 
 Binary Integer Decimal (BID):
    MongoDB uses this binary encoding for the coefficient as specified in ``IEEE
-   754-2008``. The byte order is little-endian, like the rest of the BSON types.
+   754-2008`` section 3.5.2 using method 2 "binary encoding" rather than method 1
```
JohnAD, line 79: For any who do not have access to the `IEEE Std. 754-2008` document, a quote of the key section:

from 3.5.2 on page 11:

> c. For finite numbers, r is (S, E − bias, C) and v = (−1) S × 10 (E−bias) × C, where C is the concatenation of the leading significand digit or bits from the combination field G and the trailing significand field T, and where the biased exponent E is encoded in the combination field. The encoding within these fields depends on whether the implementation uses the decimal or the binary encoding for the significand.

> > 1) If the implementation uses the decimal encoding for the significand, then the least significant w bits of the exponent are G 5 through G w +4 . The most significant two bits of the biased exponent and the decimal digit string d 0 d 1 ...d p −1 of the significand are formed from bits G 0 through G 4 and T as follows:

> > > i) When the most significant five bits of G are 110xx or 1110x, the leading significand digit d 0 is 8 + G 4 , a value 8 or 9, and the leading biased exponent bits are 2G 2 + G 3 , a value 0, 1, or 2.

> > > ii) When the most significant five bits of G are 0xxxx or 10xxx, the leading significand digit d 0 is 4G 2 + 2G 3 + G 4 , a value in the range 0 through 7, and the leading biased exponent bits are 2G 0 + G 1 , a value 0, 1, or 2. Consequently if T is 0 and the most significant five bits of G are 00000, 01000, or 10000, then v = (−1) S × (+0). 

> > The p −1 = 3 × J decimal digits d 1 ...d p −1 are encoded by T, which contains J declets encoded in densely-packed decimal.
> > A canonical significand has only canonical declets, as shown in Tables 3.3 and 3.4. Computational operations produce only the 1000 canonical declets, but also accept the 24 non-canonical declets in operands.

> > 2) Alternatively, if the implementation uses the binary encoding for the significand, then:

> > > i) If G 0 and G 1 together are one of 00, 01, or 10, then the biased exponent E is formed from G 0 through G w +1 and the significand is formed from bits G w +2 through the end of the encoding (including T).

> > > ii) If G 0 and G 1 together are 11 and G 2 and G 3 together are one of 00, 01, or 10, then the biased exponent E is formed from G 2 through G w +3 and the significand is formed by prefixing the 4 bits (8 + G w +4 ) to T.

> > The maximum value of the binary-encoded significand is the same as that of the corresponding decimal-encoded significand; that is, 10 (3 × J + 1) −1 (or 10 (3 × J ) −1 when T is used as the payload of a NaN). If the value exceeds the maximum, the significand c is non-canonical and the value used for c is zero.

> Computational operations generally produce only canonical significands, and always accept non-canonical significands in operands.

JohnAD:
Changed the title as a person at mongodb.com closed the DOCS-13645 ticket as it was mis-classified. She instead opened a new issue with the correct team. But she did not mention a new ticket number.

jmikola:
Reviewing the comments in [DOCS-13645](https://jira.mongodb.org/browse/DOCS-13645), I don't think any ticket was created. The issue was closed because our docs team does not maintain this repository and I believe the only action taken was a poke to the drivers team to look at this.

I opened [DRIVERS-1313](https://jira.mongodb.org/browse/DRIVERS-1313) to track this PR and hopefully it will be assigned a reviewer after next week's triage meeting. Apologies for the delay in getting someone to look at this.

PR #796, opened 2020-05-04
DRIVERS-1189 Redaction of AWS session token

bazile-clyde:


p-mongo:
I do not believe the access key id should be redacted. The access key id is the username and there are many places where the Ruby driver, for example, explicitly  outputs the username being authenticated with for diagnostic purposes ([example](https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/auth.rb#L165)). While it is possible for the driver to redact the username, I think it is sensible to treat access key id as non-confidential. (Same thing for AWS account id which is revealed by the server in the error message if the aws-mapped user does not exist on the server.)

With respect to the redaction process:

Although I specifically said "error messages" in the ticket, probably because I noticed the redaction behavior in an error message, I think if such behavior is required it should be described more generally. https://github.com/mongodb/specifications/blob/master/source/command-monitoring/command-monitoring.rst#security for example says:

> Some commands and replies will contain sensitive data and in order to not risk the leaking of this data to external sources or logs their commands AND replies MUST be redacted from the events. The value MUST be replaced with an empty BSON document. 

I am not seeing existing language in the specs repo that would be more relevant (i.e. specifically regarding password redaction), so I would suggest something like the following:

> The driver MUST treat the session token, and the secret access key, with the same confidentiality as an application-provided password. The driver SHOULD take steps to prevent unintentional exposure of the secret access key and the session token, such as by replacing them with the text `[REDACTED]` in log and error message output.

I don't know how feasible it is to replace the SHOULD with the MUST in the last sentence.

jmikola:
I agree that the key ID need not be redacted if it's analogous to a username.

Thanks for suggesting the text. I think we can use MUST instead of SHOULD. The only oddity is that we're not prescribing a specific course of action and just requiring some steps to be taken, which is in contrast to the other example you gave when we instructed drivers about replacing BSON documents (very specific course of action). That said, it's reasonable to say that all drivers must do _something_ to protect these secrets from being exposed.

p-mongo:
:+1: Sounds good.

bazile-clyde:
Thanks for the suggestions. After doing a bit more research in AWS documentation, AFAIK, the session token isn't sensitive and can't actually do anything either on its own in combination with the access key ID. The only sensitive portion of the credentials is the secret key since its used for signing. @p I could definitely be missing something but why exactly do you believe that this should be redacted?

p-mongo:
I do not know whether the session token should or should not be redacted. In https://jira.mongodb.org/browse/DRIVERS-1189 my intention was to document the requirement _if_ it was a requirement. It is my impression that the session token is useless without the secret key in any event.

I would also support a spec change stating that the session token is not confidential.

bazile-clyde:
> I do not know whether the session token should or should not be redacted. In https://jira.mongodb.org/browse/DRIVERS-1189 my intention was to document the requirement _if_ it was a requirement. It is my impression that the session token is useless without the secret key in any event.
> 
> I would also support a spec change stating that the session token is not confidential.

Gotcha. Add a line stating that only the secret access key is sensitive to avoid having the same questions asked about the access key ID.

PR #798, opened 2020-05-05
SPEC-1673 allowDiskUse on find is only supported in MongoDB 4.4+

patrickfreed:
[SPEC-1673](https://jira.mongodb.org/browse/SPEC-1673)

This PR documents that `allowDiskUse` is only supported on MongoDB 4.4+, and it adds guidance + a test around error reporting when the value is specified against pre 4.4 servers. This guidance is intended to be similar to the guidance provided for other options introduced in newer server versions.

### DIFF
--- a/source/crud/crud.rst
+++ b/source/crud/crud.rst
@@ -415,8 +415,10 @@ Read
      * can write temporary data to disk while executing the find operation.
      *
      * This option is sent only if the caller explicitly provides a value. The default
-     * is to not send a value. For servers < 3.2, this option is ignored and not sent
-     * as allowDiskUse does not exist in the OP_QUERY wire protocol.
+     * is to not send a value.
+     *
+     * This option is only supported by servers >= 4.4. Older servers >= 3.2 will report an error for using this option.
+     * For servers < 3.2, the driver MUST raise an error if the caller explicitly provides a value.
      *
      * @see https://docs.mongodb.com/manual/reference/command/find/
      */
--- a/source/crud/tests/v2/find-allowdiskuse-clientError.yml
+++ b/source/crud/tests/v2/find-allowdiskuse-clientError.yml
@@ -1,11 +1,11 @@
 runOn:
-  - { maxServerVersion: "4.2" }
+  - { maxServerVersion: "3.0.99" }
 
-collection_name: &collection_name 'test_find_allowdiskuse_error'
+collection_name: &collection_name 'test_find_allowdiskuse_clienterror'
 
 tests:
   -
-    description: "Find fails when allowDiskUse true is specified against pre 4.4 server"
+    description: "Find fails when allowDiskUse true is specified against pre 3.2 server"
     operations:
       -
         object: collection
@@ -15,7 +15,7 @@ tests:
           allowDiskUse: true
         error: true
   -
-    description: "Find fails when allowDiskUse false is specified against pre 4.4 server"
+    description: "Find fails when allowDiskUse false is specified against pre 3.2 server"
     operations:
       -
         object: collection
--- a/source/crud/tests/v2/find-allowdiskuse-clientError.yml
+++ b/source/crud/tests/v2/find-allowdiskuse-clientError.yml
@@ -14,6 +14,7 @@ tests:
           filter: { }
           allowDiskUse: true
         error: true
+    expctations: []
   -
     description: "Find fails when allowDiskUse false is specified against pre 3.2 server"
     operations:
@@ -24,3 +25,4 @@ tests:
           filter: { }
           allowDiskUse: false
         error: true
+    expectations: []
--- a/source/crud/tests/v2/find-allowdiskuse-clientError.yml
+++ b/source/crud/tests/v2/find-allowdiskuse-clientError.yml
@@ -14,7 +14,7 @@ tests:
           filter: { }
           allowDiskUse: true
         error: true
-    expctations: []
+    expectations: []
   -
     description: "Find fails when allowDiskUse false is specified against pre 3.2 server"
     operations:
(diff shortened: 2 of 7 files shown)


patrickfreed:
We don't run the crud v2 tests in either Swift or Rust, so I wasn't able to actually run this test. It's simple enough though that there shouldn't be any issues.

source/crud/crud.rst
```diff
@@ -415,8 +415,10 @@ Read
      * can write temporary data to disk while executing the find operation.
      *
      * This option is sent only if the caller explicitly provides a value. The default
-     * is to not send a value. For servers < 3.2, this option is ignored and not sent
-     * as allowDiskUse does not exist in the OP_QUERY wire protocol.
+     * is to not send a value.
+     *
+     * This option is only supported by servers >= 4.4. Older servers >= 3.2 will report an error for using this option.
+     * For servers < 3.2, the driver MUST raise an error if the caller explicitly provides a value.
```
jmikola, line 421: Reading [SPEC-1673](https://jira.mongodb.org/browse/SPEC-1673), I think you may have missed the context for the current behavior from the original code review for [SPEC-1418](https://jira.mongodb.org/browse/SPEC-1418). See: https://github.com/mongodb/specifications/pull/692#discussion_r366563400 and follow-up PR in #716.

To recap: The behavior to silently ignore `allowDiskUse` for older server versions using OP_QUERY is more akin to `maxAwaitTimeMS`, which we silently ignore for OP_GET_MORE. This is contrast to `hint`, where we want to guarantee users see an error (either from client or server) because the option could entirely change the results returned by the server based on the index used.

In the case of `allowDiskUse`, failure to apply this option to old queries at worse could lead to a memory error. With `maxAwaitTimeMS`, a query without it would theoretically return the same result if the cursor was iterated further.

cc: @jyemin in case he has additional context to share

source/crud/tests/v2/find-allowdiskuse-error.yml
```diff
@@ -0,0 +1,26 @@
+runOn:
+  - { maxServerVersion: "4.2" }
```
jmikola, line 2: Depending on whether this is still relevant in light of my earlier comment, I think prior art would suggest separate tests for client and server-side errors (with appropriate APM assertions), as in [deleteOne-hint-clientError.yml](https://github.com/mongodb/specifications/blob/master/source/crud/tests/v2/deleteOne-hint-clientError.yml) and 
[deleteOne-hint-serverError.yml](https://github.com/mongodb/specifications/blob/master/source/crud/tests/v2/deleteOne-hint-serverError.yml).

source/crud/crud.rst
patrickfreed: I did miss the context of that PR, though having read through it I can't say I agree with its conclusion. Specifically, the rationale hinges on the assertion that "Application sets AllowDiskUse and server does not support it and ignores it: not a problem", which I do not believe to be true from several angles:
- a user specifying `allowDiskUse` against a server that doesn't support it is _always_ a bug in the user's code. We should ideally notify them of this ASAP.
- It's inconsistent from an API perspective for the driver to not report errors conditionally when an option isn't supported.
- "you're going to get the exact same result set whether you specify AllowDiskUse or not (assuming sufficient memory)." - the "assuming" clause assumes too much here, seeing as the purpose of the option is to obviate concerns about having sufficient memory. Without the assumption, the presence of the option absolutely can affect the outcome of a query.
- It would be surprising behavior for a user to have ticked the option for paging to disk to see an out of memory error. It would not be surprising behavior to see that they got an error for specifying an invalid option.

I do agree that given 3.0 has been EOL for a while, we shouldn't necessarily be making lots of exceptions for supporting it. The fact of the matter is that we already are doing that in other places though (e.g. `hint`), so we should be consistent about it.

jmikola: No objections on my end. If this is merged, I'd suggest requesting language tickets ASAP so some drivers have a chance to make the corresponding change in their 4.4-compat release (alongside previous work for SPEC-1418).

I stand by my suggestion to split up client and server error tests below.

source/crud/tests/v2/find-allowdiskuse-error.yml
patrickfreed: done

source/crud/crud.rst
patrickfreed: Sounds good. I think the others involved in the original decision will probably want to weigh in here before we can merge anything though.

source/crud/tests/v2/find-allowdiskuse-clientError.yml
```diff
@@ -1,11 +1,11 @@
 runOn:
-  - { maxServerVersion: "4.2" }
+  - { maxServerVersion: "3.0.99" }
 
-collection_name: &collection_name 'test_find_allowdiskuse_error'
+collection_name: &collection_name 'test_find_allowdiskuse_clienterror'
```
jmikola, line 4: Is this field even necessary if we're not expecting the collection name in an APM assertion?

patrickfreed: I wasn't sure if the test runners needed it for running the operation against _some_ collection. e.g. the `deleteOne-hint-clientError.yml` test specifies it despite not having any APM assertoins. I'm happy to remove it if test runners have a fallback collection they operate on; I just wasn't sure if that is the case.

source/crud/crud.rst
mbroadst: I'm fine with the change as it is, let's see if @jyemin has any opinion here

jyemin: There are enough cooks in this kitchen.  I'll leave it to you all to make the call.

source/crud/tests/v2/find-allowdiskuse-clientError.yml
jmikola: Feel free to leave as-is, then. All of the `v2` tests appear to have it even though the field is documented as optional. I can't speak for other drivers, but I definitely use my own default if this field in not available.

PR #802, opened 2020-05-11
SPEC-1687 Add SDAM isMaster cancellation test

ShaneHarvey:
This change adds `tests/integration/cancel-server-check.yml`. This test verifies the "isMaster Cancellation" section of the spec. If a driver mistakenly did not cancel the isMaster check, the server would remain in the Unknown state for 10 seconds (until the in-progress streaming isMaster checks succeeds)  and thus the retried "insertOne" would fail during server selection (serverSelectionTimeoutMS=5000).

I have verified that the test succeeds in Python (https://evergreen.mongodb.com/version/5eb9d90a0ae6063d1374c3e1). I've also verified that the test fails when the cancellation logic is removed.

### DIFF
--- a/source/server-discovery-and-monitoring/tests/integration/cancel-server-check.yml
+++ b/source/server-discovery-and-monitoring/tests/integration/cancel-server-check.yml
@@ -1,0 +1,96 @@
+# Test SDAM error handling.
+runOn:
+  # General failCommand requirements (this file does not use appName
+  # with failCommand).
+  - minServerVersion: "4.0"
+    topology: ["replicaset"]
+  - minServerVersion: "4.2"
+    topology: ["sharded"]
+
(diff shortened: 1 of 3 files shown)


source/server-discovery-and-monitoring/tests/integration/cancel-server-check.yml
```diff
@@ -0,0 +1,96 @@
+# Test SDAM error handling.
+runOn:
+  # General failCommand requirements (this file does not use appName
+  # with failCommand).
+  - minServerVersion: "4.0"
+    topology: ["replicaset"]
+  - minServerVersion: "4.2"
+    topology: ["sharded"]
```
ShaneHarvey, line 8: Note that this test passes on both 4.4+ and <4.4 servers (both streaming and polling).

PR #803, opened 2020-05-13
Fix typo in UUID spec prose tests

prashantmital:
`JAVA_LEGACY` was mistakenly omitted from this prose test.

### DIFF
--- a/source/uuid.rst
+++ b/source/uuid.rst
@@ -305,7 +305,7 @@ Implicit Decoding
    a. Assert that the value of the "standard" field is of the native UUID type and is equal to the given UUID
    b. Assert that the value of the "legacy" field is of type BsonBinary and is equal to the inserted value.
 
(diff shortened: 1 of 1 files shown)


source/uuid.rst
```diff
@@ -305,7 +305,7 @@ Implicit Decoding
    a. Assert that the value of the "standard" field is of the native UUID type and is equal to the given UUID
    b. Assert that the value of the "legacy" field is of type BsonBinary and is equal to the inserted value.
 
-3. Set the uuidRepresentation of the client to "unspecified". Insert a document containing two fields. The "standard" field should contain a BSON Binary created by creating a BsonBinary instance with the given UUID and the STANDARD UuidRepresentation.  The "legacy" field should contain a BSON Binary created by creating a BsonBinary instance with the given UUID and the PYTHON_LEGACY UuidRepresentation. Find the document.
+3. Set the uuidRepresentation of the client to "unspecified". Insert a document containing two fields. The "standard" field should contain a BSON Binary created by creating a BsonBinary instance with the given UUID and the STANDARD UuidRepresentation.  The "legacy" field should contain a BSON Binary created by creating a BsonBinary instance with the given UUID and the JAVA_LEGACY UuidRepresentation. Find the document.
```
jyemin, line 308: I'm not clear why you think this is incorrect.  I think any of the legacy representations would be sufficient for the purposes of this test, which is about what happens in decoding when the uuidRepresentation of the client is "unspecified"

PR #805, opened 2020-05-14
SPEC-1691 test arrayFilters on subsequent ops

rcsanchez97:


source/crud/tests/v2/bulkWrite-arrayFilters2.yml
```diff
@@ -0,0 +1,37 @@
...
+              # UpdateMany with arrayFilters
+              name: "updateMany"
+              arguments:
+                filter: {}
+                update: { $set: { "y.$[i].b": 2 } }
+                arrayFilters: [ { "i.b": 1 } ]
+          options: { ordered: true }
+        result:
+          # Client-side error.
+          errorContains: The selected server does not support array filters
```
kevinAlbs, line 37: This error message is C driver specific. I think we should just assert that an error occurs to keep this driver agnostic. Example:
https://github.com/mongodb/specifications/blob/master/source/crud/tests/v2/bulkWrite-delete-hint-clientError.yml#L61

rcsanchez97: Done

source/crud/tests/v2/bulkWrite-arrayFilters-clientError.yml
```diff
@@ -0,0 +1,35 @@
...
+    -
+        # arrayFilters support first introduced in 3.5.6
+        maxServerVersion: "3.5.5"
+
+data:
+    - {_id: 1, y: [{b: 3}, {b: 1}]}
+    - {_id: 2, y: [{b: 0}, {b: 1}]}
+
+collection_name: &collection_name "test"
+database_name: &database_name "crud-tests"
```
jmikola, line 11: If this test isn't doing APM assertions (where a collection or database name could show up in the command body), I don't think we need to specify either of these.

source/crud/tests/v2/bulkWrite-arrayFilters-clientError.yml
```diff
@@ -0,0 +1,35 @@
...
+                update: { $set: { "y.0.b": 2 } }
+            -
+              # UpdateMany with arrayFilters
+              name: "updateMany"
+              arguments:
+                filter: {}
+                update: { $set: { "y.$[i].b": 2 } }
+                arrayFilters: [ { "i.b": 1 } ]
+          options: { ordered: true }
+          error: true
```
jmikola, line 35: Since this is asserting a client-side error, I think the test can also use `expectations: []` (as other client tests), which would assert that no command is ever sent to the server.

source/crud/tests/v2/bulkWrite-arrayFilters-clientError.yml
```diff
@@ -0,0 +1,35 @@
...
+data:
+    - {_id: 1, y: [{b: 3}, {b: 1}]}
+    - {_id: 2, y: [{b: 0}, {b: 1}]}
+
+collection_name: &collection_name "test"
+database_name: &database_name "crud-tests"
+
+tests:
+  -
+    description: "BulkWrite on server (that doesn't support arrayFilters) with arrayFilters on second op"
```
jmikola, line 15: I realize it's outside of the scope of [SPEC-1691](https://jira.mongodb.org/browse/SPEC-1691), but I don't believe we have an existing test for raising an error if arrayFilters is used in the first op. Would you mind adding another test case for that while you're creating this file?

source/crud/tests/v2/bulkWrite-arrayFilters-clientError.yml
```diff
@@ -0,0 +1,35 @@
+runOn:
+    -
```
jmikola, line 2: I'm not sure if it's significant for YAML, but in `tests:` later on you're using a two-space indent, while `runOn:` and `data:` uses four. I think we can use two throughout the file.

I realize this might have been copied from [`bulkWrite-arrayFilters.yml`](https://github.com/mongodb/specifications/blob/master/source/crud/tests/v2/bulkWrite-arrayFilters.yml), which has the same inconsistency. No objections if you want to fix that in a separate commit within this PR.

source/crud/tests/v2/bulkWrite-arrayFilters-clientError.yml
rcsanchez97: Quite right.  I forgot to remove them when I trimmed down the file.  They are now gone.

source/crud/tests/v2/bulkWrite-arrayFilters-clientError.yml
rcsanchez97: Quite right.  I forgot to remove them when I trimmed down the file.  They are now gone.

source/crud/tests/v2/bulkWrite-arrayFilters-clientError.yml
rcsanchez97: You are correct that I copy/pasted that.  I'm fixing the spacing in both files.

source/crud/tests/v2/bulkWrite-arrayFilters-clientError.yml
rcsanchez97: I was not aware of that.  I've added the `expectations: []` as you have suggested.

source/crud/tests/v2/bulkWrite-arrayFilters-clientError.yml
rcsanchez97: Done.

source/crud/tests/v2/bulkWrite-arrayFilters-clientError.yml
```diff
@@ -0,0 +1,50 @@
...
+              name: "updateOne"
+              arguments:
+                filter: {}
+                update: { $set: { "y.0.b": 2 } }
+                arrayFilters: [ { "i.b": 1 } ]
+          options: { ordered: true }
+          error: true
+    expectations: []
+  -
+    description: "BulkWrite on server (that doesn't support arrayFilters) with arrayFilters on second op"
```
jmikola, line 29: Can we remove the parenthetical, so this becomes the previous test description with some extra text to follow?

source/crud/tests/v2/bulkWrite-arrayFilters-clientError.yml
```diff
@@ -0,0 +1,50 @@
...
+          requests:
+            -
+              # UpdateOne with with arrayFilters
+              name: "updateOne"
+              arguments:
+                filter: {}
+                update: { $set: { "y.0.b": 2 } }
+                arrayFilters: [ { "i.b": 1 } ]
+          options: { ordered: true }
+          error: true
```
jmikola, line 26: I missed this in the first review, but `error` should appear at the top level of the `operation` element. This has it placed within `arguments`.

source/crud/tests/v2/bulkWrite-arrayFilters-clientError.yml
rcsanchez97: Done.

source/crud/tests/v2/bulkWrite-arrayFilters-clientError.yml
rcsanchez97: I missed that as well.  Auto-indent probably "helped" me out there.  Fixed.

rcsanchez97:
@kevinAlbs @jmikola since I do not have write permissions in this repo, could one of you rebase/merge/whatever and then close the PR?

PR #807, opened 2020-05-19
SPEC-1690 Add average RTT tracking to server monitoring pseudocode

ShaneHarvey:
Jira: [SPEC-1690](https://jira.mongodb.org/browse/SPEC-1690)

### DIFF
--- a/source/server-discovery-and-monitoring/server-monitoring.rst
+++ b/source/server-discovery-and-monitoring/server-monitoring.rst
@@ -700,7 +700,7 @@ The event API here is assumed to be like the standard `Python Event
                 response = read next isMaster exhaust response
             elif previousDescription.topologyVersion:
                 # Initiate streaming isMaster
-                rttMonitor.start()
+                rttMonitor.run()
                 set connection timeout to connectTimeoutMS+heartbeatFrequencyMS
                 response = call {isMaster: 1, topologyVersion: previousDescription.topologyVersion, maxAwaitTimeMS: heartbeatFrequencyMS}
             else:
--- a/source/server-discovery-and-monitoring/server-monitoring.rst
+++ b/source/server-discovery-and-monitoring/server-monitoring.rst
@@ -706,7 +706,7 @@ The event API here is assumed to be like the standard `Python Event
                 response = call {isMaster: 1}
 
             return ServerDescription(response, rtt=rttMonitor.average())
-        except (NetworkError, CommandError) as exc:
+        except Exception as exc:
             close connection
             rttMonitor.reset()
             clear connection pool for the server
@@ -784,7 +784,7 @@ connection, for example:
             try:
                 rtt = pingServer()
                 addSample(rtt)
-            except (NetworkError, CommandError) as exc:
+            except Exception as exc:
                 # Don't call reset() here. The Monitor thread is responsible
                 # for resetting the average RTT.
                 close connection
--- a/source/server-discovery-and-monitoring/server-monitoring.rst
+++ b/source/server-discovery-and-monitoring/server-monitoring.rst
@@ -785,6 +785,8 @@ connection, for example:
                 rtt = pingServer()
                 addSample(rtt)
             except (NetworkError, CommandError) as exc:
+                # Don't call reset() here. The Monitor thread is responsible
+                # for resetting the average RTT.
                 close connection
                 connection = Null
 
--- a/source/server-discovery-and-monitoring/server-monitoring.rst
+++ b/source/server-discovery-and-monitoring/server-monitoring.rst
@@ -547,6 +547,8 @@ Errors encountered when running a isMaster command MUST NOT update the topology.
 When constructing a ServerDescription from a streaming isMaster response,
 clients MUST use the current roundTripTime from the RTT task.
 
+See the pseudocode in the `RTT thread`_ section for an example implementation.
+
 SDAM Monitoring
 ```````````````
 
@@ -630,10 +632,11 @@ Most platforms can use an event object to control the monitor thread.
 The event API here is assumed to be like the standard `Python Event
 <https://docs.python.org/2/library/threading.html#event-objects>`_.
 `heartbeatFrequencyMS`_ is configurable,
-`minHeartbeatFrequencyMS`_ is always 500 milliseconds::
+`minHeartbeatFrequencyMS`_ is always 500 milliseconds:
+
+.. code-block:: python
 
   class Monitor(Thread):
-
     def __init__():
         # Monitor options:
         serverAddress = serverAddress
@@ -645,6 +648,7 @@ The event API here is assumed to be like the standard `Python Event
         connection = Null
         description = default ServerDescription
         lock = Mutex()
+        rttMonitor = RttMonitor(serverAddress)
 
     def run():
         while this monitor is not stopped:
@@ -690,26 +694,28 @@ The event API here is assumed to be like the standard `Python Event
             # check or the previous check was cancelled.
             if not connection or connection.isClosed():
                 setUpConnection()
-                return new ServerDescription from handshake response
-
-            if connection.moreToCome:
-                read next isMaster exhaust response
-                return new ServerDescription
-
-            if previousDescription.topologyVersion:
+                rttMonitor.addSample(connection.handshakeDuration)
+                response = connection.handshakeResponse
+            elif connection.moreToCome:
+                response = read next isMaster exhaust response
+            elif previousDescription.topologyVersion:
                 # Initiate streaming isMaster
+                rttMonitor.start()
                 set connection timeout to connectTimeoutMS+heartbeatFrequencyMS
-                call {isMaster: 1, topologyVersion: previousDescription.topologyVersion, maxAwaitTimeMS: heartbeatFrequencyMS}
-                return new ServerDescription
+                response = call {isMaster: 1, topologyVersion: previousDescription.topologyVersion, maxAwaitTimeMS: heartbeatFrequencyMS}
+            else:
+                # The server does not support topologyVersion.
+                set connection timeout to connectTimeoutMS
+                start = time()
+                response = call {isMaster: 1}
+                rttMonitor.addSample(time()-start)
 
-            # The server does not support topologyVersion.
-            set connection timeout to connectTimeoutMS
-            call {isMaster: 1}
-            return new ServerDescription
+            return ServerDescription(response, rtt=rttMonitor.average())
         except (NetworkError, CommandError) as exc:
             close connection
+            rttMonitor.reset()
             clear connection pool for the server
-            return new ServerDescription with type=Unknown, error=exc
+            return ServerDescription(type=Unknown, error=exc)
 
     def wait():
         start = gettime()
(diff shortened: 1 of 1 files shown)


source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -682,27 +686,29 @@ The event API here is assumed to be like the standard `Python Event
...
-                return new ServerDescription
-
-            if previousDescription.topologyVersion:
+                rttMonitor.addSample(connection.handshakeDuration)
+                response = connection.handshakeResponse
+            elif connection.moreToCome:
+                response = read next isMaster exhaust response
+            elif previousDescription.topologyVersion:
                 # Initiate streaming isMaster
+                rttMonitor.start()
```
divjotarora, line 695: nit: In the changes for the RTT thread description below, this method is called `run`.

ShaneHarvey:
> The current spec wording for the Measuring RTT section says that clients can use a dedicated RTT connection for all servers, even those that don't support the streamable protocol. I found it easier to implement that way in Go, but the pseudocode seems to only use the RTT thread for streaming. Was this intentional? 

Yes it was intentional. I modeled this after the Python implementation which only starts the RttMonitor when it is discovered that the server supports streaming. Either way is fine so I'm open to changing the pseudocode to always start the RttMonitor.

> Also, the RTT monitor should be stopped if the server replies without a topologyVersion. 

Do we need to? It seems fine to continue using the RttMonitor after a downgrade because it's valid to use the RttMonitor on all server versions. 

> As an aside, the pseudocode doesn't handle the case of a server omitting topologyVersion when one was previously included. The spec says that this isn't an expected state and the driver should restart monitoring on a new connection if it happens.

I think we should remove the line that talks about this case because it is [guaranteed by the server not to happen](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-monitoring.rst#awaitable-ismaster-protocol):
```
A server that implements the new protocol follows these rules:

- Always include the server's topologyVersion in isMaster and State Change
  Error replies.
```

source/server-discovery-and-monitoring/server-monitoring.rst
ShaneHarvey: Fixed.

divjotarora:
> Do we need to? It seems fine to continue using the RttMonitor after a downgrade because it's valid to use the RttMonitor on all server versions.

I agree it's not incorrect to keep it running. We'll just end up taking more samples than previous releases because both the RTTMonitor and the non-streaming/awaitable heartbeat paths add samples. I suggested stopping it on downgrades because my understanding is that starting it only when we detect that the server supports streaming is useful for making sure we don't use multiple dedicated connections if they're not needed, but this wouldn't be true for downgrades.

> I think we should remove the line that talks about this case because it is guaranteed by the server not to happen

Agreed. If the server is already asserting that a 4.4+ node can't return an `{ok: 1}` response without `topologyVersion`, we shouldn't be checking driver-side. @jyemin @mbroadst Thoughts on removing that line?

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -682,27 +686,29 @@ The event API here is assumed to be like the standard `Python Event
...
-            return new ServerDescription
+                response = call {isMaster: 1, topologyVersion: previousDescription.topologyVersion, maxAwaitTimeMS: heartbeatFrequencyMS}
+            else:
+                # The server does not support topologyVersion.
+                set connection timeout to connectTimeoutMS
+                start = time()
+                response = call {isMaster: 1}
+                rttMonitor.addSample(time()-start)
+
+            return ServerDescription(response, rtt=rttMonitor.average())
```
kevinAlbs, line 705: Apologies for barging, but this seemed like an apt PR to bring this up. I interpret this to mean that we should only be using the RTT from the server monitor. Is it permitted for application handshakes to update RTT?

ShaneHarvey: I think that's allowed. In fact, if an application handshake can transition a server from Unknown to Known then it might be required to update the RTT as well (to avoid having a 0 value for average RTT). FYI Python/Java do not update the topology from application handshakes.

kevinAlbs: In that case, I may diverge a little from pseudocode to implement RTT monitoring. The server description on the topology may have an existing running average updated by application handshakes, so I cannot overwrite it.

A related question came to mind. Are we explicitly deciding not to update the topology description when RTT receives a response? There is the note to not mark servers unknown on RTT failure:
> Errors encountered when running a "isMaster" command MUST NOT update the topology.

If it's safe to do so, I could just update the RTT of in the topology description as soon as the RTT monitor thread receives an ismaster reply (including the handshake).

ShaneHarvey: > Are we explicitly deciding not to update the topology description when the RTT monitor receives a response?

Yes. We decided not to update the topology to avoid churn and to avoid races (incorrectly going from Unknown to known or vice versa). The RttThread can't just publish new ServerDescriptions the same way the Monitor does without adding races. There would need to be a new method that simply updates a server's average rtt.

Could you describe how your idea would work in more detail?

kevinAlbs: Ah gotcha. I can simply update the server's average RTT then. I can take a stab at noting that in the "Measuring RTT" section if that seems worth calling out.

What I have working now is an RTT monitor thread that behaves close to the pseudocode, except shared ServerDescription in the TopologyDescription is responsible for the RTT tracking. When an RTT ismaster or polling ismaster returns, it will add to the shared ServerDescription's RTT tracking. 

Updating just the RTT in a ServerDescription seems natural in libmongoc. In the current behavior, ServerDescription's are not immutable and are responsible for updating the RTT. When a server monitor finishes a polling ismaster, the topology description is fed the BSON reply + RTT or an error. [This is the code path](https://github.com/mongodb/mongo-c-driver/blob/9f43168cfa94313d9e7aeffa75d95f76fb25db50/src/libmongoc/src/mongoc/mongoc-topology.c#L90-L91) where either a server monitor and application handshake updates the topology description after an ismaster attempt. The optional RTT there is the RTT for that single ismaster. Though, FWIW I'm considering making a ServerDescription immutable in libmongoc. It'd be easier to reason about, and is more in line with the SDAM spec's recommendations. But I think the logic for updating just the RTT could be the same.

kevinAlbs: Made a note in https://github.com/mongodb/specifications/pull/810

ShaneHarvey: Thanks, I think that sounds reasonable given that C's ServerDescription is mutable.

ShaneHarvey:
Rebased to resolve the merge confilcts.

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -742,6 +752,66 @@ The event API here is assumed to be like the standard `Python Event
...
+    def average():
+        with lock:
+            return movingAverage.get()
+
+    def run():
+        while this monitor is not stopped:
+            try:
+                rtt = pingServer()
+                addSample(rtt)
+            except (NetworkError, CommandError) as exc:
```
divjotarora, line 792: Do we need to reset the average RTT here or is it enough for the server monitor to reset it when it errors?

ShaneHarvey: I don't think we should. If this thread calls reset() then it might lead to this race:
```
    def run():
....
          else:
              # The server does not support topologyVersion.
              set connection timeout to connectTimeoutMS
              start = time()
              response = call {isMaster: 1}
              rttMonitor.addSample(time()-start)
            # <--- RttMonitor calls reset() here. rttMonitor.average() is now 0.
            return ServerDescription(response, rtt=rttMonitor.average())
```

divjotarora: Agreed. I asked because I talked to @jyemin about the Java code and he had structured it in a way where both the server and RTT monitors reset the RTT, so we weren't sure which one(s) should.

mbroadst:
@divjotarora yes, let's go ahead and remove that line from the spec. It's quite difficult to test, and the server guarantees the behavior.

@ShaneHarvey @divjotarora Regarding the RTT tracking: I think it was a mistake of convenience that we every tied RTT tracking to server heartbeats, we happened to be doing something every 10s that was close to an "immediate" command, so why not use it to measure round trip time? It's a side effect.

What if instead we _always_ require an "RTT Monitor" and treat it as the source of truth for these calculations? I think that makes the design easier to reason about, and makes it very explicit where and how the RTT is measured and calculated. We could leave in an acceptable deviation for drivers who wish to do something more complicated (switch between RTT tracker and `isMaster` calls, do both at the same time, etc). Thoughts?

ShaneHarvey:
> What if instead we always require an "RTT Monitor" and treat it as the source of truth for these calculations? I think that makes the design easier to reason about, and makes it very explicit where and how the RTT is measured and calculated. We could leave in an acceptable deviation for drivers who wish to do something more complicated (switch between RTT tracker and isMaster calls, do both at the same time, etc). Thoughts?

I've removed the missing topologyVersion line and updated the code to always start the RttMonitor. However, I still think that the RttMonitor cannot reset the average RTT to 0 without introducing the race mentioned above ([0 RTT with a known ServerDescription](https://github.com/mongodb/specifications/pull/807#discussion_r432687114)).

divjotarora:
I agree that we should always require the RTT monitor and that the RTT monitor should not call reset. If there's a long-lived issue on the server, the server monitor will reset. If it's transient and only hit by the RTT monitor once, resetting doesn't really benefit us anyway.

ShaneHarvey:
> I agree that we should always require the RTT monitor and that the RTT monitor should not call reset. If there's a long-lived issue on the server, the server monitor will reset. If it's transient and only hit by the RTT monitor once, resetting doesn't really benefit us anyway.

So you propose that we never reset the RTT and remove [this requirement from the server selection spec](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#calculation-of-average-round-trip-times)?:
> An Unknown server has no average RTT. When a server becomes unavailable, its average RTT MUST be cleared. Clients MAY implement this idiomatically (e.g nil, -1, etc.)

divjotarora:
No, I was just agreeing that the RTT monitor should not reset the value because it makes things harder to reason about and also causes the race you mentioned. I think it's fine for the server monitor  to reset the average RTT, though.

On an unrelated note, is there any way we can test that drivers use the initial connection handshake from the ServerMonitor to update the average RTT?

ShaneHarvey:
Anything else @mbroadst?

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -742,6 +747,66 @@ The event API here is assumed to be like the standard `Python Event
           interrupt connection read
           close tempConnection
 
+RTT thread
+``````````
+
+The requirements in the `Measuring RTT`_ section can be statisfied with an
```
jyemin, line 753: statisfied => satisfied

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -742,6 +747,66 @@ The event API here is assumed to be like the standard `Python Event
...
+    def average():
+        with lock:
+            return movingAverage.get()
+
+    def run():
+        while this monitor is not stopped:
+            try:
+                rtt = pingServer()
+                addSample(rtt)
+            except (NetworkError, CommandError) as exc:
```
DmitryLukyanov, line 787: as I understand any other exceptions except `NetworkError, CommandError` is not possible here. if so, will it be more accurate to have only `except (Exception)` line here?

ShaneHarvey: Good point, I updated to catch all exceptions.

PR #808, opened 2020-05-19
Add mutex to server monitoring pseudocode

divjotarora:


source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -674,13 +675,21 @@ The event API here is assumed to be like the standard `Python Event
...
+        # Take the mutex to avoid a data race becauase this code writes to the connection field and a concurrent
+        # cancelCheck call could be reading from it.
+        mutex.lock()
         connection = new Connection(serverAddress)
         set connection timeout to connectTimeoutMS
-        perform connection handshake
+        mutex.unlock()
+
+        # Do any potentially blocking operations after releasing the mutex.
+        create the socket and perform connection handshake
```
divjotarora, line 686: Changed from `perform connection handshake` to this because the dialing of the socket should also be done outside the lock.

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -674,13 +675,21 @@ The event API here is assumed to be like the standard `Python Event
...
 
     def checkServer(previousDescription):
         try:
-            if not connection:
+            # The connection is null if this is the first check. It's closed if there was an error during the previous
+            # check or the previous check was cancelled.
+            if not connection or connection.isClosed():
```
divjotarora, line 692: I think this change is needed because `cancelCheck` closes the connection but does not set it to `null`, so the connection won't be re-created on the next check unless `connection.isClosed` is checked.

Opted to make this change instead of having `cancelCheck` assign the connection to `null` because we'd need to take the lock here too if it did assignment.

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -700,7 +709,6 @@ The event API here is assumed to be like the standard `Python Event
             return new ServerDescription
         except (NetworkError, CommandError) as exc:
             close connection
-            connection = Null
```
divjotarora, line 703: Removed the assignment to avoid needing a lock here. Closing the connection is enough as it will be re-created on the next check. This also makes the code easier to reason about imo because `connection` is only null if it's the first check and closed if the previous check was cancelled.

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -674,13 +675,21 @@ The event API here is assumed to be like the standard `Python Event
             wait()
 
     def setUpConnection():
+        # Take the mutex to avoid a data race becauase this code writes to the connection field and a concurrent
+        # cancelCheck call could be reading from it.
+        mutex.lock()
```
ShaneHarvey, line 680: In SPEC-1690 I've also introduced a lock. Can we use it the same in the pseudocode? IE rename it from "mutex" to "lock" and use a with-statement. Like this:
```python
with lock:
    connection = new Connection(serverAddress)
    set connection timeout to connectTimeoutMS
# Do any potentially blocking operations after releasing the mutex.
create the socket and perform connection handshake
```

source/server-discovery-and-monitoring/server-monitoring.rst
ShaneHarvey: Acknowledged. In python we often say "Do not perform I/O while holding the lock."

source/server-discovery-and-monitoring/server-monitoring.rst
ShaneHarvey: Acknowledged.

source/server-discovery-and-monitoring/server-monitoring.rst
ShaneHarvey: Acknowledged.

source/server-discovery-and-monitoring/server-monitoring.rst
divjotarora: Thanks! I was just guessing on reasonable lock syntax. I pushed a new commit to match the pseudocode syntax from SPEC-1690.

source/server-discovery-and-monitoring/server-monitoring.rst
mbroadst: LGTM, I already had code like this in the node implementation so I'm glad we're codifying it here

PR #810, opened 2020-05-20
make minor clarifications to server monitoring

kevinAlbs:
- Note that the RTT monitor should ignore the ismaster response.
- Typo fix in test readme example "waitForEvent" => "assertEventCount".
- Note in tests that a ServerChangedEvent is expected in the beginning.

### DIFF
--- a/source/server-discovery-and-monitoring/tests/integration/isMaster-command-error.yml
+++ b/source/server-discovery-and-monitoring/tests/integration/isMaster-command-error.yml
@@ -27,7 +27,7 @@ tests:
       appname: commandErrorHandshakeTest
     operations:
       # The command error on the initial handshake should mark the server
-      # Unknown and clear the pool.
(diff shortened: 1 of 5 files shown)


source/server-discovery-and-monitoring/tests/integration/isMaster-command-error.yml
```diff
@@ -27,7 +27,7 @@ tests:
       appname: commandErrorHandshakeTest
     operations:
       # The command error on the initial handshake should mark the server
-      # Unknown and clear the pool.
+      # Unknown (emitting a ServerDescriptionChangedEvent) and clear the pool.
```
kevinAlbs, line 30: Related to my question in slack. I initially thought this would not emit a ServerDescriptionChangedEvent until I realized that a ServerDescription with error != ServerDescription without error.

PR #811, opened 2020-05-20
fix: ensure viable servers are used for latency window calculation

mbroadst:
By slightly modifying these two tests we can ensure only viable servers are used to calculate the latency window. In the first set the RTT values are slightly modified such that only the RSSecondary servers are used to calculate the latency window. In the second case we introduce an RSOther server, and ensure that its RTT is not used for the latency window calculation.

SPEC-1694

### DIFF
--- a/source/server-selection/tests/server_selection/ReplicaSetWithPrimary/read/Secondary.yml
+++ b/source/server-selection/tests/server_selection/ReplicaSetWithPrimary/read/Secondary.yml
@@ -3,18 +3,18 @@ topology_description:
   servers:
   - &1
     address: b:27017
-    avg_rtt_ms: 5
+    avg_rtt_ms: 21 # outside the latency window if primary is considered
     type: RSSecondary
     tags:
       data_center: nyc
   - &2
     address: c:27017
-    avg_rtt_ms: 100
+    avg_rtt_ms: 100 # outside the latency window if both secondaries are considered
     type: RSSecondary
     tags:
       data_center: nyc
   - address: a:27017
-    avg_rtt_ms: 26
+    avg_rtt_ms: 5
     type: RSPrimary
     tags:
       data_center: nyc
--- a/source/server-selection/tests/server_selection/ReplicaSetWithPrimary/read/Nearest.json
+++ b/source/server-selection/tests/server_selection/ReplicaSetWithPrimary/read/Nearest.json
@@ -4,7 +4,7 @@
     "servers": [
       {
         "address": "b:27017",
-        "avg_rtt_ms": 5,
+        "avg_rtt_ms": 21,
         "type": "RSSecondary",
         "tags": {
           "data_center": "nyc"
@@ -20,8 +20,16 @@
       },
       {
         "address": "a:27017",
-        "avg_rtt_ms": 26,
+        "avg_rtt_ms": 37,
         "type": "RSPrimary",
+        "tags": {
+          "data_center": "nyc"
+        }
+      },
+      {
+        "address": "d:27017",
+        "avg_rtt_ms": 5,
+        "type": "RSOther",
         "tags": {
           "data_center": "nyc"
         }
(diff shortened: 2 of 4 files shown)


source/max-staleness/tests/ReplicaSetWithPrimary/Secondary.json
```diff
@@ -0,0 +1,39 @@
...
+        "type": "RSSecondary",
+        "avg_rtt_ms": 21,
+        "maxWireVersion": 5
+      }
+    ]
+  },
+  "read_preference": {
+    "mode": "Secondary"
+  },
+  "suitable_servers": [
```
jyemin, line 23: If we're trying to test that the latency window check is applied after the suitable server check, let's make this test have two suitable servers instead of just one, where only one is in the latency window.

source/max-staleness/tests/ReplicaSetWithPrimary/Nearest3.json
```diff
@@ -0,0 +1,39 @@
...
+        "type": "RSSecondary",
+        "avg_rtt_ms": 21,
+        "maxWireVersion": 5
+      }
+    ]
+  },
+  "read_preference": {
+    "mode": "Nearest"
+  },
+  "suitable_servers": [
```
jyemin, line 23: With Nearest mode, isn't "a:27017" also a suitable server?  Java driver fails this test now because it thinks that it is.

mbroadst:
@jyemin sorry I meant to make this a draft yesterday, I wasn't quite finished with the test cases. I've reworked two existing tests to gain the coverage I was looking for PTAL

source/max-staleness/tests/ReplicaSetWithPrimary/Nearest3.json
mbroadst: closing because for re-review

source/max-staleness/tests/ReplicaSetWithPrimary/Secondary.json
mbroadst: resolving for re-review

mbroadst:
@jyemin should we find another reviewer for this?

jyemin:
Sure.

source/server-selection/tests/server_selection/ReplicaSetWithPrimary/read/Secondary.yml
```diff
@@ -3,18 +3,18 @@ topology_description:
...
     address: b:27017
-    avg_rtt_ms: 5
+    avg_rtt_ms: 21 # outside the latency window if primary is considered
     type: RSSecondary
     tags:
       data_center: nyc
   - &2
     address: c:27017
-    avg_rtt_ms: 100
+    avg_rtt_ms: 100 # outside the latency window if both secondaries are considered
```
jyemin, line 12: I don't understand the comment.  This server is outside the latency window regardless of whether the primary is considered.  Consider having three secondaries, where the second is in the latency window, and the third is outside.

mbroadst: That's right, it's the same as the test used to be: this server will _always_ be outside the latency window, and so never show up in the `in_latency_window` list. Maybe it would be more clear if I removed the comments? I'm not sure that adding another secondary gives us more coverage

jyemin: Yeah, the comment is definitely confusing and should be changed.

Regarding the test coverage: I reviewed all of the existing tests and found that there are no tests that have more than one server within the latency window.  Given the semantics of the latency window (which is all about which servers that have an RTT higher than the fastest suitable server should be selected), that seems like an oversight that should be corrected.

jyemin: To close the loop on this.

1. Please update the comment
2. You asked whether adding another secondary gives us more coverage.  I think the answer to that is yes, since we have no existing tests that assert that any more than the fastest server is considered in the latency window.   But even if that's the case, if you consider this additional coverage as out of scope for what you're trying to accomplish here, then I'll accept that and LGTM.

PR #812, opened 2020-05-26
Simplify logic for sessions support detection

mbroadst:
Some standalone servers do support sessions, we should instead depend strictly on the
presence of `logicalSessionsTimeoutMinutes` in the server handshake to determine if a 
server supports sessions.

DRIVERS-1230

### DIFF
--- a/source/sessions/driver-sessions.rst
+++ b/source/sessions/driver-sessions.rst
@@ -3,7 +3,7 @@ Driver Sessions Specification
 =============================
 
 :Spec Title: Driver Sessions Specification (See the registry of specs)
-:Spec Version: 1.7.0
+:Spec Version: 1.7.1
 :Author: Robert Stam
 :Spec Lead: A\. Jesse Jiryu Davis
 :Advisory Group: Jeremy Mikola, Jeff Yemin, Samantha Ritter
@@ -59,7 +59,7 @@ Deployment
     word "cluster" because some people interpret "cluster" to mean "sharded cluster".
 
 Explicit session
-    A session that was started explicitly by the application by calling ``startSession`` 
+    A session that was started explicitly by the application by calling ``startSession``
     and passed as an argument to an operation.
 
 MongoClient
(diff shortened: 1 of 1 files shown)


source/sessions/driver-sessions.rst
```diff
@@ -1269,3 +1267,4 @@ Change log
 :2019-05-15: A ServerSession that is involved in a network error MUST be discarded
 :2019-10-22: Drivers may defer checking if a deployment supports sessions until the first
 operation performed with a session
+:2020-05-26: Simplify logic for determining sessions support, some standalones do support sessions
```
jyemin, line 1270: Can we be more precise about which standalone servers support sessions?  Was support added in some release after the initial one that included sessions for replica sets?

mbroadst: Do you think there are servers which report `logicalSessionsTimeoutMinutes` and don't support sessions? Otherwise, I say we depend completely on its existence for feature detection here, no need to complicate things further

jyemin: That's fine, but it's still confusing to talk about "some standalones".    Maybe just remove everything after the comma?

mbroadst: done

source/sessions/driver-sessions.rst
```diff
@@ -979,19 +977,19 @@ Test Plan
 =========
 
 The test plan SHOULD be run against both replica set and sharded cluster
-topologies.  It MUST NOT be run against a standalone server.
+topologies.
```
jyemin, line 980: Currently there is only one test file: dirty-session-errors.yml, which only runs on replica sets and sharded clusters.  Most of those tests will pass against a standalone (the ones that don't I think is because they use retryable writes).  

I request that we modify that test file (perhaps it will have to be split in two) so we can verify with tests that implicit and explicit sessions work against standalone servers.

PR #817, opened 2020-06-09
SPEC-1646 Add region examples for AWS auth

bazile-clyde:


source/auth/auth.rst
```diff
@@ -863,18 +863,15 @@ Body                     Action=GetCallerIdentity&Version=2011-06-15\\n
...
-first             us-east-1
-================= ========= 
+        Region is not a header, but simply part of the authorization header. The default region for the host ``sts.amazonaws.com`` and all other hosts without a full stop "``.``" character (*dot* or *period*) as a seperator between labels is ``us-east-1``. Drivers MUST derive the region for all other hosts by using the second label as the region as in the example below.
+
+=========================== =========
+Host                        Region
+=========================== =========
+sts.amazonaws.com           us-east-1
+sts.us-west-2.amazonaws.com us-west-2
+no-period-present           us-east-1
```
ShaneHarvey, line 873: Can you add a host with only one dot, eg `"one-period.present"`?

source/auth/auth.rst
```diff
@@ -863,18 +863,15 @@ Body                     Action=GetCallerIdentity&Version=2011-06-15\\n
...
-        endpoints and test endpoints that will not follow this rule.
-
-================= =========
-Host              Region
-================= =========
-sts.amazonaws.com us-east-1
-first.second      second
-first             us-east-1
-================= ========= 
+        Region is not a header, but simply part of the authorization header. The default region for the host ``sts.amazonaws.com`` and all other hosts without a full stop "``.``" character (*dot* or *period*) as a seperator between labels is ``us-east-1``. Drivers MUST derive the region for all other hosts by using the second label as the region as in the example below.
```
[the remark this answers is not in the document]
bazile-clyde: @ShaneHarvey Moved the host validation section. @p I clarified the general syntax of the host so the relationship between the region and host is more obvious.

source/auth/auth.rst
```diff
@@ -863,18 +863,21 @@ Body                     Action=GetCallerIdentity&Version=2011-06-15\\n
...
+Host                        Region
+=========================== =========
+sts.amazonaws.com           us-east-1
+sts.us-west-2.amazonaws.com us-west-2
+no-period-present           us-east-1
+one-period.present          present
+invalid..host               <Error>
+.invalid.host               <Error>
+invalid.host.               <Error>
+"<empty string>"            <Error>
```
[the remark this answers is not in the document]
bazile-clyde: I moved the section of text enforcing the length requirements to right above this table per Shane's comment above.

source/auth/auth.rst
```diff
@@ -863,18 +863,21 @@ Body                     Action=GetCallerIdentity&Version=2011-06-15\\n
...
+=========================== =========
+Host                        Region
+=========================== =========
+sts.amazonaws.com           us-east-1
+sts.us-west-2.amazonaws.com us-west-2
+no-period-present           us-east-1
+one-period.present          present
+invalid..host               <Error>
+.invalid.host               <Error>
+invalid.host.               <Error>
```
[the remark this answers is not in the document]
bazile-clyde: These were supposed to represent empty labels. I added better examples to the table with explanations for why these labels are invalid

bazile-clyde:
Also, after speaking to Mark I realized I could remove the `first.second` example since the general form of the host will always be something like `service-code.region-code.amazonaws.com`. With the exception of `localhost` for testing. Some hosts may appear without a region code, i.e., `sts.amazonaws.com` and others may have extra labels .e.g., `sts.<region-code>.amazonaws.com.ch`

source/auth/auth.rst
```diff
@@ -863,18 +860,27 @@ Body                     Action=GetCallerIdentity&Version=2011-06-15\\n
...
-first.second      second
-first             us-east-1
-================= ========= 
+        Region is not a header, but simply part of the authorization header. The general syntax of a regional endpoint (or host) is as follows.
+
+.. code:: javascript
+
+	service-code.region-code.amazonaws.com
+|
+Drivers MUST derive the region (the ``region-code`` above) from the second label of the host. If a region is not present, or does not contain a full stop "``.``" character (*dot* or *period*) as a seperator, then the default region, ``us-east-1`` MUST be used. Drivers MUST error on any empty labels or if the host itself is empty or longer than 255 bytes. Examples are provided below.
```
ShaneHarvey, line 869: Suggest adding back the "longer than 255 bytes per RFC 1035" link so it's obviously where the limitation comes from.

source/auth/auth.rst
```diff
@@ -863,18 +860,27 @@ Body                     Action=GetCallerIdentity&Version=2011-06-15\\n
         ``*``, Denotes a header that MUST be included in SignedHeaders, if present.
 
 .. note::
```
ShaneHarvey, line 862: `.. note::` is not rendered correctly by Github. Any chance you can use something else? Maybe a new section:
```
Region                   Derived from Host - see `Region Calculation`_ below.
...

Region Calculation
~~~~~~~~~~~~~~~~~~

``Region`` is not a header, ...
```

Or make this a new section:

p-mongo:
I think the previous version of the prose was more clear than the current version. The current version uses undefined terms ("regional endpoint") and contains a circular definition ("Drivers MUST derive the region (the ``region-code`` above) from the second label of the host. If a region is not present, [...] ``us-east-1`` MUST be used.).

I suggest:

1. Putting the `host` validation rules back where they were, in the paragraph that describes the server response that returns `host`.
2. Adding the number-of-periods requirements to the `host` validation rules.

Proposed new language:

> Drivers MUST validate that the server nonce is exactly 64 bytes and the first 32 bytes are the same as the client nonce. Drivers must also validate that the length of  `host`  is greater than 0 and less than or equal to 255 bytes per 
`RFC 1035 <https://tools.ietf.org/html/rfc1035>`_.  Drivers MUST reject FQDN names with empty labels (e.g., "abc..def"), names that start with a period (e.g. ".abc.def") and names that end with a period (e.g. "abc.def.").

3. Reference the `host` validation rules in the `region` derivation section.
4. Define the procedure for calculating `region` from `host` using only the terms defined elsewhere in the spec. Suggested language:

`region` is derived from the value of `host` using the following rules:

- If `host` is invalid according to the validation rules described earlier, the corresponding `region` is not defined and the driver must raise an error.
- If `host` is equal to `sts.amazonaws.com`, `region` is `us-east-1`.
- If `host` does not contain any periods (`.`), `region` is `us-east-1`.
- Otherwise, split `host` at the periods (`.`). `region` is the second component.

For avoidance of doubt, I suggest removing the sentence "Region is not a header, but simply part of the authorization header. " entirely, also removing the sentence "The general syntax of a regional endpoint (or host) is as follows." and the " 	service-code.region-code.amazonaws.com" mention.

I also suggest creating a section for the region definition, similarly to Conversation/MongoCredential sections.

source/auth/auth.rst
bazile-clyde: Done

source/auth/auth.rst
bazile-clyde: Done

bazile-clyde:
@p 
> I think the previous version of the prose was more clear than the current version. The current version uses undefined terms ("regional endpoint") 

This is another name for the host according to amazon's docs. If a developer is following the documentation then they might see what we call host called a "regional endpoint". I mention both terms for convenience.


> and contains a circular definition ("Drivers MUST derive the region (the `region-code` above) from the second label of the host. If a region is not present, [...] `us-east-1` MUST be used.).

I'm not understanding how this is circular. The text in parentheses clarifies (not defines) the term to the left of it. In other words, "region" _is_ the "region-code" in the example above. They're synonyms. 

> 2. Adding the number-of-periods requirements to the `host` validation rules.

There is no "number-of-periods" requirement. The only requirement is that there can not be empty labels. This requirement is clarified in the examples below.

> I also suggest, per Shane's request, adding `example.com` as another example host with `com` as the corresponding calculated region, noting also that this example came up in Slack today.

The original purpose of this ticket was, in part, to reference this issue: 

>the spec currently gives no examples of actual AWS STS host names that contain a region.

So I've decided to stick to examples that drivers could expect to see. "com" is not a region and will cause confusion as to what type of host drivers should expect.

I'm also not sure how deleting the section describing the format of the host, per Amazon's docs, would help. Drivers need to know what format the host will have if they're required to parse it for the region. I put it here for convenience.

This seems to be going beyond the scope of the original ticket. Are there any issues described in [SPEC-1646](https://jira.mongodb.org/browse/SPEC-1646) that still need to be addressed?

bazile-clyde:
@p Just following up on this.

bazile-clyde:
@p-mongo Are there any more points of confusion you'd like to address before I merge this?

bazile-clyde:
@p-mongo updated!

bazile-clyde:
@p-mongo following up on this.

PR #818, opened 2020-06-10
Remove \n from STS request body


p-mongo:


p-mongo:
The length of string "Action=GetCallerIdentity&Version=2011-06-15"
    is 43 characters which matches the stated Content-Length header value.
    The trailing newline is also not present in AWS documentation for
    GetCallerIdentity.

PR #820, opened 2020-06-12
Add heartbeatFrequencyMS default value for retryable read/writes and transaction tests.

DmitryLukyanov:
Questions:
1. Should the same change be added for transactions-convenient-api?
2. Should I create a DRIVER ticket?

### DIFF
--- a/source/retryable-reads/tests/README.rst
+++ b/source/retryable-reads/tests/README.rst
@@ -148,7 +148,7 @@ data.
 .. _GridFSBucket spec: https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst#configurable-gridfsbucket-class
     
 Speeding Up Tests
------------------
+===================
 
 Drivers may benefit reducing `minHeartbeatFrequencyMS`_ in order to speed up
 tests. Python was able to decrease the run time of the tests greatly by lowering
--- a/source/retryable-reads/tests/README.rst
+++ b/source/retryable-reads/tests/README.rst
@@ -156,6 +156,9 @@ the SDAM's ``minHeartbeatFrequencyMS`` from 500ms to 50ms, thus decreasing the
 waiting time after a "not master" error:
 
 .. _minHeartbeatFrequencyMS: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#minheartbeatfrequencyms
+
+Drivers should set the default heartbeatFrequencyMS to 5ms in order to take into account the latest changes regarding streaming protocol. If a test has an explicit heartbeatFrequencyMS value, drivers should use the explicit value.
+
 Optional Enumeration Commands
 =============================
 
@@ -171,3 +174,5 @@ Changelog
              now expressed within ``runOn`` elements.
 
              Add test-level ``useMultipleMongoses`` field.
+
+:2020-06-12: Add the default value for heartbeatFrequencyMS.
--- a/source/retryable-reads/tests/README.rst
+++ b/source/retryable-reads/tests/README.rst
@@ -150,15 +150,13 @@ data.
 Speeding Up Tests
 -----------------
 
-Drivers may benefit reducing `minHeartbeatFrequencyMS`_ in order to speed up
-tests. Python was able to decrease the run time of the tests greatly by lowering
-the SDAM's ``minHeartbeatFrequencyMS`` from 500ms to 50ms, thus decreasing the
-waiting time after a "not master" error:
+Drivers can greatly reduce the execution time of tests by setting `heartbeatFrequencyMS`_
+and `minHeartbeatFrequencyMS`_ (internally) to a small value (e.g. 5ms), below what
+is normally permitted in the SDAM spec. If a test specifies an explicit value for
+heartbeatFrequencyMS (e.g. client or URI options), drivers SHOULD use that value.
 
-Also, similar improvement can be added for `heartbeatFrequencyMS`_ by setting the default heartbeatFrequencyMS to 5ms in order to take into account the latest changes regarding streaming protocol. If a test has an explicit heartbeatFrequencyMS value, drivers should use the explicit value.
-
-.. _minHeartbeatFrequencyMS: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#minheartbeatfrequencyms	
-.. _heartbeatFrequencyMS: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#heartbeatfrequencyms
+.. _minHeartbeatFrequencyMS: ../../server-discovery-and-monitoring/server-discovery-and-monitoring.rst#minheartbeatfrequencyms
+.. _heartbeatFrequencyMS: ../../server-discovery-and-monitoring/server-discovery-and-monitoring.rst#heartbeatfrequencyms
 
 Optional Enumeration Commands
 =============================
--- a/source/retryable-reads/tests/README.rst
+++ b/source/retryable-reads/tests/README.rst
@@ -148,16 +148,17 @@ data.
 .. _GridFSBucket spec: https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst#configurable-gridfsbucket-class
     
 Speeding Up Tests
-===================
+-----------------
 
 Drivers may benefit reducing `minHeartbeatFrequencyMS`_ in order to speed up
 tests. Python was able to decrease the run time of the tests greatly by lowering
 the SDAM's ``minHeartbeatFrequencyMS`` from 500ms to 50ms, thus decreasing the
 waiting time after a "not master" error:
 
-.. _minHeartbeatFrequencyMS: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#minheartbeatfrequencyms
+Also, similar improvement can be added for `heartbeatFrequencyMS`_ by setting the default heartbeatFrequencyMS to 5ms in order to take into account the latest changes regarding streaming protocol. If a test has an explicit heartbeatFrequencyMS value, drivers should use the explicit value.
 
-See also `Speeding Up Tests </source/transactions/tests/README.rst#speeding-up-tests>`_ in the Transactions spec tests.
+.. _minHeartbeatFrequencyMS: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#minheartbeatfrequencyms	
+.. _heartbeatFrequencyMS: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#heartbeatfrequencyms
 
 Optional Enumeration Commands
 =============================
@@ -175,4 +176,4 @@ Changelog
 
              Add test-level ``useMultipleMongoses`` field.
 
-:2020-06-12: Add the default value for heartbeatFrequencyMS.
+:2020-09-16: Suggest lowering heartbeatFrequencyMS in addition to minHeartbeatFrequencyMS.
--- a/source/retryable-reads/tests/README.rst
+++ b/source/retryable-reads/tests/README.rst
@@ -157,7 +157,7 @@ waiting time after a "not master" error:
 
 .. _minHeartbeatFrequencyMS: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#minheartbeatfrequencyms
 
-Drivers should set the default heartbeatFrequencyMS to 5ms in order to take into account the latest changes regarding streaming protocol. If a test has an explicit heartbeatFrequencyMS value, drivers should use the explicit value.
+See also `Speeding Up Tests </source/transactions/tests/README.rst#speeding-up-tests>`_ in the Transactions spec tests.
 
 Optional Enumeration Commands
 =============================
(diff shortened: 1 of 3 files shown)


source/retryable-reads/tests/README.rst
```diff
@@ -156,6 +156,9 @@ the SDAM's ``minHeartbeatFrequencyMS`` from 500ms to 50ms, thus decreasing the
 waiting time after a "not master" error:
 
 .. _minHeartbeatFrequencyMS: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#minheartbeatfrequencyms
+
+Drivers should set the default heartbeatFrequencyMS to 5ms in order to take into account the latest changes regarding streaming protocol. If a test has an explicit heartbeatFrequencyMS value, drivers should use the explicit value.
```
jmikola, line 160: I'm on the fence about sprinkling this throughout all of the test files. It seems like something that would be better documented in a "best practices" document at the top of the repository, since it might as well apply to any test runner working with a server. Drivers would also probably want to do this in their own test suites (I've done so in my PHPUnit configuration for a few years now).

DmitryLukyanov: >a "best practices" document
do you mean this doc: https://github.com/mongodb/specifications/blob/master/README.rst ?

jmikola: The main README may be the best place for now. I'll defer to @jyemin and @ShaneHarvey, though.

My main concern is wanting to avoid duplicating the same text in multiple places.

ShaneHarvey: I think this guidance would be out of place in the main readme. What if we add the description of "Speeding Up Tests" to only the transaction spec file and then link to that section in the other files?

DmitryLukyanov: sgtm, @jmikola ? :)

jmikola: I don't object. I'd also suggest noting this in SPEC ticket and linking it up to [DRIVERS-709](https://jira.mongodb.org/browse/DRIVERS-709). From there, I can include it in the final README for the unified test runner.

DmitryLukyanov:
Ready for review :)

source/retryable-reads/tests/README.rst
DmitryLukyanov: done

source/retryable-reads/tests/README.rst
```diff
@@ -148,14 +148,17 @@ data.
 .. _GridFSBucket spec: https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.rst#configurable-gridfsbucket-class
     
 Speeding Up Tests
------------------
+===================
```
jmikola, line 151: This should line up with the heading text, but I also don't see any reason this needs to change. The README already had a section for this, so we can leave it at its current nesting level, since this applies within the "Network Error Tests" section.

source/transactions/tests/README.rst
```diff
@@ -77,6 +77,11 @@ control the fail point's behavior. ``failCommand`` supports the following
   blocked for. Required when blockConnection is true.
   `New in mongod 4.3.4 <https://jira.mongodb.org/browse/SERVER-41070>`_.
 
+Speeding Up Tests
+=================
+
+Drivers should set the default heartbeatFrequencyMS to 5ms in order to take into account the latest changes regarding streaming protocol. If a test has an explicit heartbeatFrequencyMS value, drivers should use the explicit value.
```
jmikola, line 83: Reading this alongside the existing paragraph on tweaking `minHeartbeatFrequencyMS` in Retryable Reads, how would drivers be able to tweak `heartbeatFrequencyMS` if `minHeartbeatFrequencyMS` is already [fixed at 500ms](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-monitoring.rst#minheartbeatfrequencyms)?

I think this text should state that drivers manually lower `minHeartbeatFrequencyMS` if possible _and_ reduce `heartbeatFrequencyMS` to that same value -- with an exception if the test already has an explicit `heartbeatFrequencyMS` value. You can suggest 5ms for both if you wish.

That said, I'm not sure why we should bother adding anything to Transactions if we already have a section in the Retryable Reads spec. Perhaps just have everything in this PR refer to the Retryable Read section instead (and modify it as you like).

source/transactions/tests/README.rst
```diff
@@ -640,3 +645,4 @@ Changelog
 :2019-02-13: Modify test format for 4.2 sharded transactions, including
              "useMultipleMongoses", ``object: testRunner``, the
              ``targetedFailPoint`` operation, and recoveryToken assertions.
+:2020-06-12: Add the default value for heartbeatFrequencyMS.
```
jmikola, line 648: I don't think any of these additions need a changelog entry, since we're just provide some advice and this not anything that requires driver changes.

source/retryable-reads/tests/README.rst
```diff
@@ -148,14 +148,17 @@ data.
...
 
 Drivers may benefit reducing `minHeartbeatFrequencyMS`_ in order to speed up
 tests. Python was able to decrease the run time of the tests greatly by lowering
 the SDAM's ``minHeartbeatFrequencyMS`` from 500ms to 50ms, thus decreasing the
 waiting time after a "not master" error:
 
 .. _minHeartbeatFrequencyMS: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#minheartbeatfrequencyms
+
+See also `Speeding Up Tests </source/transactions/tests/README.rst#speeding-up-tests>`_ in the Transactions spec tests.
+
```
jmikola, line 161: Since this document already talks about adjusting `minHeartbeatFrequencyMS`, I'd propose you modify it _and_ also suggest that drivers use a default `heartbeatFrequencyMS` value to match (which was never discussed here, but probably should have been).

source/transactions/tests/README.rst
ShaneHarvey: Agreed.

source/transactions/tests/README.rst
ShaneHarvey: > I think this text should state that drivers manually lower minHeartbeatFrequencyMS if possible and reduce heartbeatFrequencyMS to that same value -- with an exception if the test already has an explicit heartbeatFrequencyMS value. You can suggest 5ms for both if you wish.

Agreed. @DmitryLukyanov can you confirm that reducing both these values to 5ms does not cause any tests to fail?

DmitryLukyanov: @ShaneHarvey , I can confirm this: https://evergreen.mongodb.com/version/5f5ffde90ae6064926c283ff. 

`heartbeatFrequencyMS` has been reduced during the last month, for `minHeartbeatFrequencyMS`, I needed to make a code change. So, I'm not sure how strong required reducing minHeartbeatFrequencyMS, I would say not really

ShaneHarvey: Thanks for testing that. I second Jermey's request that we mention Drivers may reduce `minHeartbeatFrequencyMS`as well (if possible in the language). Note that we will not require drivers to reduce `minHeartbeatFrequencyMS`, nor force drivers to make `minHeartbeatFrequencyMS` configurable by users (because it shouldn't be). We can just suggest it as a possible way to speed up slow tests.

source/retryable-reads/tests/README.rst
jmikola: @DmitryLukyanov ping

source/retryable-reads/tests/README.rst
```diff
@@ -171,3 +174,5 @@ Changelog
              now expressed within ``runOn`` elements.
 
              Add test-level ``useMultipleMongoses`` field.
+
+:2020-06-12: Add the default value for heartbeatFrequencyMS.
```
jmikola, line 178: This change description doesn't match up with the diff. I'd suggest:

> Suggest lowering heartbeatFrequencyMS in addition to minHeartbeatFrequencyMS

And please update the date.

source/retryable-writes/tests/README.rst
```diff
@@ -91,6 +91,11 @@ disabled like so::
         mode: "off"
     });
 
+Speeding Up Tests
+=================
+
+See `Speeding Up Tests </source/transactions/tests/README.rst#speeding-up-tests>`_ in the Transactions spec tests.
```
jmikola, line 97: The retryable reads spec already has the most verbose explanation. Please link to that instead.

Also, the CWD of this file is `source/retryable-writes/tests/`, so the URL will likely need to be `../../retryable-reads/tests/README.rst#speeding-up-tests`.

source/retryable-writes/tests/README.rst
```diff
@@ -337,3 +342,5 @@ Changelog
              which are now expressed within ``runOn`` elements.
 
              Add test-level ``useMultipleMongoses`` field.
+
+:2020-06-12: Add the default value for heartbeatFrequencyMS.
```
jmikola, line 346: Per my [comment](https://github.com/mongodb/specifications/pull/820/files#r458261818) in the transactions README, this can be removed.

source/transactions/tests/README.rst
jmikola: Please make the changes in my earlier comment (https://github.com/mongodb/specifications/pull/820/files#r458262676) for retryable reads and change this to refer to that spec. I think the content can match what I suggested for retryable writes, which is just a reference to the retryable reads section.

DmitryLukyanov: done

DmitryLukyanov:
NOTE: the new reviewers were added automatically. I didn't add them intentionally.

source/retryable-reads/tests/README.rst
```diff
@@ -148,16 +148,17 @@ data.
...
-===================
+-----------------
 
 Drivers may benefit reducing `minHeartbeatFrequencyMS`_ in order to speed up
 tests. Python was able to decrease the run time of the tests greatly by lowering
 the SDAM's ``minHeartbeatFrequencyMS`` from 500ms to 50ms, thus decreasing the
 waiting time after a "not master" error:
 
-.. _minHeartbeatFrequencyMS: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#minheartbeatfrequencyms
+Also, similar improvement can be added for `heartbeatFrequencyMS`_ by setting the default heartbeatFrequencyMS to 5ms in order to take into account the latest changes regarding streaming protocol. If a test has an explicit heartbeatFrequencyMS value, drivers should use the explicit value.
```
jmikola, line 158: The original paragraph still states 50ms for `minHeartbeatFrequencyMS`, which contradicts 5ms here.

I propose consolidating both paragraphs into one:

```
Drivers can greatly reduce the execution time of tests by setting `heartbeatFrequencyMS`_
and `minHeartbeatFrequencyMS`_ (internally) to a small value (e.g. 5ms), below what
is normally permitted in the SDAM spec. If a test specifies an explicit value for
heartbeatFrequencyMS (e.g. client or URI options), drivers SHOULD use that value.
```

source/retryable-reads/tests/README.rst
```diff
@@ -148,16 +148,17 @@ data.
...
 
-See also `Speeding Up Tests </source/transactions/tests/README.rst#speeding-up-tests>`_ in the Transactions spec tests.
+.. _minHeartbeatFrequencyMS: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#minheartbeatfrequencyms	
+.. _heartbeatFrequencyMS: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#heartbeatfrequencyms
```
jmikola, line 161: This can also use `../../server-discovery-and-monitoring/server-discovery-and-monitoring.rst#` as a base (before adding the anchors). No need to use absolute URLs.

source/retryable-reads/tests/README.rst
jmikola: I don't think the anecdote about the Python driver using 50ms is particular relevant anymore, so I have no qualms about removing that. Also, I think "(internally)" is sufficient to explain that we're doing something that wouldn't be permitted by a driver's public API.

cc: @ShaneHarvey

DmitryLukyanov: sgtm, done

source/retryable-reads/tests/README.rst
DmitryLukyanov: done

jmikola (approved):
@isabelatkinson @durran I think your LGTMs may be required as code owners of transactions, but the changes are quite insignificant so it should be quick.

isabelatkinson (approved):

source/retryable-reads/tests/README.rst
```diff
@@ -150,12 +150,14 @@ data.
...
 -----------------
 
-Drivers may benefit reducing `minHeartbeatFrequencyMS`_ in order to speed up
-tests. Python was able to decrease the run time of the tests greatly by lowering
-the SDAM's ``minHeartbeatFrequencyMS`` from 500ms to 50ms, thus decreasing the
-waiting time after a "not master" error:
+Drivers can greatly reduce the execution time of tests by setting `heartbeatFrequencyMS`_
+and `minHeartbeatFrequencyMS`_ (internally) to a small value (e.g. 5ms), below what
+is normally permitted in the SDAM spec. If a test specifies an explicit value for
+heartbeatFrequencyMS (e.g. client or URI options), drivers SHOULD use that value.
```
ShaneHarvey, line 156: drivers SHOULD use that value -> drivers MUST use that value

DmitryLukyanov: done

ShaneHarvey (approved):

durran (approved):

PR #821, opened 2020-06-15
SPEC-1684: Testing Atlas Data Lake with Drivers

jstewart-mongo:
https://jira.mongodb.org/browse/SPEC-1684

This PR includes a README for the tests provided. The tests cover all of the commands found in the scope as well as the commands used as arguments to `runCommand`. In addition, there are connectivity tests for SCRAM authentication.

### DIFF
--- a/source/atlas-data-lake-testing/tests/README.rst
+++ b/source/atlas-data-lake-testing/tests/README.rst
@@ -1,0 +1,51 @@
+=====================
+Atlas Data Lake Tests
+=====================
+
+.. contents::
+
+----
+
+Introduction
+============
+
+The YAML and JSON files in this directory tree are platform-independent tests
+that drivers can use to assert compatibility with `Atlas Data Lake <https://docs.mongodb.com/datalake>`_.
+
+Running these integration tests will require a running ``mongohoused``
+with data available in its ``test.driverdata`` collection. See the
+`ADL directory in drivers-evergreen-tools <https://github.com/mongodb-labs/drivers-evergreen-tools/tree/master/.evergreen/atlas_data_lake>`_
+and `10gen/mongohouse README <https://github.com/10gen/mongohouse/blob/master/README.md>`_
+for more information.
+
+Several prose tests, which are not easily expressed in YAML, are also presented
+in this file. Those tests will need to be manually implemented by each driver.
+
+Test Format
+===========
+
+The same as the `CRUD Spec Test format <../../crud/tests/README.rst#Test-Format>`_.
+
+Prose Tests
+===========
+
+The following tests MUST be implemented to fully test compatibility with
+Atlas Data Lake.
+
+#. Test that the ``killCursors`` command works with Atlas Data Lake.
+   For this test, start a query but do not fully iterate over the results
+   (e.g. specify batchSize=2 for a query that would match 3+ documents).
+   Execute a ``killCursors`` command before executing a ``getMore`` command.
+   Continue iterating the cursor to trigger a getMore command and assert
+   that ADL raises an error.
+   
+#. Test that the driver can establish a connection with Atlas Data Lake
+   without authentication. For these tests, create a MongoClient using a
+   valid connection string without auth credentials and execute a ping
+   command.
+
+#. Test that the driver can establish a connection with Atlas Data Lake
+   with authentication. For these tests, create a MongoClient using a
+   valid connection string with SCRAM-SHA-1 and credentials from the
+   drivers-evergreen-tools ADL configuration and execute a ping command.
+   Repeat this test using SCRAM-SHA-256.
--- a/source/atlas-data-lake-testing/tests/README.rst
+++ b/source/atlas-data-lake-testing/tests/README.rst
@@ -32,13 +32,28 @@ Prose Tests
 The following tests MUST be implemented to fully test compatibility with
 Atlas Data Lake.
 
-#. Test that the ``killCursors`` command works with Atlas Data Lake.
-   For this test, start a query but do not fully iterate over the results
-   (e.g. specify batchSize=2 for a query that would match 3+ documents).
-   Execute a ``killCursors`` command before executing a ``getMore`` command.
-   Continue iterating the cursor to trigger a getMore command and assert
-   that ADL raises an error.
-   
+#. Test that the driver properly constructs and issues a
+   `killCursors <https://docs.mongodb.com/manual/reference/command/killCursors/>`_
+   command to Atlas Data Lake. For this test, configure an APM listener on a
+   client and execute a query that will leave a cursor open on the server (e.g.
+   specify ``batchSize=2`` for a query that would match 3+ documents). Drivers
+   MAY iterate the cursor if necessary to execute the initial ``find`` command
+   but MUST NOT iterate further to avoid executing a ``getMore``.
+
+   Observe the CommandSucceededEvent event for the ``find`` command and extract
+   the cursor's ID and namespace from the response document's ``cursor.id`` and
+   ``cursor.ns`` fields, respectively. Destroy the cursor object and observe
+   a CommandStartedEvent and CommandSucceededEvent for the ``killCursors``
+   command. Assert that the cursor ID and target namespace in the outgoing
+   command match the values from the ``find`` command's CommandSucceededEvent.
+   Assert that the ``killCursors`` CommandSucceededEvent indicates that the
+   expected cursor was killed in the ``cursorsKilled`` field.
+
+   Note: this test assumes that drivers only issue a ``killCursors`` command
+   internally when destroying a cursor that may still exist on the server. If
+   a driver constructs and issues ``killCursors`` commands in other ways (e.g.
+   public API), this test MUST be adapted to test all such code paths.
+
 #. Test that the driver can establish a connection with Atlas Data Lake
    without authentication. For these tests, create a MongoClient using a
    valid connection string without auth credentials and execute a ping
--- a/source/atlas-data-lake-testing/tests/aggregate.yml
+++ b/source/atlas-data-lake-testing/tests/aggregate.yml
@@ -1,0 +1,19 @@
+collection_name: &collection_name "driverdata"
+database_name: &database_name "test"
+
+tests:
+-
+  description: "Aggregate with pipeline (project, sort, limit)"
+  operations:
+    -
+      object: collection
+      name: aggregate
+      arguments:
+        pipeline:
+          - $project: { _id: 0 }
+          - $sort: { a: 1 }
+          - $limit: 2
+  outcome:
+    result:
+      - { a: 1, b: 2, c: 3 }
+      - { a: 2, b: 3, c: 4 }
--- a/source/atlas-data-lake-testing/tests/getMore.yml
+++ b/source/atlas-data-lake-testing/tests/getMore.yml
@@ -1,0 +1,33 @@
+collection_name: &collection_name "driverdata"
+database_name: &database_name "test"
+
+tests:
+  -
+    description: "A successful find event with getMore"
+    operations:
+      -
+        object: collection
+        name: find
+        arguments:
+          filter: { a: { $gte: 2 }}
+          sort: { a: 1 }
+          batchSize: 3
+          limit: 4
+    expectations:
+      -
+        command_started_event:
+          command:
+            find: *collection_name
+            filter: { a: { $gte : 2 }}
+            sort: { a: 1 }
+            batchSize: 3
+            limit: 4
+          command_name: "find"
+          database_name: *database_name
+      -
+        command_started_event:
+          command:
+            batchSize: 1
+          command_name: "getMore"
+          # mongohoused always expects getMores on the "cursors" database
+          database_name: "cursors"
--- a/source/atlas-data-lake-testing/tests/getMore.yml
+++ b/source/atlas-data-lake-testing/tests/getMore.yml
@@ -27,6 +27,8 @@ tests:
       -
         command_started_event:
           command:
+            # Assertions for "getMore" and "collection" fields are omitted, as they will have arbitrary values
+            # TODO: Those assertions can be added after wildcard matchers are available via SPEC-1215
             batchSize: 1
           command_name: "getMore"
           # mongohoused always expects getMores on the "cursors" database
--- a/source/atlas-data-lake-testing/tests/listCollections.yml
+++ b/source/atlas-data-lake-testing/tests/listCollections.yml
@@ -1,0 +1,17 @@
+database_name: &database_name "test"
+
+tests:
+  -
+    description: "ListCollections succeeds"
+    operations:
+      -
+        name: listCollections
+        object: database
+    expectations:
+      -
+        command_started_event:
+          command_name: "listCollections"
+          database_name: *database_name
+          command:
+            listCollections: 1
+
--- a/source/atlas-data-lake-testing/tests/listDatabases.yml
+++ b/source/atlas-data-lake-testing/tests/listDatabases.yml
@@ -1,0 +1,15 @@
+tests:
+  -
+    description: "ListDatabases succeeds"
+    operations:
+      -
+        name: listDatabases
+        object: client
+    expectations:
+      -
+        command_started_event:
+          command_name: "listDatabases"
+          database_name: "admin"
+          command:
+            listDatabases: 1
+
--- a/source/atlas-data-lake-testing/tests/runCommand.yml
+++ b/source/atlas-data-lake-testing/tests/runCommand.yml
@@ -1,0 +1,20 @@
+database_name: &database_name "test"
+
+tests:
+  -
+    description: "ping succeeds using runCommand"
+    operations:
+      -
+        name: runCommand
+        object: database
+        command_name: ping
+        arguments:
+          command:
+            ping: 1
+    expectations:
+      -
+        command_started_event:
+          command_name: ping
+          database_name: *database_name
+          command:
+            ping: 1
--- a/source/atlas-data-lake-testing/tests/aggregate.json
+++ b/source/atlas-data-lake-testing/tests/aggregate.json
@@ -1,0 +1,46 @@
+{
+  "collection_name": "driverdata",
+  "database_name": "test",
+  "tests": [
+    {
+      "description": "Aggregate with pipeline (project, sort, limit)",
+      "operations": [
+        {
+          "object": "collection",
+          "name": "aggregate",
+          "arguments": {
+            "pipeline": [
+              {
+                "$project": {
+                  "_id": 0
+                }
+              },
+              {
+                "$sort": {
+                  "a": 1
+                }
+              },
+              {
+                "$limit": 2
+              }
+            ]
+          }
+        }
+      ],
+      "outcome": {
+        "result": [
+          {
+            "a": 1,
+            "b": 2,
+            "c": 3
+          },
+          {
+            "a": 2,
+            "b": 3,
+            "c": 4
+          }
+        ]
+      }
+    }
+  ]
+}
--- a/source/atlas-data-lake-testing/tests/count.json
+++ b/source/atlas-data-lake-testing/tests/count.json
@@ -1,0 +1,25 @@
+{
+  "collection_name": "driverdata",
+  "database_name": "test",
+  "tests": [
+    {
+      "description": "estimatedDocumentCount succeeds",
+      "operations": [
+        {
+          "object": "collection",
+          "name": "estimatedDocumentCount",
+          "result": 15
+        }
+      ],
+      "expectations": [
+        {
+          "command_started_event": {
+            "command": {
+              "count": "driverdata"
+            }
+          }
+        }
+      ]
+    }
+  ]
+}
--- a/source/atlas-data-lake-testing/tests/count.yml
+++ b/source/atlas-data-lake-testing/tests/count.yml
@@ -1,0 +1,16 @@
+collection_name: &collection_name "driverdata"
+database_name: &database_name "test"
+
+tests:
+  -
+    description: "estimatedDocumentCount succeeds"
+    operations:
+      -
+        object: collection
+        name: estimatedDocumentCount
+        result: 15
+    expectations:
+      -
+        command_started_event:
+          command:
+            count: *collection_name
--- a/source/atlas-data-lake-testing/tests/find.json
+++ b/source/atlas-data-lake-testing/tests/find.json
@@ -1,0 +1,67 @@
+{
+  "collection_name": "driverdata",
+  "database_name": "test",
+  "tests": [
+    {
+      "description": "Find with projection and sort",
+      "operations": [
+        {
+          "object": "collection",
+          "name": "find",
+          "arguments": {
+            "filter": {
+              "b": {
+                "$gt": 5
+              }
+            },
+            "projection": {
+              "_id": 0
+            },
+            "sort": {
+              "a": 1
+            },
+            "limit": 5
+          }
+        }
+      ],
+      "expectations": [
+        {
+          "command_started_event": {
+            "command": {
+              "find": "driverdata"
+            }
+          }
+        }
+      ],
+      "outcome": {
+        "result": [
+          {
+            "a": 5,
+            "b": 6,
+            "c": 7
+          },
+          {
+            "a": 6,
+            "b": 7,
+            "c": 8
+          },
+          {
+            "a": 7,
+            "b": 8,
+            "c": 9
+          },
+          {
+            "a": 8,
+            "b": 9,
+            "c": 10
+          },
+          {
+            "a": 9,
+            "b": 10,
+            "c": 11
+          }
+        ]
+      }
+    }
+  ]
+}
--- a/source/atlas-data-lake-testing/tests/find.yml
+++ b/source/atlas-data-lake-testing/tests/find.yml
@@ -1,0 +1,27 @@
+collection_name: &collection_name "driverdata"
+database_name: &database_name "test"  
+
+tests:
+  -
+    description: "Find with projection and sort"
+    operations:
+      -
+        object: collection
+        name: find
+        arguments:
+          filter: { b: { $gt: 5 } }
+          projection: { _id: 0 }
+          sort: { a: 1 }
+          limit: 5
+    expectations:
+      -
+        command_started_event:
+          command:
+            find: *collection_name
+    outcome:
+      result:
+        - {"a": 5, "b": 6, "c": 7}
+        - {"a": 6, "b": 7, "c": 8}
+        - {"a": 7, "b": 8, "c": 9}
+        - {"a": 8, "b": 9, "c": 10}
+        - {"a": 9, "b": 10, "c": 11}
--- a/source/atlas-data-lake-testing/tests/getMore.json
+++ b/source/atlas-data-lake-testing/tests/getMore.json
@@ -1,0 +1,57 @@
+{
+  "collection_name": "driverdata",
+  "database_name": "test",
+  "tests": [
+    {
+      "description": "A successful find event with getMore",
+      "operations": [
+        {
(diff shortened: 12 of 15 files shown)


jyemin:
Can you provide a link to your POC of these tests?

jstewart-mongo:
> Can you provide a link to your POC of these tests?

https://github.com/jstewart-mongo/mongo-java-driver/tree/spec1684

source/adl-standalone/tests/aggregate.yml
```diff
@@ -0,0 +1,292 @@
+database_name: "test"
+
+collection_name: &collection_name 'quux'
+
+tests:
```
jyemin, line 5: Similarly, should we be testing all options for all of these commands, e.g. for aggregate:

* batchSize
* maxTimeMS
* collation
* comment

And similarly for all of the other commands.

source/adl-standalone/tests/connection.yml
```diff
@@ -0,0 +1,42 @@
+tests:
```
jyemin, line 1: This test format is not described in the README.  

I'm also not sure what the point of it is, as they look to be just unit tests.

jyemin: If the intention is that they are integration tests, we would have to say what to do in the integration test.   Presumably it would involve running at least one command for each, and either getting an exception or not depending on whether it's valid or not.

source/adl-standalone/tests/getMore.yml
```diff
@@ -0,0 +1,34 @@
+collection_name: &collection_name "quux"
+database_name: &database_name "test"
+namespace: &namespace "test.quux"
```
jmikola, line 3: `namespace` is neither used in this test file, nor is it defined anywhere in the CRUD spec test format. Please strike this from any test file in this PR.

source/adl-standalone/tests/getMore.yml
```diff
@@ -0,0 +1,34 @@
...
+            batchSize: 3
+          command_name: "find"
+          database_name: *database_name
+      -
+        command_started_event:
+          command:
+            getMore: { $numberLong: "42" }
+            batchSize: 1
+          command_name: "getMore"
+          database_name: "cursors"
```
jmikola, line 34: "cursors" looks to be incorrect, as `database_name` is defined as "test" at the top of this file.

source/adl-standalone/tests/runCommand.yml
```diff
@@ -0,0 +1,54 @@
+tests:
+  -
+    description: "ListCollections succeeds using runCommand"
```
jmikola, line 3: The [scope](https://docs.google.com/document/d/1eXwy-7v_7BlPpcr7emXWSJZWm4V47ZuH5lvBj-XnOmY/edit#) only stated that:

> Drivers should be able to execute a command through existing driver runCommand helper method

@jyemin: Was the intention of that line to say that each of the commands tested by explicit helper methods should also be run through `runCommand` (as is the case here), or would it be sufficient to execute a single command (perhaps `ping` if that's supported).

cc: @craiggwilson

source/adl-standalone/tests/connection.yml
jmikola: I don't think these tests add much value. To relate this back to the [scope](https://docs.google.com/document/d/1eXwy-7v_7BlPpcr7emXWSJZWm4V47ZuH5lvBj-XnOmY/edit#), we had the following requirements:

> * Drivers should be able to connect to ADL. ADL pretends to be a standalone MongoDB server.
> * Drivers should be able to authenticate using both SCRAM-SHA-1 and SCRAM-SHA-256.

I don't think the first bullet needs its own test, as the CRUD tests in this PR adequately cover that.

I would propose adding two prose tests for drivers to manually configure the appropriate authentication for mongohouse and then connect to it and issue a single command to verify that authentication was successful (e.g. `ping` if that's allowed).

https://github.com/10gen/mongohouse/blob/master/README.md#getting-started states that "All the sample config files have authentication enabled", but I didn't find reference to auth mechanisms in the repo apart from some files in the [`internal/` directory](https://github.com/10gen/mongohouse/search?q=SCRAM-SHA-256&unscoped_q=SCRAM-SHA-256). @craiggwilson: Can you chime in here? Are there documented instructions for configuring mongohouse with SCRAM-SHA-1 and SCRAM-SHA-256?

craiggwilson: The auth for the sample/used configs all have auth just built-in. Nothing needs to happen other than to use one of the config files.

source/adl-standalone/tests/README.rst
```diff
@@ -0,0 +1,133 @@
...
+
+Running these integration tests will require a running ``mongohoused``. Requirements
+for building ``mongohoused`` are as follows:
+
+  - Go version 1.13
+  - Building ``mqlrun``
+
+See the README file at ``https://github.com/10gen/mongohouse`` for more information.
+
+Building ``mongohoused`` locally
```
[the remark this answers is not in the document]
craiggwilson: The intent was to have a `driver-evergreen-tools` script that drivers could just use. All the build instructions are great, but simply doing this similar to atlas-proxy is by far a better setup. We do have artifacts that can be downloaded from S3 for both mongohoused and mqlrun. Building from master is fairly straightforward, which could be put into a driver-evergreen-tools script. Either way can be fine.

source/adl-standalone/tests/connection.yml
jmikola: Ah, I missed the link to https://github.com/10gen/mongohouse/tree/master/testdata/config in https://github.com/10gen/mongohouse/blob/master/README.md#configuration.

So, to confirm, if our tests use [single-local-local.yaml](https://github.com/10gen/mongohouse/blob/master/testdata/config/single-local-local.yaml) for everything, we can trust that this has _both_ SCRAM-SHA-1 _and_ SCRAM-SHA-256 available for `mhuser:pencil` based on https://github.com/10gen/mongohouse/blob/4e6c3c13b085545e573ba2e41e66ae7de03c8211/testdata/config/single-local-local.yaml#L35?

@jstewart-mongo: I see that your "To build ``mongohoused`` for local testing" section already loads this configuration, so if you do get rid of that section (per my earlier comment about build instructions being redundant if we provision through drivers-evergreen-tools scripts), I think you'll still want to leave in the bit about how to launch `mongohoused` with the appropriate configuration file.

source/adl-standalone/tests/README.rst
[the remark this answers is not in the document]
jmikola: I think relying on scripts to provision existing build artifacts is preferable to duplicating the build process for this project, so I'd like to see the original scope goal implemented. I'll leave it to @jstewart-mongo to sort that out with your team in the event he needs some pointers to the S3 sources.

source/adl-standalone/tests/connection.yml
craiggwilson: Yes, if you use `single-local-local.yaml`, then it has that user attached. 

That being said, I'd highly suggest creating your own config and referencing your own "data" files. We on ADL may change this config file and or the data files it points to for our own tests every once in a while and that will definitely break this. Should be easy enough to copy `single-local-local` and some of the tests into a new place and just use those.

jmikola: I concur, it seems most reasonable to have drivers-evergreen-tools contain its own ADL configuration _and_ a script to provision ADL from your S3 bucket.

source/adl-standalone/tests/README.rst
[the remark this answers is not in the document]
jstewart-mongo: I have created a script for `driver-evergreen-tools` for this. The PR is located at https://github.com/jstewart-mongo/drivers-evergreen-tools/pull/1.

source/adl-standalone/tests/README.rst
```diff
@@ -0,0 +1,133 @@
+=====================
+Atlas Data Lake Tests
```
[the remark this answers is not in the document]
jmikola: Was that commit not pushed? I still see `adl-standaloen` in the PR.

source/adl-standalone/tests/README.rst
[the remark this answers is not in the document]
jmikola: I received an invite to your fork (not sure why it's private since https://github.com/mongodb-labs/drivers-evergreen-tools is already public), but I would have expected a PR against https://github.com/mongodb-labs/drivers-evergreen-tools directly. That way, any review stays on the common repository and it can be merged directly once approved (vs. presumably having to open a new PR after reviewing on your own fork).

source/adl-standalone/tests/runCommand.yml
jyemin: I single command.

source/atlas-data-lake-testing/tests/crud/listDatabases.yml
```diff
@@ -0,0 +1,13 @@
+tests:
```
jmikola, line 1: Note: this file can omit `database_name` and `collection_name`, since it won't be used for anything.

source/atlas-data-lake-testing/tests/connection/connection.yml
```diff
@@ -0,0 +1,42 @@
+tests:
+  -
+    description: "should use the default test database"
+    uri: "mongodb://mhuser:pencil@localhost"
+    valid: true
```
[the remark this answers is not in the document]
craiggwilson: yes, ADL implements the ping command.

source/adl-standalone/tests/runCommand.yml
jmikola: Assuming `ping` can be used, this would probably address most of my existing comments on the revised `runCommand.yml` test file. I'll leave this comment thread open pending @jstewart-mongo's response, even though the original target here is outdated.

source/atlas-data-lake-testing/tests/crud/getMore.yml
```diff
@@ -0,0 +1,32 @@
...
+            limit: 4
+            batchSize: 3
+          command_name: "find"
+          database_name: *database_name
+      -
+        command_started_event:
+          command:
+            batchSize: 1
+          command_name: "getMore"
+          database_name: "cursors"
```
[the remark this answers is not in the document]
jmikola: Looks like this typo persisted from my previous review: https://github.com/mongodb/specifications/pull/821#discussion_r440966275

source/adl-standalone/tests/README.rst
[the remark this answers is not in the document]
jmikola: I'll resolve this as I opened a new comment thread on the revised README in this PR (and did suggest referencing the drivers-evergreen-tools repo).

source/adl-standalone/tests/runCommand.yml
jmikola: Ah, Craig confirmed that `ping` is supported in https://github.com/mongodb/specifications/pull/821#discussion_r451174446.

source/atlas-data-lake-testing/tests/connection/connection.yml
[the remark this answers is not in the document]
jyemin: If we go with prose tests, we can configure all the URIs as project properties, similar to Atlas connectivity tests.  I don't think we need any of the ones where valid is false.  They don't add much value IMO.

source/atlas-data-lake-testing/tests/crud/runCommand.yml
```diff
@@ -0,0 +1,54 @@
...
+      -
+        name: runCommand
+        arguments:
+          command: { aggregate: "driverdata", pipeline: [ { $group: { _id: 1, n: { $sum: 1 } } }, { $sort: { a: 1 } } ], cursor: { } }
+        object: collection
+    expectations:
+      -
+        command_started_event:
+          command:
+            aggregate: "driverdata"
```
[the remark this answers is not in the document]
jstewart-mongo: I have updated these tests.

source/atlas-data-lake-testing/tests/crud/getMore.yml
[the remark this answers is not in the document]
craiggwilson: Yes... that is currently what we send back. It might not always be what we send back, so I'm hesitant to suggest that you validate it. But for now, (and maybe forever), it is "cursors".

The collection name is autogenerated, so there isn't anyway to validate that.

source/atlas-data-lake-testing/tests/listDatabases.yml
```diff
@@ -1,13 +1,15 @@
...
-        name: listDatabaseNames
+        name: listDatabases
         object: client
     expectations:
       -
         command_started_event:
+          command_name: "listDatabases"
           command:
             listDatabases: 1
+            $db: "admin"
```
jmikola, line 14: Likewise, I suggest using `database_name` instead of the OP_MSG `$db` field since we already have that assertion available to us.

```yaml
command_name: "listCollections"
database_name: "admin"
command:
```

source/atlas-data-lake-testing/tests/getMore.yml
```diff
@@ -0,0 +1,32 @@
...
+            batchSize: 3
+            limit: 4
+          command_name: "find"
+          database_name: *database_name
+      -
+        command_started_event:
+          command:
+            batchSize: 1
+          command_name: "getMore"
+          database_name: "cursors"
```
jmikola, line 32: I think there's value in asserting this (as you referenced [RUBY-2211](https://jira.mongodb.org/browse/RUBY-2211)), but let's add a comment above to capture @craiggwilson's comment in https://github.com/mongodb/specifications/pull/821#discussion_r452493691. I suggest adding the following above this line:

```yaml
command_name: "getMore"
# mongohoused always expects getMores on the "cursors" database
database_name: "cursors"
```

source/atlas-data-lake-testing/tests/README.rst
```diff
@@ -12,14 +12,32 @@ Introduction
...
+
+#. Test that the ``killCursors`` command works with Atlas Data Lake.
+   For this test, start a query but do not fully iterate over the results.
+   Execute a ``killCursors`` command before executing a ``getMore`` command.
+   Ensure that Atlas Data Lake does not respond to the ``getMore`` command.
+
+#. Test that a connection is established with Atlas Data Lake. For these
+   tests, create a MongoClient using a valid connection string and execute
+   a ``ping`` command. Tests MUST also authenticate using both SCRAM-SHA-1
+   and SCRAM-SHA-256.
```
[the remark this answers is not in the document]
jstewart-mongo: Updated. I have prose tests that connect with the following URIs:
```
       mongodb://mhuser:pencil@localhost
       mongodb://localhost/
       mongodb://mhuser:pencil@localhost/test?authSource=admin
       mongodb://mhuser:pencil@localhost/?authMechanism=SCRAM-SHA-1&authSource=admin
       mongodb://mhuser:pencil@localhost/test?authMechanism=SCRAM-SHA-1&authSource=admin
       mongodb://mhuser:pencil@localhost/?authMechanism=SCRAM-SHA-256&authSource=admin
       mongodb://mhuser:pencil@localhost/test?authMechanism=SCRAM-SHA-256&authSource=admin
```

source/atlas-data-lake-testing/tests/getMore.yml
jstewart-mongo: I have added the comment.

source/atlas-data-lake-testing/tests/listCollections.yml
```diff
@@ -8,6 +10,8 @@ tests:
     expectations:
       -
         command_started_event:
+          command_name: "listCollections"
           command:
             listCollections: 1
+            $db: *database_name
```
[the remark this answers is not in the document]
jstewart-mongo: Updated

source/atlas-data-lake-testing/tests/listDatabases.yml
jstewart-mongo: Updated

jmikola:
@jstewart-mongo: FYI, I reviewed [9 hours ago](https://github.com/mongodb/specifications/pull/821#pullrequestreview-446095788).

source/atlas-data-lake-testing/tests/runCommand.yml
```diff
@@ -0,0 +1,82 @@
+collection_name: &collection_name "driverdata"
+database_name: &database_name "test"
+
+tests:
```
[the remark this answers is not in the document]
jstewart-mongo: Updated to one test using `ping` command.

source/atlas-data-lake-testing/tests/README.rst
```diff
@@ -0,0 +1,51 @@
...
+   Execute a ``killCursors`` command before executing a ``getMore`` command.
+   Continue iterating the cursor to trigger a getMore command and assert
+   that ADL raises an error.
+   
+#. Test that the driver can establish a connection with Atlas Data Lake
+   without authentication. For these tests, create a MongoClient using a
+   valid connection string without auth credentials and execute a ping
+   command.
+
+   Test that the driver can establish a connection with Atlas Data Lake
```
jmikola, line 47: I originally tried to write this all in one paragraph but it seemed clearer to split them up for non-auth and auth. I think this can get its own bullet with `#.` since it does repeat all of the necessary details and stands on its own.

source/atlas-data-lake-testing/tests/README.rst
```diff
@@ -12,14 +12,32 @@ Introduction
...
+Prose Tests
+===========
+
+The following tests MUST be implemented to fully test compatibility with
+Atlas Data Lake.
+
+#. Test that the ``killCursors`` command works with Atlas Data Lake.
+   For this test, start a query but do not fully iterate over the results.
+   Execute a ``killCursors`` command before executing a ``getMore`` command.
+   Ensure that Atlas Data Lake does not respond to the ``getMore`` command.
```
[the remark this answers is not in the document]
jmikola: I haven't seen the Java POC for this PR, but I don't think asserting `killCursors` was issued with APM would be sufficient to assert the last line that ADL won't respond to the `getMore` command. Wouldn't that require iterating the cursor _after_ `killCursors` is called and asserting that the driver gets some error?

source/atlas-data-lake-testing/tests/aggregate.yml
```diff
@@ -0,0 +1,16 @@
+tests:
```
jmikola, line 1: I think this comment was missed: https://github.com/mongodb/specifications/pull/821#discussion_r452606100

Please add:

```yaml
collection_name: &collection_name "driverdata"
database_name: &database_name "test"  
```

How was this passing in the Java driver? I doubt the CRUD spec test runner happens to use `test.driverdata` as its default namespace.

source/atlas-data-lake-testing/tests/runCommand.yml
```diff
@@ -0,0 +1,19 @@
+database_name: &database_name "test"
```
jmikola, line 1: `database_name` is no longer used in this file. You can either add `database_name: *database_name` to the APM assertion below for delete this line.

source/atlas-data-lake-testing/tests/README.rst
jstewart-mongo: Ok. I have added the bullet.

source/atlas-data-lake-testing/tests/aggregate.yml
jstewart-mongo: Not sure how it passes in the Java driver. I have added the collection and database names.

source/atlas-data-lake-testing/tests/runCommand.yml
jstewart-mongo: I have added the assertion.

jmikola:
Ah, forgot that https://github.com/mongodb/specifications/pull/821#discussion_r452600491 was still outstanding. I'll hold off on LGTM until we can address that.

source/atlas-data-lake-testing/tests/README.rst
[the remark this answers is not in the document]
jmikola: Ah, noted that the last sentence was changed to:

> Continue iterating the cursor to trigger a getMore command and assert that ADL raises an error.

That works for me. We don't need to specify exactly which error code is returned.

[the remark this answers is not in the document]
craiggwilson: ADL will return a cursor not found error (43).

source/adl-standalone/tests/aggregate.yml
jyemin: It's not.  But the scope just says: "and query".  So there is a lot of latitude. @craiggwilso, are you comfortable with the level of coverage we have now?

source/atlas-data-lake-testing/tests/README.rst
```diff
@@ -12,14 +12,32 @@ Introduction
...
-The keys found in the YAML files for the connection tests are a subset of the keys found in
-the `Test Plan for Connection Strings <../../connection-string/tests/README.rst#Format>`_.
 The keys found in the YAML files for the CRUD tests are a subset of the keys found in the
 `Test Plan for CRUD <../../crud/tests/README.rst#Test-Format>`_.
+
+Prose Tests
+===========
+
+The following tests MUST be implemented to fully test compatibility with
+Atlas Data Lake.
```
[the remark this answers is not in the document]
jyemin: ok

source/atlas-data-lake-testing/tests/README.rst
```diff
@@ -0,0 +1,51 @@
...
+
+The same as the `CRUD Spec Test format <../../crud/tests/README.rst#Test-Format>`_.
+
+Prose Tests
+===========
+
+The following tests MUST be implemented to fully test compatibility with
+Atlas Data Lake.
+
+#. Test that the ``killCursors`` command works with Atlas Data Lake.
```
jyemin, line 35: @jstewart-mongo I thought that this turned out to be difficult to implement, because the test won't have any way of knowing what namespace to specify when constructing a killCursors command manually.

jstewart-mongo: For the Java driver, the namespace needed to pass to `killCursors` is in the batch cursor, which is not accessible from `MongoCursor`, so the Java driver has to revert to closing the cursor and checking the command listener to check if `killCursors` was executed. For other drivers, if access to the batch cursor is available, then the driver can call `killCursors` directly with the included namespace from the batch cursor.

jyemin: @jmikola do you expect any drivers to be able to access the namespace like that?  I would think that for most it would be hidden inside a cursor object that is not accessible from tests.

jmikola: I glossed over "batch cursor" in John's original comments. Is that something specific to the Java driver?

I assumed on my earlier review that this would be like `getMore` commands, which always used the "cursors" database; however, that may not be the case and I remember now that [`killCursors`](https://docs.mongodb.com/manual/reference/command/killCursors/) also requires a collection name.

Is there a concrete example of what needs to happen here? For instance, if the driver executes a `find` on `test.driverdata`, what do we expect the returned response to `find` and the subsequent `killCursors` command to look like? I assume "batch cursor" is referring to details we'd observe in the response to `find`, so perhaps drivers could rely on APM to capture that if their cursor doesn't provide a public API to get the namespace we need for `killCursors`.

source/atlas-data-lake-testing/tests/getMore.yml
```diff
@@ -0,0 +1,33 @@
...
+            filter: { a: { $gte : 2 }}
+            sort: { a: 1 }
+            batchSize: 3
+            limit: 4
+          command_name: "find"
+          database_name: *database_name
+      -
+        command_started_event:
+          command:
+            batchSize: 1
```
jmikola, line 30: Let's add a comment here.

```
# Assertions for "getMore" and "collection" fields are omitted, as they will have arbitrary values
# TODO: This can be changed to use wildcard matchers when SPEC-1215 is implemented
```

source/atlas-data-lake-testing/tests/README.rst
jmikola: Let's add a new paragraph under this list item:

> Drivers that need to manually execute a `killCursors` command for this test should note that the namespace of the cursor returned by ADL (and expected for `killCursors` and `getMore` commands) will differ from the namespace of the original `find` operation. Drivers may need to utilize APM to capture the `find` command response, access this namespace from the `cursor.ns` field, and extract database and collection names to use for the `killCursors` command.

jmikola: Ah, @jyemin explained that this is less about testing ADL receives a `killCursors` command and more about the driver properly constructing any `killCursors` command it might send. I'm not aware of any drivers that have public APIs to send `killCursors`, so let's focus on the general case of a driver that sends it when a cursor object is destroyed and the driver cannot definitively conclude that the server has already closed it (e.g. already in the final batch, or iterated it to completion).

Given that, I think this prose test should do the following:

 * Execute `find` with the necessary options to guarantee that full iteration would require a `getMore`
 * Drivers can use APM to capture the `find` response and extract the namespace for the internal cursor (aka "batch cursor")
 * The test then destroys the cursor object and uses APM to assert that the driver sends a `killCursors` command with the appropriate namespace. They can also assert the response of this command to verify that the cursor was actually killed in ADL.
 * The driver can manually construct a `getMore` command and execute it with `runCommand` and expect an error, since the cursor will not exist. Manually executing `getMore` is likely necessary because the iterable cursor object will have been destroyed to accomplish the previous step of triggering `killCursors`.

This can be repeated/adapted for any code paths for which the driver might issue a `killCursors` command. I expect for most drivers, there's only one code path when destroying a cursor, but we can leave the wording open to refer to hypothetical public APIs for killing the cursor.

source/atlas-data-lake-testing/tests/README.rst
```diff
@@ -0,0 +1,66 @@
...
+   a CommandStartedEvent and CommandSucceededEvent for the ``killCursors``
+   command. Assert that the cursor ID and target namespace in the outgoing
+   command match the values from the ``find`` command's CommandSucceededEvent.
+   Assert that the ``killCursors`` CommandSucceededEvent indicates that the
+   expected cursor was killed in the ``cursorsKilled`` field.
+
+   Note: this test assumes that drivers only issue a ``killCursors`` command
+   internally when destroying a cursor that may still exist on the server. If
+   a driver constructs and issues ``killCursors`` commands in other ways (e.g.
+   public API), this test MUST be adapted to test all such code paths.
```
jmikola, line 55: This paragraph is my attempt to cover all the bases about other possible code paths where drivers may construct and issue `killCursors`.

source/atlas-data-lake-testing/tests/README.rst
```diff
@@ -0,0 +1,66 @@
...
+   but MUST NOT iterate further to avoid executing a ``getMore``.
+
+   Observe the CommandSucceededEvent event for the ``find`` command and extract
+   the cursor's ID and namespace from the response document's ``cursor.id`` and
+   ``cursor.ns`` fields, respectively. Destroy the cursor object and observe
+   a CommandStartedEvent and CommandSucceededEvent for the ``killCursors``
+   command. Assert that the cursor ID and target namespace in the outgoing
+   command match the values from the ``find`` command's CommandSucceededEvent.
+   Assert that the ``killCursors`` CommandSucceededEvent indicates that the
+   expected cursor was killed in the ``cursorsKilled`` field.
```
jmikola, line 50: My [previous comment](https://github.com/mongodb/specifications/pull/821#discussion_r453078602) referred to executing a manual `getMore`, but I actually don't think that's relevant to drivers. Inspecting the CommandSucceededEvent for `killCursors` is sufficient to assert that ADL killed the cursor.

Also, if a driver did manually execute a `getMore` and got the namespace/ID wrong, the end result would be no different than if we got it right and the cursor was successfully killed.

PR #822, opened 2020-06-15
SPEC-1708 Retry replSetStepDown in SDAM and connection failover tests

ShaneHarvey:
Jira: [SPEC-1708](https://jira.mongodb.org/browse/SPEC-1708)

### DIFF
--- a/source/connections-survive-step-down/tests/README.rst
+++ b/source/connections-survive-step-down/tests/README.rst
@@ -83,7 +83,7 @@ transient error (see `SERVER-48154`_)::
 
 When running the "replSetStepDown" command, drivers MUST retry until the
 command succeeds. The number of retries should be limited to avoid an infinite
-failure loop.
+failure loop. For example, the Python driver uses a 10 second retry period.
 
 Not Master - Keep Connection Pool
 `````````````````````````````````
--- a/source/connections-survive-step-down/tests/README.rst
+++ b/source/connections-survive-step-down/tests/README.rst
@@ -71,8 +71,19 @@ Perform the following operations:
 - If the driver implements the `CMAP`_ specification, verify that no new `PoolClearedEvent`_ has been
   published. Otherwise verify that `connections.totalCreated`_ in `serverStatus`_ has not changed.
 
+**Note:** The "replSetStepDown" command often fails with the following
+transient error (see `SERVER-48154`_)::
 
+  {
+    "ok" : 0,
+    "errmsg" : "Unable to acquire X lock on '{4611686018427387905: ReplicationStateTransition, 1}' within 1000ms. opId: 922, op: conn30, connId: 30.",
+    "code" : 24,
+    "codeName" : "LockTimeout",
+  }
 
+When running the "replSetStepDown" command, drivers MUST retry until the
+command succeeds. The number of retries should be limited to avoid an infinite
+failure loop.
 
 Not Master - Keep Connection Pool
 `````````````````````````````````
--- a/source/connections-survive-step-down/tests/README.rst
+++ b/source/connections-survive-step-down/tests/README.rst
@@ -186,3 +186,4 @@ server communication.
 .. _PoolClearedEvent: /source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#events
 .. _serverStatus: https://docs.mongodb.com/manual/reference/command/serverStatus
 .. _connections.totalCreated: https://docs.mongodb.com/manual/reference/command/serverStatus/#serverstatus.connections.totalCreated
+.. _SERVER-48154: https://jira.mongodb.org/browse/SERVER-48154
--- a/source/server-discovery-and-monitoring/tests/README.rst
+++ b/source/server-discovery-and-monitoring/tests/README.rst
@@ -302,6 +302,20 @@ MongoClient from the one used for test operations. For example::
             replSetStepDown: 20
             force: false
 
+**Note:** The "replSetStepDown" command often fails with the following
+transient error (see `SERVER-48154`_)::
+
+  {
+    "ok" : 0,
+    "errmsg" : "Unable to acquire X lock on '{4611686018427387905: ReplicationStateTransition, 1}' within 1000ms. opId: 922, op: conn30, connId: 30.",
+    "code" : 24,
+    "codeName" : "LockTimeout",
+  }
+
+When running the "replSetStepDown" command, drivers MUST retry until the
+command succeeds. The number of retries should be limited to avoid an infinite
+failure loop.
+
 waitForPrimaryChange
 ''''''''''''''''''''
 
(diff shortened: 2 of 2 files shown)


source/connections-survive-step-down/tests/README.rst
```diff
@@ -71,8 +71,19 @@ Perform the following operations:
...
 
+  {
+    "ok" : 0,
+    "errmsg" : "Unable to acquire X lock on '{4611686018427387905: ReplicationStateTransition, 1}' within 1000ms. opId: 922, op: conn30, connId: 30.",
+    "code" : 24,
+    "codeName" : "LockTimeout",
+  }
 
+When running the "replSetStepDown" command, drivers MUST retry until the
+command succeeds. The number of retries should be limited to avoid an infinite
```
jyemin, line 85: What should the limit be?

source/server-discovery-and-monitoring/tests/README.rst
```diff
@@ -302,6 +302,20 @@ MongoClient from the one used for test operations. For example::
...
+transient error (see `SERVER-48154`_)::
+
+  {
+    "ok" : 0,
+    "errmsg" : "Unable to acquire X lock on '{4611686018427387905: ReplicationStateTransition, 1}' within 1000ms. opId: 922, op: conn30, connId: 30.",
+    "code" : 24,
+    "codeName" : "LockTimeout",
+  }
+
+When running the "replSetStepDown" command, drivers MUST retry until the
```
jyemin, line 315: Same as above.  It's unfortunate to have to do this in a generic test runner, but I don't see the alternative.

ShaneHarvey:
@kevinAlbs Do you mind reviewing this PR since you are seeing these failures in [CDRIVER-3662](https://jira.mongodb.org/browse/CDRIVER-3662)?

source/connections-survive-step-down/tests/README.rst
ShaneHarvey: I think it needs to be at least 10 seconds total given the description in SERVER-48154:
> If there are hundreds of idents to drop, the dropper can take 10 seconds or more, during which time stepdown and shutdown are blocked.

I added "For example, the Python driver uses a 10 second retry period."

kevinAlbs: Instead of doing a retry, can we increase the time value in the `replSetStepdown` command? Something like: `{replSetStepDown: 60, force: true}`?

The server test mentioned in https://jira.mongodb.org/browse/SERVER-48107 uses a time period of 1. The stepdown tests use a [time period of 5](https://github.com/mongodb/specifications/blob/master/source/connections-survive-step-down/tests/README.rst#getmore-iteration).

It appears that `replSetStepDown` will wait for the given time period before failing if unable to obtain the necessary lock.

The logs in the failing C driver tests for the stepdown test show a jump of 5 seconds, indicating it is blocking for the full 5 seconds before returning an error.

And testing with the shell, if I lock with `db.fsyncLock()`, then attempt to stepdown with `db.adminCommand({replSetStepDown: 30, force: true})`, the `replSetStepDown` will block for 30 seconds before failing with the same error. If I `db.fsyncUnlock()` while the `replSetStepDown` is blocked, it succeeds.

I'm not sure if the manual's description of the time period of `replSetStepDown` lines up with my understanding. https://docs.mongodb.com/manual/reference/command/replSetStepDown/

> The number of seconds to step down the primary, during which time the stepdown member is ineligible for becoming primary. If you specify a non-numeric value, the command uses 60 seconds.

I interpret that as: "the maximum amount of time to attempt to step down the primary, during which it is ineligible for re-election". So I _think_ we can rely on that as a timeout.

ShaneHarvey: That would also be a viable solution but increasing the stepdown time makes subsequent tests (back-to-back test runs) take longer. I prefer the retry solution because it does not slow down the test suite unnecessarily.

kevinAlbs: Oh good point, I'll make the proposed retry change in libmongoc to validate.

source/connections-survive-step-down/tests/README.rst
```diff
@@ -71,8 +71,19 @@ Perform the following operations:
 - If the driver implements the `CMAP`_ specification, verify that no new `PoolClearedEvent`_ has been
   published. Otherwise verify that `connections.totalCreated`_ in `serverStatus`_ has not changed.
 
+**Note:** The "replSetStepDown" command often fails with the following
+transient error (see `SERVER-48154`_)::
 
+  {
+    "ok" : 0,
+    "errmsg" : "Unable to acquire X lock on '{4611686018427387905: ReplicationStateTransition, 1}' within 1000ms. opId: 922, op: conn30, connId: 30.",
```
kevinAlbs, line 79: Not a big deal, but I think the error message here would be "within 5000ms", though I've observed slight discrepencies:

```
test> db.adminCommand({replSetStepDown: 5, force: true})
{
	"ok" : 0,
	"errmsg" : "Unable to acquire X lock on '{4611686018427387905: ReplicationStateTransition, 1}' within 4999ms. opId: 2588184, op: conn1288, connId: 1288.",
	"code" : 24,
	"codeName" : "LockTimeout",
	"$clusterTime" : {
		"clusterTime" : Timestamp(1592331047, 1),
		"signature" : {
			"keyId" : NumberLong(0)
		}
	},
	"operationTime" : Timestamp(1592331047, 1)
}
```

source/connections-survive-step-down/tests/README.rst
kevinAlbs: Testing with a 10 second retry at least works in libmongoc. But this test does not fail very often to begin with, so I'm not sure if that's much assurance.

source/server-discovery-and-monitoring/tests/README.rst
```diff
@@ -302,6 +302,20 @@ MongoClient from the one used for test operations. For example::
             replSetStepDown: 20
             force: false
 
+**Note:** The "replSetStepDown" command often fails with the following
+transient error (see `SERVER-48154`_)::
+
+  {
+    "ok" : 0,
+    "errmsg" : "Unable to acquire X lock on '{4611686018427387905: ReplicationStateTransition, 1}' within 1000ms. opId: 922, op: conn30, connId: 30.",
```
kevinAlbs, line 310: Here too, I don't think the `1000ms` would appear if the limit is 20s in the above example.

However, the rediscover-quickly-after-step-down.yml file uses `1000ms`. Perhaps the example should be updated to match that for consistency?

source/connections-survive-step-down/tests/README.rst
ShaneHarvey: The error is just an example failure, I don't think the actually time in the error message matters.

source/server-discovery-and-monitoring/tests/README.rst
ShaneHarvey: Updated runAdminCommand example to match what's used in rediscover-quickly-after-step-down.yml.

PR #825, opened 2020-06-24
SPEC-1710: Test update/replacement validation

jmikola:
https://jira.mongodb.org/browse/SPEC-1710

Confirmed that tests pass in PHPLIB.

### DIFF
--- a/source/crud/tests/v2/bulkWrite-update-validation.yml
+++ b/source/crud/tests/v2/bulkWrite-update-validation.yml
@@ -1,0 +1,75 @@
+data:
+  - {_id: 1, x: 11}
+  - {_id: 2, x: 22}
+  - {_id: 3, x: 33}
+
+tests:
+  -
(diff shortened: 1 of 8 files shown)


source/crud/tests/v2/bulkWrite-update-validation.yml
```diff
@@ -0,0 +1,75 @@
...
+            -
+              name: "replaceOne"
+              arguments:
+                filter: { _id: 1 }
+                # Only the first field is tested, as the spec permits drivers to
+                # only check that and rely on the server to check subsequent
+                # fields.
+                replacement: { $set: { x: 22 }}
+        error: true
+    expectations: []
```
jmikola, line 23: Some drivers may only consider `expectations` a subset of observed events, in which case this will have no effect, but for others this can be used to assert that no commands were executed and that the error was reported entirely on the client side. Existing "clientError" tests do this.

jmikola:
@mbroadst ping

PR #826, opened 2020-06-26
SPEC-1708 Increase replSetStepDown timeout and run replSetFreeze instead of retrying

ShaneHarvey:
Taking Eric's advice on [DRIVERS-1290](https://jira.mongodb.org/browse/DRIVERS-1290). @DmitryLukyanov can you POC these changes to see if it resolves the Windows C# issues?

Instead of retrying after LockTimeout errors, we increase the replSetStepDown timeout to avoid them altogether. Before we run replSetStepDown, run `{replSetFreeze:0}` to reset a secondary's timeout to ensure there is an electable node. This ensures a speedy election when the tests are run multiple times in a row (as is the case in C# for testing both sync and async).

We also increase the waitForPrimaryChange timeout from 5 seconds to 15 seconds to workaround slow elections on Windows (again found by C#). The slow elections are possibly caused by SERVER-48154.

### DIFF
--- a/source/connections-survive-step-down/tests/README.rst
+++ b/source/connections-survive-step-down/tests/README.rst
@@ -67,7 +67,7 @@ Perform the following operations:
 - Send a ``{replSetFreeze: 0}`` command to any secondary and verify that the
   command succeeded. This command will unfreeze the secondary and ensure that
   it will be eligible to be elected immediately.
-- Send a ``{replSetStepDown: 20, force: true}`` command to the current primary and verify that
+- Send a ``{replSetStepDown: 30, force: true}`` command to the current primary and verify that
   the command succeeded.
 - Retrieve the next batch of results from the cursor obtained in the find
   operation, and verify that this operation succeeded.
--- a/source/connections-survive-step-down/tests/README.rst
+++ b/source/connections-survive-step-down/tests/README.rst
@@ -64,26 +64,16 @@ Perform the following operations:
 - Insert 5 documents into a collection with a majority write concern.
 - Start a find operation on the collection with a batch size of 2, and
   retrieve the first batch of results.
-- Send a ``{replSetStepDown: 5, force: true}`` command to the current primary and verify that
+- Send a ``{replSetFreeze: 0}`` command to any secondary and verify that the
+  command succeeded. This command will unfreeze the secondary and ensure that
+  it will be eligible to be elected immediately.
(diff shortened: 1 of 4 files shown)


DmitryLukyanov:
It looks like it solves the issue with `Command replSetStepDown failed: Unable to acquire X lock on` error. At least last 4 patches didn't contain this fails. Also, I didn't see `The primary has not been changed from` errors. I will provide more results on Monday.

DmitryLukyanov:
So, I've implemented this POC, but It helped only to decrease the number of failed tasks. Before it was failed like in 2-3 tasks for a patch, now it's about 1 failed task for 2-3 patches. See for example:
https://evergreen.mongodb.com/task_log_raw/dot_net_driver_unsecure_tests__version~4.2_os~windows_64_topology~replicaset_auth~noauth_ssl~nossl_test_netstandard15_patch_297fcd723ff32aac47b7b018c978ec4baf0773d7_5ef7e66e30661563e09e75f6_20_06_28_00_38_06/0?type=T#L1836 and
https://evergreen.mongodb.com/task_log_raw/dot_net_driver_unsecure_tests__version~4.2_os~windows_64_topology~replicaset_auth~noauth_ssl~nossl_test_netstandard20_patch_297fcd723ff32aac47b7b018c978ec4baf0773d7_5ef7e67a30661563e09e75fa_20_06_28_00_38_18/0?type=T#L2252

ShaneHarvey:
Thanks @DmitryLukyanov I've opened  https://jira.mongodb.org/browse/SERVER-49273 for the server to investigate the timeout errors you're seeing. For now I've increased the timeout from 20 to 30 seconds. Would you be able to test if that's enough and if the test still fails could you suggest a more suitable timeout?

DmitryLukyanov:
> 
> 
> Thanks @DmitryLukyanov I've opened https://jira.mongodb.org/browse/SERVER-49273 for the server to investigate the timeout errors you're seeing. For now I've increased the timeout from 20 to 30 seconds. Would you be able to test if that's enough and if the test still fails could you suggest a more suitable timeout?

will check today

ShaneHarvey:
@DmitryLukyanov were you able to test this out?

DmitryLukyanov:
lately, i didn't see this problem even with `"replSetStepDown": 20,` however, I think it will be safer to increase this value just in case.

ShaneHarvey:
@kevinAlbs Do you mind reviewing this one as well?

PR #828, opened 2020-07-01
Simplify logic for sessions support detection

mbroadst:
Some standalone servers do support sessions, we should instead depend strictly on the presence of `logicalSessionsTimeoutMinutes` in the server handshake to determine if a server supports sessions.

DRIVERS-1230


source/sessions/driver-sessions.rst
```diff
@@ -979,19 +977,19 @@ Test Plan
 =========
 
 The test plan SHOULD be run against both replica set and sharded cluster
-topologies.  It MUST NOT be run against a standalone server.
```
jmikola, line 982: If you're removing the prohibition about running on a standalone, I would venture that the topology really doesn't matter at all. In that case, should the first sentence be changed as well? Perhaps:

> The test plan SHOULD be run against deployments that support sessions

Or just remove this sentence as well.

source/sessions/driver-sessions.rst
```diff
@@ -501,8 +501,6 @@ include a session ID in a ``KILLCURSORS`` command.
 How to Check Whether a Deployment Supports Sessions
 ===================================================
 
-Standalone servers do not support sessions.
-
```
jmikola, line 505: I think you should also replace "whether a replica set or sharded cluster deployment" with "whether a deployment" if we don't care about topologies anymore.

source/sessions/tests/server-support.yml
```diff
@@ -0,0 +1,102 @@
+runOn:
+  - minServerVersion: "3.6.0"
+    topology: ["single"]
+  - minServerVersion: "4.0"
+    topology: ["replicaset"]
```
jmikola, line 5: What about these tests requires a 4.0 replica set or 4.1.8 sharded cluster?

We're not using fail points, so this looks like basic session support that would be 3.6+. Let me know if I'm missing something.

source/sessions/tests/dirty-session-errors.yml
```diff
@@ -409,32 +198,3 @@ tests:
       collection:
         data:
           - {_id: 1}
-
-  - description: Dirty implicit session is discarded (non-cursor returning read)
```
jmikola, line 413: Why were these four tests removed?

* Clean explicit session is not discarded
* Clean implicit session is not discarded
* Dirty explicit session is discarded (non-bulk write)
* Dirty implicit session is discarded (non-bulk write)

Were they redundant? They seem unrelated to relaxing the prohibition on standalone deployments.

source/sessions/driver-sessions.rst
mbroadst: Good idea, I'll just remove it

source/sessions/tests/server-support.yml
mbroadst: fixed

source/sessions/tests/dirty-session-errors.yml
mbroadst: hm I have no idea what happened here, restoring them now

mbroadst: This was unintentional, I was copying from a branch that hadn't updated its local tests to include these more recent test additions. Fixed

source/sessions/tests/dirty-session-errors.yml
```diff
@@ -11,96 +11,6 @@ data:
   - {_id: 1}
 
 tests:
-  - description: Clean explicit session is not discarded
```
jmikola, line 14: Two tests are still being removed here:

 * Clean explicit session is not discarded
 * Clean implicit session is not discarded

Is that intentional?

mbroadst: Yes it was intentional, they were actually just renamed to be the tests in the `server-support` tests

PR #829, opened 2020-07-01
SPEC-1712: Default OCSP to "off" for drivers that hard-fail when an OCSP responder is unavailable

vincentkam:
https://jira.mongodb.org/browse/SPEC-1712
See for the motivation behind the change: https://jira.mongodb.org/browse/DRIVERS-1311

### DIFF
--- a/source/ocsp-support/ocsp-support.rst
+++ b/source/ocsp-support/ocsp-support.rst
@@ -128,7 +128,7 @@ invalid, the driver SHOULD end the connection.
 10. Finally, the driver SHOULD continue the connection, even if the
     status of all the unvalidated certificates has not been
     confirmed yet. This means that the driver SHOULD default to
-    “soft fail” behavior, connecting as long as there are no
+    "soft fail" behavior, connecting as long as there are no
     explicitly invalid certificates—i.e. the driver will connect
     even if the status of all the unvalidated certificates has not
     been confirmed yet (e.g. because an OCSP responder is down).
@@ -196,19 +196,18 @@ be disabled.  When set to true, a driver MUST NOT reach out to OCSP
 endpoints. When set to false, a driver MUST reach out to OCSP
 endpoints if needed (as described in
 `Specification: Suggested OCSP Behavior <ocsp-support.rst#id1>`__).
-This option MUST default to false except in the case that a driver' TLS
-library exhibits hard-fail behavior (i.e. the driver fails the
-`"Soft Fail Test
-"<https://github.com/mongodb/specifications/tree/master/source/ocsp-support/tests#integration-tests-permutations-to-be-tested>`__
-).
 
-In the case of a driver whose TLS library exhibits hard-fail behavior,
-this value MUST default to true, and a driver MUST document this
-behavior.  If this hard-failure behavior is specific to a particular
-platform (e.g. the TLS library hard-fails only on Windows) then this
-boolean MUST default to true only on the platform where the driver
-exhibits hard-fail behavior, and a driver MUST document this behavior.
+For drivers that pass the `"Soft Fail Test"
+<tests/README.rst#integration-tests-permutations-to-be-tested>`__ , this
+option MUST default to false.
 
+For drivers that fail the "Soft Fail Test" because their TLS library
+exhibits hard-fail behavior when a responder is unreachable, this option
+MUST default to true, and a driver MUST document this behavior. If this
+hard-failure behavior is specific to a particular platform (e.g. the TLS
+library hard-fails only on Windows) then this option MUST default to
+true only on the platform where the driver exhibits hard-fail behavior,
+and a driver MUST document this behavior.
 
 tlsDisableCertificateRevocationCheck
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -225,17 +224,18 @@ MUST NOT check certificate revocation status via CRLs or OCSP.  When
 set to false, a driver MUST check certificate revocation status, reach
 out to OCSP endpoints if needed (as described in
 `Specification: Suggested OCSP Behavior <ocsp-support.rst#id1>`__).
-This option MUST default to false except in the case that a driver's TLS
-library exhibits hard-fail behavior (i.e. the driver fails the
-`"Soft Fail Test
-"<https://github.com/mongodb/specifications/tree/master/source/ocsp-support/tests#integration-tests-permutations-to-be-tested>`__
 
-In the case of a driver whose TLS library exhibits hard-fail behavior,
-this value MUST default to true, and a driver MUST document this
-behavior.  If this hard-failure behavior is specific to a particular
-platform (e.g. the TLS library hard-fails only on Windows) then this
-boolean MUST default to true only on the platform where the driver
-exhibits hard-fail behavior, and a driver MUST document this behavior.
+For drivers that pass the `"Soft Fail Test"
+<tests/README.rst#integration-tests-permutations-to-be-tested>`__ , this
+option MUST default to false.
+
+For drivers that fail the "Soft Fail Test" because their TLS library
+exhibits hard-fail behavior when a responder is unreachable, this option
+MUST default to true, and a driver MUST document this behavior. If this
+hard-failure behavior is specific to a particular platform (e.g. the TLS
+library hard-fails only on Windows) then this option MUST default to
+true only on the platform where the driver exhibits hard-fail behavior,
+and a driver MUST document this behavior.
 
 Naming Deviations
 ^^^^^^^^^^^^^^^^^^
@@ -277,7 +277,7 @@ expose an option to enable/disable certificate revocation checking on a
 per MongoClient basis.
 
 1. Driver MUST enable OCSP support (with stapling if possible) when
-   certificate revocation checking is enabled UNLESS their driver
+   certificate revocation checking is enabled **unless** their driver
    exhibits hard-fail behavior (see
    `tlsDisableCertificateRevocationCheck`_). In such a case, a driver
    MUST disable OCSP support on the platforms where its TLS library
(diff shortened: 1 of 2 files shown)


source/ocsp-support/ocsp-support.rst
```diff
@@ -3,14 +3,14 @@ OCSP Support
 ============
 
 :Spec Title: OCSP Support
-:Spec Version: 1.3.1
+:Spec Version: 2.0.0
```
vincentkam, line 6: I feel like changing whether OCSP is enabled by default is a breaking change and thus requires a major version bump.

source/ocsp-support/ocsp-support.rst
```diff
@@ -194,9 +194,21 @@ This boolean option determines whether a MongoClient should refrain from
...
+library exhibits hard-fail behavior (i.e. the driver fails the
+`"Soft Fail Test
+"<https://github.com/mongodb/specifications/tree/master/source/ocsp-support/tests#integration-tests-permutations-to-be-tested>`__
+).
+
+In the case of a driver whose TLS library exhibits hard-fail behavior,
+this value MUST default to true and a driver MUST document this
+behavior.  If this hard-failure behavior is specific to a particular
+platform (e.g. the TLS library hard-fails only on Windows) then this
+boolean MUST default to true only on the platform where the exhibits
```
behackett, line 209: where the -> where it

source/ocsp-support/ocsp-support.rst
```diff
@@ -211,10 +223,19 @@ This boolean option determines whether a MongoClient should refrain
...
+This option MUST default to false except in the case that a driver's TLS
+library exhibits hard-fail behavior (i.e. the driver fails the
+`"Soft Fail Test
+"<https://github.com/mongodb/specifications/tree/master/source/ocsp-support/tests#integration-tests-permutations-to-be-tested>`__
+
+In the case of a driver whose TLS library exhibits hard-fail behavior,
+this value MUST default to true and a driver MUST document this
+behavior.  If this hard-failure behavior is specific to a particular
+platform (e.g. the TLS library hard-fails only on Windows) then this
+boolean MUST default to true only on the platform where the exhibits
```
behackett, line 237: where the -> where it

source/ocsp-support/ocsp-support.rst
```diff
@@ -756,8 +783,13 @@ of checking this are:
 Changelog
 ==========
 
+**2020-07-01**: 2.0.0: Default tlsDisableCertificateRevocationCheck
+and tlsDisableOCSPEndpointCheck to false in the case that a driver's
```
behackett, line 787: to false -> to true

source/ocsp-support/ocsp-support.rst
vincentkam: Oops, thanks! I changed "where the" to be  "where the driver" since I felt "it" was very mildly ambiguous.

source/ocsp-support/ocsp-support.rst
vincentkam: Ditto above :)

source/ocsp-support/ocsp-support.rst
vincentkam: Fixed!

source/ocsp-support/ocsp-support.rst
```diff
@@ -194,9 +194,21 @@ This boolean option determines whether a MongoClient should refrain from
...
 be disabled.  When set to true, a driver MUST NOT reach out to OCSP
 endpoints. When set to false, a driver MUST reach out to OCSP
-endpoints if needed (as described in 
+endpoints if needed (as described in
 `Specification: Suggested OCSP Behavior <ocsp-support.rst#id1>`__).
-This option MUST default to false.
+This option MUST default to false except in the case that a driver' TLS
+library exhibits hard-fail behavior (i.e. the driver fails the
+`"Soft Fail Test
+"<https://github.com/mongodb/specifications/tree/master/source/ocsp-support/tests#integration-tests-permutations-to-be-tested>`__
```
jmikola, line 202: I think the double quote character should be adjacent to "Test" on the previous line. As-is, you'll get a space between "Test" and the closing quote. You probably do want a space between the link text and the `<` character as well.

source/ocsp-support/ocsp-support.rst
```diff
@@ -256,7 +277,11 @@ expose an option to enable/disable certificate revocation checking on a
 per MongoClient basis.
 
 1. Driver MUST enable OCSP support (with stapling if possible) when
-   certificate revocation checking is enabled.
+   certificate revocation checking is enabled UNLESS their driver
```
jmikola, line 280: "UNLESS" is not a special term from RFC 2119, so I don't think this needs to be capitalized. Feel free to add italics for emphasis.

source/ocsp-support/ocsp-support.rst
```diff
@@ -339,8 +364,10 @@ a response and the OCSP responder is down.
 Drivers that fail “Soft Fail Test” MUST document that their driver’s
 TLS library utilizes “hard fail” behavior in the case of an
 unavailable OCSP responder in contrast to the mongo shell and drivers
-that utilize “soft-fail” behavior. Such drivers MUST also document the
-potential backwards compatibility issues as noted in the `Backwards
+that utilize “soft fail” behavior. They also MUST document the change
```
jmikola, line 367: Please replace with non-directional double quote characters. Bonus points if you catch any other instances of this in the document (I see a few in the lines above).

source/ocsp-support/ocsp-support.rst
```diff
@@ -339,8 +364,10 @@ a response and the OCSP responder is down.
...
+in defaults for the applicable options (see `MongoClient
+Configuration`_). Such drivers MUST also document the potential
+backwards compatibility issues as noted in the `Backwards
```
jmikola, line 370: AFAIK, no driver has had a stable release with OCSP. Why would any driver need to document a BC issue?

source/ocsp-support/ocsp-support.rst
```diff
@@ -451,7 +478,7 @@ disabling non-stapled OCSP via ``tlsDisableOCSPEndpointCheck`` or by
 disabling certificate revocation checking altogether
 via ``tlsDisableCertificateRevocationCheck``.
 
-An application that uses a driver that utilizes hard-fail behavior
+An application that uses a driver that utilizes hard fail behavior
```
jmikola, line 481: The changelog entry says "hard-fail", but you purposely removed the hyphen here. Should this be consistent throughout?

source/ocsp-support/tests/README.rst
```diff
@@ -85,6 +88,9 @@ extra cases and may help reveal unexpected behavior.
 library’s implementation of OCSP will need to document these failures as
 described under `Documentation
 Requirements <../ocsp-support.rst#documentation-requirements>`__
+Additionally, drivers that fail the "Soft Fail" will need to change any applicable defaults as described under
```
jmikola, line 91: Should this be "Soft Fail Test", as written in the spec document? As-is, "fail the soft fail" reads oddly.

divjotarora:
As a side note, how many drivers that default to hard-fail behavior even have enough control to implement `tlsDisableOCSPEndpointCheck`?

source/ocsp-support/ocsp-support.rst
vincentkam: Ah true, fixed (went with a bold).

source/ocsp-support/ocsp-support.rst
vincentkam: Done, with the bonus points :)  I think they were probably left over when this was ported from google docs. Out of curiosity, why is having directional quotes considered undesirable?

source/ocsp-support/ocsp-support.rst
vincentkam: Fixed! Note, currently we tend to use "hard fail" when it's in quotes and and hard-fail when it's not quoted. Ditto with "soft fail" vs soft-fail.

source/ocsp-support/ocsp-support.rst
vincentkam: Fixed. I currently have a new line between the link test and the <, which appears to work when viewing it on Github.

source/ocsp-support/ocsp-support.rst
```diff
@@ -194,9 +194,21 @@ This boolean option determines whether a MongoClient should refrain from
 reaching out to an OCSP endpoint i.e.  whether non-stapled OCSP should
 be disabled.  When set to true, a driver MUST NOT reach out to OCSP
 endpoints. When set to false, a driver MUST reach out to OCSP
-endpoints if needed (as described in 
+endpoints if needed (as described in
 `Specification: Suggested OCSP Behavior <ocsp-support.rst#id1>`__).
-This option MUST default to false.
+This option MUST default to false except in the case that a driver' TLS
```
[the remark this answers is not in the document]
vincentkam: I like your suggestion and have taken it, with somem inor tweaks.

source/ocsp-support/ocsp-support.rst
vincentkam: So the line about drivers documenting potential backwards breaking changes was originally in the spec because we were okay with drivers with low amounts of control over their OCSP implementation defaulting to turning OCSP on even if they were hard-failing. Such a change would also be a backwards breaking change for applications deployed behind a restrictive allow-list.

Now that we're explicitly disallowing drivers from defaulting to OCSP to on when they exhibit hard-fail, rewording this section a bit for clarity is probably a reasonable idea. I've taken Divjot's suggestion and tweaked it mildly :) 

Some extra .NET specific history that inspired the original wording:
Historically, the .NET driver changed the default for revocation checking several times, and when the default changes from false to true, it can cause breaking changes in the case of applications deployed behind a restrictive allow-list because the driver defaults to hard-fail (and we've had a HELP ticket come in about such a case).

source/ocsp-support/tests/README.rst
vincentkam: Fixed!

source/ocsp-support/ocsp-support.rst
```diff
@@ -437,11 +467,11 @@ Backwards Compatibility
 
 An application behind a firewall with an outbound whitelist that
 upgrades to a driver implementing this specification may experience
-connectivity issues. This is because the driver may need to contact
+connectivity issues when OCSP is enabled. This is because the driver may need to contact
```
vincentkam, line 470: Clarifying that these backwards breaking changes should only apply when OCSP is enabled. Previously, we were defaulting to enabling OCSP even if drivers hard-failed. Now, this  applies only if an application opts in to OCSP and a driver exhibits hard-fail behavior.

vincentkam:
Offline discussion w/ Divjot summary: Specifying both OCSP related TLS options to "true" in the case of a driver that hard fails could cause issues in the future if a driver were to implement both options because currently our URI options tests disallow setting both options to false, which would be needed if an application wanted to "opt in" to OCSP, despite the hard fail.
Solution: We will specify that a driver should preferentially set tlsDisableOcspEndpointCheck to true if it exhibits hard fail behavior. If a driver has not already set this option to true (e.g. it doesn't support it), it should set tlsDisableCertificateRevocationCheck to true. This means that an application can simply specify one URI option and opt into OCSP w/hard-fail as needed. The bigger question of whether or not to allow multiple TLS options to be specified in a connection string at once already has an open drivers ticket from discussion on this topic

vincentkam:
Re requesting review due to the tweak to the spec specifying that only ONE uri option needs to be set to true in the case of hard fail.

PR #833, opened 2020-07-07
SPEC-1715 add additional timestamp test for not 2^32-1

Neilshweky:
https://jira.mongodb.org/browse/SPEC-1715

If the 64 bit number 2^32-1 is not stored in an unsigned number, it will wrap around to -1, which is a special value. I added another test that has the high-order bit set that doesn't yield a special value on error.

### DIFF
--- a/source/bson-corpus/tests/timestamp.json
+++ b/source/bson-corpus/tests/timestamp.json
@@ -18,6 +18,11 @@
             "description": "Timestamp with high-order bit set on both seconds and increment",
             "canonical_bson": "10000000116100FFFFFFFFFFFFFFFF00",
             "canonical_extjson": "{\"a\" : {\"$timestamp\" : {\"t\" : 4294967295, \"i\" :  4294967295} } }"
+        },
+        {
+            "description": "Timestamp with high-order bit set: (4000000000, 4000000000)",
+            "canonical_bson": "1000000011610000286BEE00286BEE00", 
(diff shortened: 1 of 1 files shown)


source/bson-corpus/tests/timestamp.json
```diff
@@ -18,6 +18,11 @@
             "description": "Timestamp with high-order bit set on both seconds and increment",
             "canonical_bson": "10000000116100FFFFFFFFFFFFFFFF00",
             "canonical_extjson": "{\"a\" : {\"$timestamp\" : {\"t\" : 4294967295, \"i\" :  4294967295} } }"
+        },
+        {
+            "description": "Timestamp with high-order bit set: (4000000000, 4000000000)",
```
jmikola, line 23: Let's rename this to "Timestamp with high-order bit set on both seconds and increment (not UINT32_MAX)", since it's very much like the preceding test apart from the value being tested.

Neilshweky: Done!

Neilshweky:
Hey Jeremy, thanks so much for taking the time to look through this... I just updated everything!

jmikola:
LGTM. Feel free to squash the commits when merging, since the overall change is pretty small and the description rename won't be relevant to consumers of the specs repo (as opposed to other cases where it may be helpful to keep multiple commits for context when merging a PR).

jmikola:
For the benefit of anyone else that comes across this PR, I did test the spec changes with PHPC (implicitly libmongoc) and they passed.

Neilshweky:
Awesome, I just merged... Should I close the spec ticket?

PR #835, opened 2020-07-10
SPEC-1712: Default OCSP to "off" for drivers that hard-fail when an OCSP responder is unavailable

divjotarora:
The changes proposed here apply only to drivers whose TLS libraries hard-fail when an OCSP responder is unavailable. To ensure an OCSP responder outage doesn't affect Atlas connectivity, affected drivers should:

1. Set the `tlsDisableOCSPEndpointCheck` option to true by default if it is supported. This will allow the driver to support stapled OCSP when available and avoid reaching out to responders altogether when a stapled response isn't provided.

2. Set the `tlsDisableCertificateRevocationCheck` option to true by default if `tlsDisableOCSPEndpointCheck` is not supported. This will turn off all revocation checks, including non-OCSP checks, but is necessary if the driver doesn't have granular control of stapled vs non-stapled OCSP.

Note that the initial approach presented in the previous PR for this ticket required defaulting both of the options mentioned above to true. This wouldn't work if a driver were to support both of these options because a URI with both set to false is considered invalid, so a user wouldn't be able to opt-in to non-stapled OCSP and other revocation checks without running into a configuration error. Such a driver does not currently exist, but it's important to ensure our specs are consistent with each other.

### DIFF
--- a/source/ocsp-support/ocsp-support.rst
+++ b/source/ocsp-support/ocsp-support.rst
@@ -229,8 +229,8 @@ For drivers that pass the `"Soft Fail Test"
 <tests/README.rst#integration-tests-permutations-to-be-tested>`__ , this
 option MUST default to false.
 
-If a driver has not already defaulted `tlsDisableOCSPEndpointCheck` to
-true, and if that driver fails the "Soft Fail Test" because their TLS
+If a driver does not support ``tlsDisableOCSPEndpointCheck`` and
+that driver fails the "Soft Fail Test" because their TLS
 library exhibits hard-fail behavior when a responder is unreachable,
 then that driver must default `tlsDisableCertificateRevocationCheck` to
 true. Such a driver also MUST document this behavior. If this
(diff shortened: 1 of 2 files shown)


source/ocsp-support/ocsp-support.rst
```diff
@@ -211,10 +222,22 @@ This boolean option determines whether a MongoClient should refrain
...
-`Specification: Suggested OCSP Behavior <ocsp-support.rst#id1>`__).  
-This option MUST default to false.
+out to OCSP endpoints if needed (as described in
+`Specification: Suggested OCSP Behavior <ocsp-support.rst#id1>`__).
 
+For drivers that pass the `"Soft Fail Test"
+<tests/README.rst#integration-tests-permutations-to-be-tested>`__ , this
+option MUST default to false.
+
+If a driver has not already defaulted `tlsDisableOCSPEndpointCheck` to
```
behackett, line 232: It might be better to say "If a driver does not support tlsDisableEndpointCheck". As written this might be misunderstood to imply that an implementer can arbitrarily pick which of the options to default to true, but I don't think that's your intent. Is it?

divjotarora: That's not my intent. I updated the phrasing in the new commit.

PR #836, opened 2020-07-12
Revert addition of "single" topology type to dirty sessions test

jyemin:
The tests assume that the server supports retryable writes,
and even though standalones support sessions, they don't actually
support retryable writes (which require an oplog)

DRIVERS-1230


jmikola:
Cross-referencing with #828.

PR #837, opened 2020-07-14
SPEC-1718: Sync bson corpus tests for invalid regexp tests fix

nbbeeken:
I could update the name of the ticket but essentially we want drivers to make sure that when running BSON corpus that they're not failing on a JSON decoding exception rather than a real corpus mismatch issue. All of the stringify-ed JSON in the corpus tests are valid, I checked.


source/bson-corpus/bson-corpus.rst
```diff
@@ -178,6 +178,11 @@ this means that the corpus JSON will appear to have double-escaped characters
 printable ASCII without embedded null characters to ensure maximum portability
 to different language JSON or extended JSON decoders.
 
+Drivers MUST ensure that Extended JSON in
+``canonical_extjson``, ``relaxed_extjson``, and ``degenerate_extjson``
+are valid JSON strings. A driver shouldn't have to skip a test case because
+it cannot decode the JSON in the test.
```
p-mongo, line 184: I think it would be more effective for this language to be prescriptive, for example, in https://github.com/mongodb/specifications/blob/master/source/bson-corpus/bson-corpus.rst#parse-error-case-keys add something like:

> Parse the extended JSON input using a regular JSON parser (not an extended JSON one) and verify the input is parsed successfully. This serves to verify that the parse error test cases test extended JSON-specific error conditions and that they do not have, for example, unintended spelling errors.

nbbeeken: Thanks, yea I think that is more clear. I left it as drivers SHOULD because is matches up with the other sections about decoding and parsing errors. I suppose should leaves the freedom for some drivers to not have to explicitly do this separate check if their parser will implicitly throw the decoding error.

p-mongo:
The language looks good, the placement seems a bit weird - move it just up of the "Extended JSON encoding, escaping and ordering" section heading?

PR #839, opened 2020-07-15
SPEC-1684: Clarify that collection under test must not be dropped

jmikola:
https://jira.mongodb.org/browse/SPEC-1684


jmikola:
> Perhaps this should be added as a new parameterization in the unified test runner, so that you can control the drop behavior via the spec test YAML.

Tracking that in [SPEC-1723](https://jira.mongodb.org/browse/SPEC-1723), which I've linked with the unified test format project.

PR #847, opened 2020-08-17
SPEC-1746 SDAM spec tests without directConnection=true shouldn't exp…

iwysiu:
…ect Single topology


p-mongo:
This was done in https://github.com/mongodb/specifications/pull/844 as far as I can tell.

iwysiu:
yeah, looks like it. thanks!

PR #848, opened 2020-08-26
SPEC-1412 Clarify Connection creation wording and terminology

patrickfreed:
SPEC-1412

This PR makes a number of modifications to the CMAP spec to clarify the lifecycle of a `Connection` and the terminology used to describe it. Besides resyncing the tests, this ideally should not require any changes in drivers that currently implement a CMAP-based connection pool.

This originally started out as fixing a minor issue with the wording of "creating a connection" that resulted in ConnectionReady events being fired before ConnectionCreated events according to a literal interpretation of the spec. In fixing this issue, I realized that a solution could not easily be found that worked for both drivers that had a background thread and/or async I/O and for drivers that were single threaded. Seeing as no single-threaded drivers currently implement CMAP or plan to do so, this PR also adds a requirement to the CMAP pool that a background thread and/or async I/O is used. See the comments on SPEC-1412 (and the rationale included in the diff) for details on this part. 

Because this change involved a lot of search-and-replace, the diff is a little ugly. I've highlighted the actual changes here to aid in reviewing:
- Defined `Connection` specifically as a "Connection object" and disambiguated it from a TCP/IP connection
    - `code-formatted` "Connection" everywhere in the spec where it was referenced in this context.
- exhaustively listed the possible "states" that a `Connection` can be in and described them clearly
    - the states: "unestablished", "available", "in use", and "closed"
    - removed references to "readyToUse" and "setting up" a connection, replaced with "established" and "establish(ing)"
- added separate sections for "Populating the Pool with a Connection" and "Marking a Connection as Available"
- Required ensuring minPoolSize and clearing of perished connections to be non-blocking or done by the background thread
- Added test verifying ConnectionReady events are fired after ConnectionCreated events
- Added test verifying connections are established as a part of minPoolSize population

### DIFF
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -475,7 +475,7 @@ non-perished available `Connection <#connection>`_ is found or the list of
 available `Connections <#connection>`_ is exhausted. If no `Connections
 <#connection>`_ are available and the total number of `Connections
 <#connection>`_ is less than maxPoolSize, the pool MUST create and return a new
-`Connection <#connection>`_.
+established `Connection <#connection>`_.
 
 If the pool is closed, any attempt to check out a `Connection <#connection>`_ MUST throw an Error, and any items in the waitQueue MUST be removed from the waitQueue and throw an Error.
 
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -329,14 +329,13 @@ The SDAM specification defines `when
 <https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#connection-pool-creation>`_
 the driver should create connection pools.
 
-Once a pool is created, if minPoolSize is set, the pool MUST
-immediately begin being "populated" with enough `Connections
-<#connection>`_ such that totalConnections >= minPoolSize. The Pool
-MUST be "populated" in a non-blocking manner, such as via the use of a
-background thread or asynchronous I/O. See `Populating the Pool with a
-Connection
-<#populating-the-pool-with-a-connection-internal-implementation>`_ for
-more details.
+Once a pool is created, if minPoolSize is set, the pool MUST immediately begin
+being "populated" with enough `Connections <#connection>`_ such that
+totalConnections >= minPoolSize. The Pool MUST be "populated" in a non-blocking
+manner, such as via the use of a background thread or asynchronous I/O. See
+`Populating the Pool with a Connection
+<#populating-the-pool-with-a-connection-internal-implementation>`_ for more
+details.
 
 .. code::
 
@@ -442,17 +441,16 @@ Populating the Pool with a Connection (Internal Implementation)
 ---------------------------------------------------------------
 
 "Populating" the pool involves creating and establishing a `Connection
-<#connection>`_ that will not immediately be checked out and will
-instead be marked as "available" and kept in the pool. This type of
-`Connection <#connection>`_ is created to ensure minPoolSize and is
-distinct from from the type that is created, established, and
-immediately checked out in checkOut. See `Checking Out a Connection
-<#populating-the-pool-with-a-connection-internal-implementation>`_ for
-more information on that type of `Connection <#connection>`_ creation.
+<#connection>`_ that will not immediately be checked out and will instead be
+marked as "available" and kept in the pool. This type of `Connection
+<#connection>`_ is created to ensure minPoolSize and is distinct from from the
+type that is created, established, and immediately checked out in checkOut. See
+`Checking Out a Connection
+<#populating-the-pool-with-a-connection-internal-implementation>`_ for more
+information on that type of `Connection <#connection>`_ creation.
 
-Populating the pool MUST NOT block any application threads. For
-example, it could be performed on a background thread or via the use
-of non-blocking I/O.
+Populating the pool MUST NOT block any application threads. For example, it
+could be performed on a background thread or via the use of non-blocking I/O.
 
 .. code::
 
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -235,10 +235,10 @@ A driver-defined wrapper around a single TCP/IP connection to an Endpoint. A ``C
 WaitQueue
 ---------
 
-A concept that represents pending requests for Connections. When a thread requests a ``Connection`` from a Pool, the thread enters the Pool's WaitQueue. A thread stays in the WaitQueue until it either receives a ``Connection`` or times out. A WaitQueue has the following traits:
+A concept that represents pending requests for ``Connections``. When a thread requests a ``Connection`` from a Pool, the thread enters the Pool's WaitQueue. A thread stays in the WaitQueue until it either receives a ``Connection`` or times out. A WaitQueue has the following traits:
 
 -  **Thread-Safe**: When multiple threads attempt to enter or exit a WaitQueue, they do so in a thread-safe manner.
--  **Ordered/Fair**: When connections are made available, they are issued out to threads in the order that the threads entered the WaitQueue.
+-  **Ordered/Fair**: When ``Connections`` are made available, they are issued out to threads in the order that the threads entered the WaitQueue.
 -  **Timeout aggressively:** If **waitQueueTimeoutMS** is set, members of a WaitQueue MUST timeout if they are enqueued for longer than waitQueueTimeoutMS. Members of a WaitQueue MUST timeout aggressively, and MUST leave the WaitQueue immediately upon timeout.
 
 The implementation details of a WaitQueue are left to the driver.
@@ -258,10 +258,10 @@ A driver-defined entity that encapsulates all non-monitoring Connections associa
 -  **Emit Events:** A Pool MUST emit pool events when dictated by this spec (see `Connection Pool Monitoring <#connection-pool-monitoring>`__). Users MUST be able to subscribe to emitted events in a manner idiomatic to their language and driver.
 -  **Closeable:** A Pool MUST be able to be manually closed. When a Pool is closed, the following behaviors change:
 
-   -  Checking in a ``Connection`` to the Pool automatically closes the Connection
+   -  Checking in a ``Connection`` to the Pool automatically closes the ``Connection``
    -  Attempting to check out a ``Connection`` from the Pool results in an Error
 
--  **Capped:** a pool is capped if **maxPoolSize** is set to a non-zero value. If a pool is capped, then its total number of Connections (including available and in use) MUST NOT exceed **maxPoolSize**
+-  **Capped:** a pool is capped if **maxPoolSize** is set to a non-zero value. If a pool is capped, then its total number of ``Connections`` (including available and in use) MUST NOT exceed **maxPoolSize**
 
 .. code:: typescript
 
@@ -329,7 +329,7 @@ immediately begin populating enough ``Connections`` such that
 totalConnections >= minPoolSize. These ``Connections`` MUST be
 populated in a non-blocking manner, such as via the use of a
 background thread or asynchronous I/O. See `Populating the Pool with a
-Connection <#deprecated-options>`_ for more details.
+Connection <##populating-the-pool-with-a-connection-internal-implementation>`_ for more details.
 
 .. code::
 
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -10,7 +10,7 @@ Connection Monitoring and Pooling
 :Type: Standards
 :Minimum Server Version: N/A
 :Last Modified: June 11, 2019
-:Version: 1.1.0
+:Version: 1.2.0
 
 .. contents::
 
@@ -895,5 +895,9 @@ Exhaust Cursors may require changes to how we close `Connections <#connection>`_
 
 Change log
 ==========
+:2020-09-03: Clarify Connection states and definition. Require the use of a
+             background thread and/or async I/O. Add tests to ensure
+             ConnectionReadyEvents are fired after ConnectionCreatedEvents.
 
-:2019-06-06: Add "connectionError" as a valid reason for ConnectionCheckOutFailedEvent
+:2019-06-06: Add "connectionError" as a valid reason for
+             ConnectionCheckOutFailedEvent
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -31,13 +31,13 @@ Connection
 ~~~~~~~~~~~~~~
 
 A Connection (when linked) refers to the ``Connection`` type defined
-in this specification. It does not refer to an actual TCP/IP
-connection to an Endpoint. A ``Connection`` will attempt to create and
-wrap such a connection over the course of its existence, but it is not
+in this specification. It does not refer to an actual TCP connection
+to an Endpoint. A ``Connection`` will attempt to create and wrap such
+a TCP connection over the course of its existence, but it is not
 equivalent to one nor does it wrap an active one at all times.
 
 For the purposes of testing, a mocked ``Connection`` type could be used
-with the pool that never actually creates a TCP/IP connection or
+with the pool that never actually creates a TCP connection or
 performs any I/O.
 
 Endpoint
@@ -91,7 +91,7 @@ Drivers that implement a Connection Pool MUST support the following ConnectionPo
 
     interface ConnectionPoolOptions {
       /**
-       *  The maximum number of connections that may be associated
+       *  The maximum number of Connections that may be associated
        *  with a pool at a given time. This includes in use and
        *  available connections.
        *  If specified, MUST be an integer >= 0.
@@ -101,7 +101,7 @@ Drivers that implement a Connection Pool MUST support the following ConnectionPo
       maxPoolSize?: number;
 
       /**
-       *  The minimum number of connections that MUST exist at any moment
+       *  The minimum number of Connections that MUST exist at any moment
        *  in a single connection pool.
        *  If specified, MUST be an integer >= 0. If maxPoolSize is > 0
        *  then minPoolSize must be <= maxPoolSize
@@ -110,7 +110,7 @@ Drivers that implement a Connection Pool MUST support the following ConnectionPo
       minPoolSize?: number;
 
       /**
-       *  The maximum amount of time a connection should remain idle
+       *  The maximum amount of time a Connection should remain idle
        *  in the connection pool before being marked idle.
        *  If specified, MUST be a number >= 0.
        *  A value of 0 means there is no limit.
@@ -151,7 +151,7 @@ The following ConnectionPoolOptions are considered deprecated. They MUST NOT be 
     interface ConnectionPoolOptions {
       /**
        *  The maximum number of threads that can simultaneously wait
-       *  for a connection to become available.
+       *  for a Connection to become available.
        */
       waitQueueSize?: number;
 
@@ -169,7 +169,8 @@ Connection Pool Members
 Connection
 ----------
 
-A driver-defined wrapper around a single TCP/IP connection to an Endpoint. A `Connection <#connection>`_ has the following properties:
+A driver-defined wrapper around a single TCP connection to an
+Endpoint. A `Connection <#connection>`_ has the following properties:
 
 -  **Single Endpoint:** A `Connection <#connection>`_ MUST be associated with a single Endpoint. A `Connection <#connection>`_ MUST NOT be associated with multiple Endpoints.
 -  **Single Lifetime:** A `Connection <#connection>`_ MUST NOT be used after it is closed.
@@ -177,11 +178,15 @@ A driver-defined wrapper around a single TCP/IP connection to an Endpoint. A `Co
 -  **Single Track:** A `Connection <#connection>`_ MUST limit itself to one request / response at a time. A `Connection <#connection>`_ MUST NOT multiplex/pipeline requests to an Endpoint.
 -  **Monotonically Increasing ID:** A `Connection <#connection>`_ MUST have an ID number associated with it. `Connection <#connection>`_ IDs within a Pool MUST be assigned in order of creation, starting at 1 and increasing by 1 for each new Connection.
 -  **Valid Connection:** A connection MUST NOT be checked out of the pool until it has successfully and fully completed a MongoDB Handshake and Authentication as specified in the `Handshake <https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.rst>`__, `OP_COMPRESSED <https://github.com/mongodb/specifications/blob/master/source/compression/OP_COMPRESSED.rst>`__, and `Authentication <https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst>`__ specifications.
--  **Perishable**: it is possible for a connection to become **Perished**. A connection is considered perished if any of the following are true:
+-  **Perishable**: it is possible for a `Connection <#connection>`_ to become **Perished**. A `Connection <#connection>`_ is considered perished if any of the following are true:
 
-   -  **Stale:** The connection's generation does not match the generation of the parent pool
-   -  **Idle:** The connection is currently "available" and has been for longer than **maxIdleTimeMS**.
-   -  **Errored:** The connection has experienced an error that indicates the connection is no longer recommended for use. Examples include, but are not limited to:
+   - **Stale:** The `Connection <#connection>`_ 's generation does not
+      match the generation of the parent pool
+   - **Idle:** The `Connection <#connection>`_ is currently
+      "available" and has been for longer than **maxIdleTimeMS**.
+   - **Errored:** The `Connection <#connection>`_ has experienced an
+      error that indicates it is no longer recommended for
+      use. Examples include, but are not limited to:
 
       -  Network Error
       -  Network Timeout
@@ -193,18 +198,18 @@ A driver-defined wrapper around a single TCP/IP connection to an Endpoint. A `Co
 
     interface Connection {
       /**
-       *  An id number associated with the connection
+       *  An id number associated with the Connection
        */
       id: number;
 
       /**
-       *  The address of the pool that owns this connection
+       *  The address of the pool that owns this Connection
        */
       address: string;
 
       /**
        *  An integer representing the “generation” of the pool
-       *  when this connection was created
+       *  when this Connection was created
        */
       generation: number;
 
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -445,9 +445,8 @@ Populating the Pool with a Connection (Internal Implementation)
 marked as "available" and kept in the pool. This type of `Connection
 <#connection>`_ is created to ensure minPoolSize and is distinct from from the
 type that is created, established, and immediately checked out in checkOut. See
-`Checking Out a Connection
-<#populating-the-pool-with-a-connection-internal-implementation>`_ for more
-information on that type of `Connection <#connection>`_ creation.
+`Checking Out a Connection <#checking-out-a-connection>`_ for more information on
+that type of `Connection <#connection>`_ creation.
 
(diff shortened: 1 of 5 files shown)


source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -194,15 +208,37 @@ A driver-defined wrapper around a single TCP/IP connection to an Endpoint. A Con
...
+       *   - "available":     The Connection has been established and is waiting in the pool to be checked
+       *                      out. Contributes to both totalConnectionCount and availableConnectionCount.
+       *
+       *   - "in use":        The Connection has been established, checked out from the pool, and has yet
+       *                      to be checked back in. Contributes to totalConnectionCount.
+       *
+       *   - "closed":        The Connection has had its socket closed and cannot be used for any future
+       *                      operations. Does not contribute to any connection counts.
+       *
+       * Note: this field is mainly used for the purposes of describing state
```
patrickfreed, line 228: I hadn't seen this done before in other specs, but having this "spec-only" enumeration of states helps me visualize the lifecycle of a connection, and it also helps make the spec language / psuedocode more precise.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -344,7 +386,6 @@ Before a connection can be returned outside of the pool, it must be “set up”
       perform connection handshake
       handle OP_COMPRESSED
       perform connection authentication
-      mark connection as readyToUse
```
patrickfreed, line 347: a connection is now marked as "available" or "in use" depending on the context it's created in. No need for an additional state.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -214,24 +250,24 @@ Example implementations include:
...
 
 -  **Thread Safe:** All Pool behaviors MUST be thread safe.
 -  **Not Fork-Safe:** A Pool is explicitly not fork-safe. If a Pool detects that is it being used by a forked process, it MUST immediately clear itself and update its pid
 -  **Single Owner:** A Pool MUST be associated with exactly one Endpoint, and MUST NOT be shared between Endpoints.
 -  **Emit Events:** A Pool MUST emit pool events when dictated by this spec (see `Connection Pool Monitoring <#connection-pool-monitoring>`__). Users MUST be able to subscribe to emitted events in a manner idiomatic to their language and driver.
 -  **Closeable:** A Pool MUST be able to be manually closed. When a Pool is closed, the following behaviors change:
 
-   -  Checking in a Connection to the Pool automatically closes the connection
-   -  Attempting to check out a Connection from the Pool results in an Error
+   -  Checking in a ``Connection`` to the Pool automatically closes the ``Connection``
```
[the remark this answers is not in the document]
patrickfreed: updated to use links

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -214,24 +250,24 @@ Example implementations include:
 Connection Pool
 ---------------
 
-A driver-defined entity that encapsulates all non-monitoring connections associated with a single Endpoint. The pool has the following properties:
+A driver-defined entity that encapsulates all non-monitoring Connections associated with a single Endpoint. The pool has the following properties:
```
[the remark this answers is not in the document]
patrickfreed: yep, fixed.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -664,7 +766,7 @@ a manner idiomatic to the Driver and Language.
      *  Connection from a closed Connection Pool
      */
     interface PoolClosedError {
-      message: 'Attempted to check out a connection from closed connection pool';
+      message: 'Attempted to check out a Connection from closed connection pool';
```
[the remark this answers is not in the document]
patrickfreed: "Connection" is capitalized in some of the code sections and lower case in others, so I just capitalized all of them for consistency. For example, the first half of the [Events](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#events) section has it capitalized on master, but the second half does not (presumably different spec authors). Capitalizing them all also makes it consistent with the usage of "Endpoint" (one of the other defined terms), which is capitalized in all the code sections. Also, I wouldn't say these are like whitespace changes, since there is potential semantic meaning to differentiate here. I could go back and change them if you'd like, but I think there's value in keeping them (also it's extra work to undo them).

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -27,6 +27,19 @@ The keywords “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL N
 Definitions
 ===========
 
+Connection
+~~~~~~~~~~~~~~
+
+A Connection (when linked) refers to the ``Connection`` type defined
+in this specification. It does not refer to an actual TCP/IP
+connection to an Endpoint. A ``Connection`` will attempt to create and
```
p-mongo, line 35: Could "TCP/IP connection" be shortened to "TCP connection" and then used everywhere instead of plain lowercased "connection", such as on line 36 immediately below?

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -156,18 +170,18 @@ Connection Pool Members
...
 -  **Valid Connection:** A connection MUST NOT be checked out of the pool until it has successfully and fully completed a MongoDB Handshake and Authentication as specified in the `Handshake <https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.rst>`__, `OP_COMPRESSED <https://github.com/mongodb/specifications/blob/master/source/compression/OP_COMPRESSED.rst>`__, and `Authentication <https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst>`__ specifications.
 -  **Perishable**: it is possible for a connection to become **Perished**. A connection is considered perished if any of the following are true:
 
    -  **Stale:** The connection's generation does not match the generation of the parent pool
-   -  **Idle:** The connection is currently available and readyToUse, and has been for longer than **maxIdleTimeMS**.
+   -  **Idle:** The connection is currently "available" and has been for longer than **maxIdleTimeMS**.
```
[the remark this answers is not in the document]
p-mongo: I think using `"available" (as defined below)` would clarify that the definition is forthcoming.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -194,15 +207,37 @@ A driver-defined wrapper around a single TCP/IP connection to an Endpoint. A Con
        *  when this connection was created
        */
       generation: number;
+
+      /**
+       * The current state of the Connection.
+       *
+       * Possible values are the following:
+       *   - "unestablished": The Connection is empty. It has been created but has not opened a socket or
+       *                      performed any I/O. Contributes to totalConnectionCount.
```
p-mongo, line 216: I suggest avoiding negatives ("un"). Also, what is the state for when a socket was created but before the connection is established?

Possible wording:

> "pending": The Connection has been created but it is not yet established. Contributes to totalConnectionCount.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
p-mongo: :+1:

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -241,29 +278,29 @@ A driver-defined entity that encapsulates all non-monitoring connections associa
       generation: number;
     
       /**
-       *  An integer expressing how many total connections
+       *  An integer expressing how many total Connections
        *  (active + in use) the pool currently has
```
p-mongo, line 282: "active" is now not a defined state, I think this should say "pending" + "available" + "in use".

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -288,30 +325,31 @@ The SDAM specification defines `when
...
 the driver should create connection pools.
 
-Once a pool is created, if minPoolSize is set, the pool MUST immediately
-begin creating enough connections such that totalConnections >= minPoolSize.
-These connections MUST be created in a non-blocking manner. The driver SHOULD
-set up (connect, handshake and authenticate) these connections in a non-blocking
-manner, such as via the use of a background thread or asynchronous I/O.
+Once a pool is created, if minPoolSize is set, the pool MUST
+immediately begin populating enough `Connections <#connection>`_ such that
+totalConnections >= minPoolSize. These `Connections <#connection>`_ MUST be
```
p-mongo, line 330: I think the intention is to populate such that the number of available + in use connections is at least min pool size. Simply creating dummy connection objects doesn't achieve the goal of min pool size of having that many *ready* connections.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -288,30 +325,31 @@ The SDAM specification defines `when
...
 
 .. code::
 
     set generation to 0
     emit PoolCreatedEvent
     if minPoolSize is set:
+      # this MAY be performed on a background thread
+      # if it is not performed on a background thread, this MUST
+      # utilize non-blocking I/O.
       while totalConnectionCount < minPoolSize:
```
p-mongo, line 343: I suggest changing this to `while total count <= max pool size and available + in use count < min pool size`.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -355,29 +397,85 @@ Before a connection can be returned outside of the pool, it must be “set up”
 Closing a Connection (Internal Implementation)
 ----------------------------------------------
 
-When a Connection is closed, it MUST first be marked as closed, removing it from being counted as available, in use, or setup. Once the connection is marked as closed, the Connection can perform whatever teardown is necessary. The Driver SHOULD perform the teardown in a non-blocking manner.
+When a `Connection <#connection>`_ is closed, it MUST first be marked as "closed",
+removing it from being counted as "available" or "in use". One that is
+complete, the `Connection <#connection>`_ can perform whatever teardown is
+necessary to close its underlying socket. The Driver MUST perform this
```
p-mongo, line 403: The SHOULD -> MUST change here is significant and if it is desired to require that connections are closed in a non-blocking manner, I suggest this change is made in its own DRIVERS ticket.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -355,29 +397,85 @@ Before a connection can be returned outside of the pool, it must be “set up”
...
+teardown in a non-blocking manner, such as via the use of a background
+thread or async I/O.
 
 .. code::
 
-    mark connection as closed
+    decrement total connection count
+    if connection state is "available":
+      decrement available connection count
+    set connection state to "closed"
```
p-mongo, line 412: I believe the state should be changed first, otherwise connection accounting is going to be incorrect. It is more sensible to have some tracked connections that are closed than to have disappearing open connections.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -355,29 +397,85 @@ Before a connection can be returned outside of the pool, it must be “set up”
...
+   increment available connection count
+   set connection state to "available"
+   add connection to availableConnections
+
+
+Populating the Pool with a Connection (Internal Implementation)
+---------------------------------------------------------------
+
+If minPoolSize is set, the pool MUST ensure that it always has at
+least minPoolSize total `Connections <#connection>`_. In order to achieve this, it
```
p-mongo, line 438: The wording earlier in the spec ("the pool MUST immediately begin populating") was used to reflect the fact that connections are not created instantly. A pool that provides a minimum number of *usable* connections cannot *always* have that many connections.

A pool that provides any number of dummy connection objects can make them as soon as the pool is constructed. This is how the Ruby driver worked prior to 2.11. We do not consider this useful behavior because connection objects are lightweight - if the driver does not understand min pool size to mean that many *usable* connections, it's pointless to have a min pool size at all.

For this hunk, I suggest not requiring the min pool size with a MUST. Either omitting this sentence entirely or repeating "the pool MUST immediately begin populating" version could be viable alternatives.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -355,29 +397,85 @@ Before a connection can be returned outside of the pool, it must be “set up”
...
+needs to be able to create `Connections <#connection>`_ that begin as "available"
+instead of "in use" (i.e. "populate the pool"). Performing this
```
p-mongo, line 440: The "beginning as available vs in use" distinction is confusing to me. Conceptually all connections go through both states, i.e. each connection follows pending -> available -> in use. A driver could create all connections in background and block main thread until there is a connection created in the background, a driver could create connections in foreground only and a driver could do both. In Ruby we considered all of these approaches and settled on the hybrid one which provides the best responsiveness to application needs. But this is an implementation detail and I am of the opinion it does not belong in the spec.

Looking at this section (populating the pool), perhaps it can be removed completely because  I believe all of its content is already covered elsewhere in the spec.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -355,29 +397,85 @@ Before a connection can be returned outside of the pool, it must be “set up”
...
+process MUST NOT block any application threads. For example, it could
+be performed on a background thread or via the use of non-blocking
+I/O.
+
+.. code::
+
+   create connection
+   establish connection
+   mark connection as available
```
p-mongo, line 449: Maybe take this pseudocode, combine it with the previous section about making the connection available, and retitle that section something like "establishing a connection".

Note that, if I am not mistaken, connection accounting (i.e. counts of various connection types) needs to be performed by all of the code that works with connections, i.e. both background thread and main thread.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -355,29 +397,85 @@ Before a connection can be returned outside of the pool, it must be “set up”
...
+A Pool MUST have a method of allowing the driver to check out a `Connection <#connection>`_. Checking out a `Connection <#connection>`_ involves entering the WaitQueue and waiting for a `Connection <#connection>`_ to become available. If the thread times out in the WaitQueue, an error is thrown.
 
-If, in the process of iterating available connections in the pool by the checkOut method, a perished connection is encountered, such a connection MUST be closed and the iteration of available connections MUST continue until either a non-perished available connection is found or the list of available connections is exhausted. If no connections are available and the total number of connections is less than maxPoolSize, the pool MUST create and return a new Connection.
+If, in the process of iterating available `Connections <#connection>`_ in the pool
+by the checkOut method, a perished `Connection <#connection>`_ is encountered, such
+a `Connection <#connection>`_ MUST be closed and the iteration of available
```
p-mongo, line 459: The "MUST be closed" wording suggests the close is synchronous. If the close here is intended to be asynchronous, I suggest either using wording like "MUST be marked closed and scheduled for teardown" or referencing the closure section of the spec which discusses the background closure.

If the language mandating background closure is removed (the MUST reverted to SHOULD per my earlier comment), no change here is needed.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -355,29 +397,85 @@ Before a connection can be returned outside of the pool, it must be “set up”
...
+`Connections <#connection>`_ MUST continue until either a non-perished available
+`Connection <#connection>`_ is found or the list of available `Connections <#connection>`_ is
+exhausted. If no `Connections <#connection>`_ are available and the total number of
+`Connections <#connection>`_ is less than maxPoolSize, the pool MUST create and
+return a new `Connection <#connection>`_.
```
p-mongo, line 464: You may consider saying "a new *established* Connection" here.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -355,29 +397,85 @@ Before a connection can be returned outside of the pool, it must be “set up”
...
 
-If the pool is closed, any attempt to check out a connection MUST throw an Error, and any items in the waitQueue MUST be removed from the waitQueue and throw an Error.
+If the pool is closed, any attempt to check out a `Connection <#connection>`_ MUST throw an Error, and any items in the waitQueue MUST be removed from the waitQueue and throw an Error.
 
-If minPoolSize is set, the Connection Pool must have at least minPoolSize total connections. If the pool does not implement a background thread as specified here, the checkOut method is responsible for ensuring this requirement.
+If minPoolSize is set, the `Connection <#connection>`_ Pool MUST always have at
+least minPoolSize total `Connections <#connection>`_. If the pool does not
```
p-mongo, line 469: I believe the requirement of the original CMAP spec was min pool size of *available + in use* connections per another earlier comment, and I also find this requirement impractical to implement.

I believe simply changing to "total" connections also produces a poorly specified connection pool.

I suggest either changing this paragraph to use available + in use connections as per the original intent of the spec, or perhaps removing this paragraph completely.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -398,10 +496,11 @@ A Connection MUST NOT be checked out until it is readyToUse. In addition, the Po
         else if totalConnectionCount < maxPoolSize:
           connection = create connection
         # If there is no background thread, the pool MUST ensure that
-        # at least minPoolSize connections are in the pool.
-        # This SHOULD be done in a non-blocking manner
+        # there are at least minPoolSize total connections.
```
p-mongo, line 499: Same as previous comment, either use "available + in use" instead of "total" or remove the requirement completely.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -412,32 +511,36 @@ A Connection MUST NOT be checked out until it is readyToUse. In addition, the Po
       # This must be done in all drivers
       leave wait queue
 
-    # If the connection has not been connected yet, the connection
-    # (TCP, TLS, handshake, compression, and auth) must be performed
-    # before the connection is returned. This MUST NOT block other threads
-    # from acquiring connections.
-    if connection is not readyToUse:
+    # If the Connection has not been established yet (TCP, TLS,
+    # handshake, compression, and auth), it must be before it is returned.
```
p-mongo, line 515: Suggest "it must be established"

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -412,32 +511,36 @@ A Connection MUST NOT be checked out until it is readyToUse. In addition, the Po
...
 ------------------------
 
-A Pool MUST have a method of allowing the driver to check in a Connection. The driver MUST NOT be allowed to check in a Connection to a Pool that did not create that Connection, and MUST throw an Error if this is attempted.
+A Pool MUST have a method of allowing the driver to check in a
+`Connection <#connection>`_. The driver MUST NOT be allowed to check in a
+`Connection <#connection>`_ to a Pool that did not create that `Connection <#connection>`_, and
+MUST throw an Error if this is attempted.
 
-When the Connection is checked in, it is closed if any of the following are true:
+When the `Connection <#connection>`_ is checked in, it is closed if any of the following are true:
```
p-mongo, line 538: Suggest "it MUST be closed"

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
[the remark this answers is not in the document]
p-mongo: I like the capitalized Connection.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -716,6 +819,29 @@ Why is waitQueueTimeoutMS optional for some drivers?
 
 We are anticipating eventually introducing a single client-side timeout mechanism, making us hesitant to introduce another granular timeout control. Therefore, if a driver/language already has an idiomatic way to implement their timeouts, they should leverage that mechanism over implementing waitQueueTimeoutMS.
 
+Why must ensuring minPoolSize require the use of a background thread or async I/O?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Without the use of a background thread, minPoolSize is ensured during
+checkOut. `Connections <#connection>`_ may not be checked into the pool before
+being established, however, so establishment also happens during
```
p-mongo, line 827: Connections cannot be checked in before they are checked out, therefore  it does not make sense to say  a connection is checked in before being established. I suggest simply saying that a connection cannot be checked out prior to being established. 

Separately, I would say this sentence is not needed to clarify why background thread or async i/o is required - jumping from first to third sentence seems perfectly fine to me. The issue is not waiting for one connection to be established but waiting for a large number of connections to be established.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -716,6 +819,29 @@ Why is waitQueueTimeoutMS optional for some drivers?
...
+Why must closing a connection be non-blocking?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Because idle and perished `Connections <#connection>`_ are cleaned up as part of
+checkOut, performing blocking I/O while closing such `Connections <#connection>`_
+would block application threads, introducing unnecessary latency. Once
+a `Connection <#connection>`_ is marked as "closed", it will not be checked out
+again, so ensuring the socket is torn down does not need to happen
+immediately and can happen at a later time, either via async I/O or a
+background thread.
```
p-mongo, line 844: I agree with this rationale, however I think that calling `close` on a socket is actually already asynchronous, so that the drivers do not necessarily need to call close in a background thread.

Again it may be  a good idea to cover asynchronous close in a separate DRIVERS ticket.

patrickfreed: Sounds good to me. I updated the rationale title to be "Why should closing ..." so I could keep it in here for now, is that okay? I'll file a DRIVERS ticket once this PR is merged.

p-mongo: :+1:

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
p-mongo: :+1:

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -288,30 +329,32 @@ The SDAM specification defines `when
...
 the driver should create connection pools.
 
-Once a pool is created, if minPoolSize is set, the pool MUST immediately
-begin creating enough connections such that totalConnections >= minPoolSize.
-These connections MUST be created in a non-blocking manner. The driver SHOULD
-set up (connect, handshake and authenticate) these connections in a non-blocking
-manner, such as via the use of a background thread or asynchronous I/O.
+Once a pool is created, if minPoolSize is set, the pool MUST immediately begin
+being "populated" with enough `Connections <#connection>`_ such that
+totalConnections >= minPoolSize. The Pool MUST be "populated" in a non-blocking
```
p-mongo, line 334: The 'total >= min'  is  a valid requirement but  a weak one. I suggest either adding "The pool SHOULD populate its connections such that available + in use >= min" as a separate sentence or otherwise incorporating this language somehow.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -355,29 +402,93 @@ Before a connection can be returned outside of the pool, it must be “set up”
...
+
+
+Populating the Pool with a Connection (Internal Implementation)
+---------------------------------------------------------------
+
+"Populating" the pool involves creating and establishing a `Connection
+<#connection>`_ that will not immediately be checked out and will instead be
+marked as "available" and kept in the pool. This type of `Connection
+<#connection>`_ is created to ensure minPoolSize and is distinct from from the
+type that is created, established, and immediately checked out in checkOut. See
```
p-mongo, line 447: Conceptually I consider all connections to be of the same type and go through the same exact lifecycle. Hence it is weird to read "this type of connection is created...". I agree that the processes by which a Connection may be created can vary based on whether it is created in main thread or in a background thread.

Again I would simply drop this sentence, I am not seeing the value it adds to the spec.

p-mongo (approved):
I left two comments which are variations of comments from my previous review.

With that said, I believe this PR represents significant improvement as it is and I approve merging it in its current state if you have a decently strong preference for the current language.

mbroadst (approved):

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
patrickfreed: Assuming the steps in "populating the pool" are followed, `totalConnections >= minPoolSize` should be a sufficient requirement. To better reinforce this, I've hyperlinked "populated" to the "Populating the Pool" section and also added some wording within that section mentioning that the intention is that it ensures available + in use >= min (phrased a little differently in English though).

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
patrickfreed: Yeah, I don't think classifying them according to type is the best way to word this. I was attempting to classify the context of the creation rather than the type of the Connection itself, since as you point out, there aren't different types of Connections. The two creation contexts I'm trying to disambiguate are "on-demand" (e.g. checkOut is called and pool has no available connections) and "preemptive" (e.g. ensuring minPoolSize, which is done during population). 

To that end, I've struck the sentence about different types of Connections and rewritten the section to focus more on the preemptive nature of these Connection creations. I hope that clarifies this section a little more.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -156,19 +168,24 @@ Connection Pool Members
...
-   -  **Stale:** The connection's generation does not match the generation of the parent pool
-   -  **Idle:** The connection is currently available and readyToUse, and has been for longer than **maxIdleTimeMS**.
-   -  **Errored:** The connection has experienced an error that indicates the connection is no longer recommended for use. Examples include, but are not limited to:
+   - **Stale:** The `Connection <#connection>`_ 's generation does not
+      match the generation of the parent pool
+   - **Idle:** The `Connection <#connection>`_ is currently
+      "available" (as defined below) and has been for longer than **maxIdleTimeMS**.
+   - **Errored:** The `Connection <#connection>`_ has experienced an
+      error that indicates it is no longer recommended for
+      use. Examples include, but are not limited to:
```
ShaneHarvey, line 188: The formatting of Stale/Idle/Errored is a bit off. Can you fix it? See the rendered view: https://github.com/mongodb/specifications/blob/ee56397b89b2da14b9114d0d71f59edec694fbaf/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#id2

ShaneHarvey (changes requested):

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
patrickfreed: I think this might be a bug in GitHub's rst renderer? I can't seem to get this to format right.

patrickfreed: Okay looks like I got it. It was missing a space in the list of perished states.

mbroadst:
@patrickfreed looks good to merge, please go ahead and do so

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
p-mongo: :+1: The revised wording is great.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
p-mongo: :+1:  I think this is good now also.

PR #849, opened 2020-08-26
SPEC-1582 Parse $uuid fields as Binary subtype 4

prashantmital:
POC'ing this with PyMongo uncovered an issue - `$uuid` is also how UUIDs were represented in PyMongo's legacy extJSON format so  this might result in a backward-breaking change in the Python driver.

Also, there is no mention in this spec whether extJSON parsers should offer configuration options that allow the user to set UuidRepresentation (https://github.com/mongodb/specifications/blob/master/source/uuid.rst) and accordingly decode UUID fields (I'd guess that UNSPECIFIED would decode as Binary subtype 4, any other representation would decode as native UUID; on the generation/encoding side we'd need to decide whether native UUIDs will be encoded as subtype 4 or 3).

### DIFF
--- a/source/bson-corpus/tests/binary.json
+++ b/source/bson-corpus/tests/binary.json
@@ -42,6 +42,7 @@
         {
             "description": "subtype 0x04 UUID",
             "canonical_bson": "1D000000057800100000000473FFD26444B34C6990E8E7D1DFC035D400",
+            "canonical_extjson": "{\"x\" : { \"$binary\" : {\"base64\" : \"c//SZESzTGmQ6OfR38A11A==\", \"subType\" : \"04\"}}}",
             "degenerate_extjson": "{\"x\" : { \"$uuid\" : \"73ffd264-44b3-4c69-90e8-e7d1dfc035d4\"}}"
         },
         {
@@ -88,10 +89,6 @@
         }
     ],
     "parseErrors": [
-        {
-            "description": "$uuid missing hyphens",
-            "string": "{\"x\" : { \"$uuid\" : \"73ffd26444b34c6990e8e7d1dfc035d4\"}}"
-        },
         {
             "description": "$uuid wrong type",
             "string": "{\"x\" : { \"$uuid\" : { \"data\" : \"73ffd264-44b3-4c69-90e8-e7d1dfc035d4\"}}}"
--- a/source/bson-corpus/tests/binary.json
+++ b/source/bson-corpus/tests/binary.json
@@ -37,7 +37,8 @@
         {
             "description": "subtype 0x04",
             "canonical_bson": "1D000000057800100000000473FFD26444B34C6990E8E7D1DFC035D400",
-            "canonical_extjson": "{\"x\" : { \"$binary\" : {\"base64\" : \"c//SZESzTGmQ6OfR38A11A==\", \"subType\" : \"04\"}}}"
+            "canonical_extjson": "{\"x\" : { \"$binary\" : {\"base64\" : \"c//SZESzTGmQ6OfR38A11A==\", \"subType\" : \"04\"}}}",
+            "degenerate_extjson": "{\"x\" : { \"$uuid\" : \"73ffd264-44b3-4c69-90e8-e7d1dfc035d4\"}}"
         },
         {
             "description": "subtype 0x05",
--- a/source/bson-corpus/tests/binary.json
+++ b/source/bson-corpus/tests/binary.json
@@ -37,7 +37,11 @@
         {
             "description": "subtype 0x04",
             "canonical_bson": "1D000000057800100000000473FFD26444B34C6990E8E7D1DFC035D400",
-            "canonical_extjson": "{\"x\" : { \"$binary\" : {\"base64\" : \"c//SZESzTGmQ6OfR38A11A==\", \"subType\" : \"04\"}}}",
+            "canonical_extjson": "{\"x\" : { \"$binary\" : {\"base64\" : \"c//SZESzTGmQ6OfR38A11A==\", \"subType\" : \"04\"}}}"
+        },
+        {
+            "description": "subtype 0x04 UUID",
+            "canonical_bson": "1D000000057800100000000473FFD26444B34C6990E8E7D1DFC035D400",
             "degenerate_extjson": "{\"x\" : { \"$uuid\" : \"73ffd264-44b3-4c69-90e8-e7d1dfc035d4\"}}"
         },
         {
@@ -81,6 +85,20 @@
         {
             "description": "subtype 0x02 length negative one",
             "bson": "130000000578000600000002FFFFFFFFFFFF00"
+        }
+    ],
+    "parseErrors": [
+        {
+            "description": "$uuid missing hyphens",
+            "string": "{\"x\" : { \"$uuid\" : \"73ffd26444b34c6990e8e7d1dfc035d4\"}}"
+        },
+        {
+            "description": "$uuid wrong type",
+            "string": "{\"x\" : { \"$uuid\" : { \"data\" : \"73ffd264-44b3-4c69-90e8-e7d1dfc035d4\"}}}"
+        },
+        {
+            "description": "$uuid invalid value",
+            "string": "{\"x\" : { \"$uuid\" : \"73ffd264-44b3-90e8-e7d1dfc035d4\"}}"
         }
     ]
 }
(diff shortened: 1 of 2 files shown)


ShaneHarvey:
> POC'ing this with PyMongo uncovered an issue - $uuid is also how UUIDs were represented in PyMongo's legacy extJSON format so this might result in a backward-breaking change in the Python driver.

What is the BC break?

prashantmital:
@ShaneHarvey so far we've honored the UuidRepresentation when decoding `$uuid` fields. If representation is UNSPECIFIED, it is loaded as a Binary subtype 4, else it is loaded as native UUID. The spec change as currently written mandates that `$uuid` fields be decoded as Binary subtype 4 which means that it is no longer possible to retrieve `$uuid` fields as native UUIDs - this is the BC break.
We could continue to honor the representation but that would result in another strange behavior - round-tripping a `$uuid` field using any representation other than UNSPECIFIED would result in a Binary subtype 3 UUID field (since *_LEGACY representations are all encoded to extJSON as subtype 3).

ShaneHarvey:
This is my understanding of what the behavior should be for decoding `$uuid` for drivers that support uuidRepresentation:
```
unspecified -> Binary subtype 4
standard -> native uuid
pythonLegacy -> Binary subtype 4
javaLegacy -> Binary subtype 4
csharpLegacy -> Binary subtype 4
```

I think this behavior conforms to both specs. For pymongo specifically we can't do this until 4.0 because it would break JSON roundtripping of native uuids. So for pymongo 3.X we may need to tweak the behavior a bit. Maybe like this:
```
unspecified -> Binary subtype 4
standard -> native uuid
pythonLegacy -> native uuid
javaLegacy -> native uuid
csharpLegacy -> native uuid
```

prashantmital:
@ShaneHarvey yes, that is my assessment as well. 

@jyemin Even though the extJSON spec doesn't specify if/how to provide the user knobs to configure things like UuidRepresentation, I was wondering if I should language like this to the UUID parsing section:

> Parsers MAY allow users to configure the `UuidRepresentation <https://github.com/mongodb/specifications/blob/master/source/uuid.rst#specification>`_ to use when parsing UUID fields. UUIDs represented in extJSON as `$uuid`-fields should be transformed to Binary subtype 4 for all UuidRepresentations other than `STANDARD`, in which case it MAY be transformed to a language-native UUID. UUIDs represented using `$binary`-fields should follow the implicit decoding rules laid out in the `UUID specification <https://github.com/mongodb/specifications/blob/master/source/uuid.rst#specification>`_.

source/extended-json.rst
```diff
@@ -338,6 +338,48 @@ to parsing JSON numbers:
...
+in Groovy with the MongoDB Java Driver::
+
+  {"Binary": UUID.fromString("c8edabc3-f738-4ca3-b68d-ab92a91478a3")}
+
+is transformed into the following (newlines and spaces added for readability)::
+
+  {"Binary": {
+      "$binary": {
+          "base64": "o0w498Or7cijeBSpkquNtg==",
+          "subType": "03"}
```
jyemin, line 356: Let's change this to subtype 4, since that's the future for the java driver.  I recommend changing it here and in the complete example above.

Sad that we're showing Groovy syntax...  Feel free to change the whole thing to Python if you want.

source/extended-json.rst
```diff
@@ -338,6 +338,48 @@ to parsing JSON numbers:
...
+      }
+  }
+
+While this transformation preserves BSON subtype information (since
+UUIDs can be represented as BSON subtype 3 *and* 4), it is not very
+human readable. Since extended JSON representations of UUIDs feature
```
jyemin, line 362: Is `not very human readable` accurate?  It seems more relevant (and sticking more to fact than opinion) to say that base64 is not the standard way of representing UUIDs, so it's harder to compare against the string representation coming from platform libraries.

source/bson-corpus/tests/binary.json
```diff
@@ -37,7 +37,8 @@
         {
             "description": "subtype 0x04",
             "canonical_bson": "1D000000057800100000000473FFD26444B34C6990E8E7D1DFC035D400",
-            "canonical_extjson": "{\"x\" : { \"$binary\" : {\"base64\" : \"c//SZESzTGmQ6OfR38A11A==\", \"subType\" : \"04\"}}}"
+            "canonical_extjson": "{\"x\" : { \"$binary\" : {\"base64\" : \"c//SZESzTGmQ6OfR38A11A==\", \"subType\" : \"04\"}}}",
```
jyemin, line 40: Rather than change an existing test, add a new one that is equivalent but uses `$uuid`

source/bson-corpus/tests/binary.json
```diff
@@ -37,7 +37,8 @@
...
+            "degenerate_extjson": "{\"x\" : { \"$uuid\" : \"73ffd264-44b3-4c69-90e8-e7d1dfc035d4\"}}"
```
jyemin, line 41: Add tests in decodeErrors for various forms of bad `$uuid values, e.g. wrong type, wrong number of characters, missing dashes, etc.

jyemin:
> Even though the extJSON spec doesn't specify if/how to provide the user knobs to configure things like UuidRepresentation, I was wondering if I should language like this to the UUID parsing section:

I don't see the need for such language.  Transformation to platform types is at a level above the concerns of this specification.

source/extended-json.rst
prashantmital: Done.

mbroadst:
removing my requested approval, I think we have enough cooks in this kitchen :smile:

source/bson-corpus/tests/binary.json
```diff
@@ -82,5 +86,19 @@
             "description": "subtype 0x02 length negative one",
             "bson": "130000000578000600000002FFFFFFFFFFFF00"
         }
+    ],
+    "parseErrors": [
+        {
+            "description": "$uuid missing hyphens",
```
jyemin, line 92: I like this test, and think we should keep it, but seems out of line with the current MAY support strings without hyphens that you added to the spec.

source/bson-corpus/tests/binary.json
jyemin: See below, where I respectfully disagree about missing dashes.

source/bson-corpus/tests/binary.json
```diff
@@ -39,6 +39,11 @@
             "canonical_bson": "1D000000057800100000000473FFD26444B34C6990E8E7D1DFC035D400",
             "canonical_extjson": "{\"x\" : { \"$binary\" : {\"base64\" : \"c//SZESzTGmQ6OfR38A11A==\", \"subType\" : \"04\"}}}"
         },
+        {
+            "description": "subtype 0x04 UUID",
+            "canonical_bson": "1D000000057800100000000473FFD26444B34C6990E8E7D1DFC035D400",
+            "degenerate_extjson": "{\"x\" : { \"$uuid\" : \"73ffd264-44b3-4c69-90e8-e7d1dfc035d4\"}}"
```
jyemin, line 45: Any reason not to add canonical_extjson?

source/extended-json.rst
```diff
@@ -377,6 +378,10 @@ representation of the UUID from the previous example::
   }
 
 Parsers MUST interpret the ``$uuid`` key as BSON Binary subtype 4.
+Parsers MUST accept textual representations of UUIDs that omit the
+URN prefix (usually ``urn:uuid:``). Parsers MAY also accept textual
```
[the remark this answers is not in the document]
ShaneHarvey: Python already accepts with/without hyphens. I suspect that's what motivated Prashant to include this

prashantmital (commented):
Address comments and also did a quick POC on Python.

jmikola (approved):

jmikola:
FYI: I just approved the specific changes I had commented on. Will defer to @jyemin for his full review.

source/extended-json.rst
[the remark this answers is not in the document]
jyemin: OK, I accept this reasoning.  If we ever add generation of $uuid we can debate about whether canonical representation includes the `urn:uuid:` prefix, but that's for another day.

jyemin:
Changes look good, but I'd like to do a quick POC to make sure I didn't miss anything

jyemin (approved):

ShaneHarvey (approved):
LGTM

PR #852, opened 2020-09-01
SPEC-1747 Specify error label in InsertOne test

isabelatkinson:
https://jira.mongodb.org/browse/SPEC-1747

### DIFF
--- a/source/retryable-writes/retryable-writes.rst
+++ b/source/retryable-writes/retryable-writes.rst
@@ -3,14 +3,14 @@ Retryable Writes
 ================
 
 :Spec Title: Retryable Writes
-:Spec Version: 1.5.2
+:Spec Version: 1.5.3
 :Author: Jeremy Mikola
 :Lead: \A. Jesse Jiryu Davis
 :Advisors: Robert Stam, Esha Maharishi, Samantha Ritter, and Kaloian Manassiev
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: 3.6
-:Last Modified: 2020-02-25
+:Last Modified: 2020-09-01
 
 .. contents::
 
@@ -806,6 +806,9 @@ inconsistent with the server and potentially confusing to developers.
 
 Changes
 =======
+
+2020-09-01: State the the driver should only add the RetryableWriteError label
+to network errors when connected to a 4.4+ server.
 
 2020-02-25: State that the driver should only add the RetryableWriteError label
 when retryWrites is on, and make it clear that mongos will sometimes perform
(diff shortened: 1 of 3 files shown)


source/retryable-writes/tests/insertOne-serverErrors.json
```diff
@@ -966,7 +966,10 @@
           ],
           "writeConcernError": {
             "code": 91,
-            "errmsg": "Replication is being shut down"
+            "errmsg": "Replication is being shut down",
+            "errorLabels": [
+              "RetryableWriteError"
```
egiurleo, line 971: Can you remind me why it is necessary to place the error label inside the write concern error when it is returned as a top-level field on the error object itself?

isabelatkinson: Hm it looks like we do it both ways in these tests; for write concern errors specifically, the error is always embedded (example [here](https://github.com/mongodb/specifications/blob/e0f32a668c514ef5268c92e334ec48ba7b962790/source/retryable-writes/tests/deleteOne-serverErrors.json#L80)), but there are other failpoints where we set the label in the `data` field instead (example [here](https://github.com/mongodb/specifications/blob/e0f32a668c514ef5268c92e334ec48ba7b962790/source/retryable-writes/tests/bulkWrite-errorLabels.json#L34)). I embedded the error here for consistency with other write concern error failpoints, but it might be worth filing another ticket to make the label placement consistent across the board/with the server, which does return the label as a top-level field.

egiurleo: Ah okay that makes a lot of sense. Thank you for catching that. I have filed [DRIVERS-1385](https://jira.mongodb.org/browse/DRIVERS-1385) to escalate that issue.

source/retryable-writes/retryable-writes.rst
```diff
@@ -226,8 +226,8 @@ configuration option set to true.
 For server versions 4.4 and newer, MongoDB will add a RetryableWriteError label to
 errors or server responses that it considers retryable before returning them to the
 driver. As new server versions are released, the errors that are labeled with the
-RetryableWriteError label may change. When receiving a command result
-with an error from a 4.4+ server that supports retryable writes, the driver
+RetryableWriteError label may change. When receiving any result with an error other
+than a network error from a 4.4+ server that supports retryable writes, the driver
```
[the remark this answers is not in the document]
jmikola: Sorry for the delay. I wrote this yesterday and forgot to publish.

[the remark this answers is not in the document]
isabelatkinson: Updated

kmahar:
(removing myself, I'm unfamiliar with the spec and I think the other two reviewers should be sufficient)

jmikola (approved):

PR #854, opened 2020-09-16
Clarify that "invalid" cert means a revoked one


p-mongo:


kevinAlbs (approved):
LGTM. Seems like a good clarification.

alcaeus (approved):

PR #855, opened 2020-09-17
command_name is not an operations field

jmikola:
This might have been copypasta, but AFAIK `command_name` only exists within a command expectation. It was never a field for documents in `operations`.

### DIFF
--- a/source/client-side-encryption/tests/README.rst
+++ b/source/client-side-encryption/tests/README.rst
@@ -96,8 +96,6 @@ Each YAML file has the following keys:
 
     - ``collectionOptions``: |txn|
 
-    - ``command_name``: |txn|
-
     - ``arguments``: |txn|
 
     - ``result``: |txn|


kevinAlbs:
It is part of the operations for the `runCommand` operation: https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/tests/explain.yml#L18:L26

The transactions test runner has it too, but has more of a rationale why it's there: https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst#test-format

> command_name: Present only when name is "runCommand". The name of the command to run. Required for languages that are unable preserve the order keys in the "command" argument when parsing JSON/YAML.

jmikola:
Ah, yes -- thanks for the reminder. I didn't see this used anywhere in the FLE spec, but I could have missed it.

I'll just close this.

FWIW, this should be a problem for the unified format as I moved `commandName` into the args for the `runCommand` operation.

PR #857, opened 2020-09-24
SPEC-1753 Rate limit new connection creations (maxConnecting)

patrickfreed:
SPEC-1753

This PR updates the CMAP spec to require that drivers limit the number of concurrently establishing connections.

### DIFF
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -406,8 +406,8 @@ performs no I/O.
 .. code::
 
     connection = new Connection()
-    increment total connection count
-    increment pending connection count
+    increment totalConnectionCount
+    increment pendingConnectionCount
     set connection state to "pending"
     emit ConnectionCreatedEvent
     return connection
@@ -453,12 +453,12 @@ ConnectionCounts in any way and MUST NOT emit a ConnectionClosedEvent.
     set connection state to "closed"
 
     if original state is "available":
-      decrement available connection count
+      decrement availableConnectionCount
     else if original state is "pending":
-      decrement pending connection count
+      decrement pendingConnectionCount
 
     if original state != "unmanaged":
-      decrement total connection count
+      decrement totalConnectionCount
       emit ConnectionClosedEvent
 
     # The following can happen at a later time (i.e. in background
@@ -475,7 +475,7 @@ available `Connections <#connection>`_.
 
 .. code::
 
-   increment available connection count
+   increment availableConnectionCount
    set connection state to "available"
    add connection to availableConnections
 
@@ -596,14 +596,14 @@ Before a given `Connection <#connection>`_ is returned from checkOut, it must be
     if connection state is "pending":
       try:
         establish connection
-        decrement pending connection count
+        decrement pendingConnectionCount
       except connection establishment error:
         emit ConnectionCheckOutFailedEvent(reason="error")
-        decrement total connection count
-        decrement pending connection count
+        decrement totalConnectionCount
+        decrement pendingConnectionCount
         throw
     else:
-        decrement available connection count
+        decrement availableConnectionCount
     set connection state to "in use"
     emit ConnectionCheckedOutEvent
     return connection
@@ -660,14 +660,14 @@ monitoring threads.
 
    wait until pendingConnectionCount < maxConnecting
    connection = new Connection()
-   increment pending connection count
+   increment pendingConnectionCount
    try:
      connect connection via TCP / TLS
      set connection state to "unmanaged"
-     decrement pending connection count
+     decrement pendingConnectionCount
      return connection
    except error:
-     decrement pending connection count
+     decrement pendingConnectionCount
      throw error
 
 
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -528,7 +528,7 @@ are met or until a `Connection <#connection>`_ becomes available, re-entering
 the checkOut loop once it finishes waiting. This waiting MUST NOT block other
 threads from checking in `Connections <#connection>`_ to the pool. Threads that
 are waiting on the maxConnecting requirement to be met MUST receive the newly
-available `Connections <#connection>`_ in order that they entered the WaitQueue. 
+available `Connections <#connection>`_ in order that they entered the WaitQueue.
 
 If the pool is closed, any attempt to check out a `Connection <#connection>`_ MUST throw an Error, and any items in the waitQueue MUST be removed from the waitQueue and throw an Error.
 
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -566,12 +566,12 @@ Before a given `Connection <#connection>`_ is returned from checkOut, it must be
               close connection
               connection = Null
         else if totalConnectionCount < maxPoolSize:
-          if numConnecting < maxConnecting:
+          if pendingConnectionCount < maxConnecting:
             connection = create connection
           else:
             # this waiting MUST NOT prevent other threads from checking Connections
             # back in to the pool.
-            wait until numConnecting < maxConnecting or a connection is available
+            wait until pendingConnectionCount < maxConnecting or a connection is available
             continue
           
     except pool is closed:
@@ -587,7 +587,7 @@ Before a given `Connection <#connection>`_ is returned from checkOut, it must be
     # If there is no background thread, the pool MUST ensure that
     # there are at least minPoolSize total connections.
     # This MUST be done in a non-blocking manner
-    while totalConnectionCount < minPoolSize and numConnecting < maxConnecting:
+    while totalConnectionCount < minPoolSize and pendingConnectionCount < maxConnecting:
       populate the pool with a connection
 
     # If the Connection has not been established yet (TCP, TLS,
@@ -600,6 +600,7 @@ Before a given `Connection <#connection>`_ is returned from checkOut, it must be
       except connection establishment error:
         emit ConnectionCheckOutFailedEvent(reason="error")
         decrement total connection count
+        decrement pending connection count
         throw
     else:
         decrement available connection count
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -695,6 +695,9 @@ monitoring the state of all available `Connections <#connection>`_. This backgro
 thread SHOULD
 
 -  Populate `Connections <#connection>`_ to ensure that the pool always satisfies **minPoolSize**
+    - The background thread SHOULD just go back to sleep instead of waiting for
+      pendingConnectionCount to become less than maxConnecting when satisfying
+      minPoolSize.
 -  Remove and close perished available `Connections <#connection>`_.
 
 withConnection
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -595,12 +595,12 @@ Before a given `Connection <#connection>`_ is returned from checkOut, it must be
     if connection state is "pending":
       try:
         establish connection
-        decrement pendingConnectionCount
       except connection establishment error:
         emit ConnectionCheckOutFailedEvent(reason="error")
         decrement totalConnectionCount
-        decrement pendingConnectionCount
         throw
+      finally:
+        decrement pendingConnectionCount
     else:
         decrement availableConnectionCount
     set connection state to "in use"
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -526,11 +526,10 @@ maxConnecting, then the thread MUST wait until either both of those conditions
 are met or until a `Connection <#connection>`_ becomes available, re-entering
 the checkOut loop once it finishes waiting. This waiting MUST NOT block other
 threads from checking in `Connections <#connection>`_ to the pool. Threads that
-are waiting on the maxConnecting requirement to be met MUST receive the newly
-available `Connections <#connection>`_ in order that they entered the
-WaitQueue. For drivers that implement the WaitQueue via a fair semaphore, a
-second semaphore may be required to implement this waiting. Waiting on this
-semaphore SHOULD be limited by the WaitQueueTimeout, if the driver supports one.
+are waiting MUST be notified in order that they entered the WaitQueue. For
+drivers that implement the WaitQueue via a fair semaphore, a second semaphore
+may be required to implement this. Waiting on this second semaphore SHOULD be
+limited by the WaitQueueTimeout, if the driver supports one.
 
 If the pool is closed, any attempt to check out a `Connection <#connection>`_ MUST throw an Error, and any items in the waitQueue MUST be removed from the waitQueue and throw an Error.
 
@@ -586,7 +585,7 @@ Before a given `Connection <#connection>`_ is returned from checkOut, it must be
     # If there is no background thread, the pool MUST ensure that
     # there are at least minPoolSize total connections.
     # This MUST be done in a non-blocking manner
-    while totalConnectionCount < minPoolSize and pendingConnectionCount < maxConnecting:
+    while totalConnectionCount < minPoolSize:
       populate the pool with a connection
 
     # If the Connection has not been established yet (TCP, TLS,
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -547,7 +547,7 @@ Before a given `Connection <#connection>`_ is returned from checkOut, it must be
 "in use", and the pool's availableConnectionCount MUST be decremented.
 
 maxConnecting Implementation Note
-*********************************
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 For drivers that implement the WaitQueue via a fair semaphore, a second
 semaphore may be required to implement the waiting a thread needs to do if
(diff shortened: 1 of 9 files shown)


source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -455,18 +503,33 @@ I/O.
...
+If the list is exhausted, the total number of `Connections <#connection>`_ is
+less than maxPoolSize, and pendingConnectionCount < maxConnecting, the pool MUST
+create a `Connection <#connection>`_, establish it, mark it as "in use" and
+return it. If totalConnectionCount == maxPoolSize or pendingConnectionCount ==
+maxConnecting, then the thread MUST wait until either both of those conditions
+are met or until a `Connection <#connection>`_ becomes available, re-entering
+the checkOut loop once it finishes waiting. This waiting MUST NOT block other
+threads from checking in `Connections <#connection>`_ to the pool. Threads that
+are waiting MUST be notified in order that they entered the WaitQueue. For
+drivers that implement the WaitQueue via a fair semaphore, a second semaphore
```
ShaneHarvey, line 530: Using two semaphores with two distinct waiting steps can lead to unfair behavior. For example,

1. Thread A enters the wait queue 
1. Thread B enters the wait queue 
1. Thread A needs to create a connection but pendingConnectionCount == maxConnecting.
1. Another thread finishes an operation and checks a connection back into the pool.
1. Thread B is unblocked and checks out the connection before Thread A.
1. Another thread finishes creating its connection and decrements pendingConnectionCount.
1. Thread A is unblocked and creates a new connection (or grabs an available one) after Thread B.

Is that acceptable? I would argue yes and contend that strict fairness should not be a major concern. My reasoning is that the order the threads enter the wait queue is already non-deterministic. Actually this behavior exists even without maxConnecting: during the time it takes to create a new connection, an existing connection may be checked back into the pool which will unblock a "later" thread.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -853,6 +959,19 @@ again, so ensuring the socket is torn down does not need to happen
...
 
+What does the pool need to be able to create "Unmanaged Connections"?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In order to ensure that all TCP connections opened against the endpoint,
+including those that are created for the purposes of SDAM, are restricted and
+moderated equally (e.g. by maxConnecting), all TCP connection creation needs to
+go through the Pool. It is undesirable to use regular pooled connections for
+SDAM though, since the TCP connections used for monitoring do not need to
+authenticate, emit events, or compete for resources with application
```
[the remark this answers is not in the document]
patrickfreed: Yeah, I think better wording as to which resources we want to compete for and which we don't is needed here. That being said, I'm rethinking the purpose / necessity of having unmanaged connections at all given the new direction I'm thinking of taking the backoff portion in, so for now I'm going to hold off on updating their sections (bar a commit I already had locally and forgot to push).

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -455,18 +503,33 @@ I/O.
...
 
-A Pool MUST have a method of allowing the driver to check out a `Connection <#connection>`_. Checking out a `Connection <#connection>`_ involves entering the WaitQueue and waiting for a `Connection <#connection>`_ to become available. If the thread times out in the WaitQueue, an error is thrown.
-
-If, in the process of iterating available `Connections <#connection>`_ in the
-pool by the checkOut method, a perished `Connection <#connection>`_ is
-encountered, such a `Connection <#connection>`_ MUST be closed (as described in
-`Closing a Connection <#closing-a-connection-internal-implementation>`_) and the
-iteration of available `Connections <#connection>`_ MUST continue until either a
+A Pool MUST have a method of allowing the driver to check out a `Connection
+<#connection>`_. Checking out a `Connection <#connection>`_ involves entering
```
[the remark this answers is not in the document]
ShaneHarvey: Can we at least use this less verbose link format: 
```
`Connection`_
```

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -455,18 +503,33 @@ I/O.
...
+Connection <#closing-a-connection-internal-implementation>`_) and the iteration
+of available `Connections <#connection>`_ MUST continue until either a
 non-perished available `Connection <#connection>`_ is found or the list of
-available `Connections <#connection>`_ is exhausted. If no `Connections
-<#connection>`_ are available and the total number of `Connections
-<#connection>`_ is less than maxPoolSize, the pool MUST create a `Connection
-<#connection>`_, establish it, mark it as "in use" and return it.
+available `Connections <#connection>`_ is exhausted.
+
+If the list is exhausted, the total number of `Connections <#connection>`_ is
```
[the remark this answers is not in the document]
ShaneHarvey: Acknowledged.

ShaneHarvey (commented):
Not sure if you're aware but the travis rst job is failing https://travis-ci.org/github/mongodb/specifications/jobs/731034866:
```
$ rstcheck -r source/ --ignore-language c,python,java --report error
source/server-discovery-and-monitoring/server-monitoring.rst:148: (ERROR/3) Unknown target name: "a monitor should not use the client's regular connection pool".
```

source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-is-enforced.yml
```diff
@@ -0,0 +1,75 @@
...
+  - name: wait
+    thread: thread3
+    ms: 100
+  - name: checkOut
+    thread: thread3
+  # wait until all Connections have been created.
+  - name: waitForEvent
+    event: ConnectionReady
+    count: 3
+events:
```
ShaneHarvey, line 49: Have you implemented these tests? They are consistently failing for me. For example in this case the actual events I get before filtering out ignored events are:
```
PoolCreatedEvent(('localhost', 27017), {'maxPoolSize': 10, 'waitQueueTimeoutMS': 5000.0})
ConnectionCheckOutStartedEvent(('localhost', 27017))
ConnectionCreatedEvent(('localhost', 27017), 1)
ConnectionReadyEvent(('localhost', 27017), 1)
ConnectionCheckedOutEvent(('localhost', 27017), 1)
ConnectionCheckOutStartedEvent(('localhost', 27017))
ConnectionCreatedEvent(('localhost', 27017), 2)
ConnectionReadyEvent(('localhost', 27017), 2)
ConnectionCheckedOutEvent(('localhost', 27017), 2)
ConnectionCheckOutStartedEvent(('localhost', 27017))
ConnectionCreatedEvent(('localhost', 27017), 3)
ConnectionReadyEvent(('localhost', 27017), 3)
ConnectionCheckedOutEvent(('localhost', 27017), 3)
```

Interestingly, when I increase the failCommand `times` from 3 to 20, the test succeeds. This indicates that other connections are consuming the failpoint. In python's test suite there are many clients running at a time. Ideally we could use appName to restrict this fail point to only a single client but that is blocked by this bug: https://jira.mongodb.org/browse/SERVER-49220

Can you increase  `times`?

source/connection-monitoring-and-pooling/tests/pool-checkout-returned-connection-maxConnecting.yml
```diff
@@ -0,0 +1,106 @@
...
+  # hold it until the other threads are blocked
+  # on maxConnecting.
+  - name: start
+    target: thread1
+  - name: checkOut
+    thread: thread1
+    label: conn0
+  # Start 2 new threads that wait for thread1 to check
+  # out its connection before attempting to check out their own.
+  # This fills maxConnecting for the next 750ms.
```
ShaneHarvey, line 30: Why do these threads need to call waitForEvent? Can we instead wait for ConnectionCheckedOut on the main thread before starting these threads?
```
  - name: waitForEvent
    event: ConnectionCheckedOut
    count: 1
# Start 2 new threads to fill maxConnecting for the next 750ms.
...
```

source/connection-monitoring-and-pooling/tests/pool-checkout-returned-connection-maxConnecting.yml
```diff
@@ -0,0 +1,106 @@
...
+  mode: { times: 3 }
+  data:
+    failCommands: ["isMaster"]
+    closeConnection: false
+    blockConnection: true
+    blockTimeMS: 750
+poolOptions:
+  maxPoolSize: 10
+  waitQueueTimeoutMS: 5000
+operations:
```
ShaneHarvey, line 19: This test fails for me (even with `times: 10`). Could we simplify this entire test by:
- create one connection on the main thread, `conn0`
- start 3 threads that all perform a single `checkOut`
- Wait for 3 `ConnectionCheckOutStarted` events on the main thread 
- checkin `conn0` on the main thread
- Wait for all 3 threads to complete
- Assert that only 3 total connections were created, not 4.

patrickfreed: Done, although I don't think I needed to change the event assertions to match this scheme.

Re: failure with high `times`, should I set this to `alwaysOn` as well?

source/connection-monitoring-and-pooling/tests/pool-checkout-returned-connection-maxConnecting.yml
patrickfreed: Done

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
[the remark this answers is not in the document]
ShaneHarvey: Oh for the plural form you can do this:

```
Definitions
===========

.. _Connections: #Connection

Connection
~~~~~~~~~~
```

source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-is-enforced.yml
ShaneHarvey: > We don't use clients to run the CMAP tests, so we don't have any interference coming from background monitors

In python we have to use a MongoClient because that's the entity which handles the minPoolSize/maxIdleTime requirements. Besides that we also have various other shared MongoClients open, for example our test suite creates one meta client that stays open for all tests. This meta client can inadvertently trigger the failpoint. 

> Should I set it to alwaysOn to ensure there's absolutely no number of concurrent clients that could interfere? Or is it safer to specify a high-but-finite value?

high-but-finite value is safe imo.

source/connection-monitoring-and-pooling/tests/pool-checkout-returned-connection-maxConnecting.yml
ShaneHarvey: high-but-finite value is safer than alwaysOn imo.

source/connection-monitoring-and-pooling/tests/pool-checkout-returned-connection-maxConnecting.yml
```diff
@@ -0,0 +1,83 @@
...
+    address: 42
+  - type: ConnectionCreated
+    address: 42
+  # main thread checking its Connection back in
+  - type: ConnectionCheckedIn
+    connectionId: 1
+    address: 42
+  # remaining thread checking out the returned Connection
+  - type: ConnectionCheckedOut
+    connectionId: 1
```
ShaneHarvey, line 72: In Python this test fails here because we can't guarantee that `connectionId: 1` is checked out first. The issue is that using two semaphores is not fair (same problem discussed here: https://github.com/mongodb/specifications/pull/857#discussion_r495277704)

To address this I think we have to remove this line and omit connectionId.

patrickfreed: I left a similar comment on the Python PR about this--I think we should strive to pass the test as it is currently written if possible. It could really help out in situations where a ton of threads are queued up behind maxConnecting.

The following line in the version of the spec in this PR was meant to guard against this situation in most cases:

>Threads that are waiting [for pendingConnectionCount < maxConnecting or an available Connection] MUST be notified in order that they entered the WaitQueue."

To clarify my comment in the other thread: I don't think we need to absolutely guarantee fairness in every scenario, but I think we should try to eliminate unneeded waiting as much as possible and also ensure we avoid extremely unfair situations. In a scenario where we have a huge burst of operations that doesn't remain constant, we could have a lot of threads queued up for maxConnecting that aren't taking advantage of returned connections as quickly as they could be and subsequently receive their connection much later than newer threads. 

I now realize this is trickier to implement than just using another semaphore though, so if this ends up being really tricky / impossible to implement in Python, then I think we can just make it a SHOULD in the spec and edit this test as you suggest.

source/connection-monitoring-and-pooling/tests/pool-checkout-returned-connection-maxConnecting.yml
patrickfreed: increased to 10. I imagine once we hash out the fairness issue in the other thread, 10 should be high enough to pass this test too.

source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-is-enforced.yml
patrickfreed: Increased `times` to 10.

I eagerly await the resolution of SERVER-49220. This seems like a dirty solution lol.

source/connection-monitoring-and-pooling/tests/pool-checkout-returned-connection-maxConnecting.yml
ShaneHarvey: Yeah I agree. I fixed this in the python implementation by using a condition variable instead of a semaphore.

patrickfreed:
The travis failure appears to be something I broke in the Server Monitoring spec, which I'll just revert if we remove "Unmanaged Connections".

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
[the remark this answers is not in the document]
patrickfreed: Oh sweet, I didn't know about that functionality. I'll follow up with another PR and do all of them at once then.

patrickfreed:
A couple of updates based on the Python POC:
- Changed the suggestion of a second semaphore to a condition variable
- Suggested that the background thread, if present, simply go back to sleep instead of waiting on maxConnecting

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -527,9 +527,9 @@ a `Connection`_ becomes available, re-entering the checkOut loop once it
 finishes waiting. This waiting MUST NOT block other threads from checking in a
 `Connection`_ to the pool. Threads that are waiting MUST be notified in order
 that they entered the WaitQueue. For drivers that implement the WaitQueue via a
-fair semaphore, a second semaphore may be required to implement this. Waiting on
-this second semaphore SHOULD be limited by the WaitQueueTimeout, if the driver
-supports one.
+fair semaphore, a condition variable may also be required to implement
+this. Waiting on the condition variable SHOULD be limited by the
+WaitQueueTimeout, if the driver supports one.
```
ShaneHarvey, line 532: Can you add test coverage for this timeout behavior? It should be possible by blocking two connection establishments longer than the configured waitQueueTimeoutMS (using failCommand block) and then running a 3rd checkout attempt which will timeout waiting on maxConnecting.

patrickfreed (commented):
A few more updates:
- removed the "Unmanaged Connections" diffs
- added test for timeouts while blocked on maxConnecting
- upped the times on the failpoints to 20 after POCing in Rust

The only remaining test I think we might want is to check if minPoolSize population respects maxConnecting, but given the differing ways in which drivers implement the minPoolSize requirement, I'm not sure if a single spec test is possible.

source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-timeout.yml
```diff
@@ -0,0 +1,71 @@
...
+  # start a third thread that will be blocked waiting for
+  # one of the other two to finish
+  - name: start
+    target: thread3
+  - name: checkOut
+    thread: thread3
+  - name: waitForEvent
+    event: ConnectionCheckOutFailed
+    count: 1
+  # rejoin thread3, should experience error
```
ShaneHarvey, line 43: There's a slight problem with this test. This `waitForThread` operation does correctly cause Python's test runner to raise a WaitQueueTimeoutError exception which matches the `error` field. However, this means that the final `waitForEvent` (for ConnectionReady) never runs and therefore ConnectionReady does not appear in `events`.

I think python is following the spec test readme's behavior. Should we remove the `waitForEvent` and the `ConnectionReady` events? Or we could remove the `waitForThread` (thread 3) and the `error` field since we already validate that a ConnectionCheckOutFailed is emitted.

patrickfreed: In the Rust runner we actually continue to execute operations after receiving any errors from `waitForThread` and then compare at the end, though it does seem that it was the intention of the readme that the main thread just stops executing once it gets an error. I think the easiest approach here is just to drop the stuff that comes after the `waitForThread`. This doesn't really reduce our coverage at all either.

ShaneHarvey: SGTM. This test now passes in Python.

ShaneHarvey (approved):
LGTM!

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -455,18 +469,33 @@ I/O.
...
-<#connection>`_ are available and the total number of `Connections
-<#connection>`_ is less than maxPoolSize, the pool MUST create a `Connection
-<#connection>`_, establish it, mark it as "in use" and return it.
+A Pool MUST have a method of allowing the driver to check out a `Connection`_.
+Checking out a `Connection`_ involves entering the WaitQueue, waiting to be
+granted access to the list of available connections, and finding or creating a
+`Connection`_ to be returned. If the thread times out in the WaitQueue, an error
+is thrown.
+
+Once reaching the front of the WaitQueue, a thread begins iterating over the
```
ShaneHarvey, line 478: Discussion in meeting, we should replace "thread" with "request":
> Once reaching the front of the WaitQueue, a request begins iterating over the

patrickfreed: I opted for slightly different wording (i.e. "the pool begins iterating...") to avoid having the request do any work / seem like a thread. Let me know what you think.

ShaneHarvey (approved):
Tests pass in python (https://evergreen.mongodb.com/version/5f909deb2a60ed7372d2a6c5).

mbroadst (approved):
LGTM!

PR #859, opened 2020-09-30
SPEC-1548 Allow 0 for min/max pool size and fully test cmap URI options

saghm:
Since this bug happened due to the CMAP options being added without sufficient tests, I added specific tests for both minPoolSize and maxPoolSize to the comprehensive test of CMAP URI options well as for the specific values of 0.

### DIFF
--- a/source/uri-options/uri-options.rst
+++ b/source/uri-options/uri-options.rst
@@ -194,10 +194,10 @@ pertaining to URI options apply here.
      - The maximum replication lag, in wall clock time, that a secondary can suffer and still be eligible for server selection
 
    * - minPoolSize
-     - positive integer
+     - non-negative integer
      - defined in the `Connection Pooling spec`_
      - required for drivers with connection pools
-     - The maximum number of clients or connections able to be created by a pool at a given time
+     - The number of connections the driver should create and main in the pool even when no operations are occurring 
 
    * - readConcernLevel
      - any string (`to allow for forwards compatibility with the server <https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#unknown-levels-and-additional-options-for-string-based-readconcerns>`_)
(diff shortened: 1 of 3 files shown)


divjotarora (approved):

behackett:
What does 0 mean in most drivers? "No limit" or "no pooled connections"?

saghm:
maxPoolSize=0 is already described in the CMAP spec as meaning "no limit". For some historical context, when the URI options spec was first written, the CMAP spec didn't exist yet, and `maxPoolSize` was defined as being only a positive integer. CMAP was later written, and from the start defined `maxPoolSize=0` as meaning no limit on the pool size, but the PR to the URI options spec add the new information from the CMAP spec mistakenly omitted updating the description of what values are allowed. SPEC-1548 was filed after @dmitry-lukyanov noticed this discrepancy, and based on both the fact that disallowing 0 would cause breaking changes in drivers that followed the CMAP spec and the assumption that drivers are more likely to have implemented their connections pools following the CMAP spec's guidelines rather than the URI option spec's description, this PR is updating the URI options spec to follow to the language of the CMAP spec.

behackett:
Great. Thanks for the explainer. Carry on.

alcaeus:
@saghm shouldn't the language for `minPoolSize` also be changed to "non-negative integer", especially since you add tests confirming that specifying "0" for `minPoolSize` shouldn't break?

saghm:
Yes, you're correct! I've updated it now

DmitryLukyanov (approved):
LGTM

source/uri-options/uri-options.rst
```diff
@@ -194,10 +194,10 @@ pertaining to URI options apply here.
      - The maximum replication lag, in wall clock time, that a secondary can suffer and still be eligible for server selection
 
    * - minPoolSize
-     - positive integer
+     - non-negative integer
      - defined in the `Connection Pooling spec`_
      - required for drivers with connection pools
-     - The maximum number of clients or connections able to be created by a pool at a given time
+     - The number of connections the driver should create and main in the pool even when no operations are occurring
```
alcaeus, line 200: Wording is off here. "should create and keep in the pool" sounds more like what's intended.

alcaeus (changes requested):

alcaeus (approved):

PR #861, opened 2020-10-12
DRIVERS-1204 Add general logging specification

kmahar:
PR containing everything from the current Google doc that is related to logging in general (so nothing specific to commands).

Leaving as a draft for now since I am not sure the new truncation rules have been POCed anywhere besides Swift yet so those may be subject to change if we run into any major issues. I think those are the only recent major changes.

There are some very minor changes here from the Google doc in terms of formatting, e.g. I couldn't figure out how to get nested bullet point lists within a table so I got rid of those and just used new lines for each one.

I noticed some slightly more substantial issues doing this transcription to RST (e.g. the max document length variable did not start with `MONGODB_LOGGING`, only `MONGODB`) so I corrected those both here and in the Google doc.

### DIFF
--- a/source/logging.rst
+++ b/source/logging.rst
@@ -1,0 +1,309 @@
+=======
+Logging
+=======
+
+:Spec Title: Logging
+:Spec Version: 1.0.0
+:Author: Kaitlin Mahar
+:Spec Lead: Jeff Yemin
+:Advisors: Andreas Braun, James Kovacs, Rachelle Palmer, Isabella Siu
+:Status: Draft
+:Type: Standards
+:Minimum Server Version: N/A
+:Last Modified: 2020-10-12
+
+.. contents::
+
+--------
+
+Abstract
+========
+
+This specification defines requirements for drivers' logging architecture and
+behavior.
+
+META
+====
+
+The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in `RFC 2119 <https://www.ietf.org/rfc/rfc2119.txt>`__.
+
+Specification
+=============
+
+Terms
+-----
+
+Structured/Unstructured Logging
+  *Structured* logging refers to producing log messages in a structured format,
+  i.e. a series of key-value pairs, which can be converted to external
+  representations such as JSON. *Unstructured* logging refers to producing
+  string log messages.
+
+Implementation Requirements
+---------------------------
+Drivers SHOULD implement support for logging in a manner that is idiomatic for
+their language and ecosystem. 
+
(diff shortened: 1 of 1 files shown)


p-mongo:
Was LOG considered instead of LOGGING to save 4 characters on each environment variable name? If yes and LOG was rejected, can the rationale for that be added to the spec please?

kmahar:
@p-mongo I don't think we discussed it, but that makes sense to me. I'm fine with dropping the extra letters if no one objects.

alcaeus:
> I'm fine with dropping the extra letters if no one objects.

No objections from my end.

source/logging.rst
```diff
@@ -0,0 +1,309 @@
...
+   :widths: 1 1 1 1
+
+   * - Code
+     - Level Name
+     - Meaning
+     - Environment Variable value (case-insensitive)
+
+   * - N/A
+     - Off
+     - Nothing is logged.
```
alcaeus, line 180: I don't think `off` should be listed here, as it isn't a severity but a special configuration value. `off` may not be used when logging messages, but can be used by user to explicitly disable logging.

My suggestion would be to remove it here and add a paragraph on how to disable logging in the environment variable section above.

source/logging.rst
```diff
@@ -0,0 +1,309 @@
...
+     - trace
+
+Note that the Emergency, Alert, and Critical levels have been intentionally left
+undefined. At the time of writing this specification, we do not expect any
+driver specifications to need to log at these levels, but we have included them
+in the list of permitted levels for consistency with Syslog and so that they
+may be used in the future if needed.
+
+Not all logging frameworks will necessarily support all of these levels. If an
+equivalent level is not available, drivers SHOULD emit messages for that level
```
alcaeus, line 240: I think "MUST" is more appropriate here in favour of "SHOULD", as the rules for dealing with missing severities are quite clear.

At the same time, I'm wondering whether downgrading is the most sensible options, or whether we want error messages (severities "error", "critical", "alert", and "emergency") to be upgraded if they don't exist. For informational messages, it's ok if they get missed but we should err on the side of caution when logging error messages.

kmahar: Using MUST over SHOULD is fine with me.

Re missing messages: from my research into the logging frameworks various drivers have said they would use, I believe the existence of an "error" level is universal.

So I guess the question is really what do we do with "critical", "alert", and "emergency" level messages if their corresponding levels don't exist. I'm fine with mapping them to more severe levels if so, though we could also defer a decision on that until if the time ever comes when we actually need a level higher than "error".

alcaeus: Deferring that decision until we have a need for those levels sounds fine to me.

kmahar: I've switched to using MUST. I also added a note in the future work section about defining the unused levels, which mentions the spec should define how to map them to other levels if they are unsupported.

kmahar:
I'm going to close this, will reopen when it's ready for reviewers.

PR #863, opened 2020-10-15
Change targetedFailPoint operation to allow creating a new MongoClient

divjotarora:


source/unified-test-format/unified-test-format.rst
```diff
@@ -1465,20 +1468,18 @@ The following arguments are supported:
 
 - ``session``: Required string. See `commonOptions_session`_.
 
-  The client entity associated with this session SHOULD specify true for
```
divjotarora, line 1468: This first paragraph doesn't belong here IMO because it's talking about options for a client entity. I moved it up to where the `useMultipleMongoses` option for client entities is described.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1465,20 +1468,18 @@ The following arguments are supported:
...
+to a mongos server at the time this operation is executed. The new
+MongoClient MUST be used to run the ``configureFailPoint`` command and MUST
+be closed after the command has finished executing executing.
 
 When executing this operation, the test runner MUST keep a record of both the
-fail point and session (or pinned mongos server) so that the fail point can be
-disabled on the same mongos server after the test. The test runner MUST also
-ensure that the ``configureFailPoint`` command is excluded from the list of
-observed command monitoring events for this client (if applicable).
+fail point and pinned mongos server address so that the fail point can be
```
divjotarora, line 1479: The spec previously said that the test runner should track either the session of the pinned server. The session's state can change, however, so I don't think it's safe to assume that the test runner will be able to extract the previously pinned server's address from the session once the test is done. I changed this to just track the pinned server's address.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1465,20 +1468,18 @@ The following arguments are supported:
...
+Once the ``session`` argument has been resolved to a session entity, the test
+runner MUST create a new MongoClient that is directly connected to the
+session's pinned server. Test runners MUST error if the session is not pinned
+to a mongos server at the time this operation is executed. The new
+MongoClient MUST be used to run the ``configureFailPoint`` command and MUST
+be closed after the command has finished executing executing.
 
 When executing this operation, the test runner MUST keep a record of both the
-fail point and session (or pinned mongos server) so that the fail point can be
-disabled on the same mongos server after the test. The test runner MUST also
```
divjotarora, line 1479: The bit about ignoring the `configureFailPoint` command no longer applies because we're not creating a client entity that could be referenced in a `expectEvents` directive.

source/unified-test-format/unified-test-format.rst
jmikola: Why is it necessary to use a new MongoClient for this operation? That was never mandated by the [`targetedFailPoint` operation in the Transactions test runner](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/README.rst#targetedfailpoint).

Perhaps this didn't come up in the PHP POC because I was able to directly execute `configureFailPoint` on the pinned server without creating a new MongoClient with a direct connection to the mongos (but that may just be thanks to functionality in libmongoc and PHP that allows us to bypass server selection if we already have a specific server in mind).

source/unified-test-format/unified-test-format.rst
```diff
@@ -1465,20 +1468,18 @@ The following arguments are supported:
...
-The MongoClient and mongos on which to run the ``configureFailPoint`` command is
-determined by the ``session`` argument (after resolution to a session entity).
-Test runners MUST error if the session is not pinned to a mongos server at the
-time this operation is executed.
+Once the ``session`` argument has been resolved to a session entity, the test
+runner MUST create a new MongoClient that is directly connected to the
+session's pinned server. Test runners MUST error if the session is not pinned
+to a mongos server at the time this operation is executed. The new
+MongoClient MUST be used to run the ``configureFailPoint`` command and MUST
+be closed after the command has finished executing executing.
```
jmikola, line 1476: "executing executing"

source/unified-test-format/unified-test-format.rst
divjotarora: I didn't realize that PHP had this functionality but I talked to @alcaeus about PHP's implementation earlier and I think it's fine to do it that way if the driver provides that kind of functionality. This wouldn't be possible in Go, though. I'm going to change the PR to be more like the original prose but provide an alternative implementation option for drivers like Go.

alcaeus (approved):
👍

ShaneHarvey (approved):

PR #864, opened 2020-10-15
SPEC-1758 MONGODB-AWS test auth via EC2 Instance with IMDSv2

bazile-clyde:
So this PR clarifies how to use the IDMSv2 and adds "MUST" for the PUT requests. It also adds a mock EC2 server for testing but I'm not sure how to (or if we should) enforce testing against the mock server. The shell does enforce this though with integration tests using this mock and shell scripts. The mock fails to return a session token for any requests that aren't PUT requests but EC2 instances (as you've already seen) do allow other types of requests. 

Some background info form [here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html) and [here](https://aws.amazon.com/blogs/security/defense-in-depth-open-firewalls-reverse-proxies-ssrf-vulnerabilities-ec2-instance-metadata-service/) if you'd like to read more about it.

> HTTP GET and HEAD methods are allowed in IMDSv2 instance metadata requests. PUT requests are rejected if they contain an X-Forwarded-For header.

The reason Drivers use PUT requests is to mitigate SSRF attacks

> IMDSv2’s combination of beginning a session with a PUT request, and then requiring the secret session token in other requests, is always strictly more effective than requiring only a static header. AWS analysis of real-world vulnerabilities found that this combination protects against the vast majority of SSRF vulnerabilities.


ShaneHarvey:
Removing myself because I won't have time to review this soon. My only comment for now is that it would be great to document how the driver can use this new script. An example of running the mock server and communicating to it with CURL, showing the expected output for PUT and expected error for POST, would be great.

source/auth/auth.rst
```diff
@@ -954,25 +954,35 @@ credentials. Querying the URI will return the JSON response:
...
+The curl recipe below demonstrates the above. It retrieves a secret token that's valid for 30 seconds. It then uses that token to access the EC2 instance's credentials:
 
 .. code:: shell-session
 
     $ TOKEN=`curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds: 30"`
     $ ROLE_NAME=`curl http://169.254.169.254/latest/meta-data/iam/security-credentials/ -H "X-aws-ec2-metadata-token: $TOKEN"`
     $ curl http://169.254.169.254/latest/meta-data/iam/security-credentials/$ROLE_NAME -H "X-aws-ec2-metadata-token: $TOKEN"
-
-The JSON response will have the format:
+Drives can test this process using the mock EC2 server in `aws_services_lib <https://github.com/mongodb/specifications/blob/master/source/auth/tests/aws_services_lib>`_. The JSON response will have the format:
```
emadum, line 985:
```suggestion
Drivers can test this process using the mock EC2 server in `aws_services_lib <https://github.com/mongodb/specifications/blob/master/source/auth/tests/aws_services_lib>`_. The JSON response will have the format:
```

source/auth/tests/aws_services_lib/README.md
```diff
@@ -0,0 +1,6 @@
+=================
+Mock AWS Services
+=================
+
+This folder contains files from [mongo-enterprise-modules](https://github.com/10gen/mongo-enterprise-modules/tree/master/jstests/external_auth/lib). 
+The files simulate AWS services, and implementers should use them for local testing.
```
emadum, line 6: It looks like `mongo-enterprise-modules` is a private repo, is it OK/necessary to put those files into this public repo? To me it would seem sufficient to provide the above link and some simple instructions about using the `ec2_metadata_http_server.py` script for any drivers engineers who need it.

bazile-clyde: That makes sense. Done

bazile-clyde:
@emadum I also added headers to the `Obtaining Credentials` sections. I was starting to get lost in all the text.

emadum (approved):
LGTM

PR #865, opened 2020-10-15
Remove extraneous expectError field from poc-crud test files

divjotarora:


divjotarora:
Closing as this was intentional. The `expectedError` type has an `expectResult` field that applies to operations that can return both a result and an error (e.g. `bulkWrite`).

PR #868, opened 2020-10-19
Use $$unsetOrMatches to match insertedCount field

isabelatkinson:


alcaeus (approved):

source/unified-test-format/tests/valid-pass/poc-retryable-writes.yml
```diff
@@ -141,9 +141,7 @@ tests:
             - { _id: 4, x: 44 }
           ordered: true
         expectResult:
-          # insertMany returns a BulkWriteResult, but there is no reason to
-          # assert other fields
-          insertedCount: 2
+          insertedCount: { $$unsetOrMatches: 2 }
```
jmikola, line 144: @isabelatkinson: InsertManyResult in the [CRUD spec](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst) doesn't actually define this field at all. Should it have just been removed entirely?

PR #873, opened 2020-10-29
DRIVERS-555 Add spec/prose tests for client-side operations timeout

divjotarora:
Right now, all of the new tests are in the `client-side-operations-timeout` directory. When we convert the spec from Google Docs to RST, however, we'll make the required changes to the existing specs and move tests to other directories as needed.

Salient points for reviewers:

- The `generate-basic-tests.py` and all `.yml.template` files are used for test generation. Each `.yml.template` file defines a set of tests to run and the corresponding generated `.yml` file contains each of those tests for each operation. It might be easier to review the template files and just skim over the generated ones.
- I've added a schema-1.2.json to reflect the changes to the unified test format. All YAML files that use these new features should declare a `schemaVersion` of 1.2. Note that 1.1 was skipped because it's being taken by the versioned API project, which will almost definitely close out before this one.

TODO:

- [x] Figure out if the `generate-basic-tests.py` script needs to account for the `createIndexes` operation. The current version only tests `createIndex`.
- [x] Modify Travis to fail if the YAML templates are out of sync with the generated files.
- [x] Add new required features to the unified test format spec.
- [x] Add a new JSON schema to reflect the unified test format changes.

Once we know where these tests are actually ending up, we should also update the Makefile to run JSON schema checks against the test files.


divjotarora:
The changes to the unified test runner spec and the new schema version 1.2 JSON file have been added to this PR. @alcaeus Adding you as a reviewer for these changes.

source/client-side-operations-timeout/tests/client-side-encryption.yml
```diff
@@ -0,0 +1,79 @@
+runOn:
```
kevinAlbs, line 1: This seems fine for now. In the future we'll be moving tests to their respective directories. At that point, this can be converted to the `.yml.template` format. Right?

source/client-side-operations-timeout/tests/client-side-encryption.yml
```diff
@@ -0,0 +1,79 @@
...
+            $db: keyvault
+          command_name: listCollections
+      # Then key is fetched from the key vault.
+      - command_started_event:
+          command:
+            find: datakeys
+            filter: {"$or": [{"_id": {"$in": [ {'$binary': {'base64': 'AAAAAAAAAAAAAAAAAAAAAA==', 'subType': '04'}} ] }}, {"keyAltNames": {"$in": []}}]}
+            $db: keyvault
+            readConcern: { level: "majority" }
+          command_name: find
```
kevinAlbs, line 79: Each of the `listCollections`, `find` and `insert` commands are all under the same timeout. Is it worth adding a test that blocks each for 10ms, then sets `timeoutMS` to 25ms, to check that the running sum of the timeout is accounted for?

source/client-side-operations-timeout/tests/client-side-encryption.yml
```diff
@@ -0,0 +1,79 @@
...
+              name: "datakeys"
+            $db: keyvault
+          command_name: listCollections
+      # Then key is fetched from the key vault.
+      - command_started_event:
+          command:
+            find: datakeys
+            filter: {"$or": [{"_id": {"$in": [ {'$binary': {'base64': 'AAAAAAAAAAAAAAAAAAAAAA==', 'subType': '04'}} ] }}, {"keyAltNames": {"$in": []}}]}
+            $db: keyvault
+            readConcern: { level: "majority" }
```
kevinAlbs, line 78: [listCollections](https://docs.mongodb.com/manual/reference/command/listCollections/) does not support maxTimeMS from the looks of it. But I think this `find` should include `maxTimeMS`, right?

kevinAlbs (changes requested):
Reviewed FLE related files.

source/client-side-operations-timeout/tests/client-side-encryption.yml
divjotarora: I assume you mean the `.yml.template` format in the FLE tests directory. IIUC, that format is used when testing all operations (e.g. if we wanted to assert that timeout is applied during the FLE section of each possible operation), but we're only testing `insertOne`. Does this need to be templated?

source/client-side-operations-timeout/tests/client-side-encryption.yml
divjotarora: From talking to the server team, my understanding was that all commands support `maxTimeMS` as it's a top-level field that's parsed before the server gets to command-specific logic. I updated the tests to include it in command monitoring expectations. This also required a small update to the FLE test format to support the array-based `$$type`` syntax like the unified test format does.

source/client-side-operations-timeout/tests/client-side-encryption.yml
divjotarora: Yeah, this should've been done differently. I removed this test and added a different one that blocks `listCollections` and `find` for 10ms each and runs the `insert` with `timeoutMS=25`. In total, the FLE portion should take 30ms (2 `listCollections` and 1 `find`).

source/unified-test-format/schema-1.2.json
```diff
@@ -0,0 +1,300 @@
...
+              "items": {
+                "type": "string",
+                "enum": ["commandStartedEvent", "commandSucceededEvent", "commandFailedEvent"]
+              }
+            },
+            "ignoreCommandMonitoringEvents": {
+              "type": "array",
+              "minItems": 1,
+              "items": { "type": "string" }
+            }
```
alcaeus, line 83: In combination with the new definition above:
```suggestion
            "ignoreCommandMonitoringEvents": { "$ref": "#/definitions/ignoreCommandMonitoringEvents" }
```

source/unified-test-format/schema-1.2.json
```diff
@@ -0,0 +1,300 @@
...
+        "topologies": {
+          "type": "array",
+          "minItems": 1,
+          "items": {
+            "type": "string",
+            "enum": ["single", "replicaset", "sharded", "sharded-replicaset"]
+          }
+        }
+      }
+    },
```
alcaeus, line 55: Consider adding a definition for the `ignoreCommandMonitoringEvents` field:
```suggestion
    },
    
    "ignoreCommandMonitoringEvents": {
        "type": "array",
        "minItems": 1,
        "items": { "type": "string" }
    },    
```

source/unified-test-format/schema-1.2.json
```diff
@@ -0,0 +1,300 @@
...
+      "type": "object",
+      "additionalProperties": false,
+      "required": ["client", "events"],
+      "properties": {
+        "client": { "type": "string" },
+        "ignoreCommandMonitoringEvents": {
+          "type": "array",
+          "minItems": 1,
+          "items": { "type": "string" }
+        },
```
alcaeus, line 167: In combination with the definition above:
```suggestion
        "ignoreCommandMonitoringEvents": { "$ref": "#/definitions/ignoreCommandMonitoringEvents" },
```

source/unified-test-format/schema-1.2.json
```diff
@@ -0,0 +1,300 @@
...
+        "expectResult": {},
+        "isTimeoutError": {
+          "type": "boolean",
+          "const": true
+        },
+        "isNetworkError": {
+          "type": "boolean",
+          "const": true
+        }
```
alcaeus, line 268: The spec explicitly mentions specifying `isTimeoutError: false`, so `const` must be omitted as it only allows specifying `isTimeoutError: true`.

```suggestion
        "isTimeoutError": { "type": "boolean" },
        "isNetworkError": { "type": "boolean" }
```

Also, consider grouping with the `isClientError` property a few lines above for readability.

source/unified-test-format/unified-test-format.rst
```diff
@@ -755,6 +787,13 @@ The structure of each object is as follows:
   executing `operations`_. If the array is empty, the test runner MUST assert
   that no events were observed on the client (excluding ignored events).
 
+- ``ignoreCommandMonitoringEvents``: Optional array of one or more strings.
```
alcaeus, line 790: It may be worth expanding this paragraph with an instruction that the commands will most likely be logged but must be filtered out when comparing the list of actual events to this list of expected events. Most drivers will most likely first compare the count of both lists and fail the check if it isn't the same, so this technical detail may be helpful.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1006,7 +1045,7 @@ has been created.
 
 This operation proxies the client's ``watch`` method and supports the same
 arguments and options. Test files SHOULD NOT use the client's ``watch``
-operation directly for reasons discussed in `ChangeStream Operations`_. Test
+operation directly for reasons discussed in `Supported Entity Types`_. Test
```
alcaeus, line 1048: Consider creating an anchor and linking to the change stream entity type directly (same for the following two links).

alcaeus:
One more thing to note: you may want to change the schema version in the unified test formats's `Makefile` and also check the new tests against the schema. You can take the changes from the versioned API PR for guidance: https://github.com/mongodb/specifications/pull/877/files#diff-cb6c7b8f4cd901e87aa6f787bf2f4dbb3f90e5aab2ad6165fef180731d1b9108.

source/unified-test-format/unified-test-format.rst
divjotarora: Done.

divjotarora:
@alcaeus I haven't updated the Makefile yet because these tests will likely end up in different directories when we do the PR for the actual spec. It's not clear how that's going to work yet, though. I added a TODO to the PR description to figure out the Makefile changes before we merge this.

source/client-side-operations-timeout/tests/client-side-encryption.yml
kevinAlbs: Ah I see. Thanks for adding it.

source/client-side-operations-timeout/tests/client-side-encryption.yml
kevinAlbs: Correct, I was referring to the files in the `source/client-side-encryption/etc/test-templates` directory. Those templated files were used to keep the key vault data, JSON schemas, and ciphertexts out of the files so they could be regenerated if the keys changed (e.g. if AWS / Azure / GCP credentials needed to be regenerated).

kevinAlbs (approved):
The Client Side Encryption tests and associated readme changes LGTM.

alcaeus (approved):
Unified test format changes look good.

divjotarora:
@alcaeus Requesting another review for the `initialData` changes that we discussed offline.

source/unified-test-format/schema-1.2.json
```diff
@@ -0,0 +1,293 @@
...
+    },
+
+    "collectionData": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": ["collectionName", "databaseName", "documents"],
+      "properties": {
+        "collectionName": { "type": "string" },
+        "databaseName": { "type": "string" },
+        "collectionOptions": { "type": "object" },
```
alcaeus, line 152: The `collectionData` definition is used for both `initialData` as well as `outcome` in tests. I would suggest splitting this up into `initialData` and `outcomeData` definitions, as specifying `collectionOptions` in an `outcome` field should not be possible.

source/unified-test-format/unified-test-format.rst
```diff
@@ -298,9 +323,12 @@ The top-level fields of a test file are as follows:
 - ``initialData``: Optional array of one or more `collectionData`_ objects. Data
   that will exist in collections before each test case is executed.
 
-  Before each test and for each `collectionData`_, the test runner MUST drop the
-  collection and insert the specified documents (if any) using a "majority"
-  write concern. If no documents are specified, the test runner MUST create the
+  Before each test and for each `collectionData`_, the test runner MUST drop
+  the collection. If a ``collectionOptions`` document is present, the test
+  runner MUST execute a ``create`` command to create the collection with the
+  specified options. The test runner MUST then insert the specified documents
```
alcaeus, line 329: We may want to be explicit here that the `create` should be run with a "majority" write concern. Consider:
```suggestion
  specified options and a "majority" write concern. The test runner MUST then insert the specified documents
```

alcaeus (changes requested):
The `collectionData` definition should be split up to not allow `collectionOptions` on the `outcome` field.

source/unified-test-format/schema-1.2.json
divjotarora: Done. I split them into `initialCollectionData` and `outcomeCollectionData` types.

source/unified-test-format/unified-test-format.rst
divjotarora: I realized the information about creating collections is actually duplicated here and in the `Executing a Test` section. I removed it from here and updated the other one instead.

alcaeus (approved):
Unified test format changes LGTM

source/client-side-operations-timeout/etc/templates/deprecated-options.yml.template
```diff
@@ -0,0 +1,233 @@
+description: "operations error if timeoutMS is used with a deprecated timeout option"
```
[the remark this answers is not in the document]
divjotarora: I didn't see much value in just making this the file name. That spec section also says

> This SHOULD describe the common purpose of tests in this file and MAY refer to the filename (e.g. "updateOne-hint").

I personally find these descriptions to be more meaningful, but I'm open to changing them if you feel strongly about it.

source/client-side-operations-timeout/etc/templates/deprecated-options.yml.template
```diff
@@ -0,0 +1,233 @@
...
+
+initialData:
+  - collectionName: &collectionName coll
+    databaseName: &databaseName test
+    documents: []
+
+tests:
+  # One-off tests to assert that drivers error if both maxCommitTimeMS and timeoutMS are set.
+
+  - description: "abortTransaction errors if maxCommitTimeMS and timeoutMS are both specified"
```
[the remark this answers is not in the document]
divjotarora: Good catch. The transactions spec says:

> startTransaction SHOULD report an error if the driver can detect that transactions are not supported by the deployment.

To avoid adding too many special cases to the template, I set the min server version to 4.2 for the test file as that's the lowest version which supports transactions.

source/client-side-operations-timeout/etc/templates/deprecated-options.yml.template
```diff
@@ -0,0 +1,233 @@
...
+    operations:
+      - name: createEntities
```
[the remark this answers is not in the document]
divjotarora: It doesn't matter for these tests, but the tests that are generated by the `for` loop below need to declare entities on a per-test basis because they configure clients differently. For example, some tests create a client with `socketTimeoutMS` while others create one with `waitQueueTimeoutMS`. I modified the Unified Test Format spec to add a `createEntities` operation to do this (see the changes in `unified-test-format.rst`).

source/client-side-operations-timeout/etc/templates/deprecated-options.yml.template
```diff
@@ -0,0 +1,233 @@
+description: "operations error if timeoutMS is used with a deprecated timeout option"
+
+schemaVersion: "1.2"
+
+initialData:
+  - collectionName: &collectionName coll
+    databaseName: &databaseName test
```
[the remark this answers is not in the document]
divjotarora: Because this file doesn't use a file-level `createEntities` directive, the db/collection entities for these aren't declared until each test. I used YAML anchors here so the tests that need these entities can reference them by anchors instead of re-declaring the names everywhere.

source/client-side-operations-timeout/etc/templates/override-collection-timeoutMS.yml.template
```diff
@@ -0,0 +1,135 @@
...
+
+initialData:
+  - collectionName: &collectionName coll
+    databaseName: *databaseName
+    documents: []
+
+tests:
+  # For each collection-level operation, we execute two tests:
+  #
+  # 1. timeoutMS can be overridden to a non-zero value for a MongoCollection. Each test constructs uses the client
```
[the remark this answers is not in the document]
divjotarora: Yep the `constructs` isn't needed. Fixed.

source/client-side-operations-timeout/etc/templates/override-collection-timeoutMS.yml.template
```diff
@@ -0,0 +1,135 @@
...
+  # entity defined above to construct a collection entity with timeoutMS=1000 and configures a fail point to block the
+  # operation for 15ms so the operation succeeds.
```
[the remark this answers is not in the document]
divjotarora: If we don't set a fail point, the test would pass even if a driver had a bug that caused the collection timeoutMS to be 10 (inherited from the client) because the operation wouldn't time out. By configuring the operation to block for longer than the client's timeoutMS, we're making sure that the collection's timeoutMS was correctly overridden to infinite.

source/unified-test-format/schema-1.2.json
divjotarora: When test runners create a collection on the server, they use a test file's `initialData` directive. This directive accepts collection/database names, not entity references, so it doesn't actually know what options were used to configure the entity. Note that it can't be changed to accept entity references because the changes in this PR allow test files to create new entities at the test-level, so the collection being referenced might not even exist as an entity yet.

Also, we discussed some workarounds to try to reuse the existing `collectionOptions` stuff, but ultimately decided that it's purpose is to declare options to configure a client-side `MongoCollection` object, not a server-side collection.

source/client-side-operations-timeout/etc/templates/deprecated-options.yml.template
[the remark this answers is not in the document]
bazile-clyde: Ah, nevermind then. Yeah, this is a lot more meaningful.

source/unified-test-format/unified-test-format.rst
```diff
@@ -218,7 +218,32 @@ Test runners MUST support the following types of entities:
...
+  `client_createChangeStream`_, `database_createChangeStream`_, or
+  `collection_createChangeStream`_ operation.
+  
+  Test files SHOULD NOT use a ``watch`` operation to create a change
+  stream, as the implementation of that method may vary among drivers. For
+  example, some implementations of ``watch`` immediately execute ``aggregate``
+  and construct the server-side cursor, while others may defer ``aggregate``
+  until the change stream object is iterated.
+
+  See `Cursor Operations`_ for a list of operations.
```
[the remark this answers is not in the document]
divjotarora: I renamed `ChangeStream Operations` to a more generic `Cursor Operations` because it defines operations that pertain to both `ChangeStream` and `FindCursor` entities. Because of that, information that only applies to change streams (e.g. the guidance about not using a `watch` operation) had to be moved here.

source/unified-test-format/unified-test-format.rst
```diff
@@ -218,7 +218,32 @@ Test runners MUST support the following types of entities:
...
+- FindCursor. These entities are not defined in `createEntities`_ but are
+  instead created by using `operation.saveResultAsEntity
+  <operation_saveResultAsEntity_>`_ with a `collection_createFindCursor`_
+  operation.
+
+  Test files SHOULD only use this entity type to store tailable cursor
+  entities. Non-tailable cursors should be created and iterated via the
+  ``find`` operation instead.
+
+  See `Cursor Operations`_ for a list of operations.
```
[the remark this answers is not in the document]
divjotarora: Can you explain what you mean?

source/client-side-operations-timeout/etc/templates/deprecated-options.yml.template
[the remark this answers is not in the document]
bazile-clyde: Gotcha

bazile-clyde (approved):
Just one question but otherwise LGTM.

source/client-side-operations-timeout/etc/generate-basic-tests.py
```diff
@@ -0,0 +1,142 @@
+from collections import namedtuple
+from jinja2 import Template
+import os
+import sys
+
+# TODO: do tests distinguish between createIndex and createIndexes?
```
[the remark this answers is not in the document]
divjotarora: Seems like existing tests just use `createIndex` as the operation name, so I removed this TODO.

source/client-side-encryption/etc/test-templates/timeoutMS.yml.template
```diff
@@ -0,0 +1,65 @@
+runOn:
```
kevinAlbs, line 1: Ah nice, thanks for template-ifying this!

kevinAlbs (approved):
Client Side Encryption timeoutMS test and related changes LGTM. I did not review the rest of the changes.

source/unified-test-format/schema-1.2.json
```diff
@@ -0,0 +1,307 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
jmikola, line 2: See [DRIVERS-1485](https://jira.mongodb.org/browse/DRIVERS-1485) and https://github.com/mongodb/specifications/pull/895. A new minor version should rename the existing schema file and then incorporate any relevant changes. Doing so should result in this showing up as a rename/modified diff instead of a new file.

durran (approved):
TXN looks good.

iwysiu (approved):
transactions look good

divjotarora:
I'm going to close out this PR because it's gotten very large. I will rebase everything on the main branch, update the schema version for the unified test format spec, and make a new PR in draft mode with the final changes.

PR #875, opened 2020-11-04
FLE - maxWireVersion.json test should bypass spawning the mongocryptd

rozza:


source/client-side-encryption/tests/maxWireVersion.yml
```diff
@@ -12,6 +12,8 @@ tests:
       autoEncryptOpts:
         kmsProviders:
           aws: {} # Credentials filled in from environment.
+        extraOptions:
+          mongocryptdBypassSpawn: true # mongocryptd probably won't be on the path
```
kevinAlbs, line 16: This makes sense, if you're testing against pre-4.2, you won't have mongocryptd unless you also have a 4.2+ enterprise download.

Instead of updating the yml directly, can you update the template file: https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/etc/test-templates/maxWireVersion.yml.template#L1:L1

Then regenerate with:
```
cd source/client-side-encryption
python ./etc/generate-test.py ./etc/test-templates/maxWireVersion.yml.template ./tests
```

That should fix Travis.

kevinAlbs (changes requested):

source/client-side-encryption/tests/maxWireVersion.yml
rozza: Ah thanks - I thought Make did all the magic but thats good to know. Updated.

kevinAlbs (approved):
LGTM, thanks for fixing this!

PR #876, opened 2020-11-04
SPEC-1555 Consider connection pool health during server selection

patrickfreed:
SPEC-1555

This PR updates the Server Selection algorithm to consider connection pool health as per the design in DRIVERS-781. It also updates the CMAP spec to require liveness checking of pooled connections.

A first draft of this PR was filed on my fork while the maxConnecting changes were still in flux. See the initial rounds of discussion there: https://github.com/patrickfreed/specifications/pull/1


patrickfreed:
Due to discussion on the other draft PR, I updated the "health" determination to include the WaitQueue length of a particular pool. This will help in situations where all pools are full and have WaitQueues, and it will also help in situations where a non-full pool has developed a WaitQueue due to waiting on maxConnecting.

source/server-selection/server-selection.rst
```diff
@@ -1041,12 +1043,57 @@ The shortest average round trip time (RTT) from among suitable servers anchors
...
+1. Select two servers at random from the set of available servers in the latency
+   window. If there is only 1 server in the latency window, just return that server
+   without proceeding to the next steps.
+
+2. Calculate each server's ``operationCount``, which is the sum of the pool's
+   activeConnectionCount and the length of the pool's WaitQueue. If the
+   difference between the two server's ``operationCounts`` is greater than 5% of
+   maxPoolSize, then choose the server with the smaller ``operationCount``.
+
+3. Otherwise, choose the server with the higher ``availableConnectionCount``.
```
p-mongo, line 1070: This should also say what happens if availableConnectionCount are the same.

p-mongo:
I suggest adding rationales/descriptions for:

- Prioritizing even load distribution across servers over reusing connections
- Prioritizing even load distribution across servers once connection counts have reached steady state, even at the expense of exacerbating connection storms when the application or database is started or restarted (note: previous spec change to cmap limited how many concurrent connections can be outstanding to one node from one client to mitigate the storms)

source/server-selection/tests/README.rst
```diff
@@ -65,3 +65,43 @@ the TopologyDescription. Each YAML file contains a key for these stages of serve
...
+    for this server's pool.
+
+- ``max_pool_size``: the maximum number of connections allowed in a server's
+  connection pool.
+
+- ``expected_frequencies``: a document whose keys are the server addresses from the
+  ``in_window`` array and values are numbers in [0, 1] indicating the frequency
+  at which the server should have been selected.
+
+For each file, pass the information from `in_window` to whatever function is
```
ShaneHarvey, line 97: The test should also configure the read preference used for server selection. I suggest adding a `read_preference` field to be consistent with the existing tests.

source/server-selection/tests/README.rst
```diff
@@ -65,3 +65,43 @@ the TopologyDescription. Each YAML file contains a key for these stages of serve
...
+used to select a server from within the latency window 1000 times, counting how
+many times each server is selected.  Once 1000 selectoins have been made, verify
```
ShaneHarvey, line 99: "1000 selectoins" -> "1000 selections"

source/server-selection/tests/README.rst
```diff
@@ -65,3 +65,43 @@ the TopologyDescription. Each YAML file contains a key for these stages of serve
...
+that each server was selected at a frequency within 0.05 of the frequency
```
ShaneHarvey, line 100: 5% tolerance works okay for me in python so far but I'm a little worried about false positives here. For example I've seen the delta reach 4%. So I suspect we might need to raise the tolerance to 6% or 7% (or increase the # of iterations) in the future. Can we increase the # of iterations to 2000 to avoid the need to adjust it later?

ShaneHarvey (changes requested):
I've implemented the tests add they are passing as expected. A few comments on the test format.

source/server-selection/tests/README.rst
patrickfreed: I think we discussed this in the meeting, but since this test is verifying what happens after the set of available servers is determined and the topologies will always be sharded, a read preference shouldn't be necessary here.

source/server-selection/tests/README.rst
```diff
@@ -65,3 +65,37 @@ the TopologyDescription. Each YAML file contains a key for these stages of serve
 Drivers implementing server selection MUST test that their implementation
 correctly returns the set of servers in ``in_latency_window``. Drivers SHOULD also test
 against ``suitable_servers`` if possible.
+
+Selection Within Latency Window Tests
```
patrickfreed, line 69: Now that the "algorithm" is much simpler, is it still worth it to have all drivers implement these unit tests?

source/server-selection/server-selection.rst
patrickfreed: We decided to omit the consideration of availableConnectionCounts altogether to simplify the algorithm and remove the need for server selection to have a dependency on CMAP.

source/server-selection/server-selection.rst
```diff
@@ -1041,12 +1043,57 @@ The shortest average round trip time (RTT) from among suitable servers anchors
...
+
+        if len(in_window) == 1:
+           return in_window[0]
+
+        server1, server2 = random two entries from in_window
+
+        server1_op_count = server1.pool.active_connection_count + len(server1.pool.wait_queue)
+        server2_op_count = server2.pool.active_connection_count + len(server2.pool.wait_queue)
+        if abs(server1_op_count - server2_op_count) > max(0.05 * max_pool_size, 1):
+            if server1_op_count < server_2_op_count:
```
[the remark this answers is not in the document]
patrickfreed: This psuedocode has been removed.

source/server-selection/tests/README.rst
patrickfreed: fixed

source/server-selection/server-selection.rst
```diff
@@ -1431,6 +1475,39 @@ servers to come and go.  Making newly available servers either first or last
 could lead to unbalanced work.  Random selection has a better fairness
 guarantee and keeps the design simpler.
 
+operationCount-based selection within the latency window (multi-threaded or async)
```
patrickfreed, line 1478: @p-mongo This section should contain the missing rationale you requested. Let me know if it is still missing anything. Also, I know you mentioned that the dimensionality of our algorithm was perhaps too high--this was similar feedback to what others on the review crew had, so we reduced the various different inputs down to just a single `operationCount` which should simplify this greatly.

source/server-selection/server-selection.rst
```diff
@@ -798,8 +798,24 @@ a pool; or (b) governed by a global or client-wide limit on number of
...
+
+Multi-threaded or async drivers MUST keep track of the number of operations that
+a given server is currently executing (the server's ``operationCount``). This
+value MUST be incremented once a server is selected for an operation and MUST be
+decremented once that operation has completed, regardless of its outcome. This
+value SHOULD be stored on the ``Server`` type that also owns the connection pool
+for the server, if there exists such a type in the driver's
+implementation. Incrementing or decrementing a server's ``operationCount`` MUST
+NOT wake up any threads that are waiting for a topology update as part of server
+selection. See `operationCount-based selection within the latency window
```
patrickfreed, line 812: When implementing this in the Rust driver, I realized that putting `operationCount` on the ServerDescription would turn server selection into a mutable operation (requiring an exclusive lock instead of a read-only lock), which seemed to not be the best option in hindsight. So in the spec, I opted to just include the soft requirement that drivers SHOULD put it on the `Server` type if they have one, and if they don't, just make sure not to wake up threads waiting in server selection when they modify the value.

patrickfreed (commented):
This PR has been rewritten to consolidate the various counts we retrieved from the pool to a single `operationCount` that is incremented after a server is selected and decremented once the operation it was selected for is completed. Additionally, the consideration of availableConnectionCount was dropped, so no changes to or interaction with CMAP is required for this work. This should make the changes here much simpler and easier to understand.

source/server-selection/tests/README.rst
patrickfreed: fixed

patrickfreed:
Also, our discussions made me realize these changes have no effect on single threaded drivers, so I updated the spec to require this only for multi-threaded or async drivers.

source/server-selection/server-selection.rst
ShaneHarvey: It might be nice to mention explicitly that we use operationCount as an approximation of load and that this algorithm is meant to distribute load more fairly than random.

source/server-selection/server-selection.rst
```diff
@@ -1431,6 +1475,39 @@ servers to come and go.  Making newly available servers either first or last
 could lead to unbalanced work.  Random selection has a better fairness
 guarantee and keeps the design simpler.
 
+operationCount-based selection within the latency window (multi-threaded or async)
+----------------------------------------------------------------------------------
+
+When a node begins slows down, any pooled connections opened against it which
```
ShaneHarvey, line 1481: This is a good explanation, this algorithm should handle the case where a node's performance degrades better than random selection. In fact, this algorithm handles even more cases than described here.  For example,  "node begins slows down" could be generalized to "operation execution slows down". "Operation execution" includes factors other than the remote server's performance, like the latency of creating connections and the artificial latency incurred by maxConnecting.

ShaneHarvey (commented):
> Also, our discussions made me realize these changes have no effect on single threaded drivers, so I updated the spec to require this only for multi-threaded or async drivers.

Interesting, this is in line with our scope which says:
> [Non-goals]: 
> Change the behavior of single-threaded drivers
> Single-threaded driver instances do not maintain connection pools and thus are unlikely to cause connection storms on their own

A quick note that if single threaded drivers begin supporting any feature that require connection pinning (like OP_MSG exhaust cursors), then they will actually need to implement operationCount.

source/server-selection/server-selection.rst
patrickfreed: done

source/server-selection/server-selection.rst
patrickfreed: Updated it to be a bit more general and use some of the examples you provided.

source/server-selection/tests/README.rst
patrickfreed: We're limited a bit by the topology. We could include some of those experiments I did as prose tests--for example, describe a test against a two mongos sharded topology where one of the mongoses has a failpoint that makes every operation take 500ms, then do a ton of concurrent stuff and then assert that the non-failpoint node got picked a lot more. While this does verify the operationCount behavior, I'm not sure whether this is preferable to having a bunch of unit spec tests, though.

source/server-selection/tests/README.rst
patrickfreed: Ah I see, yeah that makes sense. In my implementation I skipped right to the in window portion, so I didn't need a TopologyDescription or a read preference. I imagine most drivers will probably implement your way though. 

Rather than including a read preference in each spec test, I just updated the runner's description to say use a default or "primary" read preference since it shouldn't matter for sharded topologies.

p-mongo (approved):
The spec changes look great, thank you.

I have not attempted to implement the tests.

source/server-selection/server-selection-tests.rst
```diff
@@ -257,6 +258,26 @@ return a set of three suitable servers within the latency window::
...
 and ``secondary_2`` at random.
 
+operationCount-based Selection Within Latency Window (multi-threaded or async drivers)
+======================================================================================
+
+The Server Selection spec mandates that multi-threaded or async
+drivers select a server from within the latency window according to
+their operationCounts. There are YAML tests verifying that drivers
+implement this selection correctly. Multi-threaded or async drivers
+implementing the spec MUST use them test their implementations.
```
mbroadst, line 268:
```suggestion
implementing the spec MUST use them to test their implementations.
```

source/server-selection/server-selection.rst
mbroadst: SGTM

mbroadst (approved):
Spec changes look great, I'll LGTM but we should hold off on merge until Shane validates the tests

source/server-selection/server-selection.rst
```diff
@@ -1211,7 +1247,15 @@ Pseudocode for `multi-threaded or asynchronous server selection`_::
 
             if servers is not empty:
                 in_window = servers within the latency window
-                selected = random entry from in_window
+                if len(in_window) == 1:
+                    selected = in_window[0]
+                else:
+                    server1, server2 = random two entries from in_window
+                    if server1.operation_count >= server2.operation_count:
```
ShaneHarvey, line 1254: This should be `<=` since.

source/server-selection/tests/in_window/three-choices.yml
```diff
@@ -0,0 +1,24 @@
...
+    - address: a:27017
+      operation_count: 3
+    - address: b:27017
+      operation_count: 6
+    - address: c:27017
+      operation_count: 20
+expected_frequencies:
+    a:27017: 0.0
+    b:27017: 0.33
+    c:27017: 0.66
```
ShaneHarvey, line 24: I think this should be:
```

File: source/server-selection/tests/README.rst (hunk 5 above)
Comment (ShaneHarvey) [reply source/server-selection/tests/README.rst:69]: Please open a jira ticket so we can further discuss adding that prose test. I'm in favor of it.

File: source/server-selection/server-selection.rst (hunk 8 above)
Comment (ShaneHarvey) [reply source/server-selection/server-selection.rst:812]: This sounded good to me earlier but when I implemented this in Python I realized it's much simpler to track ``operationCount`` on our Pool class rather than at the Server level:
- The pool already has checkout/checkin methods which can increment/decrement ``operationCount``.
- The pool already has a lock which can be reused for mutating ``operationCount``.
- In python, servers can be selected without actually used to run an operation (`MongoClient.primary` for example).

Putting the operationCount in the pool also transparently deals with pymongo's legacy exhaust cursor feature where a connection remains checked out until the cursor is exhausted.

Please update the guidance (and the pseudocode) to say that operationCount can live on the Pool or Server. Or perhaps we should always make it live in the pool and update the CMAP spec too?

**Review by ShaneHarvey — CHANGES_REQUESTED** [review]

File: source/server-selection/tests/in_window/three-choices.yml (hunk 12 above)
Comment (patrickfreed) [reply source/server-selection/tests/in_window/three-choices.yml]: fixed

File: source/server-selection/tests/in_window/maxConnecting-waitQueue.yml [hunk 13]
```diff
@@ -0,0 +1,24 @@
+description: waitQueue length caused by maxConnecting is considered
```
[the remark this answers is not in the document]
patrickfreed: updated the tests to be more generic

source/server-selection/tests/in_window/many-choices.yml
```diff
@@ -0,0 +1,60 @@
+description: Selections from many choices occur at correct frequencies
```
patrickfreed, line 1: @ShaneHarvey 

This test includes some really small frequencies, so I upped the iteration count to 10,000 and decreased the threshold to 2%. Let me know if this works for your runner okay and doesn't take too long to run.

source/server-selection/tests/README.rst
ShaneHarvey: Yeah let's just add it now. I doubt the unified test format is expressive enough for a test like this so it'll need to be a prose test like you said. You can use the blockConnection option for failCommand like this:
```
db.adminCommand({
    mode: {times: 100000},  # or "alwaysOn"
    data: {
      failCommands: ["find"],
      blockConnection: true,
      blockTimeMS: 500,
      appName: "loadBalancingTest",
    },
});
```

source/server-selection/tests/README.rst
ShaneHarvey: It still feels a little odd that we have no explicit tests for replica sets. Perhaps the prose test can cover it?

source/server-selection/tests/in_window/many-choices.yml
ShaneHarvey: Note that even with 10000 iterations I got this false positive [in this evergreen patch](https://evergreen.mongodb.com/version/5fadb207c9ec440ad07f3cb1):
```
 [2020/11/12 22:16:08.832] FAIL [0.230s]: test_in_window_one_least_two_tied_Least_operations_gets_most_selections,_two_tied_share_the_rest (test_server_selection_in_window.TestAllScenarios)
 [2020/11/12 22:16:08.832] ----------------------------------------------------------------------
 [2020/11/12 22:16:08.832] Traceback (most recent call last):
 [2020/11/12 22:16:08.832]   File "/data/mci/491f6078eca3c0414292a7732afbf55b/src/test/__init__.py", line 444, in wrap
 [2020/11/12 22:16:08.832]     return f(*args, **kwargs)
 [2020/11/12 22:16:08.832]   File "/data/mci/491f6078eca3c0414292a7732afbf55b/src/test/test_server_selection_in_window.py", line 92, in run_scenario
 [2020/11/12 22:16:08.832]     self.run_scenario(scenario_def)
 [2020/11/12 22:16:08.832]   File "/data/mci/491f6078eca3c0414292a7732afbf55b/src/test/test_server_selection_in_window.py", line 87, in run_scenario
 [2020/11/12 22:16:08.832]     self.assertAlmostEqual(actual_freq, freq, delta=0.02)
 [2020/11/12 22:16:08.832] AssertionError: 0.6805 != 0.66 within 0.02 delta (0.020499999999999963 difference)
```

source/server-selection/tests/in_window/one-least-two-tied.json
```diff
@@ -0,0 +1,42 @@
+{
+  "description": "Least operations gets most selections, two tied share the rest",
```
ShaneHarvey, line 2: Note that even with 10000 iterations I got this false positive [in this evergreen patch](https://evergreen.mongodb.com/version/5fadb207c9ec440ad07f3cb1):
```
 [2020/11/12 22:16:08.832] FAIL [0.230s]: test_in_window_one_least_two_tied_Least_operations_gets_most_selections,_two_tied_share_the_rest (test_server_selection_in_window.TestAllScenarios)
 [2020/11/12 22:16:08.832] ----------------------------------------------------------------------
 [2020/11/12 22:16:08.832] Traceback (most recent call last):
 [2020/11/12 22:16:08.832]   File "/data/mci/491f6078eca3c0414292a7732afbf55b/src/test/__init__.py", line 444, in wrap
 [2020/11/12 22:16:08.832]     return f(*args, **kwargs)
 [2020/11/12 22:16:08.832]   File "/data/mci/491f6078eca3c0414292a7732afbf55b/src/test/test_server_selection_in_window.py", line 92, in run_scenario
 [2020/11/12 22:16:08.832]     self.run_scenario(scenario_def)
 [2020/11/12 22:16:08.832]   File "/data/mci/491f6078eca3c0414292a7732afbf55b/src/test/test_server_selection_in_window.py", line 87, in run_scenario
 [2020/11/12 22:16:08.832]     self.assertAlmostEqual(actual_freq, freq, delta=0.02)
 [2020/11/12 22:16:08.832] AssertionError: 0.6805 != 0.66 within 0.02 delta (0.020499999999999963 difference)
```

source/server-selection/tests/README.rst
patrickfreed: done. When you run it, can you let me know what you get for % of operations routed to the slow node? Curious to see if theres any differences across drivers.

source/server-selection/tests/in_window/two-choices.yml
```diff
@@ -0,0 +1,21 @@
...
+  - address: a:27017
+    avg_rtt_ms: 35
+    type: Mongos
+  - address: b:27017
+    avg_rtt_ms: 35
+    type: Mongos
+  - address: c:27017
+    avg_rtt_ms: 35
+    type: Mongos
+in_window:
```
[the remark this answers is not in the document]
ShaneHarvey: Any thoughts on the "mock_server_state" proposal?

source/server-selection/tests/in_window/many-choices.yml
ShaneHarvey: So we can still implement this proposal, let's just increase `outcome.tolerance` for this test and/or reduce the number of servers in the topology.

source/server-selection/tests/README.rst
ShaneHarvey: Implemented. I ran it a few times and see:
```
{('localhost', 27017): 11, ('localhost', 27018): 89}
{('localhost', 27017): 12, ('localhost', 27018): 88}
{('localhost', 27017): 13, ('localhost', 27018): 87}
{('localhost', 27017): 15, ('localhost', 27018): 85}
{('localhost', 27017): 17, ('localhost', 27018): 83}
```

Exciting to see this feature in action!

source/server-selection/tests/in_window/many-choices.yml
patrickfreed: Increased tolerance to 0.03, which seems to work fine for me. Let me know if that has any issues for your runner.

Note: I left `iterations` outside of `outcome` since it had more to do with running the test then inspecting the outcome. `tolerance` is under there though.

source/server-selection/tests/in_window/one-least-two-tied.json
patrickfreed: done

source/server-selection/tests/in_window/two-choices.yml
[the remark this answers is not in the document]
patrickfreed: updated, though the tests don't currently have any servers that aren't in the latency windows.

source/server-selection/tests/README.rst
patrickfreed: I get around 12-15% too, so it seems like our implementations are consistent, nice!

source/server-selection/server-selection-tests.rst
```diff
@@ -257,6 +258,59 @@ return a set of three suitable servers within the latency window::
...
+4. Start 10 concurrent threads / tasks that each run 10 `findOne` operations
+   using that client.
+
+5. Assert that fewer than 25% of the CommandStartedEvents occurred on the mongos
+   that the failpoint was enabled on.
+
+6. Disable the failpoint.
+
+7. Repeat this test without any failpoints and assert that each mongos was
+   selected roughly 50% of the time.
```
ShaneHarvey, line 311: Can you add `roughly 50% of the time within +/-10% tolerance.`? I initially implemented this with 5% but that failed with: `AssertionError: 0.55 != 0.5 within 0.05 delta (0.050000000000000044 difference)`

ShaneHarvey (approved):
LGTM with a minor suggestion!

source/server-selection/server-selection-tests.rst
patrickfreed: done

source/server-selection/tests/README.rst
```diff
@@ -65,3 +65,45 @@ the TopologyDescription. Each YAML file contains a key for these stages of serve
...
+  - ``address``: a unique address identifying this server
+
+  - ``operation_count``: the ``operationCount`` for this server
+
+- ``iterations``: the number of selections that should be run as part of this
+   test
+
+- ``outcome``: an object describing the expected outcome of the selections
+
+  - ``tolerance``: the maximum difference beterrn an observed frequency and an
```
mbroadst, line 93:
```suggestion
  - ``tolerance``: the maximum difference between an observed frequency and an
```
I think this is what you meant?

patrickfreed: hah yeah, not sure what happened there.

patrickfreed:
@mbroadst @ShaneHarvey 
I updated the part about where the `operationCount` is stored, let me know what you all think.

ShaneHarvey (approved):

mbroadst:
@patrickfreed looks good!

PR #877, opened 2020-11-06
DRIVERS-996: Versioned API

alcaeus:
DRIVERS-996

This PR contains the versioned API specification, as well as the changes necessary to test versioned API with the unified test runner.

### DIFF
--- a/source/unified-test-format/unified-test-format.rst
+++ b/source/unified-test-format/unified-test-format.rst
@@ -3,13 +3,13 @@ Unified Test Format
 ===================
 
 :Spec Title: Unified Test Format
-:Spec Version: 1.0.0
+:Spec Version: 1.1.0
 :Author: Jeremy Mikola
 :Advisors: Prashant Mital, Isabel Atkinson, Thomas Reggi
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: N/A
-:Last Modified: 2020-10-16
+:Last Modified: 2020-11-06
 
 .. contents::
 
@@ -308,6 +308,11 @@ The top-level fields of a test file are as follows:
 - ``tests``: Required array of one or more `test`_ objects. List of test cases
   to be executed independently of each other.
 
+.. _yamlAnchors
+
+- ``_yamlAnchors``: Optional object containing arbitrary data. This is only used
+  to define anchors within the YAML files and MUST NOT be used by test runners.
+
 
 runOnRequirement
 ~~~~~~~~~~~~~~~~
@@ -343,6 +348,11 @@ The structure of this object is as follows:
   "sharded" topology, test runners MUST accept any type of sharded cluster (i.e.
   "sharded" implies "sharded-replicaset", but not vice versa).
 
+- ``serverParameters``: Optional object of server parameters to check against.
+  To check server parameters, drivers send a ``{ getParameter: '*' }`` command
+  to the server using the internal MongoClient. The result SHOULD be cached to
+  avoid repeated calls. The result is then compared to the expected server
+  parameters as described in the `Evaluating Matches`_ section.
 
 entity
 ~~~~~~
@@ -439,6 +449,14 @@ The structure of this object is as follows:
 
     Test files SHOULD NOT use this option unless one or more command monitoring
     events are specified in `observeEvents <entity_client_observeEvents_>`_.
+
+  - ``serverApi``: Optional object to declare an API version on the client
+    entity. A ``version`` string is required, and test runners MUST fail if the
+    given version string is not supported by the driver. The ``strict`` and
+    ``deprecationErrors`` members take an optional boolean that is passed to the
+    ``ServerApi`` object. See the
+    `Versioned API <../versioned-api/versioned-api.rst>`__ spec for more details
+    on these fields.
 
 .. _entity_database:
 
@@ -2644,3 +2662,7 @@ spec changes developed in parallel or during the same release cycle.
 
 Change Log
 ==========
+
+:2020-11-06: Added ``serverApi`` option for client entities, ``_yamlAnchors``
+             property to define values for later use in YAML tests, and
+             ``serverParameters`` property for ``runOnRequirements``.
--- a/source/unified-test-format/unified-test-format.rst
+++ b/source/unified-test-format/unified-test-format.rst
@@ -349,10 +349,17 @@ The structure of this object is as follows:
   "sharded" implies "sharded-replicaset", but not vice versa).
 
 - ``serverParameters``: Optional object of server parameters to check against.
-  To check server parameters, drivers send a ``{ getParameter: '*' }`` command
-  to the server using the internal MongoClient. The result SHOULD be cached to
-  avoid repeated calls. The result is then compared to the expected server
-  parameters as described in the `Evaluating Matches`_ section.
+  To check server parameters, drivers send a
+  ``{ getParameter: 1, <parameter>: 1 }`` command to the server using the
+  internal MongoClient. Drivers MAY also choose to send a
+  ``{ getParameter: '*' }`` command and fetch all parameters at once. The result
+  SHOULD be cached to avoid repeated calls to fetch the same parameter. Test
+  runners MUST apply the rules specified in `Flexible Numeric Comparisons`_ when
+  comparing values. If a server does not support a parameter, test runners MUST
+  treat the comparison as not equal and skip the test. This includes errors that
+  occur when fetching a single parameter using ``getParameter``.
+
+Test runners MUST evaluate these conditions in the order specified above.
 
 entity
 ~~~~~~
--- a/source/versioned-api/tests/README.rst
+++ b/source/versioned-api/tests/README.rst
@@ -22,13 +22,15 @@ Drivers MUST run their test suite against a cluster with the
 ``requireApiVersion`` parameter enabled and also requires authentication.
 
 To run this test, proceed as follows:
-* Start a standalone mongod instance
-** This can be done using ``mongo-orchestration`` or manually
-* Connect to the standalone instance and run the following command on the
+- Start a standalone mongod instance
+
+- Connect to the standalone instance and run the following command on the
   ``admin`` database: ``{ setParameter: 1, requireApiVersion: true }``
-* Declare an API version for the test run through the ``MONGODB_API_VERSION``
+
+- Declare an API version for the test run through the ``MONGODB_API_VERSION``
   environment variable.
-* If the environment variable is set, all clients created in tests MUST declare
+
+- If the environment variable is set, all clients created in tests MUST declare
   the ``ServerApiVersion`` specified.
 
 No other topologies must be tested until ``mongo-orchestration`` can handle
--- a/source/versioned-api/tests/README.rst
+++ b/source/versioned-api/tests/README.rst
@@ -14,3 +14,12 @@ implemented in the `Unified Test Format <../../unified-test-format/unified-test-
 and require schema version 1.1. Note that to run these tests, the server must be
 started with both ``enableTestCommands`` and ``acceptAPIVersion2`` parameters
 set to true.
+
+Testing with required API version
+=================================
+
+Drivers MUST test against a server with the ``requireApiVersion`` parameter
+enabled that also requires authentication. Since API versioning options can't be
+specified using the connection string, drivers MUST rely on environment
+variables to receive the API version to use in tests. The ``apiStrict`` and
+``apiDeprecationErrors`` options are not required for this test.
--- a/source/versioned-api/tests/README.rst
+++ b/source/versioned-api/tests/README.rst
@@ -18,8 +18,18 @@ set to true.
 Testing with required API version
 =================================
 
-Drivers MUST test against a server with the ``requireApiVersion`` parameter
-enabled that also requires authentication. Since API versioning options can't be
-specified using the connection string, drivers MUST rely on environment
-variables to receive the API version to use in tests. The ``apiStrict`` and
-``apiDeprecationErrors`` options are not required for this test.
+Drivers MUST run their test suite against a cluster with the
+``requireApiVersion`` parameter enabled and also requires authentication.
+
+To run this test, proceed as follows:
+* Start a standalone mongod instance
+** This can be done using ``mongo-orchestration`` or manually
+* Connect to the standalone instance and run the following command on the
+  ``admin`` database: ``{ setParameter: 1, requireApiVersion: true }``
+* Declare an API version for the test run through the ``MONGODB_API_VERSION``
+  environment variable.
+* If the environment variable is set, all clients created in tests MUST declare
+  the ``ServerApiVersion`` specified.
+
+No other topologies must be tested until ``mongo-orchestration`` can handle
+servers with ``requireApiVersion`` enabled.
--- a/source/versioned-api/tests/README.rst
+++ b/source/versioned-api/tests/README.rst
@@ -1,0 +1,16 @@
+===================
+Versioned API Tests
+===================
+
+.. contents::
+
+----
+
+Notes
+=====
+
+This directory contains tests for the Versioned API specification. They are
+implemented in the `Unified Test Format <../../unified-test-format/unified-test-format.rst>`__,
+and require schema version 1.1. Note that to run these tests, the server must be
+started with both ``enableTestCommands`` and ``acceptAPIVersion2`` parameters
+set to true.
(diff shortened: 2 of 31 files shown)


source/versioned-api/versioned-api.rst
```diff
@@ -0,0 +1,273 @@
...
+participate in executing operations. Therefore, when determining application
+compatibility with MongoDB, both the driver and the server behavior must be
+taken into account.
+
+An application can specify the server API version when creating MongoClient.
+When this is done:
+
+- The client sends the specified API version to the server, causing the server
+  to behave in a manner compatible with that API version.
+- The driver will behave in a manner compatible with the driver's behavior as of
```
patrickfreed, line 55: I'm finding this line is a bit hard to parse, could it be reworded to be a bit more simple / straightforward?

source/versioned-api/versioned-api.rst
```diff
@@ -0,0 +1,273 @@
...
+
+Drivers SHOULD group the ``serverApi`` option with other similar client options
+like ``autoEncryptionOpts``.
+
+
+ServerApiVersion enumeration
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This specification and subsequent specifications will define the known API
+versions. Drivers SHOULD define an enumeration containing the known API
```
patrickfreed, line 99: One issue with using an enum for this is that a lot of languages allow exhaustive `switch` statements over the various cases, meaning adding a new case constitutes a breaking change (Rust doesn't have this problem, by the way). If we intend to add support for new API versions in non-major releases, then we might want to include the addition requirement here that "drivers MUST be able to introduce new versions without breaking source compatibility" or something like that.

source/versioned-api/versioned-api.rst
```diff
@@ -0,0 +1,273 @@
...
+**ServerApi field**   **Command option**
+``version``           ``apiVersion``
+``strict``            ``apiStrict``
+``deprecationErrors`` ``apiDeprecationErrors``
+===================== ========================
+
+If an API version was declared, drivers MUST add the ``apiVersion`` option to
+every command that is sent to a server. Drivers MUST add the ``apiStrict`` and
+``apiDeprecationErrors`` options if they were specified by the user, even when
+the specified value is equal to the server default. Drivers SHOULD NOT add these
```
patrickfreed, line 158: can we make this "SHOULD NOT" a "MUST NOT"? Or is there a circumstance where it might be necessary to send a value the user didn't specify?

source/versioned-api/versioned-api.rst
```diff
@@ -0,0 +1,273 @@
...
+   }
+
+   class ServerApi {
+       version: string|ServerApiVersion;
+       strict: Optional<Boolean>; // Default false
+       deprecationErrors: Optional<Boolean>; // Default false
+   }
+
+Drivers SHOULD group the ``serverApi`` option with other similar client options
+like ``autoEncryptionOpts``.
```
patrickfreed, line 92: can we add an explicit "Drivers MUST NOT allow specification of any versioned API options via the connection string" (or something like that) here? I think the first time this requirement shows up in this document is in the rationale, so we should probably call it out more directly here.

source/versioned-api/versioned-api.rst
```diff
@@ -0,0 +1,273 @@
...
+Hardcode supported versions in drivers
+--------------------------------------
+
+Since a new API version might require driver changes (e.g. to account for
+removed commands), we don't yet know what changes drivers must make for a future
+version. Until we do, we must prevent users from choosing any unknown API
+version.
+
+
+Backward Compatibility
```
patrickfreed, line 244: Should we mention the related policy changes in the spec at all? I imagine in the future we'll add spec language for an "experimental" feature, so we'll need to have that term defined in the specs somewhere.

source/versioned-api/versioned-api.rst
```diff
@@ -0,0 +1,273 @@
...
+  version.
+
+Presently there is no specification for how a driver must behave when a
+particular server API version is requested, or what driver operations are
+subject to API compatibility guarantees. Such requirements may be stipulated in
+subsequent specifications.
+
+This specification requires MongoClient to validate that it supports the
+specified server API version, if any, but does not define what such support
+means.
```
patrickfreed, line 66: It might be helpful for new readers of this spec if this was more prescriptive about support for version 1. Maybe something like "Drivers that support MongoDB 4.4 automatically support API version 1, but this specification does not define what such support means for later API versions."

source/versioned-api/versioned-api.rst
alcaeus: What about this?
> The driver will behave in a manner compatible with the driver's original behavior at the time the specified API version was released.

source/versioned-api/versioned-api.rst
alcaeus: I used "SHOULD NOT" to allow drivers to use a `bool` for `strict` and `deprecationErrors` without tracking whether the user supplied it. If I change this to "MUST NOT", I'd also remove the default values for the above options from the `ServerApi` class to ensure that they are empty by default, which indicates that the driver must not add them.

source/versioned-api/versioned-api.rst
alcaeus: Yep, seems sensible to explicitly mention this here 👍

source/versioned-api/versioned-api.rst
alcaeus: Since the policy is not public, I wanted to avoid mentioning it here. I can add a note about it for all driver engineers that will be reading this.

source/versioned-api/versioned-api.rst
alcaeus: Technically you are correct - there are a few specialties though. With `count` being removed from API Version 1 before the release, we'll have to update the CRUD specification to change how `estimatedDocumentCount` works. Similarly, `distinct` is not in API Version 1, so the `distinct` helper would also need a change.

With these two issues open, I wouldn't yet add this chapter, but deal with this once we get to the DRIVERS tickets for the two cases mentioned above.

source/versioned-api/versioned-api.rst
alcaeus: Technically, drivers don't have to add compatibility for a new API Version in the current major release series, but can do so in a new major release if that's required. I've asked the leads for input on how their languages treat this to make an informed decision and will revisit this once I have some answers.

alcaeus: After discussion we've come to the conclusion that it makes sense to let drivers handle this individually without prescribing a one-size-fits-all solution. For example, a new enum value could produce a warning in C/C++ (via `-Wswitch-enum` in gcc), which in turn can cause compiler errors if warnings are treated as errors. However, since it is the user's choice to treat this strictly it is up to them to take care of this and such a change would be allowed in a minor release. Swift is a different story, as users may want to work with the enumeration in a `switch` statement, so adding a new version could be a BC break.

With this in mind, I'd suggest adding the following wording:
> Drivers MUST ensure that adding new API versions to this enumeration does not result in backward compatibility breaks in non-major releases. This can be the case in languages that allow exhaustive ``switch`` statements (e.g. Swift). In these cases, drivers should either refrain from using enumerations or only introduce new API versions in major releases.

This ensures that readers of the spec are aware of the potential issue, but can decide how they want to deal with it: either by not using enumerations (as is done in the Swift driver in some instances), by choosing to not treat it as a BC breaking change (as the C/C++ driver might want to do), or by avoiding the issue altogether and releasing a new major version.

source/versioned-api/tests/README.rst
```diff
@@ -0,0 +1,16 @@
+===================
+Versioned API Tests
+===================
+
+.. contents::
+
+----
+
+Notes
```
jyemin, line 9: We also must require that drivers test against a server running with

    --setParameter requireApiVersion=true

with auth enabled.  Otherwise, we won't be testing that drivers will work in this scenario.

source/versioned-api/versioned-api.rst
jyemin: I'm fine with the decision to provide flexibility, but I don't think we should mention the option of a major release just for a new API version, which likely otherwise wouldn't require any other driver changes except the addition of the enumeration.  Major releases are painful for users and require a lot of planning, and that's precisely what we are trying to avoid.

jyemin: I would just remove the last sentence.

source/versioned-api/versioned-api.rst
```diff
@@ -0,0 +1,277 @@
...
+
+
+ServerApi class
+~~~~~~~~~~~~~~~
+
+The ``ServerApi`` class stores an API version, along with flags that decide
+whether or not unknown or deprecated commands in the specified API version
+trigger a server-side error. A ``version`` MUST be specified when declaring an
+API version, while the ``strict`` and ``deprecationErrors`` options are both
+optional and default to ``false``. The ``ServerApi`` class is considered
```
jyemin, line 120: > are both optional and default to ``false``

If it's optional then it defaults to `unset` or `not present`.  And it contradicts what is said below:

> Drivers SHOULD NOT add these options if the user did not specify a value.

jyemin: Do you mean to say:

> are either optional, or non-optional and default to ``false``

?

source/versioned-api/versioned-api.rst
```diff
@@ -0,0 +1,277 @@
...
+every command that is sent to a server. Drivers MUST add the ``apiStrict`` and
+``apiDeprecationErrors`` options if they were specified by the user, even when
+the specified value is equal to the server default. Drivers SHOULD NOT add these
+options if the user did not specify a value.
+
+
+Cursors
+~~~~~~~
+
+The ``getMore`` command does not accept API parameters; cursors inherit their
```
jyemin, line 167: Consider an explicit note about `killCursors` accepting API parameters.  It's weird that one does and the other doesn't, so I think it's worth calling out.

jyemin (changes requested):

source/versioned-api/versioned-api.rst
patrickfreed: Hm, I think it's still a bit vague, e.g. what does "compatible" or "original behavior" mean in this case? This requirement  _has_ to be vague though because there isn't anything special for drivers to do yet, as is pointed out below. I think the issue here is that the only requirement we can feasibly impose is that the driver is compatible with a _server_ using that API version, not with _other versions of the driver_ using that API version. For instance, version 5.0 of a driver configured with `serverApiVersion: 1` does not necessarily need to exhibit the same behavior of version 6.0 of a driver configured with `serverApiVersion: 1` (e.g if 6.0 removed the `distinct` helper or something), though both versions do need to be able to _function_ with a server configured with `serverApiVersion: 1`.

So what about something as simple as the following:

>The driver will behave in a manner compatible with a server configured with that API version, regardless of the server's actual release version.

It's pretty weak, but I think it's the only meaningful requirement we can impose for now until more requirements emerge.

source/versioned-api/versioned-api.rst
patrickfreed: Yeah, I agree. Updating the default version API should certainly require a major release (imo), but simply adding support  for a new one shouldn't require one. The rest of the wording lgtm.

source/versioned-api/versioned-api.rst
patrickfreed: I think unless we wish to deviate from the server's default values for these, we should just not send them if the user didn't specify them. This would be consistent with how we handle other options that we forward along to the server.

source/versioned-api/versioned-api.rst
patrickfreed: Yeah true. I think whenever we do add the first "experimental" API to the specifications we can just tackle this issue then.

source/versioned-api/versioned-api.rst
p-mongo: I support Patrick's suggested wording.

source/versioned-api/versioned-api.rst
p-mongo: On line 160 what does "these" refer to? strict & deprecation errors or all 3 options?

p-mongo: I am not following the reason for should vs must but regarding this:

>  I'd also remove the default values for the above options from the ServerApi class to ensure that they are empty by default, which indicates that the driver must not add them.

We already require drivers not sending default values for `multi` and `upsert` in crud. Therefore I expect it should be possible for drivers to not send default values here.

p-mongo (commented):
I read the PR and I support the open comment threads. LGTM with the changes in comments applied.

source/versioned-api/versioned-api.rst
alcaeus: Sounds good, I applied your suggested wording.

source/versioned-api/versioned-api.rst
alcaeus: Removed the last sentence as per your suggestions.

source/versioned-api/versioned-api.rst
alcaeus: Changed "SHOULD NOT" to "MUST NOT" and reworded the last sentence for clarity:
> Drivers MUST NOT add any API versioning options if the user did not specify them.

source/unified-test-format/unified-test-format.rst
```diff
@@ -343,6 +348,11 @@ The structure of this object is as follows:
   "sharded" topology, test runners MUST accept any type of sharded cluster (i.e.
   "sharded" implies "sharded-replicaset", but not vice versa).
 
+- ``serverParameters``: Optional object of server parameters to check against.
+  To check server parameters, drivers send a ``{ getParameter: '*' }`` command
+  to the server using the internal MongoClient. The result SHOULD be cached to
+  avoid repeated calls. The result is then compared to the expected server
+  parameters as described in the `Evaluating Matches`_ section.
```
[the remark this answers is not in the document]
alcaeus: That sounds sensible. I've changed the paragraph to suggest either fetching single parameters or all of them and treat missing parameters as not equal. I've added a sentence to require test runners to evaluate these conditions in the order of the list above (min/max version, topology, parameters) which should eliminate cases where parameters don't exist, but I figured that if we're doing an exact match on parameters, a missing parameter should always cause tests to be skipped.

source/versioned-api/versioned-api.rst
alcaeus: Removed any mention of a default value corresponding with the above change requiring drivers to not send apiStrict and apiDeprecationErrors unless they were explicitly specified.

source/versioned-api/versioned-api.rst
alcaeus: Added in a new paragraph:
> In contrast, the ``killCursors`` command accepts API parameters and drivers MUST include them as they would with all other commands.

source/versioned-api/tests/README.rst
alcaeus: Yes, I'll add that.

While setting up the mongo-orchestration config locally I realised that this doesn't work completely yet: the server is started correctly, but the `mo` binary does not return. This is due to mongo-orchestration not sending an api version for any commands that it wants to run, thus hanging. This can't be resolved until PyMongo is updated to support the versioned API.

source/versioned-api/versioned-api.rst
jyemin: lgtm, but leaving for patrick to resolve

source/versioned-api/tests/README.rst
jyemin: Oh, ouch.  For purposes of this PR, it's sufficient to just add the prose here.  We can work out orchestration issues separately.

alcaeus: Right. Added the following paragraph:
> Drivers MUST test against a server with the ``requireApiVersion`` parameter
> enabled that also requires authentication. Since API versioning options can't be
> specified using the connection string, drivers MUST rely on environment
> variables to receive the API version to use in tests. The ``apiStrict`` and
> ``apiDeprecationErrors`` options are not required for this test.

jyemin: This still leaves a small hole in test coverage, as we don't ensure that apiVersion is included for all auth mechanisms. But I think it's the best we can do right now.

jyemin (approved):
LGTM!

patrickfreed (approved):
LGTM, looks great!

p-mongo (approved):

source/versioned-api/versioned-api.rst
```diff
@@ -0,0 +1,282 @@
+===================
+Unified Test Format
```
patrickfreed, line 2: Oh, just noticed that this needs to be updated. Surprised we all missed the biggest text in the spec :smile:

PR #878, opened 2020-11-20
SPEC-1530 / SPEC-1242 Allow connection pool to be paused

patrickfreed:
SPEC-1530 / SPEC-1242

This PR introduces a new "paused" state to the CMAP connection pool. When paused, the pool halts creating background connections and returns an error for any checkOut requests (SPEC-1242). A pool becomes paused by being cleared (e.g. due to a connection error) and is marked as "ready" again when a monitor performs a successful check. Also, as part of clearing, all requests in the WaitQueue will be eagerly evicted and return with retryable errors (SPEC-1530).

This is the third milestone in DRIVERS-781: Avoiding Connection Storms.


source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -502,11 +547,14 @@ semaphore, a condition variable may also be needed to to meet this
 requirement. Waiting on the condition variable SHOULD also be limited by the
 WaitQueueTimeout, if the driver supports one and it was specified by the user.
 
-If the pool is closed, any attempt to check out a `Connection <#connection>`_ MUST throw an Error, and any items in the waitQueue MUST be removed from the waitQueue and throw an Error.
+If the pool is "closed" or "paused", any attempt to check out a `Connection
+<#connection>`_ MUST throw an Error, and any items in the waitQueue MUST be
+removed from the waitQueue and throw an Error. The error thrown as a result of
+the pool being "paused" MUST be retryable.
```
patrickfreed, line 553: I wanted to add a unified format test to the retryable-reads spec for this, but they don't have any unified format tests yet. Should I just write a prose test for it now that can be converted once that spec is moved to unified format, just write the unified format one now, or don't write any test and wait? All seem reasonable to me.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst
```diff
@@ -178,3 +178,35 @@ Attempt a write to a collection.
 
 Outcome: Verify that the write succeeded or failed depending on existing
 driver behavior with respect to the starting topology.
+
+Connection Pool Management
```
patrickfreed, line 182: The Rust driver doesn't have SDAM monitoring yet, so I could only implement a portion of this test. Let me know how the full implementation goes in Python.

ShaneHarvey (commented):
Still need to finish the tests.

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -667,6 +667,12 @@ The event API here is assumed to be like the standard `Python Event
                 # Wait before running the next check.
                 wait()
                 continue
+
+            # for drivers that implement CMAP, mark the connection pool as ready
+            # once the server transitions to a known state.
```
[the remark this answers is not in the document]
patrickfreed: Ah good catch. Modifying connection pool state in `onServerDescriptionChanged` might be slightly problematic for some drivers though, since that is probably a method on the `TopologyDescription` instead of on the monitor. Would Python's implementation be able to plumb the connection pool through to its equivalent of that method? I think Rust could, but I'm not sure if that's true for every driver.

I'm wondering if marking the pool as ready here before calling that method might be okay, seeing as marking the pool as ready is correct so long as the server is _up_ (doesn't matter if it's a secondary / primary or if the status of another server changes). The only bad case I can think of is if between the server responding to the check successfully and that response actually getting back to the monitor, the server being checked went down and it was reported via another means. This seems rare enough to me that it might be okay to just mark the pool as ready without checking the topology version. What do you think?

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -607,7 +658,23 @@ Otherwise, the `Connection <#connection>`_ is marked as available.
...
+to Connection Monitoring events, a PoolClearedEvent MUST be emitted after
+incrementing the generation / marking the pool as "paused".
+
+As part of clearing the pool, the WaitQueue MUST also be cleared, meaning all
+requests in the WaitQueue MUST fail with errors indicating that the pool was
+cleared while the checkOut was being performed. The error returned MUST be
+considered a retryable error. Clearing the WaitQueue MUST happen eagerly so that
+any operations waiting on `Connections <#connection>`_ can retry as soon as
+possible. The pool MUST NOT rely on WaitQueueTimeoutMS to clear requests from
+the WaitQueue.
```
nbbeeken, line 677: Wouldn't this be the first driver made retryable error? I assume it should follow the same rules as other retryable errors, so would retryReads=false make this throw?

patrickfreed: Drivers already attach `RetryableWriteError` label to network errors (which this basically is). See here: https://github.com/mongodb/specifications/blob/master/source/retryable-writes/retryable-writes.rst#determining-retryable-errors

And yeah, in the case of retryWrites=false, this would fail those operations. Note that these operations would likely fail anyways if retryWrites=false, just that it happens sooner now and without attempting a connection creation.

source/server-discovery-and-monitoring/server-monitoring.rst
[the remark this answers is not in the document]
patrickfreed: Updated the server monitoring spec to explicitly require this happening before `onServerDescriptionChanged`. I also updated the CMAP spec to more clearly describe the various behaviors of the pool.

source/connection-monitoring-and-pooling/tests/pool-clear-clears-waitqueue.yml
```diff
@@ -0,0 +1,54 @@
...
+    thread: thread1
+  - name: start
+    target: thread2
+  - name: checkOut
+    thread: thread2
+  - name: start
+    target: thread3 
+  - name: checkOut
+    thread: thread3
+  # clear the pool, ejecting the previous three threads
```
ShaneHarvey, line 25: There is a race here where the clear() can happen before the 3 threads emit ConnectionCheckOutStarted which messes up the order in the event assertions. I suggest adding:
```
  # wait for the 3 threads to start checking out a connection
  - name: waitForEvent
    event: ConnectionCheckOutStarted
    count: 4
  # clear the pool, ejecting the previous three threads
  # from the WaitQueue
  - name: clear
  # The 3 threads are immediately ejected
  - name: waitForEvent
    event: ConnectionCheckOutFailed
    count: 4
```

Also, I believe this test will succeed regardless if the driver clears the wait queue or not because the checkout attempts will timeout after 1000ms. To fix this I think we need to check the original connection back in:
```
# ... continuing snippet from above
  - name: checkIn
    connection: conn
```

ShaneHarvey (changes requested):
I commented on two of the tests that are failing in my implementation. There are 6 more failing spec tests:
```
FAIL: test_cmap_pool_clear_min_size_pool_clear_halts_background_minPoolSize_establishments (test.test_cmap.TestCMAP)
  File "/Users/shane/git/mongo-python-driver/test/__init__.py", line 444, in wrap
    return f(*args, **kwargs)
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 405, in run_scenario
    self.run_scenario(scenario_def, test)
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 246, in run_scenario
    self.check_events(test['events'], test['ignore'])
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 184, in check_events
    self.fail('missing events: %r' % (events[len(actual_events):],))
AssertionError: missing events: [{'type': 'ConnectionCreated', 'address': 42}, {'type': 'ConnectionReady', 'address': 42}]

======================================================================
FAIL: test_cmap_pool_clear_ready_after_clear,_cannot_check_out_connections_until_pool_ready (test.test_cmap.TestCMAP)
  File "/Users/shane/git/mongo-python-driver/test/__init__.py", line 444, in wrap
    return f(*args, **kwargs)
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 405, in run_scenario
    self.run_scenario(scenario_def, test)
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 244, in run_scenario
    self.run_operations(test['operations'])
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 152, in run_operations
    self.run_operation(op)
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 146, in run_operation
    meth(op)
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 108, in wait_for_event
    wait_until(lambda: self.listener.event_count(event) >= count,
  File "/Users/shane/git/mongo-python-driver/test/utils.py", line 698, in wait_until
    raise AssertionError("Didn't ever %s" % success_description)
AssertionError: Didn't ever find 1 <class 'pymongo.monitoring.ConnectionCheckOutFailedEvent'> event(s)

======================================================================
FAIL: test_cmap_pool_close_must_be_able_to_manually_close_a_pool (test.test_cmap.TestCMAP)
  File "/Users/shane/git/mongo-python-driver/test/__init__.py", line 444, in wrap
    return f(*args, **kwargs)
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 405, in run_scenario
    self.run_scenario(scenario_def, test)
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 246, in run_scenario
    self.check_events(test['events'], test['ignore'])
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 181, in check_events
    self.check_event(actual, expected)
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 169, in check_event
    self.check_object(actual, expected)
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 156, in check_object
    self.assertEqual(type(actual), OBJECT_TYPES[expected['type']])
AssertionError: <class 'pymongo.monitoring.PoolReadyEvent'> != <class 'pymongo.monitoring.PoolClosedEvent'>

======================================================================
FAIL: test_cmap_pool_create_min_size_error_error_during_minPoolSize_population_clears_pool (test.test_cmap.TestCMAP)
  File "/Users/shane/git/mongo-python-driver/test/__init__.py", line 444, in wrap
    return f(*args, **kwargs)
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 405, in run_scenario
    self.run_scenario(scenario_def, test)
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 244, in run_scenario
    self.run_operations(test['operations'])
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 152, in run_operations
    self.run_operation(op)
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 146, in run_operation
    meth(op)
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 108, in wait_for_event
    wait_until(lambda: self.listener.event_count(event) >= count,
  File "/Users/shane/git/mongo-python-driver/test/utils.py", line 698, in wait_until
    raise AssertionError("Didn't ever %s" % success_description)
AssertionError: Didn't ever find 1 <class 'pymongo.monitoring.ConnectionClosedEvent'> event(s)

======================================================================
FAIL: test_cmap_pool_create_must_be_able_to_create_a_pool (test.test_cmap.TestCMAP)
  File "/Users/shane/git/mongo-python-driver/test/__init__.py", line 444, in wrap
    return f(*args, **kwargs)
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 405, in run_scenario
    self.run_scenario(scenario_def, test)
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 246, in run_scenario
    self.check_events(test['events'], test['ignore'])
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 186, in check_events
    self.fail('extra events: %r' % (actual_events[len(events):],))
AssertionError: extra events: [PoolReadyEvent(('localhost', 27019))]

======================================================================
FAIL: test_cmap_pool_create_with_options_must_be_able_to_start_a_pool_with_various_options_set (test.test_cmap.TestCMAP)
  File "/Users/shane/git/mongo-python-driver/test/__init__.py", line 444, in wrap
    return f(*args, **kwargs)
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 405, in run_scenario
    self.run_scenario(scenario_def, test)
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 246, in run_scenario
    self.check_events(test['events'], test['ignore'])
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 186, in check_events
    self.fail('extra events: %r' % (actual_events[len(events):],))
AssertionError: extra events: [PoolReadyEvent(('localhost', 27019))]

======================================================================
FAIL: test_cmap_pool_ready_pool_starts_as_cleared_and_becomes_ready (test.test_cmap.TestCMAP)
  File "/Users/shane/git/mongo-python-driver/test/__init__.py", line 444, in wrap
    return f(*args, **kwargs)
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 405, in run_scenario
    self.run_scenario(scenario_def, test)
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 244, in run_scenario
    self.run_operations(test['operations'])
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 152, in run_operations
    self.run_operation(op)
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 146, in run_operation
    meth(op)
  File "/Users/shane/git/mongo-python-driver/test/test_cmap.py", line 108, in wait_for_event
    wait_until(lambda: self.listener.event_count(event) >= count,
  File "/Users/shane/git/mongo-python-driver/test/utils.py", line 698, in wait_until
    raise AssertionError("Didn't ever %s" % success_description)
AssertionError: Didn't ever find 1 <class 'pymongo.monitoring.ConnectionCheckOutFailedEvent'> event(s)
```

source/connection-monitoring-and-pooling/tests/pool-clear-clears-waitqueue.yml
patrickfreed: Good catch, fixed.

Note: I put the check in before waiting for the events so that the even expectations can ensure the evictions happen immediately.

source/connection-monitoring-and-pooling/tests/pool-checkout-no-stale.json
```diff
@@ -3,6 +3,9 @@
   "style": "unit",
   "description": "must destroy and must not check out a stale connection if found while iterating available connections",
   "operations": [
+    {
+      "name": "ready"
+    },
```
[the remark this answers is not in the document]
patrickfreed: Fixed.

source/connection-monitoring-and-pooling/tests/pool-clear-min-size.yml
```diff
@@ -0,0 +1,36 @@
...
+  minPoolSize: 1
+operations:
+  - name: ready
+  - name: waitForEvent
+    event: ConnectionReady
+    count: 1
+  - name: clear
+  # ensure no connections created after clear
+  - name: wait
+    ms: 200
```
ShaneHarvey, line 14: In Python our minPoolSize thread runs with a 1 second interval so this test doesn't really work. I think some drivers have a 60 second interval, should we provide some guidance in the README to reduce the frequency to ~50ms?

source/connection-monitoring-and-pooling/tests/pool-clear-min-size.yml
```diff
@@ -0,0 +1,36 @@
...
+  - name: ready
+  - name: waitForEvent
+    event: ConnectionReady
+    count: 1
```
ShaneHarvey, line 18: This should be `count: 2` otherwise we won't actually wait for the 2nd connection to be created.

source/connection-monitoring-and-pooling/tests/pool-clear-ready.yml
```diff
@@ -0,0 +1,36 @@
...
+    event: ConnectionCheckOutFailed
+    count: 1
+  - name: ready
+  - name: checkOut
+events:
+  - type: ConnectionPoolReady
+    address: 42
+  - type: ConnectionCheckedOut
+    address: 42
+    id: 42
```
ShaneHarvey, line 22: This should be `connectionId: 42` not `id`.

source/connection-monitoring-and-pooling/tests/pool-clear-clears-waitqueue.yml
ShaneHarvey: Hmm I think there's still one more problem. In sync drivers pool.clear() notifies the waiting threads to wake up but doesn't wait for them. This means that the checkIn races with the threads and the order of the ConnectionCheckOutFailed/ConnectionCheckedIn events is non-deterministic. 

I think the only way to fix this is to change the test to use a high waitQueueTimeoutMS (>10seconds) and ignore the ConnectionCheckedIn event.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -261,6 +261,13 @@ has the following properties:
    -  Checking in a `Connection <#connection>`_ to the Pool automatically closes the `Connection <#connection>`_
    -  Attempting to check out a `Connection <#connection>`_ from the Pool results in an Error
 
+-  **Clearable:** A Pool MUST be able to be cleared. Clearing the pool marks all pooled and checked out `Connections <#connection>`_ as stale and lazily closes them as they are checkedIn or encountered in checkOut. Additionally, all requests are evicted from the WaitQueue and returned a retryable error.
```
ShaneHarvey, line 264: returned -> return?

patrickfreed: fixed

source/connection-monitoring-and-pooling/tests/pool-clear-clears-waitqueue.yml
patrickfreed: Fixed.

source/connection-monitoring-and-pooling/tests/pool-clear-min-size.yml
patrickfreed: Done. This is only a temporary fix for now, I think we should standardize on this across drivers, especially if the amount seems to vary by so much. Filed DRIVERS-1465 to track the follow on work for that.

source/connection-monitoring-and-pooling/tests/pool-clear-min-size.yml
patrickfreed: good catch, fixed.

source/connection-monitoring-and-pooling/tests/pool-clear-ready.yml
patrickfreed: fixed

source/connection-monitoring-and-pooling/tests/pool-clear-clears-waitqueue.yml
```diff
@@ -31,8 +31,7 @@ operations:
   # from the WaitQueue
   - name: clear
   # the 3 threads are immediately ejected from the WaitQueue,
-  # so the ConnectionCheckOutFailed events should be emitted
-  # before the ConnectionCheckIn event here.
+  # so this checkIn should cause any checkOuts to succeed.
```
ShaneHarvey, line 34: "should cause" -> "should not cause"

source/connection-monitoring-and-pooling/tests/pool-clear-clears-waitqueue.yml
```diff
@@ -60,10 +59,9 @@ events:
...
     reason: connectionError
     address: 42
-  - type: ConnectionCheckedIn
-    address: 42
 ignore:
   - ConnectionPoolReady
   - ConnectionPoolCreated
   - ConnectionCreated
   - ConnectionReady
+  - ConnectionCheckedIn
```
ShaneHarvey, line 67: I think we also need to ignore ConnectionClosed event. Isn't the connection closed during the checkin process because we're checking in a "stale" connection?

source/connection-monitoring-and-pooling/tests/pool-clear-clears-waitqueue.yml
ShaneHarvey: Upon reviewing this again I think we need to remove the `checkIn` and increase waitQueueTimeoutMS from 10000 to 30000. Including the `checkIn` hides a bug in the python implementation. The bug is that checkIn wakes up the waiting threads but clear() does not.

patrickfreed: Okay, `waitForEvent` to include an optional timeout used to ensure the events are seen quickly (within 1s which seemed generous enough).

source/connection-monitoring-and-pooling/tests/pool-clear-clears-waitqueue.yml
patrickfreed: this comment removed altogether now

source/connection-monitoring-and-pooling/tests/pool-clear-clears-waitqueue.yml
patrickfreed: Yep, nice catch. This uncovered a bug in the Rust runner where we weren't asserting that extra events were seen after matching all the expected ones. Fixed.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -459,14 +499,21 @@ connections managed by the pool is at least minPoolSize.
...
-   wait until pendingConnectionCount < maxConnecting
+   wait until pendingConnectionCount < maxConnecting and pool is "ready"
    create connection
-   establish connection
-   mark connection as available
+   try:
+     establish connection
+     mark connection as available
+   except error:
+     clear pool
```
ShaneHarvey, line 516: This case also needs to check generation number to avoid clearing the pool multiple times. Also, are these errors supposed to mark the server unknown too in addition to clearing the pool?

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -261,6 +261,13 @@ has the following properties:
    -  Checking in a `Connection <#connection>`_ to the Pool automatically closes the `Connection <#connection>`_
    -  Attempting to check out a `Connection <#connection>`_ from the Pool results in an Error
 
+-  **Clearable:** A Pool MUST be able to be cleared. Clearing the pool marks all pooled and checked out `Connections <#connection>`_ as stale and lazily closes them as they are checkedIn or encountered in checkOut. Additionally, all requests are evicted from the WaitQueue and return retryable errors.
+
+-  **Pausable:** A Pool MUST be able to be paused and resumed. A Pool is paused automatically when it is cleared, and it can be resumed by being marked as "ready". While the Pool is paused, it exhibits the following behaviors:
```
ShaneHarvey, line 266: Can we include the justification/motivation for the paused state in the "Design Rationale" section?

ShaneHarvey (commented):
The tests LGTM so far. A few comments on the prose.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
patrickfreed: >This case also needs to check generation number to avoid clearing the pool multiple times.

Good point, fixed.

>Also, are these errors supposed to mark the server unknown too in addition to clearing the pool?

I left this as a "SHOULD", since I'm unsure if every driver is capable of implementing this, and if a driver doesn't implement it, SDAM will pick up the error nonetheless through the next `checkOut` and update the topology accordingly. I added some more explicit prose to make this clear and also added a line to the psuedocode. 

Additionally, to prevent repeated PoolCleared events in the case that the background thread doesn't update the topology, I added some prose requiring that paused pools don't emit PoolCleared events but do increment their generation count. I added rationale for this to the rationale section.

mbroadst (approved):
LGTM

patrickfreed:
@ShaneHarvey this is RFAL. I think I've responded to all the open comments / suggestions.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -901,6 +988,41 @@ again, so ensuring the socket is torn down does not need to happen
...
+2. The pool can evict requests that enter the WaitQueue after the pool was
+   cleared but before the server was in a known state again. Such requests can
+   occur when a server is selected at the same time as it becomes marked as
+   Unknown in highly concurrent workloads. Without the "paused" state, the pool
+   would attempt to service these requests, since it would assume they were
+   routed to the pool because its endpoint was available, not because of a race
+   between SDAM and Server Selection. These requests would then likely fail with
+   potentially high latency, again wasting resources both server and driver side.
+
+Why not emit PoolCleared events when clearing a paused pool?
```
ShaneHarvey, line 1014: This change breaks the following SDAM tests:
- https://github.com/mongodb/specifications/blob/f123e3a/source/server-discovery-and-monitoring/tests/integration/isMaster-command-error.yml#L39
- https://github.com/mongodb/specifications/blob/f123e3a/source/server-discovery-and-monitoring/tests/integration/isMaster-network-error.yml#L38
- https://github.com/mongodb/specifications/blob/f123e3a/source/server-discovery-and-monitoring/tests/integration/isMaster-timeout.yml#L39

See the follow test failure: https://evergreen.mongodb.com/task/mongo_python_driver_tests_python_version_rhel62_test_ssl__platform~rhel62_auth~auth_ssl~ssl_python_version~2.7_coverage~coverage_test_latest_standalone_patch_61232b7f21d8f0ad9023c8ef0912017d062fafc0_5fd02078c9ec4456b5c14943_20_12_09_00_55_21

Since the pool initially starts in the paused state the PoolClearedEvent which those tests expect is no longer emitted. I propose we revert this change and continue to emit PoolCleared events when clearing a paused pool. This way we won't create a dependency on https://jira.mongodb.org/browse/DRIVERS-1314.

patrickfreed: I'd rather update the tests than update the proposed driver functionality. After reading over the various tickets there, it seems Go's not having SDAM monitoring events was one of the hangups, but that doesn't seem to be an issue any more. The other hangup is that we wanted to resolve SPEC-1474 first, but now that we have the "paused" state, we can resolve DRIVERS-1314 without actually addressing SPEC-1474, since regardless of whether the pool is created or not, no PoolClearedEvent will be emitted anyways. The test update should then be pretty simple and localized to just removing these PoolClearedEvents.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -459,14 +499,27 @@ connections managed by the pool is at least minPoolSize.
...
-I/O.
+I/O. Populating the pool MUST NOT be performed unless the pool is "ready".
+
+If an error is encountered while populating a connection, the pool MUST be
+cleared if the `Connection's <#connection>`_ generation is greater than or equal
+to the pool's generation. The error SHOULD be handled like a network error that
+occurred *Before the handshake completed* according to the `Network error when
+reading or writing`_ section in the SDAM specification, if possible in the
+driver's implementation. If not possible, then the error will get handled that
+way by the existing SDAM machinery next time checkOut is called.
```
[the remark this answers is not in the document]
patrickfreed: I incorporated this link into the existing language to ensure this error still gets treated as a network error. Let me know what you think.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -459,14 +499,27 @@ connections managed by the pool is at least minPoolSize.
...
    create connection
-   establish connection
-   mark connection as available
+   try:
+     establish connection
+     mark connection as available
+   except error:
+     if connection.generation >= pool.generation:
+       clear pool
+       handle pre handshake network error # if possible
```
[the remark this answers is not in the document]
patrickfreed: Updated. I named it `topology.handle_pre_handshake_error` to be a little more specific (that's also what we call it in Rust).

source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst
```diff
@@ -209,3 +209,35 @@ This test requires MongoDB 4.9.0+.
...
+--------------------------
+
+This test will be used to ensure monitors properly create and unpause connection
+pools when they discover servers.
+
+1. Create a client with directConnection=true, appName="SDAMPoolManagementTest",
+   and heartbeatFrequencyMS=500 (or lower if possible).
+
+2. Verify via SDAM and CMAP event monitoring that a ConnectionPoolCreatedEvent occurs
```
ShaneHarvey, line 222: `ConnectionPoolCreatedEvent` -> `ConnectionPoolReadyEvent`

source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst
```diff
@@ -209,3 +209,35 @@ This test requires MongoDB 4.9.0+.
...
+   after the first ServerHeartbeatSucceededEvent event does.
+
+3. Enable the following failpoint::
+
+     {
+         configureFailPoint: "failCommand",
+         mode: { times: 1 },
```
ShaneHarvey, line 229: I think we should bump this to `times: 2` to account for the possibility that the RTT monitor thread might trigger the failCommand once.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst
ShaneHarvey: It works with the changes I suggested below. We should also mention this test requires failCommand appName support which is now >=4.2.9 (see SERVER-47195).

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
ShaneHarvey: Okay SGTM, let's remove the PoolClearedEvents from the failing "command/network error/timeout on Monitor handshake" tests.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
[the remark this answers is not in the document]
ShaneHarvey: I think we should remove the "network error" wording since this could fail with a command error as well. When it fails with a command error I expect the topology to inspect both generation and topologyVersion (if present in the response).

source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst
patrickfreed: Fixed

source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst
patrickfreed: Done

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
patrickfreed: Done

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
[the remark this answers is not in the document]
patrickfreed: I think I had this section confused with the section describing the error we return from `checkOut`. I removed all the extra stuff I kept about pre-handshake / network error so it looks more like your original suggestion.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst
patrickfreed: Done

patrickfreed:
I know we talked about moving the tests out of `server-discovery-and-monitoring-tests.rst` and into `test/README.rst#prose-tests`, but it seems like prose tests might be more suited to the former location. I think the README is intended to just describe how to run the spec tests from this line in `server-discovery-and-monitoring-tests.rst`:

>See also the YAML test files and their accompanying README in the "tests"
directory.

Also note that there's a few other existing prose tests in `server-discovery-and-monitoring-tests.rst` already (excluding the one I put there).

ShaneHarvey (approved):
LGTM!

I'll open a drivers ticket to address the prose test location issue.

patrickfreed:
Going to leave this PR up for a few days while the Rust PR goes through review in case anything comes up.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst
patrickfreed: Could this test be written as an SDAM integration spec test?

source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst
```diff
@@ -209,3 +209,38 @@ This test requires MongoDB 4.9.0+.
```
patrickfreed, line 209: I think we're missing a test here that verifies the minPoolSize -> SDAM communication. 

What do you think about this test:
1. Create setup client with directConnection=true
2. Enable the following failpoint using the setup client: 
```
     {
         mode: { skip: 2 },
         data: {
             failCommands: ["isMaster"],
             errorCode: 1234,
             appName: "SDAMMinPoolSizeError"
         }
     }
```
3. Create client with directConnection=true, appName="SDAMMinPoolSizeError", minPoolSize=10, serverSelectionTimeoutMS=100, heartbeatFrequencyMS=10000

4. Verify via SDAM and CMAP event monitoring on the client that a ConnectionPoolReadyEvent occurs
   after the first ServerHeartbeatSucceededEvent event does.

5. Verify that an PoolClearedEvent is emitted

6. Execute a `ping` command using the client and assert that it fails due to server selection timeout

7. Disable the failpoint using the setup client

8. Verify via SDAM and CMAP event monitoring that a ConnectionPoolReadyEvent occurs
   after the next ServerHeartbeatSucceededEvent event.

9. Execute a `ping` command using the client and assert that it succeeds.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -607,7 +667,26 @@ Otherwise, the `Connection <#connection>`_ is marked as available.
...
+until it is marked as "ready" again. The checkOut and checkIn algorithms will
+handle clearing out stale `Connections <#connection>`_. If a user is subscribed
+to Connection Monitoring events, a PoolClearedEvent MUST be emitted after
+incrementing the generation / marking the pool as "paused". If the pool is
+already "paused" when it is cleared, then the pool MUST NOT emit a PoolCleared
+event.
+
+As part of clearing the pool, the WaitQueue MUST also be cleared, meaning all
+requests in the WaitQueue MUST fail with errors indicating that the pool was
+cleared while the checkOut was being performed. The error returned as a result
```
patrickfreed, line 684: I updated the description of the error here to match the description of the error in `checkOut`.

patrickfreed (commented):
@ShaneHarvey I left a few comments based on some things I noticed while updating my implementation (a new test, a minor wording change). PTAL when you get a chance.

Also, we may have an issue here when it comes to handling authentication errors (or generally, any command errors that occur during the handshake). If drivers aren't marking the server as Unknown in response to them, then we will be in a bad situation where the pool is paused (since drivers MUST clear the pool after auth errors) but the server is still selectable (monitors don't perform auth). The server selection spec seems to imply that auth errors can mark the server as Unknown, and that would jibe nicely with the pool changes if so. I'm just not sure if drivers actually do that because it isn't explicitly mandated by the specs anywhere. What does Python do in this case?

At any rate, I filed DRIVERS-1476 to discuss this issue. This is a bit of a blocker for this PR though, so I'd like to expedite the discussion on this if at all possible to get this project closed out next week.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst
ShaneHarvey: The key assertion that "ConnectionPoolReadyEvent occurs after the first ServerHeartbeatSucceededEvent" is not possible in the SDAM test format. So I would say no.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst
ShaneHarvey: I think this test can be implemented with the SDAM integration format because the assertion in step 4) isn't that critical to this test. The main point is that the background thread clears the pool, right?

ShaneHarvey (commented):
> What does Python do in this case?

Currently python does not clear the pool or mark the server unknown when auth fails with command errors. I propose we send _any_ error that happens before the handshake completes (including auth) to SDAM which will mark the server unknown + clear the pool as appropriate.

patrickfreed:
I agree with the all errors go through SDAM approach. Do we need to consult the wider drivers team or can I just go ahead and make that clarification to SDAM? Without it, we could end up in a situation where a driver clears a pool due to auth error but doesn't mark the server as Unknown, leading to errors.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst
patrickfreed: Yeah the assertion in step 4 isn't necessary to this test, and the one in step 8 is also not really necessary, though it would be nice to have. I went ahead and took a stab at writing an SDAM spec test, but Rust doesn't have the runner (nor SDAM monitoring) implemented, so I was driving a bit blind. Let me know if the test parses + passes in Python.

source/server-discovery-and-monitoring/tests/integration/minPoolSize-error.json
```diff
@@ -0,0 +1,104 @@
...
+        {
+          "name": "runCommand",
+          "object": "database",
+          "command_name": "ping",
+          "arguments": {
+            "command": {
+              "ping": {}
+            }
+          },
+          "error": false
```
ShaneHarvey, line 67: This needs to be `"error": true`.  The yaml file is correct and this json file is out of sync.

source/server-discovery-and-monitoring/tests/integration/minPoolSize-error.yml
```diff
@@ -0,0 +1,79 @@
...
+        error: true
+      # Disable the failpoint, allowing the monitor to discover the primary again.
+      - name: configureFailPoint
+        object: testRunner
+        arguments:
+          failPoint:
+            configureFailPoint: failCommand
+            mode: off
+            data:
+              errorCode: 91 # ShutdownInProgress
```
ShaneHarvey, line 65: Remove `data` here, it's not needed when disabling the failpoint.

source/server-discovery-and-monitoring/tests/integration/minPoolSize-error.yml
```diff
@@ -0,0 +1,79 @@
...
+        arguments:
+          event: PoolReadyEvent
+          count: 1
+      # Perform an operation to ensure the node is discovered.
+      - name: runCommand
+        object: database
+        command_name: ping
+        arguments:
+          command:
+            ping: {}
```
ShaneHarvey, line 78: I think this should be `ping: 1`.

source/server-discovery-and-monitoring/tests/integration/minPoolSize-error.yml
```diff
@@ -0,0 +1,79 @@
...
+
+data: []
+
+tests:
+  - description: Command error on minPoolSize background creation
+    # Configure the initial monitor handshake to succeed but the
+    # first or second background minPoolSize establishments to fail.
+    failPoint:
+      configureFailPoint: failCommand
+      mode: { skip: 2 }
```
ShaneHarvey, line 17: I think this should be `skip: 3` to account for the initial monitor handshake, the initial streaming hello, and the initial RTT connection attempt.

source/server-discovery-and-monitoring/tests/integration/minPoolSize-error.yml
```diff
@@ -0,0 +1,79 @@
+# Test SDAM error handling.
+runOn:
+    # failCommand appName requirements
+    - minServerVersion: "4.4"
```
ShaneHarvey, line 4: This test only works correctly on 4.9+ because of [SERVER-49336](https://jira.mongodb.org/browse/SERVER-49336).

source/server-discovery-and-monitoring/tests/integration/minPoolSize-error.yml
```diff
@@ -0,0 +1,79 @@
...
+    # Configure the initial monitor handshake to succeed but the
+    # first or second background minPoolSize establishments to fail.
+    failPoint:
+      configureFailPoint: failCommand
+      mode: { skip: 2 }
+      data:
+        failCommands: ["isMaster"]
+        appName: SDAMminPoolSizeError
+        closeConnection: false
+        errorCode: 91 # ShutdownInProgress
```
ShaneHarvey, line 22: When this test uses a command error, the driver does not cancel the inprogress streaming isMaster which messes up the assertions in this test. I suggest we change this to be `closeConnection: true`.

source/server-discovery-and-monitoring/tests/integration/minPoolSize-error.yml
```diff
@@ -0,0 +1,79 @@
...
+            configureFailPoint: failCommand
+            mode: off
+            data:
+              errorCode: 91 # ShutdownInProgress
+      # Wait for monitor to discover the primary and mark the pool as ready again.
+      - name: waitForEvent
+        object: testRunner
+        arguments:
+          event: PoolReadyEvent
+          count: 1
```
ShaneHarvey, line 71: This should be `count:2` because we want to wait for the 2nd PoolReadyEvent.

source/server-discovery-and-monitoring/tests/integration/minPoolSize-error.yml
patrickfreed: fixed

source/server-discovery-and-monitoring/tests/integration/minPoolSize-error.yml
patrickfreed: fixed

source/server-discovery-and-monitoring/tests/integration/minPoolSize-error.yml
patrickfreed: Fixed

source/server-discovery-and-monitoring/tests/integration/minPoolSize-error.yml
patrickfreed: fixed

source/server-discovery-and-monitoring/tests/integration/minPoolSize-error.yml
patrickfreed: fixed

source/server-discovery-and-monitoring/tests/integration/minPoolSize-error.json
patrickfreed: fixed

patrickfreed (commented):
Fixed the issues with the test, sorry for its rough state.

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -667,6 +678,12 @@ The event API here is assumed to be like the standard `Python Event
...
+            if previousDescription.type == Unknown and description.type != Unknown:
+                mark connection pool for server as "ready"
```
ShaneHarvey, line 685: I think it would be safer to mark the pool ready every time the server is in a known state:
```
            if description.type != Unknown:
                mark connection pool for server as "ready"
```

This would be a safeguard in case a driver has a bug where they pause the pool without marking the server unknown. In the current design such a bug would result in the pool being paused forever. With this change, the pool would only be paused for at most heartbeatFrequencyMS.

We should update the prose above to remove "`that was previously
Unknown`".

ShaneHarvey (commented):
> I agree with the all errors go through SDAM approach. Do we need to consult the wider drivers team or can I just go ahead and make that clarification to SDAM? Without it, we could end up in a situation where a driver clears a pool due to auth error but doesn't mark the server as Unknown, leading to errors.

Let's address this problem in https://jira.mongodb.org/browse/DRIVERS-1476.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -335,25 +368,18 @@ The SDAM specification defines `when
...
-MUST be done in a non-blocking manner, such as via the use of a background
-thread or asynchronous I/O. See `Populating the Pool with a Connection
-<#populating-the-pool-with-a-connection-internal-implementation>`_ for more
-details on the steps involved.
+When a pool is created, its state MUST initially be set to "paused". Even if
+minPoolSize is set, the pool MUST NOT begin being `populated
+<#populating-the-pool-with-a-connection-internal-implementation>`_ with
+`Connections <#connection>`_ until it has been marked as "ready". Monitoring threads will
+ready the pool once they have successfully performed a check and
+transitioned the server to a known ServerType.
```
ShaneHarvey, line 376: `Monitoring threads` -> `SDAM will mark the pool as "ready" on each successful check.`

Also I think it would help to link to the SDAM `Connection Pool Management` section.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -261,6 +261,13 @@ has the following properties:
    -  Checking in a `Connection <#connection>`_ to the Pool automatically closes the `Connection <#connection>`_
    -  Attempting to check out a `Connection <#connection>`_ from the Pool results in an Error
 
+-  **Clearable:** A Pool MUST be able to be cleared. Clearing the pool marks all pooled and checked out `Connections <#connection>`_ as stale and lazily closes them as they are checkedIn or encountered in checkOut. Additionally, all requests are evicted from the WaitQueue and return retryable errors.
+
+-  **Pausable:** A Pool MUST be able to be paused and resumed. A Pool is paused automatically when it is cleared, and it can be resumed by being marked as "ready". While the Pool is paused, it exhibits the following behaviors:
+
+   -  Attempting to check out a `Connection <#connection>`_ from the Pool results in a retryable Error
```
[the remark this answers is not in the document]
patrickfreed: fixed

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
patrickfreed: done

source/server-discovery-and-monitoring/server-monitoring.rst
patrickfreed: done

source/server-discovery-and-monitoring/tests/integration/minPoolSize-error.yml
```diff
@@ -14,12 +14,11 @@ tests:
     # first or second background minPoolSize establishments to fail.
```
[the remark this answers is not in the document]
patrickfreed: done

patrickfreed (commented):
I included some rationale about explicitly synchronizing pool clears with SDAM updates and updated the psuedocode to reflect that, let me know if it seems okay.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1384,7 +1397,11 @@ recovering" error::
 
         # Mark the server Unknown
         unknown = new ServerDescription(type=Unknown, error=message, code=code, topologyVersion=response["topologyVersion"])
-        onServerDescriptionChanged(unknown)
+        with client.lock:
```
ShaneHarvey, line 1400: It may make more sense to hold the lock throughout this entire method to ensure that the topology is not updated after the `isStaleError` check but before the `client.lock`.

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -667,10 +679,17 @@ The event API here is assumed to be like the standard `Python Event
...
+            with client.lock:
+                # for drivers that implement CMAP, mark the connection pool as ready after
+                # performing a successful check.
+                if description.type != Unknown:
+                    mark connection pool for server as "ready"
+
+                topology.onServerDescriptionChanged(description)
+                if description.error != Null:
+                    # Clear the connection pool only after the server description is set to Unknown.
+                    clear connection pool for server
```
ShaneHarvey, line 692: We should make `onServerDescriptionChanged` call pool.ready(). The Monitor should not call this method directly because `onServerDescriptionChanged` deals with topolgyVersion checks and the Monitor does not. (see this comment: https://github.com/mongodb/specifications/pull/878/#discussion_r528914452)

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
patrickfreed: done

source/server-discovery-and-monitoring/server-monitoring.rst
patrickfreed: done

ShaneHarvey (commented):
One last thing, can you bump the minor version of SDAM and add a changelog entry to the server-monitoring and SDAM docs for the pool synchronization + pool.ready() change?

patrickfreed:
Done. Also updated CMAP

ShaneHarvey (approved):
LGTM

PR #879, opened 2020-11-24
SPEC-1764 Test that monitors wait minHeartbeatFrequencyMS between failed server checks

patrickfreed:
SPEC-1764

This PR adds a prose integration test to the SDAM spec to ensure that monitors wait at least 500ms between server checks even when a thread in server selection is requesting immediate checks.

### DIFF
--- a/source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst
+++ b/source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst
@@ -196,16 +196,16 @@ This test requires MongoDB 4.9.0+.
          data: {
              failCommands: ["isMaster"],
              errorCode: 1234,
-             appName: "SDAMSleepTest"
+             appName: "SDAMMinHeartbeatFrequencyTest"
          }
      }
 
-2. Create a client with directConnection=true, appName="SDAMSleepTest", and
-   serverSelectionTimeoutMS=unlimited.
+2. Create a client with directConnection=true, appName="SDAMMinHeartbeatFrequencyTest", and
+   serverSelectionTimeoutMS=10000.
 
 3. Start a timer.
 
-4. Execute a ``ping`` command with unlimited ServerSelectionTimeoutMS.
+4. Execute a ``ping`` command.
 
 5. Stop the timer. Assert that the ``ping`` took between 4.5 seconds and 6.5
    seconds to complete.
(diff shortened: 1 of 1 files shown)


source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst
```diff
@@ -178,3 +178,34 @@ Attempt a write to a collection.
 
 Outcome: Verify that the write succeeded or failed depending on existing
 driver behavior with respect to the starting topology.
+
+Monitors sleep at least minHeartbeatFreqencyMS between checks
```
patrickfreed, line 182: @ShaneHarvey This test is similar to the one you proposed, but it doesn't rely on setting heartbeatFrequencyMS to 500. This way it tests that minHeartbeatFrequencyMS is enforced even when immediate checks are being requested. Let me know what you think.

ShaneHarvey: This test seems reasonable to me. I'm testing it here: https://github.com/mongodb/mongo-python-driver/pull/524

A little disappointing that the test takes so long, could we reduce `times:10` to reduce the runtime?

patrickfreed: Reduced times to 5 and the expected time between 2 seconds and 3.5 seconds. I worry if we go too much shorter than that and we might lost some accuracy with the test.

As an aside, should we also add a test for the happy case to ensure heartbeatFrequencyMS is respected? Something like set it to 750ms and wait for 3.5 seconds and ensure at between 3 and 5 heartbeats occurred?

ShaneHarvey: Thanks, I'm testing the updated test here: https://evergreen.mongodb.com/version/5fc5a093d6d80a31f056c6c5

Hmmm, there may already be such a test for the happy path. I'll check for one. If there isn't one I'll open a new ticket for it instead of adding it here.

ShaneHarvey: I've double check are there isn't any existing test for the happy path behavior. I will create a drivers ticket for this tomorrow.

ShaneHarvey (approved):
LGTM! I implemented this in Python and it works as expected.

nbbeeken (approved):

mbroadst (approved):
LGTM

PR #880, opened 2020-11-24
SPEC-1668 conflicting requirements for AWS creds

bazile-clyde:
@p-mongo, I'm tagging you here just in case you wanted to review the fix to your reported issue. Feel free to ignore.


p-mongo:
I believe the language as written is now unambiguous.

With that said, I suggest adding a note or rationale explaining why the EC2 credentials are to be ignored when ECS environment variable is set, since I believe this is a valid deployment option (as described in the spec ticket).

emadum (approved):
LGTM 👍

p-mongo (approved):
Thanks!

PR #882, opened 2020-12-02
SPEC-1767 Fix race in CMAP test 'maxConnecting is enforced'

ShaneHarvey:
Tested in python here: https://evergreen.mongodb.com/version/5fc6afbdd1fe0771a322eaec
And a second time for good measure: https://evergreen.mongodb.com/version/5fc6edb032f417146256a3aa

### DIFF
--- a/source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-is-enforced.yml
+++ b/source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-is-enforced.yml
@@ -33,7 +33,7 @@ operations:
   - name: waitForEvent
     event: ConnectionCreated
     count: 1
-  # wait some more time to ensure thread1's Connection is ready first
+  # wait some more time to ensure thread1 has begun establishing a Connection
   - name: wait
     ms: 100
   # start 2 check out requests. Only one thread should
--- a/source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-is-enforced.yml
+++ b/source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-is-enforced.yml
@@ -18,29 +18,30 @@ poolOptions:
   maxPoolSize: 10
   waitQueueTimeoutMS: 5000
 operations:
-  # start creating a Connection. This will take a while
-  # due to the fail point.
+  # start 3 threads
   - name: start
     target: thread1
+  - name: start
+    target: thread2
+  - name: start
+    target: thread3
+  # start creating a Connection. This will take a while
+  # due to the fail point.
   - name: checkOut
     thread: thread1
-  # Start 2 new threads that wait for a little then try
-  # to check out connections. Only one thread should
+  # wait for thread1 to actually start creating a Connection
+  - name: waitForEvent
+    event: ConnectionCreated
+    count: 1
+  # wait some more time to ensure thread1's Connection is ready first
+  - name: wait
+    ms: 100
+  # start 2 check out requests. Only one thread should
   # start creating a Connection and the other one should be
   # waiting for pendingConnectionCount to be less than maxConnecting,
   # only starting once thread1 finishes creating its Connection.
-  - name: start
-    target: thread2
-  - name: wait
-    thread: thread2
-    ms: 100
   - name: checkOut
     thread: thread2
-  - name: start
-    target: thread3
-  - name: wait
-    thread: thread3
-    ms: 100
   - name: checkOut
     thread: thread3
   # wait until all Connections have been created.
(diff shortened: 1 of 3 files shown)


source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-is-enforced.yml
```diff
@@ -18,29 +18,30 @@ poolOptions:
...
+  # due to the fail point.
   - name: checkOut
     thread: thread1
-  # Start 2 new threads that wait for a little then try
-  # to check out connections. Only one thread should
+  # wait for thread1 to actually start creating a Connection
+  - name: waitForEvent
+    event: ConnectionCreated
+    count: 1
+  # wait some more time to ensure thread1's Connection is ready first
```
patrickfreed, line 36: The waiting here is to ensure thread1 has begun establishing a connection, not that the connection is ready (aka established).

source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-is-enforced.yml
```diff
@@ -18,29 +18,30 @@ poolOptions:
   maxPoolSize: 10
   waitQueueTimeoutMS: 5000
 operations:
-  # start creating a Connection. This will take a while
-  # due to the fail point.
+  # start 3 threads
   - name: start
     target: thread1
+  - name: start
```
patrickfreed, line 24: The way the Rust driver runs these tests is that it essentially separates each threads workloads into separate queues and executes them all concurrently, so having the starts happen before the main thread waits will cause all the checkOuts to happen simultaneously. This does seem like an odd way to perform these tests however. I think they're supposed to be run like the main thread dispatching tasks sequentially as they're listed in the format. Is this how Python runs the tests?

patrickfreed: So it looks like Java, Go, and Node dispatch the threaded tasks in order from the main thread, and Ruby separates them per thread at the beginning and begins execute them one after another once the `start` operation for a given thread is called like Rust does.

I believe the tests were _intended_ to be executed like the former since it makes them much easier to write and read. I'm happy to update the Rust runner to accommodate this, but it is a little worrisome that we'd also make Ruby do the same change. If we think it constitutes a bug in the two implementations, then its probably worth fixing though.

patrickfreed: After thinking about it some more, I feel like this is a bug in the Ruby and Rust implementations and that drivers should be dispatching threaded tasks from the main thread as opposed to preprocessing them on a per-thread basis, and that we should make this explicit in the spec readme. cc @divjotarora do you agree?

also cc @p-mongo

source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-is-enforced.yml
ShaneHarvey: Fixed.

source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-is-enforced.yml
ShaneHarvey: Yeah I agree that the operations should be dispatched in order.

divjotarora: Also agreed. I haven't read the spec tests README yet, but it's probably worth updating it to clarify this if we're considering it a bug in those test runners.

patrickfreed:
I updated the rust driver to dispatch tasks in order and pulled in your test, evg patch here: https://evergreen.mongodb.com/version/5fc832813e8e861c6404f5a1

If all goes well, even the windows tasks should all succeed. I'll probably run them a few times if they come back green to make sure.

source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-is-enforced.yml
patrickfreed: Yep, @ShaneHarvey could you add that as part of this PR then? I think this would be the first test that actually requires this ordering.

patrickfreed (commented):
Just restarted my evergreen patch, but the first run seemed to have no maxConnecting failures.

source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-is-enforced.yml
ShaneHarvey: Updated, PTAL.

patrickfreed (approved):

source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-is-enforced.yml
patrickfreed: looks good to me!

ShaneHarvey:
Waiting for @divjotarora 's review before merging.

divjotarora:
@ShaneHarvey I haven't been reviewing the other changes related to DRIVERS-781, so no need to wait for my review.

PR #881, opened 2020-12-01
DRIVERS-555 Client-side operations timeout spec

divjotarora:

PR #883, opened 2020-12-02
DRIVERS-555 Fix retryable reads/writes tests

divjotarora:
DRIVERS-555 changes retryability behavior to retry indefinitely regardless of whether or not the new `timeoutMS` option is used. There are tests added in https://github.com/mongodb/specifications/pull/873 for `timeoutMS` being set to both 0 (infinite) and non-zero. To check behavior when `timeoutMS` is unset, this PR modifies the existing retryable reads/writes tests to assert that an operation can succeed even if it encounters two retryable errors. 

As of commit 6b05e2abd458b93351fb48a91e4d05e17d82424f, this PR also contains changes to the transactions tests in order to fix retryable behavior for the abortTransaction and commitTransaction operations.

This PR must be combined with the changes in #881 as that PR makes changes to the actual retryable reads/writes specs to allow multiple retries.


source/retryable-writes/tests/README.rst
```diff
@@ -171,6 +171,10 @@ Each YAML file has the following keys:
 
   - ``description``: The name of the test.
 
+  - ``skipReason`` (optional): A string describing why the test should be
```
divjotarora, line 174: This was required so we could skip some bulkWrite tests due to https://jira.mongodb.org/browse/SERVER-53186.

source/retryable-writes/tests/insertMany.yml
```diff
@@ -47,28 +47,3 @@ tests:
                     - { _id: 1, x: 11 }
                     - { _id: 2, x: 22 }
                     - { _id: 3, x: 33 }
-    -
-        description: "InsertMany fails after multiple network errors"
```
divjotarora, line 51: I removed this test because it sets the fail point to "alwaysOn", so it loops forever after the changes in DRIVERS-555. Also, we're already testing insertMany with multiple network errors in `insertMany-serverErrors.yml` (test name is `InsertMany succeeds after two connection failures`), so I don't think removing this loses any test coverage.

divjotarora:
@jmikola @isabelatkinson You both were automatically added as reviewers due to the code owners setup. There was a small change to the retryable writes test format to add a `skipReason` field which behaves the same way as the unified test format. Feel free to remove yourselves from the review if you don't have any comments.

source/retryable-reads/retryable-reads.rst
```diff
@@ -471,7 +471,7 @@ for tests.
 
 At a high level, the test plan will cover executing supported read operations
 within a MongoClient where retryable reads have been enabled, ensuring that
-reads are retried exactly once.
+reads are retried.
```
bazile-clyde, line 474: These tests contradict the "retry once" behavior outlined above. For example, the abstract reads,

> This specification is about the ability for drivers to automatically retry any read operation that has not yet received any results—due to a transient network error, a "not master" error after a replica set failover, etc.—**exactly once**.

The pseudocode also shows a single retry. This spec should be updated too. Otherwise, Drivers will sync new tests, have those tests fail, and only have incorrect guidance on how retryable reads are supposed to behave.

bazile-clyde: Yes. Could you add a reference to it in your top comment? I guess I'll hold off on this until that PR is reviewed since changes in #881 could cause downstream changes.

divjotarora: Done.

durran (approved):
Looks good to me on the transaction tests side.

ShaneHarvey:
As outlined in https://jira.mongodb.org/browse/DRIVERS-1807, the default behavior of retryable reads/writes is not going to be changed. By default they will only retry once. If timeoutMS is enabled, then they will retry repeatedly until the timeout expires.

PR #886, opened 2020-12-08
DRIVERS-1460: Fix wrong description for waitQueueTimeoutMS in URI options spec

alcaeus:
DRIVERS-1460

### DIFF
--- a/source/uri-options/uri-options.rst
+++ b/source/uri-options/uri-options.rst
@@ -339,7 +339,8 @@ pertaining to URI options apply here.
      - positive number
      - defined in the `Connection Pooling spec`_
      - required for drivers with connection pools, with exceptions described in the `Connection Pooling spec`_
-     - The maximum number of clients or connections able to be created by a pool at a given time
+     - Amount of time spent attempting to check out a connection from a server's
+       connection pool before timing out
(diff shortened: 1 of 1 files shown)


saghm (approved):

source/uri-options/uri-options.rst
```diff
@@ -339,7 +339,8 @@ pertaining to URI options apply here.
      - positive number
      - defined in the `Connection Pooling spec`_
      - required for drivers with connection pools, with exceptions described in the `Connection Pooling spec`_
-     - The maximum number of clients or connections able to be created by a pool at a given time
+     - If specified, requests MUST time out after spending waitQueueTimeoutMS or
```
divjotarora, line 342: nit: The other timeout options don't reference themselves in their definitions. For example, the description for `socketTimeoutMS` is
> Amount of time spent attempting to send or receive on a socket before timing out

Maybe change this one to `Amount of time spent attempting to check out a connection from a server's connection pool before timing out`?

alcaeus: Makes sense, I applied your suggestion.

divjotarora (approved):

PR #887, opened 2020-12-10
Skip runCommand tests when requireApiVersion is true

alcaeus:
With requireApiVersion enabled, this test is no longer valid as the test client would always have an API version declared. As per the spec, declaring an API version in a command document for a command helper and declaring a version on the client is invalid and has undefined behaviour.

### DIFF
--- a/source/versioned-api/tests/runcommand-helper-no-api-version-declared.json
+++ b/source/versioned-api/tests/runcommand-helper-no-api-version-declared.json
@@ -3,7 +3,10 @@
   "schemaVersion": "1.1",
   "runOnRequirements": [
     {
-      "minServerVersion": "4.7"
+      "minServerVersion": "4.7",
+      "serverParameters": {
+        "requireApiVersion": false
+      }
     }
   ],
   "createEntities": [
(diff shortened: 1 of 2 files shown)


source/versioned-api/tests/runcommand-helper-no-api-version-declared.json
```diff
@@ -4,6 +4,11 @@
   "runOnRequirements": [
     {
       "minServerVersion": "4.7"
+    },
```
jyemin, line 7: The requirements are OR'd together.  In this case, we need an AND:

     {
       "minServerVersion": "4.7",
       "serverParameters": {
          "requireApiVersion": false
       }
     }

jyemin (changes requested):

source/versioned-api/tests/runcommand-helper-no-api-version-declared.json
alcaeus: Whoopsie - there are things I shouldn't do past 8 pm. Thanks for noticing, I've updated this.

jyemin (approved):

source/versioned-api/tests/runcommand-helper-no-api-version-declared.json
jyemin: Do you have an environment where you can execute these tests?  I didn't catch this by inspection, I had to run them, and even then it took me a minute to spot it.

PR #888, opened 2020-12-14
SPEC-1768 add a metadataClient for CSFLE

kevinAlbs:
Summary of changes:
- Create an internal client when necessary to use as the `metadataClient`, `keyVaultClient`, or both. If both require an internal client, they share one.

The rationale for adding an internal client is that it fixes the issue in the default case (`maxPoolSize=100` by default).

Caveats:
- The internal client does not expose configuration options.

[This branch of mongo-go-driver](https://github.com/mongodb/mongo-go-driver/compare/master...kevinAlbs:spec1768_poc?expand=1) contains a POC implementation and implementation of prose tests.

The command started expectations in spec tests will also be updated to remove `listCollections` commands to the key vault collection (which is a nice side-benefit of this change). I have those changes in a separate PR: https://github.com/mongodb/specifications/pull/892

### DIFF
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -336,7 +336,8 @@ it is set to an internal ``MongoClient``. If ``bypassAutomaticEncryption=true``,
 the option is ignored since ``listCollections`` is only run during automatic
 encryption.
 
-If neither are passed, they are set to the same internal ``MongoClient``.
+The ``keyVaultClient`` and ``metadataClient`` MUST be set to the same internal
+``MongoClient`` if neither are passed in and ``bypassAutomaticEncryption=false``.
 
 The internal ``MongoClient`` MUST be configured with the same options as the
 parent ``MongoClient`` with the ``AutoEncryptionOpts`` excluded.
@@ -350,6 +351,7 @@ The following pseudo-code describes the configuration behavior:
       internalClientOpts = copy(clientOpts)
       internalClientOpts.autoEncryptionOpts = None
       client.internalClient = MongoClient (internalClientOpts)
+      return client.internalClient
 
    def configureAutoEncryptionClients (client, clientOpts):
       if clientOpts.autoEncryptionOpts.keyVaultClient != None:
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -345,6 +345,7 @@ parent ``MongoClient`` with the ``AutoEncryptionOpts`` excluded.
 The following pseudo-code describes the configuration behavior:
 
 .. code::
+
    def getOrCreateInternalClient (client, clientOpts):
       if client.internalClient != None:
          return client.internalClient
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -249,7 +249,6 @@ MongoClient Changes
       schemaMap: Optional<Map<String, Document>>; // Maps namespace to a local schema
       bypassAutoEncryption: Optional<Boolean>; // Default false.
       extraOptions: Optional<Map<String, Value>>;
-      metadataClient: Optional<MongoClient>;
    }
 
 A MongoClient can be configured to automatically encrypt collection
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -294,16 +294,16 @@ metadataClient
 A client used to run ``listCollections`` to determine whether a collection has
 an associated JSON schema with encrypted fields specified.
 
-If a ``metadataClient`` is not set, and ``bypassAutomaticEncryption=false`` then
-an internal ``MongoClient`` is used with the same options as the parent ``MongoClient``
-excluding the ``AutoEncryptionOpts``.
+If a ``metadataClient`` is not passed in, and
+``bypassAutomaticEncryption=false``, the ``metadataClient`` is set to an
+internal ``MongoClient``. See `keyVaultClient, metadataClient, and the internal
+MongoClient`_ for configuration behavior.
 
 Drivers MUST document this behavior, using the following as a template:
 
-   If the ``MongoClient`` is configured with ``AutoEncryptionOpts`` with
-   ``bypassAutomaticEncryption=false`` then a ``metadataClient`` is necessary. If a
-   ``metadataClient`` is not passed as an option, a separate ``MongoClient`` is
-   used internally. It is configured using the options of the parent ``MongoClient``
+   If ``metadataClient`` is not passed as an option, and
+   ``bypassAutomaticEncryption=false``, a separate ``MongoClient`` is used
+   internally. It is configured using the options of the parent ``MongoClient``
    with the ``AutoEncryptionOpts`` omitted.
 
 See `What's the deal with metadataClient, keyVaultClient, and the internal client?`_
@@ -315,18 +315,56 @@ cluster as indicated by the connecting URI. But the optional
 keyVaultClient can be used to route data key queries to a separate
 MongoDB cluster.
 
-A ``keyVaultClient`` is necessary for both automatic encryption and decryption.
-If a ``keyVaultClient`` is not set then an internal ``MongoClient`` is used with
-the same options as the parent ``MongoClient`` excluding the ``AutoEncryptionOpts``.
+If a ``keyVaultClient`` is not passed, the ``keyVaultClient`` is set to an
+internal ``MongoClient``. See `keyVaultClient, metadataClient, and the internal
+MongoClient`_ for configuration behavior.
 
 Drivers MUST document this behavior, using the following as a template:
 
-   If the ``MongoClient`` is configured with ``AutoEncryptionOpts`` then a
-   ``keyVaultClient`` is necessary. If a ``keyVaultClient`` is not passed as an
-   option, a separate ``MongoClient`` is used internally. It is configured using the
-   options of the parent ``MongoClient`` with the ``AutoEncryptionOpts`` omitted.
+   If ``keyVaultClient`` is not passed as an option, a separate ``MongoClient``
+   is used internally. It is configured using the options of the parent
+   ``MongoClient`` with the ``AutoEncryptionOpts`` omitted.
 
 See `What's the deal with metadataClient, keyVaultClient, and the internal client?`_
+
+keyVaultClient, metadataClient, and the internal MongoClient
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+If a ``keyVaultClient`` is not passed, it is set to an internal ``MongoClient``.
+
+If a ``metadataClient`` is not passed, and ``bypassAutomaticEncryption=false``,
+it is set to an internal ``MongoClient``. If ``bypassAutomaticEncryption=true``,
+the option is ignored since ``listCollections`` is only run during automatic
+encryption.
+
+If neither are passed, they are set to the same internal ``MongoClient``.
+
+The internal ``MongoClient`` MUST be configured with the same options as the
+parent ``MongoClient`` with the ``AutoEncryptionOpts`` excluded.
+
+The following pseudo-code describes the configuration behavior:
+
+.. code::
+   def getOrCreateInternalClient (client, clientOpts):
+      if client.internalClient != None:
+         return client.internalClient
+      internalClientOpts = copy(clientOpts)
+      internalClientOpts.autoEncryptionOpts = None
+      client.internalClient = MongoClient (internalClientOpts)
+
+   def configureAutoEncryptionClients (client, clientOpts):
+      if clientOpts.autoEncryptionOpts.keyVaultClient != None:
+         client.keyVaultClient = clientOpts.autoEncryptionOpts.keyVaultClient
+      else:
+         client.keyVaultClient = getOrCreateInternalClient (client, clientOpts)
+
+      if clientOpts.autoEncryptionOpts.bypassAutomaticEncryption:
+         # metadataClient not needed. It is not an error if a user passed metadataClient.
+         return
+
+      if clientOpts.autoEncryptionOpts.metadataClient != None:
+         client.metadataClient = clientOpts.autoEncryptionOpts.metadataClient
+      else:
+         client.metadataClient = getOrCreateInternalClient (client, clientOpts)
 
 kmsProviders
 ^^^^^^^^^^^^
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -332,6 +332,11 @@ The following pseudo-code describes the configuration behavior for the three ``M
       else:
          client.metadataClient = getOrCreateInternalClient (client, clientOpts)
 
+Configuring the internal ``MongoClient`` MUST match the parent ``MongoClient``,
+except ``minPoolSize`` is set to ``0`` and ``AutoEncryptionOpts`` is omitted.
+This includes copying the options and host information from the URI, and other
+non-URI configuration (monitoring callbacks, versioned API, etc.).
+
 Drivers MUST document that an additional ``MongoClient`` may be created, using
 the following as a template:
 
@@ -343,8 +348,8 @@ the following as a template:
    - ``AutoEncryptionOpts.bypassAutomaticEncryption`` is ``false``.
 
    If an internal ``MongoClient`` is created, it is configured with the same
-   options as the parent ``MongoClient`` except ``minPoolSize=0`` and
-   ``AutoEncryptionOpts`` is omitted.
+   options as the parent ``MongoClient`` except ``minPoolSize`` is set to ``0``
+   and ``AutoEncryptionOpts`` is omitted.
 
 See `What's the deal with metadataClient, keyVaultClient, and the internal client?`_
 
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -1430,13 +1430,17 @@ options, instead of a client object.
 Can the metadataClient serve as the internal client when no keyVaultClient is set?
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Technically yes, but it adds complexity to the API without a clear benefit.
+No, it was decided against since it adds complexity to the API without a clear
+benefit.
 
 The ``metadataClient`` and ``keyVaultClient`` currently serve separate distinct
 purposes from the user's perspective.
 
 The ``keyVaultClient`` fetches keys from the key vault collection.
 The ``metadataClient`` runs ``listCollections`` to check for remote schemas.
+
+The behavior is that if a ``metadataClient`` is set, but a ``keyVaultClient``
+is not set, an internal client will be created to serve as the ``keyVaultClient``.
 
 Future work
 ===========
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -1371,48 +1371,30 @@ What's the deal with metadataClient, keyVaultClient, and the internal client?
 -----------------------------------------------------------------------------
 
 When automatically encrypting a command, the driver runs:
-- a ``listCollections`` command against to determine if the target collection
+- a ``listCollections`` command to determine if the target collection
 has a remote schema. This uses the ``metadataClient``.
-- a ``find`` against the ``keyVaultClient`` to fetch keys. This uses the
+- a ``find`` against the key vault collection to fetch keys. This uses the
 ``keyVaultClient``.
 
 **Why not reuse the parent MongoClient?**
-These operations MUST NOT reuse the same connection pool as the parent ``MongoClient`` configured with automatic encryption to avoid possible deadlock situations.
+
+These operations MUST NOT reuse the same connection pool as the parent
+``MongoClient`` configured with automatic encryption to avoid possible deadlock
+situations.
 
 Drivers supporting a connection pool (see `CMAP specification
 </source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst>`_)
 support an option for limiting the connection pool size: ``maxPoolSize``.
 
-Drivers require checking out a connection before serializing the command If the
-``listCollections`` command sent during automatic encryption uses the same
+Drivers need to check out a connection before serializing the command. If the
+``listCollections`` or ``find`` command during automatic encryption uses the same
 connection pool as the parent MongoClient, the application is susceptible to
-deadlocks. The relevant logic looks like:
+deadlocks.
 
-.. code::
-
-   server := client.SelectServer()
-   conn := server.Pool.CheckOut()
-   if client configured with automatic encryption {
-      // Begin the automatic encryption state machine.
-      // ...
-      switch (state) {
-         case need listCollections {
-            reply := client.Database(dbname).Command ({"listCollection": 1})
-            // ERROR, using the same client will check out another connection from the pool! This can cause a deadlock.
-         }
-         case need keys {
-            cursor := client.Database(dbname).Find (filter)
-            // ERROR, using the same client will check out another connection from the pool! This can cause a deadlock.
-         }
-      }
-      // ...
-   }
-   server.Pool.CheckIn (conn)
-
-A single operation checks out two connections from the same pool. If
-maxPoolSize=1, this is an immediate deadlock. If maxPoolSize=2, and two threads
-check out the first connection, they will deadlock attempting to check out the
-second.
+Using the same connection pool causes automatic encryption to check out multiple
+connections from the pool when processing a single command. If maxPoolSize=1,
+this is an immediate deadlock. If maxPoolSize=2, and two threads check out the
+first connection, they will deadlock attempting to check out the second.
 
 **Why are keyVaultClient and metadataClient separate exposed options?**
 
@@ -1424,8 +1406,8 @@ data-bearing cluster.
 
 ``listCollections`` responses are cached by libmongocrypt for one minute.
 
-The use pattern of the ``metadataClient`` will likely differ from the parent
-``MongoClient``
+The use pattern of the ``metadataClient`` will likely greatly differ from
+the parent ``MongoClient``.
 
 This is an exposed option to provide a way for users to configure (e.g. set a
 lower ``minPoolSize`` to avoid additional connections).
@@ -1511,7 +1493,7 @@ Changelog
 =========
 
 +------------+------------------------------------------------------------+
-| 2020-12-12 | Add metadataClient option and internal clients             |
+| 2020-12-12 | Add metadataClient option and internal client              |
 | 2020-10-19 | Add 'azure' and 'gcp' KMS providers                        |
 | 2019-10-11 | Add 'endpoint' to AWS masterkey                            |
 | 2019-12-17 | Clarified bypassAutoEncryption and managing mongocryptd    |
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -11,7 +11,7 @@ Client Side Encryption
 :Type: Standards
 :Minimum Server Version: 4.2
 :Last Modified: October 19, 2020
-:Version: 1.2.0
+:Version: 1.3.0
 
 .. contents::
 
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -295,13 +295,13 @@ A client used to run ``listCollections`` to determine whether a collection has
 an associated JSON schema with encrypted fields specified.
 
 If a ``metadataClient`` is not set, and ``bypassAutomaticEncryption=false`` then
-an internal ``MongoClient`` is used internally with the same options as the
-parent ``MongoClient`` excluding the ``AutoEncryptionOpts``.
+an internal ``MongoClient`` is used with the same options as the parent ``MongoClient``
+excluding the ``AutoEncryptionOpts``.
 
 Drivers MUST document this behavior, using the following as a template:
 
    If the ``MongoClient`` is configured with ``AutoEncryptionOpts`` with
-   ``bypassAutomaticEncryption=false``, a ``metadataClient`` is necessary. If a
+   ``bypassAutomaticEncryption=false`` then a ``metadataClient`` is necessary. If a
    ``metadataClient`` is not passed as an option, a separate ``MongoClient`` is
    used internally. It is configured using the options of the parent ``MongoClient``
    with the ``AutoEncryptionOpts`` omitted.
@@ -316,12 +316,12 @@ keyVaultClient can be used to route data key queries to a separate
 MongoDB cluster.
 
 A ``keyVaultClient`` is necessary for both automatic encryption and decryption.
-If a ``keyVaultClient`` is not set then an internal ``MongoClient`` is used the
-same options as the parent ``MongoClient`` excluding the ``AutoEncryptionOpts``.
+If a ``keyVaultClient`` is not set then an internal ``MongoClient`` is used with
+the same options as the parent ``MongoClient`` excluding the ``AutoEncryptionOpts``.
 
 Drivers MUST document this behavior, using the following as a template:
 
-   If the ``MongoClient`` is configured with ``AutoEncryptionOpts`` with a
+   If the ``MongoClient`` is configured with ``AutoEncryptionOpts`` then a
    ``keyVaultClient`` is necessary. If a ``keyVaultClient`` is not passed as an
    option, a separate ``MongoClient`` is used internally. It is configured using the
    options of the parent ``MongoClient`` with the ``AutoEncryptionOpts`` omitted.
@@ -1376,7 +1376,8 @@ has a remote schema. This uses the ``metadataClient``.
 - a ``find`` against the key vault collection to fetch keys. This uses the
 ``keyVaultClient``.
 
-**Why not reuse the parent MongoClient?**
+Why not reuse the parent MongoClient?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 These operations MUST NOT reuse the same connection pool as the parent
 ``MongoClient`` configured with automatic encryption to avoid possible deadlock
@@ -1396,7 +1397,8 @@ connections from the pool when processing a single command. If maxPoolSize=1,
 this is an immediate deadlock. If maxPoolSize=2, and two threads check out the
 first connection, they will deadlock attempting to check out the second.
 
-**Why are keyVaultClient and metadataClient separate exposed options?**
+Why are keyVaultClient and metadataClient separate exposed options?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 The ``keyVaultClient`` supports the use case where the key vault collection is
 stored on a MongoDB cluster separate from the data-bearing cluster.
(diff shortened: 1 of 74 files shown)


source/client-side-encryption/client-side-encryption.rst
```diff
@@ -286,13 +289,45 @@ Data keys are stored as documents in a special MongoDB collection. Data
 keys are protected with encryption by a KMS provider (AWS KMS, Azure key
 vault, GCP KMS, or a local master key).
 
+metadataClient
+^^^^^^^^^^^^^^
+A client used to run ``listCollections`` to determine whether a collection has
+an associated JSON schema with encrypted fields specified.
+
+If a ``metadataClient`` is not set, and ``bypassAutomaticEncryption=false`` then
+an internal ``MongoClient`` is used internally with the same options as the
```
jyemin, line 298: That's one too many "internals" in this sentence.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -286,13 +289,45 @@ Data keys are stored as documents in a special MongoDB collection. Data
...
+
 keyVaultClient
 ^^^^^^^^^^^^^^
 The key vault collection is assumed to reside on the same MongoDB
 cluster as indicated by the connecting URI. But the optional
 keyVaultClient can be used to route data key queries to a separate
 MongoDB cluster.
 
+A ``keyVaultClient`` is necessary for both automatic encryption and decryption.
+If a ``keyVaultClient`` is not set then an internal ``MongoClient`` is used the
```
jyemin, line 319: "is used with"

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -286,13 +289,45 @@ Data keys are stored as documents in a special MongoDB collection. Data
...
+same options as the parent ``MongoClient`` excluding the ``AutoEncryptionOpts``.
+
+Drivers MUST document this behavior, using the following as a template:
+
+   If the ``MongoClient`` is configured with ``AutoEncryptionOpts`` with a
```
jyemin, line 324: with -> then

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -1332,6 +1367,67 @@ on sockets. See the SDAM spec description of `cooldownMS <../source/server-disco
...
+What's the deal with metadataClient, keyVaultClient, and the internal client?
+-----------------------------------------------------------------------------
+
+When automatically encrypting a command, the driver runs:
+- a ``listCollections`` command to determine if the target collection
+has a remote schema. This uses the ``metadataClient``.
+- a ``find`` against the key vault collection to fetch keys. This uses the
+``keyVaultClient``.
+
+**Why not reuse the parent MongoClient?**
```
jyemin, line 1379: Funny character?

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -1332,6 +1367,67 @@ on sockets. See the SDAM spec description of `cooldownMS <../source/server-disco
...
+situations.
+
+Drivers supporting a connection pool (see `CMAP specification
+</source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst>`_)
+support an option for limiting the connection pool size: ``maxPoolSize``.
+
+Drivers need to check out a connection before serializing the command. If the
+``listCollections`` or ``find`` command during automatic encryption uses the same
+connection pool as the parent MongoClient, the application is susceptible to
+deadlocks.
```
jyemin, line 1392: It it, technically, a deadlock given that the checkout will eventually time out?

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -1332,6 +1367,67 @@ on sockets. See the SDAM spec description of `cooldownMS <../source/server-disco
...
+
+Using the same connection pool causes automatic encryption to check out multiple
+connections from the pool when processing a single command. If maxPoolSize=1,
+this is an immediate deadlock. If maxPoolSize=2, and two threads check out the
+first connection, they will deadlock attempting to check out the second.
+
+**Why are keyVaultClient and metadataClient separate exposed options?**
```
jyemin, line 1399: Funny character

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -1332,6 +1367,67 @@ on sockets. See the SDAM spec description of `cooldownMS <../source/server-disco
...
+Because automatic decryption does not require the JSON schema.
+``listCollections`` is not run during automatic encryption.
+
+**Can the keyVaultClient and the metadataClient be the same?**
+
+Technically yes, but not all drivers support this. In some drivers, the
+``keyVaultClient`` and ``metadataClient`` is configured by passing client
+options, instead of a client object.
+
+If a ``metadataClient`` is passed, but a ``keyVaultClient`` is set, it would be
```
jyemin, line 1426: I don't understand this paragraph.

jyemin (changes requested):

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: Doh, fixed.

source/client-side-encryption/client-side-encryption.rst

source/client-side-encryption/client-side-encryption.rst

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: It's a `?`. I'm not sure why it appears that way in the diff. Either way, I changed this to be a sub-heading instead of bold, so it can be linked to.

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: If there is a `waitQueueTimeoutMS` set, then it would be a timeout instead of a deadlock. The default behavior is a deadlock though, since `waitQueueTimeoutMS` defaults to "no limit": https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#connection-pool-options-1

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: Updated to a heading.

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: The point I want to make is that we could optimize out an extra client, but I'm proposing not to.

In particular, if you set a `metadataClient`, but do not set a `keyVaultClient`, we _could_ internally reuse the `metadataClient` to serve as the `keyVaultClient` (instead of creating an internal client for `keyVaultClient`).

It's a caveat I'm willing to hash out if there is disagreement. But I tried clarifying this paragraph.

jyemin (commented):
LGTM on the prose.

Are you going to add another commit with test changes?

kevinAlbs:
The prose tests are part of this PR. The updates to the spec tests are on-deck in this PR: https://github.com/mongodb/specifications/pull/892

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -286,13 +289,45 @@ Data keys are stored as documents in a special MongoDB collection. Data
 keys are protected with encryption by a KMS provider (AWS KMS, Azure key
 vault, GCP KMS, or a local master key).
 
+metadataClient
```
jyemin, line 292: This deserves a minor version bump.  Also, it's a big enough change that at least one other driver should implement before we push this to master

source/client-side-encryption/tests/README.rst
```diff
@@ -769,4 +769,172 @@ The following tests that setting ``bypassAutoEncryption=true`` really does bypas
...
+Running a test case
+```````````````````
+- Create a ``MongoClient`` named ``client_encrypted`` configured as follows:
+   - ``AutoEncrytionOpts.keyVaultNamespace="keyvault.datakeys"``
+   - ``AutoEncrytionOpts.kmsProviders``=``{ "local": { "key": <base64 decoding of LOCAL_MASTERKEY> } }``
+   - ``maxPoolSize=1``
+   - Capture command started events.
+   - Append ``TestCase.AutoEncryptionOpts`` (defined below)
+- If the testcase sets ``AutoEncryptionOpts.bypassAutoEncryption=true``:
+   - Use ``client_test`` to insert ``{ "_id": 0, "encrypted": <ciphertext> }``.
```
jyemin, line 818: Any reason not to just always do this?  Why the branch?

jyemin (commented):
Prose tests looks good, but I don't feel confident enough to LGTM it without actually implementing it.  I think we need to find a reviewer to do that.

source/client-side-encryption/tests/README.rst
kevinAlbs: Yes it's needed.

If `bypassAutoEncryption=true`, this should use a pre-encrypted ciphertext. Otherwise the insert won't encrypt and, the insert will fail JSON schema validation.

If `bypassAutoEncryption=false`, this should go through the auto encryption logic (which is the only place where `listCollections` is run.

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: Oh good point, I bumped the version to 1.3.0.

I agree, I'll hold off merging until another driver implements this. It was a small change in Go (mostly following suit with how the `keyVaultClient` is handled).

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -237,7 +237,9 @@ MongoClient Changes
       // Implementation details.
       private mongocrypt_t libmongocrypt_handle; // Handle to libmongocrypt.
       private Optional<MongoClient> mongocryptd_client; // Client to mongocryptd.
-      private Optional<MongoClient> keyvault_client; // Optional external client containing the key vault collection.
+      private MongoClient keyvault_client; // Client used to run find on the key vault collection. This is either an external MongoClient or internal_client.
```
saghm, line 240: If I had been a reviewer on the original spec, I would have asked that this be named `key_vault_client`, but since it predates this change, I'll accept it as it is

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -286,13 +289,45 @@ Data keys are stored as documents in a special MongoDB collection. Data
 keys are protected with encryption by a KMS provider (AWS KMS, Azure key
 vault, GCP KMS, or a local master key).
 
+metadataClient
+^^^^^^^^^^^^^^
+A client used to run ``listCollections`` to determine whether a collection has
+an associated JSON schema with encrypted fields specified.
+
+If a ``metadataClient`` is not set, and ``bypassAutomaticEncryption=false`` then
```
saghm, line 297: From reading this section several times, I'm still not sure I understand exactly what the configuration will be in every case:

1. metadataClient is passed in by user, bypassAutomaticEncryption=false
2. metadataClient is passed in by user, bypassAutomaticEncryption=true
3. metadataClient is not passed in by user, bypassAutomaticEncryption=false
4. metadataClient is not passed in by user, bypassAutomaticEncryption=true

As best as I can tell, the first two cases are the same, i.e. the metadataClient is used for running `listCollections`. In case 3, it sounds like the driver is supposed to create a new client interally and set `metadataClient` to that. I'm not sure about the fourth case; does `bypassAutomaticEncryption` mean that `listCollections` won't occur?

Either way, only one of the four cases is explicitly defined here (case 3), and the other cases are left implicit. One thing that's not clear is if the driver is supposed to actually populate `metadataClient` in the case that none is passed in or if it should just substitute `internal_client` at the places where `metadataClient` would normally be used. If the former, I think the language here should be updated to "If a `metadataClient` is not passed in..." (rather than "is not set"), and if the latter, think the docstring on `metadataClient` above should be revised to be clear that the value is not actually set to the value of internal_client.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -286,13 +289,45 @@ Data keys are stored as documents in a special MongoDB collection. Data
...
+an internal ``MongoClient`` is used with the same options as the parent ``MongoClient``
+excluding the ``AutoEncryptionOpts``.
+
+Drivers MUST document this behavior, using the following as a template:
+
+   If the ``MongoClient`` is configured with ``AutoEncryptionOpts`` with
+   ``bypassAutomaticEncryption=false`` then a ``metadataClient`` is necessary. If a
```
saghm, line 304: By "is necessary", does that mean drivers should throw an error if none is passed in? If so, I think that should be explicitly stated

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -286,13 +289,45 @@ Data keys are stored as documents in a special MongoDB collection. Data
...
 MongoDB cluster.
 
+A ``keyVaultClient`` is necessary for both automatic encryption and decryption.
+If a ``keyVaultClient`` is not set then an internal ``MongoClient`` is used with
+the same options as the parent ``MongoClient`` excluding the ``AutoEncryptionOpts``.
+
+Drivers MUST document this behavior, using the following as a template:
+
+   If the ``MongoClient`` is configured with ``AutoEncryptionOpts`` then a
+   ``keyVaultClient`` is necessary. If a ``keyVaultClient`` is not passed as an
```
saghm, line 325: Ditto above comment about "is necessary"

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: Great points, I think enumerating the cases is helpful. I described this behavior with pseudo-code in a separate section to show the configuration behavior with `metadataClient`, `keyVaultClient`, and the internal client. Hopefully that clarifies.

> does bypassAutomaticEncryption mean that listCollections won't occur?

Yes, `listCollections` is only run for automatic encryption.

> One thing that's not clear is if the driver is supposed to actually populate metadataClient in the case that none is passed in or if it should just substitute internal_client at the places where metadataClient would normally be used. If the former, I think the language here should be updated to "If a metadataClient is not passed in..." (rather than "is not set"), and if the latter, think the docstring on metadataClient above should be revised to be clear that the value is not actually set to the value of internal_client.

The former, I've updated the language.

saghm (approved):

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -286,13 +289,86 @@ Data keys are stored as documents in a special MongoDB collection. Data
 keys are protected with encryption by a KMS provider (AWS KMS, Azure key
 vault, GCP KMS, or a local master key).
 
+metadataClient
```
alcaeus, line 292: I get why we expose the `metadataClient` option, but I think this needs some additional language to guard against (very) invalid use-cases. Consider the following snippet:

```
metadataClient = new Client("mongodb://server1:27017", minPoolSize=2);
client = new Client("mongodb://server2:27017", maxPoolSize=1, metadataClient);
```

The fact that users would be able to direct the metadata client to a separate server concerns me. Depending on how feasible it is to guard against this in different drivers, we may want to consider exposing only a `metadataClientOptions` options that allows overriding options used for constructing the internal metadata client instead of allowing to specify a client object. This would allow users to do this:
```
client = new Client("mongodb://server2:27017", maxPoolSize=1, metadataClientOptions={minPoolSize=2, maxPoolSize=5})
```

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -286,13 +289,86 @@ Data keys are stored as documents in a special MongoDB collection. Data
...
+``MongoClient`` if neither are passed in and ``bypassAutomaticEncryption=false``.
+
+The internal ``MongoClient`` MUST be configured with the same options as the
+parent ``MongoClient`` with the ``AutoEncryptionOpts`` excluded.
+
+The following pseudo-code describes the configuration behavior:
+
+.. code::
+
+   def getOrCreateInternalClient (client, clientOpts):
```
alcaeus, line 349: Food for thought: from my understanding, the problem arises when specifying `maxPoolSize` in the original client options. If these are not passed, is there a reason for creating a separate client? Could we reuse the original client in those cases to save resources?

alcaeus (commented):
A couple of questions regarding the metadataClient option and a possible optimisation. I think solves the original problem at hand, but I agree with @jyemin that we may want a second implementation given the nature of the spec.

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: Yes, we could reuse the original client.

It might not be a less important use case to account for. `maxPoolSize` defaults to `100`. If a user explicitly opts into `maxPoolSize=0`, they may not be concerned with extra connections. OTOH, they may be configuring with a large `minPoolSize` as well.

I'll account for this if it does not add much more complexity.

kevinAlbs (commented):
As a heads up, @prashantmital, you may want to hold off POCing until I've updated to accept client options instead of a `MongoClient`.

source/client-side-encryption/client-side-encryption.rst
jyemin: This could be a big headache for Java and .NET drivers.  Both of them use a MongoClientSettings class that encompasses optional values and required values like credentials and hosts.  Would this require duplicating that whole structure into one that included everything but the hosts and credentials?

kevinAlbs: Yes I think so. Is that any harder than copying a `MongoClientSettings` and excluding `AutoEncyptionOpts`? That's already necessary for the current proposed change to configure the internal client.

jyemin: I think that's different.  In that case we're just copying the values from one MongoClientSettings into another, and excluding AutoEntrypOps.  in this case, do drivers have to provide a public API that excludes the possibility of configuring different hosts?  Or could the driver just ignore the configured hosts? Or am I misunderstanding this completely?

kevinAlbs: I see what you mean.

Ignoring the configured `hosts` and `credentials` seems like a weird API that ignores a core part of the `MongoClientSettings`.

We want to expose an option that:
1. Let's users completely configure the internal `MongoClient` (notably to configure the connection pool sizes, since `listCollections` / `find` on the key vault will likely be much less frequent then normal `MongoClient` operations)
2. Does not have ignored options
3. Does not permit misuse (via connecting to the wrong cluster)

If that requires non-trivial driver changes, I am in favor of not exposing any configuration, and instead choosing sensible defaults.

If we need an internal client, copy the `MongoClient` options. Exclude `AutoEncryptionOpts`. Set `minPoolSize` to 0.

It's one less knob. And if we determine that we need additional configuration later, we can add it.

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: This has been accounted for in 68374f8

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: 68374f8 proposes not exposing `metadataClient`.

kevinAlbs (commented):
Updated to hide the `metadataClient`. Tests have been updated and POCed.

source/client-side-encryption/tests/README.rst
```diff
@@ -769,4 +769,168 @@ The following tests that setting ``bypassAutoEncryption=true`` really does bypas
...
+
+.. code:: typescript
+
+   {"create": "coll", "validator": {"$jsonSchema": <json_schema>}}
+
+Create a ``ClientEncryption`` object, named ``client_encryption`` configured with:
+- ``keyVaultClient``=``client_test``
+- ``keyVaultNamespace``="keyvault.datakeys"
+- ``kmsProviders``=``{ "local": { "key": <base64 decoding of LOCAL_MASTERKEY> } }``
+- ``readConcern=majority`` and ``writeConcern=majority``
```
DmitryLukyanov, line 798: as far as I understand, these options cannot be applied to `ClientEncryption` itself

source/client-side-encryption/tests/README.rst
```diff
@@ -769,4 +769,168 @@ The following tests that setting ``bypassAutoEncryption=true`` really does bypas
...
+Running a test case
+```````````````````
+- Create a ``MongoClient`` named ``client_encrypted`` configured as follows:
+   - ``AutoEncrytionOpts.keyVaultNamespace="keyvault.datakeys"``
+   - ``AutoEncrytionOpts.kmsProviders``=``{ "local": { "key": <base64 decoding of LOCAL_MASTERKEY> } }``
+   - Capture command started events.
+   - Set ``maxPoolSize=TestCase.MaxPoolSize``
+   - Append ``TestCase.AutoEncryptionOpts`` (defined below)
+- If the testcase sets ``AutoEncryptionOpts.bypassAutoEncryption=true``:
+   - Use ``client_test`` to insert ``{ "_id": 0, "encrypted": <ciphertext> }``.
```
DmitryLukyanov, line 817: into `db.coll`?

source/client-side-encryption/tests/README.rst
```diff
@@ -769,4 +769,168 @@ The following tests that setting ``bypassAutoEncryption=true`` really does bypas
...
+- Otherwise:
+   - Use ``client_encrypted`` to insert ``{ "_id": 0, "encrypted": "string0" }``.
+- Run a ``findOne`` operation, with the filter ``{ "_id": 0 }``.
```
DmitryLukyanov, line 820: nit: on which client and collection? I guess it's based on the logic from lines 816-820, but maybe it can be represented somehow in wording?

source/client-side-encryption/tests/README.rst
```diff
@@ -769,4 +769,168 @@ The following tests that setting ``bypassAutoEncryption=true`` really does bypas
...
+Each test must assert the number of unique ``MongoClient``s created. This can be accomplished by capturing ``TopologyOpeningEvent``, or by checking command started events for a client identifier (not possible in all drivers).
+
+Running a test case
+```````````````````
+- Create a ``MongoClient`` named ``client_encrypted`` configured as follows:
+   - ``AutoEncrytionOpts.keyVaultNamespace="keyvault.datakeys"``
+   - ``AutoEncrytionOpts.kmsProviders``=``{ "local": { "key": <base64 decoding of LOCAL_MASTERKEY> } }``
+   - Capture command started events.
+   - Set ``maxPoolSize=TestCase.MaxPoolSize``
+   - Append ``TestCase.AutoEncryptionOpts`` (defined below)
```
DmitryLukyanov, line 815: Just to confirm that `Append` means simply initialization for `AutoEncrytionOpts`, so `AutoEncrytionOpts` configuration on each test case is not dependent on how these options were configured in previous cases

source/client-side-encryption/tests/README.rst
```diff
@@ -769,4 +769,168 @@ The following tests that setting ``bypassAutoEncryption=true`` really does bypas
...
+- ``kmsProviders``=``{ "local": { "key": <base64 decoding of LOCAL_MASTERKEY> } }``
+- ``readConcern=majority`` and ``writeConcern=majority``
+
+Use ``client_encryption`` to encrypt the value "string0" with ``algorithm``="AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic" and ``keyAltName``="local". Store the result in a variable named ``ciphertext``.
+
+Proceed to run the test case.
+
+Each test case configures a ``MongoClient`` with automatic encryption (named ``client_encrypted``).
+
+Each test must assert the number of unique ``MongoClient``s created. This can be accomplished by capturing ``TopologyOpeningEvent``, or by checking command started events for a client identifier (not possible in all drivers).
```
DmitryLukyanov, line 806: `TopologyOpeningEvent` do we have such event? I guess at least in c# it will be implemented with `ServerDescriptionChanges` + some additional conditions

DmitryLukyanov (commented):
Now, I'm working on tests, will provide the results when it will be ready

source/client-side-encryption/tests/README.rst
kevinAlbs: Ah yes, good point. I meant this only for the `MongoClient`. Removed.

source/client-side-encryption/tests/README.rst
kevinAlbs: Yes, for drivers that support SDAM monitoring: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst#events

I believe Java may do this by inspecting the `ConnectionID` fields of [CommandStarted events|https://github.com/mongodb/specifications/blob/master/source/command-monitoring/command-monitoring.rst#api] but that is particular to their implementation of `ConnectionID`.

source/client-side-encryption/tests/README.rst
kevinAlbs: Correct, I tried clarifying by combining the instructions and saying to set `AutoEncryptionOpts`.

source/client-side-encryption/tests/README.rst
kevinAlbs: Yes, added.

source/client-side-encryption/tests/README.rst
kevinAlbs: Good point, I made it explicit now.

source/client-side-encryption/tests/README.rst
ShaneHarvey: Pymongo implements this event: https://pymongo.readthedocs.io/en/stable/api/pymongo/monitoring.html?highlight=TopologyOpenedEvent#pymongo.monitoring.TopologyOpenedEvent

It's published when the MongoClient starts monitoring the cluster.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -293,6 +296,59 @@ cluster as indicated by the connecting URI. But the optional
...
+
+   def configureAutoEncryptionClients (client, clientOpts):
+      if clientOpts.autoEncryptionOpts.keyVaultClient != None:
+         client.keyVaultClient = clientOpts.autoEncryptionOpts.keyVaultClient
+      elif clientOpts.maxPoolSize == 0:
+         client.keyVaultClient = client
+      else:
+         client.keyVaultClient = getOrCreateInternalClient (client, clientOpts)
+
+      if clientOpts.autoEncryptionOpts.bypassAutomaticEncryption:
```
DmitryLukyanov, line 329: the new value from AutoEncryptionOpts is not used?

DmitryLukyanov:
my implementation: https://github.com/DmitryLukyanov/mongo-csharp-driver/tree/fle_poc 
1. The current c# implementation doesn't allow using maxPoolSize = 0, the supported values should be higher than 0. For the POC reasons I made these changes:
https://github.com/DmitryLukyanov/mongo-csharp-driver/commit/01c9f6eaa634616ebc948c82161d8710bf3bb74c#diff-2884a1d093312e76faa3b78739399556a6d8f7683520cf3987e50cecfb5191bfR61
https://github.com/DmitryLukyanov/mongo-csharp-driver/commit/01c9f6eaa634616ebc948c82161d8710bf3bb74c#diff-895481e7156a55dc4ea1a1f09cffa3b7f66437177e3d1eb1e9e9713bc857af24R71
https://github.com/DmitryLukyanov/mongo-csharp-driver/commit/01c9f6eaa634616ebc948c82161d8710bf3bb74c#diff-0537777d32b586e768ac1e74c10677b06f8933c41fba72c3f11a6f74862925e7R915
where instead of unlimited pool size, I used just a big enough value.
2. Previous prose tests passed
3. Some of json tests with updated description fail with a timeout exception (cannot select server), not sure about reasons, should be additionally investigated.
4. only 3 of the new test cases passed. The rest of them failed in this line: https://github.com/mongodb/specifications/pull/888/files#diff-c57f446f49351e6f7432a2d3028c880a728f1a33031510a7c376aed157d6de00R821 because `encrypted` field contains an encrypted binary value. See here https://github.com/DmitryLukyanov/mongo-csharp-driver/commit/01c9f6eaa634616ebc948c82161d8710bf3bb74c#diff-0537777d32b586e768ac1e74c10677b06f8933c41fba72c3f11a6f74862925e7R880
5. I'm still not sure about this asserting requirement: https://github.com/mongodb/specifications/pull/888/files#r547968309. Which clients should take part in this asserting?
it's definitely possible that I missed something, so let's discuss it when you will be ready.

kevinAlbs (commented):
Thanks for doing a POC! I am not sure about the timeout issue yet.

> where instead of unlimited pool size, I used just a big enough value.

If there is no way to make an unlimited `maxPoolSize` in the driver, then the rules around it do not apply. An internal `MongoClient` must always be created. I tried clarifying this, and changed the prose tests to note drivers that do not support `maxPoolSize=0` must skip those cases.

You can remove the `else if` in your PoC and remove the prose tests for `maxPoolSize=0`.

> Some of json tests with updated description fail with a timeout exception (cannot select server), not sure about reasons, should be additionally investigated.

I did not experience timeouts with the Go POC. Let's investigate offline.

> only 3 of the new test cases passed. The rest of them failed in this line: https://github.com/mongodb/specifications/pull/888/files#diff-c57f446f49351e6f7432a2d3028c880a728f1a33031510a7c376aed157d6de00R821 because encrypted field contains an encrypted binary value. See here DmitryLukyanov/mongo-csharp-driver@01c9f6e#diff-0537777d32b586e768ac1e74c10677b06f8933c41fba72c3f11a6f74862925e7R880

I noted in a comment why I think those are failing: https://github.com/DmitryLukyanov/mongo-csharp-driver/commit/01c9f6eaa634616ebc948c82161d8710bf3bb74c#r45614798

> I'm still not sure about this asserting requirement: https://github.com/mongodb/specifications/pull/888/files#r547968309. Which clients should take part in this asserting?
it's definitely possible that I missed something, so let's discuss it when you will be ready.

The event listener should only be registered on `client_encrypted`. Client configuration is copied onto the internal client. So events should fire from the internal client as well.

This may be the C# driver's equivalent to `TopologyOpeningEvent`: https://mongodb.github.io/mongo-csharp-driver/2.11/apidocs/html/T_MongoDB_Driver_Core_Events_ClusterOpeningEvent.htm

source/client-side-encryption/client-side-encryption.rst
DmitryLukyanov: I meant this value: https://github.com/mongodb/specifications/pull/888/files#diff-de01d398114f63480af23798e09ffca599aa758efd04f15a9beec89b4566041bR252

kevinAlbs: Ah right! That was leftover from a prior change and should not be in `AutoEncryptionOpts`. It is removed.

DmitryLukyanov (approved):
My implementation: https://github.com/DmitryLukyanov/mongo-csharp-driver/commits/fle_poc
All tests pass, but just to mention that with suggested changes, the FLE json tests are really slowed down. The whole json tests suite is launched during 20 mins (without these changes it took like 1 min). Since Kevin didn't see such problem in Go, mostlikely this is c# related issue and we will work on it in the scope of new c# ticket

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -293,6 +295,59 @@ cluster as indicated by the connecting URI. But the optional
...
+keyVaultClient, metadataClient, and the internal MongoClient
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+The following pseudo-code describes the configuration behavior for the three ``MongoClient``s:
+
+.. code::
+
+   def getOrCreateInternalClient (client, clientOpts):
+      if client.internalClient != None:
+         return client.internalClient
+      internalClientOpts = copy(clientOpts)
```
prashantmital, line 314: Does any spec standardize what the `clientOpts` are? Seems like this refers to URI options + other options that can be supplied to MongoClient?
When creating the new client, are we expected to use the host/port supplied to the original MongoClient or are expected instead to create a new topology for the seeds/nodes that are known at the time of internal MongoClient creation? AFAICT, those two could be different so having the spec be more explicit in this regard would be helpful.

kevinAlbs: Good question. AFAIK, no spec standardizes all client options. I added a section below to describe the intended behavior. That the internal client should use the same configuration supplied to the parent `MongoClient`.

I do not think it makes much of an observable difference whether we use the original host information, or if we use the seeds/nodes known at the time of constructing the internal `MongoClient`. But if the internal `MongoClient` is configured during construction, I think it may make implementations simpler to always use the original host info from the URI.

prashantmital: I just want the spec to specifically mention what do. I agree that since the internal client will be created at (almost) the same time as the parent MongoClient bearing the `AutoEncryptionOpts`, use of the original host/port vs the currently discovered nodes wouldn't necessarily matter.

prashantmital (approved):
Rough POC for Python lives here: https://github.com/mongodb/mongo-python-driver/pull/539

jyemin (approved):
LGTM

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -237,7 +237,9 @@ MongoClient Changes
       // Implementation details.
       private mongocrypt_t libmongocrypt_handle; // Handle to libmongocrypt.
       private Optional<MongoClient> mongocryptd_client; // Client to mongocryptd.
-      private Optional<MongoClient> keyvault_client; // Optional external client containing the key vault collection.
+      private MongoClient keyvault_client; // Client used to run find on the key vault collection. This is either an external MongoClient, the parent MongoClient, or internal_client.
+      private MongoClient metadata_client; // Client used to run listCollections. This is either the parent MongoClient or internal_client.
+      private Optional<MongoClient> internal_client; // An internal MongoClient. An internal MongoClient. Created if no external keyVaultClient was set, or if a metadataClient is needed
```
alcaeus, line 242: Nit: duplicated "An internal MongoClient."
```suggestion
      private Optional<MongoClient> internal_client; // An internal MongoClient. Created if no external keyVaultClient was set, or if a metadataClient is needed
```

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -1332,6 +1392,60 @@ on sockets. See the SDAM spec description of `cooldownMS <../source/server-disco
...
+
+The ``metadataClient`` is not an exposed option because a user could
+misconfigure it to point to another MongoDB cluster, which could be a 
+security risk.
+
+Why is the metadataClient not needed if bypassAutoEncryption=true
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Because automatic decryption does not require the JSON schema.
+``listCollections`` is not run during automatic encryption.
```
alcaeus, line 1447: I believe this paragraph is incorrect. `listCollections` is run during automatic encryption, hence it being unnecessary if `bypassAutoEncryption` is true. Suggested wording:

```suggestion
JSON schema data is only needed for automatic encryption but not for automatic
decryption. ``listCollections`` is not run when ``bypassAutoEncryption`` is
``true``, making a metadataClient unnecessary.
```

source/client-side-encryption/tests/README.rst
```diff
@@ -769,4 +769,174 @@ The following tests that setting ``bypassAutoEncryption=true`` really does bypas
...
 #. Use ``client_encrypted`` to insert the document ``{"unencrypted": "test"}`` into ``db.coll``. Expect this to succeed. 
 
-#. Validate that mongocryptd was not spawned. Create a MongoClient to localhost:27021 (or whatever was passed via ``--port``) with serverSelectionTimeoutMS=1000. Run an ``isMaster`` command and ensure it fails with a server selection timeout.
\ No newline at end of file
+#. Validate that mongocryptd was not spawned. Create a MongoClient to localhost:27021 (or whatever was passed via ``--port``) with serverSelectionTimeoutMS=1000. Run an ``isMaster`` command and ensure it fails with a server selection timeout.
+
+Deadlock tests
+~~~~~~~~~~~~~~
+
+There are multiple parameterized test cases. Before each test case, perform the setup.
```
alcaeus, line 777: It may be worth adding a note to explain that these tests don't apply for drivers that don't implement connection pools (e.g. PHP)

alcaeus (commented):
LGTM, except for the `bypassAutoEncryption` explanation.

source/client-side-encryption/tests/README.rst
kevinAlbs: Done.

alcaeus (approved):
LGTM!

PR #889, opened 2020-12-14
SPEC-1752 Fix benchmark deep_bson.json data file

saghm:
Previously, the file named `deep_bson.json` in the `extended_bson.tgz` tarball was actually just a copy of the tarball itself; it contained another directory called `extended_bson` with the files `flat_bson.json`, `full_bson.json`, and `deep_bson.json`, the last of which was _also_ a copy of the original tarball. I only checked two levels deep, but it looks like tarballs can infinitely recurse, which is both kind of cool and moderately horrifying.


kmahar (approved):
LGTM. agree the possibly infinite recursive tarball is terrifying...

kmahar:
@saghm, were you waiting on something to merge this?

saghm:
Oops, didn't even notice it was approved. I'll merge now!

PR #890, opened 2020-12-16
Add top level unsetOrMatches to insertMany operation

alcaeus:
The .NET driver does not return a result for the insertOne and insertMany methods, which is allowed in the CRUD spec (as all fields of the result are optional). To reflect this, we need to add a top-level `$$unsetOrMatches` to the expected result for such operations.


MikalaiMazurenka (approved):
LGTM

PR #891, opened 2020-12-16
Drop versioned API tests that test undefined behaviour

alcaeus:
As per the versioned API spec, declaring an API version on the client and specifying API versioning options in the command document is not support and leads to undefined behaviour. As such, the tests in question need to be removed, as they define a behaviour that drivers may not be able to adhere to.


benjirewis (approved):

divjotarora (approved):
I haven't implemented versioned API for Go, but the tests you removed definitely wouldn't have passed for us, so LGTM.

PR #892, opened 2020-12-16
SPEC-1768 Update CSFLE spec tests

kevinAlbs:
Remove command started events for listCollections to the key vault collection
since find commands to the key vault operations will by default use an internal
client that is not configured with automatic encryption.


kevinAlbs:
Closing in favor of moving test files in: https://github.com/mongodb/specifications/pull/888

PR #893, opened 2020-12-18
Skip deprecation errors test when requireApiVersion is true

benjirewis:
The `test-commands-deprecation-errors` test should skip when requireApiVersion is true since it uses `runCommand` without setting an `apiVersion` (this issue [was also discovered](https://github.com/mongodb/specifications/pull/887) for other `runCommand` tests).  When requireApiVersion is true for `test-commands-deprecation-errors`, commands are sent to the server for test setup that lack a `ServerAPIVersion` and the server errors incorrectly before the `runCommand` is sent.

Skips the test when `requireApiVersion` is true.

### DIFF
--- a/source/versioned-api/tests/test-commands-deprecation-errors.json
+++ b/source/versioned-api/tests/test-commands-deprecation-errors.json
@@ -6,7 +6,8 @@
       "minServerVersion": "4.7",
       "serverParameters": {
         "enableTestCommands": true,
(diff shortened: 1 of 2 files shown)


alcaeus (approved):

divjotarora (approved):

jyemin (approved):
LGTM

PR #895, opened 2020-12-23
SPEC-1770: Clarify how JSON schema is renamed for new minor versions

jmikola:
Note that the current system we have in the [Makefile](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/tests/Makefile) doesn't check the `schemaVersion` in the test files it processes. The Makefile itself also only refers to a single schema file. If ever there comes a time when we introduce a new major version, we'll need to revisit the logic for collecting which files are validated and associate them with different schemas.

As an aside, I noticed https://github.com/mongodb/specifications/commit/784857f548747ee6f894a5b6f7b6e5bb89103709#diff-cb6c7b8f4cd901e87aa6f787bf2f4dbb3f90e5aab2ad6165fef180731d1b9108R17 modified this spec's Makefile. My original intention was that each spec using the unified format would have its own Makefile to validate its own tests (rather than having references to other specs herein).

A spec using the unified format could then refer to the unified spec's schema; however, that could be problematic if the schema filename will change with each new minor version. We can address that by using a constant filename (e.g. `schema-1.x.json`), but then we should probably ensure the actual version is specified somewhere in the file so a developer doesn't have to consult the RST to infer the schema version of the JSON file (nor assume that they're in sync). [This Stack Overflow question](https://stackoverflow.com/questions/61077293/is-there-a-standard-for-specifying-a-version-for-json-schema) discusses how there isn't a single convention for doing so, but I think we could get by with a top-level field (e.g. `schemaVersion`), which would be ignored by a parser (JSON schema's own schema permits extra fields, unlike our own schema). I'd be hesitant to use [`$comment`](https://json-schema.org/understanding-json-schema/reference/generic.html#comments), since it "should not be used to communicate to users of the schema."


alcaeus (approved):
LGTM on these changes. We can revisit Makefile setup in a separate PR, I'm not opposed to having separate Makefiles per spec, but I'd prefer being able to trigger them from the main Makefile (e.g. by having `make test-all`, `make test-versioned-api`, etc. targets in a root Makefile which we currently don't have).

divjotarora (approved):
Changes here LGTM and I agree with @alcaeus's idea of addressing Makefile setup in a separate issue/PR.

jmikola:
SGTM. I'll make a separate ticket to track the idea for revising the Makefile, which can encompass renaming the schema file (e.g. `schema-1.x.json`), noting its exact version within the file, and creating a root Makefile target to validate unified tests files across all specs.

PR #897, opened 2021-01-13
DRIVERS-1504 Update versioned api transaction-handling spec test

benjirewis:
[DRIVERS-1504](https://jira.mongodb.org/browse/DRIVERS-1504)

Committing a transaction should not append server API options, but committing will move the transaction state from "in progress" to "committed" (see an FSA [here](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#clientsession-changes)). Any other commits directly following should not append server API options, but a driver may accidentally add them, as the transaction state is no longer "in progress".

Adds a new test case to the `transaction-handling` spec test for versioned api to make sure that committing a transaction more than once will not accidentally append server API options.

### DIFF
--- a/source/versioned-api/tests/transaction-handling.json
+++ b/source/versioned-api/tests/transaction-handling.json
@@ -227,6 +227,162 @@
           ]
         }
       ]
+    },
+    {
+      "description": "Committing a transaction twice does not append server API options",
+      "runOnRequirements": [
+        {
+          "topologies": [
+            "replicaset",
+            "sharded-replicaset"
+          ]
+        }
+      ],
+      "operations": [
+        {
+          "name": "startTransaction",
+          "object": "session"
+        },
+        {
+          "name": "insertOne",
+          "object": "collection",
+          "arguments": {
+            "session": "session",
+            "document": {
+              "_id": 6,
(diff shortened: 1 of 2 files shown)


source/versioned-api/tests/transaction-handling.json
```diff
@@ -227,6 +227,162 @@
           ]
         }
       ]
+    },
+    {
+      "description": "Committing a transaction twice does not append server API options",
```
benjirewis, line 232: Instead of adding a new test case, I could have just added another `commitTransaction` operation to the end of the previous test case. To be exhaustive/descriptive, I've added another case, but let me know if it's overkill.

ShaneHarvey (approved):
LGTM. The new test passes in Python.

source/versioned-api/tests/transaction-handling.json
alcaeus: I like it - this way the test name makes it clear that the duplicate `commit` command is not an oversight 👍

alcaeus (approved):

kevinAlbs (approved):
LGTM!

PR #898, opened 2021-01-15
DRIVERS-1505 Update versioned api database aggregate strict test

benjirewis:
[DRIVERS-1505](https://jira.mongodb.org/browse/DRIVERS-1505)

In the `crud-api-version-1-strict` spec test, a $listLocalSessions stage is used in an aggregate test case. That stage was [excluded](https://docs.google.com/document/d/10heZ18gZQw3x4MFPaTcY1Pwb0XCW7aSsEX5-cSe-nvs/edit#heading=h.wl7sg62rrat1) from api version 1, so the test fails with an API strict error.

Updates the test to expect this error.

### DIFF
--- a/source/versioned-api/tests/crud-api-version-1-strict.json
+++ b/source/versioned-api/tests/crud-api-version-1-strict.json
@@ -154,6 +154,11 @@
                 "$limit": 1
               }
             ]
+          },
+          "expectError": {
+            "isError": true,
+            "errorContains": "stage $listLocalSessions is not allowed with 'apiStrict: true' in API Version 1",
+            "errorCodeName": "APIStrictError"
           }
         }
       ],
--- a/source/versioned-api/tests/crud-api-version-1-strict.json
+++ b/source/versioned-api/tests/crud-api-version-1-strict.json
@@ -156,8 +156,6 @@
             ]
           },
           "expectError": {
-            "isError": true,
-            "errorContains": "stage $listLocalSessions is not allowed with 'apiStrict: true' in API Version 1",
             "errorCodeName": "APIStrictError"
           }
         }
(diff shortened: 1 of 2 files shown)


source/versioned-api/tests/crud-api-version-1-strict.json
```diff
@@ -154,6 +154,11 @@
                 "$limit": 1
               }
             ]
+          },
```
benjirewis, line 157: The name of this test is still "aggregate on database appends declared API version" which is inaccurate as it's also checking that the correct error is generated when an unsupported stage is used in a `db.aggregate()`. I used the same name as before because "distinct appends declared API version" does the same thing. Let me know if I should change the name of both tests.

ShaneHarvey: I think the name is still okay.

source/versioned-api/tests/crud-api-version-1-strict.yml
```diff
@@ -69,6 +69,10 @@ tests:
           pipeline: &pipeline
             - $listLocalSessions: {}
             - $limit: 1
+        expectError:
+          isError: true
+          errorContains: "stage $listLocalSessions is not allowed with 'apiStrict: true' in API Version 1"
```
ShaneHarvey, line 74: I think we should avoid `errorContains` here so that the test won't start failing when the server inevitably changes the error message for APIStrictError. Instead let's just include:
```
expectError:
  errorCodeName: APIStrictError

ShaneHarvey (changes requested):

source/versioned-api/tests/crud-api-version-1-strict.yml
benjirewis: Sounds good, I changed it to what you have above. I could also change the "distinct appends declared API version" test to avoid `errorContains`, too.

alcaeus (approved):

ShaneHarvey (approved):

PR #899, opened 2021-01-19
Document that the abortTransaction command accepts recoveryToken

ShaneHarvey:
Minor change to document that the abortTransaction command accepts recoveryToken. The spec already mandates this and has tests to verify the behavior. 

Motivation: In [SERVER-53811](https://jira.mongodb.org/browse/SERVER-53811), the server asked me to explain which command args drivers use for commit and abort. I was planning to share this section of the spec but was surprised to see recoveryToken was not present in the abortTransaction example.


durran (approved):
LGTM

PR #901, opened 2021-01-22
SPEC-1777 Add more spec tests for invalid relaxed UUIDs

benjirewis:
[SPEC-1777](https://jira.mongodb.org/browse/SPEC-1777)

[DRIVERS-848](https://jira.mongodb.org/browse/DRIVERS-848) adds support for parsing a new, relaxed $uuid extended JSON representation of binary subtype 4. This change requires drivers to validate passed in UUIDs to make sure they conform to RFC 4122 standards.

The current spec tests only have one invalid UUID case. The Go driver had to add more test coverage for invalid UUIDs, so it would probably be a good idea to add these to spec tests.

### DIFF
--- a/source/bson-corpus/tests/binary.json
+++ b/source/bson-corpus/tests/binary.json
@@ -94,8 +94,20 @@
             "string": "{\"x\" : { \"$uuid\" : { \"data\" : \"73ffd264-44b3-4c69-90e8-e7d1dfc035d4\"}}}"
         },
         {
-            "description": "$uuid invalid value",
+            "description": "$uuid invalid value--too short",
             "string": "{\"x\" : { \"$uuid\" : \"73ffd264-44b3-90e8-e7d1dfc035d4\"}}"
+        },
+        {
+            "description": "$uuid invalid value--too long",
+            "string": "{\"x\" : { \"$uuid\" : \"73ffd264-44b3-4c69-90e8-e7d1dfc035d4-789e4\"}}"
+        },
+        {
+            "description": "$uuid invalid value--misplaced hyphens",
+            "string": "{\"x\" : { \"$uuid\" : \"73ff-d26444b-34c6-990e8e-7d1dfc035d4\"}}"
+        },
+        {
+            "description": "$uuid invalid value--too many hyphens",
+            "string": "{\"x\" : { \"$uuid\" : \"----d264-44b3-4--9-90e8-e7d1dfc0----\"}}"
         }
     ]
 }


source/bson-corpus/tests/binary.json
```diff
@@ -94,8 +94,20 @@
...
+            "description": "$uuid invalid value--too short",
             "string": "{\"x\" : { \"$uuid\" : \"73ffd264-44b3-90e8-e7d1dfc035d4\"}}"
+        },
+        {
+            "description": "$uuid invalid value--too long",
+            "string": "{\"x\" : { \"$uuid\" : \"73ffd264-44b3-4c69-90e8-e7d1dfc035d4-789e4\"}}"
+        },
+        {
+            "description": "$uuid invalid value--misplaced hyphens",
+            "string": "{\"x\" : { \"$uuid\" : \"73ff-d26444b-34c6-990e8e-7d1dfc035d4\"}}"
```
prashantmital, line 106: While this is indeed an invalid UUID as it violates the RFC, Python's `uuid` implementation still correctly parses this UUID and does not cause an error. I suspect we will have to skip this test on Python as the alternative would be to implement our own UUID parser which wouldn't make much sense to do.
```
In [1]: import uuid

In [2]: uuid.UUID('73ff-d26444b-34c6-990e8e-7d1dfc035d4')
Out[2]: UUID('73ffd264-44b3-4c69-90e8-e7d1dfc035d4')
```

It is possible that other drivers which rely upon their respective standard library UUID implementations might run into the same issue and I just want to call that out here.

prashantmital (approved):

source/bson-corpus/tests/binary.json
benjirewis: Interesting... I'll look at native UUID parsers in other languages before I merge this. If there's a good amount of variation between parsers, then these tests might be moot.

alcaeus: Skipping these tests is fine if an implementation is more lenient, but these tests help show this difference in behaviour, especially if it results in segmentation faults.

@benjirewis maybe you could add a comment to the tests explaining that some libraries may consider these UUIDs as valid despite them not conforming to the RFC?

benjirewis: Discussed offline: can't add comments to the JSON tests unfortunately.

Did some more research; UUID parsing is pretty inconsistent across languages and even within languages (Go, for example, has many external UUID parsing libraries that all seem to have different adherences to RFC 4122). As @alcaeus  notes, it's probably a good idea to add these just to show difference in behavior. Wouldn't want an unexpected seg fault on a simple UUID 😮 .

alcaeus (approved):

PR #903, opened 2021-01-28
DRIVERS-828 Unified test format changes for Atlas planned maintenance testing


p-mongo:


source/unified-test-format/unified-test-format.rst
```diff
@@ -1798,6 +1885,92 @@ Use a ``listIndexes`` command to check whether the index exists. Note that it is
...
+  
+- When the termination request is received, the runner SHOULD skip
+  any non-loop operations not yet started and terminate as soon as practical.
+  
+- Termination request MUST cause a successful termination of the test
+  overall, i.e., receiving the termination request MUST NOT by itself be
+  considered an error.
+
+The exact mechanism by which the termination signal is delivered to the
+unified test runner, including the respective API, is left to the driver.
```
jmikola, line 1949: Related to my earlier suggestion for accessing the entity map, I could imagine using callbacks to control loop execution. I'm not sure if that's something you'd want to discuss here, or just leave this open-ended and up to each implementation to figure out.

source/unified-test-format/unified-test-format.rst
```diff
@@ -462,6 +477,78 @@ The structure of this object is as follows:
...
+          - ``address``: the address of the server to which the command
+             was sent, e.g. ``localhost:27017``.
+        
+        - For ``CommandSucceededEvent``:
+       
+          - ``name``: the name of the event, i.e. ``CommandSucceededEvent``.
+          - ``commandName``: the name of the command, e.g. ``insert``.
+          - ``duration``: the time, in (floating-point) seconds, it took for
+            the command to execute.
+          - ``startTime``: the (floating-point) number of seconds since
```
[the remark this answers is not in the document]
jmikola: Reading this again, I think the intention was for the workload executor to read this entity and create its output file. If the test runner is going to allow logging of CommandStartedEvent via `storeEventsAsEntities`, I would advocate for removing `startTime` here and adding `requestId` (and `operationId`) in its place. From there, the workload executor can calculate `startTime` for a succeeded/failed event on its own. That removes additional complexity a test runner might need to keep previous CommandStartedEvents in memory to report this field on their own.

source/unified-test-format/unified-test-format.rst
```diff
@@ -462,6 +477,78 @@ The structure of this object is as follows:
...
+          - ``duration``: the time, in (floating-point) seconds, it took for
+            the command to execute.
+          - ``failure``: this field MUST contain a textual description
+             of the error encountered while executing the command.
+          - ``startTime``: the (floating-point) number of seconds since
+            the Unix epoch when the command began executing.
+          - ``address``: the address of the server to which the command
+             was sent, e.g. ``localhost:27017``.
+             
+        - For connection pool events (``PoolCreatedEvent``,
```
jmikola, line 527: Adding this will also require that we discuss how to enable CMAP and revise the current advice not to mix different types of observed events. IIRC, I also suggested coordinating with someone from the CMAP spec (e.g. @patrickfreed) so the approach taken here will also work with that project migrating to the unified format (assuming that's still planned).

source/unified-test-format/unified-test-format.rst
```diff
@@ -1798,6 +1885,92 @@ Use a ``listIndexes`` command to check whether the index exists. Note that it is
 currently not possible to run ``listIndexes`` from within a transaction.
 
 
+loop
+~~~~
+
+The ``loop`` operation executes sub-operations in a loop until a termination
+signal is received by the unified test runner. It supports the following
+arguments:
```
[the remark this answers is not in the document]
p-mongo: I think this is a reasonable proposal but there currently isn't a need for (the existing scenarios seem to be sufficient) and it's not part of the scope. I can copy this comment into DRIVERS-828 to be referenced in the future when follow-up work is scoped.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1798,6 +1885,92 @@ Use a ``listIndexes`` command to check whether the index exists. Note that it is
...
+
+- ``operations``: the sub-operations to run on each loop iteration.
+  Each sub-operation must be a valid operation as described in this
+  specification.
+
+- ``storeErrorsAsEntity``: if specified, the runner MUST handle errors
+  arising during sub-operation execution and append a document with error
+  information to the array stored in the specified entity. If
+  ``storeFailuresAsEntity`` is specified, the runner MUST NOT include
+  failures in the errors. If ``storeFailuresAsEntity`` is not specified,
```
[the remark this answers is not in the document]
p-mongo: I added language referencing this allowance and permitting the test runner to use the same reporting categories as it would in the driver's test suite.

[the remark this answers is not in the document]
p-mongo: It is possible for the test scenario to route both errors and failures to the same entity, if the distinction is immaterial. However for drivers that do maintain the distinction I think it is worthwhile to keep errors and failures separate.

source/unified-test-format/unified-test-format.rst
p-mongo: This section has now been made more generic. I haven't made any prescriptions as far as the termination mechanism therefore I think it makes sense to not mention callbacks either.

source/unified-test-format/unified-test-format.rst
```diff
@@ -462,6 +477,78 @@ The structure of this object is as follows:
...
+        MUST serialize the events it receives into a document and append
+        the document to the array stored in the specified entity.
+        The fields in the document MUST be as follows:
+        
+        - For ``CommandStartedEvent``:
+       
+          - ``name``: the name of the event, i.e. ``CommandStartedEvent``.
+          - ``commandName``: the name of the command, e.g. ``insert``.
+          - ``startTime``: the (floating-point) number of seconds since
+            the Unix epoch when the command began executing.
```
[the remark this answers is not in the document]
p-mongo: I added language explicitly requiring the test runner to correlate the started/succeeded and started/failed events.

source/unified-test-format/unified-test-format.rst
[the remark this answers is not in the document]
jmikola: I requested that TPMs create a JIRA component for this project so we can log this as a ticket. Will resolve for now.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1798,6 +1897,119 @@ Use a ``listIndexes`` command to check whether the index exists. Note that it is
...
+The following termination behavior MUST be implemented by the unified test
+runner:
+
+- There MUST be a way to request termination of the loops. This request
+  will be made by the Atlas testing workload executor in response to
+  receiving the termination signal from Astrolabe.
+  
+- When the termination request is received, the workload executor MUST
+  stop looping. The current loop iteration SHOULD complete to its natural
+  conclusion (success or failure).
```
jmikola, line 1974: I don't think you mean to refer to "workload executor" here, since we're in the test runner. I also suggest rewording the second sentence to prefer MUST instead of SHOULD, unless you think some languages won't be able to accommodate this:

> When the termination request is received, the test runner MUST NOT continue to new iteration within the current loop. The sequence of operations in the current loop iteration MUST be allowed to continue executing to its natural conclusion (success or failure).

source/unified-test-format/unified-test-format.rst
```diff
@@ -1798,6 +1897,119 @@ Use a ``listIndexes`` command to check whether the index exists. Note that it is
...
+
+- After the termination request is received, the runner MUST NOT start
+  any further loops.
```
jmikola, line 1977: Can we replace this with two requirements for test file structure?

 * The list of operations for a test MUST NOT include multiple `loop` operations. Furthermore, `loop` operations MUST NOT be nested within other `loop` operations either directly or indirectly (e.g. `loop` within `withTransaction` within `loop`).

I don't expect to require the test runner to enforce this (folks could if they so choose) but it would provide guidance for how we write tests. We could always revise this later if a use case arises, but for now I think this will free us from having to worry about some edge cases.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1798,6 +1897,119 @@ Use a ``listIndexes`` command to check whether the index exists. Note that it is
...
+  
+- When the termination request is received, the runner SHOULD skip
+  any non-loop operations not yet started and terminate as soon as practical.
```
jmikola, line 1980: These last two bullets suggest that the signal isn't merely to stop the loop but rather the entire test. This seems at odds with the first bullet about "request(ing) termination of the loops" and allowing the current iteration to run its course. I propose we delete this.

I'll elaborate on my reasoning below (let me know if you disagree with anything):

 * This spec does not discuss how signals (i.e. requests for termination) are provided to the test runner. They could either be proactively sent from an external source (e.g. IPC) or periodically checked by the test runner (e.g. asking a callback if it's OK to continue).
 * Any request for termination is effectively a request to terminate the current loop being executed (if any). If the test runner is not presently in a loop, the request has no effect (and may very well be ignored entirely).
 * Acknowledging a request for termination should have no impact on future tests or loops. FWIW, I also suggest we prohibit multiple or nested loops in another comment.
 * A request for termination allows the current iteration (if any) to finish executing.

If we allow the current iteration to wrap up, I don't see any problem with allowing the test to finish executing the remainder of its `operations` array. In practice, I don't expect the workload executor will care one way of another, since its tests are likely to consist of a single `loop` operation. Thoughts?

source/unified-test-format/unified-test-format.rst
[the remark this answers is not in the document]
jmikola: As mentioned in https://github.com/mongodb/specifications/pull/903/files#r567703784, if we're not going to confine this logic to a monolithic special operation, I think it's preferable to offload the logic for correlating CommandStarted and CommandSucceeded/Failed events into whatever consumes the entity map (i.e. workload executor).

Could this section be simplified to just refer to the event objects as defined in their respective specs?

 * Connection monitoring event objects: https://github.com/mongodb/specifications/blob/master/source/command-monitoring/command-monitoring.rst#api
 * CMAP event objects: https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#events

On top of the specified fields, you can require the test runner to tack on `name` and `time` (or `observedAt`) fields to each record. We can also allow test runners to omit fields if they know they won't be needed by whatever consumes the entity. For instance, `command` and `reply` are huge and aren't relevant to the current design of the workload executor. However, `requestId` would be necessary for the workload executor to correlate command events.

If you're adamant about enumerating the fields and not merely referring to the other specs, that's fine.

source/unified-test-format/unified-test-format.rst
```diff
@@ -462,6 +477,90 @@ The structure of this object is as follows:
 
     Test files SHOULD NOT use this option unless one or more command monitoring
     events are specified in `observeEvents <entity_client_observeEvents_>`_.
+    
+  - ``storeEventsAsEntities``: Optional map of event names to entity names.
+    If provided the test runner MUST:
+    
+      - For each entity name, create the respective entity with a type of
```
jmikola, line 484: Can we omit the bullets here and just rely on paragraphs? I think that'd help readability. See `useMultipleMongoses` for an example.

source/unified-test-format/unified-test-format.rst
```diff
@@ -462,6 +477,90 @@ The structure of this object is as follows:
...
+        "event list". If multiple event names map to the same entity name,
+        exactly one entity MUST be created. If the entity already exists
+        (such as from a previous ``storeEventsAsEntities`` declaration from
+        another client), the test runner MUST raise an error.
+      - Set up an event subscriber for each event named. The event subscriber
+        MUST serialize the events it receives into a document and append
+        the document to the array stored in the specified entity.
+        The fields in the document MUST be as follows:
+        
+        - For ``CommandStartedEvent``:
```
jmikola, line 494: I'd suggest dropping this bullet as well. Bullets for the fields below make sense, though.

Optionally, you can add an earlier sentence or list that enumerates the supported event types. For example:

The following event names are supported:

 * From the [Command Monitoring](https://github.com/mongodb/specifications/blob/master/source/command-monitoring/command-monitoring.rst#api) spec: CommandStartedEvent, ...
 * From the [CMAP](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#events) spec: PoolCreatedEvent, ...

jmikola, line 494: I'd suggest dropping this bullet as well. Bullets for the fields below make sense, though.

Optionally, you can add an earlier sentence or list that enumerates the supported event types. For example:

The following event names are supported:

 * From the [Command Monitoring](https://github.com/mongodb/specifications/blob/master/source/command-monitoring/command-monitoring.rst#api) spec: CommandStartedEvent, ...
 * From the [CMAP](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#events) spec: PoolCreatedEvent, ...

source/unified-test-format/unified-test-format.rst
```diff
@@ -462,6 +477,90 @@ The structure of this object is as follows:
...
+       
+          - ``name``: the name of the event, i.e. ``CommandStartedEvent``.
+          - ``commandName``: the name of the command, e.g. ``insert``.
+          - ``startTime``: the (floating-point) number of seconds since
```
jmikola, line 498: Later, I suggest making this field consistent with whatever is used in the CMAP events (e.g. `time`, `observedAt`).

If we can offload the started and succeeded/failed correlation to the workload executor, then this can always be documented as the "time that the event was observed". Since the command monitoring spec doesn't actually report this field, the best we can do is calculate it in the listener.

source/unified-test-format/unified-test-format.rst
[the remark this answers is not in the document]
p-mongo: My goal primarily is to make whatever events are collected useful. Without correlating started and succeeded/failed events, I find either one individually to not particularly useful. Currently this correlation is primarily happening by hand by the person looking at event logs (as I understand it), but I think this manual correlation is less feasible for DAT due to the volume of events that could be produced.

Therefore, I think it would be valuable to have the events be pre-correlated by *something*.

As to what should perform the correlation, I think it should really be the driver when it publishes the events in the first place. But, since we are not changing the driver in the present project, the closest point would be the test runner (which is still in the driver's code base and hence the correlation code perhaps has a chance of being moved to the driver proper  down the road).

I am specifically having a hard time understanding why you propose the test runner performs filtering (which requires knowledge of events and a processing step) but not correlation which seems to be quite similar. Can you elaborate on why you don't want the test runner to correlate?

source/unified-test-format/unified-test-format.rst
```diff
@@ -462,6 +477,78 @@ The structure of this object is as follows:
...
+          ``ConnectionCheckOutStartedEvent``, ``ConnectionCheckOutFailedEvent``,
+          ``ConnectionCheckedOutEvent``, ``ConnectionCheckedInEvent``):
+               
+          - ``name``: the name of the event, e.g. ``ConnectionCreatedEvent``.
+          - ``time``: the (floating-point) number of seconds since the
+            Unix epoch when the event was published.
+          - ``address``: the address of the server that the command was
+            published for, e.g. ``localhost:27017``.
+          - ``connectionId``: the identifier for the connection for which
+            the event was published.
```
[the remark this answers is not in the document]
p-mongo: Looking at the history of command monitoring spec, references to connenctionId go back to 2015 when the spec was written hence it seems to me they aren't specifically referencing the same connection id that CMAP uses, necessarily.

I imagine it would be sensible for this connection id to be the same as the CMAP ones, but this doesn't seem to be strictly required.

source/unified-test-format/unified-test-format.rst
jmikola: Let's leave this as-is and just add "any further loops in either the current test or subsequent tests for the lifetime of the test runner".

source/unified-test-format/unified-test-format.rst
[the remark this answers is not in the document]
p-mongo: I created https://jira.mongodb.org/browse/DRIVERS-1548 for this.

source/unified-test-format/unified-test-format.rst
```diff
@@ -195,6 +195,21 @@ error if an attempt is made to store an entity with a name that already exists
 in the map and MUST raise an error if an entity is not found for a name or is
 found but has an unexpected type.
 
+Test runners MUST provide a mechanism to retrieve entities from the entity
+map. There MUST be a way to:
```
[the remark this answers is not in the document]
p-mongo: Per our discussion earlier today, I revised the language pertaining to entity retrieval to specify when the retrieval must be possible and to provide options as to how it might be implemented.

source/unified-test-format/unified-test-format.rst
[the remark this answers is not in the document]
p-mongo: Per our discussion, I removed command monitoring event storage, leaving only CMAP events, and changed the fields to be exactly the specified fields + name + observedAt.

source/unified-test-format/unified-test-format.rst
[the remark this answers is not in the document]
p-mongo: The command monitoring event requirements have now been removed.

source/unified-test-format/unified-test-format.rst
p-mongo: I updated the termination behavior to match our discussion earlier. The case of receiving termination request prior to looping is explicitly handled and the success upon receiving termination request is no longer mandated.

source/unified-test-format/unified-test-format.rst
p-mongo: Done.

p-mongo:
I added storeSuccessesAsEntity for the benefit of astrolabe validator and the requirement to stop executing loop sub-operations when one fails or errors.

source/unified-test-format/unified-test-format.rst
jmikola: To clarify, this PR does not need CMAP integrated into it in the same way we currently handle command monitoring (i.e. `observeEvents` and 'expectEvents`). @p-mongo introduced syntax to client entity creation to collect CMAP events and append them to a list/array in the entity map where they can be retrieved by some other code (e.g. workload executor, which invokes the test runner).

It's understandable if there's no need/benefit to port CMAP tests to the unified test runner. I haven't worked with them and didn't now they were possible unit tests.

Looking at the [SDAM integration tests README](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/tests/README.rst), I see some reference to CMAP events but it looks like they're mostly queried through special operations (not like `expectEvents`). If/when that spec is ported to the unified test runner, we can revisit this.

I'll resolve this thread. Thanks for the feedback.

source/unified-test-format/unified-test-format.rst
```diff
@@ -195,6 +195,22 @@ error if an attempt is made to store an entity with a name that already exists
 in the map and MUST raise an error if an entity is not found for a name or is
 found but has an unexpected type.
 
+Test runners MUST provide a mechanism to retrieve entities from the entity
+map prior to the clearing of the entity map, as discussed in
+`Executing a Test`_. There MUST be a way to:
+
+- Retrieve an entity by its name (for example, to support retrieving the
+  iteration count stored by the ``storeIterationsAsEntity`` option.
```
jmikola, line 203: Missing a closing paren.

source/unified-test-format/unified-test-format.rst
```diff
@@ -195,6 +195,22 @@ error if an attempt is made to store an entity with a name that already exists
...
+  
+- Retrieve all entities of a particular type as a mapping of each entity name
+  to the respective documents (for example, to support retrieving
+  the event documents stored by the ``storeEventsAsEntities`` option).
```
jmikola, line 207: In the case of `storeEventsAsEntities`, doesn't that create one or more list/array entities to which events will be appended? If so, wouldn't the previous bullet be sufficient?

Given the following example:

```
storeEventsAsEntities:
  events: [PoolCreatedEvent, ConnectionCreatedEvent]
```

The workload executor would just need to access the "events" entity to obtain the list of serialized event documents.

source/unified-test-format/unified-test-format.rst
```diff
@@ -195,6 +195,22 @@ error if an attempt is made to store an entity with a name that already exists
...
+
+Test runner MAY restrict retrieving entities to those which are plain
+documents and not driver objects (e.g., it MAY permit retrieving event
+documents generated by a client but not the client object itself, though
+both are entities).
```
jmikola, line 212: The iteration count set by `storeIterationsAsEntity` won't be a "plain document". I think this can be reworded to say:

> Test runners MAY restrict access to driver objects (e.g. MongoClient, ChangeStream) and only allow access to BSON types (see: [Supported Entity Types](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#supported-entity-types)). This restriction may be necessary if the test runner needs to ensure driver objects in its entity map are properly freed/destroyed between tests.

source/unified-test-format/unified-test-format.rst
```diff
@@ -462,6 +478,35 @@ The structure of this object is as follows:
 
     Test files SHOULD NOT use this option unless one or more command monitoring
     events are specified in `observeEvents <entity_client_observeEvents_>`_.
+    
+  - ``storeEventsAsEntities``: Optional map of entity names to an an array of
+    event names.
+    
+    For each entity name, the test runner MUST create the respective entity
+    with a type of "event list". If the entity already exists
```
jmikola, line 486: Since [Supported Entity Types](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#supported-entity-types) exhaustively lists what we're allowed to store in the entity map, I suggest you refer to "BSON array" here. The following paragraph should then make it clear that we'll be appending documents to this array.

I'm not sure if some folks might have read "event list" here and assumed they could use a thread-safe list object. I suppose there's nothing wrong with doing so as long as it's ultimately converted to a BSON array before we expose it to the workload executor.

Alternatively, if you want to allow flexibility here (e.g. let drivers use a thread-safe list and never convert it to a BSON array), I'd suggest adding "event list" as an explicitly supported type. There, you'd mention that drivers can implement it as they want and I'd suggest providing a thread-safe list or BSON array as possibilities.

source/unified-test-format/unified-test-format.rst
```diff
@@ -462,6 +478,35 @@ The structure of this object is as follows:
...
+    Additionally, the following fields MUST be stored with each
+    event document:
+    
+    - ``name``: The name of the event, such as ``PoolCreatedEvent``.
+      The name of the event MUST be the name used in the respective
+      specification that defines the event in question.
+    - ``observedAt``: The floating-point time since the Unix epoch
+      when the event was observed by the test runner.
+        
+    Currently, only CMAP events MUST be supported. Other events could be
```
jmikola, line 503: As a convenience to readers, I suggest linking "CMAP" to the spec (as we do elsewhere for other specs).

source/unified-test-format/unified-test-format.rst
```diff
@@ -1798,6 +1843,134 @@ Use a ``listIndexes`` command to check whether the index exists. Note that it is
...
+  Each sub-operation must be a valid operation as described in this
+  specification.
+  
+  If, in the course of executing sub-operations, a sub-operation produces
+  an error or a failure, the test runner MUST NOT execute subsequent
+  sub-operations in the same loop iteration. Depending on the presence of
+  ``storeErrorsAsEntity`` and ``storeFailuresAsEntity`` options, as
+  described below, either the iteration MUST complete and the next iteration
+  MUST be started, or the entire ``loop`` operation MUST complete with the
+  produced failure or error.
```
jmikola, line 1862: `storeFailureAsEntity` says:

> If (`storeFailureAsEntity` is) not specified, the runner MUST treat failures as errors, and either handle them following the logic described in ``storeErrorsAsEntity`` or cause them to terminate execution, if ``storeErrorsAsEntity`` is not specified.

Is that the only text that instructs the test runner to bubble up errors/failures and abort the loop entirely? I worry it might be overlooked if so.

Apart from my feelings on treating errors/failures the same, I think it'd be clearer to _not_ have 
`storeErrorsAsEntity` and `storeFailureAsEntity` influence loop behavior and instead control that via a dedicated `continueOnError` option. I'm OK to punt on this for now and address it with [DRIVERS-1550](https://jira.mongodb.org/browse/DRIVERS-1550) later in the POC cycle.

source/unified-test-format/unified-test-format.rst
```diff
@@ -2273,6 +2446,14 @@ match exactly; however, test runners MUST permit variations in document key
 order or otherwise normalize the documents before comparison. If the list of
 documents is empty, the test runner MUST assert that the collection is empty.
 
+The test runner MUST now permit the test's entity map to be accessed and
+the entities to be retrieved. The exact mechanism by which the entities are
+to be retrieved is not prescribed by this specification. As an example,
+the test runner MAY invoke an externally-provided callback with either the
+entity map as is or with only the documents. The test runner MAY also copy
+the document entities from the entity map into a data structure that can
+be later accessed to retrieve the entities.
```
jmikola, line 2455: Suggested text below losens the language about sharing the entity map itself (since you mention it could be another data structure). I also add a note to consider concurrency and more detail about this all happening at the end of the test before the map is cleared.

> Before clearing the entity map at the end of each test, the test runner MUST allow its entities to be accessed externally. The exact mechanism for facilitating this access is not prescribed by this specification, but drivers should be mindful of concurrency if applicable. As an example, the test runner MAY be configured with a callback method, which will be invoked at the end of each test and provided with the entity map (or an equivalent data structure). As previously discussed in Entity Map (link), test runners MAY restrict access to driver objects and only expose BSON types if necessary.

This assumes that event lists will be BSON types. If you end up creating a special "event list" type (or if you just happen to prefer conciseness), then I'd suggest rewriting the last sentence as:

> As previously discussed in Entity Map (link), test runners MAY restrict access to driver objects if necessary.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1798,6 +1843,134 @@ Use a ``listIndexes`` command to check whether the index exists. Note that it is
...
+- When the termination request is received, the test runner MUST
+  stop looping. If the test runner is looping when the termination request
+  is received, the current loop iteration MUST complete to its natural
+  conclusion (success or failure). If the test runner is not looping
+  when the termination request is received, it MUST NOT start any new
+  loop iterations in either the current test or subsequent tests for the
+  lifetime of the test runner; the test runner MAY skip any non-loop
+  operations and terminate as soon as is practical, or it MAY execute
+  any non-loop operations until either a loop is encountered or the
+  list of operations to execute is exhausted.
```
jmikola, line 1941: I don't feel strongly about this, but to anticipate feedback others may have: is there any reason _not_ to be prescriptive about whether the test runner is allowed to continue executing operations? If there's not a technical challenge, wouldn't it be preferable for all drivers to guarantee the same behavior?

The termination itself does not directly influence success/failure; however, allowing flexibility here could lead to the same test producing different results in different drivers. Consider a test with the following operations:

 * `loop` with one operation, which inserts a document into a collection
 * `aggregate` that counts documents in the collection and emits a single result document based on the number of documents previously inserted (e.g. `{ over9000: <boolean> }`). This operation has an `expectResult` assertion (e.g. `{ over9000: true }`).

Possible behavior:

 * If drivers terminate by only prohibiting future loop iterations but not subsequent non-loop operations, `aggregate` may pass or fail depending on how long the loop ran before terminating.
 * If drivers terminate by aborting the loop _and_ skipping subsequent operations in the same test, the test is likely to always succeed
 * Suppose the `expectResult` in the post-loop operation was written to always fail. The test could still pass depending on whether drivers decide to skip non-loop operations after a loop termination signal.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1798,6 +1843,134 @@ Use a ``listIndexes`` command to check whether the index exists. Note that it is
...
+  - ``time``: the number of (floating-point) seconds since the Unix epoch
+    when the error was encountered.
+
+- ``storeFailuresAsEntity``: if specified, the runner MUST handle failures
+  arising during sub-operation execution and append a document with failure
+  information to the array stored in the specified entity. If
+  not specified, the runner MUST treat failures as errors, and either
+  handle them following the logic described in ``storeErrorsAsEntity``
+  or cause them to terminate execution, if ``storeErrorsAsEntity`` is not
+  specified. The failure document MUST contain the following fields:
```
jmikola, line 1884: Based on my previous suggestion for `storeErrorsAsEntity`:

> if specified, the test runner MUST capture failures arising during sub-operation execution and append a document with its information to an array stored in the specified entity. If the entity does not exist, the test runner MUST initialize it as an empty array before appending the error document. If the entity does exist and is not an array, the test runner MUST raise an error.
>
> If this option is specified and `storeErrorsAsEntity` is not, errors must also be captured and appended to the array.
>
> Documents appended to the array MUST contain the following fields:

source/unified-test-format/unified-test-format.rst
```diff
@@ -1798,6 +1843,134 @@ Use a ``listIndexes`` command to check whether the index exists. Note that it is
...
+
+- ``storeErrorsAsEntity``: if specified, the runner MUST handle errors
+  arising during sub-operation execution and append a document with error
+  information to the array stored in the specified entity. If
+  ``storeFailuresAsEntity`` is specified, the runner MUST NOT include
+  failures in the errors, to the extent the test runner distinguishes
+  errors and failures as further described below.
+  If ``storeFailuresAsEntity`` is not specified,
+  the runner MUST include failures in the errors. The error document
+  MUST contain the following fields:
```
jmikola, line 1872: If we explicitly discuss loop behavior on error/failure in an earlier paragraph, I think we can simplify the text here.

> if specified, the test runner MUST capture errors arising during sub-operation execution and append a document with its information to an array stored in the specified entity. If the entity does not exist, the test runner MUST initialize it as an empty array before appending the error document. If the entity does exist and is not an array, the test runner MUST raise an error.
>
> If this option is specified and `storeFailuresAsEntity` is not, failures must also be captured and appended to the array.
>
> Documents appended to the array MUST contain the following fields:

I changed "handle" to "capture" since the current behavior is basically suppressing the error/failure from interrupting the greater test. Also added a note about checking the type of the entity before we try to append (akin to safeguards we have in `saveResultAsEntity`). I didn't bother to distinguish the possible error for a non-array type from the errors we catch here, as I think it's clear that's not a sub-operation error that might be caught.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1798,6 +1843,134 @@ Use a ``listIndexes`` command to check whether the index exists. Note that it is
...
+- ``storeIterationsAsEntity``: if specfied, the runner MUST keep track of
+  the number of iterations of the loop performed, and store that number
+  in the specified entity.
+
+A *failure* is when an operation performed by the test runner has a
+different result from what the test expected. For example, a test
+containing an expectation that a particular document has a particular value
+in a particular field, and the actual value being different, would
+be a failure. An *error* is any other type of error. For example,
+attempting to invoke an undefined operation would be an error.
```
jmikola, line 1905: Suggested rephrase:

> A *failure* is when the result or outcome of an operation executed by the test runner differs from its expected outcome. For example, an `expectResult` assertion failing to match a BSON document or an `expectError` assertion failing to match an error message would be considered a failure. An *error* is any other type of error raised by the test runner. For example, an unsupported operation or inability to resolve an entity name would be considered an error.

I originally wrote "failure to resolve an entity name" but didn't want to repeat "failure" there to avoid possible confusion.

source/unified-test-format/unified-test-format.rst
p-mongo: Fixed, thank you.

source/unified-test-format/unified-test-format.rst
p-mongo: This is true, the issue is that workload executor doesn't know that the entity is called `events`.

source/unified-test-format/unified-test-format.rst
p-mongo: Inserted your suggested language.

source/unified-test-format/unified-test-format.rst
jmikola: Understood, but I don't see why we need an API to fetch "all entities of type X" from the entity map to address this.

If the unified test exists within the `driverWorkload` key of the workload file (e.g. `retryReads-testFailover.yml`) you can standardize on the name there. I don't believe a YAML alias/reference can be used as a key, so you can't do:

```yml
storeEventsAsEntities:
  *eventListEntityName: [ PoolCreatedEvent, ConnectionCreatedEvent]]
```

However, you could have a separate option (possible sibling of `driverWorkload`) that the executor can read to infer the entity name to be fetched. For example:

```yml
cmapEventsEntity: events

...

driverWorkload:
  ...
    storeEventsAsEntities:
      events: [ PoolCreatedEvent, ConnectionCreatedEvent]
```

Alternatively:

 * the executor could decide to peek into the `driverWorkload` document itself and try to read `storeEventsAsEntities` for the client(s)
 * you can specify exactly what entity names `driverWorkload` configurations should use for capturing each type of CMAP event, and then executors can reliably assume where to look

That seems more foolproof than attempting to fetch, for example, all BSON arrays in the entity map. IIRC, drivers aren't obligated to define an "event list" entity type. Separately, if workload authors decide they'd like to split up pool and connection events into separate lists, I think the executor itself would have an easier time deciding to directly query for both lists explicitly instead of fetching all lists and then having to differentiate them from one another by inspecting the elements within.

source/unified-test-format/unified-test-format.rst
p-mongo: I added the event list entity as you suggested.

source/unified-test-format/unified-test-format.rst
p-mongo: Done.

source/unified-test-format/unified-test-format.rst
p-mongo: I think this is a fair point. I think for now I'll just mandate specific entity names in the DAT scenarios and in the future the flexible names could be implemented on the DAT side. I removed the requirement to retrieve the mapping and added an option to constrain retrieved entities with their expected type.

source/unified-test-format/unified-test-format.rst
p-mongo: I used your suggested language and noted in https://jira.mongodb.org/browse/DRIVERS-1550 the idea to use the continueOnError option.

source/unified-test-format/unified-test-format.rst
p-mongo: I used the second version of your suggested language to account for the event list type.

source/unified-test-format/unified-test-format.rst
p-mongo: This is a good point and I think there may be a use case for post-loop operations for cleanup. I changed this section to mandate running all non-loop operations with a soft prohibition on the test runner containing post-loop operations whose success depends on the number of loop iterations performed, what do you think?

source/unified-test-format/unified-test-format.rst
```diff
@@ -1798,6 +1843,134 @@ Use a ``listIndexes`` command to check whether the index exists. Note that it is
...
+  
+  - ``error``: the textual description of the failure encountered.
+  - ``time``: the number of (floating-point) seconds since the Unix epoch
+    when the failure was encountered.
+
+- ``storeSuccessesAsEntity``: if specfied, the runner MUST keep track of
+  the number of sub-operations that completed successfully, and store
+  that number in the specified entity. For example, if the loop contains
+  two sub-operations, and they complete successfully, each loop execution
+  would increment the number of successes by 2.
```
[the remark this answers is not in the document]
p-mongo: I thought the way I did it was the better one but I'll change to your wording.

source/unified-test-format/unified-test-format.rst
p-mongo: I updated the language with one change - I think it's important to verify the type of entity (to guard against name collisions) prior to executing loop operations and not when appending an error, to provide consistent diagnostics when the test reuses entity names inappropriately.

source/unified-test-format/unified-test-format.rst
p-mongo: Added thanks!

source/unified-test-format/unified-test-format.rst
```diff
@@ -195,6 +195,19 @@ error if an attempt is made to store an entity with a name that already exists
 in the map and MUST raise an error if an entity is not found for a name or is
 found but has an unexpected type.
 
+Test runners MUST provide a mechanism to retrieve entities from the entity
+map prior to the clearing of the entity map, as discussed in
+`Executing a Test`_. There MUST be a way to retrieve an entity by its name
+(for example, to support retrieving the iteration count stored by the
+``storeIterationsAsEntity`` option). Test runners MAY permit or require
+constraining the entity to be retrieved with its expected type.
```
jmikola, line 203: Having trouble parsing this sentence after "constraining". Trying to break this down: what does "require constraining the (retrieved) entity with its expected type" mean?

The paragraph that follows talks about allowing test runners to _not_ share certain driver objects in the entity map. For the remaining objects that _are_ shared, is there anything to "constrain"?

source/unified-test-format/unified-test-format.rst
```diff
@@ -462,6 +478,38 @@ The structure of this object is as follows:
 
     Test files SHOULD NOT use this option unless one or more command monitoring
     events are specified in `observeEvents <entity_client_observeEvents_>`_.
+    
+  - ``storeEventsAsEntities``: Optional map of entity names to an an array of
```
jmikola, line 482: "an an" typo

source/unified-test-format/unified-test-format.rst
```diff
@@ -462,6 +478,38 @@ The structure of this object is as follows:
...
+    - ``name``: The name of the event, such as ``PoolCreatedEvent``.
+      The name of the event MUST be the name used in the respective
+      specification that defines the event in question.
+    - ``observedAt``: The floating-point time since the Unix epoch
+      when the event was observed by the test runner.
+        
+    Currently, only
+    `CMAP <../connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst>`__
+    events MUST be supported. Other events could be
+    added to the list of supported events in the future.
```
jmikola, line 507: In the interest of future-proofing this spec and any test runner implementation against later changes to the CMAP spec, I think it's worth enumerating the supported event names here.

> Currently, only the following [CMAP] events MUST be supported:
>
> * PoolCreatedEvent
> * ...
>

I'd propose removing the second sentence, since we generally avoid talking about future possibilities _within_ the body of the spec. If you'd like to keep a note about this, a new section under [Future Work](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#future-work) may be more appropriate.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1798,6 +1846,159 @@ Use a ``listIndexes`` command to check whether the index exists. Note that it is
...
+            arguments:
+              filter: { _id: { $gt: 1 }}
+              sort: { _id: 1 }
+            expectResult:
+              -
+                _id: 2
+                x: 22
+              -
+                _id: 3
+                x: 33
```
jmikola, line 1999: nit: I think these can be condensed to a single line, like you did for `arguments`.

source/unified-test-format/unified-test-format.rst
jmikola: New text looks good and addresses my concern above. Resolving.

source/unified-test-format/unified-test-format.rst
p-mongo: What I meant by this was when retrieving the entity, the test runner may require to be told both the type and the name of the entity to be retrieved, rather than the name only. It's not strictly needed but I thought it could be a useful additional diagnostic.

source/unified-test-format/unified-test-format.rst
p-mongo: Fixed, thanks.

source/unified-test-format/unified-test-format.rst
p-mongo: OK enumerated the known CMAP events.

source/unified-test-format/unified-test-format.rst
p-mongo: I actually copy-pasted that from somewhere but I'll use a single line here as you are suggesting.

source/unified-test-format/unified-test-format.rst
jmikola: In PHP's case, I'd expect for my workload executor to fetch the entity and _then_ assert that it's the expected type before I attempt to work with it -- just in the habit of defensive programming. But this text doesn't make me any more or less inclined to do that.

I'll have to defer to stakeholder working with a statically typed languages to chime in on this. My thinking if that they might not have a choice in the matter.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1798,6 +1857,155 @@ Use a ``listIndexes`` command to check whether the index exists. Note that it is
...
+and some errors as failures. When the test runner stores errors and
+failures as entities it MAY classify conditions as errors and failures in
+the same way as it would when used in the driver's test suite.
+This includes reporting all errors as failures or all failures as errors.
+
+If the test runner does not distinguish errors and failures in its reporting,
+it MAY report both conditions under either category, but it MUST report
+any given condition in at most one category.
+
+The same entity name MAY be used for both ``storeErrorsAsEntity``
```
alcaeus, line 1954: I'm not sure about this. This requires special handling by a driver to not blindly overwrite an entity value, and it's an extra effort done to undo treating errors and failures differently. I'd vote for either treating errors and as separate categories and omitting this option for writing them to the same entity, or to treat failures and errors exactly the same, which removes the need for this special handling.

source/unified-test-format/unified-test-format.rst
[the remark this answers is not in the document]
jmikola: FYI: I opened [DRIVERS-1556](https://jira.mongodb.org/browse/DRIVERS-1556) to track this idea.

source/unified-test-format/unified-test-format.rst
```diff
@@ -462,6 +478,49 @@ The structure of this object is as follows:
...
+    
+    - ``name``: The name of the event, such as ``PoolCreatedEvent``.
+      The name of the event MUST be the name used in the respective
+      specification that defines the event in question.
+    - ``observedAt``: The floating-point time since the Unix epoch
+      when the event was observed by the test runner.
+        
+    Currently, only the following
+    `CMAP <../connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst>`__
+    events MUST be supported:
```
jmikola, line 506: @p-mongo: Am I misremembering, or did you recently say that command monitoring events _would_ need to be re-introduced to `storeEventsAsEntities`? If so, does this section need an update?

source/unified-test-format/unified-test-format.rst
```diff
@@ -195,6 +195,19 @@ error if an attempt is made to store an entity with a name that already exists
...
+map prior to the clearing of the entity map, as discussed in
+`Executing a Test`_. There MUST be a way to retrieve an entity by its name
+(for example, to support retrieving the iteration count stored by the
+``storeIterationsAsEntity`` option). Test runners MAY permit or require
+constraining the entity to be retrieved with its expected type.
+
+Test runners MAY restrict access to driver objects (e.g. MongoClient,
+ChangeStream) and only allow access to BSON types (see:
+`Supported Entity Types`_). This restriction may be necessary if the
+test runner needs to ensure driver objects in its entity map are properly
```
prashantmital, line 208: > This restriction may be necessary if the test runner needs to ensure driver objects in its entity map are properly freed/destroyed between tests.

This sentence seems to indicate that it should be possible to access to BSON types stored in the entity map after all the tests in the `tests` array have been run by the test runner? But, this section also states:

> To ensure each test is executed in isolation, test runners MUST NOT share entity maps between tests.

These two statements seem to be contradictory. I think access to the BSON data in entity map would be accessible while running a `test` (https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#test) but not necessarily after depending on the implementation. In Python, for instance, we would facilitate any access to BSON members of the entity map prior to performing post-test cleanup. 

I suggest removing this sentence entirely.

source/unified-test-format/unified-test-format.rst
```diff
@@ -225,6 +238,9 @@ Test runners MUST support the following types of entities:
 - ClientSession. See `entity_session`_ and `Session Operations`_.
 - GridFS Bucket. See `entity_bucket`_ and `Bucket Operations`_.
 - ChangeStream. See `ChangeStream Operations`_.
+- Event list. The event list MUST store BSON documents. The type of the list
```
prashantmital, line 241: Can you link to the `storeEventsAsEntities` option here so readers are not left wondering how they might come to see event lists stored in the entity map?

source/unified-test-format/unified-test-format.rst
```diff
@@ -225,6 +238,9 @@ Test runners MUST support the following types of entities:
...
+  itself is not prescribed by this specification. Test runner MAY use a
```
prashantmital, line 242: > The type of the list itself is not prescribed by this specification

Is this necessary to state? The spec is language-agnostic so I am not sure how much more prescriptive you _can_ be in terms of type beyond saying that this is a list.

source/unified-test-format/unified-test-format.rst
```diff
@@ -462,6 +478,49 @@ The structure of this object is as follows:
 
     Test files SHOULD NOT use this option unless one or more command monitoring
     events are specified in `observeEvents <entity_client_observeEvents_>`_.
+    
+  - ``storeEventsAsEntities``: Optional map of entity names to an array of
+    event names.
+    
+    For each entity name, the test runner MUST create the respective entity
```
prashantmital, line 485: AFAICT this field does not reuse/rely on information specified in the `observeEvents` and `ignoreCommandMonitoringEvents` however this is not clear on first reading this section. I suggest rephrasing this as:

For each entity name, the test runner MUST:

1. create the respective entity with a type of "event list", as described in `Supported Entity Types`_. If the entity already exists (such as from a previous ``storeEventsAsEntities`` declaration from another client), the test runner MUST raise an error.
2. set up an event subscriber for each event named. The event subscriber MUST serialize the events it receives into a document, using the documented properties of the event as field names, and append the document to the event list created in the previous step.
...

source/unified-test-format/unified-test-format.rst
```diff
@@ -462,6 +478,49 @@ The structure of this object is as follows:
...
+    with a type of "event list", as described in `Supported Entity Types`_.
+    If the entity already exists (such as from a previous
```
prashantmital, line 487: > If the entity already exists (such as from a previous ``storeEventsAsEntities`` declaration from another client),

This is already enforced by the spec. It is stated in the `Entity Map` section:
```
Test runners MUST raise an error if an attempt is made to store an entity with a name that already exists in the map
```

I think re-stating it here does not add any value.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1798,6 +1857,155 @@ Use a ``listIndexes`` command to check whether the index exists. Note that it is
...
 
+loop
+~~~~
+
+The ``loop`` operation executes sub-operations in a loop.
+It supports the following arguments:
+
+- ``operations``: the sub-operations to run on each loop iteration.
+  Each sub-operation must be a valid operation as described in this
+  specification.
```
prashantmital, line 1868: Can we link to https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#entity-test-operations here to reduce ambiguity?

source/unified-test-format/unified-test-format.rst
```diff
@@ -1798,6 +1857,155 @@ Use a ``listIndexes`` command to check whether the index exists. Note that it is
...
+specified, the errors and failures MUST be written in chronological order
+to the extent possible, interleaving the error and failure entries
+if necessary. Note that the textual description field for both errors
+and failures is ``error``.
+
+The following termination behavior MUST be implemented by the test
+runner:
+
+- The test runner MUST provide a way to request termination of loops. This
+  request will be made by the Atlas testing workload executor in response to
```
prashantmital, line 1966: Can we link to the Drivers Atlas Testing spec or documentation here so that it is clear what is meant by 'Atlas testing workload executor' since this is not something we mention in the `Terms` section .

prashantmital (changes requested):

source/unified-test-format/unified-test-format.rst
```diff
@@ -1798,6 +1857,155 @@ Use a ``listIndexes`` command to check whether the index exists. Note that it is
 currently not possible to run ``listIndexes`` from within a transaction.
 
 
+loop
+~~~~
+
+The ``loop`` operation executes sub-operations in a loop.
+It supports the following arguments:
+
+- ``operations``: the sub-operations to run on each loop iteration.
```
jmikola, line 1866: @p-mongo: Did we previously talk about advising tests not to nest `loop` operations? If so, I'd suggest adding:

> Tests SHOULD NOT include multiple loop operations (nested or sequential).

Note: this spec uses SHOULD for things it doesn't programmatically enforce. In this case, given how termination works, a test would either get stuck in the first or most-nested loop until a termination signal is received. And subsequent loops would be skipped. So there isn't a technical problem with having multiple loops per test. Rather, it just doesn't make sense.

I don't think we need to say "Test files SHOULD NOT" instead. If multiple tests in the file each use a single loop operation, that's fine, since tests might be skipped. Obviously, the runner wouldn't be able to execute both tests to completion, though.

source/unified-test-format/unified-test-format.rst
p-mongo: My understanding is you want the type verification to be done by the workload executor. The language here permits type verification to be done by the test runner which is how I implemented it.

source/unified-test-format/unified-test-format.rst
p-mongo: I removed this provision.

p-mongo: The changes to test format on DAT side mandate distinct entity names for errors and failures. The shared name was something I thought might be useful in the future, but since we have clarified why the errors and the failures must be separate, the use case for combining them into one entity doesn't appear to be strong anymore.

source/unified-test-format/unified-test-format.rst
p-mongo: I added command events back. Since command events are defined more flexibly, I also added a note requesting the drivers use the field names used in the specification even if their event objects have different fields, as well as mentioning type conversion to BSON types if event objects' fields are not of BSON types.

source/unified-test-format/unified-test-format.rst
p-mongo: It is possible for a test runner to expose the entity map after the test completes, for example in garbage-collected languages. In languages with explicit memory management this may be impossible. For this reason the time when entity map can be accessed is explicitly defined (it's right before it would be cleared at the end of each test).

A particular language implementation can copy the entity map at the required time and access the copy indefinitely from that point onward.

source/unified-test-format/unified-test-format.rst
p-mongo: Done.

source/unified-test-format/unified-test-format.rst
p-mongo: The following sentence explicitly defines two possible storage implementations - a simple array and an array that synchronizes access.

source/unified-test-format/unified-test-format.rst
p-mongo: I added a note stating that storeEventsAsEntities is independent from observeEvents and ignoreCommandMonitoringEvents, does that address your concern?

source/unified-test-format/unified-test-format.rst
p-mongo: To me the difference between event lists and, say, client entities is that the client entities are fully constructed when placed in the entity map but event lists are placeholders and are populated later. Given this distinction I think it's helpful to remind implementers that each operation requires its own entity.

source/unified-test-format/unified-test-format.rst
p-mongo: Added.

source/unified-test-format/unified-test-format.rst
p-mongo: Added.

source/unified-test-format/unified-test-format.rst
p-mongo: The language in DAT mandates exactly one loop but there aren't any constraints in unified test format. I added your suggested text.

source/unified-test-format/unified-test-format.rst
prashantmital: Since this is specified as a "MAY" clause, I am fine with keeping it as-is.

source/unified-test-format/unified-test-format.rst
prashantmital: Yes, that does it. Resolving.

source/unified-test-format/unified-test-format.rst
prashantmital: SGTM. Resolving.

prashantmital (approved):

alcaeus (approved):

source/unified-test-format/unified-test-format.rst
jmikola: I don't feel strongly about how other languages implement this. My opinion was just that this sentence wasn't necessary to grant Ruby permission to do this. [Entity Map](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#entity-map) already gives languages a lot of leeway:

> Test runners MAY choose to implement the entity map in a fashion most suited to their language, but implementations MUST enforce both uniqueness of entity names and referential integrity when fetching an entity.

If no one else is going to chime in on this, feel free to resolve this thread after reading and leave the text as-is.

jmikola (approved):

source/unified-test-format/unified-test-format.rst
p-mongo: I removed it.

source/unified-test-format/unified-test-format.rst
jmikola: AFAIK, this is the only type in the entity map that doesn't have a prescribed type. Everything else is either a specific driver object or BSON type. The entity map is one of the few places in the spec where we _do_ talk about types, because some languages were unable to implement it as a generic dictionary.

I don't think there's any harm in keeping this note, although see your point that it may be redundant in light of the sentence that follows.

PR #905, opened 2021-02-04
SPEC-1780 Change estimatedDocumentCount to use $collStats on greater than 4.9.0

benjirewis:
[SPEC-1780](https://jira.mongodb.org/browse/SPEC-1780)

Updates CRUD spec description of `estimatedDocumentCount` to specify that on wire versions greater than 4.9.0, the function must use an aggregate with `$collStats` instead of the `count` command. Adds two new tests that conform to the [unified test format](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst) to verify `estimatedDocumentCount` uses the correct behavior under the hood.

Updates CRUD test README to acknowledge new `crud/unified/` directory for future CRUD spec tests that conform to the unified test format.

Updates existing retryable reads and versioned api tests that test `estimatedDocumentCount` to have two versions: one for expected behavior before 4.9.0 (the `count` command), and one for expected behavior on 4.9+ (`aggregate` with `$collStats`).

An implementation in Go is available on my branch [here](https://github.com/benjirewis/mongo-go-driver/tree/estimatedDocCountCollStats).

cc: @kevinAlbs 

------

I'll keep an updated patch link [here](https://spruce.mongodb.com/version/6021975fc9ec4431c6a906cb/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) to show my implementation passing all new and updated tests.


source/retryable-reads/tests/estimatedDocumentCount-serverErrors.yml
```diff
@@ -1,9 +1,11 @@
 runOn:
     -
         minServerVersion: "4.0"
+        maxServerVersion: "4.9.0"
```
jmikola, line 4: Note, the [Test Format](https://github.com/mongodb/specifications/blob/master/source/retryable-reads/tests/README.rst#test-format) for Retryable Reads states that `maxServerVersion` is "inclusive". You should probably use something like "4.8.99" here if you want to exclude 4.9+.

source/retryable-reads/tests/estimatedDocumentCount-serverErrors.yml
```diff
@@ -1,9 +1,11 @@
...
         topology: ["single", "replicaset"]
     -
         minServerVersion: "4.1.7"
+        maxServerVersion: "4.9.0"
```
jmikola, line 8: Note: In contrast to my earlier comments, this is a justified case of using two runOn requirements since `minServerVersion` differs between topologies.

source/retryable-reads/tests/estimatedDocumentCount.yml
```diff
@@ -1,9 +1,11 @@
 runOn:
     -
         minServerVersion: "4.0"
+        maxServerVersion: "4.9.0"
```
jmikola, line 4: Same note about `maxServerVersion` being inclusive. This should probably be "4.8.99".

source/versioned-api/tests/crud-api-version-1.yml
```diff
@@ -209,7 +209,9 @@ tests:
                 key: x
                 <<: *expectedApiVersion
 
-  - description: "estimatedDocumentCount appends declared API version"
+  - description: "estimatedDocumentCount appends declared API version on less than 4.9.0"
+    runOnRequirements:
+      - maxServerVersion: "4.9.0"
```
jmikola, line 214: This is also inclusive and should probably be "4.8.99". See: https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#runonrequirement

source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -0,0 +1,63 @@
+description: "estimatedDocumentCount"
+
+schemaVersion: "1.1"
```
jmikola, line 3: Does this require anything in schema version 1.1? If not, please use 1.0.

Quoting [Top-level Fields: schemaVersion](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#top-level-fields):

source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -0,0 +1,63 @@
...
+                  - $group: {
+                    _id: { $const: total },
+                    count: { $sum: $count },
+                  }
+              commandName: aggregate
+              databaseName: *database0Name
+
+  - description: "estimatedDocumentCount uses count on less than 4.9.0"
+    runOnRequirements:
+      - maxServerVersion: "4.9.0"
```
jmikola, line 51: `maxServerVersion` is inclusive, so this should be "4.8.99".

source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -0,0 +1,63 @@
...
+
+initialData:
+  - collectionName: *collection0Name
+    databaseName: *database0Name
+    documents:
+      - { _id: 1, x: 11 }
+      - { _id: 2, x: 22 }
+      - { _id: 3, x: 33 }
+
+tests:
```
jmikola, line 26: I suggest adding two tests for the maxTimeMS option (pre-4.9 and 4.9+).

source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -0,0 +1,63 @@
...
+  - description: "estimatedDocumentCount uses $collStats on 4.9.0 or greater"
+    runOnRequirements:
+      - minServerVersion: "4.9.0"
+    operations:
+      - name: estimatedDocumentCount
+        object: *collection0
+        expectResult: { $numberLong: 3 }
```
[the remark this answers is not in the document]
jmikola: Created: [DRIVERS-1553](https://jira.mongodb.org/browse/DRIVERS-1553).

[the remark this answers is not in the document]
benjirewis: Ah good to get context on that. Changed to just `3`; and thanks for filing ticket.

source/crud/tests/unified/estimatedDocumentCount.yml
benjirewis: Great point, changed all `maxServerVersion=4.9.0`s to `maxServerVersion=4.8.99`s.

source/retryable-reads/tests/estimatedDocumentCount-4.9.yml
```diff
@@ -0,0 +1,68 @@
+runOn:
+    -
+        minServerVersion: "4.9.0"
+        topology: ["single", "replicaset"]
+    -
+        minServerVersion: "4.9.0"
+        topology: ["sharded"]
```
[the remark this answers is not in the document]
benjirewis: Sounds good!

source/crud/crud.rst
```diff
@@ -661,9 +661,27 @@ older sharded clusters.
 estimatedDocumentCount
 ~~~~~~~~~~~~~~~~~~~~~~
 
-The estimatedDocumentCount function is implemented using the `count` command
-with no query filter, skip, limit, or other options that would alter the
-results. As documented above, the only supported option is maxTimeMS.
+On server versions greater than or equal to 4.9.0 (wire version 12 or higher),
+the estimatedDocumentCount function is implemented using the ``$collStats``
+aggregate pipeline stage with ``$group``. As documented above, the only supported
+option is maxTimeMS::
```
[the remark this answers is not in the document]
jmikola: > I wrote 5.0, but why wouldn't it be 4.9?

I suggested that for consistency with existing entries, since the table only refers to stable server releases.

I realize that differs from how we refer to pre-release versions in `minServerVersion` test requirements (e.g. "4.1.8" for some sharded transaction tests).

source/retryable-reads/tests/estimatedDocumentCount-4.9.yml
```diff
@@ -0,0 +1,64 @@
+runOn:
+    -
+        minServerVersion: "4.9.0"
```
jmikola, line 3: You can make this a one-liner as you did in the unified CRUD test:

```
- minServerVersion: "4.9.0"
```

Likewise for other instances where it spans multiple lines.

jmikola (approved):
LGTM with YAML formatting change.

source/crud/crud.rst
```diff
@@ -661,9 +661,23 @@ older sharded clusters.
 estimatedDocumentCount
 ~~~~~~~~~~~~~~~~~~~~~~
 
-The estimatedDocumentCount function is implemented using the `count` command
-with no query filter, skip, limit, or other options that would alter the
-results. As documented above, the only supported option is maxTimeMS.
+On server versions greater than or equal to 4.9.0 (wire version 12 or higher),
```
jyemin, line 664: I tried this out against a non-existent collection, and got different results with aggregate and count:

```
MongoDB Enterprise > db.foo.count()
0
```

```
MongoDB Enterprise > pipeline = [
...     { '$collStats': { 'count': {} } },
...     { '$group': { '_id': 1, 'n': { '$sum': '$count' } } }
...   ]
[
	{
		"$collStats" : {
			"count" : {

			}
		}
	},
	{
		"$group" : {
			"_id" : 1,
			"n" : {
				"$sum" : "$count"
			}
		}
	}
]
MongoDB Enterprise > db.foo.aggregate(pipeline)
uncaught exception: Error: command failed: {
	"ok" : 0,
	"errmsg" : "PlanExecutor error during aggregation :: caused by :: Unable to retrieve count in $collStats stage: Collection [test.foo] not found.",
	"code" : 40481,
	"codeName" : "Location40481"
} : aggregate failed :
```

I think one way or another this behavioral difference needs to be hidden from applications.  Our options are:

1. Request a server change in the $collStats stage to not error for non-existent collections, at least when count is the only thing needed
2. Handle the specific error in drivers and swallow it, returning a count of 0.

If we choose the second option, we should request a stable error code with a label for this error.  Note that drivers already check for error code 26 in order to ignore failures to drop non-existent collections or indexes.

jmikola: @jyemin What server version were you testing on? My understanding is that [SEVER-35479](https://jira.mongodb.org/browse/SERVER-35479) was supposed to address this and return an empty result.

@benjirewis: Either way, I think we'd benefit from adding another test that uses a non-existent collection. You should be able to do that by creating a separate collection entity with a new namespace and _not_ specifying that namespace in `initialData`.

jyemin: I'm testing on 4.9.0-alpha-696-gb4de2e9

benjirewis: Discussed offline. [SERVER-35479](https://jira.mongodb.org/browse/SERVER-35479) was resolved by making sure that a consistent `NamespaceNotFound` error was returned in the event of an aggregate with $collStats being run on a non-existent namespace. Since there appears to be precedent, drivers should swallow this specific error when running a `count` aggregate with $collStats, and return a count of 0.

Added two new spec tests to ensure a count of `0` with no error on 4.9+ and < 4.9. Added a line in the spec about swallowing the error.

source/retryable-reads/tests/estimatedDocumentCount-4.9.yml
benjirewis: Done!

isabelatkinson (approved):
retryability stuff lgtm!

source/crud/tests/unified/estimatedDocumentCount.json
```diff
@@ -0,0 +1,302 @@
...
+                    "client": "client0",
+                    "events": [
+                        {
+                            "commandStartedEvent": {
+                                "command": {
+                                    "aggregate": "coll0",
+                                    "pipeline": [
+                                        {
+                                            "$collStats": {
+                                                "count": {
```
jyemin, line 81: The spec says this should be:

```
"count": {}
```

source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -0,0 +1,149 @@
+description: "estimatedDocumentCount"
```
jyemin, line 1: I think we should add a test that asserts that any server error that is NOT NamespaceNotFound is correctly reported (I had a bug in my implementation of this that wouldn't have been caught by any of these tests).  We can do this with a failpoint fairly easily I think.

jyemin (changes requested):

source/crud/tests/unified/estimatedDocumentCount.json
benjirewis: You're right... 🤦  Changed to `"count": {}` in all occurrences.

source/crud/tests/unified/estimatedDocumentCount.yml
benjirewis: Sounds good! Added two new tests, one for 4.9+ and < 4.9.

source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -0,0 +1,213 @@
...
+
+schemaVersion: "1.0"
+
+createEntities:
+  - client:
+      id: &client0 client0
+      uriOptions: { retryReads: false } # needed to set failPoint
```
benjirewis, line 8: Needed to set `retryReads` to `false` on the client so that the failpoints I set for `aggregate` and `count` aren't retried (as they would be by default). I could alternatively set a failpoint with `times: 2` and expect two attempts, but this seems easier.

source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -0,0 +1,213 @@
...
+              databaseName: *database0Name
+
+  - description: "estimatedDocumentCount errors correctly on less than 4.9.0"
+    runOnRequirements:
+      - minServerVersion: "4.0.0"
+        maxServerVersion: "4.8.99"
+        topologies: [ single, replicaset ]
+      - minServerVersion: "4.2.0"
+        maxServerVersion: "4.8.99"
+        topologies: [ sharded ]
```
benjirewis, line 191: `failCommand` isn't available until 4.0 for replset/standalone, and 4.2 for sharded, so I have to set `minServerVersion`s for this test.

source/crud/tests/unified/estimatedDocumentCount.yml
jyemin: Can you add one more test: one where the failpoint doesn't force the connection to close, but rather replies with an error code that is not NamespaceNotFound?  

Rationale is that some drivers will be catching CommandException explicitly, and then ignoring NamespaceNotFound.  With the error that just closes the connection, that code is skipped entirely since a SocketException is what will be thrown instead.

jyemin (changes requested):
Sorry, requesting one additional test.  Almost there!

source/crud/tests/unified/estimatedDocumentCount.yml
benjirewis: Good point! Added. Now there are four tests for correct error handling: socket error in 4.9+ and < 4.9, and command error in 4.9+ and < 4.9.

jyemin: Thanks for adding these.  All passing in Java driver.

jyemin (approved):

source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -0,0 +1,276 @@
...
+      - name: failPoint
+        object: testRunner
+        arguments:
+          client: *client0
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+              failCommands: [ aggregate ]
+              errorCode: 11000
```
jmikola, line 114: When specifying an `errorCode` in a test file, please add a comment with the readable name. Quoting [expectedError](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#expectederror):

> YAML files SHOULD use a comment to note the corresponding code name (e.g. `errorCode: 26 # NamespaceNotFound`).

That was written with `expectError` in mind but I think it's good practice everywhere, especially if the name isn't already in the test description.

I realize it's only important that the error code here is not retryable, but since 11000 is DuplicateKey (write related) I think it'd be clearer to use something more generic. Looking through [error_codes.yml](https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.yml), InternalError(1) or UnknownError(8) seem preferable (neither is retryable).

source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -0,0 +1,276 @@
...
+      - name: estimatedDocumentCount
+        object: *collection0
+        expectError: { isError: true }
```
jmikola, line 117: I propose replacing with an error code assertion. For example:

```yaml
expectError:
  errorCode: 8 # UnknownError
```

source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -0,0 +1,276 @@
...
+          client: *client0
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+              failCommands: [ aggregate ]
+              closeConnection: true
+      - name: estimatedDocumentCount
+        object: *collection0
+        expectError: { isError: true }
```
jmikola, line 149: Since we're expecting a network error, I propose using a strict assertion:

```yaml
expectError:
  isClientError: true
```

Existing unified spec tests only used this assertion for argument validation errors, but [expectError](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#expectederror) states:

> Client errors include, but are not limited to: parameter validation errors before a command is sent to the server; network errors.

This would be a good opportunity to test that unified test runner implementations actually consider network errors to be client errors (i.e. it is not derived from a server response).

source/crud/tests/unified/estimatedDocumentCount.yml
jmikola: Were the network error tests particularly relevant? Reading https://github.com/mongodb/specifications/pull/905#discussion_r572211505, it looks like @jyemin was more concerned with testing a _server error_ other than NamespaceNotFound(26).

That said, I don't mind if we leave the network error tests in place, since it will give us an opportunity to use the `isClientError: true` assertion.

I would suggest revising the comment to be more clear/accurate, since this isn't relevant to all `failPoint` operations. For example:

```
# Avoid retrying fail points with closeConnection
```

source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -0,0 +1,276 @@
...
+      client: *client0
+      databaseName: &database0Name edc-tests
+  - collection:
+      id: &collection0 collection0
+      database: *database0
+      collectionName: &collection0Name coll0
+  - collection:
+      id: &collection1 collection1
+      database: *database0
+      collectionName: &collection1Name coll1
```
jmikola, line 21: I propose adding a comment here (perhaps above `id:`):

```
# Nonexistent collection intentionally omitted in initialData
```

source/retryable-reads/tests/estimatedDocumentCount-serverErrors.yml
```diff
@@ -1,9 +1,11 @@
 runOn:
     -
         minServerVersion: "4.0"
+        maxServerVersion: "4.8.99"
```
jmikola, line 4: I propose adding a "-pre4.9" suffix to this filename. Likewise for `estimatedDocumentCount.yml`.

Alternatively, suffix both sets of files with the API used internally (e.g. "-count_cmd", "-aggregate_cmd").

This will ensure that anyone coming across these tests later on immediately knows what they pertain to, and won't mistake `estimatedDocumentCount.yml` (or this one) as the main/current tests for that method.

Note: I realize we don't have prior art in place that would have suggested doing this. The closest thing I can think of is CRUD v2 tests where certain methods started taking a `hint` option and we needed to split them into separate files with "-hint" suffixes. That said, this shouldn't be a concern with the unified format going forward since we can simply use test-level `runOnRequirements` (as you did in the new CRUD tests within this PR).

source/crud/tests/unified/estimatedDocumentCount.yml
benjirewis: We discussed offline; they seemed important to keep. Java and Go have three potential error paths for the new aggregate estimatedDocumentCount: 

1. `NamespaceNotFound` error (tested by nonexistent collection)
2. A command error that is not `NamespaceNotFound` (tested by the new failpoints w/ error codes)
3. All other errors (tested by this network failure)

I've revised the comment above, but the `isClientError: true` assertion fails. Will elaborate in the other thread about `isClientError`.

source/crud/tests/unified/estimatedDocumentCount.yml
benjirewis: Added!

source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -0,0 +1,213 @@
...
+        expectResult: 3
+    expectEvents:
+      - client: *client0
+        events:
+          - commandStartedEvent:
+              command:
+                aggregate: *collection0Name
+                pipeline: &pipeline
+                  - $collStats: { count: {} }
+                  - $group: {
```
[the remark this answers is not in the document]
benjirewis: Good point! Reduced to a single line here and in other, identical places.

source/crud/tests/unified/estimatedDocumentCount.yml
benjirewis: That sounds good! I've switched to `UnknownError(8)` and left a comment at each point in the yaml.

source/crud/tests/unified/estimatedDocumentCount.yml
benjirewis: Done!

source/crud/tests/unified/estimatedDocumentCount.yml
benjirewis: The `isClientError: true` assertion seems to fail on the Go test runner with
```
expected error connection(localhost:27017[-21]) socket was unexpectedly closed: EOF to be a client error: true, is client error: false
```
I'm not sure if this kind of network error (unexpected socket closure) qualifies as a client error.

Keeping just `isError` for now.

source/retryable-reads/tests/estimatedDocumentCount-serverErrors.yml
benjirewis: Ah that sounds good! I've added the "-pre4.9" suffix to this file, its JSON, `estimatedDocumentCount.yml`, and its JSON.

And yeah, this was much simpler in the one versioned api test that uses `estimatedDocumentCount`, as I could use test-level `runOnRequirements` since it was in unified format.

jyemin (approved):
All tests still passing.  Though I had to add errorCode support to the unified runner, since this is the first test that uses it.

source/crud/tests/unified/estimatedDocumentCount.yml
jmikola: I created [DRIVERS-1555](https://jira.mongodb.org/browse/DRIVERS-1555) to track this, and left a todo to come back and update this test at a later point.

source/crud/tests/README.rst
```diff
@@ -19,20 +19,21 @@ version requirements as noted by the ``runOn`` section, if provided.
...
+- ``v2/``: Legacy-v2 format tests
+- ``unified/``: Tests using the `unified test format <../../../../unified-test-format/unified-test-format.rst>`_
 
 Since some drivers may not have a unified test runner capable of executing tests
-in both formats, segregating tests in this manner will make it easier for
+in all three formats, segregating tests in this manner will make it easier for
 drivers to sync and feed test files to different test runners.
 
-Test Format
+Legacy-v2 Test Format
```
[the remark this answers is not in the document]
alcaeus: Apologies for missing this. I'm to sure why this wasn't caught by rstcheck, but I'm not very familiar with the tool. My only interaction was to migrate it to GH actions. From what I can see, the tool did not fail, so we may want to revisit our RST linting (@benjirewis feel free to suggest an alternative if you found that this catches errors more reliably).

source/crud/crud.rst
```diff
@@ -661,9 +661,26 @@ older sharded clusters.
...
+
+  pipeline = [
+    { '$collStats': { 'count': {} } },
+    { '$group': { '_id': 1, 'n': { '$sum': '$count' } } }
+  ]
+
+Similar to the count command, the estimated count of documents is returned
+in the ``n`` field.
+
+In the event this aggregation is run against a non-existent namespace, a NamespaceNotFound
```
jmikola, line 677: I thought I commented on this before but might have missed it. Please replace with NamespaceNotFound(28) so readers immediately know the code as well. I assume most drivers check the `code` field in the server response instead of `codeName`.

jmikola (approved):
One small suggestion in the CRUD spec. LGTM otherwise, so feel free to just merge after making that change.

source/crud/crud.rst
benjirewis: Sounds good! Changed to `NamespaceNotFound(28)`.

source/crud/tests/README.rst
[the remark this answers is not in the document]
benjirewis: @alcaeus  `rstcheck` works great! The reason it wasn't caught is because it's [called](https://github.com/mongodb/specifications/blob/master/.github/workflows/rstcheck.yml#L25) with the `--report error` flag on the GH action. "Title underline too short" is a warning, not an error. So, we'd have to change the action to use `--report warning` to catch it. Not sure if you want all the other warnings, though...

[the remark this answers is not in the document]
alcaeus: Ah, I see. I'll take a look through the warnings and see what we can do 👍

source/crud/crud.rst
jmikola: Oops, that was a typo on my part. NamespaceNotFound should be 26: https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.yml#L62

benjirewis: Fixed in this small [commit](https://github.com/mongodb/specifications/commit/5f9afa7d32dff0e25f44ceb74311e57682f67782). Also specified that drivers should compare against the error code.

PR #907, opened 2021-02-06
SPEC-1771 Mark server unknown after auth failure

ShaneHarvey:
For [SPEC-1771](https://jira.mongodb.org/browse/SPEC-1771)


source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -561,8 +561,8 @@ WaitQueueTimeout, if the driver supports one and it was specified by the user.
 
 If the pool is "closed" or "paused", any attempt to check out a `Connection
 <#connection>`_ MUST throw an Error. The error thrown as a result of the pool
-being "paused" MUST be considered a non-timeout network error for the purposes
-of retryability and monitoring.
+being "paused" MUST be considered a retryable error and MUST NOT be an error
+that marks the SDAM state unknown.
```
ShaneHarvey, line 565: @patrickfreed do you have any ideas for how to test this change? What are your thoughts on my alternative idea where pool.ready() increments the generation number when the state changes from "paused" to "ready"?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -2257,12 +2327,13 @@ These errors indicate one of these:
...
   or a read was attempted on a read-only server without the slaveOk bit set.
+* An operation was attempted on a server that is now shutting down.
 
 In any case the error is a symptom that
 a ServerDescription's type no longer reflects reality.
 
-A primary closes its connections when it steps down,
-so in many cases the next operation causes a network error
+On MongoDB 4.0 and earlier, a primary closes its connections when it steps
+down, so in many cases the next operation causes a network error
```
ShaneHarvey, line 2336: This is some misc. outdated prose I discovered while reading the spec but it's otherwise unrelated to this PR.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1235,6 +1235,55 @@ the pool's generation number then error handling MUST continue according to
 stale and the client MUST NOT update any topology state.
 (See `Why ignore errors based on CMAP's generation number?`_)
 
+Error handling behavior
+~~~~~~~~~~~~~~~~~~~~~~~
```
ShaneHarvey, line 1239: This section still needs to be fleshed out a bit more. My goal is to replace the long-winded "Network error when reading or writing" and "not master" and "node is recovering" error handling sections with a more straightforward section focused on pseudocode.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
ShaneHarvey: Great points all around. I changed this section to a MUST because I think we should standardize error handling in SDAM. I haven't changed the generation behavior since I don't think it's necessary in this PR.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -920,6 +920,15 @@ a manner idiomatic to the Driver and Language.
       address: <pool address>;
     }
 
+    /**
+     *  Thrown when the driver attempts to check out a
+     *  Connection from a paused Connection Pool
+     */
+    interface PoolClearedError extends RetryableError {
+      message: 'Attempted to check out a Connection from paused connection pool';
```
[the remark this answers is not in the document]
ShaneHarvey: I like it. Even better if the driver can include the reason why the pool was cleared to begin with. Updated to:
```
    interface PoolClearedError extends RetryableError {
      message: 'Connection pool for <pool address> was cleared because another operation failed with: <original error which cleared the pool>';
```

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1235,6 +1235,55 @@ the pool's generation number then error handling MUST continue according to
...
+                return
+
+            if isNetworkError(error) or (not error.completedHandshake and (isNetworkTimeout(error) or isAuthError(error))):
+                # Mark the server Unknown
+                unknown = new ServerDescription(type=Unknown, error=error)
+                onServerDescriptionChanged(unknown, connection pool for server)
+                clear connection pool for server
+                # Cancel inprogress check
+                cancel monitor check
+            elif isStateChangeError(error):
```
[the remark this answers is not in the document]
ShaneHarvey: Done.

source/server-discovery-and-monitoring/tests/integration/auth-network-error.yml
```diff
@@ -0,0 +1,81 @@
...
+
+database_name: &database_name "sdam-tests"
+collection_name: &collection_name "auth-network-error"
+
+data: &data
+  - {_id: 1}
+  - {_id: 2}
+
+tests:
+  - description: Reset server and pool after network error during authentication
```
patrickfreed, line 15: a similar one for network timeouts (both auth and non-auth) could be nice to cover the last bit of the branch for handling pre-handshake errors.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
jyemin: > do the Java or C# drivers update the topology based on errors in the background thread while ensuring minPoolSize?

No for Java.  Probably not for C# but haven't confirmed.

bq. would it be possible for them to do so?

I don't see why not.  Currently, the connection pool implementations have no reference to the topology.  The reference is the other way round.  But we can introduce some sort of invalidation callback.

patrickfreed: Awesome, then it seems like we're safe to upgrade this SHOULD -> MUST then.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1235,6 +1235,122 @@ the pool's generation number then error handling MUST continue according to
...
 
+Error handling pseudocode
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Application operations can fail in various places, for example:
+
+- A network error, network timeout, or command error may occur while
+  establishing a new connection. Establishing a connection includes the
+  MongoDB handshake and completing authentication (if configured).
+- A network error, network timeout may occur while reading or writing to an
```
patrickfreed, line 1246: the comma here could be replace with "or" and would make this a little easier to read.

divjotarora:
I don't have the bandwidth to review this over the next couple of days, so I'll defer to @patrickfreed's judgement for the relevant CMAP portions.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
ShaneHarvey: Done.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -2172,6 +2209,27 @@ duplicate (or stale) State Change Errors can be identified (changes in
    the error's topologyVersion (tv2) is not greater than the current
    ServerDescription (tv2).**
 
+Why mark a server Unknown after an auth error?
+''''''''''''''''''''''''''''''''''''''''''''''
+
+The `Authentication spec`_ requires that when authentication fails on a server,
```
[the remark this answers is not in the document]
ShaneHarvey: Updated with this fact.

source/server-discovery-and-monitoring/tests/README.rst
```diff
@@ -193,7 +193,14 @@ Test Format
 ~~~~~~~~~~~
 
 The same as the `Transactions Spec Test format
-</source/transactions/tests/README.rst#test-format>`_.
+</source/transactions/tests/README.rst#test-format>`_ with the following
+additions:
+
+- The ``runOn`` requirement gains a new field:
+
+  - ``authEnabled`` (optional): If True, skip this test if auth is not enabled.
```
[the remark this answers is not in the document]
ShaneHarvey: Fixed.

source/server-discovery-and-monitoring/tests/integration/auth-network-error.yml
ShaneHarvey: I added two more tests, one for network timeouts during auth and one for network timeouts post-auth.

source/server-discovery-and-monitoring/tests/integration/pool-cleared-error.yml
```diff
@@ -0,0 +1,197 @@
...
+        arguments:
+          name: thread6
+      # Perform concurrent insert operations. The first one to execute will
+      # fail with a network error, mark the server Unknown, clear the pool,
+      # and retry.
+      # The other operations will either:
+      # - Notice the pool is paused, fail with a PoolClearedError, and retry.
+      # - Or block waiting in server selection until the server is rediscoverd.
+      #
+      # Note that this test does not gaurantee that a PoolClearedError will be
```
patrickfreed, line 69: We could guarantee this by adding a `blockTimeMS` to the failpoint, which should give enough time for the other threads to enter the wait queue.

Also, a few minor typos here "gaurantee" -> "guarantee", "rediscoverd" -> "rediscovered"

source/server-discovery-and-monitoring/tests/integration/pool-cleared-error.yml
```diff
@@ -0,0 +1,197 @@
...
+        object: testRunner
+        arguments:
+          name: thread5
+      - name: startThread
+        object: testRunner
+        arguments:
+          name: thread6
+      # Perform concurrent insert operations. The first one to execute will
+      # fail with a network error, mark the server Unknown, clear the pool,
+      # and retry.
```
patrickfreed, line 64: This test is covering the DRIVERS-1483 / DRIVERS-746 retry behavior, but the retryability specs haven't yet been updated to account for pre-handshake errors yet. Should we tweak this test to not have any dependency on retry behavior / skip it until those have been done?

patrickfreed (commented):
Prose changes LGTM, just have some comments on one of the tests.

source/server-discovery-and-monitoring/tests/integration/pool-cleared-error.yml
ShaneHarvey: No this is not testing DRIVERS-746 because in this case the handshake succeeds and it's the insert command that fails. As for DRIVERS-1483, I suppose this test is very similar. Still I think we should keep this test and also complete DRIVERS-1483. I think the spec work for DRIVERS-746 can follow after.

patrickfreed: The only reason I bundle the two tickets together is that their implementation / spec work would be basically identical except for differing tests. That being said, we can probably get spec work for at least DRIVERS-1483 done before any drivers have to implement the changes for this ticket, so I'm okay with leaving the test as-is then.

patrickfreed (approved):
Besides the merge conflict, LGTM! Great job with this, it ended being a lot more nuanced/complicated than I originally thought. 

Note: I wasn't able to run the tests in Rust since we don't have the runner for SDAM integration tests yet, so maybe another driver should verify.

As a follow on, I think I can pick up the work for DRIVERS-1483, which seems like it will be required now that we don't explicitly classify the PoolClearedError as a network error for the sake of retryability. I don't think it should be too involved. cc @kmahar

ShaneHarvey:
Is the PoolClearedError test I added in this PR sufficient to cover DRIVERS-1483? I think it might be.

patrickfreed:
So the title of that ticket is specific to testing, but if we're going to be doing it before and separately from DRIVERS-746, the retryability specs will need to be updated with prose/psuedocode to reflect that a failed check out could trigger a retry as part of it too. Ideally those specs would get their own tests of this behavior too, since it will be behavior required by them.

ShaneHarvey (commented):
@bazile-clyde, @emadum, @nangnguyen404 would any of you like to review this?

ShaneHarvey:
I fixed the merge conflict and added one extra test to ensure miscellaneous command errors during auth will also clear the pool. Not planning to add anymore tests else unless requested.

bazile-clyde (approved):
The auth changes LGTM.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1235,6 +1235,122 @@ the pool's generation number then error handling MUST continue according to
...
+
+- A network error, network timeout, or command error may occur while
+  establishing a new connection. Establishing a connection includes the
+  MongoDB handshake and completing authentication (if configured).
+- A network error or network timeout may occur while reading or writing to an
+  established connection.
+- A command error may be returned from the server.
+- A "writeConcernError" field may be included in the command response.
+
+Depending on the context, these errors may update update SDAM state by marking
```
nbbeeken, line 1251:
```suggestion
Depending on the context, these errors may update SDAM state by marking
```

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1235,6 +1235,122 @@ the pool's generation number then error handling MUST continue according to
...
+
+        with client.lock:
+            # Ignore stale errors based on generation and topologyVersion.
+            if isStaleError(client.topologyDescription, error)
+                return
+
+            if isStateChangeError(error):
+                # Mark the server Unknown
+                unknown = new ServerDescription(type=Unknown, error=error, topologyVersion=topologyVersion)
+                onServerDescriptionChanged(unknown, connection pool for server)
```
nbbeeken, line 1269: Is `onServerDescriptionChanged` supposed to mean 'emit the serverDescription change' event?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1235,6 +1235,122 @@ the pool's generation number then error handling MUST continue according to
...
+        currentGeneration = currentServer.pool.generation
+        generation = get connection generation from error
+        if generation < currentGeneration:
+            # Stale generation number.
+            return True
+
+        currentTopologyVersion = currentServer.topologyVersion
+        # True if the current error's topologyVersion is greater than the server's
+        # We use >= instead of > because any state change should result in a new topologyVersion
+        return compareTopologyVersion(currentTopologyVersion, error.commandResponse.get("topologyVersion")) >= 0
```
nbbeeken, line 1300: What does `compareTopologyVersion` do on the inside? Is it not a simple number comparison?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -2174,6 +2211,29 @@ duplicate (or stale) State Change Errors can be identified (changes in
...
+- A network error, or network timeout error may occur.
+- The server may return a `State Change Error`_.
+- The server may return a AuthenticationFailed command error (error code 18)
+  indicating that the provided credentials are invalid.
+
+Does this mean that authentication failures due to invalid credentials will
+manifest as server selection timeout errors? No, authentication errors are
+still returned to the application immediately. A subsequent operation will
+block until the server is rediscovered and immediately attempt
+authentication on a new connection.
```
nbbeeken, line 2235: How is the initial client.connect treated differently so that auth errors don't turn into server selection errors? Or is that supposed to go through selection?

If I'm already connected and there is an auth error when trying the insertOne, is this correct?
```ts
try { await coll.insertOne({ a: 1 }) } catch (e) { e === AuthenticationError }
// Servers are marked as unknown
try { await coll.findOne({ a: 1 }) } catch (e) { e === ServerSelectionError }
```

nbbeeken (commented):
Some clarifying Qs

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
ShaneHarvey: Yes it means "process the server description". It's defined here: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#replacing-the-topologydescription

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
ShaneHarvey: It's defined here: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#replacing-the-topologydescription

Note I didn't change isStaleError at all in this pr, I only shuffled the code from down below.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
ShaneHarvey: If the credentials are totally invalid, every operation will fail with an auth error (as before) expect that each failure will also mark the server unknown. The behavior will be:
```js
try { await coll.insertOne({ a: 1 }) } catch (e) { e === AuthenticationError }
// Servers are marked as unknown
// Next operation waits for Monitor to rediscover the server
try { await coll.insertOne({ a: 1 }) } catch (e) { e === AuthenticationError }
```

nbbeeken: Oh I see now, selection will run with that second op, since we have unknowns we wait for the monitor, but when it does attempt a connection we will get the Auth error.

nbbeeken (approved):
👍

PR #909, opened 2021-02-08
SPEC-1779 Add the UpdateDescription.truncatedArrays field

stIncMale:
SPEC-1779


source/change-streams/change-streams.rst
```diff
@@ -178,13 +181,29 @@ If an aggregate command with a ``$changeStream`` stage completes successfully, t
...
+     * Truncations of arrays may be reported using one of the following methods:
+     * either via this field or via the ‘updatedFields’ field. In the latter case the entire array is considered to be replaced.
+     *
+     * Example: an `update` which shrinks the array arrayField.0.nestedArrayField from size 8 to 5 may be reported as
+     *   "truncatedArrays": [{field: "arrayField.0.nestedArrayField", newSize: 5}]
+     *
+     * @note The method used to report a truncation is a server implementation detail.
+     * @since 4.7.0
+     */
+    truncatedArrays: Array<String>;
```
divjotarora, line 206: Based on your example, should this be `Array<Document>` where each document is of the form `{field: <string>, newSize: <int>}`?

divjotarora:
@prashantmital Do you think a spec test is needed for this? I'm leaning towards no because it doesn't seem like we have spec tests for other event fields like `updatedFields`. Maybe drivers that provide helper types to represent events could write their own tests for this?

source/change-streams/change-streams.rst
stIncMale: I apologize, this is my mistake caused by copy-pasting. Yes, the type is `Array<Document`.

source/change-streams/change-streams.rst
```diff
@@ -129,6 +129,9 @@ If an aggregate command with a ``$changeStream`` stage completes successfully, t
 
     /**
      * Describes the type of operation represented in this change notification.
+     *
+     * @note: Whether a change is reported as an event of the operation type
+     *       `update` or `replace` is a server implementation detail.
```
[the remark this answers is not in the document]
stIncMale: The indentation is intentional. It is consistent with the indentation used in similar situations in the Java driver code. Please tell me if this style is surprising or unwelcome, and I will remove the indentation.

stIncMale:
@divjotarora
> it doesn't seem like we have spec tests for other event fields like `updatedFields`

There appears to be one: https://github.com/mongodb/specifications/blob/master/source/change-streams/tests/change-streams.json#L547. Thank you for raising a concern about spec tests, I will try to come up with one for this change.

source/change-streams/change-streams.rst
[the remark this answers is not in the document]
divjotarora: I'm not sure why the Java driver docs do this, but unless there's precedent for doing so in this spec, I'd prefer to remove the additional spaces.

divjotarora (commented):
@stIncMale Thanks for pointing out a test that makes assertions about `updatedFields`. I'll wait for you to add another test before reviewing again.

source/change-streams/change-streams.rst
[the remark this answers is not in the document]
jmikola: I understand why this was originally indented and I think it has more to do with `@note` being an annotation instead of just prose text within the doc block. If we look at [doc blocks in the CRUD spec](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#read), you'll see that most notes are written as a regular paragraph starting with "Note:". Some `@note` annotations were added to the spec in https://github.com/mongodb/specifications/commit/bb8677b9abb3440ca7e843d5f4ca554b8844e905 but they were one-liners.

IMO, it'd be preferable to just change this to "Note:" and leave it as a prose paragraph.

divjotarora:
@stIncMale @jmikola How do you feel about the new test being written using the new [unified test format](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst)? The current change stream tests are not in the unified format, but we can account for this by splitting up the `tests` directory into `unified` and `legacy` sub-directories where `legacy` has all of the existing files and `unified` contains the new one. Thoughts?

stIncMale:
@divjotarora I am familiar with neither the change stream-specific test format (legacy), nor with the unified test format. I also do not know at this moment whether or not using the unified format will introduce a noticeable work overhead. However, I like the idea of using the unified format for new tests.

I will familiarize myself with the legacy and the unified formats and try to use the unified one.

jmikola:
If a new test is warranted, it absolutely makes sense to create it in the unified format. I concur with splitting them up by directory. For prior art, you can look at what was recently done in https://github.com/mongodb/specifications/pull/905. Note that this also came with some changes to the test README. Existing docs were left in place and revised to refer to the old tests in their respective directories. Another note was added about tests in the `unified/` directory with a link to the [Unified Test Format](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst) spec.

Take a look at https://github.com/mongodb/specifications/blob/master/source/unified-test-format/tests/valid-pass/poc-change-streams.yml for some guidance on writing a new test. Those are some existing change stream tests that were ported to the unified format (strictly for testing the test runner implementation and its operations related to change streams). I believe the test names are the same if you'd like to compare the old (under the CS spec's current test directory) and unified tests with each other.

stIncMale:
@divjotarora, @jmikola After thinking more about the specification test (with help from @jyemin) I think such a test would check nothing new:
1. We already have specification tests that check that drivers correctly send commands that update or replace documents/fields, and that drivers are able to receive server responses.
2. This PR is about a new field in the responses sent by the MongoDB server. The only thing a new specification test could possibly verify in addition to 1 is that the server does indeed send the new field in some cases. In other words, such a new specification test would test solely the MongoDB server behavior, which is not what driver specification tests should do.

The only change required in a driver as a result of this spec change, is supporting the new BSON field in the codec for `ChangeStreamDocument`, provided that a driver converts change stream documents to a custom type (the change stream specification does not require doing so, but mentions it as a possibility, see the first NOTE in the [Response Format](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#response-format)).

With this in mind, I am re-requesting your reviews.

source/change-streams/change-streams.rst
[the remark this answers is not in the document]
stIncMale: I removed indentation.

@jmikola 
Extracting "Note:" out of `/** ... */` scatters the specification of `ChangeStreamDocument.operationType` to multiple places. This, I believe, hurts readability by improving the chances to miss a piece of information about `operationType` when looking through the document.

jyemin:
Just to give a little more background on what @stIncMale said above:  In the Java driver, by default when you iterate a change stream using `watch` you get an instance of `MongoCursor<ChangeStreamDocument<Document>>`.  But for purposes of implementing change stream unified tests, we actually change that to a `MongoCursor<BsonDocument>`, because that allows us to easily compare the actual results against the expected ones.  So at least in the Java driver the unified tests don't actually test `ChangeStreamDocument` at all.  If we wanted to test `ChangeStreamDocument` we would have to implement some change stream-specific comparison code in the test runner that allowed us to compare a `ChangeStreamDocument` with a `BsonDocument`.

However, other drivers with classes similar to `ChangeStreamDocument` may be different, so maybe it's still worth it to have a test.  Or perhaps we should actually change the Java driver's unified runner to use `ChangeStreamDocument`.

divjotarora:
Such a test wouldn't be useful for Go either because our ChangeStream returns the raw BSON documents from the server, but like @jyemin said, it might still be useful to have a test in case other drivers have special behavior.

jmikola (approved):
Points noted. I concur that we don't need spec tests that only verify server behavior.

And it's reasonable to assume that any drivers modeling the server's return value should test this independently.

prashantmital:
FWIW the python driver explicitly checks for the presence of this field when testing against 4.5+ - https://github.com/mongodb/mongo-python-driver/blob/c15028a6c714519e5ec86d1a9bb50e95a04de36b/test/test_change_stream.py#L373

stIncMale:
Thanks, @prashantmital. I have two questions regarding this fact you mentioned:

1. The functionality is implemented in MongoDB 4.7 according to https://jira.mongodb.org/browse/SERVER-47354, the corresponding https://jira.mongodb.org/browse/DOCSP-13958 also specifies 4.7. Backporting to 4.4 [was declined](https://jira.mongodb.org/browse/BACKPORT-6610). Why does the Python driver check the field since 4.5?
2. It appears that the Python driver explicitly expects an empty array `[]`, but found nothing in the [design](https://docs.google.com/document/d/1sxGrck8X0r0W37UVLFwV8vaUQv3OM8xlyWWvHB5M2ew/edit#bookmark=kix.7cw4ualn66jl) specifying that `truncatedArrays ` cannot be `null` instead of being empty. Did you find this specified somewhere else?

source/change-streams/tests/unified/change-streams.json
```diff
@@ -0,0 +1,115 @@
+{
```
stIncMale, line 1: This file was generated from `change-streams.yaml` via [`js-yaml`](https://github.com/nodeca/js-yaml).

source/change-streams/tests/unified/change-streams.yml
```diff
@@ -0,0 +1,74 @@
...
+        arguments:
+          document: {
+            "_id": 1,
+            "a": 1,
+            "array": ["foo", {"a": "bar"}, 1, 2, 3]
+          }
+      - name: createChangeStream
+        object: *collection0
+        saveResultAsEntity: &changeStream0 changeStream0
+      - name: updateOne
```
stIncMale, line 37: This `updateOne` operation is not supported by the unified test runner in the Java driver. @divjotarora Helped me with this operation and confirmed it is supported by the Go driver.

The reason this can't be a simple `findOneAndUpdate` operation is that the MongoDB server currently may consider using `truncatedArrays` only if the update is done via an aggregation pipeline.

@divjotarora Could you please try the committed version of the test with the Go driver to make sure it still works? (I cannot currently run the test with the Java driver).

jyemin: You should add support for updateOne in the Java driver's unified test runner and confirm that the test passes.

prashantmital:
> Thanks, @prashantmital. I have two questions regarding this fact you mentioned:
> 
> 1. The functionality is implemented in MongoDB 4.7 according to https://jira.mongodb.org/browse/SERVER-47354, the corresponding https://jira.mongodb.org/browse/DOCSP-13958 also specifies 4.7. Backporting to 4.4 [was declined](https://jira.mongodb.org/browse/BACKPORT-6610). Why does the Python driver check the field since 4.5?

IIRC our tests started failing (due to absence of the `truncatedArrays` field in our expected response) on the `latest` version of the server available in Evergreen after 4.4 was tagged. Consequently, we started testing the presence of `truncatedArrays` on MongoDB 4.5+. There might be some dissonance in terms of the exact server fixVersion we should expect for this feature as it is during this time that the target release versions were refreshed to come inline with the continuous delivery initiative (see Steven's email dated 9/3/2020).

> 2. It appears that the Python driver explicitly expects an empty array `[]`, but found nothing in the [design](https://docs.google.com/document/d/1sxGrck8X0r0W37UVLFwV8vaUQv3OM8xlyWWvHB5M2ew/edit#bookmark=kix.7cw4ualn66jl) specifying that `truncatedArrays ` cannot be `null` instead of being empty. Did you find this specified somewhere else?

The empty array was expected in the test based on what the server was returning and not based on any specification/design.

source/change-streams/tests/unified/change-streams.yml
stIncMale: Yes, done in https://github.com/mongodb/mongo-java-driver/pull/667.

jyemin: Ack.  Thanks!

source/change-streams/tests/unified/change-streams.yml
```diff
@@ -0,0 +1,74 @@
...
+      - name: iterateUntilDocumentOrError
+        object: *changeStream0
+        expectResult: {
+          "operationType": "update",
+          "ns": {
+            "db": "database0",
+            "coll": "collection0"
+          },
+          # This expectation is based on the MongoDB server behavior
+          # (as opposed to be based on its API),
```
jyemin, line 59: The server doesn't have a formal API specification.  Can you elaborate on what you mean by this so we can decide whether this comment makes sense?  This is a public repository so we have to be clear as to the meaning of the terms we use.

stIncMale: I think it does. The API<sup>1</sup> is specified in multiple places in the official docs, but mostly by https://docs.mongodb.com/manual/reference/command/. The [design](https://docs.google.com/document/d/1sxGrck8X0r0W37UVLFwV8vaUQv3OM8xlyWWvHB5M2ew/edit#heading=h.22yp33tzmzpp) for the functionality that caused this PR also uses the term "API".

----
<sup>1</sup> I believe, previously people used the term "protocol" for this, and API was used as a name for *programming* interfaces. Now the terms "protocol" and "API" seem to be mixed for no good reason known to me.

stIncMale: > what you mean by this

It's up to the server to decide whether to report a change as truncation or not, so even if the change can be represented as truncation, it does not mean it will. How the decision is made is an implementation detail which may change.

source/change-streams/tests/unified/change-streams.yml
```diff
@@ -0,0 +1,74 @@
+description: "change-streams"
+schemaVersion: "1.0"
+createEntities:
+  - client:
+      id: &client0 client0
+      observeEvents: []
+      ignoreCommandMonitoringEvents: []
```
jmikola, line 7: `observeEvents` and `ignoreCommandMonitoringEvents` are both optional. If you're not specifying any events here, I suggest just omitting the fields entirely. In fact, empty arrays should fail to validate with the schema as it [requires `minItems: 1`](https://github.com/mongodb/specifications/blob/c6645365fb68e9427f8599196a790819942f4e64/source/unified-test-format/schema-1.1.json#L79) for both fields.

I realize the `client` [entity](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#entity) definition doesn't explicitly say these default to empty arrays. Rather, it says:

> Optional array of one or more strings.

If you think that's unclear, we can amend the spec language to explicitly say they default to an empty array. I was previously just relying on the schema validation itself to enforce this, and prevent folks from redundantly specifying an empty array.

source/change-streams/tests/unified/change-streams.yml
```diff
@@ -0,0 +1,74 @@
...
+  - collection:
+      id: &collection0 collection0
+      database: *database0
+      collectionName: *collection0
+initialData:
+  - collectionName: *collection0
+    databaseName: *database0
+    documents: []
+tests:
+  - description: "Test array truncation"
```
jmikola, line 21: When I last reviewed this PR, I thought there was consensus about not introducing a test that only asserted server behavior. Previous comments from [Jeff](https://github.com/mongodb/specifications/pull/909#issuecomment-778212888) and [Divjot](https://github.com/mongodb/specifications/pull/909#issuecomment-778315408) suggested this wouldn't be useful for Java and Go, respectively, since they both inspect change stream events as raw BSON instead of using their model classes. AFAIK, most drivers don't model change stream events, so I'm wondering if there are actually any other drivers that would benefit from this test.

Alternatively, did folks change their mind about it being generally useful to have a spec test that asserts server behavior?

source/change-streams/tests/unified/change-streams.yml
stIncMale: Thank you,
> Optional array of one or more strings.
is clear. The reason I violated the schema is that I either had not noticed this phrase in the unified format, or had not paid attention. This said, I agree that the JSON validation agains the schema should be made automatic.

For now I validated the compliance of `change-streams.json` to `unified-test-format/schema-1.1.json` with [justify](https://github.com/leadpony/justify#command-line-interface).

divjotarora: @stIncMale pointed out that we do have some tests which assert server behavior and there are some drivers that model change stream documents as native types (e.g. structs/classes), so having a test might benefit at least a subset of drivers. It's also a relatively simple test to add so I'm fine having it.

source/change-streams/tests/unified/change-streams.yml
divjotarora: @stIncMale pointed out that we have some tests which assert server behavior. Also, even if there are only a handful of drivers that convert change stream results to native types, it seems useful to have this test rather than have each of those drivers figure out how to force the server to return `truncatedArrays` individually.

source/change-streams/tests/unified/change-streams.yml
jyemin: OK, I see.  I agree with you about the command reference docs being the closest thing we have to an API specification for the server.  So even if we wanted to we couldn't base this test on the API, since the docs for truncatedArrays has yet to be published or written yet.  But once it has it may turn out to be sufficient as the basis for this test. In other words it may be inaccurate to say this is not part of the API, since we don't yet know what the API will specify (see https://jira.mongodb.org/browse/DOCS-14126).

Any other reviewers have thoughts on this?  It's a minor point, and I'm ok to resolve if no one else thinks it's important.

source/change-streams/tests/unified/change-streams.yml
stIncMale: @divjotarora contacted me in Slack and we agreed that I will try and create a unified test.

Speaking about the usefulness, it appears to me that different people who have participated in the discussion have different opinions:
- I do not see much usefulness from the test regardless of whether a driver's unified runner deserializes `ChangeStreamDocument`s or not. However, I also do not have anything agains having a unified test.
- @jyemin appears to see usefulness for drivers that deserialize `ChangeStreamDocument` in unified tests, and proposed (in Slack) a way to achieve this in the Java driver (has not been done; Jeff, I apologize if I interpreted you incorrectly).
- @divjotarora appears to lean towards having a test based on our dialog in Slack (again, I apologize if I interpreted that incorrectly).

source/change-streams/tests/unified/change-streams.yml
```diff
@@ -0,0 +1,72 @@
...
+    operations:
+      - name: insertOne
+        object: *collection0
+        arguments:
+          document: {
+            "_id": 1,
+            "a": 1,
+            "array": ["foo", {"a": "bar"}, 1, 2, 3]
+          }
+      - name: createChangeStream
```
divjotarora, line 32: This test fails in Go because the `pipeline` argument is missing and the test runner considers it a required argument. The change stream tests in [poc-change-streams](https://github.com/mongodb/specifications/blob/2ce041277c07b2f67b0dff73f1b04fa7e10f2af8/source/unified-test-format/tests/valid-pass/poc-change-streams.yml) all have it, so can you add it here as well?

source/change-streams/tests/unified/change-streams.yml
stIncMale: I am claiming that judging based on the [design](https://docs.google.com/document/d/1sxGrck8X0r0W37UVLFwV8vaUQv3OM8xlyWWvHB5M2ew/edit#heading=h.c4056nub8lzp) of the server feature, what we relying on in this test is implementation-specific behavior. I also expressed this in the specification change: https://github.com/mongodb/specifications/pull/909/files#diff-b1cdd6bd41c94cc2b4ba9bef6f89e372fcc59b67dc96f349a95a85f9178095d4R218

source/change-streams/tests/unified/change-streams.yml
divjotarora: The test currently fails in Go (see my comment on the `createChangeStream` operation above). If I fix that issue locally, though, the test passes.

source/change-streams/tests/unified/change-streams.yml
divjotarora: I also agree that command reference docs are the closest to a server API spec in today's world and that the design doc for this feature didn't fully clarify which situations would lead to array changes being reported in `truncatedArrays` vs `updatedFields` or `removedFields`. However, the design isn't meant to be public documentation and as @jyemin mentioned, the public documentation for this feature hasn't been written yet, so maybe we should avoid saying this in a public comment.

If we feel that this comment is useful, maybe we can change it to "the inserted document and the update pipeline were determined empirically by testing out different combinations until change stream event returned a "truncatedArrays" field"? This isn't necessarily the exact wording, but it avoids the possibly-contentious mentioning of "server API".

divjotarora (commented):
Two overall comments on tests:

1. Should we move the existing tests to a new `change-streams/tests/legacy` directory? @jmikola might be able to provide guidance on this. 
2. The Change Streams test [README](https://github.com/mongodb/specifications/blob/2ce041277c07b2f67b0dff73f1b04fa7e10f2af8/source/change-streams/tests/README.rst) should be updated to reflect the final structure of the tests directory.

source/change-streams/tests/unified/change-streams.yml
stIncMale: Done. I also updated the unified runner in the Java driver.

source/change-streams/tests/unified/change-streams.yml
stIncMale: Done.

source/change-streams/tests/unified/change-streams.yml
stIncMale: Done.

jyemin (approved):
LGTM

stIncMale:
@divjotarora 
> Two overall comments on tests:
> 
> 1. Should we move the existing tests to a new `change-streams/tests/legacy` directory? @jmikola might be able to provide guidance on this.
> 2. The Change Streams test [README](https://github.com/mongodb/specifications/blob/2ce041277c07b2f67b0dff73f1b04fa7e10f2af8/source/change-streams/tests/README.rst) should be updated to reflect the final structure of the tests directory.

2 is done.

@jmikola could you please provide your input about 1? We may also just leave the legacy tests be where they currently reside.

prashantmital (commented):
I would be in favor of moving all existing tests to a new directory `source/change-streams/tests/legacy/` and adding the new (unified) tests directly to `source/change-streams/tests/`. (I think) The baseline expectation after Unified Test Format is that all tests adhere to this unified format unless otherwise specified so it makes more sense to sequester the old-format tests under a separate directory.

stIncMale:
@divjotarora I moved the legacy tests and updated the README.

divjotarora (approved):
Code changes LGTM. To avoid blocking this PR much longer, we decided to move forward with the addition of a new `tests/legacy` sub-directory to hold the old tests. As for the location of the new test file, I have no preference between putting it in `tests/` vs a new `tests/unified` sub-directory.

prashantmital (approved):

PR #910, opened 2021-02-10
SPEC-1780 Fix typos in estimatedDocumentCount spec and tests

benjirewis:
Adds a line clarifying an assumption drivers can make about the result of an aggregate with $collStats.

Disables multiple mongoses on the recently added spec tests for estimatedDocumentCount so that failpoints work as expected on sharded topologies.

### DIFF
--- a/source/crud/crud.rst
+++ b/source/crud/crud.rst
@@ -672,8 +672,10 @@ As documented above, the only supported option is maxTimeMS::
   ]
 
 Similar to the count command, the estimated count of documents is returned
-in the ``n`` field. Drivers can assume that the first field of the first batch
-returned from the cursor contains the ``n`` field.
+in the ``n`` field. Implementations can assume that the document containing
+the single result of the aggregation pipeline is contained in the first batch of
+the server's reply to the aggregate command. It is not necessary to execute a getMore
+operation to ensure that the result is available.
 
 In the event this aggregation is run against a non-existent namespace, a NamespaceNotFound(26)
 error will be returned during execution. Drivers MUST interpret the server error code 26 as
--- a/source/crud/crud.rst
+++ b/source/crud/crud.rst
@@ -672,7 +672,8 @@ As documented above, the only supported option is maxTimeMS::
   ]
 
 Similar to the count command, the estimated count of documents is returned
-in the ``n`` field.
+in the ``n`` field. Drivers can assume that the first field of the first batch
+returned from the cursor contains the ``n`` field.
 
 In the event this aggregation is run against a non-existent namespace, a NamespaceNotFound(26)
 error will be returned during execution. Drivers MUST interpret the server error code 26 as
--- a/source/crud/tests/unified/estimatedDocumentCount.yml
+++ b/source/crud/tests/unified/estimatedDocumentCount.yml
@@ -5,6 +5,7 @@ schemaVersion: "1.0"
 createEntities:
   - client:
       id: &client0 client0
+      useMultipleMongoses: false # Avoid setting fail points with multiple mongoses
       uriOptions: { retryReads: false } # Avoid retrying fail points with closeConnection
       observeEvents: [ commandStartedEvent ]
   - database:
--- a/source/crud/tests/unified/estimatedDocumentCount.json
+++ b/source/crud/tests/unified/estimatedDocumentCount.json
@@ -10,7 +10,8 @@
                 ],
                 "uriOptions": {
                     "retryReads": false
-                }
+                },
+                "useMultipleMongoses": false
             }
         },
         {
--- a/source/versioned-api/tests/crud-api-version-1-strict.json
+++ b/source/versioned-api/tests/crud-api-version-1-strict.json
@@ -596,7 +596,12 @@
     },
     {
       "description": "estimatedDocumentCount appends declared API version",
-      "skipReason": "DRIVERS-1437 count was removed from API version 1",
+      "skipReason": "DRIVERS-1561 collStats is not in API version 1",
+      "runOnRequirements": [
+        {
+          "minServerVersion": "4.9.0"
+        }
+      ],
       "operations": [
         {
           "name": "estimatedDocumentCount",
@@ -611,7 +616,22 @@
             {
               "commandStartedEvent": {
                 "command": {
-                  "count": "test",
+                  "aggregate": "test",
+                  "pipeline": [
+                    {
+                      "$collStats": {
+                        "count": {}
+                      }
+                    },
+                    {
+                      "$group": {
+                        "_id": 1,
+                        "n": {
+                          "$sum": "$count"
+                        }
+                      }
+                    }
+                  ],
                   "apiVersion": "1",
                   "apiStrict": true,
                   "apiDeprecationErrors": {
--- a/source/versioned-api/tests/crud-api-version-1-strict.yml
+++ b/source/versioned-api/tests/crud-api-version-1-strict.yml
@@ -216,7 +216,9 @@ tests:
                 <<: *expectedApiVersion
 
   - description: "estimatedDocumentCount appends declared API version"
-    skipReason: "DRIVERS-1437 count was removed from API version 1"
+    skipReason: "DRIVERS-1561 collStats is not in API version 1"
+    runOnRequirements:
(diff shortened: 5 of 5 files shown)


source/crud/crud.rst
```diff
@@ -672,7 +672,8 @@ As documented above, the only supported option is maxTimeMS::
   ]
 
 Similar to the count command, the estimated count of documents is returned
-in the ``n`` field.
+in the ``n`` field. Drivers can assume that the first field of the first batch
+returned from the cursor contains the ``n`` field.
```
benjirewis, line 676: Is this a reasonable assumption @dstorch ?

source/crud/crud.rst
```diff
@@ -672,7 +672,8 @@ As documented above, the only supported option is maxTimeMS::
   ]
 
 Similar to the count command, the estimated count of documents is returned
-in the ``n`` field.
+in the ``n`` field. Drivers can assume that the first field of the first batch
```
jyemin, line 675: This language is a bit confusing, as it's not the first field of the first batch.  How about something like:

Implementations can assume that the document containing the single result of the aggregation pipeline is contained in the first batch in the server's reply to the aggregate command.  It is not necessary to execute a getMore operation to ensure that the result is available.

jyemin (changes requested):

source/crud/crud.rst
benjirewis: Yes, that sounds much better. Used almost exactly what you wrote.

jyemin (approved):
LGTM

source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -5,6 +5,7 @@ schemaVersion: "1.0"
 createEntities:
   - client:
       id: &client0 client0
+      useMultipleMongoses: false # Avoid setting fail points with multiple mongoses
```
jmikola, line 8: Sorry for not catching this in #905. FWIW, this is already advised in the Unified Test Format spec under [failPoint](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#failpoint).

source/crud/crud.rst
```diff
@@ -672,7 +672,10 @@ As documented above, the only supported option is maxTimeMS::
   ]
 
 Similar to the count command, the estimated count of documents is returned
-in the ``n`` field.
+in the ``n`` field. Implementations can assume that the document containing
+the single result of the aggregation pipeline is contained in the first batch of
+the server's reply to the aggregate command. It is not necessary to execute a getMore
+operation to ensure that the result is available.
```
jmikola, line 678: I didn't follow the entire conversation you had with @jyemin in Slack, but why is this a concern? We don't have any language like this for [countDocuments](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#countdocuments), which also uses an aggregation pipeline. In both cases, we're `$group`-ing into a single bucket. Is there any conceivable reason to expect that the single result document would _not_ be in the immediate command response to `aggregate`?

jmikola (approved):
Question about whether the text about assuming the nature of `aggregate`'s response is necessary, but LGTM either way.

jmikola:
Ah, one more question. [DRIVERS-1437](https://jira.mongodb.org/browse/DRIVERS-1437) includes a link to https://github.com/mongodb/specifications/commit/066df641d28ccf9409e1f69b11b1e39dfcd4fb9a, which added some skips to `crud-api-version-1-strict.yml` in the version API tests. Those skips reference DRIVERS-1437, so I assume they may be temporary and just depending on DRIVERS-1437 to be resolved.

If the purpose of DRIVERS-1437 is to allow `estimatedDocumentCount` to work with versioned API, should those skips now be removed? The file in question wasn't modified in #905, although new tests were added to `crud-api-version-1.yml`. This may be a question for @ShaneHarvey and/or @alcaeus.

If this needs to be addressed, I suppose you can do so in this PR. And you might as well rename it and its commits to refer to [SPEC-1780](https://jira.mongodb.org/browse/SPEC-1780) so the githook bot can pick them up in JIRA (now that the PR is becoming a bit more than a typo correction).

Lastly, once everything is wrapped up, please ask Esha/Alex about updating the description for DRIVERS-1437 to clearly instruct drivers where to look for spec changes and the most recent commit that should be used to sync their spec tests. That's typically done in smaller "spec change" tickets by just adding a paragraph to the issue description once the SPEC ticket is resolved, but DRIVERS-1437 is an epic with different formatting so I'm not sure if the TPMs may have something else in mind.

alcaeus:
Thanks for pointing this out @jmikola! Yes, the skips were indeed temporary until we have a solution for `estimatedDocumentCount`. Any driver correctly implementing this spec should also be able to pass the strict versioned API tests, so those skips should be removed.

Since those tests only run on 4.9+, we don't need to have two separate tests as is done in the CRUD tests, but the command expectation there will need adjusting.

source/crud/crud.rst
jyemin: > Why is this a concern?

It's a concern because the most natural way to implement estimatedDocumentCount (and the way both Benji and I did it without really thinking about the ramifications) is to check out a connection, check the wire version, and then send either a count or an aggregate command.  But once you go down that road, it's actually quite challenging from a design perspective to fully iterate any cursor returned in the reply to aggregate, and neither of our implementations attempted to do so.  countDocuments is different because there is no wire version check required.

> Is there any conceivable reason to expect that the single result document would not be in the immediate command response to aggregate?

Yes, because there is nothing documented anywhere that requires a server to return any results in the first batch.  The fact that it does so could be considered an implementation detail of MongoDB.  That's why we want someone from server query team to sign off on this assumption.

dstorch (approved):
This LGTM. I think for non-tailable cursors, the server should guarantee that it returns a non-empty batch, with the exception of a possibly empty final batch. Note that no such guarantee exists for tailable cursors.

CC @cswanson310 -- Do you know if we added a test for this scenario under https://jira.mongodb.org/browse/PM-2031?

benjirewis:
As discussed offline, we still can't run the `estimatedDocumentCount` test for versioned API, as `$collStats` (along with `count`) is also not a part of API version 1. This seems like an oversight from the server, so we'll have to wait for the resolution of [SERVER-54470](https://jira.mongodb.org/browse/SERVER-54470). I've filed [DRIVERS-1561](https://jira.mongodb.org/browse/DRIVERS-1561) to track the server change, and updated `skipReason` of the test to reference this driver ticket. I also modified the test to use expect the new aggregate and run on only 4.9+, so we'll just have to remove the `skipReason` eventually.

cswanson310:
@dstorch I would amend that guarantee to exclude usages of `batchSize: 0`. Those cursors are expected to return an empty first batch. Otherwise looks good. 

It looks like `{$collStats: {count: {}}` in apiVersion 1 was a fairly blatant oversight of https://jira.mongodb.org/browse/PM-2031. https://jira.mongodb.org/browse/SERVER-54470 is helpful to track that work. Is that what you meant by "this scenario" @dstorch ?

dstorch:
@cswanson310 ah yeah, good point about `batchSize:0`. https://jira.mongodb.org/browse/SERVER-54470 is another good catch!

By "this scenario", I was wondering whether our tests for fast counts expressed as $collStats use the cursor API and therefore obscure whether or not the lone result document might appear in a getMore. It would be better for us to test by issuing an aggregate command directly, circumventing the cursor protocol, and asserting that we get back a non-empty batch. That way we have a server test which proves that the assumption made by the drivers is valid.

cswanson310:
Oh! Great idea. I added a comment to SERVER-54470, I think we can add this case there.

dstorch:
Sounds good.

source/crud/crud.rst
jmikola: Your response made me wonder if there's a lingering concern for retryable reads, and possibly needing to recreate the command on a second attempt if the wire version changes. Is that outstanding, or would you say it's already covered by [3b. Sending an equivalent command for the second attempt](https://github.com/mongodb/specifications/blob/master/source/retryable-reads/retryable-reads.rst#b-sending-an-equivalent-command-for-the-second-attempt)?

jmikola (approved):

PR #911, opened 2021-02-17
SPEC-1759 Clarify that SDAM should ignore the writeErrors field

isabelatkinson:


source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1340,6 +1340,8 @@ Additionally, if the response includes a write concern error, then the code
 and message of the write concern error MUST be checked the same way a response
 error is checked above.
 
+Errors contained within the writeErrors field MUST NOT be checked.
```
ShaneHarvey, line 1343: Is it possible to test this to make it easier for the team to validate each driver's behavior? Perhaps we need https://jira.mongodb.org/browse/SERVER-51297?

divjotarora: If we don't want to block this ticket on the server, would it be possible to do this in the existing SDAM error handling unit tests? For example, I see [this phase](https://github.com/mongodb/specifications/blob/2ce041277c07b2f67b0dff73f1b04fa7e10f2af8/source/server-discovery-and-monitoring/tests/errors/stale-generation-beforeHandshakeCompletes-NotMasterNoSlaveOk.yml#L73) in a test specifies a `NotMasterNoSlaveOk` error in an `{ok: 0}` server response. Could we add another test that instead has `{ok: 1, writeErrors: [ <error doc for NotMasterNoSlaveOk> ]}`?

ShaneHarvey: Great idea. Yes I think it makes sense to add a test like that in the `errors` directory. Note the test format for "applicationErrors" is described here: https://github.com/mongodb/specifications/tree/master/source/server-discovery-and-monitoring/tests#format

isabelatkinson: Added a test to the `errors` directory. Would one of you be able to test this against your driver? We don't have a runner for this test format implemented in the Rust driver.

source/server-discovery-and-monitoring/tests/errors/write_errors_ignored.yml
```diff
@@ -0,0 +1,41 @@
+description: writeError field is ignored
```
divjotarora, line 1: nit: `writeError` -> `writeErrors`

source/server-discovery-and-monitoring/tests/errors/write_errors_ignored.yml
```diff
@@ -0,0 +1,41 @@
...
+- description: Ignore command error with writeErrors field
+  applicationErrors:
+  - address: a:27017
+    when: afterHandshakeCompletes
+    maxWireVersion: 9
+    type: command
+    response:
+      ok: 1
+      writeErrors:
+      - { ok: 0, errmsg: NotMasterNoSlaveOk, code: 13435 }
```
divjotarora, line 40: This shouldn't matter much, but I don't think individual documents in the `writeErrors` array have `ok` fields.

divjotarora (approved):
New test passes for Go. Couple of minor comments, but no need for another round of review.

ShaneHarvey (approved):
LGTM! This test actually worked to fix this bug in pymongo. We actually do mark a server unknown if when there's a not primary error in the writeErrors array, which is a bug: https://github.com/mongodb/mongo-python-driver/pull/570

ShaneHarvey (approved):
Still LGTM

nbbeeken (approved):
LGTM

PR #916, opened 2021-03-01
DRIVERS-828 Update spec changelog and schema-1.2

jmikola:
2ce041277c07b2f67b0dff73f1b04fa7e10f2af8 (#903) missed updating the spec version and changelog for storeEventsAsEntities and loop. Additionally, storeEventsAsEntities required a schema update and corresponding tests.

### DIFF
--- a/source/unified-test-format/tests/valid-fail/entity-client-storeEventsAsEntity-conflict_with_client_id.yml
+++ b/source/unified-test-format/tests/valid-fail/entity-client-storeEventsAsEntity-conflict_with_client_id.yml
@@ -1,0 +1,15 @@
+description: "entity-client-storeEventsAsEntity-conflict_with_client_id"
+
+schemaVersion: "1.2"
+
+createEntities:
+  - client:
+      id: &client0 client0
+      storeEventsAsEntities:
+        # Using the client ID here will ensure that test runners also detect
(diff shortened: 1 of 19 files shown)


alcaeus (approved):
Changes look good. Should we add a `valid-pass` as well to have a working example and prevent against regressions?

jmikola:
> Should we add a valid-pass as well to have a working example and prevent against regressions?

Can do. There's probably not a reliable way to test that events are written to an entity short of abusing [`$$matchesEntity`](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#matchesentity), but a passing test is still worthwhile.

source/unified-test-format/tests/valid-fail/entity-client-storeEventsAsEntity-conflict_with_client_id.yml
```diff
@@ -0,0 +1,15 @@
+description: "entity-client-storeEventsAsEntity-conflict_with_client_id"
+
+schemaVersion: "1.1"
```
jmikola, line 3: This is incorrect copypasta. I'll fix this.

alcaeus (approved):

p-mongo (approved):

PR #920, opened 2021-03-03
SPEC-1056 Drivers should not inspect error messages for "not master" or "node is recovering"

ShaneHarvey:
Stop inspecting error messages for "not master" or "node is recovering"
when the error includes a code.
Add 10058 as a "not master" error code to account for MongoDB<=3.2 errors.


source/server-discovery-and-monitoring/tests/errors/generate-error-tests.py
```diff
@@ -40,6 +40,7 @@ def write_test(filename, data):
     'ShutdownInProgress': (91,),
     'NotMaster': (10107,),
     'NotMasterNoSlaveOk': (13435,),
+    'LegacyNotPrimary': (10058,),
```
ShaneHarvey, line 43: All of the spec test changes except for `prefer-error-code.yml` are from the two changes in this file.

source/server-discovery-and-monitoring/tests/errors/generate-error-tests.py
```diff
@@ -129,6 +130,7 @@ def create_stale_generation_tests():
     # Stale network errors
     for network_error_type, when in itertools.product(
             ['network', 'timeout'], WHEN):
+        error_name = network_error_type
```
ShaneHarvey, line 133: This is a minor fix for the `stale-generation-beforeHandshakeCompletes-timeout.yml`-like tests. It only changes the descriptions in the tests and doesn't change any behavior.

source/server-discovery-and-monitoring/tests/errors/prefer-error-code.yml
```diff
@@ -0,0 +1,53 @@
+description: Do not check errmsg when code exists
```
ShaneHarvey, line 1: This is the one test that ensure that errmsg strings are no longer checked when the code field exists.

nbbeeken (approved):
LGTM, Node fails the new LegacyNotPrimary tests but I imagine thats expected, we need to add a check for that error code.

PR #921, opened 2021-03-04
DRIVERS-828 Use array of objects for storeEventsAsEntities instead of map

jmikola:
This makes the schema definition for intuitive and is more consistent with other parts of the test format where dynamic strings are specified in values rather than keys.

Related to https://github.com/mongodb-labs/drivers-atlas-testing/pull/105


alcaeus (approved):

DmitryLukyanov (approved):
I didn't run the provided tests, so I assume they passed. LGTM

p-mongo (approved):

PR #926, opened 2021-03-05
DRIVERS-1595 Verify that abortTransaction does not include Versioned API options

alcaeus:
DRIVERS-1595

Verified the test locally, but it's almost an exact copy of the `commitTransaction` test above so I skipped the evergreen patch build.


jyemin (approved):
LGTM, though I haven't run it myself.

PR #927, opened 2021-03-05
DRIVERS-1594 Only run db.aggregate Versioned API test on 4.9

alcaeus:
DRIVERS-1594

This follows the work done by @benjirewis in #898 to expect a strict error when running `db.aggregate`. This is only true for 4.9 and newer, so this PR splits the test into two tests for 4.7 and 4.9.

### DIFF
--- a/source/versioned-api/tests/crud-api-version-1-strict.json
+++ b/source/versioned-api/tests/crud-api-version-1-strict.json
@@ -140,7 +140,63 @@
       ]
     },
     {
-      "description": "aggregate on database appends declared API version",
+      "description": "aggregate on database appends declared API version on servers before 4.9.0",
+      "runOnRequirements": [
+        {
+          "maxServerVersion": "4.8.99"
+        }
+      ],
+      "operations": [
+        {
+          "name": "aggregate",
+          "object": "adminDatabase",
+          "arguments": {
+            "pipeline": [
+              {
+                "$listLocalSessions": {}
+              },
+              {
+                "$limit": 1
+              }
+            ]
+          }
+        }
+      ],
+      "expectEvents": [
+        {
+          "client": "client",
+          "events": [
+            {
+              "commandStartedEvent": {
+                "command": {
+                  "aggregate": 1,
+                  "pipeline": [
+                    {
+                      "$listLocalSessions": {}
+                    },
+                    {
+                      "$limit": 1
+                    }
+                  ],
+                  "apiVersion": "1",
+                  "apiStrict": true,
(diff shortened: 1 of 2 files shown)


source/versioned-api/tests/crud-api-version-1-strict.yml
```diff
@@ -61,7 +61,28 @@ tests:
                 pipeline: *pipeline
                 <<: *expectedApiVersion
 
-  - description: "aggregate on database appends declared API version"
+  - description: "aggregate on database appends declared API version on servers before 4.9.0"
```
jyemin, line 64: I don't understand why we need this test.  When would any driver be executing it, given that we all use "latest" (which I think is 4.9)?

alcaeus: This came up when @p-mongo started testing on 4.7.0. Not sure if we explicitly need the test, I'm personally fine restricting the other test to 4.9 and newer.

jyemin: That's what i would prefer.  We shouldn't need to specify behavior for internal releases where the behavior changes in the next internal release.

alcaeus: I removed the extra test, so the only change that remains is restricting the existing test to 4.9.0 and above.

benjirewis (approved):
LGTM—thanks for the fix! I agree that the split was not needed.

source/versioned-api/tests/crud-api-version-1-strict.json
```diff
@@ -141,6 +141,11 @@
     },
     {
       "description": "aggregate on database appends declared API version",
+      "runOnRequirements": [
+        {
+          "minServerVersion": "4.9.0"
```
jyemin, line 146: I would just change minServerVersion to 4.9.0 on line 6 and remove this.  There is no reason any of these tests need to run on 4.7.0 anymore.

alcaeus: If you don't mind I'd merge this as-is until we've sorted that topic.

p-mongo (approved):

source/versioned-api/tests/crud-api-version-1-strict.json
jyemin: It's ok, but is there a rush to get this done?  I'd rather avoid churn.

alcaeus:
Closing in favour of #929 which drops testing on 4.7 altogether.

PR #928, opened 2021-03-05
Clarify when API Version should be sent

alcaeus:
This clarifies when an API Version should be sent to the server. I've tried to word this in a way that conveys that this only applies to commands sent through `OP_MSG` without placing too much emphasis on the wire protocol.

### DIFF
--- a/source/versioned-api/versioned-api.rst
+++ b/source/versioned-api/versioned-api.rst
@@ -135,12 +135,18 @@ rationale for more details.
 Sending Declared API Version to the Server
 ------------------------------------------
 
-The declared API version MUST be sent to the server with every command, with the
-exception of the cases listed below. Drivers MUST NOT use a server's reported
-``maxWireVersion`` to decide whether it supports the versioned API. The server
-will reply with an error if the declared API version is not supported. If the
+The declared API version MUST be sent to the server as part of every command
+request, with the exception of the cases listed below. Drivers MUST NOT use a
+server's reported ``maxWireVersion`` to decide whether it supports the versioned
+API. The server will reply with an error if the declared API version is not
+supported, or if the command does not support API versioning options. If the
 user does not declare an API version, the driver MUST NOT send any API
 versioning options to the server.
+
+This requirement applies to all command requests, regardless of whether they are
+sent using ``OP_MSG`` or ``OP_QUERY`` (e.g. during the initial handshake). For
+all other opcodes, including pre-command usage of ``OP_QUERY``, drivers MUST NOT
+attempt to send API version parameters to the server.
 
 
 Command Syntax


p-mongo (approved):

source/versioned-api/versioned-api.rst
```diff
@@ -135,12 +135,15 @@ rationale for more details.
...
-versioning options to the server.
+The declared API version MUST be sent to any server that supports ``OP_MSG``
+as part of the command request, with the exception of the cases listed below.
+Drivers MUST NOT use a server's reported ``maxWireVersion`` to decide whether
+it supports the versioned API. The server will reply with an error if the
+declared API version is not supported. If the user does not declare an API
+version, the driver MUST NOT send any API versioning options to the server.
+Note that per the ``OP_MSG`` spec the driver is required to use ``OP_MSG``
+when communicating with an ``OP_MSG`` capable server. Versioned API options
+MUST NOT be sent to pre-``OP_MSG`` servers.
```
ShaneHarvey, line 146: > Versioned API options MUST NOT be sent to pre-``OP_MSG`` servers.

So does this mean that the ServerApi option is ignored when talking to <=3.4 servers?

p-mongo: I can see how this is still potentially confusing. The intention of the PR was to not require any changes when OP_MSG is not used. Are you sending server API parameters using legacy protocol? If yes how do you send them in e.g. OP_QUERY?

ShaneHarvey: > Are you sending server API parameters using legacy protocol?

Yes. The OP_QUERY format supports sending arbitrary command arguments in the same way that OP_MSG does (the runCommand helper works with OP_QUERY after all).  But now that you mention it, non-command based protocols do not support these arguments at all, like OP_INSERT or a find() on a <=3.0 server using OP_QUERY (it's a bit confusing to talk because OP_QUERY has 2 uses). Now I understand what this is trying to clarify.

I am under the impression that we want every operation on <=4.4 to raise an error when versioned api configured. Personally I think we should do DRIVERS-1484 and ban sending apiVersion to <=4.4 servers altogether. Then all non-OP_MSG would be banned (except of course the isMaster handshake command which still uses OP_QUERY).

p-mongo: So, when pymongo is connected to a 2.6 server, does it send api version with every command as the spec requires?

ShaneHarvey: Every command, yes. But pymongo won't send it with non-command messages like OP_INSERT or find on <=3.0.

p-mongo: Okay, I see. I haven't been around when the legacy op codes were used so I was asking around how they are supposed to be handled and as far as I know there wasn't a helpful list of which op codes should have api parameters added to them and which shouldn't.

If the idea here is to have every driver figure out how the legacy op codes work independently I'll work on that, otherwise the thought with this PR was to exempt pre-3.6 servers (and thus legacy op codes) from needing to be handled by the drivers.

alcaeus:
@p-mongo @ShaneHarvey: After discussion with @jyemin, I've changed the wording a bit. I've now clarified that the API versioning options apply to all commands, regardless of how they're sent to the server (e.g. any command sent to the server using `OP_QUERY` should contain API versioning options).

I'll note that the sentence of not checking `maxWireVersion` may change based on DRIVERS-1484, but for the time being this clarification should help reduce confusion as to when and how to send an API version. Please let me know if it doesn't.

source/versioned-api/versioned-api.rst
```diff
@@ -135,13 +135,19 @@ rationale for more details.
...
+The declared API version MUST be sent to the server as part of every command
+request, with the exception of the cases listed below. Drivers MUST NOT use a
+server's reported ``maxWireVersion`` to decide whether it supports the versioned
+API. The server will reply with an error if the declared API version is not
+supported, or if the command does not support API versioning options. If the
 user does not declare an API version, the driver MUST NOT send any API
 versioning options to the server.
 
+This requirement applies to all command requests, regardless of whether they are
+sent using ``OP_MSG`` or ``OP_QUERY`` (e.g. during the initial handshake). For
```
p-mongo, line 147: This reads to me as no action is needed for any opcodes other than these two. If that is the intention, I approve the PR.

p-mongo (approved):

source/versioned-api/versioned-api.rst
alcaeus: That is correct - the change only affects command documents, not wire protocol messages.

ShaneHarvey: Thank you for clarifying.

ShaneHarvey (approved):

PR #929, opened 2021-03-09
Use 4.9 as minServerVersion for versioned API tests

alcaeus:
After some discussion on whether versioned API tests should even run on 4.7, we've come to the conclusion that there's no point testing 4.7 as it was mainly an internal test release. This PR changes the `minServerVersion` requirement for all tests to 4.9 and drops testing `estimatedDocumentCount` with the `count` command on 4.7.


jyemin (approved):

benjirewis (approved):
LGTM! Thanks 😄

PR #930, opened 2021-03-09
Regenerate JSON test files

alcaeus:
This commit regenerates syncs JSON test files to avoid errors during CI runs. The process was as follows:
* Re-generate all automatically regenerated test files:
```
python3 ./source/server-discovery-and-monitoring/tests/errors/generate-error-tests.py
python3 ./source/client-side-encryption/etc/generate-corpus.py ./source/client-side-encryption/corpus
python3 ./source/client-side-encryption/etc/generate-test.py ./source/client-side-encryption/etc/test-templates/*.template ./source/client-side-encryption/tests
```
* Touch all YAML files to force regeneration of JSON files:
```
find . -iname '*.yml' | xargs touch
```
* Run make to regenerate all files:
```
cd source && make
```

During this process, I found that a usage of `$$unsetOrMatches` was inconsistent between JSON and YAML files. I've cleaned up this issue in a separate commit.

@benjirewis looks like a chunk of the diff is due to different indentation levels. Did you by chance write the JSON files yourself or were those generated from YAML?


benjirewis (approved):
Thanks for syncing everything! Yes, I did write the JSON myself instead of generating it from YAML; so that's why the indentation is totally off 🤦 . Will just generate from YAML in the future haha.

PR #931, opened 2021-03-09
DRIVERS-1591 clarify comparison behaviour for version strings

alcaeus:
DRIVERS-1591

### DIFF
--- a/source/unified-test-format/unified-test-format.rst
+++ b/source/unified-test-format/unified-test-format.rst
@@ -479,9 +479,9 @@ The structure of this object is as follows:
 
     Test files SHOULD NOT use this option unless one or more command monitoring
     events are specified in `observeEvents <entity_client_observeEvents_>`_.
-    
+
   .. _entity_client_storeEventsAsEntities:
-  
+
   - ``storeEventsAsEntities``: Optional array of one or more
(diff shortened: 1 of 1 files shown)


source/unified-test-format/unified-test-format.rst
```diff
@@ -1044,6 +1044,12 @@ non-negative integer:
 - ``<major>.<minor>`` (``<patch>`` is assumed to be zero)
 - ``<major>`` (``<minor>`` and ``<patch>`` are assumed to be zero)
 
+Any component other than ``major``, ``minor``, and ``patch`` MUST be discarded
```
jyemin, line 1047: Consider whether to be more proscriptive of how the actual server version is obtained.  If the spec says explicitly that it's obtained by taking the first three elements of the `versionArray` field in the `buildInfo` reply, we can avoid any mention of discarding other components.

DmitryLukyanov (approved):
LGTM

source/unified-test-format/unified-test-format.rst
alcaeus: I've expanded the wording to explain how to obtain this information from the `versionArray` field.

jyemin (approved):
LGTM

PR #933, opened 2021-03-18
DRIVERS-1561 Remove skipReason on edc versioned API test

benjirewis:
[DRIVERS-1561](https://jira.mongodb.org/browse/DRIVERS-1561)

Removes skipReason on estimatedDocumentCount test for versioned API. This test was being skipped because the $collStats aggregate pipeline stage was not a part of API version 1, but it was just added in [SERVER-54470](https://jira.mongodb.org/browse/SERVER-54470).


alcaeus (approved):
LGTM! Drivers should be able to make this change right away, as we only test versioned API on latest which by now should include the relevant commit.

PR #935, opened 2021-03-19
DRIVERS-720 Add unified tests for load balancer support

divjotarora:
This PR includes spec tests for load balancer support written in the unified test format as well as additive changes to the format itself.

Notes on setting up the relevant servers to run the tests:

1. Start a 4.9 ("latest") sharded cluster with two mongoses and authentication enabled.
2. Start two LBs: one fronting a single mongos and one fronting both mongos's. The test runner should choose the single-mongos LB for clients where `useMultiMongoses` is explicitly set to `false` and the multi-mongos LB if it is unset or explicitly set to `true`.
3. The cluster URI passed to the test runner should be `mongodb://<single-mongos LB address>/?loadBalanced=true`.
4. The test runner must mock server support for returning `serverId` in `isMaster` responses. This can be done by using `topologyVersion.processId` as `serverId`. The mock return should be enabled for all MongoClient instances, including the internal client, if the target cluster is behind a load balancer (i.e. if the cluster URI contains `loadBalanced=true`).

Tests that are expected to fail or are otherwise noteworthy have been called out in PR comments.


source/load-balancers/tests/lb-connection-establishment.yml
```diff
@@ -0,0 +1,28 @@
...
+      uriOptions:
+        # Explicitly set loadBalanced to false to override the option from the global URI.
+        loadBalanced: false
+  - database:
+      id: &database0 database0
+      client: *client0
+      databaseName: &database0Name database0
+
+tests:
+  - description: operations against load balancers fail if URI contains loadBalanced=false
```
divjotarora, line 20: This test should fail. It expects the server to fail the connection handshake because it will not contain `loadBalanced: true`, but support for this has not been implemented in the server yet.

source/load-balancers/tests/sdam-error-handling.yml
```diff
@@ -0,0 +1,192 @@
...
+          - connectionCheckedInEvent: {}
+          - connectionClosedEvent:
+              reason: stale
+          # Second insertOne.
+          - connectionCheckedOutEvent: {}
+          - connectionCheckedInEvent: {}
+
+  # This test uses singleClient to ensure that connection attempts are routed
+  # to the same mongos on which the failpoint is set.
+  - description: errors during the initial connection hello are ignored
```
divjotarora, line 135: This test should fail when running against a 4.4 server. It needs to run against a latest server due to a failCommand bug (SERVER-49220) but cannot due to SERVER-55211.

source/load-balancers/tests/wait-queue-timeouts.yml
```diff
@@ -0,0 +1,77 @@
+description: wait queue timeout errors include details about checked out connections
```
divjotarora, line 1: Both tests in this file can't be run in Go because we don't support the `waitQueueTimeoutMS` option. I added some code to mock support for this option to validate that these can pass when I wrote them, but it's since been removed. It would be very helpful if another driver could validate them.

source/unified-test-format/unified-test-format.rst
```diff
@@ -408,7 +439,13 @@ The structure of this object is as follows:
...
 .. _entity_client:
 
-- ``client``: Optional object. Defines a MongoClient object.
+- ``client``: Optional object. Defines a MongoClient object. In addition to
+  the configuration defined below, test runners MUST track the number of
+  connections checked out at any given time for the constructed MongoClient.
+  This can be done using a single counter and
+  `CMAP events <../connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#events>`__.
+  Each ``ConnectionCheckedOutEvent`` should increment the counter and each
+  ``ConnectionCheckedInEvent`` should decrement it.
```
jmikola, line 448: I would propose putting the connection-tracking stuff in its own paragraph to preserve the one-line explanation for readability.

Should we clarify that this only applies to drivers that support connection pools? Additionally, I would assume it's only necessary for tests that include one or more `assertNumberConnectionsCheckedOut` operations.

source/unified-test-format/unified-test-format.rst
```diff
@@ -3,7 +3,7 @@ Unified Test Format
 ===================
 
 :Spec Title: Unified Test Format
-:Spec Version: 1.2.0
+:Spec Version: 1.3.0
```
jmikola, line 6: Don't forget to bump the last modified date as well.

source/unified-test-format/schema-1.3.json
```diff
@@ -263,6 +348,7 @@
         "name": { "type": "string" },
         "object": { "type": "string" },
         "arguments": { "type": "object" },
+        "ignoreResultAndError": { "type": "boolean" },
```
jmikola, line 351: Since this is meant to be mutually exclusive with `expectError` and `expectResult`, I would suggest adding additional rules to the `allOf` definition for `operation`. See: https://github.com/mongodb/specifications/blob/master/source/unified-test-format/schema-1.2.json#L270

IMO, this would be a fruitful learning exercise in writing the JSON schema, since everyone is likely to make changes to the test format down the line. This would also allow you to add tests to the test runner itself to ensure your schema changes are correct.

See https://github.com/mongodb/specifications/commit/62c6bb9ed030e91cdab8622299bb1bf970b7f68d#diff-25ec19412b8ee6ccb0d93ebcd467ba7958e722164db42c5fbcc9e38de3ae2fd8 for prior art.

source/unified-test-format/schema-1.3.json
```diff
@@ -213,23 +234,87 @@
...
+        },
+        "connectionCheckedOutEvent": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {}
+        },
+        "connectionCheckedInEvent": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {}
```
jmikola, line 317: Related to my comment about `ignoreResultAndError` below, I think any schema changes should be accompanied by extra tests. See https://github.com/mongodb/specifications/commit/62c6bb9ed030e91cdab8622299bb1bf970b7f68d#diff-25ec19412b8ee6ccb0d93ebcd467ba7958e722164db42c5fbcc9e38de3ae2fd8.

This is tedious, but I think it's as worthwhile as spec tests themselves.

source/load-balancers/tests/non-lb-connection-establishment.yml
```diff
@@ -0,0 +1,52 @@
+description: connection establishment if loadBalanced is specified for non-load balanced clusters
```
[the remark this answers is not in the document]
divjotarora: Yes, the main concern was that every current and future test file would need to specify `loadBalanced: false`. Let me know what you think the best way forward is here.

source/load-balancers/tests/sdam-error-handling.yml
```diff
@@ -0,0 +1,192 @@
+description: state change errors are correctly handled
+
+schemaVersion: '1.3'
+
+runOnRequirements:
+  - loadBalanced: true
+
+_yamlAnchors:
+  observedEvents:
+    - &observedEvents
```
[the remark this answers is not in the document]
divjotarora: In the JSON, I see
```
{
      "client": {
        "id": "singleClient",
        "useMultipleMongoses": false,
        "uriOptions": {
          "appname": "lbSDAMErrorTestClient",
          "retryWrites": false
        },
        "observeEvents": [
          "connectionCreatedEvent",
          "connectionReadyEvent",
          "connectionCheckedOutEvent",
          "connectionCheckOutFailedEvent",
          "connectionCheckedInEvent",
          "connectionClosedEvent",
          "poolClearedEvent"
        ]
      }
    },
```

Is this not correct?

source/load-balancers/tests/sdam-error-handling.yml
```diff
@@ -0,0 +1,192 @@
...
+          filter: {}
+          batchSize: 2
+        saveResultAsEntity: &cursor1 cursor1
+      # Close both cursors to return the connections to the pool.
+      - name: close
+        object: *cursor0
+      - name: close
+        object: *cursor1
+      # Fail an operation with a state change error.
+      - name: failPoint
```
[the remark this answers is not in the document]
divjotarora: It's not a concern because only 5.0+ servers will support being behind an LB.

source/load-balancers/tests/server-selection.yml
```diff
@@ -0,0 +1,49 @@
+description: server selection for load-balanced clusters
+
+schemaVersion: '1.3'
+
+runOnRequirements:
+  - loadBalanced: true
+
+createEntities:
+  - client:
```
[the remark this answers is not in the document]
divjotarora: It's an expectation that the cluster URI will include `loadBalanced=true` because this is needed for the internal client as well as all client entities. The only time an explicit `loadBalanced` URI option is needed is if it's intentionally being set to an incorrect value (e.g. being set to true for a cluster that's not behind an LB) to test error cases.

source/load-balancers/tests/transactions.yml
```diff
@@ -0,0 +1,447 @@
...
+          document: *insertDocument
+          session: *session0
+      - name: assertNumberConnectionsCheckedOut
+        object: testRunner
+        arguments:
+          client: *client0
+          connections: 1
+      - name: commitTransaction
+        object: *session0
+        ignoreResultAndError: true
```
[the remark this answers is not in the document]
divjotarora: It doesn't always succeed. When testing locally, I saw the retry fail with `NoSuchTransaction` if it targetted a different mongos than the one that serviced the other transaction commands. Additionally, when servers support LB mode, the server-side reaper thread will reap the transaction if it detects that the socket associated with the transaction has been disconnected. In this case, the retry races with the reaper thread and could potentially fail. We only care that the retry was sent, which is asserted in event expectations, not that it actually succeeded.

source/load-balancers/tests/transactions.yml
```diff
@@ -0,0 +1,447 @@
...
+      - name: assertNumberConnectionsCheckedOut
+        object: testRunner
+        arguments:
+          client: *client0
+          connections: 0
+    expectEvents:
+      - client: *client0
+        events:
```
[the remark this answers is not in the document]
divjotarora: I don't expect test runners to store multiple event types in a single array and correlate the timing. Instead, I expect APM events and CMAP events to be stored in separate arrays and assertions to be made separately. The reasoning behind writing out the expectations separately was just to make it easier to read. I can clarify this in the actual test format spec if you think that would be helpful.

source/load-balancers/tests/wait-queue-timeouts.yml
```diff
@@ -0,0 +1,77 @@
...
+        arguments:
+          filter: {}
+          batchSize: 2
+        saveResultAsEntity: &cursor0 cursor0
+      - name: insertOne
+        object: *collection0
+        arguments:
+          document: { x: 1 }
+        expectError:
+          errorContains: 'maxPoolSize: 1, connections in use by cursors: 1, connections in use by transactions: 0, connections in use by other operations: 0'
```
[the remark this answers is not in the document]
divjotarora: It's client-side and all drivers are mandated to use this format per the design. Updated the tests to include `isClientError`.

source/unified-test-format/unified-test-format.rst
```diff
@@ -946,6 +1011,92 @@ The structure of this object is as follows:
...
+
+- ``connectionCheckedInEvent``: Optional object. If present, this object
+  MUST be an empty document as all fields in this event are non-deterministic.
+
+
+hasServerId
+```````````
+
+This field is an optional boolean that specifies whether or not the ``serverId``
+field of an event is set. If true, test runners MUST assert that the field is
```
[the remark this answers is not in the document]
divjotarora: Unfortunately, that spec doesn't currently exist. I don't think it's necessary to reference any specs here as the list of supported events explicitly says which event types can include a `serverId`.

source/unified-test-format/unified-test-format.rst
```diff
@@ -946,6 +1011,92 @@ The structure of this object is as follows:
...
+set and is a non-empty BSON ObjectId (i.e. all values of the ObjectId are not 0).
```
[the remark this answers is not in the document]
divjotarora: I only made this distinction because a reasonable Go API could be something like

```
type CommandStartedEvent struct {
    ServerID primitive.ObjectID
}
```
In this case, an "unset" server ID would actually be an array of 12 0's. For languages that do something similar, I wanted to make sure they don't consider such an ID to be a set value.

Updated to "all bytes".

source/unified-test-format/unified-test-format.rst
```diff
@@ -946,6 +1011,92 @@ The structure of this object is as follows:
...
+
+.. _expectedEvent_poolClearedEvent:
+
+- ``poolClearedEvent``: Optional object. Assertions for one or more
+  `PoolClearedEvent <../connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#events>`__
+  fields.
+
+  The structure of this object is as follows:
+
+  - ``serverId``: Defined in `hasServerId`_.
```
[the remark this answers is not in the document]
divjotarora: That's the only event that uses it in the LB design doc right now. If that changes, I'll update tests to match.

source/unified-test-format/unified-test-format.rst
```diff
@@ -758,6 +807,18 @@ The structure of this object is as follows:
 
   The ``session`` parameter is handled specially (see `commonOptions_session`_).
 
+.. _operation_ignoreResultAndError:
+
+- ``ignoreResultAndError``: Optional boolean. If true, both the error and result
+  for the operation MUST be ignored.
```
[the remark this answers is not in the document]
divjotarora: Discussed offline. This PR also updates the "Executing an Operation" section to check `ignoreResultAndError` before checking `expectResult` or `expectError` so I don't think it's necessary.

source/load-balancers/tests/sdam-error-handling.yml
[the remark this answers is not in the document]
jmikola: Ah, that is indeed correct.

I was looking at the JSON for `_yamlAnchors`:

```json
  "_yamlAnchors": {
    "observedEvents": [
      [
        "connectionCreatedEvent",
        "connectionReadyEvent",
        "connectionCheckedOutEvent",
        "connectionCheckOutFailedEvent",
        "connectionCheckedInEvent",
        "connectionClosedEvent",
        "poolClearedEvent"
      ]
    ]
```

I'm aware that JSON is irrelevant but it looks odd to me.

Is the extra array nesting in the YAML actually necessary to resolve correctly? I would have thought this would work:

```yml
_yamlAnchors:
  observedEvents: &observedEvents
    - connectionCreatedEvent
    - connectionReadyEvent
    - connectionCheckedOutEvent
    - connectionCheckOutFailedEvent
    - connectionCheckedInEvent
    - connectionClosedEvent
    - poolClearedEvent
```

source/load-balancers/tests/transactions.yml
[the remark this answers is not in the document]
jmikola: That sounds like a great explanation and would be helpful to share in a comment above this line.

source/load-balancers/tests/sdam-error-handling.yml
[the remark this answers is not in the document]
divjotarora: That seems to work too. Updated in a new commit.

source/load-balancers/tests/transactions.yml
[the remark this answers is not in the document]
jmikola: This definitely needs to be documented.

Given that [`expectEvents`](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#test) is an array of one or more [`expectedEventsForClient`](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#expectedeventsforclient) objects, I think it'd be preferable to segregate these events into their own `expectedEventsForClient` objects.

Then, we can maintain language advising tests not to mix event types within `expectedEventsForClient` and suggest the approach above. I'm OK with removing the advice against mixing event types in `observeEvents`.

One last idea is that we may want to enforce the type of events within `expectedEventsForClient`. We could do that by introducing a new optional field (e.g. `type`), which is an enum ("command" and "cmap" for now, potentially "sdam" down the line) and defaults to "command" if omitted for BC. This would mean that drivers could never mix events from different specs in the same list of expectations, but I think that's reasonable since there is likely too much variation to ever standardize on that.

source/unified-test-format/unified-test-format.rst
```diff
@@ -2050,6 +2219,28 @@ An example of this operation follows::
               - _id: 3, x: 33
 
 
+assertNumberConnectionsCheckedOut
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The ``assertNumberConnectionsCheckedOut`` operation instructs the test runner
+to assert that the given number of connections are currently checked out for
+the specified client entity.
```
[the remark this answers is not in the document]
jmikola: Single-threaded drivers would have at most one socket per server, which is shared for application and monitoring usage. I'm not sure if SDAM requires sharing application/monitoring sockets, but it's what libmongoc does and that's the only actively maintained implementation with Perl gone. Streamable isMaster also doesn't apply.

source/unified-test-format/unified-test-format.rst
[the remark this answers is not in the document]
jmikola: OK. I had a feeling some implementation might be using an "empty" ObjectId as its unset value. That makes sense. I'm good with just changing this to refer to "bytes" then.

source/unified-test-format/schema-1.3.json
jmikola: As discussed in Slack, this can wait until other points in the PR are sorted out to reduce churn.

source/load-balancers/tests/non-lb-connection-establishment.yml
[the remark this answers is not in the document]
jmikola: Trying to wrap my head around this alongside your [other comment](https://github.com/mongodb/specifications/pull/935#discussion_r599835910) that explains a test environment using a load balancer would include `loadBalanced=true` in its cluster URI. If that's the case, couldn't most test files remain agnostic about the whether a load balancer is being used?

Consider an arbitrary CRUD test (e.g. update with hint). Is there any reason something like that could not run irrespective of a load balancer?

source/unified-test-format/schema-1.3.json
```diff
@@ -81,7 +87,22 @@
...
+                  "poolReadyEvent",
+                  "poolClearedEvent",
+                  "poolClosedEvent",
+                  "connectionCreatedEvent",
+                  "connectionReadyEvent",
+                  "connectionClosedEvent",
+                  "connectionCheckOutStartedEvent",
+                  "connectionCheckOutFailedEvent",
+                  "connectionCheckedOutEvent",
+                  "connectionCheckedInEvent"
```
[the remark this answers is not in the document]
p-mongo: The reason why I used capitalized names was because this is how the events were defined in the specs and how they were defined in the code (class names are capitalized). Hence it seemed to me that beginning with lower case would 1) require transformations throughout the test runner to convert back and forth to the actual event names used, and 2) would raise the issue of the names in tests not matching exactly the names defined in specifications.

What was the rationale for using lowercased names? Perhaps I am missing something.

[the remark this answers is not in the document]
divjotarora: There was no rationale on my end besides the fact that the existing events (`commandStartedEvent`, `commandSucceededEvent`, and `commandFailedEvent`) are lowercased.

source/load-balancers/tests/transactions.yml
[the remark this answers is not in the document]
divjotarora: Added.

source/unified-test-format/unified-test-format.rst
```diff
@@ -382,6 +404,15 @@ The structure of this object is as follows:
...
   occur when fetching a single parameter using ``getParameter``.
 
+- ``loadBalanced``: Optional boolean. Specifies whether or not the tests can
+  be executed against a load-balanced cluster. If true, the relevant tests MUST
+  only be run against a load balancer. If false or omitted, the tests MUST only
+  be run against non-load balanced clusters.
+
+- ``auth``: Optional boolean. Specifies whether or not authentication must be
+  enabled for the tests to run successfully. If this field is omitted, there is
+  no authentication requirement for the tests.
```
alcaeus, line 414: Reading it, I'd take this as:
* `auth` missing: always run test
* `auth: true`: only run test when auth is enabled
* `auth: false`: only run test when auth is disabled

Is that assumption correct, or does `auth: false` mean that the test does not require auth but may very well run against a cluster with authentication enabled?

source/unified-test-format/unified-test-format.rst
```diff
@@ -382,6 +404,15 @@ The structure of this object is as follows:
   treat the comparison as not equal and skip the test. This includes errors that
   occur when fetching a single parameter using ``getParameter``.
 
+- ``loadBalanced``: Optional boolean. Specifies whether or not the tests can
+  be executed against a load-balanced cluster. If true, the relevant tests MUST
+  only be run against a load balancer. If false or omitted, the tests MUST only
+  be run against non-load balanced clusters.
```
alcaeus, line 410: This echoes what Jeremy already said, but I'm concerned about the different handling of the `loadBalanced` flag compared to other options.

From what I gather, this means that any `runOnRequirements` entry implicitly has `loadBalanced: false` if it's not specified. However, what does this mean for tests where no requirements have been set? At the moment, these tests are always run, which according to the above should now change. If so, this must be clarified accordingly.

I understand that this may require a large number of test changes to exclude them from being run against load balanced setups, but unless there is good reason to not run a test (e.g. functionality not supported), I'd opt for running as many of our tests against a load balancer as well, removing the need for this different behaviour, the implicit `runOnRequirement`, or adding `loadBalanced: false` to all test files.

source/unified-test-format/schema-1.3.json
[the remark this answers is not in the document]
jmikola: I think we talked about this, but the original rationale for the unified test runner was that the names from the older test format were `snake_case` (see [CRUD v2](https://github.com/mongodb/specifications/tree/master/source/crud/tests#legacy-v2-test-format)). I swapped the underscores for capitalization, but left the first letter as-is.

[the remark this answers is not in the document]
p-mongo: In my opinion  starting with capital letter would be consistent with both specs and driver code for events. Starting with a small letter might be consistent with other (non-event) values. I think the first consistency is more important in this case but perhaps others disagree.

source/unified-test-format/unified-test-format.rst
jmikola: I would be in favor of adding `loadBalanced: false` as needed vs. the alternative of changing how `runOnRequirement` is evaluated.

Having some implicit requirement check seems a like a can of worms, since we'd also then need to specify that it applies only to the top-level `runOnRequirements` and the corresponding test-level field. Consider if we only specified "if `runOnRequirements` is omitted, test runners must ensure the cluster is not load balanced." Then, even if a test-level `runOnRequirements` had `loadBalanced: true`, the test might still be skipped because there was no `runOnRequirements` at the top of the file. The same could apply vice versa.

source/unified-test-format/unified-test-format.rst
jmikola: I'd be in favor of having `auth: false` require auth to be disabled. If that was the original intention of this text, then I think we can just improve it to clarify (as was done for `loadBalanced`).

source/load-balancers/tests/transactions.yml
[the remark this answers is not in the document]
jmikola: Not pushed yet?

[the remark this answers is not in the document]
divjotarora: Just pushed.

source/unified-test-format/unified-test-format.rst
divjotarora: Your understanding is correct and that was the original intention of the text. I updated the wording to clarify.

divjotarora (commented):
I updated the test format changes to remove the new ``loadBalanced`` requirement and instead use a new ``load-balanced`` topology type that fits into the existing ``topologies`` requirement.

@alcaeus I also resolved your comment thread regarding concerns about ``loadBalanced`` as it's no longer present.

@jmikola As discussed offline, I'm working improving the structure for ``expectedEventsForClient``.

source/load-balancers/tests/transactions.yml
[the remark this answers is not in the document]
divjotarora: I pushed a change to add a new `eventType` field and specify which events can be present based on the type.

jmikola:
Will `load-balanced` need to be differentiate between sharding and replica sets? If so, I'd propose making `load-balanced` agnostic for any type of load balanced cluster and then we can consider adding specific variants down the line (e.g. `load-balanced-replicaset`, `load-balanced-sharded`, `load-balanced-sharded-replicaset`).

divjotarora:
> Will we ever need to differentiate between load balanced sharded clusters and replica sets?

It will not. We only support mongos servers being behind a load balancer. Replica sets cannot be behind an LB because the LB wouldn't have knowledge of the primary/secondary architecture and therefore wouldn't be able to route requests to the correct node.

source/unified-test-format/unified-test-format.rst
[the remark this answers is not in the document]
divjotarora: In LB mode, there is only one server, which is the LB itself. I think this just means that single-threaded drivers will only have at most one socket at any given time and the concept of pinning connections is not necessary. I updated the operation text to state that it only applies to drivers that implement connection pooling and added a note to the LB design doc to make sure the final spec PR (not currently up) mentions single-threaded drivers.

source/unified-test-format/unified-test-format.rst
```diff
@@ -715,10 +772,10 @@ The structure of this object is as follows:
...
   in order to perform these assertions.
 
-  Test files SHOULD NOT expect events from multiple specs (e.g. command
-  monitoring *and* SDAM events) for a single client. See
-  `Mixing event types in observeEvents and expectEvents`_ for more
-  information.
+  Tests SHOULD NOT specify multiple `expectedEventsForClient`_ objects for a
+  single client entity with the same ``eventType`` field. For example, a test
+  containing two `expectedEventsForClient`_ objects with the ``eventType`` set
+  to ``cmap`` for both would be invalid.
```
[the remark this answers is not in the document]
divjotarora: Done.

source/unified-test-format/schema-1.3.json
divjotarora: Added three new clauses.

source/unified-test-format/schema-1.3.json
divjotarora: Added tests to `valid-pass`, `valid-fail`, and `invalid` in my new commits.

source/unified-test-format/unified-test-format.rst
alcaeus: Thank you for clarifying!

jmikola:
> Replica sets cannot be behind an LB because the LB wouldn't have knowledge of the primary/secondary architecture and therefore wouldn't be able to route requests to the correct node.

What about differentiating single mongod shards vs. replica set shards? IIRC, that's the reason we differentiate between `sharded` and `sharded-replicaset`, since retryable writes and transactions always require replication.

divjotarora:
> What about differentiating between the LB proxying standalone shards vs. replica set shards?

Per [MongoDB docs](https://docs.mongodb.com/manual/tutorial/convert-shard-standalone-to-shard-replica-set/), versions 3.6+ only support replica set shards, so this won't be a concern for LB support. If we distinguished between standalone and RS shards due to retryability and transactions, we might be able to remove that distinction altogether as those are both 3.6+ features as well.

alcaeus:
> Per [MongoDB docs](https://docs.mongodb.com/manual/tutorial/convert-shard-standalone-to-shard-replica-set/), versions 3.6+ only support replica set shards

Just a heads up that I'm able to start a sharded cluster backed by standalone servers in 4.4.3 with [this configuration](https://github.com/alcaeus/dotfiles/blob/ced7c7647afc158aa3af1064bdaac67d311f1e32/mongo-orchestration/configurations/sharded_clusters/cluster.json#L41..L54)
```
MongoDB Enterprise mongos> db.adminCommand({buildInfo: 1})
{
	"version" : "4.4.3",
	"gitVersion" : "913d6b62acfbb344dde1b116f4161360acd8fd13",
	"modules" : [
		"enterprise"
	],
	"allocator" : "system",
	"javascriptEngine" : "mozjs",
	"sysInfo" : "deprecated",
	"versionArray" : [
		4,
		4,
		3,
		0
	],
	"openssl" : {
		"running" : "Apple Secure Transport"
	}
}
MongoDB Enterprise mongos> sh.status()
--- Sharding Status --- 
  sharding version: {
  	"_id" : 1,
  	"minCompatibleVersion" : 5,
  	"currentVersion" : 6,
  	"clusterId" : ObjectId("605d99c2bc817aa19e8c4202")
  }
  shards:
        {  "_id" : "sh01",  "host" : "localhost:4100",  "state" : 1 }
        {  "_id" : "sh02",  "host" : "localhost:4200",  "state" : 1 }
  active mongoses:
        "4.4.3" : 2
  autosplit:
        Currently enabled: yes
  balancer:
        Currently enabled:  yes
        Currently running:  no
        Failed balancer rounds in last 5 attempts:  0
        Migration Results for the last 24 hours: 
                No recent migrations
  databases:
        {  "_id" : "config",  "primary" : "config",  "partitioned" : true }
```

I didn't investigate any further whether this should still be possible, and I also didn't attempt to actually shard data on this cluster, but at least starting such a cluster is not a problem.

source/load-balancers/tests/transactions.yml
```diff
@@ -0,0 +1,473 @@
...
+          - connectionCheckedOutEvent: {}
+          - connectionCheckedInEvent: {}
+          - connectionClosedEvent:
+              reason: error
+          # Events for the abortTransaction attempt.
+          - connectionReadyEvent: {}
+          - connectionCheckedOutEvent: {}
+          - connectionCheckedInEvent: {}
+
+  # This is similar to the previous test, but it asserts that the connection pinned by the transaction is returned to
```
jyemin, line 382: I'm not sure that this is a reasonable thing for drivers to do.  I think we should consider whether we will require the cursor to be closed in order for it to unpin its connection.

If we decide that this behavior is what's desired, it's going to be tricky to implement in drivers that are using reference counting idioms to track when to return a connection to the pool.  Java driver certainly has this problem.

For now, I was able to get this test passing by adding a `close() operation after the failed `updateOne`.

jyemin (commented):
Initial comment on one of the transactions tests.

source/unified-test-format/schema-1.3.json
```diff
@@ -52,13 +52,14 @@
...
-            "enum": ["single", "replicaset", "sharded", "sharded-replicaset"]
+            "enum": ["single", "replicaset", "sharded", "sharded-replicaset", "load-balanced"]
           }
         },
         "serverParameters": {
           "type": "object",
           "minProperties": 1
-        }
+        },
+        "auth": { "type": "boolean" }
```
jmikola, line 62: New `invalid` test: `entity-client-auth-type.yml`

source/unified-test-format/schema-1.3.json
```diff
@@ -194,14 +210,47 @@
       "required": ["client", "events"],
       "properties": {
         "client": { "type": "string" },
-        "events": {
-          "type": "array",
-          "items": { "$ref": "#/definitions/expectedEvent" }
+        "eventType": {
```
jmikola, line 213: New `invalid` tests:

 * `expectedEventsForClient-eventType-enum.yml` with invalid string value
 * `expectedEventsForClient-eventType-type.yml` with invalid type

source/unified-test-format/schema-1.3.json
```diff
@@ -194,14 +210,47 @@
...
+          "type": "string",
+          "enum": ["command", "cmap"]
+        },
+        "events": { "type": "array" }
+      },
+      "oneOf": [
```
jmikola, line 219: New `invalid` tests:

* `expectedEventsForClient-events_conflicts_with_default_eventType.yml` would have cmap events with `eventType` omitted
* `expectedEventsForClient-events_conflicts_with_cmap_eventType.yml` would have command events with "cmap" `eventType`
* `expectedEventsForClient-events_conflicts_with_command_eventType.yml` would have cmap events with "command" `eventType`

source/unified-test-format/schema-1.3.json
```diff
@@ -194,14 +210,47 @@
...
+              "items": { "$ref": "#/definitions/expectedCommandEvent" }
+            }
+          }
         }
-      }
+      ]
     },
 
-    "expectedEvent": {
+    "expectedCommandEvent": {
```
jmikola, line 253: Please rename all of the `expectedEvent` files in `invalid` to `expectedCommandEvent`.

source/unified-test-format/schema-1.3.json
```diff
@@ -213,27 +262,100 @@
           "properties": {
             "command": { "type": "object" },
             "commandName": { "type": "string" },
-            "databaseName": { "type": "string" }
+            "databaseName": { "type": "string" },
+            "hasServerId": { "type": "boolean" }
```
jmikola, line 266: New `invalid` test: `expectedCommandEvent-commandStartedEvent-hasServerId-type.yml`

source/unified-test-format/schema-1.3.json
```diff
@@ -213,27 +262,100 @@
...
           }
         },
         "commandSucceededEvent": {
           "type": "object",
           "additionalProperties": false,
           "properties": {
             "reply": { "type": "object" },
-            "commandName": { "type": "string" }
+            "commandName": { "type": "string" },
+            "hasServerId": { "type": "boolean" }
```
jmikola, line 275: New `invalid` test: `expectedCommandEvent-commandSucceededEvent-hasServerId-type.yml`

source/unified-test-format/schema-1.3.json
```diff
@@ -213,27 +262,100 @@
...
           }
         },
         "commandFailedEvent": {
           "type": "object",
           "additionalProperties": false,
           "properties": {
-            "commandName": { "type": "string" }
+            "commandName": { "type": "string" },
+            "hasServerId": { "type": "boolean" }
```
jmikola, line 283: New `invalid` test: `expectedCommandEvent-commandFailedEvent-hasServerId-type.yml`

source/unified-test-format/schema-1.3.json
```diff
@@ -213,27 +262,100 @@
...
           }
         }
       }
     },
 
+    "expectedCmapEvent": {
```
jmikola, line 289: New `invalid` tests prefixed with `expectedCmapEvent-` and based on existing tests for `expectedCommandEvent-*`.

source/unified-test-format/schema-1.3.json
```diff
@@ -263,13 +385,17 @@
...
         "expectResult": {},
         "saveResultAsEntity": { "type": "string" }
       },
       "allOf": [
         { "not": { "required": ["expectError", "expectResult"] } },
-        { "not": { "required": ["expectError", "saveResultAsEntity"] } }
+        { "not": { "required": ["expectError", "saveResultAsEntity"] } },
+        { "not": { "required": ["ignoreResultAndError", "expectResult"] } },
+        { "not": { "required": ["ignoreResultAndError", "expectError"] } },
+        { "not": { "required": ["ignoreResultAndError", "saveResultAsEntity"] } }
```
jmikola, line 398: New `invalid` tests:

 * `operation-ignoreResultAndError-conflicts_with_expectResult.yml`
 * `operation-ignoreResultAndError-conflicts_with_expectError.yml`
 * `operation-ignoreResultAndError-conflicts_with_saveResultAsEntity.yml`

jmikola: I realize you added these in `ignoreResultAndError.yml` but they should be split up into separate files. We currently cannot assert multiple JSON schema failures in a single file.

source/unified-test-format/schema-1.3.json
jmikola: Related to my comment about `ignoreResultAndError`, each file in `invalid` should have a single JSON schema failure.

divjotarora:
@jmikola I added the new invalid tests you proposed in your last round of review. I added `runOnRequirement-auth-type.yml` instead of the proposed `entity-client-auth-type.yml` because the `auth` requirement is not tied to a client entity.

For the new CMAP event test in `invalid`, I added one test per event type. For events that support no properties (e.g. `poolCreatedEvent`), the test has a random property like `foo: bar`. Each event that supports a property (e.g. the `reason` property for `connectionClosedEvent`) has a test like `expectedCmapEvent-<eventType>-<propertyName>-type.yml` where the property has an invalid type.

source/unified-test-format/tests/invalid/expectedCommandEvent-commandFailedEvent-hasServerId-type.yml
```diff
@@ -0,0 +1,16 @@
+description: expectedEvent-commandFailedEvent-hasServerId-type
```
jmikola, line 1: Prefix should be `expectedCommandEvent-`.

I think you'll need to do this for the renamed files as well.

source/unified-test-format/tests/invalid/expectedCommandEvent-commandStartedEvent-hasServerId-type.yml
```diff
@@ -0,0 +1,16 @@
+description: expectedEvent-commandStartedEvent-hasServerId-type
```
jmikola, line 1: `expectedCommandEvent-`

source/unified-test-format/tests/invalid/expectedCommandEvent-commandSucceededEvent-hasServerId-type.yml
```diff
@@ -0,0 +1,16 @@
+description: expectedEvent-commandSucceededEvent-hasServerId-type
```
jmikola, line 1: `expectedCommandEvent-`

source/unified-test-format/tests/invalid/expectedCommandEvent-commandSucceededEvent-hasServerId-type.yml
```diff
@@ -0,0 +1,16 @@
...
+      id: &client0 client0
+
+tests:
+  - description: invalid type for commandSucceededEvent hasServerId field
+    operations: []
+    expectEvents:
+      - client: *client0
+        events:
+          - commandSucceededEvent:
+              hasServerId: foo
```
jmikola, line 16: FYI: missing a newline at end of file. Doesn't make a difference, but the other files you created seem to have one.

source/unified-test-format/tests/invalid/expectedEventsForClient-eventType-enum.yml
```diff
@@ -0,0 +1,16 @@
...
+      id: &client0 client0
+
+tests:
+  - description: invalid eventType value
+    operations: []
+    expectEvents:
+      - client: *client0
+        eventType: foo
+        events:
+          - commandStartedEvent: {}
```
jmikola, line 16: I suggest using an empty array for `events` here. That guarantees that the only schema validation error is the `eventType` value and not a `oneOf` requirement that enforces consistency between `eventType` and `events`.

source/unified-test-format/tests/invalid/expectedEventsForClient-eventType-type.yml
```diff
@@ -0,0 +1,16 @@
...
+      id: &client0 client0
+
+tests:
+  - description: invalid eventType type
+    operations: []
+    expectEvents:
+      - client: *client0
+        eventType: 10
+        events:
+          - commandStartedEvent: {}
```
jmikola, line 16: Related to my previous comment, I think `events: []` works better here.

source/unified-test-format/unified-test-format.rst
```diff
@@ -959,6 +1050,95 @@ The structure of this object is as follows:
   - ``commandName``: Optional string. Test runners MUST assert that the command
     name matches this value.
 
+  - ``serverId``: Defined in `hasServerId`_.
+
+expectedCmapEvent
```
jyemin, line 1055: Can you create happy-path tests for all the expected events? I see load-balancer tests for these events, but not in valid-pass.

jyemin (changes requested):

source/unified-test-format/tests/invalid/expectedCommandEvent-commandFailedEvent-hasServerId-type.yml
```diff
@@ -0,0 +1,16 @@
+description: expectedCommandEvent-commandFailedEvent-hasServerId-type
+
+schemaVersion: '1.3'
+
+createEntities:
+  - client:
+      id: &client0 client0
+
+tests:
+  - description: invalid type for commandFailedEvent hasServerId field
```
jmikola, line 10: No objection to leaving this as-is, but you can use "foo" (or similar filler) for the "invalid" test descriptions since these will never be executed by a test runner.

source/unified-test-format/tests/valid-fail/assertNumberConnectionsCheckedOut.yml
```diff
@@ -0,0 +1,38 @@
+description: assertNumberConnectionsCheckedOut
+
+schemaVersion: '1.3'
+
+createEntities:
+  - client:
+      id: &client0 client0
+      useMultipleMongoses: true
```
jmikola, line 8: Is this line necessary?

source/unified-test-format/tests/valid-fail/entity-find-cursor.yml
```diff
@@ -0,0 +1,41 @@
+description: entity-find-cursor
+
+schemaVersion: '1.3'
+
+createEntities:
+  - client:
+      id: &client0 client0
+      observeEvents:
```
jmikola, line 8: This seems irrelevant since there are no event assertions below.

source/unified-test-format/tests/valid-fail/ignoreResultAndError.yml
```diff
@@ -0,0 +1,43 @@
+description: ignoreResultAndError  
+
+schemaVersion: '1.3'
+
+createEntities:
+  - client:
+      id: &client0 client0
+      useMultipleMongoses: true
```
jmikola, line 8: Is this relevant? I don't think it's necessary unless we're using transactions and want to test pinning.

source/unified-test-format/tests/valid-pass/assertNumberConnectionsCheckedOut.yml
```diff
@@ -0,0 +1,17 @@
+description: assertNumberConnectionsCheckedOut
+
+schemaVersion: '1.3'
+
+createEntities:
+  - client:
+      id: &client0 client0
+      useMultipleMongoses: true
```
jmikola, line 8: Irrelevant?

source/unified-test-format/tests/valid-pass/entity-client-cmap-events.yml
```diff
@@ -0,0 +1,40 @@
+description: entity-client-cmap-events
+
+schemaVersion: '1.3'
+
+createEntities:
+  - client:
+      id: &client0 client0
+      useMultipleMongoses: true
```
jmikola, line 8: Irrelevant?

source/unified-test-format/tests/valid-pass/expectedEventsForClient-eventType.yml
```diff
@@ -0,0 +1,66 @@
+description: expectedEventsForClient-eventType
+
+schemaVersion: '1.3'
+
+createEntities:
+  - client:
+      id: &client0 client0
+      useMultipleMongoses: true
```
jmikola, line 8: Irrelevant?

source/unified-test-format/tests/valid-pass/ignoreResultAndError.yml
```diff
@@ -0,0 +1,34 @@
+description: ignoreResultAndError
+
+schemaVersion: '1.3'
+
+createEntities:
+  - client:
+      id: &client0 client0
+      useMultipleMongoses: true
```
jmikola, line 8: Irrelevant?

source/unified-test-format/unified-test-format.rst
```diff
@@ -442,7 +475,14 @@ The structure of this object is as follows:
...
 
-    This option has no effect for non-sharded topologies.
+    If this option is true or unset and the topology type is
+    ``LoadBalanced``, the test runner MUST connect the MongoClient to a load
+    balancer that is balancing multiple servers. If it is false and the
+    topology type is ``LoadBalanced``, the test runner MUST connect the
+    MongoClient to a load balancer that is only fronting a single server.
+
+    This option has no effect for topologies that are not sharded or load
+    balanced.
```
jmikola, line 485: I think I missed this addition in an earlier review. Perhaps it explains all of the `useMultipleMongoses: true` lines that I asked about in the test files above.

It seems odd to expand the use of `useMultipleMongoses` in this way, as it was previously only used to test transactions and session pinning (unless I'm mistaken). Reading #939, I see load balancers currently front:

> Any MongoDB service that can run behind a load balancer, including but not limited to mongos, Atlas Proxy, Elastic Front End.

How feasible is it for drivers to detect what (or how many) servers are behind the load balancer?

And with regard to existing use of this field for transaction tests, do you anticipate a conflict if we start running those tests against a load balancer? Perhaps it's not a concern if none of those tests include `load-balancer` in their accepted topologies?

I wonder if it'd be safer to introduce a separate option for this (e.g. `loadBalanceMultipleServers`), which is only relevant if the topology is load-balanced (similar to how `useMultipleMongoses` is currently ignored for non-sharded topologies).

source/unified-test-format/unified-test-format.rst
```diff
@@ -959,6 +1050,95 @@ The structure of this object is as follows:
...
+  MUST be an empty document as all fields in this event are non-deterministic.
+
+
+hasServerId
+```````````
+
+This field is an optional boolean that specifies whether or not the ``serverId``
+field of an event is set. If true, test runners MUST assert that the field is
+set and is a non-empty BSON ObjectId (i.e. all bytes of the ObjectId are not 0).
+If false, test runners MUST assert that the field is not set.
```
jmikola, line 1140: Should you clarify that "is not set" allows an ObjectId with all zero bytes? Or rather say "is not set or is an empty BSON ObjectId"?

source/unified-test-format/unified-test-format.rst
divjotarora: Does the `valid-pass/entity-client-cmap-events.yml` file address this?

source/load-balancers/tests/cursors.yml
```diff
@@ -0,0 +1,562 @@
...
+              command:
+                find: *collection0Name
+                filter: {}
+              commandName: find
+          - commandSucceededEvent:
+              reply:
+                cursor:
+                  id: 0
+                  firstBatch: { $$type: array }
+                  ns: { $$type: string }
```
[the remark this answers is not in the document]
divjotarora: Yes, it could be done that way, but I'd prefer to leave it as-is. It doesn't really add anything to the test.

source/load-balancers/tests/lb-connection-establishment.yml
```diff
@@ -0,0 +1,29 @@
...
+    skipReason: servers have not implemented LB support yet so they will not fail the connection handshake in this case
+    operations:
+      - name: runCommand
+        object: *database0
+        arguments:
+          commandName: ping
+          command: { ping: 1 }
+        expectError:
+          isClientError: false
```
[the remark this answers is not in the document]
divjotarora: I'm not sure I understand. This is asserting that the error is _not_ a client error because it's caused by the server. Also, small clarification: in LB mode, there is no server discovery. We assume that the load balancer is always selectable, so there are no server selection timeout errors either. I added an `expectEvents` block to assert that no events are sent because the connection handshake fails, so we never send the `ping` command.

source/unified-test-format/unified-test-format.rst
jyemin: That file doesn't test all the event types listed here.

source/load-balancers/tests/server-selection.yml
```diff
@@ -0,0 +1,50 @@
...
+      databaseName: &database0Name database0Name
+  - collection:
+      id: &collection0 collection0
+      database: *database0
+      collectionName: &collection0Name coll0
+      collectionOptions:
+        readPreference:
+          # Use secondaryPreferred to ensure that operations can succeed even if the shards are only comprised of one
+          # server.
+          mode: &readPrefMode secondaryPreferred
```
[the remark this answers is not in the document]
divjotarora: Some drivers may append `primary` read preferences when communicating with mongos, but it's not clear to me that all drivers will do so based on the OP_MSG and server selection specs. I hope that DRIVERS-1640 and/or DRIVERS-1182 will clarify the intended behavior. In the meantime, we know that using a non-primary read preference will always result in a `$readPreference` command field when communicating with mongos and we're not making assertions about operation results in these tests, so I don't think there's any risk in doing this.

source/unified-test-format/tests/invalid/expectedCommandEvent-commandFailedEvent-hasServerId-type.yml
divjotarora: Changed to `foo` here and in other files for consistency.

source/unified-test-format/tests/valid-fail/entity-find-cursor.yml
divjotarora: Thanks for catching that. Removed the `observeEvents` array.

source/unified-test-format/unified-test-format.rst
divjotarora: > How feasible is it for drivers to detect what (or how many) servers are behind the load balancer?

AFAIK, neither of these are possible or desirable as we don't want to special case any mongos-like services and mongos servers are not aware of each other.

> do you anticipate a conflict if we start running those tests against a load balancer

If we find that any of the existing transactions tests fail against a load balancer, we should explicitly mark them as only being valid for replica sets and/or sharded clusters via the `topologies` requirement. My guess is many of these tests will already be skipped against LB clusters because they include topology-specific runOnRequirement blocks to account for the existence of `failCommand` on different server versions.

> I wonder if it'd be safer to introduce a separate option for this (e.g. loadBalanceMultipleServers)

I briefly discussed this with @jyemin when designing the unified test format changes for this spec. We opted to reuse `useMultipleMongoses` because the spirit of the constraint is the same: we want the test to be communicating with more than one mongos server to test some special behavior. There is a concern that it's not possible to verify that there are actually multiple mongos's behind the LB during testing, so we have to rely on orchestration scripts to ensure that. I'm not sure how adding a new option that's basically the same thing but only applies to LB clusters would help.

source/unified-test-format/unified-test-format.rst
divjotarora: Done.

divjotarora (commented):
@jmikola I'm ignoring the comments about `useMultipleMongoses` in tests being irrelevant until we resolve the thread about using it for LB clusters in the first place. PTAL

source/load-balancers/tests/lb-connection-establishment.yml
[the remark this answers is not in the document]
jmikola: Ignore my last comment. I must have misread this as `isClientError: true` when I wrote that.

source/unified-test-format/unified-test-format.rst
jmikola: Points noted on the first two replies.

> We opted to reuse useMultipleMongoses because the spirit of the constraint is the same: we want the test to be communicating with more than one mongos server to test some special behavior. There is a concern that it's not possible to verify that there are actually multiple mongos's behind the LB during testing, so we have to rely on orchestration scripts to ensure that. I'm not sure how adding a new option that's basically the same thing but only applies to LB clusters would help.

I'm reading that you want to test communication with "more than one mongos" and are concerned about not being able to verify the number of mongos nodes behind a LB. That suggests to me that `useMultipleMongoses` is still intended to apply to mongos specifically.

I originally misread this change as amending `useMultipleMongoses` to also cover "multiple load balancers", but that seems not to be the case. Re-reading the text above, I realize it isn't actually saying anything about the number of load balancers -- just the number of servers behind them. Furthermore, I gather it's assuming that the type of server will always be mongos and disregard other possible types noted in #939 (e.g. elastic).

Here is a very direct series of questions to help me understand more clearly:

 * With the suggested changes above, does `useMultipleMongoses` remain _only_ concerned with whether or not the driver is communicating with multiple mongos nodes (either directly or through one or more load balancers)? If "yes", why is the presence and/or number of load balancers irrelevant?
 * `useMultipleMongoses` has historically been used to test session pinning. I see a few references to "pin" in #939 but it's not clear to me how much it differs from what is currently dictated by [Transactions: Mongos Pinning](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#mongos-pinning). Does pinning apply to drivers that connect to multiple load balancers? If so, is the type of server(s) behind the load balancers irrelevant to this pinning?

If there is going to be an assumption that any load balancer topology is fronting multiple mongoses (such that a load balancer topology is sufficient to satisfy `useMultipleMongoses: true` with no further introspection), I'd argue that the right course of action would be to add `load-balanced` as a permitted `runOnRequirement` for those existing tests that currently require a sharded cluster and specify `useMultipleMongoses: true`. AFAIK, the `runOnRequirement` would need to be changed anyway to permit those tests to run against an LB (see: [transactions/tests/mongos-pin-auto.yml](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/mongos-pin-auto.yml)). Furthermore, this would entail no changes to how the test runner evaluates `useMultipleMongoses` (it would continue to be ignored for non-sharded toplogies).

source/unified-test-format/unified-test-format.rst
divjotarora: Talked offline. The `entity-client-cmap-events` file tests the simplest ones to ensure drivers are examining CMAP events. I don't want to spend more time on this right now as there are higher priority things to do for the spec tests.

source/unified-test-format/unified-test-format.rst
divjotarora: I pushed some changes to describe passing in multiple URIs to the test runner in the "Initializing a Test Runner" section and modified the text for ``useMultipleMongoses`` to match. I skipped the bit about defining "mongos" at the top of the spec because we're still only talking about testing against "real" mongos servers right now. The serverless testing project will be the first to test against mongos-like servers so I'd like to defer this change until that project is done.

source/unified-test-format/unified-test-format.rst
```diff
@@ -475,11 +475,12 @@ The structure of this object is as follows:
     connecting to multiple mongos servers is necessary to test session
     pinning.
 
-    If this option is true or unset and the topology type is
-    ``LoadBalanced``, the test runner MUST connect the MongoClient to a load
-    balancer that is balancing multiple servers. If it is false and the
-    topology type is ``LoadBalanced``, the test runner MUST connect the
-    MongoClient to a load balancer that is only fronting a single server.
+    If the topology type is ``LoadBalanced``, the test runner MUST use one of
```
jmikola, line 478: I just noticed that the spec refers to "load-balanced" (`runOnRequirements`), "LoadBalanced" (prose sections), and "loadBalanced" (changelog). Can we standardized on "load-balanced" everywhere to agree with the topology type referenced in `runOnRequirements`?

source/unified-test-format/unified-test-format.rst
```diff
@@ -2649,6 +2650,14 @@ to elsewhere in the specification as the internal MongoClient.
 Determine the server version and topology type using the internal MongoClient.
 This information will be used to evaluate any future `runOnRequirement`_ checks.
 
+In addition to the cluster URI, the test runner MUST also be configurable
```
jmikola, line 2653: I think this may be the only reference to "cluster URI". Can we change this to "aforementioned connection string" so it's clear we're referring back to the thing mentioned in the first paragraph of this section?

source/unified-test-format/unified-test-format.rst
```diff
@@ -2649,6 +2650,14 @@ to elsewhere in the specification as the internal MongoClient.
...
+with two other connection strings (or equivalent configuration) that point to
+TCP load balancers: one fronting multiple servers and one fronting a single
+server. These will be used to initialize client entities when executing tests
+against a load balanced cluster. If the topology type is ``LoadBalanced``,
+the test runner MUST error if either of these URIs is not provided. For all
+other topology types, these URIs MUST be ignored.
```
jmikola, line 2659: "For all other topology types, these URIs SHOULD NOT be provided and MUST be ignored if provided."

Alternatively, would it make sense to just say they "MUST NOT be provided", or are you thinking that some CI configurations need to set these environment variables as empty strings (or equivalent)?

source/unified-test-format/unified-test-format.rst
```diff
@@ -2649,6 +2650,14 @@ to elsewhere in the specification as the internal MongoClient.
...
+
```
jmikola, line 2660: In the paragraph under "Executing a Test" that contains "If createEntities is specified, the test runner MUST create each entity accordingly...", I think you can add a sentence reminding implementations to consult these connection strings when creating client entities for `load-balanced` topolgies.

```
If the topology type is `load-balanced`, client entities MUST be initialized with the appropriate load balancer connection string as discussed in `useMultipleMongoses <entity_client_useMultipleMongoses_>`_.
```

source/unified-test-format/unified-test-format.rst
divjotarora: The actual topology type in the SDAM spec will be called `LoadBalanced` so I'd like to keep the prose as-is. For `runOnRequirements`, I used `load-balanced` because the existing topologies requirements are all lowercase and `sharded-replicaset` sets a precedent for hyphenating multiple words. The changelog entry was referring to an older change that has since been removed so I updated that to remove the `loadBalanced` wording.

source/unified-test-format/unified-test-format.rst
divjotarora: Done. The change looks much bigger than it is due to editor line wrapping at 80 characters.

source/unified-test-format/unified-test-format.rst
divjotarora: The latter. At least for Go, we don't usually do empty string checks to determine if certain environment variables should be passed into the test runner. Instead, we'll just access the ones we know _can_ exist and they'll be empty strings that get ignored by the runner if it doesn't need them.

I updated the text to "For all other topology types, these URIs SHOULD NOT be provided and MUST be ignored if provided".

source/unified-test-format/unified-test-format.rst
divjotarora: Added with one small change: `LoadBalanced` rather than `load-balanced`. See [this comment thread](https://github.com/mongodb/specifications/pull/935/files/a264791eace83406fdb1057d6eb78670759b6fce..2a666f2108a5c6c028baac88c454d1ee786eff06#r608093800) for an explanation.

source/unified-test-format/unified-test-format.rst
jmikola: Works for me. Thanks for the explanation.

jmikola (approved):

divjotarora:
I'm closing this PR and will be merging the unified test format changes into the `master` branch. I will put up a new PR based on the new `master` branch that contains just the LB test YAML/JSON files.

PR #936, opened 2021-03-24
Test intended invalid syntax for expectedEventsForClient

jmikola:
These tests previously nested expectEvents within an operation, which was invalid but meant the intended invalid syntax was never tested.

Note: I'm forgoing a DRIVERS ticket here as drivers don't sync these tests. They are only run from the specification repository's CI.

### DIFF
--- a/source/unified-test-format/tests/invalid/expectedEvent-additionalProperties.json
+++ b/source/unified-test-format/tests/invalid/expectedEvent-additionalProperties.json
@@ -11,18 +11,13 @@
   "tests": [
     {
       "description": "foo",
-      "operations": [
+      "operations": [],
+      "expectEvents": [
         {
-          "name": "foo",
-          "object": "client0",
-          "expectEvents": [
+          "client": "client0",
+          "events": [
             {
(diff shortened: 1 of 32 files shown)


source/unified-test-format/tests/invalid/do.php
```diff
@@ -0,0 +1,17 @@
+<?php
```
alcaeus, line 1: Not sure I understand what this file is for. Is this to be used during development? If so, could you add a comment or other way to explain its usage? Also, while I appreciate building the popularity in PHP, I'm not sure if it's going to be the most portable, as I don't expect everybody to have PHP ready (especially with MacOS not shipping PHP by default in the near future).

alcaeus (approved):
LGTM besides the PHP file that seems a little out of place.

source/unified-test-format/tests/invalid/do.php
jmikola: Oops, I didn't mean to add this.

I used this early on to sync descriptions in each file with the filename.

PR #938, opened 2021-03-24
DRIVERS-1483 PoolClearedErrors must be retried

patrickfreed:
DRIVERS-1483

This PR updates the language of the retryability specifications to require that [`PoolClearedErrors`](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#connection-pool-errors) are retried. This PR also adds a prose test to each spec so drivers can confirm they conform to the requirement correctly.

POC implementation in Rust: https://github.com/patrickfreed/mongo-rust-driver/commit/2ed7132e9cd936ca071ffdc7f072036150fc7115

This is the final spec change in the Avoiding Connection Storms project (DRIVERS-781)

### DIFF
--- a/source/retryable-reads/retryable-reads.rst
+++ b/source/retryable-reads/retryable-reads.rst
@@ -79,8 +79,9 @@ SocketException                 9001
   listed above, but with an error message containing the substring "not
   master" or "node is recovering"
 
-- a ``PoolClearedError`` as `defined in the CMAP specification
-  <https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#connection-pool-errors>`__.
+- a `PoolClearedError`_
+
+  .. _PoolClearedError: ../connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#connection-pool-errors
 
 MongoClient Configuration 
 --------------------------
--- a/source/retryable-reads/retryable-reads.rst
+++ b/source/retryable-reads/retryable-reads.rst
@@ -384,11 +384,10 @@ and reflects the flow described above.
       return executeCommand(server, command);
     } catch (NetworkException originalError) {
       updateTopologyDescriptionForNetworkError(server, originalError);
-      return executeRetry(command, session, originalError);
     } catch (NotMasterException originalError) {
       updateTopologyDescriptionForNotMasterError(server, originalError);
-      return executeRetry(command, session, originalError);
     }
+    return executeRetry(command, session, originalError);
   }
   
   /**
--- a/source/retryable-reads/retryable-reads.rst
+++ b/source/retryable-reads/retryable-reads.rst
@@ -3,7 +3,7 @@ Retryable Reads
 ===============
 
 :Spec Title: Retryable Reads
-:Spec Version: 1.1
+:Spec Version: 1.2
 :Author: Vincent Kam 
 :Lead: Bernie Hackett
 :Advisory Group: Shane Harvey, Scott L’Hommedieu, Jeremy Mikola
@@ -11,7 +11,7 @@ Retryable Reads
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: 3.6
-:Last Modified: 2019-05-29
+:Last Modified: 2021-03-24
    
 .. contents::
 
@@ -78,6 +78,9 @@ SocketException                 9001
 - a server error response without an error code or one different from those
   listed above, but with an error message containing the substring "not
   master" or "node is recovering"
+
+- a ``PoolClearedError`` as `defined in the CMAP specification
+  <https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#connection-pool-errors>`__.
 
 MongoClient Configuration 
 --------------------------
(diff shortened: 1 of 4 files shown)


source/retryable-reads/retryable-reads.rst
```diff
@@ -356,26 +360,43 @@ and reflects the flow described above.
...
      * can then assume that no attempts were made. */
     server = selectServer();
-    connection = server.getConnection()   
+
+    /* PoolClearedException indicates the operation did not even attempt to
+     * create a connection, let alone execute the operation. This means we
+     * are always safe to attempt a retry. We do not need to update SDAM,
+     * since whatever error caused the pool to be cleared will do so itself. */
+    try {
+      connection = server.getConnection()
```
patrickfreed, line 369: The psuedocode between retryable-reads and retryable-writes is a bit divergent. Here, the connection check out step is shown explicitly and happens before retryablity support is determined, but in retryable-writes checkout is done as part of execution. I didn't bother trying to unify them in this PR and instead just updated each as necessary to fit the behavioral changes specified in the prose. In DRIVERS-746, we'll likely have to do the full reunification.

In a related note: I think we should consider prioritizing DRIVERS-1401. It's a bit strange that we have error labels for determining retryable writes but don't have the same machinery for reads.

source/retryable-reads/retryable-reads.rst
```diff
@@ -356,26 +360,43 @@ and reflects the flow described above.
...
+    } catch (PoolClearedException poolClearedError) {
+      return executeRetry(command, session, poolClearedError);
```
patrickfreed, line 371: This unconditionally retries, which should be safe as per the comment in the code. That being said, I did not make this a requirement, as it could be hard to specify, at least for `PoolClosedError`s specifically. When we do the more general DRIVERS-746, I think this would be worth revisiting.

source/retryable-writes/retryable-writes.rst
```diff
@@ -216,72 +216,84 @@ This error label can be found in the top-level "errorLabels" field of the error.
 In a server error response with a writeConcernError field the top level document
 or the writeConcernError document may contain the RetryableWriteError error label.
 
+RetryableWriteError Labels
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
 The RetryableWriteError label might be added to an error in a variety of ways:
 
-When the driver encounters a network error communicating with any server version
```
patrickfreed, line 221: This section was a bit hard to scan, so I broke it down to a bulleted list, which I found easier to read, most of the text is unchanged. Let me know if this new organization seems okay.

source/retryable-writes/retryable-writes.rst
```diff
@@ -441,22 +453,20 @@ Consider the following pseudo-code:
...
-      updateTopologyDescriptionForNetworkError(server, originalError);
-    } catch (NotMasterException originalError) {
-      updateTopologyDescriptionForNotMasterError(server, originalError);
-    } catch (IllegalOperation originalError) {
-      if ( originalError.code == 20 && originalError.errmsg.startsWith("Transaction numbers") ) {
-        originalError.errmsg = "This MongoDB deployment does not support retryable...";
-        throw originalError;
+    } catch (Exception originalError) {
+      handleError(e);
+      if (!e.hasErrorLabel("RetryableWriteError")) {
```
patrickfreed, line 465: the psuedocode hadn't actually been updated yet to include the newly specified label-based behavior, so I did so here. No changes specific to `PoolClearedError` were really necessary as a result.

jmikola: Oh, is this now covered by `handleError`? I just saw that below, and the reference to SDAM.

source/retryable-writes/retryable-writes.rst
```diff
@@ -485,17 +495,20 @@ Consider the following pseudo-code:
      * propagate. */
     try {
       return executeCommand(server, retryableCommand);
-    } catch (NetworkException secondError) {
-      updateTopologyDescriptionForNetworkError(server, secondError);
-      throw secondError;
-    } catch (NotMasterException secondError) {
-      updateTopologyDescriptionForNotMasterError(server, secondError);
+    } catch (Exception secondError) {
+      if (secondError instanceof DriverException) {
```
jmikola, line 499: I would suggest the following to avoid nesting and also give more context to the DriverException being ignored (as we do above when server selection fails):

```
} catch (DriverException ignoredError) {
    throw originalError;
} catch (Exception secondError) {
    handleError(secondError);
    throw secondError;
}
```

source/retryable-writes/tests/README.rst
```diff
@@ -327,10 +327,45 @@ and sharded clusters.
...
+           }
+       }
+
+   3. Start two threads and attempt to perform an ``insertOne`` simultaneously on both.
+
+   4. Verify that both ``insertOne`` attempts succeed.
+
+   5. Via CMAP monitoring, assert that the first check out succeeds.
+
+   6. Via CMAP monitoring, assert that the second check out fails due to a
```
jmikola, line 356: Same comments as earlier on the retryable reads prose test:

 * Should we expect a certain number of CommandStartedEvents?
 * Should we expect a PoolClearedEvent?

source/retryable-reads/tests/README.rst
```diff
@@ -164,6 +165,45 @@ Optional Enumeration Commands
...
+3. Start two threads and attempt to perform a ``findOne`` simultaneously on both.
+
+4. Verify that both ``findOne`` attempts succeed.
+
+5. Via CMAP monitoring, assert that the first check out succeeds.
+
+6. Via CMAP monitoring, assert that a PoolClearedEvent is then emitted.
+
+7. Via CMAP monitoring, assert that the second check out then fails due to a
+   connection error.
```
patrickfreed, line 199: The CMAP spec doesn't require that drivers include any information about the actual error that occurred, should we update it to do so? Could drivers do that without a breaking change? cc @divjotarora

jmikola (approved):

isabelatkinson (approved):

patrickfreed:
pinging @ShaneHarvey in case this got lost in the mix over the skunkworks break.

ShaneHarvey (approved):
LGTM, apologies for my delay.

PR #939, opened 2021-03-30
WRITING-6786: Load Balancer Spec

durran:
Includes the load balancer spec itself, as well as modifications to other related specifications.

Also includes a second commit to handle DRIVERS-1293 in the specs modified by the load balancer work.

### DIFF
--- a/source/command-monitoring/command-monitoring.rst
+++ b/source/command-monitoring/command-monitoring.rst
@@ -11,8 +11,8 @@ Command Monitoring
 :Status: Approved
 :Type: Standards
 :Minimum Server Version: 2.4
-:Last Modified: Apr 16, 2018
-:Version: 1.8
+:Last Modified: Apr 15, 2021
+:Version: 1.9
 
 .. contents::
 
@@ -256,6 +256,8 @@ value MUST be replaced with an empty BSON document. The list is as follows:
 API
 ---
 
+See the `Load Balancer Specification <../load-balancers/load-balancers.rst#events>`__ for details on the ``serviceId`` field.
+
 .. code:: typescript
 
   interface CommandStartedEvent {
@@ -292,6 +294,13 @@ API
      * The name of this field is flexible to match the object that is returned from the driver.
      */
     connectionId: ConnectionId;
+
+    /**
+     * Returns the service id for the command when the driver is in load balancer mode.
+     * For drivers that wish to include this in their ConnectionId object, this field is
+     * optional.
+     */
+    serviceId: Optional<ObjectId>;
   }
 
   interface CommandSucceededEvent {
@@ -331,6 +340,13 @@ API
      * The name of this field is flexible to match the object that is returned from the driver.
      */
     connectionId: ConnectionId;
+
+    /**
+     * Returns the service id for the command when the driver is in load balancer mode.
+     * For drivers that wish to include this in their ConnectionId object, this field is
+     * optional.
+     */
+    serviceId: Optional<ObjectId>;
   }
 
   interface CommandFailedEvent {
@@ -371,6 +387,13 @@ API
      * The name of this field is flexible to match the object that is returned from the driver.
      */
     connectionId: ConnectionId;
+
+    /**
+     * Returns the service id for the command when the driver is in load balancer mode.
+     * For drivers that wish to include this in their ConnectionId object, this field is
+     * optional.
+     */
+    serviceId: Optional<ObjectId>;
   }
 
 
@@ -450,3 +473,6 @@ Changelog
 
 12 FEB 2020:
   - Added ``isMaster.speculativeAuthenticate`` to the list of values that should be redacted.
+
+15 APR 2021:
+  - Added ``serviceId`` field to events.
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -9,8 +9,8 @@ Connection Monitoring and Pooling
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: N/A
-:Last Modified: September 24, 2020
-:Version: 1.4.0
+:Last Modified: 2021-04-12
+:Version: 1.5.0
 
 .. contents::
 
@@ -203,7 +203,7 @@ A driver-defined wrapper around a single TCP connection to an Endpoint. A `Conne
 
       /**
        *  An integer representing the “generation” of the pool
-       *  when this Connection was created
+       *  when this Connection was created.
        */
       generation: number;
 
@@ -281,9 +281,15 @@ has the following properties:
       waitQueue: WaitQueue;
     
       /**
-       *  A generation number representing the SDAM generation of the pool
+       *  A generation number representing the SDAM generation of the pool.
        */
       generation: number;
+
+      /**
+       * A map representing the various generation numbers for various services
+       * when in load balancer mode.
+       */
+      serviceGenerations: Map<ObjectId, [number, number]>;
 
       /**
        * The state of the pool.
@@ -675,6 +681,10 @@ incrementing the generation / marking the pool as "paused". If the pool is
 already "paused" when it is cleared, then the pool MUST NOT emit a PoolCleared
 event.
 
+A Pool MUST also have a method of clearing all `Connections <#connection>`_ for a
+specific ``serviceId`` for use when in load balancer mode. This method increments
+the generation of the pool for that specific ``serviceId`` in the generation map.
+
 As part of clearing the pool, the WaitQueue MUST also be cleared, meaning all
 requests in the WaitQueue MUST fail with errors indicating that the pool was
 cleared while the checkOut was being performed. The error returned as a result
@@ -683,6 +693,20 @@ an error that marks the SDAM state unknown. Clearing the WaitQueue MUST happen
 eagerly so that any operations waiting on `Connections <#connection>`_ can retry
 as soon as possible. The pool MUST NOT rely on WaitQueueTimeoutMS to clear
 requests from the WaitQueue.
+
+Load Balancer Mode
+------------------
+
+For load-balanced deployments, pools MUST maintain a map from ``serviceId`` to a
+tuple of (generation, connection count) where the connection count refers to the
+total number of connections that exist for a specific ``serviceId``. The pool MUST
+remove the entry for a ``serviceId`` once the connection count reaches 0.
+Once the MongoDB handshake is done, the connection MUST get the
+generation number that applies to its ``serviceId`` from the map and update the
+map to increment the connection count for this ``serviceId``.
+
+See the `Load Balancer Specification <../load-balancers/load-balancers.rst#connection-pooling>`__ for details.
+
 
 Forking
 -------
@@ -729,6 +753,7 @@ All drivers that implement a connection pool MUST provide an API that allows use
 Events
 ------
 
+See the `Load Balancer Specification <../load-balancers/load-balancers.rst#events>`__ for details on the ``serviceId`` field.
 
 .. code:: typescript
 
@@ -765,6 +790,12 @@ Events
        *  The ServerAddress of the Endpoint the pool is attempting to connect to.
        */
       address: string;
+
+      /**
+       * The service id for which the pool was cleared for in load balancing mode.
+       * See load balancer specification for more information about this field.
+       */
+      serviceId: Optional<ObjectId>
     }
 
     /**
@@ -1077,6 +1108,8 @@ Change log
 
 :2019-06-06: Add "connectionError" as a valid reason for
              ConnectionCheckOutFailedEvent
+
+:2021-4-12: Adding in behaviour for load balancer mode.
 
 .. Section for links.
 
--- a/source/load-balancers/load-balancers.rst
+++ b/source/load-balancers/load-balancers.rst
@@ -1,0 +1,416 @@
+=====================
+Load Balancer Support
+=====================
+
+:Spec Title: Load Balancer Support
+:Spec Version: 1.0.0
+:Author: Durran Jordan
+:Advisors: Jeff Yemin, Divjot Arora, Andy Schwerin, Cory Mintz
+:Status: Accepted
+:Type: Standards
+:Minimum Server Version: 5.0
+:Last Modified: 2021-04-06
+
+.. contents::
+
+--------
+
+Abstract
+========
+
+This specification defines driver behaviour when connected to MongoDB services
+through a load balancer.
+
+META
+====
+
+The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in `RFC 2119 <https://www.ietf.org/rfc/rfc2119.txt>`__.
+
+Specification
+=============
+
+
+Terms
+-----
+
+SDAM
+^^^^
+
+An abbreviated form of "Server Discovery and Monitoring", specification defined
+in `Server Discovery and Monitoring Specification <../server-discovery-and-monitoring/server-discovery-and-monitoring.rst>`__.
+
+Service
+^^^^^^^
+
+Any MongoDB service that can run behind a load balancer.
+
+
+MongoClient Configuration
+-------------------------
+
+loadBalanced
+^^^^^^^^^^^^
+
+To specify to the driver to operate in load balancing mode, a connection string
+option of :code:`loadBalanced=true` MUST be added to the connection string.
+This boolean option specifies whether or not the driver is connecting to a
+MongoDB cluster through a load balancer. The default value MUST be false.
+This option MUST only be configurable at the level of a :code:`MongoClient`.
+
+URI Validation
+^^^^^^^^^^^^^^
+
+When :code:`loadBalanced=true` is provided in the connection string, the driver
+MUST throw an exception in the following cases:
+
+- The connection string contains more than one host/port.
+- The connection string contains a replicaSet option.
+- The connection string contains a directConnection option.
+
+If a URI is provided with the ``mongodb+srv`` scheme, the driver MUST first do the
+SRV and TXT lookup and then perform the validation. For drivers that do SRV
+lookup asynchrounously this may result in a ``MongoClient`` being instantiated
+but erroring later during operation execution.
+
+
+DNS Seedlist Discovery
+^^^^^^^^^^^^^^^^^^^^^^
+
+The connection string option for :code:`loadBalanced=true` MUST be valid in a
+TXT record and when present MUST be validated as defined in the URI Validation section.
+
+When a MongoClient is configured with an SRV URI and :code:`loadBalanced=true`, the
+driver MUST NOT poll for changes in the SRV record as is done for non-load balanced
+sharded clusters.
+
+Server Discovery and Monitoring
+-------------------------------
+
+Monitoring
+^^^^^^^^^^
+
+When :code:`loadBalanced=true` is specified in the URI the topology MUST start
+in type :code:`LoadBalanced` and MUST remain as :code:`LoadBalanced` indefinitely.
+The topology MUST contain 1 :code:`ServerDescription` with a :code:`ServerType` of
+:code:`LoadBalancer`. The "address" field of the :code:`ServerDescription` MUST be
+set to the address field of the load balancer. All other fields in the
+:code:`ServerDescription` MUST remain unset. In this mode the driver MUST NOT
+start a monitoring connection. The :code:`TopologyDescription`'s :code:`compatible`
+field MUST always be :code:`true`.
+
+Although there is no monitoring connection in load balanced mode, drivers MUST emit
+the following series of SDAM events:
+
+- :code:`TopologyOpeningEvent` when the topology is created.
+- :code:`TopologyDescriptionChangedEvent`. The :code:`previousDescription` field MUST
+  have :code:`TopologyType` :code:`Unknown` and no servers. The :code:`newDescription`
+  MUST have :code:`TopologyType` :code:`LoadBalanced` and one server with
+  :code:`ServerType` :code:`Unknown`.
+- :code:`ServerOpeningEvent` when the server representing the load balancer is created.
+- :code:`ServerDescriptionChangedEvent`. The :code:`previousDescription` MUST have
+  :code:`ServerType` :code:`Unknown`. The :code:`newDescription` MUST have
+  :code:`ServerType` :code:`LoadBalancer`.
+- :code:`TopologyDescriptionChangedEvent`. The :code:`newDescription` MUST have
+  :code:`TopologyType` :code:`LoadBalanced` and one server with :code:`ServerType`
+  :code:`LoadBalancer`.
+
+Drivers MUST also emit a :code:`ServerClosedEvent` and :code:`TopologyClosedEvent` when
+the topology is closed and MUST NOT emit any other events when operating in this mode.
+
+
+Driver Sessions
+---------------
+
+Session Support
+^^^^^^^^^^^^^^^
+
+When the :code:`TopologyType` is :code:`LoadBalanced`, sessions are always supported.
+
+Session Expiration
+^^^^^^^^^^^^^^^^^^
+
+When in load balancer mode, drivers MUST ignore :code:`logicalSessionTimeoutMinutes`
+and MUST NOT prune client sessions from the session pool when implemented by the driver.
+
+Data-Bearing Server Type
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+A :code:`ServerType` of :code:`LoadBalancer` MUST be considered a data-bearing server.
+
+
+Server Selection
+----------------
+
+A deployment of topology type Load Balanced contains one server of type :code:`LoadBalancer`.
+
+For read and write operations, the single server in the topology MUST always be selected.
+
+During command construction, the LoadBalancer server MUST be treated like a mongos and
+drivers MUST add a $readPreference field to the command when required by
+`Passing read preference to mongos and load balancers <../server-selection/server-selection.rst#passing-read-preference-to-mongos-and-load-balancers>`_.
+
+
+Connection Pooling
+------------------
+
+Connection Establishment
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+In the case of the driver having the :code:`loadBalanced=true` connection string option
+specified, every pooled connection MUST add a :code:`loadBalanced` field to the
+:code:`hello` or legacy hello command in its `handshake <../mongodb-handshake/handshake.rst#connection-handshake>`__.
+The value of the field MUST be :code:`true`.
+
+Example:
+
+Driver connection string contains :code:`loadBalanced=true`:
+
+.. code:: typescript
+
+    { hello: 1, loadBalanced: true }
+
+Driver connection string contains :code:`loadBalanced=false` or no
+:code:`loadBalanced` option:
+
+.. code:: typescript
+
+    { hello: 1 }
+
+When the server's hello response does not contain a :code:`serviceId` field,
+the driver MUST throw an exception with the message "Driver attempted to initialize
+in load balancing mode, but the server does not support this mode."
+
+For single threaded drivers that do not use a connection pool, the driver MUST have only 1
+socket connection to the load balancer in load balancing mode.
+
+Connection Pinning
+^^^^^^^^^^^^^^^^^^
+
+Some features in MongoDB such as cursors and transactions require sending multiple
+commands to the same mongos in a sharded cluster. In load balanced mode, it is not
+possible to target the same mongos behind a load balancer when pooling connections.
+To account for this, drivers MUST pin to a single connection for these features.
+When using a pinned connection, the driver MUST emit only 1
+:code:`ConnectionCheckOutStartedEvent`, and only 1 :code:`ConnectionCheckedOutEvent`
+or :code:`ConnectionCheckOutFailedEvent`. Similarly, the driver MUST only publish 1
+:code:`ConnectionCheckedInEvent`.
+
+Behaviour With Cursors
+^^^^^^^^^^^^^^^^^^^^^^
+
+When the driver is in load balancing mode and executing any cursor-initiating command, the driver
+MUST NOT check the connection back into the pool unless the command fails or the server
+returns a cursor ID of :code:`0` (i.e. all documents are returned in a single batch).
+Otherwise, the driver MUST continue to use the same connection for all subsequent
+:code:`getMore` commands for the cursor. The driver MUST check the connection back
+into the pool if the server returns a cursor ID of :code:`0` in a :code:`getMore`
+response (i.e. the cursor is drained). When the cursor's :code:`close` method is
+invoked, either explicitly or via an implicit resource cleanup mechanism, the driver
+MUST use the same connection to execute a :code:`killCursors` command if necessary
+and then check the connection back into the pool regardless of the result.
+
+If a :code:`getMore` fails with a network error, drivers MUST leave the connection pinned
+to the cursor. When the cursor's :code:`close` method is invoked, drivers MUST NOT execute
+a :code:`killCursors` command because the pinned connection is no longer valid and MUST
+return the connection back to the pool.
+
+Behaviour With Transactions
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When executing a transaction in load balancing mode, drivers MUST follow the rules outlined
+in `Sharded Transactions <../transactions/transactions.rst#sharded-transactions>`__ with one
+exception: drivers MUST use the same connection for all commands in the transaction
+(excluding retries of commitTranscation and abortTransaction in some cases). Pinning
+to a single connection ensures that all commands in the transaction target the same
+service behind the load balancer. The rules for pinning to a connection and releasing
+a pinned connection are the same as those for server pinning in non-load balanced sharded
+transactions as described in `When to unpin <../transactions/transactions.rst#when-to-unpin>`__.
+
+Connection Tracking
+^^^^^^^^^^^^^^^^^^^
+
+The driver connection pool MUST track the purpose for which connections are checked out
+in the following 3 categories:
+
+- Connections checked out for cursors
+- Connections checked out for transactions
+- Connections checked out for operations not falling under the previous 2 categories
+
+When the connection pool's :code:`maxPoolSize` is reached and the pool times out waiting
+for a new connection the :code:`WaitQueueTimeoutError` MUST include a new detailed message,
+"Timeout waiting for connection from the connection pool. maxPoolSize: n, connections
+in use by cursors: n, connections in use by transactions: n, connections in use by other
+operations: n".
+
+
+Error Handling
+--------------
+
+Initial Handshake Errors
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+When establishing a new connection in load balanced mode, drivers MUST NOT perform SDAM
+error handling for any errors that occur before the MongoDB Handshake
+(i.e. :code:`hello` or legacy hello command) is complete. Errors during the MongoDB
+Handshake MUST also be ignored for SDAM error handling purposes. Once the initial
+handshake is complete, the connection MUST determine its generation number based
+on the :code:`serviceId` field in the handshake response. Any errors that occur
+during the rest of connection establishment (e.g. errors during authentication commands)
+MUST go through the SDAM error handling flow but MUST NOT mark the server as
+:code:`Unknown` and when requiring the connection pool to be cleared, MUST only
+clear connections for the :code:`serviceId`.
+
+Post-Handshake Errors
+^^^^^^^^^^^^^^^^^^^^^^
+
+When the driver is operating in load balanced mode and an application operation receives a
+state change error, the driver MUST NOT make any changes to the :code:`TopologyDescription`
+or the :code:`ServerDescription` of the load balancer (i.e. it MUST NOT mark the load
+balancer as :code:`Unknown`). If the error requires the connection pool to be cleared,
+the driver MUST only clear connections with the same :code:`serviceId` as the connection
+which errored.
+
+
+Events
+------
+
+When in load balancer mode the driver MUST now include the :code:`serviceId` in the
+:code:`CommandStartedEvent`, :code:`CommandSucceededEvent`, and
+:code:`CommandFailedEvent`. The driver MAY decide how to expose this information.
+Drivers that have a :code:`ConnectionId` object for example, MAY choose to provide a
+:code:`serviceId` in that object. The :code:`serviceId` field is only present when
+in load balancer mode and connected to a service that is behind a load balancer.
+
+Additionally the :code:`PoolClearedEvent` MUST also contain a :code:`serviceId`
+field.
+
+
+Downstream Visible Behavioral Changes
+-------------------------------------
+
+Services MAY add a command line option or other configuration parameter, that tells the service
+it is running behind a load balancer. Services MAY also dynamically determine whether they are
+behind a load balancer.
+
+All services which terminate TLS MUST be configured to return a TLS certificate for a hostname
+which matches the hostname the client is connecting to.
+
+All services behind a load balancer that have been started with the aforementioned option MUST
+add a top level :code:`serviceId` field to their response to the :code:`hello` or legacy hello
+command. This field MUST be a BSON :code:`ObjectId` and SHOULD NOT change while the service is running.
+When a driver is configured to not be in load balanced mode and the service is configured behind
+a load balancer, the service MAY return an error from the driver's :code:`hello` command that
+the driver is not configured to use it properly.
+
+All services that have the behaviour of reaping idle cursors after a specified period of time MAY
+also close the connection associated with the cursor when the cursor is reaped. Conversely, those
+services MAY reap a cursor when the connection associated with the cursor is closed.
+
+All services that have the behaviour of reaping idle transactions after a specified period of time
+MAY also close the connection associated with the transaction when the transaction is reaped.
+Conversely, those services must abort a transaction when the connection associated with the
+transaction is closed.
+
+Any applications that connect directly to services and not through the load balancer MUST also
+supply the :code:`loadBalanced=true` option to the driver they use to connect.
+
+
+Q&A
+---
+
+Why use a connection string option instead of a new URI scheme?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Use of a connection string option would allow the driver to continue to use SRV records that
+pointed at a load balancer instead of a replica set without needing to change the URI provided
+to the :code:`MongoClient`. The SRV records could also provide the default :code:`loadBalanced=true`
+in the TXT records.
+
+Why explicitly opt-in to this behaviour instead of letting mongos inform the driver of the load balancer?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Other versions of this design proposed a scheme in which the application does not have to opt-in to
+load balanced mode. Instead, the server would send a special field in :code:`hello` and legacy hello
+command responses to indicate that it was running behind a load balancer and the driver would change
+its behavior accordingly. We opted to take an approach that required code changes instead because
+load balancing changes driver behavior in ways that could cause unexpected application errors, so
+it made sense to have applications consciously opt-in to this mode. For example, connection pinning
+creates new stresses on connection pools because we go from a total of :code:`numMongosServers * maxPoolSize`
+connections to simply maxPoolSize. Furthermore, connections get pinned to open cursors and transactions,
+further straining resource availability. Due to this change, applications may also need to increase
+the configured :code:`maxPoolSize` when opting into this mode.
+
+Why does this specification instruct drivers to not check connections back into the connection pool in some circumstances?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In the case of a load balancer fronting multiple services, it is possible that a connection to the
+load balancer could result in a connection behind the load balancer to a different service. In
+order to guarantee these operations execute on the same service they need to be executed on the
+same socket - not checking a connection back into the pool for the entire operation guarantees this.
+
+What reason has a client side connection reaper for idle cursors not been put into this specification?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+It was discussed as a potential solution for maxed out connection pools that the drivers could
+potentially behave similar to the server and close long running cursors after a specified time
+period and return their connections to the pool. Due to the high complexity of that solution
+it was determined that better error messaging when the connection pool was maxed out would
+suffice in order for users to easily debug when the pool ran out of connections and fix their
+applications or adjust their pool options accordingly.
+
+Why are we requiring mongos servers to add a new serviceId field in hello responses rather than reusing the existing topologyVersion.processId?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This option was previously discussed, but we opted to add a new :code:`hello` response field in
+order to not mix intentions.
+
+Why does this specification not address load balancer restarts or maintenance?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The Layer 4 load balancers that would be in use for this feature lack the
+ability that a layer 7 load balancer could potentially have to be able to
+understand the MongoDB wire protocol and respond to monitoring requests.
+
+
+Design Rationales
+-----------------
+
+Services cannot dynamically switch from running behind a load balancer and not running behind
+a load balancer. Based on that, this design forces the application to opt-in to this behaviour
+and make potential changes that require restarts to their applications. If this were to change,
+see alternative designs below.
+
+
+Alternative Designs
+-------------------
+
+Service PROXY Detection
+^^^^^^^^^^^^^^^^^^^^^^^
+
+An alternative to the driver using a connection string option to put it into load balancing
+mode would be for the service the driver is connected to to inform the driver it is behind
+a load balancer. A possible solution for this would be for all services to understand the
+PROXY protocol such as Data Lake does, and to alter their hello responses to inform the
+driver they are behind a load balancer, potentially with the IP address of the load balancer itself.
+
+The benefit of this solution would be that no changes would be required from the application
+side, and could also not require a restart of any application. A single request to the service
+through the load balancer could automatically trigger the change in the hello response and
+cause the driver to switch into load balancing mode pointing at the load balancer's IP address.
+Also with this solution it would provide services the ability to record the original IP addresses
+of the application that was connecting to it as they are provided the PROXY protocol's header bytes.
+
+The additional complexity of this alternative on the driver side is that instead of starting
+in a single mode and remaining there for the life of the application, the driver would need
+to deal with additional state changes based on the results of the server monitors. From a
+service perspective, every service would need to be updated to understand the PROXY protocol
+header bytes prepended to the initial connection and modify their states and hello responses
+accordingly. Additionally load balancers would need to have additional configuration as noted
+in the reference section below, and only load balancers that support the PROXY protocol would
+be supported.
+
+
+Change Log
+==========
--- a/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
+++ b/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
@@ -8,8 +8,8 @@ Server Discovery And Monitoring
 :Advisors: David Golden, Craig Wilson
 :Status: Accepted
 :Type: Standards
-:Version: 2.21
-:Last Modified: 2020-06-08
+:Version: 2.30
+:Last Modified: 2021-04-12
 
 .. contents::
 
@@ -147,6 +147,7 @@ A server type from which a client can receive application data:
 * RSPrimary
 * RSSecondary
 * Standalone
+* LoadBalanced
 
 Round trip time
 ```````````````
@@ -280,7 +281,7 @@ ServerType
 
 Standalone, Mongos,
 PossiblePrimary, RSPrimary, RSSecondary, RSArbiter, RSOther, RSGhost,
-or Unknown.
+LoadBalancer or Unknown.
 
 See `parsing an ismaster response`_.
 
@@ -471,6 +472,8 @@ Drivers MUST enforce:
   and possibly Single,
   are allowed.
   (See `verifying setName with TopologyType Single`_.)
+* ``loadBalanced=true`` cannot be used in conjunction with
+  ``directConnection=true`` or ``replicaSet``
 
 Handling of SRV URIs resolving to single host
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -595,6 +598,8 @@ are not replica set member states at all.
 |                   | secondary, nor arbiter.                                       |
 +-------------------+---------------------------------------------------------------+
 | RSGhost           | "isreplicaset: true" in response.                             |
++-------------------+---------------------------------------------------------------+
+| LoadBalanced      | "loadBalanced=true" in URI.                                   |
 +-------------------+---------------------------------------------------------------+
 
 A server can transition from any state to any other.  For example, an
@@ -749,6 +754,12 @@ response to the current ServerDescription's topologyVersion:
 See `Replacing the TopologyDescription`_ for an example implementation of
 topologyVersion comparison.
 
+serviceId
+````````
+
+MongoDB 5.0 and later, as well as any mongos-like service, include a ``serviceId``
+field when the service is configured behind a load balancer.
+
 Other ServerDescription fields
 ``````````````````````````````
 
@@ -797,6 +808,13 @@ the ServerDescription in TopologyDescription.servers
 MUST be replaced with the new ServerDescription.
 
 .. _is compatible:
+
+
+TopologyType LoadBalanced
+`````````````````````````
+
+See the `Load Balancer Specification <../load-balancers/load-balancers.rst#server-discovery-and-monitoring>`__ for details.
+
 
 Checking wire protocol compatibility
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -908,7 +926,7 @@ TopologyType explanations
 ~~~~~~~~~~~~~~~~~~~~~~~~~
 
 This subsection complements the `TopologyType table`_
-with prose explanations of the TopologyTypes (besides Single).
+with prose explanations of the TopologyTypes (besides Single and LoadBalanced).
 
 TopologyType Unknown
   A starting state.
@@ -1264,12 +1282,17 @@ implementation::
                 return
 
             if isStateChangeError(error):
-                # Mark the server Unknown
-                unknown = new ServerDescription(type=Unknown, error=error, topologyVersion=topologyVersion)
-                onServerDescriptionChanged(unknown, connection pool for server)
+                # Don't mark server unknown in load balanced mode.
+                if type != LoadBalanced
+                  # Mark the server Unknown
+                  unknown = new ServerDescription(type=Unknown, error=error, topologyVersion=topologyVersion)
+                  onServerDescriptionChanged(unknown, connection pool for server)
                 if isShutdown(code) or (error was from <4.2):
                   # the pools must only be cleared while the lock is held.
-                  clear connection pool for server
+                  if type == LoadBalanced:
+                    clear connection pool for serviceId
+                  else:
+                    clear connection pool for server
                 if multi-threaded:
                     request immediate check
                 else:
@@ -1279,10 +1302,14 @@ implementation::
                     if isNotMaster(error):
                         check failing server
             elif isNetworkError(error) or (not error.completedHandshake and (isNetworkTimeout(error) or isAuthError(error))):
-                # Mark the server Unknown
-                unknown = new ServerDescription(type=Unknown, error=error)
-                onServerDescriptionChanged(unknown, connection pool for server)
-                clear connection pool for server
+                if type != LoadBalanced
+                  # Mark the server Unknown
+                  unknown = new ServerDescription(type=Unknown, error=error)
+                  onServerDescriptionChanged(unknown, connection pool for server)
+                  clear connection pool for server
+                else
+                  if serviceId
+                    clear connection pool for serviceId
                 # Cancel inprogress check
                 cancel monitor check
 
@@ -1369,8 +1396,12 @@ we distinguish two phases of connecting to a server and using it for application
 
 If there is a network error or timeout on the connection before the handshake completes,
 the client MUST replace the server's description
-with a default ServerDescription of type Unknown,
-and fill the ServerDescription's error field with useful information.
+with a default ServerDescription of type Unknown when the TopologyType is not
+LoadBalanced, and fill the ServerDescription's error field with useful information.
+
+If there is a network error or timeout on the connection before the handshake completes,
+and the TopologyType is LoadBalanced, the client MUST keep the ServerDescription
+as LoadBalancer.
 
 If there is a network timeout on the connection after the handshake completes,
 the client MUST NOT mark the server Unknown.
@@ -1378,8 +1409,8 @@ the client MUST NOT mark the server Unknown.
 rather than an unavailable server.)
 If, however, there is some other network error on the connection after the
 handshake completes, the client MUST replace the server's description
-with a default ServerDescription of type Unknown,
-and fill the ServerDescription's error field with useful information,
+with a default ServerDescription of type Unknown if the TopologyType is not
+LoadBalanced, and fill the ServerDescription's error field with useful information,
 the same as if an error or timeout occurred before the handshake completed.
 
 When the client marks a server Unknown due to a network error or timeout,
@@ -1516,8 +1547,8 @@ Authentication errors
 ~~~~~~~~~~~~~~~~~~~~~
 
 If the authentication handshake fails for a connection, drivers MUST mark the
-server Unknown and clear the server's connection pool. (See
-`Why mark a server Unknown after an auth error?`_)
+server Unknown and clear the server's connection pool if the TopologyType is
+not LoadBalanced. (See `Why mark a server Unknown after an auth error?`_)
 
 Monitoring SDAM events
 ''''''''''''''''''''''
@@ -2483,3 +2514,5 @@ mark the server Unknown and clear the pool.
 .. _Connection Monitoring and Pooling spec: /source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
 .. _CMAP spec: /source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
 .. _Authentication spec: /source/auth/auth.rst
+
+2021-4-12: Adding in behaviour for load balancer mode.
--- a/source/sessions/driver-sessions.rst
+++ b/source/sessions/driver-sessions.rst
@@ -3,7 +3,7 @@ Driver Sessions Specification
 =============================
 
 :Spec Title: Driver Sessions Specification (See the registry of specs)
-:Spec Version: 1.7.2
+:Spec Version: 1.8.0
 :Author: Robert Stam
 :Spec Lead: A\. Jesse Jiryu Davis
 :Advisory Group: Jeremy Mikola, Jeff Yemin, Samantha Ritter
@@ -12,7 +12,7 @@ Driver Sessions Specification
 :Status: Accepted (Could be Draft, Accepted, Rejected, Final, or Replaced)
 :Type: Standards
 :Minimum Server Version: 3.6 (The minimum server version this spec applies to)
-:Last Modified: 2021-04-08
+:Last Modified: 2021-04-12
 
 .. contents::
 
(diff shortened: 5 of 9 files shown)


source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+emit only 1 :code:`ConnectionCheckOutStartedEvent`, and only 1 :code:`ConnectionCheckedOutEvent`
+or :code:`ConnectionCheckOutFailedEvent`. Similarly, the driver MUST only publish 1
+:code:`ConnectionCheckedInEvent`.
+
+For single threaded drivers that do not use a connection pool, the driver MUST have only 1
+socket connection to the load balancer in load balancing mode.
+
+In the case of network errors for all operations that create a cursor, the driver MUST behave
+as defined in the Retryable Reads specification. If a getMore fails with a network error,
+drivers MUST unpin the connection that’s pinned to the cursor and return it to the pool
```
divjotarora, line 227: We changed the rule about unpinning the connection after a network error on `getMore`. What do you think of this proposed wording:

> If a ``getMore`` fails with a network error, drivers MUST leave the connection pinned to the cursor. When the cursor is explicitly closed, drivers MUST NOT execute a ``killCursors`` command because the pinned connection is no longer valid and MUST return the connection back to the pool.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+so it does not count against maxPoolSize. Future calls to close() on the cursor MUST NOT
+attempt to execute a killCursors command on the server.
+
+When the driver is in load balancing mode, drivers MUST NOT check the connection back into
+the connection pool while executing a transaction. The driver MUST continue to use the
+same connection to execute all commands for a given transaction with the exception
+of ::code:`commitTransaction` and :code:`abortTransaction`, if either one of these
```
divjotarora, line 234: This wording implies that the rules in the paragraph below only apply to `commitTransaction` and `abortTransaction`, but the rule about unpinning on errors with the `TransientTransactionError` label actually applies to *all* commands in a transaction. We should probably also clarify that the rules for unpinning connections are the same as those for unpinning servers in non-LB sharded transactions and maybe link back to the transactions spec for those rules rather than duplicating them here.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+:code:`commitTransaction` command attempt fails with an :code:`UnknownTransactionCommitResult`
+error label. In cases where the :code:`UnknownTransactionCommitResult` causes an automatic
+retry attempt, drivers MUST unpin the :code:`ClientSession` before performing server selection
+for the retry.
+
+Starting a new transaction on a pinned :code:`ClientSession` MUST unpin the session.
+Additionally, any non-transaction operation using a pinned :code:`ClientSession` MUST unpin
+the session and the operation MUST perform normal server selection.*
+
+In the case of a network errors in transactions, the driver MUST implement the following cases:
```
divjotarora, line 254: I think we can safely remove any information about network errors in transactions because it should already be covered by the prose in the transactions spec about adding the `TransientTransactionError` label and unpinning the server/connection when an error with this label occurs.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+**Error Handling**
+------------------
+
+When the driver is in load balanced mode and encounters any error described here as a state
+change error, the driver MUST NOT make any changes to the :code:`TopologyDescription` or the
+:code:`ServerDescription` of the load balancer (i.e. it MUST NOT mark the load balancer as Unknown).
+If the error requires the connection pool to be cleared, the driver MUST only clear connections
+with the same :code:`serviceId` as the connection which errored. The connection pool MUST emit a
+:code:`PoolClearedEvent` event with the :code:`serviceId` of the connection that errored.
+The :code:`PoolClearedEvent` MUST have a new field of :code:`serviceId` with type :code:`ObjectId`.
```
divjotarora, line 288: This is already mentioned in the CMAP spec and can be removed from here.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+
+If there is a network error or timeout on the connection before the initial handshake completes,
```
divjotarora, line 290: Change "initial handshake" to "initial hello or legacy hello command" for clarity.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+The driver MUST also close the associated connection. If the network error occurred on the handshake’s
+:code:`hello` command, the connection pool MUST NOT be cleared. If the network error occurred after
+the handshake’s hello command, the connection POOL must clear close all connections with the matching
+:code:`serviceId`.
+
+
+**Implementation Notes**
+------------------------
+
+All connections MUST track a generation number. For non-load balanced deployments, the connection’s
```
divjotarora, line 301: As I mentioned in the CMAP spec comment, this information about generation numbers should probably go somewhere else. Not sure where all of it goes, though.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -455,6 +456,9 @@ It is allowed to use ``directConnection=true`` in conjunction with the
 verify that setName matches the specified name, as per
 `verifying setName with TopologyType Single`_.
 
+When the ``replicaSet`` URI option is provided, the driver MUST throw an
```
divjotarora, line 459: I'd prefer not to duplicate this here as it's already mentioned in the LB spec.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -466,11 +470,14 @@ Allowed configuration combinations
 Drivers MUST enforce:
 
 * TopologyType Single cannot be used with multiple seeds.
-* ``directConnection=true`` cannot be used with multiple seeds.
+* ``directConnection=true`` and ``loadBalanced=true`` cannot be
```
divjotarora, line 473: Same comment about avoiding duplication. We discuss `directConnection` validation here because there's no spec for it, but the new info is already expressed in the LB spec.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -596,6 +630,8 @@ are not replica set member states at all.
 +-------------------+---------------------------------------------------------------+
 | RSGhost           | "isreplicaset: true" in response.                             |
 +-------------------+---------------------------------------------------------------+
+| LoadBalanced      | "loadBalanced=true" in URI and "serviceId" in response.       |
```
divjotarora, line 633: Should we remove the "serviceId in response" requirement? There's no monitoring in LB mode and the topology transitions from Unknown to LoadBalanced before any commands are sent to the server, so the server responses are not used to determine SDAM transitions.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1259,6 +1308,10 @@ implementation::
         topologyVersion = error.topologyVersion
 
         with client.lock:
+            # Don't perform any changes if type=LoadBalanced
```
divjotarora, line 1311: I don't think this is correct for a couple of reasons:

1. For post-handshake state change errors, we only want to skip marking the server Unknown. We still want to clear the connection pool using the serviceId of the errored connection.
2. For pre-handshake errors, we want to clear the connection pool if the error occurred after the initial hello/legacy hello command completed.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -684,6 +684,12 @@ eagerly so that any operations waiting on `Connections <#connection>`_ can retry
 as soon as possible. The pool MUST NOT rely on WaitQueueTimeoutMS to clear
 requests from the WaitQueue.
 
+Load Balancer Mode
```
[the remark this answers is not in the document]
durran: Added to CMAP spec and linked back as well.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -765,6 +771,11 @@ Events
        *  The ServerAddress of the Endpoint the pool is attempting to connect to.
        */
       address: string;
+
+      /**
+       * The server id for which the pool was cleared for in load balancing mode.
```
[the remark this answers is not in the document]
durran: Done.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+:Status: Accepted
+:Type: Standards
+:Minimum Server Version: 5.0
+:Last Modified: 2021-04-06
+
+.. contents::
+
+--------
+
+**Abstract**
```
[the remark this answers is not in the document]
durran: Unbolded.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+interpreted as described in `RFC 2119 <https://www.ietf.org/rfc/rfc2119.txt>`__.
+
+**Specification**
+=================
+
+
+**Terms**
+---------
+
+**The PROXY Protocol**
```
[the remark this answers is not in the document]
durran: Removed.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+`The PROXY protocol <http://www.haproxy.org/download/1.8/doc/proxy-protocol.txt>`__
+provides a convenient way to safely transport connection information such
+as a client's address across multiple layers of NAT or TCP proxies.
+It is designed to require little changes to existing components and to limit the
+performance impact caused by the processing of the transported information.
+
+**SDAM**
+^^^^^^^^
+
+An abbreviated form of “Server Discovery and Monitoring”, specification defined
```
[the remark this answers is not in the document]
durran: Changed all instances of this as well as single quotes.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+in `this document <https://bit.ly/3fsOEmo>`__.
+
+**Service**
+^^^^^^^^^^^
+
+Any MongoDB service that can run behind a load balancer, including but not
+limited to mongos, Atlas Proxy, Elastic Front End.
+
+
+**Connection String Options**
```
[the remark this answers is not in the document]
durran: Added option to URI options spec.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+
+- The connection string contains more than one host/port.
+- The connection string contains a replicaSet option.
+- The connection string contains a directConnection option.
+
+
+**DNS Seedlist Discovery**
+--------------------------
+
+**Default Connection String Options**
```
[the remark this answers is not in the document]
durran: Initial dns seedlist discovery spec updated.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A default connection string option for :code:`loadBalanced=true` MUST be valid in a
+TXT record, and driver validation of the TXT records MUST allow for this option to
+also be a part of the connection string.
+
+When :code:`loadBalanced=true` is an option in the TXT record, the driver MUST
```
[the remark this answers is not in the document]
durran: Text added.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+validate the connection string as defined in the URI Validation section.
+
+When :code:`loadBalanced=true` is an option in the TXT record, the driver MUST
+NOT poll for changes in the SRV record.
+
+
+**Command Monitoring**
```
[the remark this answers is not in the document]
durran: Done.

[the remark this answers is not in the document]
durran: Changed to events section and linked back to it.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+**Server Discovery and Monitoring**
+-----------------------------------
+
+**Monitoring**
+^^^^^^^^^^^^^^
+
+In the case of the driver having the :code:`loadBalanced=true` connection string option
+specified, every pooled connection MUST add a loadBalanced field to the hello command
+in its `handshake <https://bit.ly/3w9VGCC>`__. The value of the field MUST be true.
+When the driver is in load balanced mode, the value of this parameter MUST be true.
```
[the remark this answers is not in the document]
durran: Yes good catch. Removed the previous.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+
+A new :code:`TopologyType` MUST be implemented to indicate to the driver that it is
+connected to a topology that is deployed behind a load balancer. The
+:code:`TopologyType` MUST be named :code:`LoadBalanced`. A new :code:`ServerType` MUST
+be implemented to represent a load balancer. The :code:`ServerType` MUST be named
+:code:`LoadBalancer`.
+
+When :code:`loadBalanced=true` the topology MUST start in type :code:`LoadBalanced`
```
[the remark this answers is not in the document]
durran: Done.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+and MUST remain as :code:`LoadBalanced` indefinitely. The topology MUST contain 1
+:code:`ServerDescription` with a :code:`ServerType` of :code:`LoadBalancer`. The
+"address" field of the :code:`ServerDescription` MUST be set to the address field
+of the load balancer. All other fields in the :code:`ServerDescription` MUST remain unset.
+
+When :code:`loadBalanced=true`, the driver MUST NOT start a monitoring connection,
```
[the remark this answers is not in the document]
durran: Merged into the previous paragraph.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+however drivers MUST emit the following series of SDAM events:
+
+- :code:`TopologyOpeningEvent` when the topology is created.
+- :code:`TopologyDescriptionChangedEvent`. The :code:`previousDescription` field MUST
+  have :code:`TopologyType` :code:`Unknown` and no servers. The :code:`newDescription`
+  MUST have :code:`TopologyType` :code:`LoadBalanced` and one server with
+  :code:`ServerType` :code:`Unknown`.
+- :code:`ServerOpeningEvent` when the server representing the load balancer is created.
+- :code:`ServerDescription`ChangedEvent. The :code:`previousDescription` MUST have
```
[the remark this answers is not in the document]
durran: Fixed.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+:code:`serviceId` in that object.
+
+
+**Server Discovery and Monitoring**
+-----------------------------------
+
+**Monitoring**
+^^^^^^^^^^^^^^
+
+In the case of the driver having the :code:`loadBalanced=true` connection string option
```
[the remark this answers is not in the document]
durran: Combined and moved.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+
+Drivers MUST also emit a :code:`ServerClosedEvent` and :code:`TopologyClosedEvent` when
+the topology is closed and MUST NOT emit any other events when operating in this mode.
+
+When :code:`loadBalanced=true` and the server’s hello response does not contain a
+:code:`serviceId` field, the driver MUST throw an exception with the message
+*“Driver attempted to initialize in load balancing mode, but the server does not
+support this mode.”*
+
+When :code:`loadBalanced=false` or the option is not present, the driver MUST NOT
```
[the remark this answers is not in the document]
durran: Moved.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+change any existing behaviour when connected to a non-load balanced service.
+If the driver is connected to a service that is configured behind a load balancer,
+and the service supports running behind a load balancer, the service MAY return an
+error that the driver is not configured to use it properly.
+
+If the :code:`loadBalanced=true` connection string option is not specified, the
```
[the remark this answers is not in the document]
durran: Removed.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+
+.. code:: typescript
+
+    { hello: 1 }
+
+Driver connection string contains :code:`loadBalanced=true`:
+
+.. code:: typescript
+
+    { hello: 1, loadBalanced: 1 }
```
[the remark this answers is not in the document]
durran: Yes. Changed.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+
+
+**Server Discovery and Monitoring**
+-----------------------------------
+
+**Monitoring**
+^^^^^^^^^^^^^^
+
+In the case of the driver having the :code:`loadBalanced=true` connection string option
+specified, every pooled connection MUST add a loadBalanced field to the hello command
```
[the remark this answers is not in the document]
durran: Yup. Updated.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+A :code:`ServerType` of :code:`LoadBalancer` MUST be considered a data-bearing server.
+
+
+**Driver Sessions**
+-------------------
+
+**How To Check Whether A Deployment Supports Sessions**
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Having verified in step 1 that the :code:`TopologyDescription` includes at least one
```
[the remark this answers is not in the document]
durran: Done.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+**Server Selection**
+--------------------
+
+Topology type: Load Balanced
+
+A deployment of topology type Load Balanced contains one server of type :code:`LoadBalancer`.
+
+For read and write operations, the single server in the topology MUST always be selected.
+
+For read operations the :code:`mode`, :code:`tag_sets`, and :code:`maxStalenessSeconds`
```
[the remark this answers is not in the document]
durran: Changed to suggested paragraph.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,564 @@
...
+provides a convenient way to safely transport connection information such
+as a client's address across multiple layers of NAT or TCP proxies.
+It is designed to require little changes to existing components and to limit the
+performance impact caused by the processing of the transported information.
+
+**SDAM**
+^^^^^^^^
+
+An abbreviated form of “Server Discovery and Monitoring”, specification defined
+in `this document <https://bit.ly/3fsOEmo>`__.
```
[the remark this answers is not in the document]
durran: Changed and updated all new links in modified specs to be relative.

source/load-balancers/load-balancers.rst
durran: I like that suggestion. Updated to that.

source/load-balancers/load-balancers.rst
durran: Linked back to transactions spec for these cases.

source/load-balancers/load-balancers.rst
durran: Done.

source/load-balancers/load-balancers.rst
durran: Removed.

source/load-balancers/load-balancers.rst
durran: Done.

source/load-balancers/load-balancers.rst
durran: Have noted this in the CMAP spec now.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
durran: Removed.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
durran: Removed.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
durran: Removed.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
durran: Changed logic to simply not mark server unknown instead inside the next block.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+:Last Modified: 2021-04-06
+
+.. contents::
+
+--------
+
+Abstract
+========
+
+This specification defines driver behaviour when connected to mongodb services
```
divjotarora, line 21: nit: mongodb -> MongoDB for consistency

divjotarora (commented):
I haven't looked through the revised version of the SDAM spec changes yet, so these comments are only for changes to other existing specs and the new LB spec.

divjotarora:
@durran Reminder to update either the LB spec to mention that session expiration should not be checked in LB mode.

source/command-monitoring/command-monitoring.rst
```diff
@@ -292,6 +294,13 @@ API
      * The name of this field is flexible to match the object that is returned from the driver.
      */
     connectionId: ConnectionId;
+
+    /**
+     * Returns the service id for the command when the driver is in load balancer mode.
+     * For drivers that wish to include this in their ConnectionId object, this field is
+     * optional.
+     */
+    serviceId: ObjectId;
```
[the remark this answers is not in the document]
durran: Updated.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -684,6 +694,25 @@ eagerly so that any operations waiting on `Connections <#connection>`_ can retry
...
 requests from the WaitQueue.
 
+Load Balancer Mode
+------------------
+
+For load-balanced deployments, pools MUST maintain a map from ``serviceId`` to a
+tuple of (generation, connection count) where the connection count refers to the
+total number of connections that exist for a specific ``serviceId``. The pool MUST
+remove the entry for a ``serviceId`` once the connection count reaches 0.
+Connection attempts MUST ignore any errors encountered when dialing a socket
```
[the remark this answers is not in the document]
durran: Updated.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+^^^^^^^
+
+Any MongoDB service that can run behind a load balancer, including but not
+limited to mongos, Atlas Proxy, Elastic Front End.
+
+
+Connection String Options
+-------------------------
+
+To specify to the driver to operate in load balancing mode, a connection string
```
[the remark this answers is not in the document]
durran: Renamed and subsections split.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+
+
+DNS Seedlist Discovery
+----------------------
+
+Default Connection String Options
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A default connection string option for :code:`loadBalanced=true` MUST be valid in a
+TXT record, and driver validation of the TXT records MUST allow for this option to
```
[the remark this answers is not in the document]
durran: That was rather redundant. :) I reworded.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+validate the connection string as defined in the URI Validation section.
+
+When :code:`loadBalanced=true` is an option in the TXT record, the driver MUST
+NOT poll for changes in the SRV record.
+
+
+Events
+------
+
+When in load balancer mode the driver MUST now include the :code:`serviceId` in the
```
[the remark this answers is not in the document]
durran: Moved down below error handling.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+- Connections checked out for operations not falling under the previous 2 categories
+
+When the connection pool's :code:`maxPoolSize` is reached and the pool times out waiting
+for a new connection the :code:`WaitQueueTimeoutError` MUST include a new detailed message,
+as well as the published :code:`ConnectionCheckOutFailedEvent`: "Timeout waiting for
+connection from the connection pool. maxPoolSize: n, connections in use by cursors: n,
+connections in use by transactions: n, connections in use by other operations: n".
+
+
+Error Handling
```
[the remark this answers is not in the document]
durran: Split up the two sections and renamed second to initial handshake errors.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+also close the connection associated with the cursor when the cursor is reaped. Conversely, those
+services MAY reap a cursor when the connection associated with the cursor is closed.
+
+All services that have the behaviour of reaping idle transactions after a specified period of time
+MUST also close the connection associated with the transaction when the transaction is reaped.
+Conversely, those services must abort a transaction when the connection associated with the
+transaction is closed.
+
+When a driver is configured to not be in load balanced mode and the service is configured behind
+a load balancer, the service MUST return an error from the driver's :code:`hello` command that
```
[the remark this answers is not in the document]
durran: Changed.

source/sessions/driver-sessions.rst
```diff
@@ -1262,3 +1263,5 @@ Change log
 :2019-10-22: Drivers may defer checking if a deployment supports sessions until the first
 operation performed with a session
 :2020-05-26: Simplify logic for determining sessions support
+:2020-05-26: Simplify logic for determining sessions support
```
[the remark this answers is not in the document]
durran: That was an accident - I've removed the duplicate.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+
+For single threaded drivers that do not use a connection pool, the driver MUST have only 1
+socket connection to the load balancer in load balancing mode.
+
+If a :code:`getMore` fails with a network error, drivers MUST leave the connection pinned
+to the cursor. When the cursor is explicitly closed, drivers MUST NOT execute a :code:`killCursors`
+command because the pinned connection is no longer valid and MUST return the connection
+back to the pool.
+
+When the driver is in load balancing mode, drivers MUST NOT check the connection back into
```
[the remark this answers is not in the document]
durran: Updated to suggestion.

source/load-balancers/load-balancers.rst
durran: Done.

durran:
> @durran Reminder to update either the LB spec to mention that session expiration should not be checked in LB mode.

Added a note in the LB spec as well as the sessions spec in the algorithm with link back to the LB spec.

source/load-balancers/load-balancers.rst
durran: I've removed it now.

source/load-balancers/load-balancers.rst
durran: A missed that. Now removed.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+  :code:`ServerType` :code:`Unknown`. The :code:`newDescription` MUST have
+  :code:`ServerType` :code:`LoadBalancer`.
+- :code:`TopologyDescriptionChangedEvent`. The :code:`newDescription` MUST have
+  :code:`TopologyType` :code:`LoadBalanced` and one server with :code:`ServerType`
+  :code:`LoadBalancer`.
+
+Drivers MUST also emit a :code:`ServerClosedEvent` and :code:`TopologyClosedEvent` when
+the topology is closed and MUST NOT emit any other events when operating in this mode.
+
+Example:
```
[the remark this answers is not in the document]
durran: Moved into `Connection Pooling` under the `Connection Establishment` subsection.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+^^^^^^^^^^^^^^
+
+When :code:`loadBalanced=true` is provided in the connection string, the driver
+MUST throw an exception in the following cases:
+
+- The connection string contains more than one host/port.
+- The connection string contains a replicaSet option.
+- The connection string contains a directConnection option.
+
+If a URI is provided with the ``mongodb+srv`` scheme, the driver SHOULD first to the
```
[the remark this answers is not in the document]
durran: Both changed.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+
+
+Connection Pooling
+------------------
+
+In the case of the driver having the :code:`loadBalanced=true` connection string option
+specified, every pooled connection MUST add a loadBalanced field to the hello or legacy
+hello command in its `handshake <../mongodb-handshake/handshake.rst#connection-handshake>`__.
+The value of the field MUST be true.
+When the driver is in load balanced mode, the value of this parameter MUST be true.
```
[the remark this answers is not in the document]
durran: Removed.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+A deployment of topology type Load Balanced contains one server of type :code:`LoadBalancer`.
+
+For read and write operations, the single server in the topology MUST always be selected.
+
+During command construction, the LoadBalancer server MUST be treated like a mongos and
+drivers MUST add a $readPreference field to the command when required by
+`Passing read preference to mongos and load balancers <../server-selection/server-selection.rst#passing-read-preference-to-mongos-and-load-balancers>`_.
+
+
+Connection Pooling
```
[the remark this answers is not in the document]
durran: I actually split this up into 4 subsections now: `Connection Establishment`, `Behaviour With Cursors`, `Connection Pinning` and `Connection Tracking`

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+hello command in its `handshake <../mongodb-handshake/handshake.rst#connection-handshake>`__.
+The value of the field MUST be true.
+When the driver is in load balanced mode, the value of this parameter MUST be true.
+When the server's hello response does not contain a :code:`serviceId` field,
+the driver MUST throw an exception with the message "Driver attempted to initialize
+in load balancing mode, but the server does not support this mode."
+
+When :code:`loadBalanced=false` or the option is not present, the driver MUST NOT
+change any existing behaviour when connected to a non-load balanced service.
+If the driver is connected to a service that is configured behind a load balancer,
```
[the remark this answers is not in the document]
durran: Removed.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+and the service supports running behind a load balancer, the service MAY return an
+error that the driver is not configured to use it properly.
+
+When the driver is in load balancing mode and executing any cursor-initiating command, the driver
+MUST NOT check the connection back into the connection pool until the initial cursor
```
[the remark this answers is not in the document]
durran: Suggestion brought in, much clearer, thanks.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+
+For drivers that implement a connection pool and when using a pinned connection, the driver MUST
+emit only 1 :code:`ConnectionCheckOutStartedEvent`, and only 1 :code:`ConnectionCheckedOutEvent`
+or :code:`ConnectionCheckOutFailedEvent`. Similarly, the driver MUST only publish 1
+:code:`ConnectionCheckedInEvent`.
+
+For single threaded drivers that do not use a connection pool, the driver MUST have only 1
+socket connection to the load balancer in load balancing mode.
+
+If a :code:`getMore` fails with a network error, drivers MUST leave the connection pinned
```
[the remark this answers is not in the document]
durran: Moved and changed the explicit wording.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+The driver connection pool MUST track the purpose for which connections are checked out
+in the following 3 categories:
+
+- Connections checked out for cursors
+- Connections checked out for transactions
+- Connections checked out for operations not falling under the previous 2 categories
+
+When the connection pool's :code:`maxPoolSize` is reached and the pool times out waiting
+for a new connection the :code:`WaitQueueTimeoutError` MUST include a new detailed message,
+as well as the published :code:`ConnectionCheckOutFailedEvent`: "Timeout waiting for
```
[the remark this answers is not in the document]
durran: I've reworded to indicate both events must be published with the same error message.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,553 @@
...
+the driver MUST NOT make any changes to the :code:`TopologyDescription` or mark the server as Unknown.
+The driver MUST also close the associated connection. If the network error occurred on the handshake's
+:code:`hello` command, the connection pool MUST NOT be cleared. If the network error occurred after
+the handshake's hello command, the connection POOL must clear close all connections with the matching
+:code:`serviceId`.
+
+Post-Handshake Errors
+^^^^^^^^^^^^^^^^^^^^^^
+
+When the driver is in load balanced mode and encounters any error described here as a state
```
divjotarora, line 266: The "described here" sentence doesn't link to anything. Maybe we can avoid linking at all by saying "When the driver is operating in load balanced mode and an application operation receives a state change error, the driver MUST NOT...".

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,553 @@
...
+in load balancer mode and connected to a service that is behind a load balancer.
+
+Additionally the :code:`PoolClearedEvent` MUST also contain a :code:`serviceId`
+field.
+
+
+Implementation Notes
+--------------------
+
+All connections MUST track a generation number. For non-load balanced deployments, the connection's
```
divjotarora, line 290: Most of the information in this section is duplicated across parts of this spec and other specs like CMAP. Can we disperse it into the relevant places and basically remove the "Implementation Notes" section completely?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -482,6 +485,15 @@ driver MUST start in Unknown topology, and follow the rules in the
 the driver MUST NOT use the number of hosts from the initial SRV lookup
 to decide what topology to start in.
 
+A default connection string option for ``loadBalanced=true`` MUST be valid
```
divjotarora, line 488: Both this paragraph and the next are duplicated from the LB spec. Can they be removed from here?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -798,6 +818,13 @@ MUST be replaced with the new ServerDescription.
 
 .. _is compatible:
 
+
+TopologyType LoadBalanced
+`````````````````````````
+
+See the `Load Balancer Specification <https://github.com/mongodb/specifications/blob/master/source/load-balancers/load-balancers.rst#server-discovery-and-monitoring>`__ for details.
```
divjotarora, line 825: This should be a relative URI for consistency: `../load-balancers/load-balancers.rst#server-discovery-and-monitoring`.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1264,9 +1291,11 @@ implementation::
                 return
 
             if isStateChangeError(error):
-                # Mark the server Unknown
-                unknown = new ServerDescription(type=Unknown, error=error, topologyVersion=topologyVersion)
-                onServerDescriptionChanged(unknown, connection pool for server)
+                # Don't mark server unknown in load balanced mode.
+                if type!=LoadBalanced
```
divjotarora, line 1295: nit: Add spacing to make this `if type != LoadBalanced`. Also, should there be a corresponding check in the `elif` code block?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1369,17 +1398,21 @@ we distinguish two phases of connecting to a server and using it for application
...
 If there is a network error or timeout on the connection before the handshake completes,
 the client MUST replace the server's description
-with a default ServerDescription of type Unknown,
-and fill the ServerDescription's error field with useful information.
+with a default ServerDescription of type Unknown when the TopologyType is not
+LoadBalanced, and fill the ServerDescription's error field with useful information.
+
+If there is a network error or timeout on the connection before the handshake completes,
+and the TopologyType is LoadBalanced, the client MUST keep the ServerDescription
+as LoadBalancer and fill the ServerDescription's error field with useful information.
```
divjotarora, line 1406: Why are we modifying the description's `error` field? The server remains selectable and we don't publish `ServerDescriptionChanged` events, so the error would never be surfaced to the user as part of a server selection error or SDAM event. It seems simpler to leave the entire description as-is.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1516,8 +1549,12 @@ Authentication errors
 ~~~~~~~~~~~~~~~~~~~~~
 
 If the authentication handshake fails for a connection, drivers MUST mark the
-server Unknown and clear the server's connection pool. (See
-`Why mark a server Unknown after an auth error?`_)
+server Unknown and clear the server's connection pool if the TopologyType is
+not LoadBalanced. (See `Why mark a server Unknown after an auth error?`_)
+
+If the authentication handshake fails for a connection and the TopologyType is
```
divjotarora, line 1555: Mentioned previously as well: this makes it sound like we're unconditionally clearing the pool. What we actually want is for the error to go through all of the checks in the SDAM error handling process and then clear the pool if necessary. Also, this paragraph wasn't here before, so can we just remove it? It doesn't add any LB-specific info.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,553 @@
...
+SDAM
+^^^^
+
+An abbreviated form of "Server Discovery and Monitoring", specification defined
+in `Server Discovery and Monitoring Specification <../server-discovery-and-monitoring/server-discovery-and-monitoring.rst>`__.
+
+Service
+^^^^^^^
+
+Any MongoDB service that can run behind a load balancer, including but not
```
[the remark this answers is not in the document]
durran: I've updated to just be any service behind a load balancer.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,553 @@
...
+MUST throw an exception in the following cases:
+
+- The connection string contains more than one host/port.
+- The connection string contains a replicaSet option.
+- The connection string contains a directConnection option.
+
+If a URI is provided with the ``mongodb+srv`` scheme, the driver MUST first do the
+SRV and TXT lookup and then perform the validation. For drivers that do SRV
+lookup asynchrounously this may result in a ``MongoClient`` being instantiated
+but erroring later.
```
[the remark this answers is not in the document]
durran: Updated.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,553 @@
...
+
+
+DNS Seedlist Discovery
+^^^^^^^^^^^^^^^^^^^^^^
+
+The connection string option for :code:`loadBalanced=true` MUST be valid in a
+TXT record and when present MUST be validated as defined in the URI Validation section.
+
+When :code:`loadBalanced=true` is an option in the TXT record, the driver MUST
```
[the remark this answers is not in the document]
durran: Updated to suggestion.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,553 @@
...
+NOT poll for changes in the SRV record.
+
+
+Server Discovery and Monitoring
+-------------------------------
+
+Monitoring
+^^^^^^^^^^
+
+When :code:`loadBalanced=true` is specified in the URI the topology MUST start
```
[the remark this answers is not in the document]
durran: Updated that this must always be true.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,553 @@
...
+  :code:`ServerType` :code:`LoadBalancer`.
+- :code:`TopologyDescriptionChangedEvent`. The :code:`newDescription` MUST have
+  :code:`TopologyType` :code:`LoadBalanced` and one server with :code:`ServerType`
+  :code:`LoadBalancer`.
+
+Drivers MUST also emit a :code:`ServerClosedEvent` and :code:`TopologyClosedEvent` when
+the topology is closed and MUST NOT emit any other events when operating in this mode.
+
+
+Initial Handshake
```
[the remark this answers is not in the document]
durran: I missed the top level one, so I removed that now and put this in the driver sessions subsection.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,553 @@
...
+Data-Bearing Server Type
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+A :code:`ServerType` of :code:`LoadBalancer` MUST be considered a data-bearing server.
+
+
+Driver Sessions
+---------------
+
+How To Check Whether A Deployment Supports Sessions
```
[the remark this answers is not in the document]
durran: Fine with Sessions Support, so renamed.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,553 @@
...
+
+Connection Pinning
+^^^^^^^^^^^^^^^^^^
+
+For drivers that implement a connection pool and when using a pinned connection, the driver MUST
+emit only 1 :code:`ConnectionCheckOutStartedEvent`, and only 1 :code:`ConnectionCheckedOutEvent`
+or :code:`ConnectionCheckOutFailedEvent`. Similarly, the driver MUST only publish 1
+:code:`ConnectionCheckedInEvent`.
+
+When executing a transaction in load balancing mode, drivers MUST follow the rules outlined
```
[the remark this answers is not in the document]
durran: Done.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,553 @@
...
+
+If a :code:`getMore` fails with a network error, drivers MUST leave the connection pinned
+to the cursor. When the cursor's :code:`close` method is invoked, drivers MUST NOT execute
+a :code:`killCursors` command because the pinned connection is no longer valid and MUST
+return the connection back to the pool.
+
+Connection Pinning
+^^^^^^^^^^^^^^^^^^
+
+For drivers that implement a connection pool and when using a pinned connection, the driver MUST
```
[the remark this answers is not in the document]
durran: Moved up with the suggested wording.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,553 @@
...
+same message.
+
+
+Error Handling
+--------------
+
+Initial Handshake Errors
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+If there is a network error or timeout on the connection before the hello or legacy hello completes,
```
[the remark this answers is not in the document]
durran: Updated per suggestion.

source/load-balancers/load-balancers.rst
durran: Updated per suggestion.

source/load-balancers/load-balancers.rst
durran: I've removed the implementation notes now, and stuck with the duplicated information in the CMAP spec and added the error handling notes in that spec below it.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
durran: Removed.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
durran: Updated to relative URI.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
durran: Made formatting change and added the logic to the `elif` block as well.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
durran: Removed the bit about updating the error field.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
durran: Removed.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,535 @@
...
+
+- Connections checked out for cursors
+- Connections checked out for transactions
+- Connections checked out for operations not falling under the previous 2 categories
+
+When the connection pool's :code:`maxPoolSize` is reached and the pool times out waiting
+for a new connection the :code:`WaitQueueTimeoutError` MUST include a new detailed message,
+"Timeout waiting for connection from the connection pool. maxPoolSize: n, connections
+in use by cursors: n, connections in use by transactions: n, connections in use by other
+operations: n". A :code:`ConnectionCheckOutFailedEvent` MUST also be published with the
```
divjotarora, line 245: Per the CMAP spec, `ConnectionCheckoutFailedEvent` events have `address` and `reason` fields. The `reason` is an enum and there's nothing about using custom strings. How do we expect drivers to include this new error message in the event as required by the last sentence of this paragraph?

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,535 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+When establishing a new connection in load balanced mode, drivers MUST NOT perform SDAM
+error handling for any errors that occur before the MongoDB Handshake
+(i.e. :code:`hello` or legacy hello command) is complete. Errors during the MongoDB
+Handshake MUST also be ignored for SDAM error handling purposes. Once the initial
+handshake is complete, the connection MUST determine its generation number based
+on the :code:`serviceId` field in the handshake response. Any errors that occur
+during the rest of connection establishment (e.g. errors during authentication commands)
+MUST go through the SDAM error handling flow but MUST NOT mark the server as
```
divjotarora, line 262: Forgot to include this in my previous suggestion, but we should also say that only connections with matching `serviceId` should be cleared if a pool clearing is required as part of the SDAM error handling process. The `Post-Handshake Errors` section correctly mentions this so you can match that.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,535 @@
...
+
+
+Downstream Visible Behavioral Changes
+-------------------------------------
+
+Services MAY add a command line option or other configuration parameter, that tells the service
+it is running behind a load balancer. Services MAY also dynamically determine whether they are
+behind a load balancer.
+
+When the server supports the :code:`loadBalanced` option, the server MUST respond to the command
```
divjotarora, line 297: This information is duplicated in the fourth paragraph. I recommend removing this paragraph because the other is more descriptive.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,535 @@
...
+All services behind a load balancer that have been started with the aforementioned option MUST
+add a top level :code:`serviceId` field to their response to the :code:`hello` command. This field
+MUST be a BSON :code:`ObjectId` and SHOULD NOT change while the service is running.
+
+All services that have the behaviour of reaping idle cursors after a specified period of time MAY
+also close the connection associated with the cursor when the cursor is reaped. Conversely, those
+services MAY reap a cursor when the connection associated with the cursor is closed.
+
+All services that have the behaviour of reaping idle transactions after a specified period of time
+MUST also close the connection associated with the transaction when the transaction is reaped.
```
divjotarora, line 312: Here, we say the service MUST close the connection when reaping the transaction, but the previous paragraph says the service MAY close the connection when reaping cursors. Should we standardize on MAY? IIRC, these were both previously MUST but we realized that servers may not need to invalidate the connection as part of their implementation.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,535 @@
...
+Conversely, those services must abort a transaction when the connection associated with the
+transaction is closed.
+
+When a driver is configured to not be in load balanced mode and the service is configured behind
```
divjotarora, line 316: This can be combined with the paragraph about returning `serviceId`. The new paragraph would contain all of the information about how the server should respond to `hello`/legacy `hello` commands.

source/load-balancers/load-balancers.rst
durran: Oh I missed that part. What do you think about adding a new reason enum or reusing "poolClosed". I'm leaning towards the former here, maybe something like "poolClosedForService"?

source/load-balancers/load-balancers.rst
durran: Moved into the previous paragraph.

source/load-balancers/load-balancers.rst
divjotarora: The spec uses "closed" to mean that the pool itself has been closed, which would probably be done via a call like `MongoClient.close()` in drivers. This doesn't fit with the scenario here, though, because the pool is still open. We just can't get a connection because we've reached the max size.

Also, aside from wording, I'd prefer to not add a new reason here. Reaching `maxPoolSize` is already possible today. When this happens, we return a `waitQueueTimeoutError` and publish an event with `reason=timeout`. I see no reason to change that - we're just making the error message more descriptive because it's "easier" to reach `maxPoolSize` in LB mode, but I don't think we need to change the event structure.

source/load-balancers/load-balancers.rst
durran: Changed to MAY in this case, nice catch.

source/load-balancers/load-balancers.rst
durran: Removed.

source/load-balancers/load-balancers.rst
durran: Included this in this paragraph now.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -684,6 +694,24 @@ eagerly so that any operations waiting on `Connections <#connection>`_ can retry
...
+tuple of (generation, connection count) where the connection count refers to the
+total number of connections that exist for a specific ``serviceId``. The pool MUST
+remove the entry for a ``serviceId`` once the connection count reaches 0.
+Once the MongoDB handshake is done, the connection MUST get the
+generation number that applies to its ``serviceId`` from the map and update the
+map to increment the connection count for this ``serviceId``.
+
+See the `Load Balancer Specification <../load-balancers/load-balancers.rst#connection-pooling>`__ for details.
+
+Errors encountered when performing handshake commands after the initial :code:`hello`
```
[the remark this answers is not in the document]
durran: Removed.

source/load-balancers/load-balancers.rst
durran: Removed the reference to the event itself so behaviour remains the same.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -749,6 +754,12 @@ response to the current ServerDescription's topologyVersion:
 See `Replacing the TopologyDescription`_ for an example implementation of
 topologyVersion comparison.
 
+serviceId
+````````
+
+MongoDB 5.0 and later, as well as any mongos-like service include a ``serviceId``
```
jyemin, line 760: Comma between `service` and `include`.  Without the comma, I totally misinterpreted this sentence.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1264,9 +1282,11 @@ implementation::
                 return
 
             if isStateChangeError(error):
-                # Mark the server Unknown
```
jyemin, line 1267: In Java driver I implemented error handling for load balancer in a separate method, so I can't compare it directly to this pseudocode.  But pseudocode doesn't look right.  For example:

- There's no clearing of connection pool on network error when in LoadBalanced mode.
- The clearing of the pool when there is a shutdown error clears the whole pool and doesn't mention anything about serviceId

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+be supported.
+
+
+References
+----------
+
+Data Lake PROXY Detection
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+`Github Link <https://github.com/10gen/mongohouse/blob/master/internal/util/proxy/conn.go#L108>`__
```
jyemin, line 422: Is this link available to users outside mongodb org?  If not, please remove the reference

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+
+Load Balancers PROXY Protocol Configuration
```
jyemin, line 424: I think we can remove this whole section as well.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -281,10 +281,16 @@ has the following properties:
...
     
       /**
-       *  A generation number representing the SDAM generation of the pool
+       *  A generation number representing the SDAM generation of the pool.
        */
       generation: number;
 
+      /**
+       * A map representing the various generation numbers for various services
+       * when in load balancer mode..
```
divjotarora, line 290: nit: Extra period.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+
+Services MAY add a command line option or other configuration parameter, that tells the service
+it is running behind a load balancer. Services MAY also dynamically determine whether they are
+behind a load balancer.
+
+All services which terminate TLS MUST be configured to return a TLS certificate for a hostname
+which matches the hostname the client is connecting to.
+
+All services behind a load balancer that have been started with the aforementioned option MUST
+add a top level :code:`serviceId` field to their response to the :code:`hello` command. This field
```
jyemin, line 301: or legacy hello I would think

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+
+Design Rationales
+-----------------
+
+This design is based on the simplest approach to implementation in both the drivers and all services,
+when taking into consideration that it is NOT a requirement that applications connected to any service
+should not need to make changes and/or restart their applications. If this requirement were to change,
+see alternative designs below.
+
+Connection pinning behaviour is not altered in this specification and it is understood that this can
```
jyemin, line 382: `not altered`?  Isn't it altered?

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+Why are we requiring mongos servers to add a new serviceId field in hello responses rather than reusing the existing topologyVersion.processId?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This option was previously discussed, but we opted to add a new :code:`hello` response field in
+order to not mix intentions.
+
+Why does this specification not address load balancer restarts or maintenance?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+It was decided that the core problem that was needed to be solved was to introduce durable/resumable
```
divjotarora, line 370: IIRC, resumable cursors are intended to make cursor iteration more robust because it would tolerate various transient errors. LB maintenance/restarts pose a whole different problem: because there's no monitoring routine, we have no idea if the LB itself is even running and accepting new connections. I'm not sure if there was ever a solution for this problem, but we did throw around the idea of having a "smarter" layer 7 LB that understood the wire protocol and could respond to monitoring requests.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+cursors to some services. The related ticket is `DRIVERS-1549 <https://jira.mongodb.org/browse/DRIVERS-1549>`__.
+
+
+Design Rationales
+-----------------
+
+This design is based on the simplest approach to implementation in both the drivers and all services,
```
jyemin, line 377: Let's not say it's the simplest approach.  It's too subjective.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+
+Use of a connection string option would allow the driver to continue to use SRV records that
+pointed at a load balancer instead of a replica set without needing to change the URI provided
+to the :code:`MongoClient`. The SRV records could also provide the default :code:`loadBalanced=true`
+in the TXT records.
+
+Why explicitly opt-in to this behaviour instead of letting mongos inform the driver of the load balancer?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This was a case of complexity in implementation and the number of affected services, including
```
divjotarora, line 334: This paragraph is saying that there would be changes to start-up params and `hello` responses for upstream services, but we ended up doing this anyway. Can we remove this paragraph? The second paragraph is really the important bit: it adds a lot of complexity to the driver to start in one mode and then have to transition into LB mode, so we require the user to configure the driver appropriately and just start in LB mode from the beginning.

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+configuration options that are read during start-up. From there they would need to alter their
+hello responses in order to tell the driver they were behind a load balancer.
+
+From a driver's view there is a change in behaviour in that once in load balanced mode, the driver
+can never transition out of that mode, unlike other states in the SDAM specification.
+
+Why does this specification instruct drivers to not check connections back into the connection pool in some circumstances?
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+In the case of a load balancer fronting multiple services, it is possible that a connection to th
```
divjotarora, line 346: Typo: there's a space between "th" and "e".

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,531 @@
...
+
+It was decided that the core problem that was needed to be solved was to introduce durable/resumable
+cursors to some services. The related ticket is `DRIVERS-1549 <https://jira.mongodb.org/browse/DRIVERS-1549>`__.
+
+
+Design Rationales
+-----------------
+
+This design is based on the simplest approach to implementation in both the drivers and all services,
+when taking into consideration that it is NOT a requirement that applications connected to any service
```
jyemin, line 378: I'm not sure I understand this design rationale.  Can you say more clearly what it is you mean?

The rationale that I recall is that we wanted people to opt in to load balancing mode because it may require changes in other settings, in particular max pool size.  For the same reason, we didn't just want to start pinning connections/sessions in all modes because of connection pool pressure.

jyemin (changes requested):

source/load-balancers/load-balancers.rst
divjotarora, line 382: I'm not sure I understand this paragraph. It says that connection pinning behavior is not altered in this spec, but I'm not aware of any previous features that required drivers to pin connections in the first place. Also, the wording about session and cursor leaks reads as if the driver is intentionally leaking resources, which is not true. Overall, I'm not sure this paragraph adds much value so maybe we can simply remove it?

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,553 @@
...
+Implementation Notes
+--------------------
+
+All connections MUST track a generation number. For non-load balanced deployments, the connection's
+generation number MUST equal the generation number of the pool at the time that the connection attempt
+was started. Any errors encountered when dialing a connection or performing any MongoDB commands
+for the connection handshake in this mode MUST be processed per the "Error Handling" section of
+the SDAM spec.
+
+For load-balanced deployments, pools MUST maintain a map from :code:`serviceId` to a tuple of
```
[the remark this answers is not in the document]
divjotarora: FYI this comment is outdated. I think this is referring to the "Implementation Notes" section, which has already been removed from the LB spec (see [this comment thread](https://github.com/mongodb/specifications/pull/939#discussion_r621327114)).

source/load-balancers/load-balancers.rst
```diff
@@ -0,0 +1,553 @@
...
+--------------
+
+Initial Handshake Errors
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+If there is a network error or timeout on the connection before the hello or legacy hello completes,
+the driver MUST NOT make any changes to the :code:`TopologyDescription` or mark the server as Unknown.
+The driver MUST also close the associated connection. If the network error occurred on the handshake's
+:code:`hello` command, the connection pool MUST NOT be cleared. If the network error occurred after
+the handshake's hello command, the connection POOL must clear close all connections with the matching
```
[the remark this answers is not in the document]
durran: This code is outdated and that text is no longer present.

source/load-balancers/load-balancers.rst
durran: I've removed this paragraph now.

source/load-balancers/load-balancers.rst
durran: Fixed.

source/load-balancers/load-balancers.rst
durran: Removed.

source/load-balancers/load-balancers.rst
durran: I've reworded this paragraph now and split into 2 questions.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
durran: Fixed.

source/load-balancers/load-balancers.rst
durran: I've reworded this sections now to be more clear.

source/load-balancers/load-balancers.rst
durran: Section reworded.

source/load-balancers/load-balancers.rst
durran: I've actually now removed this paragraph per @divjotarora 's suggestion.

source/load-balancers/load-balancers.rst
durran: Added legacy hello here.

source/load-balancers/load-balancers.rst
durran: Removed.

source/load-balancers/load-balancers.rst
durran: Reference removed.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
durran: I've added the conditions for clearing the pool for service id now in both blocks.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
durran: Fixed.

source/load-balancers/load-balancers.rst
```diff
@@ -367,22 +361,26 @@ order to not mix intentions.
 Why does this specification not address load balancer restarts or maintenance?
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-It was decided that the core problem that was needed to be solved was to introduce durable/resumable
-cursors to some services. The related ticket is `DRIVERS-1549 <https://jira.mongodb.org/browse/DRIVERS-1549>`__.
+The Layer 4 load balancers that would be in use for this feature lack the
+ability that a layer 7 load balancer could potentially have to be able to
+understand the MongoDB wire protocol and respond to monitoring requests.
+
+Why was connection pinning for cursors added to this spec?
```
divjotarora, line 368: I'm not following how resumable cursors are related to connection pinning behavior. The former would make cursors more robust while the latter is needed to ensure cursor-related commands are all sent to the same backing server. Also, durable/resumable cursors would be useful in all contexts, not just LB mode. Can we remove this paragraph for simplicity?

durran: I've removed this section now. I wanted to keep it around for context on discussions we had but I think since it's still in the google doc it was ok to remove here.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1264,9 +1282,13 @@ implementation::
...
-                unknown = new ServerDescription(type=Unknown, error=error, topologyVersion=topologyVersion)
-                onServerDescriptionChanged(unknown, connection pool for server)
+                # Don't mark server unknown in load balanced mode.
+                if type != LoadBalanced
+                  # Mark the server Unknown
+                  unknown = new ServerDescription(type=Unknown, error=error, topologyVersion=topologyVersion)
+                  onServerDescriptionChanged(unknown, connection pool for server)
+                else
```
divjotarora, line 1290: This `else` statement is not correct - we only want to clear the pool for the specific `serviceId` on shutdown errors or pre-4.2 state change errors. That's already covered by the next `if` statement, but it needs to be amended to mention `serviceId`. I'd recommend removing this `else` and modifying the `if` to:

```
if isShutdown(code) or (error was from <4.2):
  if type == LoadBalanced:
    clear connection pool for serviceId
  else:
    clear connection pool for server
```

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1279,12 +1301,15 @@ implementation::
...
-                onServerDescriptionChanged(unknown, connection pool for server)
-                clear connection pool for server
-                # Cancel inprogress check
-                cancel monitor check
+                if type != LoadBalanced
+                  # Mark the server Unknown
+                  unknown = new ServerDescription(type=Unknown, error=error)
+                  onServerDescriptionChanged(unknown, connection pool for server)
+                  clear connection pool for server
+                  # Cancel inprogress check
```
divjotarora, line 1309: We want to do this for LB mode as well. Can we move it outside the if/else so it's done unconditionally?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1279,12 +1301,15 @@ implementation::
...
+                  cancel monitor check
+                else
+                  clear connection pool for serviceId
```
divjotarora, line 1312: If `not error.completedHandshake`, this should only happen if `serviceId` is set as that indicates that the initial `hello` succeeded.

source/load-balancers/load-balancers.rst
jyemin: I still don't get it.  Let's discuss it in person to figure out what we really mean here.

jyemin: Thought about this a bit while out on evening stroll.  I think these are the points we want to make:

Q.  Why do we pin connections to open transactions and cursors

A Because that's the only way to ensure that we can send subsequent messages with same txn/cursor id to the same mongos server that initiated the txn/cursor.

Q. Why do applications have to opt-in to this mode?

A. Load balancing and pinning creates new stresses on connection pools.  We go from a total number of connections to the cluster of num-mongos-servers * max-pool-size to simply max-pool-size.  And in addition, connections get pinned to open cursors and transactions, further straining resource availability.  So in addition to opting in applications my need to bump up the max pool size.

divjotarora: IMO the first question (why do we pin) is already answered in the spec's [Connection Pinning](https://github.com/mongodb/specifications/blob/79c138069b47612b610cbb9cf7a972408d0069ea/source/load-balancers/load-balancers.rst#connection-pinning) section. For the second question, I agree that this should be made clearer in the "Design Rationale" section. Suggested wording for this section:

> Other versions of this design proposed a scheme in which the application does not have to opt-in to load balanced mode. Instead, the server would send a special field in ``hello`` and legacy ``hello`` command responses to indicate that it was running behind a load balancer and the driver would change its behavior accordingly. We opted to take an approach that required code changes instead because load balancing changes driver behavior in ways that could cause unexpected application errors, so it made sense to have applications consciously opt-in to this mode.  For example, connection pinning creates new stresses on connection pools because we go from a total of ``numMongosServers * maxPoolSize`` connections to simply ``maxPoolSize``. Furthermore, connections get pinned to open cursors and transactions, further straining resource availability. Due to this change, applications may also need to increase the configured ``maxPoolSize`` when opting into this mode.

jyemin: True about the connection pinning rationale.  It's a little strange to have a rationale in the main portion of the spec.  Maybe it should move here?  Either way is ok with me though.

divjotarora: I suggested adding it in the spec itself because it felt weird to say "pin connections in these cases" without at least giving a basic explanation of the requirements. I'd prefer to keep it there, but we can duplicate it here as well if you think that would make it more readable.

jyemin: I don't disagree with that.

durran: I've updated the opt-in question to the suggested wording.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
durran: This has now been updated.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
durran: Added check for serverId existence.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
durran: Moved out to cover all cases.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1516,8 +1547,8 @@ Authentication errors
 ~~~~~~~~~~~~~~~~~~~~~
 
 If the authentication handshake fails for a connection, drivers MUST mark the
-server Unknown and clear the server's connection pool. (See
-`Why mark a server Unknown after an auth error?`_)
+server Unknown and clear the server's connection pool if the TopologyType is
```
divjotarora, line 1550: Nit: This makes it sound like we also don't clear the connection pool in LB mode. Can we add a second sentence right after this:

> If the TopologyType is LoadBalanced, drivers MUST NOT mark the server as Unknown for authentication failures but MUST clear the connection pool for the serviceId of the connection.

divjotarora (approved):
One minor nit, but LGTM! Thanks for addressing all of the comments!

jyemin (approved):
LGTM

PR #940, opened 2021-04-06
Change the required MongoDB version to 4.9.0+ in the "Connection Pool Management" test

stIncMale:
DRIVERS-1650

### DIFF
--- a/source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst
+++ b/source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst
@@ -8,7 +8,7 @@ Server Discovery And Monitoring -- Test Plan
 :Advisors: David Golden, Craig Wilson
 :Status: Draft
 :Type: Standards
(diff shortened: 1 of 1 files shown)


patrickfreed (commented):
appName support actually was added to failCommand in 4.2.9, however it has a bug that causes it to not work on initial handshake `isMaster` / `hello` commands until 4.9, which is why some other tests require 4.9. This test in particular only triggers the failpoint on post-handshake `isMaster` commands though, which means it can run on 4.2.9+. I updated the [failCommand wiki page](https://github.com/mongodb/mongo/wiki/The-%22failCommand%22-fail-point) to reflect this information.

stIncMale:
@patrickfreed Thank you for the details! Is my understanding correct that your answer implies that we need to close the PR?

patrickfreed:
I personally think the test is fine as-is and we can close the PR, though if you think it will cause similar confusion in the future, maybe its worth updating.

stIncMale:
Then I am closing it.

PR #941, opened 2021-04-06
DRIVERS-1638 Add more tests for sharded transaction unpinning

divjotarora:
Most of the changes are due to restructuring the `transactions/tests` directory into `legacy` and `unified` sub-directories. The only new changes are in `transactions/tests/README.rst` and `transactions/tests/unified/mongos-unpin.yml`.


divjotarora:
I added a requirement about unpinning the server on once the transaction has been aborted because per the transactions FSM, doing any transactional operations after aborting will result in a client-side error so the pinned server is no longer needed.

source/transactions/tests/unified/mongos-unpin.yml
```diff
@@ -0,0 +1,118 @@
...
+          errorCode: *lockTimeoutErrorCode
+          errorLabelsContain: [ TransientTransactionError ]
+          errorLabelsOmit: [ UnknownTransactionCommitResult ]
+      - &assertNoPinnedServer
+        name: assertSessionUnpinned
+        object: testRunner
+        arguments:
+          session: *session0
+
+  - description: unpin on successful abort
```
ShaneHarvey, line 70: This is a new requirement in the spec that requires driver changes even for drivers that otherwise follow the spec before. Can you call this change out in the spec changelog and bump the version number?

source/transactions/transactions.rst
```diff
@@ -824,8 +824,10 @@ error label. In cases where the UnknownTransactionCommitResult causes an
 automatic retry attempt, drivers MUST unpin the ClientSession before performing
 server selection for the retry.
 
-Starting a new transaction on a pinned ClientSession MUST unpin the
-session. Additionally, any non-transaction operation using a pinned
+Aborting a transaction or starting a new transaction on a pinned
+ClientSession MUST unpin the session. Committing a transaction on a pinned
+ClientSession MUST NOT unpin the session as transactions can be committed
```
ShaneHarvey, line 829: "as transactions can be committed multiple times" is a bit ambiguous and could be interpreted as some kind of non-idempotent behavior where a transaction gets applied multiple times on the server. Can we change it to "as commitTransaction may be called multiple times"?

ShaneHarvey (commented):
Tests pass in python.

source/transactions/tests/unified/mongos-unpin.yml
divjotarora: Done.

source/transactions/transactions.rst
divjotarora: Done.

divjotarora:
I also updated the description for DRIVERS-1638 to describe the new tests and the prose change for `abortTransaction`.

source/transactions/transactions.rst
```diff
@@ -1398,6 +1400,7 @@ durable, which achieves the primary objective of avoiding duplicate commits.
 **Changelog**
 -------------
 
+:2020-04-07: Specify that servers should be unpinned once a transcation is aborted.
```
ShaneHarvey, line 1403: servers->sessions
transcation->transaction

divjotarora: Done.

ShaneHarvey (approved):
LGTM! Please note the test directory structure refactor and the new abortTransaction requirement  in commit message+jira ticket.

durran (approved):
LGTM

PR #942, opened 2021-04-07
DRIVERS-1293: Remove deprecated terms from unowned specs

JamesKovacs:
These updates remove oppressive language from specs unowned by team members. The one exception is the handshake spec, which will receive its own PR due to the more substantial changes to discuss hello, helloOk, Versioned API, and legacy hello.

One point to note: The OP_COMPRESSED spec still has examples with `isMaster` present. I grabbed the latest server build (4.9.0-alpha7), but the `mongo` shell in that build still uses `isMaster` for the handshake even when `--apiStrict` is provided. The resulting server logs still show compression being negotiated via `isMaster`. We also don't have any driver implementations yet using the new handshake protocol (e.g. `hello`/`helloOk`). Thus I couldn't get clean logs not containing `isMaster`. Options to consider:

1. Wait for Server to implement the `hello`/`helloOk` handshake in the shell.
2. Update the examples once we prototype the `hello`/`helloOk` handshake in one of our drivers.
3. Remove the `mongod` log lines from the spec.


alcaeus:
> Thus I couldn't get clean logs not containing `isMaster`. Options to consider:

I'd add option 4: update the logs to use `hello`, add the line info "log output may differ depending on server version", and be done with it. Since those are examples on what this looks like, I feel we don't necessarily need to be precise to the exact command being used.

source/benchmarking/benchmarking.rst
```diff
@@ -489,10 +489,10 @@ the database and reading a response.
 Dataset:  n/a
 
 Dataset size: While there is no external dataset, for score calculation
-purposes use 160,000 bytes (10,000 x the size of a BSON {ismaster:true}
+purposes use 130,000 bytes (10,000 x the size of a BSON {hello:true}
```
kevinAlbs, line 492: Oh interesting, no change needed, but I wonder if this will impact these benchmarks.

kevinAlbs (commented):
Looking good.

- driver-bulk-update.rst has a reference to `isMaster()` and no spec owner.
- GridFS refers to "black box".

> I'd add option 4: update the logs to use `hello`, add the line info "log output may differ depending on server version", and be done with it. Since those are examples on what this looks like, I feel we don't necessarily need to be precise to the exact command being used.

Should we regenerate those logs? Those logs show the old, non-JSON format. Using the C driver's example client, and replacing `isMaster` with `hello` generated this: https://gist.github.com/kevinAlbs/fa7ec139016d606da24be9cddd77e049

source/compression/OP_COMPRESSED.rst
```diff
@@ -194,7 +194,7 @@ Compressor IDs
 Each compressor is assigned a predefined compressor ID.
 
 +-----------------+----------------+-------------------------------------------------------+
-| compressorId    | isMaster Value |  Description                                          |
+| compressorId    | hello Value    |  Description                                          |
```
jmikola, line 197: At risk of a larger diff, would "handshake value" be more consistent with what you wrote above?

This would also allow an opportunity to remove the extra space before "noop" in the following row.

source/find_getmore_killcursors_commands.rst
```diff
@@ -468,7 +468,7 @@ getMore
 -------
 
 The **getMore** command replaces the **OP_GET_MORE** wire protocol message.
-The query flags passed to OP_QUERY for a getMore command MUST be slave_ok=True
+The query flags passed to OP_QUERY for a getMore command MUST be secondary_ok=True
```
jmikola, line 471: Since the same document uses camelCase above, would it make sense to use "secondaryOk" here and in the instances below?

source/compression/OP_COMPRESSED.rst
jmikola, line 197: Should this be "handshake value" for consistency with the above, or was that intentionally not done to avoid a larger diff?

If you do go the route of a larger diff, you could also remove the leading space before "noop" in the row below.

source/benchmarking/benchmarking.rst
JamesKovacs: BSON `{ismaster:true}` will be serialized on the wire as:
payload size as int32
single boolean element (\x08) with value true (\x01): \x08 ismaster (8 bytes + \x00) \x01
terminal null: \x00
Total Size: 16 bytes

BSON `{hello:true}` will be serialized on the wire as:
payload size as int32
single boolean element (\x08) with value true (\x01): \x08 hello (5 bytes + \x00) \x01
terminal null: \x00
Total Size: 13 bytes

So yes, the data size will therefore be changed for this test case, which could affect the results. We can either decide to accept that `{ismaster:true}` test results can't be compared directly to `{hello:true}` results or we can artificially pad out `{hello:true}` by adding another 3 bytes such as `{a:null}` (`\x0a \x61 \x00`). My preference is to acknowledge that the test results aren't directly comparable rather than artificially padding the `hello` test to match the previous size. I suppose we could also change the number of iterations, but that would likely affect the results more than the difference in BSON document size.

source/find_getmore_killcursors_commands.rst
JamesKovacs: Since the original spec text used `slaveOk` and later `slave_ok=True`, I thought there was some significance to the use of snake_case. But I guess the reason was the original spec author probably worked on the Python driver. Changed to `secondaryOk=true` to maintain consistency with previous usages in the spec.

source/compression/OP_COMPRESSED.rst
JamesKovacs: Done. And the phrase `Handshake Value` is just long enough to force me to reformat the entire table. 😛  So the `noop` fix got "added" for free.

JamesKovacs:
> > Thus I couldn't get clean logs not containing `isMaster`. Options to consider:
> 
> I'd add option 4: update the logs to use `hello`, add the line info "log output may differ depending on server version", and be done with it. Since those are examples on what this looks like, I feel we don't necessarily need to be precise to the exact command being used.

Option 4 works for me.

JamesKovacs:
Feedback incorporated. Re-requesting reviews.

source/benchmarking/benchmarking.rst
kevinAlbs: Hah, I hadn't thought about that artificial padding. I agree with your preference of keeping it as is.

source/compression/OP_COMPRESSED.rst
```diff
@@ -284,64 +285,84 @@ Connection strings, and results
...
-   2017-04-17T15:04:37.756-0700 D NETWORK  [conn6] Compressing message with snappy
-   2017-04-17T15:04:37.757-0700 D NETWORK  [conn6] Decompressing message with snappy
-   2017-04-17T15:04:37.757-0700 D COMMAND  [conn6] run command test.$cmd { ping: 1 }
-   2017-04-17T15:04:37.757-0700 I COMMAND  [conn6] command test.$cmd command: ping { ping: 1 } numYields:0 reslen:37 locks:{} protocol:op_query 0ms
-   2017-04-17T15:04:37.757-0700 D NETWORK  [conn6] Compressing message with snappy
-   2017-04-17T15:04:37.757-0700 D NETWORK  [conn6] Socket recv() conn closed? 127.0.0.1:34294
-   2017-04-17T15:04:37.757-0700 D NETWORK  [conn6] SocketException: remote: 127.0.0.1:34294 error: 9001 socket exception [CLOSED] server [127.0.0.1:34294]
-   2017-04-17T15:04:37.757-0700 I -        [conn6] end connection 127.0.0.1:34294 (1 connection now open)
+  mongod should have logged the following (the exact log ouptut may differ depending on server version)::
```
alcaeus, line 288: Typo: "output"
```suggestion
  mongod should have logged the following (the exact log output may differ depending on server version)::
```

alcaeus (approved):
Found one typo, but rest LGTM!

source/compression/OP_COMPRESSED.rst
```diff
@@ -284,64 +285,84 @@ Connection strings, and results
...
-
-   2017-02-27T04:00:00.000-0700 D COMMAND  [conn654] run command admin.$cmd { isMaster: 1, client: { ... }, compression: [] }
+  mongod should not have logged anything as drivers should reject invalid compressor values.
 
-  e.g., empty compression: [] array. No operations should have been compressed.
   The results of the program should have been::
 
-   WARNING: Unsupported compressor: 'snoopy'
-   { "ok" : { "$numberDouble" : "1.0" } }
+   Failed global initialization: BadValue Invalid network message compressor specified in configuration: snoopy
```
kevinAlbs, line 340: Is this the behavior for .NET? I interpreted the spec to proceed with a [warning on an invalid compressor](https://github.com/mongodb/specifications/blob/master/source/compression/OP_COMPRESSED.rst#compressors)

> Comma separated list of compressors the client should present to the server. Unknown compressors MUST yield a warning, as per the Connection String specification, and MUST NOT be included in the handshake.

kevinAlbs (commented):
I have a question about OP_COMPRESSED on unrecognized compressors.

I also think driver-bulk-update.rst may need to be updated to replace `isMaster()` with `hello()`.

source/compression/OP_COMPRESSED.rst
```diff
@@ -284,64 +285,84 @@ Connection strings, and results
...
-   { "ok" : 1.0 }
+   { "ok" : 1 }
 
 
 * mongodb://localhost:27017/?compressors=snoopy
 
-  mongod should have logged the following::
-
-   2017-02-27T04:00:00.000-0700 D COMMAND  [conn654] run command admin.$cmd { isMaster: 1, client: { ... }, compression: [] }
+  mongod should not have logged anything as drivers should reject invalid compressor values.
```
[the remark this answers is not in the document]
JamesKovacs: Oh, I need to correct this. I was testing with the `mongo` shell and (foolishly) assumed that the behaviour would be the same. I'll double check with a driver or two and fix this section. Thanks for catching this.

source/compression/OP_COMPRESSED.rst
JamesKovacs: As noted above, this was based on the `mongo` shell. I'll re-run it with an actual driver and fix this.

source/compression/OP_COMPRESSED.rst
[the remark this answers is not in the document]
JamesKovacs: Fixed.

source/compression/OP_COMPRESSED.rst
JamesKovacs: Fixed.

kevinAlbs (approved):
LGTM!

alcaeus:
Merging to unblock DRIVERS-1633.

PR #944, opened 2021-04-07
DRIVERS-1603 Serverless drivers testing spec

patrickfreed:
[DRIVERS-1603](https://jira.mongodb.org/browse/DRIVERS-1603)

This PR adds a testing-only spec which requires that drivers run a certain set of tests against a live Atlas serverless instance. As part of this, the unified test format was updated to include a Serverless-specific runOnRequirement, and some tests were updated to use it as necessary.

Example patch: https://evergreen.mongodb.com/version/606e3ba332f4171c4c8b406e

### DIFF
--- a/source/crud/tests/README.rst
+++ b/source/crud/tests/README.rst
@@ -61,6 +61,15 @@ Each YAML file has the following keys:
     and "sharded". If this field is omitted, the default is all topologies (i.e.
     ``["single", "replicaset", "sharded"]``).
 
+  - ``serverlessMode``: Optional string, only applicable to "sharded"
+    topologies. Whether or not the test should be run on serverless instances
+    imitating sharded clusters. Valid modes are "requireServerless",
+    "forbidServerless", and "allowServerless". If "requireServerless", the test
+    MUST only be run on serverless instances. If "forbidServerless", the test
+    MUST only be run on actual sharded deployments. If omitted or
+    "allowServerless", the test can be run on either serverless instances or on
+    real sharded deployments.
+
 - ``collection_name`` (optional): The collection to use for testing.
 
 - ``database_name`` (optional): The database to use for testing.
--- a/source/crud/tests/README.rst
+++ b/source/crud/tests/README.rst
@@ -64,9 +64,8 @@ Each YAML file has the following keys:
   - ``serverless``: (optional) Whether or not the test should be run on
     serverless instances imitating sharded clusters. Valid values are "require",
     "forbid", and "allow". If "require", the test MUST only be run on serverless
-    instances. If "forbid", the test MUST only be run on actual sharded
-    deployments. If omitted or "allow", the test can be run on either serverless
-    instances or on real sharded deployments.
+    instances. If "forbid", the test MUST NOT be run on Serverless instances. If
+    omitted or "allow", this option has no effect.
 
 - ``collection_name`` (optional): The collection to use for testing.
 
--- a/source/crud/tests/README.rst
+++ b/source/crud/tests/README.rst
@@ -61,14 +61,13 @@ Each YAML file has the following keys:
     and "sharded". If this field is omitted, the default is all topologies (i.e.
     ``["single", "replicaset", "sharded"]``).
 
-  - ``serverlessMode``: Optional string, only applicable to "sharded"
-    topologies. Whether or not the test should be run on serverless instances
-    imitating sharded clusters. Valid modes are "requireServerless",
-    "forbidServerless", and "allowServerless". If "requireServerless", the test
-    MUST only be run on serverless instances. If "forbidServerless", the test
-    MUST only be run on actual sharded deployments. If omitted or
-    "allowServerless", the test can be run on either serverless instances or on
-    real sharded deployments.
+  - ``serverlessMode``: (optional) Whether or not the test should be run on
+    serverless instances imitating sharded clusters. Valid modes are
+    "requireServerless", "forbidServerless", and "allowServerless". If
+    "requireServerless", the test MUST only be run on serverless instances. If
+    "forbidServerless", the test MUST only be run on actual sharded
+    deployments. If omitted or "allowServerless", the test can be run on either
+    serverless instances or on real sharded deployments.
 
 - ``collection_name`` (optional): The collection to use for testing.
 
--- a/source/crud/tests/README.rst
+++ b/source/crud/tests/README.rst
@@ -61,13 +61,12 @@ Each YAML file has the following keys:
     and "sharded". If this field is omitted, the default is all topologies (i.e.
     ``["single", "replicaset", "sharded"]``).
 
-  - ``serverlessMode``: (optional) Whether or not the test should be run on
-    serverless instances imitating sharded clusters. Valid modes are
-    "requireServerless", "forbidServerless", and "allowServerless". If
-    "requireServerless", the test MUST only be run on serverless instances. If
-    "forbidServerless", the test MUST only be run on actual sharded
-    deployments. If omitted or "allowServerless", the test can be run on either
-    serverless instances or on real sharded deployments.
+  - ``serverless``: (optional) Whether or not the test should be run on
+    Serverless instances imitating sharded clusters. Valid values are "require",
+    "forbid", and "allow". If "require", the test MUST only be run on Serverless
+    instances. If "forbid", the test MUST only be run on actual sharded
+    deployments. If omitted or "allow", the test can be run on either Serverless
+    instances or on real sharded deployments.
 
 - ``collection_name`` (optional): The collection to use for testing.
 
--- a/source/crud/tests/README.rst
+++ b/source/crud/tests/README.rst
@@ -62,10 +62,10 @@ Each YAML file has the following keys:
     ``["single", "replicaset", "sharded"]``).
 
   - ``serverless``: (optional) Whether or not the test should be run on
-    Serverless instances imitating sharded clusters. Valid values are "require",
-    "forbid", and "allow". If "require", the test MUST only be run on Serverless
+    serverless instances imitating sharded clusters. Valid values are "require",
+    "forbid", and "allow". If "require", the test MUST only be run on serverless
     instances. If "forbid", the test MUST only be run on actual sharded
-    deployments. If omitted or "allow", the test can be run on either Serverless
+    deployments. If omitted or "allow", the test can be run on either serverless
     instances or on real sharded deployments.
 
 - ``collection_name`` (optional): The collection to use for testing.
--- a/source/crud/tests/unified/aggregate.yml
+++ b/source/crud/tests/unified/aggregate.yml
@@ -10,7 +10,7 @@ createEntities:
   - database:
       id: &database0 database0
       client: *client0
-      databaseName: &database0Name edc-tests
+      databaseName: &database0Name aggregate-tests
   - collection:
       id: &collection0 collection0
       database: *database0
--- a/source/crud/tests/unified/aggregate.yml
+++ b/source/crud/tests/unified/aggregate.yml
@@ -1,0 +1,79 @@
+description: "aggregate"
+
+schemaVersion: "1.0"
+
+createEntities:
+  - client:
+      id: &client0 client0
+      useMultipleMongoses: true # ensure cursors pin to a single server
+      observeEvents: [ commandStartedEvent ]
+  - database:
+      id: &database0 database0
+      client: *client0
+      databaseName: &database0Name edc-tests
+  - collection:
+      id: &collection0 collection0
+      database: *database0
+      collectionName: &collection0Name coll0
+
+initialData:
+  - collectionName: *collection0Name
+    databaseName: *database0Name
+    documents:
+      - { _id: 1, x: 11 }
+      - { _id: 2, x: 22 }
+      - { _id: 3, x: 33 }
+      - { _id: 4, x: 44 }
+      - { _id: 5, x: 55 }
+      - { _id: 6, x: 66 }
+      - { _id: 7, x: 77 }
+      - { _id: 8, x: 88 }
+
+tests:
+  - description: "aggregate with multiple batches works"
+    operations:
+      - name: aggregate
+        arguments:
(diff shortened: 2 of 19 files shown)


source/crud/tests/README.rst
```diff
@@ -61,6 +61,14 @@ Each YAML file has the following keys:
     and "sharded". If this field is omitted, the default is all topologies (i.e.
     ``["single", "replicaset", "sharded"]``).
 
+  - ``serverlessMode``: (optional) Whether or not the test should be run on
```
patrickfreed, line 64: This was modeled after the "sslMode" option from the server, let me know if it the terms sound okay. It's the same exact field in the unified format.

source/crud/tests/unified/aggregate.yml
```diff
@@ -0,0 +1,79 @@
+description: "aggregate"
```
patrickfreed, line 1: This and `find.yml` ensure that cursors work even with multiple batches (test required by scope).

source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -34,6 +34,7 @@ tests:
   - description: "estimatedDocumentCount uses $collStats on 4.9.0 or greater"
     runOnRequirements:
       - minServerVersion: "4.9.0"
+        serverlessMode: forbidServerless
```
patrickfreed, line 37: Ironically, serverless does not support counting via `$collStats`, so these need to be skipped until that has been fixed ([CLOUDP-86318](https://jira.mongodb.org/browse/CLOUDP-86318))

source/crud/tests/v2/db-aggregate.yml
```diff
@@ -1,6 +1,7 @@
 runOn:
     -
         minServerVersion: "3.6.0"
+        serverlessMode: "forbidServerless"
```
patrickfreed, line 4: `$listLocalSessions` is not supported in serverless.

source/serverless-testing/tests/README.rst
```diff
@@ -0,0 +1,115 @@
...
+scripts for creating and deleting Serverless instances, and the ``config.yml``
+contains an example evergreen configuration that uses them to run the tests.
+
+.. _serverless directory in the drivers-evergreen-tools repository: https://github.com/mongodb-labs/drivers-evergreen-tools/tree/master/.evergreen/serverless
+
+All tests MUST be run with wire protocol compression and authentication
+enabled. Additionally, Serverless requires the usage of the versioned API, so
+the tests MUST be run with a server API version specified.
+
+Required Variables
```
patrickfreed, line 31: I'm wondering what would be the best way to store the values of these for drivers to refer to. The atlas connectivity ones are in a google doc, so maybe we should go that route too.

source/serverless-testing/tests/README.rst
```diff
@@ -0,0 +1,115 @@
...
+Note that the formats for the JSON/YAML tests of these specifications were
+updated to include a new ``runOnRequirement`` specifically for serverless
+testing, so make sure to update the affected test runners to support this
+requirement and then sync the tests.
+
+The Serverless proxy presents itself as a mongos, so any test meant to run
+against sharded clusters will be executed by the runners, with the exception of
+the tests affected by the previously mentioned ``runOnRequirement``.
+
+Prose Tests
```
patrickfreed, line 78: killCursors tests required by the scope. These are heavily modeled after similar tests for Atlas Data Lake.

source/serverless-testing/tests/README.rst
```diff
@@ -0,0 +1,115 @@
...
+   the collection name and the database may be inferred from either the ``$db``
+   field or accessed via the CommandStartedEvent directly. Finally, assert that
+   the ``killCursors`` CommandSucceededEvent indicates that the expected cursor
+   was killed in the ``cursorsKilled`` field.
+
+#. Repeat test 1 on a cursor that is still left open after both the initial
+   ``find`` and a single ``getMore`` (e.g. specify ``batchSize=1`` for a query
+   that would match 3+ documents).
+
+Other Tests
```
patrickfreed, line 109: Required by the "existing tests using cursors" portion of the spec.

source/crud/tests/README.rst
alcaeus, line 64: May I humbly suggest `serverless: "allow|require|forbid"`? I feel it's less verbose and less confusing (`serverlessMode` makes it sound like it's about the configuration of the serverless proxy, not whether we're testing against it)

source/unified-test-format/unified-test-format.rst
```diff
@@ -3,13 +3,13 @@ Unified Test Format
 ===================
 
 :Spec Title: Unified Test Format
-:Spec Version: 1.2.3
+:Spec Version: 1.3.0
```
alcaeus, line 6: Please remember to add a changelog entry for this change.

source/crud/tests/README.rst
patrickfreed: Sounds good, updated.

source/unified-test-format/unified-test-format.rst
patrickfreed: Done

source/serverless-testing/tests/README.rst
jmikola: @agolin95: this sounds like another thing that could ultimately benefit from some structure/process

source/unified-test-format/unified-test-format.rst
```diff
@@ -369,6 +369,13 @@ The structure of this object is as follows:
   "sharded" topology, test runners MUST accept any type of sharded cluster (i.e.
   "sharded" implies "sharded-replicaset", but not vice versa).
 
+- ``serverless``: Optional string. Whether or not the test should be run on
+  Serverless instances imitating sharded clusters. Valid values are "require",
```
jmikola, line 373: Unless "Serverless" is going to be a defined term or proper noun for branding, I think lowercase is fine. Note that it is lowercased in https://www.mongodb.com/serverless (unless that has yet to be updated for what this is referring to).

source/unified-test-format/unified-test-format.rst
```diff
@@ -369,6 +369,13 @@ The structure of this object is as follows:
...
+  "forbid", and "allow". If "require", the test MUST only be run on Serverless
+  instances. If "forbid", the test MUST only be run on actual sharded
+  deployments. If omitted or "allow", the test can be run on either Serverless
```
jmikola, line 376: > If "forbid",  the test MUST NOT be run on Serverless instances. If omitted or "allow", this option has no effect. 
 
This will future-proof the text in case something else comes along that masquerades as a mongos. Also, I think "no effect" is easier to understand for anyone implementing this (and also avoids enumerating the current alternatives).

source/unified-test-format/unified-test-format.rst
jmikola, line 6: FYI: this and the corresponding schema version are going to conflict https://github.com/mongodb/specifications/pull/935. Please reach out to @divjotarora and decide if you should use 1.4.0 here or vice versa.

I'll note that the schema versioning doesn't force drivers to implement the projects in a particular order. IIRC, I was working on Atlas testing (1.2 with `loop`, etc.) in PHPLIB before we had implemented versioned API (1.1).

source/serverless-testing/tests/README.rst
agolin95: Agreed @jmikola thank you for the tag! Taking a reminder to investigate what kind of secret sharing platform options we have available to us.

source/crud/tests/unified/aggregate.yml
```diff
@@ -0,0 +1,79 @@
...
+
+createEntities:
+  - client:
+      id: &client0 client0
+      useMultipleMongoses: true # ensure cursors pin to a single server
+      observeEvents: [ commandStartedEvent ]
+  - database:
+      id: &database0 database0
+      client: *client0
+      databaseName: &database0Name edc-tests
```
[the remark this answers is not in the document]
patrickfreed: Updated. (didn't use the "commit suggestion" to group the JSON changes in the same commit).

source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -34,6 +34,7 @@ tests:
   - description: "estimatedDocumentCount uses $collStats on 4.9.0 or greater"
     runOnRequirements:
       - minServerVersion: "4.9.0"
+        serverless: forbid
```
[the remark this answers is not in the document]
patrickfreed: As discussed a little bit in slack, the proxy doesn't yet support count via `$collStats` yet (https://jira.mongodb.org/browse/CLOUDP-86318)

source/crud/tests/unified/find.yml
```diff
@@ -0,0 +1,78 @@
...
+
+createEntities:
+  - client:
+      id: &client0 client0
+      useMultipleMongoses: true # ensure cursors pin to a single server
+      observeEvents: [ commandStartedEvent ]
+  - database:
+      id: &database0 database0
+      client: *client0
+      databaseName: &database0Name edc-tests
```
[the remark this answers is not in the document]
patrickfreed: Updated

source/crud/tests/v2/db-aggregate.yml
```diff
@@ -1,6 +1,7 @@
 runOn:
     -
         minServerVersion: "3.6.0"
+        serverless: "forbid"
```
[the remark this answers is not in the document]
patrickfreed: done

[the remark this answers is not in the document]
patrickfreed: I think the only options for db-level aggregate are `$listLocalSessions` and `$currentOp`, both of which are unsupported by the proxy.

And yeah, I believe whatever restrictions apply to the serverless proxy probably also apply to the free-tier proxy, since I think the former is a descendant of / shares a lot of code with the latter, not sure about how tightly coupled they are though.

source/unified-test-format/tests/valid-pass/poc-crud.yml
```diff
@@ -170,6 +170,7 @@ tests:
   - description: "Aggregate with $listLocalSessions"
     runOnRequirements:
       - minServerVersion: "3.6.0"
+        serverless: forbid
```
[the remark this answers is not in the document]
patrickfreed: done

source/unified-test-format/unified-test-format.rst
patrickfreed: Good point. I think whichever one is ready to merge first ought to take the 1.3 and the other can do 1.4. I'll check with him before merging anything to ensure we're on the same page.

source/unified-test-format/unified-test-format.rst
patrickfreed: Updated.

source/unified-test-format/unified-test-format.rst
patrickfreed: updated

source/versioned-api/tests/crud-api-version-1-strict.yml
```diff
@@ -62,6 +62,8 @@ tests:
                 <<: *expectedApiVersion
 
   - description: "aggregate on database appends declared API version"
+    runOnRequirements:
```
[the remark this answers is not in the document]
patrickfreed: Yep, I don't think we can test db-aggregate on serverless for the time being.

source/unified-test-format/unified-test-format.rst
```diff
@@ -369,6 +369,12 @@ The structure of this object is as follows:
   "sharded" topology, test runners MUST accept any type of sharded cluster (i.e.
   "sharded" implies "sharded-replicaset", but not vice versa).
 
+- ``serverless``: Optional string. Whether or not the test should be run on
+  serverless instances imitating sharded clusters. Valid values are "require",
+  "forbid", and "allow". If "require", the test MUST only be run on serverless
+  instances. If "forbid", the test MUST NOT be run on Serverless instances. If
+  omitted or "allow", this option has no effect.
```
jmikola, line 376: Can you confirm that serverless is entirely distinct from load balancers, in that a load balancer can front mongos _or_ serverless instances, but drivers can also connect directly to serverless without a load balancer?

I just want to make sure I'm understanding both projects correctly before we codify any `runOnRequirement` changes.

For context, let me quote @patrickfreed's previous comment (now hidden) from [this thread](https://github.com/mongodb/specifications/pull/944#discussion_r610796103):

> And yeah, I believe whatever restrictions apply to the serverless proxy probably also apply to the free-tier proxy, since I think the former is a descendant of / shares a lot of code with the latter, not sure about how tightly coupled they are though.

In #935, @divjotarora explained that drivers cannot infer what is behind a load balancer. This came up with respect with `useMultipleMongoses`. The solution there was to feed test runners two additional connection strings (one for an LB fronting a single mongos, and another for a LB fronting multiple mongoses), which can then be used as needed based on whether `useMultipleMongoses` is true or false. While this works in practice, it does invert the typical behavior for `useMultipleMongoses` in that drivers simply decide which connection string to use instead of inspecting the topology and possibly raising an error.

If serverless presents itself as mongos and cannot be differentiated, how is the test runner supposed to enforce the `serverless` option? I assume the test runner (either unified or legacy) will need to consult some explicit indicator (e.g. environment variable, configuration option) and decide based on that. If so, I think we need to clarify that. The test runners themselves need not be prescriptive beyond saying that it must be possible to inform them whether or not serverless is being used. The serverless testing doc itself can advise on a particular method for doing so.

Whatever mechanism is suggested will need to play nice with #935 in the event we ever test load balancers fronting serverless. #935 initially assumes that drivers will _only_ test load balancers with mongos, but I think we should consider how the `serverless` option will interact should we ever add load balancers fronting serverless to the test matrix. Expecting serverless usage to be communicated explicitly (e.g. environment variable or otherwise configuring the test runner) seems like a good way to avoid potential conflict there.

jyemin: By the time Serverless is GA, the only way to interact with it will be via a load balancer.

patrickfreed:
Continued in #964 (couldn't re-open this since a force-push occurred while it was closed)

PR #950, opened 2021-04-08
DRIVERS-1293: Updated CSFLE spec to use hello and legacy hello

JamesKovacs:


source/client-side-encryption/client-side-encryption.rst
```diff
@@ -1079,7 +1079,8 @@ update                   AUTOENCRYPT
 authenticate             BYPASS
 getnonce                 BYPASS
 logout                   BYPASS
-isMaster                 BYPASS
+hello                    BYPASS
```
kevinAlbs, line 1082: This will require a change in libmongocrypt as well so the "hello" command does not undergo the automatic encryption process needlessly. I've created MONGOCRYPT-316 to track that.

kevinAlbs (approved):
LGTM!

source/client-side-encryption/client-side-encryption.rst
JamesKovacs: Good catch!

PR #951, opened 2021-04-08
DRIVERS-1293: Updated APM spec to use hello and legacy hello

JamesKovacs:


kmahar (approved):

source/command-monitoring/command-monitoring.rst
```diff
@@ -250,7 +250,8 @@ value MUST be replaced with an empty BSON document. The list is as follows:
    * - ``copydbgetnonce``
    * - ``copydbsaslstart``
    * - ``copydb``
-   * - ``isMaster`` or ``ismaster`` when ``speculativeAuthenticate`` is present
+   * - ``hello`` or ``hello`` when ``speculativeAuthenticate`` is present
```
behackett, line 253: This and the line after should just be combined. "hello or legacy hello...". You might have to deal with the camel case vs. lower case issue in your "legacy hello" definition elsewhere.

behackett (changes requested):

source/command-monitoring/command-monitoring.rst
alcaeus: I was going to suggest switching to a case-insensitive comparison here to cover this, then realised that the server only accepts `isMaster` and `ismaster`, but rejects `isMAster`. Are there any other cases where we might have to account for different capitalisation of the legacy hello command?

JamesKovacs: Combined and re-worded those lines as well as adding a link to the handshake spec.

source/command-monitoring/command-monitoring.rst
```diff
@@ -250,7 +250,10 @@ value MUST be replaced with an empty BSON document. The list is as follows:
    * - ``copydbgetnonce``
    * - ``copydbsaslstart``
    * - ``copydb``
-   * - ``isMaster`` or ``ismaster`` when ``speculativeAuthenticate`` is present
+   * - ``hello`` or legacy hello, regardless of whether ``speculativeAuthenticate`` is present
```
kmahar, line 253: the previous wording was a little ambiguous but my reading was that we only redact the body of the command (with either capitalization) when speculativeAuthenticate is present. 

this seems to change that, so that we redact the body of this command regardless of if speculativeAuthenticate is present.

is that intentional?

JamesKovacs: That is ambiguous. I parsed it the other way as I glossed over the capitalization difference in `isMaster`. I think you're right that the intention is to redact only when `spectulativeAuthenticate` is present. I'll change that.

kmahar (approved):

alcaeus (approved):

behackett (approved):

durran (approved):

PR #952, opened 2021-04-08
DRIVERS-1577 Note caveats for killAllSessions

jmikola:
https://jira.mongodb.org/browse/DRIVERS-1577

### DIFF
--- a/source/unified-test-format/unified-test-format.rst
+++ b/source/unified-test-format/unified-test-format.rst
@@ -3,13 +3,13 @@ Unified Test Format
 ===================
 
 :Spec Title: Unified Test Format
-:Spec Version: 1.2.3
+:Spec Version: 1.2.4
 :Author: Jeremy Mikola
 :Advisors: Prashant Mital, Isabel Atkinson, Thomas Reggi
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: N/A
-:Last Modified: 2021-03-24
+:Last Modified: 2021-04-08
 
 .. contents::
 
@@ -2635,11 +2635,11 @@ sections.
 Terminating Open Transactions
 `````````````````````````````
 
-Open transactions can cause tests to block indiscriminately. Test runners SHOULD
-terminate all open transactions at the start of a test suite and after each
-failed test by killing all sessions in the cluster. Using the internal
-MongoClient, execute the ``killAllSessions`` command on either the primary or,
-if connected to a sharded cluster, all mongos servers.
+Open transactions can cause tests to block indiscriminately. When connected to
+MongoDB 3.6 or later, test runners SHOULD terminate all open transactions at the
+start of a test suite and after each failed test by killing all sessions in the
+cluster. Using the internal MongoClient, execute the ``killAllSessions`` command
+on either the primary or, if connected to a sharded cluster, all mongos servers.
 
 For example::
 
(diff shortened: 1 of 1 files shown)


source/unified-test-format/unified-test-format.rst
```diff
@@ -3,13 +3,13 @@ Unified Test Format
 ===================
 
 :Spec Title: Unified Test Format
-:Spec Version: 1.2.3
+:Spec Version: 1.2.4
```
jmikola, line 6: I was on the fence about bumping the minor or patch version, but this would avoid conflicts with the other larger PRs on the spec (both of which are introducing schema changes).

source/unified-test-format/unified-test-format.rst
```diff
@@ -2635,22 +2635,34 @@ sections.
 Terminating Open Transactions
 `````````````````````````````
 
-Open transactions can cause tests to block indiscriminately. Test runners SHOULD
-terminate all open transactions at the start of a test suite and after each
-failed test by killing all sessions in the cluster. Using the internal
-MongoClient, execute the ``killAllSessions`` command on either the primary or,
-if connected to a sharded cluster, all mongos servers.
+Open transactions can cause tests to block indiscriminately. When connected to
+MongoDB 3.6 or later, test runners SHOULD terminate all open transactions at the
```
jmikola, line 2639: Just adding "When connected to MongoDB 3.6 or later". I realized Ruby was possibly [ignoring CommandNotFound(59)](https://github.com/mongodb/mongo-ruby-driver/blob/95de4ad759a3b6959493e6d4952239a1aff7827d/spec/runners/unified/test.rb#L358) because it was testing against older servers, so this might address that edge case.

source/unified-test-format/unified-test-format.rst
```diff
@@ -2635,22 +2635,34 @@ sections.
...
       killAllSessions: []
     });
 
-The test runner MAY ignore any command failure with error Interrupted(11601) to
-work around `SERVER-38335`_.
+The test runner MAY ignore the following command failures:
+
+- Interrupted(11601) to work around `SERVER-38335`_.
+- Unauthorized(13) to work around `SERVER-54216`_.
+- CommandNotFound(59) if the command is executed on a pre-3.6 server
```
jmikola, line 2654: Just being flexible in case a test runner does decide to call `killAllSessions` regardless of server version.

source/unified-test-format/unified-test-format.rst
```diff
@@ -2635,22 +2635,34 @@ sections.
...
 .. _SERVER-38335: https://jira.mongodb.org/browse/SERVER-38335
+.. _SERVER-54216: https://jira.mongodb.org/browse/SERVER-54216
+
+Note that Atlas, by design, does not allow database users to kill sessions
+belonging to other users. This makes it impossible to guarantee that an existing
+transaction will not block test execution. To work around this, test runners
+SHOULD either ignore Unauthorized(13) command failures or avoid calling
+``killAllSessions`` altogether when connected to Atlas (e.g. by detecting
+``mongodb.net`` in the hostname or allowing the test runner to be configured
+externally).
```
jmikola, line 2665: The goal of this paragraph is to advise test runners not to call the command on Atlas or at least ignore a failure if they do. I'm intentionally not being prescriptive about how the test runner decides to do that, since I imagine some folks might strongly prefer once mechanism over another. IMO, detecting the hostname is probably the most flexible solution.

p-mongo (approved):

source/unified-test-format/unified-test-format.rst
p-mongo: We ran crud v1, v2 and transaction tests with the same base runner which is probably where this error handling came from.

alcaeus (approved):

PR #953, opened 2021-04-09
DRIVERS-1647: Add appName to fail point at pool-create-min-size-error

BorisDog:
[DRIVERS-1647](https://jira.mongodb.org/browse/DRIVERS-1647)
Updating pool-create-min-size-error test to include appName in the fail point.

### DIFF
--- a/source/connection-monitoring-and-pooling/tests/pool-create-min-size-error.yml
+++ b/source/connection-monitoring-and-pooling/tests/pool-create-min-size-error.yml
@@ -3,8 +3,8 @@ style: integration
 description: error during minPoolSize population clears pool
 runOn:
   -
-    # required for blockConnection in fail point
-    minServerVersion: "4.2.0"
+    # required for appName in fail point
+    minServerVersion: "4.9.0"
 failPoint:
   configureFailPoint: failCommand
   # high amount to ensure not interfered with by monitor checks.
@@ -12,8 +12,10 @@ failPoint:
   data:
     failCommands: ["isMaster"]
     closeConnection: true
+    appName: "poolCreateMinSizeErrorTest"
 poolOptions:
   minPoolSize: 1
+  appName: "poolCreateMinSizeErrorTest"
 operations:
   - name: ready
   - name: waitForEvent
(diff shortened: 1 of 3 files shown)


source/connection-monitoring-and-pooling/tests/pool-create-min-size-error.yml
```diff
@@ -3,15 +3,16 @@ style: integration
...
+    # required for appName in fail point
+    minServerVersion: "4.9.0"
 failPoint:
   configureFailPoint: failCommand
   # high amount to ensure not interfered with by monitor checks.
   mode: { times: 50 }
   data:
     failCommands: ["isMaster"]
     closeConnection: true
+    appName: "poolCreateMinSizeErrorTest"
```
ShaneHarvey, line 15: How is the driver supposed to know to use this appName when creating a client (or pool) for this test?

BorisDog: I don't see clientOptions in integration test format.
So we can add `appName` to `poolOptions` or instruct to extract if from fail point data in 
[Integration Test Runner fail-point section](https://github.com/mongodb/specifications/blob/7bb11d1b0bfc48052de23ccb1493a81d12e9725c/source/connection-monitoring-and-pooling/tests/README.rst#integration-test-runner)

divjotarora: I'd prefer to either put it in `poolOptions` or add a new `clientOptions` document. Extracting it from the fail point data seems hacky and is not extensible because we won't be able to easily support other such options in the future if we need them.

patrickfreed: I'd rather put it in the `poolOptions` than introduce a new `clientOptions`, since these tests can technically be run using just the pool. This is what Rust does, for example.

divjotarora: Adding it to `poolOptions` and documenting that in the tests README SGTM.

BorisDog: Thanks, added to Integration Test Format section. Please lmk if the appName comment is expressive enough.

divjotarora (approved):

PR #955, opened 2021-04-09
DRIVERS-1649 acceptAPIVersion2 -> acceptApiVersion2

kmahar:
https://jira.mongodb.org/browse/DRIVERS-1649

The server renamed this parameter.


alcaeus (approved):
LGTM, thanks!

kmahar:
@alcaeus, good catch, done... should I update the spec version too? I'm not really sure how one would classify this change under semver 🤔

alcaeus:
Tough one...because it seems like a breaking change. Since it's based on a server change I think we should increase the patch version and call it a day.

kmahar:
sgtm. done

PR #956, opened 2021-04-12
include rationale for CSFLE cmds to mongocryptd on unencrypted colls

kevinAlbs:
Not urgent – but I wanted to capture this rationale somewhere.


jyemin (approved):
LGTM

PR #959, opened 2021-04-13
DRIVERS-555 Unified test format changes for CSOT

divjotarora:
This PR is a rebased and updated version of #873. It only contains Unified Test Format changes and modifications/additions to tests in the `valid-pass` and `invalid` directories. The actual CSOT spec tests will be introduced in another PR to avoid making this one larger than necessary.

Notes for reviewers:

1. This PR bumps the schema version in the unified test format to 1.8 because 1.9
2. Many of the existing tests for `collectionData` in the `invalid` directory weren't functioning as intended because they had an additional `foo` property in the collection entity declaration. As a result, they were considered invalid because the collection entity was not valid and the `initialData` field was not being checked. These have been fixed in this PR.
3. ~The aforementioned `collectionData` test files were renamed from `collectionData*` to `initialCollectionData*` to reflect a similar change in the test format spec. A new set of `outcomeCollectionData*` tests have also been added.~

PR #960, opened 2021-04-13
DRIVERS-555 Add spec and prose tests for CSOT

divjotarora:
This PR is a rebased and updated version of #873 and contains the spec/prose tests for CSOT as well as an update to the Unified Test Format Makefile to validate the tests in the `client-side-operations-timeout/tests` directory. Implementing the new CSOT tests also requires the unified test format changes described in #959.

One of the linked GitHub checks fail because the Unified Test Format Makefile is validating using `schema-1.3.json`, which does not contain the new features from #959, so the CSOT tests fail validation. I have verified locally that schema validation succeeds if the changes from #959 are incorporated.

Note: Some of the new tests may need an additional `serverless` requirement if we find that they fail due to known limitations of serverless clusters.


source/transactions-convenient-api/tests/README.rst
```diff
@@ -196,8 +196,8 @@ should be checked:
    the retry timeout has been exceeded, ``withTransaction`` should propagate the
    error to its caller.
  * If committing raises an error with the UnknownTransactionCommitResult label,
-   the error is not a write concern timeout, and the retry timeout has been
-   exceeded, ``withTransaction`` should propagate the error to its caller.
+   and the retry timeout has been exceeded, ``withTransaction`` should
+   propagate the error to its caller.
```
jmikola, line 200: If you're revising a prose test, I think this warrants a changelog entry. This should also be highlighted in the final description of DRIVERS-555 so languages know to consider this change and adjust their prose tests accordingly.

alcaeus (approved):

rozza (approved):

jmikola (approved):
Noticed missing newline at end of a file (perhaps from template generation), but LGTM.

DmitryLukyanov (approved):
LGTM

source/client-side-encryption/tests/timeoutMS.yml
```diff
@@ -0,0 +1,65 @@
...
+      autoEncryptOpts:
+        kmsProviders:
+          aws: {} # Credentials filled in from environment.
+      timeoutMS: 50
+    operations:
+      - name: insertOne
+        arguments:
+          document: *doc0
+        result:
+          isTimeoutError: true
```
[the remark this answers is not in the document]
divjotarora: This is from template generation and matches the formatting of the other YAML files in `client-side-encryption/tests`. Adding it manually causes the linked GitHub action to fail and I don't want to modify the template generation script in this PR, so I'd prefer to leave this as-is.

durran (approved):
LGTM

ShaneHarvey:
Updated to fix merge conflicts.

source/client-side-operations-timeout/tests/global-timeoutMS.yml
```diff
@@ -0,0 +1,3142 @@
...
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+              failCommands: ["aggregate"]
+              blockConnection: true
+              blockTimeMS: 15
+      - name: createChangeStream
+        object: *client
+        arguments:
+          pipeline: []
```
benjirewis, line 311: The Go driver expects a `saveResultAsEntity` field for all the `createChangeStream` operations in these new tests and errors in its absence. The unified test format does not seem to imply that there _must_ be a `saveResultAsEntity` field for `createChangeStream`, so I _could_ just throw away the result of the `watch` command, but wanted to point out that this deviates from existing `createChangeStream` operations (by nature of the template used to create the tests).

source/client-side-operations-timeout/tests/global-timeoutMS.yml
```diff
@@ -0,0 +1,3142 @@
...
+        arguments:
+          client: *failPointClient
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+              failCommands: ["count"]
+              blockConnection: true
+              blockTimeMS: 60
+      - name: count
```
benjirewis, line 937: There is no `count` command in the Go driver (nor in a couple other, new drivers), so I think we'll skip tests with `count`.

source/client-side-operations-timeout/tests/global-timeoutMS.yml
```diff
@@ -0,0 +1,3142 @@
...
+      - name: estimatedDocumentCount
+        object: *collection
+        
+        expectError:
+          isTimeoutError: true
+    expectEvents:
+      - client: *client
+        events:
+          - commandStartedEvent:
+              commandName: count
```
benjirewis, line 1140: Tests for `estimatedDocumentCount` expect a `count` `commandStartedEvent` in this PR. Depending on the wire version/whether drivers have implemented DRIVERS-2228, an `aggregate` may be started for `estimatedDocumentCount`. I think tests with `estimatedDocumentCount` will need some `min/maxServerVersion` logic.

source/client-side-operations-timeout/tests/global-timeoutMS.yml
```diff
@@ -0,0 +1,3142 @@
...
+        arguments:
+          client: *failPointClient
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+              failCommands: ["listIndexes"]
+              blockConnection: true
+              blockTimeMS: 60
+      - name: listIndexNames
```
benjirewis, line 1610: The Go driver has no `listIndexNames` operation, so we'll skip tests with this operation. We have `ListIndexSpecifications`, but it's not quite the same.

source/client-side-operations-timeout/tests/override-operation-timeoutMS.yml
```diff
@@ -0,0 +1,1920 @@
+# Tests in this file are generated from override-operation-timeoutMS.yml.template.
+
+description: "timeoutMS can be overriden for an operation"
```
benjirewis, line 3: I don't think we'll run any of the tests in this file in the Go driver. Users can pass a `Context` with a deadline to any operation to specify an operation-level timeout. Allowing users to set `timeoutMS` on an operation _and_ pass a `Context` with a deadline would be redundant and confusing.

benjirewis (commented):
I'm slowly POC'ing these tests in the Go driver. Left a few comments I had after syncing `global-timeoutMS`, `override-collection-timeoutMS`, `override-database-timeoutMS` and `override-operation-timeoutMS`.

source/client-side-operations-timeout/tests/override-operation-timeoutMS.yml
jmikola: On an unrelated note, I realized there's a typo in the test name: overridden

source/client-side-operations-timeout/tests/global-timeoutMS.yml
ShaneHarvey: Yep drivers without the deprecated count helper will skip tests for count. Note there are already tests for count:
```
git grep ' name: count$'
source/client-side-encryption/etc/test-templates/count.yml.template:19:      - name: count
source/client-side-encryption/etc/test-templates/count.yml.template:50:      - name: count
source/client-side-encryption/tests/count.yml:19:      - name: count
source/client-side-encryption/tests/count.yml:50:      - name: count
source/crud/tests/v1/read/count-collation.yml:20:            name: count
source/crud/tests/v1/read/count-empty.yml:24:            name: count
source/crud/tests/v1/read/count.yml:48:            name: count
source/crud/tests/v1/read/count.yml:57:            name: count
source/crud/tests/v1/read/count.yml:67:            name: count
source/retryable-reads/tests/legacy/count-serverErrors.yml:26:                    name: count
source/retryable-reads/tests/legacy/count.yml:22:                    name: count
source/transactions/tests/legacy/count.yml:25:      - name: count
```

source/client-side-operations-timeout/tests/global-timeoutMS.yml
ShaneHarvey: Good catch. The unified test format says (emphasis mine):
> ... before the resulting change stream *_might be saved_* with [operation.saveResultAsEntity](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#operation-saveresultasentity).

So I think saveResultAsEntity can be optional. I tried adding saveResultAsEntity but it's tedious and error prone because it's mutually exclusive with expectError. Can we leave it optional for now?

source/client-side-operations-timeout/tests/global-timeoutMS.yml
ShaneHarvey: We can't use wireVersions because the count/aggregate behavior depends on the driver, not the server. Drivers will just need to implement [DRIVERS-2228](https://jira.mongodb.org/browse/DRIVERS-2228) first.

source/client-side-operations-timeout/tests/global-timeoutMS.yml
ShaneHarvey: Yep I believe that's expected. Python also does not have a listIndexNames helper.

source/client-side-operations-timeout/tests/override-operation-timeoutMS.yml
ShaneHarvey: What do you think about running these tests by having the test runner translate the `timeoutMS` argument into a Context with an equivalent deadline. This is what I've done in Python. And I've fixed the typo.

source/client-side-operations-timeout/tests/global-timeoutMS.yml
jmikola: Not my intention to create additional work for anyone, but there may be value in introduce a new `valid-pass` test in the unified spec to demonstrate `createChangeStream` being used w/o `saveResultAsEntity` so that the only example of that usage isn't confined to the CSOT spec.

Feel free to ignore if this seems superfluous.

source/client-side-operations-timeout/tests/global-timeoutMS.yml
jmikola: cc: @kmahar re: https://github.com/mongodb/specifications/pull/1206#discussion_r865129646

source/client-side-operations-timeout/tests/global-timeoutMS.yml
jmikola: FWIW, `listIndexNames` has existed in the legacy retryable reads spec tests for some time. This is likely the first instance of that operation being used in a unified spec test, but I concur that drivers can/should just skip this as needed.

source/client-side-operations-timeout/tests/override-operation-timeoutMS.yml
jmikola: This might require more discussion in the unified spec, but would the proposal be that drivers implementing `timeoutMS` via a context modify their execution such that each operation with a `timeoutMS` option runs within its own context?

ShaneHarvey: Yes, but I think the description of that behavior should live in the CSOT test readme, not the unified spec.

benjirewis: I can try to do that in Go and report back. @jmikola yep that's my understanding of the proposal.

source/client-side-operations-timeout/tests/global-timeoutMS.yml
benjirewis: As described in that linked thread, I'll add the logic in Go to automatically skip tests with unsupported operations. Other drivers that decode entire test files and the contained operations into native types will likely find issue with the `count` and `listIndexNames` tests.

source/client-side-operations-timeout/tests/global-timeoutMS.yml
ShaneHarvey: Done.

source/client-side-operations-timeout/tests/override-operation-timeoutMS.yml
jmikola: > but I think the description of that behavior should live in the CSOT test readme, not the unified spec.

No objections to hosting that documentation in the CSOT spec, but I think it'd be prudent to reference the CSOT spec's explanation from [Executing an Operation](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#executing-an-operation) in the unified spec. In the event `timeoutMS` ever shows up in non-CSOT spec tests, it'd be helpful to have this mentioned _somewhere_ in the test runner implementation directions.

ShaneHarvey: My idea doesn't work (at least for Go) so drivers won't be using this approach. Given that I don't think there's anything special to document in the unified format regarding the timeoutMS option.

PR #962, opened 2021-04-15
DRIVERS-1130 Convert crud-v2 tests to the unified format

alcaeus:
[DRIVERS-1130](https://jira.mongodb.org/browse/DRIVERS-1130)

This PR ports all crud-v2 tests to the unified format. The transformation was done using an tool (needs some polishing before being published) with some manual corrections to fix issues.

I've done a quick test of most tests locally and PHP passes almost all of them. The failing tests are expected due to us not implementing the functionality. I'll be running the tests on evergreen to see if any other issues crop up. However, I'd appreciate it if other folks could run these tests and let me know if they encounter issues.


source/crud/tests/unified/aggregate-out-readConcern.yml
```diff
@@ -0,0 +1,200 @@
...
+      collectionName: *collection_name
+      collectionOptions:
+        readConcern: { level: "linearizable" }
+  -
+    collection:
+      id: &collection_readConcern_invalid collection_readConcern_invalid
+      database: database0
+      collectionName: *collection_name
+      collectionOptions:
+        readConcern: { level: "!invalid123" }
```
jyemin, line 59: This is problematic for any driver than uses an enum for ReadConcern level, as there is no way to create a ReadConcern with an invalid level.  Because of that, the test runner will fail to create this entity, and every test in this file will fails as a result.

If it's important to test invalid read concern levels for drivers that allow them, I suggest moving the test that uses this entity (and the entity itself) to a separate file, so that drivers can skip the test.

I'm also fine with not testing this case.

source/crud/tests/unified/bulkWrite-arrayFilters-clientError.yml
```diff
@@ -0,0 +1,103 @@
...
+            -
+              updateOne:
+                filter: {  }
+                update:
+                  $set:
+                    y.0.b: 2
+                arrayFilters:
+                  -
+                    i.b: 1
+          options:
```
jyemin, line 65: To match existing poc-crud.yml, remove the `options` wrapper for `ordered` in all the bulk write tests.

jyemin (changes requested):
Looking great.

kmahar:
Synced the tests to Swift. I ran into the same bulk issue as Jeff mentioned above, and had to skip all of the `unacknowledged-X-hint-clientError.json` ones as I imagine you probably did too in PHP, due to [CDRIVER-3630](https://jira.mongodb.org/browse/CDRIVER-3630) being closed as won't fix. 

I also had to skip `updateWithPipeline.json`  as we haven't added that API to Swift yet ([SWIFT-560](https://jira.mongodb.org/browse/SWIFT-560)). 

Everything else passes 🥳

Thanks @alcaeus!

source/crud/tests/unified/bulkWrite-arrayFilters-clientError.yml
jmikola: See also: [Expressing Required and Optional Parameters](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#expressing-required-and-optional-parameters)

source/crud/tests/unified/aggregate-out-readConcern.yml
jmikola: How were you previously passing the [v2 test](https://github.com/mongodb/specifications/blob/master/source/crud/tests/v2/aggregate-out-readConcern.yml#L95)? I assume `error: true` was satisfied by an error attempting to construct the collection object for the operation, without ever getting around to invoking the `aggregate` method?

I don't have any objections to removing this test, since it seems like something that belongs in the [read/write concern spec tests](https://github.com/mongodb/specifications/tree/master/source/read-write-concern/tests) if anything.

jyemin: Grepping, I found this:

                        skipTest(testDocument, test.asDocument())
                                // Skip test because is assumes the driver sends an invalid read concern level to the server.  But the
                                // Java driver uses an enum for read concern level so it's a client-side not a server-side error
                                || description.equals("invalid readConcern with out stage")});

jyemin: I can't do the same thing here because the invalid read concern is in the entities section, which applies to all the tests in the file not just the one that uses that entity

alcaeus: I can remove that test.

alcaeus:
Latest files pass for PHP: https://spruce.mongodb.com/version/607946ece3c331099632f200/tasks

jyemin (approved):

source/crud/tests/unified/aggregate-merge.yml
```diff
@@ -0,0 +1,188 @@
...
+  -
+    collection:
+      id: &collection_readConcern_available collection_readConcern_available
+      database: database0
+      collectionName: *collection_name
+      collectionOptions:
+        readConcern: { level: "available" }
+runOnRequirements:
+  -
+    minServerVersion: 4.1.11
```
jmikola, line 48: I assume we don't have much control over the YAML produced by `mongodb-spec-converter`, but is it possible to do without the line break and indent for array elements (e.g. `- minServerVersion: 4.1.11` here instead)? That would be consistent with most of our hand-written files and save space.

Additionally, I'd propose ordering the top-level fields in the tests as:

 * description
 * schemaVersion
 * runOnRequirements
 * createEntites
 * initialData
 * tests

Having `runOnRequirements` up near the top helps when scanning test files.

source/crud/tests/unified/aggregate-merge.yml
```diff
@@ -0,0 +1,188 @@
...
+
+description: aggregate-merge
+schemaVersion: '1.1'
+createEntities:
+  -
+    client:
+      id: &client0 client0
+      useMultipleMongoses: false
+      uriOptions:
+        retryReads: false
```
jmikola, line 12: Was `retryReads: false` and `useMultipleMongoses: false` added by default to all tests? I don't think either should be necessary.

`retryReads` in particular doesn't make sense in this file because `$merge` is a write anyway. And with respect to mongos, I don't think any of the CRUD tests care about the number of mongoses.

source/crud/tests/unified/aggregate-merge.yml
```diff
@@ -0,0 +1,188 @@
...
+        events:
+          -
+            commandStartedEvent:
+              command:
+                aggregate: *collection_name
+                pipeline: *pipeline
+                cursor: {  }
+    outcome: *outcome
+  -
+    description: 'Aggregate with $merge and majority readConcern'
```
jmikola, line 133: I was surprised that a standalone server accepts "majority" readConcern, so I double-checked this on a 4.9.0 standalone. No problems there, which explains why the original test didn't have specific topology requirements.

source/crud/tests/unified/aggregate-out-readConcern.yml
```diff
@@ -0,0 +1,173 @@
...
+        client: *client0
+        events:
+          -
+            commandStartedEvent:
+              command:
+                aggregate: *collection_name
+                pipeline: &pipeline
+                  - { $sort: { x: 1 } }
+                  - { $match: { _id: { $gt: 1 } } }
+                  - { $out: other_test_collection }
```
jmikola, line 102: Why is it that the converter script left this as-is from the original, but expanded the pipeline in the "Aggregate with $merge" test? IMO, the concise format we have here is preferred.

source/crud/tests/unified/bulkWrite-arrayFilters-clientError.yml
```diff
@@ -0,0 +1,101 @@
...
+  -
+    maxServerVersion: 3.5.5
+initialData:
+  -
+    collectionName: *collection_name
+    databaseName: *database_name
+    documents:
+      -
+        _id: 1
+        'y':
```
jmikola, line 35: Why is the "y" key quoted, but not "_id" above? I would expect these to be consistent if they were both produced by the converter script. Previous files didn't quote fields (e.g. `x`) in `intialData.documents` either.

source/crud/tests/unified/bulkWrite-arrayFilters-clientError.yml
```diff
@@ -0,0 +1,101 @@
...
+      -
+        _id: 2
+        'y':
+          -
+            b: 0
+          -
+            b: 1
+tests:
+  -
+    description: 'BulkWrite on server that doesn''t support arrayFilters'
```
jmikola, line 49: This is odd, as the original `description` used double quotes around the string to avoid a conflict with the single qutoe within. Is the sequence of two single quotes an escaping mechanism in YAML?

source/crud/tests/unified/bulkWrite-arrayFilters-clientError.yml
```diff
@@ -0,0 +1,101 @@
...
+        object: *collection0
+        name: bulkWrite
+        arguments:
+          requests:
+            -
+              updateOne:
+                filter: {  }
+                update:
+                  $set:
+                    y.0.b: 2
```
jmikola, line 61: Interesting that this is a case where the original test _did_ quote the field name. I suppose it makes no different in practice, so long as the field name doesn't contain a colon, though.

source/crud/tests/unified/bulkWrite-arrayFilters.yml
```diff
@@ -0,0 +1,180 @@
...
+        object: *collection0
+        name: bulkWrite
+        arguments:
+          requests:
+            -
+              updateOne:
+                filter: {  }
+                update:
+                  $set:
+                    'y.$[i].b': 2
```
jmikola, line 61: Did the dollar or brackets cause the spec converter to quote this field, or is that due to internal logic in Symfony's YAML library? If this is something you have control over, I'd suggest maybe just applying the quoting if the field contains _any_ special characters. That'd probably be consistent with most of our hand-written files.

source/crud/tests/unified/bulkWrite-delete-hint-serverError.yml
```diff
@@ -0,0 +1,145 @@
...
+                hint: &hint_string _id_
+            -
+              deleteOne:
+                filter: &deleteOne_filter2
+                  _id: 2
+                hint: &hint_doc
+                  _id: 1
+          ordered: true
+        expectError:
+          isError: true
```
jmikola, line 68: This would be a good use of `isClientError: false`, but I realize it's outside the scope of this PR. Perhaps the best course of action would be a ticket to go through and revise the converted tests after the fact. I know we don't use `commandFailedEvent` very much in spec tests, but that may be another thing to explore for server-side error tests.

source/crud/tests/unified/bulkWrite-update-hint-clientError.yml
```diff
@@ -0,0 +1,151 @@
...
+                hint: &hint_string _id_
+            -
+              updateOne:
+                filter: *updateOne_filter
+                update: *updateOne_update
+                hint: &hint_doc
+                  _id: 1
+          ordered: true
+        expectError:
+          isError: true
```
jmikola, line 70: In contrast to my previous comment about server-side errors, this would be a case to use `isClientError: true` instead.

source/crud/tests/unified/db-aggregate.yml
```diff
@@ -0,0 +1,73 @@
...
+    database:
+      id: &database0 database0
+      client: client0
+      databaseName: &database_name admin
+  -
+    collection:
+      id: &collection0 collection0
+      database: database0
+      collectionName: &collection_name crud-v2
+runOnRequirements:
```
jmikola, line 25: This is relevant to @patrickfreed since he was in the process of adding `serverless: false` to the legacy CRUD test format. If we migrate these tests to the unified runner that should no longer be necessary; however, it'd require bumping the `schemaVersion` of some of these tests to incorporate his unified format changes.

source/crud/tests/unified/updateMany-hint-serverError.yml
```diff
@@ -0,0 +1,120 @@
...
+                updates:
+                  -
+                    q: *filter
+                    u: *update
+                    multi: true
+                    hint:
+                      _id: 1
+                    upsert:
+                      $$unsetOrMatches: false
+    outcome: *outcome
```
jmikola, line 120: I noticed most files had a trailing newline. Should all of them, or is this something being inherited from the original input file?

kmahar (approved):
tried out the updated bulk write tests - all good on [Swift](https://spruce.mongodb.com/version/6079b7df3e8e86054fbb2236/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

source/crud/tests/unified/bulkWrite-arrayFilters-clientError.yml
alcaeus: y is a shorthand for true in some parsers, hence the quotes applied by symfony/yaml

source/crud/tests/unified/aggregate-merge.yml
```diff
@@ -0,0 +1,188 @@
+# This file was created automatically using mongodb-spec-converter.
+# Please review the generated file, then remove this notice.
```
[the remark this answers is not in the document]
alcaeus: Since I didn't really work on the files (e.g. fix formatting, etc.) I decided to leave the note in so the first person that manually changes this file can do their thing.

source/crud/tests/unified/aggregate-merge.yml
alcaeus: I can try to influence the order, but will have to see how difficult this would be.

source/crud/tests/unified/aggregate-merge.yml
alcaeus: I believe I copied these from existing tests. Happy to drop them if they are not necessary.

source/crud/tests/unified/bulkWrite-arrayFilters-clientError.yml
alcaeus: Yes, that's how it's escaped. I couldn't find a way to change what quotes are used when dumping YAML, which is why everything is quoted using single quotes now.

source/crud/tests/unified/bulkWrite-arrayFilters.yml
alcaeus: There's no control over this, and the library seems to only quote if the string contains problematic characters.

source/crud/tests/unified/bulkWrite-delete-hint-serverError.yml
alcaeus: Yes, I noticed that the error expectations in the unified format are significantly more advanced than what we had. It's difficult for the converter to make these changes and I didn't want to stray too far from the automated files. We might be able to use the presence of a `result` field to decide that we're expecting a server-side error, but can't really use the absence of one to expect a client-side error (as the test may be expecting a server-side error without checking the command result).

source/crud/tests/unified/updateMany-hint-serverError.yml
alcaeus: The trailing newlines were added by PhpStorm whenever I made manual changes to the files. I'll ensure every file has a trailing newline.

source/crud/tests/unified/aggregate-out-readConcern.yml
alcaeus: The symfony/yaml library can only switch to inline dumping based on the level and no other criteria. In addition, it does not support this concise format out of the box, and this may have been one of the few places where I bothered to reformat the output.

I'd have to do some testing to make sure I'm not breaking anything, but in theory I could replace `-\n` with a single `-` and drop all but one of the spaces that follow it. Alternatively, we could try to find a tool that is able to format YAML files how we want them and enforce that across the board. I'm not aware of any such tool, but maybe someone else is.

source/crud/tests/unified/aggregate-merge.yml
jmikola: I'm not sure where they would have been copied from, as I can't find any reference to "retryReads" or "mongos" in the v2 CRUD tests, nor the CRUD POC in the unified test format spec. I vote on removing these.

source/crud/tests/unified/aggregate-out-readConcern.yml
jmikola: No worries. If this is something you can easily fix with a search/replace feel free; otherwise, I think we can just get to it as needed when we start modifying the converted files down the line.

source/crud/tests/unified/bulkWrite-delete-hint-serverError.yml
jmikola: Feel free to log a ticket for improving the error expectations and I can triage it down the line.

jmikola:
I repurposed [DRIVERS-1130](https://jira.mongodb.org/browse/DRIVERS-1130) to use for this task.

source/crud/tests/unified/aggregate-merge.yml
alcaeus: I've applied the order you suggested.

source/crud/tests/README.rst
```diff
@@ -19,118 +19,19 @@ version requirements as noted by the ``runOn`` section, if provided.
...
 and placed under ``unified/``. Until such time that all original tests have been ported
 to the unified test format, tests in each format will be grouped in their own subdirectory:
 
 - ``v1/``: Legacy-v1 format tests
-- ``v2/``: Legacy-v2 format tests
 - ``unified/``: Tests using the `unified test format <../../../../unified-test-format/unified-test-format.rst>`_
 
 Since some drivers may not have a unified test runner capable of executing tests
-in all three formats, segregating tests in this manner will make it easier for
+in all two formats, segregating tests in this manner will make it easier for
```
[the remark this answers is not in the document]
alcaeus: Removed the paragraph entirely.

source/crud/tests/unified/aggregate-merge.yml
alcaeus: Removed.

source/crud/tests/unified/aggregate-out-readConcern.yml
alcaeus: I found [yamlfmt](https://pypi.org/project/yamlfmt/) which can format the files in place, but I dislike the indentation it suggests:
```diff
 runOnRequirements:
-  -
-    minServerVersion: 4.1.11
+- minServerVersion: 4.1.11
```

If we're fine with that style, I'm happy to apply this to all generated files. This could also be done separately, so no need for a decision now.

source/crud/tests/unified/bulkWrite-delete-hint-serverError.yml
alcaeus: I created [DRIVERS-1698](https://jira.mongodb.org/browse/DRIVERS-1698) to track the effort.

source/crud/tests/unified/aggregate-out-readConcern.yml
jmikola: Very odd that it doesn't indent. Let's punt on this for now.

jmikola (approved):

PR #964, opened 2021-04-19
DRIVERS-1603 Test format updates for serverless testing

patrickfreed:
DRIVERS-1603

This PR updates the unified test format to support serverless testing.

Any changes to the CRUD v2 tests were dropped off this pending the merge of #962

### DIFF
--- a/source/crud/tests/unified/aggregate.yml
+++ b/source/crud/tests/unified/aggregate.yml
@@ -1,0 +1,79 @@
+description: "aggregate"
+
+schemaVersion: "1.0"
+
+createEntities:
+  - client:
+      id: &client0 client0
+      useMultipleMongoses: true # ensure cursors pin to a single server
+      observeEvents: [ commandStartedEvent ]
+  - database:
+      id: &database0 database0
+      client: *client0
+      databaseName: &database0Name aggregate-tests
+  - collection:
+      id: &collection0 collection0
+      database: *database0
+      collectionName: &collection0Name coll0
+
+initialData:
+  - collectionName: *collection0Name
+    databaseName: *database0Name
+    documents:
+      - { _id: 1, x: 11 }
+      - { _id: 2, x: 22 }
+      - { _id: 3, x: 33 }
+      - { _id: 4, x: 44 }
+      - { _id: 5, x: 55 }
+      - { _id: 6, x: 66 }
+      - { _id: 7, x: 77 }
+      - { _id: 8, x: 88 }
+
+tests:
+  - description: "aggregate with multiple batches works"
+    operations:
+      - name: aggregate
+        arguments:
+          pipeline: [ { $match: { _id: { $gt: 1 } }} ]
+          batchSize: 2
+        object: *collection0
+        expectResult:
+          - { _id: 2, x: 22 }
+          - { _id: 3, x: 33 }
+          - { _id: 4, x: 44 }
+          - { _id: 5, x: 55 }
+          - { _id: 6, x: 66 }
+          - { _id: 7, x: 77 }
+          - { _id: 8, x: 88 }
+    expectEvents:
+      - client: *client0
+        events:
+          - commandStartedEvent:
+              command:
+                aggregate: *collection0Name
+                pipeline: [ { $match: { _id: { $gt: 1 } }} ]
+                cursor: { batchSize: 2 }
+              commandName: aggregate
+              databaseName: *database0Name
+          - commandStartedEvent:
+              command:
+                getMore: { $$exists: true }
+                collection: *collection0Name
+                batchSize: 2
+              commandName: getMore
+              databaseName: *database0Name
+          - commandStartedEvent:
+              command:
+                getMore: { $$exists: true }
+                collection: *collection0Name
+                batchSize: 2
+              commandName: getMore
+              databaseName: *database0Name
+          - commandStartedEvent:
+              command:
+                getMore: { $$exists: true }
+                collection: *collection0Name
+                batchSize: 2
+              commandName: getMore
+              databaseName: *database0Name
+
--- a/source/crud/tests/unified/aggregate.yml
+++ b/source/crud/tests/unified/aggregate.yml
@@ -57,21 +57,21 @@ tests:
               databaseName: *database0Name
           - commandStartedEvent:
               command:
-                getMore: { $$exists: true }
+                getMore: { $$type: [ int, long ] }
                 collection: *collection0Name
                 batchSize: 2
               commandName: getMore
               databaseName: *database0Name
           - commandStartedEvent:
               command:
-                getMore: { $$exists: true }
+                getMore: { $$type: [ int, long ] }
                 collection: *collection0Name
                 batchSize: 2
               commandName: getMore
               databaseName: *database0Name
           - commandStartedEvent:
               command:
-                getMore: { $$exists: true }
+                getMore: { $$type: [ int, long ] }
                 collection: *collection0Name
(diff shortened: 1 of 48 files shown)


source/crud/tests/unified/aggregate.yml
```diff
@@ -0,0 +1,79 @@
...
+          - commandStartedEvent:
+              command:
+                aggregate: *collection0Name
+                pipeline: [ { $match: { _id: { $gt: 1 } }} ]
+                cursor: { batchSize: 2 }
+              commandName: aggregate
+              databaseName: *database0Name
+          - commandStartedEvent:
+              command:
+                getMore: { $$exists: true }
```
jmikola, line 60: Other unified tests use `getMore: { $$type: [ int, long ] }` if you'd like to be strict (see command monitoring POC).

source/crud/tests/unified/find.yml
```diff
@@ -0,0 +1,78 @@
...
+    databaseName: *database0Name
+    documents:
+      - { _id: 1, x: 11 }
+      - { _id: 2, x: 22 }
+      - { _id: 3, x: 33 }
+      - { _id: 4, x: 44 }
+      - { _id: 5, x: 55 }
+      - { _id: 6, x: 66 }
+      - { _id: 7, x: 77 }
+      - { _id: 8, x: 88 }
```
jmikola, line 30: Are eight documents necessary? Five seems sufficient if you wanted to cover two cases of going from initial response to first getMore _and_ a subsequent getMore.

source/unified-test-format/schema-1.4.json
```diff
@@ -55,6 +55,10 @@
             "enum": ["single", "replicaset", "sharded", "sharded-replicaset", "load-balanced"]
           }
         },
+        "serverless": {
+          "type": "string",
+          "enum": ["require", "forbid", "allow"]
```
jmikola, line 60: Please add two `invalid/` tests:

 * runOnRequirement-serverless-type.yml with a non-string value
 * runOnRequirement-serverless-enum.yml with an unacceptable string value

There should be enough prior art to base these off of.

patrickfreed (commented):
A few of the blocker issues in the proxy were fixed, so I updated the various skips to reflect that. Drivers should now be able to run the retryable reads/writes, versioned API, and sessions tests now. I updated the google doc and tests accordingly.

source/crud/tests/unified/find.yml
jmikola: Quoting [`find` command docs](https://docs.mongodb.com/manual/reference/command/find/):

> Unlike the previous wire protocol version, a batchSize of 1 for the find command does not close the cursor.

I wasn't aware that the `find` command differed in this respect from OP_QUERY, but given that this test file doesn't specify a `minServerVersion` I think we should continue to use `batchSize: 2` for compatibility unless you'd like to specify `minServerVersion: "3.2" for these tests.

source/crud/tests/unified/aggregate.yml
```diff
@@ -0,0 +1,64 @@
...
+          - { _id: 3, x: 33 }
+          - { _id: 4, x: 44 }
+    expectEvents:
+      - client: *client0
+        events:
+          - commandStartedEvent:
+              command:
+                aggregate: *collection0Name
+                pipeline: [ { $match: { _id: { $gt: 1 } }} ]
+                cursor: { batchSize: 1 }
```
jmikola, line 47: [Aggregate Data Specifying Batch Size](https://docs.mongodb.com/manual/reference/command/aggregate/#aggregate-data-specifying-batch-size) doesn't talk about any special handling for `batchSize: 1` so I think this is OK regardless of server version (unlike `find`).

source/crud/tests/unified/find.yml
```diff
@@ -0,0 +1,63 @@
...
+      - { _id: 3, x: 33 }
+      - { _id: 4, x: 44 }
+
+tests:
+  - description: "find with multiple batches works"
+    operations:
+      - name: find
+        arguments:
+          filter: { _id: { $gt: 1 } }
+          batchSize: 1
```
jmikola, line 34: Quoting [`find` command docs](https://docs.mongodb.com/manual/reference/command/find/):

> Unlike the previous wire protocol version, a batchSize of 1 for the find command does not close the cursor.

I wasn't aware that the `find` command differed in this respect from OP_QUERY, but given that this test file doesn't specify a `minServerVersion` I think we should continue to use `batchSize: 2` for compatibility unless you'd like to specify `minServerVersion: "3.2" for these tests.

source/unified-test-format/schema-1.4.json
jmikola, line 60: One last thought before we codify this: would "deny" be preferable since that's the inverse of allow in other contexts (e.g. allowlist/denylist)? I don't feel strongly, but curious if others do.

source/versioned-api/tests/test-commands-strict-mode.yml
```diff
@@ -6,6 +6,8 @@ runOnRequirements:
   - minServerVersion: "4.9"
     serverParameters:
       enableTestCommands: true
+    # serverless does not support the testVersion2 command
```
kevinAlbs, line 9: Nit: this sounds as if `testVersion2` is supported in some cases. In this case serverless gives a different error:
```
{
        "ok" : 0,
        "errmsg" : "command not found",
        "code" : 59,
        "codeName" : "CommandNotFound"
}
```

Perhaps reword to this?

```suggestion
    # serverless gives a different error for unrecognized testVersion2 command
```

kevinAlbs (approved):
LGTM with Jeremy's comment about batchSize:1 addressed.

source/crud/tests/unified/find.yml
patrickfreed: Updated to use batchSize 2

source/unified-test-format/schema-1.4.json
patrickfreed: I don't really have any strong feelings either on this and would be happy to change it to whatever we think is best.

jmikola: Fine to leave as-is.

jmikola (approved):

patrickfreed:
I just updated the transactions tests to include the necessary skips and updated the Google doc to require running them as per discussion in slack. Note that I couldn't verify the convenient API ones because the Swift driver doesn't have the convenient API yet, so I excluded those for now from the doc.

source/transactions/tests/legacy/mongos-recovery-token.yml
```diff
@@ -2,6 +2,8 @@ runOn:
     -
         minServerVersion: "4.1.8"
         topology: ["sharded"]
+        # serverless proxy doesn't use recovery tokens
+        serverless: "forbid"
```
kevinAlbs, line 6: This made me think it may be worth testing a retry scenarios specific to serverless.

I thought it could be a risk that a driver assumes that responses to a sharded cluster always include the `recoveryToken` field. But I think the existing `tests/legacy/retryable-commit.yml` and `tests/legacy/retryable-writes.yml` should cover those cases.

kevinAlbs (approved):

source/transactions/tests/legacy/mongos-recovery-token.yml
patrickfreed: Yeah I think for now the coverage of that case is acceptable. Once it moves behind the LB and we have to write new tests anyways, that would probably be a good time to revisit this.

PR #965, opened 2021-04-19
DRIVERS-720 small fix to unified test format changes for LB support

jyemin:
Divjot, 

Java driver fails this type assertion on 2.6 and 3.0, since the cursorsKilled field isn't manufactured when translating from OP_KILL_CURSORS.  This changes works around that.  

How is Go driver dealing with this this?

### DIFF
--- a/source/unified-test-format/tests/valid-pass/entity-find-cursor.yml
+++ b/source/unified-test-format/tests/valid-pass/entity-find-cursor.yml
@@ -84,5 +84,6 @@ tests:
               commandName: killCursors
           - commandSucceededEvent:
               reply:
-                cursorsKilled: { $$unsetorMatches: { $$type: array } }
+                # No cursorsKilled field expected on pre-3.2 servers
+                cursorsKilled: { $$unsetOrMatches: { $$type: array } }
               commandName: killCursors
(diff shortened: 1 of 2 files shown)


source/unified-test-format/tests/valid-pass/entity-find-cursor.yml
```diff
@@ -84,5 +84,5 @@ tests:
               commandName: killCursors
           - commandSucceededEvent:
               reply:
-                cursorsKilled: { $$type: array }
```
jmikola, line 87: It's quite possible Go simply wasn't running this test on a pre-3.2 server. An alternative solution may be to specify a runOnRequirement such that this only runs when `killCursors` is actually used. 

If this line stays, please add a comment above this line explaining why you're using `$$unsetOrMatches`. Also, please correct the capitalization: [`$unsetOrMatches`](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#unsetormatches). The spec notes when certain things are case-insensitive (e.g. `errorContains`), but we should assume case matters otherwise.

jyemin: Done

jmikola (approved):

divjotarora (approved):
I haven't put up a Go PR for the v1.3 unified test runner changes, so this test might've failed in Go once I got there. Changes LGTM

PR #967, opened 2021-04-21
Change serviceId to hasServiceId in expected events

divjotarora:
The unified test format changes for DRIVERS-720 introduced a ``hasServiceId`` assertion in expected events to assert whether or not the event published by the driver contains a ``serviceId`` field. The test files use this field correctly but it was incorrectly called ``serviceId`` in the spec rather than ``hasServiceId``.


jmikola (approved):
No need for a changelog entry or version bump IMO, since this is a retcon and drivers have yet to implement anything.

PR #966, opened 2021-04-20
DRIVERS-1685: Add transaction unpinning test using a read operation

jyemin:

PR #968, opened 2021-04-21
DRIVERS-1655 Unpin sessions after all abortTransaction attempts

divjotarora:


source/transactions/transactions.rst
```diff
@@ -810,26 +810,30 @@ certain retries of commitTransaction and abortTransaction) to the same mongos.
 When to unpin
 ^^^^^^^^^^^^^
 
-Drivers MUST unpin a ClientSession when a command within a transaction,
-including commitTransaction and abortTransaction, fails with a
-TransientTransactionError. Transient errors indicate that the transaction
-in question has already been aborted or that the pinned mongos is
+Drivers MUST unpin a ClientSession in the following situations:
+
+#. The ``abortTransaction`` command is executed. The session MUST be unpinned
```
divjotarora, line 815: This is the only new criteria for unpinning that was added in this PR. All of the other changes in this file are to restructure the existing criteria in list form.

source/unified-test-format/tests/Makefile
```diff
@@ -1,8 +1,8 @@
 SCHEMA=../schema-1.3.json
 
-.PHONY: all invalid valid-fail valid-pass versioned-api load-balancers gridfs HAS_AJV
+.PHONY: all invalid valid-fail valid-pass versioned-api load-balancers gridfs transcations HAS_AJV
```
divjotarora, line 3: I realized that this was missed when the `mongos-unpin.yml` file was first added.

source/unified-test-format/tests/Makefile
```diff
@@ -23,6 +23,9 @@ load-balancers: HAS_AJV
 gridfs: HAS_AJV
 	@ajv test -s $(SCHEMA) -d "../../gridfs/tests/*.yml" --valid
 
+transactions: HAS_AJV
+	@ajv test -s $(SCHEMA) -d "../../transactions/tests/unified/*.yml" --valid
```
jmikola, line 27: I assume this was missed in https://github.com/mongodb/specifications/commit/4f9ca96588cf075a640d2332c0d5d3f59cb8df65 for [DRIVERS-1638](https://jira.mongodb.org/browse/DRIVERS-1638)?

source/transactions/transactions.rst
```diff
@@ -810,26 +810,30 @@ certain retries of commitTransaction and abortTransaction) to the same mongos.
...
+regardless of whether or the abort operation succeeds or fails. If the
+operation fails with a retryable error, the session MUST be unpinned before
+performing server selection for the retry.
```
jmikola, line 818: This language differs from what was introduced in [DRIVERS-1638](https://jira.mongodb.org/browse/DRIVERS-1638):

> Aborting a transaction or starting a new transaction on a pinned ClientSession MUST unpin the session.

Was it an intentional change to refer to execution of `abortTransaction` rather than the ClientSession method? I ran into this while implementing the libmongoc changes in https://github.com/mongodb/mongo-c-driver/pull/783/files#diff-035ed3213d4e0cc642c9c77d8a28c5a9f3a2531851f92d4a11d755bc1cf2bd9dR1269, as there are two cases where abort can do something:

 * `startTransaction` method was previously called and the transaction is in a starting state, but has yet to send anything to the server. The abort happens entirely client-side.
 * The transaction is in progress (i.e. commands have been sent to the server), so an `abortTransaction` command must be sent.

I opted to unpin in both cases, even though the first was likely a NOP since the session should already have been unpinned by a call to `startTransaction`.

jmikola (commented):
This is the third ticket/PR to come in with related changes:

 * [DRIVERS-1638](https://jira.mongodb.org/browse/DRIVERS-1638) created the first unified transactions test with some abort tests in #941
 * [DRIVERS-1685](https://jira.mongodb.org/browse/DRIVERS-1685) added a "read" test in #966
 * [DRIVERS-1655](https://jira.mongodb.org/browse/DRIVERS-1655) is adding more tests in this PR

Is there some way we can group the work so drivers can know to implement these all at once, or at least do so in a particular order? Or is it too late for that and the best we can do is ensure all of the issues are linked to one another? I'd also be curious to know if there's another ticket lingering that will bring more changes to this file.

source/unified-test-format/tests/Makefile
divjotarora: Yeah

source/transactions/transactions.rst
divjotarora: The first case you mentioned was not intentionally excluded - IMO the session should be unpinned in that case as well. I updated the wording to "The transaction is aborted" instead.

divjotarora (commented):
I realize there's been a bit of churn with this file, but it was not intentional. DRIVERS-1638 was the original ticket, DRIVERS-1685 added a new test because @jyemin noticed a testing gap for the Java driver when implementing the original ticket, and this ticket adds new criteria that hadn't been considered previously. I'm not sure how much churn could have been avoided.

jmikola (approved):

source/transactions/transactions.rst
```diff
@@ -810,26 +810,30 @@ certain retries of commitTransaction and abortTransaction) to the same mongos.
...
 ^^^^^^^^^^^^^
 
-Drivers MUST unpin a ClientSession when a command within a transaction,
-including commitTransaction and abortTransaction, fails with a
-TransientTransactionError. Transient errors indicate that the transaction
-in question has already been aborted or that the pinned mongos is
+Drivers MUST unpin a ClientSession in the following situations:
+
+#. The transaction is aborted. The session MUST be unpinned regardless of
+whether or the abort succeeds or fails. If the operation fails with a
```
ShaneHarvey, line 816: `the abort` -> `the abortTransaction command`.

source/transactions/transactions.rst
```diff
@@ -810,26 +810,30 @@ certain retries of commitTransaction and abortTransaction) to the same mongos.
 When to unpin
```
ShaneHarvey, line 810: Can you bump the transaction spec and add a changelog entry?

ShaneHarvey (commented):
Tests pass in Python.

source/transactions/transactions.rst
```diff
@@ -810,26 +810,30 @@ certain retries of commitTransaction and abortTransaction) to the same mongos.
...
 ^^^^^^^^^^^^^
 
-Drivers MUST unpin a ClientSession when a command within a transaction,
-including commitTransaction and abortTransaction, fails with a
-TransientTransactionError. Transient errors indicate that the transaction
-in question has already been aborted or that the pinned mongos is
+Drivers MUST unpin a ClientSession in the following situations:
+
+#. The transaction is aborted. The session MUST be unpinned regardless of
```
ShaneHarvey, line 815: The numbered list here does not render correctly: https://github.com/mongodb/specifications/blob/fc561d9216774fba026bf210f283a5881d09b373/source/transactions/transactions.rst#when-to-unpin

The issue is that items that span multiple lines need to be indented: https://docutils.sourceforge.io/docs/user/rst/quickref.html#enumerated-lists

source/transactions/transactions.rst
jmikola: This might have been intentional, as I [previously asked](https://github.com/mongodb/specifications/pull/968#discussion_r617933562) about the case where no command is executed (e.g. abort immediately after starting a transaction). Perhaps revise to say "regardless of whether the `abortTransaction` command succeeds or fails, or was executed at all"?

ShaneHarvey: Oh I see. The current wording or your suggestion SGTM in that case.

divjotarora: Revised.

source/transactions/transactions.rst
divjotarora: Done.

source/transactions/transactions.rst
divjotarora: Sorry, I meant to check the rendering when I first put up the PR but forgot.

ShaneHarvey (approved):

source/unified-test-format/tests/Makefile
iwysiu: transcations -> transactions

divjotarora: Done @iwysiu

PR #969, opened 2021-04-22
DRIVERS-1633 Require using "hello" command with versioned API

alcaeus:
The legacy hello command will not be part of API Version 1. This requires drivers to exclusively use "hello" when an API Version is declared (as any server that supports the Versioned API is also expected to support "hello"). In the case that users misconfigure their clients and declare an API Version when connecting to servers that don't support "hello", the server type is expected to be "unknown", eventually triggering errors.

Note that the "Remove oppressive language" project will change the usage of hello vs. legacy hello when not using the Versioned API.

I previously considered requiring using `OP_MSG` with the Versioned API, as all servers that would support "hello" also support `OP_MSG`. However, while building the implementation in libmongoc I realised that this may cause significant workload for drivers, so I removed the requirement. Instead, the handshake spec allows for using `OP_MSG` for the initial handshake using "hello" but does not require it. This requirement will be added in a separate step after the 5.0 release.

### DIFF
--- a/source/message/OP_MSG.rst
+++ b/source/message/OP_MSG.rst
@@ -10,9 +10,9 @@ OP_MSG
 :Informed: Bryan Reinero, Chris Hendel, drivers@
 :Status: Approved
 :Type: Standards
-:Last Modified: 2021-04-06
+:Last Modified: 2021-04-20
 :Minimum Server Version: 3.6
-:Version: 1.2
+:Version: 1.3
 
 
 
@@ -51,7 +51,9 @@ Usage
 -----
 
 ``OP_MSG`` is only available in MongoDB 3.6 (``maxWireVersion >= 6``) and later.
-MongoDB drivers MUST continue to perform the MongoDB Handshake using ``OP_QUERY``
+MongoDB drivers SHOULD perform the MongoDB handshake using ``OP_MSG`` if an API
+version was declared on the client, but MAY decide to use ``OP_QUERY``. If no
+API version was declared, drivers MUST perform the handshake using ``OP_QUERY``
 to determine if the node supports ``OP_MSG``.
 
 If the node supports ``OP_MSG``, any and all messages MUST use ``OP_MSG``,
@@ -616,6 +618,7 @@ Q & A
 Changelog
 =========
 
+- 2021-04-20 Suggest using OP_MSG for initial handshake when using versioned API
 - 2021-04-06 Updated to use hello and not writable primary
 - 2017-11-12 Specify read preferences for OP_MSG with direct connection
 - 2017-08-17 Added the ``User originating command`` section
--- a/source/message/OP_MSG.rst
+++ b/source/message/OP_MSG.rst
@@ -52,9 +52,12 @@ Usage
 
 ``OP_MSG`` is only available in MongoDB 3.6 (``maxWireVersion >= 6``) and later.
 MongoDB drivers SHOULD perform the MongoDB handshake using ``OP_MSG`` if an API
-version was declared on the client, but MAY decide to use ``OP_QUERY``. If no
-API version was declared, drivers MUST perform the handshake using ``OP_QUERY``
-to determine if the node supports ``OP_MSG``.
+version was declared on the client, but MAY decide to use ``OP_QUERY``.
+
+If no API version was declared, drivers that support MognoDB 3.4 and earlier
+MUST perform the handshake using ``OP_QUERY`` to determine if the node supports
+``OP_MSG``. Drivers that only support MongoDB 3.6 and newer MAY default to using
+``OP_MSG``.
 
 If the node supports ``OP_MSG``, any and all messages MUST use ``OP_MSG``,
 optionally compressed with ``OP_COMPRESSED``.
--- a/source/server-discovery-and-monitoring/tests/integration/connectTimeoutMS.json
+++ b/source/server-discovery-and-monitoring/tests/integration/connectTimeoutMS.json
@@ -42,7 +42,8 @@
               },
               "data": {
                 "failCommands": [
-                  "isMaster"
+                  "isMaster",
+                  "hello"
                 ],
                 "appName": "connectTimeoutMS=0",
                 "blockConnection": true,
--- a/source/versioned-api/versioned-api.rst
+++ b/source/versioned-api/versioned-api.rst
@@ -172,7 +172,7 @@ API versioning options if the user did not specify them.
 Handshake behavior
 ~~~~~~~~~~~~~~~~~~
 
-if an API version was declared, drivers MUST NOT use the legacy hello command
+If an API version was declared, drivers MUST NOT use the legacy hello command
 during the initial handshake or afterwards. Instead, drivers MUST use the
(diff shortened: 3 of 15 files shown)


source/message/OP_MSG.rst
```diff
@@ -51,7 +51,9 @@ Usage
 -----
 
 ``OP_MSG`` is only available in MongoDB 3.6 (``maxWireVersion >= 6``) and later.
-MongoDB drivers MUST continue to perform the MongoDB Handshake using ``OP_QUERY``
+MongoDB drivers SHOULD perform the MongoDB handshake using ``OP_MSG`` if an API
+version was declared on the client, but MAY decide to use ``OP_QUERY``. If no
+API version was declared, drivers MUST perform the handshake using ``OP_QUERY``
```
jyemin, line 56: In practice we are developing drivers (e.g. Rust) that only support 3.6+ servers.  Let's clarify here that OP_QUERY is only required for drivers that support 3.4-

source/message/OP_MSG.rst
```diff
@@ -170,20 +172,20 @@ and MUST ignore all undefined optional bits.
 
 The currently defined flags are:
 
-===== ==================== ========= ========== =========================== 
+===== ==================== ========= ========== ===========================
```
jyemin, line 175: Please exclude formatting changes.  If you still want to do them, make them a separate commit.

source/versioned-api/versioned-api.rst
```diff
@@ -169,6 +169,16 @@ the specified value is equal to the server default. Drivers MUST NOT add any
 API versioning options if the user did not specify them.
 
 
+Handshake behavior
+~~~~~~~~~~~~~~~~~~
+
+Since the legacy hello command is not part of the versioned API, drivers MUST
```
jyemin, line 175: It's not clear that this requirement only applies when an API version is declared.  Can we prefix with

> If an API version was declared

as in the previous paragraph?

jyemin (changes requested):

source/versioned-api/versioned-api.rst
ajdavis: I appreciate your motivation for calling isMaster "legacy hello", but I think "isMaster" is much clearer to future readers. Could you say "isMaster" here? Or if necessary, define "legacy hello" as "IsMaster" once before continuing to use "legacy hello".

ajdavis: Also, do we plan to update handshake.rst to replace isMaster with hello there? I see that we updated a bunch of specifications this morning 99c9ee188a4bc4fc095837ae0be2451c30761c9b

source/versioned-api/versioned-api.rst
```diff
@@ -169,6 +169,16 @@ the specified value is equal to the server default. Drivers MUST NOT add any
...
+NOT use this command during the initial handshake or afterwards. Instead,
+drivers MUST use the ``hello`` command exclusively. If the server does not
+support ``hello``, the server description MUST reflect this with an ``unknown``
```
ajdavis, line 178: Capital ``Unknown`` for consistency with SDAM.

ajdavis (changes requested):
Can you avoid the extraneous whitespace changes?

source/server-discovery-and-monitoring/tests/integration/connectTimeoutMS.json
```diff
@@ -42,7 +42,8 @@
               },
               "data": {
                 "failCommands": [
-                  "isMaster"
+                  "isMaster",
+                  "hello"
```
ShaneHarvey, line 46: I see we're updating these tests but haven't updated the SDAM spec itself yet. Is that going to happen in a different ticket?

source/message/OP_MSG.rst
alcaeus: Clarified the wording.

alcaeus: Clarified the wording.

source/message/OP_MSG.rst
alcaeus: Apologies, this came in due to my IDE truncating trailing spaces. I've extracted the changes to a separate commit.

source/versioned-api/versioned-api.rst
alcaeus: These will be changed as part of DRIVERS-1293.

source/versioned-api/versioned-api.rst
alcaeus: Changed. Thanks!

source/versioned-api/versioned-api.rst
alcaeus: I've asked @JamesKovacs for clarification. IIRC, we intended for one spec to define the term "legacy hello" so we can link to it from other specs. I'll update this once I know what to link to.

source/server-discovery-and-monitoring/tests/integration/connectTimeoutMS.json
alcaeus: I've added a change to cover the `isWritablePrimary` field. The SDAM spec will also see changes related to this change (in a separate DRIVERS ticket) to provide better error reporting in case the handshake command completes but yields an unexpected result (e.g. "hello" command not found when using Versioned API).

ajdavis (approved):
I approve! Assuming that the two remaining issues with this patch (extraneous whitespace changes, undefined term "legacy hello") are resolved now or later.

ShaneHarvey (changes requested):
Can you update the following spec tests to fail hello commands as well:
```
$ git grep -A5 failCommand '*.rst' | grep 'isMaster'
source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst:197:             failCommands: ["isMaster"],
source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst:234:             failCommands: ["isMaster"],
source/server-discovery-and-monitoring/tests/README.rst:427:               failCommands: ["isMaster"],
```

source/versioned-api/versioned-api.rst
alcaeus: See https://github.com/mongodb/specifications/pull/975.

ShaneHarvey (approved):

source/versioned-api/versioned-api.rst
```diff
@@ -169,6 +169,15 @@ the specified value is equal to the server default. Drivers MUST NOT add any
 API versioning options if the user did not specify them.
 
 
+Handshake behavior
+~~~~~~~~~~~~~~~~~~
+
+if an API version was declared, drivers MUST NOT use the legacy hello command
```
jyemin, line 175: Nit: Capitalize `if`

source/message/OP_MSG.rst
jyemin: Sorry, still seeing the whitespace.

jyemin (changes requested):

jyemin (changes requested):
I'm still seeing a lot of white space changes.  I know it's a bit nit-picky but there are enough that they obscure the substantive changes.  Can you revert them?

jyemin (approved):
LGTM

PR #970, opened 2021-04-23
DRIVERS-1703: CMAP integration test waits for wrong event

BorisDog:
DRIVERS-1703: Wait for ConnectionPoolCleared in pool-create-min-size-error test


divjotarora (approved):
The change LGTM but @patrickfreed will have a better idea of how this might affect other drivers because I don't have as much context regarding the original test, so please wait for his review.

patrickfreed (approved):

PR #973, opened 2021-04-26
DRIVERS-1293: Remove deprecated terms from transaction specs

JamesKovacs:


source/transactions/tests/legacy/mongos-recovery-token.yml
```diff
@@ -194,7 +194,7 @@ tests:
         result:
           insertedId: 1
       # Enable the fail point only on the Mongos that session0 is pinned to.
-      # Fail isMaster to prevent the heartbeat requested directly after the
+      # Fail isMaster/hello to prevent the heartbeat requested directly after the
```
durran, line 197: Should this be hello/legacy hello?

source/transactions/tests/legacy/pin-mongos.yml
```diff
@@ -461,7 +461,7 @@ tests:
         result:
           insertedId: 3
       # Enable the fail point only on the Mongos that session0 is pinned to.
-      # Fail isMaster to prevent the heartbeat requested directly after the
+      # Fail isMaster/hello to prevent the heartbeat requested directly after the
```
durran, line 464: hello/legacy hello

durran (changes requested):

source/transactions/tests/legacy/mongos-recovery-token.yml
JamesKovacs: Good point. Updated.

source/transactions/tests/legacy/pin-mongos.yml
JamesKovacs: Good point. Updated.

durran (approved):
LGTM

iwysiu (approved):

PR #974, opened 2021-04-27
DRIVERS-1293: Updated retryable reads/writes specs with modern terminology

JamesKovacs:


source/retryable-reads/retryable-reads.rst
```diff
@@ -677,6 +673,10 @@ degraded performance can simply disable ``retryableReads``.
 Changelog 
 ==========
 
+2021-04-26: Replaced deprecated terminology
+            Removed requirement to parse error message text as MongoDB 3.6+ servers
+            will always return an error code
```
jmikola, line 678: Formatting for this looks a bit off. See: https://github.com/mongodb/specifications/blob/cbe79a1fb120ec50c612deee12ff60663b1b24eb/source/retryable-reads/retryable-reads.rst#changelog

I think removing indentation for subsequent lines would fix it, as that's what we do in the retryable writes spec.

jmikola (approved):
One comment regarding changelog formatting for retryable reads, but LGTM. I don't need to review the correction.

source/retryable-reads/retryable-reads.rst
JamesKovacs: Fixed the formatting.

isabelatkinson (approved):

PR #975, opened 2021-04-27
DRIVERS-1293: Updated to define hello and legacy hello (aka isMaster)

JamesKovacs:
The intent is to define `isMaster` as legacy hello and use legacy hello in other specs rather than referring to `isMaster`.

### DIFF
--- a/source/mongodb-handshake/handshake.rst
+++ b/source/mongodb-handshake/handshake.rst
@@ -414,15 +414,15 @@ that can be formatted and value extracted from.
 Backwards Compatibility
 =======================
 
-The ``isMaster`` command currently ignores arguments. (i.e. If arguments are
-provided the ``isMaster`` command discards them without erroring out). Adding
+The legacy hello command currently ignores arguments. (i.e. If arguments are
+provided the legacy hello command discards them without erroring out). Adding
 client metadata functionality has therefore no backwards compatibility concerns.
 
 This also allows a driver to determine if the ``hello`` command is supported. On
-server versions that support the ``hello`` command, ``isMaster`` with
+server versions that support the ``hello`` command, the legacy hello command with
 ``helloOk: true`` will respond with ``helloOk: true``. On server versions that do
 not support the ``hello`` command, the ``helloOk: true`` argument is ignored and
-the ``isMaster`` response will not contain ``helloOk: true``.
+the legacy hello response will not contain ``helloOk: true``.
 
 Reference Implementation
 ========================
@@ -440,8 +440,8 @@ Q&A
    * Just the subdocument
 * Should I really try to fill the 512 bytes with data?
    * Not really. The server does not attempt to normalize or compress this data in anyway, so it will hold it in memory as-is per connection. 512 bytes for 20,000 connections is ~ 10mb of memory the server will need.
-* What happens if I pass this new ``isMaster`` argument to previous MongoDB versions?
-   * Nothing. Arguments passed to ``isMaster`` prior to MongoDB 3.4 are not treated in any special way and have no effect one way or other
+* What happens if I pass new arguments in the legacy hello command to previous MongoDB versions?
+   * Nothing. Arguments passed to the legacy hello command to prior versions of MongoDB are not treated in any special way and have no effect one way or another.
 * Are there wire version bumps or anything accompanying this specification?
    * No
 * Is establishing the handshake required for connecting to MongoDB 3.4?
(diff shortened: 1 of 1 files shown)


behackett (approved):

source/mongodb-handshake/handshake.rst
```diff
@@ -52,22 +53,43 @@ Specification
...
+monitoring. ``hello`` is the modern and preferred command. ``isMaster`` is referred
+to as "legacy hello" and is maintained for backwards compatibility with servers
+that do not support the ``hello`` command.
+
+NOTE: The legacy hello command name is ``isMaster``, but servers will accept the
+alternate casing ``ismaster``. Other case variations result in ``CommandNotFound``.
+Drivers MUST take this case variation into account when determining which commands
+to encrypt, redact, or otherwise treat specially.
+
+If Versioned API is requested, drivers MUST use the ``hello`` command for the
```
kevinAlbs, line 66: I suggest linking to the versioned API spec. Since drivers specify a `serverApi` class, I think the preferred term would be "server API version".

```suggestion
If a `server API version <../versioned-api/versioned-api.rst>`__ is requested, drivers MUST use the ``hello`` command for the
```

source/mongodb-handshake/handshake.rst
```diff
@@ -52,22 +53,43 @@ Specification
...
+initial handshake. If Versioned API is not requested, drivers MUST use legacy
+hello (``isMaster``) for the initial handshake and include ``helloOk: true``
+in the handshake request.
+
+ASIDE: If the legacy handshake response includes ``helloOk: true``, then
```
kevinAlbs, line 71: This seems less like an aside since this has a "MUST" statement.

```suggestion
If the legacy handshake response includes ``helloOk: true``, then
```

JamesKovacs: The reason I made it an aside is because this is really an SDAM issue - and will be dealt with in detail in the SDAM spec - but felt it should be mentioned in the handshake spec for context on why `helloOk:true` is necessary.

source/mongodb-handshake/handshake.rst
```diff
@@ -3,14 +3,14 @@ MongoDB Handshake
...
-:Spec Version: 1.2.0
+:Spec Version: 1.3.0
 :Author: Hannes Magnusson
 :Kernel Advisory: Mark Benvenuto
 :Driver Advisory: Anna Herlihy, Justin Lee
 :Status: Approved
 :Type: Standards
 :Minimum Server Version: 3.4
-:Last Modified: 2020-02-12
+:Last Modified: 2021-04-27
```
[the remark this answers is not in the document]
JamesKovacs: That's a clearer indication than mixing it into the specification section. I'll add the terms section as suggested.

I'm going to remove the part about hello being supported in 5.0+ because it was backported to late point releases of 3.6+. It is not usable for initial handshake until 5.0 (because we require server API version support to know that hello is supported), but can be used for heartbeats if helloOk is returned from isMaster. Because helloOk was only backported to 4.2 and 4.4, but backports declined to 3.6 and 4.0 (see https://jira.mongodb.org/browse/SERVER-50412), there is no way for drivers to use the hello command on 3.6 and 4.0 even if the patch release supports it.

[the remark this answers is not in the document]
alcaeus: Looking at the issues, I didn't find an indication why the backports to 3.6 and 4.0 were rejected. Do we have more info on that? I'm wondering if the wording change had anything to do with it, and whether we could get `helloOk` backported to 3.6 and 4.0 to complement the existence of the `hello` command.

alcaeus (approved):

source/mongodb-handshake/handshake.rst
kevinAlbs: That makes sense!

source/mongodb-handshake/handshake.rst
```diff
@@ -52,22 +69,40 @@ Specification
 Connection handshake
 --------------------
 
-The ``isMaster`` handshake MUST be performed on every socket to any and all servers
+MongoDB uses the ``hello`` or ``isMaster`` commands for handshakes and topology
+monitoring. ``hello`` is the modern and preferred command. ``isMaster`` is referred
+to as "legacy hello" and is maintained for backwards compatibility with servers
```
kevinAlbs, line 74: With the terminology introduced here, I'd vote for replacing ``isMaster`` with `legacy hello` in the "Backwards Compatibility" section and "Q&A".

source/mongodb-handshake/handshake.rst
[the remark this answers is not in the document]
kevinAlbs: The terms section looks great.

@alcaeus the backport tickets linked in [SERVER-50412](https://jira.mongodb.org/browse/SERVER-50412) includes rationale for why the server team chose not to.

kevinAlbs (approved):
LGTM with additional replacements in the "Backwards Compatibility" and "Q&A" sections!

source/mongodb-handshake/handshake.rst
JamesKovacs: Fair point. Done.

kmahar:
removed request for review from me, I think the others have probably given sufficient coverage and I don't want to hold this up further.

PR #976, opened 2021-04-30
DRIVERS-1293: Updated Server Selection spec to use modern terminology.

JamesKovacs:
Updated the terminology to use "not writable primary", "SecondaryOk", and similar in place of deprecated terms.

One notable change that may deserve some discussion. Since the `slaveOK` URI option is deprecated and no longer in `source/uri-options/uri-options.rst` I simply removed the section discussing it. Since this URI option has been deprecated for a long time, drivers should remove it from their connection string parsing logic in their next major release anyways. There is really no good way to change `slaveOk=true` into something else. Even if we rename the URI option, we will force customers using it to update their connection strings - and they should switch to using read preference anyways.

### DIFF
--- a/source/server-selection/server-selection.rst
+++ b/source/server-selection/server-selection.rst
@@ -9,8 +9,8 @@ Server Selection
 :Advisors: \A. Jesse Jiryu Davis, Samantha Ritter, Robert Stam, Jeff Yemin
 :Status: Accepted
 :Type: Standards
-:Last Modified: 2021-05-12
-:Version: 1.13.1
+:Last Modified: 2021-05-13
+:Version: 1.13.2
 
 .. contents::
 
@@ -167,10 +167,10 @@ Terms
     Abbreviation for "round trip time".
 
 **Round trip time**
-    The time in milliseconds to execute an ``ismaster`` command and
+    The time in milliseconds to execute a ``hello`` or legacy hello command and
     receive a response for a given server.  This spec differentiates between
-    the RTT of a single ``ismaster`` command and a server's *average* RTT over
-    several such commands.
+    the RTT of a single ``hello`` or legacy hello command and a server's *average*
+    RTT over several such commands.
 
 **Secondary**
     A server of type RSSecondary.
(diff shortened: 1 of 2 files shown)


source/server-selection/server-selection.rst
```diff
@@ -1157,7 +1152,7 @@ individual read operation.
 Reference Implementation
 ========================
 
-The single-threaded reference implementation is the Perl master branch (work
+The single-threaded reference implementation is the Perl main branch (work
```
p-mongo, line 1155: This branch does not exist in https://github.com/mongodb-labs/mongo-perl-driver/branches, where would one find it?

p-mongo (commented):
I'm afraid I don't follow the reasoning for removing portions of the specification. If they are no longer relevant, I think there should be respective drivers ticket(s) stating that the requirements being removed no longer apply. Otherwise I would think the sections should be retained.

source/server-selection/server-selection.rst
JamesKovacs: We are not renaming git branches yet. That's being left for a future project. I should probably have left this alone. I can revert this change.

JamesKovacs: Switched back from main to master.

p-mongo:
I'm afraid I don't follow the reasoning for removing portions of the specification. If they are no longer relevant, I think there should be respective drivers ticket(s) stating that the requirements being removed no longer apply. Otherwise I would think the sections should be retained.

JamesKovacs:
> I'm afraid I don't follow the reasoning for removing portions of the specification. If they are no longer relevant, I think there should be respective drivers ticket(s) stating that the requirements being removed no longer apply. Otherwise I would think the sections should be retained.

Fair point about creating a DRIVERS ticket for this removal. https://jira.mongodb.org/browse/DRIVERS-1738 created along with reasoning why it is safe to remove. Thanks for the feedback.

JamesKovacs:
@p-mongo: Separated out the slaveOk URI option changes into PR #992. I haven't reintroduced those changes here, but they'll go away once I rebase this PR on #992.

p-mongo:
Okay that works for me, if you prefer to remove those changes from this PR I'll approve this PR as it is.

JamesKovacs:
This PR has been rebased on the PR #992 changes and the commits squashed. Ready for final review by @p-mongo and @patrickfreed.

p-mongo (approved):

patrickfreed (approved):

PR #977, opened 2021-04-30
DRIVERS-1293: Added hello to the list of fail points where needed in …

JamesKovacs:
There were no changes required to the CMAP spec for deprecated language, but the failpoints in certain tests needed `hello` added to them. Without these additional failpoints, these tests will fail when drivers start using `hello` for heartbeats to supporting servers.

### DIFF
--- a/source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-is-enforced.json
+++ b/source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-is-enforced.json
@@ -14,7 +14,8 @@
     },
     "data": {
       "failCommands": [
-        "isMaster"
+        "isMaster",
+        "hello"
       ],
       "closeConnection": false,
       "blockConnection": true,
--- a/source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-is-enforced.yml
+++ b/source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-is-enforced.yml
@@ -10,7 +10,7 @@ failPoint:
   # high amount to ensure not interfered with by monitor checks.
   mode: { times: 50 }
   data:
-    failCommands: ["isMaster"]
+    failCommands: ["isMaster","hello"]
     closeConnection: false
     blockConnection: true
     blockTimeMS: 750
--- a/source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-timeout.json
+++ b/source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-timeout.json
@@ -14,7 +14,8 @@
     },
     "data": {
       "failCommands": [
-        "isMaster"
+        "isMaster",
+        "hello"
       ],
       "closeConnection": false,
       "blockConnection": true,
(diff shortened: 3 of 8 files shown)


source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-is-enforced.json
```diff
@@ -14,7 +14,8 @@
     },
     "data": {
       "failCommands": [
-        "isMaster"
+        "isMaster",
+        "hello"
```
kevinAlbs, line 18: Have these tests been POC'ed in a driver yet?

I wonder if the exact change will depend on the SDAM / server monitoring spec changes.

I was thinking the first handshake command will continue to be `isMaster` with `helloOK` to determine if the server supports `hello`. If so, I wonder if this is necessary to change.

alcaeus: That does not hold true when testing with a declared api version, in which case we'll use hello for the initial handshake.

JamesKovacs: These tests have not been POC'd in a driver yet. I'm going to do that when I start making these changes in the C# driver. I have validated the failpoint behaviour though. Given the following failpoint:

```
db.adminCommand({ configureFailPoint: "failCommand", mode: {times:3}, data: {errorCode: 42, failCommands: ["ping","hello"]}})
```

This will result in 3 failures total from either command. (I used `ping` and `hello` because the `mongo` shell sends `isMaster` for heartbeats.) This could be 3 pings or 3 hellos or a combination of the two. You will receive 3 failures and subsequent commands will succeed.

When drivers update their SDAM implementations to use `hello` for heartbeats (whether it be because server API versioning was specified or isMaster/helloOk was echoed), these tests will continue to pass because our intent in these tests is to fail a certain number of heartbeats, not test whether a specific heartbeat command was sent.

I should go back and add some tests to the Handshake and SDAM specs to make sure server API version and `isMaster`/`helloOk` both behave correctly and the correct commands and heartbeats are sent.

alcaeus (approved):

source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-is-enforced.json
kevinAlbs: OK, thank you for verifying the failpoint behavior. If you have not already, can you run the updated tests to verify they pass in the C# driver as-is?

I prefer waiting until a POC to merge test changes. (I've been bitten before when modifying tests without running them). But if we verify this does not break existing driver tests, and since this is a straightforward change, I think this is OK to merge. In the worst case, we can modify them again once the SDAM and handshake spec changes are POC'd.

kevinAlbs (approved):

source/connection-monitoring-and-pooling/tests/pool-checkout-maxConnecting-is-enforced.json
patrickfreed: I just ran these tests and they pass on the Rust driver (which doesn't use `hello` at all currently), so existing drivers are not broken by this change at least.

patrickfreed (approved):

PR #978, opened 2021-05-02
Set the `minServerVersion` requirement to 4.9 in isMaster-network/command-error tests

stIncMale:
The complete explanation is in DRIVERS-1722. In short, this change is related to the fact that servers <4.9 do not activate fail points configured with `appName`.

DRIVERS-1722

### DIFF
--- a/source/server-discovery-and-monitoring/tests/integration/isMaster-command-error.json
+++ b/source/server-discovery-and-monitoring/tests/integration/isMaster-command-error.json
@@ -1,7 +1,7 @@
 {
   "runOn": [
     {
-      "minServerVersion": "4.4"
+      "minServerVersion": "4.9"
     }
   ],
   "database_name": "sdam-tests",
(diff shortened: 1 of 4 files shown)


ShaneHarvey (approved):
LGTM but I think we should proceed as follows:
- Commit this change as something like "DRIVERS-1722 Skip isMaster-network/command-error tests on 4.4" to more clearly express the motivation.
- Open a new drivers ticket to revert this change once SERVER-49336 is backported and released to 4.4.

This way we won't forget to reenable these tests.

stIncMale:
Thanks, @ShaneHarvey, I filed DRIVERS-1725 (Enable isMaster-network-error, isMaster-command-error tests on servers before 4.9) and renamed DRIVERS-1722 to "Disable isMaster-network-error, isMaster-command-error tests on servers before 4.9".

stIncMale:
How do we usually go about such changes: do we wait for the second reviewer (Neal was added automatically as a code owner) to also approve or leave comments?

ShaneHarvey:
Thanks! We usually wait for the reviewer to either approve or recuse themselves but in this case the change is so small I think you're good to merge whenever.

PR #980, opened 2021-05-03
DRIVERS-1237: Remove driver restrictions for dollar-prefixed and dotted keys

jmikola:
https://jira.mongodb.org/browse/DRIVERS-1237

### DIFF
--- a/source/bson-corpus/tests/document.json
+++ b/source/bson-corpus/tests/document.json
@@ -17,6 +17,26 @@
             "description": "Single-character key subdoc",
             "canonical_bson": "160000000378000E0000000261000200000062000000",
             "canonical_extjson": "{\"x\" : {\"a\" : \"b\"}}"
+        },
+        {
+            "description": "Dollar-prefixed key in sub-document",
+            "canonical_bson": "170000000378000F000000022461000200000062000000",
+            "canonical_extjson": "{\"x\" : {\"$a\" : \"b\"}}"
+        },
+        {
+            "description": "Dollar as key in sub-document",
+            "canonical_bson": "160000000378000E0000000224000200000061000000",
+            "canonical_extjson": "{\"x\" : {\"$\" : \"a\"}}"
+        },
+        {
+            "description": "Dotted key in sub-document",
+            "canonical_bson": "180000000378001000000002612E62000200000063000000",
+            "canonical_extjson": "{\"x\" : {\"a.b\" : \"c\"}}"
+        },
+        {
+            "description": "Dot as key in sub-document",
+            "canonical_bson": "160000000378000E000000022E000200000061000000",
+            "canonical_extjson": "{\"x\" : {\".\" : \"a\"}}"
         }
     ],
     "decodeErrors": [
--- a/source/crud/tests/unified/bulkWrite-replaceOne-dots_and_dollars.yml
+++ b/source/crud/tests/unified/bulkWrite-replaceOne-dots_and_dollars.yml
@@ -1,0 +1,165 @@
+description: "bulkWrite-replaceOne-dots_and_dollars"
+
+schemaVersion: "1.0"
+
+createEntities:
+  - client:
+      id: &client0 client0
+      observeEvents: [ commandStartedEvent ]
+  - database:
+      id: &database0 database0
+      client: *client0
+      databaseName: &database0Name crud-tests
+  - collection:
+      id: &collection0 collection0
+      database: *database0
+      collectionName: &collection0Name coll0
+
+initialData: &initialData
+  - collectionName: *collection0Name
+    databaseName: *database0Name
+    documents:
+      - { _id: 1 }
+
+tests:
+  - description: "Replacing document with top-level dotted key on 3.6+ server"
+    runOnRequirements:
+      - minServerVersion: "3.6"
+    operations:
+      - name: bulkWrite
+        object: *collection0
+        arguments:
+          requests:
+            - replaceOne:
+                filter: { _id: 1 }
+                replacement: &dottedKey { _id: 1, a.b: 1 }
+        expectResult: &bulkWriteResult
+          deletedCount: 0
+          insertedCount: 0
+          insertedIds: { $$unsetOrMatches: { } }
+          matchedCount: 1
+          modifiedCount: 1
+          upsertedCount: 0
+          upsertedIds: { }
+    expectEvents: &expectEventsDottedKey
+      - client: *client0
+        events:
+          - commandStartedEvent:
+              command:
+                update: *collection0Name
+                updates:
+                  - q: { _id: 1 }
+                    u: *dottedKey
+                    multi: { $$unsetOrMatches: false }
+                    upsert: { $$unsetOrMatches: false }
+    outcome:
+      - collectionName: *collection0Name
+        databaseName: *database0Name
+        documents:
+          - *dottedKey
+
+  - description: "Replacing document with top-level dotted key on pre-3.6 server yields server-side error"
+    runOnRequirements:
+      - maxServerVersion: "3.4.99"
+    operations:
+      - name: bulkWrite
+        object: *collection0
+        arguments:
+          requests:
+            - replaceOne:
+                filter: { _id: 1 }
+                replacement: *dottedKey
+        expectError:
+          isClientError: false
+    expectEvents: *expectEventsDottedKey
+    outcome: *initialData
+
+  - description: "Replacing document with dollar-prefixed key in embedded doc on 5.0+ server"
+    runOnRequirements:
+      - minServerVersion: "5.0"
+    operations:
+      - name: bulkWrite
+        object: *collection0
+        arguments:
+          requests:
+            - replaceOne:
+                filter: { _id: 1 }
+                replacement: &dollarPrefixedKeyInEmbedded { _id: 1, a: { $b: 1 } }
+        expectResult: *bulkWriteResult
+    expectEvents: &expectEventsDollarPrefixedKeyInEmbedded
+      - client: *client0
+        events:
+          - commandStartedEvent:
+              command:
+                update: *collection0Name
+                updates:
+                  - q: { _id: 1 }
+                    u: *dollarPrefixedKeyInEmbedded
+                    multi: { $$unsetOrMatches: false }
+                    upsert: { $$unsetOrMatches: false }
+    outcome:
+      - collectionName: *collection0Name
+        databaseName: *database0Name
+        documents:
+          - *dollarPrefixedKeyInEmbedded
+
+  - description: "Replacing document with dollar-prefixed key in embedded doc on pre-5.0 server yields server-side error"
+    runOnRequirements:
+      - maxServerVersion: "4.99"
+    operations:
+      - name: bulkWrite
+        object: *collection0
+        arguments:
+          requests:
+            - replaceOne:
+                filter: { _id: 1 }
+                replacement: *dollarPrefixedKeyInEmbedded
+        expectError:
+          isClientError: false
+    expectEvents: *expectEventsDollarPrefixedKeyInEmbedded
+    outcome: *initialData
+
+  - description: "Replacing document with dotted key in embedded doc on 3.6+ server"
+    runOnRequirements:
+      - minServerVersion: "3.6"
+    operations:
+      - name: bulkWrite
+        object: *collection0
+        arguments:
+          requests:
+            - replaceOne:
+                filter: { _id: 1 }
+                replacement: &dottedKeyInEmbedded { _id: 1, a: { b.c: 1 } }
+        expectResult: *bulkWriteResult
+    expectEvents: &expectEventsDottedKeyInEmbedded
+      - client: *client0
+        events:
+          - commandStartedEvent:
+              command:
+                update: *collection0Name
+                updates:
+                  - q: { _id: 1 }
+                    u: *dottedKeyInEmbedded
+                    multi: { $$unsetOrMatches: false }
+                    upsert: { $$unsetOrMatches: false }
+    outcome:
+      - collectionName: *collection0Name
+        databaseName: *database0Name
+        documents:
+          - *dottedKeyInEmbedded
+
+  - description: "Replacing document with dotted key in embedded doc on pre-3.6 server yields server-side error"
+    runOnRequirements:
+      - maxServerVersion: "3.4.99"
+    operations:
+      - name: bulkWrite
+        object: *collection0
+        arguments:
+          requests:
+            - replaceOne:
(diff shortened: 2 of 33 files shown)


source/crud/tests/unified/bulkWrite-update-validation.yml
```diff
@@ -1,140 +1,73 @@
-# This file was created automatically using mongodb-spec-converter.
-# Please review the generated file, then remove this notice.
+description: "bulkWrite-update-validation"
+
+schemaVersion: "1.0"
```
jmikola, line 3: I'm not sure why the converted CRUD v2 tests all used "1.1" (perhaps it was just in the template), but these validate against the original schema so I've lowered the version. To quote the unified test format spec:

I came across this after restoring the older schemas for [DRIVERS-1754](https://jira.mongodb.org/browse/DRIVERS-1754).

source/crud/tests/unified/bulkWrite-update-validation.yml
```diff
@@ -1,140 +1,73 @@
...
-                  _id: 1
-                replacement:
-                  $set:
-                    x: 22
+            - replaceOne:
+                filter: { _id: 1 }
+                replacement: { $set: { x: 22 } }
         expectError:
-          isError: true
+          isClientError: true
```
jmikola, line 37: Switching from `isError` to `isClientError` is the only significant change for the `*-validation.yml` files.

source/crud/tests/unified/insertOne-dots_and_dollars.yml
```diff
@@ -0,0 +1,241 @@
...
+      - name: insertOne
+        object: *collection0
+        arguments:
+          document: *dollarPrefixedKey
+        expectError:
+          isClientError: false
+    expectEvents: *expectEventsDollarPrefixedKey
+    outcome: *initialData
+
+  - description: "Inserting document with top-level dotted key"
```
jmikola, line 68: The original scope talked about creating error-expecting tests for pre-5.0 servers, but I learned that older server versions are actually quite permissive in what they accept. Although [3.6 relaxed dot/dollar restrictions at the storage level](https://docs.mongodb.com/v3.6/reference/limits/#Restrictions-on-Field-Names), inserts on pre-3.6 server versions seem to have few restrictions.

When writing these tests, I ran various server versions (2.6, 3.0, 3.4, 3.6, 4.4, and 5.0) to determine boundaries for each test case and whether we should expect success/error. Note: 2.6 is no longer supported by libmongoc/PHPC, so I happened to use the shell in that case.

source/crud/tests/unified/insertOne-dots_and_dollars.yml
```diff
@@ -0,0 +1,241 @@
...
+                documents:
+                  - *dottedKeyInEmbedded
+    outcome:
+      - collectionName: *collection0Name
+        databaseName: *database0Name
+        documents:
+          - *dottedKeyInEmbedded
+
+  - description: "Inserting document with dollar-prefixed key in _id yields server-side error"
+    # Note: 5.0+ did not remove restrictions on dollar-prefixed keys in _id documents
```
jmikola, line 132: The scope originally expected this would work, but the functionality was removed in the final server implementation. The note here is just to clarify why we're not limiting this test to a particular server version.

source/crud/tests/unified/insertOne-dots_and_dollars.yml
```diff
@@ -0,0 +1,241 @@
...
+        arguments:
+          # Note: an incomplete DBRef document may cause issues loading the test
+          # file with an Extended JSON parser, since the presence of one DBRef
+          # key may cause the parser to require others and/or enforce expected
+          # types (e.g. $ref and $db must be strings).
+          #
+          # Using "$db" here works for libmongoc so long as it's a string type;
+          # however, neither $ref nor $id would be accepted on their own.
+          #
+          # See https://github.com/mongodb/specifications/blob/master/source/extended-json.rst#parsers
```
jmikola, line 201: To clarify, this limitation has nothing to do with the driver's insert functionality. It's a problem with using an Extended JSON parser to load the test file itself.

I realize this comment is pretty specific to libmongoc, but I thought it was helpful to explain why I couldn't use `$ref` or `$id`. I'm hoping this doesn't cause issues for other drivers, but if so we can consider removing this test entirely if so.

source/crud/tests/unified/insertOne-dots_and_dollars.yml
```diff
@@ -0,0 +1,241 @@
...
+      - collectionName: *collection0Name
+        databaseName: *database0Name
+        documents:
+          - *dbrefLikeKey
+
+  # Note: the description refers to "pre-5.0" purely for consistency with the
+  # related replaceOne test. Document validation on [3.6, 5.0) server versions
+  # is less strict for insert than it is for replacement documents so this test
+  # actually requires a pre 3.6-server.
+  - description: "Unacknowledged write using dollar-prefixed or dotted keys may be silently rejected on pre-5.0 server"
```
jmikola, line 222: The title here also aligns with the original goal from the scope:

> Document that unacknowledged writes using dollar-prefixed or dotted keys may be silently rejected by pre-5.0 servers.

Alternatively, I could swap the document for `*dollarPrefixedKeyInId` since that fails everywhere (pre- and post-5.0). In that case we could probably remove this comment entirely.

source/crud/tests/unified/replaceOne-dots_and_dollars.yml
```diff
@@ -0,0 +1,180 @@
...
+        object: *collection0
+        arguments:
+          filter: { _id: 1 }
+          replacement: *dottedKeyInEmbedded
+        expectError:
+          isClientError: false
+    expectEvents: *expectEventsDottedKeyInEmbedded
+    outcome: *initialData
+
+  - description: "Unacknowledged write using dollar-prefixed or dotted keys may be silently rejected on pre-5.0 server"
```
jmikola, line 157: This corresponds to the similarly-named test in `insertOne-dots_and_dollars.yml`.

source/crud/tests/unified/updateMany-dots_and_dollars.yml
```diff
@@ -0,0 +1,138 @@
...
+      collectionName: &collection0Name coll0
+
+initialData: &initialData
+  - collectionName: *collection0Name
+    databaseName: *database0Name
+    documents:
+      - { _id: 1, foo: {} }
+
+tests:
+  - description: "Updating document to set top-level dollar-prefixed key on 5.0+ server"
```
jmikola, line 25: Note that this file has no error-expecting tests, which was an original goal from the scope:

> Create CRUD spec tests for pre-5.0 servers that expect server-side errors when inserting, updating, or replacing documents with dollar-prefixed and dotted keys. These tests may complement the previous set of tests for 5.0+ servers, which do not expect an error.

While implementing update tests, I realized that update operations have historically had no dot/dollar validation since both characters are used throughout for field paths and operators. One could argue that the successful tests here also aren't doing much to test the driver, but I do think it's helpful to have some examples of the new server syntax handy (especially since this isn't yet documented). Also, successful tests for update _were_ an original goal from the scope.

source/crud/tests/unified/updateMany-dots_and_dollars.yml
```diff
@@ -0,0 +1,138 @@
...
+    expectEvents:
+      - client: *client0
+        events:
+          - commandStartedEvent:
+              command:
+                update: *collection0Name
+                updates:
+                  - q: { _id: 1 }
+                    u: *dollarPrefixedKey
+                    multi: true
```
jmikola, line 48: The updateMany tests (both here and for bulkWrite) don't actually try to update multiple documents at once. I didn't think that was important to test, since it really doesn't concern the driver beyond how we construct the outgoing command.

source/crud/tests/unified/findOneAndReplace-dots_and_dollars.yml
```diff
@@ -0,0 +1,140 @@
...
+        expectResult: *initialDocument
+    expectEvents: &expectEventsDollarPrefixedKeyInEmbedded
+      - client: *client0
+        events:
+          - commandStartedEvent:
+              command:
+                findAndModify: *collection0Name
+                query: { _id: 1 }
+                update: *dollarPrefixedKeyInEmbedded
+                new: { $$unsetOrMatches: false }
```
jmikola, line 82: These tests do not use `returnDocument`, which allows them to easily expect the initial document as a result.

According to the CRUD spec, the `new` option should be omitted if unspecified by the application (server defaults to false); however, I noticed that PHPLIB ran afoul of that (opened [PHPLIB-658](https://jira.mongodb.org/browse/PHPLIB-658) to track). I suppose other drivers might as well, so I'd like to be flexible here.

I reckon this is on par with what we're doing elsewhere in update tests for `upsert` and `multi` options (which have the same guidance in the CRUD spec).

source/driver-bulk-update.rst
```diff
@@ -13,6 +13,10 @@ Changes from previous versions
 
 Deprecated in favor of the *Driver CRUD API*.
 
+v0.8
```
jmikola, line 16: AFAIK, these version numbers have no significance but this was the closest thing to a changelog.

source/crud/tests/unified/bulkWrite-update-validation.yml
alcaeus: That is correct, I used schemaVersion 1.1 in those tests. I recall there being a reason for this but can't remember, so if 1.0 works as well that's fine

alcaeus:
I took a look through the tests, and at first glance they look sane. I noticed today that 5.0.0-alpha0-506 started appearing in builds which caused build failures in libmongoc (see mongodb/mongo-c-driver#795). This should also allow us to POC the changes in libmongoc and verify that the tests work as expected.

source/crud/tests/unified/bulkWrite-update-validation.yml
jmikola: Will be handled by https://github.com/mongodb/specifications/pull/1000 for [DRIVERS-1775](https://jira.mongodb.org/browse/DRIVERS-1775).

source/bson-corpus/tests/document.json
```diff
@@ -17,6 +17,26 @@
...
+        },
+        {
+            "description": "Dollar-prefixed key in sub-document",
+            "canonical_bson": "170000000378000F000000022461000200000062000000",
+            "canonical_extjson": "{\"x\" : {\"$a\" : \"b\"}}"
+        },
+        {
+            "description": "Dollar as key in sub-document",
+            "canonical_bson": "160000000378000E0000000224000200000061000000",
+            "canonical_extjson": "{\"x\" : {\"$\" : \"a\"}}"
```
benjirewis, line 29: Curious here and elsewhere, do we also need to test allowing dots and dollars in sub-arrays? Some drivers may have different logical paths for parsing sub-arrays vs. embedded documents.

source/crud/tests/unified/insertOne-dots_and_dollars.yml
benjirewis: The explicit failure of a dollar-prefixed key in ID might show this silent rejection better and allow you to remove the comment, but I don't have a strong preference.

source/crud/tests/unified/updateMany-dots_and_dollars.yml
benjirewis: Tests as examples/documentation (as opposed to actual invocations of expected driver behavior) seem good to me.

source/driver-bulk-update.rst
```diff
@@ -895,13 +899,7 @@ Test Case 2:
...
         Same for initializeOrderedBulkOp().    
 
-Test Case 3:
-    Key validation, no $-prefixed keys allowed:
-        batch = initializeUnorderedBulkOp().insert({$key: 1})
-        
-        bulk.execute() throws reasonable error (server does the validation)
-        
-        Same for initializeOrderedBulkOp().
+Test Case 3: *Removed*
```
benjirewis, line 902: Why not just remove the test case entirely?

benjirewis (commented):
Excellent attention to detail... 🧑‍🔧  just a few questions.

source/crud/tests/unified/bulkWrite-replaceOne-dots_and_dollars.yml
```diff
@@ -0,0 +1,165 @@
...
+                    upsert: { $$unsetOrMatches: false }
+    outcome:
+      - collectionName: *collection0Name
+        databaseName: *database0Name
+        documents:
+          - *dottedKey
+
+  - description: "Replacing document with top-level dotted key on pre-3.6 server yields server-side error"
+    runOnRequirements:
+      - maxServerVersion: "3.4.99"
```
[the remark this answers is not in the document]
jmikola: This is intentional. 3.5.x is the development release series for what ultimately becomes 3.6.x, so I'm allowing this to run on any stable release series before that. In practice, this makes little difference since a driver is unlikely to ever test on development versions before the latest release (e.g. 4.3.x became irrelevant as soon as 4.4.0 was tagged).

source/bson-corpus/tests/document.json
jmikola: Do you mean as the field in a BSON array itself, or an embedded document within a BSON array?

In the case of the field in a BSON array, those are typically assigned as numeric strings starting at `"0"`, but there is really no enforcement of that. When iterating a BSON array (i.e. type 0x04), the keys should be ignored. Such a test would likely require we use `degenerate_bson` (e.g. ["Single Element Array with index set incorrectly to ab"](https://github.com/mongodb/specifications/blob/master/source/bson-corpus/tests/array.json#L23)).

As for an embedded document _within_ a BSON array, I'm not sure why any driver would have a different code path for that vs. an embedded document anywhere else. If anyone is aware of such a code path, I'm happy to add an extra test case; otherwise, I don't think it's necessary.

source/driver-bulk-update.rst
jmikola: This follows what was done in [Change Streams: Prose Tests](https://github.com/mongodb/specifications/tree/master/source/change-streams/tests#prose-tests). See https://github.com/mongodb/specifications/commit/66c6e9177709230483e914ec7773b5a996c89554 for an explanation.

source/bson-corpus/tests/document.json
benjirewis: That's true; it would have to be an embedded document within a BSON array, and drivers shouldn't have different paths for those vs. regular embedded documents. Probably not necessary, then.

benjirewis (approved):
LGTM after merge conflicts !

alcaeus (approved):

PR #981, opened 2021-05-04
Fix a race condition in pool-checkout-no-idle, pool-checkout-no-stale tests

stIncMale:
https://jira.mongodb.org/browse/DRIVERS-1724

### DIFF
--- a/source/connection-monitoring-and-pooling/tests/pool-checkout-no-idle.json
+++ b/source/connection-monitoring-and-pooling/tests/pool-checkout-no-idle.json
@@ -23,6 +23,16 @@
     },
     {
       "name": "checkOut"
+    },
+    {
+      "name": "waitForEvent",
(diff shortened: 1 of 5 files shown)


stIncMale:
This [failure](https://github.com/mongodb/specifications/pull/981/checks?check_run_id=2504259982) is weird. The file `pool-create-min-size-error.json` is not touched by this PR, and yet it causes build failures. The error reported is correct, `js-yaml` indeed spits out
```json
"closeConnection": true,
"appName": "poolCreateMinSizeErrorTest"
```
which is different from how these fields are ordered in the `master` branch
```json
"appName": "poolCreateMinSizeErrorTest"
"closeConnection": true
```.

I guess have to also fix `pool-create-min-size-error.json` in this PR.

patrickfreed (approved):

PR #984, opened 2021-05-05
DRIVERS-1720 Require passing Versioned API options to getMore and transaction-continuing commands

alcaeus:
DRIVERS-1720

The server already allows passing Versioned API options to transaction-continuing commands (see https://github.com/mongodb/mongo/commit/4aa27885874b90e098c1225fccb10f4daa3b3d38), and ``getMore`` will get the same treatment soon. The server will require these options no later than RC0, so we can already make the spec change.

Note that drivers implementing this will have to wait until the latest build includes the commit above as well as the commit to change ``getMore`` behaviour.

### DIFF
--- a/source/versioned-api/tests/crud-api-version-1-strict.yml
+++ b/source/versioned-api/tests/crud-api-version-1-strict.yml
@@ -244,7 +244,7 @@ tests:
                   - $group: { _id: 1, n: { $sum: $count }}
                 <<: *expectedApiVersion
 
-  - description: "find command with declared API version appends to the command, but getMore does not"
+  - description: "find and getMore append API version"
     operations:
       - name: find
         object: *collection
--- a/source/versioned-api/tests/crud-api-version-1-strict.yml
+++ b/source/versioned-api/tests/crud-api-version-1-strict.yml
@@ -268,9 +268,7 @@ tests:
           - commandStartedEvent:
               command:
                 getMore: { $$type: [ int, long ] }
-                apiVersion: { $$exists: false }
-                apiStrict: { $$exists: false }
-                apiDeprecationErrors: { $$exists: false }
+                <<: *expectedApiVersion
 
   - description: "findOneAndDelete appends declared API version"
     operations:
--- a/source/versioned-api/tests/crud-api-version-1.yml
+++ b/source/versioned-api/tests/crud-api-version-1.yml
@@ -262,9 +262,7 @@ tests:
           - commandStartedEvent:
               command:
                 getMore: { $$type: [ int, long ] }
-                apiVersion: { $$exists: false }
-                apiStrict: { $$exists: false }
-                apiDeprecationErrors: { $$exists: false }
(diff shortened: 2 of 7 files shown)


jyemin (approved):
LGTM! Let's hold off on merging until we can test the changes against a server that supports the new behavior.

source/versioned-api/versioned-api.rst
```diff
@@ -181,20 +184,18 @@ server description MUST reflect this with an ``Unknown`` server type.
...
 
 
 Transactions
 ~~~~~~~~~~~~
 
-When running commands as part of a transaction, drivers MUST NOT send API
-parameters after the initial command that includes the ``startTransaction``
-option.
+When running commands as part of a transaction, drivers MUST send API parameters
+with all commands that are part of a transaction. If the API parameters for a
```
ajdavis, line 196: Mention "including commitTransaction and abortTransaction"?

ajdavis (approved):
I approve, tiny suggestion that you are free to ignore.

source/versioned-api/versioned-api.rst
alcaeus: Good point, added 👍

source/versioned-api/tests/crud-api-version-1.yml
```diff
@@ -262,9 +262,7 @@ tests:
           - commandStartedEvent:
               command:
                 getMore: { $$type: [ int, long ] }
-                apiVersion: { $$exists: false }
-                apiStrict: { $$exists: false }
-                apiDeprecationErrors: { $$exists: false }
+                <<: *expectedApiVersion
```
benjirewis, line 265: Same comment about the test name here.

source/versioned-api/tests/transaction-handling.yml
```diff
@@ -48,7 +44,7 @@ initialData:
       - { _id: 5, x: 55 }
 
 tests:
-  - description: "Only the first command in a transaction declares an API version"
+  - description: "All commands in a transaction declares an API version"
```
benjirewis, line 47:
```suggestion
  - description: "All commands in a transaction declare an API version"
```

source/versioned-api/tests/crud-api-version-1-strict.yml
```diff
@@ -268,9 +268,7 @@ tests:
           - commandStartedEvent:
               command:
                 getMore: { $$type: [ int, long ] }
-                apiVersion: { $$exists: false }
-                apiStrict: { $$exists: false }
-                apiDeprecationErrors: { $$exists: false }
+                <<: *expectedApiVersion
```
benjirewis, line 271: Should probably change the name of this test to reflect that `getMore` _does_ append API version

ShaneHarvey:
LGTM, this passes on ubuntu-18.04 and amazon1-2018 in evg for python. I haven't tested other distros which might be lagging behind on their server-latest builds.

PR #986, opened 2021-05-05
DRIVERS-1730 clarify OCSP stapling tests are not supported on all Linux platforms

kevinAlbs:
Not urgent.

### DIFF
--- a/source/ocsp-support/tests/README.rst
+++ b/source/ocsp-support/tests/README.rst
@@ -211,9 +211,10 @@ duration even after a driver encounters a TLS error early).
 Testing on Linux
 ----------------
 
-Drivers MUST test OCSP stapling behavior on a Linux platform other than Ubuntu
-18.04. `SERVER-51364 <https://jira.mongodb.org/browse/SERVER-51364>`__ disables
-OCSP stapling on the server for Ubuntu 18.04.
+Drivers MUST test on Linux platforms that have server support for OCSP stapling.
+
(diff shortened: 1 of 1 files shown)


source/ocsp-support/tests/README.rst
```diff
@@ -235,6 +243,8 @@ to simplify the testing procedure.
 Changelog
 ==========
 
+**2021-11-??**: Clarify that not all Linux platforms support server stapling.
```
kevinAlbs, line 246:
```suggestion
**2021-11-08**: Clarify that not all Linux platforms support server stapling.
```

source/ocsp-support/tests/README.rst
```diff
@@ -208,6 +208,14 @@ Drivers may wish to use a smaller value for ``serverSelectionTimeoutMS`` to
 speed up tests (otherwise server selection will spin for the entire
 duration even after a driver encounters a TLS error early).
 
+Testing on Linux
+----------------
+
+Drivers MUST test on Linux platforms that have server support for OCSP stapling.
```
benjirewis, line 214: Should we suggest what linux platforms drivers should test against? Go currently tests against RHEL 7.0 (+ MacOS and Windows).

kevinAlbs: Good point. I added a note to consider using Ubuntu 20.04 (C uses `ubuntu2004-small`) or RHEL 7.

benjirewis (approved):

PR #987, opened 2021-05-06
DRIVERS-1603 Update converted crud-v2 tests for serverless testing

patrickfreed:
DRIVERS-1603

This covers the crud-v2 tests which were recently converted to the unified format. It also fixes an existing aggregate test which had wrong assertions.


source/crud/tests/unified/aggregate.json
```diff
@@ -42,10 +42,6 @@
         {
           "_id": 3,
           "x": 33
-        },
```
patrickfreed, line 45: The cursor actually needs to do another getmore after the final document in the result set, so we only need two documents to generate the expected getmore events.

jmikola: For the record, this test was only introduced in https://github.com/mongodb/specifications/commit/3586db8b1ed90d835737f6226bb3ff9012cfc5de#diff-665043dea9fd9951f2063ea5827d4e888cfdfe44e7deda6201f688e63f9dadc0 for #964 so most drivers didn't catch this yet?

I just want to make sure we don't need to look into any drivers that were incorrectly passing the previous test by inadvertently ignoring extra observed events.

jmikola (approved):
One question but LGTM if the answer is as expected.

source/crud/tests/unified/aggregate.json
patrickfreed: I don't think any drivers have pulled in that test yet since it's also part of DRIVERS-1603, which does not yet have unblocked language tickets.

PR #988, opened 2021-05-07
Rework GitHub reviews for specifications

alcaeus:
This PR changes the CODEOWNERS file that is currently used to assign reviews to spec owners. Instead of assigning reviewers in the CODEOWNERS file directly, these are now assigned using GitHub Teams. The teams are nested in a spec reviewers team, so it's easy to get an overview: https://github.com/orgs/mongodb/teams/dbx-spec-reviewers/teams.
I've split the work into individual commits to make it easier to track changes.

This has multiple advantages. From GitHub's point of view, the review of each code owner is requested (and optionally required). For specs with two owners this would mean that both owners would have to approve a PR, which is not always what we want. Teams work around this, as there's a single code owner (the team), and each team member can approve the PR on the team's behalf. Furthermore, teams allow to configure review assignments to request a review from a team member directly instead of putting the request on the team.

The most important improvement is the experience around unowned specifications. Authors would usually assign their lead to review the PR. The new "dbx-spec-reviewers-unowned" team takes over this role and distributes the reviews around the leads. The list of people available for review can be expanded to other interested engineers as well. Essentially, this means that engineers won't have to think about who should review a PR. In some cases, they will still want to add their own reviewers, but the code ownership setup ensures that there will always be a pre-selected reviewer.


kmahar (approved):

.github/CODEOWNERS
```diff
@@ -0,0 +1,47 @@
...
+/source/transactions-convenient-api/ @mongodb/dbx-spec-owners-transactions
+
+# Unified Test Format
+/source/unified-test-format/ @mongodb/dbx-spec-owners-unified-test-format
+
+# URI Options
+/source/uri-options/ @mongodb/dbx-spec-owners-uri-options
+
+# Workflows and code ownership file
+/.github/ @mongodb/dbx-leadership
```
kevinAlbs, line 47: > Furthermore, teams allow to configure review assignments to request a review from a team member directly instead of putting the request on the team.

That's good, everyone in dbx-leadership seems like too much for reviewing changes in the workflows.

kevinAlbs (approved):
LGTM!

> The most important improvement is the experience around unowned specifications. Authors would usually assign their lead to review the PR. The new "dbx-spec-reviewers-unowned" team takes over this role and distributes the reviews around the leads.

That is awesome.

behackett (approved):

.github/CODEOWNERS
alcaeus: Even then, only a single lead needs to approve the PR on behalf of the team, but it creates noise for all leads.

PR #989, opened 2021-05-10
DRIVERS-1739 Fix broken aggregation test

patrickfreed:
DRIVERS-1739

This PR fixes the "aggregate with multiple batches works" test in `unified/aggregate.yml` test by updating it to match the "find with multiple batches works" test in `unified/find.yml`, hopefully fixing this for good. The latter uses an odd number of result documents with a batch size of 2, which should make it so that regardless of whether the server requires an extra round trip at the end of the results, the test should still work. I've tested in on 3.6+ standalones and it seems to work on all those versions.

### DIFF
--- a/source/crud/tests/unified/aggregate.json
+++ b/source/crud/tests/unified/aggregate.json
@@ -42,6 +42,18 @@
         {
           "_id": 3,
           "x": 33
+        },
+        {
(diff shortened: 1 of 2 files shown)


jyemin (approved):
Looks reasonable.

PR #990, opened 2021-05-11
DRIVERS-1293: Updated SDAM spec and tests to use modern terminology.

JamesKovacs:
The following are the major updates made to the SDAM spec and tests:

- Updated specs to use `hello` and legacy hello instead of `isMaster`
- Updated specs and tests to use `isWritablePrimary` instead of `ismaster`.
- Created a `legacy-hello` directory and made copies of all `ismaster`-related tests using the same directory structure.
- Removed master-slave tests as master-slave clusters pre-date MongoDB 2.6.
- For pre-3.6 support, I left in error message checking related to "not master" in the error text.

Note that the updated spec tests will fail until DRIVERS-1633 and DRIVERS-1750 have been implemented.


source/server-discovery-and-monitoring/tests/errors/non-stale-topologyVersion-greater-NotMaster.json
```diff
@@ -1,5 +1,5 @@
 {
-  "description": "Non-stale topologyVersion greater NotMaster error",
+  "description": "Non-stale topologyVersion greater NotWritablePrimary error",
```
ShaneHarvey, line 2: Many of these files still have "Master" in the name.

JamesKovacs: Good point. I'll check and update the filenames.

JamesKovacs: Pushed filename changes to the PR branch. I've left the filenames as-is in `legacy-hello`.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1347,15 +1342,15 @@ recovering" error::
...
         if code:
-            if code in notMasterCodes:
+            if code in notWritablePrimaryCodes:
               return true
         else:
           # if no code, use the error message.
           if isRecovering(message, None):
               return false
-          return ("not master" in message)
+          return ("not writable primary" in message)
```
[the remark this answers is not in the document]
JamesKovacs: I went through `server-discovery-and-monitoring.rst` to ensure that "not writable primary" error refers to the error but that when parsing error strings "not master" is used. Changes pushed to PR branch.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -2512,6 +2507,7 @@ mark the server Unknown and clear the pool.
```
kevinAlbs, line 2507: Should there be a change log entry?

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -78,34 +76,40 @@ set tags. In SDAM terms, a significant topology change on the server means the
...
+monitor uses the `Streaming Protocol`_ to continuously stream hello or legacy hello
 responses from the server. In MongoDB <= 4.2, a monitor uses the
 `Polling Protocol`_ pausing heartbeatFrequencyMS between `checks`_.
 Clients check servers sooner in response to certain events.
 
+If a `server API version`_ is requested, then the driver must use hello for monitoring.
+If a server API version is not requested, the initial handshake using the legacy hello
+command must include `helloOk: true`. If the response contains `helloOk: true`, then the
+driver must use the `hello` command for monitoring. If the response does not contain
+`helloOk: true`, then the driver must use the legacy hello command for monitoring.
```
kevinAlbs, line 111: Nice, does this section cover the spec changes for DRIVERS-1750?

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -657,6 +662,12 @@ The event API here is assumed to be like the standard `Python Event
         heartbeatFrequencyMS = heartbeatFrequencyMS
         minHeartbeatFrequencyMS = 500
 
+        # Heartbeat command
+        if versionedApi or handshakeResponse.helloOk:
```
kevinAlbs, line 666: There won't be a `handshakeResponse` available in the monitor's constructor since the first legacy hello hasn't been called yet. And if the server is marked unknown, I think we'll want to send a legacy hello command again with `helloOk` (in case it was downgraded).

Would it make sense to store a bool on the ServerDescription to indicate whether the server can accept a `hello` command? Something like `ServerDescription.helloOk`?

A response to a legacy hello would set it to true if the response includes `helloOk`. And any response to `hello` would set it to true regardless.

Then, the logic here would follow suit with how `topologyVersion` is handled.

cc @ShaneHarvey in case you'd like to weigh in.

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -772,8 +784,8 @@ RTT thread
 ``````````
 
 The requirements in the `Measuring RTT`_ section can be satisfied with an
-addtional thread that periodically runs the isMaster command on a dedicated
-connection, for example:
+addtional thread that periodically runs the hello or legacy hello command
```
kevinAlbs, line 787:
```suggestion
additional thread that periodically runs the hello or legacy hello command
```

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -783,6 +795,13 @@ connection, for example:
         serverAddress = serverAddress
         connectTimeoutMS = connectTimeoutMS
         heartbeatFrequencyMS = heartbeatFrequencyMS
+
+        # Heartbeat command
+        if versionedApi or handshakeResponse.helloOk:
```
kevinAlbs, line 800: Similar comment here about `handshakeResponse`. Since the `RttMonitor` is independent. To go along with the suggestion about storing a `ServerDescription.helloOk`, the `RttMonitor` may need to start storing a `ServerDescription` to track whether it can use `hello`, and reset that on an exception in `run`.

source/server-discovery-and-monitoring/tests/README.rst
```diff
@@ -73,7 +73,7 @@ An "applicationError" object has the following keys:
   - "timeout": A network timeout error.
 
 - response: (optional) A command error response, for example
-  ``{ok: 0, errmsg: "not master"}``. Present if and only if ``type`` is
+  ``{ok: 0, errmsg: "not writable primary"}``. Present if and only if ``type`` is
```
kevinAlbs, line 76: Should this remain as-is? I may be misremembering. Does the server replace this error message with "not writable primary" if the client connects performs the handshake with `hello`?

source/server-discovery-and-monitoring/tests/README.rst
```diff
@@ -417,14 +417,14 @@ Run the following test(s) on MongoDB 4.4+.
...
-       which should add extra latency to each RTT check::
+    #. Configure the following failpoint to block hello or legacy hello commands
+       for 250ms which should add extra latency to each RTT check::
 
          db.adminCommand({
              configureFailPoint: "failCommand",
              mode: {times: 1000},
              data: {
-               failCommands: ["isMaster", "hello"],
+               failCommands: ["hello"], // or the legacy hello command
```
kevinAlbs, line 427: Oh good thinking.

kevinAlbs (commented):
Looking good!

> Note that the updated spec tests will fail until DRIVERS-1633 and DRIVERS-1750 have been implemented.

OK, are the spec changes of DRIVERS-1750 covered in this PR? Otherwise, I would suggest we hold off on merging until we a POC can verify the new tests pass.

---

The scope lists a goal to remove tests of master/slave direct_connection_slave.*. Those can be removed instead of moved.

---

I did not review the tests yet. Are any tests that have changed in a significant way (aside from the error code change and moving)?

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -1087,11 +1107,11 @@ server's RTT by listening to TopologyDescriptionChangedEvents.
...
-ServerHeartbeatSucceededEvents published from awaitable isMaster
+ServerHeartbeatSucceededEvents published from awaitable hello or legacy hello
 responses will regularly have 10 second durations. The spec introduces
 the "awaited" field on server heartbeat events so that applications can
 differentiate a slow heartbeat in the polling protocol from a normal
-awaitable isMaster heartbeat in the new protocol.
+awaitable hello or legacy hello heartbeat in the new protocol.
 
 
 Changelog
```
kevinAlbs, line 1117: Should this have a changelog entry?

comandeo:
I copied SDAM specs from this PR to Ruby driver - https://github.com/mongodb/mongo-ruby-driver/pull/2271
Everything looks good and green 👍

p-mongo:
Can the test files be organized in such a way that tests that test the same functionality are on the same level please.

Specifically, https://github.com/JamesKovacs/specifications/tree/drivers1293/sdam/source/server-discovery-and-monitoring/tests and https://github.com/JamesKovacs/specifications/tree/drivers1293/sdam/source/server-discovery-and-monitoring/tests/legacy-hello are in some ways peers and in other ways not. The files that are in https://github.com/JamesKovacs/specifications/tree/drivers1293/sdam/source/server-discovery-and-monitoring/tests that are equivalent to the files in https://github.com/JamesKovacs/specifications/tree/drivers1293/sdam/source/server-discovery-and-monitoring/tests/legacy-hello should be moved in such a way as to be peers of https://github.com/JamesKovacs/specifications/tree/drivers1293/sdam/source/server-discovery-and-monitoring/tests/legacy-hello in the directory structure. This would also likely create a good label for this group of tests which presently appears to be missing.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst
```diff
@@ -427,6 +427,7 @@ See the `README <https://github.com/mongodb/specifications/server-discovery-and-
 Changelog
 =========
 
+- 30 APR 2021: Updated to use modern terminology.
```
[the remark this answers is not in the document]
JamesKovacs: Fixed.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-summary.rst
```diff
@@ -74,9 +74,6 @@ and monitor their availability and round trip time.
 This spec defines how mongoses are discovered and monitored,
 but does not define which mongos is selected for a given operation.
 
-A client MUST be able to directly connect to a mongod begun with "--slave".
-No additional master-slave features are described in this spec.
```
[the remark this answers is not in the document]
JamesKovacs: @kevinAlbs: The change is to delete those lines. Are you suggesting that we re-add them but refer to it as "legacy replication"?

source/server-discovery-and-monitoring/server-discovery-and-monitoring-summary.rst
```diff
@@ -207,16 +204,16 @@ MUST abort and raise an exception if it was a write.
...
+The ``hello`` or legacy hello call is retried once, immediately,
 before the server is marked "down".
 
 In either case the client MUST clear its connection pool for the server:
 if one socket is bad, it is likely that all are.
 
-An algorithm is specified for parsing
-"not master" and "node is recovering" errors.
+An algorithm is specified for inspecting error codes (MongoDB 3.6+) and
+falling back to parsing "not master" and "node is recovering" errors (MongoDB 3.4 and earlier).
```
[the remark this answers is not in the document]
JamesKovacs: 👍

source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst
```diff
@@ -33,29 +33,29 @@ Outcome: The former primary's ServerType MUST become Unknown.
...
+-----------------------------------------------------------------
 
 Scenario: With TopologyType ReplicaSetWithPrimary, we read from a server we
-thought was RSPrimary. Thus the slaveOk bit is not set.
+thought was RSPrimary. Thus the SecondaryOk bit is not set.
 
 The response's QueryFailure bit is set and the response document is:
 
-    {$err: "not master and slaveOk=false"}
+    {$err: "not writable primary and SecondaryOk=false"}
```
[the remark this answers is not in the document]
JamesKovacs: Changed to prose description of error as discussed in Slack.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-tests.rst
```diff
@@ -33,29 +33,29 @@ Outcome: The former primary's ServerType MUST become Unknown.
 The TopologyType MUST change to ReplicaSetNoPrimary.
 The client MUST NOT immediately re-check the former primary.
 
-"Not master" error when reading without SlaveOkay bit
------------------------------------------------------
+"Not writable primary" error when reading without SecondaryOk bit
```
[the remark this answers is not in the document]
JamesKovacs: Do you think that the test title should be changed to something else? Or is it fine the way it is?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
JamesKovacs: Added change log entry.

source/server-discovery-and-monitoring/server-monitoring.rst
JamesKovacs: Yes, this is the spec change to go along with DRIVERS-1750.

source/server-discovery-and-monitoring/server-monitoring.rst
JamesKovacs: Valid point. I've reworked the logic a bit to add `ServerDescription.helloOk` as this is a sensible place to hang this per-server information. The one drawback is that it adds a field to `ServerDescription`, which is a breaking - though additive - change.

source/server-discovery-and-monitoring/server-monitoring.rst
JamesKovacs: Fixed.

source/server-discovery-and-monitoring/server-monitoring.rst
JamesKovacs: Thinking this through a bit, I believe we should leave these threads independent from each other. My suggestion is to initialize the `RttMonitor` with `helloOk` based on `apiVersion` - same as the main monitoring thread. The `RttMonitor` can then examine the response to determine if subsequent pings should use hello or legacy hello. This way the `RttMonitor` does not need access to the `ServerDescription` and we don't have to worry about synchronization between the main monitor thread attempting to update the `RttMonitor` with the current state of `helloOk`.

source/server-discovery-and-monitoring/tests/README.rst
JamesKovacs: As soon as you run `hello` on a connection, the error message changes from `not master` to `not primary`. The error code remains 10107 and `NotWritablePrimary` regardless. Only `errmsg` changes based on whether `hello` has been called on that connection.

I'll fix the text to make mention of this fact and correct error message text.

source/server-discovery-and-monitoring/server-monitoring.rst
JamesKovacs: Added.

JamesKovacs:
> Can the test files be organized in such a way that tests that test the same functionality are on the same level please.
> 
> Specifically, https://github.com/JamesKovacs/specifications/tree/drivers1293/sdam/source/server-discovery-and-monitoring/tests and https://github.com/JamesKovacs/specifications/tree/drivers1293/sdam/source/server-discovery-and-monitoring/tests/legacy-hello are in some ways peers and in other ways not. The files that are in https://github.com/JamesKovacs/specifications/tree/drivers1293/sdam/source/server-discovery-and-monitoring/tests that are equivalent to the files in https://github.com/JamesKovacs/specifications/tree/drivers1293/sdam/source/server-discovery-and-monitoring/tests/legacy-hello should be moved in such a way as to be peers of https://github.com/JamesKovacs/specifications/tree/drivers1293/sdam/source/server-discovery-and-monitoring/tests/legacy-hello in the directory structure. This would also likely create a good label for this group of tests which presently appears to be missing.

The reason for the current organization is to facilitate eventual deprecation and removal of the `legacy-hello` functionality. I can re-arrange the folder structure if desired. What do other people think? Re-arrange to place `hello` and related `legacy-hello` tests side-by-side or keep the current structure for easier removal of `legacy-hello`?

JamesKovacs:
Incorporated feedback. Note that how `helloOk` is managed in the monitor and rttMonitor threads has been reworked based on discussions with @kevinAlbs.

ShaneHarvey:
> Re-arrange to place hello and related legacy-hello tests side-by-side or keep the current structure for easier removal of legacy-hello?

Can you explain why we need two copies of the tests at all? I'd except we'd need to have two copies of only a specific subset of tests where one copy uses isMaster and only runs on <5.0 and the other uses hello and only runs on >=5.0.

JamesKovacs:
> > Re-arrange to place hello and related legacy-hello tests side-by-side or keep the current structure for easier removal of legacy-hello?
> 
> Can you explain why we need two copies of the tests at all? I'd except we'd need to have two copies of only a specific subset of tests where one copy uses isMaster and only runs on <5.0 and the other uses hello and only runs on >=5.0.

If you diff `tests/` with `tests/legacy-hello/`, you will notice that only tests that required changes (for example the response containing `isWritablePrimary` rather than `ismaster`) were duplicated and modified. (The folder structure was maintained between the two.)

Regarding running isMaster on <5.0 and hello >=5.0, unfortunately things aren't that simple for two reasons:

1. MongoDB 4.2 and 4.4 both support hello in patch releases and we should test this.

ASIDE: Actually hello was backported all the way to 3.6 and 4.0, but helloOk was not - which is what allows drivers to detect support for hello without failing a command. It was decided that drivers would only use hello on servers supporting helloOk and server declined the backport requests of helloOk to 3.6 and 4.0.

2. MongoDB 5.0 still supports `isMaster` and without server API versioning configured even modern drivers have to perform initial negotiation with `isMaster/helloOk`.

If there is a better way to organize the tests or if you see a flaw in my reasoning, definitely happy to incorporate the feedback. Thanks for taking the time to review these changes.

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -734,11 +734,15 @@ The event API here is assumed to be like the standard `Python Event
...
+                response = call {helloCommand: 1, helloOk: True, topologyVersion: previousDescription.topologyVersion, maxAwaitTimeMS: heartbeatFrequencyMS}
             else:
                 # The server does not support topologyVersion.
-                response = call {helloCommand: 1}
+                response = call {helloCommand: 1, helloOk: True}
 
+            # if server supports hello, then response.helloOk will be true
+            # and hello will be used for subsequent monitoring commands.
+            # if server does not support hello, then response.helloOk will be undefined
+            # and legacy hello will be used for subsequent monitoring commands.
```
kevinAlbs, line 745: To avoid adding `helloOk` to the ServerDescription:
```suggestion
            # and legacy hello will be used for subsequent monitoring commands.
            helloOk = response.helloOk
```

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -734,11 +734,15 @@ The event API here is assumed to be like the standard `Python Event
...
             return ServerDescription(response, rtt=rttMonitor.average())
         except Exception as exc:
```
kevinAlbs, line 747: To skip storing on the ServerDescription, reset `helloOk`:
```suggestion
        except Exception as exc:
        helloOk = versionedApi != null
```

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -662,17 +662,12 @@ The event API here is assumed to be like the standard `Python Event
...
-            helloCommand = legacy hello
-
         # Internal Monitor state:
         connection = Null
-        description = default ServerDescription
+        # server API versioning implies that the server supports hello
+        description = default ServerDescription(helloOk=versionedApi)
         lock = Mutex()
-        rttMonitor = RttMonitor(serverAddress)
+        rttMonitor = RttMonitor(serverAddress, helloOk=description.helloOk)
```
kevinAlbs, line 670:
```suggestion
        rttMonitor = RttMonitor(serverAddress, helloOk)
```

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -662,17 +662,12 @@ The event API here is assumed to be like the standard `Python Event
...
-        if versionedApi or handshakeResponse.helloOk:
-            helloCommand = hello
-        else
-            helloCommand = legacy hello
-
         # Internal Monitor state:
         connection = Null
-        description = default ServerDescription
+        # server API versioning implies that the server supports hello
+        description = default ServerDescription(helloOk=versionedApi)
```
kevinAlbs, line 668: To skip storing on the ServerDescription:
```suggestion
        description = default ServerDescription()
        helloOk = versionedApi != Null
```

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -723,7 +718,12 @@ The event API here is assumed to be like the standard `Python Event
         try:
             # The connection is null if this is the first check. It's closed if there was an error during the previous
             # check or the previous check was cancelled.
-            # NOTE: helloCommand is defined in __init__() above
+
+            if previousDescription.helloOk:
```
kevinAlbs, line 722: To skip storing on the ServerDescription:
```suggestion
            if helloOk:
```

source/server-discovery-and-monitoring/server-monitoring.rst
kevinAlbs: Good thinking.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-summary.rst
```diff
@@ -213,7 +213,7 @@ In either case the client MUST clear its connection pool for the server:
 if one socket is bad, it is likely that all are.
 
 An algorithm is specified for inspecting error codes (MongoDB 3.6+) and
-falling back to parsing "not master" and "node is recovering" errors (MongoDB 3.4 and earlier).
+falling back to parsing errors when error codes are unavailable (MongoDB 3.4 and earlier).
```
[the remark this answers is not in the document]
JamesKovacs: Done.

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -734,11 +734,15 @@ The event API here is assumed to be like the standard `Python Event
...
                 if connectTimeoutMS != 0:
                     set connection timeout to connectTimeoutMS+heartbeatFrequencyMS
-                response = call {helloCommand: 1, topologyVersion: previousDescription.topologyVersion, maxAwaitTimeMS: heartbeatFrequencyMS}
+                response = call {helloCommand: 1, helloOk: True, topologyVersion: previousDescription.topologyVersion, maxAwaitTimeMS: heartbeatFrequencyMS}
             else:
                 # The server does not support topologyVersion.
-                response = call {helloCommand: 1}
+                response = call {helloCommand: 1, helloOk: True}
 
+            # if server supports hello, then response.helloOk will be true
```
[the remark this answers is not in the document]
JamesKovacs: Fixed.

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -734,11 +734,15 @@ The event API here is assumed to be like the standard `Python Event
...
+            # and hello will be used for subsequent monitoring commands.
+            # if server does not support hello, then response.helloOk will be undefined
```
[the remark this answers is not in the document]
JamesKovacs: Fixed.

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -662,17 +662,12 @@ The event API here is assumed to be like the standard `Python Event
...
-        # Heartbeat command
-        if versionedApi or handshakeResponse.helloOk:
-            helloCommand = hello
-        else
-            helloCommand = legacy hello
-
         # Internal Monitor state:
         connection = Null
-        description = default ServerDescription
+        # server API versioning implies that the server supports hello
```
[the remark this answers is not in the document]
JamesKovacs: Fixed.

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -796,12 +800,6 @@ on a dedicated connection, for example:
         connectTimeoutMS = connectTimeoutMS
         heartbeatFrequencyMS = heartbeatFrequencyMS
```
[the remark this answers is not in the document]
JamesKovacs: Good catch. Fixed.

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -844,10 +842,15 @@ on a dedicated connection, for example:
...
+            helloCommand = hello
+        else
+            helloCommand = legacy hello
+
         start = time()
-        # NOTE: helloCommand is defined in __init__() above
-        call {helloCommand: 1}
+        response = call {helloCommand: 1, helloOk: True}
         rtt = time() - start
+        helloOk = response.helloOk
```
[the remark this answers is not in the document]
JamesKovacs: Valid point and great catch. I've reworked the `RttMonitor` logic to take `versionedApi` as a ctor param so that it can be properly reset to its original state when an exception occurs. This should let us handle a server downgrade correctly. Before the `RttMonitor` could be stuck using `hello` even if the downgraded server no longer supported it.

source/server-discovery-and-monitoring/server-monitoring.rst
JamesKovacs: I had added `helloOk` to `ServerDescription`:
https://github.com/JamesKovacs/specifications/blob/drivers1293/sdam/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#serverdescription

But thinking about it more, I believe we can implement this using a `helloOk` field in `Monitor` and `RttMonitor` thus not requiring changes to the public API for `ServerDescription`. The C# driver exposes `ServerDescription` in our eventing API and I suspect many/all other drivers do so as well. So better not to modify the public API for `ServerDescription` if we don't need to.

source/server-discovery-and-monitoring/server-monitoring.rst
JamesKovacs: I added this logic to `setupConnection` so we would only have to do it in one place.

source/server-discovery-and-monitoring/server-monitoring.rst
JamesKovacs: Modified to pass in `versionedApi` instead so that `RttMonitor` could make its own independent decisions about `helloOk`.

source/server-discovery-and-monitoring/server-monitoring.rst
JamesKovacs: `helloOk` initialization/reset logic centralized in `setupConnection`.

source/server-discovery-and-monitoring/server-monitoring.rst
JamesKovacs: Fixed.

kevinAlbs (commented):
I looked over the test changes and used BeyondCompare to compare the directories:

- `source/server-discovery-and-monitoring/tests/legacy-hello`
- `source/server-discovery-and-monitoring/tests`

I only have the following minor comments.

1. The following files are missing in the `legacy-hello` tests:

    - `load-balanced/load-balanced.yml`
    - `rs/non-rs-member.yml`
    - `sharded/normalize_uri_case.yml`
    - `single/direct_connection_unavailable_seed.yml`
    - `single/discover_unavailable_seed.yml`

	I think that is OK, and going forward no new tests need to be added to the `legacy-hello` directory.

	Can you add a README.rst to the `legacy-hello` directory to note that:
	- The purpose of the directory is to test that drivers can process legacy hello responses on 4.0 and prior servers.
	- No new tests should be added to the `legacy-hello` directory.

2. Should we remove `./source/server-discovery-and-monitoring/tests/legacy-hello/single/direct_connection_slave.yml` to align with the scope goal of "Remove tests for the long-deprecated master/slave replication"?

3. (Optional) `generate-error-tests.py` is missing from `legacy-hello/errors`, but there are `*.template` files. Should we add it back in case we need to regenerate those tests? I suspect it is low risk we will, so I am OK with leaving it out.

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -708,6 +710,8 @@ The event API here is assumed to be like the standard `Python Event
         # Take the mutex to avoid a data race becauase this code writes to the connection field and a concurrent
         # cancelCheck call could be reading from it.
         with lock:
+            # Server API versioning implies that the server supports hello.
+            helloOk = versionedApi != null
```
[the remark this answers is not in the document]
JamesKovacs: Fixed.

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -661,13 +661,15 @@ The event API here is assumed to be like the standard `Python Event
...
         heartbeatFrequencyMS = heartbeatFrequencyMS
         minHeartbeatFrequencyMS = 500
+        versionedApi = versionedApi
 
         # Internal Monitor state:
         connection = Null
-        # server API versioning implies that the server supports hello
-        description = default ServerDescription(helloOk=versionedApi)
+        # Initialized/reset in setupConnection.
+        helloOk = False
```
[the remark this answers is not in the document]
JamesKovacs: I thought I could be clever and initialize it only once in `setUpConnection`. And there was an ambiguity in `RttMonitor.setUpConnection` where `helloCommand` would be used in "perform connection handshake" before it was initialized. I've fixed that issue too and gone with the more verbose but correct version that resets `helloOk = versionedApi != Null` where needed.

source/server-discovery-and-monitoring/tests/integration/minPoolSize-error.yml
```diff
@@ -16,7 +16,7 @@ tests:
       configureFailPoint: failCommand
       mode: { skip: 3 }
       data:
-        failCommands: ["isMaster", "hello"]
+        failCommands: ["hello"]
```
prashantmital, line 19: @JamesKovacs I think this should be `["hello", "isMaster"]` ?

JamesKovacs:
Based on a Slack discussion with Shane and Kevin, I'm going to rework the `tests/legacy-hello` tests. Basically if the test is setting up a mocked cluster in a known state, the test will use `isWritablePrimary: true` because how the cluster gets into the correct state is irrelevant. The test's purpose is to verify error handling or some other aspect. I will also ensure that there are tests that verify that drivers can talk to a cluster using `hello` or `isMaster/helloOk`.

source/server-discovery-and-monitoring/tests/integration/minPoolSize-error.yml
JamesKovacs: I'm going to review the integration tests tomorrow. They should include both `hello` and `isMaster` in the list of failpoints.

JamesKovacs:
Reworked the spec tests as discussed:

- Removed duplicated tests from `tests/legacy-hello`.
- Added `tests/single/standalone_using_legacy_hello.[yml|json]` to verify that legacy hello responses can still be parsed.
- Added `helloOk: true` to hello responses as this field will be included in actual server responses.

Feedback (as always) is much appreciated.

source/server-discovery-and-monitoring/tests/errors/error_handling_handshake.yml
```diff
@@ -5,6 +5,7 @@ phases:
   responses:
   - - a:27017
     - ok: 1
+      helloOk: true
```
kevinAlbs, line 8: That makes sense. Since `helloOk` is always sent in monitoring, new servers will always return this in monitoring responses.

Can you also update the `*.template` files in this directory and regenerate with generate-error-tests.py to ensure those are in sync?

source/server-discovery-and-monitoring/tests/errors/non-stale-network-timeout-error.yml
```diff
@@ -5,6 +5,7 @@ phases:
   responses:
   - - a:27017
     - ok: 1
+      helloOk: true
```
kevinAlbs, line 8: The following tests appear to be missing `helloOk` in some mocked responses:
   - `tests/monitoring/required_replica_set.yml`
   - `tests/monitoring/standalone_suppress_equal_description_changes.yml`
   - `tests/monitoring/required_replica_set.yml`
   - `tests/monitoring/standalone.yml`

Should those be updated?

Edit: I believe the rationale for _not_ updating those is because they mock a `maxWireVersion: 4`. No need to change if that is the case.

source/server-discovery-and-monitoring/tests/errors/error_handling_handshake.yml
JamesKovacs: Good catch. I updated the `yml` files and then regenerated the `json` files, but forgot about the `template` files.

source/server-discovery-and-monitoring/tests/errors/non-stale-network-timeout-error.yml
JamesKovacs: No, this was an accidental omission on my part. I had my working directory set incorrectly in vim, which resulted in my grep missing files in other folders. Fixing this now.

source/server-discovery-and-monitoring/tests/errors/error_handling_handshake.yml
JamesKovacs: Ran `generate-error-tests.py` and confirmed that everything is in sync.

ShaneHarvey (approved):
LGTM with 2 questions:
- Are there any tests which verify that the driver actually upgrades from "isMaster" with helloOk to "hello" on monitoring connections? If not we should open a follow up ticket to add them.
- Are drivers supposed to only send "helloOk:true" on the handshake or send it with all heartbeat commands?

kevinAlbs (approved):
LGTM!

source/server-discovery-and-monitoring/tests/integration/minPoolSize-error.json
```diff
@@ -17,7 +17,6 @@
         },
         "data": {
           "failCommands": [
-            "isMaster",
```
ShaneHarvey, line 20: This change needs to be reverted or else this test will fail.

ShaneHarvey (changes requested):
Found one issue in the Python implementation.

source/server-discovery-and-monitoring/tests/integration/minPoolSize-error.yml
prashantmital: Actually, it should just be `isMaster` since we want the handshake to fail.

ShaneHarvey: I think both hello and isMaster are needed because we can run these tests with apiVersion implicitly set by the test suite.

prashantmital: Ah yes, you are right... without `hello`, running with versioned API will cause this test to fail.

source/server-discovery-and-monitoring/tests/integration/minPoolSize-error.json
JamesKovacs: The integration tests should all have both `hello` and `isMaster` in the fail commands when needed. Looks like this one got lost during a merge. Fixed.

prashantmital (approved):

JamesKovacs:
> * Are there any tests which verify that the driver actually upgrades from "isMaster" with helloOk to "hello" on monitoring connections? If not we should open a follow up ticket to add them.

No, we don't have tests that verify this. Because the spec tests start at receiving a response to a handshake/monitoring command, I didn't see a quick or easy way to implement a test that asserted that a certain command was used for handshake and another command used for monitoring. Filed https://jira.mongodb.org/browse/DRIVERS-1824 so this can be considered.

> * Are drivers supposed to only send "helloOk:true" on the handshake or send it with all heartbeat commands?

Drivers should send `helloOk:true` on all handshake and heartbeat commands in order to handle node upgrade/downgrade more gracefully. This also keeps the various threads relatively independent of one another. The monitor and RTT monitor threads for each node can independently determine whether `hello` is supported to avoid locking of thread state around which command should be used for subsequent heartbeats.

ShaneHarvey (approved):
LGTM!

JamesKovacs:
Manually rebased and squashed into `master`.

PR #991, opened 2021-05-12
Update uri options test

jyemin:
Allow loadBalanced=true&directConnection=false

This test was missed in the original commit.

DRIVERS-1721

### DIFF
--- a/source/uri-options/tests/connection-options.json
+++ b/source/uri-options/tests/connection-options.json
@@ -181,6 +181,18 @@
       }
     },
     {
+      "description": "loadBalanced=true with directConnection=false",
+      "uri": "mongodb://example.com/?loadBalanced=true&directConnection=false",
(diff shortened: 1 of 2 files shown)


ShaneHarvey (approved):
LGTM! The old test fails with this error in python:
```
FAIL: test_test_uri_options_connection-options_loadBalanced=true_with_directConnection=false_causes_an_error (__main__.TestAllScenarios)
  File "test/test_uri_spec.py", line 84, in modified_test_scenario
    func(*args, **kwargs)
  File "test/test_uri_spec.py", line 109, in run_scenario
    self.assertEqual(
AssertionError: True != False : Expected error for test 'loadBalanced=true with directConnection=false causes an error'
```

The new test passes.

durran (approved):
LGTM. Thanks!

PR #995, opened 2021-05-19
DRIVERS-1438 Add tests for creating time-series collections

benjirewis:
DRIVERS-1438

Adds a test to verify that drivers can support the new `timeseries` option through `createCollection`.

Here's my [POC](https://github.com/benjirewis/mongo-go-driver/tree/timeSeries.1871) in Go, and [here](https://spruce.mongodb.com/version/60a53d8f3627e0207ab872aa/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) is that POC passing the new tests.

### DIFF
--- a/source/collection-management/tests/timeseries-collection.yml
+++ b/source/collection-management/tests/timeseries-collection.yml
@@ -55,3 +55,61 @@ tests:
                 create: *collection0Name
                 timeseries: *timeseries0
               databaseName: *database0Name
+
+  # Unlike regular collections, time-series collections allow duplicate ids.
+  - description: "insertMany with duplicate ids"
+    operations:
+      - name: dropCollection
+        object: *database0
+        arguments:
+          collection: *collection0Name
+      - name: createCollection
+        object: *database0
+        arguments:
+          collection: *collection0Name
+          timeseries: *timeseries0
+      - name: assertCollectionExists
+        object: testRunner
+        arguments:
+          databaseName: *database0Name
+          collectionName: *collection0Name
+      - name: insertMany
+        object: *collection0
+        arguments:
+          documents: &docs
+            - {
+                _id: 1,
+                time: {
+                  $date: {
+                    $numberLong: "1552949630482"
+                  }
+                }
+              }
+            - {
+                _id: 1,
+                time: {
+                  $date: {
+                    $numberLong: "1552949630483"
+                  }
+                }
+              }
+    expectEvents:
+      - client: *client0
+        events:
+          - commandStartedEvent:
+              command:
+                drop: *collection0Name
+              databaseName: *database0Name
+          - commandStartedEvent:
+              command:
+                create: *collection0Name
+                timeseries: *timeseries0
+              databaseName: *database0Name
+          - commandStartedEvent:
+              command:
+                insert: *collection0Name
+                documents: *docs
+    outcome:
+    - collectionName: *collection0Name
+      databaseName: *database0Name
+      documents: *docs
--- a/source/collection-management/tests/timeseries-collection.yml
+++ b/source/collection-management/tests/timeseries-collection.yml
@@ -3,11 +3,7 @@ description: "timeseries-collection"
 schemaVersion: "1.0"
 
 runOnRequirements:
-  - minServerVersion: "4.9"
-    serverParameters:
-      featureFlagTimeseriesCollection:
-        value: true
-        version: "5.0"
+  - minServerVersion: "5.0"
 
 createEntities:
   - client:
--- a/source/collection-management/tests/timeseries-collection.yml
+++ b/source/collection-management/tests/timeseries-collection.yml
@@ -49,11 +49,9 @@ tests:
           - commandStartedEvent:
               command:
                 drop: *collection0Name
-              commandName: drop
               databaseName: *database0Name
           - commandStartedEvent:
               command:
                 create: *collection0Name
                 timeseries: *timeseries0
-              commandName: create
               databaseName: *database0Name
--- a/source/collection-management/timeseries-collection.json
+++ b/source/collection-management/timeseries-collection.json
@@ -1,0 +1,110 @@
+{
+  "description": "timeseries-collection",
+  "schemaVersion": "1.0",
+  "runOnRequirements": [
+    {
+      "minServerVersion": "4.9",
+      "serverParameters": {
+        "featureFlagTimeseriesCollection": {
+          "value": true,
+          "version": "5.0"
+        }
+      }
+    }
+  ],
+  "createEntities": [
+    {
+      "client": {
+        "id": "client0",
+        "observeEvents": [
+          "commandStartedEvent"
+        ]
+      }
+    },
+    {
+      "database": {
+        "id": "database0",
+        "client": "client0",
+        "databaseName": "ts-tests"
+      }
+    },
+    {
+      "collection": {
+        "id": "collection0",
+        "database": "database0",
+        "collectionName": "test"
+      }
+    }
+  ],
+  "initialData": [
+    {
+      "collectionName": "test",
+      "databaseName": "ts-tests",
+      "documents": []
+    }
+  ],
+  "tests": [
+    {
+      "description": "createCollection with all options",
+      "operations": [
+        {
+          "name": "dropCollection",
+          "object": "database0",
+          "arguments": {
+            "collection": "test"
+          }
+        },
+        {
+          "name": "createCollection",
+          "object": "database0",
+          "arguments": {
+            "collection": "test",
+            "timeseries": {
+              "timeField": "time",
+              "metaField": "meta",
+              "expireAfterSeconds": 604800
+            }
+          }
+        },
+        {
+          "name": "assertCollectionExists",
+          "object": "testRunner",
+          "arguments": {
+            "databaseName": "ts-tests",
+            "collectionName": "test"
+          }
+        }
+      ],
+      "expectEvents": [
+        {
+          "client": "client0",
+          "events": [
+            {
+              "commandStartedEvent": {
+                "command": {
+                  "drop": "test"
+                },
+                "commandName": "drop",
+                "databaseName": "ts-tests"
+              }
+            },
+            {
+              "commandStartedEvent": {
+                "command": {
+                  "create": "test",
+                  "timeseries": {
+                    "timeField": "time",
+                    "metaField": "meta",
+                    "expireAfterSeconds": 604800
+                  }
+                },
+                "commandName": "create",
+                "databaseName": "ts-tests"
+              }
+            }
+          ]
+        }
+      ]
+    }
+  ]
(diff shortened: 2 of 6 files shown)


source/collection-management/timeseries-collection.json
```diff
@@ -0,0 +1,110 @@
...
+  "description": "timeseries-collection",
+  "schemaVersion": "1.0",
+  "runOnRequirements": [
+    {
+      "minServerVersion": "4.9",
+      "serverParameters": {
+        "featureFlagTimeseriesCollection": {
+          "value": true,
+          "version": "5.0"
+        }
```
benjirewis, line 11: This `featureFlag` seems to be required to create time-series collections. Unfortunately, this means these tests can only run against a config started with `--setParameter featureFlagTimeseriesCollection=true`, which may mean Evergreen changes for some drivers.

source/collection-management/timeseries-collection.json
```diff
@@ -0,0 +1,110 @@
...
+  "initialData": [
+    {
+      "collectionName": "test",
+      "databaseName": "ts-tests",
+      "documents": []
+    }
+  ],
+  "tests": [
+    {
+      "description": "createCollection with all options",
```
benjirewis, line 48: I've only added a single test to make sure drivers can create a collection with the `timeseries` option without error, and that the collection actually exists. 

I could add tests to expect errors from an empty `timeField` or an unsupported option (like `capped`) used in conjunction with `timeseries`, but the server will error in these cases. I could also add tests for inserting and finding on a `timeseries` collection, but from the drivers-side, these new collections function similarly to normal ones.

benjirewis (commented):
I don't _think_ other spec changes are required here. 

The `createCollection` helper is not documented in specifications. And, while drivers will need to update their unified test runners to handle the new `timeseries` option for the `createCollection` operation, `createCollection` is also not documented in the [unified test format spec](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst), so a version bump/change is probably not required.

source/collection-management/timeseries-collection.json
jmikola: I know the feature flag for dots and dollars ([DRIVERS-1237](https://jira.mongodb.org/browse/DRIVERS-1237)) was recently removed. Is it likely this feature flag will also be removed before RC0 is tagged next week? Or is this feature likely to sit behind a feature flag for all of the 5.0.x series?

source/collection-management/timeseries-collection.json
jmikola: Assuming the server strictly validates options for the `create` command and raises errors for unsupported options, I don't think we need an additional error test here.

source/collection-management/timeseries-collection.yml
```diff
@@ -0,0 +1,63 @@
+description: "timeseries-collection"
```
jmikola, line 1: I take it folks decided to punt on writing a collection management spec document for the time being. IMO, it still makes sense to keep these tests under a `tests/` sub-directory. Also, you can create a basic `tests/README.rst` file that explains the tests in this directory use the unified format.

Consider [`versioned-api/tests/README.rst`](https://github.com/mongodb/specifications/blob/master/source/versioned-api/tests/README.rst) as a model. That would also be a good place to note that a certain server parameter or feature flag might be required to enable time-series collections on the server.

source/collection-management/timeseries-collection.yml
```diff
@@ -0,0 +1,63 @@
...
+
+schemaVersion: "1.0"
+
+runOnRequirements:
+  - minServerVersion: "4.9"
```
jmikola, line 6: Something else that occurred to me. Even if this functionality is in the v4.9 branch, since we already have 5.0 releases I think it makes sense to just use 5.0 here.

source/collection-management/timeseries-collection.yml
```diff
@@ -0,0 +1,63 @@
...
+    serverParameters:
+      featureFlagTimeseriesCollection:
+        value: true
+        version: "5.0"
```
jmikola, line 10: I realize the schema only defines [`serverParameters`](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/schema-1.1.json#L58) as an object type, but this seems very different from existing uses of `serverParameter` in the Versioned API tests, where all keys in the object have a scalar value.

Does `featureFlagTimeseriesCollection` actually have an object as its value? I would expect it to be a boolean like other feature flags.

[`runOnRequirement.serverParameter`](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#runonrequirement) doesn't say much about _how_ to match parameter values other than using flexible numeric comparisons. If some parameters can have non-scalar values I think we need to revise that text (and maybe suggest using document comparison rules defined elsewhere in the spec) -- and if so I'd suggest creating a ticket to track that work. cc: @alcaeus

alcaeus: I had originally wanted to use the document matchers used elsewhere in the unified test format, but there were concerns around that. I believe equality matches on non-scalar values should still work, but we won't be able to use any matchers here.

jmikola: No objection on equality matches (and not supporting matching syntax), but we'd do well to use similar language to the `outcome` assertions:

> When comparing collection data, the rules in Evaluating Matches do not apply and the documents MUST match exactly; however, test runners MUST permit variations in document key order or otherwise normalize the documents before comparison.

yada yada JSON objects are unordered...

source/collection-management/timeseries-collection.json
benjirewis: Discussed offline: `featureFlagTimeseriesCollection` is now enabled by default on the latest server versions, so we don't need to check for this server parameter. I did bump the `minServerVersion` to `5.0`, though.

source/collection-management/timeseries-collection.yml
benjirewis: Yep, no full collection management spec for the time being. Created a `tests/` subdirectory with a small `README`. Since `featureFlagTimeseriesCollection` is now enabled by default, the `README` just specifies that these tests are in the unified test format and require schema version 1.0.

source/collection-management/timeseries-collection.yml
```diff
@@ -0,0 +1,63 @@
+description: "timeseries-collection"
+
+schemaVersion: "1.0"
+
+runOnRequirements:
+  - minServerVersion: "4.9"
+    serverParameters:
```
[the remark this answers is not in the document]
benjirewis: Since we no longer require `serverParameters`, keeping the `schemaVersion` at 1.0. 

Added validation for these tests in the Makefile.

source/collection-management/timeseries-collection.yml
benjirewis: Again, we no longer need to check this server parameter, but it would be good to update the spec language to clarify behavior with server parameters that are not just `true` or `false`.

jmikola: Will do by [DRIVERS-1765](https://jira.mongodb.org/browse/DRIVERS-1765).

source/collection-management/tests/timeseries-collection.yml
```diff
@@ -0,0 +1,59 @@
...
+        arguments:
+          databaseName: *database0Name
+          collectionName: *collection0Name
+    expectEvents:
+      - client: *client0
+        events:
+          - commandStartedEvent:
+              command:
+                drop: *collection0Name
+              commandName: drop
```
jmikola, line 52: FWIW, I don't think `commandName` is very useful for APM assertions. I'd be in favor of dropping this, as most of our other tests don't use it. I actually don't recall why it was originally introduced in the older test formats, and I neglected to answer that question when pulling it forward into the unified format.

That said, this _is_ relevant for the `runCommand` operation (same option name, entirely different context) to overcome JSON/YAML key order concerns.

jmikola (approved):
LGTM. Optional suggestion to remove `commandName` from the APM assertions.

source/collection-management/tests/timeseries-collection.yml
benjirewis: Ah interesting; seems like it `commandName` serves no purpose here. Removed the assertions.

source/collection-management/tests/timeseries-collection.yml
```diff
@@ -0,0 +1,57 @@
...
+        arguments:
+          collection: *collection0Name
+      - name: createCollection
+        object: *database0
+        arguments:
+          collection: *collection0Name
+          timeseries: &timeseries0
+            timeField: "time"
+            metaField: "meta"
+            expireAfterSeconds: 604800
```
kevinAlbs, line 40: Optional: When playing around with this, it surprised me is that `_id` no longer has a uniqueness constraint in a timeseries collection.

```
> db.runCommand({create: "timecoll", timeseries: {timeField: "t"}})
{
	"ok" : 1,
	"$clusterTime" : {
		"clusterTime" : Timestamp(1621606645, 2),
		"signature" : {
			"keyId" : NumberLong(0)
		}
	},
	"operationTime" : Timestamp(1621606645, 2)
}
> db.timecoll.insert({_id:1, t: new Date()})
WriteResult({ "nInserted" : 1 })
> db.timecoll.insert({_id:1, t: new Date()})
WriteResult({ "nInserted" : 1 })
> db.timecoll.find().pretty()
{ "t" : ISODate("2021-05-21T14:17:31.341Z"), "_id" : 1 }
{ "t" : ISODate("2021-05-21T14:17:31.728Z"), "_id" : 1 }
```

I do not think this impacts driver API or CRUD results, so it does not strictly warrant additional tests. But since this is the first time that assumption no longer holds, if it is straightforward, it might be worthwhile to have a copy of this test with an `InsertMany` and `Find` using two documents with a duplicate `_id`.

kevinAlbs (approved):
LGTM with an optional suggestion for adding an additional test with duplicate `_id`.

I was able to run this locally with the C driver test runner.

source/collection-management/tests/timeseries-collection.yml
```diff
@@ -0,0 +1,57 @@
+description: "timeseries-collection"
+
+schemaVersion: "1.0"
+
+runOnRequirements:
+  - minServerVersion: "5.0"
```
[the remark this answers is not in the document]
benjirewis: Yep! I believe DRIVERS-1237 will also add `minServerVersion: "5.0"` tests. So long as everyone implements 1591 we should be set.

source/collection-management/tests/timeseries-collection.yml
benjirewis: Ah good find, it does look like they support duplicate `_id` fields. I've added another test for `InsertMany` with duplicate `_id`s.

kevinAlbs (approved):
LGTM even more!

PR #996, opened 2021-05-21
DRIVERS-1173: DBRef spec

jmikola:
https://jira.mongodb.org/browse/DRIVERS-1173

### DIFF
--- a/source/dbref.rst
+++ b/source/dbref.rst
@@ -266,8 +266,9 @@ code paths as needed.
    c. ``{ "$db": "db0" }``
 
    Assert that each document is not decoded to a DBRef model. In the context of
-   implicit decoding, the document MUST be left as-is. In the context of
-   explicit decoding, the DBRef decoding method MUST raise an error.
+   implicit decoding, the document MUST be decoded like any other embedded
+   document. In the context of explicit decoding, the DBRef decoding method MUST
+   raise an error.
 
 5. Documents with invalid types for ``$ref`` or ``$db`` MUST NOT be decoded to a
    DBRef model. For each of the following:
@@ -276,8 +277,9 @@ code paths as needed.
    b. ``{ "$ref": "coll0", "$id": 1, "$db": 1 }``
 
    Assert that each document is not decoded to a DBRef model. In the context of
-   implicit decoding, the document MUST be left as-is. In the context of
-   explicit decoding, the DBRef decoding method MUST raise an error.
+   implicit decoding, the document MUST be decoded like any other embedded
+   document. In the context of explicit decoding, the DBRef decoding method MUST
+   raise an error.
 
 
 Encoding
--- a/source/dbref.rst
+++ b/source/dbref.rst
@@ -143,8 +143,8 @@ example of implicit decoding might be configuring the driver's BSON codec to
 automatically convert embedded documents that comply with the `DBRef Structure`_
 into a DBRef model.
 
-Drivers that provide implicit decoding MUST provide some way for applications to
-opt out and allow DBRefs to be decoded like any other embedded document.
+Drivers that provide implicit decoding SHOULD provide some way for applications
+to opt out and allow DBRefs to be decoded like any other embedded document.
 
 When decoding a BSON document to a DBRef model:
 
--- a/source/dbref.rst
+++ b/source/dbref.rst
@@ -30,9 +30,10 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
 "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
 interpreted as described in `RFC 2119 <https://www.ietf.org/rfc/rfc2119.txt>`__.
 
-This specification presents documents as Extended JSON. Although JSON fields are
-unordered, the order of fields presented herein should be considered pertinent.
-This is especially relevant for the `Test Plan`_.
+This specification presents documents as Extended JSON for readability and
+expressing special types (e.g. ObjectId). Although JSON fields are unordered,
+the order of fields presented herein should be considered pertinent. This is
+especially relevant for the `Test Plan`_.
 
 
 Specification
@@ -199,7 +200,9 @@ to drivers that provide a DBRef model class.
 
 The documents in these tests are presented as Extended JSON for readability;
 however, readers should consider the field order pertinent when translating to
-BSON (or their language equivalent).
+BSON (or their language equivalent). These tests are not intended to exercise a
+driver's Extended JSON parser. Implementations SHOULD construct the documents
+directly using native BSON types (e.g. Document, ObjectId).
 
 
 Decoding
--- a/source/dbref.rst
+++ b/source/dbref.rst
@@ -165,6 +165,13 @@ If a BSON document cannot be implicitly decoded to a DBRef model, it MUST be
 left as-is (like any other embedded document). If a BSON document cannot be
 explicitly decoded to a DBRef model, the driver MUST raise an error.
 
+Since DBRefs are a special type of embedded document, a DBRef model class used
+for decoding SHOULD inherit the class used to represent an embedded document
+(e.g. Hash in Ruby). This will allow applications to always expect an instance
+of a common class when decoding an embedded document (if desired) and should
+also support the requirement for DBRef models to provide access to any extra,
+optional fields.
+
 
 Encoding a DBRef model to a BSON document
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/source/dbref.rst
+++ b/source/dbref.rst
@@ -1,0 +1,349 @@
+==================
+Handling of DBRefs
+==================
+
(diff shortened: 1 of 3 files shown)


alcaeus:
LGTM pending additional tests

source/dbref.rst
```diff
@@ -0,0 +1,349 @@
...
+
+
+DBRef Structure
+---------------
+
+A DBRef is an embedded document with the following fields:
+
+- ``$ref``: required string field. Contains the name of the collection where
+  the referenced document resides. This MUST be the first field in the DBRef.
+- ``$id``: required field. Contians the value of the ``_id`` field of the
```
kmahar, line 49:
```suggestion
- ``$id``: required field. Contains the value of the ``_id`` field of the
```

source/dbref.rst
```diff
@@ -0,0 +1,349 @@
...
+constituent parts. If so:
+
+- Drivers MUST solicit a string value for ``$ref``.
+
+- Drivers MUST solicit an arbitary value for ``$id``. Drivers SHOULD NOT enforce
+  any restrictions on this value; however, this may be necessary if the driver
+  is unable to differentiate between certain BSON types (e.g. ``null``,
+  ``undefined``) and the parameter being unspecified.
+
+- Drivers SHOULD solicit an optional string value for ``$db``.
```
kmahar, line 127: why is this a SHOULD rather than a MUST?

source/dbref.rst
```diff
@@ -0,0 +1,349 @@
...
+
+- Drivers MUST require ``$ref`` and ``$db`` (if specfied) to be strings but
```
kmahar, line 129:
```suggestion
- Drivers MUST require ``$ref`` and ``$db`` (if specified) to be strings but
```

source/dbref.rst
```diff
@@ -0,0 +1,349 @@
...
+Decoding a BSON document to a DBRef model
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Drivers MAY support explicit and/or implicit decoding. An example of explicit
+decoding might be a DBRef model constructor that takes a BSON document. An
+example of implicit decoding might be configuring the driver's BSON codec to
+automatically convert embedded documents that comply with the `DBRef Structure`_
+into a DBRef model.
+
+Drivers that provide implicit decoding MUST provide some way for applications to
```
kmahar, line 145: why is this a requirement?

kmahar: oh, I see this mentioned in the rationale section, so users can avoid changing the order of fields?

kmahar (commented):
looks great! just a couple small questions and typo fixes.

source/dbref.rst
jmikola: It certainly helps address that, but I think it's also generally useful because DBRef really is just a convention and not a proper type that _needs_ a special class. I don't object to changing this to SHOULD if it's problematic as-is.

I'd be curious to know if this would require work for existing implementations, or if it's already supported. I suppose you can speak to Node, but that leaves C#, Java, Python, and Ruby.

cc: @ShaneHarvey, @p-mongo, @jyemin, @rstam

source/dbref.rst
jmikola: I was keeping this flexible in case some implementations didn't support this and didn't care to add it. I'm not sure how common cross-DB references (or even cross-DB applications) are in the wild, but I figured even if the model class didn't support this an application could always construct the embedded document on its own.

In my mind, `$ref` and `$id` are the essentials (MUST), `$db` was relevant but still optional (SHOULD), and the extra fields are nice-to-haves. This also only pertains to constructing a DBRef model from individual parameters. The decoding code paths MUST support all of these to ensure DBRefs can be round-tripped without data loss (excluding possible field re-ordering).

source/dbref.rst
```diff
@@ -0,0 +1,349 @@
...
+- Drivers MUST accept any BSON type for ``$id`` and MUST NOT enforce any
+  restrictions on its value.
+
+- Drivers MUST preserve extra, optional fields (beyond ``$ref``, ``$id``, and
+  ``$db``) and MUST provide some way to access those fields via the DBRef model.
+  For example, an accessor method that returns the original BSON document
+  (including ``$ref``, etc.) would fulfill this requirement.
+
+If a BSON document cannot be implicitly decoded to a DBRef model, it MUST be
+left as-is (like any other embedded document). If a BSON document cannot be
```
p-mongo, line 164: I think it would be more correct to say something like "if bson document cannot be implicitly decoded to a dbref it must be decoded to a mapping like an embedded document would", since every single bit of data in a bson document is decoded somehow in languages other than C.

source/dbref.rst
p-mongo: Let me think about this some more. There currently seems to be the following issue:

- The dbref class in Ruby isn't derived from Hash.
- When Hash.from_bson is called, it may return a dbref based on the contents of bson document.
- This makes it possible for Hash.from_bson to not simply return different types but to return an unexpected type out of hierarchy.

Given that this is existing behavior I think some thought is need on my part as to how this might best be addressed.

p-mongo: Are we anticipating a use case where, once dollars and dots are allowed in field names, applications may insert documents containing $ref and $id fields that they intend to simply be hashes, and not dbrefs? If yes it seems reasonable to require drivers to provide dbref-free decoding to support this use case.

Ruby would need to do some work to implement this since currently bson decoding isn't configurable, but there is precedent for similar functionality with extended json encoding (which can be canonical or relaxed) (https://www.rubydoc.info/github/mongodb/bson-ruby/BSON/Object#as_extended_json-instance_method) and I think it would be reasonable to implement something similar for bson decoding.

p-mongo:
In light of https://jira.mongodb.org/browse/RUBY-2639 I think it would be useful to have cross-driver guidance as to whether DBRef should be a subclass of Hash (or the language's mapping type).

source/dbref.rst
```diff
@@ -0,0 +1,349 @@
...
+  There are no inherent restrictions on extra field names; however, older server
+  versions may impose their own restrictions (e.g. no dots or dollars).
+
+DBRefs have no relation to the deprecated DBPointer BSON type (i.e. type 0x0C).
+
+
+Examples of Valid DBRefs
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following examples are all valid DBRefs:
```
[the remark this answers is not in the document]
jmikola: I was using Extended JSON here purely to represent the ObjectId value for `$oid`. I'm happy to clarify that if you think it's helpful. Otherwise I can just change it to `ObjectId("...")` if that would avoid any confusion.

source/dbref.rst
jmikola: My intention with "as-is" was to indicate it would be treated as an arbitrary embedded document. Would "MUST be decoded in the same manner as embedded document" work?

source/dbref.rst
jmikola: > Are we anticipating a use case where, once dollars and dots are allowed in field names, applications may insert documents containing $ref and $id fields that they intend to simply be hashes, and not dbrefs? If yes it seems reasonable to require drivers to provide dbref-free decoding to support this use case.

It's quite possible for applications to start doing that with a 5.0-compat driver, but I wouldn't really anticipate it as a common behavior.

Since the spec will also require DBRef models allow access to any extra fields (e.g. `foo` for `{ $ref: "coll", $id: 1, "foo": 2 }`), I'm not worried about that either.

My reasoning is just that a DBRef model isn't a necessity, so giving (power?) users the ability to opt out of it is a good idea. Can we settle on changing this to SHOULD so Ruby (and others if needed) can punt on this for the time being?

source/dbref.rst
jmikola: Changing to:

> In the context of implicit decoding, the document MUST be decoded like any other embedded document.

source/dbref.rst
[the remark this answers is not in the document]
jmikola: Added a clarification to META:

> This specification presents documents as Extended JSON for readability and expressing special types (e.g. ObjectId).

Also clarified in the Test Plan section that drivers should not use an Extended JSON parser but rather construct the corresponding BSON using their native types directly.

jmikola:
@alcaeus: FYI, additional corpus tests are added. They currently fail in libmongoc due to some JSON parser issues, but I'm moving on to resolve those next.

source/dbref.rst
p-mongo: SHOULD works for me also.

p-mongo (approved):

source/dbref.rst
jmikola: @p-mongo: Changed this in the latest commit.

kmahar: thanks for the explanation! given what you said and the conversation above the SHOULD makes sense to me. I will let @p-mongo resolve in case there's any further discussion to have here

source/dbref.rst
kmahar: that makes sense. thanks for the explanation!

kmahar (approved):

alcaeus (approved):

PR #998, opened 2021-05-21
DRIVERS-1724 Add `backgroundThreadIntervalMS` pool option to CMAP test specification

stIncMale:
Use the new option in `pool-checkout-no-idle.yml`, `pool-checkout-no-stale.yml` to disable background thread runs.
Use the new option in `pool-clear-min-size.yml` and `pool-create-min-size-error.yml` to specify 50 ms delay between runs.

DRIVERS-1724

### DIFF
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -736,24 +736,15 @@ A Pool SHOULD have a background Thread that is responsible for
 monitoring the state of all available `Connections <#connection>`_. This background
 thread SHOULD
 
--  Populate `Connections <#connection>`_ to ensure that the pool always satisfies **minPoolSize**
-    - The background thread SHOULD just go back to sleep instead of waiting for
-      pendingConnectionCount to become less than maxConnecting when satisfying
-      minPoolSize.
+-  Populate `Connections <#connection>`_ to ensure that the pool always satisfies minPoolSize
 -  Remove and close perished available `Connections <#connection>`_.
 
-Conceptually, the aforementioned activities are organized into sequential Background Thread Runs of unspecified finite duration.
-Each Run MUST do as much work as possible, for example
+Conceptually, the aforementioned activities are organized into sequential Background Thread Runs of unspecified duration.
+A Run MUST do as much work as readily available and then end instead of waiting for more work.
+For example, instead of waiting for pendingConnectionCount to become less than maxConnecting when satisfying minPoolSize,
+a Run MUST either proceed with the rest of its duties, e.g., closing available perished connections, or end.
 
--  if at the beginning of a Run a Connection Pool needed three more Connections to satisfy minPoolSize, then the Run MUST try to populate
-   the Connection Pool with three more connections instead of populating it with only one and ending,
-   unless it encounters a situation preventing it from continuing;
--  if at the beginning of a Run a Connection Pool had three perished available connections, the Run MUST try to close all of them
-   instead of closing only one and ending.
-
-The first Run MUST start as soon after a Connection Pool becomes `ready <#marking-a-connection-pool-as-ready>`__ as practically possible.
-Neither the minimum delay between Runs (the minimum interval between the end of a Run and the beginning of the next Run)
-nor the desired period of Runs (the desired interval between the beginning of a Run and the beginning of the next Run) is specified.
+Marking a pool as `ready <#marking-a-connection-pool-as-ready>`__ MUST start a new Run as soon as possible.
 The
 `Test Format and Runner Specification <https://github.com/mongodb/specifications/tree/master/source/connection-monitoring-and-pooling/tests>`__
 may change the specified restrictions or introduce new ones to facilitate testing.
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -417,7 +417,7 @@ method MUST immediately return and MUST NOT emit a PoolReadyEvent.
 
    mark pool as "ready"
    emit PoolReadyEvent
-   resume background thread
+   allow background thread to create connections
 
 Note that resuming the background thread after emitting PoolReadyEvent is of the essence,
 and it must be the case that no observer is able to observe actions of the background thread
@@ -743,8 +743,6 @@ Conceptually, the aforementioned activities are organized into sequential Backgr
 A Run MUST do as much work as readily available and then end instead of waiting for more work.
 For example, instead of waiting for pendingConnectionCount to become less than maxConnecting when satisfying minPoolSize,
 a Run MUST either proceed with the rest of its duties, e.g., closing available perished connections, or end.
-
-Marking a pool as `ready <#marking-a-connection-pool-as-ready>`__ MUST start a new Run as soon as possible.
 
 The duration of intervals between the end of one Run and the beginning of the next Run is not specified,
 but the
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -9,8 +9,8 @@ Connection Monitoring and Pooling
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: N/A
-:Last Modified: 2021-04-12
-:Version: 1.5.0
+:Last Modified: 2021-05-26
+:Version: 1.5.1
 
 .. contents::
 
@@ -745,17 +745,17 @@ thread SHOULD
 Conceptually, the aforementioned activities are organized into sequential Background Thread Runs of unspecified finite duration.
 Each Run MUST do as much work as possible, for example
 
--  if at the beginning of the Run the Connection Pool needed three more Connections to satisfy minPoolSize, then the Run MUST try to populate
-   the Connection Pool with three more connections instead of populating it with only one,
+-  if at the beginning of a Run a Connection Pool needed three more Connections to satisfy minPoolSize, then the Run MUST try to populate
+   the Connection Pool with three more connections instead of populating it with only one and ending,
    unless it encounters a situation preventing it from continuing;
--  if at the beginning of the Run the Connection Pool had three perished available connections, the Run MUST try to close all of them
-   instead of closing only one.
+-  if at the beginning of a Run a Connection Pool had three perished available connections, the Run MUST try to close all of them
+   instead of closing only one and ending.
 
 The first Run MUST start as soon after a Connection Pool becomes `ready <#marking-a-connection-pool-as-ready>`__ as practically possible.
 Neither the minimum delay between Runs (the minimum interval between the end of a Run and the beginning of the next Run)
 nor the desired period of Runs (the desired interval between the beginning of a Run and the beginning of the next Run) is specified.
 The
-`Test Format and Runner specification <https://github.com/mongodb/specifications/tree/master/source/connection-monitoring-and-pooling/tests>`__
+`Test Format and Runner Specification <https://github.com/mongodb/specifications/tree/master/source/connection-monitoring-and-pooling/tests>`__
 may change the specified restrictions or introduce new ones to facilitate testing.
 
 withConnection
@@ -1130,6 +1130,8 @@ Change log
              ConnectionCheckOutFailedEvent
 
 :2021-4-12: Adding in behaviour for load balancer mode.
+
+:2021-05-26: Formalize the behavior of a `Background Thread <#background-thread>`__.
 
 .. Section for links.
 
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -742,6 +742,22 @@ thread SHOULD
       minPoolSize.
 -  Remove and close perished available `Connections <#connection>`_.
 
+Conceptually, the aforementioned activities are organized into sequential Background Thread Runs of unspecified finite duration.
+Each Run MUST do as much work as possible, for example
+
+-  if at the beginning of the Run the Connection Pool needed three more Connections to satisfy minPoolSize, then the Run MUST try to populate
+   the Connection Pool with three more connections instead of populating it with only one,
+   unless it encounters a situation preventing it from continuing;
+-  if at the beginning of the Run the Connection Pool had three perished available connections, the Run MUST try to close all of them
+   instead of closing only one.
+
+The first Run MUST start as soon after a Connection Pool becomes `ready <#marking-a-connection-pool-as-ready>`__ as practically possible.
+Neither the minimum delay between Runs (the minimum interval between the end of a Run and the beginning of the next Run)
+nor the desired period of Runs (the desired interval between the beginning of a Run and the beginning of the next Run) is specified.
+The
+`Test Format and Runner specification <https://github.com/mongodb/specifications/tree/master/source/connection-monitoring-and-pooling/tests>`__
+may change the specified restrictions or introduce new ones to facilitate testing.
+
 withConnection
 ^^^^^^^^^^^^^^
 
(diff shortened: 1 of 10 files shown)


source/connection-monitoring-and-pooling/tests/README.rst
```diff
@@ -36,7 +36,21 @@ Unit Test Format:
...
 All Unit Tests have some of the following fields:
 
-- ``poolOptions``: if present, connection pool options to use when creating a pool
+- ``poolOptions``: If present, connection pool options to use when creating a pool;
+  both `standard ConnectionPoolOptions <https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#connection-pool-options-1>`__
+  and the following test-specific options are allowed:
+
+  - ``backgroundThreadDelayMS``: An artificial delay before starting observable activities in a
+    `Background Thread <https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#background-thread>`__.
+    The Test Runner / Connection Pool MUST try to start counting time towards this delay
```
patrickfreed, line 45: Can we change this to start counting when the pool transitions from paused to ready, since the thread won't be running when the pool is paused anyways? (only make this change assuming we stick with a delay here instead of an on/off flag per my other comment)

source/connection-monitoring-and-pooling/tests/README.rst
```diff
@@ -36,7 +36,21 @@ Unit Test Format:
...
+    at an instant as close to starting executing ``operations`` as reasonably possible.
+    If a Connection Pool does not implement a Background Thread, the Test Runner MUST ignore the option.
+    Possible values:
+
+    - 0 (default)—no delay;
```
patrickfreed, line 50: After looking at the changes, the only value that seems to be needed is -1, so I'm wondering if it would be better to make this just a boolean indicating whether the background thread should run at all. I'm struggling to come up with a case where a hard-coded delay would actually be used.

BorisDog: @patrickfreed maybe this boolean flag can be replaced by [frequency setting](https://jira.mongodb.org/browse/DRIVERS-1465) with 0 or -1 values?

patrickfreed: Yeah that sounds like a good idea. That'll also allow us to remove the 50ms guidance from the README and move it to the tests that require it.

source/connection-monitoring-and-pooling/tests/README.rst
stIncMale: I went the path proposed in the other comment.

source/connection-monitoring-and-pooling/tests/README.rst
BorisDog: I agree that we don't need to implement DRIVERS-1465 now. Evolving switch option to frequency option in DRIVERS-1465 sounds good to me. @patrickfreed I believe there is no concern of significant overhead in evolving switch to frequency option in driver implementations?

patrickfreed: I agree that we don't need to implement DRIVERS-1465 now, but I do think we should implement the frequency option now (for tests only) to prevent churn in the various drivers' test runners in the future. For an individual driver it isn't _significant_ overhead to make this change and then later convert to the frequency option, but when multiplied across every language team it is significant for the DBX team as a whole. Also, if a driver lags in converting its runner to the frequency option, it could end up in a state where it can't sync new unrelated tests due to an outdated runner. Also, existing runners likely have support for tuning the frequency anyways, as the README currently requires it be set to 50ms. The proposed option would just make that variable per test, whereas the on/off flag could require different codepaths.

stIncMale: The concept of frequency as it is currently mentioned in the test runner spec is vague. I introduced `backgroundThreadRunPeriodMS`, which hopefully meets the desired intent.

source/connection-monitoring-and-pooling/tests/README.rst
```diff
@@ -36,7 +36,22 @@ Unit Test Format:
 
 All Unit Tests have some of the following fields:
 
-- ``poolOptions``: if present, connection pool options to use when creating a pool
+- ``poolOptions``: If present, connection pool options to use when creating a pool;
+  both `standard ConnectionPoolOptions <https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#connection-pool-options-1>`__
+  and the following test-specific options are allowed:
+
+  - ``backgroundThreadIntervalMS``: A desired time interval between the end of a
```
BorisDog, line 43: @patrickfreed  should we also mention `backgroundThreadIntervalMS` as internal option or interval between Runs in general in [background thread definition](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#background-thread), as it has direct impact on background thread implementation?

stIncMale: I preemptively (while there is no response) fleshed out a bit possible restrictions a test runner may introduce.

source/connection-monitoring-and-pooling/tests/README.rst
```diff
@@ -36,7 +36,22 @@ Unit Test Format:
...
+    `Background Thread Run <https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#background-thread>`__
+    and the beginning of the next Run. If a Connection Pool does not implement a Background Thread, the Test Runner MUST ignore the option.
+    If the option is not specified, an implementation is free to use any value it finds reasonable.
+    An implementation is also free to use an interval smaller than the desired one. Once the specified time interval since the end of a Run passes,
```
BorisDog, line 47: > An implementation is also free to use an interval smaller than the desired one

Do we want to define Interval as the minimal interval between the runs? Or clarify why would implementation use smaller interval than specified.

BorisDog (changes requested):

source/connection-monitoring-and-pooling/tests/README.rst
stIncMale: It is of the essence to ask and answer a question: how can such a restriction benefit the tests? As far as I know, so far such a benefit has not been expressed.

Here are the benefits of not specifying such a restriction:

1. The requirement _"Marking a pool as ready MUST start a new Run as soon as possible."_ requires violating the "minimal interval" restriction.
2. If a pool implementation wants to guarantee a Run while the pool is paused to clear resources, then having the "minimal interval" restriction complicates the implementation.
3. Without the "minimal interval" restriction, an implementation is free to implement the desired interval as desired period, a.k.a. frequency. Currently the Java driver implements it as desired frequency, and the corresponding Java API is intentionally public. Having the "minimal interval" restriction would require the Java driver to either change the specification of its public API, or introduce an additional internal API with a different implementation, which clogs up the implementation.

source/connection-monitoring-and-pooling/tests/README.rst
```diff
@@ -36,7 +36,22 @@ Unit Test Format:
...
+    an implementation MUST begin the next Run as soon as possible.
+
+    Possible values (0 is not allowed):
+
+    - a negative value—never begin a Run;
+    - a positive value—the desired interval between Runs in milliseconds.
```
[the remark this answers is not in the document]
stIncMale: Done.

source/connection-monitoring-and-pooling/tests/README.rst
```diff
@@ -36,7 +36,22 @@ Unit Test Format:
...
+  - ``backgroundThreadIntervalMS``: A desired time interval between the end of a
+    `Background Thread Run <https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#background-thread>`__
+    and the beginning of the next Run. If a Connection Pool does not implement a Background Thread, the Test Runner MUST ignore the option.
+    If the option is not specified, an implementation is free to use any value it finds reasonable.
+    An implementation is also free to use an interval smaller than the desired one. Once the specified time interval since the end of a Run passes,
+    an implementation MUST begin the next Run as soon as possible.
+
+    Possible values (0 is not allowed):
+
+    - a negative value—never begin a Run;
```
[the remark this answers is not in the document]
stIncMale: Done.

source/connection-monitoring-and-pooling/tests/README.rst
stIncMale: Changed the wording.

source/connection-monitoring-and-pooling/tests/README.rst
```diff
@@ -36,7 +36,23 @@ Unit Test Format:
 
 All Unit Tests have some of the following fields:
 
-- ``poolOptions``: if present, connection pool options to use when creating a pool
+- ``poolOptions``: If present, connection pool options to use when creating a pool;
+  both `standard ConnectionPoolOptions <https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#connection-pool-options-1>`__
+  and the following test-specific options are allowed:
+
+  - ``backgroundThreadIntervalMS``: A desired minimal time interval between the end of a
```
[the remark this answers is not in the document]
stIncMale: @BorisDog wants wording that discourages implementations from choosing an interval much smaller than the desired one. This is an attempt to prevent implementations from being too inefficient by deciding to use a too short interval or no interval at all (with DRIVERS-1465 the interval will affect not only tests). Given that the whole CMAP spec is purely about efficiency, this is not unreasonable.

I am fine with both "minimal desired" and just "desired" wordings.

source/connection-monitoring-and-pooling/tests/README.rst
```diff
@@ -36,7 +36,23 @@ Unit Test Format:
...
+    `Background Thread Run <https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#background-thread>`__
+    and the beginning of the next Run. If a Connection Pool does not implement a Background Thread, the Test Runner MUST ignore the option.
+    If the option is not specified, an implementation is free to use any value it finds reasonable.
+    An implementation SHOULD respect the lower bound restriction,
```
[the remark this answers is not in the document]
stIncMale: At least items 2 and 3 from https://github.com/mongodb/specifications/pull/998#discussion_r643131171 still hold without the requirement _"Marking a pool as ready MUST start a new Run as soon as possible"_.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -736,12 +736,21 @@ A Pool SHOULD have a background Thread that is responsible for
...
 thread SHOULD
 
--  Populate `Connections <#connection>`_ to ensure that the pool always satisfies **minPoolSize**
-    - The background thread SHOULD just go back to sleep instead of waiting for
-      pendingConnectionCount to become less than maxConnecting when satisfying
-      minPoolSize.
+-  Populate `Connections <#connection>`_ to ensure that the pool always satisfies minPoolSize
 -  Remove and close perished available `Connections <#connection>`_.
 
+Conceptually, the aforementioned activities are organized into sequential Background Thread Runs of unspecified duration.
```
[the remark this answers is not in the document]
stIncMale: Done.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -736,12 +736,21 @@ A Pool SHOULD have a background Thread that is responsible for
...
+A Run MUST do as much work as readily available and then end instead of waiting for more work.
+For example, instead of waiting for pendingConnectionCount to become less than maxConnecting when satisfying minPoolSize,
+a Run MUST either proceed with the rest of its duties, e.g., closing available perished connections, or end.
+
+Marking a pool as `ready <#marking-a-connection-pool-as-ready>`__ MUST start a new Run as soon as possible.
```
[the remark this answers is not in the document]

source/connection-monitoring-and-pooling/tests/README.rst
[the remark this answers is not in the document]
patrickfreed: Maybe we ought to drop "desired" then too? There shouldn't be any choice being made here by the drivers--if an interval is specified they must adhere to it as exactly as they reasonably can (barring any noise from waiting on lock acquisitions and what not).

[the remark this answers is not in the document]
stIncMale: At this point I am fine with any wording once a consensus is reached between you and Boris. Currently it appears to me that there is no consensus. @BorisDog, are you fine with just saying _"A time interval"_ without any further clarification on what is expected from the implementation?

source/connection-monitoring-and-pooling/tests/README.rst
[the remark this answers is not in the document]
stIncMale: > Regarding 2 and the now removed requirement, drivers should refrain from implementing that behavior until it is specified in CMAP to ensure a consistent implementation across all drivers

The specification currently does not forbid releasing resources when a pool is paused. I would be surprised if any driver implemented an approach that guarantees that the background thread does nothing while a pool is paused (for example, checking from the background thread whether a pool is paused is not enough to achieve such a guarantee, no matter how often the check is done, even if it is done before clearing each connection). I am expecting that all current implementation may occasionally release some resources after pausing a pool, even if it was not intended.

> if Java implements it, you'll run the (probably small) risk of failing a new test written without that assumption in mind

I don't think this is the case. As I explained above, there it is impossible to derive a guarantee that a paused pool does not release resources based on the current specification. Therefore, no test may rely on such a guarantee, and implementation that always releases resources in the paused state cannot not fail a test, provided that a test is correct.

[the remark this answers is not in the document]
patrickfreed: >I am expecting that all current implementation may occasionally release some resources after pausing a pool, even if it was not intended.

The behavior I'm referring to is specifically beginning new runs immediately in `ready()` and `clear()`, which is yet to be specified. It is totally acceptable and expected that background threads will continue to run while a pool is paused for the purposes of resource cleanup; it's just not expected that they will start new runs immediately on state changes. Given that when runs are to be started is unspecified, drivers technically can start them in `ready()` and `clear()` if they want, but as I was saying earlier they run the risk of needing to be updated once such behavior does become specified in DRIVERS-1465.

>I don't think this is the case. As I explained above, it is impossible to derive a guarantee that a paused pool does not release resources based on the current specification. Therefore, no test may rely on such a guarantee, and implementation that always releases resources in the paused state cannot not fail a test, provided that the test is correct.

Right I agree that the risk for this is minimal. The more realistic risk is that Java's implementation will have to change in DRIVERS-1465.

source/connection-monitoring-and-pooling/tests/README.rst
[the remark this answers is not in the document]
stIncMale: Done.

stIncMale:
I believe I addressed all concerns that have been expressed until now.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -736,12 +736,19 @@ A Pool SHOULD have a background Thread that is responsible for
 monitoring the state of all available `Connections <#connection>`_. This background
 thread SHOULD
 
--  Populate `Connections <#connection>`_ to ensure that the pool always satisfies **minPoolSize**
-    - The background thread SHOULD just go back to sleep instead of waiting for
-      pendingConnectionCount to become less than maxConnecting when satisfying
-      minPoolSize.
+-  Populate `Connections <#connection>`_ to ensure that the pool always satisfies minPoolSize
```
BorisDog, line 739: minor: period.

patrickfreed (commented):
Tried it out in the Rust driver, all tests pass, just have one minor comment on some existing wording.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -417,7 +417,7 @@ method MUST immediately return and MUST NOT emit a PoolReadyEvent.
 
    mark pool as "ready"
    emit PoolReadyEvent
-   resume background thread
+   allow background thread to create connections
 
 Note that resuming the background thread after emitting PoolReadyEvent is of the essence,
```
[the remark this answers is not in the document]

patrickfreed (approved):
LGTM! Thanks for taking this on!

BorisDog (approved):

PR #1002, opened 2021-05-26
DRIVERS-1781 Specify 5.0 cursor behaviour

alcaeus:
DRIVERS-1781

As discovered recently, cursor behaviour has changed in 5.0. When `getMore` has a full batch, the cursor stays open, even if the original limit on the cursor was reached. Since the server team has decided to keep these changes in place, a new spec test is added to test the new 5.0 behaviour, and a note is added to document this change in behaviour.

Note: while part of the APM test suite, these tests cover the "find, getMore, and killCursors" specification. I've created DRIVERS-1782 to move these tests once they have been converted to the unified format.


source/command-monitoring/tests/find.json
```diff
@@ -526,6 +527,150 @@
...
+            "database_name": "command-monitoring-tests"
+          }
+        },
+        {
+          "command_succeeded_event": {
+            "reply": {
+              "ok": 1,
+              "cursor": {
+                "id": {
+                  "$numberLong": "42"
```
jyemin, line 663: Wouldn't we expect "0" here?

alcaeus: Yes. Tests still pass because "42" matches "0" as well, but the intention was to be more specific. Thanks for pointing this out.

source/command-monitoring/tests/find.yml
```diff
@@ -195,8 +195,9 @@ tests:
               - { $numberLong: "42" }
           command_name: "killCursors"
   -
-    description: "A successful find event with a getmore and the server kills the cursor"
+    description: "A successful find event with a getmore and the server kills the cursor (<= 4.4)"
     ignore_if_server_version_less_than: "3.1"
+    ignore_if_server_version_greater_than: "4.4"
```
jmikola, line 200: For the benefit of other reviewers, this covers all 4.4.x minor versions (per [README](https://github.com/mongodb/specifications/blob/master/source/command-monitoring/tests/README.rst#ignoring-tests-based-on-server-version-or-topology-type)). It's been a while since I looked at this test format so I wanted to double-check.

source/command-monitoring/tests/find.yml
```diff
@@ -247,6 +248,76 @@ tests:
...
+    ignore_if_server_version_less_than: "5.0"
+    ignore_if_topology_type:
+      - "sharded"
+    operation:
+      name: "find"
+      arguments:
+        filter: { _id: { $gte: 1 }}
+        sort: { _id: 1 }
+        batchSize: {"$numberLong": "3"}
+        limit: {"$numberLong": "4"}
```
jmikola, line 262: Is it necessary to use `$numberLong`, or is this just copying what was done in other tests?

I notice we don't use it for all numerics (`filter` and `sort` above for instance).

source/command-monitoring/tests/find.yml
```diff
@@ -247,6 +248,76 @@ tests:
               nextBatch:
                 - { _id: 4, x: 44 }
           command_name: "getMore"
+  -
+    description: "A successful find event with a multiple getMore (>= 5.0)"
```
jmikola, line 252: Since this is a new test, does it make sense to add it using the old format instead of creating a new `unified` sub-directory for this spec and adding it there? IIRC, one of the expectations of the unified format would be that all new tests would only be written in the unified format (excluding cases like FLE, where the spec itself isn't yet supported).

source/find_getmore_killcursors_commands.rst
```diff
@@ -365,18 +365,35 @@ More detailed information about the interaction of the **secondaryOk** with **OP
...
-
-When setting the **batchSize** on the **find** and **getMore** command the value MUST be based on the cursor limit calculations specified in the `CRUD`_ specification. 
-
-In the following example the **limit** is set to **4** and the **batchSize** is set to **3** the following commands are executed.
+The new **find** command has different semantics to the existing 3.0 and earlier
+**OP_QUERY** wire protocol message. The **limit** field is a hard limit on the
+total number of documents returned by the cursor no matter what **batchSize** is
+provided. This includes other limiting operations, such as the **$limit**
+aggregation pipeline stage. This differs from existing **OP_QUERY** behavior
+where there is no server side concept of limit and where the driver **MUST**
```
jmikola, line 373: "server-side"

source/find_getmore_killcursors_commands.rst
```diff
@@ -365,18 +365,35 @@ More detailed information about the interaction of the **secondaryOk** with **OP
...
+keep track of the limit on the client side and **MUST** send a
+**OP_KILL_CURSORS** wire protocol message when it limit is reached.
+
+When setting the **batchSize** on the **find** and **getMore** command the value
+MUST be based on the cursor limit calculations specified in the `CRUD`_
```
jmikola, line 378: Since "MUST" isn't bolded here, I'd suggest removing the bold formatting from any META terms in the previous text changed by this PR as well.

source/find_getmore_killcursors_commands.rst
```diff
@@ -365,18 +365,35 @@ More detailed information about the interaction of the **secondaryOk** with **OP
...
+The new **find** command has different semantics to the existing 3.0 and earlier
+**OP_QUERY** wire protocol message. The **limit** field is a hard limit on the
+total number of documents returned by the cursor no matter what **batchSize** is
+provided. This includes other limiting operations, such as the **$limit**
+aggregation pipeline stage. This differs from existing **OP_QUERY** behavior
+where there is no server side concept of limit and where the driver **MUST**
+keep track of the limit on the client side and **MUST** send a
+**OP_KILL_CURSORS** wire protocol message when it limit is reached.
+
+When setting the **batchSize** on the **find** and **getMore** command the value
```
jmikola, line 377: "commands" since you're referring to two types. I'd also suggest writing "the value of limit" since this section refers to "limit, skip, and batchSize". This is the first reference to a value in this paragraph, so I think it's helpful to clarify in case someone immediately jumps into this and starts reading from here.

source/find_getmore_killcursors_commands.rst
```diff
@@ -365,18 +365,35 @@ More detailed information about the interaction of the **secondaryOk** with **OP
...
+MUST be based on the cursor limit calculations specified in the `CRUD`_
+specification.
+
+Note that with 5.0, the server-side handling of cursors with a limit has
+changed. Before 5.0, some cursors were automatically closed when the limit was
+reached (e.g. when running **find** with **limit**), and the reply document did
+not include a cursor ID. Starting with 5.0, all cursor-producing operations will
```
jmikola, line 384: Suggestion to clarify what "did not include" actually means: "did not include a cursor ID (i.e. `cursor.id` was `0`)".

source/find_getmore_killcursors_commands.rst
```diff
@@ -365,18 +365,35 @@ More detailed information about the interaction of the **secondaryOk** with **OP
...
+return a cursor ID if the end of the batch being returned lines up with the
+limit on the cursor. In this case, drivers MUST send an additional **getMore**
+request with **batchSize** 1 to have the cursor closed on the server side.
```
jmikola, line 387: Is the `batchSize` relevant here? Presumably _any_ further iteration would result in the server closing the cursor. We just need this to overcome the new query engine's behavior that avoids peaking ahead.

source/find_getmore_killcursors_commands.rst
```diff
@@ -365,18 +365,35 @@ More detailed information about the interaction of the **secondaryOk** with **OP
...
+
+In the following example the **limit** is set to **4** and the **batchSize** is
+set to **3** the following commands are executed.
 
 .. code:: javascript
 
     {find: ..., batchSize:3}
-    {getMore: ..., batchSize:1}
+    {getMore: ..., batchSize:1} // Returns remaining items but leaves cursor open on 5.0
```
jmikola, line 395: Apologies if this was covered in the HELP ticket, but what would happen if drivers purposely added `1` to the `batchSize`? Couldn't that guarantee that the query engine _would_ look ahead as necessary and immediately close the cursor? Provided the server enforces the original `find` command's `limit`, I don't think there's be any impact on the returned batch contents.

source/find_getmore_killcursors_commands.rst
```diff
@@ -462,7 +479,7 @@ Semantics of maxTimeMS for a Driver
 
 In the case of  a **non-tailable cursor query** OR **a tailable cursor query with awaitData == false**, the driver MUST set maxTimeMS on the **find** command and MUST NOT set maxTimeMS on the **getMore** command.
 
-In the case of **a tailable cursor with awaitData == true** the driver MUST provide a Cursor level option named **maxAwaitTimeMS** (See CRUD specification for details). The **maxTimeMS** option on the **getMore** command MUST be set to the value of the option **maxAwaitTimeMS**. If no **maxAwaitTimeMS** is specified, the driver SHOULD not set **maxTimeMS** on the **getMore** command. 
+In the case of **a tailable cursor with awaitData == true** the driver MUST provide a Cursor level option named **maxAwaitTimeMS** (See CRUD specification for details). The **maxTimeMS** option on the **getMore** command MUST be set to the value of the option **maxAwaitTimeMS**. If no **maxAwaitTimeMS** is specified, the driver SHOULD not set **maxTimeMS** on the **getMore** command.
```
jmikola, line 482: "cursor-level" would seem consistent with earlier references to "server-side". I realize it's in the original text, so no objections if you want to leave as-is.

Optionally link to the CRUD spec, since we already have an anchor defined from an earlier paragraph. "See" can also be made lowercase.

Edit: Just realized the only change here was removing trailing whitespace. Feel free to ignore my suggestions.

source/unified-test-format/tests/valid-pass/poc-command-monitoring.yml
```diff
@@ -29,9 +29,10 @@ initialData:
       - { _id: 5, x: 55 }
 
 tests:
-  - description: "A successful find event with a getmore and the server kills the cursor"
+  - description: "A successful find event with a getmore and the server kills the cursor (<= 4.4)"
     runOnRequirements:
       - minServerVersion: "3.1"
+        maxServerVersion: "4.8.99"
```
jmikola, line 35: Since min/max are inclusive, I think `4.4.99` makes more sense.

source/find_getmore_killcursors_commands.rst
```diff
@@ -365,18 +365,35 @@ More detailed information about the interaction of the **secondaryOk** with **OP
...
+When setting the **batchSize** on the **find** and **getMore** command the value
+MUST be based on the cursor limit calculations specified in the `CRUD`_
+specification.
+
+Note that with 5.0, the server-side handling of cursors with a limit has
+changed. Before 5.0, some cursors were automatically closed when the limit was
+reached (e.g. when running **find** with **limit**), and the reply document did
+not include a cursor ID. Starting with 5.0, all cursor-producing operations will
+return a cursor ID if the end of the batch being returned lines up with the
+limit on the cursor. In this case, drivers MUST send an additional **getMore**
```
jyemin, line 386: The way this is written, it seems like a special case that needs to be added.  But don't we want to remove the special case?  Should drivers just exhaust the cursor by executing getMore with the same batch size, up until the server returns a cursorId of 0?  And just let the server worry about enforcing the limit?

source/command-monitoring/tests/find.yml
```diff
@@ -247,6 +248,76 @@ tests:
...
+          reply:
+            ok: 1.0
+            cursor:
+              id: { $numberLong: "42" }
+              ns: *namespace
+              nextBatch:
+                - { _id: 4, x: 44 }
+          command_name: "getMore"
+      -
+        command_started_event:
```
jyemin, line 305: I commented on this below, but just wanted to note that he Java driver fails this assertion.  Instead of sending a final getMore in order to exhaust the cursor, it notices that it has reached the limit and instead kills the cursor.

jyemin: I tried to get this to pass in Java driver and the changes just got more and more intrusive.  I think we should discuss.

source/command-monitoring/tests/find.yml
alcaeus: Yes, this was pure copy/paste here.

source/command-monitoring/tests/find.yml
alcaeus: I was considering using the new format, but given that I just want to fix the test I opted for not requiring drivers to set up a unified runner for this one test. DRIVERS-1448 will bring in more changes for this, so I only focused on getting the test to run.

source/find_getmore_killcursors_commands.rst
alcaeus: I'll make sure the new content is formatted consistently what's already there and can update the formatting to match our other specs in a separate PR for easier reviewing.

source/find_getmore_killcursors_commands.rst
alcaeus: Strictly speaking the batchSize is completely irrelevant here as long as we're fetching at least one result. I used `batchSize` 1 because that's what libmongoc uses, but I'm happy to accept any `batchSize` for the extra batch.

source/find_getmore_killcursors_commands.rst
alcaeus: Yes, that would work, but I believe that to be a more invasive spec change, considering that the current behaviour is designed to match the `OP_QUERY` / `OP_GET_MORE` behaviour. DRIVERS-1448 will change this considerably so I was deferring major changes to then.

source/find_getmore_killcursors_commands.rst
alcaeus: Yes, that's my IDE automatically trimming trailing whitespace. I can remove the change from the diff and take care of this in the separate PR that brings in formatting.

source/find_getmore_killcursors_commands.rst
alcaeus: The goal here is to not change the behaviour entirely (removing limit tracking), as that'd be more effort than fixing the tests. Since the cursor is open, either `killCursors` or `getMore` are expected. For any driver that doesn't track the limit, the presence of a cursor ID means that they'll do another `getMore` request, which is why I added this behaviour. I'm all for dropping this tracking entirely, especially since this behaviour doesn't cover cursors returned from `aggregate` (where we don't have an explicit limit), but would like to defer that to Q3 and DRIVERS-1448. For now I only want to ensure that the behaviour is documented and drivers don't leave any open cursors on servers.

source/command-monitoring/tests/find.yml
alcaeus: Ah, I was afraid there would be drivers that send `killCursors` instead of an additional `getMore`. I could relax the spec language to require closing the cursor, either by sending off an additional `getMore`, or by sending `killCursors`. This would document the behaviour, instruct drivers to ensure cursors are closed, but testing would be more difficult. I can see if I can cover both behaviours with a single test, but we may have to either not have any tests, not assert a whole lot about the additional command being sent (other than it being successful), or have two tests (one with getMore, one with killCursors) with drivers having to skip the one that doesn't apply to them.

jyemin: I think a goal here is to not require any changes to existing cursor logic, since the current logic, whatever it may be, should still be correct with 5.0.  I'd be ok not testing with 5.0 and defer spec/test changes to DRIVERS-1448.

alcaeus: So essentially:
* we limit the test to <= 4.4
* we clarify the spec to instruct drivers to close cursors (using {{getMore}} or {{killCursors}})
* creating tests and changing the specification (e.g. by dropping the driver-side limit logic) are deferred to DRIVERS-1448

What do you think @jmikola?

kmahar:
just a heads up I'm not going to have time to review this until next Tuesday or Wednesday. I think Jeremy and Jeff can probably give sufficient review coverage so don't feel you need to wait on me if this needs to move along more quickly.

source/find_getmore_killcursors_commands.rst
jmikola: Ok, I don't feel strongly. It might be awkward to have to split this up (e.g. MUST got `getMore`, SHOULD for a particular batchSize). Perhaps: "MUST send an additional `getMore` with a positive number for `batchSize` (e.g. 1)."

source/command-monitoring/tests/find.yml
jmikola: SGTM.

source/find_getmore_killcursors_commands.rst
ShaneHarvey: I believe that pymongo will send a killCursors command in this case to cleanup the cursor which works on all server versions. Adding a requirement that we use getMore with batchSize=1 instead adds extra complexity since we'll still need to use killCursors on pre-3.2 servers. If we use the getMore batchSize=1 approach on older servers we could actually get 1 document which is wasteful, especially if the document is large.

Can we simplify this by always sending killCursors when the limit is reached instead of sometimes killCursors and sometimes getMore?

jyemin: As I said on another comment thread, I would prefer that this PR not require _any_ changes to any driver's cursor implementation.

ShaneHarvey: Yes I would prefer no driver changes as well.

source/command-monitoring/tests/find.yml
alcaeus: I've removed the test as per the discussion below.

source/find_getmore_killcursors_commands.rst
alcaeus: Added as suggested

source/find_getmore_killcursors_commands.rst
alcaeus: I've updated this to no longer prescribe a single method. The spec now requires drivers (the language uses "MUST", but I have no objections to change this to "SHOULD") to either use `getMore` or `killCursors` to close the server-side cursor if a cursor ID is returned.

I've also used this to clarify what the driver should do if the additional `getMore` (unexpectedly) yields more documents in the batch or still includes a cursor ID. I expect drivers to have varying implementations here and we don't test for this yet, so I'm fine with drivers not complying with this yet, especially since that is a very rare scenario (the driver would have to assume there's a limit on the cursor which the server doesn't see). I thought it was a good idea to be defensive here and provide this guidance.

source/command-monitoring/tests/find.yml
alcaeus: I've made the changes as suggested above.

source/find_getmore_killcursors_commands.rst
```diff
@@ -365,18 +365,54 @@ More detailed information about the interaction of the **secondaryOk** with **OP
...
+**OP_KILL_CURSORS** wire protocol message when it limit is reached.
+
+When setting the **batchSize** on the **find** and **getMore** commands the
+value of **batchSize** **MUST** be based on the cursor limit calculations
+specified in the `CRUD`_ specification.
+
+Note that with 5.0, the server-side handling of cursors with a limit has
+changed. Before 5.0, some cursors were automatically closed when the limit was
+reached (e.g. when running **find** with **limit**), and the reply document did
+not include a cursor ID (i.e. ``cursor.id`` was ``0`). Starting with 5.0, all
```
jmikola, line 384: Missing an extra backtick after "0".

jmikola (approved):

alcaeus:
@ShaneHarvey @jyemin on the subject of having to change existing behaviour in drivers, I believe that any driver that handles this case correctly will not need changing. Strictly speaking, we should always ensure that a cursor is either exhausted or closed when we no longer read from it. A driver that neither exhausts a limited cursor (i.e. does not send an additional `getMore`) nor uses `killCursors` to close that cursor if the last `getMore` returned a cursor ID will produce a large number of open cursors on 5.0+ servers because the condition that leaves the server-side cursor open is triggered on every `find` with a `limit`. These drivers don't need changes now, but they should fix this behaviour to avoid allocating server resources longer than necessary, hence the suggestion in the spec to exhaust the cursor.

More fundamental changes will come in DRIVERS-1448, which will require all drivers to make changes to their cursor implementation, so I'm confident that each driver can decide whether they are handling this case appropriately or whether they want to make changes before we address DRIVERS-1448. If this PR is approved, I'd leave guidance in DRIVERS-1781 to ensure drivers don't feel compelled to make changes to their cursor logic when their behaviour is correct.

source/find_getmore_killcursors_commands.rst
```diff
@@ -365,18 +365,54 @@ More detailed information about the interaction of the **secondaryOk** with **OP
...
+being returned lines up with the limit on the cursor. In this case, drivers
+**MUST** ensure the cursor is closed on the server. To do this, drivers **MUST**
+use one of the options provided below:
+
+* send an additional **getMore** command with **batchSize** 1. Since the limit
+  has been reached, the **getMore** response is expected to contain neither a
+  cursor ID, nor any elements in the batch. If results are returned in the
+  **getMore** call, these **MUST** be discarded. If the result contains a cursor
+  ID, the driver **MUST NOT** attempt another **getMore**, but **MUST** use
+  **killCursors** as prescribed below.
```
ShaneHarvey, line 395: Apologies if this is a repeat question but I see this in the help ticket:
> Drivers that don't track the limit themselves will send out an additional getMore command for every cursor when the end of a batch lines up with the size of the cursor (e.g. count(cursor) == n * cursor.batchSize)

Do any drivers *not* track the limit? The existing test seems to suggest that all drivers do track the limit because of the updated getMore batchSize in the CommandStartedEvent. If so, would it be possible to add an extra command-monitoring test for >=5.0 which asserts that the driver sends a killCursors command?

ShaneHarvey (commented):
Thanks @alcaeus! I have one question regarding the tests and driver limit-tracking.

source/find_getmore_killcursors_commands.rst
```diff
@@ -365,18 +365,54 @@ More detailed information about the interaction of the **secondaryOk** with **OP
...
+Note that with 5.0, the server-side handling of cursors with a limit has
+changed. Before 5.0, some cursors were automatically closed when the limit was
+reached (e.g. when running **find** with **limit**), and the reply document did
+not include a cursor ID (i.e. ``cursor.id`` was ``0``). Starting with 5.0, all
+cursor-producing operations will return a cursor ID if the end of the batch
+being returned lines up with the limit on the cursor. In this case, drivers
+**MUST** ensure the cursor is closed on the server. To do this, drivers **MUST**
+use one of the options provided below:
+
+* send an additional **getMore** command with **batchSize** 1. Since the limit
```
jyemin, line 390: I think this options is only included because the C driver currently does this.  But would we want any driver that didn't already do this to choose this option?  I would think not, as option 2 is much simpler.  If there's agreement on this point, I think we should consider just removing this option from the PR, and just call the C driver out of spec until DRIVERS-1448.

jyemin (changes requested):

source/find_getmore_killcursors_commands.rst
alcaeus: You are correct, drivers do track the limit, I believe this was an oversight when I wrote this.

As for the additional test, I had that in but @jyemin raised the point that drivers might not conform to this. For example, libmongoc sends an additional `getMore`, which means that adding a test that requires `killCursors` to be sent requires work in libmongoc, which we want to avoid here. I've only added guidance to the spec to make sure drivers check whether they handle this use-case correctly.

Once this PR is merged, I'll update the DRIVERS ticket with guidance for drivers what to do here, which is essentially:
* Check if the offending spec test fails on 5.0.0rc0
* Sync spec tests

If the spec test changed in this PR does not fail for a driver, it means that the driver does not send an additional command to exhaust or kill the cursor on the server. I'll leave it to each language team to decide the best course of action; it's not optimal to leave cursors open, but considering our busy workload it may not be possible to fix this at this time. These changes will be necessary once DRIVERS-1448 is done.

source/find_getmore_killcursors_commands.rst
alcaeus: The reason I offered both is because they achieve the same result.

If we're looking at DRIVERS-1448, then the result of not tracking the limit on the cursor means that an additional `getMore` will be required, as the only indication that we've reached the limit will be the absence of a cursor ID in the reply document. With that in mind, I would continue to suggest both options, meaning that every driver conforms to the spec until changes may become necessary with DRIVERS-1448.

jyemin: Discussed in person, and agreed that some changes will be made here to simplify the requirements.

jyemin (changes requested):
Awaiting updates.

source/find_getmore_killcursors_commands.rst
alcaeus: Updated to remove the detailed suggestions. The spec now asks drivers to ensure the cursor is closed by exhausting it or using `killCursors`.

source/find_getmore_killcursors_commands.rst
```diff
@@ -365,18 +365,38 @@ More detailed information about the interaction of the **secondaryOk** with **OP
...
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-The new **find** command has different semantics to the existing 3.0 and earlier **OP_QUERY** wire protocol message. The **limit** field is a hard limit on the total number of documents returned by the cursor no matter what **batchSize** is provided.
-
-Once the limit on the cursor has been reached the server will destroy the cursor and return a **cursorId** of **0** in the **OP_REPLY**. This differs from existing **OP_QUERY** behavior where there is no server side concept of limit and where the driver **MUST** keep track of the limit on the client side and **MUST** send a **OP_KILL_CURSORS** wire protocol message when it limit is reached.
-
-When setting the **batchSize** on the **find** and **getMore** command the value MUST be based on the cursor limit calculations specified in the `CRUD`_ specification. 
-
-In the following example the **limit** is set to **4** and the **batchSize** is set to **3** the following commands are executed.
+The new **find** command has different semantics to the existing 3.0 and earlier
```
jyemin, line 368: Nit: remove "new".  It's not new anymore! :)

jyemin (approved):
LGTM

source/find_getmore_killcursors_commands.rst
jmikola: Do we need to run this by marketing?

source/find_getmore_killcursors_commands.rst
```diff
@@ -365,18 +365,38 @@ More detailed information about the interaction of the **secondaryOk** with **OP
...
+**killCursors** or **getMore**, depending on how a driver ensures the cursor is
+closed on 5.0:
 
 .. code:: javascript
 
-    {find: ..., batchSize:3}
-    {getMore: ..., batchSize:1}
+    {find: ..., batchSize:3, limit:4}
+    {getMore: ..., batchSize:1} // Returns remaining items but leaves cursor open on 5.0
+    {...}          // Kills server-side cursor. Necessary on
```
ShaneHarvey, line 399: I think there's some missing text here: "Necessary on 5.0+?"

alcaeus: Yes. Thank you

PR #1003, opened 2021-05-27
DRIVERS-1774 Move expireAfterSeconds in time-series tests

benjirewis:
DRIVERS-1774

Moves the `expireAfterSeconds` time-series option to be a top-level `createCollection` option instead of one under `timeseries`. Changes its value to a `$numberLong` to specify that it should be stored as an `int64` or equivalent. Removes `outcome` expectation from the duplicate `_id` test because of potential variations in ordering of documents with duplicate `_id` fields in test runners.

### DIFF
--- a/source/collection-management/tests/timeseries-collection.yml
+++ b/source/collection-management/tests/timeseries-collection.yml
@@ -34,10 +34,11 @@ tests:
         object: *database0
         arguments:
           collection: *collection0Name
+          expireAfterSeconds:
+            $numberLong: "604800"
           timeseries: &timeseries0
             timeField: "time"
             metaField: "meta"
-            expireAfterSeconds: 604800
       - name: assertCollectionExists
         object: testRunner
         arguments:
@@ -67,7 +68,11 @@ tests:
         object: *database0
         arguments:
           collection: *collection0Name
-          timeseries: *timeseries0
+          expireAfterSeconds:
+            $numberLong: "604800"
+          timeseries: &timeseries0
+            timeField: "time"
+            metaField: "meta"
       - name: assertCollectionExists
         object: testRunner
         arguments:
@@ -109,7 +114,6 @@ tests:
               command:
                 insert: *collection0Name
                 documents: *docs
-    outcome:
-    - collectionName: *collection0Name
-      databaseName: *database0Name
-      documents: *docs
+
+   # Expect no particular outcome from the collection since ordering of
+   # documents may vary given duplicate _ids.
--- a/source/collection-management/tests/timeseries-collection.yml
+++ b/source/collection-management/tests/timeseries-collection.yml
@@ -34,8 +34,8 @@ tests:
         object: *database0
         arguments:
           collection: *collection0Name
-          expireAfterSeconds:
-            $numberLong: "604800"
+          # expireAfterSeconds should be an int64 (as it is stored on the server).
+          expireAfterSeconds: 604800
           timeseries: &timeseries0
             timeField: "time"
             metaField: "meta"
@@ -54,6 +54,8 @@ tests:
           - commandStartedEvent:
               command:
                 create: *collection0Name
+                expireAfterSeconds:
+                  $numberLong: "604800"
                 timeseries: *timeseries0
               databaseName: *database0Name
 
@@ -68,11 +70,9 @@ tests:
         object: *database0
         arguments:
           collection: *collection0Name
-          expireAfterSeconds:
-            $numberLong: "604800"
-          timeseries: &timeseries0
-            timeField: "time"
-            metaField: "meta"
+          # expireAfterSeconds should be an int64 (as it is stored on the server).
+          expireAfterSeconds: 604800
+          timeseries: *timeseries0
       - name: assertCollectionExists
         object: testRunner
         arguments:
@@ -98,6 +98,12 @@ tests:
                   }
                 }
               }
+      - name: find
+        object: *collection0
+        arguments:
+          filter: {}
+          sort: { time: 1 }
+        expectResult: *docs
     expectEvents:
       - client: *client0
         events:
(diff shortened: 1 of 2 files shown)


kevinAlbs (approved):
LGTM! I verified this locally with the C driver against `5.0.0-alpha0-602-gb334917`

source/collection-management/tests/timeseries-collection.yml
```diff
@@ -34,10 +34,11 @@ tests:
         object: *database0
         arguments:
           collection: *collection0Name
+          expireAfterSeconds:
+            $numberLong: "604800"
```
jmikola, line 38: Can you confirm whether or not the server actually requires this to be an Int64 type (i.e. 0x12 instead of 0x10 for Int32)? I would expect the server to accept either integer type, in which case I don't think there is any need to use `$numberLong` here.

Keep in mind that drivers are likely to parse this Extended JSON into their native integer type. That may have no direct impact on the type used for the outgoing command, particular if drivers decide to encode integers into Int32 or Int64 BSON types based on the integer's value.

If you're concerned with asserting the outgoing type, you should consider using `$numberLong` in the APM assertion to match the outgoing command.

----

If the intention was the document the type until the actual spec is written, I think that can be accomplished with a comment instead.

I'm actually not sure if the server would even end up documenting the particular type, as the similarly named option for [`createIndexes`](https://docs.mongodb.com/v5.0/reference/command/createIndexes/) is only reported as an "integer". That's not to say we shouldn't do so ourselves (e.g. CRUD mentions when a value can be 64-bit).

source/collection-management/tests/timeseries-collection.yml
```diff
@@ -67,7 +68,11 @@ tests:
         object: *database0
         arguments:
           collection: *collection0Name
-          timeseries: *timeseries0
+          expireAfterSeconds:
+            $numberLong: "604800"
+          timeseries: &timeseries0
```
jmikola, line 73: I'm not sure if you intended to redefine this anchor, but I assume not since you're using the name numbering as the previous (zero). This is apparently [valid to do](https://stackoverflow.com/a/40376576); however, we don't redefine anchors in other tests AFAIK.

In this particular case, the value is exactly the same as the previous usage, so I would expect you could just replace this line with `timeseries: *timeseries0` and call it a day.

source/collection-management/tests/timeseries-collection.yml
```diff
@@ -109,7 +114,6 @@ tests:
               command:
                 insert: *collection0Name
                 documents: *docs
-    outcome:
-    - collectionName: *collection0Name
-      databaseName: *database0Name
-      documents: *docs
+
+   # Expect no particular outcome from the collection since ordering of
+   # documents may vary given duplicate _ids.
```
jmikola, line 119: I realize you can't use `outcome` here, but you can still do an assertion with a sorted `find` operation and `expectResult`. I've done this elsewhere in other tests where insert order might be ambiguous.

source/collection-management/tests/timeseries-collection.yml
benjirewis: For context, this switch to `$numberLong` came out of a [discussion](https://github.com/mongodb/mongo-go-driver/pull/675#discussion_r640190211) in the Go driver implementation.

The server will accept either integer type, but it does [store](https://github.com/mongodb/mongo/blob/1a6e08a61c138e7da0500c11de78d3489b7024d4/src/mongo/db/catalog/collection_options.h#L156) the type in collection_options as an int64. Since it does accept both types, I won't use `$numberLong` here, but I did add a comment saying that the type should be int64 and used `$numberLong` in the APM assertion for the outgoing command for the sake of consistency across drivers

source/collection-management/tests/timeseries-collection.yml
benjirewis: Ah I did not, but good to know about yaml anchor re-definitions 🤔 

Just used `*timeseries0` here.

source/collection-management/tests/timeseries-collection.yml
benjirewis: Gotcha, removed this comment in favor of a sorted `find` operation with `expectResult: *docs`.

source/collection-management/tests/timeseries-collection.yml
jmikola: In light of [Flexible Numeric Comparisons](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#flexible-numeric-comparisons), I actually think it's pointless to use `$numberLong` to assert the outgoing type. It's rare that we do this for spec tests, although I do see things like `{$numberLong: "1" }` used for `lsid` in some of our legacy transaction spec tests. IMO, it makes no sense to do this unless we actually need to represent a large number in Extended JSON.

If we actually wanted to match a particular BSON type, we have [`$$type`](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#type), but that can't currently be combined with a value check. It's essentially a wildcard (like `$$exists: true`) with a type requirement.

If it helps to have prior art for this, consider all of the instances of "Int64" for various options in the [CRUD](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst) spec. We don't ever assert those types in our spec tests, nor do we have tests that use values larger than `INT32_MAX`. It's purely documentation for the driver's implementation.

Since there's not yet a spec here, I'd suggest just having a comment and avoiding `$numberLong` altogether.



I gave some more thought to using `$numberLong` for the APM assertion (because we very seldom do that in other spec tests) because we don't do that at all for

source/collection-management/tests/timeseries-collection.yml
```diff
@@ -103,13 +113,17 @@ tests:
...
-    outcome:
-    - collectionName: *collection0Name
-      databaseName: *database0Name
-      documents: *docs
+          - commandStartedEvent:
+              command:
+                find: *collection0Name
+                filter: {}
+                sort: { time: 1 }
+              commandName: find
```
jmikola, line 128: I think we covered this in a previous PR, but `commandName` really servers no purpose in `expectEvents` and I regret blindly keeping it around from the legacy APM tests. The only necessary use of `commandName` is with `runCommand` operations.

I'd suggest removing this, as you don't have it for the other commands in this test.

jmikola (approved):
LGTM with `commandName` removed from APM expectations.

source/collection-management/tests/timeseries-collection.yml
benjirewis: Ah yep; since I'm matching the `command` already, removed `commandName` from this assertion.

PR #1004, opened 2021-05-28
DRIVERS-820: WriteError.details and clarify WriteError construction

jmikola:
https://jira.mongodb.org/browse/DRIVERS-820

### DIFF
--- a/source/crud/crud.rst
+++ b/source/crud/crud.rst
@@ -2020,7 +2020,7 @@ Q: Why are client-side errors raised when options are provided for unacknowledge
 Changes
 =======
 
-* 2021-05-27: Introduce WriteError.details and clarify how WriteError construction
+* 2021-06-02: Introduce WriteError.details and clarify WriteError construction
 * 2021-06-01: Add let to AggregateOptions
 * 2021-01-21: Update estimatedDocumentCount to use $collStats stage for servers >= 4.9
 * 2020-04-17: Specify that the driver must raise an error for unacknowledged hints on any write operation, regardless of server version.
--- a/source/crud/crud.rst
+++ b/source/crud/crud.rst
@@ -12,7 +12,7 @@ Driver CRUD API
 :Status: Approved
 :Type: Standards
 :Minimum Server Version: 2.6
-:Last Modified: May 27, 2021
+:Last Modified: June 2, 2021
 
 .. contents::
 
@@ -766,9 +766,9 @@ The server supports several collection-less aggregation source stages like ``$cu
 Write
 -----
 
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-Insert, Update, Delete, and Bulk Writes
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Insert, Update, Replace, Delete, and Bulk Writes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. code:: typescript
 
@@ -1501,7 +1501,7 @@ WriteConcernError
 Drivers MUST construct a ``WriteConcernError`` from a server reply as follows:
 
 - Set ``code`` to ``writeConcernError.code``.
-- Set ``message`` to ``writeConcernError.errmsg``.
+- Set ``message`` to ``writeConcernError.errmsg`` if available.
 - Set ``details`` to ``writeConcernError.errInfo`` if available. Drivers MUST NOT parse inside ``errInfo``.
 
 See the `Read/Write Concern specification </source/read-write-concern/read-write-concern.rst#writeconcernerror-examples>`_
--- a/source/crud/crud.rst
+++ b/source/crud/crud.rst
@@ -12,7 +12,7 @@ Driver CRUD API
 :Status: Approved
 :Type: Standards
 :Minimum Server Version: 2.6
-:Last Modified: June 1, 2021
+:Last Modified: May 27, 2021
 
 .. contents::
 
@@ -766,9 +766,9 @@ The server supports several collection-less aggregation source stages like ``$cu
 Write
 -----
 
-~~~~~
-Basic
-~~~~~
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Insert, Update, Delete, and Bulk Writes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 .. code:: typescript
 
@@ -1295,8 +1295,8 @@ Bulk Write Models
   }
 
 
-Results
-~~~~~~~
+Write Results
+~~~~~~~~~~~~~
 
 The acknowledged property is defined for languages/frameworks without a sufficient optional type. Hence, a driver may choose to return an Optional<BulkWriteResult> such that unacknowledged writes don't have a value and acknowledged writes do have a value.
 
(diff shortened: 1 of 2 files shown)


source/crud/crud.rst
```diff
@@ -1438,35 +1438,47 @@ Any result class with all parameters marked NOT REQUIRED is ultimately NOT REQUI
 
   }
 
-
+~~~~~~~~~~~~~~
```
jmikola, line 1441: While I was modifying WriteError, I realized that "Error Handling" applies just as much to `findAndModify` as it does to the "Basic" (insert/update/delete) section.

source/crud/crud.rst
```diff
@@ -1438,35 +1438,47 @@ Any result class with all parameters marked NOT REQUIRED is ultimately NOT REQUI
...
 Error Handling
 ~~~~~~~~~~~~~~
 
-Defined below are error and exception types that should be reported from the various write methods. Since error types across languages would be impossible to reconcile, the below definitions represent the fields and names for the information that should be present. Structure isn't important as long as the information is available.
+Defined below are error and exception types that should be reported from the
+various write methods. Since error types across languages would be impossible to
+reconcile, the below definitions represent the fields and names for the
+information that should be present. Structure isn't important as long as the
+information is available.
```
jmikola, line 1449: Just whitespace changes to apply word wrap.

source/crud/crud.rst
```diff
@@ -1438,35 +1438,47 @@ Any result class with all parameters marked NOT REQUIRED is ultimately NOT REQUI
...
 
-Drivers SHOULD report errors however they report other server errors: by raising an exception, returning "false", or another idiom that is consistent with other server errors.
+Drivers SHOULD report errors however they report other server errors: by raising
+an exception, returning "false" and populating an error struct, or another idiom
```
jmikola, line 1452: "returning false" didn't make sense on its own, as there's way to actually report an error in that case. I assumed this was referring to libmongoc, so I'm clarifying that it returns false _and_ populates an error struct.

source/crud/crud.rst
```diff
@@ -1476,25 +1488,58 @@ Drivers SHOULD report errors however they report other server errors: by raising
...
 Drivers MUST construct a ``WriteConcernError`` from a server reply as follows:
 
-- set ``code`` to ``writeConcernError.code``.
-- set ``message`` to ``writeConcernError.errmsg`` if available.
-- set ``details`` to ``writeConcernError.errInfo`` if available. Drivers MUST NOT parse inside ``errInfo``.
+- Set ``code`` to ``writeConcernError.code``.
+- Set ``message`` to ``writeConcernError.errmsg``.
+- Set ``details`` to ``writeConcernError.errInfo`` if available. Drivers MUST NOT parse inside ``errInfo``.
+
+See the `Read/Write Concern specification </source/read-write-concern/read-write-concern.rst#writeconcernerror-examples>`_
```
jmikola, line 1495: Changed the link fragment to a more relevant jump point in that spec.

source/crud/crud.rst
```diff
@@ -1476,25 +1488,58 @@ Drivers SHOULD report errors however they report other server errors: by raising
...
 
-See the `Read/Write Concern specification </source/read-write-concern/read-write-concern.rst#Errors>`_ for examples of how a server represents write concern errors in replies.
+For single-statement writes (i.e. ``insertOne``, ``updateOne``, ``updateMany``,
+``replaceOne``, ``deleteOne``, and ``deleteMany``), a single write error may be
+reported in the array and ``writeErrors[0].index`` will be zero.
+
+For multi-statement writes (i.e. ``insertMany`` and ``bulkWrite``), potentially
+many write errors may be reported in the array and the ``index`` property will
+be set accordingly. Since the reported ``index`` is specific to each command,
+drivers MUST adjust the index accordingly for ``BulkWriteError.index``.
```
jmikola, line 1517: It seemed worth clarifying how we construct WriteError objects. Also, the spec didn't say much about how the `index` is calculated. This is another gap in the CRUD spec where the actual details are likely in the deprecated [Bulk API spec](https://github.com/mongodb/specifications/blob/master/source/driver-bulk-update.rst).

source/crud/crud.rst
```diff
@@ -1476,25 +1488,58 @@ Drivers SHOULD report errors however they report other server errors: by raising
...
 
     /**
-     * A description of the error.
+     * A document providing more information about the write error (e.g. details
+     * pertaining to document validation). Corresponds to the
+     * "writeErrors[].errInfo" field in the command response.
+     *
+     * @see https://docs.mongodb.com/manual/reference/method/WriteResult/
+     */
+    details: Document;
```
jmikola, line 1538: This is the crucial spec change that allows drivers to expose the reason documents fail validation.

That said, there are no tests for this as the unified test format does not support assertions for specific error types beyond [`expectedError.expectResult`](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#expectederror), which is used for asserting the incomplete WriteResult within a BulkWriteException. That was implemented mainly because we already had legacy tests making those assertions.

I'm creating DRIVERS-TBD to track better assertions for WriteError(s) and WriteConcernError, but I don't think we need to complicate an otherwise trivial spec change this late into the release cycle with test runner changes. While something like `expectedError: { hasDetails: true }` could get the job done, I think it'd be hastily designed and likely something we'd come to regret.

----

As far as drivers actually testing this change, I intend to provide some guidance in the description of DRIVERS-820 to suggest that drivers should create a prose test that creates a collection with documentation validation and inserts an invalid document to check that the new error property gets set. My previous investigation in [DRIVERS-1643](https://jira.mongodb.org/browse/DRIVERS-1643) should be helpful for that.

alcaeus: Agree with not making more changes to the unified test format this late in the cycle. A single prose test to ensure the property gets populated should be enough until we can make the spec change to the test format.

jmikola: Ah, I didn't realize the CRUD spec had existing prose tests. I'll add something similar to the existing ["errInfo" is propagated](https://github.com/mongodb/specifications/tree/master/source/crud/tests#errinfo-is-propagated) test in a bit.

jmikola: New prose test in https://github.com/mongodb/specifications/pull/1004/commits/24177fac949c0f9ec732a7f587c84668f6eb6c2c

alcaeus (approved):

jyemin (approved):
LGTM

source/crud/crud.rst
```diff
@@ -1960,6 +2008,7 @@ Q: Why are client-side errors raised when options are provided for unacknowledge
 Changes
 =======
 
+* 2021-05-27: Introduce WriteError.details and clarify how WriteError construction
```
kevinAlbs, line 2011:
```suggestion
* 2021-05-27: Introduce WriteError.details and clarify WriteError construction
```

source/crud/tests/README.rst
```diff
@@ -239,4 +243,34 @@ Test that a writeConcernError "errInfo" is propagated to the user in whatever wa
...
+   {
+     "create": "test",
+     "validator": {
+       "x": { $type: "string" }
+     }
+   }
+
+Enable `command monitoring <../../command-monitoring/command-monitoring.rst>`_
+to observe CommandSucceededEvents. Then, insert an invalid document (e.g.
+``{x: 1}}``) and assert that a WriteError occurs, that its code is ``121``
```
kevinAlbs, line 272:
```suggestion
``{x: 1}``) and assert that a WriteError occurs, that its code is ``121``
```

kevinAlbs (approved):
LGTM! I left a few minor comments.

source/crud/crud.rst
```diff
@@ -1476,25 +1488,58 @@ Drivers SHOULD report errors however they report other server errors: by raising
 
 Drivers MUST construct a ``WriteConcernError`` from a server reply as follows:
 
-- set ``code`` to ``writeConcernError.code``.
-- set ``message`` to ``writeConcernError.errmsg`` if available.
-- set ``details`` to ``writeConcernError.errInfo`` if available. Drivers MUST NOT parse inside ``errInfo``.
+- Set ``code`` to ``writeConcernError.code``.
+- Set ``message`` to ``writeConcernError.errmsg``.
```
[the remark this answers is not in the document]
jmikola: I'll restore this, as the removal was unintentional on my part.

PR #1005, opened 2021-05-28
DRIVERS-1791 Fix load balancer spec test ConnectionClosedEvent reason

ShaneHarvey:
In this test a connection is closed because the insert command fails with a NotPrimary error so the ConnectionClosedEvent reason field should be "error" not "stale".

### DIFF
--- a/source/load-balancers/tests/sdam-error-handling.json
+++ b/source/load-balancers/tests/sdam-error-handling.json
@@ -241,7 +241,7 @@
             },
             {
               "connectionClosedEvent": {
-                "reason": "stale"
+                "reason": "error"
               }
             },
             {
--- a/source/load-balancers/tests/sdam-error-handling.yml
+++ b/source/load-balancers/tests/sdam-error-handling.yml
@@ -128,7 +128,7 @@ tests:
           - poolClearedEvent: {}
           - connectionCheckedInEvent: {}
           - connectionClosedEvent:
-              reason: stale
+              reason: error
           # Second insertOne.
           - connectionCheckedOutEvent: {}
           - connectionCheckedInEvent: {}


ShaneHarvey:
Hmm. Actually now that I'm reviewing this again I notice that the insert command is not failing with a network error, it's failing with an InterruptedAtShutdown error code. In PyMongo it just happens to be that network errors and NotPrimary-like errors closed the connection in the same way. I suspect that this behavior is non-standard between drivers.

What if we just remove the "reason" assertion here altogether?

jyemin:
@durran can you take first crack at this?  At first glance, it seems important that we specify the reason as "stale", but I would need to dig in.  

@divjotarora if you recall anything about this, please chime in.

ShaneHarvey:
I'd like to know as well. IMO both "error" and "stale" are valid reasons here. "error" is valid because the connection did experience an error, quoting CMAP:
>    *   - "error":           The Connection experienced an error, making it no longer valid

"stale" is also valid because the pool will be cleared after a InterruptedAtShutdown error.

divjotarora:
The Go Driver publishes `ConnectionClosedEvent` with `reason = error` only if the connection encounters a network error. Well-formed server errors that later result in the connection being cleared end up with `reason = stale` because we first go through the error handling code path, which bumps the pool's generation number, and then put the connection back into the pool, where it gets closed because its generation number is less than the pool's current generation (i.e. it thinks it's stale). I don't think the CMAP spec is quite clear here. This [section](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#connection-1) says:

> Errored: The Connection has experienced an error that indicates it is no longer recommended for use. Examples include, but are not limited to:
- Network Error
- Network Timeout
- Endpoint closing the connection
- Driver-Side Timeout
- Wire-Protocol Error

I'm not sure where errors like `ShutdownInProgress` fit in here. Are those considered "wire-protocol errors"?

ShaneHarvey:
Since both "error" and "stale" make sense and the behavior is somewhat unspecified how about we do this:
- create 4 connections (open and close 4 cursors)
- fail an insert with InterruptedAtShutdown
- Assert that 2 connections get closed.
- Omit the "reason" field on the first because it's ambiguous
- Add "reason: stale" to the second 

My only concern with this idea is that it may not be defined which connection gets closed first.

divjotarora:
@ShaneHarvey Could we instead remove the `reason` field from this test and file a DRIVERS ticket to better clarify the field in the CMAP spec? Part of the ticket could also be to add back the field here.

durran:
I think it makes sense to remove the `reason` field from the test for now and do what @ShaneHarvey suggests.

ShaneHarvey:
I'm now revisiting this after implementing more of the load balancer spec in Python. I've come to the conclusion that Python's old behavior is a bug. We shouldn't always close the connection when receiving a state change error and should instead rely on SDAM to clear the pool. I've updated python such that it passes this spec test as written.

Hopefully Python is the only driver with this behavior (it wasn't a conscious choice but a design quirk because our Network exception and State change exception classes inherit from the same base class) so I'm going to close this issue. If it happens that other drivers run into the same issue then we can revisit.

Note the reason I'm hesitant to make this change is that it will defeat the purpose of this test; that only connections to a given serviceId are cleared. The `reason: "stale"` indicates that the serviceId which experienced the error was cleared but the other connection was not.

PR #1007, opened 2021-06-01
DRIVERS-1653 Add tests for redactions in command monitoring

benjirewis:
DRIVERS-1653

Adds tests to verify that security-sensitive commands (as described [here](https://github.com/mongodb/specifications/blob/master/source/command-monitoring/command-monitoring.rst#security) in the command monitoring specification) are redacted from command monitoring.

### DIFF
--- a/source/command-monitoring/tests/README.rst
+++ b/source/command-monitoring/tests/README.rst
@@ -12,6 +12,9 @@ Command Monitoring
 Testing
 =======
 
+Tests in ``unified`` are implemented in the `Unified Test Format <../../unified-test-format/unified-test-format.rst>`__ and require
+schema version 1.0. Other tests should be run as described below.
+
 Tests are provided in YML and JSON format to assert proper upconversion of commands.
 
 Database and Collection Names
--- a/source/command-monitoring/tests/unified/redacted-commands.json
+++ b/source/command-monitoring/tests/unified/redacted-commands.json
@@ -337,7 +337,7 @@
       ]
     },
     {
-      "description": "hello with speculativeAuth",
+      "description": "hello with speculative authenticate",
       "operations": [
         {
           "name": "runCommand",
@@ -413,6 +413,72 @@
                   "isMaster": {
                     "$$exists": false
                   }
+                }
+              }
+            }
+          ]
+        }
+      ]
+    },
+    {
+      "description": "hello without speculative authenticate is not redacted",
+      "operations": [
+        {
+          "name": "runCommand",
+          "object": "database",
+          "arguments": {
+            "commandName": "hello",
+            "command": {
+              "hello": "public"
+            }
+          }
+        },
+        {
+          "name": "runCommand",
+          "object": "database",
+          "arguments": {
+            "commandName": "ismaster",
+            "command": {
+              "ismaster": "public"
+            }
+          }
+        },
+        {
+          "name": "runCommand",
+          "object": "database",
+          "arguments": {
+            "commandName": "isMaster",
+            "command": {
+              "isMaster": "public"
+            }
+          }
+        }
+      ],
+      "expectEvents": [
+        {
+          "client": "client",
+          "events": [
+            {
+              "commandStartedEvent": {
+                "commandName": "hello",
+                "command": {
+                  "hello": "public"
+                }
+              }
+            },
+            {
+              "commandStartedEvent": {
+                "commandName": "ismaster",
+                "command": {
+                  "ismaster": "public"
+                }
+              }
+            },
+            {
+              "commandStartedEvent": {
+                "commandName": "isMaster",
+                "command": {
+                  "isMaster": "public"
                 }
               }
             }
--- a/source/command-monitoring/tests/unified/redacted-commands.json
+++ b/source/command-monitoring/tests/unified/redacted-commands.json
@@ -1,5 +1,5 @@
 {
-  "description": "runCommand",
+  "description": "redacted-commands",
   "schemaVersion": "1.0",
   "runOnRequirements": [
     {
@@ -25,7 +25,7 @@
   ],
   "tests": [
     {
-      "description": "authenticate is redacted",
+      "description": "authenticate",
       "operations": [
         {
           "name": "runCommand",
@@ -33,7 +33,6 @@
           "arguments": {
             "commandName": "authenticate",
             "command": {
-              "otherField": "public",
               "authenticate": "private"
             }
           },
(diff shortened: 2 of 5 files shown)


source/command-monitoring/tests/unified/redacted-commands.yml
```diff
@@ -0,0 +1,166 @@
...
+      - name: runCommand
+        object: *database
+        arguments:
+          commandName: authenticate
+          command:
+            authenticate: "private"
+        # Malformed authentication commands will fail against the server, but we
+        # just want to assert that security-related commands are redacted
+        # in command monitoring.
+        expectError:
```
benjirewis, line 30: I thought it would be easier to "fake" most of these authentication related commands and just expect the server error since most of them are malformed (missing fields like `mechanism`, "private" is an invalid value, or the command cannot be run out of order as in `saslContinue`).

I _could_ make sure each command is run with the correct context, so they all can succeed, but we're not really testing authentication in drivers here, just command monitoring redaction.

source/command-monitoring/tests/unified/redacted-commands.yml
```diff
@@ -0,0 +1,166 @@
...
+          isError: true
+    expectEvents: &redacted-event
+      - client: *client
+        events:
+          # Expect empty (redacted) event.
+          - commandStartedEvent: {}
```
jmikola, line 36: [Security](https://github.com/mongodb/specifications/blob/master/source/command-monitoring/command-monitoring.rst#security) in the spec states that the command and/or reply documents are supposed to be redacted. In this case, I think it makes sense to have `commandStartedEvent` use `commandName` to assert a particular command was observed and _then_ use `$$exists: false` on a field within the command/reply body that would otherwise be expected. For instance:

```
- commandStartedEvent:
    commandName: redactedCommandName
    command: { redactedCommandName: { $$exists: false } }
```

source/command-monitoring/tests/unified/redacted-commands.yml
jyemin: I agree.

jyemin (commented):
Waiting on changes asked for by Jeremy.

source/command-monitoring/tests/unified/redacted-commands.yml
benjirewis: That makes sense, thanks for the explanation here and offline! Changed the expected event to be
```
- commandStartedEvent:
    commandName: redactedCommandName
    command: { redactedCommandName: { $$exists: false } }
```
And left a comment similar to the one above. I agree that it's unlikely a driver would redact only the command name field from the published `command`. I _could_ add further `{ $$exists: false }` assertions for other fields like `mechanism` and `payload` if that level of detail seems important to others.

source/command-monitoring/tests/unified/redacted-commands.json
```diff
@@ -0,0 +1,424 @@
...
+                  }
+                }
+              }
+            }
+          ]
+        }
+      ]
+    },
+    {
+      "description": "hello with speculativeAuth",
```
jyemin, line 340: Let's add one more test that asserts hello/legacy hello _without_ speculativeAuthenticate is _not_ redacted.

jyemin, line 340: Nit: change `speculativeAuth` to either `speculativeAuthenticate" or "speculative authenticate"

jyemin (changes requested):

source/command-monitoring/tests/unified/redacted-commands.json
benjirewis: Added `hello without speculative authenticate is not redacted`.

benjirewis: Changed to "speculative authenticate" here and in new test.

jyemin (approved):
LGTM

source/command-monitoring/tests/unified/redacted-commands.yml
kmahar: for the `command` assertion, I assume we need to do it that way (with `$$exists`) because commands are considered [root documents](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#allowing-extra-fields-in-root-level-documents) and allowed to have extra fields?

this seems fine to me, as I think adding to the format just for this to support e.g. an empty doc check would probably be overkill. but I don't think it's necessarily obvious that the test is only written in this less optimal way because of that restriction. perhaps worth mentioning in the comment as well?

> I _could_ add further `{ $$exists: false }` assertions for other fields like `mechanism` and `payload` if that level of detail seems important to others

I agree with @jmikola that it's unlikely someone would remove the command name but omit the other fields, so IMO this isn't needed

benjirewis: Yep, I think that section about root documents is why we can't just say `command: {}`. I expanded on the comment to explain that restriction. I think @jyemin agreed offline that adding to the format for this situation would not have much benefit for us in the future.

source/command-monitoring/tests/README.rst
```diff
@@ -12,6 +12,9 @@ Command Monitoring
 Testing
 =======
 
+Tests in ``unified`` are implemented in the `Unified Test Format <../../unified-test-format/unified-test-format.rst>`__ and require
```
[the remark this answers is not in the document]
benjirewis: Yep! I can do this in a separate commit (move old tests to `legacy` and update the test README to reflect the new location).

kmahar (approved):
lgtm! nice work

source/command-monitoring/tests/unified/redacted-commands.yml
```diff
@@ -213,3 +213,36 @@ tests:
...
               command: { isMaster: { $$exists: false } }
+
+  - description: "hello without speculative authenticate is not redacted"
+    operations:
+      - name: runCommand
+        object: *database
+        arguments:
+          commandName: hello
+          command:
+            hello: "public"
```
jmikola, line 226: IIRC, the server never checks the _value_ of these fields; however, I'm not sure why we'd use `"public"` here when the documented value is `1`.

 * https://docs.mongodb.com/manual/reference/command/hello/
 * https://docs.mongodb.com/manual/reference/command/isMaster/

benjirewis: Yeah I believe most of these authentication-related commands use `1` as a value normally. I used `public` just to emphasize that `hello`'s value will be visible in the published commands (as opposed to `private` in the redacted tests above).

jmikola: I just noticed you were using `private` in other tests when syncing these with PHP. SGTM.

PR #1009, opened 2021-06-01
DRIVERS-1779 Add granularity option to time-series tests

benjirewis:
DRIVERS-1779

Adds the `granularity` option to the `createCollection` operations for time-series tests to ensure that drivers can support non-default granularities (like "minutes").


kevinAlbs (approved):
LGTM!

jmikola (approved):

PR #1011, opened 2021-06-04
DRIVERS-1237: Do not rely on dotted field for client-side error

jmikola:
https://jira.mongodb.org/browse/DRIVERS-1237

Since dots/dollars are no longer prohibited, these tests can use update/replace validation to yield a client-side error.


jyemin (approved):
All passing.  LGTM!

iwysiu (approved):

PR #1012, opened 2021-06-04
DRIVERS-1173: Remove DBRef parse error tests

jmikola:
https://jira.mongodb.org/browse/DRIVERS-1173

These were missed in the original PR (#996). Invalid DBRef documents are no longer parse errors but should be left as-is.

### DIFF
--- a/source/bson-corpus/tests/top.json
+++ b/source/bson-corpus/tests/top.json
@@ -215,14 +215,6 @@
             "string": "{\"a\" : {\"$date\" : {\"$numberLong\" : \"1356351330501\"}, \"unrelated\": true}}"
         },
         {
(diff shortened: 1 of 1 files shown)


jyemin (approved):
LGTM.

I note that Java driver's bson library, where these tests are implemented, has no notion of DBRef.  That is a concept introduced in the driver library.  So these tests are not useful for asserting correct DBRef behavior.  I wonder if other drivers are similar.

jmikola:
> I wonder if other drivers are similar.

FWIW, libmongoc's extended JSON parser _did_ perform validation of DBRef documents, despite no DBRef type/model formally existing elsewhere.

PR #1014, opened 2021-06-07
DRIVERS-776: Add tests for aggregate with $out using let option

jyemin:
For historical reasons, some drivers have different code paths for aggregating to a collection.

### DIFF
--- a/source/crud/tests/unified/aggregate-let.yml
+++ b/source/crud/tests/unified/aggregate-let.yml
@@ -14,12 +14,19 @@ createEntities:
       id: &collection0 collection0
       database: *database0
       collectionName: &collection0Name coll0
+  - collection:
+      id: &collection1 collection1
+      database: *database0
+      collectionName: &collection1Name coll1
 
 initialData: &initialData
   - collectionName: *collection0Name
     databaseName: *database0Name
     documents:
       - { _id: 1 }
+  - collectionName: *collection1Name
+    databaseName: *database0Name
+    documents: [ ]
 
 tests:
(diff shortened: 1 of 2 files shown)


source/crud/tests/unified/aggregate-let.yml
```diff
@@ -115,3 +122,52 @@ tests:
...
+    runOnRequirements:
+      - minServerVersion: "5.0"
+    operations:
+      - name: aggregate
+        object: *collection0
+        arguments:
+          pipeline: &pipeline2
+            - $match: { $expr: { $eq: ["$_id", "$$id"] } }
+            - $project: { _id: 1 }
+            - $out: *collection1Name
```
jmikola, line 136: @jyemin IIRC, you originally hit the error from [SERVER-57403](https://jira.mongodb.org/browse/SERVER-57403) using `$merge`. Is it at all relevant that this test is using `$out` instead?

I'm also curious because [SERVER-46998](https://jira.mongodb.org/browse/SERVER-46998) doesn't mention `$out` (only `$merge`). Perhaps this doesn't matter at all since the purpose of this test is to exercise the _driver's_ code path(s).

jmikola (approved):
Tested locally with PHPLIB.

One question about use of `$out` vs. `$merge`, but LGTM.

source/crud/tests/unified/aggregate-let.yml
jyemin: Yes, that's right.  I'm just trying to exercise the driver here.

PR #1016, opened 2021-06-09
DRIVERS-1804 Connection pools must be created and eventually marked ready for any server if a direct connection is used

stIncMale:
DRIVERS-1804

### DIFF
--- a/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
+++ b/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
@@ -1703,7 +1703,7 @@ must do no I/O::
 
         # for drivers that implement CMAP, mark the connection pool as ready after a successful check
         if (server.type in (Mongos, RSPrimary, RSSecondary, Standalone, LoadBalanced))
-                or (server.type != Unknown and directConnection):
+                or (server.type != Unknown and newTopologyDescription.type == Single):
             pool.ready()
 
         take any additional actions,
--- a/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
+++ b/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
@@ -8,8 +8,8 @@ Server Discovery And Monitoring
 :Advisors: David Golden, Craig Wilson
 :Status: Accepted
 :Type: Standards
-:Version: 2.31
-:Last Modified: 2021-05-03
+:Version: 2.32
+:Last Modified: 2021-06-09
 
 .. contents::
 
@@ -1220,8 +1220,11 @@ Connection Pool Management
 ''''''''''''''''''''''''''
 
 For drivers that support connection pools, after a server check is
-complete, if the server is determined to be `data-bearing
-<https://github.com/mongodb/specifications/blob/masterserver-discovery-and-monitoring.rst#data-bearing-server-type>`_
+completed successfully, if the server is determined to be
+`data-bearing <https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#data-bearing-server-type>`_
+or a
+`direct connection <https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#general-requirements>`__
+to the server is requested,
 and does not already have a connection pool, the driver MUST create
 the connection pool for the server. Additionally, if a driver
 implements a CMAP compliant connection pool, the server's pool (even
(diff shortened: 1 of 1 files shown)


source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1698,9 +1701,9 @@ must do no I/O::
         address = server.address
         newTopologyDescription.servers[address] = server
 
-        # for drivers that implement CMAP, mark the connection pool as ready after
-        # a successful check on a data bearing server.
-        if server.type in (Mongos, RSPrimary, RSSecondary, Standalone, LoadBalanced):
+        # for drivers that implement CMAP, mark the connection pool as ready after a successful check
+        if (server.type in (Mongos, RSPrimary, RSSecondary, Standalone, LoadBalanced))
+                or (server.type != Unknown and directConnection):
```
ShaneHarvey, line 1706: directConnection -> "newTopologyDescription.type == Single"

stIncMale: `newTopologyDescription.type == Single` only if `server.type` is `Standalone` and is covered by the spec in its state even before the PR . My understanding is that `directConnection` may be requested for any deployment (not only standalone), and in that case we must create + unpause a pool regardless of the discovered server type - this is what this PR tries to achieve.

ShaneHarvey: > newTopologyDescription.type == Single only if server.type is Standalone.

"TopologyType==single" is not the same as "server.type==Standalone". With TopologyType == single, the single server can be of _any_ type.

stIncMale: > With TopologyType == single, the single server can be of any type.

It appears to me that the spec says otherwise. When I look at https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#topologytype-table, I see that `updateUnknownWithStandalone` is called only if `server.type==Standalone"`. [`updateUnknownWithStandalone`](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#updateunknownwithstandalone) sets `topologyDescription.type = Single` and as far as I can see there are no other ways to have `newTopologyDescription.type == Single`.

stIncMale: On the other hand, there is also this: _"[TopologyType Single](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#topologytype-single). The TopologyDescription's type was initialized as Single and remains Single forever. "_

Does the Python driver sometimes sets `TopologyType Single` at the very beginning, i.e., does not start with topology type `Unknown` without user requesting `directConnection`? If yes, I am curious when this happen.

ShaneHarvey (approved):
LGTM!

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1220,8 +1220,11 @@ Connection Pool Management
 ''''''''''''''''''''''''''
 
 For drivers that support connection pools, after a server check is
-complete, if the server is determined to be `data-bearing
-<https://github.com/mongodb/specifications/blob/masterserver-discovery-and-monitoring.rst#data-bearing-server-type>`_
+completed successfully, if the server is determined to be
+`data-bearing <https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#data-bearing-server-type>`_
+or a
+`direct connection <https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#general-requirements>`__
+to the server is requested,
```
BorisDog, line 1227: Should this be TopologyType == Single instead of direct connection?

stIncMale: The specified [link](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#general-requirements) says the following: _"Direct connections: A client MUST be able to connect to a single server of any type. This includes querying hidden replica set members, and connecting to uninitialized members (see RSGhost) in order to run "replSetInitiate"._ As was pointed by others, specifying [`directConnection` in the connection string](https://github.com/mongodb/specifications/blob/master/source/uri-options/uri-options.rst#srv-uri-with-directconnection-uri-option) is just one of possible ways to fulfill the "direct connection" requirement, and this is why the pseudocode check was changed. However, the actual intent is to create/unpause pools when "direct connection" is requested (regardless of how this may be done in drivers, and how it is better to express the check in the pseudocode), so I think it is important to leave the actual intent clearly specified as a plaintext requirement.

BorisDog: Thanks for the walkthrough!

BorisDog (approved):

PR #1017, opened 2021-06-10
DRIVERS-1789: Add guidance on concurrent cursor ops

durran:
Drivers in load balanced mode must either document this is not allowed or explicitly prevent it.


ShaneHarvey (approved):

jyemin (approved):
LGTM

PR #1018, opened 2021-06-11
DRIVERS-1806 Increase `waitQueueTimeoutMS` in `wait-queue-timeout.yml`, `wait-queue-timeouts.yml`

stIncMale:
This change reduces the probability of timing out accidentally

DRIVERS-1806


ShaneHarvey (approved):
I haven't implemented the load balancer wait-queue-timeouts.json test yet but the CMAP test passes. I'd expect the LB one to pass too.

BorisDog (approved):

PR #1022, opened 2021-06-16
DRIVERS-1607: Snapshot reads on Secondaries

BorisDog:


source/sessions/tests/unified/snapshot-sessions.json
```diff
@@ -0,0 +1,706 @@
+{
```
BorisDog, line 1: `snapshot-sessions.yml` will be added after json test version will be finalized.

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,358 @@
...
+Session
+    A session is an abstract concept that represents a set of sequential
+    operations executed by an application that are related in some way. This
+    specification defines how sessions are used to implement snapshot reads.
+
+Snapshot reads
+    Reads with readconcern level ``snapshot`` that occur outside of transactions on
+    both the primary and secondary nodes, including in sharded clusters.
+
+Snapshot cluster time
```
ajdavis, line 67: How about "Snapshot timestamp"? It keeps confusing everyone that the param is called "atClusterTime" but its type is just Timestamp, not the signed cluster time document that we usually call a "cluster time". So I propose calling this a timestamp.

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,358 @@
...
+---------
+
+Applications set ``isSnapshot`` when starting a session to
+indicate whether they want snapshot reads.
+
+Note that the ``isSnapshot`` property is optional. The default value of
+this property is false.
+
+Snapshot reads and causal consistency are mutually exclusive. Therefore if ``isSnapshot`` is set to true,
+``causalConsistency`` property is set to false. Client MUST throw an Error if both ``isSnapshot`` and ``causalConsistency`` are set to true.
```
ajdavis, line 138: The server permits a snapshot read to **begin** with afterClusterTime so it's causally after some previous operation. Within a snapshot read, of course, you can't be causally consistent. This API doesn't expose the "begin a snapshot read causally after an operation" feature. That might be fine, and we could enhance the API later if it turns out that users want that feature. I want to confirm it's our intention to omit the feature now.

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,358 @@
...
+
+ClientSession changes
+=====================
+
+``ClientSession`` changes summary
+
+.. code:: typescript
+
+    interface ClientSession {
+        Optional<BsonTimestamp> snapshotClusterTime;
```
ajdavis, line 149: "snapshotTimestamp"? I now forget whether we've discussed the user-facing terminology. It could be "snapshotClusterTime" to discourage users from providing their own timestamps (which might not work), but "snapshotTimestamp" seems more understandable and accurate to me.

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,358 @@
...
+
+        // other members as defined in other specs
+    }
+
+Each new member is documented below.
+
+snapshotClusterTime
+-------------------
+
+This property returns the cluster time of the first find/aggregate operation performed
```
ajdavis, line 159: Regardless of the property name, let's call it a timestamp. "This property returns the timestamp of the first...."

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,358 @@
...
+There are no additional API changes to ``MongoCollection`` beyond those specified
+in the Sessions Specification. All ``MongoCollection`` methods that talk to the
+server have been overloaded to take a session parameter. If that session was
+started with ``isSnapshot = true`` then all operations using that
+session will share the same snapshot.
+
+ReadConcern changes
+===================
+
+``shapshot`` added to `ReadConcernLevel enumeration <../read-write-concern/read-write-concern.rst#read-concern>`_.`.
```
ajdavis, line 184: Typo, "shapshot"

ajdavis (changes requested):
I'm so happy that we're choosing the ClientSession API!

source/sessions/snapshot-sessions.rst
BorisDog: Thanks, `Snapshot timestamp` sounds more natural.

source/sessions/snapshot-sessions.rst
BorisDog: Done.

source/sessions/snapshot-sessions.rst
BorisDog: Done.

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,358 @@
...
+1.  | The first read in a snapshot session must not send ``atClusterTime``
+    | to the server (because the ``atClusterTime`` has not yet been determined)
+
+    * ``session = client.startSession(isSnapshot = true)``
+    * ``document = collection.anyReadOperation(session, ...)``
+    * capture the command sent to the server (using APM or other mechanism).
+    * assert that the command does not have an ``atClusterTime``.
+
+2.  | Subsequent snapshot reads on a ``ClientSession`` should read from the snapshot of the first read in that session.
+    | Test SHOULD introduce two variations one for primaries and additional for secondary reads
```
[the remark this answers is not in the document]
jyemin: I think that would be more a test of the server than of drivers, so I don't think it's necessary.  What do you think, @ajdavis?

source/sessions/tests/unified/snapshot-sessions.json
```diff
@@ -0,0 +1,706 @@
...
+  "description": "snapshot-sessions",
+  "schemaVersion": "1.0",
+  "runOnRequirements": [
+    {
+      "minServerVersion": "4.9.0",
```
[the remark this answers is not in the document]
jyemin: But no one needs to, so make it 5.0 to avoid confusion for people that don't understand our wacky versioning scheme (which is changing after 5.0 anyway)

source/sessions/snapshot-sessions.rst
jyemin: Confirmed.

source/sessions/tests/unified/snapshot-sessions.json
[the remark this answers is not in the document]
BorisDog: Done.

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,358 @@
...
+``isSnapshot`` field in the options passed to the ``startSession`` method.
+
+SessionOptions changes
+======================
+
+``SessionOptions`` change summary
+
+.. code:: typescript
+
+    class SessionOptions {
```
jyemin, line 114: To support both multi-threaded and multi-process use cases, do we need to make snapshotTimestamp a session option as well?  If so, we should document (for now) that the option should only be set to a value that was read from `ClientSession.snapshotTimestamp`

CC @ajdavis

source/sessions/snapshot-sessions.rst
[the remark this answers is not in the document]
ajdavis: Yeah, this is tested on the server in snapshot_reads.js, no need for drivers to do it.

ajdavis (approved):

source/read-write-concern/read-write-concern.rst
```diff
@@ -12,8 +12,8 @@ Read and Write Concern
 :Status: Approved
 :Type: Standards
 :Server Versions: 2.4+
-:Last Modified: 2021-04-07
-:Version: 1.5.5
+:Last Modified: 2021-06-15
+:Version: 1.5.6
```
alcaeus, line 16: This should be bumped to 1.6 as we're adding a new read concern mode.

source/read-write-concern/tests/README.rst
```diff
@@ -1,5 +1,5 @@
 =======================
```
alcaeus, line 1: Nit:
```suggestion
============================
```

source/read-write-concern/tests/README.rst
```diff
@@ -1,5 +1,5 @@
...
-Connection String Tests
+Read and Write Concern Tests
 =======================
```
alcaeus, line 3: Nit:
```suggestion
============================
```

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,358 @@
+================================
+Snapshot Reads Specification
+================================
```
alcaeus, line 3: Nit:
```suggestion
============================
Snapshot Reads Specification
============================
```

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,358 @@
...
+
+:Spec Title: Snapshot Reads Specification (See the registry of specs)
```
alcaeus, line 5: I don't think we need to call out a registry of specs here
```suggestion
:Spec Title: Snapshot Reads Specification
```

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,358 @@
...
+:Spec Version: 1.0
+:Author: Boris Dogadov
+:Advisors: Jeff Yemin, A. Jesse Jiryu Davis, Judah Schvimer
+:Status: Draft (Could be Draft, Accepted, Rejected, Final, or Replaced)
```
alcaeus, line 9: Reminder to change this to "Accepted" before merging.

```suggestion
:Status: Accepted
```

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,358 @@
...
+:Type: Standards
+:Minimum Server Version: 5.0 (The minimum server version this spec applies to)
```
alcaeus, line 11: Nit:
```suggestion
:Minimum Server Version: 5.0
```

source/sessions/tests/unified/snapshot-sessions.json
```diff
@@ -0,0 +1,706 @@
...
+          "expectResult": [
+            {
+              "_id": 1,
+              "x": 13
+            }
+          ]
+        }
+      ],
+      "outcome": [
+      ]
```
alcaeus, line 270: This empty list is not necessary.
```suggestion
      ]
```

alcaeus (changes requested):

source/sessions/snapshot-sessions.rst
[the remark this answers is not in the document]
BorisDog: Thanks

source/sessions/tests/unified/snapshot-sessions.json
BorisDog: Thanks, removed.

source/sessions/tests/unified/snapshot-sessions.json
```diff
@@ -0,0 +1,706 @@
...
+            "session": "session_distinct1"
+          },
+          "expectResult": [
+            11,
+            12
+          ]
+        }
+      ],
+      "outcome": [
+      ]
```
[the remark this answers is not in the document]
BorisDog: Thanks, removed.

source/sessions/tests/unified/snapshot-sessions.json
```diff
@@ -0,0 +1,706 @@
...
+          "expectResult": [
+            {
+              "_id": 3,
+              "x": 12
+            }
+          ]
+        }
+      ],
+      "outcome": [
+      ]
```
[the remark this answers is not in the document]
BorisDog: Thanks, removed.

source/sessions/tests/unified/snapshot-sessions.json
```diff
@@ -0,0 +1,706 @@
...
+            },
+            "session": "session_mixed"
+          },
+          "expectResult": [
+            11
+          ]
+        }
+      ],
+      "outcome": [
+      ]
```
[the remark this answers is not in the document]
BorisDog: Thanks, removed.

source/sessions/snapshot-sessions.rst
BorisDog: Done

source/sessions/snapshot-sessions.rst
BorisDog: Not sure what's the convention here as this referral appears in `causal-consistency.rst` and `driver-sessions.rst`.
Removed for now.

source/sessions/snapshot-sessions.rst
BorisDog: Thanks.

source/read-write-concern/tests/README.rst
BorisDog: Thanks.

source/read-write-concern/tests/README.rst
BorisDog: Thanks.

source/read-write-concern/read-write-concern.rst
BorisDog: Done.

source/sessions/snapshot-sessions.rst
BorisDog: @jyemin  Do we want to allow setting `snapshotTimestamp` now, or later in a different scope?
If so, that would require additional tests.

source/sessions/snapshot-sessions.rst
BorisDog: @jyemin @ajdavis
Making sure that we want drivers forbid requesting both ``causalConsistency`` and ``isSnapshot`` (need to add a test for that).
I lean to raising an error in such case in driver, as otherwise due to drivers possible implementation differences it could be that some would send both `atClusterTime` and `afterClusterTime`, and some only one of the fields. Which will result in inconsistent behavior.

jyemin: Let's start with making it an error.  It's easier to go from error to not error than the other way around.

source/sessions/snapshot-sessions.rst
jyemin: I went back to look at the google discussion doc, and it doesn't seem like we decided one way or the other.

I guess let's leave it out and consider it in a future spec update. 

CC @ajdavis

source/sessions/snapshot-sessions.rst
alcaeus: I believe those are leftovers from the templates. I'll give the specs a quick look to clean these up. Thanks for pointing out other instances.

source/unified-test-format/unified-test-format.rst
```diff
@@ -624,6 +624,7 @@ The structure of this object is as follows:
     specifications:
 
     - `Causal Consistency <../causal-consistency/causal-consistency.rst#sessionoptions-changes>`__
+    - `Snapshot Reads <../sessions/snapshot-sessions.rst#sessionoptions-changes>`__
```
alcaeus, line 627: I forgot to mention this, but you should add a make target for session tests in the [unified test runner Makefile](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/tests/Makefile). This ensures all unified session tests are checked against the schema during the CI build.

BorisDog: Great, thanks! Added, and will verify when yml tests are added.

source/sessions/tests/unified/snapshot-sessions.json
```diff
@@ -0,0 +1,698 @@
+{
+  "description": "snapshot-sessions",
+  "schemaVersion": "1.0",
+  "runOnRequirements": [
+    {
+      "minServerVersion": "5.0",
+      "topologies": [
+        "sharded-replicaset"
```
jyemin, line 8: I think this should be:

```topologies: [ "replicaset", "sharded-replicaset" ]```

**Review by jyemin — CHANGES_REQUESTED** [review]

File: source/sessions/tests/unified/snapshot-sessions.json [hunk 23]
```diff
@@ -0,0 +1,698 @@
...
+        }
+      }
+    },
+    {
+      "session": {
+        "id": "session_invalid_snapshot",
+        "client": "client0",
+        "sessionOptions": {
+          "isSnapshot": true,
+          "snapshotTimestamp": 0
```
jyemin, line 118: I'm not clear what your intention is here.  In the spec this option is not defined.

source/sessions/tests/unified/snapshot-sessions.json
```diff
@@ -0,0 +1,698 @@
...
+      ],
+      "expectEvents": [
+        {
+          "client": "client0",
+          "events": [
+            {
+              "commandStartedEvent": {
+                "command": {
+                  "find": "collection0",
+                  "readConcern": {
```
jyemin, line 685: Let's add commandStartedEvent assertions to all the tests, demonstrating that atClusterTime is included when appropriate.

source/sessions/tests/unified/snapshot-sessions.json
```diff
@@ -0,0 +1,698 @@
...
+        "collectionName": "collection0",
+        "collectionOptions": {
+          "readPreference": {
+            "mode": "secondary"
+          }
+        }
+      }
+    },
+    {
+      "session": {
```
jyemin, line 41: Is there really a need for all these session objects?  Each test runs in isolation, so a new session instance is created for each test.  So there is no re-use issues between tests.  I suspect we can have just one session with an isSnapshot: true option.

jyemin (changes requested):

source/sessions/tests/unified/snapshot-sessions.json
BorisDog: [I've marked this test as optional](https://github.com/mongodb/specifications/pull/1022/files#diff-6c9d82f6b269142ff8fe9f0ef8d7c90f9e2761c947b5caf92e6820124b81fc2bR299-R300), for drivers that can alter `snapshotTimestamp` field internally.

Not sure what's the best approach here: Mandate drivers to provide functionality to set `snapshotTimestamp` or extract this test as optional prose test.

source/sessions/tests/unified/snapshot-sessions.json
BorisDog: No need, removed.

source/sessions/tests/unified/snapshot-sessions.json
BorisDog: That's problematic as we don't have a way to obtain actual `atClusterTime`.

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,360 @@
...
+
+ClientSession changes
+=====================
+
+``ClientSession`` changes summary
+
+.. code:: typescript
+
+    interface ClientSession {
+        Optional<BsonTimestamp> snapshotTimestamp;
```
jyemin, line 149: Now I think about it, it's not clear why we need this to be a public property if there is no way to initialize a new ClientSession with its value.

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,351 @@
...
+        }
+    }
+
+Lists of commands that support snapshot reads:
+
+1. find
+2. aggregate
+3. distinct
+
+Snapshot read commands errors
```
jyemin, line 275: I'm not convinced its necessary to specify this, since retryable reads are only enabled via an allow-list anyway.

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,351 @@
...
+=============================
+
+Drivers MUST NOT retry errors in SnapshotError category for snapshot reads if ``atClusterTime`` is supplied.
+
+Test Plan
```
jyemin, line 280: Check with others, but I think this whole section should move to the README.rst in the tests subdirectory.

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,351 @@
...
+=========
+
+Note: some tests are only relevant to certain deployments. For the purpose of deciding
+which tests to run assume that any deployment that is version 5.0 or higher and is either a
+replica set or a sharded cluster supports snapshot reads.
+The server ``minSnapshotHistoryWindowInSeconds`` parameter SHOULD be configured to match the test execution time.
+
+1.  | The first read in a snapshot session must not send ``atClusterTime``
```
jyemin, line 288: This only needs to be included for prose tests.  You don't have to duplicate what's already specified in YAML.

jyemin (changes requested):

source/sessions/snapshot-sessions.rst
ajdavis: Sounds good to me.

source/sessions/snapshot-sessions.rst
BorisDog: Thanks, removed.

source/read-write-concern/read-write-concern.rst
```diff
@@ -131,6 +136,14 @@ considered the server’s default ``ReadConcern``.
 default ``ReadConcern`` while the latter is the user explicitly specifying a
 ``ReadConcern`` with a ``level`` of “local”.
 
+Snapshot Read Concern
+---------------------
+
+When a ``ReadConcern`` ``level`` ``snapshot`` is used, ``atClusterTime`` may be specified to indicate
+the desired point in time for reading. ``find`` and ``aggregate`` operations executed with ``ReadConcern`` ``snapshot`` but without ``atClusterTime``
```
benjirewis, line 143: I assume `distinct` is not mentioned here because it does not have a `cursor` field in the reply and `atClusterTime` would just be in the top level?

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,339 @@
...
+.. contents::
+
+--------
+
+Abstract
+========
+
+Version 5.0 of the server introduces support for readConcern level "snapshot" (non-speculative)
+for read commands outside of transactions, including on secondaries.
+This spec builds upon the Sessions Specification to define how an application
```
benjirewis, line 23: Could do a relative link to the [driver-sessions](https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst) specification here.

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,339 @@
...
+
+Applications will start a new client session for snapshot reads like
+this:
+
+.. code:: typescript
+
+    options = new SessionOptions(isSnapshot = true);
+    session = client.startSession(options);
+
+All read operations performed using this session will be read from same snapshot.
```
benjirewis, line 95:
```suggestion
All read operations performed using this session will be read from the same snapshot.
```

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,339 @@
...
+======================
+
+``SessionOptions`` change summary
+
+.. code:: typescript
+
+    class SessionOptions {
+        Optional<bool> isSnapshot;
+
+        // other options defined by other specs
```
benjirewis, line 117: Is this to leave room for a potential `snapshotTimestamp` option in the future? Why do we need this comment?

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,339 @@
...
+In order to support snapshot reads a new property named
+``isSnapshot`` is added to ``SessionOptions``. Applications set
+``isSnapshot`` when starting a client session to indicate
+whether they want snapshot reads. All read operations performed
+using that client session will share the same snapshot.
+
+Each new member is documented below.
+
+isSnapshot
+---------
```
benjirewis, line 129:
```suggestion
----------
```

source/sessions/snapshot-sessions.rst
benjirewis: Just FYI, libmongoc defines `mongoc_session_opt_t` with a field `flags` that is an enum [`mongoc_session_flag_t`](https://github.com/mongodb/mongo-c-driver/blob/master/src/libmongoc/src/mongoc/mongoc-client-session-private.h#L41-L44). I'll add another variant for `isSnapshot`, and by nature of the enum, there will be no way to specify `causalConsistency` and `isSnapshot` simultaneously. I'll still throw an error if users try to, but worth noting that if other drivers use a similar enum, users won't be able to do this anyways.

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,339 @@
...
+
+``snapshot`` added to `ReadConcernLevel enumeration <../read-write-concern/read-write-concern.rst#read-concern>`_.`.
+
+Server Commands
+===============
+
+There are no new server commands related to snapshot reads. Instead,
+snapshot reads are implemented by:
+
+1. Saving the ``atClusterTime`` returned by 5.0+ servers for the first find/aggregate operation in a
```
benjirewis, line 170: What about `distinct`? Or, can only `find` or `aggregate` be the first read operation to fetch and set the `atClusterTime`?

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,339 @@
...
+
+Server Errors
+=============
+1. The server may reply to read commands with a ``SnapshotTooOld`` error if the client's ``atClusterTime`` value is not available in the server's history.
+2. The server will return ``InvalidOptions`` error if both ``atClusterTime`` and ``afterClusterTime`` options are set to true.
+
+Snapshot read commands
+======================
+
+For snapshot reads the driver MUST first obtain ``atClusterTime`` from cursor response of find/aggregate command,
```
benjirewis, line 216: Same question about `distinct` here. Can only `find` or `aggregate` fetch the first `atClusterTime` to use in subsequent read operations?

source/sessions/snapshot-sessions.rst
benjirewis: I agree; seems like the normal place for prose tests.

benjirewis (commented):
Looking great! Have some comments now, and I might add some more when I finish the C implementation ([CDRIVER-4034](https://jira.mongodb.org/browse/CDRIVER-4034)), particularly with reference to the new prose and spec tests.

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,339 @@
...
+
+.. code:: typescript
+
+    {
+        find : <string>, // or other read command
+        ... // the rest of the command parameters
+        readConcern :
+        {
+            level : "snapshot",
+            afterClusterTime : <BsonTimestamp>
```
benjirewis, line 243: Should this be `atClusterTime`?

source/sessions/tests/unified/snapshot-sessions.json
```diff
@@ -0,0 +1,926 @@
...
+        "id": "session0",
+        "client": "client0",
+        "sessionOptions": {
+          "isSnapshot": true
```
benjirewis, line 51: Can we specify `"causalConsistency": false` here and on `session1`? `causalConsistency` is [true by default ](https://github.com/mongodb/specifications/blob/master/source/causal-consistency/causal-consistency.rst#causalconsistency) which will cause an error with `"isSnapshot": true`.

source/sessions/tests/unified/snapshot-sessions.json
```diff
@@ -0,0 +1,926 @@
...
+                  }
+                }
+              }
+            }
+          ]
+        }
+      ]
+    },
+    {
+      "description": "Distinct operation with snapshot",
```
benjirewis, line 275: I cannot run this test and `Mixed operation with snapshot` in libmongoc since we don't have a dedicated `distinct` helper, so I can't _verify_ that these work, but they look good to me.

benjirewis (commented):
Barring tests with `distinct` (see my other comment), spec tests seem to pass in C. Will implement prose tests and re-review.

source/sessions/tests/unified/snapshot-sessions.json
BorisDog: The default `causalConsistency` value is "not supplied". So in that case it should be translated to `!isSnapshot` (now it's translated to `true`)?
I think drivers should account for this case, as we don't want mandate users to always  specify `"causalConsistency": false` if snapshot session is requested.

benjirewis: I agree that specifying `causalConsistency = false` just for `isSnapshot = true` should not be a burden on the users. 

But if a user (or a test) sets `isSnapshot` to true, drivers should set `causalConsistency` to false? I suppose I'm confused about when drivers should explicitly _error_ because of simultaneously true `isSnapshot` and `causalConsistency` values, and when drivers should quietly switch one or the other off.

rstam (approved):
LGTM

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,289 @@
...
+``snapshot`` added to `ReadConcernLevel enumeration <../read-write-concern/read-write-concern.rst#read-concern>`_.`.
+
+Server Commands
+===============
+
+There are no new server commands related to snapshot reads. Instead,
+snapshot reads are implemented by:
+
+1. Saving the ``atClusterTime`` returned by 5.0+ servers for the first find/aggregate/distinct operation in a
+   property ``snapshotTimestamp`` of the ``ClientSession`` object. Drivers MUST save the ``atClusterTime``
```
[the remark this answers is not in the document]
BorisDog: Good point. `snapshotTimestamp` is not part of public API now, and in fact in C# it's called `snapshotTime` internally for this consistency reason.
@ajdavis do you have any objections to `snapshotTime` name?

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,289 @@
...
+
+All read operations performed using this session will be read from the same snapshot.
+
+If no value is provided for ``isSnapshot`` a value of false is
+implied.
+
+MongoClient changes
+===================
+
+There are no API changes to ``MongoClient`` to support snapshot reads.
```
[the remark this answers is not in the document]
BorisDog: I agree, we might not need even that cliet/db/collection section.
@rstam do you think we could skip those no-changes sections?

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,289 @@
...
+ServerSession
+    The driver object representing a server session.
+
+Session
+    A session is an abstract concept that represents a set of sequential
+    operations executed by an application that are related in some way. This
+    specification defines how sessions are used to implement snapshot reads.
+
+Snapshot reads
+    Reads with readconcern level ``snapshot`` that occur outside of transactions on
```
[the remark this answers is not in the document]
BorisDog: Done.

source/sessions/tests/README.rst
```diff
@@ -78,7 +79,21 @@ the given session is *not* marked dirty::
         arguments:
           session: session0
 
+Snapshot session tests
+======================
+Snapshot sessions tests require server of version 5.0 or higher and 
+replica set or a sharded cluster deployment.
+The server ``minSnapshotHistoryWindowInSeconds`` parameter SHOULD be configured to match the test execution time.
```
benjirewis, line 86: Does this need to be set? Why is this important for the new tests, and by "test execution time", do you mean how long it takes the spec + prose tests to run?

benjirewis (approved):
LGTM given new changes to `causal-consistency`. Prose test looks good in libmongoc; just had a question about the set server parameter.

source/sessions/tests/unified/snapshot-sessions.json
BorisDog: You are right, this point is not clear.
[I suggest](https://github.com/mongodb/specifications/pull/1022/files#diff-d87cc9942c0585f49a1d22a6ea656ef51d4a57c27e87b0f147a39890c7617f6eR133-R134) that `causalConsistency`  defaults to true only if no snapshot where requested.

source/sessions/tests/README.rst
BorisDog: We want to ensure that the snapshot the test reads still exists on the server.
IIRC the default history is 10 seconds, which was not sufficient at least for .Net tests. 
So drivers should be aware about this parameter and set it as needed.

ShaneHarvey: Why was 10 seconds not long enough? Are there snapshot reads tests that take longer than 10 seconds?

source/sessions/snapshot-sessions.rst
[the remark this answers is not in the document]
ShaneHarvey: Hmm okay. This line gave me the impression that the snapshotTimestamp needs to be a public session property:
> property ``snapshotTimestamp`` of the ``ClientSession`` object

Can you clarify that it must be private? Perhaps just saying "...private ``snapshotTime`` property..."

I still think it's more consistent to name it "snapshotTime" though.

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,289 @@
...
+========================================================
+
+Snapshot reads are built on top of client sessions.
+
+Applications will start a new client session for snapshot reads like
+this:
+
+.. code:: typescript
+
+    options = new SessionOptions(isSnapshot = true);
```
ShaneHarvey, line 93: Can we rename `isSnapshot` to `snapshot`? Including `is` in the name doesn't add any meaning, it's just a filler word. It's also inconsistent with other boolean options in the driver api. For example `causalConsistency` isn't called `isCausalConsistency` or `isCausallyConsistent`.

source/sessions/tests/README.rst
BorisDog: It was not sufficient when running in Debug mode.
Apparently it's 5 seconds and .Net tests in release are much quicker. So no changes to EG needed (at least in .Net)

I'd still suggest to track this parameter for such slower dubug mode runs. Not sure what's the best wording though, maybe:
> The server ``minSnapshotHistoryWindowInSeconds`` parameter MAY be configured to match the tests execution time.

?

source/sessions/snapshot-sessions.rst
BorisDog: Good point re. consistency. Renamed.

source/sessions/snapshot-sessions.rst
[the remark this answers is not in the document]
BorisDog: Done.

source/sessions/tests/README.rst
ShaneHarvey: Perhaps something like:
"Note that by default the server only keeps the snapshot alive for 10 seconds. Debugging a driver in the middle of a snapshot session may lead to SnapshotTooOld errors. Drivers can work around this issue by increasing the server's minSnapshotHistoryWindowInSeconds parameter, for example:
```python
client.admin.command('setParameter', 1, minSnapshotHistoryWindowInSeconds=60)
```

source/sessions/tests/README.rst
```diff
@@ -78,7 +79,26 @@ the given session is *not* marked dirty::
...
           session: session0
 
+Snapshot session tests
+======================
+Snapshot sessions tests require server of version 5.0 or higher and 
+replica set or a sharded cluster deployment.
+Default snapshot history window on the server is 5 seconds. Running the test in debug mode, or in any other slow configuration
+may lead to `SnapshotTooOld` errors. Drivers can work around this issue by increasing the server's `minSnapshotHistoryWindowInSeconds` parameter, for example:
+
+.. code:: typescript
```
ShaneHarvey, line 89: Should be "python".

source/sessions/tests/README.rst
```diff
@@ -78,7 +79,26 @@ the given session is *not* marked dirty::
...
+
+    client.admin.command('setParameter', 1, minSnapshotHistoryWindowInSeconds=60)
+
+Prose tests
+```````````
+-  Setting both ``snapshot`` and ``causalConsistency`` is not allowed
+
+    * ``client.startSession(snapshot = true, causalConsistency = true)``
+    * Assert that an error was raised by driver
```
ShaneHarvey, line 98: Can we add a unified test for this? It could run the client.startSession operation with `snapshot = true, causalConsistency = true` arguments and assert that we raise a client-side error with `isClientError: true`.

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,265 @@
...
+snapshot reads are implemented by:
+
+1. Saving the ``atClusterTime`` returned by 5.0+ servers for the first find/aggregate/distinct operation in a
+   private property ``snapshotTime`` of the ``ClientSession`` object. Drivers MUST save the ``atClusterTime``
+   in the ``ClientSession`` object.
+
+2. Passing that ``snapshotTime`` in the ``atClusterTime`` field of the ``readConcern`` field
+   for subsequent snapshot read operations (for find/aggregate/distinct commands).
+
+Server Command Responses
```
ShaneHarvey, line 154: When the driver starts a snapshot session but the server does not return the atClusterTime field the driver needs to raise a client side error indicating that server does not support snapshot reads. If we don't do this then snapshot reads can silently fail.

ShaneHarvey (changes requested):

source/sessions/snapshot-sessions.rst
ShaneHarvey: False alarm. The problem was that my implementation neglected to set readConcern.level to ‘snapshot’. When the driver correctly sends ‘snapshot’ old servers will reject the request with this error:
```
{'operationTime': Timestamp(1624572085, 12), 'ok': 0.0, 'errmsg': 'read concern level snapshot is only valid in a transaction', 'code': 72, 'codeName': 'InvalidOptions', '$clusterTime': {'clusterTime': Timestamp(1624572085, 12), 'signature': {'hash': b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00', 'keyId': 0}}}
```

source/sessions/tests/README.rst
```diff
@@ -78,7 +79,26 @@ the given session is *not* marked dirty::
         arguments:
           session: session0
 
+Snapshot session tests
+======================
+Snapshot sessions tests require server of version 5.0 or higher and 
+replica set or a sharded cluster deployment.
+Default snapshot history window on the server is 5 seconds. Running the test in debug mode, or in any other slow configuration
```
ShaneHarvey, line 86: The default was updated to 5 _minutes_ in SERVER-47855 so this section is probably not needed at all anymore.

source/sessions/snapshot-sessions.rst
```diff
@@ -0,0 +1,265 @@
...
+ServerSession
+    The driver object representing a server session.
+
+Session
+    A session is an abstract concept that represents a set of sequential
+    operations executed by an application that are related in some way. This
+    specification defines how sessions are used to implement snapshot reads.
+
+Snapshot reads
+    Reads with read concern level ``snapshot`` that occur outside of transactions on
```
benjirewis, line 64: Should we add a test to assert that `snapshot` read concern level will not be appended during a transaction?

BorisDog: Not sure. Now that transactions are not allowed with snapshot reads, and we got a test for that. I don't see how snapshot readconcern can end up being appended in transaction.

benjirewis: Yes I think this is covered by the new "StartTransaction fails in snapshot session" test; thanks.

source/sessions/tests/README.rst
ShaneHarvey: > and we don't have startSession operation.
The unified test runner allows use to call any method on the client including `startSession`. I don't think it's overkill but we don't need to do it in this PR. I'll open a drivers ticket for it.

ShaneHarvey (changes requested):
Can you add the `.yml` versions of the two new tests?:
```
 source/sessions/tests/unified/snapshot-sessions-not-supported-server-error.yml
 source/sessions/tests/unified/snapshot-sessions.yml
 ```

**Review by ShaneHarvey — APPROVED** [review]
LGTM

**Comment by BorisDog** [comment]
Rebased and merged [here](https://github.com/mongodb/specifications/commit/1a5ddeefb4c01180bf06d06ac651e7571ab29910).

File: source/unified-test-format/tests/Makefile [hunk 50]
```diff
@@ -32,6 +32,9 @@ crud: HAS_AJV
 collection-management: HAS_AJV
 	@ajv test -s $(SCHEMA) -d "../../collection-management/tests/*.yml" --valid
 
+sessions: HAS_AJV
```
jmikola, line 35: FYI: "sessions" should also have been added to the `all` target above. See: https://github.com/mongodb/specifications/pull/1026#discussion_r660939153

PR #961, opened 2021-04-13
DRIVERS-720 Change existing tests to run in LB mode

divjotarora:
This PR is related to 958 and modifies existing spec tests for versioned API, change streams, retryable reads, and retryable writes to run against load balancers. Most of these tests files predate the unified test format and have not been ported over yet, but all of the formats contain a `runOn` directive with a `topology` option. I modified any test that was already running against `sharded` topologies to also include `load-balanced` unless the test had a `maxServerVersion` < 5.0 because LB support will only be present in 5.0+ servers.

PR #1023, opened 2021-06-16
Fix schema version in versioned API transaction handling test file

kmahar:
It seems in #961 where load balancer testing support was added the schema version should have been bumped in these files as they run against the load-balanced topology.

### DIFF
--- a/source/versioned-api/tests/transaction-handling.json
+++ b/source/versioned-api/tests/transaction-handling.json
@@ -1,6 +1,6 @@
 {
   "description": "Transaction handling",
-  "schemaVersion": "1.1",
+  "schemaVersion": "1.3",
   "runOnRequirements": [
     {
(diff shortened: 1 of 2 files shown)


alcaeus (approved):

kmahar:
@alcaeus do you think I should make a DRIVERS ticket for this / spin out language tickets to sync it? I'm inclined to not bother generating the noise because it doesn't seem like there's much risk in anyone having the file with a too-low version if it's not causing them problems at the moment... this was only an issue in Swift because we decode the topologies to an enum and didn't have a "load-balanced" case yet.

alcaeus:
Maybe a message in driver-devs is sufficient to let people know there might be a problem. Let's not create more 5.0 noise

kmahar:
sgtm 👍

PR #1024, opened 2021-06-21
DRIVERS-1483 Add error label to retryable writes `PoolClearedError` prose test

patrickfreed:
This PR adds an error label to the failpoint used in the `PoolClearedError` retryable writes test. Without it, the test would fail on MongoDB 4.4+. This PR also adds information regarding the minimum MongoDB version required to run the retryability tests against.

### DIFF
--- a/source/retryable-writes/tests/README.rst
+++ b/source/retryable-writes/tests/README.rst
@@ -330,7 +330,7 @@ and sharded clusters.
 
 #. Test that drivers properly retry after encountering PoolClearedErrors. This
    test MUST be implemented by any driver that implements the CMAP
-   specification.
+   specification. This test requires MongoDB 4.2.9+ for `blockConnection` support in the failpoint.
 
    1. Create a client with maxPoolSize=1 and retryWrites=true. If testing
       against a sharded deployment, be sure to connect to only a single mongos.
@@ -344,7 +344,8 @@ and sharded clusters.
                failCommands: ["insert"],
                errorCode: 91,
                blockConnection: true,
-               blockTimeMS: 1000
+               blockTimeMS: 1000,
+               errorLabels: ["RetryableWriteError"]
            }
        }
 
--- a/source/retryable-writes/tests/README.rst
+++ b/source/retryable-writes/tests/README.rst
@@ -330,7 +330,7 @@ and sharded clusters.
 
 #. Test that drivers properly retry after encountering PoolClearedErrors. This
    test MUST be implemented by any driver that implements the CMAP
-   specification. This test requires MongoDB 4.2.9+ for `blockConnection` support in the failpoint.
+   specification. This test requires MongoDB 4.2.9+ for ``blockConnection`` support in the failpoint.
 
    1. Create a client with maxPoolSize=1 and retryWrites=true. If testing
       against a sharded deployment, be sure to connect to only a single mongos.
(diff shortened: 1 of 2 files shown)


source/retryable-writes/tests/README.rst
```diff
@@ -330,7 +330,7 @@ and sharded clusters.
 
 #. Test that drivers properly retry after encountering PoolClearedErrors. This
    test MUST be implemented by any driver that implements the CMAP
-   specification.
+   specification. This test requires MongoDB 4.2.9+ for `blockConnection` support in the failpoint.
```
jmikola, line 333: RST uses double backticks for monospace formatting.

source/retryable-writes/tests/README.rst
```diff
@@ -344,7 +344,8 @@ and sharded clusters.
                failCommands: ["insert"],
                errorCode: 91,
                blockConnection: true,
-               blockTimeMS: 1000
+               blockTimeMS: 1000,
+               errorLabels: ["RetryableWriteError"]
```
jmikola, line 348: > Without it, the test would fail on MongoDB 4.4+.

Is this because the fail point never adds a label on the error code, and drivers won't add the label themselves for 4.4+ servers (per [RetryableWriteError Labels](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/retryable-writes.rst#retryablewriteerror-labels))?

Is this redundant for pre-4.4 servers?

We document the `failCommand` fail point in [UTF spec](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#failcommand), but `errorLabels` is never mentioned there. Cross-referencing with https://github.com/mongodb/mongo/wiki/The-%22failCommand%22-fail-point leads me to [SERVER-43941](https://jira.mongodb.org/browse/SERVER-43941), which shows this was introduced in 4.3.1 and never backported. That presumably means this test can only be run on a 4.4+ server.

I also noticed that `errorLabels` was used in [poc-retryable-writes.yml](https://github.com/mongodb/specifications/blob/72249aad5068f9a4d24930e723e90d92bb82b6cb/source/unified-test-format/tests/valid-pass/poc-retryable-writes.yml#L134) and that test is permitted to run on 4.0+ servers. That doesn't make sense to me, unless `configureFailPoint` is not strictly validating `data` options and ignoring unsupported values.

jmikola: Created [DRIVERS-1814](https://jira.mongodb.org/browse/DRIVERS-1814) to track the UTF documentation issue.

patrickfreed: I think the option is just ignored pre 4.4, which is fine as the server doesn't attach the label on those versions anyways and the driver is required to.

source/retryable-writes/tests/README.rst
patrickfreed: fixed

source/retryable-writes/tests/README.rst
jmikola: Aye, @ldennis confirmed as much in [DRIVERS-1814](https://jira.mongodb.org/browse/DRIVERS-1814).

jmikola (approved):

ShaneHarvey (approved):

PR #1026, opened 2021-06-24
DRIVERS-1816 Test redaction of replies to security sensitive commands

alcaeus:
DRIVERS-1816

This does not cover replies to all commands, since we're only able to make assertions against successful commands. With some authentication commands, getting them to work is a sizeable effort. For that reason, it's up to engineers to ensure they're not just redacting replies to the commands where we test this, but rather all commands. This includes replies that are potentially sensitive, e.g. replies for `saslStart` and `saslContinue`.

Note that due to a limitation of the unified test format, we can't make assertions against the contents of CommandFailedEvents. While the spec doesn't explicitly mention this event, it is assumed that drivers redact replies when they are exposed in CommandFailedEvents. That said, most if not all replies will not contain sensitive information, so I believe it's ok to leave this untested for the time being.

### DIFF
--- a/source/command-monitoring/tests/unified/redacted-commands.yml
+++ b/source/command-monitoring/tests/unified/redacted-commands.yml
@@ -234,8 +234,10 @@ tests:
               reply:
                 # We expect the reply not to be redacted, as the server can't
                 # successfully authenticate us, which means the reply does not
-                # need to be redacted
-                isWritablePrimary: { $$exists: true }
+                # need to be redacted. However, drivers may still choose to
+                # redact the reply based on the fact that the command contained
+                # sensitive information
+                isWritablePrimary: { $$unsetOrMatches: true }
                 # This assertion will currently always hold true since we're
                 # not expecting successful authentication, in which case this
                 # field is missing anyways.
@@ -270,7 +272,7 @@ tests:
           - commandSucceededEvent:
               commandName: ismaster
               reply:
-                ismaster: { $$exists: true }
+                ismaster: { $$unsetOrMatches: true }
                 speculativeAuthenticate: { $$exists: false }
           - commandStartedEvent:
               commandName: isMaster
@@ -280,7 +282,7 @@ tests:
           - commandSucceededEvent:
               commandName: isMaster
               reply:
-                ismaster: { $$exists: true }
+                ismaster: { $$unsetOrMatches: true }
                 speculativeAuthenticate: { $$exists: false }
 
   - description: "hello without speculative authenticate is not redacted"
--- a/source/command-monitoring/tests/unified/redacted-commands.yml
+++ b/source/command-monitoring/tests/unified/redacted-commands.yml
@@ -11,6 +11,7 @@ createEntities:
       id: &client client
       observeEvents:
         - commandStartedEvent
+        - commandSucceededEvent
       observeSensitiveCommands: true
   - database:
       id: &database database
@@ -25,10 +26,12 @@ tests:
         arguments:
           commandName: authenticate
           command:
-            authenticate: "private"
-        # Malformed authentication commands will fail against the server, but we
-        # just want to assert that security-related commands are redacted
-        # in command monitoring.
+            authenticate: 1
+            mechanism: "MONGODB-X509"
+            user: "CN=myName,OU=myOrgUnit,O=myOrg,L=myLocality,ST=myState,C=myCountry"
+            db: "$external"
+        # An authentication error is expected, but we want to check that the
+        # CommandStartedEvent is redacted
         expectError:
           isError: true
     expectEvents:
@@ -36,11 +39,14 @@ tests:
         events:
           - commandStartedEvent:
               commandName: authenticate
-              # This only asserts that the command name has been redacted from the published command;
-              # however, it's unlikely that a driver would redact this but leave other, sensitive fields.
-              # We cannot simply assert that command is an empty document because it's at root-level, so
-              # additional fields in the actual document will be permitted.
-              command: { authenticate: { $$exists: false } }
+              # We cannot simply assert that command is an empty document
+              # because it's at root-level, so we make a best effort to make
+              # sure sensitive fields are redacted.
+              command:
+                authenticate: { $$exists: false }
+                mechanism: { $$exists: false }
+                user: { $$exists: false }
+                db: { $$exists: false }
 
   - description: "saslStart"
     operations:
@@ -49,7 +55,9 @@ tests:
         arguments:
           commandName: saslStart
           command:
-            saslStart: "private"
+            saslStart: 1
+            payload: "definitely-invalid-payload"
+            db: "admin"
         expectError:
           isError: true
     expectEvents:
@@ -57,7 +65,10 @@ tests:
         events:
           - commandStartedEvent:
               commandName: saslStart
-              command: { saslStart: { $$exists: false } }
+              command:
+                saslStart: { $$exists: false }
+                payload: { $$exists: false }
+                db: { $$exists: false }
 
   - description: "saslContinue"
     operations:
@@ -66,7 +77,9 @@ tests:
         arguments:
           commandName: saslContinue
           command:
-            saslContinue: "private"
+            saslContinue: 1
+            conversationId: 0
+            payload: "definitely-invalid-payload"
         expectError:
           isError: true
     expectEvents:
@@ -74,7 +87,10 @@ tests:
         events:
           - commandStartedEvent:
               commandName: saslContinue
-              command: { saslContinue: { $$exists: false } }
+              command:
+                saslContinue: { $$exists: false }
+                conversationId: { $$exists: false }
+                payload: { $$exists: false }
 
   - description: "getnonce"
     operations:
@@ -83,13 +99,18 @@ tests:
         arguments:
           commandName: getnonce
           command:
-            getnonce: "private"
+            getnonce: 1
     expectEvents:
       - client: *client
         events:
           - commandStartedEvent:
               commandName: getnonce
               command: { getnonce: { $$exists: false } }
+          - commandSucceededEvent:
+              commandName: getnonce
+              reply:
+                ok: { $$exists: false }
+                nonce: { $$exists: false }
 
   - description: "createUser"
     operations:
@@ -99,6 +120,10 @@ tests:
           commandName: createUser
           command:
             createUser: "private"
+            # Passing an object is prohibited and we want to trigger a command
+            # failure
+            pwd: {}
+            roles: []
         expectError:
           isError: true
     expectEvents:
@@ -106,7 +131,10 @@ tests:
         events:
           - commandStartedEvent:
               commandName: createUser
-              command: { createUser: { $$exists: false } }
+              command:
+                createUser: { $$exists: false }
+                pwd: { $$exists: false }
+                roles: { $$exists: false }
 
   - description: "updateUser"
     operations:
@@ -116,6 +144,8 @@ tests:
           commandName: updateUser
           command:
             updateUser: "private"
+            pwd: {}
+            roles: []
         expectError:
           isError: true
     expectEvents:
@@ -123,7 +153,10 @@ tests:
         events:
           - commandStartedEvent:
               commandName: updateUser
-              command: { updateUser: { $$exists: false } }
+              command:
+                updateUser: { $$exists: false }
+                pwd: { $$exists: false }
+                roles: { $$exists: false }
 
   - description: "copydbgetnonce"
     operations:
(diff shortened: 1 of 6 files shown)


source/command-monitoring/tests/unified/redacted-commands.yml
```diff
@@ -177,76 +210,131 @@ tests:
               command: { copydb: { $$exists: false } }
 
   - description: "hello with speculative authenticate"
+    runOnRequirements:
+      - minServerVersion: 4.9
```
benjirewis, line 214: I think this should be a string `"4.9"` here and on L288.

source/command-monitoring/tests/unified/redacted-commands.yml
```diff
@@ -177,76 +210,131 @@ tests:
...
+              command:
+                hello: { $$exists: false }
+                speculativeAuthenticate: { $$exists: false }
+          - commandSucceededEvent:
+              commandName: hello
+              reply:
+                # We expect the reply not to be redacted, as the server can't
+                # successfully authenticate us, which means the reply does not
+                # need to be redacted
+                isWritablePrimary: { $$exists: true }
```
benjirewis, line 238: Hmm confused what's going on here. The Go driver will redact the `reply` field here because this is a `hello` with `speculativeAuthenticate`, `isWritablePrimary` will not exist, and this test will not pass. Why is the redaction conditional on the success of the command?

source/command-monitoring/tests/unified/redacted-commands.yml
```diff
@@ -177,76 +210,131 @@ tests:
...
           - commandStartedEvent:
               commandName: isMaster
-              command: { isMaster: { $$exists: false } }
+              command:
+                isMaster: { $$exists: false }
+                speculativeAuthenticate: { $$exists: false }
+          - commandSucceededEvent:
+              commandName: isMaster
+              reply:
+                ismaster: { $$exists: true }
```
benjirewis, line 283: Uppercase `isMaster` here?

source/command-monitoring/tests/unified/redacted-commands.yml
```diff
@@ -177,76 +210,131 @@ tests:
...
           - commandStartedEvent:
               commandName: isMaster
-              command: { isMaster: "public" }
+              command:
+                isMaster: 1
+          - commandSucceededEvent:
+              commandName: isMaster
+              reply:
+                ismaster: { $$exists: true }
```
benjirewis, line 340: Uppercase `isMaster`?

benjirewis (commented):
Thanks for adding these tests! To be clear, do these changes assert the following?
1. _Most_ security-sensitive fields are redacted from security-sensitive commands (you've added assertions for `payload`, `mechanism`, etc.)
2. Server replies to `hello`s and legacy hellos are similarly redacted.

And, it's up to individual drivers to assert all other replies are redacted (including failures)?

source/command-monitoring/tests/unified/redacted-commands.yml
dariakp: seconding this comment, as the same is true for the node driver

source/command-monitoring/tests/unified/redacted-commands.yml
alcaeus: The response is always lowercased, regardless of the command being invoked (output truncated to show relevant fields only):
```
MongoDB Enterprise > db.runCommand({ismaster: 1})
{
	"ismaster" : true,
	"ok" : 1
}
MongoDB Enterprise > db.runCommand({isMaster: 1})
{
	"ismaster" : true,
	"ok" : 1
}
```

source/command-monitoring/tests/unified/redacted-commands.yml
alcaeus: `hello` (and legacy hello for the rest of this comment) need to be handled slightly differently. While for all other commands we make the decision to redact based on the command name alone, `hello` only needs to be redacted when it contains sensitive information, i.e. the `speculativeAuthenticate` field. However, the reply to a sensitive command may not be necessarily sensitive (as in this case where we fail speculative authentication and the `hello` reply is as if we hadn't attempted speculative authentication in the first place).

Due to the architecture of the C driver (where I built these tests), we can no longer inspect the original command document when evaluating whether a reply needs to be redacted. Thus, the C driver only redacts replies to `hello` if the reply itself contains a `speculativeAuthenticate` field. While I believe this behaviour is fine and still fulfils the intention of the spec (redacting sensitive information), this behaviour doesn't necessarily apply to other drivers.

I've updated the tests to use `$$unsetOrMatches`: this allows drivers to always redact replies to `hello` when the command document used speculative authentication even when the reply does not contain sensitive information. It also accepts the behaviour of drivers that only redact the reply if it contains sensitive information.

Note that this only applies to `hello` - other replies are always redacted in the C driver. I'll make a second attempt at creating tests that allow successful authentication (e.g. to be able to test the response to `saslStart` and `hello` with speculative authentication), but I believe this to be outside of the scope of this ticket as it will most likely require more changes to the unified test format, which I'd like to avoid at this time.

source/unified-test-format/schema-1.5.json
```diff
@@ -0,0 +1,462 @@
+{
```
alcaeus, line 1: While working on the recent changes, I realised that the original PR that added support for the `observeSensitiveCommands` client entity option didn't create the 1.5 schema version and also didn't add the tests to the unified test format's Makefile.

source/unified-test-format/tests/Makefile
```diff
@@ -1,8 +1,8 @@
-SCHEMA=../schema-1.4.json
+SCHEMA=../schema-1.5.json
 
-.PHONY: all invalid valid-fail valid-pass versioned-api load-balancers gridfs transactions crud collection-management HAS_AJV
+.PHONY: all invalid valid-fail valid-pass versioned-api load-balancers gridfs transactions crud collection-management command-monitoring HAS_AJV
```
alcaeus, line 3: I've love to find a different option for this at some point in time. Ideally, we'd go through all directories, grab every `.yml` file, attempt to read a schema version for it (skipping the file if we don't find one) and validating the file against the explicit schema version indicated in the file. @jmikola do you think this is feasible? If so I'd file a ticket so we can eventually return to this.

This solution would fix two issues:
1. it would no longer require us to manually add new folders with unified test files here, and
2. it validates files against the schema version they claim to be, which can catch errors when somebody adds functionality from newer schema versions without bumping the `schemaVersion` in the file.

source/command-monitoring/tests/unified/redacted-commands.yml
benjirewis: Yep, in the Go driver, the `redacted` flag in a `commandSucceededEvent` is directly inherited from the corresponding `commandStartedEvent`, so the logic isn't granular enough to recognize when a command is security-sensitive but its reply is not. Using `$$unsetOrMatches` makes sense to me.

source/command-monitoring/tests/unified/redacted-commands.yml
```diff
@@ -177,76 +210,133 @@ tests:
...
+                # We expect the reply not to be redacted, as the server can't
+                # successfully authenticate us, which means the reply does not
+                # need to be redacted. However, drivers may still choose to
+                # redact the reply based on the fact that the command contained
+                # sensitive information
+                isWritablePrimary: { $$unsetOrMatches: true }
+                # This assertion will currently always hold true since we're
+                # not expecting successful authentication, in which case this
+                # field is missing anyways.
+                speculativeAuthenticate: { $$exists: false }
```
benjirewis, line 244: Even if I remove the logic for redacting replies, this test (and the corresponding one for legacy hello) still passes in the Go driver. The reply still exists (because I'm no longer redacting), `isWritablePrimary` is true, and `speculativeAuthenticate` does not exist because authentication failed.

source/command-monitoring/tests/unified/redacted-commands.yml
benjirewis: I'm not sure this will work actually, see my [most recent comment](https://github.com/mongodb/specifications/pull/1026#discussion_r658799927).

source/unified-test-format/tests/Makefile
jmikola: [DRIVERS-1767](https://jira.mongodb.org/browse/DRIVERS-1767) already exists to track this. If there's some variation to the behavior (e.g. skipping files that don't have a `schemaVersion` field) feel free to note that as a suggested approach in the ticket.

source/unified-test-format/schema-1.5.json
jmikola: I definitely missed that while reviewing #1015.

@prashantmital: Any idea what happened there? You updated the test file itself to use 1.5 but the PR never included a `schema-1.5.json`. Did you actually make one and just forget to `git add` it?

source/command-monitoring/tests/unified/redacted-commands.yml
alcaeus: Yes, that is unfortunately true. I can either remove the reply assertions in this test or leave them in to show intent.

I've tried to create a user beforehand so we can actually test the auth process (and thus get a speculativeAuthenticate in the reply) but wasn't able to get it to pass consistently without spending more time. With the tight timeline on this I opted to take a more pragmatic approach.

source/unified-test-format/tests/Makefile
alcaeus: I'll check it out, thanks for pointing me in the right direction

source/command-monitoring/tests/unified/redacted-commands.yml
alcaeus: Resolving this comment thread as this issue is addressed in a separate comment.

source/unified-test-format/schema-1.5.json
prashantmital: Actually, I didn't create `schema-1.5.json` at all. Completely slipped my mind that this was something I needed to do.

source/command-monitoring/tests/unified/redacted-commands.yml
```diff
@@ -25,22 +26,27 @@ tests:
...
-        # Malformed authentication commands will fail against the server, but we
-        # just want to assert that security-related commands are redacted
-        # in command monitoring.
+            authenticate: 1
+            mechanism: "MONGODB-X509"
+            user: "CN=myName,OU=myOrgUnit,O=myOrg,L=myLocality,ST=myState,C=myCountry"
+            db: "$external"
+        # An authentication error is expected, but we want to check that the
+        # CommandStartedEvent is redacted
         expectError:
```
jmikola, line 35: Is this a case where you might want to use `ignoreResultAndError` (introduced in schema version 1.3)?

source/command-monitoring/tests/unified/redacted-commands.yml
```diff
@@ -177,76 +210,131 @@ tests:
...
+                # not expecting successful authentication, in which case this
+                # field is missing anyways.
+                speculativeAuthenticate: { $$exists: false }
+
+  - description: "legacy hello with speculative authenticate"
+    operations:
       - name: runCommand
         object: *database
         arguments:
           commandName: ismaster
```
jmikola, line 249: Noted that you're using the all-lowercase variant of `isMaster` here in order to test both forms accepted by the server (and dictated by the spec).

source/command-monitoring/tests/unified/redacted-commands.yml
```diff
@@ -177,76 +210,131 @@ tests:
...
+                # contain sensitive information, we're expecting the reply to be
+                # redacted as well.
+                isWritablePrimary: { $$exists: false }
+                # This assertion will currently always hold true since we're
+                # not expecting successful authentication, in which case this
+                # field is missing anyways.
+                speculativeAuthenticate: { $$exists: false }
+
+  - description: "legacy hello with speculative authenticate"
+    operations:
```
jmikola, line 245: Would it make sense to give this test a `maxServerVersion` requirement, or are there no plans for the server to remove support for `isMaster` entirely in the foreseeable future?

source/unified-test-format/tests/Makefile
```diff
@@ -1,8 +1,8 @@
-SCHEMA=../schema-1.4.json
+SCHEMA=../schema-1.5.json
 
-.PHONY: all invalid valid-fail valid-pass versioned-api load-balancers gridfs transactions crud collection-management HAS_AJV
+.PHONY: all invalid valid-fail valid-pass versioned-api load-balancers gridfs transactions crud collection-management sessions command-monitoring HAS_AJV
 
-all: invalid valid-fail valid-pass versioned-api load-balancers gridfs transactions crud collection-management
+all: invalid valid-fail valid-pass versioned-api load-balancers gridfs transactions crud collection-management sessions command-monitoring
```
jmikola, line 5: Re: "sessions", I assume that was missed in https://github.com/mongodb/specifications/pull/1022?

cc: @BorisDog

source/unified-test-format/tests/invalid/entity-client-observeSensitiveCommands-type.yml
```diff
@@ -0,0 +1,12 @@
+description: "entity-client-observeSensitiveCommands-type"
+
+schemaVersion: "1.0"
```
jmikola, line 3: This should use schema version 1.5.

source/unified-test-format/tests/Makefile
ShaneHarvey: Boris is OOO but yes we forgot to add it in the snapshot sessions PR.

source/command-monitoring/tests/unified/redacted-commands.yml
alcaeus: I know of no plans to drop legacy hello, and even the work to deprecated legacy opcodes makes an exception for using `OP_QUERY` to send an initial handshake command, so I don't think we'll have an issue.

If the legacy hello command were indeed removed from the server, I think some more of our tests need updating to account for this so I'd rather defer for the time being.

source/unified-test-format/tests/Makefile
alcaeus: That is correct, I figured I'd sneak it in there.

durran:
Tests pass on Node: https://github.com/mongodb/node-mongodb-native/pull/2872

source/unified-test-format/tests/invalid/entity-client-observeSensitiveCommands-type.yml
durran: Same.

durran (changes requested):
Just a schema version bump to 1.5 in `entity-client-observeSensitiveCommands-type.yml`

source/command-monitoring/tests/unified/redacted-commands.yml
jmikola: As discussed in Slack, doing so would complicate integrating these tests into older driver versions for the patch. Fine to keep as-is, since we do know these commands will error anyway.

durran (approved):

PR #1032, opened 2021-06-29
DRIVERS-1820 Snapshot reads raise client-side error on <5.0

ShaneHarvey:
Updated to require drivers raise a client-side error "snapshot reads require MongoDB 5.0 or later" when a snapshot read is attempted on < 5.0.

Note: I have a POC of this change here: https://github.com/mongodb/mongo-python-driver/pull/659

### DIFF
--- a/source/sessions/snapshot-sessions.rst
+++ b/source/sessions/snapshot-sessions.rst
@@ -238,7 +238,7 @@ Requires MongoDB 5.0+
 
 Snapshot reads require MongoDB 5.0+. When the connected server's
 maxWireVersion is less than 13, drivers MUST throw an exception with the
-message "snapshot reads require MongoDB 5.0 or later".
+message "Snapshot reads require MongoDB 5.0 or later".
 
 Motivation
 ==========
--- a/source/sessions/snapshot-sessions.rst
+++ b/source/sessions/snapshot-sessions.rst
@@ -3,7 +3,7 @@ Snapshot Reads Specification
 ============================
 
 :Spec Title: Snapshot Reads Specification (See the registry of specs)
-:Spec Version: 1.0
+:Spec Version: 1.1
 :Author: Boris Dogadov
 :Advisors: Jeff Yemin, A. Jesse Jiryu Davis, Judah Schvimer
 :Status: Draft (Could be Draft, Accepted, Rejected, Final, or Replaced)
@@ -233,6 +233,13 @@ Lists of commands that support snapshot reads:
 2. aggregate
 3. distinct
 
+Requires MongoDB 5.0+
+=====================
+
+Snapshot reads require MongoDB 5.0+. When the connected server's
+maxWireVersion is less than 13, drivers MUST throw an exception with the
+message "snapshot reads require MongoDB 5.0 or later".
+
 Motivation
 ==========
 
@@ -269,3 +276,4 @@ Changelog
 =========
 
 :2021-06-15: Initial version.
+:2021-06-28: Raise client side error on < 5.0.
--- a/source/sessions/tests/unified/snapshot-sessions-not-supported-server-error.json
+++ b/source/sessions/tests/unified/snapshot-sessions-not-supported-server-error.json
@@ -43,6 +43,20 @@
       }
     },
     {
+      "database": {
+        "id": "missingDB",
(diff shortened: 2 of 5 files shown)


jyemin (commented):
Seems like a lot of semi-related changes are mixed in to this PR.  Having a bit of trouble figuring out what's what.  Can you clarify?

ShaneHarvey:
This PR only makes the one change described above. I did remove some cruft in one of the test files but nothing significant.

source/sessions/tests/unified/snapshot-sessions-not-supported-server-error.yml
```diff
@@ -3,24 +3,21 @@ description: snapshot-sessions-not-supported-server-error
...
       client: *client0
-      databaseName: &database0Name database0
+      databaseName: *database0Name
   - collection:
-      id: &collection0 collection0
-      database: *database0
-      collectionName: &collection0Name collection0
+      id: &collection0Name collection0
+      database: *database0Name
+      collectionName: *collection0Name
```
alcaeus, line 20: I'm not sure I understand the point of this. Previously, `database0` pointed to the entity ID (to be used in operations), while `database0Name` pointed to the name of the database (which doesn't necessarily have to match). However, the entity ID was never referenced in the tests below, instead using `database0` as hard-coded identifier. IMO, these entities should remain the same (`database0` for the entity, `database0Name` for the name of the database backing the entity), but the tests below should use references instead of hardcoding entity IDs.

As a quick resolution to not stall the downstream work based on this, I'd be fine with undoing this change and filing a follow-up ticket to ensure we're using references when using entities.

jyemin (changes requested):

source/sessions/tests/unified/snapshot-sessions-not-supported-server-error.yml
ShaneHarvey: &database0Name and &database0 are exactly the same (they're both "database0"). All I did was remove the redundant label. Is that a problem?

source/sessions/tests/unified/snapshot-sessions-not-supported-server-error.json
```diff
@@ -3,11 +3,7 @@
   "schemaVersion": "1.0",
   "runOnRequirements": [
     {
-      "minServerVersion": "3.6",
-      "maxServerVersion": "4.4.99"
-    },
-    {
-      "minServerVersion": "3.6",
+      "minServerVersion": "5.0",
```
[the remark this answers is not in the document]
ShaneHarvey: This change is required. We rely on the server to report an error on 5.0+ so these tests can't run on <5.0. I moved the testing for <5.0 servers to the new file "snapshot-sessions-not-supported-client-error.yml".

source/sessions/snapshot-sessions.rst
```diff
@@ -233,6 +233,13 @@ Lists of commands that support snapshot reads:
 2. aggregate
 3. distinct
 
+Requires MongoDB 5.0+
+=====================
+
+Snapshot reads require MongoDB 5.0+. When the connected server's
+maxWireVersion is less than 13, drivers MUST throw an exception with the
+message "snapshot reads require MongoDB 5.0 or later".
```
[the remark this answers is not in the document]
ShaneHarvey: Done.

jyemin (approved):
LGTM!

rstam (approved):
LGTM

source/sessions/tests/unified/snapshot-sessions-not-supported-server-error.yml
ShaneHarvey: I'm going to merge this with the changes I already made. I found it confusing to have multiple labels for the same string especially when they're barely used. If you'd like to change it back or standardize on one way or the other that's fine by me. Perhaps there's even a tool to automatically "condense" yaml files.

alcaeus: Yeah, no need to change it here, just wanted to point it out. Use of these are spotty at best across our spec tests (see this file not using the entity reference and instead hardcoding the name), so no need to enforce consistency when there isn't any to begin with.

PR #1033, opened 2021-06-29
DRIVERS-1819 Snapshot session sends readConcern with all commands even writes

ShaneHarvey:
This PR requires drivers to send readConcern with all commands in a snapshot session (even writes). This effectively bans all unsupported operations in a snapshot session because the server will reject commands that do not support readConcern.level "snapshot". For example:
```
MongoDB Enterprise repl0:PRIMARY> db.runCommand({'listCollections': 'test', readConcern: {level: 'snapshot'}})
{
"ok" : 0,
"errmsg" : "Command listCollections does not support { readConcern: { level: \"snapshot\", provenance: \"clientSupplied\" } } :: caused by :: read concern not supported",
"code" : 72,
"codeName" : "InvalidOptions"
}
MongoDB Enterprise repl0:PRIMARY> db.runCommand({'insert': 'test', documents: [{}], readConcern: {level: 'snapshot'}})
{
"ok" : 0,
"errmsg" : "Command insert does not support { readConcern: { level: \"snapshot\", provenance: \"clientSupplied\" } } :: caused by :: read concern not supported",
"code" : 72,
"codeName" : "InvalidOptions"
}
```

I've added tests for a variety of methods to ensure we send readConcern.level correctly and that the server rejects the field. I can add tests for more operations once we decide this approach is reasonable.

### DIFF
--- a/source/sessions/snapshot-sessions.rst
+++ b/source/sessions/snapshot-sessions.rst
@@ -3,13 +3,13 @@ Snapshot Reads Specification
 ============================
 
 :Spec Title: Snapshot Reads Specification (See the registry of specs)
-:Spec Version: 1.1
+:Spec Version: 1.2
 :Author: Boris Dogadov
 :Advisors: Jeff Yemin, A. Jesse Jiryu Davis, Judah Schvimer
 :Status: Draft (Could be Draft, Accepted, Rejected, Final, or Replaced)
 :Type: Standards
 :Minimum Server Version: 5.0
-:Last Modified: 15-Jun-2021
+:Last Modified: 29-Jun-2021
 
 .. contents::
 
@@ -192,6 +192,7 @@ Server Errors
 =============
 1. The server may reply to read commands with a ``SnapshotTooOld`` error if the client's ``atClusterTime`` value is not available in the server's history.
 2. The server will return ``InvalidOptions`` error if both ``atClusterTime`` and ``afterClusterTime`` options are set to true.
+3. The server will return ``InvalidOptions`` error if the command does not support readConcern.level "snapshot".
 
 Snapshot read commands
 ======================
@@ -211,7 +212,7 @@ by specifying ``readConcern`` with ``snapshot`` level field, and store it as ``s
         }
     }
 
-For subsequent reads from same snapshot driver MUST send the ``snapshotTime`` saved in
+For subsequent reads in the same session, the driver MUST send the ``snapshotTime`` saved in
 the ``ClientSession`` as the value of the ``atClusterTime`` field of the
 ``readConcern`` with ``snapshot`` level field:
 
@@ -232,6 +233,12 @@ Lists of commands that support snapshot reads:
 1. find
 2. aggregate
 3. distinct
+
+Sending readConcern to the server on all commands
+=================================================
+
+Drivers MUST set the readConcern ``level`` and ``atClusterTime`` fields (as
+outlined above) on all commands in a snapshot session.
 
 Requires MongoDB 5.0+
 =====================
@@ -277,3 +284,4 @@ Changelog
 
 :2021-06-15: Initial version.
 :2021-06-28: Raise client side error on < 5.0.
+:2021-06-29: Send readConcern with all snapshot session commands.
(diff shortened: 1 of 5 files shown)


jyemin (commented):
I like the approach.  I implemented the changes in the Java driver, though, and some of the test fail on sharded 5.0 clusters.  For example "Server returns an error on insertOne with snapshot" fails because the server does not throw an exception.  I confirmed in the shell:

```
MongoDB Enterprise mongos> db.runCommand({'insert': 'test', documents: [{}], readConcern: {level: 'snapshot'}})
{
	"n" : 1,
	"ok" : 1,
	"$clusterTime" : {
		"clusterTime" : Timestamp(1625064185, 7),
		"signature" : {
			"keyId" : NumberLong(0)
		}
	},
	"operationTime" : Timestamp(1625064185, 7)
}
```

Not sure what we should do.  I suppose we could disable these tests on sharded clusters and report this as a server bug.

source/sessions/tests/unified/snapshot-sessions-unsupported-ops.json
```diff
@@ -0,0 +1,468 @@
...
+              "commandFailedEvent": {
+                "commandName": "listIndexes"
+              }
+            }
+          ]
+        }
+      ]
+    },
+    {
+      "description": "Server returns an error on runComand with snapshot",
```
jyemin, line 420: Nit: runComand => runCommand

jyemin (changes requested):

ShaneHarvey:
AH that's 100% correct. I had forgot to check the sharded test results. I also see mongos not validating readConcern on insert, update, and delete commands. I just opened [SERVER-58176](https://jira.mongodb.org/browse/SERVER-58176).

jyemin:
Cool.  Can you re-arrange the tests so that they are skipped on sharded clusters for now?

ShaneHarvey:
> Can you re-arrange the tests so that they are skipped on sharded clusters for now?

Done.

source/sessions/tests/unified/snapshot-sessions-unsupported-ops.json
```diff
@@ -457,7 +482,7 @@
             },
             {
               "commandFailedEvent": {
-                "commandName": "insert"
+                "listCollections": "insert"
```
jyemin, line 485: I think this should be:

     commandName: "listCollections"

jyemin (changes requested):

ShaneHarvey:
@ajdavis Does this approach look good to you? For more background see the description in SERVER-58176.

source/sessions/tests/unified/snapshot-sessions-unsupported-ops.json
ShaneHarvey: Fixed.

jyemin (approved):
This required a change in approach for the implementation, but it wasn't too big a deal in the end.  All the tests are passing.

LGTM

source/sessions/snapshot-sessions.rst
```diff
@@ -233,6 +234,12 @@ Lists of commands that support snapshot reads:
 2. aggregate
 3. distinct
 
+Sending readConcern to the server on all commands
+=================================================
+
+Drivers MUST set the readConcern ``level`` and ``atClusterTime`` fields (as
+outlined above) on all commands in a snapshot session.
```
ajdavis, line 241: I'd add, "even commands like insert and update that do not accept a readConcern. This ensures the server will return an error for invalid operations, such as writes, within a session configured for snapshot reads." Make the implicit explicit.

ajdavis:
Yeah, if we can fix SERVER-58176 in time, then I like this approach a lot.

source/sessions/snapshot-sessions.rst
ShaneHarvey: Done.

ShaneHarvey (commented):
Thanks for the quick reviews!

rstam (approved):
Makes sense. LGTM

ajdavis (approved):

PR #1035, opened 2021-07-02
DRIVERS-1835 add a load balancer DNS test case resolving to 0 hosts

kevinAlbs:
__Motivation:__ Connecting to a load balanced deployment requires one host in the resolved URI. There is a test to assert that SRV records do not add more than one host: https://github.com/mongodb/specifications/blob/44f83106033a609f2af40e3e4d99852a03096ba5/source/initial-dns-seedlist-discovery/tests/load-balanced/loadBalanced-true-multiple-hosts.json. It seems natural to test the case that no hosts are returned.

This is copied over from the corresponding replica-set test:
https://github.com/mongodb/specifications/blob/44f83106033a609f2af40e3e4d99852a03096ba5/source/initial-dns-seedlist-discovery/tests/replica-set/no-results.yml

Validated with C drivers tests: https://spruce.mongodb.com/version/60dfaaed9ccd4e340bf1e357


durran (approved):
LGTM

PR #1036, opened 2021-07-07
DRIVERS-1836 Skip tests with $out stage on serverless

alcaeus:
DRIVERS-1836

Note that the CRUD v1 tests still contain unavailable functionality (namely `collate` which has also been disabled), but I don't believe we've added support for the `serverless` runOnRequirement in that test format, so I'd avoid touching those tests. Drivers that run the CRUD v1 tests on serverless should skip or exclude the offending tests separately.

### DIFF
--- a/source/retryable-reads/tests/aggregate.json
+++ b/source/retryable-reads/tests/aggregate.json
@@ -12,8 +12,7 @@
       "topology": [
         "sharded",
         "load-balanced"
-      ],
-      "serverless": "forbid"
+      ]
     }
   ],
   "database_name": "retryable-reads-tests",
--- a/source/retryable-reads/tests/aggregate.json
+++ b/source/retryable-reads/tests/aggregate.json
@@ -12,7 +12,8 @@
       "topology": [
         "sharded",
         "load-balanced"
-      ]
+      ],
+      "serverless": "forbid"
     }
   ],
   "database_name": "retryable-reads-tests",
--- a/source/retryable-reads/tests/aggregate.yml
+++ b/source/retryable-reads/tests/aggregate.yml
@@ -5,6 +5,7 @@ runOn:
     -
         minServerVersion: "4.1.7"
         topology: ["sharded", "load-balanced"]
+        serverless: "forbid"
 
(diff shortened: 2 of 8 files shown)


jmikola:
> Drivers that run the CRUD v1 tests on serverless should skip or exclude the offending tests separately.

No objections there, but it should be clearly specified in the DRIVERS ticket when languages sync up with these tests. https://github.com/mongodb/specifications/commit/3586db8b1ed90d835737f6226bb3ff9012cfc5de added `serverless` to the legacy transactions format, but that already included something similar to `runOn` in the unified format.

CRUD v1 just has top-level fields (https://github.com/mongodb/specifications/tree/master/source/crud/tests#legacy-v1-test-format-for-single-operations) with no topology checks IIRC.

source/retryable-reads/tests/aggregate.yml
```diff
@@ -5,6 +5,7 @@ runOn:
     -
         minServerVersion: "4.1.7"
         topology: ["sharded", "load-balanced"]
+        serverless: "forbid"
```
jmikola, line 8: `topology: ["single", "replicaset"]` in the other requirement would never match serverless since it'd either identify as sharded or load-balanced, correct?

jmikola (approved):

source/retryable-reads/tests/aggregate.json
```diff
@@ -12,7 +12,8 @@
       "topology": [
         "sharded",
         "load-balanced"
-      ]
+      ],
+      "serverless": "forbid"
```
jyemin, line 16: These tests aren't failing for me.  Do they need to be disabled?

source/retryable-reads/tests/aggregate.yml
alcaeus: From my understanding, yes.

source/retryable-reads/tests/aggregate.json
alcaeus: You're right, these tests set a fail point, so they will never trigger the underlying error. I've removed the requirement for this test.

PR #1041, opened 2021-07-15
DRIVERS-1851 Skip crud-v1 tests for $out and collation on Serverless

patrickfreed:
DRIVERS-1851

Atlas Serverless doesn't currently support `$out` aggregations or collation, so the associated crud-v1 tests need to be skipped. This involved slightly amending the crud-v1 format.

### DIFF
--- a/source/crud/tests/README.rst
+++ b/source/crud/tests/README.rst
@@ -21,12 +21,12 @@ Subdirectories for Test Formats
 
 This document describes a legacy format for CRUD tests: legacy-v1, which dates back
 to the first version of the CRUD specification. New CRUD tests should be written
-in the `unified test format <../../../../unified-test-format/unified-test-format.rst>`_
+in the `unified test format <../../../unified-test-format/unified-test-format.rst>`_
 and placed under ``unified/``. Until such time that all original tests have been ported
 to the unified test format, tests in each format will be grouped in their own subdirectory:
 
 - ``v1/``: Legacy-v1 format tests
-- ``unified/``: Tests using the `unified test format <../../../../unified-test-format/unified-test-format.rst>`_
+- ``unified/``: Tests using the `unified test format <../../../unified-test-format/unified-test-format.rst>`_
 
 Since some drivers may not have a unified test runner capable of executing tests
 in all two formats, segregating tests in this manner will make it easier for
@@ -59,7 +59,7 @@ single operation. Notable differences from the legacy-v2 format are as follows:
   there is no corresponding bound on the required server version. The
   ``serverless`` requirement behaves the same as the ``serverless`` field of the
   `unified test format's runOnRequirement
-  <../../../../unified-test-format/unified-test-format.rst#runonrequirement>`_.
+  <../../../unified-test-format/unified-test-format.rst#runonrequirement>`_.
 
 The legacy-v1 format should not conflict with the newer, multi-operation format
 used by other specs (e.g. Transactions). It is possible to create a unified test
--- a/source/crud/tests/README.rst
+++ b/source/crud/tests/README.rst
@@ -21,12 +21,12 @@ Subdirectories for Test Formats
 
 This document describes a legacy format for CRUD tests: legacy-v1, which dates back
 to the first version of the CRUD specification. New CRUD tests should be written
-in the `unified test format <../../../unified-test-format/unified-test-format.rst>`_
+in the `unified test format <../../unified-test-format/unified-test-format.rst>`_
 and placed under ``unified/``. Until such time that all original tests have been ported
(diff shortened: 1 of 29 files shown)


source/crud/tests/v1/write/bulkWrite-collation.json
```diff
@@ -22,6 +22,7 @@
     }
   ],
   "minServerVersion": "3.4",
+  "serverless": "forbid",
```
MikalaiMazurenka, line 25: I missed this test because it passed in C# driver.
The `bulkWrite` operation didn't fail despite having a `collation` present. The difference with other operations is that they had `collation` specified as operation option while `bulkWrite` has this option set for the individual requests.
Which means that it is either an intended behavior or a bug in serverless implementation.

source/crud/tests/README.rst
```diff
@@ -21,12 +21,12 @@ Subdirectories for Test Formats
...
 This document describes a legacy format for CRUD tests: legacy-v1, which dates back
 to the first version of the CRUD specification. New CRUD tests should be written
-in the `unified test format <../../../../unified-test-format/unified-test-format.rst>`_
+in the `unified test format <../../unified-test-format/unified-test-format.rst>`_
 and placed under ``unified/``. Until such time that all original tests have been ported
 to the unified test format, tests in each format will be grouped in their own subdirectory:
 
 - ``v1/``: Legacy-v1 format tests
-- ``unified/``: Tests using the `unified test format <../../../../unified-test-format/unified-test-format.rst>`_
+- ``unified/``: Tests using the `unified test format <../../unified-test-format/unified-test-format.rst>`_
```
benjirewis, line 29: Ah thanks for fixing these.

benjirewis (approved):

source/crud/tests/v1/write/deleteMany-collation.json
```diff
@@ -14,6 +14,7 @@
     }
   ],
   "minServerVersion": "3.4",
+  "serverless": "forbid",
```
MikalaiMazurenka, line 17: This test didn't fail in C# driver either, because the `deleteMany` is executed as `bulkWrite` operation with a single `deleteMany` request.

source/crud/tests/v1/write/deleteOne-collation.json
```diff
@@ -14,6 +14,7 @@
     }
   ],
   "minServerVersion": "3.4",
+  "serverless": "forbid",
```
MikalaiMazurenka, line 17: Same as for `deleteMany`

source/crud/tests/v1/write/replaceOne-collation.json
```diff
@@ -10,6 +10,7 @@
     }
   ],
   "minServerVersion": "3.4",
+  "serverless": "forbid",
```
MikalaiMazurenka, line 13: Same behavior with the operation being an individual request of `bulkWrite` operation.

source/crud/tests/v1/write/updateMany-collation.json
```diff
@@ -14,6 +14,7 @@
     }
   ],
   "minServerVersion": "3.4",
+  "serverless": "forbid",
```
MikalaiMazurenka, line 17: Same behavior with the operation being an individual request of `bulkWrite` operation.

source/crud/tests/v1/write/updateOne-collation.json
```diff
@@ -10,6 +10,7 @@
     }
   ],
   "minServerVersion": "3.4",
+  "serverless": "forbid",
```
MikalaiMazurenka, line 13: Same behavior with the operation being an individual request of `bulkWrite` operation.

MikalaiMazurenka (commented):
Looks good, but before approving I will need to check with serverless team whether `bulkWrite` operation succeeding on serverless when request has a `collation` is an intended behavior. If it is, changes in `bulkWrite` and adjacent tests should be reverted. If it's a server bug, then no further changes required.

emadum (approved):
Crud v1 passes in Node with these changes 👍

source/crud/tests/v1/write/bulkWrite-collation.json
emadum: To add another data point, these tests were [failing](https://evergreen.mongodb.com/task_log_raw/mongo_node_driver_ubuntu1804_test_serverless_test_serverless_patch_fa4352d297e70b83e48c1cb225eaab7cbe7404ab_60f07dbf0ae6061fe3feb9a7_21_07_15_18_26_13/0?type=T) in Node.

MikalaiMazurenka (approved):
LGTM!
This indeed was a bug

PR #1042, opened 2021-07-22
Test top-level errorLabel with writeConcernError

jyemin:
Add some unified tests that add RetryableWriteError as a top-level
errorLabel of a writeConcernError

DRIVERS-1385

I added these now because I'm fixing a user-reported bug in the Java driver and needed some tests to confirm that it's fixed.

Isabel suggested that I add them as the first unified tests in the retryable-writes test directory.  I didn't move the legacy tests to a legacy directory but I can if you'd like.

This probably isn't a complete fix for DRIVERS-1385 but it's sufficient for my needs.


jmikola:
> I didn't move the legacy tests to a legacy directory but I can if you'd like.

Please do and revise the README as needed. That would be consistent with what has been done for other specs when unified tests are introduced (e.g. https://github.com/mongodb/specifications/commit/39ccfce33adae2940ee24e7a7ad68fb3f9125b64 from https://github.com/mongodb/specifications/pull/1010).

jyemin:
Moved legacy tests and updated test README

jyemin:
Instead of a new ticket, and two synchronizations for every driver, can we consider pushing this patch, and hold of on language tickets until you two decide that sufficient tests have been implemented for DRIVERS-1385 (incorporating these initial tests).  Once all tests are written we can move 1385 to Implementing and open language tickets.

source/retryable-writes/tests/README.rst
```diff
@@ -9,8 +9,14 @@ Retryable Write Tests
...
 ============
 
-The YAML and JSON files in this directory tree are platform-independent tests
-that drivers can use to prove their conformance to the Retryable Writes spec.
+Tests in this directory are platform-independent tests that drivers can use to
+prove their conformance to the Retryable Writes specification.
+
+Tests in the ``unified`` directory are implemented in the
+`Unified Test Format <../../unified-test-format/unified-test-format.rst>`__
+and require schema version 1.0.
```
jmikola, line 17: I'd suggest removing "and require schema version 1.0", as this seems very likely to get out-of-date with a test change (e.g. needing to skip something for serverless). And it doesn't add much since the schema version for each test is already expressed within.

jmikola (commented):
> Instead of a new ticket

In that case, please make sure DRIVERS-1385 is referenced in the final commit when this gets merged.

source/retryable-writes/tests/unified/insertOne-serverErrors.yml
```diff
@@ -0,0 +1,80 @@
...
+      - { _id: 2, x: 22 }
+
+tests:
+  - description: "InsertOne succeeds after retryable writeConcernError"
+    runOnRequirements:
+      - minServerVersion: "4.0"
+        topologies: [ replicaset ]
+      - minServerVersion: "4.1.7"
+        # Original test uses "sharded", but retryable writes requires a sharded
+        # cluster backed by replica sets
```
jmikola, line 37: What is "original test" referring to? There is no existing test with this title. I'd suggest removing that line. No objections if you want to still mention that "retryable writes requires a sharded cluster backed by replica sets", but I don't think that comment provides much value since the `topologies` line says the same thing. The bigger question for a casual reader might be why 4.1.7 is required here. I assume that is for [failCommand](https://github.com/mongodb/mongo/wiki/The-%22failCommand%22-fail-point), based on it appearing in other test files, although the wiki suggests that was added in mongos 4.1.5. I'm not suggesting we change the `minServerVersion` here, though.

jyemin:
> > Instead of a new ticket
> 
> In that case, please make sure [DRIVERS-1385](https://jira.mongodb.org/browse/DRIVERS-1385) is referenced in the final commit when this gets merged.

It is referenced already, in the last line of the commit message.  That's enough for it to get added to the ticket.  But I'll move it to the first line when I squash since that's the practice in this repo.

source/retryable-writes/tests/README.rst
jyemin: I copied that from another spec which had it, but I agree with you.  

Done.

source/retryable-writes/tests/unified/insertOne-serverErrors.yml
jyemin: Ah that was copied from poc-retryable-writes.yml. Happy to remove.

jyemin: Done

source/retryable-writes/tests/unified/bulkWrite-serverErrors.yml
```diff
@@ -0,0 +1,96 @@
...
+                document: { _id: 3, x: 33 }
+            - deleteOne:
+                filter: { _id: 2 }
+        expectResult:
+          deletedCount: 1
+          insertedCount: 1
+          matchedCount: 0
+          modifiedCount: 0
+          upsertedCount: 0
+          insertedIds: { "0": 3 }
```
jmikola, line 65: `insertedIds` in InsertManyResult and BulkWriteResult should be wrapped with `$$unsetOrMatches` since it's optional (see https://github.com/mongodb/specifications/commit/63d3de940b8a8aca43242db6707df84dc472449b for context).

```suggestion
          insertedIds: { $$unsetOrMatches: { 0: 3 } }
```

Also using a numeric literal instead of a string for the key since the map is indexed numerically.

source/retryable-writes/tests/unified/insertOne-serverErrors.yml
```diff
@@ -0,0 +1,78 @@
...
+              errorLabels: [RetryableWriteError] # top-level error labels
+              writeConcernError:
+                code: 91 # ShutdownInProgress
+                errmsg: "Replication is being shut down"
+      - name: insertOne
+        object: *collection0
+        arguments:
+          document: { _id: 3, x: 33 }
+        expectResult:
+          insertedId: 3
```
jmikola, line 56:
```suggestion
          $$unsetOrMatches: { insertedId: { $$unsetOrMatches: 3 } }
```

See https://github.com/mongodb/specifications/commit/44f83106033a609f2af40e3e4d99852a03096ba5#diff-7b80332d6e4f34d2b705cb9fab2c176444ad002089466b6cf47b3dd24b0b8383R38 for context.

jmikola (commented):
LGTM after both suggestions are applied.

source/retryable-writes/tests/unified/bulkWrite-serverErrors.yml
jyemin: Done

source/retryable-writes/tests/unified/insertOne-serverErrors.yml
jyemin: That's quite a mouthful.

Done.

jyemin:
@isabelatkinson should I wait for your review as well?

isabelatkinson (approved):
LGTM, just tested these out in the Rust driver and they pass :)

PR #1047, opened 2021-08-06
DRIVERS-1879 Remove modifiers option from command monitoring spec test

ShaneHarvey:
The modifiers option was removed from the crud spec long ago. See https://jira.mongodb.org/browse/DRIVERS-1879.

I'm testing this change in https://github.com/mongodb/mongo-python-driver/pull/696


kmahar (approved):
LGTM, thanks Shane! test passes in Swift and this allowed me to delete some code from our runner that was hacking around the options being specified like this.

PR #1048, opened 2021-08-12
DRIVERS-1864: bump maxWireVersion in SDAM tests

nbbeeken:
Bumps maxWireVersion in SDAM tests to 6 to prevent failures when DRIVERS-1315 is implemented.


ShaneHarvey (approved):
LGTM

PR #1049, opened 2021-08-12
DRIVERS-1883 Split unified runner valid-fail malformed operation tests into separate files

kmahar:
Opening a PR for your consideration, as I manually split up the tests myself for Swift testing purposes already; see DRIVERS-1883 for details. (I have left it in "needs triage" for now to avoid spinning out language tickets in the event we don't end up going forward with this.)

### DIFF
--- a/source/unified-test-format/tests/valid-fail/entity-findCursor-malformed.json
+++ b/source/unified-test-format/tests/valid-fail/entity-findCursor-malformed.json
@@ -1,5 +1,5 @@
 {
-  "description": "entity-find-cursor",
+  "description": "entity-findCursor-malformed",
   "schemaVersion": "1.3",
   "createEntities": [
     {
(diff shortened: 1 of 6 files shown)


kmahar:
I ran into a similar issue with `valid-fail/entity-find-cursor.yml`. "createFindCursor fails if filter is not specified" produces a decoding error for us prior to test runs as the operation is malformed, whereas the other two tests check behavior that doesn't fail until we actually try to execute tests (since the failures result from entities being unexpectedly missing from the map).

I'm happy to update that as well (and any other cases like this, where tests failing at different stages are mixed in one file, if I find them), but I'll wait to see what you two think on this subject.

jmikola:
No objections to splitting up `ignoreResultAndError` and `entity-find-cursor`. In that case, I would suggest adding a comment to the top of the YAML file explaining that you did so as a convenience for drivers that decode the entire test file up front and are liable to error before any evaluation begins.

I'd also suggest renaming `entity-find-cursor` to `entity-findCursor` (in both tests after your split) since the entity itself is called a [FindCursor](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#supported-entity-types) in the spec. We don't need a hyphen interrupting a spec term.

kmahar:
@jmikola SGTM, I've added comments to the YAML files, and replaced `find-cursor` instances with `findCursor`.

isabelatkinson (approved):

jmikola (approved):

PR #1050, opened 2021-08-19
DRIVERS-1141 Add serverConnectionId to command monitoring events

benjirewis:
DRIVERS-1141

Adds a new `serverConnectionId` field to `CommandStartedEvent`, `CommandSucceededEvent` and `CommandFailedEvent` so users of APM can correlate client-side `connectionId`s with server-side `serverConnectionId`s (see ticket for more details). Adds a new unified command monitoring test to verify that `serverConnectionId` is present in command monitoring events. 

Adds a new `hasServerConnectionId` field to `commandStartedEvent`, `commandSucceededEvent` and `commandFailedEvent` in the unified test format. This field is similar to the existing [`hasServiceId`](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#hasserviceid) field and lets us assert the existence of `serverConnectionId` without explicitly checking its value (which will always be dependent on the connected server).

My implementation in the Go driver is available [here](https://github.com/benjirewis/mongo-go-driver/tree/serverConnectionID.2085). And here is a [patch](https://spruce.mongodb.com/version/611ec60e30661523556162d2/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) of that implementation passing our latest tests.

### DIFF
--- a/source/command-monitoring/command-monitoring.rst
+++ b/source/command-monitoring/command-monitoring.rst
@@ -304,7 +304,7 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
      * the connection id and is returned by the hello or legacy hello response as "connectionId"
      * from the server on 4.2+.
      */
-    serverConnectionId: Int64;
+    serverConnectionId: Optional<Int64>;
 
     /**
      * Returns the service id for the command when the driver is in load balancer mode.
@@ -357,7 +357,7 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
      * the connection id and is returned by the hello or legacy hello response as "connectionId"
      * from the server on 4.2+.
      */
-    serverConnectionId: Int64;
+    serverConnectionId: Optional<Int64>;
 
     /**
      * Returns the service id for the command when the driver is in load balancer mode.
--- a/source/command-monitoring/command-monitoring.rst
+++ b/source/command-monitoring/command-monitoring.rst
@@ -300,6 +300,13 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
     connectionId: ConnectionId;
 
     /**
+     * Returns the server connection id for the command. The server connection id is distinct from
+     * the connection id and is returned by the hello or legacy hello response as "connectionId"
+     * from the server on 4.2+.
+     */
+    serverConnectionId: Int64;
+
+    /**
      * Returns the service id for the command when the driver is in load balancer mode.
      * For drivers that wish to include this in their ConnectionId object, this field is
      * optional.
@@ -344,6 +351,13 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
      * The name of this field is flexible to match the object that is returned from the driver.
      */
     connectionId: ConnectionId;
+
+    /**
+     * Returns the server connection id for the command. The server connection id is distinct from
+     * the connection id and is returned by the hello or legacy hello response as "connectionId"
+     * from the server on 4.2+.
+     */
+    serverConnectionId: Int64;
 
     /**
      * Returns the service id for the command when the driver is in load balancer mode.
@@ -391,6 +405,13 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
      * The name of this field is flexible to match the object that is returned from the driver.
      */
     connectionId: ConnectionId;
+
+    /**
+     * Returns the server connection id for the command. The server connection id is distinct from
+     * the connection id and is returned by the hello or legacy hello response as "connectionId"
+     * from the server on 4.2+.
+     */
+    serverConnectionId: Optional<Int64>;
 
     /**
      * Returns the service id for the command when the driver is in load balancer mode.
--- a/source/command-monitoring/command-monitoring.rst
+++ b/source/command-monitoring/command-monitoring.rst
@@ -302,7 +302,8 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
     /**
      * Returns the server connection id for the command. The server connection id is distinct from
      * the connection id and is returned by the hello or legacy hello response as "connectionId"
-     * from the server on 4.2+.
+     * from the server on 4.2+. Drivers MAY use a wider type to represent the server connection ID
+     * value, but the server's behavior is to return an Int32.
      */
     serverConnectionId: Optional<Int32>;
 
@@ -355,7 +356,8 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
     /**
      * Returns the server connection id for the command. The server connection id is distinct from
      * the connection id and is returned by the hello or legacy hello response as "connectionId"
-     * from the server on 4.2+.
+     * from the server on 4.2+. Drivers MAY use a wider type to represent the server connection ID
+     * value, but the server's behavior is to return an Int32.
      */
     serverConnectionId: Optional<Int32>;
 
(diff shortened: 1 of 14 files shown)


ShaneHarvey:
Should we add this field to CMAP and ServerHeartbeat events as well?

source/command-monitoring/command-monitoring.rst
```diff
@@ -392,6 +406,13 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
      */
     connectionId: ConnectionId;
 
+    /**
+     * Returns the server connection id for the command. The server connection id is distinct from
+     * the connection id and is returned by the hello or legacy hello response as "connectionId"
+     * from the server on 4.2+.
+     */
+    serverConnectionId: Optional<Int64>;
```
jmikola, line 414: Why is this the only case where `serverConnectionId` is optional?

source/unified-test-format/unified-test-format.rst
```diff
@@ -1170,6 +1176,14 @@ that the field is set and is a non-empty BSON ObjectId (i.e. all bytes of the
 ObjectId are not 0). If false, test runners MUST assert that the field is not
 set or is an empty BSON ObjectId.
 
+hasServerConnectionId
+`````````````````````
+
+This field is an optional boolean that specifies whether or not the
+``serverConnectionId`` field of an event is set. If true, test runners MUST
+assert that the field is set and is an Int64. If false, test runners MUST assert
```
jmikola, line 1184: Is there any restriction on the value if the field is set (i.e. non-zero)? Just curious if there's relevance for drivers that might not actually implement the field as an optional type. For example, libmongoc might want to return an `int64_t` and use zero (or a constant defined to -1) to signify an unset value, rather than have to create a struct that represents an optional int64_t like it did for booleans in Versioned API.

Perhaps this comment is more relevant to the APM spec itself where the fields are documented.

source/command-monitoring/command-monitoring.rst
benjirewis: Typo... added `Optional` to all.

source/unified-test-format/unified-test-format.rst
benjirewis: The [underlying type](https://github.com/mongodb/mongo/blob/master/src/mongo/db/client.h#L60) of `ConnectionId` in the server code is a `long long int`, so `serverConnectionId` _could_ be <= 0. That said, judging from [this internal logic](https://github.com/mongodb/mongo/blob/master/src/mongo/db/client.h#L184-L189), I believe a connection to an external user (i.e. driver) invariantly has a positive `serverConnectionId`. A `ConnectionId` of `0` seems to be used for internal system connections.

I've added some wiggle room for drivers without native Optional-like structures similar to the qualifications of  `hasServiceId` above.

source/unified-test-format/schema-1.6.json
```diff
@@ -0,0 +1,466 @@
+{
```
benjirewis, line 1: I had to add a new schema version 1.6 for `hasServerConnectionId` per the rules on UTF schema version bumps [here](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#impact-of-spec-changes-on-schema-version).

source/unified-test-format/unified-test-format.rst
jyemin: Looking at the `hello` response, the server is returning the value as a BSON Int32.  In practice, it's hard to imagine ever needing a value greater than max Int32, but I suppose we can be flexible, but FYI Java driver (which implemented this a long time ago) models the connectionId as an Int32.

benjirewis: It looks like the server is returning a BSON Int32 despite it being stored as a `long long int` server-side. I also don't think the value will exceed max Int32, so I've switched to an Int32 instead of Int64. That way, the Java driver is already compliant with the spec, too. If anyone else has a strong opinion against I'll leave this thread open.

benjirewis:
I'm investigating adding `serverConnectionId` to relevant SDAM events (ServerHeartbeatStartedEvent/ServerHeartbeatSucceededEvent/ServerHeartbeatFailedEvent) and relevant CMAP events
(ConnectionReadyEvent/ConnectionClosedEvent/ConnectionCheckedOutEvent/ConnectionCheckedInEvent) and testing its presence with spec/prose tests. I'll move this PR out of draft when I've finished.

source/unified-test-format/unified-test-format.rst
jmikola: Int32 sounds fine to me. If we codify that in a spec, it gives us an excuse should the server ever decide to deviate from its current behavior (server folks do reference our specs from time to time). You can easily include a note that drivers MAY use a wider type to represent the value but the server's behavior is to return an Int32.

benjirewis: That sounds good. I've included a note that drivers may use a wider type.

benjirewis:
We may add `serverConnectionId` to relevant SDAM events and relevant CMAP events and testing its presence with spec/prose tests as part of a separate ticket DRIVERS-1897. DRIVERS-1141 will only add the field to command monitoring events.

source/unified-test-format/schema-1.6.json
```diff
@@ -0,0 +1,466 @@
...
+      "properties": {
+        "commandStartedEvent": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "command": { "type": "object" },
+            "commandName": { "type": "string" },
+            "databaseName": { "type": "string" },
+            "hasServiceId": { "type": "boolean" },
+            "hasServerConnectionId": { "type": "boolean" }
```
jmikola, line 272: Please add tests for the schema itself (one per event type), similar to what was done for `hasServiceId`. See: [invalid/expectedCommandEvent-commandFailedEvent-hasServiceId-type.yml](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/tests/invalid/expectedCommandEvent-commandFailedEvent-hasServiceId-type.yml).

jmikola (commented):
LGTM once "invalid" tests are added for the schema changes.

source/command-monitoring/tests/unified/server-connection-id.yml
```diff
@@ -0,0 +1,56 @@
+description: "server-connection-id"
```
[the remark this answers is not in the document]
benjirewis: Good idea. Added `pre-42-server-connection-id` tests with `hasServerConnectionId: false`.

source/unified-test-format/schema-1.6.json
benjirewis: Sounds good. Added three new `invalid` tests for `hasServerConnectionId` with `commandStartedEvent`, `commandSucceededEvent` and `commandFailedEvent`.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1170,6 +1176,14 @@ that the field is set and is a non-empty BSON ObjectId (i.e. all bytes of the
 ObjectId are not 0). If false, test runners MUST assert that the field is not
 set or is an empty BSON ObjectId.
 
+hasServerConnectionId
+`````````````````````
+
+This field is an optional boolean that specifies whether or not the
+``serverConnectionId`` field of an event is set. If true, test runners MUST
+assert that the field is set and is a positive Int32. If false, test runners
+MUST assert that the field is not set or is a nonpositive Int32.
```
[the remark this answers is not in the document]
benjirewis: Yep! Added an elaboration similar to the one you wrote.

kmahar (approved):
lgtm, nice work!

kevinAlbs (approved):
LGTM!

jmikola (approved):

kmahar:
@benjirewis sorry I just realized, but I think the spec versions, last updated dates, and changelogs may need to be updated as well?

benjirewis:
@kmahar all good; I missed those as well. Opened #1061 to add those changes in with the same DRIVERS-1141 ticket.

PR #1052, opened 2021-08-23
DRIVERS-1885: Allow runOnRequirement conditions to be evaluated in any order

jmikola:
https://jira.mongodb.org/browse/DRIVERS-1885

### DIFF
--- a/source/unified-test-format/unified-test-format.rst
+++ b/source/unified-test-format/unified-test-format.rst
@@ -3,13 +3,13 @@ Unified Test Format
 ===================
 
 :Spec Title: Unified Test Format
-:Spec Version: 1.5.4
+:Spec Version: 1.5.5
 :Author: Jeremy Mikola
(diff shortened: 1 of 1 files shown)


source/unified-test-format/unified-test-format.rst
```diff
@@ -3,13 +3,13 @@ Unified Test Format
 ===================
 
 :Spec Title: Unified Test Format
-:Spec Version: 1.5.4
+:Spec Version: 1.5.5
```
jmikola, line 6: Note: I opted to make this a patch bump since I did not expect the change to have any effect on test file execution. All conditions within a `runOnRequirement` must be satisfied for the requirement itself to be considered satisfied, so ordering should make no difference.

Avoiding a bump to 1.6.0 here always avoids a potential conflict with #1050.


 change has no effect on test file execution. to avoid a potential conflict with #1050

kmahar (approved):

kevinAlbs (approved):

PR #1053, opened 2021-08-23
DRIVERS-1603 add serverless test README

kevinAlbs:
**Summary**
- Add serverless-testing/README.rst.
- Update serverless-testing/README.rst add load balancer tests.
- Add `serverless: forbid` to spec tests that cannot run against serverless instances behind a load balancer.

**Motivation**
Serverless tests are currently spawning instances with the GCP backing provider. The GCP backing provider does not run behind a load balancer. That will change soon to run behind a load balancer. This updates specification tests noted in the [Spec: Atlas Serverless Testing](https://docs.google.com/document/d/1Eo8bYnWYeeTEETa1CtIfO1BfVednhLEkpI4OUTYyJec/edit) to exclude tests that cannot succeed against a load balancer fronting multiple Atlas Proxy instances.

**Caveats**
- Running against load balanced serverless instances further restricts the set of tests we can run against serverless. But there is no foreseeable alternative.
- The analog of `SINGLE_MONGOS_LB_URI` for a serverless instance is `SINGLE_ATLASPROXY_SERVERLESS_URI` which is a URI to one Atlas Proxy. It is not in front of a load balancer, but still responds with a `serviceId`

**Testing**
A POC of these tests was run with the Go driver in: https://github.com/mongodb/mongo-go-driver/pull/719 against the AWS backing provider (proposed in https://github.com/mongodb-labs/drivers-evergreen-tools/pull/168).

**Changelog**
- Friday, August 27, 2021
    - Add the `serverless` field to the `runOn` block for the retryable reads test runner.
    - Clarify behavior of `useMultipleMongoses` in retryable reads / writes tests for `LoadBalanced` topologies.
- Thursday, August 26, 2021
    - Add `MULTI_ATLASPROXY_SERVERLESS_URI` and `SINGLE_ATLASPROXY_SERVERLESS_URI`
    - Add back retryable reads and writes tests.
    - Remove `serverless: forbid` on tests relying on a URI to a single service.
- Wednesday, August 25, 2021
	- Revert changes to retryable reads and retryable writes. Tests relying on failpoints cannot run 	against a load balancer fronting multiple servers.
	- Add serverless-testing/README.rst

### DIFF
--- a/source/load-balancers/tests/sdam-error-handling.yml
+++ b/source/load-balancers/tests/sdam-error-handling.yml
@@ -65,6 +65,10 @@ initialData:
 
 tests:
   - description: only connections for a specific serviceId are closed when pools are cleared
+    runOnRequirements:
+    # This test assumes that two sequential connections receive different serviceIDs.
+    # Sequential connections to a serverless instance may receive the same serviceID.
+    - serverless: forbid
     operations:
       # Create two cursors to force two connections.
       - name: createFindCursor
--- a/source/load-balancers/tests/sdam-error-handling.yml
+++ b/source/load-balancers/tests/sdam-error-handling.yml
@@ -4,6 +4,7 @@ schemaVersion: '1.3'
 
 runOnRequirements:
   - topologies: [ load-balanced ]
+    serverless: forbid # Test sets useMultipleMongoses: false. Serverless only exposes a load balancer fronting multiple Atlas Proxy instances. See DRIVERS-1882.
 
 _yamlAnchors:
   observedEvents: &observedEvents
--- a/source/load-balancers/tests/sdam-error-handling.yml
+++ b/source/load-balancers/tests/sdam-error-handling.yml
@@ -4,7 +4,6 @@ schemaVersion: '1.3'
 
 runOnRequirements:
   - topologies: [ load-balanced ]
-    serverless: forbid # Test sets useMultipleMongoses: false. Serverless only exposes a load balancer fronting multiple Atlas Proxy instances. See DRIVERS-1882.
 
 _yamlAnchors:
   observedEvents: &observedEvents
--- a/source/retryable-reads/tests/aggregate-merge.yml
+++ b/source/retryable-reads/tests/aggregate-merge.yml
@@ -1,7 +1,6 @@
 runOn:
   -
     minServerVersion: "4.1.11"
-    serverless: "forbid"
 
 database_name: &database_name "retryable-reads-tests"
 collection_name: &collection_name "coll"
--- a/source/retryable-reads/tests/aggregate-merge.yml
+++ b/source/retryable-reads/tests/aggregate-merge.yml
@@ -1,6 +1,7 @@
 runOn:
   -
     minServerVersion: "4.1.11"
+    serverless: "forbid"
 
 database_name: &database_name "retryable-reads-tests"
 collection_name: &collection_name "coll"
--- a/source/retryable-writes/tests/README.rst
+++ b/source/retryable-writes/tests/README.rst
@@ -191,10 +191,18 @@ Each YAML file has the following keys:
 
   - ``clientOptions``: Parameters to pass to MongoClient().
 
-  - ``useMultipleMongoses`` (optional): If ``true``, the MongoClient for this
-    test should be initialized with multiple mongos seed addresses. If ``false``
-    or omitted, only a single mongos address should be specified. This field has
-    no effect for non-sharded topologies.
+  - ``useMultipleMongoses`` (optional): If ``true``, and the topology type is
+    ``Sharded``, the MongoClient for this test should be initialized with multiple
+    mongos seed addresses. If ``false`` or omitted, only a single mongos address
+    should be specified.
+
+    If ``true``, and the topology type is ``LoadBalanced``, the MongoClient for
+    this test should be initialized with the URI of the load balancer fronting
+    multiple servers. If ``false`` or omitted, the MongoClient for this test
+    should be initialized with the URI of the load balancer fronting a single
+    server.
+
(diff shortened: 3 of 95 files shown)


source/retryable-reads/tests/aggregate-merge.yml
```diff
@@ -1,6 +1,7 @@
 runOn:
   -
     minServerVersion: "4.1.11"
+    serverless: "forbid"
```
jmikola, line 4: Is this missing a comment about why serverless is forbidden here? I assume it's because of fail points.

patrickfreed (commented):
It may be simpler to update the prose of the serverless testing spec to just omit retryable reads and writes rather than manually annotating all the tests, since I think all of them will need to be skipped due to failpoint usage.

jmikola:
> It may be simpler to update the prose of the serverless testing spec to just omit retryable reads and writes rather than manually annotating all the tests

I just realized that [Retryable Writes: Test Format](https://github.com/mongodb/specifications/blob/master/source/retryable-writes/tests/README.rst#test-format) was never updated with `serverless` syntax, as was done for [Retryable Reads: Test Format](https://github.com/mongodb/specifications/blob/master/source/retryable-reads/tests/README.rst#test-format). I agree that omitting both specs entirely fomr the serverless testing spec might be easier.

I suppose the `serverless` syntax for retryable reads would be left in place, since that has already been done by https://github.com/mongodb/specifications/commit/3586db8b1ed90d835737f6226bb3ff9012cfc5de for [DRIVERS-1603](https://jira.mongodb.org/browse/DRIVERS-1603); however, this would save us from having to modify retryable writes.

source/serverless-testing/README.rst
```diff
@@ -0,0 +1,115 @@
...
+
+Existing Spec Tests
+===================
+
+Tests defined in the following specifications MUST be included in a driver's
+Atlas Serverless testing suite, including prose tests:
+
+- CRUD, including the v1 and unified tests
+- Retryable Reads
+- Retryable Writes
```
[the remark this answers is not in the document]
kevinAlbs: Ah that was my mistake in the PR description. I clarified the description to note:

> Tests relying on failpoints cannot run against a load balancer fronting multiple servers.

Failpoints are handled within mongos or Atlas Proxy. If a failpoint is configured on a load balancer fronting multiple servers, the next operation may go to another server and not trigger the intended failpoint.

source/serverless-testing/README.rst
```diff
@@ -0,0 +1,125 @@
...
+The URIs can be read by drivers via the ``expansions.update`` Evergreen command.
+This will store the URIs into the ``SINGLE_ATLASPROXY_SERVERLESS_URI`` and
+``MULTI_ATLASPROXY_SERVERLESS_URI`` Evergreen expansions.
+
+.. _serverless directory in the drivers-evergreen-tools repository: https://github.com/mongodb-labs/drivers-evergreen-tools/tree/1ca6209825b6ed07ce90e24cda659143443709c8/.evergreen/serverless
+
+All tests MUST be run with wire protocol compression and authentication
+enabled.
+
+Test Runner Configuration
```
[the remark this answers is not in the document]
kevinAlbs: I noted that these URIs to satisfy load balancer URIs in test runners and that they are analogous to `SINGLE_LB_MONGOS_URI` and `MULTI_LB_MONGOS_URI`.

I considered mentioning these variables in the Retryable Reads/Writes or Unified Test Format instructions, but I chose not to. Those instructions do not mention `SINGLE_LB_MONGOS_URI` or `MULTI_LB_MONGOS_URI` specifically, so I do not think those test runners need to treat these any differently.

I considered reusing the variable names `SINGLE_LB_MONGOS_URI` and `MULTI_LB_MONGOS_URI`. Since mongos is not part of serverless, I chose not to.

patrickfreed (approved):
LGTM!

source/retryable-writes/tests/README.rst
```diff
@@ -169,6 +169,18 @@ Each YAML file has the following keys:
     the default is all topologies (i.e. ``["single", "replicaset", "sharded",
     "load-balanced"]``).
 
+  - ``serverless``: Optional string. Whether or not the test should be run on
```
jmikola, line 172: Just to confirm, this is being added with the assumption that it _may_ be needed down the line to exclude certain tests? As-is, it seems that all retryable writes tests are expected to pass on serverless.

kevinAlbs: Correct, no retryable writes tests currently need this. I have added the `serverless` field since these are run as part of serverless.

jmikola (approved):

source/load-balancers/tests/sdam-error-handling.yml
```diff
@@ -65,6 +65,10 @@ initialData:
 
 tests:
   - description: only connections for a specific serviceId are closed when pools are cleared
+    runOnRequirements:
+    # This test assumes that two sequential connections receive different serviceIDs.
+    # Sequential connections to a serverless instance may receive the same serviceID.
+    - serverless: forbid
```
kevinAlbs, line 71: For posterity, this test had inconsistently failed on the Go driver POC.

This test assumes that two sequential connections get different service IDs.

Serverless currently has multiple load balancers for a given instance. A client may not connect to the same load balancer on each connection. Therefore, we cannot guarantee that two sequential connections get different service IDs.

See [this slack conversation](https://mongodb.slack.com/archives/G01304NFS3C/p1630621416055700).

kevinAlbs (commented):
I add `serverless: forbid` to the test `only connections for a specific serviceId are closed when pools are cleared` after observing spurious failures in the Go POC. After checking with the Cloud team, the behavior is expected.

Proceeding with merging. Thank you for the reviews!

PR #1054, opened 2021-08-24
DRIVERS-1554: Clarify how internal client interacts with mongoses

jmikola:
https://jira.mongodb.org/browse/DRIVERS-1554

### DIFF
--- a/source/unified-test-format/unified-test-format.rst
+++ b/source/unified-test-format/unified-test-format.rst
@@ -3,13 +3,13 @@ Unified Test Format
 ===================
 
 :Spec Title: Unified Test Format
-:Spec Version: 1.5.4
+:Spec Version: 1.5.6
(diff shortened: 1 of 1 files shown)


source/unified-test-format/unified-test-format.rst
```diff
@@ -3,13 +3,13 @@ Unified Test Format
 ===================
 
 :Spec Title: Unified Test Format
-:Spec Version: 1.5.4
+:Spec Version: 1.5.6
```
jmikola, line 6: This assumes that #1052 will bump to 1.5.5 before this is merged. I'm opting to bump the patch version for the same reason cited in https://github.com/mongodb/specifications/pull/1052#discussion_r694333526.

This spec change is also mostly a clarification and may not warrant changes in some drivers.

PR #1055, opened 2021-08-24
DRIVERS-1896 Clarify server selection process for change stream resume

benjirewis:
DRIVERS-1896

Clarifies that server selection during change stream resumption should not only respect the original `aggregate`'s read preference but also follow the regular rules of server selection.

It seems that this detail was implied, but we may as well make it explicit and verify that all drivers have the same behavior.

### DIFF
--- a/source/change-streams/change-streams.rst
+++ b/source/change-streams/change-streams.rst
@@ -335,8 +335,8 @@ Driver API
     /**
      * Allowed values: ‘updateLookup’.  When set to ‘updateLookup’, the change notification
      * for partial updates will include both a delta describing the changes to the document,
-     * as well as a copy of the entire document that was changed from some time after the 
-     * change occurred. The default is to not send a value. 
+     * as well as a copy of the entire document that was changed from some time after the
+     * change occurred. The default is to not send a value.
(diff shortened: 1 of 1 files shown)


benjirewis:
Server selection during change stream _creation_ (the initial `aggregate` executed in `Watch`) should also follow the regular rules of server selection. I did not see a place in the spec to specify this, but let me know if you think I should add it somewhere.

prashantmital (approved):

prashantmital:
Let's open (or repurpose this) drivers validation ticket so drivers can manually verify that their change streams implementations are resilient against changes to the replica set. I don't think we have the capability in the Unified Test Format to add this as a spec test so a manual spot check will have to do at this point.

benjirewis:
As discussed offline, we'll use DRIVERS-1896 to have individual languages verify that their implementation of change stream server selection is compliant with the spec (in particular, that no driver has interpreted the spec to imply ignoring `localThresholdMS` for change stream server selection). DRIVERS-1805 may eventually track the work of drivers manually verifying that their change streams implementations are resilient against changes to replica sets.

PR #1060, opened 2021-08-30
DRIVERS-1554 Clarify internal per-mongos clients may be used for targetedFailPoint

kmahar:
I left spec version as-is and didn't bother changing the last updated date or change log since this is part of the same drivers ticket/clarification work you just did, but let me know if you think I should do so.

### DIFF
--- a/source/unified-test-format/unified-test-format.rst
+++ b/source/unified-test-format/unified-test-format.rst
@@ -3286,7 +3286,8 @@ Change Log
 :2021-08-30: Test runners may create an internal MongoClient for each mongos.
              Better clarify how internal MongoClients may be used.
              Clarify that drivers creating an internal MongoClient for each
-             mongos should use clients for ``targetedFailPoint`` operations.
+             mongos should use those clients for ``targetedFailPoint``
+             operations.
 
 :2021-08-23: Allow ``runOnRequirement`` conditions to be evaluated in any order.
 
--- a/source/unified-test-format/unified-test-format.rst
+++ b/source/unified-test-format/unified-test-format.rst
@@ -1893,7 +1893,10 @@ test runner SHOULD use the client entity associated with the session
 to execute the ``configureFailPoint`` command. In this case, the test runner
 MUST also ensure that this command is excluded from the list of observed
 command monitoring events for this client (if applicable). If such an API is
-not available, test runners MUST create a new MongoClient that is directly
+not available, but the test runner creates an internal MongoClient for each
+mongos, the test runner MAY use the internal MongoClient corresponding to
+the session's pinned server for this operation.
+Otherwise, test runners MUST create a new MongoClient that is directly
 connected to the session's pinned server for this operation. The new
 MongoClient instance MUST be closed once the command has finished executing.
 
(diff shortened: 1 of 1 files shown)


source/unified-test-format/unified-test-format.rst
```diff
@@ -1893,7 +1893,10 @@ test runner SHOULD use the client entity associated with the session
 to execute the ``configureFailPoint`` command. In this case, the test runner
 MUST also ensure that this command is excluded from the list of observed
 command monitoring events for this client (if applicable). If such an API is
-not available, test runners MUST create a new MongoClient that is directly
+not available, but the test runner creates an internal MongoClient for each
+mongos, the test runner MAY use the internal MongoClient corresponding to
```
kmahar, line 1897: I wasn't sure whether this should be a SHOULD or a MAY. It could be a SHOULD just for the sake of lessening the number of created clients, but it doesn't make a functional difference I don't think. Any opinions?

jmikola: The `failPoint` operation states:

> (fail points are not configured using an internal MongoClient)

I don't recall the reason for that, since `configureFailPoint` commands are already ignored by default in `observeEvents`. I presume the original text here suggested creating a new MongoClient only because we never clarified how the internal MongoClient(s) dealt with multiple mongos nodes.

In that case, I think it makes sense to use SHOULD here. If drivers have an API to target specific servers, they probably won't have any reason to create one internal MongoClient for mongos in the first place.

jmikola (commented):
Keeping the same version number seems fine, but I would bump the changelog entry and amend it to mention that targetedFailPoint was also revised.

source/unified-test-format/unified-test-format.rst
kmahar: I changed this to a SHOULD. 

> I presume the original text here suggested creating a new MongoClient only because we never clarified how the internal MongoClient(s) dealt with multiple mongos nodes.

given this, is it worth keeping the "Otherwise, test runners MUST..." segment below?

kmahar (commented):
👍 updated changelog entry and revised the date

source/unified-test-format/unified-test-format.rst
jmikola: No harm in leaving it there, but it also seems like it won't really be relevant so I have no objections to deleting it. I defer to you since the Swift implementation is one of the target audiences of this section :)

source/unified-test-format/unified-test-format.rst
```diff
@@ -3280,8 +3283,10 @@ Change Log
 :2021-08-30: Add ``hasServerConnectionId`` field to ``commandStartedEvent``,
              ``commandSuccededEvent`` and ``commandFailedEvent``.
 
-:2021-08-24: Test runners may create an internal MongoClient for each mongos.
+:2021-08-30: Test runners may create an internal MongoClient for each mongos.
```
jmikola, line 3286: Feel free to leave this as "08-30" as merging so we don't have to deal with it appearing _after_ @benjirewis's entry that bumped to 1.6.0.

jmikola (approved):

PR #1061, opened 2021-08-30
DRIVERS-1141 Update command monitoring and UTF spec versions

benjirewis:
DRIVERS-1141

Updates spec versions, last modified dates and change logs for the command monitoring and unified test format specs. Forgot to do this as part of #1050.

### DIFF
--- a/source/unified-test-format/unified-test-format.rst
+++ b/source/unified-test-format/unified-test-format.rst
@@ -3,7 +3,7 @@ Unified Test Format
 ===================
 
 :Spec Title: Unified Test Format
-:Spec Version: 1.5.7
+:Spec Version: 1.6.0
 :Author: Jeremy Mikola
 :Advisors: Prashant Mital, Isabel Atkinson, Thomas Reggi
 :Status: Accepted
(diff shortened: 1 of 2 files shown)


source/unified-test-format/unified-test-format.rst
```diff
@@ -3,13 +3,13 @@ Unified Test Format
 ===================
 
 :Spec Title: Unified Test Format
-:Spec Version: 1.5.6
+:Spec Version: 1.5.7
```
kmahar, line 6: this seems like it could be minor too, but I am not totally sure on how spec versions here change vs  schema versions so I will defer to @jmikola

kmahar (approved):
lgtm mod whatever Jeremy says on the unified format versioning

jmikola (approved):

source/unified-test-format/unified-test-format.rst
jmikola: This should definitely have been 1.6.0, since new schema changes were introduced.

benjirewis: Yeah changed it to `1.6.0` with the "Use minor versions" commit.

PR #1062, opened 2021-09-02
DRIVERS-823: Support $out and $merge on secondaries

jmikola:
https://jira.mongodb.org/browse/DRIVERS-823
https://jira.mongodb.org/browse/DRIVERS-1909

POC: https://github.com/mongodb/mongo-php-library/pull/861


source/crud/crud.rst
```diff
@@ -1927,8 +1928,63 @@ The ``update`` family of operations require that the update document parameter M
...
+  available to inherit (e.g. from the Collection), drivers MUST attempt to use
+  it. If that would result in a pre-5.0, secondary server being selected,
+  drivers MUST instead select a server using a primary read preference.
+
+Since the server version may not be known prior to server selection, this
+specification is not prescriptive about how drivers should select a server for
+an ``aggregate`` command including a write stage. For example, drivers MAY use a
+custom server selector to consider server/wire version when matching a read
+preference and fall back to selecting a primary. Alternatively, drivers MAY
+invoke server selection an additional time if necessary.
```
jmikola, line 1975: Server selection currently does not wire version into account at all. This is a complex issue that really wasn't _completely_ addressed in the driver scope. The scope states:

> Rather than requiring drivers to check out a connection from the selected server’s pool to determine its wire version, we will instead use the wire versions stored on the ServerDescription. This represents the information retrieved from the last successful SDAM check. We recognize that doing this is generally discouraged to avoid upgrade/downgrade race conditions, but actually encountering this race for this feature would require a double-downgrade of the entire cluster (5.0 to 4.2), which should be exceedingly rare. By doing this, drivers can support this feature by going through server selection once.

This citation is discussing connection check-out (assuming multi-threaded/async SDAM), but I'm more concerned with server selection at the moment. Server selection is what returns a ServerDescription, which we'd need to check to determine wire version. This made me concerned that drivers might need to invoke server selection _twice_ for this feature. The first attempt would respect a user-specified read preference and if that returns a pre-5.0 secondary we'd need to make a second selection to obtain the primary.

I asked @divjotarora about this and he clarified his original intentions while writing the scope:

> @divjotarora: IIRC, the hope was to use a "custom" server selector that is slightly more complicated than the read preference selection algorithm. I don't remember the specifics, but instead of just considering a server to be a candidate if it matches the read preference, it could be something like:
> ```
> isCandidate := (server matches readpref && server.maxWireVersion >= 13) || (server is primary)
> ```
> In the case of this being the first $out operation, you run the small risk of there being a 5.0+ secondary that hasn't been discovered yet. You could maybe handle this by removing the primary from the list of candidates if it's the only candidate and is 5.0+. You'd end up with an empty list of candidates and loop through server selection again. But regardless, I think it's possible to do the whole thing in a single 30-second selection loop rather than calling `topology.selectServer()` twice.

I don't follow the point in the second paragraph here (citing from our Slack conversation), but I think the paragraph I'm adding to the spec here is open-ended enough to allow for the pseudo-code he proposed for custom server selection.

In PHPLIB, customizing server selection in this manner isn't really an option so I think we'd need to conditionally invoke server selection a second time. We technically _could_ support it, but it would entail making non-trivial API changes to libmongoc and PHPC, which I don't think are worth the trouble in this case.

source/crud/crud.rst
```diff
@@ -1927,8 +1928,63 @@ The ``update`` family of operations require that the update document parameter M
...
+
+Drivers MUST follow the rules for
+`Passing read preference to mongos and load balancers <../server-selection/server-selection.rst#passing-read-preference-to-mongos-and-load-balancers>`_.
+If an explicit or inherited read preference was used and the selected server is
+a pre-5.0 mongos or a load balancer backing a pre-5.0 mongos, drivers MUST
+behave as if a primary read preference was used. This is especially important if
+drivers utilize a custom server selector, since they may need to infer whether a
+primary read preference was used as a fall back.
```
jmikola, line 1983: The goal of this paragraph is just the remind drivers that the effective read preference may need to be passed along to mongos or a load balancer. If the user-specified read preference ultimately wasn't used (e.g. it selected a pre-5.0 server), we should assume that a primary read preference is needed as a fall back. The [upgrade process for sharded clusters](https://docs.mongodb.com/manual/tutorial/upgrade-revision/#upgrade-sharded-clusters) sees mongos processes upgraded _after_ shards. Given that, I think the only safe assumption is that if mongos is 5.0+, then we can assume the servers behind it are also 5.0+ (but not vice versa).

I realize LBs were introduced with MongoDB 5.0, but I saw no harm in being pedantic about a "load balancer backing a pre-5.0 mongos".

source/server-selection/server-selection.rst
```diff
@@ -730,33 +730,35 @@ the command and how it is invoked:
...
+      - collStats
+      - count
+      - dbStats
+      - distinct
+      - find
+      - geoNear
+      - geoSearch
+      - group
+      - mapReduce where the ``out`` option is ``{ inline: 1 }``
+      - parallelCollectionScan
```
jmikola, line 743: Just fixing indentation here. Also fixed some grammar and capitalization (in the mapReduce paragraph) below.

source/server-selection/server-selection.rst
```diff
@@ -730,33 +730,35 @@ the command and how it is invoked:
...
-      secondary.
+      database, or collection configuration.
+
+      For pre-5.0 servers, an aggregate command is "must-use-primary" if its
+      pipeline contains a write stage (e.g. ``$out``, ``$merge``); otherwise, it
+      is "may-use-secondary". For 5.0+ servers, secondaries can execute an
+      aggregate command with a write stage and all aggregate commands are
+      "may-use-secondary". This is discussed in more detail in
+      `Read preferences and server selection <../crud/crud.rst#read-preferences-and-server-selection>`_
+      in the CRUD spec.
```
jmikola, line 755: For the record, I haven't found these terms (e.g. "must-use-primary") used in other specs, despite the text below:

> Command helper specifications SHOULD use those terms for clarity.

Because this feature may require some deviation to the normal rules for server selection, I also opted not to introduce these terms in my CRUD spec additions.

I'm not even sure this paragraph makes sense in the scope of the server selection spec, because it's referring to server versions as an influencing factor of the command helper's type (e.g. "must-use-primary", "may-use-secondary"). Rather than try to explain that here, I thought it was easier to consolidate everything in the CRUD spec.

If this issue comes up down the line with _another_ helper, I suppose we can revisit and restructure the server selection spec. For now, I think we can avoid doing so in the spirit of "defy augury".

source/crud/crud.rst
p-mongo: I am having a bit of trouble following this paragraph (the proposed change in the diff). Is this referring to the case of $merge/$out being redirected to primary on pre-5.0 and, subsequently, pointing out that since the query is being sent to primary the read preference that is sent along with the query must therefore also be primary?

p-mongo (commented):
The current content LGTM unless I misunderstood the one section I commented on.

I looked through our (Ruby's) present server selection implementation and it seems to me that it would be relatively straightforward to add a condition along the lines of "if wire version is less than X use primary". However, our present implementation performs server selection twice if $out/$merge ends up being directed to  a primary (hence could be subject to twice the timeout). I believe it would be  more work to reduce this to  a single round since we would need to add a mechanism for returning the primary read preference back from the server selector so that we could send the correct one to mongos/load balancers as mentioned.

source/crud/crud.rst
```diff
@@ -1927,8 +1928,63 @@ The ``update`` family of operations require that the update document parameter M
...
+
+
+Returning a cursor on the output collection
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+As noted in the documentation for the ``aggregate`` helper earlier in this
+document, ``$out`` and ``$merge`` are special pipeline stages that cause no
+results to be returned from the server. As such, drivers MAY setup a cursor to
+be executed upon iteration against the output collection and return that instead
+of an iterable that would otherwise have no results.
```
alcaeus, line 1946: For `$out` I see no problems, but I'm not sure if we should assume that the output of `$merge` is the entire list of documents in the output collection. No objection to keeping this in, but maybe it's worth pointing this out in the documentation for drivers that do return a cursor from the output collection.

source/crud/crud.rst
jmikola: > Is this referring to the case of $merge/$out being redirected to primary on pre-5.0 and, subsequently, pointing out that since the query is being sent to primary the read preference that is sent along with the query must therefore also be primary?

Going back to the scope for this feature, we are cautiously assuming that a read preference can only be used for `$out`/`$merge` if the server itself is 5.0+. We aren't differentiating between the kind of read preference being used at all. If the driver ultimately selects a pre-5.0 mongos (either directly or through the LB), we will not pass a read preference at all. Erring on the side of caution, we're going to assume that a pre-5.0 mongos is fronting pre-5.0 mongod servers.

But if a 5.0+ mongos is selected, we can assume it's safe to pass along _any_ read preference. In this case, that will either be what was explicitly provided to the operation or inherited through some driver object (e.g. collection). That correlates with the two bullets in "Drivers MUST guarantee the following" a few paragraphs up.

This approach should not require any special knowledge of what happens _within_ server selection, be it a single attempt with custom logic or two separate attempts. To keep things simple, we need only consider the wire version of the server that was ultimately selected.

I suppose for drivers relying on two separate selection attempts, they can be more efficient. For example, if `mode=secondary` returns a pre-5.0 mongos, a driver could still use that mongos and just omit passing the read preference onwards. A second selection attempt with `mode=primary` would be a waste of time, since we'd likely just get another pre-5.0 mongos back (ignoring the possibility that there are mixed mongos versions available).

----

Going back to what you wrote in your other top-level PR comment:

> I looked through our (Ruby's) present server selection implementation and it seems to me that it would be relatively straightforward to add a condition along the lines of "if wire version is less than X use primary". However, our present implementation performs server selection twice if $out/$merge ends up being directed to a primary (hence could be subject to twice the timeout). I believe it would be more work to reduce this to a single round since we would need to add a mechanism for returning the primary read preference back from the server selector so that we could send the correct one to mongos/load balancers as mentioned.

I think I might have already answered this above, but there should be no need to change the server selector to return a read preference in addition to the selected server. If the first attempt comes back with a 5.0+ mongos, that's great and you can pass along the original read preference you provided the server selector (if any). If a pre-5.0 mongos is selected, you can assume no read preference should be pass on. In either case, I don't think you'd need to make a second selection attempt.

A second selection should only come up when dealing with a replica set topology where the first selection attempt returns a pre-5.0 _secondary_ server. If the user is trying to use `mode=nearest` and that happens to select a pre-5.0 primary, there's no reason to repeat server selection with `mode=primary` since we'd ultimately get the same result. That said, if `mode=nearest` happened to select a pre-5.0 secondary, we would need to repeat server selection to get a primary (or use a single attempt with custom logic).

source/crud/crud.rst
jyemin: Both.   For both out and merge, the cursor returns all documents in the collection.

jmikola: Added a paragraph below to suggest docs for this. @alcaeus please resolve at your leisure.

source/server-selection/server-selection.rst
```diff
@@ -1001,10 +1003,12 @@ If ``mode`` is 'primaryPreferred', select the primary if it is known, otherwise
...
-For all read preferences modes except 'primary', clients MUST set the ``SecondaryOk``
-wire protocol flag to ensure that any suitable server can handle the request.  Clients
-MUST NOT set the ``SecondaryOk`` wire protocol flag if the read preference mode is
-'primary'.
+For all read preferences modes except 'primary', clients MUST set the
+``SecondaryOk`` wire protocol flag (OP_QUERY) or ``$readPreference`` global
+command argument (OP_MSG) to ensure that any suitable server can handle the
+request. If the read preference mode is 'primary', clients MUST NOT set the
+``SecondaryOk`` wire protocol flag (OP_QUERY) or ``$readPreference`` global
+command argument (OP_MSG).
```
jmikola, line 1011: While working on this PR I realized the server selection spec is _very_ light on documentation for using the `$readPreference` global command argument. This change pertains to [DRIVERS-1918](https://jira.mongodb.org/browse/DRIVERS-1918). I opted not to revise [Passing read preference to mongos and load balancers](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#passing-read-preference-to-mongos-and-load-balancers) in this PR, as that would be a much more intensive change.

source/crud/tests/unified/aggregate-merge-readPreference.yml
```diff
@@ -0,0 +1,106 @@
+description: aggregate-merge-readPreference
+
+schemaVersion: '1.3'
+
+runOnRequirements:
+  # 3.6+ non-standalone is needed to utilize $readPreference in OP_MSG. 4.2+ is
+  # required for $merge.
```
jmikola, line 7: Requiring 4.2+ here is the only notable difference (apart from pipeline differences) between this file and `aggregate-out-readPreference.yml`.

source/crud/tests/unified/aggregate-merge-readPreference.yml
```diff
@@ -0,0 +1,106 @@
...
+
+createEntities:
+  - client:
+      id: &client0 client0
+      observeEvents:
+        - commandStartedEvent
+      # Used to test that read and write concerns are still inherited
+      uriOptions:
+        readConcernLevel: *readConcernLevel
+        w: *writeConcernW
```
jmikola, line 25: This was a late addition. While POC-ing this in PHPLIB, which uses libmongoc's command execution API, I realized that we couldn't use the existing read/write-aware command methods. Normally, an aggregation with `$out`/`$merge` would utilize [`mongoc_client_read_write_command_with_opts`](http://mongoc.org/libmongoc/current/mongoc_client_read_write_command_with_opts.html). That method ensures that any higher-level read/write concern is inherited, but it does not support a `readPreference` option ("primary" is assumed).

This feature breaks convention by allowing a write command to be sent to a secondary (with a read preference to boot). Addressing this in PHPLIB meant that I had to fall back to using the generic command runner ([`mongoc_client_command_with_opts`](http://mongoc.org/libmongoc/current/mongoc_client_command_with_opts.html)), which _does_ support all options (RC/RP/WC) but _does not_ handle inheritance. That ultimately wasn't a problem for PHPLIB since we also handle inheritance ourselves, but I imagine this will be more annoying for libmongoc.

In any event, covering RC/WC inheritance this this test should help libmongoc catch this during implementation and avoid a potential bug.

source/crud/crud.rst
p-mongo: I reviewed our code again and it should only redirect in replica set topology (standalones and mongoses are always accepted on the first try). So I think we are in alignment here, I will resolve this thread.

p-mongo (approved):

source/crud/crud.rst
```diff
@@ -1927,8 +1928,67 @@ The ``update`` family of operations require that the update document parameter M
...
+- If no explicit read preference is specified but a default read preference is
+  available to inherit (e.g. from the Collection), drivers MUST attempt to use
+  it. If that would result in a pre-5.0, secondary server being selected,
+  drivers MUST instead select a server using a primary read preference.
+
+Since the server version may not be known prior to server selection, this
+specification is not prescriptive about how drivers should select a server for
+an ``aggregate`` command including a write stage. For example, drivers MAY use a
+custom server selector to consider server/wire version when matching a read
+preference and fall back to selecting a primary. Alternatively, drivers MAY
```
jyemin, line 1978: Does this MAY imply that it's ok for a driver to effectively double the server selection timeout, or do you have to subtract the time for the first server selection before falling back to the second?  If so, the spec should indicate that explicitly.  Additionally, the spec should indicate what should happen if the initial server selection times out.  I imagine at that point the driver has to error, even if the second server selection would have succeeded (assuming we would have fallen back had the first selection succeeded).

jyemin (changes requested):
Looking good!  Just one question about the double server selection option.

source/crud/crud.rst
jmikola: I have no qualms about duplicating the server selection timeout in the interest of minimizing driver complexity here. If that does prove to be a concern, I expect CSOT would also address it in due time; however, I'm not expecting this to be a problem given the small audience for this feature. AFAIK, the only current demand for this feature is for internal users of the Go driver with Charts.

I do think that stopping if the first server selection attempt times out is reasonable, although it is a notable difference from the single attempt approach (with a custom selector). For drivers that do need to make two attempts, catching and ignoring the first attempt's error would also introduce undesirable complexity.

So, proposed changes would be:

 * Explicitly note that the two-attempt approach may result in the total selection time exceeding `serverSelectionTimeoutMS`
 * Require drivers using the two-attempt approach to forgo the second attempt if the first resulted in a server selection timeout. It's _possible_ we could find a primary on the second attempt but another timeout may be more likely (especially for a permissive read preference like `secondaryPreferred`).

Does anyone else have thoughts on the matter? If not, I'll follow up on Monday with spec changes.

jyemin: >  however, I'm not expecting this to be a problem given the small audience for this feature

I'm not so sure about that.  Given the way we're exposing this feature, we're effectively opting applications into it.  Any application with a non-default read preference defined for client/database/collection will hit this code path, even if they are still on 4.4-and-below server and always end up with the fallback-to-primary scenario.

jmikola: > Given the way we're exposing this feature, we're effectively opting applications into it.

I agree. "small audience" was a poor choice of words on my part. The feature itself is frustrating to me, as it seemingly violates so many existing rules we have around server selection.

I wasn't around for scoping, but I would have much preferred to never have `$out`/`$merge` inherit a read preference and only allow drivers to start accepting a per-operation read preference for them. But that would leave drivers w/o per-operation read preference options no way to support this at all.

> Any application with a non-default read preference defined for client/database/collection will hit this code path, even if they are still on 4.4-and-below server and always end up with the fallback-to-primary scenario.

I don't think there's _any_ performance regression for drivers implementing a single attempt with custom selection logic (i.e. considering read preference and wire version at once). For drivers that decide to make a second attempt as needed, let's enumerate the possible code paths by topology type:

 * Single topology: it's only possible to select a single server, so a second attempt never makes sense. It's possible that the application has a direct connection to a secondary (or some non-writable server). Perhaps the spec should clarify that no second attempt should ever be made for a single topology.
 * Replica set topology: a second attempt only happens if a pre-5.0 secondary is selected. If a primary (any version) or 5.0+ secondary is selected, that server is always usable and no second attempt is needed.
 * Sharded or load-balanced topology: the first and only selection attempt is going to return a server capable of receiving a write. Based on the wire version of mongos (either the direct node or the backing mongos), the driver will decide to forward the original read preference or substitute `mode=primary`.

If my understanding is correct, I think this only affects a subset of applications that (a) use a read preference that can potentially select a secondary _and_ (b) have pre-5.0 secondaries in the cluster _and_ (c) performance aggregations with `$out`/`$merge`.

jmikola: To recap, proposed changes to address the two-attempt approach:

 * Drivers MUST NOT make a second attempt if the first attempt encounters a server selection timeout.
 * Drivers MUST NOT make second attempts for single, sharded, or load-balanced topologies (for reasons cited above, there is no reason to do so).
 * Explicitly discuss how the two-attempt approach can result in a performance regression (as discussed in the end of [last comment](https://github.com/mongodb/specifications/pull/1062#discussion_r706937170)).

jyemin: Is it an option to make implementation with a single server selection attempt a MUST?

jmikola: I would only go as far as SHOULD. Making it a requirement would require creating an entirely new server selection API in libmongoc just to support the single-attempt approach in PHPLIB. I'm not convinced that this feature warrants the engineering effort.

jyemin: @kevinAlbs what do you think?

source/crud/crud.rst
jyemin: I think the point in the second paragraph is just that a driver might still make the "wrong" choice,  If the only servers that have been discovered so far are 4.4 servers, the driver will fallback to primary, even if there are indeed 5.0 servers in the topology

source/crud/crud.rst
alcaeus: It would be nice to have a level between SHOULD and MUST - ideally, we'd require drivers to not do two rounds of server selection as long as the technical complexity that's involved doesn't justify it.

For PHPLIB specifically, I don't think the complexity of exposing custom server selection in libmongoc and PHPC is worth it. I don't believe that a lot of users would run into the double server selection in PHPLIB, but we can always use PHP's error reporting mechanism (`trigger_error`) to warn users of a potential performance instance that they can then handle as they see fit.

Of course, we can always use MUST and allow deviations from the spec, so basically "drivers MUST NOT perform two rounds of server selection, but MAY choose to do so if the complexity involved doesn't justify it" (probably the worst wording, but I hope it gets the point across so others can improve it)

jmikola: > It would be nice to have a level between SHOULD and MUST - ideally, we'd require drivers to not do two rounds of server selection as long as the technical complexity that's involved doesn't justify it.

I do think that SHOULD fits the bill for this. It _is_ actually quite strict on its own, as drivers are not expected to deviate from a SHOULD without good reason (technical complexity in this case). To that end, I'd consider something like:

> drivers SHOULD use the single-attempt approach with custom server selection logic, but MAY use two attempts if...

benjirewis: SHOULD with a following MAY seems accurate to me; I also agree that overhauling server selection in libmongoc is not worth it for this feature. Go should be able to implement the single-attempt approach with a custom server selector.

benjirewis (commented):
Changes look good to me for now. Still implementing in Go, so I'll come back for a second round of review once [GODRIVER-1868](https://jira.mongodb.org/browse/GODRIVER-1868) is in review.

source/crud/crud.rst
jmikola: Updated these paragraphs:

 * Drivers SHOULD use the single attempt approach with custom logic
 * Drivers MAY use the two attempt approach if necessary

For the two attempt approach:

 * Drivers MUST NOT make a second attempt if the first results in a timeout error. Instead, they raise the original error.
 * Drivers MUST NOT make a second attempt for Single, Sharded, or LoadBalanced topologies. There's no reason to do so, as the first server would either be writable (most cases) or the only option (i.e. Single topology with a direct connection to a pre-5.0 secondary).

kevinAlbs: > Making it a requirement would require creating an entirely new server selection API in libmongoc just to support the single-attempt approach in PHPLIB.

I agree with that assessment. I considered whether it would be sufficient to add the [custom server selector client level option from the Server Selection specification](https://github.com/mongodb/specifications/blob/c48d7f416e51dce8fd4c8d05bdb05790508ca3bb/source/server-selection/server-selection.rst#serverselector). But the custom callback would not know if the current operation is an agg with `$out`/`$merge`. So the API addition would be specific to this feature.

I am not opposed new API to libmongoc if it is needed for correctness. But if the caveat is doubling server selection timeout in rare scenarios, my vote goes for less implementation complexity. CSOT will likely require a deadline in server selection, so we could remove this caveat later.

jmikola:
Noted the LGTM from @jyemin. I'll wait until @benjirewis comes back after progressing on the Go POC.

@alcaeus already signed off on the PHPLIB POC (https://github.com/mongodb/mongo-php-library/pull/861), so I don't think there's anything to be done there. I left a [comment](https://github.com/mongodb/mongo-php-library/pull/861#discussion_r711224550) a few days ago explaining why I plan to keep a second server selection attempt in place instead of immediately implementing a workaround to select a fall back primary outside of server selection.

source/crud/crud.rst
```diff
@@ -1927,8 +1928,72 @@ The ``update`` family of operations require that the update document parameter M
...
+``featureCompatibilityVersion``, executing an aggregation pipeline with ``$out``
+or ``$merge`` on a secondary will require a 5.0+ server (i.e. wire version >=
+13).
+
+Drivers MUST guarantee the following:
+
+- If an explicit (i.e. per-operation) read preference is specified for an
+  aggregation with a write stage, drivers MUST attempt to use it. If that would
+  result in a pre-5.0, secondary server being selected, drivers MUST instead
+  select a server using a primary read preference.
```
benjirewis, line 1968: Not sure exactly how server selection works in other drivers, but we use filtering functions in the Go driver: lists of candidate servers are narrowed down based on certain criteria (latency, read preference, etc.). It's not exactly clear _which_ server will be selected; we just remove servers from the candidates list if they do not fit the criteria. So, I interpreted this (and similar statements below) as: `remove all pre-5.0 secondary servers from the candidates list. If no candidates remain, fall back to the primary.`

Still a couple questions here, though:
1. What is the expected behavior for an aggregate with an output stage with `secondary` read preference against a single-node replica set (i.e. only one node and it's a primary)? Is this a server selection error?
2. If the user specifies `secondary` read preference, but we fall back to the primary because there are no usable secondaries, what is the `$readPreference` sent to the server in the actual command? `secondary` or `primary`?

benjirewis: > [in this case] I would always expect a `secondary` RP to result in a server selection failure.

Same here, and I agree that "fixing" server selection in this case will only further disrupt our process of server selection.

> `$readPreference` in the OP_MSG command should always correspond to the _effective_ read preference that was used.

This sounds good to me, and I do think the spec should clarify that "falling back to primary" changes both the effective RP and the `$readPreference` in the command to be `primary`. Is it possible to test that case with a spec test?

source/crud/tests/unified/db-aggregate-write-readPreference.yml
```diff
@@ -0,0 +1,145 @@
+description: db-aggregate-write-readPreference
+
+schemaVersion: '1.4'
+
+runOnRequirements:
+  # 3.6+ non-standalone is needed to utilize $readPreference in OP_MSG.
+  # Serverless does not support $listLocalSessions and $currentOp stages, and
+  # mongos does not allow combining them with $out or $merge.
```
jmikola, line 8: @benjirewis: FYI, this change should address the error you saw:

> $listLocalSessions must run on mongoS, but $out must run on a shard

kevinAlbs (approved):

source/crud/crud.rst
```diff
@@ -1927,8 +1928,72 @@ The ``update`` family of operations require that the update document parameter M
...
+- If no explicit read preference is specified but a default read preference is
+  available to inherit (e.g. from the Collection), drivers MUST attempt to use
+  it. If that would result in a pre-5.0, secondary server being selected,
+  drivers MUST instead select a server using a primary read preference.
+
+Drivers SHOULD use a custom server selector to consider server/wire version when
+matching a read preference and, if a pre-5.0 secondary would be selected, fall
+back to selecting a primary. With this approach, only a single server selection
+attempt is needed.
```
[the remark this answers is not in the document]
jmikola: I think even a one-attempt solution could still end up with the primary, depending how the custom logic is implemented. If we consider @benjirewis's earlier comment about Go's procedure for reducing the list of eligible servers, then I would definitely expect the 5.0 secondary to always be selected; however, a "custom server selector" could just as easily do the random selection between both secondaries (and that wouldn't be any less _correct_), or end up having to rule out the 5.0 secondary due to latency.

source/crud/crud.rst
jmikola: > I do think the spec should clarify that "falling back to primary" changes both the effective RP and the `$readPreference` in the command to be primary. Is it possible to test that case with a spec test?

I can amend this language to describe the concept of an _effective_ read preference. As far as testing, I think the current tests that expect `$readPreference` to be absent is already covering the "fall back to primary" case. A `primary` read preference never gets forwarded to the server, so its absence in the command document is how we assert that the command is destined for a primary (either directly in a replica set topology or via mongos/LB).

source/crud/crud.rst
```diff
@@ -1927,8 +1928,83 @@ The ``update`` family of operations require that the update document parameter M
...
+- In the absence of an explicit or inherited read preference, the effective read
+  preference is ``{ "mode": "primary" }``.
+- If an explicit or inherited read preference results in selection of an
+  ineligible server and selection falls back to a primary, the effective read
+  preference is ``{ "mode": "primary" }``.
+- If a pre-5.0 mongos or a load balancer backing a pre-5.0 mongos is selected,
+  the effective read preference is ``{ "mode": "primary" }``.
+- If an explicit or inherited read preference is used and results in selection
+  of an eligible server (excluding a pre-5.0 mongos as discussed above), that is
+  the effective read preference.
```
jmikola, line 2003: @benjirewis: I rewrote the last paragraph and added this bulleted list to discern what the _effective_ read preference is for populating `$readPreference` in the command. Please let me know if this addresses your previous questions.

source/crud/crud.rst
```diff
@@ -1983,13 +1983,24 @@ MUST NOT make a second attempt and MUST propagate the original timeout error. If
...
-Drivers MUST follow the rules for
-`Passing read preference to mongos and load balancers <../server-selection/server-selection.rst#passing-read-preference-to-mongos-and-load-balancers>`_.
-If an explicit or inherited read preference was used and the selected server is
-a pre-5.0 mongos or a load balancer backing a pre-5.0 mongos, drivers MUST
-behave as if a primary read preference was used. This is especially important if
-drivers utilize a custom server selector, since they may need to infer whether a
-primary read preference was used as a fall back.
+Regardless of whether drivers employ a custom server selector or two-attempt
+approach, they MUST discern the *effective* read preference for the operation,
+which SHALL be be used for specifying the
```
kevinAlbs, line 1988:
```suggestion
which SHALL be used for specifying the
```

kevinAlbs (approved):
LGTM with typo fix.

source/crud/crud.rst
benjirewis: Looks great to me! Thanks for this comprehensive list 🧑‍🔧

benjirewis (approved):
LGTM! Go driver implementation is now merged ([GODRIVER-1868](https://jira.mongodb.org/browse/GODRIVER-1868)).

PR #1063, opened 2021-09-03
DRIVERS-521 Allow custom service names with rawsrv URI scheme

benjirewis:
DRIVERS-521

Introduces a new URI scheme ``mongodb+rawsrv`` to be used in initial DNS seedlist discovery and subsequent SRV polling. ``mongodb+rawsrv`` functions identically to ``mongodb+srv`` but does not automatically prepend ``_mongodb._tcp`` service name and protocol to the SRV lookup. Instead, it uses the service name and protocol provided by the user in the initial URI.

Updates initial DNS seedlist discovery and SRV polling specs to describe new scheme. Adds spec tests for valid and invalid ``mongodb+rawsrv`` URI schemes and a prose test for SRV polling with ``mongodb+rawsrv``.

Prototype in Go is available [here](https://github.com/benjirewis/mongo-go-driver/tree/rawServiceName). [Here](https://spruce.mongodb.com/version/613249103066157f45794be2/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) is a patch of that implementation passing all tests.

### DIFF
--- a/source/connection-string/tests/invalid-uris.yml
+++ b/source/connection-string/tests/invalid-uris.yml
@@ -258,4 +258,13 @@ tests:
         auth: ~
         options: ~
 
-    
+    -
+        description: "mongodb+rawsrv without service name"
+        uri: "mongodb+rawsrv://tcp.test7.test.mongodb.com"
+        valid: false
+        warning: ~
+        hosts: ~
+        auth: ~
+        options: ~
+
+
--- a/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
+++ b/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
@@ -10,8 +10,8 @@ Initial DNS Seedlist Discovery
 :Authors: Derick Rethans
 :Status: Draft
 :Type: Standards
-:Last Modified: 2019-04-15
-:Version: 1.4.0
+:Last Modified: 2021-09-01
+:Version: 1.5.0
 :Spec Lead: Matt Broadstone
 :Advisory Group: \A. Jesse Jiryu Davis
 :Approver(s): Bernie Hackett, David Golden, Jeff Yemin, Matt Broadstone, A. Jesse Jiryu Davis
@@ -32,8 +32,8 @@ domain to return a list of host names. Supporting this feature would assist
 our users by decreasing maintenance load, primarily by removing the need to
 maintain seed lists at an application level.
 
-This specification builds on the `Connection String`_ specification. It adds a
-new protocol scheme and modifies how the `Host Information`_ is interpreted.
+This specification builds on the `Connection String`_ specification. It adds
+new protocol schemes and modifies how the `Host Information`_ is interpreted.
 
 .. _`Connection String`: ../connection-string/connection-string-spec.rst
 .. _`Host Information`: ../connection-string/connection-string-spec.rst#host-information
@@ -48,14 +48,15 @@ interpreted as described in `RFC 2119 <https://www.ietf.org/rfc/rfc2119.txt>`_.
 Specification
 =============
 
-The connection string parser in the driver is extended with a new protocol
-``mongodb+srv`` as a logical pre-processing step before it considers the
-connection string and SDAM specifications. In this protocol, the comma
-separated list of host names is replaced with a single host name. The
-format is::
+The connection string parser in the driver is extended with two new protocols,
+``mongodb+srv`` and ``mongodb+rawsrv``, as logical pre-processing steps before
+considering the connection string and SDAM specifications.
+
+In both protocols, the comma separated list of host names is replaced with a
+single host name. The formats are::
 
     mongodb+srv://{hostname}.{domainname}/{options}
-
+    mongodb+rawsrv://_{servicename}._{protocol}.{hostname}.{domainname}/{options}
 
 ``{options}`` refers to the optional elements from the `Connection String`_
 specification following the ``Host Information``. This includes the ``Auth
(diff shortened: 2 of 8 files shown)


source/polling-srv-records-for-mongos-discovery/tests/README.rst
```diff
@@ -116,8 +116,8 @@ The following tests MUST setup a MongoClient using the
 ``test3.test.build.10gen.cc`` SRV record. Each test MUST mock the described
 situation and make the specified assertions.
 
-9. Test that SRV polling is not done for load balalanced clusters
```
benjirewis, line 119: Typo..

source/connection-string/tests/invalid-uris.yml
```diff
@@ -258,4 +258,13 @@ tests:
         auth: ~
         options: ~
 
-    
+    -
+        description: "mongodb+rawsrv without service name"
+        uri: "mongodb+rawsrv://tcp.test7.test.mongodb.com"
+        valid: false
```
jmikola, line 264: I'm admittedly not familiar with this, but if the point of `+rawsrv` is to prevent drivers from appending `_mongodb._tcp.` on their own, is it possible to have a valid test case where we manually add `_mongodb._tcp.` with `+rawsrv`?

Or is that specific to how DNS is configured and unrelated to the host name in the connection string here?

source/polling-srv-records-for-mongos-discovery/tests/README.rst
```diff
@@ -126,3 +126,14 @@ mock the addition of the following DNS record::
 
 Wait until ``2*rescanSRVIntervalMS`` and assert that the final topology description
 only contains one server: ``localhost.test.build.10gen.cc.`` at port ``27017``.
+
+10. Test that SRV polling is done correctly for rawsrv URIs
+-----------------------------------------------------------
+
+Connect to ``mongodb+rawsrv://_customname._tcp.test3.test.build.10gen.cc/``,
```
benjirewis, line 133: I'll have to speak with the BUILD team about setting up a custom SRV record with `customname` as a service name for this prose test to work. Will update with a ticket number on this thread.

source/connection-string/tests/invalid-uris.yml
benjirewis: I did add the `rawsrv-uri` test [here](https://github.com/mongodb/specifications/pull/1063/files#diff-1f7e89e066c8273e3140ac34a2e732f6ab7a2b4d5e267be662624cd936aa1a0cR1-R9) that uses `mongodb+rawsrv://_mongodb._tcp.test3.test.build.10gen.cc/` as a URI and ensures that hosts are correctly found through SRV. Does that count as a valid test case where `_mongodb._tcp.` is already included with `+rawsrv`?

jmikola: Oh, I missed that when I wrote the above comment. That's the kind of test I was expecting. Thanks!

source/polling-srv-records-for-mongos-discovery/polling-srv-records-for-mongos-discovery.rst
```diff
@@ -9,8 +9,8 @@ Polling SRV Records for mongos Discovery
 :Author: Derick Rethans
 :Status: Accepted
 :Type: Standards
-:Last Modified: 2018-11-29
-:Version: 1.0
+:Last Modified: 2021-09-03
+:Version: 1.1
```
benjirewis, line 13: Maybe this should just be a patch bump; LMK what you think.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
```diff
@@ -65,16 +66,22 @@ Seedlist Discovery
...
-``_mongodb._tcp.{hostname}.{domainname}``. This DNS query is expected to
-respond with one or more SRV records. From the DNS result, the driver now MUST
-behave the same as if an ``mongodb://`` URI was provided with all the host
+records on ``{hostname}.{domainname}``.
+
+If ``mongodb+srv`` is used, ``{hostname}.{domainname}`` MUST be prefixed with
+``_mongodb._tcp.``: ``_mongodb._tcp.{hostname}.{domainname}``. If ``mongodb+rawsrv``
+is used, the user has supplied their own service name and protocol in the URI, and
+``_mongodb._tcp`` SHALL NOT be prepended. If the ``mongodb+rawsrv`` scheme is used
+and no service name has been provided in the URI, the driver MUST raise a parse error.
```
jmikola, line 75: I read the old comments in [DRIVERS-521](https://jira.mongodb.org/browse/DRIVERS-521) (going back to 2018) but I don't think this question was answered. Is there any reason we needed to introduce a new scheme and couldn't control this via a driver option (akin to how we configure FLE and versioned API)? For instance, a driver option that toggles whether the driver prepends `_mongodb._tcp` and defaults to true for BC.

I only bring this up because there are a host of third-party libraries that track valid schemes for MongoDB connection strings. Those presently have `mongodb://` and `mongodb+srv://` hard-coded and I presume they would eventually need to be updated to accept `mongodb+rawsrv://` as well.

source/polling-srv-records-for-mongos-discovery/polling-srv-records-for-mongos-discovery.rst
jmikola: I think a minor version bump makes sense, since this is very much a new feature for drivers.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
benjirewis: A good question. Another option may be to add service name and protocol to URI options, either as new, named options or as key-value pairs in a field like `authMechanismProperties` (you can currently set `SERVICE_NAME` through `authMechanismProperties` for GSSAPI).

source/polling-srv-records-for-mongos-discovery/polling-srv-records-for-mongos-discovery.rst
```diff
@@ -78,10 +78,15 @@ discovery section of the original specification. The behaviour of the periodic
...
 
-- The driver will query the DNS server for SRV records on
+- If the initial topology was created through a ``mongodb+srv://`` URI,
+  the driver will query the DNS server for SRV records on
   ``{hostname}.{domainname}``, prefixed with ``_mongodb._tcp.``:
   ``_mongodb._tcp.{hostname}.{domainname}``.
 
+- If the initial topology was created through a ``mongodb+rawsrv://`` URI,
+  the driver will query the DNS server for SRV records using the original
+  URI without the scheme: ``_{servicename}._{protocol}.{hostname}.{domainname}``.
```
xdg, line 88: I find this phrasing ("without the scheme" ) and the example confusing.  Maybe "using the original URI without any additional prefix" and no example would be clearer.  (If we get rid of `mongodb+srvraw`, then the example above should be changed to not hard-code _mongodb but to use the SRV service name.)

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
xdg: I was originally thinking that we could let users specify anything for an SRV record, but the advantage of underscore prefixing is that it lets us tell where the "hostname.domainname" parts are.  On reflection, I would be supportive of keeping only `mongodb+srv` and adding a URI parameter to set the service name to something other than "_mongodb".  This is comparable to how we default the Kerberos service name to "mongodb" but let people set their own.

xdg (changes requested):
The more I think about it, the more I agree with Jeremy that keeping only `mongodb+srv` and making the service name a parameter (with a default of "_mongodb") is a better approach.  It's less change to specs, code, and validation rules.

benjirewis:
Let me investigate making service name a URI option. Moving this PR back to draft for now, so you can hold off on further review.

benjirewis:
See #1067 for an alternate implementation of custom service names as URI options. I'll still leave this PR as a draft but will close it if we opt for the `srvServiceName`-URI-option solution.

benjirewis:
Closing this PR in favor of #1067.

PR #1064, opened 2021-09-04
DRIVERS-1912 Clarify maxWireVersion check for maxStalenessSeconds

kevinAlbs:
**Summary**
There is currently a topology-wide maxWireVersion check for maxStalenessSeconds support:

> Clients are required to throw an error if maxStalenessSeconds is set, and any server in the topology has maxWireVersion less than 5.

Servers of type `Unknown` or `PossiblePrimary` should be excluded from this check. Neither server type has a known maxWireVersion.

**Motivation**

A related bug in libmongoc was reported [CDRIVER-3636](https://jira.mongodb.org/browse/CDRIVER-3636). libmongoc checked `PossiblePrimary` servers, which always had maxWireVersion=0.

**Testing**

Here is a POC of the new tests the C driver: https://github.com/mongodb/mongo-c-driver/pull/860


p-mongo (approved):

source/max-staleness/tests/ReplicaSetNoPrimary/Unavailable.json
```diff
@@ -0,0 +1,60 @@
+{
```
JamesKovacs, line 1: Why is this test named `Unavailable.json`? That would seem to indicate that there are no available servers. But this test includes 2 servers that should be ignored and one that can service the request. So why `Unavailable.json`?

JamesKovacs (commented):
See comments inline.

source/max-staleness/tests/ReplicaSetNoPrimary/Unavailable.yml
```diff
@@ -0,0 +1,31 @@
+# Clients check the topology description for support of maxStalenessSeconds
+# and throw an error if any available server reports maxWireVersion < 5.
+# Unavailable servers (type Unknown or PossiblePrimary) must be ignored.
+---
```
ShaneHarvey, line 4: What are the 3 dashes?

source/server-selection/server-selection.rst
```diff
@@ -458,7 +458,7 @@ with error code 160 (SERVER-24421).
 
 During server selection,
 drivers (but not mongos) MUST raise an error if ``maxStalenessSeconds`` is a positive number,
-and any server's ``maxWireVersion`` is less than 5. [#]_
```
ShaneHarvey, line 461: We can remove this error checking rule now that we've dropped support for 3.4 (DRIVERS-1914/DRIVERS-1315)

kevinAlbs: I kept the rule but clarified that it only applies if minWireVersion < 5. I noted the rationale in https://github.com/mongodb/specifications/pull/1064/files#r707775526.

ShaneHarvey (commented):
Changes LGTM and the new test passes in Python but do you mind resolving the merge conflicts?

JamesKovacs (approved):
LGTM

ShaneHarvey (approved):

PR #1065, opened 2021-09-07
DRIVERS-1875 Add BSON corpus test for binary subtype 7

eramongodb:
The new test only validates correct subtype number encoding/decoding and binary data encoding/decoding. The binary data itself is not validated. DRIVERS-1875 states the data intended to be stored is "meant to be opaque", so I believe this should satisfy current requirements.

### DIFF
--- a/source/bson-corpus/tests/binary.json
+++ b/source/bson-corpus/tests/binary.json
@@ -51,6 +51,11 @@
             "canonical_extjson": "{\"x\" : { \"$binary\" : {\"base64\" : \"c//SZESzTGmQ6OfR38A11A==\", \"subType\" : \"05\"}}}"
         },
         {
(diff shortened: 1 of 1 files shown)


jmikola:
Copying my previous comment from https://github.com/mongodb/mongo-c-driver/pull/859#pullrequestreview-748315888:

> Passes in PHPC.
> 
> Note: after merging, make sure you update the downstream change summary on [DRIVERS-1875](https://jira.mongodb.org/browse/DRIVERS-1875) to advise drivers to create a new constant for sub-type 0x07 (please suggest a name, such as `COLUMN`) and sync their BSON corpus tests to the merge commit.

jmikola (approved):

kevinAlbs (approved):

PR #1067, opened 2021-09-10
DRIVERS-521 Allow custom service names with srvServiceName URI option

benjirewis:
DRIVERS-521

Allow the use of custom SRV service names for SRV lookup using a new `srvServiceName` URI option. Clarify that during initial DNS seedlist discovery and SRV polling, the value of `srvServiceName` should be prefixed to the URI and `mongodb` should only be used as a service name if `srvServiceName` is not specified.

Prototype available in Go [here](https://github.com/benjirewis/mongo-go-driver/tree/customServiceName).

### DIFF
--- a/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
+++ b/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
@@ -10,8 +10,8 @@ Initial DNS Seedlist Discovery
 :Authors: Derick Rethans
 :Status: Draft
 :Type: Standards
-:Last Modified: 2019-04-15
-:Version: 1.4.0
+:Last Modified: 2021-09-xx
+:Version: 1.4.1
 :Spec Lead: Matt Broadstone
 :Advisory Group: \A. Jesse Jiryu Davis
 :Approver(s): Bernie Hackett, David Golden, Jeff Yemin, Matt Broadstone, A. Jesse Jiryu Davis
@@ -65,11 +65,13 @@ Seedlist Discovery
 ------------------
 
 In this preprocessing step, the driver will query the DNS server for SRV
-records on ``{hostname}.{domainname}``, prefixed with ``_mongodb._tcp.``:
-``_mongodb._tcp.{hostname}.{domainname}``. This DNS query is expected to
-respond with one or more SRV records. From the DNS result, the driver now MUST
-behave the same as if an ``mongodb://`` URI was provided with all the host
-names and port numbers that were returned as part of the DNS SRV query result.
+records on ``{hostname}.{domainname}``, prefixed with the SRV service name
+and protocol. The SRV service name is provided in the ``srvServiceName`` URI option and
+defaults to ``mongodb``. The protocol is always ``tcp``. After prefixing, the URI
+should look like: ``_{srvServiceName}._tcp.{hostname}.{domainname}``. This DNS query
+is expected to respond with one or more SRV records. From the DNS result, the driver
+now MUST behave the same as if an ``mongodb://`` URI was provided with all the host names
+and port numbers that were returned as part of the DNS SRV query result.
 
 The priority and weight fields in returned SRV records MUST be ignored.
 
@@ -273,6 +275,10 @@ SRV records.
 
 ChangeLog
 =========
+
+2021-09-xx - 1.4.1
+    Clarify that service name only defaults to ``mongodb``, and should be
+    defined by the ``srvServiceName`` URI option.
 
 2021-04-15 - 1.4.0
     Adding in behaviour for load balancer mode.
--- a/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
+++ b/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
@@ -11,7 +11,7 @@ Initial DNS Seedlist Discovery
 :Status: Draft
 :Type: Standards
 :Last Modified: 2021-09-xx
-:Version: 1.4.1
+:Version: 1.5.0
 :Spec Lead: Matt Broadstone
 :Advisory Group: \A. Jesse Jiryu Davis
 :Approver(s): Bernie Hackett, David Golden, Jeff Yemin, Matt Broadstone, A. Jesse Jiryu Davis
@@ -276,7 +276,7 @@ SRV records.
 ChangeLog
 =========
 
-2021-09-xx - 1.4.1
+2021-09-xx - 1.5.0
     Clarify that service name only defaults to ``mongodb``, and should be
     defined by the ``srvServiceName`` URI option.
 
--- a/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
+++ b/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
@@ -69,9 +69,9 @@ records on ``{hostname}.{domainname}``, prefixed with the SRV service name
 and protocol. The SRV service name is provided in the ``srvServiceName`` URI option and
 defaults to ``mongodb``. The protocol is always ``tcp``. After prefixing, the URI
 should look like: ``_{srvServiceName}._tcp.{hostname}.{domainname}``. This DNS query
-is expected to respond with one or more SRV records. The driver MUST use all the host
-names and port numbers that were returned as part of the DNS SRV query result in the same
-way as a ``mongodb://`` URI.
+is expected to respond with one or more SRV records. The driver MUST add all the host
+names and port numbers that were returned as part of the DNS SRV query result to the
+seedlist.
 
 The priority and weight fields in returned SRV records MUST be ignored.
 
--- a/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
+++ b/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
@@ -69,9 +69,9 @@ records on ``{hostname}.{domainname}``, prefixed with the SRV service name
 and protocol. The SRV service name is provided in the ``srvServiceName`` URI option and
 defaults to ``mongodb``. The protocol is always ``tcp``. After prefixing, the URI
 should look like: ``_{srvServiceName}._tcp.{hostname}.{domainname}``. This DNS query
-is expected to respond with one or more SRV records. From the DNS result, the driver
-now MUST behave the same as if an ``mongodb://`` URI was provided with all the host names
-and port numbers that were returned as part of the DNS SRV query result.
+is expected to respond with one or more SRV records. The driver MUST use all the host
+names and port numbers that were returned as part of the DNS SRV query result in the same
+way as a ``mongodb://`` URI.
 
 The priority and weight fields in returned SRV records MUST be ignored.
 
--- a/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
+++ b/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
@@ -10,7 +10,7 @@ Initial DNS Seedlist Discovery
 :Authors: Derick Rethans
 :Status: Draft
 :Type: Standards
-:Last Modified: 2021-09-xx
+:Last Modified: 2021-09-15
 :Version: 1.5.0
 :Spec Lead: Matt Broadstone
 :Advisory Group: \A. Jesse Jiryu Davis
@@ -276,7 +276,7 @@ SRV records.
 ChangeLog
 =========
 
-2021-09-xx - 1.5.0
+2021-09-15 - 1.5.0
     Clarify that service name only defaults to ``mongodb``, and should be
     defined by the ``srvServiceName`` URI option.
 
--- a/source/polling-srv-records-for-mongos-discovery/polling-srv-records-for-mongos-discovery.rst
+++ b/source/polling-srv-records-for-mongos-discovery/polling-srv-records-for-mongos-discovery.rst
@@ -9,8 +9,8 @@ Polling SRV Records for mongos Discovery
 :Author: Derick Rethans
 :Status: Accepted
 :Type: Standards
-:Last Modified: 2018-11-29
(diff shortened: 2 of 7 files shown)


source/uri-options/tests/srv-service-name-option.json
```diff
@@ -0,0 +1,24 @@
+{
+  "tests": [
+    {
+      "description": "SRV URI with custom srvServiceName",
+      "uri": "mongodb+srv://test3.test.build.10gen.cc/?srvServiceName=customname",
```
benjirewis, line 5: I'll have to speak with the BUILD team about setting up a custom SRV record with `customname` as a service name for this spec test to work. Will update with a ticket number on this thread.

benjirewis (commented):
This PR is an alternative to #1063. #1063 allows setting custom service names with a new URI scheme `mongodb+rawsrv`. A new URI scheme may create friction in third-party libraries that expect only two valid mongodb URI schemes (`mongodb://` and `mongdb+srv://`).

source/uri-options/uri-options.rst
```diff
@@ -71,6 +71,14 @@ is specified with an SRV URI, because the URI may resolve to multiple
 hosts. The driver MUST allow specifying ``directConnection=false`` URI
 option with an SRV URI.
 
+Non-SRV URI with srvServiceName URI option
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The driver MUST report an error if the ``srvServiceName`` URI option is
+specified with a non-SRV URI (i.e. ``mongodb://``) because SRV lookup
+only occurs with SRV URIs. The driver MUST allow specifying the ``srvServiceName``
+URI option with an SRV URI.
```
xdg, line 80: Personally, I wouldn't worry about this.  I think it's OK for the option to be ignored if not needed.

source/uri-options/tests/srv-service-name-option.json
jmikola: I assume we need a custom SRV record because testing with `srvServiceName=mongodb` wouldn't really distinguish between a driver ignoring `srvServiceName` altogether and using the value explicitly. Is that correct?

source/uri-options/uri-options.rst
```diff
@@ -269,6 +277,13 @@ pertaining to URI options apply here.
      - no
      - Amount of time spent attempting to send or receive on a socket before timing out; note that this only applies to application operations, not SDAM
 
+   * - srvServiceName
+     - any string
```
jmikola, line 281: Should an empty string be prohibited? Alternatively, should an empty string result in the default value ("mongodb") being used? I assume if we actually allow an empty string as an explicit value we'll get two sequential dots after concatenating, which could be problematic.

source/uri-options/uri-options.rst
jmikola, line 80: I was going to ask if we needed to explicitly prohibit `srvServiceName` in TXT records, but I was reminded that there is already an [explicit list of supported URI options](https://github.com/mongodb/specifications/blob/master/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst#default-connection-string-options) -- so nothing to worry about there.

source/uri-options/tests/srv-service-name-option.yml
```diff
@@ -0,0 +1,16 @@
+tests:
+    - description: "SRV URI with custom srvServiceName"
+      uri: "mongodb+srv://test3.test.build.10gen.cc/?srvServiceName=customname"
+      valid: true
+      warning: false
```
jmikola, line 5: Why does this produce a warning?

source/polling-srv-records-for-mongos-discovery/polling-srv-records-for-mongos-discovery.rst
```diff
@@ -79,8 +79,10 @@ rescan is similar, but not identical to the behaviour of initial seedlist
 discovery.  Periodic scan MUST follow these rules:
 
 - The driver will query the DNS server for SRV records on
-  ``{hostname}.{domainname}``, prefixed with ``_mongodb._tcp.``:
-  ``_mongodb._tcp.{hostname}.{domainname}``.
+  ``{hostname}.{domainname}``, prefixed with the SRV service name
+  and protocol. The SRV service name is provided in the ``srvServiceName`` URI option and
+  defaults to ``mongodb``. The protocol is always ``tcp``. After prefixing, the URI
+  should look like: ``_{srvServiceName}._tcp.{hostname}.{domainname}``..
```
jmikola, line 85:
```suggestion
  should look like: ``_{srvServiceName}._tcp.{hostname}.{domainname}``.
```

source/uri-options/uri-options.rst
xdg, line 281: This should restrict to strings that are valid SRV service names.  See [RFC 6335](https://datatracker.ietf.org/doc/html/rfc6335#section-5.1) and possibly [RFC 8553](https://datatracker.ietf.org/doc/html/rfc8553), though I might suggest being more lenient with length restrictions as long as the whole label with underscore is less than the 63 character limit for DNS.

xdg (changes requested):

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
```diff
@@ -10,8 +10,8 @@ Initial DNS Seedlist Discovery
 :Authors: Derick Rethans
 :Status: Draft
 :Type: Standards
-:Last Modified: 2019-04-15
-:Version: 1.4.0
+:Last Modified: 2021-09-xx
+:Version: 1.4.1
```
jmikola, line 14: I think this warrants a minor version bump, as was done for the URI options spec -- but I won't push for that if other feel strongly about the SRV and Polling SRV specs keeping patch bumps.

source/uri-options/tests/srv-service-name-option.json
jmikola: Will tests also be added to the [SRV spec](https://github.com/mongodb/specifications/tree/master/source/initial-dns-seedlist-discovery/tests)?

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
alcaeus: I agree here - a new minor version would also indicate to teams that there may be some effort required to support the new spec version, as opposed to only syncing spec tests and hoping for the best ;)

source/uri-options/uri-options.rst
```diff
@@ -278,7 +278,8 @@ pertaining to URI options apply here.
      - Amount of time spent attempting to send or receive on a socket before timing out; note that this only applies to application operations, not SDAM
 
    * - srvServiceName
-     - any string
+     - a valid SRV service name according to `RFC 6335 <https://datatracker.ietf.org/doc/html/rfc6335#section-5.1>`_; can be longer than 15 characters as long as the 63 character DNS query limit is
+       not surpassed
```
xdg, line 282: I'd be specific that the limit is 62 because the underscore is prepended.  Otherwise LGTM

xdg (approved):
LGTM with one minor suggestion.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
JamesKovacs: I would argue that this is more a `1.4.0` -> `1.5.0`. When I see `1.4.0` -> `1.4.1`, I'm expecting spelling corrections and clarifications, not new functionality. I'm not going to insist, but just my 2 cents.

JamesKovacs: Apologies. I was looking at an outdated diff. You did bump to `1.5.0`. I'm good with that.

source/uri-options/uri-options.rst
```diff
@@ -455,13 +471,14 @@ this specification MUST be updated to reflect those changes.
 Changes
 -------
 
+- 2021-09-xx Add srvServiceName option
```
JamesKovacs, line 474: Needs correct date.

source/uri-options/uri-options.rst
```diff
@@ -3,15 +3,15 @@ URI Options Specification
...
+:Spec Version: 1.8.0
 :Author: Sam Rossi
 :Spec Lead: Bernie Hackett
 :Advisory Group: Scott L'Hommedieu
 :Approver(s): Cailin Nelson, Jeff Yemin, Matt Broadstone, Dan Pasette, Prashant Mital, Spencer Jackson
 :Informed: drivers@
 :Status: Accepted (Could be Draft, Accepted, Rejected, Final, or Replaced)
 :Type: Standards
-:Last Modified: 2021-4-15
+:Last Modified: 2021-9-xx
```
JamesKovacs, line 14: Needs correct date.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
```diff
@@ -65,11 +65,13 @@ Seedlist Discovery
...
-respond with one or more SRV records. From the DNS result, the driver now MUST
-behave the same as if an ``mongodb://`` URI was provided with all the host
-names and port numbers that were returned as part of the DNS SRV query result.
+records on ``{hostname}.{domainname}``, prefixed with the SRV service name
+and protocol. The SRV service name is provided in the ``srvServiceName`` URI option and
+defaults to ``mongodb``. The protocol is always ``tcp``. After prefixing, the URI
+should look like: ``_{srvServiceName}._tcp.{hostname}.{domainname}``. This DNS query
+is expected to respond with one or more SRV records. From the DNS result, the driver
+now MUST behave the same as if an ``mongodb://`` URI was provided with all the host names
+and port numbers that were returned as part of the DNS SRV query result.
```
JamesKovacs, line 74: This is actually technically not true as the default options for `mongodb://` and `mongodb+srv://` are different. For example `mongodb+srv://` implies `&tls=true` whereas `mongodb://` does not. As well `mongodb+srv://` must read additional options from the associated TXT record whereas `mongodb://` does not. So this statement can be misinterpreted. Please clarify.

source/uri-options/uri-options.rst
benjirewis: Done!

JamesKovacs (changes requested):
Overall looking good. The part on `mongodb+srv://` and `mongodb://` equivalency requires clarification as it can be misinterpreted.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
benjirewis: That's fair. I didn't write this language, but I've changed it to be more specific and not claim that drivers must "behave the same as a `mongodb://` URI was provided".

source/uri-options/uri-options.rst
benjirewis: I'll write in the current date when I'm about to merge; that way the last modified date is in fact accurate.

source/uri-options/uri-options.rst
benjirewis: See above.

source/uri-options/uri-options.rst
```diff
@@ -269,6 +277,14 @@ pertaining to URI options apply here.
      - no
      - Amount of time spent attempting to send or receive on a socket before timing out; note that this only applies to application operations, not SDAM
 
+   * - srvServiceName
+     - a valid SRV service name according to `RFC 6335 <https://datatracker.ietf.org/doc/html/rfc6335#section-5.1>`_; can be longer than 15 characters as long as the 63 (62 with prepended underscore)
+       character DNS query limit is not surpassed
+     - "mongodb"
+     - no
+     - the service name to use for SRV lookup in `initial DNS seedlist discovery <https://github.com/mongodb/specifications/blob/master/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst>`_
+       and `SRV polling <https://github.com/mongodb/specifications/blob/master/source/polling-srv-records-for-mongos-discovery/polling-srv-records-for-mongos-discovery.rst>`_
```
JamesKovacs, line 286: Relative URIs should be used for references to other specs. e.g. `../source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst`

JamesKovacs (approved):
LGTM. Added minor note on using relative URIs to other specs.

source/uri-options/tests/srv-service-name-option.yml
jmikola: `~` (i.e. null) is used for things like `hosts` and `auth`, but I've only seen `true` and `false` used for `warning`. That agrees with https://github.com/mongodb/specifications/tree/master/source/uri-options/tests#format

Should this be `warning: false`? I think it was `warning: true` when I originally left a comment about this.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
jmikola: This occurred to me while I was working on https://github.com/mongodb/specifications/pull/1069 as well, since I ended up modifying this paragraph in that PR. I think we should just state that the results get added to the seedlist and leave it at that.

source/uri-options/tests/srv-service-name-option.yml
benjirewis: It was originally `warning: false`, so I'll just switch back to that. It makes sense to assert the lack of warning in the success case.

jmikola: Oh ok, I totally misread the diff when I wrote my original comment. Sorry about that.

source/uri-options/uri-options.rst
jmikola: @benjirewis Don't forget the leading zero for "09" when you do. I'm not sure why some of the specs omit that, but I think leading zeroes make sense for YYYYMMDD date formats.

jmikola (approved):

PR #1068, opened 2021-09-13
Fix link to load balancer spec

jmikola:


JamesKovacs (changes requested):
Line 83 requires updating as well.

JamesKovacs (approved):
LGTM

PR #1069, opened 2021-09-14
DRIVERS-1519: srvMaxHosts option for limiting SRV seeding

jmikola:
https://jira.mongodb.org/browse/DRIVERS-1519


source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
```diff
@@ -10,8 +10,8 @@ Initial DNS Seedlist Discovery
 :Authors: Derick Rethans
 :Status: Draft
 :Type: Standards
-:Last Modified: 2019-04-15
-:Version: 1.4.0
+:Last Modified: 2019-09-XX
+:Version: 1.5.0
```
jmikola, line 14: Note: I'll rebase this after https://github.com/mongodb/specifications/pull/1067 is merged.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
```diff
@@ -97,14 +96,52 @@ It is an error to specify more than one host name in a connection string with
...
+``{domainname}``.
+
+The driver MUST NOT attempt to connect to any hosts until the DNS query has
+returned its results.
+
+If ``srvMaxHosts`` is zero or greater than or equal to the number of hosts in
+the DNS result, the driver MUST populate the seedlist with all hosts. That is to
+say, the driver now MUST behave the same as if a ``mongodb://`` URI was provided
+with all the host names and port numbers that were returned as part of the DNS
+SRV query result.
```
jmikola, line 138: This paragraph will be revised after rebasing. See: https://github.com/mongodb/specifications/pull/1067#discussion_r708624023

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
```diff
@@ -97,14 +96,52 @@ It is an error to specify more than one host name in a connection string with
...
+
+If ``srvMaxHosts`` is greater than zero and less than the number of hosts in the
+DNS result, the driver MUST randomly select that many hosts and use them to
```
behackett, line 141: What algorithm? I believe the choice made a big difference when @mhelmstetter implemented this.

jmikola: @mhelmstetter's code did the following:

```js
var numMongos = 2;
var mongoHosts = hosts
    .sort(() => Math.random() - Math.random())
    .slice(0, numMongos)
    .map(h => `${h.host}:${h.port}`)
    .join(',');
```

This approach seems like an abuse of [`sort()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#description), since the callback isn't considering the elements. This [Stack Overflow answer](https://stackoverflow.com/a/18650169) also explains why using `sort()` in this capacity may be problematic (per the [linked documentation](https://javascript.info/array-methods#shuffle-an-array)).

I suppose Mark can chime in here. I didn't put much thought into a selection algorithm, although I did consider whether this might warrant using cryptographically secure random generation to avoid cases where multiple processes seed their RNG with the same value and make the same "random" selections (akin to the issue in [CDRIVER-3978)](https://jira.mongodb.org/browse/CDRIVER-3978)). I concluded that probably wasn't necessary, as the server selection spec doesn't go into that detail when it discusses [random selection](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#random-selection-within-the-latency-window-single-threaded) either.

When you talk about an algorithm, are you thinking about something like the following?

 * Given the list of SRV results, select X indexes in the list randomly and use those hosts.
 * Given the list of SRV hosts, select an index randomly. Repeat this process X-1 times.

I'm not sure if this makes any difference in practice. Perhaps it's more significant when we get into SRV polling and needing to make random selections from the subset of "valid new hosts" -- but even then those would just be a set/array of hosts after segregating them from existing hosts in the topology.

mhelmstetter: I think that you may have been looking at an older version of code that I wrote, which as @behackett pointed out, had some issues with producing an uneven distribution.
The was a [new version of that code](https://github.com/mhelmstetter/hello_node/blob/main/app2.js) has a `getRandomSubarray()` function, which implements the Fisher-Yates shuffle. The implementation was based on [this StackOverflow answer](https://stackoverflow.com/questions/11935175/sampling-a-random-subset-from-an-array).

jmikola: @mhelmstetter: Thanks for the link. I was only looking at the snippet you shared in Slack a few weeks ago, so it's helpful to see this final copy.

mhelmstetter: I've updated the code link to the code above, as the previous link was a private repo.

jmikola: Reading through [ Fisher-Yates Shuffle: Modern Algorithm](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#The_modern_algorithm), that seems comparable to what I was thinking in my [earlier comment](https://github.com/mongodb/specifications/pull/1069#discussion_r709527132); however, I didn't clarify that we should take steps to avoid duplicate selections. In the documented algorithm, the selected element is swapped out to the end of the list, which ensures that random selection doesn't select the same index twice.

Since we're only doing a random sampling, I think a driver could also just abort the shuffle once `srvMaxHosts` choices have been made. So, given the suggested "modern algorithm" that swaps selected elements to the start of the list:

```
-- To shuffle an array a of n elements (indices 0..n-1):
for i from 0 to n−2 do
     j ← random integer such that i ≤ j < n
     exchange a[i] and a[j]
```

A driver could break out of the loop once `srvMaxHosts` have been chosen and then take that many elements from the start of the list.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
```diff
@@ -61,35 +64,40 @@ format is::
...
+
+srvServiceName
+~~~~~~~~~~~~~~
+
+This option specifies a valid SRV service name according to
+`RFC 6335 <https://datatracker.ietf.org/doc/html/rfc6335#section-5.1>`_, with
+the exception that it may exceed 15 characters as long as the 63 (62 with
+prepended underscore) character DNS query limit is not surpassed. This option
+requires a string value is defaults to "mongodb". This option MUST only be
+configurable at the level of a ``MongoClient``.
```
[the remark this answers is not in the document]
benjirewis: Great, thanks for linking the table cell to refer here.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
```diff
@@ -61,35 +64,40 @@ format is::
...
-``{domainname}``.
+
+srvServiceName
+~~~~~~~~~~~~~~
+
+This option specifies a valid SRV service name according to
+`RFC 6335 <https://datatracker.ietf.org/doc/html/rfc6335#section-5.1>`_, with
+the exception that it may exceed 15 characters as long as the 63 (62 with
+prepended underscore) character DNS query limit is not surpassed. This option
+requires a string value is defaults to "mongodb". This option MUST only be
```
benjirewis, line 92:
```suggestion
requires a string value and defaults to "mongodb". This option MUST only be
```

jmikola:
Opening this up for review. I've yet to POC this, as I expect that will need to be done in libmongoc, but folks can start reviewing the current design.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
alcaeus: I wouldn't have the spec require a specific algorithm, and instead require drivers to use the same algorithm that is used for server selection. While driver implementations would differ, each driver would have a single implementation which would allow for an easier introduction of a common randomiser spec for drivers.

behackett: We can certainly write a randomiser spec in the future, but now is not the time. For now, the algorithm should be spelled out here.

source/initial-dns-seedlist-discovery/tests/README.rst
```diff
@@ -76,28 +85,49 @@ Test Format and Use
...
 
 .. _`Connection String`: ../../connection-string/connection-string-spec.rst
+.. _`URI options`: ../../uri-options/uri-options.rst
+
+For each file, create MongoClient initialized with the ``mongodb+srv``
+connection string.
+
+If ``seeds`` is specified, drivers SHOULD verify that the set of hosts in the
+client's initial seedlist matches the list in ``seeds``. If ``numSeeds`` is
+specified, drivers SHOULD verify that the size of that set matches ``numSeeds``.
```
benjirewis, line 115: Why are these both `SHOULD` instead of `MUST`?

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
benjirewis: I'll try the modern Fisher-Yates shuffle in a Go driver POC ([GODRIVER-2092](https://jira.mongodb.org/browse/GODRIVER-2092)) and report back here. 

I agree that spelling out the algorithm here is better than leaving it up to individual drivers. Without specific instruction, I probably would have used [`rand.Shuffle`](https://pkg.go.dev/math/rand#Shuffle) and selected the first `srvMaxHosts` items in Go. That would have been `O(n^2)` time complexity which could have been a significant performance hit (the modern algorithm should be `O(n)`).

source/initial-dns-seedlist-discovery/tests/replica-set/srvMaxHosts-equal_to_srv_records.yml
```diff
@@ -0,0 +1,15 @@
+# When srvMaxHosts equals the number of SRV records, all hosts are added to the
```
benjirewis, line 1: Are you planning on adding the associated JSON test files later?

source/polling-srv-records-for-mongos-discovery/polling-srv-records-for-mongos-discovery.rst
```diff
@@ -105,11 +105,15 @@ discovery.  Periodic scan MUST follow these rules:
...
 
-  - MUST add each valid new host to the topology as Unknown
   - MUST remove all hosts that are part of the topology, but are no longer
     in the returned set of valid hosts
   - MUST NOT remove all hosts, and then re-add the ones that were returned.
     Hosts that have not changed, MUST be left alone and unchanged.
+  - If srvMaxHosts_ is zero or greater than or equal to the number of valid
+    hosts, each valid new host MUST be added to the topology as Unknown.
+  - If srvMaxHosts_ is greater than zero and less than the number of valid
+    hosts, valid new hosts MUST be randomly selected and added to the topology
```
benjirewis, line 115: I suppose you'll want to document the random selection algorithm here as well.

source/polling-srv-records-for-mongos-discovery/tests/README.rst
```diff
@@ -126,3 +126,67 @@ mock the addition of the following DNS record::
...
 Wait until ``2*rescanSRVIntervalMS`` and assert that the final topology description
 only contains one server: ``localhost.test.build.10gen.cc.`` at port ``27017``.
+
+
+SRV polling with srvMaxHosts MongoClient option
+```````````````````````````````````````````````
+
+The following tests MUST setup a MongoClient using the ``srvMaxHosts`` option
+and ``test1.test.build.10gen.cc`` SRV record. Each test MUST mock the described
+DNS changes and then verify the the new list of hosts is present.
```
benjirewis, line 136:
```suggestion
DNS changes and then verify that the new list of hosts is present.
```

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
jmikola: Added a sentence to advise use of Fisher-Yates shuffle. Opting for SHOULD, since drivers may intentionally want to use another algorithm (e.g. something in their standard library that is equivalent or better). I only require that randomization during polling use the same algorithm, since any driver would be capable of enforcing consistency there.

source/polling-srv-records-for-mongos-discovery/polling-srv-records-for-mongos-discovery.rst
jmikola: This is now done.

source/polling-srv-records-for-mongos-discovery/tests/README.rst
```diff
@@ -126,3 +126,67 @@ mock the addition of the following DNS record::
...
+Replace the following record::
+
+    _mongodb._tcp.test1.test.build.10gen.cc.  86400  IN SRV  27018  localhost.test.build.10gen.cc.
+
+with::
+
+    _mongodb._tcp.test1.test.build.10gen.cc.  86400  IN SRV  27019  localhost.test.build.10gen.cc.
+    _mongodb._tcp.test1.test.build.10gen.cc.  86400  IN SRV  27020  localhost.test.build.10gen.cc.
+
+Assert that the topology ultimately contains the following hosts:
```
kevinAlbs, line 153: Can this include a timeout?

```suggestion
Wait until ``2*rescanSRVIntervalMS`` and assert that the final topology description contains the following hosts:
```

source/polling-srv-records-for-mongos-discovery/tests/README.rst
```diff
@@ -126,3 +126,67 @@ mock the addition of the following DNS record::
...
+---------------------------------------------------------
+
+Configure the MongoClient with ``srvMaxHosts=2``.
+
+Replace both records with::
+
+    _mongodb._tcp.test1.test.build.10gen.cc.  86400  IN SRV  27019  localhost.test.build.10gen.cc.
+    _mongodb._tcp.test1.test.build.10gen.cc.  86400  IN SRV  27020  localhost.test.build.10gen.cc.
+
+Assert that the topology ultimately contains the following hosts:
```
kevinAlbs, line 170:
```suggestion
Wait until ``2*rescanSRVIntervalMS`` and assert that the final topology description contains the following hosts:
```

source/polling-srv-records-for-mongos-discovery/tests/README.rst
```diff
@@ -126,3 +126,67 @@ mock the addition of the following DNS record::
...
+Replace the following record::
+
+    _mongodb._tcp.test1.test.build.10gen.cc.  86400  IN SRV  27018  localhost.test.build.10gen.cc.
+
+with::
+
+    _mongodb._tcp.test1.test.build.10gen.cc.  86400  IN SRV  27019  localhost.test.build.10gen.cc.
+    _mongodb._tcp.test1.test.build.10gen.cc.  86400  IN SRV  27020  localhost.test.build.10gen.cc.
+
+Assert that the topology ultimately has two hosts present and that one of the
```
kevinAlbs, line 190:
```suggestion
Wait until ``2*rescanSRVIntervalMS`` and assert that the topology has two hosts present and that one of the
```

kevinAlbs (commented):
This is looking great! I appreciate the clean up in the test READMEs.

I have a question about whether this should be prohibited with non-SRV URIs, and how it plays with replicaSet.

source/initial-dns-seedlist-discovery/tests/sharded/srvMaxHosts-equal_to_srv_records.yml
```diff
@@ -0,0 +1,13 @@
+# When srvMaxHosts equals the number of SRV records, all hosts are added to the
+# seed list.
+uri: "mongodb+srv://test1.test.build.10gen.cc/?srvMaxHosts=2"
+numSeeds: 2
+seeds:
+    - localhost.test.build.10gen.cc:27017
+    - localhost.test.build.10gen.cc:27018
+hosts:
+    - localhost:27017
+    - localhost:27018
```
jmikola, line 10: @benjirewis: Since you mentioned these assertions fail in the Go driver

I see that [tests/load-balanced/loadBalanced-true-txt.yml](https://github.com/mongodb/specifications/blob/master/source/initial-dns-seedlist-discovery/tests/load-balanced/loadBalanced-true-txt.yml#L5) includes the following note:

> In LB mode, the driver does not do server discovery, so the hostname does not get resolved to localhost:27017.

When I wrote these tests, I was assuming that the mongos host name would be corrected to the canonical host reported in the `hello` response as part of SDAM, as is done in the replica set tests (e.g. [tests/replica-set/loadBalanced-false-txt.yml](https://github.com/mongodb/specifications/blob/master/source/initial-dns-seedlist-discovery/tests/replica-set/loadBalanced-false-txt.yml)).

That said, we didn't have existing tests for sharded clusters so I very well could have made a mistake here. I'll note that the [Polling SRV tests](https://github.com/mongodb/specifications/blob/master/source/polling-srv-records-for-mongos-discovery/tests/README.rst) discuss changing DNS records and go on to say "verify that the new list of hosts is present". They don't actually specify what the hosts should be, so we might assume that those host names should be identical to the seeds provided by DNS.

AFAIK, the only mention of host name validation in the SDAM spec talks about it from the context of replica set connections:

 * [Clients use the hostnames listed in the replica set config, not the seed list](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#clients-use-the-hostnames-listed-in-the-replica-set-config-not-the-seed-list)
 * [Using me field to detect seed list members that do not match host names in the replica set configuration](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#using-me-field-to-detect-seed-list-members-that-do-not-match-host-names-in-the-replica-set-configuration)

When [TopologyType explanations](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#topologytype-explanations) discussed the sharded topology, it only talks about keeping servers with the type Unknown or Mongos and removing all other servers. There is no mention of host name validation like we find in **updateRSWithoutPrimary**:

```
if description.address != description.me:
    remove this server from topologyDescription and stop monitoring it
    return
```

Given that, I think the SDAM spec's silence on the matter for mongos suggests that host name canonicalization does _not_ take place like it does for replica sets. Perhaps @ShaneHarvey or @nbbeeken can chime in and confirm.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
```diff
@@ -61,35 +64,40 @@ format is::
...
+`SRV polling <../polling-srv-records-for-mongos-discovery/polling-srv-records-for-mongos-discovery.rst>`_.
+This option requires a non-negative integer and defaults to zero (i.e. no
+limit). This option MUST only be configurable at the level of a ``MongoClient``.
 
-A driver MUST verify that in addition to the ``{hostname}``, the
-``{domainname}`` consists of at least two parts: the domain name, and a TLD.
-Drivers MUST raise an error and MUST NOT contact the DNS server to obtain SRV
-(or TXT records) if the full URI does not consists of at least three parts.
+This option is primarily used to limit the number of mongos connections that may
+be created for sharded topologies.
```
jmikola, line 82: @benjirewis pointed out that [`srvServiceName` is prohibited for non-SRV schemes](https://github.com/mongodb/specifications/blob/master/source/uri-options/uri-options.rst#non-srv-uri-with-srvservicename-uri-option) (tested by [srv-service-name-option.yml](https://github.com/mongodb/specifications/blob/master/source/uri-options/tests/srv-service-name-option.yml)). That makes sense, so I'll revise this PR to make the same prohibition for `srvMaxHosts` and add a corresponding URI test.

I do think it'd be worth repeating that in the SRV spec itself (not just URI options), so I'll amend both of these sections as well.

source/initial-dns-seedlist-discovery/tests/README.rst
```diff
@@ -10,11 +10,20 @@ Test Setup
...
-servers running on any open ports. In both cases, the clusters MUST be
-started with SSL enabled.
+replica set name ``repl0``.
+
+The tests in the ``load-balanced`` directory MUST be executed against a
+load-balanced sharded cluster with the mongos servers running on localhost ports
+27017 and 27018 (corresponding to the script in `drivers-evergreen-tools`_). The
+load balancers, shard servers, and config servers may run on any open ports.
+
+.. _`drivers-evergreen-tools`: ../../connection-string/connection-string-spec.rst
```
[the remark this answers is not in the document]
jmikola: Oh, I think so. I'm not sure where this `connection-string-spec.rst` link came from. Must be a typo on my part.

source/uri-options/uri-options.rst
```diff
@@ -277,12 +277,19 @@ pertaining to URI options apply here.
...
+   * - srvMaxHosts
+     - non-negative integer; 0 means no maximum
+     - defined in the `Initial DNS Seedlist Discovery spec <../initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst#srvmaxhosts>`_
+     - no
+     - The maximum number of SRV results to randomly select when initially
+       populating the seedlist or, during SRV polling, adding new hosts to the
+       topology.
+
    * - srvServiceName
      - a valid SRV service name according to `RFC 6335 <https://datatracker.ietf.org/doc/html/rfc6335#section-5.1>`_; can be longer than 15 characters as long as the 63 (62 with prepended underscore)
```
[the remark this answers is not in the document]
jmikola: @benjirewis: Any objections? I'd be in favor of that to reduce duplication.

source/initial-dns-seedlist-discovery/tests/README.rst
```diff
@@ -78,28 +87,49 @@ Test Format and Use
...
+specified, drivers SHOULD verify that the size of that set matches ``numSeeds``.
+
+If ``hosts`` is specified, drivers MUST verify that the set of
+ServerDescriptions in the client's TopologyDescription eventually matches the
+list in ``hosts``. If ``numHosts`` is specified, drivers MUST verify that the
+size of that set matches ``numHosts``.
+
+If ``options`` is specified, drivers MUST verify each of the values under
+``options`` match the MongoClient's parsed value for that option. There may be
+other options parsed by the Client as well, which a test does not verify.
```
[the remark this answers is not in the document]
jmikola:
```suggestion
other options parsed by the MongoClient as well, which a test does not verify.
```

source/polling-srv-records-for-mongos-discovery/tests/README.rst
jmikola: Applied manually.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
behackett: I'd get rid of the word "primarily". There isn't, currently, any other use case.

source/initial-dns-seedlist-discovery/tests/replica-set/srvMaxHosts-less_than_srv_records.yml
```diff
@@ -0,0 +1,14 @@
+# When srvMaxHosts is less than the number of SRV records, a random subset of
+# hosts are added to the seed list. We cannot anticipate which hosts will be
+# selected, so this test uses numSeeds instead of seeds. Since this is a replica
+# set, all hosts should ultimately be discovered by SDAM.
```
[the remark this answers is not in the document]
benjirewis: I agree that prohibiting the use of `srvMaxHosts` when connecting to a replica set is more effort than it's worth.

source/initial-dns-seedlist-discovery/tests/sharded/srvMaxHosts-equal_to_srv_records.yml
benjirewis: Sounds good, thanks for investigating. The Go driver implementation (still in review), does currently [use a modified version of these sharded tests](https://github.com/mongodb/mongo-go-driver/pull/764/files#diff-a5930af71de27d03199a7a6d483d4098cc02aba763d4a68b5f1d5659e86d60bdR8-R10) with the non-canonicalized host names.

source/uri-options/uri-options.rst
```diff
@@ -75,8 +75,8 @@ Non-SRV URI with srvServiceName URI option
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The driver MUST report an error if the ``srvServiceName`` URI option is
-specified with a non-SRV URI (i.e. ``mongodb://``) because SRV lookup
-only occurs with SRV URIs. The driver MUST allow specifying the ``srvServiceName``
+specified with a non-SRV URI (i.e. ``mongodb://``) because SRV lookup only
+occurs with SRV URIs. The driver MUST allow specifying the ``srvServiceName``
 URI option with an SRV URI.
```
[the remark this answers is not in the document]
benjirewis: I think so.

source/uri-options/uri-options.rst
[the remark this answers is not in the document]
benjirewis: Fine with me to remove.

benjirewis (commented):
BTW I used the native Go [`rand#Shuffle`](https://pkg.go.dev/math/rand#Shuffle) for the modern Fisher-Yates algorithm. I imagine many modern languages will have native "shuffle" functions that are `O(n)` and will not have to implement their own.

source/initial-dns-seedlist-discovery/tests/sharded/srvMaxHosts-equal_to_srv_records.yml
jmikola: Updated all sharded tests to use the same host names in `seeds` and `hosts`.

source/uri-options/uri-options.rst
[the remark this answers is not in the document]
jmikola: This is now addressed by a recent commit. I've also added URI options tests.

source/uri-options/uri-options.rst
```diff
@@ -277,12 +277,19 @@ pertaining to URI options apply here.
      - no
      - Amount of time spent attempting to send or receive on a socket before timing out; note that this only applies to application operations, not SDAM
 
+   * - srvMaxHosts
```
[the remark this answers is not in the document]
jmikola: This is now done.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
```diff
@@ -10,7 +10,7 @@ Initial DNS Seedlist Discovery
 :Authors: Derick Rethans
 :Status: Draft
 :Type: Standards
-:Last Modified: 2019-09-XX
+:Last Modified: 2019-10-XX
```
kevinAlbs, line 13:
```suggestion
:Last Modified: 2021-10-XX
```

kevinAlbs (approved):
LGTM!

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
```diff
@@ -99,14 +110,54 @@ It is an error to specify more than one host name in a connection string with
 the ``mongodb+srv`` protocol, and the driver MUST raise a parse error and MUST
 NOT do DNS resolution or contact hosts.
 
-The driver MUST NOT attempt to connect to any hosts until the DNS query has
-returned its results.
+A driver MUST verify that in addition to the ``{hostname}``, the
+``{domainname}`` consists of at least two parts: the domain name, and a TLD.
+Drivers MUST raise an error and MUST NOT contact the DNS server to obtain SRV
+(or TXT records) if the full URI does not consist of at least three parts.
```
JamesKovacs, line 116: This additional validation of the `domainname` is orthogonal to `srvMaxHosts`. We should create separate tickets for the validation and `srvMaxHosts` implementations.

source/initial-dns-seedlist-discovery/tests/README.rst
```diff
@@ -78,28 +87,49 @@ Test Format and Use
...
-  TXT records
-- ``parsed_options``: additional options present in the `Connection String`_
-  URI such as ``Userinfo`` (as ``user`` and ``password``), and ``Auth
-  database`` (as ``auth_database``).
+- ``numHosts``: the expected number of hosts discovered once SDAM completes a
+  scan. This is mainly used to test ``srvMaxHosts``, since randomly selected
+  hosts cannot be deterministically asserted.
+- ``options``: the parsed `URI options`_ as discovered from the
+  `Connection String`_'s "Connection Options" component and SRV resolution
+  (e.g. TXT records, implicit ``ssl`` default).
```
JamesKovacs, line 101: Should be "implicit `tls` default".

source/polling-srv-records-for-mongos-discovery/polling-srv-records-for-mongos-discovery.rst
```diff
@@ -79,10 +79,10 @@ rescan is similar, but not identical to the behaviour of initial seedlist
...
 
 - The driver will query the DNS server for SRV records on
-  ``{hostname}.{domainname}``, prefixed with the SRV service name
-  and protocol. The SRV service name is provided in the ``srvServiceName`` URI option and
-  defaults to ``mongodb``. The protocol is always ``tcp``. After prefixing, the URI
-  should look like: ``_{srvServiceName}._tcp.{hostname}.{domainname}``.
+``{hostname}.{domainname}``, prefixed with the SRV service name and protocol.
+  The SRV service name is provided in the srvServiceName_ URI option and
+  defaults to ``mongodb``. The protocol is always ``tcp``. After prefixing, the
+  URI should look like: ``_{srvServiceName}._tcp.{hostname}.{domainname}``.
```
JamesKovacs, line 85: Should we add the validation for `domainname` consisting of `domain` + TLD here as well?

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
```diff
@@ -61,35 +64,43 @@ format is::
...
+limit). This option MUST only be configurable at the level of a ``MongoClient``.
+Drivers MUST report an error if this option is specified and the scheme is not
+``mongodb+srv``.
 
-A driver MUST verify that in addition to the ``{hostname}``, the
-``{domainname}`` consists of at least two parts: the domain name, and a TLD.
-Drivers MUST raise an error and MUST NOT contact the DNS server to obtain SRV
-(or TXT records) if the full URI does not consists of at least three parts.
+This option is used to limit the number of mongos connections that may be
+created for sharded topologies.
```
JamesKovacs, line 84: Should we perform any validation of URI options after resolution of the TXT record? Options that seem to be incompatible with `srvMaxHosts` include:

- `loadBalanced=true`
- `replicaSet` (see comment 1 below)
- `directConnection`

I would expect the driver to throw an exception if any of those incompatible options were present along with `srvMaxHosts`.

_comment 1_ It was mentioned that checking for `srvMaxHosts` and `replicaSet` was more trouble than it is worth. Maybe I'm missing some subtlety, but why couldn't we verify that `srvMaxHosts` and `replicaSet` aren't present in the URI and/or TXT record?

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
```diff
@@ -99,14 +110,54 @@ It is an error to specify more than one host name in a connection string with
...
 
-The driver MUST NOT attempt to connect to any hosts until the DNS query has
-returned its results.
+A driver MUST verify that in addition to the ``{hostname}``, the
+``{domainname}`` consists of at least two parts: the domain name, and a TLD.
+Drivers MUST raise an error and MUST NOT contact the DNS server to obtain SRV
+(or TXT records) if the full URI does not consist of at least three parts.
+
+If ``mongodb+srv`` is used, a driver MUST implicitly also enable TLS. Clients
+can turn this off by passing ``ssl=false`` in either the Connection String,
```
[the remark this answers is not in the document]
jmikola: This was copied from the original text while I was reorganizing the headings but I can change it in this PR.

source/initial-dns-seedlist-discovery/tests/README.rst
jmikola:
```suggestion
  (e.g. TXT records, implicit ``tls`` default).
```

source/polling-srv-records-for-mongos-discovery/polling-srv-records-for-mongos-discovery.rst
jmikola: This seems outside the scope of this PR. If you'll be creating a DRIVERS ticket for https://github.com/mongodb/specifications/pull/1069/files#r724565721 would you like to lump this in as well?

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
jmikola: For context, the only options currently supported in a TXT record are `authSource`, `replicaSet`, and `loadBalanced`.

I think we could still end up connecting to a replica set topology even in the absence of a `replicaSet` URI option. Omitting that option just omits replica set name validation and initializes the topology to an Unknown state. From there, it can still transition based on the servers that are discovered.

`directConnection=true` is already prohibited for `mongodb+srv` schemes, so I don't think we need to do anything there. `directConnection=false` is not a logical conflict for SRV, and I believe the URI options spec was already amended to allow that.

That leaves `loadBalanced=true` (again, `false` is not an issue). In that case, the topology type won't be sharded so `srvMaxHosts` will have no effect (similar to using it with a replica set).

If you or anyone else is strongly in favor of this validation I can add it. I reckon we would just prohibit `srvMaxHosts` it for `loadBalanced=true` and the presence of a `replicaSet` URI option, and do nothing if the topology later becomes a replica set after SDAM kicks in.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
JamesKovacs: I agree that it is outside the scope of this PR, which is why I didn't want this change getting lost. We can create a separate DRIVERS ticket to track. You mentioned that it was "copied from the original text". Which original text are you talking about? I can create a DRIVERS ticket, but want to make sure we don't have one already if this text originated from another change.

jmikola: I believe this text goes back to the very first commits for this spec. See: https://github.com/mongodb/specifications/blame/master/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst#L84

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
jmikola: Raising an error for conflicting URI args seems like a small ask and easy to validate. I'll make the change and add a few tests. Will also explicitly note that we'll ignore the edge case where `replicaSet` is omitted but the user ends up connecting to a replica set topology anyway.

jmikola:
@JamesKovacs: Added new spec language to discuss URI Validation for `srvMaxHosts`.

@benjirewis: Since you have a running POC in Go, can you take a look at the new tests (in both SRV and URI Options specs)?

source/uri-options/tests/srv-options.yml
```diff
@@ -22,10 +22,68 @@ tests:
...
+      uri: "mongodb+srv://test1.test.build.10gen.cc/?srvMaxHosts=0&replicaSet=foo"
+      valid: true
+      warning: false
+      hosts: ~
+      auth: ~
+      options:
+          replicaSet: foo
+          srvMaxHosts: 0
+    - description: "SRV URI with srvMaxHosts=0 and loadBalanced=true"
+      uri: "mongodb+srv://test1.test.build.10gen.cc/?srvMaxHosts=0&loadBalanced=true"
```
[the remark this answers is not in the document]
jmikola: Ah, I wasn't aware that the URI Options tests were actually doing SRV lookups. I suppose this isn't really addressed by the test README since `srv-options.yml` is the first file to actually use `mongodb+srv://`.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst
```diff
@@ -276,6 +343,11 @@ SRV records.
 ChangeLog
 =========
 
+2021-10-XX - 1.6.0
```
jmikola, line 346:
```suggestion
2021-10-14 - 1.6.0
```

source/polling-srv-records-for-mongos-discovery/polling-srv-records-for-mongos-discovery.rst
```diff
@@ -229,6 +237,9 @@ No future work is expected.
 Changelog
 =========
 
+2021-09-XX - 1.2.0
```
jmikola, line 240:
```suggestion
2021-10-14 - 1.2.0
```

PR #1071, opened 2021-09-21
DRIVERS-1929 Add missing "hello" to load balancer spec test

ShaneHarvey:


source/load-balancers/tests/sdam-error-handling.yml
```diff
@@ -153,7 +153,7 @@ tests:
             configureFailPoint: failCommand
             mode: { times: 1 }
             data:
-              failCommands: [isMaster]
+              failCommands: [isMaster, hello]
```
jmikola, line 156: I haven't worked on these tests yet, but IIRC this is the same approach used in several SDAM and CMAP tests so I'll assume all is well here.

jmikola (approved):

source/load-balancers/tests/sdam-error-handling.yml
ShaneHarvey: Yep it's the same, see https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/tests/pool-create-min-size-error.yml#L13

PR #1075, opened 2021-10-04
DRIVERS-1746 Add native support for AWS IAM Roles for service accounts, EKS in particular

kekoav:
Add specification for manual STS token assumption with AWS credentials as provided to EKS via the web identity token file when AWS_ROLE_ARN and AWS_WEB_IDENTITY_TOKEN_FILE are present in the environment variables.

Also opened at https://jira.mongodb.org/browse/DRIVERS-1941

### DIFF
--- a/source/auth/auth.rst
+++ b/source/auth/auth.rst
@@ -946,6 +946,16 @@ Environment variables
 _____________________
 AWS Lambda runtimes set several `environment variables <https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime>`_ during initialization. To support AWS Lambda runtimes Drivers MUST check a subset of these variables, i.e., ``AWS_ACCESS_KEY_ID``, ``AWS_SECRET_ACCESS_KEY``, and ``AWS_SESSION_TOKEN``, for the access key ID, secret access key and session token, respectively if AWS credentials are not explicitly provided in the URI. The ``AWS_SESSION_TOKEN`` may or may not be set. However, if ``AWS_SESSION_TOKEN`` is set Drivers MUST use its value as the session token.
 
+EKS Service Account
+___________________
+Pods running in AWS EKS which are configured to `associate an IAM Role to a service account <https://docs.aws.amazon.com/eks/latest/userguide/specify-service-account-role.html>`_ will use a Web Identity to authenticate with AWS. If a username and password are not provided and the aforementioned environment variables are not set, drivers MUST check for the presence of the ``AWS_WEB_IDENTITY_TOKEN_FILE`` and ``AWS_ROLE_ARN`` variables.
+
+If both variables are provided, drivers MUST perform an STS Assume Role with the Web Identity provided by the ``AWS_WEB_IDENTITY_TOKEN_FILE`` and ``AWS_ROLE_ARN`` values. If the ``AWS_DEFAULT_REGION`` environment variable is present, drivers MUST use it to configure the region to use for STS. If ``AWS_DEFAULT_REGION`` is not present, the global STS endpoint MUST be used.
+
+When performing the Assume Role, drivers SHOULD enforce a 10 second timeout while waiting for STS token assumption. If temporary credentials cannot be obtained then drivers MUST fail authentication and raise an error.
+
+On successful token assumption, the resulting credential values ``AccessKeyID``, ``SecretAccessKey`` and ``SessionToken`` contain the credentials to be used.
+
 ECS endpoint
 ____________
 If a username and password are not provided and the aforementioned enviornment variables are not set, drivers MUST query a link-local AWS address for temporary credentials.


source/auth/auth.rst
```diff
@@ -946,6 +946,16 @@ Environment variables
 _____________________
 AWS Lambda runtimes set several `environment variables <https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime>`_ during initialization. To support AWS Lambda runtimes Drivers MUST check a subset of these variables, i.e., ``AWS_ACCESS_KEY_ID``, ``AWS_SECRET_ACCESS_KEY``, and ``AWS_SESSION_TOKEN``, for the access key ID, secret access key and session token, respectively if AWS credentials are not explicitly provided in the URI. The ``AWS_SESSION_TOKEN`` may or may not be set. However, if ``AWS_SESSION_TOKEN`` is set Drivers MUST use its value as the session token.
 
+EKS Service Account
```
kevinAlbs, line 949: IIUC the method of assuming credentials from `AWS_WEB_IDENTITY_TOKEN_FILE` is not specific to EKS.

If that is correct, I suggest renaming this to not be specific to EKS. The suggested change [matches boto3](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#assume-role-with-web-identity-provider).

```suggestion
Assume Role With Web Identity Provider
```

source/auth/auth.rst
```diff
@@ -946,6 +946,16 @@ Environment variables
...
+Pods running in AWS EKS which are configured to `associate an IAM Role to a service account <https://docs.aws.amazon.com/eks/latest/userguide/specify-service-account-role.html>`_ will use a Web Identity to authenticate with AWS. If a username and password are not provided and the aforementioned environment variables are not set, drivers MUST check for the presence of the ``AWS_WEB_IDENTITY_TOKEN_FILE`` and ``AWS_ROLE_ARN`` variables.
+
+If both variables are provided, drivers MUST perform an STS Assume Role with the Web Identity provided by the ``AWS_WEB_IDENTITY_TOKEN_FILE`` and ``AWS_ROLE_ARN`` values. If the ``AWS_DEFAULT_REGION`` environment variable is present, drivers MUST use it to configure the region to use for STS. If ``AWS_DEFAULT_REGION`` is not present, the global STS endpoint MUST be used.
+
+When performing the Assume Role, drivers SHOULD enforce a 10 second timeout while waiting for STS token assumption. If temporary credentials cannot be obtained then drivers MUST fail authentication and raise an error.
+
+On successful token assumption, the resulting credential values ``AccessKeyID``, ``SecretAccessKey`` and ``SessionToken`` contain the credentials to be used.
+
```
kevinAlbs, line 958: Since the motivation for the change is to test EKS, please add a test case is necessary in https://github.com/mongodb/specifications/blob/master/source/auth/tests/mongodb-aws.rst to test with AWS EKS.

source/auth/auth.rst
```diff
@@ -946,6 +946,16 @@ Environment variables
...
 
+EKS Service Account
+___________________
+Pods running in AWS EKS which are configured to `associate an IAM Role to a service account <https://docs.aws.amazon.com/eks/latest/userguide/specify-service-account-role.html>`_ will use a Web Identity to authenticate with AWS. If a username and password are not provided and the aforementioned environment variables are not set, drivers MUST check for the presence of the ``AWS_WEB_IDENTITY_TOKEN_FILE`` and ``AWS_ROLE_ARN`` variables.
+
+If both variables are provided, drivers MUST perform an STS Assume Role with the Web Identity provided by the ``AWS_WEB_IDENTITY_TOKEN_FILE`` and ``AWS_ROLE_ARN`` values. If the ``AWS_DEFAULT_REGION`` environment variable is present, drivers MUST use it to configure the region to use for STS. If ``AWS_DEFAULT_REGION`` is not present, the global STS endpoint MUST be used.
```
kevinAlbs, line 953: I suggest showing the full STS endpoint here, similar to the `EC2 endpoint`. That will help drivers that do not use the AWS SDK.

source/auth/auth.rst
```diff
@@ -946,6 +946,16 @@ Environment variables
 _____________________
 AWS Lambda runtimes set several `environment variables <https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime>`_ during initialization. To support AWS Lambda runtimes Drivers MUST check a subset of these variables, i.e., ``AWS_ACCESS_KEY_ID``, ``AWS_SECRET_ACCESS_KEY``, and ``AWS_SESSION_TOKEN``, for the access key ID, secret access key and session token, respectively if AWS credentials are not explicitly provided in the URI. The ``AWS_SESSION_TOKEN`` may or may not be set. However, if ``AWS_SESSION_TOKEN`` is set Drivers MUST use its value as the session token.
```
kevinAlbs, line 948: Please update the `The order in which Drivers MUST search for credentials is` section above to include the new method of obtaining credentials.

kevinAlbs (commented):
Thank you for the PR @kekoav. I have a few initial questions and suggestions.

blink1073:
I've updated the title to point to DRIVERS-1746, which is the ticket we're now using to track this feature.

blink1073:
This is now superseded by https://github.com/mongodb/specifications/pull/1299#pullrequestreview-1095341339.

PR #1076, opened 2021-10-05
DRIVERS-1918 Clean up read preference spec for mongos/load balancers

abr-egn:
DRIVERS-1918

This updates the "Passing read preference to mongos and load balancers" section to describe what should happen for OP_MSG now that OP_QUERY doesn't need to be supported.


source/server-selection/server-selection.rst
```diff
@@ -579,55 +579,13 @@ Passing read preference to mongos and load balancers
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 If a server of type Mongos or LoadBalancer is selected for a read operation, the read
-preference is passed to the selected mongos through the use of the
-``SecondaryOk`` wire protocol flag, the ``$readPreference`` query
-modifier or both, according to the following rules.
```
jmikola, line 584: Do I assume correctly that we're clear to remove all references to OP_QUERY in light of [DRIVERS-1315](https://jira.mongodb.org/browse/DRIVERS-1315) (dropping support for pre-3.6 servers)? I asked this in [DRIVERS-1918](https://jira.mongodb.org/browse/DRIVERS-1918) but I don't believe it was discussed in the comments.

If the intention is truly to strip all references to OP_QUERY, I think some other references to "SecondaryOk" need to be removed as well:

 * In [Topology Type: Single](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#topology-type-single) we still have the following bullet: "For all other types, using OP_QUERY..."
 * In [Topology types: ReplicaSetWithPrimary or ReplicaSetNoPrimary](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#topology-types-replicasetwithprimary-or-replicasetnoprimary), I recently added clarification for OP_MSG in [5f8f668](https://github.com/mongodb/specifications/commit/5f8f6689538a39a36186285e747d3ac2fbcb27d1#diff-894b3661e1f3249c636d8de2da83bca4341783fe4ec6c90d39338b4d6d4c5dedR1006) for [DRIVERS-823](https://jira.mongodb.org/browse/DRIVERS-823). That paragraph may need to be modified to remove mention of OP_QUERY.

Beyond those references, we also have the following mentions of OP_QUERY:

 * [Scope and General Requirements](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#scope-and-general-requirements)
 * [Terms](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#terms) for "Command" and "Query"

source/server-selection/server-selection.rst
```diff
@@ -579,55 +579,13 @@ Passing read preference to mongos and load balancers
...
-        },
-        $readPreference: {
-            mode: 'secondary',
-            tags: [ { 'dc': 'ny' } ],
-            maxStalenessSeconds: 120,
-            hedge: { enabled: true }
-        }
-    }
+preference is passed to the selected mongos through the use of the ``$readPreference``
+`Global Command Argument`_ according to the following rules:
```
jmikola, line 583: Note that [DRIVERS-1918](https://jira.mongodb.org/browse/DRIVERS-1918) was _not_ focused specifically on mongos and load balancers (see: [this comment](https://jira.mongodb.org/browse/DRIVERS-1918?focusedCommentId=4087768&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4087768)). [DRIVERS-1642](https://jira.mongodb.org/browse/DRIVERS-1642) was the original issue pertaining to mongos and it was later marked as a duplicate of DRIVERS-1918.

I'm in favor of updating this section, but I think other references to `SecondaryOk` and `$readPreference` will also need to be updated to ensure _all_ topologies are covered.

Also, since [Passing read preference to mongos and load balancers](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#passing-read-preference-to-mongos-and-load-balancers) is a separate section under [Read Preference](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#read-preference), I think we should also include a reference to it from [Topology type: LoadBalanced](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#topology-type-loadbalanced), as is already done for [Topology Type: Sharded](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#topology-type-sharded).

source/server-selection/server-selection.rst
```diff
@@ -1789,10 +1747,13 @@ References
 .. _Driver Authentication: https://github.com/mongodb/specifications/blob/master/source/auth
 .. _maxConnecting: /source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#connection-pool
 .. _Connection Monitoring and Pooling: /source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+.. _Global Command Argument: /source/message/OP_MSG.rst#global-command-arguments
 
 Changes
 =======
 
+2021-08-05: Updated $readPreference logic to describe OP_MSG behavior.
```
jmikola, line 1755: If we do scrub this spec of SecondaryOk and OP_QUERY mentions, I think that should also be captured in this changelog entry.

source/server-selection/server-selection.rst
jyemin: It seems like the section name is incorrect now, and it should just be `Passing read preference to MongoDB servers` or something like that.

source/server-selection/server-selection.rst
```diff
@@ -579,14 +579,68 @@ Passing read preference to mongos and load balancers
...
+OP_QUERY only, the ``SecondaryOk`` wire protocol flag, according to the following rules.
+
+For OP_MSG:
 
   - For mode 'primary', drivers MUST NOT set ``$readPreference``
 
   - For modes 'secondary', 'primaryPreferred', 'secondaryPreferred', and 'nearest', drivers
     MUST set ``$readPreference``
 
+For OP_QUERY:
```
jmikola, line 593: For readability, I'd suggest making this and "OP_MSG" above sub-headings. I think backticks would be the next level down from tildes. While this isn't really a problem for OP_MSG, which only has a list, the OP_QUERY section has quite a bit more content that a heading would help set apart.

source/server-selection/server-selection.rst
jmikola: I'm content with leaving OP_QUERY as-is and limiting this PR to just discuss OP_MSG handling.

Just to confirm, you only needed to update the documentation for mongos and LB in this PR because of the changes I made in https://github.com/mongodb/specifications/commit/5f8f6689538a39a36186285e747d3ac2fbcb27d1#diff-894b3661e1f3249c636d8de2da83bca4341783fe4ec6c90d39338b4d6d4c5dedR1006 for replica sets?

For single topologies, we still have the following bullet:

> For all other types, using OP_MSG: If no read preference is configured by the application, or if the application read preference is Primary, then $readPreference MUST be set to { "mode": "primaryPreferred" } to ensure that any server type can handle the request.

"other types" here basically refers to replica set members (since all other types are covered in preceding bullets). I'm curious if that's still insufficient, since it's only referring to "no read preference" or "primary". If _another_ read preference is configured by the application, should we be sending it as-is?

If we consider the bullet for OP_QUERY (where a read preference is never forwarded to a replica set member directly), the spec instructs drivers to always set the `SecondaryOk` wire protocol flag to ensure _any_ server can answer the request. If we follow that logic, then maybe OP_MSG should _always_ send "primaryPreferred" when directly connected to a replica set member.

I'm unsure about this myself, so it'd be worth having someone from @mongodb/dbx-spec-owners-server-selection chime in here.

patrickfreed: >If another read preference is configured by the application, should we be sending it as-is?

The SDAM spec seems to indicate that we should, but it also confusingly says that the `secondaryOk` bit won't be set as its reasoning (from the "Direction Connections" portion of the [General Requirements](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#general-requirements) section):

>Setting a read preference MUST NOT be necessary to connect to a secondary. Of course, the secondary will reject all operations done with the PRIMARY read preference because the secondaryOk bit is not set, but the initial connection itself succeeds.

I think it would probably make more sense if we just disallowed `directConnection` with `readPreference` to make it clear that the latter will be ignored, but that might be out of scope for this PR.

source/server-selection/server-selection.rst
```diff
@@ -579,9 +579,18 @@ Passing read preference to mongos and load balancers
...
-preference is passed to the selected mongos through the use of the
-``SecondaryOk`` wire protocol flag, the ``$readPreference`` query
-modifier or both, according to the following rules.
+preference is passed to the selected mongos through the use of ``$readPreference``
+(as a `Global Command Argument`_ for OP_MSG or a query modifier for OP_QUERY) and, for
+OP_QUERY only, the ``SecondaryOk`` wire protocol flag, according to the following rules.
+
+For OP_MSG:
+
+  - For mode 'primary', drivers MUST NOT set ``$readPreference``
```
patrickfreed, line 588: The rendered version of this has these bullets inside of a blockquote block (ditto for the OP_QUERY part). I think they need one more space of indentation or two fewer, I can't really remember with RST.

source/server-selection/server-selection.rst
```diff
@@ -962,7 +974,10 @@ The single server is always suitable for write operations if it is available.
 Topology type: LoadBalanced
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-See the `Load Balancer Specification <../load-balancers/load-balancers.rst#server-selection>`__ for details.
+During command construction, the LoadBalancer server MUST be treated like a
+mongos and drivers MUST add a $readPreference field to the command when
```
patrickfreed, line 978: Is the bit about treating the load balanced server like a mongos required? Or can we just say the "drivers MUST add..." bit?

source/server-selection/server-selection.rst
abr-egn: Fixed.

source/server-selection/server-selection.rst
```diff
@@ -579,9 +579,20 @@ Passing read preference to mongos and load balancers
...
+(as a `Global Command Argument`_ for OP_MSG or a query modifier for OP_QUERY) and, for
+OP_QUERY only, the ``SecondaryOk`` wire protocol flag, according to the following rules.
+
+For OP_MSG:
+```````````
+
+- For mode 'primary', drivers MUST NOT set ``$readPreference``
+
+- For modes 'secondary', 'primaryPreferred', 'secondaryPreferred', and 'nearest', drivers
+  MUST set ``$readPreference``
```
p-mongo, line 592: If this section is meant to cover all non-primary read preferences, for clarity I would say something like "For all other read preference modes (e.g. secondary, primaryPreferred, ...)" so that it is clear that there aren't any other read preference modes that the driver needs to consider.

source/server-selection/server-selection.rst
jmikola: > I've updated this to express that the read preference should be sent as-is otherwise.

That makes sense to me. I think the existing direction for sending `primaryPreferred` by default (or instead of an explicit "primary") is done for ease of use. Since "primary" is the default, applications making a direct connection to a secondary might otherwise need to explicitly specify a non-primary read preference.

> I think it would probably make more sense if we just disallowed directConnection with readPreference to make it clear that the latter will be ignored, but that might be out of scope for this PR.

I dunno if raising a client-side error when mixing `directConnection` and `readPreference` would be desirable (even if it logically makes sense), since it could break existing applications. And that would require drivers to differentiate between an explicit "primary" read preference or the default. I reckon not all drivers differentiate between explicit and default "primary" today.

source/server-selection/server-selection.rst
abr-egn: Updated.

source/server-selection/server-selection.rst
patrickfreed: Yeah, a hard error would be too heavy handed probably, but I think a log warning could be useful. This isn't really a huge concern though, since somone using `directConnection` doesn't really need a read preference anyways, so they're probably happy with it being ignored.

patrickfreed (approved):

PR #1077, opened 2021-10-06
Skip "listIndexes pins the cursor to a connection" on serverless

jyemin:
Server team is investigating a bug which this test appears to have uncovered
when running in serverless environment.  While it's being investigated,
skipping this test on serverless.

DRIVERS-1882


kevinAlbs (approved):
Thank you!

PR #1078, opened 2021-10-13
Add a CMAP test that verifies the background thread hands over connections to threads doing checkout

stIncMale:
Added @ShaneHarvey as a reviewer because introducing this test was his idea.

DRIVERS-1948

### DIFF
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -526,7 +526,7 @@ thread, the checkOut method is responsible for ensuring this requirement is met.
 When populating the Pool, pendingConnectionCount has to be decremented after
 establishing a `Connection`_ similarly to how it is done in
 `Checking Out a Connection <#checking-out-a-connection>`_ to signal that
-another `Connection`_ is allowed to be established. Such a signal must become
+another `Connection`_ is allowed to be established. Such a signal MUST become
 observable to any `Thread`_ after the action that
 `marks the established Connection as "available" <#marking-a-connection-as-available-internal-implementation>`_
 becomes observable to the `Thread`_.
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -527,8 +527,9 @@ When populating the Pool, pendingConnectionCount has to be decremented after
 establishing a `Connection`_ similarly to how it is done in
 `Checking Out a Connection <#checking-out-a-connection>`_ to signal that
 another `Connection`_ is allowed to be established. Such a signal must become
-observable to any `Thread`_ after the state of the established `Connection`_
-becomes observable to the `Thread`_ as anything but "pending".
+observable to any `Thread`_ after the action that
+`marks the established Connection as "available" <#marking-a-connection-as-available-internal-implementation>`_
+becomes observable to the `Thread`_.
 Informally, this order guarantees that no `Thread`_ tries to start
 establishing a `Connection`_ when there is an "available" `Connection`_
 established as a result of populating the Pool.
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -529,6 +529,9 @@ establishing a connection similarly to how it is done in
 another `Connection`_ is allowed to be established. Such a signal must become
 observable to any `Thread`_ after the state of the established `Connection`_
 becomes observable to the `Thread`_ as anything but "pending".
+Informally, this order guarantees that no `Thread`_ tries to start
+establishing a `Connection`_ when there is an "available" `Connection`_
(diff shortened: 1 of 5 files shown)


source/connection-monitoring-and-pooling/tests/pool-checkout-minSize-connection-maxConnecting.yml
```diff
@@ -0,0 +1,63 @@
...
+    failCommands: ["isMaster","hello"]
+    closeConnection: false
+    blockConnection: true
+    blockTimeMS: 500
+poolOptions:
+  # allows both thread1 and the background thread to start opening connections concurrently
+  minPoolSize: 2
+  # gives opportunity for the checkout in thread2 to open a new connection, which it must not do nonetheless
+  maxPoolSize: 3
+  waitQueueTimeoutMS: 5000
```
ShaneHarvey, line 21: Do we need backgroundThreadIntervalMS here?

stIncMale:
> This test fails in Java before the fix in JAVA-4316?

Yes, it fails without the changes in https://github.com/mongodb/mongo-java-driver/pull/799, which encompass both JAVA-4316 and JAVA-4316: _"expected:<class `com.mongodb.event.ConnectionCheckedOutEvent`> but was:<class `com.mongodb.event.ConnectionCreatedEvent`>"_.

source/connection-monitoring-and-pooling/tests/pool-checkout-minSize-connection-maxConnecting.yml
stIncMale: No, all we need is for the BG thread to start a single [run](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#background-thread), which it does without artificial delays, and then we do not care when the next run starts.

source/connection-monitoring-and-pooling/tests/pool-checkout-minSize-connection-maxConnecting.yml
```diff
@@ -0,0 +1,63 @@
+version: 1
+style: integration
+description: threads blocked by maxConnecting check out minSize connections
```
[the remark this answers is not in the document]
stIncMale: Done

source/connection-monitoring-and-pooling/tests/pool-checkout-minSize-connection-maxConnecting.yml
stIncMale: It is not explicitly specified that the background thread should be started ASAP ("immediately" is not an appropriate term here), but the specification of `minPoolSize` (which is unachievable, and, therefore, should have been formulated differently) implies this: [_"The minimum number of Connections that MUST exist at any moment in a single connection pool."_](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#id1)

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -523,6 +523,16 @@ occur only while the pool is "ready". If the pool implements a background
 thread, it can be used for this. If the pool does not implement a background
 thread, the checkOut method is responsible for ensuring this requirement is met.
 
+When populating the Pool, pendingConnectionCount has to be decremented after
+establishing a `Connection`_ similarly to how it is done in
```
ShaneHarvey, line 527: In my mind "after establishing a `Connection`_" does not convey the nuance required here. I think we need to say that pendingConnectionCount must not be decremented until after the established Connection is actually added to the pool.

ShaneHarvey (changes requested):

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
stIncMale: The first sentence is not there to specify a requirement (even though it must be so), but to introduce the notion of a specific signal, for which an ordering constraint is then specified. I updated the ordering constraint such that it refers to the "Marking a Connection as Available" action, which is the wording the specification uses to say "add connection to the pool".

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -523,6 +523,17 @@ occur only while the pool is "ready". If the pool implements a background
...
+`Checking Out a Connection <#checking-out-a-connection>`_ to signal that
+another `Connection`_ is allowed to be established. Such a signal must become
```
ShaneHarvey, line 529: must -> MUST

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
ShaneHarvey: Ah thanks I get it now.

ShaneHarvey (changes requested):
One minor nit otherwise LGTM

ShaneHarvey (approved):
LGTM

BorisDog (approved):
LGTM

PR #1083, opened 2021-10-20
DRIVERS-1959: Expect either valid:false or warning:true, but not both

jmikola:
https://jira.mongodb.org/browse/DRIVERS-1959

SRV option tests will either raise errors (e.g. using options with non-SRV scheme) or warnings (e.g. invalid value/type), but not both. This corrects a mistake originally introduced in b508f6d2a1819882d40c4e7e73b95fc8ae3bdfe5 (#1069).

### DIFF
--- a/source/uri-options/tests/README.rst
+++ b/source/uri-options/tests/README.rst
@@ -38,6 +38,11 @@ The ``valid`` and ``warning`` fields are boolean in order to keep the tests
 flexible. We are not concerned with asserting the format of specific error or
 warnings messages strings.
 
+Under normal circumstances, it should not be necessary to specify both
+``valid: false`` and ``warning: true``. Typically, a URI test case will either
+yield an error (e.g. options conflict) or a warning (e.g. invalid type or value
+for an option), but not both.
+
 Use as unit tests
 =================
 
--- a/source/uri-options/tests/srv-options.yml
+++ b/source/uri-options/tests/srv-options.yml
@@ -10,7 +10,7 @@ tests:
     - description: "Non-SRV URI with custom srvServiceName"
       uri: "mongodb://example.com/?srvServiceName=customname"
       valid: false
-      warning: true
+      warning: false
       hosts: ~
       auth: ~
       options: {}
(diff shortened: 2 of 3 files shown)


benjirewis (approved):
LGTM. Thanks for fixing the custom service name test, too.

JamesKovacs:
Might be worth adding some verbiage to `uri-options/tests/README.rst` to note this constraint.

source/uri-options/tests/README.rst
```diff
@@ -38,6 +38,11 @@ The ``valid`` and ``warning`` fields are boolean in order to keep the tests
 flexible. We are not concerned with asserting the format of specific error or
 warnings messages strings.
 
+Under normal circumstances, it should not be necessary to specify both
+``valid: false`` and ``warning: true``. Typically, a URI option will either
+yield an error (e.g. usage conflict) or a warning (e.g. invalid type or value),
+but not both.
```
jmikola, line 44: @JamesKovacs: Added.

source/uri-options/tests/srv-options.yml
```diff
@@ -24,39 +24,39 @@ tests:
           srvMaxHosts: 2
     - description: "SRV URI with negative integer for srvMaxHosts"
       uri: "mongodb+srv://test1.test.build.10gen.cc/?srvMaxHosts=-1"
-      valid: false
+      valid: true
```
jyemin, line 27: I was expecting:

```
valid: false
warning: false
```

source/uri-options/tests/srv-options.yml
```diff
@@ -24,39 +24,39 @@ tests:
...
       warning: true
       hosts: ~
       auth: ~
       options: {}
     - description: "SRV URI with invalid type for srvMaxHosts"
       uri: "mongodb+srv://test1.test.build.10gen.cc/?srvMaxHosts=foo"
-      valid: false
```
jyemin, line 34: I was expecting:

```
valid: false
warning: false
```

jmikola: Same as above. "foo" is an invalid value for `srvMaxHosts`, since it is documented as expecting a non-negative integer.

source/uri-options/tests/srv-options.yml
jmikola: This seems consistent with some other tests for invalid values, such as [`tlsAllowInvalidCertificates`](https://github.com/mongodb/specifications/blob/master/source/uri-options/tests/tls-options.yml#L40).

PR #1086, opened 2021-11-01
DRIVERS-1385 Always specify error labels at the top level

isabelatkinson:
Updates spec language and spec tests always to specify error labels at the top level. This behavior is verified in the comments of DRIVERS-1385.

### DIFF
--- a/source/retryable-writes/retryable-writes.rst
+++ b/source/retryable-writes/retryable-writes.rst
@@ -3,14 +3,14 @@ Retryable Writes
 ================
 
 :Spec Title: Retryable Writes
-:Spec Version: 1.6.1
+:Spec Version: 1.6.2
 :Author: Jeremy Mikola
 :Lead: \A. Jesse Jiryu Davis
 :Advisors: Robert Stam, Esha Maharishi, Samantha Ritter, and Kaloian Manassiev
 :Status: Accepted
 :Type: Standards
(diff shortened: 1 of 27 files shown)


jmikola (approved):
I read through the comments on [DRIVERS-1385](https://jira.mongodb.org/browse/DRIVERS-1385) to refresh myself on this ticket. I also noted the conclusion that no server versions ever returned error labels under the `writeConcernError` field. LGTM with requested changelog/version edits.

source/retryable-writes/retryable-writes.rst
```diff
@@ -3,14 +3,14 @@ Retryable Writes
 ================
 
 :Spec Title: Retryable Writes
-:Spec Version: 1.6.1
+:Spec Version: 1.6.2
```
jmikola, line 6: This is a patch version bump (typically used for grammar/typo fixes). I was suggesting a minor version bump (i.e. 1.7.0) since this is technically changing driver behavior.

If other folks feel this change is sufficiently insignificant I could go either way, though.

isabelatkinson: Ah whoops I meant to do a minor version bump, I'll update. I agree that a minor version bump is the right way to go.

isabelatkinson:
@durran do you want to review this as well?

PR #1093, opened 2021-11-04
Remove KMS TLS Tests

kevinAlbs:
Not urgent.

**Background & Motivation**
The `KMS TLS Options Tests` was introduced in DRIVERS-1353 to test TLS options for all KMS providers. The `KMS TLS Options Tests` include connecting to a mock server with an expired certificate and an invalid hostname certificate.

**Scope**
- Removes `KMS TLS Tests` since the `KMS TLS Options` tests test that case.

### DIFF
--- a/source/client-side-encryption/tests/README.rst
+++ b/source/client-side-encryption/tests/README.rst
@@ -1088,82 +1088,14 @@ Drivers that do not support an unlimited maximum pool size MUST skip this test.
       - a find on "keyvault".
 - ExpectedNumberOfClients: 1
 
-KMS TLS Tests
-~~~~~~~~~~~~~
+KMS TLS Options Tests
+~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


comandeo (approved):

ShaneHarvey:
I don't think we can remove these test yet. The issue is that these tests exercise the behavior of an empty kmsTlsOptions map (eg using system default certs). The new tests only cover the behavior of explicitly specifying kmsTlsOptions. If we want to remove these tests then I think we need to add a 5th ClientEncryption object in "KMS TLS Options Tests" to test the behavior with the default TLS settings.

kevinAlbs:
> I don't think we can remove these test yet. The issue is that these tests exercise the behavior of an empty kmsTlsOptions map (eg using system default certs). The new tests only cover the behavior of explicitly specifying kmsTlsOptions. If we want to remove these tests then I think we need to add a 5th ClientEncryption object in "KMS TLS Options Tests" to test the behavior with the default TLS settings.

Good catch. That would require driver changes. Filed DRIVERS-1986. Closing.

PR #1095, opened 2021-11-07
DRIVERS-1969: Revise rules for $out/$merge read preferences

jmikola:
https://jira.mongodb.org/browse/DRIVERS-1969

Only fall back to a primary read preference if there is evidence that the driver is connected to a pre-5.0 server.

### DIFF
--- a/source/crud/crud.rst
+++ b/source/crud/crud.rst
@@ -2026,54 +2026,37 @@ in the collection prior to ``$merge`` being executed).
 Read preferences and server selection
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+This section is only applicable if an explicit (i.e. per-operation) or inherited
+(e.g. from a Collection) read preference is available and it is *not* a primary
+read preference (i.e. ``{ "mode": "primary" }``).
+
 Historically, only primaries could execute an aggregation pipeline with ``$out``
-or ``$merge``. As of ``featureCompatibilityVersion`` 4.4, secondaries can now
-execute such pipelines; however, since drivers do not track
-``featureCompatibilityVersion``, executing an aggregation pipeline with ``$out``
-or ``$merge`` on a secondary will require a 5.0+ server (i.e. wire version >=
-13).
+or ``$merge`` and drivers never considered a read preference for the operation.
+As of ``featureCompatibilityVersion`` 4.4, secondaries can now execute pipelines
+with ``$out`` or ``$merge``. Since drivers do not track
+``featureCompatibilityVersion``, the decision to consider a read preference for
+such a pipeline will depend on the wire version(s) of the server(s) to which the
+driver is connected.
 
-Drivers MUST guarantee the following:
+If there are one or more available servers and one or more of those servers is
+pre-5.0 (i.e. wire version < 13), drivers MUST NOT use the available read
+preference and MUST instead select a server using a primary read preference.
+Otherwise, if there are either no available servers or all available servers are
+5.0+ (i.e. wire version >= 13), drivers MUST use the available read preference.
 
-- If an explicit (i.e. per-operation) read preference is specified for an
-  aggregation with a write stage, drivers MUST attempt to use it. If that would
-  result in a pre-5.0, secondary server being selected, drivers MUST instead
-  select a server using a primary read preference.
-- If no explicit read preference is specified but a default read preference is
-  available to inherit (e.g. from the Collection), drivers MUST attempt to use
-  it. If that would result in a pre-5.0, secondary server being selected,
-  drivers MUST instead select a server using a primary read preference.
+Drivers SHOULD augment their
+`server selection algorithm <..../server-selection/server-selection.rst#server-selection-algorithm>`_
+such that this logic can be enforced within a single server selection attempt.
 
-Drivers SHOULD use a custom server selector to consider server/wire version when
-matching a read preference and, if a pre-5.0 secondary would be selected, fall
-back to selecting a primary. With this approach, only a single server selection
-attempt is needed.
+If the driver is connected to a load balancer, it MUST consider the wire version
+of the backing server obtained from the connection handshake.
 
-If it is not possible to augment the server selector and a pre-5.0 secondary is
-selected, drivers MAY invoke server selection a second time to obtain a primary.
-However, if the first attempt results in a server selection timeout, drivers
-MUST NOT make a second attempt and MUST propagate the original timeout error. If
-the topology type is Single, Sharded, or LoadBalanced, drivers MUST NOT make a
-second attempt (there is no benefit in doing so).
-
-Regardless of whether drivers employ a custom server selector or two-attempt
-approach, they MUST discern the *effective* read preference for the operation,
-which SHALL be used for specifying the
+Drivers MUST discern the read preference used to select a server for the
+operation, which SHALL be used for specifying the
 `$readPreference global command argument <../message/OP_MSG.rst#global-command-arguments>`_
 and
 `passing read preference to mongos and load balancers <../server-selection/server-selection.rst#passing-read-preference-to-mongos-and-load-balancers>`_
-(if applicable). The effective read preference SHALL be discerned as follows:
-
-- In the absence of an explicit or inherited read preference, the effective read
-  preference is ``{ "mode": "primary" }``.
-- If an explicit or inherited read preference results in selection of an
-  ineligible server and selection falls back to a primary, the effective read
-  preference is ``{ "mode": "primary" }``.
-- If a pre-5.0 mongos or a load balancer backing a pre-5.0 mongos is selected,
-  the effective read preference is ``{ "mode": "primary" }``.
-- If an explicit or inherited read preference is used and results in selection
-  of an eligible server (excluding a pre-5.0 mongos as discussed above), that is
-  the effective read preference.
+(if applicable).
 
 
 Test Plan
@@ -2173,6 +2156,7 @@ Q: Why are client-side errors raised for some unsupported options?
 Changes
 =======
 
+* 2021-11-10: Revise rules for applying read preference for aggregations with $out and $merge.
 * 2021-11-10: Add let to FindOptions, UpdateOptions, DeleteOptions, FindOneAndDeleteOptions, FindOneAndReplaceOptions, FindOneAndUpdateOptions
 * 2021-09-28: Support aggregations with $out and $merge on 5.0+ secondaries
 * 2021-08-31: Allow unacknowledged hints on write operations if supported by server (reverts previous change).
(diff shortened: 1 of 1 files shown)


source/crud/crud.rst
```diff
@@ -1953,54 +1953,37 @@ in the collection prior to ``$merge`` being executed).
...
-  preference is ``{ "mode": "primary" }``.
-- If an explicit or inherited read preference results in selection of an
-  ineligible server and selection falls back to a primary, the effective read
-  preference is ``{ "mode": "primary" }``.
-- If a pre-5.0 mongos or a load balancer backing a pre-5.0 mongos is selected,
-  the effective read preference is ``{ "mode": "primary" }``.
-- If an explicit or inherited read preference is used and results in selection
-  of an eligible server (excluding a pre-5.0 mongos as discussed above), that is
-  the effective read preference.
+(if applicable).
```
jmikola, line 1986: I don't think the original list is necessary. The first sentence in this paragraph is clear that the read preference used to select a server (either the available RP or the primary RP we fall back to) is what the driver must pass via `$readPreference`.

source/crud/crud.rst
```diff
@@ -1953,54 +1953,37 @@ in the collection prior to ``$merge`` being executed).
...
+preference and MUST instead select a server using a primary read preference.
+Otherwise, if there are either no available servers or all available servers are
+5.0+ (i.e. wire version >= 13), drivers MUST use the available read preference.
+
+Drivers SHOULD augment their
+`server selection algorithm <..../server-selection/server-selection.rst#server-selection-algorithm>`_
+such that this logic can be enforced within a single server selection attempt.
+
+If the driver is connected to a load balancer, it MUST consider the wire version
+of the backing server obtained from the connection handshake.
```
jmikola, line 1979: This paragraph replaces a previous bullet in the removed list below to clarify how drivers should make their decision when connected to a load balancer. I noted that the [load balancer](https://github.com/mongodb/specifications/blob/master/source/load-balancers/load-balancers.rst) spec doesn't really instruct drivers on how to perform wire version checks, but consulting the handshake `hello` response seems like a general rule we'd want to follow.

source/crud/crud.rst
```diff
@@ -1953,54 +1953,37 @@ in the collection prior to ``$merge`` being executed).
...
+
+If there are one or more available servers and one or more of those servers is
+pre-5.0 (i.e. wire version < 13), drivers MUST NOT use the available read
+preference and MUST instead select a server using a primary read preference.
+Otherwise, if there are either no available servers or all available servers are
+5.0+ (i.e. wire version >= 13), drivers MUST use the available read preference.
+
+Drivers SHOULD augment their
+`server selection algorithm <..../server-selection/server-selection.rst#server-selection-algorithm>`_
+such that this logic can be enforced within a single server selection attempt.
```
jmikola, line 1976: I'm no longer going into detail about the two-attempt approach, since it may only apply to PHP and the the original text conflicted with the new rule to consider _all_ servers to which the driver is connected (to determine any evidence of a connection to a pre-5.0 server).

In the previous text, the two-attempt approach instructed the driver to fall back to a primary only if the available RP selected a pre-5.0 secondary. With the new rule, I think the two attempt approach becomes significantly more complicated:

 * Attempt to select a server using the available RP. If this is the first operation executed by the driver, this first server selection attempt will initialize SDAM and allow us to check all available servers in the next step.
 * Regardless of whether server selection succeeds or fails, check the wire version of all available servers
 * If there is no evidence of a pre-5.0 server, either use the server we selected or propagate the server selection failure.
 * If there is any evidence of a pre-5.0 server, note that `{mode: "primary"}` is the effective read preference for this operation (i.e. we must "fall back"). If we already happened to select a primary in the first attempt, we can use it without making a second selection attempt. Otherwise, we must invoke server selection again to select a primary.

One notable difference here is that we can no longer abort if the first attempt raises a server selection failure. This change in behavior should also address the BC concern I raised in [DRIVERS-1969](https://jira.mongodb.org/browse/DRIVERS-1969?focusedCommentId=4161573&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4161573):

> Row 6 (P4, S-none): P4 will be selected for most RPs; however, RP(secondary) will result in a server selection failure because a pre-5.0 secondary is never selected to warrant fall back behavior. This is potentially a behavioral BC break

source/crud/crud.rst
```diff
@@ -1953,54 +1953,37 @@ in the collection prior to ``$merge`` being executed).
 Read preferences and server selection
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
+This section is only applicable if an explicit (i.e. per-operation) or inherited
+(e.g. from a Collection) read preference is available and it is *not* a primary
+read preference (i.e. ``{ "mode": "primary" }``).
```
jmikola, line 1958: Adding this paragraph up top allows us to briefly address the concept of an explicit or inherited read preference. From this point on, we don't actually care whether the RP is explicit/inherited -- there's either a RP available for this operation or there isn't (i.e. default to primary).

I've also noted that explicit RPs aren't discussed elsewhere in the spec apart from being alluded to in "Q: Where is read preference?", so removing repetitive mentions of that concept from this section seems like an improvement.

source/crud/crud.rst
```diff
@@ -1953,54 +1953,37 @@ in the collection prior to ``$merge`` being executed).
...
+with ``$out`` or ``$merge``. Since drivers do not track
+``featureCompatibilityVersion``, the decision to consider a read preference for
+such a pipeline will depend on the wire version(s) of the server(s) to which the
+driver is connected.
+
+If there are one or more available servers and one or more of those servers is
+pre-5.0 (i.e. wire version < 13), drivers MUST NOT use the available read
+preference and MUST instead select a server using a primary read preference.
+Otherwise, if there are either no available servers or all available servers are
+5.0+ (i.e. wire version >= 13), drivers MUST use the available read preference.
```
jmikola, line 1972: @rstam originally proposed:

> If there are any available servers and all available servers are 5.0+ apply the read preference; otherwise, fallback to primary

This is the text that @jyemin and I devised after discussing over Zoom. I don't believe there's any functional difference, as the edge case if when there are no available servers with which to make a decision and in that case the user is going to encounter a server selection failure regardless.

In contrast to the original text, these new rules do mean that there is now no change that a capable 5.0+ secondary might be selected while the driver is connected to a mixed version replica set. I personally didn't see the issue with the previous behavior but I won't object if this is what others are asking for.

The only concern I had with the original text was the BC issue when the driver was connected to a single pre-5.0 primary. With the original text, the user would always encounter a server selection failure if the operation was provided RP(secondary).

source/crud/crud.rst
```diff
@@ -1953,54 +1953,37 @@ in the collection prior to ``$merge`` being executed).
...
+
+Drivers SHOULD augment their
+`server selection algorithm <..../server-selection/server-selection.rst#server-selection-algorithm>`_
+such that this logic can be enforced within a single server selection attempt.
+
+If the driver is connected to a load balancer, it MUST consider the wire version
```
jyemin, line 1978: I don't think this is feasible.  Given the current CMAP spec, there is no guarantee that the driver will have a connection the LB available at the time when server selection is performed.  I think instead we should say that a driver can assume that any server behind an LB supports read preference for $out/$merge.

jyemin (changes requested):

source/crud/crud.rst
jmikola: This may overlap with a question @comandeo asked in Slack, to which I replied:

> Assuming the LB backs a 5.0+ server certainly works for this ticket, where you just need to determine if the server is pre- or post-5.0; however, you'll inevitably run into problems down the line if you need to do wire version checks for future features (e.g. something requires 5.2+). At some point, you will need to start making decisions on wire version reported in the handshake response.

Drivers are going to need to address this at some point. I don't think "defy augury" is a suitable excuse to avoid thinking about this.

jyemin: > Drivers are going to need to address this at some point. I don't think "defy augury" is a suitable excuse to avoid thinking about this.

Perhaps, but we don't have to do it right now in this PR.

jyemin: > Doesn't that still create a problem if we ultimately discover the driver is connected to a single 4.4 primary?

I don't think it does.  Since there are no servers available, it doesn't really matter what read preference will be applied, because the list of suitable servers will be empty, and driver will enter the server selection loop.  The next time the driver tries to apply the read preference, there will be a server available and we start the process again.

jmikola: > Perhaps, but we don't have to do it right now in this PR.

Just to confirm, you're OK with just removing this paragraph? If no other specs are talking about wire version checks for load balancers, I think leaving this unsaid is more future-proof.

> it doesn't really matter what read preference will be applied, because the list of suitable servers will be empty, and driver will enter the server selection loop. The next time the driver tries to apply the read preference, there will be a server available and we start the process again.

Understood. My mind is still thinking about the try-once mode for single-threaded selection. I didn't consider that the pre-5.0 server check would happen during _each_ loop iteration.

jyemin: Leaving it unsaid is more future-proof, but driver implementations will still need to special case this, so I think we should give guidance.  Something like:

> If the driver is connected to a load balancer. it MUST treat it the same as a 5.0+ cluster

Something like that.  What do you think?

jmikola: Updated the text to add a LoadBalanced topology to the list of cases where we apply the read preference.

jyemin (approved):
LGTM

source/crud/crud.rst
benjirewis: Did you figure out a way to still use the two-attempt approach in PHP without changes to libmongoc API? Or will libmongoc API also be altered as a result of this spec change?

source/crud/crud.rst
```diff
@@ -2100,6 +2083,7 @@ Q: Why are client-side errors raised for some unsupported options?
 Changes
 =======
 
+* 2021-11-XX: Revise rules for applying read preference for aggregations with $out and $merge.
```
benjirewis, line 2086:
```suggestion
* 2021-11-09: Revise rules for applying read preference for aggregations with $out and $merge.
```

source/crud/crud.rst
```diff
@@ -12,7 +12,7 @@ Driver CRUD API
 :Status: Approved
 :Type: Standards
 :Minimum Server Version: 2.6
-:Last Modified: September 28, 2021
+:Last Modified: November XX, 2021
```
benjirewis, line 15:
```suggestion
:Last Modified: November 9, 2021
```

benjirewis (approved):

source/crud/crud.rst
jmikola: I'd rather create a convoluted approach in PHPLIB than introduce API changes in both libmongoc _and_ PHPC just so PHPLIB can utilize them. My current plan is to attempt server selection with the read preference and then regardless of whether a server is selected or we get an error, I'll check all available servers for pre-5.0 and conditionally select a primary if needed.

PR #1096, opened 2021-11-08
DRIVERS-1353 fix AWS cases KMS TLS Options Tests

kevinAlbs:
Port 8002 should be used for client_encryption_no_tls and client_encryption_with_tls.
Port 8000 should be used for client_encryption_expired.
Port 8001 should be used for client_encryption_invalid_hostname

The AWS test cases require specifying the endpoint through the masterKey document. The C PoC had tested this behavior [here](https://github.com/mongodb/mongo-c-driver/pull/881/files#diff-d363a5539f2fbabfadf2d7cc6cdb96a203fb63f2caa31bae731c3e12d97b204cR2862-R2920), but I wrote the test cases incorrectly.

### DIFF
--- a/source/client-side-encryption/tests/README.rst
+++ b/source/client-side-encryption/tests/README.rst
@@ -1343,19 +1343,43 @@ following masterKey:
 
 Expect an error indicating TLS handshake failed.
 
-Call `client_encryption_with_tls.createDataKey()` with "aws" as the provider and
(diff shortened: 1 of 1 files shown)


DmitryLukyanov (approved):
LGTM

comandeo (approved):

PR #1097, opened 2021-11-08
DRIVERS-1353 client_encryption_no_tls => client_encryption_no_client_cert

kevinAlbs:
"no_tls" is misleading. TLS is still enabled.
tlsClientCertificateKeyFile is not set.

### DIFF
--- a/source/client-side-encryption/tests/README.rst
+++ b/source/client-side-encryption/tests/README.rst
@@ -1211,7 +1211,7 @@ Create the following four ``ClientEncryption`` objects.
 
 Configure each with ``keyVaultNamespace`` set to ``keyvault.datakeys``, and a default MongoClient as the ``keyVaultClient``.
 
-1. Create a ``ClientEncryption`` object named ``client_encryption_no_tls`` with the following KMS providers:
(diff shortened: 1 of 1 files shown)


source/client-side-encryption/tests/README.rst
```diff
@@ -1211,7 +1211,7 @@ Create the following four ``ClientEncryption`` objects.
 
 Configure each with ``keyVaultNamespace`` set to ``keyvault.datakeys``, and a default MongoClient as the ``keyVaultClient``.
 
-1. Create a ``ClientEncryption`` object named ``client_encryption_no_tls`` with the following KMS providers:
+1. Create a ``ClientEncryption`` object named ``client_encryption_no_client_cert`` with the following KMS providers:
```
DmitryLukyanov, line 1214: should we somehow mention this option: `- ``tlsCAFile`` (or equivalent) set to `ca.pem`_. This MAY be configured system-wide.`?

kevinAlbs: Good catch. If `tlsCAFile` is not configured, the error may occur verifying the server certificate, instead of a failure to send a client certificate. I think either error proves that TLS options are not being passed. But having the error caused by no client certificate seems better for consistency.

PR #1098, opened 2021-11-09
DRIVERS-1943 Make maxConnecting configurable

stIncMale:
@BorisDog and @patrickfreed are owners of the CMAP spec, @JamesKovacs is an owner of the URI options spec.

DRIVERS-1943

### DIFF
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -278,7 +278,7 @@ has the following properties:
    -  Connections are not created in the background to satisfy minPoolSize
 
 -  **Capped:** a pool is capped if **maxPoolSize** is set to a non-zero value. If a pool is capped, then its total number of `Connections <#connection>`_ (including available and in use) MUST NOT exceed **maxPoolSize**
--  **Rate-limited:** A Pool MUST limit the number of `Connections <#connection>`_ being `established <#establishing-a-connection-internal-implementation>`_ concurrently.
+-  **Rate-limited:** A Pool MUST limit the number of `Connections <#connection>`_ being `established <#establishing-a-connection-internal-implementation>`_ concurrently via the **maxConnecting** `pool option <#connection-pool-options-1>`_.
 
 
 .. code:: typescript
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -9,8 +9,8 @@ Connection Monitoring and Pooling
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: N/A
-:Last Modified: 2021-06-02
-:Version: 1.5.1
+:Last Modified: 2021-11-08
+:Version: 1.5.2
 
 .. contents::
 
@@ -116,6 +116,15 @@ Drivers that implement a Connection Pool MUST support the following ConnectionPo
        *  Defaults to 0.
        */
       maxIdleTimeMS?: number;
+
+      /**
+       *  The maximum number of Connections a Pool may be establishing concurrently.
+       *  Establishment of a Connection is a part of its life cycle
+       *  starting after a ConnectionCreatedEvent and ending before a ConnectionReadyEvent.
+       *  If specified, MUST be a number > 0.
+       *  Defaults to 2.
+       */
+      maxConnecting?: number;
     }
 
 Additionally, Drivers that implement a Connection Pool MUST support the following ConnectionPoolOptions UNLESS that driver meets ALL of the following conditions:
@@ -269,7 +278,7 @@ has the following properties:
    -  Connections are not created in the background to satisfy minPoolSize
 
 -  **Capped:** a pool is capped if **maxPoolSize** is set to a non-zero value. If a pool is capped, then its total number of `Connections <#connection>`_ (including available and in use) MUST NOT exceed **maxPoolSize**
--  **Rate-limited:** A Pool MUST limit the number of connections being created at a given time to be 2 (maxConnecting). 
+-  **Rate-limited:** A Pool MUST limit the number of `Connections <#connection>`_ being `established <#establishing-a-connection-internal-implementation>`_ concurrently.
 
 
 .. code:: typescript
@@ -1137,6 +1146,8 @@ Change log
 :2021-4-12: Adding in behaviour for load balancer mode.
 
 :2021-06-02: Formalize the behavior of a `Background Thread <#background-thread>`__.
+
+:2021-11-08: Make maxConnecting configurable.
 
 .. Section for links.
 
--- a/source/connection-monitoring-and-pooling/tests/pool-checkout-custom-maxConnecting-is-enforced.json
+++ b/source/connection-monitoring-and-pooling/tests/pool-checkout-custom-maxConnecting-is-enforced.json
@@ -47,10 +47,6 @@
       "count": 1
     },
     {
-      "name": "wait",
-      "ms": 100
-    },
-    {
       "name": "checkOut",
       "thread": "thread2"
     },
--- a/source/connection-monitoring-and-pooling/tests/pool-checkout-custom-maxConnecting-is-enforced.json
+++ b/source/connection-monitoring-and-pooling/tests/pool-checkout-custom-maxConnecting-is-enforced.json
@@ -1,0 +1,85 @@
+{
+  "version": 1,
+  "style": "integration",
+  "description": "custom maxConnecting is enforced",
+  "runOn": [
+    {
+      "minServerVersion": "4.4.0"
+    }
+  ],
+  "failPoint": {
+    "configureFailPoint": "failCommand",
+    "mode": "alwaysOn",
+    "data": {
+      "failCommands": [
+        "isMaster",
+        "hello"
+      ],
+      "closeConnection": false,
+      "blockConnection": true,
+      "blockTimeMS": 500
+    }
+  },
+  "poolOptions": {
+    "maxConnecting": 1,
+    "maxPoolSize": 2,
+    "waitQueueTimeoutMS": 5000
+  },
+  "operations": [
+    {
+      "name": "ready"
+    },
+    {
+      "name": "start",
+      "target": "thread1"
+    },
+    {
+      "name": "start",
+      "target": "thread2"
+    },
+    {
+      "name": "checkOut",
+      "thread": "thread1"
+    },
+    {
+      "name": "waitForEvent",
+      "event": "ConnectionCreated",
+      "count": 1
+    },
+    {
+      "name": "wait",
+      "ms": 100
+    },
+    {
+      "name": "checkOut",
+      "thread": "thread2"
+    },
+    {
+      "name": "waitForEvent",
+      "event": "ConnectionReady",
+      "count": 2
+    }
+  ],
+  "events": [
+    {
+      "type": "ConnectionCreated"
+    },
+    {
+      "type": "ConnectionReady"
+    },
+    {
+      "type": "ConnectionCreated"
+    },
+    {
+      "type": "ConnectionReady"
+    }
+  ],
+  "ignore": [
+    "ConnectionCheckOutStarted",
+    "ConnectionCheckedIn",
+    "ConnectionCheckedOut",
+    "ConnectionClosed",
+    "ConnectionPoolCreated",
+    "ConnectionPoolReady"
+  ]
+}
--- a/source/connection-monitoring-and-pooling/tests/pool-checkout-custom-maxConnecting-is-enforced.yml
+++ b/source/connection-monitoring-and-pooling/tests/pool-checkout-custom-maxConnecting-is-enforced.yml
@@ -31,10 +31,6 @@ operations:
   - name: waitForEvent
     event: ConnectionCreated
     count: 1
-  # Wait long enough to be sure that thread1 started establishing a connection,
-  # and short enough to be sure that it is still blocked by failPoint.
-  - name: wait
-    ms: 100
   - name: checkOut
     thread: thread2
   - name: waitForEvent
--- a/source/connection-monitoring-and-pooling/tests/pool-checkout-custom-maxConnecting-is-enforced.yml
+++ b/source/connection-monitoring-and-pooling/tests/pool-checkout-custom-maxConnecting-is-enforced.yml
@@ -1,0 +1,54 @@
+version: 1
+style: integration
+description: custom maxConnecting is enforced
+runOn:
+  -
+    minServerVersion: "4.4.0"
+failPoint:
+  configureFailPoint: failCommand
+  mode: "alwaysOn"
+  data:
+    failCommands: ["isMaster","hello"]
+    closeConnection: false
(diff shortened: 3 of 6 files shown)


source/uri-options/uri-options.rst
```diff
@@ -210,6 +210,12 @@ pertaining to URI options apply here.
      - required for drivers with connection pools
      - The maximum number of clients or connections able to be created by a pool at a given time. This count includes connections which are currently checked out.
 
+   * - maxPoolConnecting
```
BorisDog, line 213: I'd suggest that the connection string option matches the `ConnectionPoolOptions` option name, same as with `maxPoolSize`/`minPoolSize`. Going with `maxPoolConnecting` might require replacing all instances of `maxConnecting` in the spec.
Would be great to get @patrickfreed opinion.

stIncMale: Using `maxConnecting` in a connection string makes what the option relates to less clear when judging based on the option name. I can rename it to `maxPoolConnecting` in the CMAP spec (I see no problem with renaming per se). Having the "pool" part in the name of a property in CMAP is a useless elaboration as all properties there are those of a pool. However, using the same name everywhere simplifies searching for it. So if I were to pick a single name, I'd pick **`maxPoolConnecting`** because of the connection string argument mentioned above.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -269,7 +278,7 @@ has the following properties:
    -  Connections are not created in the background to satisfy minPoolSize
 
 -  **Capped:** a pool is capped if **maxPoolSize** is set to a non-zero value. If a pool is capped, then its total number of `Connections <#connection>`_ (including available and in use) MUST NOT exceed **maxPoolSize**
--  **Rate-limited:** A Pool MUST limit the number of connections being created at a given time to be 2 (maxConnecting). 
+-  **Rate-limited:** A Pool MUST limit the number of `Connections <#connection>`_ being `established <#establishing-a-connection-internal-implementation>`_ concurrently.
```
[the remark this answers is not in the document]

ShaneHarvey (changes requested):
Nice work. I think the only thing missing is tests.

source/uri-options/uri-options.rst
stIncMale: Renamed to `maxConnecting`. I understand that @BorisDog and @ShaneHarvey agree with this name. @patrickfreed and @JamesKovacs, could you please tell whether you agree with this name for both the URI option and the pool option?

I am asking because we are trying to have this done in the Java driver 4.4., which is released on Thursday (I have only today and tomorrow for the implementation), and we can do that only if the parts visible to users (names of the options in this case) are approved.

patrickfreed: `maxConnecting` SGTM.

JamesKovacs: `maxConnecting` LGTM. The equivalent `mongos` server parameter that we are emulating is called [ShardingTaskExecutorPoolMaxConnecting](https://docs.mongodb.com/manual/reference/parameters/#mongodb-parameter-param.ShardingTaskExecutorPoolMaxConnecting). I don't see the need to include "pool" explicitly in the name to indicate that it is related to connection pools.

source/uri-options/uri-options.rst
```diff
@@ -210,6 +210,12 @@ pertaining to URI options apply here.
      - required for drivers with connection pools
      - The maximum number of clients or connections able to be created by a pool at a given time. This count includes connections which are currently checked out.
 
+   * - maxConnecting
+     - positive integer
+     - defined in the `Connection Pooling spec`_
+     - required for drivers with connection pools
+     - The maximum number of Connections a Pool may be establishing concurrently.
```
JamesKovacs, line 217: Please add spec tests to check that only positive integers are valid for `maxConnecting`.

JamesKovacs (changes requested):
Overall LGTM including the name `maxConnecting`. Just need some spec tests to validate URI parsing as well as `maxConnecting` behaviour.

source/uri-options/uri-options.rst
stIncMale: Done, see `connection-pool-options.yml`.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -1138,6 +1147,8 @@ Change log
 
 :2021-06-02: Formalize the behavior of a `Background Thread <#background-thread>`__.
 
+:2021-11-08: Make maxConnecting configurable.
```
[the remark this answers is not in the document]
stIncMale: Done, see `pool-checkout-custom-maxConnecting-is-enforced.yml`.

source/uri-options/uri-options.rst
```diff
@@ -482,6 +488,7 @@ this specification MUST be updated to reflect those changes.
 Changes
 -------
 
+- 2021-11-08 Add maxPoolConnecting option.
```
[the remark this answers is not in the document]
stIncMale: Done, see `connection-pool-options.yml`.

source/connection-monitoring-and-pooling/tests/pool-checkout-custom-maxConnecting-is-enforced.yml
```diff
@@ -0,0 +1,54 @@
...
+  - name: start
+    target: thread2
+  - name: checkOut
+    thread: thread1
+  - name: waitForEvent
+    event: ConnectionCreated
+    count: 1
+  # Wait long enough to be sure that thread1 started establishing a connection,
+  # and short enough to be sure that it is still blocked by failPoint.
+  - name: wait
```
patrickfreed, line 36: I think we can omit this wait, since observing the `ConnectionCreatedEvent` should be enough to ensure that the next thread cannot create a new connection per the psuedocode in [Creating a Connection (internal implementation)](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#creating-a-connection-internal-implementation).

stIncMale: Indeed, we of course grab a permit before emitting `ConnectionCreatedEvent`. Thanks for spotting this 👍

stIncMale: Done.

patrickfreed (approved):
LGTM!

source/uri-options/tests/connection-pool-options.yml
```diff
@@ -46,3 +47,21 @@ tests:
       auth: ~
       options:
           minPoolSize: 0 
+
+    -
+      description: "maxConnecting=0 causes an error"
+      uri: "mongodb://example.com/?maxConnecting=0"
+      valid: false
+      warning: false
```
ShaneHarvey, line 55: For both of these tests, this should be:
```
      valid: true
      warning: true
```

And the test names should change from "causes an error" to "causes a warning". See "Too low maxIdleTimeMS causes a warning" for example.

source/connection-monitoring-and-pooling/tests/pool-checkout-custom-maxConnecting-is-enforced.json
```diff
@@ -0,0 +1,81 @@
+{
+  "version": 1,
+  "style": "integration",
+  "description": "custom maxConnecting is enforced",
```
ShaneHarvey, line 4: This test LGTM, passes in python and fails if maxConnecting is not configurable.

ShaneHarvey (changes requested):

BorisDog (approved):
LGTM. Adding @DmitryLukyanov for validating the tests in C#.

source/uri-options/tests/connection-pool-options.yml
DmitryLukyanov: both of these tests failed in c# as well because we don't have a defined rule whether we should make the whole possible value validation on the connectionString level or no, sometimes we do it, sometimes no. To fix it, I added a value validation for this particular case

source/connection-monitoring-and-pooling/tests/pool-checkout-custom-maxConnecting-is-enforced.json
DmitryLukyanov: the test passed in c#

DmitryLukyanov (approved):
LGTM. I provided some notes regarding how `initial-dns-seedlist-discovery` passed in c# tests. I don't mind if these tests will be changed in a way suggested by @ShaneHarvey. The suggestion looks a bit better to me, because it's how most of the connectionString tests work in c#, but I will be ok with any decision here

source/uri-options/tests/connection-pool-options.yml
JamesKovacs: What is the reasoning behind not throwing an exception (or erring out) when `maxConnecting` is an invalid value? The .NET/C# driver doesn't have a mechanism to log warnings at the moment.

ShaneHarvey: I'm not sure why it's like this but that's how Python works and it's how we test the other options:
```python
>>> c = MongoClient("mongodb://example.com/?maxConnecting=-1")
/Users/shane/git/mongo-python-driver/pymongo/common.py:771: UserWarning: The value of maxconnecting must be a positive integer
  warnings.warn(str(exc))
```

stIncMale: @ShaneHarvey I also do not understand the problem with reacting to an invalid value of `maxConnecting` with an error. We do so for `srvMaxHosts`, for example. I do see that for `maxPoolSize` we produce a warning instead, but I don't see anything in https://github.com/mongodb/specifications/blob/master/source/uri-options/uri-options.rst that precludes producing an error and explains why a warning must be produced instead. I also fail to think of a good reason to produce a warning instead of an error.

ShaneHarvey: srvMaxHosts also uses warning:true, valid:true like I'm suggesting here:
[```
    {
      "description": "SRV URI with negative integer for srvMaxHosts",
      "uri": "mongodb+srv://test1.test.build.10gen.cc/?srvMaxHosts=-1",
      "valid": true,
      "warning": true,
      "hosts": null,
      "auth": null,
      "options": {}
    },
    {
      "description": "SRV URI with invalid type for srvMaxHosts",
      "uri": "mongodb+srv://test1.test.build.10gen.cc/?srvMaxHosts=foo",
      "valid": true,
      "warning": true,
      "hosts": null,
      "auth": null,
      "options": {}
    },
```](https://github.com/mongodb/specifications/blob/84bdb3f59892944efff7ad2b13d6fb1278dd81ba/source/uri-options/tests/srv-options.yml#L25-L38)

Comment (JamesKovacs) [reply source/uri-options/tests/connection-pool-options.yml:55]: I think it odd that we don't err on bad URI values in certain cases, but that's a bigger discussion not specific to `maxConnecting`. Let's go with a warning for invalid values of `maxConnecting` for consistency with other options.

**Review by JamesKovacs — APPROVED** [review]
LGTM

File: source/uri-options/tests/connection-pool-options.yml (hunk 7 above)
Comment (stIncMale) [reply source/uri-options/tests/connection-pool-options.yml:55]: And here it does not :) https://github.com/mongodb/specifications/blob/84bdb3f59892944efff7ad2b13d6fb1278dd81ba/source/uri-options/tests/srv-options.yml#L39-L42

I think we need a better argument in favour of not producing an error than the Python driver implementation detail.
There are both warnings and errors in the tests, and I think that errors are more explicit and produce less surprising results for a user than warnings (the driver does not end up configured differently from how a user expected it to be). This is why I think producing an error is preferable.

Comment (ShaneHarvey) [reply source/uri-options/tests/connection-pool-options.yml:55]: That case is different though. In that case it's the schema+srvMaxHosts that's an error. I think it's outside the scope of this ticket to change the behavior here. If you'd like to change the warning/error behavior we should create a new drivers ticket and address the problem consistently for all options. But connection string spec requires that invalid values for uri options will raise a warning, not an error:
```
Any invalid Values for a given key MUST be ignored and MUST log a WARN level message. For example:

Unsupported value for "fsync" : "ifPossible"
```
https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.rst#values

DmitryLukyanov: >What is the reasoning behind not throwing an exception

the reason that we almost never does it for just `number` values on this level, so consistency. 

>The .NET/C# driver doesn't have a mechanism to log warnings at the moment.

we consider warnings as errors if the exception is triggered and ignore it if the exception has not been raised.

stIncMale: Ok, @jyemin pointed out that the Connection String Spec says [_"Any invalid Values for a given key MUST be ignored and MUST log a WARN level message"_](https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.rst#:~:text=Any%20invalid%20Values%20for%20a%20given%20key%20MUST%20be%20ignored%20and%20MUST%20log%20a%20WARN%20level%20message). It does not explain the reasoning behind the requirement, but explains the reasoning behind a somewhat related requirement: [_"Q: Why WARN level warning on unknown options rather than throwing an exception?"_](https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.rst#:~:text=Q%3A%20Why%20WARN%20level%20warning%20on%20unknown%20options%20rather%20than%20throwing%20an%20exception%3F).

Despite there being no known explanation for the requirement of not producing errors for invalid values, we have an explicit requirement in the spec, which resolves the argument. I will change the test such that it expects a warning.

stIncMale: @ShaneHarvey I have just seen your [reply](https://github.com/mongodb/specifications/pull/1098#discussion_r749675228). Yes, that part of the spec is exactly the reason we have to produce a warning.

stIncMale: I updated the test. While doing so, I discovered the following:
1. The Java driver throws errors for invalid values of all options and ignores all spec tests that contain `"warning": true`.
2. The [URI-options test runner spec](https://github.com/mongodb/specifications/blob/master/source/uri-options/tests/README.rst) & tests are confusing regarding what it means for a connection string to be "valid". For example, a string that produces a warning during validation is invalid according to [_"Testing whether a URI is valid or not requires testing whether URI parsing (or MongoClient construction) causes a warning due to a URI option being invalid"_](https://github.com/mongodb/specifications/blob/master/source/uri-options/tests/README.rst#:~:text=Testing%20whether%20a%20URI%20is%20valid%20or%20not%20requires%20testing%20whether%20URI%20parsing%20(or%20MongoClient%20construction)%20causes%20a%20warning%20due%20to%20a%20URI%20option%20being%20invalid), yet many tests use `valid: true warning: true`.

ShaneHarvey (approved):
LGTM

source/uri-options/tests/connection-pool-options.yml
ShaneHarvey: > Testing whether a URI is valid or not requires testing whether URI parsing (or MongoClient construction) causes a warning due to a URI option being invalid

This line is incorrect and we should file a new ticket to change it.

PR #1102, opened 2021-11-23
Clarify that mongocryptdSpawnArgs may be accepted in multiple forms

kevinAlbs:
**Motivation for change**

Not all drivers accept the form ["--option", "value"].

For example, C# only accepts the following forms of `extraOptions.mongocryptdSpawnArgs`:

- The string: `--option1=Value1 --options2=Value2`
- Collection without dashes: `[ "option1=Value1", "option2=Value2" ]`
- Collection with dashes: `[ "--option1=Value1", "--option2=Value2" ]`

### DIFF
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -793,15 +793,18 @@ If a MongoClient is configured for Client Side Encryption
 (unless mongocryptdBypassSpawn=true), mongocryptd MUST be
 spawned by the driver. Spawning MUST include the command line argument
 --idleShutdownTimeoutSecs. If the user does not supply one through
-extraOptions.mongocryptdSpawnArgs (which may be either in the form
-"--idleShutdownTimeoutSecs=60" or as two consecutive arguments
-["--idleShutdownTimeoutSecs", 60], then the driver MUST append
+extraOptions.mongocryptdSpawnArgs, then the driver MUST append
 --idleShutdownTimeoutSecs=60 to the arguments. This tells mongocryptd
 to automatically terminate after 60 seconds of non-use. The stdout
 and stderr of the spawned process MUST not be exposed in the driver (e.g.
 redirect to /dev/null). Users can pass the argument --logpath to
 extraOptions.mongocryptdSpawnArgs if they need to inspect mongocryptd
 logs.
+
+Drivers MAY accept multiple forms of options in
+``extraOptions.mongocryptdSpawnArgs``. For example: both ``["--option=value"]``
+and ``["option=value"]`` MAY be accepted. Drivers MUST check all accepted forms
+when checking for the presence of ``--idleShutdownTimeoutSecs``.
 
 Upon construction, the MongoClient MUST create a MongoClient to
 mongocryptd configured with serverSelectionTimeoutMS=10000.


source/client-side-encryption/client-side-encryption.rst
```diff
@@ -793,16 +793,19 @@ If a MongoClient is configured for Client Side Encryption
...
 to automatically terminate after 60 seconds of non-use. The stdout
 and stderr of the spawned process MUST not be exposed in the driver (e.g.
 redirect to /dev/null). Users can pass the argument --logpath to
 extraOptions.mongocryptdSpawnArgs if they need to inspect mongocryptd
 logs.
 
+Drivers MAY accept multiple forms of options in
+``extraOptions.mongocryptdSpawnArgs``. For example: both ``["--option=value"]``
+and ``["option=value"]`` MAY be accepted. Drivers MUST check all accepted forms
+when checking for the presence of ``--idleShutdownTimeoutSecs``.
```
rstam, line 807: Why would we accept `option=value`. (without the `--` required by command line arguments`)

`extraArgs` should simply be strings that are passed as command line arguments to `mongocrypt`.

Therefore, only strings that are valid for `mongocrypt` should be accepted. 

For that matter, I don't see why drivers should do any validation of these arguments at all. Leave that validation for `mongocryptd` itself to do.

rstam (changes requested):

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: > Why would we accept `option=value`. (without the `--` required by command line arguments`)

@dmitry-lukyanov alerted me that this the C# driver already accepts the form `option=value`. The proposed change is not meant to be a behavior change. The intent is to clarify that the accepted forms are driver dependent.

> For that matter, I don't see why drivers should do any validation of these arguments at all. Leave that validation for `mongocryptd` itself to do.

`mongocryptd` does not terminate if `--idleShutdownTimeoutSecs` is passed.

The intended behavior is for `mongocryptd` terminate after 60 seconds of non-use unless the user requested otherwise by passing `--idleShutdownTimeoutSecs`.

Passing two occurrences of `--idleShutdownTimeoutSecs` to `mongocryptd` results in an error:

```
% mongocryptd --idleShutdownTimeoutSecs=60 --idleShutdownTimeoutSecs 30
Error parsing command line:  Multiple occurrences of option "--idleShutdownTimeoutSecs"
try 'mongocryptd --help' for more information
```

DmitryLukyanov: As I see this change says that drivers MAY allow different forms to specify mongocryptdSpawnArgs arguments. The requirement with mandatory `--` doesn't really have value (at least to me), what's the purpose for end-user to have this requirement? If someone missed specifying `--`, does it mean that we don't know how to configure mongocryptd? Should we really be dependent on this low level configuration behavior when we configure MongoClient?

However, I will be ok with any decision here, but changing this now will be a BC in c# which won't have any real value.

rstam: I consider the current behavior of the C# driver to be a bug.

We should not change the spec to condone the bug.

`extraOptions` is a way to provide extra command line arguments to `mongocryptd`. I think those argument MUST already be in a format acceptable to `mongocryptd`. Drivers should not manipulate those options in any way at all.

I understand the need to examing the `extraOptions` to see if `--idleShutdownTimeoutSecs` is present or not and add it if missing. That's unrelated to whether drivers are allowed to change the format of the `extraOptions` before passing them to the `mongocryptd` command line args.

rstam (changes requested):

source/client-side-encryption/client-side-encryption.rst
JamesKovacs: We should err on the side of having consistent behaviour across drivers unless there is an idiomatic, language-specific need otherwise. In this case, `extraOptions` is a way to pass command-line options to `mongocryptd`. Whatever options `mongocryptd` expects should be passed from `extraOptions` to `mongocryptd` unaltered.

It is common Unix practice for the full name of an argument to be prefixed by a double dash (`--`) and the abbreviated form by a single dash (`-`). `grep` is a good example of this idiom where `grep -e pattern` is the abbreviated form and `grep --regexp=pattern` is the extended form.

`mongocryptd` follows a similar idiom, but we don't provide as many abbreviated forms as common Unix commands. For example you can specify verbose mode either via `-v` or `--verbose`. You can also specify a config file using `-f` or `--config`. Our drivers should not be inserting or removing dashes from provided arguments. For example, if you specified verbose mode in `extraOptions` as `-v`, if we automatically modified this to `--v`, `mongocryptd` reports:

```
Error parsing command line: unrecognised option '--v'
try 'mongocryptd --help' for more information
```

`mongocryptd` may add more abbreviated forms in the future and we don't want to disallow valid configuration because we are modifying `extraOptions`.

Let's discuss how we can fix this issue in the .NET/C# driver and remain consistent with how other drivers expose `extraOptions` as a raw list of command line arguments to `mongocryptd`.

DmitryLukyanov: After additional discussions, we will probably update behavior on the c# side to correspond the original spec, so it's no longer required PR

kevinAlbs:
Thank you for the discussion. Closing.

PR #1104, opened 2021-11-23
DRIVERS-1991 Fix data lake tests which use legacy test format

ShaneHarvey:
The data lake tests currently use the legacy CRUD v2 (same as the legacy transactions) test format where commandName and databaseName should be named command_name and database_name.

### DIFF
--- a/source/atlas-data-lake-testing/tests/estimatedDocumentCount.json
+++ b/source/atlas-data-lake-testing/tests/estimatedDocumentCount.json
@@ -32,8 +32,8 @@
                 }
               ]
             },
-            "commandName": "aggregate",
-            "databaseName": "test"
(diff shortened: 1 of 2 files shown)


jyemin (approved):
Huh, I will have to check why Java driver passed the test this way.  I thought I copied this from another unified test.

jyemin:
Actually, can you hold off on merging this for a sec?

jyemin (approved):
Sorry, totally misunderstood.  Java driver passes because the legacy test runner doesn't check for unsupported fields like this. I'm not going to bother fixing the test runner.

LGTM

ShaneHarvey:
I opened https://jira.mongodb.org/browse/DRIVERS-1993 to migrate to the unified test format.

PR #1105, opened 2021-11-23
DRIVERS-1994 deleteOne-let, deleteMany-let, updateMany-let missing fields in events

juliusgeo:


ShaneHarvey:
LGTM but can you link to a passing patch build that tests these changes?

juliusgeo:
Here is the link to the passing build: https://evergreen.mongodb.com/version/619d8fded6d80a0f59e94262?redirect_spruce_users=true

ShaneHarvey (approved):
LGTM

source/crud/tests/unified/updateMany-let.yml
```diff
@@ -92,6 +93,7 @@ tests:
                   -
                     q: *filter1
                     u: *update1
+                    multi: true
```
jmikola, line 96: Since you're also updating the `updateMany` test here, I'd suggest revising the title of DRIVERS-1994 _and_ the commit for this PR to accurately reflect the change.

I'll note that for updateOne, existing tests use `multi: { $$unsetOrMatches: false }`, but that's not necessary for updateMany, since `multi` will be required. Likewise for deleteOne/Many, which will always specify a value for `limit`.

juliusgeo: Done.

jmikola (approved):

PR #1109, opened 2021-12-10
DRIVERS-2006: Remove srvMaxHosts tests expecting an error for invalid values

jmikola:
https://jira.mongodb.org/browse/DRIVERS-2006

These cases should raise warnings, not errors. Since they are already tested by existing uri-options spec tests, these files can simply be removed.


Neilshweky (approved):
LGTM!

p-mongo (approved):

PR #1110, opened 2021-12-14
DRIVERS-1857 relax exhaust cursor requirements

kevinAlbs:
# Summary of changes
- Require OP_MSG for exhaust cursors on 5.1+ instead of 3.6+.
- Relax requirement from MUST to SHOULD, and permit falling back to a non-exhaust cursor.

# Rationale
Exhaust cursors are [underspecified](https://jira.mongodb.org/browse/DRIVERS-535) and not implemented in all drivers.

Falling back to a non-exhaust cursor matches the proposed approach in https://github.com/mongodb/mongo-c-driver/pull/912.

Exhaust cursors with OP_QUERY only impact the C driver. Previously it was thought to affect Ruby as well. RUBY-2770 determined it is not possible to request an exhaust cursor.

### DIFF
--- a/source/find_getmore_killcursors_commands.rst
+++ b/source/find_getmore_killcursors_commands.rst
@@ -348,7 +348,8 @@ cursors. Server versions 5.1 and later do not support **OP_QUERY**.
 On server versions 5.1 and later, drivers MAY fallback to a non-exhaust cursor
 if an exhaust cursor is requested.
 
-On server versions 4.2 to 5.0 drivers SHOULD use **OP_MSG** but MAY use legacy **OP_QUERY**.
+On server versions 4.2 to 5.0 drivers SHOULD use **OP_MSG** but MAY use legacy
+**OP_QUERY**.
 
 On server versions prior to 3.6, drivers MUST fallback to the legacy **OP_QUERY**
 wire protocol messages. The **find** command does not support the exhaust flag
--- a/source/find_getmore_killcursors_commands.rst
+++ b/source/find_getmore_killcursors_commands.rst
@@ -348,6 +348,8 @@ cursors. Server versions 5.1 and later do not support **OP_QUERY**.
 On server versions 5.1 and later, drivers MAY fallback to a non-exhaust cursor
 if an exhaust cursor is requested.
 
+On server versions 4.2 to 5.0 drivers SHOULD use **OP_MSG** but MAY use legacy **OP_QUERY**.
+
 On server versions prior to 3.6, drivers MUST fallback to the legacy **OP_QUERY**
 wire protocol messages. The **find** command does not support the exhaust flag
 from **OP_QUERY**.
--- a/source/find_getmore_killcursors_commands.rst
+++ b/source/find_getmore_killcursors_commands.rst
@@ -342,9 +342,8 @@ Exhaust
 
 This section only applies to drivers that support exhaust cursors.
 
-On server versions 5.1 and later drivers SHOULD use **OP_MSG** for exhaust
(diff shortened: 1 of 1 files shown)


behackett:
PyMongo supports exhaust with OP_MSG when available: https://github.com/mongodb/mongo-python-driver/commit/d26bf933ed22f9355959ed6c957b6fcce1ead228

source/find_getmore_killcursors_commands.rst
```diff
@@ -342,7 +342,12 @@ Exhaust
 
 This section only applies to drivers that support exhaust cursors.
 
-On server versions 3.6 and later drivers MUST use **OP_MSG** for exhaust cursors.
+On server versions 5.1 and later drivers SHOULD use **OP_MSG** for exhaust
+cursors. Server versions 5.1 and later do not support **OP_QUERY** for exhaust
+cursors.
```
eramongodb, line 347:
```suggestion
On server versions 5.1 and later drivers MUSTuse **OP_MSG** for exhaust
cursors. Server versions 5.1 and later do not support **OP_QUERY**.
```
I think "SHOULD" should be "MUST" given server versions 5.1 and newer do not support `OP_QUERY` _at all_, not just for exhaust cursors. Failure to comply would constitute a faulty driver implementation for server versions 5.1 and newer (exhaust cursor that _doesn't_ use OP_MSG is not a valid option).

kevinAlbs: Good catch.

🤦 I thought to making this a `SHOULD` was required to permit the fallback behavior. But the fallback still uses OP_MSG.

source/find_getmore_killcursors_commands.rst
```diff
@@ -731,6 +735,8 @@ More in depth information about passing read preferences to Mongos can be found
 
 Changes
 =======
+2021-12-14 Exhaust cursors may fallback to non-exhaust cursors on 5.1+ servers.
```
eramongodb, line 738: In addition to permitting fallback behavior, may want to include note regarding revision of prior OP_MSG requirement on 3.6+ servers being relaxed to 5.1+.

eramongodb (approved):

source/find_getmore_killcursors_commands.rst
```diff
@@ -342,7 +342,11 @@ Exhaust
...
 This section only applies to drivers that support exhaust cursors.
 
-On server versions 3.6 and later drivers MUST use **OP_MSG** for exhaust cursors.
+On server versions 5.1 and later drivers MUST use **OP_MSG** for exhaust
+cursors. Server versions 5.1 and later do not support **OP_QUERY**.
+
+On server versions 5.1 and later, drivers MAY fallback to a non-exhaust cursor
+if an exhaust cursor is requested.
 
 On server versions prior to 3.6, drivers MUST fallback to the legacy **OP_QUERY**
```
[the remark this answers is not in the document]
kevinAlbs: Nice, I think describing the server behavior helps give context. I will do the refactor in a follow-up PR (https://github.com/mongodb/specifications/pull/1112)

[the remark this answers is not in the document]
ShaneHarvey: SGTM

ShaneHarvey (approved):

p-mongo (approved):
I have no objections because Ruby does not implement exhaust cursors and hence is not affected by the proposed change.

Producing a non-exhaust cursor when an exhaust cursor was requested seems like it could potentially be problematic but  if the drivers where exhaust cursors are implemented are OK with this I have no objections.

PR #1111, opened 2021-12-14
DRIVERS-1393 listCollectionNames should use authorizedCollections argument

Neilshweky:
A few things to ask/note about this ticket/PR:

1. The proposed changes in this spec may change the behavior of the `listCollectionNames` driver helper to list only the collections that the user has access to, as opposed to returning all collection names (which requires database wide access).
2. That said, should the `authorizedCollections` be forcibly set to true (i.e. even if the user provides the option as false) or just defaulted to true?
    - On the one hand, this is would match the behavior of the `db.getCollectionNames()` shell command, and it is still possible to retrieve all of the collections using `listCollections`.
    - On the other hand, we might not want to change the functionality of `listCollectionNames` and provide the users the option to retrieve all collection names. The Ruby driver plans on using this option.

3. Should I be adding `authorizedCollections` docs to the other parts of the spec, i.e. `listCollections`, `listMongoCollections`, etc, or should I leave it only in `listCollectionNames` since this is the only function that requires it as of now? 

EDIT: Jeremy clarified the answers to these questions in the ticket [DRIVERS-1393](https://jira.mongodb.org/browse/DRIVERS-1393).

### DIFF
--- a/source/enumerate-collections.rst
+++ b/source/enumerate-collections.rst
@@ -271,6 +271,11 @@ MongoDB 4.0 introduced a ``nameOnly`` boolean option to the ``listCollections``
 database command, which limits the command result to only include collection
 names. NOTE: ``nameOnly`` is applied before any filter is applied.
 
+MongoDB 4.0 also added an ``authorizedCollections`` boolean option to the ``listCollections``
+command, which can be used to limit the command result to only include collections
+the user is authorized to use. Drivers MUST always specify ``authorizedCollections``
+as true when they intend to access collection names from the ``listCollections`` command.
+
 Example return::
 
     [
@@ -284,11 +289,12 @@ Example return::
 
 
 Server version between 2.7.6 (inclusive) and 4.0 (exclusive) do not support
-the ``nameOnly`` option for the ``listCollections`` command and will ignore it 
-without raising an error. Therefore, drivers MUST always specify the ``nameOnly`` 
-option when they only intend to access collection names from the ``listCollections`` 
-command result, except drivers MUST NOT set ``nameOnly`` if a filter
-specifies any keys other than ``name``.
+the ``nameOnly`` and ``authorizedCollections`` options for the ``listCollections``
+command and will ignore it without raising an error. Therefore, drivers MUST
+always specify the ``nameOnly`` and ``authorizedCollections`` options when they
+only intend to access collection names from the ``listCollections`` command
+result, except drivers MUST NOT set ``nameOnly`` if a filter specifies any
+keys other than ``name``.
 
 Getting Full Collection Information
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
@@ -433,6 +439,9 @@ The shell implements the first algorithm for falling back if the
 
 Version History
 ===============
+Version 0.6.2 Changes
+    - Support ``authorizedCollections`` option in ``listCollections`` command.
+
 Version 0.6.1 Changes
     - Update to use secondaryOk.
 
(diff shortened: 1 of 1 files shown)


jmikola:
I left a response in [DRIVERS-1393](https://jira.mongodb.org/browse/DRIVERS-1393?focusedCommentId=4251284&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-4251284) that should address most of the questions above.

source/enumerate-collections.rst
```diff
@@ -362,7 +368,8 @@ Example return (in PHP, but abbreviated)::
       [5] => class MongoCollection#11 { }
     }
 
-Drivers MUST specify the ``nameOnly`` option when executing the ``listCollections`` command for this method,
+Drivers MAY allow the ``nameOnly`` and ``authorizedCollections`` options
+to be passed when executing the ``listCollections`` command for this method,
```
jmikola, line 372: This is a behavioral change for `nameOnly`. I think you should retain MUST for that option since drivers should only need the name to construct their collection objects.

If anything, you can improve the original text by stating that drivers should specify `true` for `nameOnly`. "Drivers MUST specify the ``nameOnly`` option" is a bit vague, since it's only talking about the option being specified and says nothing of its value.

You can add a separate paragraph to state that drivers MAY support `authorizedCollections`, similar to the text used in the preceding section.

source/enumerate-collections.rst
```diff
@@ -271,6 +271,11 @@ MongoDB 4.0 introduced a ``nameOnly`` boolean option to the ``listCollections``
 database command, which limits the command result to only include collection
 names. NOTE: ``nameOnly`` is applied before any filter is applied.
 
+MongoDB 4.0 also added an ``authorizedCollections`` boolean option to the ``listCollections``
+command, which can be used to limit the command result to only include collections
+the user is authorized to use. Drivers MAY allow users to set the ``authorizedCollections``
+option on the ``listCollectionNames`` method.
```
jmikola, line 277: IMO, this paragraph would read better if placed at the end of this section. The list below and subsequent paragraph are still talking about `nameOnly`.

source/enumerate-collections.rst
```diff
@@ -433,6 +440,9 @@ The shell implements the first algorithm for falling back if the
 
 Version History
 ===============
+Version 0.6.2 Changes
+    - Support ``authorizedCollections`` option in ``listCollections`` command.
```
jmikola, line 444: The previous patch version bump to 0.6.1 made sense because it was renaming `slaveOkay` to `secondaryOk`.

In this case, you're introduce a new option to the spec. That seems more like a minor version bump, so I'd suggest 0.7.0 in this case.

source/enumerate-collections.rst
Neilshweky: done

source/enumerate-collections.rst
Neilshweky: done

source/enumerate-collections.rst
```diff
@@ -362,8 +368,12 @@ Example return (in PHP, but abbreviated)::
       [5] => class MongoCollection#11 { }
     }
 
-Drivers MUST specify the ``nameOnly`` option when executing the ``listCollections`` command for this method,
-except drivers MUST NOT set ``nameOnly`` if a filter specifies any keys other than ``name``.
+Drivers SHOULD specify true for the ``nameOnly`` option when executing the
```
jmikola, line 371: SHOULD is still a change from MUST in the original text. I don't feel strongly about this, but if it wasn't intended you may want to revert.

jmikola (approved):
LGTM. I don't need to follow up on my other comment (you can decide to address it or not).

source/enumerate-collections.rst
Neilshweky: Ah I thought you wanted should based on the text of your other comment, I can change that for sure.

jmikola: Oh, I see how my last comment could have suggested that. I was just saying "should" in English and didn't mean to suggest spec language there. Apologies.

Neilshweky: No worries!

PR #1113, opened 2021-12-16
Clarify that old drivers should default to OP_QUERY handshakes

jmikola:
The original intention of this paragraph was to allow newer drivers published after MongoDB 3.6 to avoid implementing OP_QUERY purely for an initial handshake. Drivers that have historically supported pre-3.6 servers should continue to use OP_QUERY to ensure that meaningful error messages are raised when connecting to an old server instead of a silently dropped connection due to an unrecognized protocol.

----

Note: I skipped making a DRIVERS ticket for this to reduce overhead. If someone feels one is necessary I can do so.

### DIFF
--- a/source/message/OP_MSG.rst
+++ b/source/message/OP_MSG.rst
@@ -10,9 +10,9 @@ OP_MSG
 :Informed: Bryan Reinero, Chris Hendel, drivers@
 :Status: Approved
 :Type: Standards
-:Last Modified: 2021-04-20
+:Last Modified: 2021-12-16
 :Minimum Server Version: 3.6
-:Version: 1.3
+:Version: 1.3.1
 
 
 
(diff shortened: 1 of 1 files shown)


kevinAlbs (approved):
Thank you for making the clarification and capturing this rationale.

jyemin (approved):
Thanks, great idea to clarify.

source/message/OP_MSG.rst
```diff
@@ -54,10 +54,10 @@ Usage
 MongoDB drivers SHOULD perform the MongoDB handshake using ``OP_MSG`` if an API
 version was declared on the client, but MAY decide to use ``OP_QUERY``.
 
-If no API version was declared, drivers that support MongoDB 3.4 and earlier
-MUST perform the handshake using ``OP_QUERY`` to determine if the node supports
-``OP_MSG``. Drivers that only support MongoDB 3.6 and newer MAY default to using
-``OP_MSG``.
+If no API version was declared, drivers that have historically supported MongoDB
+3.4 and earlier MUST perform the handshake using ``OP_QUERY`` to determine if the
```
chardan, line 58: Although *IF* the user of the driver has specified a specific protocol, we can send OP_QUERY, so MUST doesn't seem right. As I understand it, OP_QUERY is to (always) be used if the driver intends to support <= MongoDB 3.4 servers AND the user has not specified a specific protocol, which is slightly different. (But, maybe this text isn't the right place to capture that?)

chardan (changes requested):
I'm new to the way that these documents are written, but is "MUST" the right wording? THANK YOU for doing this!

source/message/OP_MSG.rst
chardan: Ah, I see if "no API version *was declared*". Ok. +1. (This captures the case where the user has requested a specific version.) +1

chardan (approved):
+1, I hadn't fully parsed "no API version was declared". +1. Thanks again!

PR #1115, opened 2022-01-07
DRIVERS-746: Make handshake errors retryable

nbbeeken:
Add ~optional~ **REQUIRED** retryablity for network errors when establish a connection and running a handshake

[DRIVERS-746](https://jira.mongodb.org/browse/DRIVERS-746)


ShaneHarvey (changes requested):
A few questions. I haven't POCed the tests yet.

source/retryable-reads/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,71 @@
+description: "retryable reads handshake failures"
+
+schemaVersion: "1.6"
+
+runOnRequirements:
+  - minServerVersion: "4.2"
+    topologies: [replicaset]
```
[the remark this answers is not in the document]
nbbeeken: done!

source/retryable-writes/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,79 @@
+description: "retryable writes handshake failures"
+
+schemaVersion: "1.6"
+
+runOnRequirements:
+  - minServerVersion: "4.2"
+    topologies: [replicaset]
```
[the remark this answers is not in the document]
nbbeeken: done! also should load-balanced be tested?

source/retryable-reads/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,71 @@
+description: "retryable reads handshake failures"
```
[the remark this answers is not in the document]
nbbeeken: done!

source/retryable-writes/retryable-writes.rst
```diff
@@ -219,6 +219,10 @@ RetryableWriteError Labels
 
 The RetryableWriteError label might be added to an error in a variety of ways:
 
+- When the driver encounters a network error establishing an initial connection to a server,
+  it SHOULD add a RetryableWriteError label to that error if the MongoClient performing
+  the operation has the retryWrites configuration option set to true.
```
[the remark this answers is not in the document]
nbbeeken: I took a look at the txn spec is there a specific section that you're thinking of? I believe that specification of not being allowed to retry writes would exclude this behavior from working, should we add a test to show that's the case?

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -602,6 +602,9 @@ established. In addition, the Pool MUST NOT prevent other threads from checking
 out `Connections <#connection>`_ while establishing a `Connection
 <#connection>`_.
 
+If connection establishment fails due to a network error drivers SHOULD mark the
```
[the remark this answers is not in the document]
nbbeeken: I can roll back the change in here, I was thinking that it wasn't sufficient to mention the network error in the retryable spec since it would be created at the CMAP layer, but I think I'm influenced by how node does things here.

source/retryable-writes/retryable-writes.rst
```diff
@@ -219,6 +219,10 @@ RetryableWriteError Labels
 
 The RetryableWriteError label might be added to an error in a variety of ways:
 
+- When the driver encounters a network error establishing an initial connection to a server,
+  it SHOULD add a RetryableWriteError label to that error if the MongoClient performing
```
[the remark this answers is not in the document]
nbbeeken: I originally thought the ticket asked for it to be optional, but I think I misread, changed!

source/retryable-reads/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,71 @@
...
+
+schemaVersion: "1.6"
```
ShaneHarvey, line 3: Is schemaVersion 1.6 actually require here or can we use a lower version? I think the best practice is to set this to the lowest version possible.

source/retryable-reads/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,71 @@
...
+            data:
+              failCommands: [ saslContinue ] # use saslContinue here to avoid SDAM errors
+              closeConnection: true
+      -
+        name: findOne
+        object: *collection0
+        arguments:
+          filter: { _id: 2 }
+        expectResult:
+          - { _id: 2, x: 22 }
```
ShaneHarvey, line 55: The findOne result should not be a list.

source/retryable-reads/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,71 @@
...
+          - { connectionCheckOutStartedEvent: {} }
+      - client: *client0
+        events:
+          - &findAttempt
+            commandStartedEvent:
+              command:
+                find: *collectionName
+                filter: { _id: 2 }
+              databaseName: *databaseName
+          - *findAttempt
```
ShaneHarvey, line 71: There should only be one findAttempt since the first one is never actually sent to the server.

source/retryable-reads/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,71 @@
...
+        # fail the next connection establishment
+        name: failPoint
+        object: testRunner
+        arguments:
+          client: *client0
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+              failCommands: [ saslContinue ] # use saslContinue here to avoid SDAM errors
```
ShaneHarvey, line 47: This failPoint operation creates one working connection in the pool so the findOne always succeeds on the first attempt in python. We need to update this test to guarantee that findOne will open a new connection. One way is to cause the failPoint connection to be discarded:
```
    operations:
      -
        # fail the next connection establishment
        name: failPoint
        object: testRunner
        arguments:
          client: *client0
          failPoint:
            mode: { times: 2 }
            data:
              failCommands: [ saslContinue, dropCollection ] # use saslContinue here to avoid SDAM errors 
              closeConnection: true
      -
        name: dropCollection
        object: *database0
        expectError:
          isError: true
      -
        name: findOne
        object: *collection0
        arguments:
          filter: { _id: 2 }
        expectResult:
          - { _id: 2, x: 22 }
```

ShaneHarvey (changes requested):

source/retryable-reads/retryable-reads.rst
```diff
@@ -163,6 +163,7 @@ Drivers SHOULD support retryability for the following operations:
 - Any driver that provides generic command runners for read commands (with logic
   to inherit a client-level read concerns) SHOULD implement retryability for the
   read-only command runner.
+- Handshake failures due to a network error.
```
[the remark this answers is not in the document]
nbbeeken: Changed the wording but its kinda light still. I'm not fully grokking how you determine its a handshake failure at this layer. Would it be correct to say any error that checkOut throws that isn't related to pool errors or waitQueueTimeouts is retryable? and implicitly related to handshakes in that case?

Currently the pseudo code catches a "NetworkError" but am I correct in assuming a ShutdownInProgress error doesn't reset the connection? So we need to add logic/spec text to pass errors from `pool.checkOut()` through our exising retryable rules, since that should catch something like shutdownInProgress

source/retryable-writes/retryable-writes.rst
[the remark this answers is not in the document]
nbbeeken: Added commit and abort transaction tests and a mention of it in the spec!

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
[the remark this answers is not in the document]
nbbeeken: Done, removed the cmap changes!

source/retryable-reads/retryable-reads.rst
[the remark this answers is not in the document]
nbbeeken: I'll put up a test to include the ShutDownInProgress example we discussed, I think it would fit here.

source/retryable-reads/tests/unified/handshakeError.yml
nbbeeken: done!

source/retryable-reads/tests/unified/handshakeError.yml
nbbeeken: done!

source/retryable-reads/tests/unified/handshakeError.yml
nbbeeken: right, good catch thanks!

source/retryable-reads/tests/unified/handshakeError.yml
nbbeeken: I pulled in the dropCollection example, let me know how this runs on your end

source/retryable-reads/retryable-reads.rst
```diff
@@ -79,7 +79,7 @@ SocketException                 9001
 
   .. _PoolClearedError: ../connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#connection-pool-errors
 
-MongoClient Configuration 
+MongoClient Configuration
```
jmikola, line 82: If you're trimming trailing whitespace I think you also need to adjust the heading line below to avoid RST warnings. Not just here but in the trimmed headings later in this file.

source/retryable-reads/retryable-reads.rst
```diff
@@ -163,6 +163,7 @@ Drivers SHOULD support retryability for the following operations:
 - Any driver that provides generic command runners for read commands (with logic
   to inherit a client-level read concerns) SHOULD implement retryability for the
   read-only command runner.
+- Any error that occurs during a handshake. (ex. Auth, ShutdownInProgress)
```
jmikola, line 166:
```suggestion
- Any error that occurs during a handshake (eg. Auth, ShutdownInProgress).
```

source/retryable-reads/retryable-reads.rst
```diff
@@ -365,8 +366,11 @@ and reflects the flow described above.
       connection = server.getConnection()
     } catch (PoolClearedException poolClearedError) {
       return executeRetry(command, session, poolClearedError);
+    } catch (NetworkException networkError) {
+      // ConnectionPool should have handled SDAM state changes
```
jmikola, line 370: What does this comment mean? The block comment about `try` already talks about SDAM being updated by whatever causes `PoolClearedException` to be thrown, but it's not clear to me how a NetworkException here relates to a connection pool.

source/retryable-reads/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,68 @@
+description: "retryable reads handshake failures"
+
+schemaVersion: "1.0"
```
jmikola, line 3: `runOnRequirement.auth`, the "load-balanced" topology, and CMAP observation were introduced in schema version 1.3 (https://github.com/mongodb/specifications/commit/ed4e62b48939378b814cbd7d9d4939a2c1076a6b).

source/retryable-reads/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,68 @@
...
+    operations:
+      - name: failPoint # fail the next connection establishment
+        object: testRunner
+        arguments:
+          client: *client0
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 2 }
+            data:
+              failCommands: [saslContinue, dropCollection] # use saslContinue here to avoid SDAM errors
```
jmikola, line 44: If these operations execute sequentially, what necessitates the driver checking out a second connection (which will fall during `saslContinue`) instead of re-using the first already-authed connection?

Separate question: `saslContinue` isn't used for _every_ [auth mechanism](https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst). Would this test fail if the test suite was run against a cluster using X509 authentication? If so, is that a concern that'd warrant new test format syntax in `runOnRequirement` to check the authMechanism? There may be a BC-friendly way to introduce that where `auth` can start accepting an array of permitted authMechanism strings while still allowing a boolean (i.e. `true` matches any mechanism, and `false` requires none).

source/retryable-writes/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,69 @@
+description: "retryable writes handshake failures"
+
+schemaVersion: "1.0"
```
jmikola, line 3: Per my earlier comment in the retryable reads test, this needs to be bumped.

source/retryable-reads/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,68 @@
...
+initialData:
+  - collectionName: *collectionName
+    databaseName: *databaseName
+    documents:
+      - { _id: 1, x: 11 }
+      - { _id: 2, x: 22 }
+      - { _id: 3, x: 33 }
+
+tests:
+  - description: "InsertOne succeeds after retryable handshake network error"
```
jmikola, line 34: This looks like copypasta.

source/retryable-reads/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,68 @@
...
+        object: testRunner
+        arguments:
+          client: *client0
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 2 }
+            data:
+              failCommands: [saslContinue, dropCollection] # use saslContinue here to avoid SDAM errors
+              closeConnection: true
+      - name: dropCollection
```
jmikola, line 46: Why is `dropCollection` used here? This seems more like a test of a write operation than anything pertaining to `findOne`.

Also, since `findOne` technically isn't in the CRUD spec, would it make more sense to use `find` here?

source/retryable-reads/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,68 @@
...
+        arguments:
+          filter: { _id: 2 }
+        expectResult: { _id: 2, x: 22 }
+
+    expectEvents:
+      - client: *client0
+        eventType: cmap
+        events:
+          - { connectionCheckOutStartedEvent: {} }
+          - { connectionCheckOutStartedEvent: {} }
```
jmikola, line 61: Does this entire spec change only apply to drivers with connection pools? I'm reading through [DRIVERS-746](https://jira.mongodb.org/browse/DRIVERS-746) and although it does say:

> * there is no idle connection already available in the pool
> * the driver attempts to open a new connection to the server and complete the connection handshake

...I don't think it technically requires a pool. The main issue has to do with a retryable operation using an entirely new connection, which would seem applicable to _all_ drivers (even single-threaded libmongoc w/o a pool).

If I'm not mistaken, is there a more portable way we can write these tests? Perhaps setting the fail point with one client and then using a separate client to run the retryable read/write as its first operation? And if we go that route, would there still be a need to require auth and fail on `saslContinue` instead of a `hello` command?

source/transactions/tests/unified/retryable-abort-handshake.yml
```diff
@@ -0,0 +1,88 @@
+description: "retryable abortTransaction on handshake errors"
+
+schemaVersion: "1.0"
```
jmikola, line 3: `serverless` syntax was introduced in schema version 1.4 (https://github.com/mongodb/specifications/commit/3586db8b1ed90d835737f6226bb3ff9012cfc5de).

source/transactions/tests/unified/retryable-abort-handshake.yml
```diff
@@ -0,0 +1,88 @@
...
+tests:
+  - description: "AbortTransaction succeeds after handshake network error"
+    operations:
+      - name: failPoint # fail the next connection establishment
+        object: testRunner
+        arguments:
+          client: *client0
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 2 }
```
jmikola, line 44: I feel like I'm definitely misunderstanding something, but why does this use `times: 2` like the retryable reads spec's `handshakeError.yml` test? Is that necessary to apply to _both_ the `insert` and `abortTransaction` commands? I'm probably still confused as to why two sequential commands would necessitate checking out separate connections from the pool.

source/retryable-reads/retryable-reads.rst
```diff
@@ -670,9 +674,11 @@ minor risk of degraded performance. Additionally, any customers experiencing
 degraded performance can simply disable ``retryableReads``.
 
 
-Changelog 
+Changelog
 ==========
 
+2022-01-07: Note that drivers should retry handshake network failures.
```
jmikola, line 680: This should have a corresponding last modified and version bump in the header. I assume a minor version bump is appropriate since this is a bit more than a typo/clarification.

source/retryable-writes/retryable-writes.rst
```diff
@@ -791,14 +795,16 @@ Why does the driver only add the RetryableWriteError label to errors that occur
...
-only add the label to an error when the client has added a txnNumber to the 
+only add the label to an error when the client has added a txnNumber to the
 command, which only happens when the retryWrites option is true on the client.
 For the driver to add the label even if retryWrites is not true would be
 inconsistent with the server and potentially confusing to developers.
 
 Changes
 =======
 
+2022-01-07: Note that drivers should retry handshake network failures.
```
jmikola, line 806: Reminder to bump version and last modified date in header.

source/transactions/transactions.rst
```diff
@@ -438,9 +438,9 @@ have been performed on this transaction, drivers MUST NOT run the
 commitTransaction command.
 
 commitTransaction is a retryable write command. Drivers MUST retry once
-after commitTransaction fails with a retryable error according to the
-Retryable Writes Specification, regardless of whether retryWrites is set
-on the MongoClient or not.
+after commitTransaction fails with a retryable error, including a
+handshake network error, according to the Retryable Writes Specification,
+regardless of whether retryWrites is set on the MongoClient or not.
```
jmikola, line 443: This should be accompanied with a last modified and changelog update. I suppose the version can be a patch bump, since it's a subtle change (vs. the pseudo-code and implementation being updated in the retryable read/write specs).

jmikola (commented):
Sorry for the late review. I lost track of this in my notifications.

source/retryable-reads/retryable-reads.rst
nbbeeken: What warnings are you referring too, is there some local tooling that checks the rst format?

source/retryable-reads/retryable-reads.rst
nbbeeken: done!

source/retryable-reads/retryable-reads.rst
nbbeeken: This could use further clarity, but my intention was to differentiate poolClearedExceptions which can be caused by something else, such as a pool being in the paused state. Maybe it would be better to represent it as a flag upon the cleared exception? 

The comment mentioning SDAM is a kind of reminder that at this point the pool should've done its logic to mark the relevant server unknown. When I first put up the PR there was some changes I made to CMAP pseudo code so this related to that, we can remove it to keep the code brief.

source/retryable-reads/tests/unified/handshakeError.yml
nbbeeken: done!

source/retryable-reads/tests/unified/handshakeError.yml
nbbeeken: I mention the reason behind this dropCollection in another thread.

As for the `saslContinue`, this is a way to get handshakes to fail only for operations and not on monitoring connections. There's precedent for it in sdam integration tests and (where I originally got the idea from) this [lb test](https://github.com/mongodb/specifications/blob/master/source/load-balancers/tests/sdam-error-handling.yml#L187). But I do agree that being able to be specific about the authMechanism could be good for future shared coverage of those mechs. I can file a followup drivers ticket

source/retryable-writes/tests/unified/handshakeError.yml
nbbeeken: done!

source/retryable-reads/tests/unified/handshakeError.yml
nbbeeken: Fixed!

source/retryable-reads/tests/unified/handshakeError.yml
nbbeeken: @ShaneHarvey  caught the issue with the failPoint creating a usable connection, if we also make dropCollection fail and hit that we can have the usable connection get discarded. Good point, I'll update to use find

https://github.com/mongodb/specifications/pull/1115#discussion_r782642103

source/transactions/tests/unified/retryable-abort-handshake.yml
nbbeeken: done!

source/transactions/tests/unified/retryable-abort-handshake.yml
nbbeeken: Maybe I'm also misunderstanding something 🤔  I definitely think I have something wrong here with the failPoint, I want drivers to be able to successfully checkout a connection and complete the startTransaction and insertOne operation, but abortTransaction needs to have a "network error" (caused by the closeConnection: true setting) one time. 
So I've updated the test to set the failPoint after the insertOne now. I unfortunately cannot run these tests with the node driver due to a fundamental difference in how we connect a client to MongoDB, so I'm doing my best to logic these out and asking Shane to run them against python.

source/retryable-reads/retryable-reads.rst
nbbeeken: done!

source/retryable-writes/retryable-writes.rst
nbbeeken: done!

source/transactions/transactions.rst
nbbeeken: done! and patch sgtm

source/retryable-reads/tests/unified/handshakeError.yml
nbbeeken: Sorry I missed this comment in my initial responses.

That sounds like it may work, but if we set the failPoint to hello, wouldn't serverSelection fail? monitoring won't be able to identify the topology, or at least it will clear the first failpoint if we use the "times: 1" method. And it wouldn't make it to the retryablity section of executing an operation. Maybe there's a different order of operations I'm not considering?

source/retryable-reads/retryable-reads.rst
jmikola: I'm probably thinking of tooling used to build the MongoDB manual. I don't believe anything in the specs repo checks for this.

See: https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#sections:

> When an overline is used, the length and character used must match the underline.

source/retryable-reads/retryable-reads.rst
jmikola: The diff is now entirely different, which I assume is the result of rebasing/merging atop the CSOT merge. Is `executeRetry` still the correct function to call here? I don't see it anywhere else in the code example, so I'm curious if CSOT renamed `executeRetry` to `executeRetryableRead`.

If you're confident that this line is correct feel free to just resolve the comment. I don't think I'm going to be able to wrap my head around this at the moment.

source/retryable-reads/tests/unified/handshakeError.yml
jmikola: The reasoning for `dropCollection` sounds like it's worth a YAML comment for future reference.

source/retryable-reads/tests/unified/handshakeError.yml
jmikola: Feel free to resolve this after you file the separate DRIVERS ticket (if you'd like to use this thread as a reminder).

source/retryable-reads/tests/unified/handshakeError.yml
jmikola: > but if we set the failPoint to hello, wouldn't serverSelection fail?

That makes sense. Forget I suggested that. I was recently using `hello` failpoints (albeit for SDAM monitoring tests), which is probably why that was on my mind.

I reckon that failing on `hello` would actually be much worse for single-threaded clients due to [`cooldownMS`](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-monitoring.rst#cooldownms) blocking for five seconds. I don't think a failure on `saslContinue` would trigger that.

Apologies for wasting your time here. I probably have no business reviewing anything with CMAP in it.

source/transactions/tests/unified/retryable-abort-handshake.yml
```diff
@@ -0,0 +1,92 @@
...
+
+      - name: startTransaction
+        object: *session0
+
+      - name: insertOne
+        object: *collection0
+        arguments:
+          document: { _id: 2, x: 22 }
+
+      - name: failPoint # fail the next connection establishment
```
jmikola, line 47: Why does this refer to "next connection". Is the subsequent `abortTransaction` going to check a new connection out of the pool?

I read through [Checking Out a Connection](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#id32) in the CMAP spec but it's still not clear to me how the pool orders its connections for re-use.

source/transactions/tests/unified/retryable-abort-handshake.yml
```diff
@@ -0,0 +1,92 @@
...
+        object: testRunner
+        arguments:
+          client: *client0
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+              failCommands: [saslContinue]
```
jmikola, line 55: No harm in copying over the comment about why `saslContinue` is used here (to avoid errors during SDAM but still apply to handshakes for new connections)

source/retryable-writes/retryable-writes.rst
```diff
@@ -514,13 +518,16 @@ Consider the following pseudo-code:
...
-    }    
+    }
   }
 
 
 ``handleError`` in the above pseudocode refers to the function defined in the
 `Error handling pseudocode`_ section of the SDAM specification.
 
+Handshake errors which are considered retryable will be handled by the SDAM logic causing an eventual
+server selection timeout. TODO DO NOT MERGE!!!
```
jmikola, line 529: Is this referencing a comment from @ShaneHarvey's previous review?

nbbeeken: Yes I meant to leave a comment on here before logging off yesterday, in the CSOT channel there was discussion of reverting the new retryable changes. So this sentence is in place of my pseudo code changes which got removed with the merge. This is allegedly how handshake errors would be retried with the new paradigm, SDAM would be the exit condition to the indefinite loop, but the test scenarios we've come up with here, are specifically demonstrating the scenario where hellos for the monitor are succeeding, but the handshake for operations aren't. So either handshakes have to be special case like the socketTimeout errors are, or we reconsider how the loop is managed.

source/retryable-reads/tests/unified/handshakeError.yml
nbbeeken: Created: https://jira.mongodb.org/browse/DRIVERS-2043

source/retryable-reads/retryable-reads.rst
nbbeeken: Ah! the actual underline length should match, I didn't connect the two for some reason, fixed!

source/retryable-reads/tests/unified/handshakeError.yml
nbbeeken: Grasping the tricks we're using here to test took me a min since it isn't very direct, so fresh eyes/perspective is always welcome!

source/transactions/tests/unified/retryable-abort-handshake.yml
nbbeeken: Good Q here, my phrasing is strange, I need to give these transaction tests a bit more thought as opposed to the normal retryable ones. The other tests rely on connection pools being started with zero connections, so we can bank on the first operation being the one to perform the handshake. Which is why we needed to have the pool drop the connection created by the failPoint so we get back down to zero connections. 

With transaction tests, I need to think about how I can engineer the scenario where the pool is emptied before we run an abort/commit so that those operations would run a handshake. In reality I suppose this could happen due to some networking errors that reset the pool(s) before the abort/commit is run. 🤔

source/transactions/tests/unified/retryable-abort-handshake.yml
nbbeeken: done!

source/retryable-reads/retryable-reads.rst
nbbeeken: Ok now the diff is back to what it was before the retry change. I think the tricky part about changing the pseudo code is that it's not incorrect to leave exactly how it's written now. Prior to now, drivers may have taken a poolClearedError caused by a network error during handshake and retried it because its not explcitly disallowed. I'm trying to call it out in code here as a new case along with the prose change, so even adding a flag might be confusing too.

Maybe we leave the code as is, but there could be a comment in `executeRetry` that mentions originalError might be a network error. The prose already describe this as a new case that is valid to retry, so drivers have to double check the logic that handles their pool cleared scenarios doesn't avoid retrying pool clearing caused by network issues.

source/retryable-writes/retryable-writes.rst
nbbeeken: I've updated the transaction tests now, Shane cleverly suggested using multiple sessions to avoid killing the transaction before we're able to run abort or commit. Now the dropCollections trick should empty the pool right before those two final operations.

source/transactions/tests/unified/retryable-abort-handshake.yml
nbbeeken: I've updated the transaction tests now, Shane cleverly suggested using multiple sessions to avoid killing the transaction before we're able to run abort or commit. Now the dropCollections trick should empty the pool right before those two final operations.

source/retryable-writes/retryable-writes.rst
nbbeeken: This is removed now, no longer relevant due to the revert.

source/retryable-reads/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,76 @@
...
+              # this failPoint itself will create a usable connection in the connection pool
+              # so we run a dropCollection (that also fails) in order to discard the connection
+              # before testing our read operation "find"
+              failCommands: [saslContinue, dropCollection]
+              closeConnection: true
+
+      - name: dropCollection
+        object: *database0
+        arguments:
+          collection: doesNotMatter
```
jmikola, line 54: If `dropCollection` doesn't actually matter, would it be preferable to use something more benign like a `ping` command?

source/retryable-writes/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,77 @@
...
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+              failCommands: [dropCollection, saslContinue]
+              closeConnection: true
+
+      - name: dropCollection
+        object: *database0
+        arguments:
+          collection: doesNotMatter
```
jmikola, line 48: See similar comment as I brought up in `retryable-reads/tests/unified/handshakeError.yml`.

source/retryable-reads/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,76 @@
...
+  - minServerVersion: "4.2"
+    topologies: [replicaset, sharded, load-balanced]
+    auth: true
+
+createEntities:
+  - client:
+      id: &client0 client0
+      useMultipleMongoses: false
+      observeEvents: [commandStartedEvent, connectionCheckOutStartedEvent]
+      uriOptions: { retryReads: true, retryWrites: true, maxPoolSize: 5 }
```
jmikola, line 15: Are `retryReads` and `retryWrites` relevant if they both already default to `true`?

Also, apologies if I asked this before but is there a significance to `maxPoolSize: 5`? Would 2 make more sense?

source/retryable-writes/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,77 @@
...
+  - minServerVersion: "4.2"
+    topologies: [replicaset, sharded, load-balanced]
+    auth: true
+
+createEntities:
+  - client:
+      id: &client0 client0
+      useMultipleMongoses: false
+      observeEvents: [commandStartedEvent, connectionCheckOutStartedEvent]
+      uriOptions: { retryReads: true, retryWrites: true, maxPoolSize: 5 }
```
jmikola, line 15: Similar comment as I left in the retryable reads test about the default valuse and `maxPoolSize`.

source/transactions/tests/unified/retryable-commit-handshake.yml
```diff
@@ -0,0 +1,111 @@
...
+    topologies: [replicaset, sharded, load-balanced]
+    serverless: "forbid"
+    auth: true
+
+createEntities:
+  - client:
+      id: &client0 client0
+      useMultipleMongoses: false
+      observeEvents: [commandStartedEvent, connectionCheckOutStartedEvent]
+      uriOptions: { retryReads: true, retryWrites: true, maxPoolSize: 5 }
```
jmikola, line 16: Since the related spec change says `commitTransaction` is retryable "regardless of whether retryWrites is set on the MongoClient or not", would it make sense to explicitly specify `retryWrites: false` here?

Beyond that, my earlier comment about default values and `maxPoolSize` applies here.

jmikola (commented):
Just a few comments that may not require any changes. I think I've provided about all of the feedback I can on this one, so I'll defer to Shane for a complete LGTM.

source/retryable-writes/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,77 @@
...
+tests:
+  - description: "InsertOne succeeds after retryable handshake error"
+    operations:
+      - name: failPoint # fail the next connection establishment
+        object: testRunner
+        arguments:
+          client: *client0
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 1 }
```
ShaneHarvey, line 40: This needs to be times: 2 otherwise the failpoint will be disabled before saslContinue.

source/retryable-reads/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,76 @@
...
+          client: *client0
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 2 }
+            data:
+              # use saslContinue here to avoid SDAM errors
+              # this failPoint itself will create a usable connection in the connection pool
+              # so we run a dropCollection (that also fails) in order to discard the connection
+              # before testing our read operation "find"
+              failCommands: [saslContinue, dropCollection]
```
ShaneHarvey, line 48: dropCollection -> drop

source/transactions/tests/unified/retryable-abort-handshake.yml
```diff
@@ -0,0 +1,113 @@
...
+          session: *session1
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 2 }
+            data:
+              # use saslContinue here to avoid SDAM errors
+              # this failPoint itself will create a usable connection in the connection pool
+              # so we run a dropCollection (that also fails) in order to discard the connection
+              # before testing that abortTransaction gets retried
+              failCommands: [saslContinue, dropCollection]
```
ShaneHarvey, line 68: dropCollection -> drop

source/transactions/tests/unified/retryable-commit-handshake.yml
```diff
@@ -0,0 +1,111 @@
...
+          session: *session1
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 2 }
+            data:
+              # use saslContinue here to avoid SDAM errors
+              # this failPoint itself will create a usable connection in the connection pool
+              # so we run a dropCollection (that also fails) in order to discard the connection
+              # before testing that commitTransaction gets retried
+              failCommands: [saslContinue, dropCollection]
```
ShaneHarvey, line 66: dropCollection -> drop

source/retryable-writes/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,77 @@
...
+            data:
+              failCommands: [dropCollection, saslContinue]
```
ShaneHarvey, line 42: dropCollection -> drop

source/retryable-reads/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,76 @@
...
+
+    expectEvents:
+      - client: *client0
+        eventType: cmap
+        events:
+          - { connectionCheckOutStartedEvent: {} }
+          - { connectionCheckOutStartedEvent: {} }
+      - client: *client0
+        events:
+          - commandStartedEvent:
```
ShaneHarvey, line 72: Missing the drop event.

source/retryable-writes/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,77 @@
...
+
+    expectEvents:
+      - client: *client0
+        eventType: cmap
+        events:
+          - { connectionCheckOutStartedEvent: {} }
+          - { connectionCheckOutStartedEvent: {} }
+      - client: *client0
+        events:
+          - commandStartedEvent:
```
ShaneHarvey, line 65: Missing the drop event.

source/transactions/tests/unified/retryable-abort-handshake.yml
```diff
@@ -0,0 +1,113 @@
...
+        eventType: cmap
+        events:
+          - { connectionCheckOutStartedEvent: {} } # startTransaction
+          - { connectionCheckOutStartedEvent: {} } # insertOne
+          - { connectionCheckOutStartedEvent: {} } # failPoint
+          - { connectionCheckOutStartedEvent: {} } # abortTransaction
+          - { connectionCheckOutStartedEvent: {} } # abortTransaction retry
+      - client: *client0
+        events:
+          - commandStartedEvent:
```
ShaneHarvey, line 93: Missing the drop event.

source/transactions/tests/unified/retryable-commit-handshake.yml
```diff
@@ -0,0 +1,111 @@
...
+        eventType: cmap
+        events:
+          - { connectionCheckOutStartedEvent: {} } # startTransaction
+          - { connectionCheckOutStartedEvent: {} } # insertOne
+          - { connectionCheckOutStartedEvent: {} } # failPoint
+          - { connectionCheckOutStartedEvent: {} } # commitTransaction
+          - { connectionCheckOutStartedEvent: {} } # commitTransaction retry
+      - client: *client0
+        events:
+          - commandStartedEvent:
```
ShaneHarvey, line 91: Missing the drop event.

ShaneHarvey (changes requested):

source/retryable-reads/tests/unified/handshakeError.yml
nbbeeken: @ShaneHarvey Switching to ping seems like it would be a fine equivalent, anything specific about drop's behavior? we just need to run something that will have the connection get shut upon running it

source/transactions/tests/unified/retryable-commit-handshake.yml
nbbeeken: I agree with the retryWrites false, done!

source/retryable-reads/tests/unified/handshakeError.yml
ShaneHarvey: Yes it can be any command the ends of clearing the pool. Switching to ping SGTM.

source/retryable-reads/retryable-reads.rst
jmikola: @ShaneHarvey: Just leaving this open for you to acknowledge (since it pertains more to CMAP). Feel free to resolve this thread once you've read it if you have no feedback.

ShaneHarvey: I don't see any changes to the pseudocode at all in this PR anymore. Is that expected?

nbbeeken: Yep, thanks for the ping on this Jeremy, so related to my comment above, I can bring back the explicit catch block but it can be confusing, a handshake failure is a "poolClearedError" so in pseudo code how best do we want to capture this new scenario? With the prose changes we've defined handshake failures as retryable and the previous form of the spec did not explicitly disallow retrying these types of failures so the code remains "correct".

If you've been taking a bit of a stab at doing this in python how does it literally look in code, do you have a flag on your error or is there some hierarchy of instanceof that captures the new case?

ShaneHarvey: A handshake failure != "poolClearedError". From my understanding poolClearedError is a very specific exception type that's only thrown when a checkout is attempted from an already paused pool. A handshake failure on the other hand is any network/command error that occurs when establishing a new connection. In fact I'd argue that a poolClearedError is one type of handshake failure.

In Python our retry logic considers and retries connection checkout errors as if the error occurred while actually running the command, eg:

```python
try:
    with pool.checkout() as connection:  # This line may raise a retryable handshake error
        executeOperation(connection)  # This line may also raise a retryable error
except Exception as exc:
    if retryable(exc):
        retry()
```

nbbeeken: You can take a look at what I pushed but just for ease I'll paste here:
```java
    try {
      connection = server.getConnection()
    } catch (PoolClearedException poolClearedError) {
      ...
    } catch (NetworkError networkError) {
      /* NetworkError indicates the getConnection was not able to establish
       * a successfully handshake-ed connection with the server, this is retryable.
       * We do not need to update SDAM, since network errors already mark
       * servers Unknown and empty the pool. */
      return executeRetry(command, session, poolClearedError);
    }
```
Hopefully the code comment addresses the original concern of the thread around lack of clarity, if, correct me if I'm wrong, what I've written here is accurate

source/retryable-reads/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,79 @@
...
+              command:
+                find: *collectionName
+                filter: { _id: 2 }
+              databaseName: *databaseName
+          - commandStartedEvent:
+              command:
+                ping: 1
```
ShaneHarvey, line 78: The ping command needs to come first.

source/retryable-reads/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,79 @@
...
+            data:
+              # use saslContinue here to avoid SDAM errors
+              # this failPoint itself will create a usable connection in the connection pool
+              # so we run a dropCollection (that also fails) in order to discard the connection
+              # before testing our read operation "find"
+              failCommands: [saslContinue, ping]
+              closeConnection: true
+
+      - name: runCommand
+        object: *database0
```
ShaneHarvey, line 51: Missing the "commandName: ping" field.

source/retryable-writes/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,80 @@
...
+        arguments:
+          client: *client0
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 2 }
+            data:
+              failCommands: [saslContinue, ping]
+              closeConnection: true
+
+      - name: dropCollection
```
ShaneHarvey, line 44: This needs to be replaced with ping.

source/retryable-writes/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,80 @@
...
+        events:
+          - commandStartedEvent:
+              command:
+                insert: *collectionName
+                documents: [{ _id: 2, x: 22 }]
+              commandName: insert
+              databaseName: *databaseName
+          - commandStartedEvent:
+              command:
+                ping: 1
```
ShaneHarvey, line 72: The ping command needs to come first.

source/transactions/tests/unified/retryable-abort-handshake.yml
```diff
@@ -0,0 +1,116 @@
...
+          session: *session1
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 2 }
+            data:
+              # use saslContinue here to avoid SDAM errors
+              # this failPoint itself will create a usable connection in the connection pool
+              # so we run a dropCollection (that also fails) in order to discard the connection
+              # before testing that abortTransaction gets retried
+              failCommands: [saslContinue, drop]
```
ShaneHarvey, line 67: drop -> ping.

source/transactions/tests/unified/retryable-abort-handshake.yml
```diff
@@ -0,0 +1,116 @@
...
+        arguments:
+          client: *client0
+          session: *session1
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 2 }
+            data:
+              # use saslContinue here to avoid SDAM errors
+              # this failPoint itself will create a usable connection in the connection pool
+              # so we run a dropCollection (that also fails) in order to discard the connection
```
ShaneHarvey, line 65: These comments are now outdated.

source/transactions/tests/unified/retryable-abort-handshake.yml
```diff
@@ -0,0 +1,116 @@
...
+              # before testing that abortTransaction gets retried
+              failCommands: [saslContinue, drop]
+              closeConnection: true
+
+      - name: runCommand
+        object: *database0
```
ShaneHarvey, line 71: Also missing the commandName field.

source/transactions/tests/unified/retryable-commit-handshake.yml
```diff
@@ -0,0 +1,115 @@
...
+            mode: { times: 2 }
+            data:
+              # use saslContinue here to avoid SDAM errors
+              # this failPoint itself will create a usable connection in the connection pool
+              # so we run a dropCollection (that also fails) in order to discard the connection
+              # before testing that commitTransaction gets retried
+              failCommands: [saslContinue, ping]
+              closeConnection: true
+
+      - name: ping
```
ShaneHarvey, line 69: ping -> runCommand, also missing commandName.

source/retryable-reads/retryable-reads.rst
```diff
@@ -163,6 +163,7 @@ Drivers SHOULD support retryability for the following operations:
 - Any driver that provides generic command runners for read commands (with logic
   to inherit a client-level read concerns) SHOULD implement retryability for the
   read-only command runner.
+- Any error that occurs during a handshake (eg. Auth, ShutdownInProgress).
```
ShaneHarvey, line 166: I'm not sure this is the right section of the spec for this prose. This section says "Drivers SHOULD support retryability for the following operations" whereas this spec change is defining a new type of retryable error that MUST be retried. What if we add this to the "Retryable Error" definition:
```
...
- a `PoolClearedError`_
- Any of the above retryable errors that occur during a connection handshake (including the
  authentication step). For example, a network error or ShutdownInProgress error
  encountered when running the hello or saslContinue commands.
```

source/retryable-reads/retryable-reads.rst
```diff
@@ -357,16 +358,22 @@ and reflects the flow described above.
...
+      /* PoolClearedException indicates the operation did not even attempt to
+       * create a connection, let alone execute the operation. This means we
+       * are always safe to attempt a retry. We do not need to update SDAM,
+       * since whatever error caused the pool to be cleared will do so itself. */
+      return executeRetry(command, session, poolClearedError);
+    } catch (NetworkError networkError) {
+      /* NetworkError indicates the getConnection was not able to establish
+       * a successfully handshake-ed connection with the server, this is retryable.
+       * We do not need to update SDAM, since network errors already mark
+       * servers Unknown and empty the pool. */
```
ShaneHarvey, line 373: New thread because Github is not letting me reply on the existing one. Since the executeCommand case below includes updateTopologyDescriptionForNetworkError+updateTopologyDescriptionForNotWritablePrimaryError I think we should do the same here:
```
    try {
      connection = server.getConnection()
    } catch (PoolClearedException poolClearedError) {
      /* PoolClearedException indicates the operation did not even attempt to
       * create a connection, let alone execute the operation. This means we
       * are always safe to attempt a retry. We do not need to update SDAM,
       * since whatever error caused the pool to be cleared will do so itself. */
      return executeRetry(command, session, poolClearedError);
    } catch (NetworkError networkError) {
      updateTopologyDescriptionForNetworkError(server, networkError);
      return executeRetry(command, session, poolClearedError);
    } catch (NotWritablePrimaryException originalError) {
      updateTopologyDescriptionForNotWritablePrimaryError(server, originalError);
      return executeRetry(command, session, poolClearedError);
    }
```

Speaking of, is it possible to add test coverage for a NotWritablePrimaryException? Right now I think we're only testing NetworkError.

ShaneHarvey (changes requested):

source/retryable-reads/retryable-reads.rst
nbbeeken: Updaeted the code. Yes adding that coverage sounds reasonable, I'll post something Monday

source/retryable-reads/retryable-reads.rst
nbbeeken: This was still left over from when I started this spec change as a should/may 😅 good catch, moved!

source/retryable-reads/retryable-reads.rst
nbbeeken: Pushed tests for NotWritablePrimary errors, I was able to test these against node and they pass 🚀

source/retryable-reads/retryable-reads.rst
```diff
@@ -367,10 +370,10 @@ and reflects the flow described above.
...
       return executeRetry(command, session, poolClearedError);
     } catch (NetworkError networkError) {
-      /* NetworkError indicates the getConnection was not able to establish
-       * a successfully handshake-ed connection with the server, this is retryable.
-       * We do not need to update SDAM, since network errors already mark
-       * servers Unknown and empty the pool. */
+      updateTopologyDescriptionForNetworkError(server, networkError);
+      return executeRetry(command, session, poolClearedError);
+    } catch (NotWritablePrimaryException originalError) {
+      updateTopologyDescriptionForNotWritablePrimaryError(server, originalError);
```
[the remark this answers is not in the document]
nbbeeken: The indentation looks right to me, but I've fixed the comment formatting that might make it look more correct. And fixed the issue with the wrong error.

source/retryable-writes/tests/unified/notWritablePrimaryException.yml
```diff
@@ -0,0 +1,70 @@
...
+        object: testRunner
+        arguments:
+          client: *client0
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+              failCommands: [insertOne]
+              errmsg: "not master"
+              errorCode: 1
```
jmikola: I don't recall seeing this in a previous iteration of the PR.

Are drivers still in the business of checking error strings for "not master"? `errorCode: 1` looks odd.

I suppose this has to do with [Use error messages to detect "not master" and "node is recovering"](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#id114)](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#use-error-messages-to-detect-not-master-and-node-is-recovering), but that says "When error codes are not available" so isn't there an error code we could/should use here instead of relying on string detection? As-is, I might wonder if a driver would first inspect `errorCode` and then design not to bother checking `errmsg`.

Also, I'll note that `errmsg` is not documented in either [failCommand parameters](https://github.com/mongodb/mongo/wiki/The-%22failCommand%22-fail-point#failcommand-parameters) or the equivalent docs in the [Unified Test Format spec](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#failcommand). Is that actually a supported `data` option? If so, should we update both the wiki and unified spec to discuss this?

nbbeeken: I pushed the new tests across two commits, the first one was a mistake, so the error code is now 10107 if you look at the latest. I agree we should drop the errmsg from the test b/c indeed the code should take precedence, I've done so. 
After a quick test, it looks like it is not a supported param in that it is ignored.

source/retryable-reads/tests/unified/notWritablePrimaryException.yml
```diff
@@ -0,0 +1,70 @@
...
+      - name: failPoint # fail the next connection establishment
+        object: testRunner
+        arguments:
+          client: *client0
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+              failCommands: [find]
+              errorCode: 10107
```
jmikola, line 43:
```suggestion
              errorCode: 10107 # NotWritablePrimary
```

source/retryable-writes/tests/unified/notWritablePrimaryException.yml
```diff
@@ -0,0 +1,69 @@
...
+      - name: failPoint # fail the next connection establishment
+        object: testRunner
+        arguments:
+          client: *client0
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+              failCommands: [insertOne]
+              errorCode: 10107
```
jmikola, line 41:
```suggestion
              errorCode: 10107 # NotWritablePrimary
```

jmikola (commented):
Suggested comments for error code literals, as is done in other spec tests.

source/transactions/tests/unified/retryable-abort-handshake.yml
```diff
@@ -0,0 +1,117 @@
...
+
+      - name: startTransaction
+        object: *session0
+
+      - name: insertOne
+        object: *collection0
+        arguments:
+          document: { _id: 2, x: 22 }
+
+      # The following failPoint and dropCollection utilize session1 so that
```
ShaneHarvey, line 51: dropCollection -> ping

source/transactions/tests/unified/retryable-abort-handshake.yml
```diff
@@ -0,0 +1,117 @@
...
+  - description: "AbortTransaction succeeds after handshake network error"
+    operations:
+
+      - name: startTransaction
+        object: *session0
+
+      - name: insertOne
+        object: *collection0
+        arguments:
+          document: { _id: 2, x: 22 }
```
ShaneHarvey, line 49: Missing session0.

source/transactions/tests/unified/retryable-commit-handshake.yml
```diff
@@ -0,0 +1,116 @@
...
+  - description: "CommitTransaction succeeds after handshake network error"
+    operations:
+
+      - name: startTransaction
+        object: *session0
+
+      - name: insertOne
+        object: *collection0
+        arguments:
+          document: { _id: 2, x: 22 }
```
ShaneHarvey, line 48: missing session0

source/transactions/tests/unified/retryable-commit-handshake.yml
```diff
@@ -0,0 +1,116 @@
...
+              commandName: insert
+              databaseName: *databaseName
+          - commandStartedEvent:
+              command:
+                ping: 1
+              databaseName: *databaseName
+          - commandStartedEvent:
+              command:
+                commitTransaction: 1
+                lsid: *session0
```
ShaneHarvey, line 106: Should be `$$sessionLsid: *session0`

source/transactions/tests/unified/retryable-commit-handshake.yml
```diff
@@ -0,0 +1,116 @@
...
+                startTransaction: true
```
ShaneHarvey, line 107: remove startTransaction

source/transactions/tests/unified/retryable-abort-handshake.yml
```diff
@@ -0,0 +1,117 @@
...
+              databaseName: *databaseName
+          - commandStartedEvent:
+              command:
+                ping: 1
+              databaseName: *databaseName
+          - commandStartedEvent:
+              command:
+                abortTransaction: 1
+                lsid: *session0
+                startTransaction: true
```
ShaneHarvey, line 108: Remove startTransaction

source/transactions/tests/unified/retryable-abort-handshake.yml
```diff
@@ -0,0 +1,117 @@
...
+              commandName: insert
+              databaseName: *databaseName
+          - commandStartedEvent:
+              command:
+                ping: 1
+              databaseName: *databaseName
+          - commandStartedEvent:
+              command:
+                abortTransaction: 1
+                lsid: *session0
```
ShaneHarvey, line 107: Should be `$$sessionLsid: *session0`

source/transactions/tests/unified/retryable-abort-handshake.yml
```diff
@@ -0,0 +1,117 @@
...
+                startTransaction: true
+              commandName: abortTransaction
+              databaseName: admin
+
+    outcome:
+      - collectionName: *collectionName
+        databaseName: *databaseName
+        documents:
+          - { _id: 1, x: 11 }
+          - { _id: 2, x: 22 } # The write was still applied
```
ShaneHarvey, line 117: Remove, the write is not applied.

source/retryable-reads/tests/unified/handshakeError.yml
```diff
@@ -0,0 +1,80 @@
...
+        object: testRunner
+        arguments:
+          client: *client0
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 2 }
+            data:
+              # use saslContinue here to avoid SDAM errors
+              # this failPoint itself will create a usable connection in the connection pool
+              # so we run a dropCollection (that also fails) in order to discard the connection
```
ShaneHarvey, line 45: dropCollection -> ping. There are a few other references in other tests too. Can you do grep though and change them too?

ShaneHarvey (changes requested):

source/retryable-reads/tests/unified/notWritablePrimaryException.yml
```diff
@@ -0,0 +1,70 @@
...
+    operations:
+      - name: failPoint # fail the next connection establishment
+        object: testRunner
+        arguments:
+          client: *client0
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+              failCommands: [find]
```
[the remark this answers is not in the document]
ShaneHarvey: Fixed.

ShaneHarvey (commented):
Fixed some more issues in the tests and it's passing locally. LGTM if this patch succeeds: https://spruce.mongodb.com/version/620efd600ae60613663e7e89/tasks

ShaneHarvey:
Interesting, in order to get the transaction tests to pass on sharded clusters I had to implement https://jira.mongodb.org/browse/DRIVERS-2032 first. The problem is that the new test goes through the exact scenario mentioned in DRIVERS-2032. In python my fix is to make pinned server selection block until the server is readable again (or a selection timeout): https://github.com/mongodb/mongo-python-driver/pull/846/commits/c79a1f39d587de3221de29987c93a678cf0ab792

New patch: https://spruce.mongodb.com/version/620f0d267742ae451558f682/

ShaneHarvey (approved):
LGTM!

durran (approved):
LGTM on TXN changes.

PR #1116, opened 2022-01-10
DRIVERS-1873 Use OP_MSG to authenticate if the server supports OP_MSG

juliusgeo:


source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -453,13 +453,16 @@ Establishing a Connection (Internal Implementation)
 
 Before a `Connection <#connection>`_ can be marked as either "available" or "in use", it
 must be established. This process involves performing the initial
-handshake, handling OP_COMPRESSED, and performing authentication.
+handshake, inspecting the value of ``maxWireVersion`` (to decide whether to
+use ``OP_MSG``(>=3.6) or ``OP_QUERY``(<3.6) for future communications),
```
behackett, line 457: Here and in the change in OP_MSG you should say what the relevant wire version is.

juliusgeo: Done.

source/message/OP_MSG.rst
```diff
@@ -53,6 +53,11 @@ Usage
 ``OP_MSG`` is only available in MongoDB 3.6 (``maxWireVersion >= 6``) and later.
 MongoDB drivers SHOULD perform the MongoDB handshake using ``OP_MSG`` if an API
 version was declared on the client, but MAY decide to use ``OP_QUERY``.
+However, once the driver has performed the MongoDB handshake, and it has
+inspected the value of ``maxWireVersion``, it must use either ``OP_MSG``
+(``maxWireVersion >= 6``) or ``OP_QUERY``(``maxWireVersion < 6``)
+respectively. Starting in MongoDB 6.0, using ``OP_QUERY`` for anything other
+than the MongoDB handshake will result in an error.
```
ShaneHarvey, line 60: This seems to be restating this existing rule:
> If the node supports OP_MSG, any and all messages MUST use OP_MSG, optionally compressed with OP_COMPRESSED.

What do you think about mentioning the auth commands specifically in that line instead?

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -453,13 +453,21 @@ Establishing a Connection (Internal Implementation)
...
 must be established. This process involves performing the initial
-handshake, handling OP_COMPRESSED, and performing authentication.
+handshake, inspecting the value of ``maxWireVersion`` (to decide whether to
+use ``OP_MSG``(``maxWireVersion >= 6``) or ``OP_QUERY``(``maxWireVersion <6``)
+for future communications), handling OP_COMPRESSED, and performing
+authentication. It is necessary to inspect the value of ``maxWireVersion``
+because starting with MongoDB 6.0 using ``OP_QUERY`` for anything other than
+the handshake will result in an error. Once the driver has performed the
+handshake and inspected the value of ``maxWireVersion`` it MUST use
+``OP_MSG`` if supported.
```
ShaneHarvey, line 463: Should we link to the section in the OP_MSG spec instead of duplicating it? Alternatively, I would be more open to duplicating here once we make it more succinct (following my comment in OP_MSG). Or could we just remove this section altogether?

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -453,13 +453,21 @@ Establishing a Connection (Internal Implementation)
...
 
 .. code::
 
     try:
       connect connection via TCP / TLS
       perform connection handshake
+      inspect "maxWireVersion"
```
ShaneHarvey, line 470: I propose we clarify the pseudocode like this rather than just "inspect maxWireVersion":
```
      handle OP_COMPRESSED
      if maxWireVersion >= 6:  # MongoDB 3.6+
          perform connection authentication via OP_MSG
      else:
          perform connection authentication via OP_QUERY
```

ShaneHarvey (changes requested):

source/message/OP_MSG.rst
juliusgeo: Done.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
juliusgeo: Done.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
juliusgeo: I reduced it down significantly because the pseudocode offers some of the details

ShaneHarvey (approved):
LGTM, I'm not sure there are any tests we can add with this change. I assume teams will just need to audit their drivers to fix this issue.

patrickfreed:
I think this requirement is better suited for the auth spec rather than CMAP, since CMAP is more about managing the connections rather than any connection-related implementation details. I think [the section about the auth handshake](https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#authentication-handshake) seems like a good option.

patrickfreed (approved):

source/auth/auth.rst
```diff
@@ -171,6 +171,10 @@ handshake:
 #. If the server is of type RSArbiter, no authentication is possible and the
    handshake is complete.
 
+#. Inspect the value of ``maxWireVersion``. If the value is greater than or
+   equal to ``6``, then the driver MUST use ``OP_MSG`` for authentication.
+   Otherwise, it MUST use ``OP_QUERY``.
+
```
ShaneHarvey, line 177: Could you update the "Spec Version" to 1.10.5, bump the "last modified date", and add a changelog entry at the bottom?

ShaneHarvey (changes requested):
One last thing.

source/auth/auth.rst
juliusgeo: Done.

ShaneHarvey (approved):

source/message/OP_MSG.rst
```diff
@@ -60,7 +60,8 @@ node supports ``OP_MSG``. Drivers that have only ever supported MongoDB 3.6 and
 newer MAY default to using ``OP_MSG``.
 
 If the node supports ``OP_MSG``, any and all messages MUST use ``OP_MSG``,
-optionally compressed with ``OP_COMPRESSED``.
+optionally compressed with ``OP_COMPRESSED``. This includes authentication
+messages.
```
JamesKovacs, line 64: This is ambiguous. One possible reading is that authentication messages must use `OP_MSG` (which is true) and optionally be compressed with `OP_COMPRESSED` (which is false). Authentication-related messages MUST NEVER be compressed. They were explicitly excluded from compression because of concerns about unintentionally introducing security issues. The following is a list of commands that must never be compressed used by the .NET/C# driver:

hello
isMaster
saslStart
saslContinue
getnonce
authenticate
createUser
updateUser
copydbsaslstart
copydbgetnonce
copydb

If we want to reconsider compressing these messages we can do that, but it seems outside the scope of this change.

JamesKovacs (changes requested):
Clarify that auth-related messages MUST use `OP_MSG` but MUST NOT use `OP_COMPRESSED`. See below for why.

source/message/OP_MSG.rst
juliusgeo: That makes sense. Fixed.

source/message/OP_MSG.rst
```diff
@@ -60,7 +60,8 @@ node supports ``OP_MSG``. Drivers that have only ever supported MongoDB 3.6 and
 newer MAY default to using ``OP_MSG``.
 
 If the node supports ``OP_MSG``, any and all messages MUST use ``OP_MSG``,
-optionally compressed with ``OP_COMPRESSED``.
+optionally compressed with ``OP_COMPRESSED``. Authentication messages MUST
+also use ``OP_MSG`` when it is supported, but MUST not use ``OP_COMPRESSED``.
```
ShaneHarvey, line 64: "MUST not" -> "MUST NOT"

See:
> The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in RFC 2119.

ShaneHarvey (changes requested):

ShaneHarvey (approved):

source/message/OP_MSG.rst
juliusgeo: My bad. Fixed.

JamesKovacs (approved):
LGTM

PR #1118, opened 2022-01-12
DRIVERS-2027 Add 'let' option to ReplaceOptions

comandeo:


source/crud/crud.rst
```diff
@@ -2156,6 +2168,7 @@ Q: Why are client-side errors raised for some unsupported options?
 Changes
 =======
 
+* 2022-01-12: Add let to ReplaceOptions
```
jmikola, line 2171: Don't forget to bump the last modified date at the top of the spec. Thankfully we don't need to worry about any version increment.

jmikola (approved):

PR #1122, opened 2022-01-18
DRIVERS-1954: SDAM should give priority to electionId over setVersion when updating topology

nbbeeken:
DRIVERS-1954

Along with updating the section that speaks about how to detect the stale primary, I reordered our mention of the two fields to help imply the precedence along with the explcit mention of the ordering.

Points of discussion:
- I removed the mention of "big-endian integer" comparision of ObjectIds, the byte order must remain the same on all hardware so memory comparisons are consistent
- I refactored the pseudo code's formatting, let me know if this impacts readability
- I added one test that will fail on drivers currently but will pass with a change to the precedence check
- I added a line where we record the maxElectionId if we haven't yet recorded one
- I had to modify one existing test to get it to pass with the new precedence.
- Should we test the case where an electionId hasn't been seen yet (so `maxElectionId == null`) but also `maxSetVersion == null` and a new electionId comes in independently of a setVersion? I don't think this is a real scenario, I believe both have to be either null or non nullish together.

Node POC: https://github.com/mongodb/node-mongodb-native/pull/3109


source/server-discovery-and-monitoring/tests/rs/use_setversion_without_electionid.yml
```diff
@@ -97,21 +97,21 @@ phases: [
         outcome: {
             servers: {
                 "a:27017": {
+                    type: "RSPrimary",
```
ShaneHarvey, line 100: If A is supposed to become the primary I think this test needs more work. The comment above says:
```
    # Delayed response from A, reporting its reelection. Its setVersion shows
    # the election preceded B's so we ignore it.
```

source/server-discovery-and-monitoring/tests/rs/use_setversion_without_electionid.yml
```diff
@@ -97,21 +97,21 @@ phases: [
...
+                    setName: "rs",
+                    setVersion: 1
```
ShaneHarvey, line 102: Is this intentionally missing electionId? I think it should be:
```
electionId: {"$oid": "000000000000000000000002"},
```

ShaneHarvey (changes requested):
I'm going to ask someone on the server team to review this as well.

source/server-discovery-and-monitoring/tests/rs/use_setversion_without_electionid.yml
nbbeeken: Updated the comments to clarify what the test should be getting at

source/server-discovery-and-monitoring/tests/rs/use_setversion_without_electionid.yml
nbbeeken: Yep that should be there, added!

ShaneHarvey:
@jasonjhchan Could you also help us answer this question?:
> Should we test the case where an electionId hasn't been seen yet (so maxElectionId == null) but also maxSetVersion == null and a new electionId comes in independently of a setVersion? I don't think this is a real scenario, I believe both have to be either null or non nullish together.

jasonjhchan (approved):

nbbeeken:
**Do not merge**: Acceptance criteria, [replicating this server jstest in node](https://github.com/mongodb/mongo/blob/59e19dc9bff797fc1cff45d56942e5e61da36f1a/jstests/sharding/reconfig_race_with_failover.js#L2)

nbbeeken:
Synced SDAM test:
https://github.com/mongodb/mongo/blob/565f818c1277998b9712fe2c9fc8c6d6158df07b/src/mongo/client/sdam/json_tests/sdam_tests/rs/set_version_can_rollback.json

Outstanding Q:
>  I believe the maxElectionId/maxSetVersion changes the PR outlines are inconsistent with what the server does. My guess is that the server tracks the max (electionId, setVersion) tuple whereas the spec says to track each max value independently.

@shuvalov-mdb Any insight here? Drivers currently will not roll back the `maxSetVersion` as this new test expects on line 46 vs line 94

shuvalov-mdb:
Hi, it's more a question do drivers actually care if the `maxSetVersion` rolls back? Yes, this is a valid scenario that RSM in server now supports, because server cares about maxSetVersion in its logic. But for drivers, you only care if the set of servers you track is current, as long as you can discover the primary. Do you need any other scenario?

To clarify: in servers the `electionId` wins, the compare ignores `maxSetVersion` if electionId is different, which means there is a real scenario when electionId increments and maxSetVersion decrements. The bugfix I did was to support the maxSetVersion going backwards, it never did before.

shuvalov-mdb:
For the record - why maxSetVersion can rollback: a command that affects the set version can return before the state is replicated and increments the set version at the caller. If the primary fails over before the new state is replicated the new primary will have greater term (electionId) but old set version.

ShaneHarvey:
@shuvalov-mdb drivers also utilize maxSetVersion to detect stale primaries (see the pseudocode changes in this PR) . Are you suggesting that we only need to track maxElectionId and ignore maxSetVersion/setVersion altogether?

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -2420,6 +2438,17 @@ Why is auto-discovery the preferred default?
...
 
+Why is it possible for maxSetVersion to go down?
+''''''''''''''''''''''''''''''''''''''''''''''''
+
+``maxElectionId`` and ``maxSetVersion`` are actually considered a pair of values
+Drivers MAY consider implementing comparison in code as a tuple of the two to ensure their always updated together:
+
+::
+  New tuple                          old tuple
+  { electionId: 2, setVersion: 1 } > { electionId: 1, setVersion: 50 }
```
nbbeeken, line 2449: TODO fix rst

nbbeeken:
@ShaneHarvey This is RFAL, I replicated in pseudo code what I did in my driver I think the verbose variables are clearer than the deeply nested conditional, lmk what you think

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1101,30 +1102,27 @@ updateRSFromPrimary
...
         # for comparison rules.
-        if (topologyDescription.maxSetVersion is not null and
-            topologyDescription.maxElectionId is not null and (
-                topologyDescription.maxSetVersion > description.setVersion or (
-                    topologyDescription.maxSetVersion == description.setVersion and
-                    topologyDescription.maxElectionId > description.electionId
-                )
-            ):
 
+        # Null values are always considered "less than"
```
jyemin, line 1110: This comment seems unnecessary since null values are never compared.  There is a check on line 1105 for null in the description and the conditionals below guard against null for the max values.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1101,30 +1102,27 @@ updateRSFromPrimary
...
+        maxElectionIdIsGreater = topologyDescription.maxElectionId < serverDescription.electionId if topologyDescription.maxElectionId is not null else false
```
jyemin, line 1111: Can you double check these?  In a Java driver POC I ended up coding these as:

```
   maxElectionIdIsGreater = maxElectionId != null && maxElectionId > electionId
   maxElectionIdIsEqual = maxElectionId != null && maxElectionId electionId
   maxElectionIdIsLess = maxElectionId == null || maxElectionId > electionId
   maxSetVersionIsGreater = maxSetVersion != null && maxSetVersion > setVersion;
   maxSetVersionIsLess = maxSetVersion == null || maxSetVersion < setVersion
```

which passes all tests, but doesn't seem to match this pseudocode.  I may be confused by the pseudocode syntax though.  I read the first one, for example, as equivalent to this ternary logic:

```
   maxElectionIdIsGreater = 
                 maxElectionId != null ? 
                      maxElectionId < electionId : 
                      false
```

which simplifies to:

```
   maxElectionIdIsGreater = maxElectionId != null && maxElectionId < electionId
```

but I had to change the `<` to a `>` for it to work as expected.

jyemin (changes requested):

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
nbbeeken: I was thinking that this was an important note as for why the ternary would return false or true.

Like:
```
maxSetVersionIsLess = topologyDescription.maxSetVersion < serverDescription.setVersion if topologyDescription.maxSetVersion is not null else true
# maxSetVersion is automatically less when it is null (that's the "else true"), otherwise get the result of the actual comparison
```

jyemin: Oh, I see.  That makes sense, then.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
jyemin: > I can't speak to why flipping the check to assert that maxElectionId is less than electionId would make your tests pass.

I'm flipping it to `>`.  It's `<` in the pseudocode.  Just looking at the pseudocode, doesn't it seem wrong?  If the boolean is called `maxElectionIdIsGreater`, shouldn't it be:

```
maxElectionIdIsGreater = topologyDescription.maxElectionId > serverDescription.electionId if topologyDescription.maxElectionId is not null else false
```

?

nbbeeken: OMG! Yea the pseudo code is reversed 😅  wow too easy to miss, fixed! 

In node since this is an objectId I actually am using a comparator, when I copied over the code I translated the 1/0/-1 logic, and flipped the check, whoops!

jyemin: `maxElectionIdIsLess` still looks incorrect.

ShaneHarvey (changes requested):
I still see a few test failures:
```
FAIL: test_rs_electionId_setVersion (test.test_discovery_and_monitoring.TestAllScenarios)
  File "/Users/shane/git/mongo-python-driver/test/utils.py", line 965, in assertion_context
    yield
  File "/Users/shane/git/mongo-python-driver/test/test_discovery_and_monitoring.py", line 213, in run_scenario
    check_outcome(self, c, phase["outcome"])
  File "/Users/shane/git/mongo-python-driver/test/test_discovery_and_monitoring.py", line 195, in check_outcome
    self.assertEqual(outcome.get("maxSetVersion"), topology.description.max_set_version)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 912, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 905, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: 2 != 1

During handling of the above exception, another exception occurred:

  File "/Users/shane/git/mongo-python-driver/test/test_discovery_and_monitoring.py", line 213, in run_scenario
    check_outcome(self, c, phase["outcome"])
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/contextlib.py", line 131, in __exit__
    self.gen.throw(type, value, traceback)
  File "/Users/shane/git/mongo-python-driver/test/utils.py", line 970, in assertion_context
    raise exc_type(exc_val).with_traceback(exc_tb)
  File "/Users/shane/git/mongo-python-driver/test/utils.py", line 965, in assertion_context
    yield
  File "/Users/shane/git/mongo-python-driver/test/test_discovery_and_monitoring.py", line 213, in run_scenario
    check_outcome(self, c, phase["outcome"])
  File "/Users/shane/git/mongo-python-driver/test/test_discovery_and_monitoring.py", line 195, in check_outcome
    self.assertEqual(outcome.get("maxSetVersion"), topology.description.max_set_version)
AssertionError: 2 != 1

======================================================================
FAIL: test_rs_null_election_id (test.test_discovery_and_monitoring.TestAllScenarios)
  File "/Users/shane/git/mongo-python-driver/test/utils.py", line 965, in assertion_context
    yield
  File "/Users/shane/git/mongo-python-driver/test/test_discovery_and_monitoring.py", line 213, in run_scenario
    check_outcome(self, c, phase["outcome"])
  File "/Users/shane/git/mongo-python-driver/test/test_discovery_and_monitoring.py", line 164, in check_outcome
    self.assertEqual(
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 912, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 1292, in assertMultiLineEqual
    self.fail(self._formatMessage(msg, standardMsg))
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 753, in fail
    raise self.failureException(msg)
AssertionError: 'RSPrimary' != 'Unknown'
- RSPrimary
+ Unknown


During handling of the above exception, another exception occurred:

  File "/Users/shane/git/mongo-python-driver/test/test_discovery_and_monitoring.py", line 213, in run_scenario
    check_outcome(self, c, phase["outcome"])
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/contextlib.py", line 131, in __exit__
    self.gen.throw(type, value, traceback)
  File "/Users/shane/git/mongo-python-driver/test/utils.py", line 970, in assertion_context
    raise exc_type(exc_val).with_traceback(exc_tb)
  File "/Users/shane/git/mongo-python-driver/test/utils.py", line 965, in assertion_context
    yield
  File "/Users/shane/git/mongo-python-driver/test/test_discovery_and_monitoring.py", line 213, in run_scenario
    check_outcome(self, c, phase["outcome"])
  File "/Users/shane/git/mongo-python-driver/test/test_discovery_and_monitoring.py", line 164, in check_outcome
    self.assertEqual(
AssertionError: 'RSPrimary' != 'Unknown'
- RSPrimary
+ Unknown


======================================================================
FAIL: test_rs_setversion_without_electionid (test.test_discovery_and_monitoring.TestAllScenarios)
  File "/Users/shane/git/mongo-python-driver/test/utils.py", line 965, in assertion_context
    yield
  File "/Users/shane/git/mongo-python-driver/test/test_discovery_and_monitoring.py", line 213, in run_scenario
    check_outcome(self, c, phase["outcome"])
  File "/Users/shane/git/mongo-python-driver/test/test_discovery_and_monitoring.py", line 164, in check_outcome
    self.assertEqual(
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 912, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 1292, in assertMultiLineEqual
    self.fail(self._formatMessage(msg, standardMsg))
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 753, in fail
    raise self.failureException(msg)
AssertionError: 'Unknown' != 'RSPrimary'
- Unknown
+ RSPrimary


During handling of the above exception, another exception occurred:

  File "/Users/shane/git/mongo-python-driver/test/test_discovery_and_monitoring.py", line 213, in run_scenario
    check_outcome(self, c, phase["outcome"])
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/contextlib.py", line 131, in __exit__
    self.gen.throw(type, value, traceback)
  File "/Users/shane/git/mongo-python-driver/test/utils.py", line 970, in assertion_context
    raise exc_type(exc_val).with_traceback(exc_tb)
  File "/Users/shane/git/mongo-python-driver/test/utils.py", line 965, in assertion_context
    yield
  File "/Users/shane/git/mongo-python-driver/test/test_discovery_and_monitoring.py", line 213, in run_scenario
    check_outcome(self, c, phase["outcome"])
  File "/Users/shane/git/mongo-python-driver/test/test_discovery_and_monitoring.py", line 164, in check_outcome
    self.assertEqual(
AssertionError: 'Unknown' != 'RSPrimary'
- Unknown
+ RSPrimary


======================================================================
FAIL: test_rs_use_setversion_without_electionid (test.test_discovery_and_monitoring.TestAllScenarios)
  File "/Users/shane/git/mongo-python-driver/test/utils.py", line 965, in assertion_context
    yield
  File "/Users/shane/git/mongo-python-driver/test/test_discovery_and_monitoring.py", line 213, in run_scenario
    check_outcome(self, c, phase["outcome"])
  File "/Users/shane/git/mongo-python-driver/test/test_discovery_and_monitoring.py", line 164, in check_outcome
    self.assertEqual(
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 912, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 1292, in assertMultiLineEqual
    self.fail(self._formatMessage(msg, standardMsg))
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/unittest/case.py", line 753, in fail
    raise self.failureException(msg)
AssertionError: 'Unknown' != 'RSPrimary'
- Unknown
+ RSPrimary


During handling of the above exception, another exception occurred:

  File "/Users/shane/git/mongo-python-driver/test/test_discovery_and_monitoring.py", line 213, in run_scenario
    check_outcome(self, c, phase["outcome"])
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/contextlib.py", line 131, in __exit__
    self.gen.throw(type, value, traceback)
  File "/Users/shane/git/mongo-python-driver/test/utils.py", line 970, in assertion_context
    raise exc_type(exc_val).with_traceback(exc_tb)
  File "/Users/shane/git/mongo-python-driver/test/utils.py", line 965, in assertion_context
    yield
  File "/Users/shane/git/mongo-python-driver/test/test_discovery_and_monitoring.py", line 213, in run_scenario
    check_outcome(self, c, phase["outcome"])
  File "/Users/shane/git/mongo-python-driver/test/test_discovery_and_monitoring.py", line 164, in check_outcome
    self.assertEqual(
AssertionError: 'Unknown' != 'RSPrimary'
- Unknown
+ RSPrimary
```
I haven't looked into all the failures but the last one looks like a problem with the use_setversion_without_electionid.yml test. In the step for `# Reconfig, B reports as primary, B is missing the electionId but reports setVersion`, shouldn't B's response be considered stale with the updated comparison rules?


Here's how I've implemented the comparison which I think is correct. MinKey compares False for > and < for all objects and True for `==` with other MinKey which simplifies the comparison logic:
```python
    max_election_tuple = max_election_id, max_set_version
    new_election_tuple = server_description.election_id, server_description.set_version
    max_election_compare_safe = tuple(i if i is not None else MinKey() for i in max_election_tuple)
    new_election_compare_safe = tuple(i if i is not None else MinKey() for i in new_election_tuple)

    if new_election_compare_safe >= max_election_compare_safe:
        max_election_id, max_set_version = new_election_tuple
    else:
        # Stale primary, set to type Unknown.
        sds[server_description.address] = server_description.to_unknown()
        return _check_has_primary(sds), replica_set_name, max_set_version, max_election_id
```

nbbeeken:
You're missing the leading if stmt:
`if (serverDescription.electionId != null && serverDescription.setVersion != null)`
We don't enter the block where staleness is marked if the description doesn't have both of these properties defined, and that was true before this spec change, I'm still thinking if that makes sense or not 🤔

ShaneHarvey:
I believe that if stmt is incorrect. A null field always compares < a non-null field, so if the max tuple is `(OID(1), 1)` and the new one has a null, the new one should be considered stale. This is how the server works, see:
https://github.com/mongodb/mongo/blob/r5.2.0/src/mongo/client/sdam/json_tests/sdam_tests/rs/use_setversion_without_electionid.json#L52-L84

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -1088,57 +1089,59 @@ updateRSWithPrimaryFromMember
...
-        topologyDescription.maxSetVersion = description.setVersion
+    if topologyDescription.maxElectionId < serverDescription.electionId or (
+        (
+            topologyDescription.maxElectionId == serverDescription.electionId
+            or serverDescription.electionId == null
+        )
+        and topologyDescription.maxSetVersion < serverDescription.setVersion
+    ):
+        topologyDescription.maxElectionId = serverDescription.electionId
+        topologyDescription.maxSetVersion = serverDescription.setVersion
```
ShaneHarvey, line 1130: I would prefer to see this if-stmt and the "# Stale primary" be combined into one if/else for simplicity. If we are leaving the null comparison implied and not explicit (which the current code does) then what do you think about this?:
```
if (topologyDescription.electionId == serverDescription.maxElectionId and topologyDescription.setVersion >= serverDescription.maxSetVersion) or topologyDescription.electionId > serverDescription.maxElectionId:
    topologyDescription.maxElectionId = serverDescription.electionId
    topologyDescription.maxSetVersion = serverDescription.setVersion
else:
    # Stale primary....
```

ShaneHarvey (commented):
The new tests pass in Python.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
nbbeeken: Went with just the first way for now, We could maybe expand the null comment with the suggestions for zero OID/setVersion, I'm thinking its too verbose for pseudo code.

jyemin (approved):
LGTM

PR #1124, opened 2022-01-21
DRIVERS-1374: remove outdated causal consistency test

dariakp:
DRIVERS-1374

Removing test that is incorrect due to SPEC-1019


nbbeeken (approved):
LGTM as is, but just a Q to consider would it make sense to keep the title of the test for reference of _what_ was removed?

dariakp:
@nbbeeken I was using the guidance specified in the readme for this repo

> Outdated tests must not be removed completely, but may be marked as such (e.g.
> by striking through or replacing the entire test with a note (e.g. **Removed**).

PR #1125, opened 2022-01-21
DRIVERS-1030: Implicit sessions allocate underlying server session after connection checkout

nbbeeken:
https://jira.mongodb.org/browse/DRIVERS-1030

The language change is on the light side but the undertaking for this change is likely large for most drivers. I could use some discussion guidance on more rigorous prose tests to reccomend to drivers. 

The key pain point we're trying to solve is that connection pool size is the true to limit to how many concurrent operations can happen against a particular node, by obtaining implicit sessions after we've checkedOut we can be sure we're not creating a larger than necessary pool of sessions.

I looked at Transactions and LB and to my knowledge there isn't an impact to those specs, transactions are using explicit sessions, and LB will behave the same it has the continued concern of using the same connection related to a given session.

### DIFF
--- a/source/sessions/driver-sessions.rst
+++ b/source/sessions/driver-sessions.rst
@@ -503,8 +503,8 @@ Sessions and Connections
 ========================
 A driver MUST only obtain an implicit session after it successfully checks out a connection.
 This limits the number of implicit sessions to never exceed the maximum connection pool size.
-The motivation for this behavior is to prevent many sessions from being created in a scenario
-where only a limited number are actually needed to execute operations.
+The motivation for this behavior is to prevent too many sessions from being created in a scenario
+where only a limited number are actually needed to execute operations (i.e. TooManyLogicalSessions error).
 
 This only applies to implicit sessions, explicit sessions should not be modified
 to be bound by connection checkout in this way.
@@ -1260,6 +1260,21 @@ session because disconnects should be relatively rare and the server won't
 normally accumulate a large number of abandoned dirty sessions. Any abandoned
 sessions will be automatically cleaned up by the server after the
 configured ``logicalSessionTimeoutMinutes``.
+
+
+Why must driver's create an implicit session after successfully obtaining a connection from the pool?
+-----------------------------------------------------------------------------------------------------
+The problem that may occur is when the number of concurrent application requests are larger than the number of available connections,
+the driver may generate many more implicit sessions than connections.
+For example with maxPoolSize=1 and 100 threads, 100 implicit sessions may be created.
+This increases the load on the server since session state is cached in memory.
+In the worst case this kind of workload can hit the session limit and trigger TooManyLogicalSessions.
+
+In order to address this, drivers should check out an implicit session only after checking out a connection.
+This change will limit the number of implicit sessions to no greater than an application's maxPoolSize
+
+It is still possible that via explicit sessions or cursors, which hold on to the session they started with, a driver could over allocate sessions.
+But those scenarios are extenuating and outside the scope of solving in this spec.
 
 Change log
 ==========
--- a/source/sessions/driver-sessions.rst
+++ b/source/sessions/driver-sessions.rst
@@ -1109,7 +1109,7 @@ ensure that they close any explicit client sessions and any unexhausted cursors.
     * In the parent, return the ClientSession to the pool, create a new ClientSession, and assert its lsid is the same.
     * In the child, return the ClientSession to the pool, create a new ClientSession, and assert its lsid is different.
 
-13. To confirm that implicit sessions are only obtained after a successful connection checkout
+13. To confirm that implicit sessions only allocate their server session after a successful connection checkout
 
     * Create a MongoClient with maxPoolSize=1, and attach a command succeeded listener
     * Initiate 3 concurrent operations, e.g. collection.insertOne(doc)
@@ -1310,4 +1310,4 @@ Change log
 :2021-04-08: Updated to use hello and legacy hello
 :2021-04-08: Adding in behaviour for load balancer mode.
 :2020-05-26: Simplify logic for determining sessions support
-:2022-01-19: Implicit sessions MUST be obtained after connection checkout succeeds
+:2022-01-19: Implicit sessions MUST obtain server session after connection checkout succeeds
--- a/source/sessions/driver-sessions.rst
+++ b/source/sessions/driver-sessions.rst
@@ -1262,7 +1262,7 @@ sessions will be automatically cleaned up by the server after the
 configured ``logicalSessionTimeoutMinutes``.
 
 
-Why must driver's wait to consume a server session until after a connection is checked out?
+Why must drivers wait to consume a server session until after a connection is checked out?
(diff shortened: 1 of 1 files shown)


jyemin:
The key pain point you identified suggests a possible prose test.  Something like this:

1. Create a MongoClient with maxPoolSize=1, and a command listener
2. Initiate X concurrent requests, e.g. `collection.insertOne(doc)`
3. Wait for all requests to complete
4. Assert that all commands contain the same lsid

nbbeeken:
Good suggestion! In fact my current prose was very nonspecific about how to assert this, command monitoring seems the most cross driver friendly way. I was thinking for those that implement a ServerSessionPool it would also make sense to assert size checks on that. Does that make sense to include here?

jyemin:
I prefer an opaque test like the one I suggested.  The problem is that in many languages (e.g. Java!) there is literally no way to access the ServerSessionPool from the MongoClient.  And we really want this to be an integration test from the MongoClient, and not a unit test at a lower level (which maybe could access the session pool), because that's the only way to ensure all the components are working together as expected.

source/sessions/driver-sessions.rst
```diff
@@ -496,6 +499,16 @@ If a driver decides to run a ``KILLCURSORS`` command on the cursor, it also MAY
 run using the same session ID. See the Exceptions below for when it is permissible to not
 include a session ID in a ``KILLCURSORS`` command.
 
+Sessions and Connections
+========================
+A driver MUST only obtain an implicit session after it successfully checks out a connection.
+This limits the number of implicit sessions to never exceed the maximum connection pool size.
+The motivation for this behavior is to prevent many sessions from being created in a scenario
```
dariakp, line 506: I think the motivation stated here could benefit from the details actually included in DRIVERS-1030 (regarding not wanting to have more sessions than there are connections and specifically not hitting the TooManyLogicalSessions error). In theory, we could execute all operations with just one session (if we queue them up), but we don't want to do that either. So, I think we should clarify exactly why we are limiting session count to connection count, what the downside is to having too many sessions or too few, and why we only care about implicit sessions and not explicit ones. (Perhaps some of this can go in the Q&A section)

dariakp (changes requested):

source/sessions/driver-sessions.rst
nbbeeken: Added essentially what's on the ticket with a bit of phrasing, lmk what you think!

rstam:
This is going to be very hard to implement in the C# driver because the implicit session is created in High and the connection is checked out of the connection pool in Core. It is almost impossible to reverse the order.

rstam:
> This is going to be very hard to implement in the C# driver because the implicit session is created in High and the connection is checked out of the connection pool in Core. It is almost impossible to reverse the order.

Talking to Jeff Yemin about this he has proposed what sounds like a workable approach to implement this in the C# driver, involving creating a lazyily-initialized session object that delays actually acquiring a low-level server session from the session pool until the very last minute, by which time the connection will already have been acquired.

source/sessions/driver-sessions.rst
```diff
@@ -444,7 +444,9 @@ Check Whether a Deployment Supports Session). If sessions are supported, the
 driver MUST behave as if a new ``ClientSession`` was started just for this one
 operation and ended immediately after this operation completes. The actual
 implementation will likely involve calling ``client.startSession``, but that is not
-required by this spec.
+required by this spec. Regardless, please consult the startSession section to
+replicate the required steps for creating a session.
+Drivers MUST create an implicit session only after successfully checking out a connection.
```
rstam, line 449: The C# driver will have to continue to create an implicit session *before* checking out a connection, but we think we can achieve the desired result of not consuming a server session id until after the connection is checked out by making the implicit session lazily acquire a server session id at a later point in time.

Not sure if you want the spec to be as restrictive as currently written here because some drivers might actually not be able to do it exactly this way.

rstam: Maybe the wording could be "MUST NOT consume a server session id until after the connection is checked out" instead of "MUST NOT create an implicit session until after the connection is checked out"?

rstam (changes requested):

source/sessions/driver-sessions.rst
nbbeeken: Yea I think this is a good idea, and I can mention the reasoning in Q&A since other drivers may approach this with a similar design. Updated!

source/sessions/driver-sessions.rst
```diff
@@ -1240,6 +1261,26 @@ normally accumulate a large number of abandoned dirty sessions. Any abandoned
 sessions will be automatically cleaned up by the server after the
 configured ``logicalSessionTimeoutMinutes``.
 
+
+Why must driver's wait to consume a server session until after a connection is checked out?
```
dariakp, line 1265:
```suggestion
Why must drivers wait to consume a server session until after a connection is checked out?
```

source/sessions/driver-sessions.rst
```diff
@@ -1240,6 +1261,26 @@ normally accumulate a large number of abandoned dirty sessions. Any abandoned
...
+In order to address this, drivers MUST NOT consume a server session id until after the connection is checked out.
+This change will limit the number of "in use" server sessions to no greater than an application's maxPoolSize.
+
+The language here is specific about obtaining a server session as opposed to creating the implicit session
+to permit drivers to take an implementation approach where the implicit session creation logic largely remains unchanged.
+Implicit session creation can be left as is, as long as the underlying server resource isn't allocated until it
+is needed and, known it will be used, after connection checkout succeeds.
+
+It is still possible that via explicit sessions or cursors, which hold on to the session they started with, a driver could over allocate sessions.
+But those scenarios are extenuating and outside the scope of solving in this spec.
```
dariakp, line 1282: is the idea that the user of the driver can control these so it's on them to not over-allocate?

source/sessions/driver-sessions.rst
```diff
@@ -1269,3 +1310,4 @@ Change log
 :2021-04-08: Updated to use hello and legacy hello
 :2021-04-08: Adding in behaviour for load balancer mode.
 :2020-05-26: Simplify logic for determining sessions support
+:2022-01-19: Implicit sessions MUST obtain server session after connection checkout succeeds
```
dariakp, line 1313: Leaving this comment open to make sure we update the date before merging, but on the same note, should there be a minor version bump to the spec?

dariakp (changes requested):
A nit and some closing questions

source/sessions/driver-sessions.rst
nbbeeken: Yep, this would really happen if someone was leaking their explicit sessions, I don't think we can (or want to) do anything reasonable to clean them up / manage them. They're explicitly allocated, they need to be explicitly ended.

source/sessions/driver-sessions.rst
nbbeeken: I a version bump makes sense, this is a behavioral change to the spec. updated!

rstam (approved):
LGTM

PR #1126, opened 2022-01-26
fix(DRIVERS-1983): update direct connection details

durran:
Services now enable load balancing by the `--loadBalancerPort` option which requires the v1 or v2 proxy protocol to be enabled in the load balancer. Drivers wishing to bypass the load balancer and connect directly must do so now on the original service port without enabling load balancer mode. This PR updates the downstream wording to reflect this.


jyemin (approved):
LGTM

PR #1127, opened 2022-01-26
DRIVERS-1767: test all json files with a schemaVersion against their corresponding schema

nbbeeken:
I've made a script here that will grep for all json files that have a schemaVersion, and use that value in an `ajv` command to check that its a valid schema.

You can see what a failure looks like in the first commit's CI.

This uncovered missed version changes in: 
- `source/load-balancers/tests/cursors.yml`
- `source/load-balancers/tests/sdam-error-handling.yml`
- `source/load-balancers/tests/transactions.yml`

So we may want to change the ticket to drivers changes needed, but the impact might be non-existent if these tests have been sync-ed and no issues reported thus far.

Additionally made some drive-by improvements of the ci file, updated action/node/python versions simply to not let those run out of date. 

https://jira.mongodb.org/browse/DRIVERS-1767

### DIFF
--- a/.github/workflows/check_schema_version.sh
+++ b/.github/workflows/check_schema_version.sh
@@ -1,0 +1,32 @@
+#! /usr/bin/env bash
+
+# Collect error codes so we can see all the invalid json files in one run
+exitCode=0
+
+# $1 - takes a single argument of the path to the JSON file containing a schemaVersion key at the top level.
+function get_schema_version() {
+  node << EOF
+    const { readFileSync } = require('fs')
+    console.log(JSON.parse(readFileSync("./$1")).schemaVersion)
+EOF
+}
+
+function get_all_schemaVersion_defining_files () {
+  # look for all json files with 'schemaVersion": "[1-9]'
+  grep --include='*.json' --files-with-matches --recursive --word-regexp --regexp='schemaVersion": "[1-9]' source | \
+  # Remove the known invalid test files from 'unified-test-format/tests/invalid'
+  grep --word-regexp --invert-match 'unified-test-format/tests/invalid' | \
+  # sort the result!
+  sort
+}
+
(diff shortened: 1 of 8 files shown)


.github/workflows/check_schema_version.sh
```diff
@@ -0,0 +1,32 @@
...
+function get_schema_version() {
+  node << EOF
+    const { readFileSync } = require('fs')
+    console.log(JSON.parse(readFileSync("./$1")).schemaVersion)
+EOF
+}
+
+function get_all_schemaVersion_defining_files () {
+  # look for all json files with 'schemaVersion": "[1-9]'
+  grep --include='*.json' --files-with-matches --recursive --word-regexp --regexp='schemaVersion": "[1-9]' source | \
```
jmikola, line 16: `ajv` operates on YAML files, which is what we do in [source/unified-test-format/tests/Makefile](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/tests/Makefile). Can this be changed to target those instead?

I suppose it's not a big issue if that change isn't feasible, since [json-regenerate-check.yml](https://github.com/mongodb/specifications/blob/master/.github/workflows/json-regenerate-check.yml) should already enforce that YAML and JSON are in sync with one another.

jmikola (commented):
I don't feel strongly about requiring driver changes for this PR, as test runners likely aren't concerned with `schemaVersion` unless it's unsupported -- in which case the tests would be skipped. Based on existing features and driver releases, I expect all drivers are beyond 1.4 at this point.

.github/workflows/check_schema_version.sh
nbbeeken: Done, I had to add js-yaml to the deps install step, and modified the grep regex to account for the mixed quoting.

jmikola (approved):

PR #1129, opened 2022-01-27
DRIVERS-1807 Revert CSOT changes from retryable reads/writes specs

ShaneHarvey:
I'm reverting the CSOT changes from retryable reads/writes specs to unblock Neal's work on retrying handshakes in DRIVERS-746 (#1115). We're also reworking the retry policy in DRIVERS-1807 so I think it makes sense to revert and re-review these CSOT changes anyway.

Note this diff was generated by:
```
git checkout 6dc6f80026f0f8d99a8c81f996389534b14f6602~ -- source/retryable-reads source/retryable-writes
```

### DIFF
--- a/source/retryable-reads/retryable-reads.rst
+++ b/source/retryable-reads/retryable-reads.rst
@@ -3,7 +3,7 @@ Retryable Reads
 ===============
 
 :Spec Title: Retryable Reads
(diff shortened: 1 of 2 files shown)


jmikola:
I don't think I need to review anything for a revert, but let me know if I'm mistaken.

PR #1130, opened 2022-01-27
DRIVERS-1899: Permit optional results for unacknowledged write concerns

jmikola:
https://jira.mongodb.org/browse/DRIVERS-1899

For write commands, this was previously inferable by existing documentation for the "acknowledged" property of Write Results. Likewise for findAndModify, comments suggested that the return value could be null. This change makes the documentation more explicit.

Also adds explicit documentation that InsertOneResult and InsertManyResult are not required, which was previously inferable.

Note: this is a spec clarification, so no driver changes are needed.

### DIFF
--- a/source/crud/crud.rst
+++ b/source/crud/crud.rst
@@ -1375,7 +1375,7 @@ Bulk Write Models
 Write Results
 ~~~~~~~~~~~~~
 
-The acknowledged property is defined for languages/frameworks without a sufficient optional type. Hence, a driver may choose to return a optional results (e.g. ``Optional<BulkWriteResult>``) such that unacknowledged writes don't have a value and acknowledged writes do have a value.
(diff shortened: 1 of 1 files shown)


source/crud/crud.rst
```diff
@@ -1375,7 +1375,7 @@ Bulk Write Models
 Write Results
 ~~~~~~~~~~~~~
 
-The acknowledged property is defined for languages/frameworks without a sufficient optional type. Hence, a driver may choose to return an Optional<BulkWriteResult> such that unacknowledged writes don't have a value and acknowledged writes do have a value.
+The acknowledged property is defined for languages/frameworks without a sufficient optional type. Hence, a driver may choose to return a optional results (e.g. ``Optional<BulkWriteResult>``) such that unacknowledged writes don't have a value and acknowledged writes do have a value.
```
benjirewis, line 1378:
```suggestion
The acknowledged property is defined for languages/frameworks without a sufficient optional type. Hence, a driver may choose to return an optional result (e.g. ``Optional<BulkWriteResult>``) such that unacknowledged writes don't have a value and acknowledged writes do have a value.
```

benjirewis (approved):

PR #1131, opened 2022-01-28
test: remove duplicate tls option test

dariakp:
I came across this duplicate test while fixing up the node driver uri spec runner. It's a minor thing, but it does mess with our IDE test runner tooling.


JamesKovacs (approved):
LGTM

PR #1133, opened 2022-02-01
feat(DRIVERS-1108): add service host to mech props

durran:
Updates the auth spec to allow drivers to allow a `SERVICE_HOST` property in their auth mechanism properties, which replaces the default host in the service principal name to the provided attribute.

Node impl: https://github.com/mongodb/node-mongodb-native/pull/3123

### DIFF
--- a/source/auth/auth.rst
+++ b/source/auth/auth.rst
@@ -465,6 +465,9 @@ mechanism_properties
 	SERVICE_REALM
 		Drivers MAY allow the user to specify a different realm for the service. This might be necessary to support cross-realm authentication where the user exists in one realm and the service in another.
 
+	SERVICE_HOST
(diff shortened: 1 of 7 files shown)


JamesKovacs (approved):
LGTM.

Discussed with Durran whether we can just incorporate `SERVICE_HOST` into the `SERVICE_NAME` as `authMechanismProperties=SERVICE_NAME:other%40example.com`. The .NET/C# driver correctly parses the service name as `other@example.com`, but we have other code that appends the `hostname` automatically. This would have to be altered and Windows/Linux differences in `serviceName/hostname@realm` on Windows versus `serviceName@hostname` on Linux would have to be dealt with. Easier to override the `hostname` with `SERVICE_HOST` than trying to parse `SERVICE_NAME`.

In summary let's add `SERVICE_HOST` to `authMechanismProperties` as proposed.

PR #1134, opened 2022-02-02
DRIVERS-1030: Session allocation should occur after connection checkout - test improvements

nbbeeken:
Circling back after a POC/implementation has been attempted in Java.

Some improvements to the test details, as well as permitting drivers to treat explicit sessions the same way, as long as they continue to function the same from a user's perspective it should be fine to treat the logic the same for both.

### DIFF
--- a/source/sessions/driver-sessions.rst
+++ b/source/sessions/driver-sessions.rst
@@ -1110,7 +1110,7 @@ ensure that they close any explicit client sessions and any unexhausted cursors.
 
 13. To confirm that implicit sessions only allocate their server session after a successful connection checkout
 
-    * Create a MongoClient with the following options: ``maxPoolSize=1``, ``retryableReads=true``, and ``retryableWrites=true``
+    * Create a MongoClient with the following options: ``maxPoolSize=1`` and ``retryWrites=true``
     * Attach a command succeeded listener that collects each command's lsid
     * Initiate the following concurrent operations
       * insertOne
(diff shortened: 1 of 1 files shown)


source/sessions/driver-sessions.rst
```diff
@@ -1111,8 +1110,17 @@ ensure that they close any explicit client sessions and any unexhausted cursors.
 
 13. To confirm that implicit sessions only allocate their server session after a successful connection checkout
 
-    * Create a MongoClient with maxPoolSize=1, and attach a command succeeded listener
-    * Initiate 3 concurrent operations, e.g. collection.insertOne(doc)
+    * Create a MongoClient with the following options: ``maxPoolSize=1``, ``retryableReads=true``, and ``retryableWrites=true``
```
jyemin, line 1113: Read retryability is not relevant since that feature doesn't make use of sessions

Nit: the option is called `retryWrites`.  It's also enabled by default so not sure we need to even mention it.  But your call.

jyemin (changes requested):

source/sessions/driver-sessions.rst
nbbeeken: Doh! on the options name, I figure it's safer to call out what should be enabled despite it being the expected default we want to make sure that the retryable path is tested. fixed the retryRead mention

jyemin (approved):
LGTM

rstam (approved):
LGTM

source/sessions/driver-sessions.rst
```diff
@@ -1111,8 +1110,17 @@ ensure that they close any explicit client sessions and any unexhausted cursors.
...
+    * Attach a command started listener that collects each command's lsid
+    * Initiate the following concurrent operations
+      * insertOne
+      * deleteOne
+      * updateOne
+      * bulkWrite ``[ { updateOne } ]``
+      * findOneAndDelete
+      * findOneAndUpdate
+      * findOneAndReplace
+      * find
```
Neilshweky, line 1123: I also added a `replaceOne` call in here as well.

PR #1135, opened 2022-02-03
feat(DRIVERS-1803): add new canonicalize host name opts

durran:
For GSSAPI this updates the values accepted in the `CANONICALIZE_HOST_NAME` auth mechanism properties. These new values are:

- "none" : No canonicalization is performed.
- "forward" : A forward DNS lookup is performed to canonicalize the host name.
- "forwardAndReverse" : A forward DNS lookup followed by a reverse lookup on that value is performed.

Drivers may also continue to support the old boolean values if desired in addition to these new values. `true` is equivalent to "forward" and `false` is equivalent to "none".


source/auth/auth.rst
```diff
@@ -486,8 +486,8 @@ Second, the client MAY implement its own canonicalization. If so, the canonicali
     # Unspecified which CNAME is used if > 1.
     host = one of the records in cnames
 
-  reversed = do a reverse DNS lookup for address
-  if reversed:
+  if forwardAndReverse:
+    reversed = do a reverse DNS lookup for address
```
JamesKovacs, line 490: `CANONICALIZE_HOST_NAME` is currently equivalent to `forwardAndReverse`. This PR changes it to mean `forward`. Is that the correct behaviour? Why are we potentially breaking backward compat?

JamesKovacs: As well the Python impl below does not take `true` versus `forward` versus `forwardAndReverse` into account. It still implements the old pseudo code where only `true` was applicable and had the same meaning as `forwardAndReverse`.

source/auth/auth.rst
```diff
@@ -468,7 +468,7 @@ mechanism_properties
 Hostname Canonicalization
 `````````````````````````
 
-If CANONICALIZE_HOST_NAME is true, the client MUST canonicalize the name of each host it uses for authentication. There are two options. First, if the client's underlying GSSAPI library provides hostname canonicalization, the client MAY rely on it. For example, MIT Kerberos has `a configuration option for canonicalization <https://web.mit.edu/kerberos/krb5-1.13/doc/admin/princ_dns.html#service-principal-canonicalization>`_.
+If CANONICALIZE_HOST_NAME is true, "forward", or "forwardAndReverse", the client MUST canonicalize the name of each host it uses for authentication. There are two options. First, if the client's underlying GSSAPI library provides hostname canonicalization, the client MAY rely on it. For example, MIT Kerberos has `a configuration option for canonicalization <https://web.mit.edu/kerberos/krb5-1.13/doc/admin/princ_dns.html#service-principal-canonicalization>`_.
```
JamesKovacs, line 471: What are the valid values for `CANONICALIZE_HOST_NAME`. Based on this statement, valid values include `true`, `"forward"`, and `"forwardAndReverse"`. I assume `false` would also be valid. In the spec tests below, `"none"` is also used.

Should true/false be deprecated? If not, should `false` or `"none"` be used when disabling canonicalization.

What is the expected behaviour for an unknown value like `"off"` or `"on"`? Should an error be raised? Should it default to true? Or false? My concern here is what if someone uses `"forwardReverse"` rather than `"forwardAndReverse"` by mistake? Or if they type `"foward"` (misspelling)?

We should clearly enumerate the allowed values for `CANONICALIZE_HOST_NAMES` and specify whether an error or warning should be raised for unknown values.

source/auth/auth.rst
JamesKovacs: Given that we are asking drivers to implement (or wrap libgssapi) to canonicalize host names, do we want to add some host name canonicalization tests to ensure that drivers are performing canonicalization correctly? It seems dangerous to leave this behaviour testing up to each driver individually as it may give rise to differences in implementation.

JamesKovacs (changes requested):
Overall looks reasonable. Some clarification required along with fixing the Python sample impl.

source/auth/auth.rst
durran: Good catch. I was basing this off the Node implementation which doesn't do the reverse lookup so I'll update to make forwardAndReverse the default. Also yes I will add some more tests as well and ping you when the changes are made.

durran: I've updated to default `true` to being "forwardAndRevers", updated the python impl, and added spec test for raising an error on invalid values.

source/auth/auth.rst
durran: I've called out now all valid values in the spec for `CANONICALIZE_HOST_NAME` and specified that drivers MUST error if the value provided is not one of them. I've decided not to deprecate `true` and `false` and just say that drivers MAY allow them, since I figured many apps may depend on `true` and `false` and not want to change them as these changes were driven by the dev tools team.

source/auth/auth.rst
jyemin: The Java driver calls `[java.net.InetAddress#getByName(java.lang.String)](https://docs.oracle.com/javase/8/docs/api/java/net/InetAddress.html#getByName-java.lang.String-)`, followed by `[java.net.InetAddress#getCanonicalHostName](https://docs.oracle.com/javase/8/docs/api/java/net/InetAddress.html#getCanonicalHostName--)`.  I think that's the equivalent of forwardAndReverse.... Can you confirm?

durran:
Node reference impl: https://github.com/mongodb/node-mongodb-native/pull/3131

source/auth/auth.rst
durran: Yes that is the equivalent of forwardAndReverse.

JamesKovacs (changes requested):
Looking good. Just one main outstanding question around MAY/MUST and tests.

source/auth/tests/connection-string.json
```diff
@@ -89,11 +89,76 @@
...
           "SERVICE_NAME": "other",
-          "CANONICALIZE_HOST_NAME": true,
-          "SERVICE_HOST": "example.com"
+          "SERVICE_HOST": "example.com",
+          "CANONICALIZE_HOST_NAME": "forward"
         }
       }
     },
+    {
+      "description": "should accept true as hostname canonicalization (GSSAPI)",
```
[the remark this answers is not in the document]
durran: I've removed the tests now for true/false values.

source/uri-options/tests/auth-options.json
```diff
@@ -2,7 +2,7 @@
   "tests": [
     {
       "description": "Valid auth options are parsed correctly (GSSAPI)",
-      "uri": "mongodb://foo:bar@example.com/?authMechanism=GSSAPI&authMechanismProperties=SERVICE_NAME:other,CANONICALIZE_HOST_NAME:true,SERVICE_HOST:example.com&authSource=$external",
+      "uri": "mongodb://foo:bar@example.com/?authMechanism=GSSAPI&authMechanismProperties=SERVICE_NAME:other,CANONICALIZE_HOST_NAME:forwar,SERVICE_HOST:example.comd&authSource=$external",
```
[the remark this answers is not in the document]
durran: Fixed.

source/uri-options/tests/auth-options.yml
```diff
@@ -1,7 +1,7 @@
 tests:
     -
         description: "Valid auth options are parsed correctly (GSSAPI)"
-        uri: "mongodb://foo:bar@example.com/?authMechanism=GSSAPI&authMechanismProperties=SERVICE_NAME:other,CANONICALIZE_HOST_NAME:true,SERVICE_HOST:example.com&authSource=$external"
+        uri: "mongodb://foo:bar@example.com/?authMechanism=GSSAPI&authMechanismProperties=SERVICE_NAME:other,CANONICALIZE_HOST_NAME:forwar,SERVICE_HOST:example.comd&authSource=$external"
```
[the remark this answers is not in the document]
durran: Fixed.

source/auth/auth.rst
```diff
@@ -460,7 +460,7 @@ mechanism_properties
 		Drivers MUST allow the user to specify a different service name. The default is "mongodb".
 
 	CANONICALIZE_HOST_NAME
-		Drivers MAY allow the user to request canonicalization of the hostname. This might be required when the hosts report different hostnames than what is used in the kerberos database. The default is "false".
+		Drivers MAY allow the user to request canonicalization of the hostname. This might be required when the hosts report different hostnames than what is used in the kerberos database. The value is a string of either "none", "forward", or "forwardAndReverse". "none" is the default and performs no canonicalization. "forward" performs a forward DNS lookup to canonicalize the hostname. "forwardAndReverse" performs a forward DNS lookup and then a reverse lookup on that value to canonicalize the hostname. The driver MUST error if any lookup errors or returns no results. Drivers MAY decide to also keep the legacy boolean values where `true` equals the "forwardAndReverse" behaviour and `false` equals "none".
```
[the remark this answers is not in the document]
durran: I've gone and removed those tests now for the MAY wording.

JamesKovacs (approved):
LGTM

PR #1136, opened 2022-02-06
DRIVERS-2191 Remove use of example.com in Client Side Encryption prose tests

kevinAlbs:
# Summary
Replace use of "example.com" in Client Side Encryption prose tests with "doesnotexist.invalid".
- Test expectations in "Custom Endpoint Test" Cases 6, 7, and 8 change from parsing errors to DNS resolution errors.
- The error expectation for Case 9 has not changed from "Invalid KMS response". The error comes from the response to the oauth request. The oauth request is still sent to a valid endpoint, but requests permissions to an [invalid scope](https://github.com/mongodb/libmongocrypt/blob/master/src/mongocrypt-kms-ctx.c#L1305-L1311).

Here is a PoC of the updated tests with the C driver: https://github.com/mongodb/mongo-c-driver/pull/942 with passing Client Side Encryption tests.

# Motivation & Background
See DRIVERS-2191 for a repro. Sending AWS signed requests to example.com has been recently (as of 02/04/2022) observed to frequently return an HTTP 404, rather than the expected 200.

There is no reason to use "example.com". The tests only need to assert that the passed endpoint is used.

For precedence, Case 10 uses "doesnotexist.local" and expects `to fail with a network exception indicating failure to resolve "doesnotexist.local".`.

Using "doesnotexist.invalid" agrees with [RFC 2606](https://www.rfc-editor.org/rfc/rfc2606.html#section-2):

> ".invalid" is intended for use in online construction of domain names that are sure to be invalid and which it is obvious at a glance are invalid.

### DIFF
--- a/source/client-side-encryption/tests/README.rst
+++ b/source/client-side-encryption/tests/README.rst
@@ -687,12 +687,12 @@ Configure with KMS providers as follows:
             "tenantId": <set from environment>,
             "clientId": <set from environment>,
             "clientSecret": <set from environment>,
-            "identityPlatformEndpoint": "example.com:443"
+            "identityPlatformEndpoint": "doesnotexist.invalid:443"
          },
          "gcp": {
             "email": <set from environment>,
             "privateKey": <set from environment>,
-            "endpoint": "example.com:443"
+            "endpoint": "doesnotexist.invalid:443"
          },
          "kmip": {
             "endpoint": "doesnotexist.local:5698"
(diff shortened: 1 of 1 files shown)


comandeo (approved):
LGTM :+1:

source/client-side-encryption/tests/README.rst
```diff
@@ -818,7 +818,7 @@ Test cases
         "location": "global",
         "keyRing": "key-ring-csfle",
         "keyName": "key-name-csfle",
-        "endpoint": "example.com:443"
+        "endpoint": "doesnotexist.invalid:443"
       }
 
    Expect this to fail with an exception with a message containing the string: "Invalid KMS response".
```
eramongodb, line 824: Does this line not need to be updated as with others?

```suggestion
   Expect this to fail with a network exception indicating failure to resolve "doesnotexist.invalid".
```

eramongodb (approved):

PR #1137, opened 2022-02-06
DRIVERS-2192 Fix comment helper specification tests

juliusgeo:


juliusgeo:
[POC](https://github.com/mongodb/mongo-python-driver/pull/847) for these changes implemented in PyMongo. 
[Evergreen build](https://evergreen.mongodb.com/version/620072815623434782a312b2?redirect_spruce_users=true)

source/crud/tests/unified/bulkWrite-comment.yml
```diff
@@ -48,15 +48,15 @@ tests:
             - updateOne:
                 filter: &updateOne_filter
                   _id: 2
-                update: &update { x: "updated" }
+                update: &update { $set: {x: "updated"} }
```
jmikola, line 51: @comandeo: If this was missed in https://github.com/mongodb/specifications/pull/1117 I think there may be a bug in Ruby if a replacement doc with no top-level, `$`-prefixed operators was required for updateOne's `update` parameter. See [Update vs. Replace Validation](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#update-vs-replace-validation) for more info.

I also find it a bit odd that the original test would have expected separate `update` commands for the replaceOne and updateOne operations in the bulk write, especially if updateOne essentially took a replacement document; however, I don't believe there's any spec requirement to efficiently batch write operations so that's not technically a deviation (although it'd likely cause conflicts with other spec tests if other drivers are all doing optimal batching).

source/crud/tests/unified/updateMany-comment.yml
```diff
@@ -29,8 +29,7 @@ tests:
         object: *collection0
         arguments:
           filter: &filter { _id: 1 }
-          update: &update
-            - $set: { x: 22 }
```
jmikola, line 33: This technically wasn't an error. @comandeo was supplying a [pipeline argument](https://docs.mongodb.com/manual/reference/command/update/#update-with-an-aggregation-pipeline) for the update. And since this test requires 4.4 as a minimum server version, the driver can be assured that it's supported by the server.

That said, the `$set` here is simple enough that there's probably no reason to use a pipeline. I'll let @comandeo chime in, but I think this is probably reasonable to change to a non-pipeline parameter here since we are already making other spec test changes in the PR (I wouldn't consider this worth it alone, though).

source/crud/tests/unified/updateMany-comment.yml
```diff
@@ -42,6 +41,7 @@ tests:
                   -
                     q: *filter
                     u: *update
+                    multi: true
```
jmikola, line 44: This is good. `multi` should always be specified for updateMany. For updateOne and replaceOne, tests can use `$$unsetOrMatches: false` since some drivers may omit the option and rely on the server's default of `false`.

source/crud/tests/unified/bulkWrite-comment.yml
```diff
@@ -148,6 +137,7 @@ tests:
                 delete: *collection_name
                 deletes:
                   - q: *deleteOne_filter
+                    limit: 1
```
jmikola, line 140: This is good. All spec tests should explicitly specify `1` or `0` for `limit` on `delete` commands.

source/crud/tests/unified/updateOne-comment.yml
```diff
@@ -29,7 +29,7 @@ tests:
         object: *collection0
         arguments:
           filter: &filter { _id: 1 }
-          update: &update { x: 22 }
+          update: &update { $set: {x: 22} }
```
jmikola, line 32: @comandeo: Another case where Ruby might not be validating its arguments for `updateOne` (and possibly `updateMany` and `replaceOne`).

source/crud/tests/unified/bulkWrite-comment.yml
comandeo: It seems that there is a bug in the unified test runner in Ruby driver. I tested my tests against Ruby drive only. Thank you @juliusgeo for fixing it!

source/crud/tests/unified/updateOne-comment.yml
comandeo: Noted.

ShaneHarvey:
I'll defer to the others here.

jmikola (approved):

comandeo (approved):

PR #1138, opened 2022-02-09
DRIVERS-1343 Include the "to" field in ChangeStreamDocument

abr-egn:


benjirewis (changes requested):
Awesome! You'll also want to update the "last updated" date, bump the minor version of the spec, and add to the bottom of the changelog. Also, is there any way we can add a spec test to ensure that a [rename change event](https://docs.mongodb.com/manual/reference/change-events/#rename-event) will include the new `to` field?

abr-egn:
Should be good for another look - updated the spec tracking info and added a test.

benjirewis (approved):
LGTM thanks!

source/change-streams/tests/unified/change-streams.yml
```diff
@@ -226,3 +226,27 @@ tests:
...
+  - description: "to field is set in a rename change event"
+    runOnRequirements:
+      - minServerVersion: "4.0.1"
+    operations:
+      - name: createChangeStream
+        object: *collection0
+        arguments:
+          pipeline: []
+        saveResultAsEntity: &changeStream0 changeStream0
+      - name: rename
```
jmikola, line 239: Oh, I didn't realize there was already another PR that added a `rename` operation. And this test file uses `schemaVersion: "1.0"`.

I think that can change our approach in https://github.com/mongodb/specifications/pull/1156#discussion_r839959883 and we can safely just start adding `rename` to test files without bumping the schema version, as any driver implementing a downstream ticket for [DRIVERS-1343](https://jira.mongodb.org/browse/DRIVERS-1343) is going to encounter this already.

PR #1140, opened 2022-02-09
Fix test plan link in client-side-encryption

dariakp:
The test plan was hardcoded to link to a very old commit instead of to the current version of the test README.


comandeo (approved):
:+1:

PR #1141, opened 2022-02-10
DRIVERS-2189 getMore sends inherited comment

comandeo:


source/change-streams/change-streams.rst
```diff
@@ -417,6 +417,10 @@ Driver API
      * Server versions prior to 4.4 only support string as comment,
      * and providing a non-string type will result in a server-side error.
      *
+     * If a comment is provided, drivers MUST attach this comment to all
+     * subsequent getMore commands run on the same cursor.id returned from
+     * the aggregate command.
```
jmikola, line 422: I don't think we refer to `cursor.id` anywhere else like this, but this gets the same point across with less words.

```suggestion
     * subsequent getMore commands run on the same cursor.
```

Feel free to accept/reject. I'll suggest on the other lines for your convenience in case you want to apply them as a batch.

source/crud/crud.rst
```diff
@@ -524,10 +523,9 @@ Read
      * Server versions prior to 4.4 only support string as comment,
      * and providing a non-string type will result in a server-side error.
      *
-     * Any comment set on a find command is inherited by any subsequent
-     * getMore commands run on the same cursor.id returned from the
-     * find command. Therefore, drivers MUST NOT attach the comment
-     * to subsequent getMore commands on a cursor.
+     * If a comment is provided, drivers MUST attach this comment to all
+     * subsequent getMore commands run on the same cursor.id returned from
+     * the find command.
```
jmikola, line 528:
```suggestion
     * subsequent getMore commands run on the same cursor.
```

jmikola (approved):
Optional suggestion for doc block and one question, but tests LGTM.

benjirewis (approved):

source/change-streams/change-streams.rst
comandeo: Sounds much better, I agree. Thank you!

PR #1144, opened 2022-02-11
DRIVERS-2203: Rename Versioned API to Stable API

jmikola:
https://jira.mongodb.org/browse/DRIVERS-2203

### DIFF
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -10,8 +10,8 @@ Client Side Encryption
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: 4.2
-:Last Modified: 2022-01-19
-:Version: 1.4.0
+:Last Modified: 2022-02-24
+:Version: 1.4.1
 
 .. contents::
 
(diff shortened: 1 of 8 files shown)


comandeo (approved):
LGTM
I searched through all files in the project, and the only leftovers of the word 'versioned' are in file names or spec tests.
I am not sure whether such a terminology change deserves version bump, but I am not against this bump.

jmikola:
> I am not sure whether such a terminology change deserves version bump, but I am not against this bump.

I concur. These are all patch bumps, which would be what would otherwise be used for small changes like typos.

Separately, I think we should revisit the idea of spec versioning. There are places where it definitely makes sense (e.g. unified test format with schema), but with some of the recent PRs I've worked on I'm not convinced it's adding any value. A number of older specs (e.g. CRUD) don't have a version and I'm not aware of any drivers actually relying on these spec versions (e.g. advertising that they comply with version X.Y of some spec). That's a task for a later day, though.

nbbeeken (commented):
Just to keep us all on the same page, I left a comment on DRIVERS-2203 relating to taking this work further to cover the drivers changes needed work of renaming the files/directory and making the changes inside the tests.

jmikola:
> Just to keep us all on the same page, I left a comment on [DRIVERS-2203](https://jira.mongodb.org/browse/DRIVERS-2203) relating to taking this work further to cover the drivers changes needed work of renaming the files/directory and making the changes inside the tests.

Reported [DRIVERS-2216](https://jira.mongodb.org/browse/DRIVERS-2216) to track that, per @behackett's request.

PR #1145, opened 2022-02-12
DRIVERS-1712: Convert session spec tests to unified test format

jmikola:
https://jira.mongodb.org/browse/DRIVERS-1712

POC: https://github.com/mongodb/mongo-php-library/pull/889


source/sessions/tests/driver-sessions-server-support.yml
```diff
@@ -0,0 +1,123 @@
...
+      - name: assertSessionNotDirty
+        object: testRunner
+        arguments:
+          session: *session0
+      - name: insertOne
+        object: *collection0
+        arguments:
+          session: *session0
+          document: { _id: 2 }
+        expectResult: { $$unsetOrMatches: { insertedId: { $$unsetOrMatches: 2 } } }
```
juliusgeo, line 43: I'm not sure these `expectResult`'s are necessary if they're entirely composed of fields with `$$unsetOrMatches`. It doesn't actually apply any restriction on what is in the result, right?

juliusgeo (approved):
LGTM!

jyemin:
@juliusgeo did you verify that these tests pass in your driver?

juliusgeo:
> @juliusgeo did you verify that these tests pass in your driver?

Sorry about that, completely forgot. [EVG Build ](https://evergreen.mongodb.com/version/620a9af22a60ed7ce130e798)

source/sessions/tests/driver-sessions-dirty-session-errors.yml
```diff
@@ -0,0 +1,353 @@
...
+          client: *client0
+    expectEvents:
+      - client: *client0
+        events:
+          - commandStartedEvent: &findAndModify_attempt
+              command:
+                findAndModify: *collection0Name
+                query: { _id: 1 }
+                update: { $inc: { x: 1 } }
+                new: false
```
kmahar, line 168: it appears that libmongoc's `mongoc_collection_find_and_modify_with_opts` only sends `new` in a command when it is set to `true`, presumably since `false` is the server default:
https://github.com/mongodb/mongo-c-driver/blob/28c2ca827d318cb3ff3c832982e65b39408707c2/src/libmongoc/src/mongoc/mongoc-collection.c#L3335-L3337

perhaps this should use `$$unsetOrMatches`? or should we consider this a libmongoc issue since it makes libmongoc less resilient to the server hypothetically changing the default?

jmikola: Thanks for point this out. I took at look at existing specs and noted that some legacy transaction spec tests (same format used by legacy session tests) also use `new: false`. libmongoc purposely ignores assertions on `new` fields in commandStartedEvents for `findAndModify` (see: https://github.com/mongodb/mongo-c-driver/blob/1.21.0/src/libmongoc/tests/json-test-monitoring.c#L418); however, there's probably a small bug there since it completely ignores the field instead of just a `new: false` assertion. I don't think it's worth reporting for a legacy test runner, though (cc: @kevinAlbs).

This came up previously in some unified specs and I see that I used `$$unsetOrMatches` there for reasons cited in https://github.com/mongodb/specifications/pull/980/files#r635414676. I see no reason we can't do the same here.

kmahar: SGTM

dariakp (approved):
LGTM, ran these in the node driver, too - the dirty explicit session is discarded cases are failing in the nodejs driver due to a mismatch in expected session id, but the existing poc-sessions test for the same has been permanently skipped in our driver due to the same failure (which we have a ticket to investigate), which leads me to believe that it's an issue with the node driver/test runner, so not a concern for this spec update specifically

jmikola:
> which leads me to believe that it's an issue with the node driver/test runner, so not a concern for this spec update specifically

@dariakp: thanks for the review. That works for me, as they are passing in PHP and Python.

PR #1149, opened 2022-02-16
chore: suggest ports for CSFLE testing that don't conflict with loadbalancer test port

nbbeeken:
The current port selection is the same ports used by haproxy for LB testing.

I'm not sure if we need a drivers ticket here there's no functional changes needed to anyones testing if they didn't run into this themselves, changing this now would just make future implementers pick a port that doesn't conflict. 

Please let me know if I should file one.

### DIFF
--- a/source/client-side-encryption/tests/README.rst
+++ b/source/client-side-encryption/tests/README.rst
@@ -84,7 +84,7 @@ Each YAML file has the following keys:
         - ``local`` The local KMS provider.
 
           - ``key`` A 96 byte local key.
-         
+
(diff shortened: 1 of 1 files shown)


kevinAlbs (approved):
LGTM, thank you for the fix. I validated the port change the C driver [tests](https://spruce.mongodb.com/task/mongo_c_driver_asan_ubuntu_test_asan_latest_server_auth_nosasl_openssl_cse_patch_f6e6a7025d3c65a24e3c32c2763dadf3ef5b804a_620daebed1fe07588c01d210_22_02_17_02_11_11/logs?execution=0).

I suggest updating the description from "KMIP testing" to "CSFLE testing". The mock HTTP servers started on ports 8000, 8001, and 8002 are used for AWS, Azure, and GCP KMS providers as well.

PR #1151, opened 2022-02-24
DRIVERS-2214 Fix getMore with comment specs

comandeo:


benjirewis (approved):
LGTM (I _believe_ these are the only spec tests that test `getMore` `comment` functionality).

jmikola (approved):

ShaneHarvey:
I understand why we stopped testing the change stream tests on sharded clusters, but why did we stop testing find and aggregate?

comandeo:
@ShaneHarvey  My understanding of the language here - https://github.com/mongodb/specifications/tree/master/source/change-streams/tests#testing-on-sharded-clusters - is that the "problem" with `getMore` is not limited to change streams, but it can happen with any first `getMore` after data is written. Therefore limiting topologies. If I was wrong, and this situation is only related to change streams, I think we can relax topology restrictions for find/aggregate.

I also thought it was not absolutely necessary to test this behavior on all topologies. We just want to make sure that drivers attach comment to `getMore` command; we do not even check server response.

PR #1152, opened 2022-03-01
DRIVERS-2223 Fix race condition in pool-checkout-returned-connection-maxConnecting.yml.

matthewdale:


stIncMale:
[DRIVERS-2223](https://jira.mongodb.org/browse/DRIVERS-2223) does not have an explanation of how exactly 4 connections may be created instead of 3, and I fail to deduce that explanation on my own without having a driver I am familiar with that actually fails the test, so I could either debug or think based on my knowledge of the driver. It's impossible to review the change without understanding the race condition in the test. @matthewdale, could you please provide such an explanation in DRIVERS-2223?

matthewdale:
@stIncMale Yes, I will add an explanation of the race condition in [DRIVERS-2223](https://jira.mongodb.org/browse/DRIVERS-2223).

stIncMale:
Thank you, @matthewdale for explaining the situation with the Go driver, and for the interesting and helpful conversation overall.

The JSON code in the PR does not correspond to the YAML code. It's worth auto-generating the JSON from YAML and pushing it to the PR so that people supporting drivers that only work with JSON could run the test. I generated JSON locally.

The proposed test does not pass in the Java driver because:

1. Once a checkout request raises the `ConnectionCreated` event (actually, a bit earlier, when it acquires a permit to establish a connection, but this is currently not relevant for the test), this checkout cannot receive a checked in connection, and will wait until it establishes a new connection.
2. Due to №1, when `conn1` is checked in, it is passed to the `thread3` because it's the only thread which issued a checkout that has not yet raised `ConnectionCreated`. This thread then checks out `conn1`.
3. Again, due to №1, when `conn2` and `conn3` are checked in, they stay in the pool, because there are no more checkout requests that can be satisfied by the connections available in the pool. As a result, `thread1` and `thread2` checkout `conn4` and `conn5`. Another difference is that these threads checkout concurrently, i.e., there is no order based on which one may predict which of these connections is going to be checked out first. In Go the "order" imposed by small 20 ms waits in checkin actions "orders" checkout actions issues by `thread1` and `thread2` because they can benefit from the checked in connections, which is not the case in Java.

stIncMale:
I think, that we could make the original test work for everyone by changing it such that the two threads blocked on establishing connections never establish any connections while the test is running:

1. Checkout a single connection in the main thread.
2. Start checking out two more connections in new threads, make sure these are blocked till the end of the test (the current blocking time of 750 ms should be enough).
3. Start a third checkout in a new thread, it will be blocked on `maxConnecting`.
4. Checkin the connection from step №1.
5. A single checkout succeeds (either one started in step №2 (Go), or one started in step №3 (Java)) only if the mechanism that the test is intended to check is working.
6. So far the pool raised two `ConnectionCheckedOut` and one `ConnectionReady` events.
7.  If the checkouts started in step №2 ever succeed, then it appears impossible to write further assertions that all drivers pass. Therefore we need to end the test and ignore all events that the threads from step №2 may produce. It appears that if we don't have any `waitForThread`/`waitForEvent`/`wait`, the test will successfully end, and then the test runner will either dispose of the "stuck" threads, or they will eventually be unblocked because they are blocked only for 750 ms. The important part is that they will not affect the outcome of the test.

stIncMale:
I noticed that the checks have passed. This is weird given that JSON does not correspond to YAML. For example, there is no correspondence to the following checkin:
```
  - name: checkIn
    connection: conn3
```

matthewdale:
Closing this because I'm not going to get back to it anytime soon.

PR #1155, opened 2022-03-10
note that 4.2.9 server is required for server selection prose test

kevinAlbs:
# Background & Motivation

The "Prose Test" in server-selection-tests.rst specifies configuring a `failCommand` fail point with the `blockTimeMS` option.

~~[SERVER-53512](https://jira.mongodb.org/browse/SERVER-53512)~~ [SERVER-41070](https://jira.mongodb.org/browse/SERVER-41070) added a `failCommand` fail point with `blockTimeMS` in ~~4.9.0~~ 4.2.9.

### DIFF
--- a/source/server-selection/server-selection-tests.rst
+++ b/source/server-selection/server-selection-tests.rst
@@ -281,7 +281,7 @@ Prose Test
 
 Multi-threaded and async drivers MUST also implement the following prose test:
 
-1. Configure a sharded cluster with two mongoses. Use a 4.9.0 or newer server version.
+1. Configure a sharded cluster with two mongoses. Use a 4.2.9 or newer server version.
 
 2. Enable the following failpoint against exactly one of the mongoses::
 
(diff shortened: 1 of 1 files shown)


source/server-selection/server-selection-tests.rst
```diff
@@ -281,7 +281,7 @@ Prose Test
 
 Multi-threaded and async drivers MUST also implement the following prose test:
 
-1. Configure a sharded cluster with two mongoses.
+1. Configure a sharded cluster with two mongoses. Use a 4.9.0 or newer server version.
```
patrickfreed, line 284: That particular feature of blockTimeMS was added in 4.9.0, but the option itself was added in 4.2.9 in [SERVER-41070](https://jira.mongodb.org/browse/SERVER-41070), so we can use that bound here.

kevinAlbs: Oof, I misread the ticket. Updated.

patrickfreed (approved):

PR #1156, opened 2022-03-16
DRIVERS-1713 Convert change stream tests to unified format

abr-egn:
DRIVERS-1713

This PR pulls in the test conversion that was done earlier in https://github.com/mongodb/mongo-rust-driver/pull/570.  There was a little bit of merging needed in `change-streams.yml` because some new tests had been added since the conversion, but otherwise it's all a direct copy.


source/change-streams/tests/unified/change-streams-errors.yml
```diff
@@ -0,0 +1,159 @@
+description: "change-streams errors"
+schemaVersion: "1.0"
```
benjirewis, line 2: Should this be schema version 1.7? Does 1.7 just include the addition of `ignoreExtraEvents`?

source/change-streams/tests/unified/change-streams-errors.yml
```diff
@@ -0,0 +1,159 @@
...
+createEntities:
+  - client:
```
benjirewis, line 4: I'm finding the numerous client, database and collection entities here confusing. Don't you just need one of each for the following tests?

source/change-streams/tests/unified/change-streams-errors.yml
```diff
@@ -0,0 +1,159 @@
...
+      collectionName: *collection0
+  - collection:
+      id: &globalDb0Collection1 globalDb0Collection1
+      database: *globalDatabase0
+      collectionName: *collection1
+
+initialData:
+  - collectionName: *collection0
+    databaseName: *database0
+    documents: []
```
benjirewis, line 57: Is it necessary to set this collection as empty with `initialData`?

source/change-streams/tests/unified/change-streams-errors.yml
```diff
@@ -0,0 +1,159 @@
...
+    runOnRequirements:
+      - minServerVersion: 3.6.0
+        topologies:
+          - single
+    operations:
+      - name: createChangeStream
+        object: *collection0
+        arguments:
+          pipeline: []
+          options: {}
```
benjirewis, line 69: So I think `createChangeStream` options such as `batchSize` can appear under `arguments` adjacent to things like `pipeline`. Like:
```
operations:
  - name: createChangeStream
     object: *collection0
     arguments:
      pipeline: []
      batchSize: 1
```
The unified test format [describes](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#id4) this helper as having all the same options as the corresponding `watch` method. Having the options embedded under `options:` seems odd to me (we can avoid having `options: {}` whenever there are no options, too).

source/change-streams/tests/unified/change-streams-errors.yml
```diff
@@ -0,0 +1,159 @@
...
+        saveResultAsEntity: &changeStream0 changeStream0
+      - name: insertOne
+        object: *globalCollection0
+        arguments:
+          document:
+            z: 3
+      - name: iterateUntilDocumentOrError
+        object: *changeStream0
+        expectError:
+          errorCode: 216
```
benjirewis, line 159:
```suggestion
          errorCode: 216
          
```

benjirewis (commented):
Thanks for taking on this conversion! 🧑‍🔧  Haven't looked at all the converted tests yet; just a few initial comments.

source/change-streams/tests/unified/change-streams-errors.yml
abr-egn: Yes to both; fixed.

source/change-streams/tests/unified/change-streams-errors.yml
abr-egn: This was a static preamble in the converter script; turns out only `change-streams.yml` needs all of these, so I've pruned it in the other files.

source/change-streams/tests/unified/change-streams-errors.yml
abr-egn: Nope!  Removed.

source/change-streams/tests/unified/change-streams-errors.yml
abr-egn: Done.

source/change-streams/tests/unified/change-streams-errors.yml
abr-egn: Done.

source/change-streams/tests/unified/change-streams-resume-allowlist.yml
```diff
@@ -0,0 +1,1418 @@
...
+                      resumeAfter:
+                        $$unsetOrMatches:
+                          $$exists:
+                            true
+              commandName: aggregate
+              databaseName: *database0
+  - description: change stream resumes after HostUnreachable
+    runOnRequirements:
+      - minServerVersion: "4.2"
+        maxServerVersion: 4.2.99
```
benjirewis, line 110:
```suggestion
        maxServerVersion: "4.2.99"
```
Here and in all other occurrences of `maxServerVersion` in yaml.

source/change-streams/tests/unified/change-streams-resume-errorLabels.yml
```diff
@@ -0,0 +1,1342 @@
...
+          - replicaset
+          - sharded
+          - load-balanced
+    operations:
+      - name: failPoint
+        object: testRunner
+        arguments:
+          client: globalClient
+          failPoint:
+            configureFailPoint: failGetMoreAfterCursorCheckout
```
benjirewis, line 1017: There was a [link](https://github.com/mongodb/specifications/blob/master/source/change-streams/tests/legacy/change-streams-resume-errorLabels.yml#L9) to SERVER ticket here that might still be helpful to leave in.

source/change-streams/tests/unified/change-streams.yml
```diff
@@ -177,3 +217,585 @@ tests:
                 comment: "comment"
               commandName: getMore
               databaseName: *database0
+
+  - description: $changeStream must be the first stage in a change stream pipeline sent to the server
+    runOnRequirements:
+      - minServerVersion: 3.6.0
```
benjirewis, line 223:
```suggestion
      - minServerVersion: "3.6.0"
```
Here and elsewhere in file.

source/change-streams/tests/unified/change-streams.yml
```diff
@@ -177,3 +217,585 @@ tests:
...
+      - name: createChangeStream
+        object: *collection0
+        arguments:
+          pipeline: []
+        saveResultAsEntity: &changeStream0 changeStream0
+      - name: dropCollection
+        object: *database0
+        arguments:
+          collection: *collection1
+      - name: rename
```
benjirewis, line 646: I'm not sure this function exists in the unified test format yet. You might need to define it in the spec and bump the version again.

source/change-streams/tests/unified/change-streams-resume-allowlist.yml
jmikola: Indeed. I realize that any sane (and correct) YAML parser should consider this a string but there's no harm in being explicit.

source/change-streams/tests/unified/change-streams.yml
jmikola: Relevant spec section would be [Database Operations](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#database-operations), which states:

> Database operations that require special handling or are not documented by an existing specification are described below.

This is a bit tricky as there is nothing special about the rename operation (with respect to the test runner), other than the fact that we don't have an existing spec for it. I expect this may come up down the line when the retryable reads spec tests get ported, since those have a number of other operations not explicitly defined in other specs (e.g. `mapReduce`, which may very well be obsolete by the time we get to that task).

Either way, I think it makes sense to document this in the unified spec for the time being, so we can at least document the expected arguments. Are there any other special operations we need to handle? I'm not sure if other change stream tests also performed collection-level renames, but if so we should add that as well.

source/change-streams/tests/unified/change-streams-resume-errorLabels.yml
abr-egn: Derp, forgot to preserve comments in the conversion.  Fixed here and elsewhere.

source/change-streams/tests/unified/change-streams.yml
abr-egn: Done.

source/change-streams/tests/unified/change-streams.yml
abr-egn: I'm pretty sure there aren't any other special operations; `renameCollection` was the only one added to the Rust runner to support these (and apologies that it didn't occur to me when doing so that it would need a spec update), and it looks like that's the only test that uses it.

I'll put together a PR to add `rename` to the unified spec.  Given the above, should it be a version increment to 1.8, rolled into 1.7, or a doc update without a version bump?

jmikola: > should it be a version increment to 1.8, rolled into 1.7, or a doc update without a version bump?

As much as I hate the idea, this is a technically an additive change and warrants a minor version bump. This will probably be the first time we have a minor version without a corresponding schema.

Alternatively, you can float the idea in `#driver-devs` about getting rid of the "spec version" and just relying on schema version. That will require some significant language changes to [Impact of Spec Changes on Schema Version](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#impact-of-spec-changes-on-schema-version), though.

jmikola: I alluded to this discussion in [DRIVERS-2253](https://jira.mongodb.org/browse/DRIVERS-2253).

jmikola: Let's not move forward on [DRIVERS-2253](https://jira.mongodb.org/browse/DRIVERS-2253) until it gets properly scheduled. But in the mean time, I think I can rules-lawyer our way out of requiring a schema version bump.

Quoting [Executing an Operation](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#executing-an-operation):

< If operation.object is not "testRunner", this is an entity operation. If operation.object is defined in the current test's Entity Map, the test runner MUST fetch that entity and note its type; otherwise, the test runner MUST raise an error for an undefined entity. _If operation.name does not correspond to a known operation for the entity type (per Entity Test Operations), the test runner MUST raise an error for an unsupported operation_. Test runners MAY skip tests that include operations that are intentionally unimplemented (e.g. listCollectionNames).

I think this leaves two possibilities:

 * Drivers have implemented the unified test runner correctly and will see an obvious error if they encounter a new test with a `rename` operation and implement it then when syncing the tests. This can also be noted in the downstream changes for DRIVERS-1713.
 * Drivers didn't implement the unified test runner correctly and won't realize the operation is unsupported (e.g. it could be silently ignored). In that case, I'd expect these change stream tests to fail since there wouldn't be a `rename` event returned. Either way, drivers would know to address this.

Feel free to float the idea in #driver-devs. But if anyone is uncomfortable with this approach (starting with @benjirewis), then I have no qualms about just bumping to 1.8 and calling it a day.

If so, just be sure to make a copy of `schema-1.7.json` and commit it as `schema-1.8.json` in this PR. I'd also suggest using `schemaVersion: "1.8"` in _only_ the test files that use `rename` (others can use 1.7), since that might also let test runner implementations catch that _something_ is different from 1.7, even if the JSON schema itself is unchanged. And we also get to avoid having a missing schema file, which would no doubt raise questions and require redundant explanations down the line.

jmikola: @abr-egn: I just saw https://github.com/mongodb/specifications/pull/1138/files#r839991713. Since that's already merged, we can argue the damage is already done (speaking lightheartedly here) and just ignore a schema bump for `rename`. When drivers implement their downstream tickets for [DRIVERS-1343](https://jira.mongodb.org/browse/DRIVERS-1343) they will surely encounter one of the "two possibilities" I alluded to above.

@benjirewis: Do you concur?

benjirewis: Yeah I should've seen that in DRIVERS-1343 🤦 ... I think not bumping the schema version to 1.8 (and remaining on 1.7) here is fine. I worry more about the drivers that silently ignore unrecognized operations (I hope this is not many, as that would be an incorrect implementation of the UTR). I'm curious about the followup in DRIVERS-2253 and am all for dropping spec versions beyond the unified test format (as well as other, trivial [metadata](https://jira.mongodb.org/browse/DRIVERS-2253?focusedCommentId=4447139&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4447139)).

source/change-streams/tests/unified/change-streams-resume-errorLabels.yml
```diff
@@ -0,0 +1,1344 @@
+# Tests for resume behavior on server versions that support the ResumableChangeStreamError label
+description: "change-streams errors"
```
jmikola, line 2:
```suggestion
description: "change-streams-resume-errorLabels"
```

source/change-streams/tests/unified/change-streams.yml
```diff
@@ -177,3 +217,587 @@ tests:
...
+    expectEvents:
+      - client: *client0
+        ignoreExtraEvents: true
+        events:
+          - commandStartedEvent:
+              command:
+                aggregate: *collection0
+                cursor: {}
+                pipeline:
+                  - $changeStream: {}
```
jmikola, line 262: Note: drivers that have yet to implement [DRIVERS-2231](https://jira.mongodb.org/browse/DRIVERS-2231) will fail this assertion. I don't think we need to fix that here, but we should note that here, but it would be good to bring it up in the downstream changes for this issue. I'll also add a "has to be done after" relationship between the issues.

Previously, the POC change stream tests used `$$unsetOrMatches` for `fullDocument` here. I opened [DRIVERS-2261](https://jira.mongodb.org/browse/DRIVERS-2261) to clean that up.

source/change-streams/tests/unified/change-streams-errors.yml
```diff
@@ -0,0 +1,126 @@
+description: "change-streams errors"
```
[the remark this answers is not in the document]
abr-egn: Done.

source/change-streams/tests/unified/change-streams-resume-allowlist.yml
```diff
@@ -0,0 +1,1421 @@
+# Tests for resume behavior on server versions that do not support the ResumableChangeStreamError label
+description: "change-streams errors"
```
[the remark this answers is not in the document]
abr-egn: Done.

source/change-streams/tests/unified/change-streams-errors.yml
```diff
@@ -0,0 +1,126 @@
...
+        topologies:
+          - single
+    operations:
+      - name: createChangeStream
+        object: *collection0
+        arguments:
+          pipeline: []
+        expectError:
+          errorCode: 40573
+  - description: Change Stream should error when an invalid aggregation stage is passed in
```
jmikola, line 42:
```suggestion

  - description: Change Stream should error when an invalid aggregation stage is passed in
```

IMO, newlines before each test (after the first) help with readability when editing YAML files. Feel free to disregard this suggestion, though.

source/change-streams/tests/unified/change-streams-resume-allowlist.yml
```diff
@@ -0,0 +1,1421 @@
...
+            $$exists: true
+          documentKey:
+            $$exists: true
+          operationType: insert
+          ns:
+            db: *database0
+            coll: *collection0
+          fullDocument:
+            x:
+              $numberInt: "1"
```
jmikola, line 74:
```suggestion
            x: 1
```

There should never be a reason to use `$numberInt` in a match expression. See: [Flexible Numeric Comparisons](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#flexible-numeric-comparisons).

source/change-streams/tests/unified/change-streams-resume-allowlist.yml
```diff
@@ -0,0 +1,1421 @@
...
+            _id:
+              $$exists: true
```
jmikola, line 76:
```suggestion
            _id: { $$exists: true }
```

These can all be condensed to a single line for readability.

source/change-streams/tests/unified/change-streams-errors.yml
```diff
@@ -0,0 +1,126 @@
...
+          - replicaset
+          - sharded
+          - load-balanced
+    operations:
+      - name: createChangeStream
+        object: *collection0
+        arguments:
+          pipeline:
+            - $project:
+                _id: 0
```
jmikola, line 81:
```suggestion
            - $project: { _id: 0 }
```

More concise. Consider some of the existing formatting in other spec tests and `change-streams.yml`.

source/change-streams/tests/unified/change-streams-errors.yml
```diff
@@ -0,0 +1,126 @@
...
+        saveResultAsEntity: &changeStream0 changeStream0
+      - name: insertOne
+        object: *globalCollection0
+        arguments:
+          document:
+            z: 3
```
jmikola, line 87:
```suggestion
          document: { z: 3 }
```

Most spec tests inline small documents like this.

source/change-streams/tests/unified/change-streams-errors.yml
```diff
@@ -0,0 +1,126 @@
...
+        object: testRunner
+        arguments:
+          client: globalClient
+          failPoint:
+            configureFailPoint: failCommand
+            mode:
+              times: 1
+            data:
+              failCommands:
+                - getMore
```
jmikola, line 110:
```suggestion
            mode: { times: 1 }
            data:
              failCommands: ["getMore"]
```

Most other `failPoint` operations use a single line for these.

source/change-streams/tests/unified/change-streams-resume-allowlist.yml
```diff
@@ -0,0 +1,1421 @@
...
+      - minServerVersion: "4.2"
+        topologies:
+          - replicaset
+          - sharded
+          - load-balanced
+    operations:
+      - name: failPoint
+        object: testRunner
+        arguments:
+          client: globalClient
```
jmikola, line 42: Note: you're using a string literal here when you likely intended to use a reference (i.e. `*globalClient`). I don't think it matters given what I'm about to say below but you may want to check for similar mistakes made elsewhere.

I didn't realize this in previous code reviews, but why do you use a global client here? This was never necessary in the original [POC tests I ported](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/tests/valid-pass/poc-change-streams.yml).

Perhaps this was influenced by the [legacy spec test runner algorithm](https://github.com/mongodb/specifications/tree/master/source/change-streams/tests#spec-test-runner), but if it's not actually necessary I think we'd do well to drop this from the ported tests.

source/change-streams/tests/unified/change-streams-resume-allowlist.yml
```diff
@@ -0,0 +1,1421 @@
...
+              failCommands:
+                - getMore
+              closeConnection: true
+      - name: createChangeStream
+        object: *collection0
+        arguments:
+          pipeline: []
+        saveResultAsEntity: &changeStream0 changeStream0
+      - name: insertOne
+        object: *globalCollection0
```
jmikola, line 57: I realize the tests in [poc-change-streams.yml](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/tests/valid-pass/poc-change-streams.yml) did you separate collections for running `insertOne`, but I don't recall why that was done. Interestingly, the current unified tests in [change-streams.yml](https://github.com/mongodb/specifications/blob/master/source/change-streams/tests/unified/change-streams.yml) _don't_ do this and share the same collection object used to create the change stream.

I'd see if you can make do without this and, if so, remove all of the global entities added in this PR. I wouldn't object to stripping them from the POC file as well, but we can always punt on that.

source/change-streams/tests/unified/change-streams-resume-allowlist.yml
```diff
@@ -0,0 +1,1421 @@
...
+          - commandStartedEvent:
+              command:
+                aggregate: *collection0
+                cursor: {}
+                pipeline:
+                  - $changeStream:
+                      resumeAfter:
+                        $$unsetOrMatches:
+                          $$exists:
+                            true
```
jmikola, line 105: This may be more readable:

```suggestion
                  - $changeStream:
                      resumeAfter: { $$unsetOrMatches: { $$exists: true }}
```

Prior art: https://github.com/mongodb/specifications/blob/master/source/unified-test-format/tests/valid-pass/poc-change-streams.yml#L214-L217

source/change-streams/tests/unified/change-streams.yml
```diff
@@ -8,6 +8,12 @@ createEntities:
       id: &client0 client0
       observeEvents:
         - commandStartedEvent
+      ignoreCommandMonitoringEvents:
+        - killCursors
+      useMultipleMongoses: false
+  - client:
+      id: &globalClient globalClient
```
jmikola, line 15: Per my comments above, I'll note that none of the existing tests in this file required global clients, databases, or collections.

That said, if these tests will operate on multiple collections you may want to explicitly initialize those in `initialData` (as I did in [`poc-change-streams.yml`](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/tests/valid-pass/poc-change-streams.yml#L47)).

jmikola (commented):
I just noticed this PR introduces global entities, which I think may be redundant. Please check that out and report back.

Also left various formatting suggestions that might make the test files more readable/maintainable (disregard as you wish, since those aren't functional changes). I did point out one place where you used a literal string instead of a YAML reference, so that might be worth looking into even if the style suggestions are ignored.

abr-egn:
Still working on the style updates but wanted to continue the discussion of client entities while I do that :)

source/change-streams/tests/unified/change-streams-resume-allowlist.yml
abr-egn: The `globalClient` and associated data entities aren't strictly necessary AFAICT; the tests can be edited to only use `client0` and add the `insert` to the expected command events.  I'm not sure that's an improvement on the current setup, though, since it's adding noise to the expected output (these are testing change streams, not insert).

source/change-streams/tests/unified/change-streams-errors.yml
abr-egn: Agreed - done.

source/change-streams/tests/unified/change-streams-resume-allowlist.yml
abr-egn: Fixed.

source/change-streams/tests/unified/change-streams-resume-allowlist.yml
jmikola: Ah, I re-read the POC spec tests again and it looks like those separate collections did have their own client. That was likely done for ignoring `insert` in the APM events, as you pointed out.

The tests in `change-streams.yml` vary. Some don't expect events at all, and others include the `insert` command.

I suppose a third option is just to add `insert` to `ignoreCommandMonitoringEvents` if you'd like to avoid the extra entities, but I won't push for that. Totally up to you and I'm OK with leaving this as-is (thanks for indulging my questions).

source/change-streams/tests/unified/change-streams-resume-allowlist.yml
abr-egn: Done.

source/change-streams/tests/unified/change-streams-errors.yml
abr-egn: Updated.

source/change-streams/tests/unified/change-streams-errors.yml
abr-egn: Done.

source/change-streams/tests/unified/change-streams-resume-allowlist.yml
abr-egn: Fixed the references here; see below for discussion of the client usage.

source/change-streams/tests/unified/change-streams-resume-allowlist.yml
abr-egn: Done.

source/change-streams/tests/unified/change-streams-errors.yml
```diff
@@ -70,47 +64,36 @@ tests:
   - description: Change Stream should error when _id is projected out
     runOnRequirements:
       - minServerVersion: "4.1.11"
-        topologies:
-          - replicaset
-          - sharded
-          - load-balanced
+        topologies: [ replicaset, sharded, load-balanced ]
```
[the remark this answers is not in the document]
abr-egn: Done.

source/change-streams/tests/unified/change-streams-resume-allowlist.yml
```diff
@@ -31,48 +31,38 @@ tests:
   - description: change stream resumes after a network error
     runOnRequirements:
       - minServerVersion: "4.2"
-        topologies:
-          - replicaset
-          - sharded
-          - load-balanced
+        topologies: [ replicaset, sharded, load-balanced ]
```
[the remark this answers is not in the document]
abr-egn: All of the constraints here were just ported straight from the legacy tests; I don't know why they were that way.  I consolidated everything to use a top-level list that includes `load-balanced`.

source/change-streams/tests/unified/change-streams-resume-errorLabels.yml
```diff
@@ -31,47 +31,38 @@ tests:
   - description: change stream resumes after HostUnreachable
     runOnRequirements:
       - minServerVersion: "4.3.1"
-        topologies:
-          - replicaset
-          - sharded
-          - load-balanced
+        topologies: [ replicaset, sharded, load-balanced ]
```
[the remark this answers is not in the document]
abr-egn: They all had the same version requirements as well, so consolidated both to the top level.

source/change-streams/tests/unified/change-streams-resume-allowlist.yml
abr-egn: Whoops, forgot to actually reply here :)  I'd prefer to leave it as-is - none of the other options seem worth the churn.

jmikola (approved):
Tested in PHPLIB. Various tests were skipped due to my local server version and there were a few failures due to us not having implemented `comment`, but everything looks OK otherwise.

benjirewis (approved):
These look great! They seem to work in the Go driver with additional logic for `rename` and `ignoreExtraEvents`. Thanks for the careful review responses. There are some merge conflicts that I think you'll have to handle, though.

PR #1157, opened 2022-03-16
DRIVERS-2218: Clarify intended bounding of implicit session allocation and rework flaky prose test

nbbeeken:
http://jira.mongodb.org/browse/DRIVERS-2218

* Clarify the purpose and goal of session limiting
* Adjust test 13 to correctly assert the optimal solution
* ~Clean up code snippets and test bullets~ moved to #1158

### DIFF
--- a/source/sessions/driver-sessions.rst
+++ b/source/sessions/driver-sessions.rst
@@ -504,10 +504,10 @@ include a session ID in a ``KILLCURSORS`` command.
 
 Sessions and Connections
 ========================
-A driver MUST only obtain an implicit session's ``ServerSession`` after it successfully checks out a connection.
+
+To reduce the number of ``ServerSessions`` created, the driver MUST only obtain an implicit session's
+``ServerSession`` after it successfully checks out a connection.
 A driver SHOULD NOT attempt to release the acquired session before connection check in.
-Through both through the pooling mechanism and limiting acquisition to a successful
-connection checkout we can have guaranteed improvement of ``ServerSession`` reuse.
 
 Explicit sessions MAY be changed to allocate a server session similarly.
 
@@ -1155,6 +1155,9 @@ Test Plan
     * Assert the following across at least 5 retries of the above test:
 
       * Drivers MUST assert that exactly one session is used for all operations at least once across the retries of this test.
+
+        * Note that it's possible, although rare, for >1 server session to be used because the session is not released until after the connection is checked in.
+
       * Drivers MUST assert that the number of allocated sessions is strictly less than the number of concurrent operations in every retry of this test. In this instance it would less than (but NOT equal to) 8.
 
 
@@ -1332,8 +1335,6 @@ Why should drivers NOT attempt to release a serverSession before checking back i
 There are a variety of cases, such as, retryable operations or cursor creating operations
 where a ``serverSession`` must remain acquired by the ``ClientSession`` after an operation is attempted.
 Attempting to account for all these scenarios has risks that do not justify the potential guaranteed ``ServerSession`` allocation limiting.
-Drivers SHOULD attempt to release the ``ServerSession`` to the pool at the earliest possible opportunity.
-Drivers SHOULD attempt to reuse ``ServerSession`` as best as possible, concurrency fairness willing.
 
 Change log
 ==========
--- a/source/sessions/driver-sessions.rst
+++ b/source/sessions/driver-sessions.rst
@@ -12,7 +12,7 @@ Driver Sessions Specification
 :Status: Accepted (Could be Draft, Accepted, Rejected, Final, or Replaced)
 :Type: Standards
 :Minimum Server Version: 3.6 (The minimum server version this spec applies to)
-:Last Modified: 2022-03-16
+:Last Modified: 2022-03-24
 
 .. contents::
 
--- a/source/sessions/driver-sessions.rst
+++ b/source/sessions/driver-sessions.rst
@@ -1154,8 +1154,8 @@ Test Plan
     * Wait for all operations to complete successfully
     * Assert the following across at least 5 retries of the above test:
 
-        * Drivers MUST assert that exactly one session is used for all operations at least once across the 5 retries of this test.
-        * Drivers MUST assert that the number of allocated sessions is strictly less than the number of concurrent operations in every retry of this test. In this instance it would less than (but NOT equal to) 8.
+      * Drivers MUST assert that exactly one session is used for all operations at least once across the 5 retries of this test.
+      * Drivers MUST assert that the number of allocated sessions is strictly less than the number of concurrent operations in every retry of this test. In this instance it would less than (but NOT equal to) 8.
 
 
 
--- a/source/sessions/driver-sessions.rst
+++ b/source/sessions/driver-sessions.rst
@@ -1154,7 +1154,7 @@ Test Plan
     * Wait for all operations to complete successfully
     * Assert the following across at least 5 retries of the above test:
 
-      * Drivers MUST assert that exactly one session is used for all operations at least once across the 5 retries of this test.
+      * Drivers MUST assert that exactly one session is used for all operations at least once across the retries of this test.
       * Drivers MUST assert that the number of allocated sessions is strictly less than the number of concurrent operations in every retry of this test. In this instance it would less than (but NOT equal to) 8.
 
 
--- a/source/sessions/driver-sessions.rst
+++ b/source/sessions/driver-sessions.rst
@@ -3,7 +3,7 @@ Driver Sessions Specification
 =============================
 
 :Spec Title: Driver Sessions Specification (See the registry of specs)
-:Spec Version: 1.9.0
+:Spec Version: 1.9.1
 :Author: Robert Stam
 :Spec Lead: A\. Jesse Jiryu Davis
 :Advisory Group: Jeremy Mikola, Jeff Yemin, Samantha Ritter
@@ -12,7 +12,7 @@ Driver Sessions Specification
 :Status: Accepted (Could be Draft, Accepted, Rejected, Final, or Replaced)
 :Type: Standards
 :Minimum Server Version: 3.6 (The minimum server version this spec applies to)
-:Last Modified: 2022-01-28
+:Last Modified: 2022-03-16
 
 .. contents::
 
@@ -205,10 +205,12 @@ in such a way that new options can be added in a backward compatible way (it is
 acceptable for backward compatibility to be at the source level).
 
 A ``ClientSession`` MUST be associated with a ``ServerSession`` at the time
-``startSession`` is called. As an implementation optimization drivers SHOULD reuse
+``startSession`` is called. As an implementation optimization drivers MUST reuse
 ``ServerSession`` instances across multiple ``ClientSession`` instances subject
 to the rule that a server session MUST NOT be used by two ``ClientSession``
-instances at the same time (see the Server Session Pool section).
+instances at the same time (see the Server Session Pool section). Additionally,
+a ``ClientSession`` may only ever be associated with one ``ServerSession`` for
+its lifetime.
 
 Drivers MUST report an error if sessions are not supported by the deployment
 (see How to Check Whether a Deployment Supports Sessions). This error MUST either
@@ -304,8 +306,8 @@ This property returns the ``SessionOptions`` that were used to start this
 sessionId
 ---------
 
-This property returns the session ID of this session. Note that if server
-sessions are pooled, different ``ClientSession`` instances will have the same session ID,
+This property returns the session ID of this session. Note that since ``ServerSession``s
+are pooled, different ``ClientSession`` instances will have the same session ID,
 but never at the same time.
 
 advanceClusterTime
@@ -502,10 +504,17 @@ include a session ID in a ``KILLCURSORS`` command.
 
 Sessions and Connections
 ========================
-A driver MUST only obtain an implicit session after it successfully checks out a connection.
-This limits the number of implicit sessions to never exceed the maximum connection pool size.
-The motivation for this behavior is to prevent too many sessions from being created in a scenario
-where only a limited number are actually needed to execute operations (i.e. TooManyLogicalSessions error).
+A driver MUST only obtain an implicit session's ``ServerSession`` after it successfully checks out a connection.
+A driver SHOULD NOT attempt to release the acquired session before connection check in.
+There are a variety of cases, such as, retryable operations or cursor creating operations
+where a ``serverSession`` must remain acquired by the ``ClientSession`` after an operation is attempted.
+Attempting to account for all these scenarios has risks that do not justify the potential guaranteed ``ServerSession`` allocation limiting.
+
+Drivers SHOULD attempt to release the ``ServerSession`` to the pool at the earliest possible opportunity.
+Drivers SHOULD attempt to reuse ``ServerSession`` as best as possible, concurrency fairness willing.
+
+Through both through the pooling mechanism and limiting acquisition to successful
+connection checkout we can have guaranteed improvement of ``ServerSession`` recycling.
 
 Explicit sessions MAY be changed to allocate a server session similarly, but it is not required.
 
@@ -666,7 +675,9 @@ and implicit sessions.
 When wrapping commands in a ``$query`` field
 --------------------------------------------
 
-If the driver is wrapping the command in a ``$query`` field in order to pass a readPreference to a mongos (see `ReadPreference and Mongos <./find_getmore_killcursors_commands.rst#readpreference-and-mongos>`_), the driver SHOULD NOT add the ``lsid`` as a top-level field, and MUST add the ``lsid`` as a field of the ``$query``
+If the driver is wrapping the command in a ``$query`` field for non-OP_MSG messages in order to pass a readPreference to a
+mongos (see `ReadPreference and Mongos <./find_getmore_killcursors_commands.rst#readpreference-and-mongos>`_),
+the driver SHOULD NOT add the ``lsid`` as a top-level field, and MUST add the ``lsid`` as a field of the ``$query``
 
 .. code:: typescript
 
@@ -781,9 +792,7 @@ In case of an error, the server response has the following format:
 
 Drivers MUST ignore any errors returned by the ``endSessions`` command.
 
-Drivers that do not implement a server session pool MUST run the ``endSessions``
-command when the ``ClientSession.endSession`` method is called. Drivers that do
-implement a server session pool SHOULD run the ``endSessions`` command once when
+Drivers that implement a server session pool SHOULD run the ``endSessions`` command once when
 the ``MongoClient`` instance is shut down. If the number of sessions is very large
 the ``endSessions`` command SHOULD be run multiple times to end 10,000 sessions at
 a time (in order to avoid creating excessively large commands).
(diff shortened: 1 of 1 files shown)


source/sessions/driver-sessions.rst
```diff
@@ -305,8 +306,8 @@ This property returns the ``SessionOptions`` that were used to start this
 sessionId
 ---------
 
-This property returns the session ID of this session. Note that if server
-sessions are pooled, different ``ClientSession`` instances will have the same session ID,
+This property returns the session ID of this session. Note that since ``ServerSession``s
+are pooled, different ``ClientSession`` instances will have the same session ID,
```
ShaneHarvey, line 310: This line does not render correctly. Also:
"instances will have" -> "instances can have"

source/sessions/driver-sessions.rst
```diff
@@ -501,10 +504,17 @@ include a session ID in a ``KILLCURSORS`` command.
...
-The motivation for this behavior is to prevent too many sessions from being created in a scenario
-where only a limited number are actually needed to execute operations (i.e. TooManyLogicalSessions error).
+A driver MUST only obtain an implicit session's ``ServerSession`` after it successfully checks out a connection.
+A driver SHOULD NOT attempt to release the acquired session before connection check in.
+There are a variety of cases, such as, retryable operations or cursor creating operations
+where a ``serverSession`` must remain acquired by the ``ClientSession`` after an operation is attempted.
+Attempting to account for all these scenarios has risks that do not justify the potential guaranteed ``ServerSession`` allocation limiting.
+
+Drivers SHOULD attempt to release the ``ServerSession`` to the pool at the earliest possible opportunity.
+Drivers SHOULD attempt to reuse ``ServerSession`` as best as possible, concurrency fairness willing.
```
ShaneHarvey, line 514: I would omit these two lines. They seem to contradict the "SHOULD NOT attempt to release the acquired session before connection check in" rule.

source/sessions/driver-sessions.rst
```diff
@@ -501,10 +504,17 @@ include a session ID in a ``KILLCURSORS`` command.
...
 ========================
-A driver MUST only obtain an implicit session after it successfully checks out a connection.
-This limits the number of implicit sessions to never exceed the maximum connection pool size.
-The motivation for this behavior is to prevent too many sessions from being created in a scenario
-where only a limited number are actually needed to execute operations (i.e. TooManyLogicalSessions error).
+A driver MUST only obtain an implicit session's ``ServerSession`` after it successfully checks out a connection.
+A driver SHOULD NOT attempt to release the acquired session before connection check in.
+There are a variety of cases, such as, retryable operations or cursor creating operations
+where a ``serverSession`` must remain acquired by the ``ClientSession`` after an operation is attempted.
+Attempting to account for all these scenarios has risks that do not justify the potential guaranteed ``ServerSession`` allocation limiting.
```
ShaneHarvey, line 511: Let's move the justification to the Q&A and put a link to it here. That way we can elaborate more on this point without cluttering the spec.

source/sessions/driver-sessions.rst
```diff
@@ -137,7 +137,7 @@ Applications start a new session like this:
 
 .. code:: typescript
 
-    options = new SessionOptions(...);
+    options = new SessionOptions(/* various settings */);
```
ShaneHarvey, line 140: These style changes make this PR harder to review. It would have been better to put split them into a separate PR.

source/sessions/driver-sessions.rst
```diff
@@ -501,10 +504,17 @@ include a session ID in a ``KILLCURSORS`` command.
...
+where a ``serverSession`` must remain acquired by the ``ClientSession`` after an operation is attempted.
+Attempting to account for all these scenarios has risks that do not justify the potential guaranteed ``ServerSession`` allocation limiting.
+
+Drivers SHOULD attempt to release the ``ServerSession`` to the pool at the earliest possible opportunity.
+Drivers SHOULD attempt to reuse ``ServerSession`` as best as possible, concurrency fairness willing.
+
+Through both through the pooling mechanism and limiting acquisition to successful
+connection checkout we can have guaranteed improvement of ``ServerSession`` recycling.
 
 Explicit sessions MAY be changed to allocate a server session similarly, but it is not required.
```
ShaneHarvey, line 519: The "but it is not required" part is redundant since "MAY" already means not required.

source/sessions/driver-sessions.rst
```diff
@@ -751,19 +776,23 @@ The server response has the following format:
...
+      errmsg: string;
+      code: number;
+    }
 
 Drivers MUST ignore any errors returned by the ``endSessions`` command.
 
-Drivers that do not implement a server session pool MUST run the ``endSessions``
-command when the ``ClientSession.endSession`` method is called. Drivers that do
-implement a server session pool SHOULD run the ``endSessions`` command once when
+Drivers that implement a server session pool SHOULD run the ``endSessions`` command once when
```
ShaneHarvey, line 795: Remove "that implement a server session pool" since it is required now.

source/sessions/driver-sessions.rst
```diff
@@ -781,15 +810,15 @@ corresponding ``ServerSession``. However, starting a server session might requir
...
 available for reuse. A ``ServerSession`` pool MUST belong to a ``MongoClient``
 instance and have the same lifetime as the ``MongoClient`` instance.
 
-If a driver has a server session pool, then when a new ``ClientSession`` is started
-it MUST attempt to acquire a server session from the server session pool. See
-the algorithm below for the steps to follow when attempting to acquire a
-``ServerSession`` from the server session pool.
+When a new implicit ``ClientSession`` is started it MUST NOT attempt to acquire a server
+session from the server session pool. When a new explicit ``ClientSession`` is started
+it MAY attempt to acquire a server session from the server session pool.
```
ShaneHarvey, line 820: MAY -> SHOULD

source/sessions/driver-sessions.rst
```diff
@@ -908,17 +935,13 @@ as follows:
...
-        },
-        ...
-    }
+  interface ClusterTime {
+    clusterTime: Timestamp;
+    signature: {
+      hash: Binary;
+      keyId: Int64;
+    };
+  }
```
ShaneHarvey, line 944: I prefer the previous code which shows the whole response including the "$clusterTime" field.

source/sessions/driver-sessions.rst
```diff
@@ -991,95 +1014,102 @@ been upgraded to 3.6.
 Test Plan
 =========
 
-1. Pool is LIFO.
-    * This test applies to drivers with session pools.
-    * Call ``MongoClient.startSession`` twice to create two sessions, let us call them ``A`` and ``B``.
-    * Call ``A.endSession``, then ``B.endSession``.
-    * Call ``MongoClient.startSession``: the resulting session must have the same session ID as ``B``.
-    * Call ``MongoClient.startSession`` again: the resulting session must have the same session ID  as ``A``.
+1. Pool is LIFO
```
ShaneHarvey, line 1017: Similar to my style comment above, I think it would be better to make these format changes in a new PR.

source/sessions/driver-sessions.rst
```diff
@@ -1113,16 +1141,20 @@ ensure that they close any explicit client sessions and any unexhausted cursors.
...
+      * ``findOneAndDelete({ }),``
+      * ``findOneAndUpdate({ }, { $set: { a: 1 } }),``
+      * ``findOneAndReplace({ }, { a: 1 }),``
+      * ``find().toArray()``
+
     * Wait for all operations to complete
-    * Assert that all commands contain the same lsid
+    * Drivers MAY assert that exactly one session is used for all the concurrent operations listed, however this can be nondeterministic if the session isn't released before check in. Drivers SHOULD NOT attempt to release before check in.
+    * Drivers SHOULD assert that after repeated runs they are able to achieve the use of exactly one session, this will statistically prove we've reduced the allocation amount
+    * Drivers MUST assert that the number of allocated sessions is strictly less than the number of concurrent operations. In this instance it would less than (but NOT equal to) 8.
```
ShaneHarvey, line 1157: What do you think about:

```
    * Wait for all operations to complete successfully
    * Assert that all commands contain the same lsid. Note that it's possible, although very rare, for >1 server session to be used because the session is not released until after the connection is checked in. If this assertion fails, repeat this test 5 times until the assertion holds. 
```

source/sessions/driver-sessions.rst
```diff
@@ -1289,10 +1323,23 @@ is needed and, known it will be used, after connection checkout succeeds.
 It is still possible that via explicit sessions or cursors, which hold on to the session they started with, a driver could over allocate sessions.
 But those scenarios are extenuating and outside the scope of solving in this spec.
 
+Appendix
+========
+
+Drivers that DO NOT implement ServerSession pooling
```
ShaneHarvey, line 1329: I believe every driver already pools sessions. Can we remove this section?

ShaneHarvey (changes requested):
Since pooling is now required, we need to update this part of endSession as well:

> Conceptually, calling endSession implies ending the corresponding server session (by calling the endSessions command). As an implementation detail drivers SHOULD cache server sessions for reuse (see Server Session Pool).

source/sessions/driver-sessions.rst
nbbeeken: Is it not worth saving for historical reasons / new drivers? (despite it being a MUST now) 
I'm just thinking it is an important detail that you have to send the command over upon release as opposed to upon close.

source/sessions/driver-sessions.rst
nbbeeken: I added this phrasing / bullet but also kept what I had there are you thinking its overly verbose? I think it translates to actual code assertions well.

source/sessions/driver-sessions.rst
nbbeeken: Very good point! fixed.

source/sessions/driver-sessions.rst
nbbeeken: fixed the rendering here, and language

source/sessions/driver-sessions.rst
nbbeeken: I moved them to the Q&A section, I think they fit better there.

source/sessions/driver-sessions.rst
nbbeeken: Done!

source/sessions/driver-sessions.rst
nbbeeken: It's just really, really, really optional 🙂  (fixed)

source/sessions/driver-sessions.rst
nbbeeken: Done, lmk if the english here reads good 😁

source/sessions/driver-sessions.rst
nbbeeken: done

source/sessions/driver-sessions.rst
nbbeeken: reformatted (in the other PR) lmk if it captures it well, the `ClusterTime` interface is referenced in other message examples but yea it's a good reminder to bring up here that it's under a dollar prefix property.

source/sessions/driver-sessions.rst
```diff
@@ -502,12 +504,12 @@ include a session ID in a ``KILLCURSORS`` command.
 
 Sessions and Connections
 ========================
-A driver MUST only obtain an implicit session after it successfully checks out a connection.
-This limits the number of implicit sessions to never exceed the maximum connection pool size.
-The motivation for this behavior is to prevent too many sessions from being created in a scenario
-where only a limited number are actually needed to execute operations (i.e. TooManyLogicalSessions error).
+A driver MUST only obtain an implicit session's ``ServerSession`` after it successfully checks out a connection.
+A driver SHOULD NOT attempt to release the acquired session before connection check in.
+Through both through the pooling mechanism and limiting acquisition to a successful
```
dariakp, line 509: This sentence doesn't make sense to me. What exactly are we trying to say? Also, the motivation in the original text was clearly stated, and, I think helpful. So let's try to keep it, maybe clarifying that we are trying to limit the number of *server* sessions created.

source/sessions/driver-sessions.rst
```diff
@@ -801,15 +802,15 @@ corresponding ``ServerSession``. However, starting a server session might requir
...
+cache ``ServerSession`` instances for reuse. To this end drivers MUST
 implement a server session pool containing ``ServerSession`` instances
 available for reuse. A ``ServerSession`` pool MUST belong to a ``MongoClient``
 instance and have the same lifetime as the ``MongoClient`` instance.
 
-If a driver has a server session pool, then when a new ``ClientSession`` is started
-it MUST attempt to acquire a server session from the server session pool. See
-the algorithm below for the steps to follow when attempting to acquire a
-``ServerSession`` from the server session pool.
+When a new implicit ``ClientSession`` is started it MUST NOT attempt to acquire a server
```
dariakp, line 810: why did you change `MUST` to `MUST NOT`? if we can't acquire a server session from the pool, then where would we get the sessions?

source/sessions/driver-sessions.rst
dariakp: I think you should make the "Drivers MAY assert that exactly one session is used..." part a sub-bullet of the bullet above, since it clarifies the context, and then you can get rid of the SHOULD assert bullet point about repeated runs (but definitely keep the last bullet point about the strictly less part - it doesn't hurt, and it's a good sanity check)

source/sessions/driver-sessions.rst
dariakp: Maybe add a `for legacy reasons` to the first sentence? I don't have very strong feelings about this.

dariakp (changes requested):

source/sessions/driver-sessions.rst
ShaneHarvey: I don't think this section is worth keeping. I don't think there was ever a driver that didn't pool.

source/sessions/driver-sessions.rst
ShaneHarvey: I was suggesting to replace these 3 lines:
```
    * Drivers MAY assert that exactly one session is used for all the concurrent operations listed, however this can be nondeterministic if the session isn't released before check in. Drivers SHOULD NOT attempt to release before check in.
    * Drivers SHOULD assert that after repeated runs they are able to achieve the use of exactly one session, this will statistically prove we've reduced the allocation amount
    * Drivers MUST assert that the number of allocated sessions is strictly less than the number of concurrent operations. In this instance it would less than (but NOT equal to) 8.
```
I think it's a bit too verbose, for example the "Drivers SHOULD NOT attempt to release before check in." line doesn't really fit. I also believe the "strictly less than the number of concurrent operations" bullet is unneeded and confusing since we're already asserting that (eventually) only 1 session is used.

dariakp: The difference between "eventually one session is used" and "on every run the number of sessions is less than the number of concurrent operations" is that the latter guards against the unlikely but not impossible scenario of a driver having a race condition resulting in sometimes creating too many sessions. If we only keep your suggested phrasing, then the test would pass if we sometimes create one session and other times 16, and while we hope that no driver has such faulty logic, concurrent operations create the opportunity for all sorts of race conditions, so I think it's worth keeping a deterministic assertion that costs nothing to implement as a sanity check. 

But we certainly can and should update the phrasing to clarify that *this* assertion must be performed on every iteration, whereas the "exactly one" assertion just needs to happen once over the course of 5 tries.

ShaneHarvey: Okay I relent.

source/sessions/driver-sessions.rst
nbbeeken: I think the sentence cuts short, I added "immediately" the change is that specifically an implicit session needs to delay its acquiring until connection checkout but that's described in the "See the algorithm below" bit. LMK if this clears things up enough

source/sessions/driver-sessions.rst
nbbeeken: Updated the language

source/sessions/driver-sessions.rst
nbbeeken: SGTM, removed.

source/sessions/driver-sessions.rst
ShaneHarvey: Agreed. Maybe we can remove this sentence and add a bit of motivation to the first sentence:
> To reduce the number of ``ServerSession`` created, the driver MUST only obtain an implicit session's ``ServerSession`` after it successfully checks out a connection.

source/sessions/driver-sessions.rst
```diff
@@ -1143,17 +1142,21 @@ Test Plan
...
+      * ``bulkWrite([{ updateOne: { filter: { }, update: { $set: { a: 1 } } } }]),``
+      * ``findOneAndDelete({ }),``
+      * ``findOneAndUpdate({ }, { $set: { a: 1 } }),``
+      * ``findOneAndReplace({ }, { a: 1 }),``
+      * ``find().toArray()``
+
+    * Wait for all operations to complete successfully
+    * Assert the following across at least 5 retries of the above test:
+
+      * Drivers MUST assert that exactly one session is used for all operations at least once across the retries of this test.
```
ShaneHarvey, line 1157: This test should mention why we might need to repeat the test since the issue is subtle:
> Note that it's possible, although rare, for >1 server session to be used because the session is not released until after the connection is checked in.

source/sessions/driver-sessions.rst
```diff
@@ -1323,6 +1326,15 @@ is needed and, known it will be used, after connection checkout succeeds.
...
 But those scenarios are extenuating and outside the scope of solving in this spec.
 
+Why should drivers NOT attempt to release a serverSession before checking back in the operation's connection?
+-------------------------------------------------------------------------------------------------------------
+
+There are a variety of cases, such as, retryable operations or cursor creating operations
+where a ``serverSession`` must remain acquired by the ``ClientSession`` after an operation is attempted.
+Attempting to account for all these scenarios has risks that do not justify the potential guaranteed ``ServerSession`` allocation limiting.
+Drivers SHOULD attempt to release the ``ServerSession`` to the pool at the earliest possible opportunity.
+Drivers SHOULD attempt to reuse ``ServerSession`` as best as possible, concurrency fairness willing.
```
ShaneHarvey, line 1336: I don't think these two SHOULDs fit here since they appear to contradict the previous two sentences. Let's remove them?

ShaneHarvey (changes requested):

source/sessions/driver-sessions.rst
nbbeeken: Yea that is more concise, moved!

source/sessions/driver-sessions.rst
nbbeeken: agreed, done!

source/sessions/driver-sessions.rst
```diff
@@ -1143,17 +1142,24 @@ Test Plan
...
+      * ``find().toArray()``
+
+    * Wait for all operations to complete successfully
+    * Assert the following across at least 5 retries of the above test:
+
+      * Drivers MUST assert that exactly one session is used for all operations at least once across the retries of this test.
+
+        * Note that it's possible, although rare, for >1 server session to be used because the session is not released until after the connection is checked in.
+
+      * Drivers MUST assert that the number of allocated sessions is strictly less than the number of concurrent operations in every retry of this test. In this instance it would less than (but NOT equal to) 8.
```
dariakp, line 1161:
```suggestion
      * Drivers MUST assert that the number of allocated sessions is strictly less than the number of concurrent operations in every retry of this test. In this instance it would be less than (but NOT equal to) 8.
```

source/sessions/driver-sessions.rst
```diff
@@ -1323,6 +1329,13 @@ is needed and, known it will be used, after connection checkout succeeds.
 It is still possible that via explicit sessions or cursors, which hold on to the session they started with, a driver could over allocate sessions.
 But those scenarios are extenuating and outside the scope of solving in this spec.
 
+Why should drivers NOT attempt to release a serverSession before checking back in the operation's connection?
+-------------------------------------------------------------------------------------------------------------
+
+There are a variety of cases, such as, retryable operations or cursor creating operations
```
dariakp, line 1335:
```suggestion
There are a variety of cases, such as retryable operations or cursor creating operations,
```

dariakp (commented):
A couple small nits, but I am good with this as is, though let's also remember to update the change log date when this is about to merge

ShaneHarvey (approved):

rstam (approved):
LGTM

PR #1161, opened 2022-03-24
Fix getMore with comment specs

comandeo:
https://jira.mongodb.org/browse/DRIVERS-2243 - Fix OperationFailure on MongoDB <4.4.0
https://jira.mongodb.org/browse/DRIVERS-2244 - Test getMore with comment on all topologies

### DIFF
--- a/source/change-streams/tests/unified/change-streams.yml
+++ b/source/change-streams/tests/unified/change-streams.yml
@@ -140,7 +140,50 @@ tests:
   - description: "Test that comment is set on getMore"
     runOnRequirements:
       - minServerVersion: "4.4.0"
-        topologies: [ single, replicaset ]
+        topologies: [ replicaset ]
+    operations:
+      - name: createChangeStream
+        object: *collection0
+        arguments:
+          pipeline: []
+          comment: &commentDoc
+            key: "value"
+        saveResultAsEntity: &changeStream0 changeStream0
+      - name: insertOne
+        object: *collection0
+        arguments:
+          document: &new_document
+            _id: 1
+            a: 1
+      - name: iterateUntilDocumentOrError
+        object: *changeStream0
+    expectEvents:
+      - client: *client0
+        events:
+          - commandStartedEvent:
+              command:
+                aggregate: *collection0
+                pipeline:
+                  - $changeStream: {}
+                comment: *commentDoc
+          - commandStartedEvent:
+              command:
+                insert: *collection0
+                documents:
+                  - *new_document
+          - commandStartedEvent:
+              command:
+                getMore: { $$type: [ int, long ] }
+                collection: *collection0
+                comment: *commentDoc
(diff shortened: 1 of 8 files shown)


source/change-streams/tests/unified/change-streams.yml
```diff
@@ -140,7 +140,50 @@ tests:
   - description: "Test that comment is set on getMore"
     runOnRequirements:
       - minServerVersion: "4.4.0"
-        topologies: [ single, replicaset ]
```
jmikola, line 143: I've noted that `single` was introduced by https://github.com/mongodb/specifications/commit/ad1ee9b764f7d9ff193fe485d31feca536e6a2d7; however, this had no effect as the top-level `runOnRequirements` specified replica sets and sharded clusters only (https://github.com/mongodb/specifications/commit/962fb5f20e9e31b528319fe245147928a3100411#diff-ba3c94557f77e83f167fb27ed81eb223e4eb3e6178238f3d93def75f37adb301R3).

source/change-streams/tests/unified/change-streams.yml
```diff
@@ -140,7 +140,50 @@ tests:
...
+        topologies: [ replicaset ]
```
jmikola, line 143: Should you add a comment explaining why sharded clusters are excluded here (e.g. potentially empty `getMore` responses)?

If we do want to include sharded clusters, I suppose we could utilize `ignoreExtraEvents`, which was added in https://github.com/mongodb/specifications/commit/ff0c705f9cc95a5a9c8c595dbf9384cb41dfcdd2 for [DRIVERS-1713](https://jira.mongodb.org/browse/DRIVERS-1713); however, that would then require drivers to implement those test runner changes just to sync these tests.

The best course of action may be to just leave the comment here and open another DRIVERS ticket to come back to these tests and add `ignoreExtraEvents` to allow sharded clusters. That wouldn't be something you'd need to work on yourself -- reporting the issue and leaving it in the backlog would be sufficient just so we don't lose track of it.

jmikola (approved):
LGTM with added comment to the change stream test. As a sanity check I'd also ask that @ShaneHarvey POC these in Python before we merge.

baileympearson (approved):
I also tested these changes in the Node driver, the tests look good ([evergreen run](https://spruce.mongodb.com/version/623e083632f4173240ace899/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC))

juliusgeo:
Tested this in the Python driver--[evergreen run]

ShaneHarvey (approved):

source/change-streams/tests/unified/change-streams.yml
comandeo: Good point, thank you. Created a ticked and added corresponding comment to tests.

PR #1162, opened 2022-03-25
Warning-free build with Sphinx

vector-of-bool:
These changes allow all specification documents to render with Sphinx (Version 4) without generating errors or warnings. There are(should-be) no changes to the actual meaning of any documents. This is only the minimal changes to get Sphinx to generate the pages.

The current tool, `rstcheck`, is set to only report hard-errors, and only looks for semantic issues that Sphinx does not already report as errors/warnings.

### DIFF
--- a/source/auth/auth.rst
+++ b/source/auth/auth.rst
@@ -794,12 +794,12 @@ Client Second
 
 .. code:: javascript
 
-   { 
+   {
        "a" : "AWS4-HMAC-SHA256 Credential=AKIAICGVLKOKZVY3X3DA/20191107/us-east-1/sts/aws4_request, SignedHeaders=content-length;content-type;host;x-amz-date;x-mongodb-gs2-cb-flag;x-mongodb-server-nonce, Signature=ab62ce1c75f19c4c8b918b2ed63b46512765ed9b8bb5d79b374ae83eeac11f55",
        "d" : "20191107T002607Z"
        "t" : "<security_token>"
    }
-|
+
 Note that `X-AMZ-Security-Token` is required when using temporary credentials. When using regular credentials, it
 MUST be omitted. Each message above will be encoded as BSON V1.1 objects and sent to the peer as the value of
 ``payload``. Therefore, the SASL conversation would appear as:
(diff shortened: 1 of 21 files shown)


source/auth/auth.rst
```diff
@@ -834,10 +834,9 @@ Client Second:
        "conversationId" : 1,
        "payload" : new BinData(0, "LQEAAAJhAAkBAABBV1M0LUhNQUMtU0hBMjU2IENyZWRlbnRpYWw9QUtJQUlDR1ZMS09LWlZZM1gzREEvMjAxOTExMTIvdXMtZWFzdC0xL3N0cy9hd3M0X3JlcXVlc3QsIFNpZ25lZEhlYWRlcnM9Y29udGVudC1sZW5ndGg7Y29udGVudC10eXBlO2hvc3Q7eC1hbXotZGF0ZTt4LW1vbmdvZGItZ3MyLWNiLWZsYWc7eC1tb25nb2RiLXNlcnZlci1ub25jZSwgU2lnbmF0dXJlPThhMTI0NGZjODYyZTI5YjZiZjc0OTFmMmYwNDE5NDY2ZGNjOTFmZWU1MTJhYTViM2ZmZjQ1NDY3NDEwMjJiMmUAAmQAEQAAADIwMTkxMTEyVDIxMDEyMloAAA==")
    }
-|
 
 In response to the Server First message, drivers MUST send an ``authorization header``. Drivers MUST follow the
-`Signature Version 4 Signing Process <https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html>`_ to
+`Signature Version 4 Signing Process <https://docs.aws.amazon.com/general/latest/gr/signature-version-4.html>`__ to
```
ShaneHarvey, line 839: Why do these links need a double underscore? AFAIK links only have one underscore: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#hyperlinks

ShaneHarvey (commented):
Wow this is great! Any chance you can add a sphinx build to github actions?

source/auth/auth.rst
vector-of-bool: This is a subtle thing that I didn't know about until I saw warnings about it from Sphinx. A single underscore generates a named hyperlink target for the whole page, where the name of the target is the link text. There are two links in the page with the same link text but different URLs, and rST needs to pick one. Using two underscores generates an anonymous link, so no link target name is globally defined, and they are separate links.

vector-of-bool:
I have local changes that add a Sphinx HTML build, I was going to include those in a separate PR. Would you like them in this one as well?

ShaneHarvey (approved):
LGTM! A follow up PR to add CI support sounds good to me.

benjirewis (approved):
LGTM! Thanks :)

PR #1165, opened 2022-03-30
[DRIVERS-2241] Wording to add usage and API information for the csfle dynamic library

vector-of-bool:
These changes include updates for client-side-encryption on how to use and expose the `csfle` dynamic library to perform command-marking in place of `mongocryptd`.

This does _not_ yet include test cases for csfle.

Refer: [DRIVERS-2241](https://jira.mongodb.org/browse/DRIVERS-2241)

### DIFF
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -173,8 +173,8 @@ mongocryptd is a singleton local process needed for auto
 encryption. It speaks the MongoDB wire protocol and the driver uses
 mongocryptd by connecting with a MongoClient. By default, the driver
 will attempt to automatically spawn mongocryptd. If the MongoClient is
-configured with `extraOptions.mongocryptdBypassSpawn=true`, or 
-`AutoEncryptionOpts.bypassAutoEncryption=true` then the driver will not 
+configured with `extraOptions.mongocryptdBypassSpawn=true`, or
+`AutoEncryptionOpts.bypassAutoEncryption=true` then the driver will not
 attempt to spawn mongocryptd. The mongocryptd process is
 responsible for self terminating after idling for a time period.
 
@@ -305,7 +305,7 @@ See `What's the deal with metadataClient, keyVaultClient, and the internal clien
 
 keyVaultClient, metadataClient, and the internal MongoClient
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-The following pseudo-code describes the configuration behavior for the three ``MongoClient``s:
+The following pseudo-code describes the configuration behavior for the three ``MongoClient`` objects:
 
 .. code::
 
@@ -475,8 +475,8 @@ bypassAutoEncryption
 ^^^^^^^^^^^^^^^^^^^^
 
 Drivers MUST disable auto encryption when the 'bypassAutoEncryption'
-option is true and not try to spawn mongocryptd. Automatic encryption 
-may be disabled with the bypassAutoEncryption option. 
+option is true and not try to spawn mongocryptd. Automatic encryption
+may be disabled with the bypassAutoEncryption option.
 See `Why is there a bypassAutoEncryption?`_.
 
 extraOptions
@@ -792,8 +792,8 @@ encryption.
 
 Managing mongocryptd
 ====================
-If a MongoClient is configured for Client Side Encryption 
-(eg. bypassAutoEncryption=false), then by default 
+If a MongoClient is configured for Client Side Encryption
+(eg. bypassAutoEncryption=false), then by default
 (unless mongocryptdBypassSpawn=true), mongocryptd MUST be
 spawned by the driver. Spawning MUST include the command line argument
 --idleShutdownTimeoutSecs. If the user does not supply one through
@@ -864,7 +864,7 @@ The state machine is created with the libmongocrypt function
 ``mongocrypt_ctx_new`` and initialized with a ``mongocrypt_ctx_encrypt_init`` or
 ``mongocrypt_ctx_decrypt_init``. See the `libmongocrypt API documentation
 <https://github.com/mongodb/libmongocrypt/blob/master/src/mongocrypt.h.in>`_ for
-more information. 
+more information.
 
 An encrypted MongoClient configured with bypassAutoEncryption MUST NOT
 attempt automatic encryption for any command.
@@ -1330,7 +1330,7 @@ Why cache keys?
 We can't re-fetch the key on each operation, the performance goal for
 this project requires us to cache. We do need a revocation mechanism,
 based upon periodic checking from the client. Initially this window will
-not be configurable. See future work: `Make the key caching window configurable`__.
+not be configurable.
 
 Why require including a C library?
 ----------------------------------
@@ -1468,7 +1468,7 @@ Meaning if the first attempt to mongocryptd fails to connect, then the user
 would observe a 5 second delay. This is not configurable in the URI, so this
 must be overriden internally. Since mongocryptd is a local process, there should
 only be a very short delay after spawning mongocryptd for it to start listening
-on sockets. See the SDAM spec description of `cooldownMS <../source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#cooldownms>`_.
+on sockets. See the SDAM spec description of `cooldownMS`_.
 
 Because single threaded drivers may exceed ``serverSelectionTimeoutMS`` by the
 duration of the topology scan, ``connectTimeoutMS`` is also reduced.
@@ -1518,7 +1518,7 @@ The use pattern of the ``metadataClient`` will likely greatly differ from
 the parent ``MongoClient``. So it is configured with ``minPoolSize=0``.
 
 The ``metadataClient`` is not an exposed option because a user could
-misconfigure it to point to another MongoDB cluster, which could be a 
+misconfigure it to point to another MongoDB cluster, which could be a
 security risk.
 
 Why is the metadataClient not needed if bypassAutoEncryption=true
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -546,7 +546,7 @@ detail described in the `Implementation`_ section:
       mongocryptdSpawnArgs: Optional<Array[String]>
 
       // Override the path used to load the csfle library
-      csflePathOverride: Optional<string>;
+      csflePath: Optional<string>;
 
       // If 'true', refuse to continue encryption without a csfle library
       csfleRequired: boolean;
@@ -560,12 +560,12 @@ deprecating/removing options in the future without an API break, such as
 with a BSON document or map type instead of a struct type with fixed
 fields. See `Why are extraOptions and kmsProviders maps?`_.
 
-.. option:: extraOptions.csflePathOverride
+.. option:: extraOptions.csflePath
 
    :type: :ts:`undefined | string`
    :default: :ts:`undefined`
 
-   .. |opt-path-override| replace:: :option:`extraOptions.csflePathOverride`
+   .. |opt-path-override| replace:: :option:`extraOptions.csflePath`
 
    Allow the user to specify an absolute path to a csfle_ dynamic library to
    load. Refer:
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -517,9 +517,9 @@ example:
 Drivers MUST raise an error if the TLS options are set to disable TLS.
 The error MUST contain the message "TLS is required".
 
-error if insecure TLS options are set. The error MUST contain the message
-"Insecure TLS options prohibited". This includes options equivalent to the
-following :doc:`URI options </uri-options/uri-options>`:
+Drivers SHOULD raise an error if insecure TLS options are set. The error MUST
+contain the message "Insecure TLS options prohibited". This includes options
+equivalent to the following :doc:`URI options </uri-options/uri-options>`:
 
 - ``tlsInsecure``
 - ``tlsAllowInvalidCertificates``
@@ -628,7 +628,7 @@ fields.
    :type: :ts:`boolean`
    :default: |false|
 
-   If |true|, the driver should refuse to continue unless `csfle` was loaded
+   If |true|, the driver MUST refuse to continue unless `csfle` was loaded
    successfully.
 
    If, after initializing a `libmongocrypt_handle`, `csfle` is detected to be
@@ -1062,7 +1062,7 @@ Setting Search Paths
 
 The driver MUST append `csfle` search paths to the `libmongocrypt_handle`. For
 each path :math:`P` in |opt-paths|, append :math:`P` to the search paths on the
-`libmongocrypt_handle`, maintaining the order of the user-provided array..
+`libmongocrypt_handle`, maintaining the order of the user-provided array.
 
 
 .. rubric:: Explaination
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -17,7 +17,7 @@ Client Side Encryption
 
 .. _lmc-integrating: https://github.com/mongodb/libmongocrypt/blob/master/integrating.md
 
-.. default-role:: any
+.. default-role:: literal
 
 .. role:: ts(code)
    :language: typescript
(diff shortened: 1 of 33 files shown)


source/client-side-encryption/client-side-encryption.rst
```diff
@@ -770,73 +942,299 @@ behind-the-scenes operations required for encryption or decryption.
...
+After initializing the `libmongocrypt_handle`, the driver can detect whether
+`csfle` was successfully loaded by asking libmongocrypt for the csfle version
+string. If the result is an empty string, libmongocrypt did not load csfle and
+the driver must rely on `mongocryptd` to mark command documents for encryption.
+
+
+"Disabling" `csfle`
+-------------------
+
+`csfle` can be "disabled" on a `libmongocrypt_handle` by ommission:
```
addaleax, line 1118:
```suggestion
`csfle` can be "disabled" on a `libmongocrypt_handle` by omission:
```

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -770,73 +942,299 @@ behind-the-scenes operations required for encryption or decryption.
...
+   paths will have no effect.
+
+.. note::
+
+   If no |opt-paths-prefer| was specified or is an empty array, AND if no
+   |opt-paths-fallback| was specified or is an empty array, AND
+   `extraOptions.csfleDisableSystemLibrary <cse.extraOptions_>`__ was specified
+   as ``true``, the result should be that no csfle search paths will be added to
+   the `libmongocrypt_handle`, therefore libmongocrypt *will not* search for the
+   csfle library.
```
addaleax, line 1060: This means that the default behavior, with no `extraOptions` passed, will be that the driver passes only `$SYSTEM` to libmongocrypt using `mongocrypt_setopt_append_csfle_search_path()`, libmongocrypt will attempt to load the shared library, and if it fails to do so, the driver will attempt to spawn mongocryptd.

Is that correct? It means that an application can load either the shared library or mongocryptd through the same application code, only depending on whether they are using a version of the driver that includes this feature or not.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -770,73 +942,299 @@ behind-the-scenes operations required for encryption or decryption.
...
+-------------------
+
+The driver MUST append csfle search paths to the `libmongocrypt_handle` in the
+following order:
+
+1. Append each search path in |opt-paths-prefer|.
+2. If `extraOptions.csfleDisableSystemLibrary <cse.extraoptions_>`__ *was not*
+   specified as ``true`` by the user, append the literal string "``$SYSTEM``" to
+   the path list. (`Refer <cse.csfle.dollar-system_>`_)
+3. Append each search path in |opt-paths-fallback|.
```
addaleax, line 1046: In the server shared library planning docs, the expectation was that, in typical situations, the shared library would be deployed as part of application directories, not as part of the system libraries. 

Considering this, would it be worth simplifying this interface to merge the csfleSearchPathsPrefer/csfleSearchPathsFallback/csfleDisableSystemLibrary options and only specify a single list of search paths, which may or may not include `$SYSTEM`? We already need to make users aware that `$ORIGIN` exists as an expansion.

(If the desire here is to make the shared library an opt-out feature rather than an opt-in one, we could make specifying an empty search path list different from specifying no search path list at all, i.e. make the default `["$SYSTEM"]` rather than `[]`.)

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -770,73 +942,299 @@ behind-the-scenes operations required for encryption or decryption.
...
+
+If the following conditions are met:
+
+- The user's ``MongoClient`` is configured for client-side encryption (i.e.
+  ``bypassAutoEncryption != false``)
+- **AND** the user has not disabled ``mongocryptd`` spawning (i.e. with
+  ``mongocryptdBypassSpawn=true``)
+- **AND** the `cse.csfle` library is unavailable *OR* is disabled.
+
+**then** ``mongocryptd`` MUST be spawned by the driver.
```
addaleax, line 1184: Should there be a way to tell the driver “use a CSFLE shared library from a search path or fail”, i.e. to not fall back to mongocryptd? Right now, with `{ extraOptions: { mongocryptdBypassSpawn: true, csfleSearchPathsPrefer: ['/path1', '/path2'] } }`, you could tell the driver to search `/path1` and `/path2` for the shared library and not spawn mongocryptd, but you wouldn’t prevent it from attempting to connect to mongocryptd.

source/client-side-encryption/client-side-encryption.rst
vector-of-bool: This is correct according to my intentions when writing, but raises a good qeustion: I presumed "load and use csfle if possible" was the default so as to hide the command-marking process, but that may be surprising to users who currently expect `mongocrypd` to appear, even though we try to keep `mongocryptd`/`csfle` "hidden" as best we can.

source/client-side-encryption/client-side-encryption.rst
vector-of-bool: I interpretted the discussion as to whether it would be system-installed to be still tentative (although I would fall on the side of application-level deployment rather than system installation).

Given that we already need to document search path behavior, I think it would be feasible to merge these options and allow "set an empty path list" to be the "disable csfle" option. It will minimize the moving parts.

Although that still leaves an open question as to what the default should be (as related to your other comment). A default of `["$SYSTEM"]` is reasonable if we prefer a system-wide installation, but a default of `["$ORIGIN"]` would be better for application-level deployment (i.e. the user would only need to have `csfle` reside alongside the `libmongocrypt` library).

Either way, I like the idea to reduce the moving parts and merge these options and replace the "disable" toggle.

source/client-side-encryption/client-side-encryption.rst
vector-of-bool: Good point. I think this is a useful toggle to offer, although it will now require a few tweaks to the `libmongocrypt` library since I had not foreseen this.

vector-of-bool: Not sure what I was thinking earlier. This can be handled entirely within the driver since `libmongocrypt` isn't the one responsible for the spawning.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -242,70 +334,318 @@ MongoClient Changes
...
-   Automatic encryption is an enterprise only feature that only applies to
-   operations on a collection. Automatic encryption is not supported for
-   operations on a database or view, and operations that are not bypassed
-   will result in error (see `libmongocrypt: Auto Encryption Allow-List`_).
-   To bypass automatic encryption for all operations, set
-   bypassAutoEncryption=true in AutoEncryptionOpts.
+   Drivers MUST disable auto encryption when the 'bypassAutoEncryption' option
+   is |true| and not try to
+   `spawn mongocryptd <cse.managing-mongocryptd>` nor
+   `load csfle <cse.enabling-csfle>`. Automatic encryption may be
```
addaleax, line 371: Is it an explicit error to specify CSFLE library path options and `bypassAutoEncryption`, or will the CSFLE library path just be ignored? Either way is good for me.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -242,70 +334,318 @@ MongoClient Changes
...
+   :type: :ts:`boolean`
+   :default: |false|
+
+   If |true|, the driver should refuse to continue unless `csfle` was loaded
+   successfully.
+
+   If, after initializing a `libmongocrypt_handle`, `csfle` is detected to be
+   unavailable AND :option:`extraOptions.csfleRequired` is |true|, the driver
+   MUST consider the `libmongocrypt_handle` to be invalid and return an error to
+   the user. Refer:
```
addaleax, line 637: I like this solution a lot :slightly_smiling_face:

addaleax (approved):

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -752,91 +982,304 @@ See `Why do operations on views fail?`_.
...
+
+
+.. _search path:
+.. _search paths:
+.. _cse.csfle.search-paths:
+
+Setting Search Paths
+--------------------
+
+The driver MUST append `csfle` search paths to the `libmongocrypt_handle`. For
```
kevinAlbs, line 1063: This open question may be best answered from Java, C#, Python, or Node.JS teams. Those are the four drivers that [package libmongocrypt](https://docs.google.com/spreadsheets/d/1-RcZD3BqxxtiANoM5zeeye5jnkYn-bzqH_VfHXHSxhM/edit#gid=0).

Will drivers that package libmongocrypt need to prepend a custom search path to find `csfle`?

`csfle` has the [enterprise module license](https://github.com/10gen/mongo-enterprise-modules/blob/master/distsrc/LICENSE-Enterprise.txt). Whereas libmongocrypt is Apache licensed. That may motivate a separate package for `csfle`.

Consider [pymongocrypt](https://pypi.org/project/pymongocrypt/). If there is a separate `pymongocrypt-enterprise` package that only wraps `csfle`, then `libmongocrypt` and `csfle` may be installed in separate directories.

kevinAlbs (commented):
The additional refactoring is nice. Do you mind moving the larger chunks of refactoring to a separate PR? Other DBX engineers will likely refer to the PR to see the new content. I think it will be easier to interpret if the diff is limited to the new content.

source/client-side-encryption/client-side-encryption.rst
vector-of-bool: An option: Another special `"$DEFAULT"` path, which is replaced by the driver's own default, which should be documented. This could also be the sole default value of the search paths array in `extraOptions` (instead of `["$ORIGIN"]`)

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -752,91 +982,304 @@ See `Why do operations on views fail?`_.
...
-selection ever fails when connecting to mongocryptd, the server
-selection error is propagated to the user.
+
+If the following conditions are met:
+
+- The user's ``MongoClient`` is configured for client-side encryption (i.e.
+  `bypassAutoEncryption` is not |false|)
+- **AND** the user has not disabled `mongocryptd` spawning (i.e. by setting
+  `extraOptions.mongocryptdBypassSpawn` to |true|),
+- **AND** the `cse.csfle` library is unavailable,
```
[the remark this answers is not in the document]
vector-of-bool: I was curious about this aspect, as the spec doesn't mention _when_ in the process `mongocryptd` should be spawned, just that it _must_ be spawned. Should that be specified as part of the spec?

vector-of-bool:
@kevinAlbs I've stripped a lot of the not-necessary cleanup changes. Should be much more streamlined now.

source/client-side-encryption/client-side-encryption.rst
addaleax: Is the expectation here that drivers would release a separate package to their package management systems that contains the csfle shared library? If yes, then I think it would be okay to say that a driver MAY add a driver-specific path if it detects the presence of such a package?

vector-of-bool: That's an option, yes. the reason I'm thinking about ``$DEFAULT`` is to allow a user to control whether the driver behaves in this way or not, and to position the `$DEFAULT` in the priority they feels most appropriate for their application. e.g. a user can specify `["some_dir/", "$DEFAULT"]` or `["$DEFAULT", "some_dir/"]` depending on which they prefer.

kevinAlbs: I like the proposed behavior of `$DEFAULT`.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -790,10 +898,215 @@ MUST be distinguished in some way (e.g. exception type) to make it
...
+
+Setting Search Paths
+--------------------
+
+The driver MUST append csfle_ search paths to the `libmongocrypt_handle`. For
+each path :math:`P` in |opt-paths|, append :math:`P` to the search paths on the
+`libmongocrypt_handle`, maintaining the order of the user-provided array.
+
+
+.. rubric:: Explaination
```
kevinAlbs, line 938:
```suggestion
.. rubric:: Explanation
```

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -13,6 +13,19 @@ Client Side Encryption
 :Last Modified: 2022-02-24
```
kevinAlbs, line 13:
```suggestion
:Last Modified: 2022-03-31
```

kevinAlbs (approved):
LGTM pending the open question about `$DEFAULT`. Great work! Please re-request my review if you need further input.

I would like to wait until there is a C PoC to merge.

I have asked for a reviewer from the from a language team that packages libmongocrypt.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -61,14 +75,25 @@ schema
...
    A local process the driver communicates with to determine
    how to encrypt values in a command.
 
+csfle_
+   This initialism, spelled in all-lowercase, refers to the *client-side
+   field-level-encryption* dynamic library provided as part of a MongoDB
+   Enterprise distribution. It replaces mongocryptd_ as the method of
+   :ref:`marking-up a database command for encryption <subtype6.intent-to-encrypt>`.
+
+   .. seealso::
```
ShaneHarvey, line 92: Note sure if this was already mentioned but most RST features like `.. seealso` and `.. option` unfortunately do not work on Github's renderer so we will need to replace them with normal text. For example: https://github.com/mongodb/specifications/blob/9c8d7eafae935a681439113263d7a0f250923072/source/client-side-encryption/client-side-encryption.rst#csfle

ShaneHarvey (changes requested):

kevinAlbs:
Added @jyemin as a reviewer.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -242,70 +334,318 @@ MongoClient Changes
...
+
+   .. |opt-paths| replace:: :option:`extraOptions.csfleSearchPaths`
+
+   Allow the user to specify directories in which to search for `csfle`. Refer:
+
+   - :ref:`cse.csfle.search-paths`
+   - :ref:`cse.csfle.path-resolution`
+   - :ref:`cse.enabling-csfle`
+
+.. option:: extraOptions.csflePathOverride
```
kevinAlbs, line 577: Can users use `LD_LIBRARY_PATH`, `DYLD_LIBRARY_PATH`, and `PATH` to set a path to csfle? If "yes", can `csflePathOverride` be removed?

Drivers will not be packaging `csfle`. AFAIK the main use case of `csflePathOverride` is mongosh. @addaleax can mongosh use `LD_LIBRARY_PATH`, `DYLD_LIBRARY_PATH`, and `PATH` to specify the path to `csfle`?

kevinAlbs (commented):
After talking with @jyemin, I realize The DBX scope answers the question [Should csfle be available in package managers?](https://docs.google.com/document/d/1zjb7F01psR7im7SjmoqD_gR82js7T8Yt3tCcUe4RSaM/edit#heading=h.1dxovhtt6eyh) with "No".

I have two new open questions about simplifying the API. I have left comments inline.

source/client-side-encryption/client-side-encryption.rst
addaleax: > Can users use `LD_LIBRARY_PATH`, `DYLD_LIBRARY_PATH`, and `PATH` to set a path to csfle? If "yes", can `csflePathOverride` be removed?

I assume that, technically, the answer to the first question is “yes”. However, users may be reluctant – I certainly would be – to modify global state that affects external code loading (like these environment variables) for what is potentially a small part of their application.

If users were using these environment variables, I think they would replace `csfleSearchPaths` rather than `csflePathOverride`?

> AFAIK the main use case of `csflePathOverride` is mongosh. 

`csflePathOverride` is going to be used by mongosh, but I don’t see any reason why an application that bundles the shared library in a specific location and consequently can know the exact path to it wouldn’t want to use it.

> can mongosh use `LD_LIBRARY_PATH`, `DYLD_LIBRARY_PATH`, and `PATH` to specify the path to `csfle`?

No, the filename would still mismatch.

kevinAlbs: > I assume that, technically, the answer to the first question is “yes”. However, users may be reluctant – I certainly would be – to modify global state that affects external code loading (like these environment variables) for what is potentially a small part of their application.

That is a good point. It may not be a desirable option for users.

> No, the filename would still mismatch.

Good point. `csflePathOverride` is needed to specify the filename.

--

To simplify changes for drivers, is it an option is to only require the Node.JS driver support `csflePathOverride`? And not require drivers support `csfleSearchPaths`.

jyemin: What is the use case for an application (presumably mongosh or Compass?) needing a different file name than the default one?  If it's a narrow enough use case that we don't think likely to be common in normal applications, then we could leave it out of the spec and just implement it in the Node driver.

jyemin: I'm not sure how reluctant users will be to set `LD_LIBRARY_PATH`.  Any search path that is specified via this option would be externalized in _some_ way: you wouldn't want to embed the path(s) directly in the application.  If that's the case, LD_LIBRARY_PATH, etc. seems as good a mechanism as any for that, and it has the advantage of being a standard one that is already supported by dlopen.

I think we could leave this option out and wait to see if it's requested by users.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -498,13 +544,69 @@ detail described in the `Implementation`_ section:
...
+   :default: |false|
+
+   .. |opt-csfle-required| replace:: :option:`extraOptions.csfleRequired`
+
+   If |true|, the driver MUST refuse to continue unless csfle_ was loaded
+   successfully.
+
+   If, after initializing a `libmongocrypt_handle`, csfle_ is detected to be
+   unavailable AND |opt-csfle-required| is |true|, the driver
+   MUST consider the `libmongocrypt_handle` to be invalid and return an error to
```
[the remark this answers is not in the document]
addaleax: Currently, the `csfle-required` flag would not be passed to the C library. The driver would instead call `mongocrypt_init()` and check for the usage of the csfle library via `mongocrypt_csfle_version_string()`, and fail if the latter reports no csfle library being loaded.

I think this is okay – the driver will have to call the two methods in that order anyway since that’s also the way to determine whether mongocryptd needs to be spawned (or at least connected to).

source/client-side-encryption/client-side-encryption.rst
addaleax: > To simplify changes for drivers, is it an option is to only require the Node.JS driver support `csflePathOverride`? 

And let other drivers use something like `LD_LIBRARY_PATH`? I don’t know. I’m personally mostly concerned about the Node.js driver of course, but we already did the work for supporting `csflePathOverride` in libmongocrypt, and that is imo the cleanest way to specify which library to load (because, as mentioned, it’s a local option controlling local behavior). Adding driver bindings for this is trivial – for the Node.js driver it was 7 SLOC. I’d definitely be in favor of doing it.

> And not require drivers support `csfleSearchPaths`.

I think this would be fine either way – it’s also not hard to add support for it in the way that this PR currently suggests, since the bulk of the work here is already done. On the other hand, I think it’s likely that the applications which would use `csfleSearchPaths` would be able to use `csflePathOverride` just as well.

> What is the use case for an application (presumably mongosh or Compass?) needing a different file name than the default one? 

In mongosh we want to avoid possible conflicts of the csfle shared library shipped with mongosh and a possible system installation, and avoid accidentally making csfle available to users who have not specifically opted into using it by downloading it separately from mongosh (and thus entering into the enterprise license agreement). It’s the same thing we currently do with mongocryptd.

> you wouldn't want to embed the path(s) directly in the application

Why not?

> and it has the advantage of being a standard one that is already supported by dlopen.

This still feels like we’re asking for trouble. `LD_LIBRARY_PATH` needs to be set before the process starts, and its functionality can be disabled entirely depending on your selinux config. Users will set `DYLD_LIBRARY_PATH` without knowing that what they probably wanted to set was `DYLD_FALLBACK_FRAMEWORK_PATH`. We might get bug reports about `PATH` not working on Windows and have to tediously figure out that yet another, completely unrelated, part of their application overrode `PATH` after it had already been set, or accidentally used `:` instead of `;`.

I know this is all hypothetical, but I really think we should tell users to go with `csfleOverridePath`. They can still use the OS-specific library loading env vars if they really want to, but an explicit option should give us the best behavior in terms of consistency and diagnostics and is, for better or for worse, already implemented and easy to add driver bindings for.

vector-of-bool: Either way, setting one of these environment variables only has an effect on the behavior of the system's library search mechanism, and does not force the behavior i.e. it tells the loader to look in the given directories before it's own defaults, whereas `csfleOverridePath` would be an exact filepath given to `dlopen`/`LoadLibrary` and will be a clear hard-error if it names a non-existent path. If users specify any such environment variable, that will effectively inject those paths when the `$SYSTEM` search is performed.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -759,9 +861,9 @@ Drivers SHOULD take a best-effort approach to store sensitive data
 securely when interacting with KMS since responses may include decrypted
 data key material (e.g. use secure malloc if available).
 
-All errors from the MongoClient to mongocryptd MUST be distinguished in
-some way (e.g. exception type) to make it easier for users to
-distinguish when a command fails due to auto encryption limitations.
+All errors from the MongoClient to mongocryptd_ or the csfle_ error category
+MUST be distinguished in some way (e.g. exception type) to make it easier for
```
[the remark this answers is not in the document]
vector-of-bool: In the unified format, I am unsure. The modifications to this paragraph are just an update of the paragraph to include errors form `csfle`, whereas the previous was just to distinguish errors from `mongocryptd`, so I presume there was some prior confidence in the intent of this paragraph.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -790,10 +898,215 @@ MUST be distinguished in some way (e.g. exception type) to make it
...
+- If used, the `override path`_ must be given as a path to the csfle_ dynamic
+  library file *itself*, and not simply the directory that contains it.
+
+- If the `override path`_ is specified, the `search paths`_ will be ignored.
+
+- For csfle_ `search paths`_ and the `override path`_, if the given path is a
+  relative path and the first path component is the literal string
+  :ts:`"$ORIGIN"`, the :ts:`"$ORIGIN"` component will be replaced by the
+  absolute path to the directory containing the libmongocrypt_ library that is
+  performing the csfle_ search. This behavior mimics the ``$ORIGIN`` behavior of
```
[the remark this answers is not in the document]
vector-of-bool: Possibly a different solution in that case, but it may depend on the driver/appliation. The general idea is to allow applications to sit beside their dynamic libraries and enforce the dynamic linker to prefer the once that it is bundled with the application rather than the ones on the system. Of course, on systems that support `RPATH` they can just set their app's `RPATH` to `$ORIGIN/../lib`, but not all languages will be able to set the `RPATH`, in which can `LD_LIBRARY_PATH`.

source/client-side-encryption/client-side-encryption.rst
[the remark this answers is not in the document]
kevinAlbs: Currently no. Testing that errors are distinguishable is requested in [DRIVERS-2138](https://jira.mongodb.org/browse/DRIVERS-2138).

source/client-side-encryption/client-side-encryption.rst
vector-of-bool: I've pushed changes that remove `csfleSearchPaths` and I've opened a PR on `libmongocrypt` that removes the functionality as well. This necessitates a new option to disable `csfle` since "don't specify any search paths" is no longer an option.

jyemin: Are we certain it's a requirement to be able to disable it explicitly?

vector-of-bool: I seem to recall someone addressing that aspect explicitly, but I cannot find where it was. We can remove the option as well.

addaleax: I think the prior discussion you are referring to was https://github.com/mongodb/specifications/pull/1165#discussion_r838338960?

I think it makes sense to give the users the option to explicitly specify not using the shared library in some way in order to get consistent behavior across installations. It wouldn’t need to be an extra option in every driver – e.g. in Node.js it could be a tristate `'require-csfle' | 'try-load-csfle' | 'ignore-csfle'` or similar.

kevinAlbs: > Are we certain it's a requirement to be able to disable it explicitly?

It is not a requirement to have an option for disabling searching for `csfle`.

It may be a good precaution. If `csfle` and `mongocryptd` are both on the system, `csfleDisabled` may be the only way to force using `mongocryptd`. Maybe it could be useful in the event that `csfle` has a regression not present on `mongocryptd`. Are there other use-cases?

addaleax (approved):
Just approving again to re-affirm that the current state of this PR LGTM

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: I am OK with the currently proposed options (`csflePath`, `csfleRequired`, and `csfleDisabled`). I think arguably we could remove `csfleRequired` and `csfleDisabled`, but they may provide value to users in the event that mongocryptd and csfle are both on the system.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -790,10 +897,169 @@ MUST be distinguished in some way (e.g. exception type) to make it
...
+   |false|).
+2. AND the initialization of that `libmongocrypt_handle` does not successfully
+   find the same csfle_ library that was loaded by the existing
+   `libmongocrypt_handle` that is already using csfle_.
+
+Drivers MUST document this limitation for users along with the documentation on
+the ``csfle*`` options in extraOptions_ by including the following:
+
+   All `MongoClient` objects in the same process should use the same setting for
+   |opt-path-override|, as it is an error ot load more that one csfle_ dynamic
```
kevinAlbs, line 1024:
```suggestion
   |opt-path-override|, as it is an error to load more that one csfle_ dynamic
```

kevinAlbs (approved):

vector-of-bool:
I've revived the search-paths functionality in a limited fashion, not for the user-API, but to be an option to driver testing. Being able to allow drivers more control over the `csfle` search behavior is highly beneficial for testing purposes.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -498,13 +544,68 @@ detail described in the `Implementation`_ section:
...
+   unavailable AND |opt-csfle-required| is |true|, the driver
+   MUST consider the `libmongocrypt_handle` to be invalid and return an error to
+   the user. Refer:
+
+   - `Enabling csfle`_
+   - `Managing mongocryptd`_
+   - `Detecting csfle Availability`_
+
+
+.. option:: extraOptions.csfleDisabled
```
jyemin, line 598: Please consider removing this option.

Rationale: in the near future, I would like to be able to remove support for mongocryptd entirely (in both server and drivers).  An API like this (even if only in extraOptions) could make that more difficult, as it may require a round of deprecations/major releases to get there.  If we leave it out, the only new options are ones that could more easily be claimed to be supported in a future worlds with no mongocryptd (or even a world where clfle is compiled into libmongocrypt).

We can always add this option in the future if its requested by users experiencing actual deployment pain.

jyemin (changes requested):

source/client-side-encryption/client-side-encryption.rst
vector-of-bool: Removed those.

Aside: I have a branch that supports full HTML rendering and I'm also interested in getting a way to automatically publish such HTML pages to an easily-browse-able site (perhaps gh-pages?). Maybe someday...

jyemin (approved):

source/client-side-encryption/client-side-encryption.rst
ShaneHarvey: Thanks!

ShaneHarvey (approved):

PR #1166, opened 2022-03-31
DRIVERS-1926 Improve stability of server selection load balancing algorithm test

patrickfreed:
DRIVERS-1926

This PR makes some modifications to the prose test for the operationCount-based load balancing algorithm used in server selection intended to reduce the frequency of false negative failures. From [initial tests](https://evergreen.mongodb.com/version/6244f04457e85a04f1ff7ff6), they seem to work well for the Rust driver.

### DIFF
--- a/source/server-selection/server-selection-tests.rst
+++ b/source/server-selection/server-selection-tests.rst
@@ -296,7 +296,7 @@ Multi-threaded and async drivers MUST also implement the following prose test:
             blockTimeMS: 500,
             appName: "loadBalancingTest",
         },
-    }
+     }
 
 3. Insert an empty document into the collection being used for the test.
 
@@ -316,7 +316,7 @@ Multi-threaded and async drivers MUST also implement the following prose test:
 8. Disable the failpoint.
 
 9. Start 10 concurrent threads / tasks that each run 100 `findOne` operations
-    with empty filters using that client.
+   with empty filters using that client.
 
 10. Using command monitoring events, assert that each mongos was selected
     roughly 50% of the time (within +/- 10%).
--- a/source/server-selection/server-selection-tests.rst
+++ b/source/server-selection/server-selection-tests.rst
@@ -281,11 +281,11 @@ Prose Test
 
 Multi-threaded and async drivers MUST also implement the following prose test:
 
-1. Configure a sharded cluster with two mongoses. Use a 4.2.9 or newer server version.
+#. Configure a sharded cluster with two mongoses. Use a 4.2.9 or newer server version.
    This test should not be run using TLS or authentication to reduce the chance of
    a false test failure.
 
-2. Enable the following failpoint against exactly one of the mongoses::
+#. Enable the following failpoint against exactly one of the mongoses::
 
      {
         configureFailPoint: "failCommand",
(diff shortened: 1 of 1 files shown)


source/server-selection/server-selection-tests.rst
```diff
@@ -296,20 +298,28 @@ Multi-threaded and async drivers MUST also implement the following prose test:
...
     }
 
-3. Create a client with both mongoses' addresses in its seed list,
-   appName="loadBalancingTest", and command monitoring enabled.
+3. Insert an empty document into the collection being used for the test.
 
-4. Start 10 concurrent threads / tasks that each run 10 `findOne` operations
-   using that client.
+4. Create a client with both mongoses' addresses in its seed list,
+   appName="loadBalancingTest", and localThresholdMS=30000.
```
patrickfreed, line 304: the high localThresholdMS should help avoid cases where a single mongos happens to get all the selections due to a random spike in network latency.

source/server-selection/server-selection-tests.rst
```diff
@@ -296,20 +298,28 @@ Multi-threaded and async drivers MUST also implement the following prose test:
...
 
-5. Assert that fewer than 25% of the CommandStartedEvents occurred on the mongos
-   that the failpoint was enabled on.
+5. Using CMAP events, ensure the client's connection pools for both
```
patrickfreed, line 306: This helps reduce any noise introduced by connection establishment latency during the actual server selection tests.

source/server-selection/server-selection-tests.rst
p-mongo: Do you think it would be helpful to include these notes in the actual text for future implementers/to help troubleshoot failures?

p-mongo (approved):

source/server-selection/server-selection-tests.rst
patrickfreed: Good idea, added

source/server-selection/server-selection-tests.rst
```diff
@@ -294,22 +296,37 @@ Multi-threaded and async drivers MUST also implement the following prose test:
...
+   mongoses have been saturated, either via setting minPoolSize=maxPoolSize or
+   executing operations.
 
-7. Repeat this test without any failpoints and assert that each mongos was
-   selected roughly 50% (within +/- 10%) of the time.
+   * This helps reduce any noise introduced by connection establishment latency
+     during the actual server selection tests.
 
+6. Start 10 concurrent threads / tasks that each run 10 `findOne` operations
```
BorisDog, line 317: .Net implementation test got stable only at 100 iterations, should we increase the number of iterations or mention that drivers might need to fine-tune it?

source/server-selection/server-selection-tests.rst
```diff
@@ -282,6 +282,8 @@ Prose Test
 Multi-threaded and async drivers MUST also implement the following prose test:
 
 1. Configure a sharded cluster with two mongoses. Use a 4.2.9 or newer server version.
+   This test should not be run using TLS or authentication to reduce the chance of
+   a false test failure.
```
stIncMale, line 286: Why does TLS/authentication affect the probability of a failure? It'd be helpful to mention the explanation in the test.

source/server-selection/server-selection-tests.rst
```diff
@@ -294,22 +296,37 @@ Multi-threaded and async drivers MUST also implement the following prose test:
             blockTimeMS: 500,
             appName: "loadBalancingTest",
         },
-    }
+     }
 
-3. Create a client with both mongoses' addresses in its seed list,
-   appName="loadBalancingTest", and command monitoring enabled.
+3. Insert an empty document into the collection being used for the test.
```
stIncMale, line 301: Why do we need to have a document in the collection?

source/server-selection/server-selection-tests.rst
patrickfreed: I'm not actually certain--I think maybe they contributed to the random latency spikes that caused one server to fall out of the latency window? Now that we have a high localThresholdMS that might not be a problem anymore. For reference, I just tried out the new version of the test with auth + TLS enabled and had no failures, so maybe this isn't required anymore. I know .Net said skipping auth + TLS helped a lot, so I wonder if @BorisDog could try this new version out with auth/TLS enabled and see if skipping is still necessary.

source/server-selection/server-selection-tests.rst
BorisDog: Sorry, just for the second test, missed that.

source/server-selection/server-selection-tests.rst
BorisDog: Yes, will do.

BorisDog: With the proposed changes, the test passes successfully on auth topologies. I believe that non-auth requirement can be removed.

patrickfreed: Ok nice, updated to remove the requirement.

stIncMale: Thank you, Boris, for checking!

stIncMale (approved):

BorisDog (approved):

PR #1167, opened 2022-03-31
DRIVERS-2236 Clarify load-balanced pool clearing behavior

patrickfreed:
DRIVERS-2236

This PR clarifies the differences in behavior between load balanced pools and non-load balanced pools. No driver or test changes should be required for this as it's just a wording change.

### DIFF
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -699,6 +699,14 @@ Otherwise, the `Connection <#connection>`_ is marked as available.
 Clearing a Connection Pool
 --------------------------
 
+Clearing the pool involves different steps depending on whether the pool is in
+load balanced mode or not. The traditional / non-load balanced clearing behavior
+MUST NOT be used by pools in load balanced mode, and the load balanced pool
+clearing behavior MUST NOT be used in non-load balanced pools.
+
+Clearing a non-load balanced pool
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
 A Pool MUST have a method of clearing all `Connections <#connection>`_ when
 instructed. Rather than iterating through every `Connection <#connection>`_,
 this method should simply increment the generation of the Pool, implicitly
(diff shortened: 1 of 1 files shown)


source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -727,6 +728,16 @@ eagerly so that any operations waiting on `Connections <#connection>`_ can retry
 as soon as possible. The pool MUST NOT rely on WaitQueueTimeoutMS to clear
 requests from the WaitQueue.
 
+Clearning a load balanced pool
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A Pool MUST also have a method of clearing all `Connections <#connection>`_ for
+a specific ``serviceId`` for use when in load balancer mode. This method
```
ShaneHarvey, line 735: Could you outline what drivers are intended to do when clearing the pooling in load balanced mode _without_ a serviceId? For example, when a connection handshake fails.

ShaneHarvey (changes requested):

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -727,6 +728,16 @@ eagerly so that any operations waiting on `Connections <#connection>`_ can retry
...
+increments the generation of the pool for that specific ``serviceId`` in the
+generation map. A PoolClearedEvent MUST be emitteed after incrementing the
```
BorisDog, line 737: typo: `emitted`

BorisDog (approved):
minor typo and LTGM

ShaneHarvey (approved):
This change LGTM but note that it does not solve the original issue in DRIVERS-2236. 

The problem is that this load balancer test conflicts with the spec:
https://github.com/mongodb/specifications/blob/008d2f5/source/load-balancers/tests/sdam-error-handling.yml#L175-L205

The test expects a handshake error to clear the pool, but the spec says to ignore handshake errors. Feel free to reassign the ticket with the load balancer component.

patrickfreed:
Ah I see what you mean. I'll reassign the ticket then and move it back into needs triage.

patrickfreed:
Oh wait, it looks like the spec uses a slightly different notion of "pre/post handshake" than SDAM does, I [remember running into this a while back](https://app.slack.com/client/T024FNNHU/C01C3KRQZ9P/thread/C01C3KRQZ9P-1628195792.005500). LB considers the handshake to just be the `hello` part, and explicitly requires other parts of connection establishment after the initial `hello` (e.g. auth) to trigger pool clearing, meaning the test is correct:

>Once the initial handshake is complete, the connection MUST determine its generation number based on the serviceId field in the handshake response. Any errors that occur during the rest of connection establishment (e.g. errors during authentication commands) MUST go through the SDAM error handling flow but MUST NOT mark the server as Unknown and when requiring the connection pool to be cleared, MUST only clear connections for the serviceId.

From the ["Initial Handshake Errors"](https://github.com/mongodb/specifications/blob/master/source/load-balancers/load-balancers.rst#initial-handshake-errors) section.

ShaneHarvey:
Oh thanks for explaining that! The tests make sense in that case but it is a bit confusing to re-use the term "handshake" with a different meaning.

PR #1168, opened 2022-03-31
DRIVERS-1954: Minor fixes

BorisDog:


nbbeeken (approved):
LGTM, thanks for the fixes!
@ShaneHarvey I figured we don't need a new ticket for these changes since we know everyone's blocked on the other work, so no one should fall out of sync. I can update the Downstream Changes Summary field to call this out after we merge.

ShaneHarvey (approved):
SGTM.

PR #1169, opened 2022-04-04
DRIVERS-1343 Fix topology and delete rename target

abr-egn:


source/change-streams/tests/unified/change-streams.yml
```diff
@@ -227,16 +227,21 @@ tests:
...
     operations:
       - name: createChangeStream
         object: *collection0
         arguments:
           pipeline: []
         saveResultAsEntity: &changeStream0 changeStream0
+      - name: dropCollection
+        object: *database0
+        arguments:
+          collection: &collection1 collection1
```
jmikola, line 240: I'm not sure what tests introduced this, but I think it's preferable to use separate references for entity IDs and the database and collection names.

Unified tests have typically done something like the following: https://github.com/mongodb/specifications/blob/98db467cb8f2f7bf50e79fc563e59701a8b3e3bf/source/unified-test-format/tests/valid-pass/poc-change-streams.yml#L14-L21

I didn't bother to mention this in https://github.com/mongodb/specifications/pull/1156 but it applies there as well.

abr-egn: Ah, makes sense.  And yes, I'd prefer to do that as a single large batch in that PR.

source/change-streams/tests/unified/change-streams.yml
```diff
@@ -2,7 +2,7 @@ description: "change-streams"
 schemaVersion: "1.0"
 runOnRequirements:
   - minServerVersion: "3.6"
-  - topologies: [ replicaset, sharded-replicaset ]
```
jmikola, line 5: Oh, I didn't realize this typo before. Good catch!

jmikola (approved):

PR #1172, opened 2022-04-06
DRIVERS-1995:  Restrict "test server error on projecting out _id" test to 4.2 an higher

BorisDog:


benjirewis (approved):
LGTM! Thanks for fixing those typos, too.

PR #1175, opened 2022-04-08
DRIVERS-2260 Add CSFLE spec test for auto encryption on a collection …

kevinAlbs:
…with no jsonSchema

# Background & Motivation

MONGOCRYPT-378 introduced [a regression](https://github.com/mongodb/libmongocrypt/pull/282) that only applied to collections with no local or remote jsonSchema configured. This was resolved and tested within libmongocrypt. This PR adds an integration test for the missing coverage.

The new test was run with the Go driver in https://github.com/mongodb/mongo-go-driver/pull/903.


comandeo (approved):

ShaneHarvey (approved):
LGTM the test passes in pymongo

PR #1176, opened 2022-04-08
DRIVERS-1915: Change stream support for point-in-time pre and post-images

jmikola:
https://jira.mongodb.org/browse/DRIVERS-1915

POC: https://github.com/mongodb/mongo-php-library/pull/911

Note: testing may require starting mongod with `--setParameter featureFlagChangeStreamPreAndPostImages=true` if the "latest" server build does not include [SERVER-52282](https://jira.mongodb.org/browse/SERVER-52282).

### DIFF
--- a/source/change-streams/change-streams.rst
+++ b/source/change-streams/change-streams.rst
@@ -361,11 +361,11 @@ Driver API
 
   class ChangeStreamOptions {
     /**
-     * Allowed values: 'updateLookup', 'whenAvailable', 'required'.
+     * Allowed values: 'default', 'updateLookup', 'whenAvailable', 'required'.
      *
-     * The default is to not send a value. By default, the change notification
-     * for partial updates will include a delta describing the changes to the
-     * document.
+     * The default is to not send a value, which is equivalent to 'default'. By
+     * default, the change notification for partial updates will include a delta
+     * describing the changes to the document.
      *
      * When set to 'updateLookup', the change notification for partial updates
      * will include both a delta describing the changes to the document as well
--- a/source/change-streams/change-streams.rst
+++ b/source/change-streams/change-streams.rst
@@ -9,8 +9,8 @@ Change Streams
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: 3.6
-:Last Modified: 2022-03-25
-:Version: 1.13.1
+:Last Modified: 2022-04-08
+:Version: 1.14
 
 .. contents::
 
@@ -175,20 +175,34 @@ If an aggregate command with a ``$changeStream`` stage completes successfully, t
     updateDescription: Optional<UpdateDescription>;
 
     /**
-     * Always present for operations of type ‘insert’ and ‘replace’. Also
-     * present for operations of type ‘update’ if the user has specified ‘updateLookup’
-     * in the ‘fullDocument’ arguments to the ‘$changeStream’ stage.
+     * Always present for operations of type 'insert' and 'replace'. Also
+     * present for operations of type 'update' if the user has specified
+     * 'updateLookup' for the 'fullDocument' option when creating the change
+     * stream.
      *
-     * For operations of type ‘insert’ and ‘replace’, this key will contain the
-     * document being inserted, or the new version of the document that is replacing
-     * the existing document, respectively.
+     * For operations of type 'insert' and 'replace', this key will contain the
+     * document being inserted or the new version of the document that is
+     * replacing the existing document, respectively.
      *
-     * For operations of type ‘update’, this key will contain a copy of the full
+     * For operations of type 'update', this key will contain a copy of the full
      * version of the document from some point after the update occurred. If the
      * document was deleted since the updated happened, it will be null.
+     *
+     * Contains the point-in-time post-image of the modified document if the
+     * post-image is available and either "required" or "whenAvailable" was
+     * specified for the "fullDocument" option when creating the change stream.
+     * A post-image is always available for "insert" and "replace" events.
      */
     fullDocument: Document | null;
 
+    /**
+     * Contains the pre-image of the modified or deleted document if the
+     * pre-image is available for the change event and either "required" or
+     * "whenAvailable" was specified for the "fullDocumentBeforeChange" option
+     * when creating the change stream. If "whenAvailable" was specified but the
+     * pre-image is unavailable, this will be explicitly set to null.
+     */
+    fullDocumentBeforeChange: Document | null;
   }
 
   class UpdateDescription {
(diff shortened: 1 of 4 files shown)


source/change-streams/change-streams.rst
```diff
@@ -175,20 +175,34 @@ If an aggregate command with a ``$changeStream`` stage completes successfully, t
...
+     * document being inserted or the new version of the document that is
+     * replacing the existing document, respectively.
      *
-     * For operations of type ‘update’, this key will contain a copy of the full
+     * For operations of type 'update', this key will contain a copy of the full
      * version of the document from some point after the update occurred. If the
      * document was deleted since the updated happened, it will be null.
+     *
+     * Contains the point-in-time post-image of the modified document if the
+     * post-image is available and either "required" or "whenAvailable" was
```
benjirewis, line 192: [opt-nit] This was already variable in the spec, but maybe we should pick `'` or `"`. `"required"` is `'required'` in `ChangeStreamOptions` below.

source/change-streams/change-streams.rst
```diff
@@ -347,16 +361,52 @@ Driver API
...
+     * this option.
+     *
      * @note this is an option of the `$changeStream` pipeline stage.
      */
     fullDocument: Optional<String>;
 
+    /**
+     * Allowed values: 'whenAvailable', 'required', 'off'.
+     *
+     * The default is to not send a value, which is equivalent to 'off'.
```
benjirewis, line 393: Not our problem but curious why the server even has an explicit `off` option.

source/change-streams/change-streams.rst
```diff
@@ -946,3 +996,6 @@ Changelog
 +------------+------------------------------------------------------------+
 | 2022-03-25 | Do not error when parsing change stream event documents.   |
 +------------+------------------------------------------------------------+
+| 2022-04-08 | Support returning point-in-time pre and post-images with   |
+|            | fullDocumentBeforeChange and fullDocument.                 |
```
benjirewis, line 1000:
```suggestion
|            | ``fullDocumentBeforeChange`` and ``fullDocument``.                 |
```

source/change-streams/tests/unified/change-streams-pre_and_post_images.yml
```diff
@@ -0,0 +1,359 @@
...
+          - commandStartedEvent:
+              command:
+                aggregate: *collection0Name
+                pipeline:
+                  - $changeStream: { fullDocument: "required" }
+
+
+
+
+
```
benjirewis, line 166: Is this extra space necessary in the yaml? I see it's delineating `fullDocument` and `fullDocumentBeforeChange` tests but I haven't seen something like this before.

source/change-streams/tests/unified/change-streams-pre_and_post_images.yml
```diff
@@ -0,0 +1,359 @@
...
+          updateDescription: { $$type: "object" }
+          fullDocumentBeforeChange: { _id: 1 }
+    expectEvents:
+      - client: *client0
+        events:
+          - commandStartedEvent:
+              command:
+                aggregate: *collection0Name
+                pipeline:
+                  - $changeStream: { fullDocumentBeforeChange: "whenAvailable" }
```
benjirewis, line 202: We still send `fullDocument: default` by default, right? Shouldn't this be `- $changeStream: { fullDocument: "default", fullDocumentBeforeChange: "whenAvailable" }` (and similar for all tests below)? This tests fails in my POC in the Go driver because the `fullDocument: default` is seen as an extra key in the `$changeStream` pipeline stage.

jmikola: [DRIVERS-2231](https://jira.mongodb.org/browse/DRIVERS-2231) changed spec tests to no longer expect `fullDocument: "default"`. [GODRIVER-2329](https://jira.mongodb.org/browse/GODRIVER-2329) was the related Go driver ticket.

I just linked that issue as "has to be done before" [DRIVERS-1915](https://jira.mongodb.org/browse/DRIVERS-1915), as I previously did with other change stream tickets (e.g. converted unified spec tests).

source/change-streams/tests/unified/change-streams-pre_and_post_images.yml
jmikola: Unintentional. I'll remove this.

source/change-streams/change-streams.rst
jmikola: Fixing this manually in order to preserve the table alignment.

source/change-streams/change-streams.rst
jmikola: Likely the same reason `fullDocument` originally supported a "default" string. IIRC, that was only required in pre-release versions. By the time change streams appeared in a stable release, "default" was equivalent to leaving it unspecified and there was no longer any purpose for the value.

source/change-streams/change-streams.rst
jmikola: Changed to single quotes.

source/change-streams/change-streams.rst
```diff
@@ -347,16 +361,52 @@ Driver API
...
     /**
-     * Allowed values: ‘updateLookup’.  When set to ‘updateLookup’, the change notification
-     * for partial updates will include both a delta describing the changes to the document,
-     * as well as a copy of the entire document that was changed from some time after the
-     * change occurred. The default is to not send a value.
-     * For forward compatibility, a driver MUST NOT raise an error when a user provides an
-     * unknown value. The driver relies on the server to validate this option.
+     * Allowed values: 'default', 'updateLookup', 'whenAvailable', 'required'.
+     *
+     * The default is to not send a value, which is equivalent to 'default'. By
```
jmikola, line 366: Decided to document "default" for posterity, since we're doing the same for "off" with `fullDocumentBeforeChange`. The important line here is "not to send a value" (from [DRIVERS-2231](https://jira.mongodb.org/browse/DRIVERS-2231)), so this will not impact drivers.

source/change-streams/tests/unified/change-streams-pre_and_post_images.yml
benjirewis: Ah interesting; looks like we didn't expect the `fullDocument: "default"` part of the `$changeStream` stage in our legacy change stream spec test runner. I'll modify our unified test runner to not expect it either as part of https://jira.mongodb.org/browse/GODRIVER-2294.

benjirewis (approved):
LGTM. I only wonder if the file names should be `change-streams-pre-and-post-images.*` instead of `change-streams-pre_and_post_images`. The mix of hyphen and underscore seems nonstandard.

source/change-streams/tests/unified/change-streams-pre_and_post_images.yml
```diff
@@ -0,0 +1,350 @@
+description: "change-streams-pre_and_post_images"
+
+schemaVersion: "1.0"
+
+runOnRequirements:
+  - minServerVersion: "6.0.0"
+    topologies: [ replicaset, sharded-replicaset, load-balanced ]
```
benjirewis, line 7: I believe the CI is failing because of this line... Not sure why, though.

jmikola:
> I only wonder if the file names should be change-streams-pre-and-post-images.* instead of change-streams-pre_and_post_images. The mix of hyphen and underscore seems nonstandard.

I based this on `findOneAndReplace-dots_and_dollars.yml`. IMO, hyphens are for separating the logical blocks (e.g. method and option/feature) and underscores are in place of spaces or other special characters.

I object to all of the existing tests having `change-streams-` as a prefix, but I wasn't about to change that.

source/change-streams/tests/unified/change-streams-pre_and_post_images.yml
jmikola: Looking into this now. I just realized change-streams were never added to the Unified Test Format Makefile and just did so now. It validates locally, so I'll have to sort out how the CI task differs.

PR #1179, opened 2022-04-13
DRIVERS-2188 Add comment to estimatedDocumentCount

comandeo:


source/crud/tests/unified/countDocuments-comment.yml
```diff
@@ -0,0 +1,70 @@
+description: "count"
+
+schemaVersion: "1.0"
+
+createEntities:
+  - client:
+      id: &client0 client0
+      useMultipleMongoses: false # Avoid setting fail points with multiple mongoses
+      uriOptions: { retryReads: false } # Avoid retrying fail points with closeConnection
```
jmikola, line 9: Are these two lines necessary? I don't see any fail points used in this file.

source/crud/tests/unified/countDocuments-comment.yml
```diff
@@ -0,0 +1,70 @@
+description: "count"
```
jmikola, line 1:
```suggestion
description: "countDocuments-comment"
```

Observation: I see we have existing `countDocuments` tests in `tests/v1`, but this is the first unified test for the method.

source/crud/tests/unified/countDocuments-comment.yml
```diff
@@ -0,0 +1,70 @@
...
+initialData:
+  - collectionName: *collection0Name
+    databaseName: *database0Name
+    documents:
+      - { _id: 1, x: 11 }
+      - { _id: 2, x: 22 }
+      - { _id: 3, x: 33 }
+
+tests:
+  - description: "count with comment on 4.4.0 or greater"
```
jmikola, line 29:
```suggestion
  - description: "countDocuments with document comment"
```

source/crud/tests/unified/countDocuments-comment.yml
```diff
@@ -0,0 +1,70 @@
...
+        events:
+          - commandStartedEvent:
+              command:
+                aggregate: *collection0Name
+                pipeline: [{'$match': {}}, {'$group': {'_id': 1, 'n': {'$sum': 1}}}]
+                comment: *objectComment
+              commandName: aggregate
+              databaseName: *database0Name
+
+  - description: "count with comment on less than 4.4.0 - string comment"
```
jmikola, line 50:
```suggestion
  - description: "countDocuments with string comment"
```

source/crud/tests/unified/countDocuments-comment.yml
```diff
@@ -0,0 +1,70 @@
...
+
+initialData:
+  - collectionName: *collection0Name
+    databaseName: *database0Name
+    documents:
+      - { _id: 1, x: 11 }
+      - { _id: 2, x: 22 }
+      - { _id: 3, x: 33 }
+
+tests:
```
jmikola, line 28: If we look at `insertMany-comment.yml`, I see we did have a test with `maxServerVersion: 4.2.99` entitled "insertMany with comment - pre 4.4" that expects a server-side error. Can you add a similar test here?

source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -265,3 +265,65 @@ tests:
...
+        arguments:
+          comment: &objectComment { key: "value"}
+        expectResult: 3
+    expectEvents:
+      - client: *client0
+        events:
+          - commandStartedEvent:
+              command:
+                aggregate: *collection0Name
+                pipeline: *pipeline
```
jmikola, line 284: I assume this is necessary because of the `$collStats` approach discussed in [the CRUD spec](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#estimateddocumentcount).

I do think these comment tests would be better placed in their own file (e.g. `estimatedDocumentCount-comment.yml`), even if that means redefining the `$collStats` pipeline. That would be more consistent with existing tests, which tend to split out files for testing certain options.

source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -265,3 +265,65 @@ tests:
...
+                comment: *objectComment
+              commandName: aggregate
+              databaseName: *database0Name
+
+  - description: "estimatedDocumentCount with comment on less than 4.9.0"
```
jmikola, line 289:
```suggestion
  - description: "estimatedDocumentCount with document comment - pre 4.9"
```

source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -265,3 +265,65 @@ tests:
...
+        expectResult: 3
+    expectEvents:
+      - client: *client0
+        events:
+          - commandStartedEvent:
+              command:
+                count: *collection0Name
+                comment: *stringComment
+              commandName: count
+              databaseName: *database0Name
```
jmikola, line 329: I know we have _some_ tests for server-side errors using the `comment` option (e.g. `insertOne`). Can we add one more test here?

> estimatedDocumentCount with string comment - pre 4.4

This can be modeled after the "insertOne with comment - pre 4.4" test.

source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -265,3 +265,65 @@ tests:
...
+              command:
+                count: *collection0Name
+                comment: *objectComment
+              commandName: count
+              databaseName: *database0Name
+
+  - description: "estimatedDocumentCount with comment on less than 4.4.14 - string comment"
+    runOnRequirements:
+      - minServerVersion: "4.4.0"
+        maxServerVersion: "4.4.13"
```
jmikola, line 314:
```suggestion
  - description: "estimatedDocumentCount with string comment - pre 4.9"
    runOnRequirements:
      - minServerVersion: "4.4.0"
        maxServerVersion: "4.8.99"
```

Assuming 4.4.0 is the first server version where `count` accepts a `comment` option at all, I don't think we need such a restrictive upper bound on this test. Typically, we'd have a test for both string and document types, with document filling in for _all_ non-string BSON types.

source/crud/tests/unified/countDocuments-comment.yml
```diff
@@ -0,0 +1,70 @@
...
+                aggregate: *collection0Name
+                pipeline: [{'$match': {}}, {'$group': {'_id': 1, 'n': {'$sum': 1}}}]
+                comment: *objectComment
+              commandName: aggregate
+              databaseName: *database0Name
+
+  - description: "count with comment on less than 4.4.0 - string comment"
+    runOnRequirements:
+      - minServerVersion: 3.6.0
+        maxServerVersion: 4.3.99
```
jmikola, line 53: I don't think we need a `maxServerVersion` if `aggregate` supports a string comment on all server versions since 3.6. Consider the existing "find with string comment" test.

source/crud/crud.rst
```diff
@@ -404,6 +404,19 @@ Read
...
+    /**
+     * Enables users to specify an arbitrary comment to help trace the operation through
+     * the database profiler, currentOp and logs. The default is to not send a value.
+     *
+     * The comment can be any valid BSON type for server versions 4.4.14 and above.
+     * For server versions between 4.4.0 and 4.4.14 string comment is supported.
+     # Servers versions below 4.4.0  do not support comment for count command,
+     * which is used to implement estimatedDocumentCount for server versions
+     * versions less than 4.9.0. Therefore, providing a comment may result
+     * in a server-side error.
```
jmikola, line 417: Per my comment below, let's add tests for the cases where we'd expect a server-side error:

 * Using a string comment on a pre-4.4 server version.
 * Using a document comment on a pre-4.4.14 sever version.

source/crud/tests/unified/countDocuments-comment.yml
comandeo: Actually, `aggregate` supports string comments down to 3.6 - https://www.mongodb.com/docs/manual/reference/command/aggregate/ Tests to comments in `aggregate.yml` have same `runOnRequirements`.
I am not sure if we should add new tests for pre 3.6 servers.

jmikola: I agree with not testing a pre-3.6 server. Having a test for a pre-4.4 server with a non-string comment seems reasonable, though.

source/crud/tests/unified/estimatedDocumentCount-comment.yml
```diff
@@ -0,0 +1,114 @@
...
+              command:
+                count: *collection0Name
+                comment: *stringComment
+              commandName: count
+              databaseName: *database0Name
+
+  - description: "estimatedDocumentCount with document comment on less than 4.4 - server error"
+    runOnRequirements:
+      - minServerVersion: "3.6.0"
+        maxServerVersion: "4.3.99"
```
jmikola, line 94: I think this should be 4.4.13, correct?

And if so, you can change the description to 4.4.13 and add a reference to [SERVER-63315](https://jira.mongodb.org/browse/SERVER-63315) in a comment, too.

source/crud/tests/unified/estimatedDocumentCount-comment.yml
```diff
@@ -0,0 +1,114 @@
...
+              command:
+                count: *collection0Name
+                comment: *documentComment
+              commandName: count
+              databaseName: *database0Name
+
+  - description: "estimatedDocumentCount with string comment on less than 4.4.14"
+    runOnRequirements:
+      - minServerVersion: "4.4.0"
+        maxServerVersion: "4.4.13"
```
jmikola, line 74: `estimatedDocumentCount` uses `count` until server version 4.9.0. Shouldn't this be `4.8.99`?

And then the description of this test can read "less than 4.9.0", like the "document comment" test before it.

source/crud/tests/unified/countDocuments-comment.yml
```diff
@@ -0,0 +1,92 @@
...
+              command:
+                aggregate: *collection0Name
+                pipeline: &pipeline
+                  - $match: {}
+                  - $group: { _id: 1, n: { $sum: 1 } }
+                comment: *documentComment
+              commandName: aggregate
+              databaseName: *database0Name
+
+  - description: "countDocuments with string comment on less than 4.4.0"
```
jmikola, line 50: This test only specifies a `minServerVersion`. I think the test title should just be:

```suggestion
  - description: "countDocuments with string comment"
```

source/crud/tests/unified/countDocuments-comment.yml
comandeo: Spec added.

source/crud/tests/unified/countDocuments-comment.yml
comandeo: No, they are not, thanks. Removed.

source/crud/tests/unified/countDocuments-comment.yml
comandeo: Done

source/crud/tests/unified/countDocuments-comment.yml
comandeo: Done

source/crud/tests/unified/countDocuments-comment.yml
comandeo: Done

source/crud/tests/unified/estimatedDocumentCount.yml
comandeo: Good idea, thank. Extracted new tests to a separated file.

source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -265,3 +265,65 @@ tests:
                 count: *collection0Name
               commandName: count
               databaseName: *database0Name
+
+  - description: "estimatedDocumentCount with comment on 4.9.0 or greater"
```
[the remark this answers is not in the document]
comandeo: Done

source/crud/tests/unified/estimatedDocumentCount.yml
comandeo: Done

source/crud/tests/unified/estimatedDocumentCount.yml
comandeo: Done

source/crud/tests/unified/estimatedDocumentCount.yml
comandeo: Done

source/crud/tests/unified/countDocuments-comment.yml
comandeo: Done

source/crud/tests/unified/estimatedDocumentCount-comment.yml
comandeo: Done

source/crud/tests/unified/estimatedDocumentCount-comment.yml
comandeo: Done

source/crud/tests/unified/countDocuments-comment.yml
comandeo: Done

source/crud/tests/unified/estimatedDocumentCount-comment.yml
```diff
@@ -0,0 +1,116 @@
...
+      - client: *client0
+        events:
+          - commandStartedEvent:
+              command:
+                count: *collection0Name
+                comment: *stringComment
+              commandName: count
+              databaseName: *database0Name
+
+  - description: "estimatedDocumentCount with document comment - pre 4.4.13, server error"
```
jmikola, line 91: The server version in the description is _exclusive_, if we want to be consistent with earlier tests (e.g. "pre 4.9" using "4.8.99" in the requirement).

```suggestion
  - description: "estimatedDocumentCount with document comment - pre 4.4.14, server error"
```

source/crud/tests/unified/estimatedDocumentCount-comment.yml
```diff
@@ -0,0 +1,116 @@
...
+    runOnRequirements:
+      - minServerVersion: "3.6.0"
+        maxServerVersion: "4.4.13"
+    operations:
+      - name: estimatedDocumentCount
+        object: *collection0
+        arguments:
+          # Even though according to the docs count command does not support any
+          # comment for server version less than 4.4, no error is raised by such
+          # servers. Therefore, we have only one tests with document comment
```
jmikola, line 101:
```suggestion
          # servers. Therefore, we have only one test with a document comment
```

source/crud/tests/unified/estimatedDocumentCount.yml
jmikola: I noted that an error test for string comments was not implemented, per the comment you left in the document comment test expecting a server-side error.

source/crud/tests/unified/estimatedDocumentCount-comment.yml
```diff
@@ -0,0 +1,116 @@
...
+
+  - description: "estimatedDocumentCount with document comment - pre 4.4.13, server error"
+    runOnRequirements:
+      - minServerVersion: "3.6.0"
+        maxServerVersion: "4.4.13"
+    operations:
+      - name: estimatedDocumentCount
+        object: *collection0
+        arguments:
+          # Even though according to the docs count command does not support any
```
jmikola, line 99: Acknowledging that this explains why you did not add an error-expecting test for "using a string comment on a pre-4.4 server version", as I originally requested in https://github.com/mongodb/specifications/pull/1179#discussion_r849558587.

source/crud/tests/unified/estimatedDocumentCount-comment.yml
```diff
@@ -0,0 +1,116 @@
...
+      - client: *client0
+        events:
+          - commandStartedEvent:
+              command:
+                count: *collection0Name
+                comment: *documentComment
+              commandName: count
+              databaseName: *database0Name
+
+  - description: "estimatedDocumentCount with string comment - pre 4.9"
```
jmikola, line 71: Just to confirm: did you intentionally omit a "string comment - post 4.9" test? I suppose it's redundant given "estimatedDocumentCount with document comment - post 4.9", since there's no reason `aggregate` wouldn't support both types (noting `document` is just a stand-in for _any_ BSON type).

jmikola (commented):
Just some typo fixes. I'd like to ask one other driver to POC this before we merge, though.

@benjirewis: Is Go able to test this?

source/crud/tests/unified/estimatedDocumentCount-comment.yml
comandeo: That is correct. I think it is safe to assume that if a test with a document comment pass, it means that this server accepts any BSON type as a comment.
We can add a test for every BSON type as a comment, but I think it would be too much.

benjirewis (approved):
"POC"d in the Go driver. We don't really have support for the `comment` field yet (most changes will be done as part of GODRIVER-2318). I added `comment` to `CountOptions` and `EstimatedDocumentCountOptions` to test, though, and was able to get these tests to pass. So, LGTM.

Good job on the intricate server version logic, too. The existing difference in `estimatedDocumentCount`'s behavior based on server version (its manifestation as either `count` or `aggregate`) makes this a little weird to test.

jmikola (approved):

PR #1180, opened 2022-04-13
test: add retryable write label to handshake tests

durran:
Adds the "RetryableWriteError" label to handshake error unified tests that expect the write to be retryable.

### DIFF
--- a/source/retryable-writes/tests/unified/handshakeError.json
+++ b/source/retryable-writes/tests/unified/handshakeError.json
@@ -181,7 +181,10 @@
                   "saslContinue",
                   "ping"
                 ],
-                "errorCode": 91
+                "errorCode": 91,
+                "errorLabels": [
+                  "RetryableWriteError"
+                ]
               }
             }
           }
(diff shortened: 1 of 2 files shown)


jyemin:
Before we merge this, I'd like to have an understanding of why other drivers are passing this test as is (Node driver is failing it without this change).

ShaneHarvey (commented):
On second thought, this PR change doesn't make sense. The server will never add RetryableWriteError label to errors encountered when setting up a connection because it only adds that label to errors returned when processing _retryable write commands_. The driver must retry these handshake errors based on the code, not the label.

I almost think that drivers should always apply the error code checking logic on all server versions (based on this and other issues like SERVER-62175).

nbbeeken:
Drivers should be retrying this error based off the code because the scenario is that this is the initial handshake and so there is no reported wire version yet. Since we don't have a wire version greater than 9 we should perform the retryablity logic from pre-4.4 (using the codes). IIUC

PR #1182, opened 2022-04-18
DRIVERS-1853 Add tests and update spec to support API changes for clustered collections

abr-egn:
DRIVERS-1853

This adds some new spec tests to validate the API changes in support of clustered indexes as described in the [Scope Document](https://tinyurl.com/yk697poh).  It also updates the index management spec with the new field returned for indexes created in this way.

To my surprise, I couldn't find a spec that covers the behavior and options accepted by `createCollection`; it looks like that's covered by tests unassociated with any spec, so I've continued in that track.

https://github.com/abr-egn/mongo-rust-driver/tree/DRIVERS-1853/exploration is a branch of the Rust driver that passes these tests.


kmahar:
question regarding the Rust POC / what driver APIs in general need to look like:
do we need to support the "legacy" boolean or can we simplify and assume it's always a document? for the purposes of `createCollection` I think the latter would be fine. but perhaps it's possible that certain server versions can return a boolean for that value in `listIndexes`?

source/index-management.rst
```diff
@@ -821,6 +821,11 @@ Common API Components
      * This option is only supported by servers >= 4.4.
      */
     hidden: Boolean;
+
+    /**
+     * Optionally specifies that this index is clustered.
+     */
+     clustered: Boolean;
```
kmahar, line 828: we might want to call out the server version where this is introduced similar to `hidden` above this.

also, is it possible to create a clustered index on an existing collection? the [docs](https://www.mongodb.com/docs/v6.0/core/clustered-collections/#limitations) seem to imply no:
> You cannot transform a non-clustered collection to a clustered collection, or the reverse.

if not, it's a little funny to have this here since drivers typically use this type for their `createIndexes` helpers... but on the flipside, we probably want it here for anyone who uses this type in their `listIndexes` output. 

we at least probably should call this out in the docstring because I think otherwise drivers may allow users to specify it here. in Swift anyway it would be easy enough to make it so users can only read but not set/mutate this property, but we can set it internally when decoding a `listIndexes` response. could we support something similar in Rust?

abr-egn:
> question regarding the Rust POC / what driver APIs in general need to look like: do we need to support the "legacy" boolean or can we simplify and assume it's always a document? for the purposes of `createCollection` I think the latter would be fine. but perhaps it's possible that certain server versions can return a boolean for that value in `listIndexes`?

The design doc [specifies](https://docs.google.com/document/d/1hRnqe1pq7ErcguJ5Bl8si1kHGKlZg0h25qsa3roASuo/edit#bookmark=id.mq9r4uqqp1go) that `createCollection` needs to accept both formats.

source/collection-management/tests/clustered-indexes.yml
```diff
@@ -0,0 +1,102 @@
...
+        object: *collection0
+        expectResult: [
+          {
+            key: { _id: 1 },
+            name: *index0name,
+            clustered: true,
+            unique: true,
+            v: 2,
+          }
+        ]
```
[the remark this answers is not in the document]
abr-egn: Done.

source/index-management.rst
abr-egn: Added version information.

The design doc actually [explicitly says](https://docs.google.com/document/d/1hRnqe1pq7ErcguJ5Bl8si1kHGKlZg0h25qsa3roASuo/edit#bookmark=id.oh101ver26rd) that `createIndexes` should accept `clustered`; I don't know how to reconcile that with the doc you linked.

source/collection-management/tests/clustered-indexes.yml
```diff
@@ -0,0 +1,102 @@
...
+            options: {
+              clusteredIndex: {
+                key: { _id: 1 },
+                unique: true,
+                name: *index0name,
+                v: 2,
+              }
+            }
+          }
+        ]
```
jmikola, line 76: Avoiding JSON-like syntax.

```suggestion
        expectResult:
          - name: *collection0Name
            options:
              clusteredIndex:
                key: { _id: 1 }
                unique: true
                name: *index0Name
                v: 2
```

source/collection-management/tests/clustered-indexes.yml
```diff
@@ -0,0 +1,102 @@
...
+        arguments:
+          collection: *collection0Name
+      - name: createCollection
+        object: *database0
+        arguments:
+          collection: *collection0Name
+          clusteredIndex:
+            key: { _id: 1 }
+            unique: true
+            name: &index0name "test index"
```
jmikola, line 59: For consistency with camel case used in other references.

```suggestion
            name: &index0Name "test index"
```

source/collection-management/tests/clustered-indexes.yml
```diff
@@ -0,0 +1,102 @@
...
```
jmikola, line 91: No need to redefine with the same value.

```suggestion
            name: *index0Name
```

source/collection-management/tests/clustered-indexes.yml
jmikola, line 102:
```suggestion
        expectResult:
        - key: { _id: 1 }
          name: *index0name
          clustered: true
          unique: true
          v: 2
```

source/collection-management/tests/clustered-indexes.yml
```diff
@@ -0,0 +1,102 @@
...
+          filter: { name: { $eq: *collection0Name } }
+        expectResult: [
+          {
+            name: *collection0Name,
+            options: {
+              clusteredIndex: {
+                key: { _id: 1 },
+                unique: true,
+                name: *index0name,
+                v: 2,
```
jmikola, line 72: Independent of my suggested change below, should we bother asserting that `v: 2` is set in the result? This seems like an assumption on the server-side, which could possibly change in the future. I assume `unique: true` is closely related to the design of this feature, but I think we can relax the `v` assertion (or perhaps just expect `{ $$type: [ int, long ] }`).

source/index-management.rst
kmahar: I got some clarification from this slack thread: https://mongodb.slack.com/archives/C2RCHGB2L/p1650576866908169 

given this, I suggest we note in the docstring here that this is not a valid option to provide to `createIndexes`, but can appear in the options returned for an index via `listIndexes`, and that if users want to create a clustered index, they need to create a new collection using the `clusteredIndex` option.

for the Rust impl it seems like we'll have to include it in `IndexOptions`, since that type models the response of to `createIndexes`, but can we do something to prevent users from actually setting the property themselves / omit the builder method for this property somehow?

source/collection-management/tests/clustered-indexes.yml
abr-egn: Done.

As a side note, it would be really helpful if there was a style guide for test yaml - there's a remarkably wide variety of styles in use across the existing test files for things like formatting of expectation objects, quoting vs. not, etc. and it's very hard to tell which is the recommended way to do things.

source/collection-management/tests/clustered-indexes.yml
abr-egn: Done.

source/collection-management/tests/clustered-indexes.yml
abr-egn: Done.

source/collection-management/tests/clustered-indexes.yml
abr-egn: Done.

source/collection-management/tests/clustered-indexes.yml
abr-egn: Good call, the actual value here isn't relevant - relaxed to just assert on type.

kmahar:
> The design doc [specifies](https://docs.google.com/document/d/1hRnqe1pq7ErcguJ5Bl8si1kHGKlZg0h25qsa3roASuo/edit#bookmark=id.mq9r4uqqp1go) that createCollection needs to accept both formats.

Well, the server command needs to for backward compatibility reasons since someone could for example have a script that relies on that syntax being supported -- but since we never accepted a boolean in our API, I'm not sure we need to start doing so as the non-legacy format provides equivalent functionality. I would guess we can skip it, but this might be worth following up with Haley about on that thread?

source/collection-management/tests/clustered-indexes.yml
jmikola: https://jira.mongodb.org/browse/DRIVERS-2289

source/index-management.rst
jmikola: I didn't notice this earlier, but if this option is only relevant to `listIndexes` (and not `createIndexes` or any of the other helpers in this specification), should this be mentioned in the [Enumerating Indexes](https://github.com/mongodb/specifications/blob/master/source/enumerate-indexes.rst) specification? I suppose not, since that spec doesn't talk at all about modeling the response.

kmahar: it seems weird to me that we have two separate specs around indexes. why is enumerating indexes not just part of index management?
it feels like this could go either place, but since it seems like this option will mostly only matter for any drivers that use `IndexOptions` to model the result of `listIndexes` I lean toward keeping it here with the definition of the type

PR #1183, opened 2022-04-19
DRIVERS-2284 add FLE 2 behavior for `CreateCollection()` and `Collection.Drop()`.

kevinAlbs:
# Summary
- Add terms for `FLE 2`, `FLE 1`, and `EncryptedFieldConfig`.
- Add `EncryptedFieldsMap` to `AutoEncryptionOpts`.
- Add `CreateCollection` and `Drop` behavior for FLE 2.
- Add `EncryptedFields` options to `CreateCollection` and `Drop`
- Add `encryptedFieldsMap` to the Client Side Encryption specification test format.
- Add fle2-CreateCollection.yml test.

# Background & Motivation

Please see the [Go driver implementation of this behavior](https://github.com/mongodb/mongo-go-driver/pull/913) as a reference implementation.

Please see the [Technical Design: Create Collection and Compact for FLE2 Encrypted Multi Maps (EMM)](https://docs.google.com/document/d/1_XCsKrZ1V1l0rf85umaZZzm7jFccro0Oo1AWSr3H8R4/) for the description of driver requirements:

> Drivers are responsible for the following calls to “create” an encrypted collection.
> - Create collection <collection_name> with encryptedFields
> - Create Index of {__safeContent__: 1}  on <collection_name>.
> - Create collection named “enxcol_.<collection_name>.esc” for the ESC
> - Create collection named “enxcol_.<collection_name>.ecc” for the ECC
> - Create collection named “enxcol_.<collection_name>.ecoc” for the ECOC 

I chose not to amend the Unified Test Format to add the new tests. The only required change to the existing test format was the `encryptedFieldConfigMap`.

# Q & A
The following are Q&A from the scope that may be helpful for reviewers. Rationale I think is helpful for implementors is included in the source.

## Q1: Why are the new collections and index created in `CreateCollection()` not server-side operations?
Answer: See the [Technical Design: Create Collection and Compact for FLE2 Encrypted Multi Maps (EMM)](https://docs.google.com/document/d/1_XCsKrZ1V1l0rf85umaZZzm7jFccro0Oo1AWSr3H8R4/edit#heading=h.wxgl2ze3j23q) Design Rationale "Create Collection creates all collections". 

> If create collection in FLE 2 creates 4 collections and 1 secondary index in one create collection invocation this breaks the assumptions of tools like MongoRestore/MongoSync/MongoMirror in that one call to createCollection creates one collection. Based on this design, when these tools “restore” multiple collections that involve a user collection and one or more state collection, they would need to create the user data collection first, and skip creation of the state collections. After collection creation is complete, the tools can proceed as normal. This adds a lot of complexity to the tools and has been rejected

## Q2: Why are state collections created before user collection?
Answer: See [Technical Design: Create Collection and Compact for FLE2 Encrypted Multi Maps (EMM)](https://docs.google.com/document/d/1_XCsKrZ1V1l0rf85umaZZzm7jFccro0Oo1AWSr3H8R4/edit#)

> The state collections should be created before the user collection so that state collections are not implicitly created.

## Q3: Where is the "automatic key creation"?
Answer: Originally [scoped](https://docs.google.com/document/d/1duvPDqbFFEvSjTWX_8Vp-u_IjGZGdV3ZyrzCjrjbn2E/edit#heading=h.dbzfz3bbd2e7) was a requirement to create keys for a user automatically that are not present in the ``EncryptedFieldConfig``:

> It is a product requirement to create keys for the user when an FLE 2 collection is created. The ability for us to create keys for a customer enables the ideal encryption state, where there is a unique key per field/document/sub document being encrypted, while making it easier for the developer by abstracting out key creation

This requirement is removed for initial support of FLE 2 to simplify the implementation.


comandeo (approved):
LGTM!

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -591,6 +623,59 @@ the user. Refer:
...
+  If ``EncryptedFieldConfig["escCollection"]`` is not set, use the collection name ``enxcol_.<collectionName>.esc``.
+  Creating this collection MUST NOT check if the collection namespace is in the ``AutoEncryptionOpts.EncryptedFieldConfigMap``.
+- Create the collection with name ``EncryptedFieldConfig["eccCollection"]`` using default options.
+  If ``EncryptedFieldConfig["eccCollection"]`` is not set, use the collection name ``enxcol_.<collectionName>.ecc``.
+  Creating this collection MUST NOT check if the collection namespace is in the ``AutoEncryptionOpts.EncryptedFieldConfigMap``.
+- Create the collection with name ``EncryptedFieldConfig["ecocCollection"]`` using default options.
+  If ``EncryptedFieldConfig["ecocCollection"]`` is not set, use the collection name ``enxcol_.<collectionName>.ecoc``.
+  Creating this collection MUST NOT check if the collection namespace is in the ``AutoEncryptionOpts.EncryptedFieldConfigMap``.
+- Create the collection ``collectionName`` with ``collectionOptions`` and the option ``encryptedFields`` set to the ``EncryptedFieldConfig``.
+- Create the the index ``{"__safeContent__": 1}`` on collection ``collectionName``.
```
addaleax, line 665: Should there be a behavior specification for the case that one of these steps fails? E.g. be explicit about the fact that if one step fails, the former steps won’t be undone?

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -591,6 +623,59 @@ the user. Refer:
...
+EncryptedFieldConfigMap
+^^^^^^^^^^^^^^^^^^^^^^^
+
+``EncryptedFieldConfigMap`` maps a collection namespace to an ``EncryptedFieldConfig``.
+
+``EncryptedFieldConfigMap`` only applies to FLE 2.
+
+If a collection is present on both the ``EncryptedFieldConfigMap`` and ``schemaMap``, libmongocrypt_ will error on initialization. See :ref:`fle2-and-fle1-error`.
+
+If a collection is present on the ``EncryptedFieldConfigMap``, the behavior of ``CreateCollection()`` and ``Collection.Drop()`` is altered. See :ref:`fle2-createcollection-drop`.
```
addaleax, line 635: To clarify my understanding, this will affect the regular driver `createCollection()`/`drop()` methods, and users won’t need to be calling a different method for creating FLE 2 collections? How does this map to the `ClientEncryption.createEncryptedCollection()` method lined out in the scope (i.e. is the expectation that the driver `createCollection()` method will call that)?

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -591,6 +623,59 @@ the user. Refer:
...
+- Create the collection with name ``EncryptedFieldConfig["eccCollection"]`` using default options.
+  If ``EncryptedFieldConfig["eccCollection"]`` is not set, use the collection name ``enxcol_.<collectionName>.ecc``.
+  Creating this collection MUST NOT check if the collection namespace is in the ``AutoEncryptionOpts.EncryptedFieldConfigMap``.
+- Create the collection with name ``EncryptedFieldConfig["ecocCollection"]`` using default options.
+  If ``EncryptedFieldConfig["ecocCollection"]`` is not set, use the collection name ``enxcol_.<collectionName>.ecoc``.
+  Creating this collection MUST NOT check if the collection namespace is in the ``AutoEncryptionOpts.EncryptedFieldConfigMap``.
+- Create the collection ``collectionName`` with ``collectionOptions`` and the option ``encryptedFields`` set to the ``EncryptedFieldConfig``.
+- Create the the index ``{"__safeContent__": 1}`` on collection ``collectionName``.
+
+Drivers MUST NOT add a public option to directly set the ``encryptedFields`` option in the ``CreateCollection`` helper.
```
addaleax, line 667: Right now, at least the Node.js driver just passes through collection creation options, i.e. I can write `await db.createCollection('foo', { encryptedFields: { fields: [{ path: 'field', bsonType: 'string', keyId: UUID('00000000-0000-0000-0000-000000000000') }] } })` and that will work. Should the driver error out in this case?

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -294,6 +325,7 @@ MongoClient Changes
       bypassAutoEncryption: Optional<Boolean>; // Default false.
       extraOptions: Optional<Map<String, Value>>;
       tlsOptions: Optional<Map<String, TLSOptions>>; // Maps KMS provider to TLS options.
+      encryptedFieldConfigMap: Map<String, Map<String, Document>>; // Maps namespace to EncryptedFieldConfig.
```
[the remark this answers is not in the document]
kevinAlbs: I think it is OK to leave as-is. The type is expected to be the same as `schemaMap`.

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: > To clarify my understanding, this will affect the regular driver createCollection()/drop() methods, and users won’t need to be calling a different method for creating FLE 2 collections?
Yes. The proposal is to use `createCollection()` / `drop()` methods.

> How does this map to the ClientEncryption.createEncryptedCollection() method lined out in the scope (i.e. is the expectation that the driver createCollection() method will call that)?

The proposal is not to add `ClientEncryption.createEncryptedCollection()` / `ClientEncryption.dropEncryptedCollection` as is scoped. There is no longer a requirement to automatically create data keys, so I do not think the new API is necessary.

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: Good point. Added.

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: Good point. The rationale for no a public `encryptedFields` option is to prevent users from creating a collection with `encryptedFields` without the index or state collections.

IMO we should not make it a driver error if `encryptedFields` is set in `RunCommand`. If some drivers pass along options through `createCollection`, this seems like a similar case.

I have removed this sentence. I want drivers to not add an `encryptedFields` option on their `CreateCollectionOptions` type if they have it. But drivers that are passing options along (Node, and also C) should not reject it.

source/client-side-encryption/tests/fle2-CreateCollection.yml
```diff
@@ -0,0 +1,522 @@
...
+        # Index on __safeContents__ is then created.
+        - command_started_event:
+            command:
+              createIndexes: "encryptedCollection"
+              indexes:
+                - name: __safeContent___1
+                  key: { __safeContent__: 1 }
+            commandName: createIndexes
+            databaseName: *database_name
+        # events from createCollection ... end
```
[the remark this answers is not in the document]
kevinAlbs: Done.

source/client-side-encryption/client-side-encryption.rst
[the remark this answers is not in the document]
jyemin: Could the type be the same as schemaMap: `Optional<Map<String, Document>>`?  It seems similarly optional, and the generic type representing EncryptedFieldConfig could just be `Document` rather than `Map<String, Document>`.  I'm not sure what purpose the extra level of `Map` is serving.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -591,6 +623,59 @@ the user. Refer:
...
+  If ``EncryptedFieldConfig["eccCollection"]`` is not set, use the collection name ``enxcol_.<collectionName>.ecc``.
+  Creating this collection MUST NOT check if the collection namespace is in the ``AutoEncryptionOpts.EncryptedFieldConfigMap``.
+- Create the collection with name ``EncryptedFieldConfig["ecocCollection"]`` using default options.
+  If ``EncryptedFieldConfig["ecocCollection"]`` is not set, use the collection name ``enxcol_.<collectionName>.ecoc``.
+  Creating this collection MUST NOT check if the collection namespace is in the ``AutoEncryptionOpts.EncryptedFieldConfigMap``.
+- Create the collection ``collectionName`` with ``collectionOptions`` and the option ``encryptedFields`` set to the ``EncryptedFieldConfig``.
+- Create the the index ``{"__safeContent__": 1}`` on collection ``collectionName``.
+
+
+A call to a driver helper ``Collection.Drop()`` must check if the collection namespace (``<databaseName>.<collectionName>``) is in ``AutoEncryptionOpts.EncryptedFieldConfigMap``.
```
jyemin, line 669: Did we consider whether `drop` should use an ``EncryptedFieldConfig`` obtained from the server, as seems to be done for auto-encryption, according to the note:

> Supplying an ``EncryptedFieldConfigMap`` provides more security than relying on an ``EncryptedFieldConfig`` obtained from the server.

?

source/client-side-encryption/client-side-encryption.rst
addaleax: > The proposal is not to add `ClientEncryption.createEncryptedCollection()` / `ClientEncryption.dropEncryptedCollection()` as is scoped.

Thank you for clariying. One final question then: Am I understanding correctly that this means that users have to know which FLE2 collections they will be creating at `MongoClient` instantiation time? That seems fine for applications, but might not be quite what users expect when they use a tool like mongosh or Compass.

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: Good question. I had not considered it.

Here is a proposal to use `listCollections` in Go:

```go
func (coll *Collection) Drop(ctx context.Context) error {
	efcMap := coll.db.client.encryptedFieldConfigMap
	if efcMap != nil {
		// Use listCollections to check the server for an EncryptedFieldConfig.
		efc, err := coll.getEncryptedFieldConfigFromServer(ctx)
		if err != nil {
			return err
		}
		if efc != nil {
			// Drop collection and state collections from EncryptedFieldConfig.
			return coll.dropEncryptedCollection(ctx, efc)
		}
	}
	// Do a normal Drop.
	return coll.dropHelper(ctx)
}
```

The behavior is gated behind an `EncryptedFieldConfigMap` being set. This assumes it is not desirable to run the `listCollections` in the Drop helper unless there is reason to believe the user has FLE 2 collections.

I am in favor of adding the call to `listCollections`. It does not add much complexity. Does this seem OK?

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: > Am I understanding correctly that this means that users have to know which FLE2 collections they will be creating at MongoClient instantiation time?

As proposed, yes.

> That seems fine for applications, but might not be quite what users expect when they use a tool like mongosh or Compass.

Good point. Another option is to add an `EncryptedFieldConfig` option to the `createCollection()` helper directly. I briefly considered it, but thought a new option would not be necessary. Does adding an `EncryptedFieldConfig` option to `CreateCollection` seem OK for mongosh and Compass use cases?

addaleax: Yeah, I think that would make sense. For drivers that already pass through options as they are, those could inspect that field to understand whether they would need to perform the additional steps lined out here, correct?

kevinAlbs: > For drivers that already pass through options as they are, those could inspect that field to understand whether they would need to perform the additional steps lined out here, correct?

Yes, that seems OK. If the `encryptedFieldConfig` option is present, remove it, and do the steps to create the state collections and index.

I will proceed with proposing a new `EncryptedFieldConfig` option in `createCollection()`.

jyemin: For drop as well?

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -591,6 +623,59 @@ the user. Refer:
...
+- Create the collection with name ``EncryptedFieldConfig["escCollection"]`` using default options.
+  If ``EncryptedFieldConfig["escCollection"]`` is not set, use the collection name ``enxcol_.<collectionName>.esc``.
+  Creating this collection MUST NOT check if the collection namespace is in the ``AutoEncryptionOpts.EncryptedFieldConfigMap``.
+- Create the collection with name ``EncryptedFieldConfig["eccCollection"]`` using default options.
+  If ``EncryptedFieldConfig["eccCollection"]`` is not set, use the collection name ``enxcol_.<collectionName>.ecc``.
+  Creating this collection MUST NOT check if the collection namespace is in the ``AutoEncryptionOpts.EncryptedFieldConfigMap``.
+- Create the collection with name ``EncryptedFieldConfig["ecocCollection"]`` using default options.
+  If ``EncryptedFieldConfig["ecocCollection"]`` is not set, use the collection name ``enxcol_.<collectionName>.ecoc``.
+  Creating this collection MUST NOT check if the collection namespace is in the ``AutoEncryptionOpts.EncryptedFieldConfigMap``.
+- Create the collection ``collectionName`` with ``collectionOptions`` and the option ``encryptedFields`` set to the ``EncryptedFieldConfig``.
```
jyemin, line 665: If we get to this point, and then create collection fails due to the collection already existing, is the server in a weird state?  Or should we proceed with creating the index?

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: > For drop as well?

Will do.

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: > If we get to this point, and then create collection fails due to the collection already existing, is the server in a weird state?

In that situation the state collections would be created. Users could call Drop to drop any of the state collections and retry the CreateCollection.

> Or should we proceed with creating the index?

The proposed behavior is to exit early.

> If any of the following operations error, the remaining operations are not attempted:

source/client-side-encryption/client-side-encryption.rst
addaleax: That sounds okay – I think you could also just reuse `encryptedFields`, i.e. the options that are actually part of the `create` command, since they will contain the same information as the EncryptedFieldConfigMap for the collection?

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: Added a check in `Drop`. If `EncryptedFieldConfigMap` is set and the collection has no entry in the map, `listCollections` is run to check for an `EncryptedFieldConfig` on the server.

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: > That sounds okay – I think you could also just reuse encryptedFields, i.e. the options that are actually part of the create command, since they will contain the same information as the EncryptedFieldConfigMap for the collection?

Good point. I have renamed `EncryptedFieldConfig` to `EncryptedFields`.

kevinAlbs (commented):
Summary of new changes:
- Add `EncryptedFieldConfig` option to `CreateCollection` and `Drop`
- Add check in `Drop` to run `listCollections` to check for remote `EncryptedFieldConfig`
- Rename `EncryptedFieldConfig` to `EncryptedFields`.
- Add new tests.
- Update the Go implementation.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -325,7 +325,7 @@ MongoClient Changes
       bypassAutoEncryption: Optional<Boolean>; // Default false.
       extraOptions: Optional<Map<String, Value>>;
       tlsOptions: Optional<Map<String, TLSOptions>>; // Maps KMS provider to TLS options.
-      encryptedFieldConfigMap: Optional<Map<String, Document>>; // Maps namespace to EncryptedFieldConfig.
+      EncryptedFieldsMap: Optional<Map<String, Document>>; // Maps namespace to EncryptedFields.
```
jyemin, line 328: Nit: lowercase the first letter for consistency

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -649,33 +649,44 @@ FLE 2 ``CreateCollection()`` and ``Collection.Drop()``
...
+
+A call to a driver helper ``Collection.Drop(dropOptions)`` must check if the collection namespace (``<databaseName>.<collectionName>``) has an associated ``EncryptedFields``. Check for an associated ``EncryptedFields`` from the following:
+
+- The ``EncryptedFields`` option passed in ``dropOptions``.
+- The value of ``AutoEncryptionOpts.EncryptedFieldsMap[<databaseName>.<collectionName>]``.
+- Run a ``listCollections`` command on the database ``databaseName`` with the filter ``{ "name": "<collectionName>" }``. Check the returned ``options`` for the ``encryptedFields`` option. The value of ``encryptedFields`` is the ``EncryptedFields``.
 
-A call to a driver helper ``Collection.Drop()`` must check if the collection namespace (``<databaseName>.<collectionName>``) is in ``AutoEncryptionOpts.EncryptedFieldConfigMap``.
-If it is, then do the following operations. If any of the following operations error, the remaining operations are not attempted:
+If the collection namespace has an associated ``EncryptedFields``, then do the following operations. If any of the following operations error, the remaining operations are not attempted:
```
jyemin, line 681: Only do this if AutoEncryptionOpts.EncryptedFieldsMap is not null?

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -649,33 +649,44 @@ FLE 2 ``CreateCollection()`` and ``Collection.Drop()``
...
+- Create the collection with name ``EncryptedFields["escCollection"]`` using default options.
+  If ``EncryptedFields["escCollection"]`` is not set, use the collection name ``enxcol_.<collectionName>.esc``.
+  Creating this collection MUST NOT check if the collection namespace is in the ``AutoEncryptionOpts.EncryptedFieldsMap``.
+- Create the collection with name ``EncryptedFields["eccCollection"]`` using default options.
+  If ``EncryptedFields["eccCollection"]`` is not set, use the collection name ``enxcol_.<collectionName>.ecc``.
+  Creating this collection MUST NOT check if the collection namespace is in the ``AutoEncryptionOpts.EncryptedFieldsMap``.
+- Create the collection with name ``EncryptedFields["ecocCollection"]`` using default options.
+  If ``EncryptedFields["ecocCollection"]`` is not set, use the collection name ``enxcol_.<collectionName>.ecoc``.
+  Creating this collection MUST NOT check if the collection namespace is in the ``AutoEncryptionOpts.EncryptedFieldsMap``.
+- Create the collection ``collectionName`` with ``collectionOptions`` and the option ``encryptedFields`` set to the ``EncryptedFields``.
```
jyemin, line 670: Do drivers need to do anything special if the `collectionName` collection already exists?

source/client-side-encryption/client-side-encryption.rst

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: I think not. The server will error:

```
> db.runCommand({create: "foo"})
{
        "ok" : 0,
        "errmsg" : "Collection test.foo already exists.",
        "code" : 48,
        "codeName" : "NamespaceExists",
        "$clusterTime" : {
                "clusterTime" : Timestamp(1650500294, 1),
                "signature" : {
                        "keyId" : NumberLong(0)
                }
        },
        "operationTime" : Timestamp(1650500294, 1)
}
```

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: Oof, good catch. ``listCollections`` is intended to only be run if ``AutoEncryptionOpts.encryptedFieldsMap`` is not null. Updated the step for ``listCollections``.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -591,6 +623,70 @@ the user. Refer:
...
+- Create the collection ``collectionName`` with ``collectionOptions`` and the option ``encryptedFields`` set to the ``encryptedFields``.
+- Create the the index ``{"__safeContent__": 1}`` on collection ``collectionName``.
+
+Drivers MUST have a BSON document option named ``encryptedFields`` in ``Collection.Drop()``.
+
+A call to a driver helper ``Collection.Drop(dropOptions)`` must check if the collection namespace (``<databaseName>.<collectionName>``) has an associated ``encryptedFields``. Check for an associated ``encryptedFields`` from the following:
+
+- The ``encryptedFields`` option passed in ``dropOptions``.
+- The value of ``AutoEncryptionOpts.encryptedFieldsMap[<databaseName>.<collectionName>]``.
+- If ``AutoEncryptionOpts.encryptedFieldsMap`` is not null, run a ``listCollections`` command on the database ``databaseName`` with the filter ``{ "name": "<collectionName>" }``. Check the returned ``options`` for the ``encryptedFields`` option.
```
addaleax, line 679: My gut feeling here would be:
- Dropping collections is a fairly infrequent operation
- `listCollections` with a filter for a single collection is comparatively cheap
- The downsides of leaving unused collections lying around are larger than the upsides of fast collection dropping

I know it adds an extra round-trip, but I’d be okay with dropping the base condition here. (If not, that’s also fine – we’ll implement this behavior in devtools either way.)

addaleax (approved):
This looks great, thank you! With the current behavior this definitely seems better than having separate methods for FLE collections hanging off `ClientEncryption` :+1:

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -294,6 +325,7 @@ MongoClient Changes
       bypassAutoEncryption: Optional<Boolean>; // Default false.
       extraOptions: Optional<Map<String, Value>>;
       tlsOptions: Optional<Map<String, TLSOptions>>; // Maps KMS provider to TLS options.
+      encryptedFieldsMap: Optional<Map<String, Document>>; // Maps namespace to encryptedFields.
```
addaleax, line 328: Sorry for the extra comment – in the scope this is called `encryptedFieldConfigMap`. I do like thie variant a bit more, but either way, once this spec PR is merged, we can rely on this name not being changed, right?

addaleax: Perfect, thank you!

source/client-side-encryption/client-side-encryption.rst
addaleax: SGTM – Feel free to leave open if you’d like more input here, otherwise you can just resolve this comment

jyemin (approved):
LGTM!

PR #1184, opened 2022-04-19
DRIVERS-1713 Fix change stream errors test on 3.6

abr-egn:
DRIVERS-1713

Discovered when updating the Rust tests that the first test in `change-streams-errors.yml` fails with a different error code on server version 3.6 if the database doesn't exist.  Not sure why this didn't show up in previous test runs.

### DIFF
--- a/source/change-streams/tests/unified/change-streams-errors.json
+++ b/source/change-streams/tests/unified/change-streams-errors.json
@@ -49,6 +49,13 @@
       }
     }
   ],
+  "initialData": [
+    {
+      "collectionName": "collection0",
+      "databaseName": "database0",
+      "documents": []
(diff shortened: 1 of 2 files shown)


benjirewis (approved):
Seems good to me. 

Discussed offline the reason why that first test might have been failing in the Rust driver without `initialData`. My only guess was that on server version 3.6-, change stream creation first checked if the target existed (which it did not without `initialData`) and then checked if the topology even supported change streams. So, the server would return a `NamespaceNotFound` error (code 26) for the non-existent collection instead of the expected `$changeStream is only supported against replicaSets` error. The order of those conditional checks might have changed on server versions greater than 3.6.

PR #1185, opened 2022-04-20
DRIVERS-2287 Clarify interaction of bypassAutoEncryption and csfle search paths

addaleax:
Picking the behavior of _not_ setting csfle search paths when `bypassAutoEncryption` was passed, because:
- Creating a mongocrypt_t instance with an unnecessary csfle library may conflict with the creation of a mongocrypt_t instance that is supposed to use the csfle library but from another path
- It seems like `bypassAutoEncryption` could be implemented by using a `ClientEncryption` instance rather than a separate libmongocrypt handle (on what at least the Python and Node.js bindings call `AutoEncrypter`), in which case the csfle library should probably never be loaded

### DIFF
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -915,8 +915,11 @@ Setting Search Paths
 --------------------
 
 For the user-facing API the driver MUST append the literal string
-:ts:`"$SYSTEM"` to the search paths for the `libmongocrypt_handle`. For purposes
-of testing, a driver may use a different set of search paths.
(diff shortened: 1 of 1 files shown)


kevinAlbs (approved):
Thank you for the change, LGTM!

> It seems like bypassAutoEncryption could be implemented by using a ClientEncryption instance rather than a separate libmongocrypt handle (on what at least the Python and Node.js bindings call AutoEncrypter), in which case the csfle library should probably never be loaded

That is a great point I had not considered. `csfle`, like `mongocryptd`, is not needed in the ClientEncryption instance.

comandeo (approved):

PR #1186, opened 2022-04-20
DRIVERS-1967: Serverless testing uses load balancer fronting single proxy

jmikola:
https://jira.mongodb.org/browse/DRIVERS-1967

Complements https://github.com/mongodb-labs/drivers-evergreen-tools/pull/194


source/serverless-testing/README.rst
```diff
@@ -27,47 +27,42 @@ Serverless instance, so it is recommended to create one manually via the scripts
...
 
-Drivers MUST add the username and password to both
-``SINGLE_ATLASPROXY_SERVERLESS_URI`` and ``MULTI_ATLASPROXY_SERVERLESS_URI`` to
-ensure that test clients can connect to the cluster. Drivers MUST use the final
-URI stored in ``SINGLE_ATLASPROXY_SERVERLESS_URI`` to configure internal clients
-for test runners (e.g. the internal MongoClient described by the `Unified Test
-Format spec <../../unified-test-format/unified-test-format.rst>`__).
+All tests MUST be run with wire protocol compression and authentication enabled.
+
+In contrast to the `Load Balancer testing`_, which has separate URIs for load
```
patrickfreed, line 49: This was probably discussed already, but why not have drivers spin up two separate instances per test run, one of which has only the single proxy behind the LB and one that has multiple?

source/unified-test-format/unified-test-format.rst
```diff
@@ -486,17 +489,22 @@ The structure of this object is as follows:
...
-    multiple servers. Otherwise, the test runner MUST use the URI of the load
-    balancer fronting a single server.
+    If the topology type is ``LoadBalanced`` and Atlas Serverless is not being
+    used, the test runner MUST use one of the two load balancer URIs described
+    in `Initializing the Test Runner`_ to configure the MongoClient. If
+    ``useMultipleMongoses`` is true or unset, the test runner MUST use the URI
+    of the load balancer fronting multiple servers. Otherwise, the test runner
+    MUST use the URI of the load balancer fronting a single server.
+
+    If the topology type is ``LoadBalanced`` and Atlas Serverless is being used,
```
patrickfreed, line 504: rather than ignoring, should we skip these tests if we're unable to use multiple proxies?

source/serverless-testing/README.rst
jmikola: I did discuss this originally with Kevin and argued that the only reason we'd care about multiple mongoses from the driver's perspective is if we were directly connected to mongos(es) and needed to validate server pinning. If we're going through a load balancer, there is no benefit (to the driver) to have multiple mongoses behind the LB and it simply becomes a limitation for configuring fail points.

Testing a client with multiple mongoses is certainly of interest to the Atlas Serverless team, but I think it's reasonable for them to provide that test coverage themselves (and I expect they do).

----

> but why not have drivers spin up two separate instances per test run

Having Evergreen provision two different clusters for the same test run would also be a significant departure and likely complicate _every_ driver's test suite. At present, most of our Evergreen configs apart from the Atlas connectivity tests expect to feed the test suite a single connection URI (with optional variations for single/multi LB URIs). But in those cases, there's only a single cluster under test.

What you're describing is having the script provision two completely different clusters and feeding the test suite two entirely different connection strings.

That said, we _could_ introduce this new test environment (i.e. single proxy behind LB) as its own test environment, and require drivers to use the original Atlas Serverless project to provision a LB with multiple proxies; however, tests using the multi-proxy LB URI would never be able to test fail points reliably. This takes me back to the original point that there's no benefit to drivers in doing so.

source/unified-test-format/unified-test-format.rst
jmikola: My understanding is that there's only a danger in ignoring `useMultipleMongoses` if the connection string included multiple backing servers and a test required `useMultipleMongoses: false` (e.g. for fail point usage). Then, we wouldn't reliably be able to target a single mongos.

This project introduces the opposite case (only a single backing server), so I don't think there's a risk of ignoring `useMultipleMongoses` and causing a test failure. If anything, I expect we just wouldn't meaningfully test pinning when we might otherwise expect to be able to. But since this test environment is in _addition to_ existing sharded cluster and load balancer environments, I don't think we're giving up any test coverage.

I haven't discovered any complications in the PHPLIB POC, which also suggests to me that it's safe to ignore `useMultipleMongoses` for Atlas Serverless. I could be missing some edge cases where tests (a) require a sharded cluster or LB and (b) have some assertion that expects there to be multiple backing servers, but if so I expect we'll find those when other drivers with actually connection pools POC or implement this and can revisit it then if needed.

To that last point, perhaps we should have a second driver POC this before merging.

kevinAlbs (approved):

source/retryable-reads/tests/README.rst
```diff
@@ -81,17 +81,21 @@ Each YAML file has the following keys:
...
+
+    The test runner MUST be informed whether or not Atlas Serverless is being
+    used in order to determine if this requirement is met (e.g. through an
+    environment variable or configuration option).
+
+    Note: the Atlas Serverless proxy imitates mongos, so the test runner is not
+    capable of determining if Atlas Serverless is in use by issuing commands
+    such as ``buildInfo`` or ``hello``. Furthermore, connections to Atlas
+    Serverless use a load balancer, so the topology will appear as
+    "load-balanced".
```
jmikola, line 98: FYI @isabelatkinson: @kevinAlbs requested I copy/paste the new text from the unified test format spec to these older test READMEs. There's probably not much to review here if you'd rather skip.

source/unified-test-format/unified-test-format.rst
```diff
@@ -392,16 +392,19 @@ The structure of this object is as follows:
   "sharded" implies "sharded-replicaset", but not vice versa).
 
 - ``serverless``: Optional string. Whether or not the test should be run on
```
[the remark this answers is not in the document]
jmikola: . and updated.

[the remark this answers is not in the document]
jmikola: Done.

source/unified-test-format/unified-test-format.rst
```diff
@@ -486,17 +489,22 @@ The structure of this object is as follows:
...
+    because connecting to multiple mongos servers is necessary to test session
     pinning.
 
-    If the topology type is ``LoadBalanced``, the test runner MUST use one of
-    the two load balancer URIs described in `Initializing the Test Runner`_
-    to configure the MongoClient. If ``useMultipleMongoses`` is true or
-    unset, the test runner MUST use the URI of the load balancer fronting
-    multiple servers. Otherwise, the test runner MUST use the URI of the load
-    balancer fronting a single server.
+    If the topology type is ``LoadBalanced`` and Atlas Serverless is not being
```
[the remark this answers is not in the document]
jmikola: Done.

source/serverless-testing/README.rst
```diff
@@ -95,44 +92,55 @@ when using these variables to help prevent accidental leaks.
...
+.. _CLOUDP-117288: https://jira.mongodb.org/browse/CLOUDP-117288
+
 Existing Spec Tests
 ===================
 
-Tests defined in the following specifications MUST be included in a driver's
-Atlas Serverless testing suite, including prose tests:
+Unified spec tests from all specifications MUST be run against Atlas Serverless.
+Since schema version 1.4, unified tests can specify Atlas Serverless
+compatibility in their ``runOnRequirements``.
```
jmikola, line 102: cc: @jyemin

source/serverless-testing/README.rst
```diff
@@ -95,44 +92,55 @@ when using these variables to help prevent accidental leaks.
...
-Note that the formats for the JSON/YAML tests of these specifications were
+Note that the legacy JSON/YAML test formats for these specifications were
 updated to include a new ``runOnRequirement`` specifically for Atlas Serverless
-testing, so make sure to update the affected test runners to support this
-requirement and then sync the tests. To ensure these requirements are enforced
-properly, the runner MUST be informed that it is running against an Atlas
-Serverless instance through some indicator (e.g. an environment variable).
+testing. To ensure these requirements are enforced properly, the runner MUST be
+informed that it is running against an Atlas Serverless instance through some
+indicator (e.g. an environment variable).
```
jmikola, line 125: I removed the following, as it only seemed relevant to the original implementation of serverless testing:

> so make sure to update the affected test runners to support this requirement and then sync the tests

source/serverless-testing/README.rst
jmikola: I understand there's be some benefit to running those tests against Atlas Serverless, although it does still seem like that'd be a matter of testing Atlas Serverless more than the driver itself. Unless I'm mistaken, I assume the driver's code paths are sufficiently tested by our non-serverless load balancer testing.

I don't mean to discount the importance of testing Atlas Serverless, since that's the reason we have connectivity tests in the first place, but I'm definitely of the opinion that it's not worth introducing the complexity (and CI overhead) at this time. Beyond the CI script changes, I think that would also require significant changes `runOnRequirement.server`, since we'd then need to differentiate between single and multi-proxy deployments.

If you're OK with taking no action here, feel free to resolve.

source/unified-test-format/unified-test-format.rst
jmikola: Asked about a second POC in Slack and @kevinAlbs said he was OK merging as-is with the passing PHPLIB POC.

PR #1187, opened 2022-04-20
DRIVERS-2228 Revert to using the `count` command for `estimatedDocumentCount`

patrickfreed:
DRIVERS-2228

This PR updates the CRUD spec to require using the `count` server command to implement `estimatedDocumentCount` now that it's being added back to the the stable API. The motivation for this is that `$collStats` does not work on views, meaning the changes drivers made to `estimatedDocumentCount` in 5.0 broke counting for some users. This change has the side effect that `estimatedDocumentCount` will now fail with `apiStrict: true` on 5.0.0-5.0.7 and 5.1.0-5.3.1, but it was decided that this change essentially constitutes a bug fix and therefore doesn't violate semver if released in a minor or patch version.

The actual behavior of the `count` command on views is to perform a slow count, so a requirement was added for drivers to document this as well.


source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -144,16 +122,11 @@ tests:
...
-                  - $group: { _id: 1, n: { $sum: $count }}
-              commandName: aggregate
+                count: *collection0Name
+              commandName: count
               databaseName: *database0Name
 
-  - description: "estimatedDocumentCount uses count on less than 4.9.0"
-    runOnRequirements:
-      - maxServerVersion: "4.8.99"
+  - description: "estimatedDocumentCount uses count"
```
kmahar, line 129: this and the tests below seem like they are redundant with the tests above, now that the behavior is unified below and above 4.9?

source/crud/crud.rst
```diff
@@ -741,30 +742,37 @@ older sharded clusters.
...
-error will be returned during execution. Drivers MUST interpret the server error code 26 as
-a ``0`` count.
-
-For server versions less than 4.9.0 (wire version 11 or under), the estimatedDocumentCount
-function is implemented using the ``count`` command with no query filter, skip,
-limit, or other options that would alter the results. Once again, the only supported
-option is maxTimeMS.
+The estimatedDocumentCount function is implemented using the ``count`` command
+with no query filter, skip, limit, or other options that would alter the
+results. Once again, the only supported option is maxTimeMS.
```
kmahar, line 747: the "once again" in the original text seemed to reference that this had already been mentioned at the start of the section. so maybe this should be rephrased?

source/crud/crud.rst
```diff
@@ -741,30 +742,37 @@ older sharded clusters.
...
+
+Drivers MUST document that, due to a bug in versions 5.0.0-5.0.7 of MongoDB,
+estimatedDocumentCount was not included in v1 of the Stable API, so users of the
+Stable API are recommended to upgrade their server version to 5.0.8+ or set
+``apiStrict: false`` to avoid encountering errors when using
+estimatedDocumentCount.
```
JamesKovacs, line 753: "Bug" is rather strong wording. `count` was intentionally omitted from the Stable API. It was an oversight that this would break estimatedDocumentCounts on views. I'm not sure how best to word this, but am open to suggestions.

JamesKovacs (changes requested):
Minor comment about correct terminology. Also need to update spec version, last modified date, and changes.

source/crud/tests/unified/estimatedDocumentCount.yml
patrickfreed: Good catch, fixed.

source/crud/crud.rst
```diff
@@ -754,30 +754,38 @@ older sharded clusters.
...
+with no query filter, skip, limit, or other options that would alter the
+results. The only supported options are listed in the
+``EstimatedDocumentCountOptions`` type defined above.
+
+Drivers MUST document that, due to a bug in versions 5.0.0-5.0.7 of MongoDB,
+estimatedDocumentCount was not included in v1 of the Stable API, so users of the
+Stable API are recommended to upgrade their server version to 5.0.8+ or set
+``apiStrict: false`` to avoid encountering errors when using
+estimatedDocumentCount.
+
```
jmikola, line 767: Can we just delete this entire paragraph? I think the final paragraph in this section better describes the server bug/regression and also clearly instructs drivers on how they should document this.

```suggestion
```

This may also address @JamesKovacs' concerns with referring to a "bug" here in the [preceding thread](https://github.com/mongodb/specifications/pull/1187/files#r854624143).

source/atlas-data-lake-testing/tests/estimatedDocumentCount.yml
```diff
@@ -13,9 +13,6 @@ tests:
       -
         command_started_event:
           command:
-            aggregate: *collection_name
-            pipeline:
-              - $collStats: { count: {} }
```
jmikola, line 18: If you've not done so already, can you take a quick grep through the specifications repo and ensure that `$collStats` appears no where else? This request assumes any instance of it would have been related to estimatedDocumentCount, so do take that into consideration if there's another legitimate instance left behind.

source/crud/crud.rst
```diff
@@ -754,30 +754,38 @@ older sharded clusters.
...
+Drivers MUST document that estimatedDocumentCount will perform a slower but
+correct count of the documents when performed against a view. In future server
+versions, this will be updated to provide a fast but estimated count instead,
+matching the behavior of the command on regular collections.
```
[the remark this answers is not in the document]
jmikola: Then I'd suggest we do whatever we can to ensure driver's get notified of a future fix in the server, so we can come back and update this otherwise open-ended caveat.

source/crud/crud.rst
jmikola: Points noted. I missed that the later paragraph is mostly referring to "release notes". In that case I'll retract this suggestion.

source/crud/crud.rst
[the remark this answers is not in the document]
patrickfreed: So I talked to Chris again and there is no timeline or ticket (besides definitely after 7.0), so he suggested linking to a central documentation point that will get updated as things change. The shell's documentation page has already been updated to [include this info](https://www.mongodb.com/docs/manual/reference/method/db.collection.estimatedDocumentCount/#mechanics), but I hesitate to link to there from driver docs as it can easily confuse users, especially of the Node driver. Instead, I updated the requirement to link to the [manual page for the `count` command](https://www.mongodb.com/docs/manual/reference/command/count/#behavior) itself, and if we agree with this change, I'll file a DOCS ticket to get that page updated with the same info that the shell's page has.

source/crud/crud.rst
patrickfreed: I took a second pass at this, let me know what you think. I also added a section to the Q&A about why it was decided that this change doesn't require a SemVer major version.

source/crud/crud.rst
```diff
@@ -759,33 +759,35 @@ with no query filter, skip, limit, or other options that would alter the
...
+Drivers MUST document that, due to an oversight in versions 5.0.0-5.0.7 of
+MongoDB, the ``count`` command, which estimatedDocumentCount uses in its
+implementation, was not included in v1 of the Stable API, and so users of the
+Stable API with estimatedDocumentCount are recommended to upgrade their server
+version to 5.0.8+ or set ``apiStrict: false`` to avoid encountering errors.
+
+Drivers MUST document that the ``count`` server command is used to implement
+estimatedDocumentCount and that users can visit `this link
+<https://www.mongodb.com/docs/manual/reference/command/count/#behavior>`_ for
+more information about the command's behavior.
```
jmikola, line 771: Feel free to disregard, but a suggestion to use a contextual name instead of "this".

```suggestion
estimatedDocumentCount and that users can find more information via
`Count: Behavior <https://www.mongodb.com/docs/manual/reference/command/count/#behavior>`_.
```

source/crud/crud.rst
```diff
@@ -2362,6 +2364,11 @@ Q: Where is ``singleBatch`` in FindOptions?
...
 
+Q: Why does reverting to using ``count`` instead of ``aggregate`` with
+``$collStats`` for estimatedDocumentCount not require a major version bump in
+the drivers, even though it might break users of the Stable API?
+  SemVer `allows <https://semver.org/#what-if-i-inadvertently-alter-the-public-api-in-a-way-that-is-not-compliant-with-the-version-number-change-ie-the-code-incorrectly-introduces-a-major-breaking-change-in-a-patch-release>`_ for a library to include a breaking change in minor or patch version if the change is required to fix another accidental breaking change introduced in a prior version and is not expected to further break a large number of users. Given that the original switching to ``$collStats`` was a breaking change due to it not working on views, the number of users using estimatedDocumentCount with ``apiStrict: true`` is small, and the server is backporting adding ``count`` to the Stable API, we felt that this change was acceptable to make in minor version releases of the drivers per the aforementioned allowance in the SemVer spec.
```
jmikola:
```suggestion
  SemVer `allows <https://semver.org/#what-if-i-inadvertently-alter-the-public-api-in-a-way-that-is-not-compliant-with-the-version-number-change-ie-the-code-incorrectly-introduces-a-major-breaking-change-in-a-patch-release>`_ for a library to include a breaking change in a minor or patch version if the change is required to fix another accidental breaking change introduced in a prior version and is not expected to further break a large number of users. Given that the original switch to ``$collStats`` was a breaking change due to it not working on views, the number of users using estimatedDocumentCount with ``apiStrict: true`` is small, and the server is back-porting the addition of ``count`` to the Stable API, we felt that this change was acceptable to make in minor version releases of the drivers per the aforementioned allowance in the SemVer spec.
```

jmikola (approved):

JamesKovacs (approved):
LGTM

kmahar (approved):

source/crud/crud.rst
patrickfreed: GH didn't let me auto commit this, so I just manually made this change.

PR #1191, opened 2022-04-25
add minimum version to bulkWrite spec tests

prestonvasquez:
[DRIVER-2290](https://jira.mongodb.org/browse/DRIVERS-2290)

Adding a minimum version to the pre 4.4 CRUD spec tests.

### DIFF
--- a/source/crud/tests/unified/bulkWrite-comment.json
+++ b/source/crud/tests/unified/bulkWrite-comment.json
@@ -412,6 +412,7 @@
       "description": "BulkWrite with comment - pre 4.4",
       "runOnRequirements": [
         {
+          "minServerVersion": "3.4.0",
           "maxServerVersion": "4.2.99"
         }
       ],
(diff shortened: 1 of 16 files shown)


jmikola (approved):
Assuming you manually verified that 3.4.0 is, indeed, the server version where write commands are validated for unrecognized options, LGTM.

prestonvasquez:
> Assuming you manually verified that 3.4.0 is, indeed, the server version where write commands are validated for unrecognized options, LGTM.

@jmikola I manually tested that `runCommand` errors as expected on 3.4.0 and that it does not on the latest version of 3.2 (3.2.22).  I couldn't find exactly where this `Unknown option to %s command` error was introduced to the mongo server, however.

jmikola:
That works for me. Provided 3.4.0 itself was tested, I'm not worried about some edge case where a later 3.2.x patch version might also be valid.

comandeo (approved):
LGTM! :+1:

comandeo:
https://jira.mongodb.org/browse/DRIVERS-2290

PR #1194, opened 2022-04-26
DRIVERS-2290 Correct ROR in YML

prestonvasquez:
[DRIVERS-2290](https://jira.mongodb.org/browse/DRIVERS-2290)

Correct ROR YAML object array from two objects to one.  The separation is causes [this failure](https://github.com/mongodb/specifications/runs/6178549023?check_suite_focus=true).


patrickfreed (approved):
LGTM, thanks for fixing!

jmikola (approved):
Note: since this fix only required modifying the YAML files, that suggests you modified JSON files directly in https://github.com/mongodb/specifications/pull/1191. There should never be a reason to do so (at least in the specifications repository), since YAML files are the canonical test files and the JSON is only be generated _from_ YAML.

PR #1199, opened 2022-04-28
[DRIVERS-2241] Clarifications on driver CSE testing with csfle

vector-of-bool:
This is a simple clarification on drivers making use of `csfle` during CSE testing. These points were discovered while integrating `csfle` into testing with the C driver.

Questions: 
- Should drivers always *prefer* to use `csfle` during testing?
- Should we provide a more uniform way to disable `csfle` for testing purposes?
- Which tests should _require_ mongocryptd? (e.g. Tests that specifically check that it was/wasn't spawned automatically will need to prevent csfle from loading.)


kevinAlbs (commented):
Changes look good. I have an open question about the explain test.

> Should drivers always prefer to use csfle during testing?

Preferring csfle SGTM. mongocryptd is not deprecated yet. I think drivers must test against both csfle and mongocryptd. I think it is sufficient to have one Evergreen task run with mongocryptd.

> Should we provide a more uniform way to disable csfle for testing purposes?

Ideally yes. But I do not have a good idea to do so without adding public API.

> Which tests should require mongocryptd? (e.g. Tests that specifically check that it was/wasn't spawned automatically will need to prevent csfle from loading.)

I think only the "Bypass spawning mongocryptd" prose test should require mongocryptd.

I am still unsure why the [explain](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/tests/explain.yml) test failed with csfle. Can you check the output of csfle from the C driver test? The query team may be able to investigate. It may suggest a bug, or result in all drivers needing to skip the explain test.

kevinAlbs (approved):
LGTM

vector-of-bool:
Before merging I am investigating a test failure in the C driver for the `explain` CSE test that only occurs when `csfle` is used, whether it's a bug in the C driver, csfle, or the test runner itself.

PR #1202, opened 2022-04-29
DRIVERS-2228 Fix runOnRequirements for `estimatedDocumentCount` stable API tests, add test for views

patrickfreed:
DRIVERS-2228

This is a follow up to the previous PR (#1187) which corrects the server version that the addition of `count` to the stable API was backported to. It also adds a test for `estimatedDocumentCount` on views.

### DIFF
--- a/source/crud/tests/unified/estimatedDocumentCount.yml
+++ b/source/crud/tests/unified/estimatedDocumentCount.yml
@@ -1,6 +1,6 @@
 description: "estimatedDocumentCount"
 
-schemaVersion: "1.9"
+schemaVersion: "1.0"
 
 createEntities:
   - client:
@@ -33,13 +33,6 @@ initialData:
       - { _id: 1, x: 11 }
       - { _id: 2, x: 22 }
       - { _id: 3, x: 33 }
-  - collectionName: *collection0ViewName
-    databaseName: *database0Name
-    documents: []
-    createOptions:
-      viewOn: *collection0Name
-      pipeline:
-        - { $match: { _id: { $gt: 1 } } }
 
 tests:
   - description: "estimatedDocumentCount always uses count"
@@ -148,13 +141,39 @@ tests:
               databaseName: *database0Name
 
   - description: "estimatedDocumentCount works correctly on views"
+    # viewOn option was added to the create command in 3.4
+    runOnRequirements:
+      - minServerVersion: "3.4.0"
     operations:
+      - name: dropCollection
+        object: *database0
+        arguments:
+          collection: *collection0ViewName
+      - name: createCollection
+        object: *database0
+        arguments:
+          collection: *collection0ViewName
+          viewOn: *collection0Name
+          pipeline: &pipeline
+            - { $match: { _id: { $gt: 1 } } }
       - name: estimatedDocumentCount
         object: *collection0View
         expectResult: 2
     expectEvents:
       - client: *client0
         events:
+          - commandStartedEvent:
+              command:
+                drop: *collection0ViewName
+              commandName: drop
+              databaseName: *database0Name
+          - commandStartedEvent:
+              command:
+                create: *collection0ViewName
+                viewOn: *collection0Name
+                pipeline: *pipeline
+              commandName: create
+              databaseName: *database0Name
           - commandStartedEvent:
               command:
                 count: *collection0ViewName
--- a/source/crud/tests/unified/estimatedDocumentCount.yml
+++ b/source/crud/tests/unified/estimatedDocumentCount.yml
@@ -1,6 +1,6 @@
 description: "estimatedDocumentCount"
 
-schemaVersion: "1.0"
+schemaVersion: "1.9"
 
 createEntities:
   - client:
--- a/source/crud/tests/unified/estimatedDocumentCount.yml
+++ b/source/crud/tests/unified/estimatedDocumentCount.yml
@@ -21,6 +21,10 @@ createEntities:
       id: &collection1 collection1
       database: *database0
       collectionName: &collection1Name coll1
+  - collection:
+      id: &collection0View collection0View
+      database: *database0
+      collectionName: &collection0ViewName coll0view
 
 initialData:
   - collectionName: *collection0Name
@@ -29,6 +33,13 @@ initialData:
       - { _id: 1, x: 11 }
       - { _id: 2, x: 22 }
       - { _id: 3, x: 33 }
+  - collectionName: *collection0ViewName
+    databaseName: *database0Name
+    documents: []
+    createOptions:
+      viewOn: *collection0Name
+      pipeline:
+        - { $match: { _id: { $gt: 1 } } }
 
 tests:
   - description: "estimatedDocumentCount always uses count"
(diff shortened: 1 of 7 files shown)


source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -1,6 +1,6 @@
 description: "estimatedDocumentCount"
 
-schemaVersion: "1.0"
+schemaVersion: "1.9"
```
jmikola, line 3: Noted this is due to `createOptions` used below.

kmahar (approved):

source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -29,6 +33,13 @@ initialData:
       - { _id: 1, x: 11 }
       - { _id: 2, x: 22 }
       - { _id: 3, x: 33 }
+  - collectionName: *collection0ViewName
+    databaseName: *database0Name
+    documents: []
+    createOptions:
+      viewOn: *collection0Name
+      pipeline:
+        - { $match: { _id: { $gt: 1 } } }
```
ShaneHarvey, line 42: Views were added in 3.4 and some drivers still support < 3.6 (go only?). I suggest moving the view creation to the test.operations itself and adding a minServerVersion: "3.4".

ShaneHarvey (changes requested):

source/crud/tests/unified/estimatedDocumentCount.yml
kmahar: seems like moving the creation to the operations might also allow us to lower the schema version, since we wouldn't need `createOptions`? anymore?

patrickfreed: This was the original route I tried, but the unified format doesnt contain a createCollection operation from what I could tell, and createOptions seemed like the supported approach. In order to do this without modifying the unified test format, we could use a runCommand though, would that be fine?

patrickfreed: And yeah, the only reason for the bumped schema version is for createOptions.

ShaneHarvey: It looks like we support createCollection:
```
$ git grep 'name: createCollection'
source/client-side-encryption/tests/fle2-CreateCollection.yml:35:      - name: createCollection
source/client-side-encryption/tests/fle2-CreateCollection.yml:150:      - name: createCollection
source/client-side-encryption/tests/fle2-CreateCollection.yml:265:      - name: createCollection
source/client-side-encryption/tests/fle2-CreateCollection.yml:437:      - name: createCollection
source/client-side-encryption/tests/fle2-CreateCollection.yml:547:      - name: createCollection
source/client-side-encryption/tests/fle2-CreateCollection.yml:601:      - name: createCollection
source/client-side-encryption/tests/fle2-CreateCollection.yml:711:      - name: createCollection
source/client-side-encryption/tests/fle2-CreateCollection.yml:881:      - name: createCollection
source/client-side-encryption/tests/fle2-CreateCollection.yml:1061:      - name: createCollection
source/collection-management/tests/clustered-indexes.yml:32:      - name: createCollection
source/collection-management/tests/clustered-indexes.yml:52:      - name: createCollection
source/collection-management/tests/clustered-indexes.yml:79:      - name: createCollection
source/collection-management/tests/timeseries-collection.yml:33:      - name: createCollection
source/collection-management/tests/timeseries-collection.yml:69:      - name: createCollection
source/transactions/tests/legacy/create-collection.yml:21:      - name: createCollection
source/transactions/tests/legacy/create-index.yml:86:      - name: createCollection
source/unified-test-format/tests/valid-pass/poc-transactions.yml:62:      - name: createCollection
```

kmahar: Swift and Rust both have operations for it already. But it is not discussed in the unified format spec. (should it be? @jmikola)

ShaneHarvey: I asked Jeremy this same question a few days ago and this is what he said:
> If the operation would logically come under a different spec there's no need to define it in the unified test format.
> This is why I didn't suggest adding rename for collections in recent PRs.
> But stuff like iterateOnce, special ops for creating change streams, and test runner ops certainly go in the unified test format.

patrickfreed: Huh, I couldve sworn I grepped for that and didnt see any unified tests using it. Mustve spelled it wrong or something. Anyways, will update this to use that instead of createOptions.

jmikola: Worst case, you could always fall back to using `runCommand`. I recently did that for `collMod` in [`change-streams-pre_and_post_images.yml`](https://github.com/mongodb/specifications/blob/master/source/change-streams/tests/unified/change-streams-pre_and_post_images.yml).

That said, there's enough tests using `createCollection` already that I think you can rely on that. The spec requires drivers to raise errors for unsupported operations so there should never be a case where this goes unnoticed. If anything, a driver would just realize they need to support the operation when they encounter a test file using it.

patrickfreed: Updated to use createCollection, also added minServerVersion 3.4 requirement and reverted the schemaversion back to 1.0

jmikola (approved):

kmahar (approved):

source/crud/tests/unified/estimatedDocumentCount.yml
```diff
@@ -135,3 +139,43 @@ tests:
...
+  - description: "estimatedDocumentCount works correctly on views"
+    # viewOn option was added to the create command in 3.4
+    runOnRequirements:
+      - minServerVersion: "3.4.0"
+    operations:
+      - name: dropCollection
+        object: *database0
+        arguments:
+          collection: *collection0ViewName
+      - name: createCollection
```
ShaneHarvey, line 152: For legacy reasons the Python driver issues a listCollections command before running createCollection. Could we add a `ignoreCommandMonitoringEvents: ["listCollections"]` to `client0` or use runCommand instead? Otherwise the test will fail during event assertion and I can't think of a better way to workaround it on the python side.

ShaneHarvey (changes requested):

source/crud/tests/unified/estimatedDocumentCount.yml
patrickfreed: How does Python pass the collection management tests (e.g. ["insertMany with duplicate ids"](https://github.com/mongodb/specifications/blob/master/source/collection-management/tests/timeseries-collection.yml#L63)), which use `createCollection` without ignoring those events? If possible, I'd rather not introduce a Python-specific exception to the spec tests if it can be avoided.

ShaneHarvey: We manually inserted the listCollections events into those files which is unfortunate because we have to maintain those edits every-time we resync the specs. But I agree with you. I'll find a workaround in python.

ShaneHarvey (approved):

PR #1205, opened 2022-05-03
DRIVERS-2310 Include encryptFields in expected FLE2 test command

addaleax:
In the “CreateCollection from encryptedFieldsMap”, it makes sense
to also assert on the encryptedFields option that ends up being
passed to the server.

- [ ] Update changelog.

### DIFF
--- a/source/client-side-encryption/tests/fle2-CreateCollection.yml
+++ b/source/client-side-encryption/tests/fle2-CreateCollection.yml
@@ -13,7 +13,7 @@ tests:
         kmsProviders:
           aws: {} # Credentials filled in from environment.
         encryptedFieldsMap:
-          default.encryptedCollection: {
+          default.encryptedCollection: &encrypted_fields0 {
                 "escCollection": "encryptedCollection.esc",
                 "eccCollection": "encryptedCollection.ecc",
                 "ecocCollection": "encryptedCollection.ecoc",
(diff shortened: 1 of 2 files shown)


source/client-side-encryption/tests/fle2-CreateCollection.yml
```diff
@@ -602,6 +602,18 @@ tests:
...
+                "eccCollection": "encryptedCollection.ecc",
+                "ecocCollection": "encryptedCollection.ecoc",
+                "fields": [
+                    {
+                        "path": "firstName",
+                        "bsonType": "string",
+                        "keyId": { "$binary": { "subType": "04", "base64": "AAAAAAAAAAAAAAAAAAAAAA==" }}
+                    }
+                ]
+            }
```
kevinAlbs, line 616: To test the expectation, this should be in the `command_started_event` below.

I suggest adding a YAML anchor to reference the encryptedFields used in create. Here is one proposal: https://github.com/mongodb/specifications/compare/master...kevinAlbs:add-encryptedFields-to-expectations?expand=1#diff-36980134f3bcb9a64658afc5058a1d6b49fa472a09486fccee2d8ab1a71fad6b

kevinAlbs (commented):
I like this idea. Thank you for the PR!

source/client-side-encryption/tests/fle2-CreateCollection.yml
addaleax: @kevinAlbs Ah, that’s much better even! Thank you, done :sparkles:

kevinAlbs (approved):
LGTM and passes in Go driver. Thank you!

PR #1208, opened 2022-05-05
DRIVERS-2313: valid-fail test for unsupported operation

jmikola:
https://jira.mongodb.org/browse/DRIVERS-2313

PHPLIB POC: https://github.com/mongodb/mongo-php-library/pull/923


benjirewis (approved):
LGTM. I misspoke when we talked offline and I said that the Go driver does not run the `valid-fail` tests for the unified test format spec. We _do_, and this test passes in the Go driver. We _don't_ run the `invalid` tests, as it wasn't trivial to recover potential `panic`s from invalid test files.

PR #1212, opened 2022-05-10
Use 6.0.0-rc4 in `csfle` download example

kevinAlbs:
5.3.1 does not have all of FLE 2.0 functionality. Per [slack](https://mongodb.slack.com/archives/CFPHLTRMK/p1652192975727899) it may not load correctly on macOS.

- [ ] Update changelog.

### DIFF
--- a/source/client-side-encryption/tests/README.rst
+++ b/source/client-side-encryption/tests/README.rst
@@ -351,6 +351,8 @@ for information on "disabling" csfle and setting csfle search paths.
 
       $ python3 mongodl.py --component=csfle --version=6.0.0-rc4 --out=./csfle/
 
+   Other versions of `csfle` are available. Please use the `--list` option to see versions.
+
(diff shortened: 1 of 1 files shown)


addaleax (approved):

source/client-side-encryption/tests/README.rst
```diff
@@ -349,7 +349,7 @@ for information on "disabling" csfle and setting csfle search paths.
 
    .. code-block:: shell
 
-      $ python3 mongodl.py --component=csfle --version=5.3.1 --out=./csfle/
+      $ python3 mongodl.py --component=csfle --version=6.0.0-rc4 --out=./csfle/
```
vector-of-bool, line 352: Can we have a note here about the availability of newer/older versions? I don't want to imply that this example shows "the correct version" to be used.

vector-of-bool (approved):

comandeo (approved):

PR #1213, opened 2022-05-10
DRIVERS-2017 Add spec and tests for rewrapManyDataKey and custom keyMaterial

eramongodb:
This PR is a part of [DRIVERS-2017](https://jira.mongodb.org/browse/DRIVERS-2017). It is being reviewed separately to unblock Drivers waiting to implement customer-requested support for data key rewrapping and creating data keys with custom key material.

### Relocate existing CSE tests into legacy subdirectory

To limit the scope of this PR, rewriting existing CSE tests using the unified test format is postponed for a future PR. Instead, existing CSE tests are moved into a `legacy` subdirectory for consistency with other spec tests that contain both unified and non-unified test files.

### Add test format wording for client side encryption

Minor updates to the unified test format spec to account for the addition of unified test files under the CSE spec directory.

### Add `ClientEncryption.rewrapManyDataKey()` to Client Side Encryption spec

Added Client Side Encryption spec wording for the new `rewrapManyDataKey()` method. Also added unified tests to guide Driver implementations and test runners.

Some Q&A and rationale documented in the Scope document have been carried over into the spec.

Some unified tests have been split by KMS provider to reduce the amount of time an individual test spends sending and receiving KMS requests. 

### Add DataKeyOpts.keyMaterial to Client Side Encryption spec

Added Client Side Encryption spec wording for the new custom `keyMaterial` field under `dataKeyOpts`.

### Add unified test files for `ClientEncryption.createKey()`

Added unified tests for `createKey()` to guide Driver implementations and test runners.


source/client-side-encryption/client-side-encryption.rst
```diff
@@ -1274,31 +1316,39 @@ selection error is propagated to the user.
...
-ClientEncryption.createDataKey().
+explicit encryption/decryption as well as key management functions.
+
+For key management functions that requires updating key documents in the key
+vault collection, the corresponding create, update, or delete operations MUST be
+done with write concern majority.
 
-For ClientEncryption.createDataKey(), the new document MUST be inserted
-into the key vault collection with write concern majority.
+For encryption/decryption and key management functions that requires reading
```
addaleax, line 1340:
```suggestion
For encryption/decryption and key management functions that require reading
```

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -1274,31 +1316,39 @@ selection error is propagated to the user.
...
 Key Vault collection
 ====================
 The key vault collection is the specially designated collection
 containing encrypted data keys. There is no default collection (user
 must specify). The key vault collection is used for automatic and
-explicit encryption/decryption as well as
-ClientEncryption.createDataKey().
+explicit encryption/decryption as well as key management functions.
+
+For key management functions that requires updating key documents in the key
```
addaleax, line 1336:
```suggestion
For key management functions that require updating key documents in the key
```

source/client-side-encryption/tests/unified/createKey-dataKeyOpts-keyMaterial.yml
```diff
@@ -0,0 +1,66 @@
+description: createKey-dataKeyOpts-keyMaterial
```
kevinAlbs, line 1: `createKey-dataKeyOpts-keyMaterial` may pass if the keyMaterial option is not applied. I do not see a way to make an assertion to guarantee that keyMaterial is applied without significant changes to the unified test format.

I suggest adding a prose test to the test README. Here is a suggested sketch of a prose test to validate keyMaterial is applied:

```
Use ``clientEncryption`` to create a data key with these ``DataKeyOpts``:

.. code::typescript

   class DataKeyOpts {
      masterKey: { "local": {} }
      keyMaterial: <TODO: some fixed keyMaterial>
   }

Store the returned value in ``originalKeyID``:

Use ``client`` to copy the data key with a fixed ``_id``:

.. code:: typescript

   got = client["keyvault"]["datakeys"].find_one({"_id": <originalKeyID>})
   got["_id"] = <newKeyID>
   client["keyvault"]["datakeys"].insert_one(got)

.. note:: Copying the key and setting a fixed ``_id`` is required because ciphertexts depend on the contents of ``_id``. This test relies on deterministic ciphertexts.

Use ``clientEncryption`` to encrypt the value "test" with these ``EncryptOpts``

.. code:: typescript

   class EncryptOpts {
      keyId : <newKeyID>
      algorithm: "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic",
   }

Assert the returned value is equal to <TODO: some fixed value>.
```

Here is a POC of this test based on the C implementation: https://github.com/mongodb/mongo-c-driver/commit/a149a612231eb42b2c638ebfeee2c819a871e6ce#diff-d363a5539f2fbabfadf2d7cc6cdb96a203fb63f2caa31bae731c3e12d97b204cR3204

kevinAlbs (commented):
This looks great. Adding the Q&A to the spec seems helpful to future readers.

I suggest a prose test for custom keyMaterial, and have minor comments about tests.

durran (changes requested):
Just a note for NODE, spec tests are part of NODE-4410

source/client-side-encryption/tests/unified/createKey-dataKeyOpts-keyMaterial.yml
eramongodb: To confirm, this prose test is taking advantage of the fact that data encryption using `AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic` depends on the value of the data key UUID and the (custom) key material, so we are deliberately overwriting the (randomly-generated) UUID of the data key with a predetermined value so that using we get a known and deterministic result when encrypting data. Is this correct?

source/client-side-encryption/tests/unified/rewrapManyDataKey-filter.yml
```diff
@@ -0,0 +1,207 @@
...
+                ordered: true
+                updates:
+                  - q:
+                      _id: *aws_key_id
+                    u:
+                      $set:
+                        masterKey: { provider: local }
+                        keyMaterial: { $$type: binData }
+                      $currentDate: { updateDate: true }
+                    upsert: { $$unsetOrMatches: false }
```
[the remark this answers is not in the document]
jmikola: Thanks for catching this. Just to link this back to a previous PR discussion: https://github.com/mongodb/specifications/pull/1105/files#r758652544

It'd be great if we could document this somewhere, but I'm not sure where would be best to do so. The main specs repo README seems too generic for this purpose -- I reckon we'd need a "best practices" document for writing tests (perhaps a `CONTRIBUTING.md` file).

source/client-side-encryption/tests/unified/createKey-provider-invalid.yml
```diff
@@ -0,0 +1,71 @@
...
+runOnRequirements:
+  - csfle: true
+
+createEntities:
+  - client:
+      id: &client0 client0
+      observeEvents:
+        - commandStartedEvent
+        - commandSucceededEvent
+        - commandFailedEvent
```
jmikola, line 14: This file doesn't perform any APM assertions.

source/client-side-encryption/tests/unified/createKey-provider-invalid.yml
```diff
@@ -0,0 +1,71 @@
...
+      databaseName: &keyvault keyvault
+  - collection:
+      id: &collection0 collection0
+      database: *database0
+      collectionName: &datakeys datakeys
+
+initialData:
+  - databaseName: *keyvault
+    collectionName: *datakeys
+    documents: []
```
jmikola, line 36: Is this relevant if all of the tests expect client errors?

source/client-side-encryption/tests/unified/createKey-dataKeyOpts-keyMaterial.yml
```diff
@@ -0,0 +1,65 @@
...
+    operations:
+      - name: createKey
+        object: *clientEncryption0
+        arguments:
+          kmsProvider: local
+          opts:
+            # "key_material" repeated 8 times.
+            keyMaterial: { $binary: { base64: "a2V5X21hdGVyaWFsa2V5X21hdGVyaWFsa2V5X21hdGVyaWFsa2V5X21hdGVyaWFsa2V5X21hdGVyaWFsa2V5X21hdGVyaWFsa2V5X21hdGVyaWFsa2V5X21hdGVyaWFs", subType: "00" } }
+        expectResult:
+          id: { $$type: binData }
```
jmikola, line 48: `createKey` is documented as returning a Binary object. Where does the `id` field come from here?

Should this test instead have `expectResult: { $type: binData }`?

source/client-side-encryption/tests/unified/createKey-dataKeyOpts-keyMaterial.yml
kevinAlbs: Correct.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -2136,6 +2204,15 @@ command with maxMessageSizeBytes=24MB. If after marking we determine
...
 in the end libmongocrypt would create multiple OP_MSGs to send.
 
+Support sessions in key management functions
+--------------------------------------------
+
+Key management functions currently assume there are no concurrent accesses of
+the key vault collection being operated on. To support concurrent access of the
+key vault collection, the key management functions may be overloaded to take an
+explicit session parameter as described in the
+`Drivers Sessions Specification <../sessions/driver-sessions.rst>`__.
```
addaleax, line 2214: “explicit session parameter” sounds like the signature would be required to be e.g. `rewrapManyDataKey(filter, session, opts)` or `rewrapManyDataKey(filter, opts, session)`. At least in the case of Node.js, it would be significantly easier (and match existing driver API usage) to make `session` a possible entry in `opts`. Would that be considered acceptable?

addaleax (approved):

kevinAlbs (approved):
LGTM, nicely done!

source/client-side-encryption/tests/unified/rewrapManyDataKey-provider-kmip.yml
```diff
@@ -0,0 +1,167 @@
...
+      id: &clientEncryption0 clientEncryption0
+      clientEncryptionOpts:
+        keyVaultClient: *client0
+        keyVaultNamespace: keyvault.datakeys
+        kmsProviders:
+          aws: { accessKeyId: { $$placeholder: 1 }, secretAccessKey: { $$placeholder: 1 } }
+          azure: { tenantId: { $$placeholder: 1 }, clientId: { $$placeholder: 1 }, clientSecret: { $$placeholder: 1 } }
+          gcp: { email: { $$placeholder: 1 }, privateKey: { $$placeholder: 1 } }
+          kmip: { endpoint: { $$placeholder: 1 } }
+          local: { key: { $$placeholder: 1 } }
```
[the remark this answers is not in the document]
eramongodb: The KMS credentials are required for decrypting the data keys that are being rewrapped. Given each data key (`modifiedCount: 5`) has a respective KMS provider, they require appropriate KMS credentials to be loaded from the environment.

> would it be preferable to just have a single test file for rewrapManyDataKey instead of duplicating these blocks

If we do not mind the test runner spending a disproportionate amount of time (over 20 seconds) running tests within a single test file, then I can revert the split and put them into the same test file. The reason for duplication is to test the NxN matrix of KMS provider rewrapping possibilities (e.g. AWS -> AWS, AWS -> Azure, etc.). If this is considered overkill, I can reduce the number of keys required to be rewrapped per test.

source/client-side-encryption/tests/unified/rewrapManyDataKey-decrypt_failure.yml
```diff
@@ -0,0 +1,69 @@
...
+      database: *database0
+      collectionName: &datakeys datakeys
+
+initialData:
+  - databaseName: *keyvault
+    collectionName: *datakeys
+    documents:
+      - _id: { $binary: { base64: "YXdzYXdzYXdzYXdzYXdzYQ==", subType: "04" } }
+        keyAltNames: ["aws_key"]
+        keyMaterial: { $binary: { base64: AQICAHhQNmWG2CzOm1dq3kWLM+iDUZhEqnhJwH9wZVpuZ94A8gFXJqbF0Fy872MD7xl56D/2AAAAwjCBvwYJKoZIhvcNAQcGoIGxMIGuAgEAMIGoBgkqhkiG9w0BBwEwHgYJYIZIAWUDBAEuMBEEDO7HPisPUlGzaio9vgIBEIB7/Qow46PMh/8JbEUbdXgTGhLfXPE+KIVW7T8s6YEMlGiRvMu7TV0QCIUJlSHPKZxzlJ2iwuz5yXeOag+EdY+eIQ0RKrsJ3b8UTisZYzGjfzZnxUKLzLoeXremtRCm3x47wCuHKd1dhh6FBbYt5TL2tDaj+vL2GBrKat2L, subType: "00" } }
```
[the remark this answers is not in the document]
eramongodb: No reason, just an inconsistency. Will remove the unnecessary quotes.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -824,6 +833,41 @@ example shows creating and referring to a data key by alternate name:
...
+-----------------------
+
+.. code:: typescript
+
+   class RewrapManyDataKeyResult {
+      bulkWriteResult: BulkWriteResult;
+   }
+
+``bulkWriteResult`` is the result of the bulk write operation used to update the
+key vault collection with rewrapped data keys.
```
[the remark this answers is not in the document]
eramongodb: Good suggestion. Will add.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -824,6 +833,41 @@ example shows creating and referring to a data key by alternate name:
    opts = EncryptOpts(keyAltName="name1", algorithm="AEAD_AES_256_CBC_HMAC_SHA_512-Random")
    clientencryption.encrypt("457-55-5462", opts)
 
+keyMaterial
+~~~~~~~~~~~
```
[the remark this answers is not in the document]
eramongodb: Not intentional. Thank you for the spot.

source/client-side-encryption/tests/README.rst
```diff
@@ -371,6 +373,31 @@ In the prose tests LOCAL_MASTERKEY refers to the following base64:
...
 Perform all applicable operations on key vault collections (e.g. inserting an example data key, or running a find command) with readConcern/writeConcern "majority".
 
+Custom Key Material Test
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+#. Create a ``MongoClient`` object (referred to as ``client``).
+
+#. Using ``client``, drop the collection ``keyvault.datakeys``.
+
+#. Create a ``ClientEncryption`` object (referred to as ``client_encryption``) with ``keyvault.datakeys`` as the ``keyVaultClient``.
```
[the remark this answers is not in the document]
eramongodb: I mixed up the collection and the entity used to specify the collection. Will change `keyvault.datakeys` -> `client` to be consistent with spec interface.

source/client-side-encryption/tests/README.rst
```diff
@@ -371,6 +373,31 @@ In the prose tests LOCAL_MASTERKEY refers to the following base64:
...
+
+#. Using ``client_encryption``, create a data key with a ``local`` KMS provider and the following custom key material (given as base64):
+
+.. code:: javascript
+
```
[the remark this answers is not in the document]
eramongodb: All examples of `.. code::` I could find include a whitespace between the start of the code block and the code itself, so I have done the same here for consistency.

source/client-side-encryption/tests/README.rst
```diff
@@ -371,6 +373,31 @@ In the prose tests LOCAL_MASTERKEY refers to the following base64:
...
+  xPTAjBRG5JiPm+d3fj6XLi2q5DMXUS/f1f+SMAlhhwkhDRL0kr8r9GDLIGTAGlvC+HVjSIgdL+RKwZCvpXSyxTICWSXTUYsWYPyu3IoHbuBZdmw2faM3WhcRIgbMReU5
+
+#. Find the resulting key document in ``keyvault.datakeys``, save a copy of the key document, then remove the key document from the collection.
+
+#. Replace the ``_id`` field in the copied key document with a UUID with base64 value ``AAAAAAAAAAAAAAAAAAAAAA==`` (16 bytes all equal to ``0x00``) and insert the modified key document into ``keyvault.datakeys``.
+
+#. Using ``client_encryption``, encrypt the string ``"test"`` with the modified data key using the ``AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic`` algorithm and assert the resulting value is equal to the following (given as base64):
+
+.. code:: javascript
+
```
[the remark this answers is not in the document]
eramongodb: See other comment regarding whitespace.

source/client-side-encryption/tests/unified/createKey-dataKeyOpts-keyMaterial-invalid_length.yml
```diff
@@ -0,0 +1,48 @@
...
+      databaseName: &keyvault keyvault
+  - collection:
+      id: &collection0 collection0
+      database: *database0
+      collectionName: &datakeys datakeys
+
+initialData:
+  - databaseName: *keyvault
+    collectionName: *datakeys
+    documents: []
```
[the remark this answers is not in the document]
eramongodb: It does not need to exist. Will remove.

source/client-side-encryption/tests/unified/createKey-provider-invalid.yml
eramongodb: Does the absence of `expectEvents` constitute an assertion that no events occurred? Or should I be including an empty `expectEvents` in these tests to assert that no commands should be executed (my original intent)?

source/client-side-encryption/tests/unified/createKey-provider-invalid.yml
eramongodb: I suppose not. Will remove.

source/client-side-encryption/tests/unified/createKey-dataKeyOpts-keyMaterial.yml
eramongodb: This is an inconsistency introduced during rebasing of other Key Management API drafts. I had wording in the unified test format spec that instructed drivers to report the value of the Binary object as `{ id: (value) }` because I was under the impression that `expectResult: { $$type: binData }` would not work as expected. However, it seems that it _does_ work as suggested, so I will go with that instead.

source/client-side-encryption/client-side-encryption.rst
eramongodb: This wording is borrowing the specification in `driver-sessions.rst` as linked:

>  All ``MongoDatabase`` methods that talk to the server SHOULD be overloaded to take an explicit session parameter. (See `why is session an explicit parameter?`_.)

I do not know if it would be considered acceptable to use a field in `opts` as an alternative. CC @jmikola.

source/client-side-encryption/tests/README.rst
```diff
@@ -371,6 +374,31 @@ In the prose tests LOCAL_MASTERKEY refers to the following base64:
 
 Perform all applicable operations on key vault collections (e.g. inserting an example data key, or running a find command) with readConcern/writeConcern "majority".
 
+Custom Key Material Test
```
jmikola, line 377: I didn't catch this earlier, but please make sure all prose tests in this file are numbered. By numbering them and not re-using numbers from retired tests, it's easier for drivers to keep prose tests in sync and track which have and have not been implemented.

See: https://github.com/mongodb/specifications#prose-test-numbering

source/client-side-encryption/client-side-encryption.rst
jmikola: The relevant text in [that spec](https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst#new-database-methods-that-take-an-explicit-session) is:

> When overloading methods to take a session parameter, the session parameter SHOULD be the first parameter. If overloading is not possible for your language, it MAY be in a different position or MAY be embedded in an options structure.

A number of drivers just take a session through an options struct (or keyword args) instead of overloading.

With respect to unified spec tests, the session is parameter is merely expressed as the `session` key in the `arguments` object of an [operation](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#operation).

source/client-side-encryption/tests/README.rst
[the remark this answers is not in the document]
jmikola: SGTM. I wasn't aware of that.

source/client-side-encryption/client-side-encryption.rst
addaleax: Cool – I think that clarifies everything I need to know here (i.e. yes, in the Node.js driver it should probably be part of the options object). Feel free to resolve :+1:

source/client-side-encryption/tests/unified/rewrapManyDataKey-provider-kmip.yml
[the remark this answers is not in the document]
eramongodb: After out-of-band conversation, concluded merging tests into a single file to avoid duplication and test file maintenance overhead is preferable to idiosyncracies of current C driver test runner progress reporting behavior.

source/client-side-encryption/tests/README.rst
```diff
@@ -371,8 +374,33 @@ In the prose tests LOCAL_MASTERKEY refers to the following base64:
...
+
+#. Using ``client_encryption``, create a data key with a ``local`` KMS provider and the following custom key material (given as base64):
+
+.. code:: javascript
+
+  xPTAjBRG5JiPm+d3fj6XLi2q5DMXUS/f1f+SMAlhhwkhDRL0kr8r9GDLIGTAGlvC+HVjSIgdL+RKwZCvpXSyxTICWSXTUYsWYPyu3IoHbuBZdmw2faM3WhcRIgbMReU5
+
+#. Find the resulting key document in ``keyvault.datakeys``, save a copy of the key document, then remove the key document from the collection.
+
+#. Replace the ``_id`` field in the copied key document with a UUID with base64 value ``AAAAAAAAAAAAAAAAAAAAAA==`` (16 bytes all equal to ``0x00``) and insert the modified key document into ``keyvault.datakeys``.
```
kevinAlbs, line 394:
```suggestion
#. Replace the ``_id`` field in the copied key document with a UUID with base64 value ``AAAAAAAAAAAAAAAAAAAAAA==`` (16 bytes all equal to ``0x00``) and insert the modified key document into ``keyvault.datakeys`` with majority write concern.
```

Majority write concern must be used since the driver uses [majority read concern to find key documents](https://github.com/mongodb/specifications/blob/d1458823bd810014df9da16d3a5354d2269ab865/source/client-side-encryption/client-side-encryption.rst#key-vault-collection).

addaleax (approved):

jmikola (approved):

source/client-side-encryption/tests/unified/createKey-kms_providers.yml
```diff
@@ -0,0 +1,134 @@
...
+      - name: find
+        object: *collection0
+        arguments:
+          filter: {}
+          projection: { _id: 0, masterKey: 1 }
+        expectResult:
+          - masterKey:
+              provider: aws
+              <<: *new_aws_masterkey
+  - description: create datakey with Azure KMS provider
```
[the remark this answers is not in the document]
eramongodb: Oh. I was under the wrong impression that line breaks were not permitted between tests. This would definitely help with readability. Will apply.

source/client-side-encryption/tests/unified/rewrapManyDataKey.yml
```diff
@@ -0,0 +1,273 @@
...
+            modifiedCount: 4
+            deletedCount: 0
+            upsertedCount: 0
+            upsertedIds: {}
+      - name: find
+        object: *collection0
+        arguments:
+          filter: {}
+          projection: { masterKey: 1 }
+          sort: { keyAltNames: 1 }
```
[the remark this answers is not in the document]
eramongodb: Good suggestion. Will add a note at top of test file explaining rationale and behavior of result document sorting.

jmikola (approved):

kevinAlbs (approved):
LGTM with a question about a possible accidental test removal.

PR #1216, opened 2022-05-16
Add serverless: forbid for collection management tests

jyemin:
- [n/a] Bump spec version and last modified date.
- [n/a] Update changelog.

Not sure why this wasn't caught in initial testing, but serverless tests are failing in Java driver.

### DIFF
--- a/source/collection-management/tests/clustered-indexes.json
+++ b/source/collection-management/tests/clustered-indexes.json
@@ -1,6 +1,6 @@
 {
   "description": "clustered-indexes",
-  "schemaVersion": "1.0",
+  "schemaVersion": "1.4",
   "runOnRequirements": [
     {
       "minServerVersion": "5.3",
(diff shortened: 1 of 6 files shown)


jyemin:
Java runs all unified tests on serverless, and relies on runOnRequirements to restrict.  Everyone should do this IMO.

> I think you should also forbid serverless on timeseries-collection and modifyCollection-pre_and_post_images, too

Those tests are not failing so I didn't change them.  Is time series not supported on serverless? I don't know why modify... is succeeded, but it seems to be.

jyemin:
Added serverless restriction to modify collection test

benjirewis (approved):
Thanks! LGTM

jmikola:
@jyemin: Was there a DRIVERS ticket for this change? I assume the new commit will be relevant for drivers to sync to.

If there will be a ticket created for this, I think it'd also be a good opportunity to knock out [DRIVERS-2325](https://jira.mongodb.org/browse/DRIVERS-2325) since the same test file will be synced.

PR #1219, opened 2022-05-18
DRIVERS-2285 Add `wallTime` field to change stream events

abr-egn:
DRIVERS-2285

This adds the `wallTime` field now that it will always be populated in server versions 6.0+.  https://github.com/mongodb/mongo-rust-driver/pull/658 is the (very small) Rust PoC for this change.


benjirewis (approved):
LGTM and passes in the Go driver (although we not have a native type for `ChangeStreamDocument`).

PR #1220, opened 2022-05-18
Add new change stream events to spec

baileympearson:
- [x] Update changelog.

### DIFF
--- a/source/change-streams/change-streams.rst
+++ b/source/change-streams/change-streams.rst
@@ -137,7 +137,8 @@ If an aggregate command with a ``$changeStream`` stage completes successfully, t
      * MUST NOT err when they encounter a new `operationType`. Unknown `operationType`
      * values may be represented by "unknown" or the literal string value.
      */
-    operationType: "insert" | "update" | "replace" | "delete" | "invalidate" | "drop" | "dropDatabase" | "rename";
+    operationType: "insert" | "update" | "replace" | "delete" | "invalidate" | "drop" | "dropDatabase" | "rename" | "createIndex"
+        | "dropIndex" | "collMod" | "create" | "shardCollection" | "refineCollectionShardKey" | "reshardCollection" | "chunkMigrated";
 
     /**
      * Contains two fields: "db" and "coll" containing the database and
@@ -152,8 +153,21 @@ If an aggregate command with a ``$changeStream`` stage completes successfully, t
      * Only present for ops of type 'rename'.
      *
      * The namespace, in the same format as `ns`, that a collection has been renamed to.
+     * 
+     * This field is also present in the `operationDescription` field for `rename` events
+     * but has been left for backwards compatibility.
      */
     to: Optional<Document>;
+
+    /**
+     * Only present for ops of type 'rename', 'createIndexes', 'dropIndexes', 'shardCollection', 'reshardCollection', 'refineCollectionShardKey'.
+     * Only present when the `showExpandedEvents` change stream option is enabled.
+     *
+     * An description of the operation.
+     * 
+     * @since 6.1.0
+     */
+    operationDescription: Optional<Document>
 
     /**
      * Only present for ops of type ‘insert’, ‘update’, ‘replace’, and
@@ -169,10 +183,15 @@ If an aggregate command with a ``$changeStream`` stage completes successfully, t
     /**
      * Only present for ops of type ‘update’.
      *
-     * Contains a description of updated and removed fields in this
-     * operation.
+     * When the `showRawUpdateDescription` change stream pipeline stage option is set,
+     * this field contains the raw oplog entry corresponding to the update event.  The
+     * format of the oplog is internal and subject to change, and drivers should pass
+     * the data through.  
+     * 
+     * When the `showRawUpdateDescription` change stream pipeline stage option is not set,
+     * this field contains a description of updated and removed fields in this operation.
      */
-    updateDescription: Optional<UpdateDescription>;
+    updateDescription: Optional<UpdateDescription | Document>;
 
     /**
      * Always present for operations of type 'insert' and 'replace'. Also
@@ -203,6 +222,35 @@ If an aggregate command with a ``$changeStream`` stage completes successfully, t
      * pre-image is unavailable, this will be explicitly set to null.
      */
     fullDocumentBeforeChange: Document | null;
+
+    /**
+     * The `wall` time from the oplog entry corresponding to the change event.
+     *
+     * Only present when the `showExpandedEvents` change stream option is enabled.
+     *  
+     * @since 6.0.0
+     */
+    wallTime: Date;
+
+    /**
+     * The `ui` field from the oplog entry corresponding to the change event.
+     * 
+     * Only present when the `showExpandedEvents` change stream option is enabled and for the following events
+     *  - 'insert'
+     *  - 'update'
+     *  - 'delete'
+     *  - 'createIndexes'
+     *  - 'dropIndexes'
+     *  - 'modify'
+     *  - 'drop'
+     *  - 'create'
+     *  - 'shardCollection'
+     *  - 'reshardCollection'
+     *  - 'refineCollectionShardKey'
+     *  
+     * @since 6.0.0
+     */
+    collectionUUID: UUID;
   }
 
   class UpdateDescription {
@@ -490,6 +538,22 @@ Driver API
      * @note this is an aggregation command option
      */
     comment: Optional<any>
+
+    /**
+     * Enables the server to send the 'expanded' list of change stream events.
+     * The list of events included with this flag set are
+     * - createIndexes
+     * - dropIndexes
+     * - modify
+     * - create
+     * - shardCollection
+     * - reshardCollection
+     * - refineCollectionShardKey
+     * 
+     * @since 6.0.0 // todo - confirm this date
+     * @note this is an option of the change stream pipeline stage
+     */
+    showExpandedEvents: Optional<Boolean>
   }
 
 **NOTE:** The set of ``ChangeStreamOptions`` may grow over time.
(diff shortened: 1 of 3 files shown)


source/change-streams/change-streams.rst
```diff
@@ -137,7 +137,21 @@ If an aggregate command with a ``$changeStream`` stage completes successfully, t
...
+                  | "drop" 
+                  | "dropDatabase" 
+                  | "rename" 
+                  | "createIndexes"
+                  | "dropIndexes" 
+                  | "modify"
+                  | "create" 
+                  | "shardCollection" 
+                  | "refineCollectionShardKey" 
+                  | "reshardCollection";
```
baileympearson, line 155: `refineCollectionShardKey` and `reshardCollection` events are behind a (currently unsupported) feature flag.  I included them here for completeness, but notice that there are no spec tests testing for these new events.  If there are no objections, I can file a follow up ticket to add tests for these events once they're available.

source/change-streams/change-streams.rst
```diff
@@ -9,7 +9,7 @@ Change Streams
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: 3.6
-:Last Modified: 2022-05-17
+:Last Modified: 2022-05-19
 :Version: 1.15
```
benjirewis, line 13: We'll want a minor version bump for these changes, I think.

source/change-streams/change-streams.rst
```diff
@@ -137,7 +137,21 @@ If an aggregate command with a ``$changeStream`` stage completes successfully, t
...
+                  | "drop" 
+                  | "dropDatabase" 
+                  | "rename" 
+                  | "createIndexes"
+                  | "dropIndexes" 
+                  | "modify"
+                  | "create" 
+                  | "shardCollection" 
+                  | "refineCollectionShardKey" 
+                  | "reshardCollection";
```
benjirewis, line 154: What about `migrateLastChunkFromShard`?

source/change-streams/change-streams.rst
```diff
@@ -155,6 +169,16 @@ If an aggregate command with a ``$changeStream`` stage completes successfully, t
      */
     to: Optional<Document>;
 
+    /**
+     * Only present for ops of type 'rename', 'createIndexes', 'dropIndexes', 'shardCollection', 'reshardCollection', 'refineCollectionShardKey'.
```
benjirewis, line 173: In reading the syntax document, it looks like `operationDescription` is also available for `modify` and `create`?

source/change-streams/change-streams.rst
```diff
@@ -155,6 +169,16 @@ If an aggregate command with a ``$changeStream`` stage completes successfully, t
...
+     * Only present when the `showExpandedEvents` change stream option is enabled.
+     *
+     * An description of the operation.
```
benjirewis, line 176:
```suggestion
     * A description of the operation.
```

source/change-streams/change-streams.rst
```diff
@@ -496,6 +540,24 @@ Driver API
...
+     * The list of events included with this flag set are
+     * - createIndexes
+     * - dropIndexes
+     * - modify
+     * - create
+     * - shardCollection
+     * - reshardCollection
+     * - refineCollectionShardKey
+     * 
+     * This flag is available in server versions greater than 6.0.0.  `reshardCollection` and
```
benjirewis, line 555:
```suggestion
     * This flag is available in server versions greater than 6.0.0. `reshardCollection` and
```

source/change-streams/change-streams.rst
```diff
@@ -496,6 +540,24 @@ Driver API
      * @note this is an aggregation command option
      */
     comment: Optional<any>
+
+    /**
+     * Enables the server to send the 'expanded' list of change stream events.
+     * The list of events included with this flag set are
```
benjirewis, line 546:
```suggestion
     * The list of additional events included with this flag set are
```

source/change-streams/change-streams.rst
```diff
@@ -1007,3 +1069,5 @@ Changelog
 +------------+------------------------------------------------------------+
 | 2022-05-17 | Added ``wallTime`` to ``ChangeStreamDocument``.            |
 +------------+------------------------------------------------------------+
+| 2022-05-19 | Support new change stream events for C2C replication.      |
```
benjirewis, line 1072:
```suggestion
| 2022-05-19 | Support new change stream events with showExpandedEvents.      |
```
I'm not sure we even need to mention C2C replication. We're really just supporting the `showExpandedEvents` flag.

benjirewis (commented):
Looking great 🧑‍🔧 ! Some initial comments; I'll leave another round of review for the actual tests.

source/change-streams/tests/unified/change-streams-showExpandedEvents.yml
```diff
@@ -0,0 +1,299 @@
+description: "change-streams-showExpandedEvents"
+schemaVersion: "1.7"
+runOnRequirements:
+  - minServerVersion: "6.0.0"
+    topologies: [ replicaset, sharded-replicaset, sharded ]
```
benjirewis, line 5: While I don't see any failures in my Go driver POC, I worry that some of these tests will fail on sharded clusters and we may need to investigate the failures as part of DRIVERS-2323. Leaving `sharded-replicaset` and `sharded` here seems fine for now (especially since we need those topologies to test `shardCollection`).

source/change-streams/tests/unified/change-streams-showExpandedEvents.yml
```diff
@@ -0,0 +1,299 @@
...
+        object: *changeStream0
+        expectResult:
+          operationType: rename
+          ns:
+            db: *database0
+            coll: *collection0
+          to:
+            db: *database0
+            coll: *existing-collection
+          operationDescription:
```
benjirewis, line 179: "when showExpandedEvents is true, new fields on change stream events are handled appropriately" fails in the Go driver because the `operationDescription` of the `rename` change event _also_ contains `{"to": {"db": "database0","coll": "foo"}` along with a `dropTarget`. Is that expected?

source/change-streams/tests/unified/change-streams-showExpandedEvents.yml
```diff
@@ -0,0 +1,299 @@
...
+      - name: dropIndex
+        object: *collection0
+        arguments:
+          name: *index1
+      - name: iterateUntilDocumentOrError
+        object: *changeStream0
+        expectResult:
+          operationType: dropIndexes
+
+  - description: "when showExpandedEvents is true, create events are reported"
```
benjirewis, line 225: We might want a test for the creation of views, too. Note that we usually use `createCollection` with a `viewOn` argument to do this (see [here](https://github.com/mongodb/specifications/blob/master/source/crud/tests/unified/estimatedDocumentCount.yml#L152-L158)).

source/change-streams/tests/unified/change-streams-showExpandedEvents.yml
```diff
@@ -0,0 +1,299 @@
+description: "change-streams-showExpandedEvents"
+schemaVersion: "1.7"
+runOnRequirements:
+  - minServerVersion: "6.0.0"
+    topologies: [ replicaset, sharded-replicaset, sharded ]
+createEntities:
```
benjirewis, line 6: There are a lot of entities here. I don't see usages of `globalClient`, `database1`, `collection1`, `globalDatabase0`, `globalCollection0`, `globalDatabase1`, `globalCollection1`, `globalDb1Collection0` or `globalDb0Collection1` in the tests, but I might be missing something. Do we need those entities?

source/change-streams/change-streams.rst
baileympearson: `migrateLastChunkFromShard` is guarded behind `showSystemEvents`, not `showExpandedEvents`. This flag is undocumented.

I think that Go may have implemented this change event for atlas to use internally but other drivers shouldn't need to expose this event.

source/change-streams/change-streams.rst
baileympearson: Good call.  Done!

source/change-streams/change-streams.rst
baileympearson: Yeah, you're right.  I got this list from the section in the syntax document that says:

> operationDescription
> >  The specification of the event. In addition to the createIndexes, dropIndexes, shardCollection, reshardCollection,  refineCollectionShardKey and migrateLastChunkFromShard listed above, the existing rename event will also include this field. 

But I confirmed that `operationDescription` is present for `create` and `modify` as well.

source/change-streams/tests/unified/change-streams-showExpandedEvents.yml
baileympearson: You're right that those fields are expected.  I'm not sure why the Node runner doesn't error here.  That's something I'll look into and include in the Node POC.  For the spec tests, I'll add the `to` field to the events as well.

source/change-streams/tests/unified/change-streams-showExpandedEvents.yml
baileympearson: Those are leftover from my initial tests trying to write tests for `refineCollectionShardKey` and `reshardCollection` - I must have forgotten to remove them.  I remove any of the unused entities.

source/change-streams/tests/unified/change-streams-showExpandedEvents.yml
baileympearson: Great catch here - this was actually a bug in our matching function in the unified runner 😄 .

I fixed it as a part of the Node POC

source/change-streams/tests/unified/change-streams-showExpandedEvents.yml
```diff
@@ -0,0 +1,296 @@
...
+      - name: createCollection
+        object: *shardedDb
+        arguments:
+          collection: *shardedCollection
+      - name: createChangeStream
+        object: *shardedCollection
+        arguments:
+          pipeline: []
+          showExpandedEvents: true
+          showSystemEvents: true
```
abr-egn, line 284: Since `showSystemEvents` isn't mentioned in the spec yet, maybe this test should wait for a follow-up?

abr-egn (commented):
Looks good overall, and passes in the Rust driver!

abr-egn (approved):
LGTM!

benjirewis (approved):
LGTM! Great work 🧑‍🔧

nbbeeken (changes requested):

PR #1221, opened 2022-05-18
DRIVERS-1096 Convert command monitoring tests to unified format

kmahar:
I'm currently working on figuring out how to adopt the unified format to support logging assertions. Since I started with command logging (which I plan to update the command monitoring spec to cover) I started to look at the existing monitoring tests to get a sense of what to cover. I discovered there are not actually very many of these and that they are quite repetitive, mostly running the same scenario with different CRUD commands, so I just went ahead and spent a couple hours converting them over. 

I've tested these in both Rust and Swift and they pass on all topologies and server versions 3.6+. (Excluding serverless on Rust as those tests still need to be fixed per DRIVERS-1967). ~, although we have not done DRIVERS-1967 for either driver yet so I cannot test them against serverless yet. If anyone else who has done that could try them out that would be great, otherwise we can wait to merge this until Rust or Swift has done that ticket.~


source/command-monitoring/tests/unified/command.yml
```diff
@@ -0,0 +1,94 @@
...
+initialData:
+  - collectionName: *collectionName
+    databaseName: *databaseName
+    documents:
+      - { _id: 1, x: 11 }
+
+tests:
+  - description: "A successful command"
+    operations:
+      - name: distinct
```
kmahar, line 30: I switched to used `distinct` instead of `count` here because:
1) drivers don't have `count` API methods anymore
2) it seemed best to avoid any dependency on the recent changes around `estimatedDocumentCount` implementations

source/command-monitoring/tests/unified/command.yml
```diff
@@ -0,0 +1,94 @@
...
+              commandName: distinct
+              databaseName: *databaseName
+          - commandSucceededEvent:
+              reply:
+                ok: 1
+                values:
+                  - 11
+              commandName: distinct
+
+  - description: "A failed command"
```
kmahar, line 51: this was also `count`. I switched it to just be a copy of the "failed find" test from find.yml. I'm figuring we should delete one of them, but I left both in for now for discussion of whether there is any value in testing some different command here instead

source/command-monitoring/tests/unified/deleteMany.yml
```diff
@@ -0,0 +1,72 @@
...
+                deletes:
+                  - { q: { _id: { $gt: 1 }}, limit: 0 }
+                ordered: true
+              commandName: delete
+              databaseName: *databaseName
+          - commandSucceededEvent:
+              reply: { ok: 1, n: 2 }
+              commandName: delete
+
+  - description: "A successful deleteMany with write errors"
```
kmahar, line 51: this test doesn't make quite the same assertions about the reply as the legacy runner asked for:
* in command succeeded event's reply, errorCode exists
*in command succeeded event's reply,  errmsg exists and is non-empty

given that the data ends up nested in the `writeErrors` field it didn't seem that I could easily assert on it without exhaustively listing out the fields in each document in `writeErrors` (since they are not top-level). 
I also couldn't have asserted on a string being non-empty anyway, I think?
it seemed reasonable enough to me to just check that the operation throws a server-side error, and rely on the server to include errorCode and errmsg in that error, but let me know what you think. 

(same applies for the other "a successful x with write errors" tests)

source/command-monitoring/tests/unified/find.yml
```diff
@@ -0,0 +1,290 @@
...
+                filter: { _id: { $gte: 1 }}
+                sort: { _id: 1 }
+                batchSize: 3
+              commandName: find
+              databaseName: *databaseName
+          - commandSucceededEvent:
+              reply:
+                ok: 1
+                cursor:
+                  id: { $$type: [ int, long ] }
```
kmahar, line 129: the legacy runner wants you to assert consistency between non-zero cursor ids within single tests, but I couldn't think of a way to do that in the unified format, and that also seems more like a test for cursor behavior than command monitoring behavior so it didn't seem too important to try to replicate.

source/command-monitoring/tests/unified/find.yml
```diff
@@ -0,0 +1,290 @@
...
+                  id: 0
+                  ns: *namespace
+                  nextBatch:
+                    - { _id: 4, x: 44 }
+                    - { _id: 5, x: 55 }
+              commandName: getMore
+
+  - description: "A successful find with a getMore and killCursors"
+    runOnRequirements:
+      - maxServerVersion: "2.6.99"
```
kmahar, line 156: we can probably just delete this test, unless anyone is still maintaining a driver version that supports 2.6? 

if they are, they should test this, as I can't with Swift or Rust.

source/command-monitoring/tests/unified/unacknowledgedBulkWrite.yml
```diff
@@ -0,0 +1,56 @@
...
+        writeConcern: { w: 0 }
+
+initialData:
+  - collectionName: *collectionName
+    databaseName: *databaseName
+    documents:
+      - { _id: 1, x: 11 }
+
+tests:
+# On a 2.4 server, no GLE is sent and requires a client-side manufactured reply
```
kmahar, line 30: this comment is copied from the original test. I'm not sure if it's relevant for anyone anymore though

kmahar:
Update: Swift serverless testing has been fixed, and I've now confirmed these tests pass against serverless as well.

source/command-monitoring/tests/unified/command.yml
```diff
@@ -0,0 +1,94 @@
...
+
+  - description: "A successful command with a non-primary read preference"
+    operations:
+      - name: distinct
+        object: *collection
+        arguments:
+          fieldName: x
+          filter: {}
+          readPreference:
+            mode: primaryPreferred
```
jmikola, line 79: I believe this may be one of the first unified spec tests that specifies a `readPreference` argument on an operation. This is permitted per [Special Handling for Arguments](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#special-handling-for-arguments) but I wanted to confirm that it makes sense to do so.

The APM assertion below doesn't actually care about this option being specified, so perhaps we can remove it.

kmahar: I think this test is meant to check drivers correctly handle this spec requirement: https://github.com/mongodb/specifications/blob/master/source/command-monitoring/command-monitoring.rst#read-preference

that said, I think that is only relevant when drivers are using OP_QUERY and the $query wrapping of command arguments applies. so maybe it is not worth keeping around?

source/command-monitoring/tests/unified/find.yml
```diff
@@ -0,0 +1,290 @@
...
+          skip: 2
+          comment: "test"
+          hint: { _id: 1 }
+          max: { _id: 6 }
+          maxTimeMS: 6000
+          min: { _id: 0 }
+          returnKey: false
+          showRecordId: false
+          readPreference:
+            mode: primaryPreferred
```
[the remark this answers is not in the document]
kmahar: I think the old test was maybe meant to cover this language from the section linked above: 
> The read preference will subsequently be dropped as it is considered metadata and metadata is not currently provided in the command events.

If so though, it would probably make sense to assert here though that the read preference is unset in the command.

similarly to above we could use a separate collection if we want to avoid per-operation read preference

source/command-monitoring/tests/unified/find.yml
jmikola: I concur. Asserting that `getMore` uses the same cursor ID value seems beyond the scope of this spec. When/if we get a cursor spec, that seems like something we can offload to a prose test. Granted, if drivers weren't behaving correctly in that regard a _lot_ of spec tests would start failing, so I'm also not too concerned about this.

source/command-monitoring/tests/unified/find.yml
jmikola: You can ask in `#driver-devs` but I think there's a few drivers that still support older servers. I'm not sure if they care about this spec test, though, as they might have their own tests in place for that behavior.

source/command-monitoring/tests/unified/unacknowledgedBulkWrite.yml
jmikola: I'm in favor of dropping this comment. If you'd like to keep it, I'd propose moving it above the `commandSucceededEvent` line down below.

source/command-monitoring/tests/unified/command.yml
jmikola: I'd rather delete this than keep a pre-3.6 test around. Especially if the Node team doesn't doesn't care.

source/command-monitoring/tests/unified/deleteMany.yml
jmikola: > I suppose we could assert via expectError but I'm not sure that would help us verify that the command failed event contains the expected info too.

Correct. That would be more a test of CRUD than the APM spec. I'm OK with what you have here and just asserting that `writeErrors` is an array.

source/command-monitoring/tests/unified/deleteMany.yml
```diff
@@ -0,0 +1,79 @@
...
+    operations:
+      - name: deleteMany
+        object: *collection
+        arguments:
+          filter: { _id: { $nothing: 1 } }
```
jmikola, line 56: Nit: there is prior art for `$unsupportedQueryOperator` and `$unsupported` if you think that'd be more self-documenting.

source/command-monitoring/tests/unified/insertOne.yml
```diff
@@ -0,0 +1,77 @@
+description: "insertMany"
```
jmikola, line 1:
```suggestion
description: "insertOne"
```

source/command-monitoring/tests/unified/command.yml
```diff
@@ -0,0 +1,75 @@
+description: "commands"
```
jmikola, line 1: For consistency with the file name.

```suggestion
description: "command"
```

source/command-monitoring/tests/unified/unacknowledgedBulkWrite.yml
```diff
@@ -0,0 +1,55 @@
+description: "unacknowledged-bulkWrite"
```
jmikola, line 1:
```suggestion
description: "unacknowledgedBulkWrite"
```

source/command-monitoring/tests/unified/command.yml
kmahar: any opinion on keeping or deleting this test @durran?

source/command-monitoring/tests/unified/deleteMany.yml
kmahar: nice idea, updated all to `$unsupported`.

source/command-monitoring/tests/unified/command.yml
durran: While the Node 3.7 still supports pre 3.6 servers and this works, being that our current release is 4.6.0 we wouldn't backport these changes. I'd be in preference of deleting this test.

jyemin: Java and Go still support pre-3.6 servers, but I still think we should make little if any effort to maintain pre-3.6 tests

jmikola (approved):

durran (approved):

PR #1229, opened 2022-05-24
DRIVERS-2017 Add remaining Key Management API functions to spec and tests

eramongodb:
- [x] Update changelog.

### Description

This PR is the last in the series resolving the DRIVERS-2017 ticket to add Key Management API functions to Drivers for parity with mongosh. It adds the following functions to the `ClientEncryption` object:

- `rewrapManyDataKey()` (added in https://github.com/mongodb/mongo-c-driver/pull/991)
- `deleteKey()`
- `getKey()`
- `getKeys()`
- `addKeyAlternateName()`
- `removeKeyAlternateName()`
- `getKeyByAltName()`

Every one of these functions serve as lightweight wrappers around simple operations on the key vault collection that drivers are already capable of.

### Unique Index on `keyAltNames`

mongosh functions are conducted on a `KeyVault` object that is specified to create an unique index on `keyAltNames` if one does not already exist. Analagous behavior is NOT required to be implemented by drivers due to unresolved historical issues regarding demanding excessive user privileges (see the rationale under "Why aren't we creating a unique index in the key vault collection?" in the CSE spec). Instead, a prose test, "12. Unique Index on keyAltNames", is added to the CSE spec to test and ensure implementations of the key management API are capable of handling duplicate key errors triggered by a unique index. None of the currently proposed unified tests require the existence of the unique index for correctness.

### Return Types

Because the key management API functions are lightweight wrappers around collection operations, their return values are consistent with the interface of the underlying commands:

- `deleteKey()` returns the result of `deleteOne()` (a `DeleteResult`).
- `getKey()`, `getKeys()`, and `getKeyByAltName()` return the result of `find()` (an `Iterable<Document>`).
- `addKeyAlternateName()` and `removeKeyAlternateName()` return the document in the `value` field of the result of `findOneAndModify()` (an `Optional<Document>`).

### Collection Operations

Most operations require only a single operation. However, `removeKeyAlternateName()` may require a second operation to remove the `keyAltNames` field if, after removal of a keyAltName, the `keyAltNames` field would otherwise be empty. This is to satisfy the invariants of the unique index on `keyAltNames`, which are conditioned on the existence of the `keyAltNames` field.
Note again that drivers are _not_ currently being required to implement the _creation of_ the unique index (though are permitted to), but must still conform to its invariants.

The CSE spec DOES NOT currently specify what should occur if the operation to remove the `keyAltNames` field fails after successfully removing the last keyAltName on a key document (which violates the invariants of the unique index on keyAltNames).

### Unified Tests

Several tests make use of `find` or `aggregate` operations to validate the state of the collection after a given key management operation to work around having to deal with exact matching or non-deterministic results (e.g. order of resulting documents of `getKeys()` or the order of elements of `keyAltNames`).


source/client-side-encryption/client-side-encryption.rst
```diff
@@ -742,12 +766,19 @@ configuring auto encryption on a MongoClient, it is
...
-MongoClient.
+and managing data keys in the key vault collection.
 
 See `Why do we have a separate top level type for ClientEncryption?`_ and `Why do we need to pass a client to create a ClientEncryption?`_.
 
+When implementing behavior and error handling for key vault functions, Drivers
+SHOULD assume the presence of a unique index in the key vault collection on the
+``keyAltNames`` field with a partial index filter for only documents where
+``keyAltNames`` exists when implementing behavior of key management functions.
+Drivers MAY choose to not validate or enforce the existence of the unique index,
```
addaleax, line 777: Am I reading this correctly in that “MAY choose not to […] enforce the existence of the unique index” also means that they MAY create the unique index if it does not yet exist? Or is the idea that drivers would be allowed to error out if it doesn’t?

addaleax (approved):

source/client-side-encryption/client-side-encryption.rst
eramongodb: "MAY choose to not validate" means drivers are not required to assert the existence of the unique index. "MAY choose to not enforce" means drivers are not required to create the unique index if it doesn't exist. Drivers are permitted to do either of these if they so desire, but given that the reason for this clause is that it is not yet clear if/how the unique index should be created/mandated by drivers (per the "Why aren't we creating a unique index in the key vault collection?" rationale), I expect not many drivers will actually do so.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -708,20 +708,44 @@ ClientEncryption
...
 
       // An alias function equivalent to createKey.
-      createDataKey(kmsProvider: String, opts: Optional<DataKeyOpts>): Binary;
+      createDataKey(kmsProvider: String, opts: Optional<DataKeyOpts>): UUID;
 
       // Decrypts multiple data keys and (re-)encrypts them with a new masterKey, or with their current masterKey if a new one is not given.
       // Returns a RewrapManyDataKeyResult.
       rewrapManyDataKey(filter: Document, opts: Optional<RewrapManyDataKeyOpts>): RewrapManyDataKeyResult;
 
+      // Removes the key document with the given \_id from the key vault collection.
```
jmikola, line 720: Since this is appearing within a `.. code` block, is is necessary to escape `\_id` like that? This also applies to comment blocks above, which predate this particular PR.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -708,20 +708,44 @@ ClientEncryption
 
       // Creates a new key document and inserts into the key vault collection.
       // Returns the \_id of the created document as a UUID (BSON binary subtype 4).
-      createKey(kmsProvider: String, opts: Optional<DataKeyOpts>): Binary;
+      createKey(kmsProvider: String, opts: Optional<DataKeyOpts>): UUID;
```
jmikola, line 711: Since some languages have native UUID types, I think this should remain as Binary with a comment indicating that the type will (or must) be UUID with subtype 0x04.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -708,20 +708,44 @@ ClientEncryption
...
-      createDataKey(kmsProvider: String, opts: Optional<DataKeyOpts>): Binary;
+      createDataKey(kmsProvider: String, opts: Optional<DataKeyOpts>): UUID;
 
       // Decrypts multiple data keys and (re-)encrypts them with a new masterKey, or with their current masterKey if a new one is not given.
       // Returns a RewrapManyDataKeyResult.
       rewrapManyDataKey(filter: Document, opts: Optional<RewrapManyDataKeyOpts>): RewrapManyDataKeyResult;
 
+      // Removes the key document with the given \_id from the key vault collection.
+      // Returns the result of the internal deleteOne() operation on the key vault collection.
+      deleteKey(id: UUID): DeleteResult;
```
jmikola, line 722: In this case, are drivers permitted to be flexible in the argument type? If they support a native UUID type, can they accept that and internally prepare it as a Binary with subtype `0x04`? If not, I think this should also require a Binary argument, with a comment requiring the implementation to validate it as a subtype `0x04`.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -708,20 +708,44 @@ ClientEncryption
...
+
+      // Removes a keyAltName from the keyAltNames array of the key document in the key vault collection with the given \_id.
+      // Returns the previous version of the key document.
+      removeKeyAlternateName(id: UUID, keyAltName: String): Optional<Document>;
+
+      // Returns a key document in the key vault collection with the given \_id.
+      getKeyByAltName(keyAltName: String): Iterable<Document>;
+
       // Encrypts a BSONValue with a given key and algorithm.
       // Returns an encrypted value (BSON binary of subtype 6). The underlying implementation may return an error for prohibited BSON values.
```
jmikola, line 744: This predates the PR, but I think "MAY" should be capitalized since this seems to carry the weight of normal spec language.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -708,20 +708,44 @@ ClientEncryption
...
       encrypt(value: BSONValue, opts: EncryptOpts): Binary;
 
-      // Decrypts an encrypted value (BSON binary of subtype 6). Returns the original BSON value.
+      // Decrypts an encrypted value (BSON binary of subtype 6).
+      // Returns the original BSON value.
       decrypt(value: Binary): BSONValue;
```
jmikola, line 749: Is this the first use of "BSONValue" in a spec? Previous specs have used `any` (assuming that all language values have a BSON representation). I can imagine that the libmongoc API will use `bson_value_t`, but I'm curious about whether this will be clear for other languages.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -708,20 +708,44 @@ ClientEncryption
...
+
+      // Returns a key document in the key vault collection with the given \_id.
+      getKeyByAltName(keyAltName: String): Iterable<Document>;
+
       // Encrypts a BSONValue with a given key and algorithm.
       // Returns an encrypted value (BSON binary of subtype 6). The underlying implementation may return an error for prohibited BSON values.
       encrypt(value: BSONValue, opts: EncryptOpts): Binary;
 
-      // Decrypts an encrypted value (BSON binary of subtype 6). Returns the original BSON value.
+      // Decrypts an encrypted value (BSON binary of subtype 6).
```
jmikola, line 747: Are drivers expected to raise an error of the Binary parameter has the incorrect subtype?

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -1938,9 +1969,10 @@ little as possible.
 Why aren't we creating a unique index in the key vault collection?
 ------------------------------------------------------------------
 
-There should be a unique index on keyAltNames. Although GridFS
+There should be a unique index on the ``keyAltNames`` field with a partial index
+filter for only documents where ``keyAltNames`` exists. Although GridFS
 automatically creates indexes as a convenience upon first write, it has
-been problematic before. It requires the createIndex privilege, which a
+been problematic before due to requiring the ``createIndex`` privilege, which a
```
jmikola, line 1975: On the off chance you want to link this: https://www.mongodb.com/docs/manual/reference/privilege-actions/#mongodb-authaction-createIndex

No worries if not.

source/client-side-encryption/tests/README.rst
```diff
@@ -1779,3 +1779,59 @@ Use ``clientEncryption`` to encrypt the value "encrypted unindexed value" with t
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Setup
+`````
+
+1. Create a ``MongoClient`` object (referred to as ``client``).
+
+2. Using ``client``, drop the collection ``keyvault.datakeys``.
+
+3. Using ``client``, create a unique index on ``keyAltNames`` with a partial index filter for only documents where ``keyAltNames`` exists.
```
jmikola, line 1793: Should you mention a "majority" write concern here? It shows up in the command below.

source/client-side-encryption/tests/README.rst
```diff
@@ -1779,3 +1779,59 @@ Use ``clientEncryption`` to encrypt the value "encrypted unindexed value" with t
...
+4. Create a ``ClientEncryption`` object (referred to as ``client_encryption``) with ``client`` set as the ``keyVaultClient``.
+
+5. Using ``client_encryption``, create a data key with a ``local`` KMS provider and the keyAltName "def".
+
+Case 1: createKey()
+```````````````````
+
+1. Use ``client_encryption`` to create a new local data key with a keyAltName "abc" and assert the operation does not fail.
+
+2. Repeat Step 1 and assert the operation fails due to a duplicate key server error.
```
jmikola, line 1822: Do you want to specify [error code 11000](https://github.com/mongodb/mongo/blob/48d37ac38656aab14427a985653e221b6ed78550/src/mongo/base/error_codes.yml#L503)?

source/client-side-encryption/tests/unified/createKey.yml
```diff
@@ -131,3 +131,100 @@ tests:
...
+
+  - description: create datakey with single keyAltName
+    operations:
+      - name: createKey
+        object: *clientEncryption0
+        arguments:
+          kmsProvider: local
+          opts:
+            keyAltNames: ["local_key"]
+        expectResult: { $$type: binData }
```
jmikola, line 166: I'm not sure it'd be helpful/relevant here, but I just realized that there's no way to assert the subtype of a binary value. The server's [`$type`](https://www.mongodb.com/docs/manual/reference/operator/query/type/) operator doesn't support this, but I suppose we could in the Unified Test Format if there was a need. If so, feel free to log a ticket to throw in the backlog.

If so, you could always add some `# TODO` comments to these tests in places where you would have liked to use the functionality.

source/client-side-encryption/tests/unified/deleteKey.yml
```diff
@@ -0,0 +1,120 @@
...
+          - *local_key_doc
+
+  - description: delete existing AWS data key
+    operations:
+      - name: deleteKey
+        object: *clientEncryption0
+        arguments:
+          id: *aws_key_id
+        expectResult:
+          acknowledged: { $$unsetOrMatches: true }
```
jmikola, line 77: Is this necessary? If a write result was not acknowledged, all other fields would be empty. Therefore, it would not be possible to perform any assertion on `deletedCount`.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -708,20 +708,44 @@ ClientEncryption
...
+      // Adds a keyAltName to the keyAltNames array of the key document in the key vault collection with the given \_id.
+      // Returns the previous version of the key document.
+      addKeyAlternateName(id: UUID, keyAltName: String): Optional<Document>;
+
+      // Removes a keyAltName from the keyAltNames array of the key document in the key vault collection with the given \_id.
+      // Returns the previous version of the key document.
+      removeKeyAlternateName(id: UUID, keyAltName: String): Optional<Document>;
+
+      // Returns a key document in the key vault collection with the given \_id.
+      getKeyByAltName(keyAltName: String): Iterable<Document>;
```
jmikola, line 741: Can multiple keys share the same `keyAltName`? I assume not, as you're creating a partial, unique index on this field. In that case, wouldn't it make sense for this to return an `Optional<Document>`?

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -708,20 +708,44 @@ ClientEncryption
...
+      // Returns the result of the internal find() operation on the key vault collection.
+      getKeys(): Iterable<Document>;
+
+      // Adds a keyAltName to the keyAltNames array of the key document in the key vault collection with the given \_id.
+      // Returns the previous version of the key document.
+      addKeyAlternateName(id: UUID, keyAltName: String): Optional<Document>;
+
+      // Removes a keyAltName from the keyAltNames array of the key document in the key vault collection with the given \_id.
+      // Returns the previous version of the key document.
+      removeKeyAlternateName(id: UUID, keyAltName: String): Optional<Document>;
```
jmikola, line 738: Any reason these methods use "AlternateName" in the name while "AltName" is used by `getKeyByAltName`? Should they be consistent? I'm sure the common spec rules for naming deviations would permit drivers to use either, but it'd be nice for the spec example itself to be consistent.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1503,7 +1503,13 @@ Drivers MUST be running the mock `KMS KMIP server
...
 when evaluating tests that require KMS requests to a KMIP KMS provider.
 
-
+Drivers MAY enforce a unique index on ``keyAltNames`` as described in the
+`Client Side Field Level Encryption spec <../client-side-encryption/client-side-encryption.rst#why-aren-t-we-creating-a-unique-index-in-the-key-vault-collection>`_
+when running key management operations on the key vault collection. Although
+unified tests are written assuming the existence of the unique index, no unified
+test currently requires its implementation for correctness (e.g. no unified test
+currently attempts to create a data key with an existing keyAltName or add an
+existing keyAltName to a data key).
```
jmikola, line 1512: What is stopping you from doing so? We could theoretically expect a server-side error with code `11000`.

source/client-side-encryption/tests/unified/createKey.yml
eramongodb: Ticket created: https://jira.mongodb.org/browse/DRIVERS-2336

source/client-side-encryption/client-side-encryption.rst
jmikola: The `BsonValue` term in those specs seems to be the same thing that would apply here, so I'd suggest using that form of capitalization for consistency.

source/client-side-encryption/tests/unified/deleteKey.yml
jmikola: `acknowledged` itself is optional (see: [Write Results](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#write-results)). The reason for the [existing usages](https://github.com/mongodb/specifications/blob/7f37c9325b423223edf54d795afed444e6989b98/source/crud/tests/unified/deleteOne-hint-unacknowledged.yml#L66) is for asserting an unacknowledged write result. The CRUD spec permits methods to either return no result at all, or a document with an optional `acknowledged`. Also discussed in the CRUD spec: if an object has all optional fields, it may be optional itself. For that reason, we needed a convoluted way to handle an empty result or a possible `acknowledged: false` value. When `acknowledged` is `false`, "all other members of this result will be undefined."

In the case of an _acknowledged_ write result, there is really no reason to check for `acknowledged`. Other fields in the document will be accessible and can be asserted.

If you did want to assert `acknowledged`, then what you have here is correct since the field itself is optional. But I'm proposing we just remove the assertion for `acknowledged` altogether since an optional assertion for `true` is not done elsewhere (AFAIK) and is likely redundant.

eramongodb: Thank you for the explanation. Given the field is optional, I am now confident it is written correctly as-is, as it is not permitted to omit the fielddue to matcher rules forbidding extra keys. For example, omitting the field entirely yields the following error in the C driver:
```
error: BSON match failed: match error at '': key acknowledged is not present
Expected:  { "acknowledged" : true, "deletedCount" : { "$numberInt" : "0" } }
Actual:  { "deletedCount" : { "$numberInt" : "0" } }
```

source/client-side-encryption/client-side-encryption.rst
jmikola: > `Array<Document>` (for `getKeys()`)

I have no objection to using `Iterable<Document>` there, as there is precedent for doing so in other specs. Drivers are free to expose a cursor or return any iterable type.

I'm only focused on `getKeyByAltName`. This creates a more complicated API for any consumer of this method, since they need to iterate just to access a potential result. In the case of libmongoc, they might also have to free the cursor (in addition to anything they might copy out of it).

Consider the `findAndModify` methods in the CRUD spec. Implementation details aside (`findAndModify` doesn't use a cursor), consumers of those methods expect to either get a single document result or nothing at all. Returning an `Optional<Document>` here seems like something [users would most expect](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

eramongodb: I am convinced and in agreement. Thank you for pointing to `findAndModify` (i.e. `findOneAndUpdate`) as reference. I will change the return type of functions that return "zero or one" documents (both `getKey()` and `getKeyByAltName()` for consistency) from `Iterable<Document>` to `Optional<Document>` as suggested.

source/client-side-encryption/client-side-encryption.rst
jmikola: I don't think driver specs should forsake their own [consistency](https://github.com/mongodb/specifications/blob/master/README.rst#strive-to-be-idiomatic-but-favor-consistency) to comply with inconsistencies in other tools (assuming `mongosh` has different variations).

Drivers will already have the flexibility to use different names here, so this probably only directly affects how spec tests would be written. [CSFLE: User Facing API](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#user-facing-api) only talks about options, but [CRUD: Deviations](https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#id1) discusses methods as well.

It's not my intention to bikeshed, so I'm happy to defer to @kevinAlbs or another lead to make the call.

eramongodb: I am also in favor of renaming these to `addKeyAltName` and `removeKeyAltName` for consistency in Drivers spec. However, the new `createKey` function was added despite already having `createDataKey` solely for consistency with mongosh, which compelled me to side with consistency with mongosh. I would also like to defer on making this call.

source/client-side-encryption/client-side-encryption.rst
jmikola: If you want to allow flexibility (no objections there), I think that should be communicated in the comments. And if this is a general point, perhaps you can add a heading higher up explaining that language UUID types and Binary subtype `0x04` may be used interchangeably in APIs.

source/client-side-encryption/client-side-encryption.rst
eramongodb: Per design by contract, if a precondition is violated, the behavior is undefined. The spec currently requires nothing from implementations if the contract is violated. They may error (most likely), they may ignore the subtype and proceed (I hope not), they may terminate (imo most appropriate), or they may wipe the hard drive (in principle). I do not want to unnecessarily widen the contract if it is not necessary.

source/client-side-encryption/client-side-encryption.rst
jmikola: I didn't realize `createKey` was added purely for `mongosh`. That said, having an extra alias didn't trigger my spider sense as much as introducing two different naming conventions in the same PR :)

source/client-side-encryption/client-side-encryption.rst
eramongodb: I do not know if that is within the scope of this PR. That appears to be more a property of driver implementation of UUID per current UUID spec rather than something specific to the Key Management API. I will defer to the existing interface (`Binary` return value from `createDataKey`) for now, but such flexibility is _already_ permitted by the spec, then there should be no changes necessary here. If it is _not_ already permitted by the spec, I do not know if it is desirable to begin with. @kevinAlbs Thoughts?

source/client-side-encryption/tests/unified/deleteKey.yml
jmikola: Is your libmongoc branch up-to-date? Extra fields are permitted in root documents (see: [Unified Test Format spec](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#allowing-extra-fields-in-root-level-documents)). I recently made a bug fix related to this in [CDRIVER-4374](https://jira.mongodb.org/browse/CDRIVER-4374).

source/client-side-encryption/client-side-encryption.rst
addaleax: Fwiw, if this spec introduces names that deviate from the ones currently used in mongosh, we’ll probably want to add aliases for those in mongosh as well.

kevinAlbs (commented):
Looks great. I suggest adding command started expectations if it is feasible.

source/client-side-encryption/client-side-encryption.rst
addaleax: I think that’s okay, yes. If you do, the ticket would be appreciated :)

source/client-side-encryption/tests/unified/addKeyAlternateName.yml
```diff
@@ -0,0 +1,135 @@
...
+        arguments:
+          # First 3 letters of local_key_id replaced with 'A' (value: "#alkeylocalkey").
+          id: { $binary: { base64: AAAjYWxrZXlsb2NhbGtleQ==, subType: "04" } }
+          keyAltName: new_key_alt_name
+        expectResult: null
+    outcome:
+      - collectionName: *collection0Name
+        databaseName: *database0Name
+        documents:
+          - *local_key_doc
```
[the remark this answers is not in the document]
eramongodb: It depends. I will see what I can do.

source/client-side-encryption/client-side-encryption.rst
jmikola: This comment thread is marked as "Outdated", but it looks like @kevinAlbs' previous comment warrants some changes to permit drivers to use a native UUID type and Binary with subtype `0x04` interchangeably.

source/client-side-encryption/tests/README.rst
```diff
@@ -1779,3 +1779,60 @@ Use ``clientEncryption`` to encrypt the value "encrypted unindexed value" with t
 Store the result in ``payload``.
 
 Use ``clientEncryption`` to decrypt ``payload``. Assert the returned value equals "encrypted unindexed value".
+
+12. Unique Index on keyAltNames
```
jmikola, line 1783: "12. Explicit Encryption" precedes this prose test.

```suggestion
13. Unique Index on keyAltNames
```

jmikola (approved):
This is about as far as I can go with this PR.

One suggestion to rename the prose test number. I read through the RST changes and only the UUID/Binary issue that @kevinAlbs commented on stands out. I have not given the YAML tests another look.

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: Filed https://jira.mongodb.org/browse/MONGOSH-1239

source/client-side-encryption/tests/README.rst
```diff
@@ -1779,3 +1779,59 @@ Use ``clientEncryption`` to encrypt the value "encrypted unindexed value" with t
...
+The command should be equivalent to:
+
+.. code:: typescript
+
+   db.runCommand(
+     {
+        createIndexes: "datakeys",
+        indexes: [
+          {
+            key: "keyAltNames",
```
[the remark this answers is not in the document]
kevinAlbs: `key` also needs to be a document ` { "keyAltNames": 1 }`.

kevinAlbs (approved):
LGTM with a pending comment about the createIndexes command example.

eramongodb:
Added an explicit provision under "User facing API" that permits drivers to use a native UUID type in place of BSON Binary subtype 04 and includes a reference to the "Handling of Native UUID Types" spec.

PR #1232, opened 2022-05-26
Clarify how encryptedFields interacts with create and drop commands

jmikola:
Please complete the following before merging:
- [x] Bump spec version and last modified date.
- [x] Update changelog.

PR #1233, opened 2022-05-26
Add missing rationale for key management API and fix RST syntax

eramongodb:
### Description

This adds a block of rationale that was missing from https://github.com/mongodb/specifications/pull/1229.

As a drive-by fix, this PR also fixes the section title style that currently breaks HTML formatting on Github.

### RST Syntax Checking

Because this issue is not being detected by `rstcheck`, I used `rst2html`:

```
rst2html --halt=none client-side-encryption.rst > /dev/null
```

This yielded the following set of warnings:

~~~
client-side-encryption.rst:87: (ERROR/3) Unknown interpreted text role "ref".
client-side-encryption.rst:245: (ERROR/3) Unknown interpreted text role "ref".
client-side-encryption.rst:642: (ERROR/3) Unknown interpreted text role "ref".
client-side-encryption.rst:644: (ERROR/3) Unknown interpreted text role "ref".
client-side-encryption.rst:669: (SEVERE/4) Title level inconsistent:

Create Collection Helper
````````````````````````
client-side-encryption.rst:696: (SEVERE/4) Title level inconsistent:

Drop Collection Helper
``````````````````````
client-side-encryption.rst:1112: (ERROR/3) Unknown directive type "index".

.. index:: csfle
client-side-encryption.rst:1751: (ERROR/3) Unknown directive type "toctree".

.. toctree::
   :maxdepth: 2

   ./tests/README


client-side-encryption.rst:1382: (ERROR/3) Unknown target name: "why does clientencryption have key management functions when drivers can use existing crud operations instead?".
~~~

The warnings regarding "unknown interpreted text role" and "unknown directive type" appear to be due to Sphinx-specific directives, thus I have elected to ignore them.

Also, for future reference, using triple `~~~` allowed me to escape the RST code block above that would otherwise have been confused by the backticks in the section title style.

### DIFF
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -667,7 +667,7 @@ A collection supporting FLE 2 requires an index and three additional collections
 .. _drop: https://www.mongodb.com/docs/manual/reference/command/drop
 
 Create Collection Helper
-````````````````````````
+^^^^^^^^^^^^^^^^^^^^^^^^
 
 Drivers MUST support a BSON document option named ``encryptedFields`` for any
 `create`_ command helpers (e.g. ``Database.createCollection()``). This option
@@ -694,7 +694,7 @@ If the collection namespace has an associated ``encryptedFields``, then do the f
 - Create the the index ``{"__safeContent__": 1}`` on collection ``collectionName``.
 
 Drop Collection Helper
-``````````````````````
+^^^^^^^^^^^^^^^^^^^^^^
 
 Drivers MUST support a BSON document option named ``encryptedFields`` for any
 `drop`_ command helpers (e.g. ``Database.dropCollection()``,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by kevinAlbs — APPROVED** [review]
LGTM.

TIL about the ~~~ escape and rst2html.

File: source/client-side-encryption/client-side-encryption.rst [hunk 1]
```diff
@@ -694,7 +694,7 @@ If the collection namespace has an associated ``encryptedFields``, then do the f
 - Create the the index ``{"__safeContent__": 1}`` on collection ``collectionName``.
 
 Drop Collection Helper
-``````````````````````
+^^^^^^^^^^^^^^^^^^^^^^
```
jmikola, line 697: My mistake. I broke this in #1232 when looking at the previous sections for guidance. Backticks are the heading syntax used _after_ carets.

p-mongo: We made https://github.com/mongodb/mongoid/blob/master/docs/nesting-levels.txt in Mongoid to act as a quick reference for the correct heading syntax to use.

PR #1238, opened 2022-06-01
DRIVERS-1096: Split showRecordId/returnKey into separate test

jyemin:
Java driver is failing the `A successful find with options` test because it only include showRecordId and returnKey in the command if the value is true (the booleans are not nullable).  I switched up the tests a bit so that Java (and perhaps other drivers) will pass them.

DRIVERS-1096

- [ n/a] Bump spec version and last modified date.
- [ n/a] Update changelog.
- [ x] Make sure there are generated JSON files from the YAML test files.
- [ x] Test changes in at least one language driver.

### DIFF
--- a/source/command-monitoring/tests/unified/find.yml
+++ b/source/command-monitoring/tests/unified/find.yml
@@ -63,15 +63,14 @@ tests:
         object: *collection
         arguments:
           filter: { _id: { $gt: 1 } }
-          sort: { _id: 1 }
+          sort: { x: -1 }
+          projection: { _id: 0, x: 1 }
           skip: 2
           comment: "test"
           hint: { _id: 1 }
           max: { _id: 6 }
           maxTimeMS: 6000
           min: { _id: 0 }
-          returnKey: false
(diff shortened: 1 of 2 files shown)


source/command-monitoring/tests/unified/find.yml
```diff
@@ -63,31 +63,29 @@ tests:
         object: *collection
         arguments:
           filter: { _id: { $gt: 1 } }
-          sort: { _id: 1 }
+          sort: { x: -1 }
```
kmahar, line 66: are the sort and projection changes related to the original issue you ran into, or is it just to get some additional coverage?

kmahar (approved):
one question, but LGTM and the updated tests pass for both Swift and Rust.

source/command-monitoring/tests/unified/find.yml
jyemin: Unrelated.  There was no projection coverage (though really all of this should be in CRUD spec tests)

PR #1244, opened 2022-06-08
DRIVERS-555 Increase CSOT test timeouts

ShaneHarvey:
This change makes the tests more reliable in Python and Go.

- [ ] Update changelog.

Tested in python here: https://github.com/mongodb/mongo-python-driver/pull/954

### DIFF
--- a/source/client-side-operations-timeout/etc/templates/global-timeoutMS.yml.template
+++ b/source/client-side-operations-timeout/etc/templates/global-timeoutMS.yml.template
@@ -22,7 +22,7 @@ tests:
   # For each operation, we execute two tests:
   #
   # 1. timeoutMS can be configured to a non-zero value on a MongoClient and is inherited by the operation. Each test
-  # constructs a client entity with timeoutMS=50 and configures a fail point to block the operation for 60ms so
+  # constructs a client entity with timeoutMS=250 and configures a fail point to block the operation for 350ms so
   # execution results in a timeout error.
   #
   # 2. timeoutMS can be set to 0 for a MongoClient. Each test constructs a client entity with timeoutMS=0 and
@@ -39,7 +39,7 @@ tests:
             - client:
                 id: &client client
                 uriOptions:
-                  timeoutMS: 50
+                  timeoutMS: 250
                 useMultipleMongoses: false
                 observeEvents:
                   - commandStartedEvent
@@ -57,11 +57,11 @@ tests:
           client: *failPointClient
           failPoint:
             configureFailPoint: failCommand
-            mode: { times: 1 }
+            mode: { times: 2 }
             data:
               failCommands: ["{{operation.command_name}}"]
               blockConnection: true
-              blockTimeMS: 60
+              blockTimeMS: 350
       - name: {{operation.operation_name}}
         object: *{{operation.object}}
         {% if operation.arguments|length > 0 -%}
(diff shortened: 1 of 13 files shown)


kmahar (approved):
looks reasonable to me 👍

source/client-side-operations-timeout/etc/templates/global-timeoutMS.yml.template
```diff
@@ -57,11 +57,11 @@ tests:
           client: *failPointClient
           failPoint:
             configureFailPoint: failCommand
-            mode: { times: 1 }
+            mode: { times: 2 }
```
benjirewis, line 60: Why are we changing `times` for a lot of these fail points? Does it have to do with modifications to retryability you've made in Python? I haven't changed our retry logic yet in the Go driver, but the spec currently says to retry infinitely in the presence of `timeoutMS`; looks like maybe you've changed Python to retry twice.

ShaneHarvey: This was actually required for a different reason. In Python on Windows it's actually possible for the socket read (technically socket poll()) to raise a timeout 20ms _before_ the actual deadline. So a timeout gets raised but then the retry logic sees that there is 20ms remaining so a retry is attempted. Without this change the retry proceeds quickly and the operation actually succeeds within the timeout.

source/client-side-operations-timeout/tests/command-execution.yml
```diff
@@ -27,7 +27,7 @@ initialData:
 tests:
   - description: "maxTimeMS value in the command is less than timeoutMS"
     operations:
-      # Artificially increase the server RTT to ~10ms.
+      # Artificially increase the server RTT to ~20ms.
```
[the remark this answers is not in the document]
ShaneHarvey: > calculating 90th percentile RTT requires at least 10 RTT samples

If there's less than 10 samples, wouldn't the 90th percentile RTT be approximated by the max() or the 2nd max()? Using 0 when there's less than 10 samples is exactly the kind of bug that this test aims to find. 

In Python I'm currently using the average RTT (the same value used during server selection to satisfy localThresholdMS).

> Even if we use average RTT as our measurement, I'd still like to get at least 5 samples to prevent noisy samples on startup from artificially increasing RTT.

I don't agree. The worst that happens with an "artificially increased RTT" is that the operation gets rejected unnecessarily but the connection can be reused. The worst that happens with using 0 or min(RTT) is that the operation fails with a socket timeout and the connection needs to be closed. The second one is much worse since connection establishment is expensive.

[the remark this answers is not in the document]
benjirewis: > If there's less than 10 samples, wouldn't the 90th percentile RTT be approximated by the max() or the 2nd max()?

Discussing offline now, but I'd be in favor of changing Go to use average RTT for both `maxTimeMS` calculation and "short-circuiting". In that case, we wouldn't have a minimum sample requirement.

source/client-side-operations-timeout/etc/templates/global-timeoutMS.yml.template
benjirewis: Hmm ok. That does seem Python-specific; do you think other languages will have issues with socket-level early timeouts followed by a successful retry? I'm fine increasing `times` as it doesn't affect drivers that fail on the first hit to the fail point, but maybe we could leave a comment in the template explaining why it's necessary?

ShaneHarvey: Done.

benjirewis (approved):

PR #1249, opened 2022-06-14
DRIVERS-2348 add IndexKeyId

kevinAlbs:
# Summary
- Add `EncryptOpts.indexKeyId`.
- Add prose test "case 6: IndexKeyId works".

# Background & Motivation

The test "case 6: IndexKeyId works" is expected to fail if the IndexKeyID option is not passed to libmongocrypt with `mongocrypt_ctx_setopt_index_key_id`. libmongocrypt defaults the IndexKeyID to [the keyID set in `mongocrypt_ctx_setopt_key_id`](https://github.com/mongodb/libmongocrypt/blob/1ed14540db29fe52dd0ccc1208b949b8addac3fd/src/mongocrypt-ctx-encrypt.c#L1636-L1642). If the IndexKeyID is not overridden, the test will fail to find the inserted document.

A mistaken insert with an incorrect IndexKeyId results in incorrect query results and non-obvious errors. [SERVER-67263](https://jira.mongodb.org/browse/SERVER-67263) tracks a request to reject incorrect IndexKeyId from being inserted.

- [x] Update changelog.
- [ ] Make sure there are generated JSON files from the YAML test files. *Not applicable. Only prose tests were modified*.
- [x] Test changes in at least one language driver. *Tested in [Go](https://github.com/mongodb/mongo-go-driver/pull/986)*

### DIFF
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -11,7 +11,7 @@ Client Side Encryption
 :Type: Standards
 :Minimum Server Version: 4.2 (CSFLE), 6.0 (Queryable Encryption)
 :Last Modified: 2022-06-09
-:Version: 1.7.4
+:Version: 1.8.0
 
 .. _lmc-c-api: https://github.com/mongodb/libmongocrypt/blob/master/src/mongocrypt.h.in
 
@@ -944,6 +944,7 @@ EncryptOpts
 
    class EncryptOpts {
       keyId : Optional<Binary>
+      indexKeyId : Optional<Binary>
       keyAltName: Optional<String>
       algorithm: String,
       contentionFactor: Optional<Int64>,
@@ -956,6 +957,14 @@ identified by \_id or by alternate name. Exactly one is required.
 keyId
 ^^^^^
 Identifies a data key by \_id. The value is a UUID (binary subtype 4).
+
+indexKeyId
+^^^^^^^^^^
+Identifies a data key by \_id. The value is a UUID (binary subtype 4).
+It is an error to set indexKeyId when algorithm is not "Indexed".
+If indexKeyId is not set, it defaults to the value of keyId.
+The IndexKeyId must match the keyId set in encryptedFields for correctness.
+See `Why is IndexKeyId needed?`_.
 
 keyAltName
 ^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


source/client-side-encryption/client-side-encryption.rst
```diff
@@ -2302,6 +2311,15 @@ key vault collection, the key management functions may be overloaded to take an
...
 `Drivers Sessions Specification <../sessions/driver-sessions.rst>`__.
 
+Why is IndexKeyId needed?
+-------------------------
+
+Queryable Encryption supports two keys for an encrypted value: UserKey and IndexKey.
+- The UserKey encrypts user data. 
+- The IndexKey is used to derive tokens for indexing. It must match the "keyId" in encryptedFields for correct query results.
+
+Adding an EncryptOpts.indexKeyId enables multiple values with different UserKey to be inserted on the same field.
```
kevinAlbs, line 2321: @markbenvenuto is this rationale correct? If IndexKeyId is not necessary, I would like to leave it out of the API.

ShaneHarvey (approved):

source/client-side-encryption/client-side-encryption.rst
markbenvenuto: You could remove it IF you do the analysis on the client-side to find the indexKeyId from the encryptedFields in collection metadata. The server will have to do the analysis to check they are the same.

You could let the user leave it blank and then fill it in for the user based on encryptedFields.

kevinAlbs: > You could remove it IF you do the analysis on the client-side to find the indexKeyId from the encryptedFields in collection metadata. The server will have to do the analysis to check they are the same.

That may be non trivial. `ClientEncryption::Encrypt` does not know which field the payload will be applied to. To determine the field, the client would need to do query analysis.

Example with Go:

```go
// encryptedFields on server uses keyID1 on field "foo".
eo := options.Encrypt().
    SetAlgorithm("Indexed").
    SetKeyID(key2ID).
    SetIndexKeyID(key1ID)
// Encrypt() does not know which field insertPayload will be applied to.
insertPayload, err := clientEncryption.Encrypt(context.Background(), rawVal, eo)
assert.Nil(mt, err, "error on encrypt: %v", err)

coll := encryptedClient.Database("db").Collection("explicit_encryption")
_, err = coll.UpdateOne(context.Background(), bson.M{}, bson.M{"$set": bson.M{"encryptedIndexed": insertPayload}})
assert.Nil(mt, err, "error in UpdateOne: %v", err)
```

The default behavior of libmongocrypt is to assume IndexKeyId == UserKeyId. AFAICT adding an IndexKeyId option only enables the use-case for inserting data with IndexKeyId != UserKeyId. Is inserting with IndexKeyId != UserKeyId is a use-case we need to support?

addaleax (approved):

kevinAlbs:
This will not be merged. Please see DRIVERS-2348 for rationale.

PR #1250, opened 2022-06-14
DRIVERS-2356 Reduce expected removeKeyAltName operations to a single findOneAndUpdate

eramongodb:
- [x] ~Bump spec version and last modified date.~ N/A
- [x] ~Update changelog.~ N/A

## Description

This PR updates the unified spec tests for `removeKeyAltName()` to expect only a single `findOneAndUpdate()` operation that handles conditionally unsetting the `keyAltNames` field if it would otherwise be empty by using an aggregation pipeline rather than an additional and separate `updateOne()` operation. See DRIVERS-2356 for more information.

### DIFF
--- a/source/client-side-encryption/tests/unified/removeKeyAltName.json
+++ b/source/client-side-encryption/tests/unified/removeKeyAltName.json
@@ -118,11 +118,36 @@
                       }
                     }
                   },
-                  "update": {
-                    "$pull": {
-                      "keyAltNames": "does_not_exist"
+                  "update": [
+                    {
+                      "$set": {
+                        "keyAltNames": {
+                          "$cond": [
+                            {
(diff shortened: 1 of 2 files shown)


eramongodb:
@ShaneHarvey The [CSE Spec](https://github.com/mongodb/specifications/blob/ba44473e2488290e13acabbea0f32fa8c2bc4b7a/source/client-side-encryption/client-side-encryption.rst#clientencryption) does not describe implementation details regarding ClientEncryption methods. Those are expressed via the unified spec tests that implementations are expected to run and satisfy. This PR does not propose any change to the documented behavior of `removeKeyAltName()`, only the manner in which it is expected to be done.

ShaneHarvey (approved):

PR #1252, opened 2022-06-15
DRIVERS-2352 use a string for QueryType

kevinAlbs:
# Summary

- Change `EncryptOpts.QueryType` from an enum to a string.

# Background & Motivation

String values helps with forward compatibility When new values of QueryType and IndexType are added in libmongocrypt, users would only need to upgrade libmongocrypt, and not the driver, to use the new values. See [DRIVERS-2352](https://jira.mongodb.org/browse/DRIVERS-2352) for the motivation.

- [x] Update changelog.
- [ ] Make sure there are generated JSON files from the YAML test files. **Not applicable. No specification tests have changed**.
- [x] Test changes in at least one language driver. *Tested in [Go](https://github.com/mongodb/mongo-go-driver/pull/988/files)*

### DIFF
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -11,7 +11,7 @@ Client Side Encryption
 :Type: Standards
 :Minimum Server Version: 4.2 (CSFLE), 6.0 (Queryable Encryption)
 :Last Modified: 2022-06-09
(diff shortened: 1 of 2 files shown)


ShaneHarvey (approved):

vector-of-bool (approved):
LGTM

jmikola (approved):

PR #1262, opened 2022-06-24
DRIVERS-2371 add explicit encryption case 6 for default contention

kevinAlbs:
# Summary
- Add "explicit encryption" prose test "case 6".
- Use explicit contention factor of 0 in prose test "case 2" 

# Background & Motivation

The default contention factor value was raised to 4 on the server in [SERVER-66663](https://jira.mongodb.org/browse/SERVER-66663). [MONGOCRYPT-447](https://jira.mongodb.org/browse/MONGOCRYPT-447) updated the default contention factor for explicit encryption to match.

"case 6" tests the desired behavior when mixing automatic encryption and explicit encryption. For automatic encryption, default contention [comes from query analysis (mongocryptd or mongo_crypt)](https://github.com/mongodb/mongo/blob/3e3aae871cf2213d5288336f95f67c9df343a814/src/mongo/crypto/encryption_fields.idl#L66). For explicit encryption, default contention [comes from libmongocrypt](https://github.com/mongodb/libmongocrypt/blob/a6b33ea1434d13ec60772be9f8866763e6596fbd/src/mongocrypt-ctx-encrypt.c#L1650).

- [ ] Bump spec version and last modified date. **N/A. Test changes only.**
- [ ] Update changelog. **N/A. Test changes only.**
- [ ] Make sure there are generated JSON files from the YAML test files. **N/A. Prose test changes only.**
- [x] Test changes in at least one language driver. **Tested in [C here](https://github.com/mongodb/mongo-c-driver/pull/1049)**
- [ ] Test these changes against all server versions and topologies (including standalone, replica set, sharded clusters, and serverless). **C does not test CSFLE on sharded**


kevinAlbs:
No longer applicable.

PR #1263, opened 2022-06-24
DRIVERS-2329: remove optimization mention that no longer applies

nbbeeken:
Removing the mention of an optimization that doesn't apply because of DRIVERS-830

- [x] Update changelog.
- Make sure there are generated JSON files from the YAML test files.
- Test changes in at least one language driver.
- Test these changes against all server versions and topologies (including standalone, replica set, sharded clusters, and serverless).

### DIFF
--- a/source/server-discovery-and-monitoring/server-monitoring.rst
+++ b/source/server-discovery-and-monitoring/server-monitoring.rst
@@ -7,7 +7,7 @@ Server Monitoring
 :Status: Accepted
 :Type: Standards
 :Version: Same as the `Server Discovery And Monitoring`_ spec
-:Last Modified: 2022-02-24
+:Last Modified: 2022-06-24
 
 .. contents::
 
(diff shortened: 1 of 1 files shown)


source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -190,11 +190,6 @@ Each time a check completes, threads waiting for a `suitable`_ server
 are unblocked. Each unblocked thread MUST proceed if the new TopologyDescription
 now contains a suitable server.
 
-As an optimization, the client MAY leave threads blocked
-if a check completes without detecting any change besides
-roundTripTime: no operation that was blocked will
-be able to proceed anyway.
```
ShaneHarvey, line 196: This optimization could still be implemented, it would just need to ensure that even when the threads are blocked, an immediate check is still requested. That being said I'm fine with removing it entirely. I'm not sure any driver ever implemented this optimization.

ShaneHarvey (approved):

patrickfreed (approved):

PR #1264, opened 2022-06-25
Make KMS options more typescript-shaped

vector-of-bool:
- [ ] Update changelog.

### DIFF
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -10,8 +10,8 @@ Client Side Encryption
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: 4.2 (CSFLE), 6.0 (Queryable Encryption)
-:Last Modified: 2022-06-23
-:Version: 1.8.1
+:Last Modified: 2022-06-24
+:Version: 1.8.2
 
 .. _lmc-c-api: https://github.com/mongodb/libmongocrypt/blob/master/src/mongocrypt.h.in
 
@@ -2353,6 +2353,7 @@ Changelog
    :align: left
 
    Date, Description
+   24-06-22, Clean up kmsProviders to use more TypeScript-like type definitions.
    22-06-23, Make ``RewrapManyDataKeyResult.bulkWriteResult`` optional.
    22-06-16, Change ``QueryType`` to a string.
    22-06-15, Clarify description of date fields in key documents.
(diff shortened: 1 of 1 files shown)


source/client-side-encryption/client-side-encryption.rst
```diff
@@ -336,14 +336,17 @@ MongoClient Changes
...
    }
 
+.. _AutoEncryptionOpts:
+.. code-block:: typescript
+
    class AutoEncryptionOpts {
       keyVaultClient: Optional<MongoClient>;
       keyVaultNamespace: String;
-      kmsProviders: Map<String, Map<String, Value>>;
+      kmsProviders: KMSProvidersMap;
```
kevinAlbs, line 345:
```suggestion
      kmsProviders: KMSProviders;
```

Since `KMSProvidersMap` and `KMSProviderTLSOptions` represent a map from KMS provider name, suggest making the names consistent: `KMSProviders` and `KMSProvidersTLSOptions`?

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -336,14 +336,17 @@ MongoClient Changes
...
       schemaMap: Optional<Map<String, Document>>; // Maps namespace to a local schema
       bypassAutoEncryption: Optional<Boolean>; // Default false.
       extraOptions: Optional<Map<String, Value>>;
-      tlsOptions: Optional<Map<String, TLSOptions>>; // Maps KMS provider to TLS options.
+      tlsOptions?: KMSProviderTLSOptions; // Maps KMS provider to TLS options.
```
kevinAlbs, line 349:
```suggestion
      tlsOptions?: KMSProvidersTLSOptions; // Maps KMS provider to TLS options.
```

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -740,14 +761,14 @@ ClientEncryption
 
       // Creates a new key document and inserts into the key vault collection.
       // Returns the _id of the created document as a UUID (BSON binary subtype 0x04).
-      createKey(kmsProvider: String, opts: Optional<DataKeyOpts>): Binary;
+      createKey(kmsProvider: KMSProviderName, opts: DataKeyOpts | null): Binary;
```
kevinAlbs, line 764: `createKey` is expected to accept arbitrary strings for the `kmsProvider` for forward compatibility. If a KMS provider is added in the future, users may only need to upgrade libmongocrypt, not the driver, to use it.

Does using `KMSProviderName` as the type for `kmsProvider` suggest a driver may not allow arbitrary strings? If so, I suggest using `string` for `kmsProvider`.

kevinAlbs (commented):
Looks good. `KMSProviderName` may need to be changed to `string` to accept arbitrary strings.

source/client-side-encryption/client-side-encryption.rst
vector-of-bool: I've clarified the intention of `KMSProviderName`, which should hopefully avoid any possible confusion.

kevinAlbs (approved):

PR #1265, opened 2022-06-27
DRIVERS-2370 Remove createKey in favor of createDataKey

eramongodb:
- [x] Update changelog.

## Description

This PR addresses DRIVERS-2370. All instances of `createKey()` have either been removed or renamed to `createDataKey()`.

Upon reconsideration, it was decided that the engineering effort, subsequent maintenance, and clarity of usage and documentation of adding to all Drivers a new `createKey()` function that is merely an alias to the existing `createDataKey()` function may not be desirable. It was deemed preferable for mongosh, whose interface the Key Management API is aiming to reach parity with, to add `createDataKey()` for parity instead.

Other factors considered was leaving room for further key types to be managed in the key vault collection. Instead of reserving the generic `createKey` for "data keys", `createDataKey()` leaves room for additional `createXKey()` without semantic conflicts. Other Key Management API functions such as `deleteKey()` and `addKeyAltName()` do not require qualifications as their operations may apply regardless of key type (by UUID or by key alternate name).

## CSE Version Number

This PR proposes bumping the CSE spec version to 1.9.0 from 1.8.1 despite removing a public interface entity, which would normally constitute a major version bump per SemVer. This is because this change is intended to be applied as part of initial Key Management API implementation and no Driver has, or is expected to have, publically released a version with the Key Management API yet. If the version number _should_ instead be bumped to 2.0.0 regardless, please indicate so in the review.

### DIFF
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -2349,7 +2349,7 @@ Changelog
 
    Date, Description
    22-06-27, Remove ``createKey``.
-   24-06-22, Clean up kmsProviders to use more TypeScript-like type definitions.
+   22-06-24, Clean up kmsProviders to use more TypeScript-like type definitions.
    22-06-23, Make ``RewrapManyDataKeyResult.bulkWriteResult`` optional.
    22-06-16, Change ``QueryType`` to a string.
    22-06-15, Clarify description of date fields in key documents.
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -10,8 +10,8 @@ Client Side Encryption
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: 4.2 (CSFLE), 6.0 (Queryable Encryption)
-:Last Modified: 2022-06-23
-:Version: 1.8.1
+:Last Modified: 2022-06-27
+:Version: 1.9.0
 
 .. _lmc-c-api: https://github.com/mongodb/libmongocrypt/blob/master/src/mongocrypt.h.in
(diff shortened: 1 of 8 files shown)


kevinAlbs (approved):
LGTM, thank you for the quick turnaround.

ShaneHarvey (approved):
LGTM!

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -10,8 +10,8 @@ Client Side Encryption
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: 4.2 (CSFLE), 6.0 (Queryable Encryption)
-:Last Modified: 2022-06-23
-:Version: 1.8.1
+:Last Modified: 2022-06-27
+:Version: 1.9.0
```
jmikola, line 14: No objections to a minor bump here (per your explanation in the OP), but that's more because I personally don't think these versions serve much purpose.

source/client-side-encryption/tests/README.rst
```diff
@@ -1817,7 +1819,7 @@ The command should be equivalent to:
 
 5. Using ``client_encryption``, create a data key with a ``local`` KMS provider and the keyAltName "def".
 
-Case 1: createKey()
+Case 1: createDataKey()
```
jmikola, line 1822: Don't forget to keep the subsequent line consistent in length.

jmikola (approved):
LGTM with RST title length corrected.

source/client-side-encryption/tests/README.rst
eramongodb: _grumble grumble_ RST _grumble grumble_

PR #1270, opened 2022-07-07
Updated CSE tests where there is no expected result

rozza:
DRIVERS-2382

- [ ] Update changelog.


rozza:
I'm not sure this change requires a bump in the spec version / updating the change log.

eramongodb (approved):

PR #1277, opened 2022-07-22
DRIVERS-2395: Add test that read is not retried in a transaction

jyemin:
- [N/A] Bump spec version and last modified date.
- [N/A] Update changelog.
- [Done] Make sure there are generated JSON files from the YAML test files.
- [Done] Test changes in at least one language driver.
- [Done] Test these changes against all server versions and topologies (including standalone, replica set, sharded clusters, and serverless).

Though the requirement is in the retryable reads spec, the test seemed more appropriate in the transactions spec.  Happy to move if there is strong disagreement.

DRIVERS-2395

### DIFF
--- a/source/transactions/tests/unified/do-not-retry-read-in-transaction.json
+++ b/source/transactions/tests/unified/do-not-retry-read-in-transaction.json
@@ -86,7 +86,10 @@
             "session": "session0"
           },
           "expectError": {
-            "isError": true
+            "isError": true,
+            "errorLabelsContain": [
+              "TransientTransactionError"
+            ]
           }
         }
       ],
--- a/source/transactions/tests/unified/do-not-retry-read-in-transaction.json
+++ b/source/transactions/tests/unified/do-not-retry-read-in-transaction.json
@@ -3,9 +3,15 @@
   "schemaVersion": "1.4",
   "runOnRequirements": [
     {
-      "minServerVersion": "4.0",
+      "minServerVersion": "4.0.0",
       "topologies": [
-        "replicaset",
+        "single",
+        "replicaset"
+      ]
+    },
+    {
+      "minServerVersion": "4.2.0",
+      "topologies": [
         "sharded",
         "load-balanced"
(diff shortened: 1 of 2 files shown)


jyemin:
Huh, this is failing on 4.0 variants with 

```
{"ok": 0.0, "errmsg": "failCommand not found", "code": 266, "codeName": "FailPointSetFailed"
```

I'm not sure why, it's passing for me locally on a 4.0.18 replica set.

ShaneHarvey:
Should we add a test for reads too or does that already exist?

jyemin:
I removed [this check](https://github.com/mongodb/mongo-java-driver/blob/5c903986d5f5330dd390f1c354e7c746a2337142/driver-core/src/main/com/mongodb/internal/operation/OperationHelper.java#L413-L415) and a bunch of tests fail, including [this one](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/legacy/retryable-writes.json#L254).  So I think we're covered for writes.

jyemin:
Added errorLabelsContain assertion to match what's in the corresponding test for writes.

ShaneHarvey (approved):

source/transactions/tests/unified/do-not-retry-read-in-transaction.json
```diff
@@ -0,0 +1,116 @@
+{
+  "description": "do not retry read in a transaction",
+  "schemaVersion": "1.4",
+  "runOnRequirements": [
+    {
+      "minServerVersion": "4.0.0",
+      "topologies": [
+        "single",
```
jyemin, line 8: I errantly added this topology, and didn't wait for tests to complete on Evergreen.  Will remove in subsequent commit.

jmikola: https://github.com/mongodb/specifications/commit/bafb024be8202b1676a3c6632e8a3eb9cf36209b

PR #1280, opened 2022-07-25
DRIVERS-2398 add mongos limiting note

prestonvasquez:
DRIVERS-2398

- [ ] Update changelog.

### DIFF
--- a/source/sessions/tests/README.rst
+++ b/source/sessions/tests/README.rst
@@ -178,7 +178,7 @@ Skip this test if your driver does not allow forking.
 14. Implicit sessions only allocate their server session after a successful connection checkout
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-* Create a MongoClient with the following options: ``maxPoolSize=1`` and ``retryWrites=true``
+* Create a MongoClient with the following options: ``maxPoolSize=1`` and ``retryWrites=true``. If testing against a sharded deployment, be sure to connect to only a single mongos.
 * Attach a command started listener that collects each command's lsid
 * Initiate the following concurrent operations
 
--- a/source/sessions/tests/README.rst
+++ b/source/sessions/tests/README.rst
@@ -19,7 +19,7 @@ by each driver.
 
 Snapshot session tests
 ======================
-Snapshot sessions tests require server of version 5.0 or higher and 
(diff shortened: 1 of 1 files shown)


source/sessions/tests/README.rst
```diff
@@ -178,7 +178,7 @@ Skip this test if your driver does not allow forking.
 14. Implicit sessions only allocate their server session after a successful connection checkout
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-* Create a MongoClient with the following options: ``maxPoolSize=1`` and ``retryWrites=true``
+* Create a MongoClient with the following options: ``maxPoolSize=1`` and ``retryWrites=true``. If testing against a sharded deployment, be sure to connect to only a single mongos.
```
matthewdale, line 181: Is this possibly related to the need to run the test 5 times to be able to assert that only 1 session is used? If we're limiting the number of `mongos` nodes to 1, are there still cases where more than 1 session is used to run the concurrent operations in the test?

matthewdale, line 181: Should we use the MUST keyword to make it clear that this is a requirement?

For example:
> If testing against a sharded deployment, the test runner MUST ensure that the MongoClient connects to only a single mongos host.

(wording borrowed from the description of `useMultipleMongoses` in the [Unified Test Format](https://github.com/mongodb/specifications/blob/0e2f8f93c9ae2b3ce7227db1d3a370bd9246ebb6/source/unified-test-format/unified-test-format.rst) spec)

prestonvasquez: IIUC it is still possible to have more than one session checked out because part of the requirement here is to ensure that we end the implicit sessions _after_ we close the connection. The order of operations looks like this:

1. make connection
2. check out implicit session
3. close connection
4. end implicit session

For concurrent jobs I believe there is always a race condition between steps 2 and 4. In the Go driver, we can dependably illustrate this by creating some kind of delay like `time.Sleep(1)` in the `(session.Client).EndSession` method.

prestonvasquez: Yes, I agree with this.

ShaneHarvey: That's right, we still need to ru-run the test even with one mongos.

matthewdale: Sounds good, thanks for the explanation 👍

matthewdale (approved):
Looks good 👍

PR #1284, opened 2022-07-29
Error if RewrapManyDataKey is called with libmongocrypt < 1.5.2

kevinAlbs:
# Summary
- Add documentation requirement to alert users not to use RewrapManyDataKey with libmongocrypt < 1.5.2.
- Recommend drivers return an error if RewrapManyDataKey is called and libmongocrypt version is < 1.5.2.

# Background & Motivation

Returning an error is intended to limit risk of users hitting the bug of MONGOCRYPT-464.

- [x] Update changelog.
- [ ] Make sure there are generated JSON files from the YAML test files. **N/A**
- [x] Test changes in at least one language driver. **Tested in [Go](https://github.com/mongodb/mongo-go-driver/pull/1037)**


kevinAlbs:
Closing. Superseded by other PRs.

PR #1285, opened 2022-07-29
DRIVERS-2404 Fix Azure and KMIP key material for RewrapManyDataKey

eramongodb:
- [x] ~Bump spec version and last modified date.~ N/A
- [x] ~Update changelog.~ N/A
- [x] Test changes in [at least one language driver](https://github.com/mongodb/mongo-c-driver/pull/1079).

## Description

Resolves DRIVERS-2404.

As a drive-by fix, also updates the descriptions of some CSE tests to be consistent with their filenames.


kevinAlbs (approved):
Confirmed this passes locally with the Go driver using libmongocrypt built with MONGOCRYPT-464.

PR #1289, opened 2022-08-01
Remove 10 minute cursorTimeout comment from GridFSFindOptions docs

sanav33:
The cursor timeout value is now configurable on a per server basis: https://jira.mongodb.org/browse/SERVER-8188. Additionally, the spec did not mark the `skip` option in `GridFSFindOptions` as `optional`.

### DIFF
--- a/source/gridfs/gridfs-spec.rst
+++ b/source/gridfs/gridfs-spec.rst
@@ -775,7 +775,7 @@ Generic Find on Files Collection
     maxTimeMS: Int64 optional;
 
     /**
-     * The server normally times out idle cursors after an inactivity period (10 minutes)
+     * The server normally times out idle cursors after an inactivity period
      * to prevent excess memory use. Set this option to prevent that.
      */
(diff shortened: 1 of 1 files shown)


ShaneHarvey:
Changes LGTM but I believe noCursorTimeout=True doesn't actually do anything at all on 4.4.8+ servers due to the session timeout behavior described in https://jira.mongodb.org/browse/DRIVERS-1602 and https://jira.mongodb.org/browse/SERVER-6036.

alcaeus (approved):
LGTM. As @ShaneHarvey commented, the session timeout (which comes into effect for operations outside a session due to the drivers starting an implicit session) overrides the `noCursorTimeout` option, again creating an implicit timeout for cursors.

kmahar:
Thanks for the extra context @ShaneHarvey and @alcaeus. I agree we could probably skip this option then @sanav33, we've got very few users on MongoDB versions < 5.0

kmahar (approved):

PR #1291, opened 2022-08-16
[DRIVERS-2411] Add behavior for automatic Azure KMS credentials

vector-of-bool:
Refer: DRIVERS-2411

- [x] Update changelog.
- [ ] Test changes in at least one language driver. (TODO)

### DIFF
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -326,6 +326,8 @@ specified as a BSON binary with subtype 0x04 as described in
 MongoClient Changes
 -------------------
 
+.. _MongoClient:
+
 .. code:: typescript
 
    class MongoClient {
@@ -337,6 +339,10 @@ MongoClient Changes
       private MongoClient keyvault_client; // Client used to run find on the key vault collection. This is either an external MongoClient, the parent MongoClient, or internal_client.
       private MongoClient metadata_client; // Client used to run listCollections. This is either the parent MongoClient or internal_client.
       private Optional<MongoClient> internal_client; // An internal MongoClient. Created if no external keyVaultClient was set, or if a metadataClient is needed
+
+      // Exposition-only, used for caching automatic Azure credentials
+      private cachedAzureAccessToken?: AzureAccessToken;
+      private azureAccessTokenExpireTime?: PointInTime;
    }
 
 .. _AutoEncryptionOpts:
@@ -577,21 +583,22 @@ following process:
 
 5. If `K` contains an ``azure`` property, and that property is an empty map:
 
-   1. Attempt to obtain the **default** Azure VM Managed Identity Access Token
-      `T` as detailed in `Obtaining an Access Token for Azure Key Vault`_.
-   2. If a token `T` was obtained successfully, create a new AzureAccessToken_
-      object with `T` as the ``accessToken`` property. Insert that
-      AzureAccessToken_ object into `P` as the ``azure`` property.
+   1. If the current MongoClient_ has a ``cachedAzureAccessToken`` AND the
+      duration until ``azureAccessTokenExpireTime`` is greater than one minute,
+      insert ``cachedAzureAccessToken`` as the ``azure`` property on `P`.
+   2. Otherwise:
 
-6. If `K` contains an ``azure`` property, and that property is an empty map:
+      1. Let `t_0` be the current time.
+      2. Attempt to obtain an Azure VM Managed Identity Access Token `T` as
+         detailed in `Obtaining an Access Token for Azure Key Vault`_.
+      3. If a token `T` with expire duration `d_{exp}` were obtained
+         successfully, create a new AzureAccessToken_ object with `T` as the
+         ``accessToken`` property. Insert that AzureAccessToken_ object into `P`
+         as the ``azure`` property. Record the generated AzureAccessToken_ in
+         ``cachedAzureAccessToken``. Record the ``azureAccessTokenExpireTime``
+         as `t_0 + d_{exp}`.
 
-   1. Attempt to obtain an Azure VM Managed Identity Access Token `T`, as
-      detailed in `Obtaining an Access Token for Azure Key Vault`_.
-   2. If a token `T` was obtained successfully, create a new AzureAccessToken_
-      object with `T` as the ``accessToken`` property. Insert that
-      AzureAccessToken_ object into `P` as the ``azure`` property.
-
-7. Return `P` as the additional KMS providers to libmongocrypt_.
+6. Return `P` as the additional KMS providers to libmongocrypt_.
 
 __ ../auth/auth.html#obtaining-credentials
 
@@ -652,15 +659,17 @@ The below steps should be taken:
 7.  If `Resp_{status} ≠ 200`, obtaining the access token has failed, and the
     HTTP response body of `Resp` encodes information about the error that
     occurred. Return an error instead of an access token.
-8.  Otherwise, let ``J`` be the JSON document encoded in the HTTP response body
+8.  Otherwise, let `J` be the JSON document encoded in the HTTP response body
     of `Resp`.
 9.  The result access token `T` is given as the ``access_token`` string property
-    of ``J``. Return `T` as the resulting access token.
+    of `J`. Return `T` as the resulting access token.
+10. The resulting "expires in" duration `d_{exp}` is a count of seconds given as an
+    ASCII-encoded integer string ``expires_in`` property of `J`.
 
 .. note::
 
    If JSON decoding of `Resp` fails, or the ``access_token`` property is absent
-   from ``J``, this is a protocol error from IMDS. Indicate this error to the
+   from `J`, this is a protocol error from IMDS. Indicate this error to the
    requester of the access token.
 
 .. note::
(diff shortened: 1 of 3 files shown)


source/client-side-encryption/client-side-encryption.rst
```diff
@@ -597,6 +631,66 @@ message.
...
 
+
+Obtaining an Access Token for Azure Key Vault
+`````````````````````````````````````````````
+
+.. versionadded:: 1.12.0 2022/08/16
+
+Virtual machines running on the Azure platform have one or more *Managed
+Identities* associated with them. From within the VM, an identity can be used by
+obtaining a access token via HTTP from the *Azure Instance Metadata Service*
```
kevinAlbs, line 642:
```suggestion
obtaining an access token via HTTP from the *Azure Instance Metadata Service*
```

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -597,6 +631,66 @@ message.
...
+
+__ https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http
+
+.. default-role:: math
+
+If an Azure VM has more than one managed identity, requesting an access token
+requires additional query parameters |--| `objectId`, `clientId`, and `miResId`
+|--| that disambiguate the request. We'll refer to these parameters as `M`. If
+the VM only has one managed identity, `M` may be omitted.
+
```
kevinAlbs, line 653: Should the driver cache the returned `access_token`?

https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service?tabs=windows#rate-limiting suggests there are rate limits:

> In general, requests to IMDS are limited to 5 requests per second (on a per VM basis)

This was a problem for AWS and is being resolved in https://jira.mongodb.org/browse/DRIVERS-2333.

This was not a problem for GCP, since the GCP Metadata Server does not have a rate limit. The GCP Metadata Server caches access tokens. See: https://docs.google.com/document/d/1RTDp5QMg_ayYnR_T7S9SriE19doNALwqIHNxlWerFeE/edit#heading=h.z9wdvrobyao9.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -597,6 +631,66 @@ message.
...
+5. Prepare an HTTP GET request `Req` based on `U`.
+
+   .. note:: All query parameters on `U` should be appropriately percent-encoded
+
+6. Add an HTTP header ``Metadata: true`` to `Req`.
+
+7. Let `T` be a duration initialized to `0sec`.
+8. Issue `Req` to the Azure IMDS server ``168.254.169.254:80``. Let `Resp` be
+   the response from the server.
+9. If the `Resp_{status} >= 500`, wait for one second, then go to step ``8``.
```
kevinAlbs, line 673: Why add the retry? I prefer removing the retry to simplify the implementation. It may also be difficult to test.

If the retry is needed, this may need to account for `timeoutMS`. `timeoutMS` is in the [new Client-Side Operation Timeout specification](https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.rst). `timeoutMS` caps the duration of a single driver operation. E.g. `coll.insert_one({x:1}, timeout_ms=1000)` must return within 1000ms.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -597,6 +631,66 @@ message.
...
+10. If `T < 30sec` AND `Resp_{status} = 404` OR `Resp_{status} = 429`, wait for
+    the duration specified by `T`, then set `T = (T * 2) + 2sec`, then go to
+    step ``8``.
+11. If `Resp_{status} ≠ 200`, obtaining the access token has failed, and the
+    HTTP response body of `Resp` encodes information about the error that
+    occurred. Return an error instead of an access token.
+12. Otherwise, let ``J`` be the JSON document encoded in the HTTP response body
+    of `Resp`.
+13. The result access token `T` is given as the ``access_token`` string property
+    of ``J``. Return `T` as the resulting access token.
```
kevinAlbs, line 683:
```suggestion
13. The result access token `Token` is given as the ``access_token`` string property
    of ``J``. Return `Token` as the resulting access token.
```

To avoid redeclaring `T`.

kevinAlbs (commented):
Looks pretty good. Caching the access token may be a requirement. Consider removing the retry.

source/client-side-encryption/client-side-encryption.rst
vector-of-bool: The retry was added only at the recommendation of the Azure documentation, which hints at spurious failure. If it's not actually that likely to fail, we're probably fine to just drop it.

source/client-side-encryption/client-side-encryption.rst
vector-of-bool: Is request for the KMS credentials really that frequent? I honestly don't know how often it is needed.

The token response includes expiry information that can be useful to maintain a cache of the token, if necessary.

kevinAlbs: Potentially. libmongocrypt will cache the decrypted Data Encryption Key for 1 minute. So if (automatic|explicit) (encryption|decryption) has all keys cached, the `MONGOCRYPT_CTX_NEED_KMS_CREDENTIALS` state is not expected to be entered.

Other operations always need KMS. For example: [creating a data key](https://github.com/mongodb/libmongocrypt/blob/db6d3ab9a7f7d5ef20a95341a4007196c454b705/src/mongocrypt-ctx-datakey.c#L506) or rewrapping data keys.

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: If this is an Azure recommendation, I am in favor of keeping the retry logic. Consider adding a link to the Azure documentation, since other reviewers may be interested in the motivation.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -597,6 +631,66 @@ message.
...
+obtaining a access token via HTTP from the *Azure Instance Metadata Service*
+(IMDS). `See this documentation for more information`__
+
+__ https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http
+
+.. default-role:: math
+
+If an Azure VM has more than one managed identity, requesting an access token
+requires additional query parameters |--| `objectId`, `clientId`, and `miResId`
+|--| that disambiguate the request. We'll refer to these parameters as `M`. If
```
kevinAlbs, line 651: I am unsure if more than one managed identity is a common use-case we need to support. I have asked in [DRIVERS-1780](https://jira.mongodb.org/browse/DRIVERS-1780?focusedCommentId=4761705&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4761705)

kevinAlbs: After offline discussion with produce management, there are no known requests for supporting multiple user-assigned managed identities. We can proceed with not supporting multiple user-assigned managed identities.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -530,51 +539,68 @@ Automatic Credentials
...
-      GCPKMSOptions_ map from :math:`C` and insert that map onto :math:`P` as
-      the ``gcp`` property.
+   1. Attempt to obtain the **default** Azure VM Managed Identity Access Token
+      `T` as detailed in `Obtaining an Access Token for Azure Key Vault`_.
+   2. If a token `T` was obtained successfully, create a new AzureAccessToken_
+      object with `T` as the ``accessToken`` property. Insert that
+      AzureAccessToken_ object into `P` as the ``azure`` property.
 
-4. Return :math:`P` as the additional KMS providers to libmongocrypt_.
+6. If `K` contains an ``azure`` property, and that property contains a
```
kevinAlbs, line 586: Step 6 can be removed.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -597,6 +623,59 @@ message.
...
+   If JSON decoding of `Resp` fails, or the ``access_token`` property is absent
+   from ``J``, this is a protocol error from IMDS. Indicate this error to the
+   requester of the access token.
+
+.. note::
+
+   If an Azure VM has more than one managed identity, requesting an access token
+   requires additional query parameters to disambiguate the request. For
+   simplicity, these parameters are omitted, and only VMs that have a single
+   managed identity are support.
```
kevinAlbs, line 674:
```suggestion
   managed identity are supported.
```

kevinAlbs (commented):
Nice, changes look good with the removal of one step and additional behavior to cache the token.

To confirm: are test changes going to be included in a later PR?

I will ask if others team members are planning to implement this soon to get more eyes on the review.

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: The retry was removed to simplify implementation and testing. It is not known if a retry is needed.

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: I forgot about this. If a user receives an error attempting to create > 5 data keys in one second, I think this caching the token in the driver may be necessary.

vector-of-bool:
@kevinAlbs Added instructions about token caching.

blink1073 (commented):
LGTM!

PR #1293, opened 2022-08-18
DRIVERS-2417: Exclude create events from change stream when testing showExpandedEvents

alcaeus:
The test in question sometimes fails for the PHP driver, due to the `create` event being emitted in the change stream. Since the test only checks the first event, we have to exclude the `create` event to ensure we only receive the `createIndexes` event we're looking for. This fixes the issue for PHP and shouldn't introduce problems for other drivers.

Note: after discussion with @jmikola we decided to only exclude the problematic `create` event from the change stream. This was done so that any other additional commands sent to the server before the `createIndex` call would cause the test to fail. This may be very unlikely to happen, but we figured it was more sensible than only including `createIndexes` in the change stream.

- [ ] Bump spec version and last modified date. (N/A)
- [ ] Update changelog. (N/A)

### DIFF
--- a/source/change-streams/tests/unified/change-streams-showExpandedEvents.json
+++ b/source/change-streams/tests/unified/change-streams-showExpandedEvents.json
@@ -275,7 +275,15 @@
           "name": "createChangeStream",
           "object": "collection0",
           "arguments": {
-            "pipeline": [],
+            "pipeline": [
+              {
+                "$match": {
+                  "operationType": {
+                    "$ne": "create"
+                  }
+                }
+              }
+            ],
             "showExpandedEvents": true
           },
           "saveResultAsEntity": "changeStream0"
--- a/source/change-streams/tests/unified/change-streams-showExpandedEvents.yml
+++ b/source/change-streams/tests/unified/change-streams-showExpandedEvents.yml
@@ -160,7 +160,15 @@ tests:
       - name: createChangeStream
         object: *collection0
         arguments:
-          pipeline: []
+          pipeline:
+            # On sharded clusters, the create command run when loading initial
+            # data sometimes is still reported in the change stream. To avoid
+            # this, we exclude the create command when creating the change
+            # stream, but specifically don't exclude other events to still catch
+            # driver errors.
+            - $match:
+                operationType:
+                  $ne: create
           showExpandedEvents: true
         saveResultAsEntity: &changeStream0 changeStream0
       - name: createIndex


ShaneHarvey:
IIUC shared clusters should return the same events as replica sets. Is this a server bug?

alcaeus:
@ShaneHarvey that is technically true, but I'm not convinced it can be considered a server bug. For the record, I couldn't get this test to break when using a connection string with a single mongos. While the test does specify `useMultipleMongoses: false`, the internal client used to prepare initial data gets the full connection string as it may need to send commands to each mongos (e.g. `killAllSessions`). This means that while the `createIndexes` command will be sent to the same mongos where we also create the change stream, the collection may have been created on a different mongos as the internal client chooses one at random from the connection string. This may be responsible for the `create` change stream event being available on other mongoses only after we create the change stream.

Apart from ignoring the `create` command, another workaround would be to not rely on the internal client to create the collection in question (through the use of `initialData`), but to issue a `create` command as part of the test, ensuring that it gets sent to the same mongos that we then create the change stream on. Either option works fine, but I only came up with the latter one after digging deeper to be able to blame our test runner 😉

benjirewis:
I have never seen this test or other `showExpandedEvents` tests fail in the Go driver due to extra `create` events coming from the internal client. Your explanation does make sense, but it's just odd this doesn't happen in the Go driver given that we also use an internal client with the full connection string.

I think I'm ok with this change given that it shouldn't _cause_ problems for other drivers, but should you also add the `$match` to other `createChangeStream` pipelines in the test file?

ShaneHarvey:
Thanks for the explanation. I agree it's not really a server bug. Another alternative might be to use startAtOperationTime or a causally consistent session to start the change stream _after_ the create. Either way could you mention the mongos initialData issue in the comment?

alcaeus:
> Another alternative might be to use startAtOperationTime [...]

This was my first attempt at fixing it just for the PHP driver, and it fixed the test. However, it would've required significant changes for our test runner to exclude the `startAtOperationTime` from command assertions (as all other spec tests started failing). Alternatively, `startAtOperationTime` could be added to the spec test itself, but I'm not sure how we'd go about this as we'd have to read the operation time from the `create` command and use that as the `startAtOperationTime`.

> Either way could you mention the mongos initialData issue in the comment?

I updated the comment to specify that the `create` command is triggered by loading initial data.

benjirewis (approved):
This seems fine to me with the added comment; thanks @alcaeus. 

I'm still not quite sure why we wouldn't have to add this `$match` stage to any other test that runs `createChangeStream` with `showExpandedEvents` set to `true` and expects a certain order of events, but I'm assuming you just aren't seeing failures in other tests in the PHP driver?

jmikola:
> I'm assuming you just aren't seeing failures in other tests in the PHP driver?

Correct. The failure was only observed in this particular test.

If there are no objections I'm happy to add `$match: { operationType: { $ne: create }}` to all of the tests in this file. That could avoid some churn if we run into another failure down the line.

benjirewis:
Adding that stage to every test sounds like a good idea to me; thanks @jmikola .

jmikola:
@benjirewis: On second thought, I think I understand this change was only required in a single test.

"when showExpandedEvents is true, createIndex events are reported" is the only test where the first operation is `createChangeStream` _and_ `iterateUntilDocumentOrError` is used to expect an operation executed _after_ the change stream is created.

The first two tests in this file use only a single `createChangeStream` operation with an APM assertion on the outgoing `aggregate` command. `iterateUntilDocumentOrError` is never used there to assert the change stream contents.

In other tests, `createChangeStream` is executed _after_ other DDL operations in the test case (e.g. `dropCollection`, `createIndex`). It wouldn't make sense to amend those change streams to exclude "create" from `initialData`, and I think it'd be premature to exclude the particular DDL operation appearing earlier in the test operations.

I propose we leave this PR as-is and revisit as needed. A future test failure might also give us a better understanding of exactly what's going on with mongos here.

benjirewis:
@jmikola that seems like a fair theory; I'm fine with merging this as-is.

PR #1294, opened 2022-08-19
DRIVERS-2402: Clarify the behavior of comments on pre-4.4 servers

baileympearson:
This PR clarifies the intended behavior for the following commands on pre-4.4 servers when comment is attached.

- listIndexes
- listCollections
- listDatabases

- [x] Update changelog.


alcaeus (approved):
LGTM. This is consistent with other options where we send an explicitly provided value to servers that don't support it (e.g. read concern)

comandeo (approved):
LGTM :+1:

PR #1296, opened 2022-08-22
DRIVERS-2366 Use camelCase for ReadPreference mode in unified tests

patrickfreed:
This PR fixes the capitalization of a ReadPreference mode in one of the SDAM unified tests. It currently reads `Secondary`, which is consistent with existing server selection and max-staleness tests but inconsistent with existing unified format tests ([this CRUD unified test](https://github.com/mongodb/specifications/blob/master/source/crud/tests/unified/db-aggregate-write-readPreference.yml#L36) uses `secondaryPreferred`, for example).

This PR also updates the unified test format changelog to mention the addition of the `createEntities` operation.

Thanks to @dariakp for noticing these issues.

@ShaneHarvey @jyemin would you prefer if there was a separate DRIVERS ticket for this since you both already implemented the original change? Or do you just want to sync this in real quick as a one-off? Happy to do either, just figured I'd check before making some Jira noise.

### DIFF
--- a/source/server-discovery-and-monitoring/tests/unified/rediscover-quickly-after-step-down.json
+++ b/source/server-discovery-and-monitoring/tests/unified/rediscover-quickly-after-step-down.json
@@ -117,7 +117,7 @@
               "replSetFreeze": 0
             },
             "readPreference": {
-              "mode": "Secondary"
+              "mode": "secondary"
             },
             "commandName": "replSetFreeze"
           }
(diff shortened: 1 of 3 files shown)


jyemin:
Java driver does a toLower on the mode, so this didn't come up.  No need for a separate ticket.

ShaneHarvey (approved):
Python also lowercases so this works just fine.

jyemin (approved):

dariakp (approved):

jmikola (approved):

PR #1297, opened 2022-08-23
DRIVERS-2327 NoWritesPerformed Retryable Writes Error Handling

prestonvasquez:
DRIVERS-2327

- [x] Update changelog.

### DIFF
--- a/source/retryable-writes/retryable-writes.rst
+++ b/source/retryable-writes/retryable-writes.rst
@@ -3,14 +3,14 @@ Retryable Writes
 ================
 
 :Spec Title: Retryable Writes
-:Spec Version: 1.8.0
+:Spec Version: 1.8.1
 :Author: Jeremy Mikola
 :Lead: \A. Jesse Jiryu Davis
 :Advisors: Robert Stam, Esha Maharishi, Samantha Ritter, and Kaloian Manassiev
 :Status: Accepted
 :Type: Standards
 :Minimum Server Version: 3.6
-:Last Modified: 2022-01-25
+:Last Modified: 2022-08-30
 
 .. contents::
 
--- a/source/retryable-writes/retryable-writes.rst
+++ b/source/retryable-writes/retryable-writes.rst
@@ -409,7 +409,7 @@ error should be raised instead as the caller can infer that an attempt was made
 and the second error is likely more relevant (with respect to the current
 topology state).
 
-If a retry attempt fails with an error labeled `NoWritesPerformed`, drivers MUST
+If a retry attempt fails with an error labeled NoWritesPerformed, drivers MUST
 return the original error.
 
 Drivers MUST NOT attempt to retry a write command with the same transaction ID
--- a/source/retryable-writes/retryable-writes.rst
+++ b/source/retryable-writes/retryable-writes.rst
@@ -409,6 +409,9 @@ error should be raised instead as the caller can infer that an attempt was made
 and the second error is likely more relevant (with respect to the current
 topology state).
 
+If a retry attempt fails with an error labeled `NoWritesPerformed`, drivers MUST
+return the original error.
+
 Drivers MUST NOT attempt to retry a write command with the same transaction ID
 more than once.
 
--- a/source/retryable-writes/tests/README.rst
+++ b/source/retryable-writes/tests/README.rst
@@ -403,7 +403,7 @@ and sharded clusters.
       sharded deployment, the test runner MUST ensure that the client connects
       to only a single mongos host.
 
-   2. Configure a fail point with error code ``10107`` (NotWritablePrimary) error::
+   2. Configure a fail point with error code ``10107`` (NotWritablePrimary)::
 
          db.adminCommand({
                 configureFailPoint: "failCommand",
@@ -430,7 +430,8 @@ and sharded clusters.
         });
 
    4. Attempt an ``insertOne`` operation on any record for any database and
-      collection and assert that the associated error code is ``10107``.
+      collection. For the resulting error, assert that the associated error code
+      is ``10107``.
 
    5. Disable the fail point::
 
--- a/source/retryable-writes/tests/README.rst
+++ b/source/retryable-writes/tests/README.rst
@@ -399,9 +399,9 @@ and sharded clusters.
    label `NoWritesPerformed`. This test MUST be implemented by any driver that
    implements the Command Monitoring specification.
 
-   1. Create a client with ``maxPoolSize=1`` and ``retryableWrites=true``. If
-      testing against a sharded deployment, the test runner MUST ensure that the
-      client connects to only a single mongos host.
+   1. Create a client with ``retryableWrites=true``. If testing against a
+      sharded deployment, the test runner MUST ensure that the client connects
+      to only a single mongos host.
 
    2. Configure a fail point for a ``NoWritablePrimary`` error::
 
--- a/source/retryable-writes/tests/README.rst
+++ b/source/retryable-writes/tests/README.rst
@@ -428,11 +428,11 @@ and sharded clusters.
                 },
         });
 
-    4. Attempt an ``insertOne`` operation on any record for any database and
+   4. Attempt an ``insertOne`` operation on any record for any database and
        collection and assert that the associated error code is ``10107`` (i.e.,
        ``NoWritesPerformed``).
 
-    5. Disable the fail point::
+   5. Disable the fail point::
 
         db.adminCommand({
                 configureFailPoint: "failCommand",
--- a/source/retryable-writes/tests/README.rst
+++ b/source/retryable-writes/tests/README.rst
@@ -397,7 +397,11 @@ and sharded clusters.
 
 #. Test that drivers return the original error after encountering an error with
    label `NoWritesPerformed`. This test MUST be implemented by any driver that
-   implements the Command Monitoring specification.
+   implements the Command Monitoring specification. This prose test requires
+   drivers to set a fail point after an ``insertOne`` operation but before the
+   subsequent retry. Drivers that are unable to set failCommand after the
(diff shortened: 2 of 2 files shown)


prestonvasquez:
@ShaneHarvey For brevity, here is the POC in Go: https://github.com/mongodb/mongo-go-driver/pull/1052

source/retryable-writes/tests/README.rst
```diff
@@ -395,6 +395,50 @@ and sharded clusters.
...
+                configureFailPoint: "failCommand",
+                mode: {times: 1},
+                data: {
+                        errorCode: 10107,
+                        errorLabels: ["RetryableWriteError"],
+                        failCommands: ["insert"],
+                },
+        });
+
+   3. Via the command monitoring ``CommandFailedEvent``, configure a fail point
```
ShaneHarvey, line 418: I don't think this is possible in all drivers. I believe some drivers use async publishing on command events where the retry can run before the handler. Is there any other way to configure 2 different errors on subsequent attempts?

jmikola: A separate can of worms (albeit unlikely) is that running database commands from _within_ a APM handler seems a possible way to enter a feedback loop. If the `failCommand` command itself were to fail, we might re-enter the existing CommandFailedEvent handler and issue a second `failCommand`. That said, this would be pretty easy to work around in the test implementation (e.g. check that the failed command was `insert` and/or had the previously expected `10107` error code).

source/retryable-writes/tests/README.rst
```diff
@@ -395,6 +395,50 @@ and sharded clusters.
 
    9. Disable the failpoint.
 
+#. Test that drivers return the original error after encountering an error with
+   label `NoWritesPerformed`. This test MUST be implemented by any driver that
+   implements the Command Monitoring specification.
```
jmikola, line 400: Noting @ShaneHarvey's concerns about implementing this test below, it made me think back to whether we have existing tests for returning "original errors". There are several other points in the Retryable Writes spec that instruct drivers to do so, but it doesn't appear we have tests (neither spec nor prose) for any of those conditions.

source/retryable-writes/tests/README.rst
prestonvasquez: IIUC there is no way to sequentially issue fail points; the server will only register the most recent configuration. I can only think of two alternatives to using command monitor:

1. Explore using a mock dialer
2. Create a blocking server ticket to register multiple fail points

What are your thoughts?

ShaneHarvey: I think we should do both 1) and 2). That is, drivers that can use the CommandFailedEvent event option should be encouraged to do so (perhaps mention to filter based on the first failed _insert_ command only to deal with the problem Jeremy mentioned. Then we can also open a new server ticket to request the ability to set multiple different errors in failCommand. Something like:
```js
db.adminCommand({
    mode: {times: 2},
    data: {
        failCommands: ["insert"],
        errors: [
            {
                errorCode: 10107,
                errorLabels: ["RetryableWriteError"],
            }, {
                errorCode: 9001,
                errorLabels: ["RetryableWriteError", "NoWritesPerformed],
            }
        ]
    }
);
```

source/retryable-writes/tests/README.rst
```diff
@@ -395,6 +395,50 @@ and sharded clusters.
...
+   3. Via the command monitoring ``CommandFailedEvent``, configure a fail point
+      with error code ``9001`` (SocketException) and a NoWritesPerformed
+      label::
+
+         db.adminCommand({
+                configureFailPoint: "failCommand",
+                mode: {times: 1},
+                data: {
+                        errorCode: 9001,
+                        errorLabels: ["RetryableWriteError", "NoWritesPerformed],
```
ShaneHarvey, line 427: Missing a " in NoWritesPerformed.

source/retryable-writes/tests/README.rst
prestonvasquez: How prescriptive should the mocking instructions be? Should I pair with a driver that would require this implementation to make sure we have a comprehensive prose test for this case? Or is it enough to say something like "Drivers MAY use a mock connection class for testing the NoWritesPerformed error handling"?

ShaneHarvey: I don't think it needs to be prescriptive since it will vary based on the driver. We can say something like "drivers that are unable to set failCommand after the CommandFailedEvent SHOULD use mocking or write a unittest to cover the same sequence of events".

prestonvasquez: Great, thank you!

jmikola (approved):

source/retryable-writes/tests/README.rst
```diff
@@ -395,6 +395,57 @@ and sharded clusters.
...
+   drivers to set a fail point after an ``insertOne`` operation but before the
+   subsequent retry. Drivers that are unable to set a failCommand after the
+   CommandFailedEvent SHOULD use mocking or write a unit test to cover the same
+   sequence of events.
+
+   1. Create a client with ``retryableWrites=true``. If testing against a
+      sharded deployment, the test runner MUST ensure that the client connects
+      to only a single mongos host.
+
+   2. Configure a fail point with error code ``10107`` (NotWritablePrimary)::
```
ShaneHarvey, line 410: Would we be able to change this first error to a retryable write concern error so that we match Matthew's repro given in: https://jira.mongodb.org/browse/SERVER-66116?focusedCommentId=4526619&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4526619

source/retryable-writes/retryable-writes.rst
```diff
@@ -409,6 +409,9 @@ error should be raised instead as the caller can infer that an attempt was made
 and the second error is likely more relevant (with respect to the current
 topology state).
 
+If a retry attempt fails with an error labeled NoWritesPerformed, drivers MUST
+return the original error.
+
```
ShaneHarvey, line 414: Can you bump the spec version and add a changelog entry?

ShaneHarvey (changes requested):

source/retryable-writes/tests/README.rst
prestonvasquez: Absolutely!

source/retryable-writes/retryable-writes.rst
jmikola: Sorry for not requesting this before I approved. I assumed @prestonvasquez would get to it when ticking the checkboxes in the issue template.

source/retryable-writes/tests/README.rst
```diff
@@ -395,10 +395,67 @@ and sharded clusters.
...
+   unable to set a failCommand after the CommandSucceededEvent SHOULD use
+   mocking or write a unit test to cover the same sequence of events.
+
+   1. Create a client with ``retryableWrites=true``. If testing against a
+      sharded deployment, the test runner MUST ensure that the client connects
+      to only a single mongos host.
+
+   2. Configure a fail point with error code ``91`` (ShutdownInProgress)::
+
+         db.adminCommand({
```
visemet, line 412: [nit] It looks like there's an extra space at the beginning of the `db.adminCommand({` lines both here and on line 428. (9 spaces instead of 8 spaces.)

source/retryable-writes/tests/README.rst
```diff
@@ -395,10 +395,67 @@ and sharded clusters.
...
 
+#. Test that drivers return the original error after encountering a
+   WriteConcernError error with a RetryableWriteError label. This test MUST be
+   implemented by any driver that implements the Command Monitoring
+   specification. This prose test requires drivers to set a fail point after an
+   ``insertOne`` operation but before the subsequent retry. Drivers that are
+   unable to set a failCommand after the CommandSucceededEvent SHOULD use
+   mocking or write a unit test to cover the same sequence of events.
+
+   1. Create a client with ``retryableWrites=true``. If testing against a
```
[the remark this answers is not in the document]
prestonvasquez: I see, thank you for that thorough explanation.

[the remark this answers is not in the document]
ShaneHarvey: Yes mongos would need to apply the same rule as in this PR. Otherwise I believe it could return a non-retryable NoWritesPerformed-like error to the driver even though the write actually did succeed. @prestonvasquez could you open a server ticket for mongos to implement this change?

ShaneHarvey (approved):
LGTM!

source/retryable-writes/tests/README.rst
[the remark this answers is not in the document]
prestonvasquez: @ShaneHarvey @visemet [SERVER-69295](https://jira.mongodb.org/browse/SERVER-69295)

source/retryable-writes/tests/README.rst
```diff
@@ -396,20 +396,25 @@ and sharded clusters.
...
-   WriteConcernError error with a RetryableWriteError label. This test MUST be
-   implemented by any driver that implements the Command Monitoring
-   specification. This prose test requires drivers to set a fail point after an
+   WriteConcernError with a RetryableWriteError label. This test MUST
+
+   1. be implemented by any driver that implements the Command Monitoring
+      specification,
+
+   2. only be run against replica sets as mongos does not propagate the
+      RetryableWriteError label to the drivers.
```
visemet, line 405: Pedantically, mongos does propagate the RetryableWriteError label when the mongos itself is shutting down. So an InterruptedAtShutdown from the mongos itself shutting down should have the RetryableWriteError label attach but no instances of retryable errors returned by the shards to the mongos will have an error label attach if the error is propagated all the way back to the driver.

s/RetryableWriteError/NoWritesPerformed/ would allow you to omit these specifics.

visemet (approved):

PR #1300, opened 2022-09-02
Remove MongoDB < 3.6 information and front matter from command monitoring spec

kmahar:
I'm going to be making somewhat significant additions to this spec around command logging. A lot of the information in here is currently for outdated server versions that most drivers do not support and soon no drivers will support. Rather than trying to preserve it as I incorporate new information, I propose we remove it altogether. 

Rather than bumping the spec version and last modified date, I also took the liberty of removing the spec version and front matter as @jmikola will do in DRIVERS-2253 will do for all specs.

### DIFF
--- a/source/command-monitoring/command-monitoring.rst
+++ b/source/command-monitoring/command-monitoring.rst
@@ -443,3 +443,6 @@ Changelog
 
 18 MAY 2022:
   - Converted legacy tests to the unified test format.
+
+2 SEPTEMBER 2022:
+  - Remove material that only applies to MongoDB versions < 3.6.
--- a/source/command-monitoring/command-monitoring.rst
+++ b/source/command-monitoring/command-monitoring.rst
@@ -131,7 +131,7 @@ fields in the reply beyond the ``ok`` field.
 
 *2. Why are document sequences included as BSON arrays?*
 
-The OP_MSG wire protocol was introduced in MongoDB 3.6, with document sequences as an optimization for bulk writes. We have chosen to represent these OP_MSGs as single command or reply documents for now, until a need for a more accurate (and perhaps better-performing) command monitoring API for document sequences has been demonstrated.
+The OP_MSG wire protocol was introduced in MongoDB 3.6, with document sequences as an optimization for bulk writes. We have chosen to represent these OP_MSGs as single command documents for now, until a need for a more accurate (and perhaps better-performing) command monitoring API for document sequences has been demonstrated.
 
 *3. Why is BSON serialization and deserialization optional to include in durations?*
 
--- a/source/command-monitoring/command-monitoring.rst
+++ b/source/command-monitoring/command-monitoring.rst
@@ -5,14 +5,9 @@
 Command Monitoring
 ==================
 
-:Spec: 200
 :Title: Command Monitoring
-:Authors: Durran Jordan
 :Status: Approved
-:Type: Standards
 :Minimum Server Version: 2.4
-:Last Modified: 2022-05-18
-:Version: 1.10.0
 
 .. contents::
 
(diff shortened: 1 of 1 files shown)


source/command-monitoring/command-monitoring.rst
```diff
@@ -179,25 +123,10 @@ A document sequence in an OP_MSG reply to "aggregate" or "getMore" MUST be simil
 
 See "Why are document sequences included as BSON arrays?" in the `rationale`_.
 
-Read Preference
-^^^^^^^^^^^^^^^
-
-In cases where queries or commands are embedded in a ``$query`` parameter when a read preference
```
kmahar, line 185: @jmikola and I discussed this section [here](https://github.com/mongodb/specifications/pull/1221#discussion_r879756606) back when converting the tests to the unified format. The section is a bit confusing but I think it was referring to OP_QUERY behavior as described [here](https://github.com/mongodb/specifications/blob/master/source/find_getmore_killcursors_commands.rst#readpreference-and-mongos) so it does not seem relevant for OP_MSG.

source/command-monitoring/command-monitoring.rst
```diff
@@ -489,6 +415,9 @@ Changelog
 31 OCT 2015:
   - Changed find test on 3.1 and higher to ignore being run on sharded clusters.
 
+22 NOV 2015:
```
kmahar, line 418: the changelog was a little out of order so corrected that

source/command-monitoring/command-monitoring.rst
```diff
@@ -210,11 +139,11 @@ replies are received. As such, it would be innappropiate at this level for a dri
 custom logic around particular commands to determine what failure or success means for a particular
 command. Implementators of the API are free to handle these events as they see fit, which may include
 code that futher interprets replies to specific commands based on the presence or absence of other
-fields in the reply beyond the ‘ok’ field.
+fields in the reply beyond the ``ok`` field.
 
 *3. Why are document sequences included as BSON arrays?*
 
-The OP_MSG wire protocol was introduced in MongoDB 3.6, with document sequences as an optimization for bulk writes. The same optimization will be introduced for "find" and "aggregate" replies in 3.8. We have chosen to represent these OP_MSGs as single command or reply documents for now, until a need for a more accurate (and perhaps better-performing) command monitoring API for document sequences has been demonstrated.
```
kmahar, line 217: 3.8 never happened and I wasn't sure if/when the upcoming work described here actually happened (the OP_MSG spec only mentions the bulk use case and I couldn't find tickets about this). I took out the sentence for now, but if anyone knows if/when this did happen happy to add back mention of it

jyemin: It didn't happen, at least not in drivers.  I recall doing some experiments to see if it was any faster and it wasn't, so we dropped it

source/command-monitoring/command-monitoring.rst
```diff
@@ -179,25 +123,10 @@ A document sequence in an OP_MSG reply to "aggregate" or "getMore" MUST be simil
...
-In cases where queries or commands are embedded in a ``$query`` parameter when a read preference
-is provided, they MUST be unwrapped and the value of the ``$query`` attribute becomes the
-``filter`` or the command in the started event. The read preference will subsequently be dropped
-as it is considered metadata and metadata is not currently provided in the command events.
-
 ---------
 Rationale
 ---------
 
-*1. Why does the specification treat all events as commands, even those that are not sent as such?*
```
jmikola, line 194: These aren't prose tests, so I don't see a problem with re-numbering after you remove some questions. Alternatively, just remove the numbers entirely and leave the questions on their own. These are just emphasized text and not linked headers so we need not worry about anything linking directly to these.

kmahar: good catch. renumbered

source/command-monitoring/command-monitoring.rst
kmahar: I realized that a section above also discussed the use of document sequences in find/aggregate replies so I have removed that as well.

jmikola (approved):

kmahar:
@durran and @jyemin did you want to review/re-review this?

PR #1170, opened 2022-04-05
DRIVERS-1707: Preemptively cancel in progress operations when SDAM heartbeats timeout.

DmitryLukyanov:
DRIVERS-1707

PR #1303, opened 2022-09-09
DRIVERS-1204: Add general logging specification and command logging

kmahar:
This is the first in a series of PRs for the logging spec. 

Note this is on top of #1300; I'll rebase once that gets merged in.

This PR adds:
* A new general logging specification describing requirements for how drivers implement logging and defining shared concepts other specs can use, e.g. common log levels and component names
* Additions to the unified test format to support logging assertions (I know there is an existing PR that also adds a schema version 1.11; if that goes in first I will update to use 1.12 instead). 
* Rebrands the command monitoring spec as the command logging and monitoring spec (I considered CMAL but it felt like the acronym CLAM was too good to pass up...), and adds information about command logging, along with both unified format and prose tests.

Future PRs will add logging and tests for CMAP, SDAM, and server selection.


source/logging/logging.rst
```diff
@@ -0,0 +1,380 @@
...
+If possible, drivers SHOULD support the following configuration requirements in a manner that does
+not require any changes to an application's source code, and in the case of compiled languages,
+does not require recompilation.
+However, if that is impossible to do without conflicting with idiomatic logging patterns for the
+language ecosystem, it is acceptable for a driver to require minimal changes to enable logging,
+such as recompiling with a feature flag specified, or passing a logging configuration to a 
+``MongoClient`` constructor. In that case, drivers SHOULD strive to minimize the amount of code
+change needed.
+
+Drivers SHOULD take advantage of built-in support for logging configuration within logging
```
kmahar, line 58: the "fallback implementation methods" here are all things I have not actually attempted to do, because Rust has a logging framework to rely on.
I expect we might have to adjust these once a language without a logging framework actually attempts this (iirc @jmikola is going to do the libmongoc implementation later this quarter?)

source/logging/logging.rst
```diff
@@ -0,0 +1,380 @@
...
+Drivers SHOULD implement support for logging in a manner that is idiomatic for their language and
+ecosystem.
+
+
+Minimizing Required Code Changes
+--------------------------------
+If possible, drivers SHOULD support the following configuration requirements in a manner that does
+not require any changes to an application's source code, and in the case of compiled languages,
+does not require recompilation.
+However, if that is impossible to do without conflicting with idiomatic logging patterns for the
```
kmahar, line 52: while we had a scope goal around requiring no recompilation but per discussion in the project Slack channel ended up loosening the requirement.

iirc it would be un-idiomatic for .NET to support with no code change at all (@BorisDog could provide more context on this.) 
and it is infeasible for Rust to support without recompilation right now - the only crates available to us for logging/tracing support are pre-1.0 and so it is best practice to gate their usage behind a feature flag (e.g. the Tokio async runtime we depend on does the same thing).

source/command-logging-and-monitoring/command-logging-and-monitoring.rst
```diff
@@ -0,0 +1,615 @@
...
+Drivers MUST support logging of command information via the following types of log messages. These messages MUST be logged at ``Debug`` level and use
+the ``command`` log component.
+
+The log messages are intended to match the information contained in the events above. Drivers MAY implement command logging
+support via an event subscriber if it is convenient to do so.
+
+The types used in the structured message definitions below are demonstrative, and drivers MAY use similar types instead so long as the information
+is present (e.g. a double instead of an integer, or a string instead of an integer if the structured logging framework does not support numeric types.) 
+
+For the string representations of messages defined below, for any optional field that is not available, drivers MAY either substitute the value with a
```
kmahar, line 360: I can see arguments for either of these. on the one hand, it might wrongly concern users to see a value included in log messages is always null (for example if they are on < 4.2 and serverConnectionId is always null, or they aren't doing any bulk writes and so the operationId is always missing). also, it will make messages longer/arger if we always include the extra value.

on the other hand, in some cases it may give a more explicit indicator of a problem when debugging to say that something is null. a TSE or user reading a log won't necessarily know that a missing value means the driver considers the value to be null or that the value should have been included at all.

any thoughts?

source/command-logging-and-monitoring/command-logging-and-monitoring.rst
```diff
@@ -0,0 +1,615 @@
...
+language-appropriate representation of null (e.g. "null", "nil", "None") or omit the corresponding segment of the message altogether. 
+
+Drivers MUST not emit command log messages for commands issued as part of the handshake with the server, or heartbeat commands issued by server monitors.
```
kmahar, line 363: Rust doesn't emit command monitoring events for these commands already so this was easy for us and consistent with the monitoring APIs. but not sure if the same is true for other drivers?

source/command-logging-and-monitoring/command-logging-and-monitoring.rst
```diff
@@ -0,0 +1,615 @@
...
+
+   * - serverConnectionId
+     - Int
+     - The server's ID for the connection used for the command. Optional; only present for server versions 4.2+.
+
+   * - serviceId
+     - String
+     - The hex string representation of the service ID for the command. Optional; only present when the driver is in load balancer mode.
+
+The unstuctured form MUST be as follows, using the values defined in the structured format above to fill in placeholders as appropriate:
```
kmahar, line 422: I wasn't sure whether it was helpful or too prescriptive to define the exact stringified form. I do think consistency across drivers that produce strings could be helpful. but we won't actually be testing this. any thoughts?

source/command-logging-and-monitoring/tests/README.rst
```diff
@@ -0,0 +1,53 @@
...
+
+1. Configure logging with a minimum severity level of "debug" for the "command" component. Set the max document length to 5 of (your driver's unit of choice for truncation).
+2. Run the command ``{"hello": true}```.
+3. Inspect the resulting "command started" log message and assert that the "command" value is a string of length 5.
+4. Inspect the resulting "command succeeded" log message and assert that the "reply" value is a string of length 5.
+5. If the driver attaches raw server responses to failures and can access these via log messages to assert on, run the command 
+   ``{"notARealCommand": true}``. Inspect the resulting "command failed" log message and confirm that the server error is
+   a string of length 5.
+
+*Test 3: Truncation with multi-byte codepoints*
```
kmahar, line 47: I'm not really sure how drivers can write these tests in a way that is not brittle to the formatting/ordering of commands and replies, as well as the topology you are connected to, since you basically have to find particular byte counts that fall mid code point and set those counts as your max length and see how truncation goes. 

here's the one I came up with for Rust: https://github.com/kmahar/mongo-rust-driver/blob/a03a6990dd8884a93653cd968ac953ca5193dc41/src/test/spec/trace.rs#L137

source/command-logging-and-monitoring/command-logging-and-monitoring.rst
```diff
@@ -150,8 +150,9 @@ Security
 --------
 
 Some commands and replies will contain sensitive data and in order to not risk the leaking of this
-data to external sources or logs their commands AND replies MUST be redacted from the events. The
-value MUST be replaced with an empty BSON document. The list is as follows:
+data to external sources or logs their commands AND replies AND corresponding error messages/exceptions
```
kmahar, line 153: the error messages/exceptions bit is new; see the comments on DRIVERS-2394 for some context. the security team recommends we redact these just in case, so I went ahead and added the requirement for log messages and have expectations for it in the tests.

source/command-logging-and-monitoring/command-logging-and-monitoring.rst
```diff
@@ -341,42 +342,217 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
...
+
+   * - serverHost
+     - String
+     - The hostname or IP address for the server the command is being run on.
+
+   * - serverPort
+     - Int
+     - The port for the server the command is being run on. Optional; only present if a port was specified.
 
-A Ruby subscriber to a query series and how it could handle it with respect to logging.
```
kmahar, line 349: this example seemed like it would be confusing in light of actual logging info being added to the spec. I didn't bother to try to write a replacement example but I can do so if anyone thinks that is needed.

source/command-logging-and-monitoring/command-logging-and-monitoring.rst
jyemin: It's true for Java driver

source/unified-test-format/unified-test-format.rst
```diff
@@ -1147,13 +1168,13 @@ of ``eventType`` in the corresponding `expectedEventsForClient`_
 object, which can have one of the following values:
 
 - ``command`` or omitted: only the event types defined in
-`expectedCommandEvent`_ are allowed.
+  `expectedCommandEvent`_ are allowed.
```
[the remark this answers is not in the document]
jmikola: Thanks. It looks like this snuck in through https://github.com/mongodb/specifications/commit/08baffedabe6c13648f48700fd3443502b9f805a#diff-f718c246f51e05b3731a7b03b176a0d4b0c381a9261810742d2677823f963f98R1149 /cc @patrickfreed

source/command-logging-and-monitoring/command-logging-and-monitoring.rst
jmikola: IIRC libmongoc has trace-level logging that can dump all data sent over sockets. This includes messages pertaining to handshakes, auth, and monitoring. /cc @kevinAlbs

source/unified-test-format/unified-test-format.rst
```diff
@@ -1342,6 +1363,66 @@ MUST assert that the field is not set, or, if the driver uses a nonpositive Int3
...
+
+- ``client``: Required string. Client entity on which the events are expected
+  to be observed. See `commonOptions_client`_.
+
+- ``messages``: Required array of `expectedLogMessage`_ objects. List of 
+  messages, which are expected to be observed (in this order) on the corresponding
+  client while executing `operations`_. If the array is empty, the test runner
+  MUST assert that no messages were observed on the client.
+
+expectedLogMessage
```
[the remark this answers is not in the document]
jmikola: If I understand correctly, the fact that `expectedLogMessage`, and thus its `data` field, is generalized is the reason we need `$$matchAsDocument` and `$$matchAsRoot`.

I acknowledge the argument against exhaustively modeling each log message, but if we hypothetically _were_ to do that, would those special operators become unnecessary?

I also haven't diligently read through all of the new tests, but I saw some instance of the operators being used in sequence. Are they always use sequence and, if so, does it make sense to combine them into a single ``$$matchJsonAsDocument`` operator?

source/command-logging-and-monitoring/command-logging-and-monitoring.rst
jyemin: I prefer to leave them out.  On balance, it's more confusing to include a null for a value that the driver knows is always going to be null in some situations, than to leave it out and risk confusing everyone reading.  I've seen something similar with the C# drivers's representation of host names in server selection timeout exception messages.  I think the ipV6 address is always printed as null even though it rarely exists.  It always confuses me.

source/logging/logging.rst
```diff
@@ -0,0 +1,380 @@
...
+
+    If the variable is not provided, the driver MUST log to stderr.
+
+Configurable Max Document Length
+--------------------------------
+Drivers MUST support configuring the maximum logged length for extended JSON documents in log messages.
+The unit here is flexible and can be bytes, Unicode code points, or graphemes, depending on what a
+driver is able to accomplish with its language's string APIs. The default max length is 1000 of
+whichever unit is selected.
+If the chosen unit is bytes, the driver MUST ensure that it gracefully handles cases where the
```
jyemin, line 112: In Java driver's implementation, the limit is specified as number of "chars", which in Java is two bytes.  This includes all characters in the Basic Multilingual Plane (BMP).  I think this paragraph would apply to Java as well.  So maybe change to something like:

> If the chosen unit is anything other than a Unicode code point ...

source/logging/logging.rst
```diff
@@ -0,0 +1,380 @@
...
+    all.
+
+Components
+----------
+Drivers MUST support configuring minimum log severity levels on a per-component level. The below
+components currently exist and correspond to the listed specifications. This list is expected to
+grow over time as logging is added to more specifications.
+
+Drivers SHOULD specify the component names in whatever the idiomatic way is for their language.
+For example, the Java command component could be named ``org.mongodb.driver.command``.
```
jyemin, line 132: It's actually named `org.mongodb.driver.protocol.command`

source/logging/logging.rst
```diff
@@ -0,0 +1,380 @@
...
+
+   * - Component Name
+     - Specification(s)
+     - Environment Variable
+
+   * - command
+     - `Command Monitoring <../command-monitoring/command-monitoring.rst>`__
+     - ``MONGODB_LOG_COMMAND``
+
+   * - sdam
```
jyemin, line 149: Let's avoid this abbreviation.  Very few people reading the log files will know what it refers to.  How about `topology` instead?

source/logging/logging.rst
```diff
@@ -0,0 +1,380 @@
...
+these levels, but we have included them in the list of permitted levels for consistency with Syslog.
+
+The levels above are defined in order from most to least severe. Not all logging frameworks will
+necessarily support all of these levels. If an equivalent level is not available, drivers MUST emit
+messages for that level at the closest, less severe level if one is available, or the closest more
+severe level otherwise.
+
+For example:
+
+- If an Informational level is not available and Debug is, Informational messages defined as \
```
jyemin, line 244: Informational messages defined as Informational -> messages defined as Informational

source/logging/logging.rst
```diff
@@ -0,0 +1,380 @@
...
+--------------------------------------
+If structured logging is available in and idiomatic for the driver's language/ecosystem, the driver
+SHOULD produce structured log messages. Note that some ecosystems with structured logging support
+may also have support available to convert structured output to traditional unstructured messages
+for users who want it (for example, the
+`log feature <https://docs.rs/tracing/latest/tracing/#emitting-log-records>`_ in Rust's
+`tracing <https://docs.rs/tracing/latest/tracing/>`_ crate). If such support is available, drivers
+SHOULD utilize it to support both types of logging.
+
+Note that drivers implementing unstructured logging MUST still support some internal way to intercept
```
jyemin, line 258: Add something like:

> , as this is required to implement the unified tests for logging conformance.

source/logging/logging.rst
```diff
@@ -0,0 +1,380 @@
...
+the data contained in messages in a structured form. See the `unified test
+format specification <../unified-test-format/unified-test-format.rst#expectedLogMessage>`_ for details.
+
+Representing Errors in Log Messages
```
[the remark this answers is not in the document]
jyemin: CLAM spec!  I like it!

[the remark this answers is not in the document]
jyemin: I like it here.  I think it could show up in other places, like CMAP

source/logging/logging.rst
```diff
@@ -0,0 +1,380 @@
...
+-----------------------------------
+Drivers MAY represent errors in log messages in whatever format is idiomatic for their language and
+existing error types. For example, if a driver's error classes have existing ``toString()``
+implementations, those MAY be used. Alternatively, if a driver emits structured log messages, a
+structured format containing error data could be used. Any information which a driver reports via
+its error classes MUST be included in the log representations. 
+Note that if the driver includes full server responses in its errors these MUST be truncated in
```
jyemin, line 269: Is there a real possibility of essentially unbounded error replies?  I guess a bulk write where every write has a different error?

source/logging/logging.rst
```diff
@@ -0,0 +1,380 @@
...
+accordance with the max document length option.
+
+Performance Considerations
```
jyemin, line 272: Consider adding something like:

> Drivers SHOULD optimize JSON generation such that the complexity is O(N) where N = `<max document length>`, rather than N = `<actual document length>`

source/logging/logging.rst
```diff
@@ -0,0 +1,380 @@
...
+    documents by default, but allowing the user to adjust the maximum length logged. We chose the
+    third option as we felt it struck the best balance between concerns around readability and
+    usability of log messages. In the case where data is sufficiently small, the default behavior
+    will show the user the full data. In the case where data is large, the user will receive a
+    readable message with truncated data, but have the option to see more or all of the data.
+
+2. Why are the units for max document length flexible? 
+    String APIs vary across languages, and not all drivers will be able to easily and efficiently
+    truncate strings in the same exact manner. The important thing is that the option exists and
+    that its default value is reasonable, and for all possible unit choices (byte, code point, or
```
jyemin, line 320: Add `code unit` to this list, as this concept corresponds most closely to Java's String class with it's 2-byte code units.

source/logging/logging.rst
```diff
@@ -0,0 +1,380 @@
...
+    grapheme) we felt 1000 was a reasonable default. See
+    `here <https://exploringjs.com/impatient-js/ch_unicode.html>`_ for a helpful primer on related
+    Unicode concepts.
+
+--------------------------------------
+Structured versus Unstructured Logging
+--------------------------------------
+The MongoDB server produces structured logs as of 4.4, so it seems natural that MongoDB drivers
+might too. However, the idiomatic logging mechanisms of choice for some language ecosystems (e.g.
+Java) do not support structured logging yet, so we cannot require it.
```
jyemin, line 330: I'm not sure this is a strong argument.  Even though SLF4J doesn't support structured logging, the Java driver could still output log messages that _look_ like structured logs.  But still, it's not common to see logs like that in Java.

source/logging/logging.rst
```diff
@@ -0,0 +1,380 @@
...
+
+Backwards Compatibility
```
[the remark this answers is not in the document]
jyemin: This lgtm

source/logging/logging.rst
```diff
@@ -0,0 +1,380 @@
...
+    **Fallback implementation method**: If the environment variable ``MONGODB_LOG_PATH`` is provided:
+     -  If the value is "stdout" (case-insensitive), log to stdout.
+     - If the value is "stderr" (case-insensitive), log to stderr.
+     - Else, log to a file at the specified path. If the file already exists, it MUST be appended to.
+
+    If the variable is not provided, the driver MUST log to stderr.
+
+Configurable Max Document Length
+--------------------------------
+Drivers MUST support configuring the maximum logged length for extended JSON documents in log messages.
```
jyemin, line 108: Consider specifying somewhere that Extended JSON Relaxed mode MUST be used for representing BSON documents.

source/logging/logging.rst
```diff
@@ -0,0 +1,380 @@
...
+stderr, output file (path MUST be configurable).
+
+    **Fallback implementation method**: If the environment variable ``MONGODB_LOG_PATH`` is provided:
+     -  If the value is "stdout" (case-insensitive), log to stdout.
+     - If the value is "stderr" (case-insensitive), log to stderr.
+     - Else, log to a file at the specified path. If the file already exists, it MUST be appended to.
+
+    If the variable is not provided, the driver MUST log to stderr.
+
+Configurable Max Document Length
```
jyemin, line 106: The Java driver appends a trailing ` ...` when truncation was applied.  It seemed like a good idea at the time.  Should it be allowed?  Required?

source/unified-test-format/unified-test-format.rst
```diff
@@ -1342,6 +1363,66 @@ MUST assert that the field is not set, or, if the driver uses a nonpositive Int3
...
+  contained in the log message matches the expected data, and MUST treat the
+  log message data as a root-level document.
+  A suggested implementation approach is to decode ``data`` as a BSON document
+  and serialize the data attached to each log message to a BSON document, and
+  match those documents.
+  Note that for drivers that do not implement structured logging, this requires
+  designing logging internals such that data is first gathered in a structured
+  form (e.g. a document or hashmap) which can be intercepted for testing purposes.
+  A sample implementation for Java can be found on this 
+  `branch <https://github.com/jyemin/mongo-java-driver/tree/j4486>`_.
```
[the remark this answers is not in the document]
jyemin: It's almost certainly going to disappear.  I would just remove this.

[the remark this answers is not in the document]
jyemin: Maybe just link to the commit (https://github.com/jyemin/mongo-java-driver/commit/fbe9a044075cbeaddc55426fd256a463d45acbfb) in the DRIVERS ticket instead.

source/unified-test-format/unified-test-format.rst
```diff
@@ -615,6 +610,17 @@ The structure of this object is as follows:
...
 
+  .. _entity_client_observeLogMessages:
+
+  - ``observeLogMessages``: Optional object where the key names are log 
+    `components <./logging/logging.rst#components>`__ and the values are minimum
+    `log severity levels <logging/logging.rst#log-severity-levels>`__ indicating
+    which components to collect log messages for and what the minimum severity
+    level of collected messages should be. Messages for unspecified components
+    and/or with lower severity levels than those specified MUST be ignored by
+    this client's log collector(s) and SHOULD NOT be included in
```
jyemin, line 621: Why SHOULD NOT instead of MUST NOT?

source/unified-test-format/unified-test-format.rst
```diff
@@ -929,6 +935,21 @@ The structure of this object is as follows:
...
+- ``expectLogMessages``: Optional array of one or more `expectedLogMessagesForClient`_
+  objects. For one or mroe clients, a list of log messages that are expected to
+  be observed in a particular order.
+
+  If a driver only supports configuring log collectors globally (for all
+  clients), the test runner SHOULD associate each observed message with a client
+  in order to perform these assertions. One possible implementation is to add a
+  test-only option to MongoClient which 
+
+  Tests SHOULD NOT specify multiple `expectedLogMessagesForClient`_ objects for a
```
jyemin, line 950: Why not MUST NOT?

source/unified-test-format/unified-test-format.rst
```diff
@@ -929,6 +935,21 @@ The structure of this object is as follows:
...
   identical) or likely to fail (if the ``events`` arrays differed).
 
+
+.. _test_expectLogMessages:
+
+- ``expectLogMessages``: Optional array of one or more `expectedLogMessagesForClient`_
+  objects. For one or mroe clients, a list of log messages that are expected to
+  be observed in a particular order.
+
+  If a driver only supports configuring log collectors globally (for all
```
jyemin, line 945: Reviewing the Java driver POC, I recall a conversation about using `uriOptions,appName` as a sneaky way for the Java driver to get around the problem of a global log collector. There is even an assertion that requires it:

```
        this.applicationName = requireNonNull(applicationName);
```

I can't recall the outcome of our discussion, but I don't see a way to implement this in Java without using appname as a way to control which messages are intercepted.  Can we consider making it a requirement to specify appname whenever log messages are being observed?

jyemin (changes requested):
Lots of comments, but this looks fantastic.  Nice work.

source/command-logging-and-monitoring/command-logging-and-monitoring.rst
```diff
@@ -341,42 +342,217 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
...
+     - The hex string representation of the service ID for the command. Optional; only present when the driver is in load balancer mode.
 
-.. code:: ruby
+The unstuctured form MUST be as follows, using the values defined in the structured format above to fill in placeholders as appropriate:
 
-  class LoggingSubscriber
-    def logger
-      Logger.new(STDERR)
-    end
+  Command "{{commandName}}" started on database "{{databaseName}}" using a connection with driver ID {{driverConnectionId}} and server ID
```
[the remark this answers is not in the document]
jyemin: Never mind.  I see how I did this in the Java POC.

source/command-logging-and-monitoring/tests/logging/unified/operation-id.yml
```diff
@@ -0,0 +1,100 @@
+# This test only applies to drivers that generate operationIds to enable users to link
```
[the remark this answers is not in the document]
BorisDog: Yes this works with minor fix.

BorisDog (changes requested):

source/command-logging-and-monitoring/tests/logging/unified/operation-id.yml
```diff
@@ -0,0 +1,100 @@
...
+  - collectionName: *collectionName
+    databaseName: *databaseName
+    documents:
+      - { _id: 1, x: 11 }
+
+tests:
+  - description: "Successful bulk write command log messages include operationIds"
+    operations:
+      - name: bulkWrite
+        object: *collectionName
```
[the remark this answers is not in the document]
kmahar: thanks, fixed!

kmahar:
uh oh, I messed up that rebase. sorry for the noise everyone who had unrelated commits get pulled into this

source/command-logging-and-monitoring/tests/README.rst
```diff
@@ -0,0 +1,57 @@
...
+5. If the driver attaches raw server responses to failures and can access these via log messages to assert on, run the command 
+   ``{"notARealCommand": true}``. Inspect the resulting "command failed" log message and confirm that the server error is
+   a string of length 5 + (length of trailing ellipsis).
+
+*Test 3: Truncation with multi-byte codepoints*
+
+A specific test case is not provided here due to the allowed variations in truncation logic as well as varying extended JSON whitespace usage.
+Drivers MUST write language-specific tests that confirm truncation of commands, replies, and (if applicable) server responses included in error
+messages work as expected when the data being truncated includes multi-byte Unicode codepoints.
+If the driver uses bytes as the unit for max document length, there also MUST be tests confirming that cases where the max byte length falls in
```
jyemin, line 56: This needs a change similar to the one from https://github.com/mongodb/specifications/pull/1303#discussion_r970854633

source/logging/logging.rst
```diff
@@ -0,0 +1,416 @@
...
+
+   * - Component Name
+     - Specification(s)
+     - Environment Variable
+
+   * - command
+     - `Command Monitoring <../command-monitoring/command-monitoring.rst>`__
+     - ``MONGODB_LOG_COMMAND``
+
+   * - topologu
```
jyemin, line 153: I think you meant `topology` ;)

source/logging/logging.rst
jyemin: Yeah, that works.

source/unified-test-format/unified-test-format.rst
jyemin: I will defer to @jmikola on all things unified.

source/unified-test-format/unified-test-format.rst
jyemin: That could certainly be done, yes.  The only potential issue is the entity names are typically short and not unique, but then so might appName.  And in reality we should be closing all of our MongoClient instances in our tests so at most there should be just a few that are active at a time: the one currently under test, and the "global" one that is used for all the ad hoc integration tests.

source/command-logging-and-monitoring/command-logging-and-monitoring.rst
BorisDog: Same for .Net. But in my implementation logs for heartbeats commands were emitted. It was an easy fix, but wondering whether this can/worth to be tested .

source/logging/logging.rst
```diff
@@ -0,0 +1,416 @@
...
+in log messages. While this might make it easier to identify cases where a value is unexpectedly null,
+we decided against it because there are a number of values that will often be null, or even always
+be null for certain applications (e.g. ``serviceId`` when not connected to a load-balanced topology)
+and their inclusion may confuse users and lead them to think the null value is meaningful.
+Additionally, always including null values would increase the size of log messages.
+
+Backwards Compatibility
+=======================
+This specification takes the stance that the contents of log messages (both structured and
+unstructured) are _not_ covered by semantic versioning, but that logging components _are_, since
```
BorisDog, line 371: Are _not_/_are_ missing additional underscore?

source/command-logging-and-monitoring/tests/logging/operation-id.yml
```diff
@@ -0,0 +1,99 @@
...
+            component: command
+            data:
+              message: "Command succeeded"
+              commandName: delete
+              operationId: { $$type: [int, long] }
+        
+  - description: "Failed bulk write command log message includes operationId"
+    operations:
+      - name: bulkWrite
+        object: *collectionName
```
jyemin, line 75: `*collectionName` => `*collection`

jyemin (changes requested):

source/unified-test-format/unified-test-format.rst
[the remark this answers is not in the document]
jmikola: Inconsistent JSON spacing _and_ field ordering would likely be an issue.

This SGTM. Thanks for the extra context.

source/command-logging-and-monitoring/command-logging-and-monitoring.rst
```diff
@@ -341,42 +342,217 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
...
+     - Int
+     - The driver-generated request ID.
+  
+   * - operationId
+     - Int
+     - The driver-generated operation ID. Optional; only present if the driver generated operation IDs and this command has one. 
+
+   * - driverConnectionId
+     - Int
+     - The driver's ID for the connection used for the command, as defined in the `connection monitoring and pooling specification <../connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst>`_.
```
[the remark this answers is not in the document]
jmikola: > I split the connectionId assertions into their own test case to make it easy for non-CMAP drivers to skip them.

Both libmongoc and PHPLIB can skip individual test cases within a file, so there's no need to break these out to separate files if it's more maintainable to keep them together with other, related tests.

source/logging/logging.rst
```diff
@@ -0,0 +1,380 @@
...
+  component variable(s) MUST take precedence.
+
+  The default is to not log anything. 
+
+  If a variable is set to an invalid value, it MUST be treated as if it were not specified at all.
+
+Configurable Log Destination
+----------------------------
+Drivers MUST support configuring where log messages should be output, including the options: stdout,
+stderr, output file (path MUST be configurable).
```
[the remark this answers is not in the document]
jmikola: SGTM.

For the record, PHPLIB would integrate with the logging API defined in [PSR-3](https://www.php-fig.org/psr/psr-3/). Monolog is just one of the more common implementations that isn't tied to a full app framework. We'd likely use that for our testing or examples :)

source/command-logging-and-monitoring/tests/logging/redacted-commands.yml
```diff
@@ -0,0 +1,460 @@
...
+            component: command
+            data:
+              message: "Command started"
+              databaseName: *databaseName
+              commandName: authenticate
+              command:
+                $$matchAsDocument: {}
+          - level: debug
+            component: command
+            hasFailure: false
```
jyemin, line 48: Should all of the `hasFailure: false` in this file be `hasFailure: true`?

jyemin (changes requested):
Found one more potential issue while implementing the tests.

source/command-logging-and-monitoring/command-logging-and-monitoring.rst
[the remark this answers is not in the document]
kmahar: I don't think it makes much of a difference in terms of maintainability; a separate file is also consistent with what I did for other tests that test optional values like operation ID and service ID.

source/command-logging-and-monitoring/command-logging-and-monitoring.rst
kmahar: I added some tests, `no-heartbeat-messages.yml` and `no-handshake-messages.yml`, that I think at least implicitly test this. 
the heartbeat one creates a client, waits for a server description change (implying a heartbeat happened), and then runs a ping, and then asserts we have only seen command logs for the ping.
the handshake one creates a client, runs a ping, waits for connection created and ready events (implying a handshake happened), then runs a ping, and then asserts we have only seen command logs for the ping.

let me know what you think and if those pass for you! cc @jyemin could you take a look and try in Java too?

source/logging/logging.rst
kmahar: updated

source/logging/logging.rst
kmahar: ugh, I need an RST spell checker, thanks

source/command-logging-and-monitoring/tests/README.rst
kmahar: updated

source/logging/logging.rst
kmahar: I think it should actually be asterisks instead of underscores, I meant to make it italic. fixed! thanks

source/command-logging-and-monitoring/tests/logging/redacted-commands.yml
kmahar: hasFailure is meant to test that driver redacts the error for sensitive commands:

> ``hasFailure``: Optional boolean. When this field is set to ``true``, 
  the test runner MUST assert that the log message has a ``failure`` value
  in its attached data, and that the failure has not been redacted in any
  way. When this field is set to ``false``, the test runner MUST assert
  that the log message either has no ``failure`` value at all or that it
  has a redacted failure, e.g. an empty string or object. 
  The exact form of those assertions will vary based on the driver's chosen
  representations for failures and redacted failures in log messages.
  Currently, this assertion is only relevant for "command failed" log messages.

 some context on the redaction in this [comment](https://github.com/mongodb/specifications/pull/1303#discussion_r967396950).

since we want to redact the failure for all of these commands,  I think `false` is the right value.

jyemin: ok, I'm going to leave a comment on the `hasFailure` definition, which I now have some concerns about, and resolve this one.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1342,6 +1365,65 @@ MUST assert that the field is not set, or, if the driver uses a nonpositive Int3
...
+   level. Test runners MUST assert that the actual level matches this value.
+
+- ``component``: Required string. This MUST be one of the component names listed
+   in `components <../logging/logging.rst#components>`__. This specifies the
+   expected component for the log message. Note that since naming variations
+   are permitted for components, some drivers may need to map this to a
+   corresponding language-specific component name. Test runners MUST assert
+   that the actual component matches this value.
+
+- ``hasFailure``: Optional boolean. When this field is set to ``true``,
```
jyemin, line 1407: Since this has multiple conditions wrapped into a single boolean value, we should have tests for each cell in the truth table.  These are currently missing:

- assert `hasFailure: false for a non-exceptional log message (e.g. command-succeeded) 
- assert `hasFailure: true for an exceptional log message (e.g. command-failed with un-redacted failure) 

Also consider simplifying the boolean by making this one just about failure and a second one that is just about redaction.

jyemin (changes requested):

source/unified-test-format/unified-test-format.rst
kmahar: @jyemin and I discussed this offline and ended up having a broader conversation about error redaction. 

We reached the following conclusions:

* We should only redact command errors, not all errors. E.g. an IO exception in response to a sensitive command should not be redacted.
    * We can test this by using the failCommand failpoint with `closeConnection: true` to generate a non-command error.
* When we redact command errors, we should leave intact information that is definitely not going to be sensitive and could be helpful for debugging. We think [error code, error name, error labels] would be a reasonable allowlist. The error message and any other arbitrary fields should be redacted.
    * If a driver attaches the raw server response to command errors (as DRIVERS-2385 will mandate), this should be similarly redacted with only [error code, error name, error labels] present in the document.

An open question is what the right way to test the above redaction proposal is. We need to be able to test the following things when asserting a command error is redacted:
1. That an error code, error code name, and error labels are present/non-redacted.
2. That no other properties are present/unredacted.
3. That the attached server response only contains an error code, error code name, and error labels.

One potential path here is to adapt [expectedError](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#expectederror) to support this, and add an optional `expectedError` to `expectedLogMessage`, which would replace `hasFailure` as currently defined in this PR.

Currently, `expectedError` would already allow us to make assertion 1 above. 

To support assertion 2, we could potentially add a new property to `expectedError` e.g. `isRedacted`,and when present the test runner would assert that all properties besides the allowed ones are null/empty. (I do worry that such an assertion might be rather brittle since it could be easy to add new properties to the command error type and forget to update this logic, but I don't really know a good way around that.)

To support assertion 3: `isRedacted` could also require asserting on the server response and checking it only contains certain keys. 
Alternatively, @jmikola and I discussed that DRIVERS-2385 would likely add an expected server response to `expectedError`. If that expected response is _not_ matched as a root document, the expected value could simply contain only keys from the allowlist to facilitate this. But, it does seem that in _most_ cases we would probably want treat this expected server response as a root document, similar to how the reply in a CommandSucceededEvent is always matched as a root document. That said, it always being root made DRIVERS-1653 and DRIVERS-1816 difficult and we ended up writing "best effort" tests that just check a subset of fields we'd expect to be present are not. I suppose the addition of `$$matchAsRoot` to the test format via this PR makes it possible for any non-root document to be matched as root when needed, so if we _did_ make it non-root by default there would be a way to opt into root matching when needed.

To summarize, the two options I'm proposing here are:
1. Add `isRedacted` to `expectedError`, and have `isRedacted` make assertions 2 and 3 above. 
2. Add `isRedacted` to `expectedError`, and have `isRedacted` assert that the error object has a code, codeName, and errorLabels, but no other properties. Have the expected server response being added in DRIVERS-2385 be matched as non-root by default, and use that for assertion 3. 

If you can think of other alternatives that seem better than either of these, let me know.

@jyemin let me know if I missed anything; @jmikola tagging you since this relates to DRIVERS-2385, and also for general input on the testing points.

Aside: the command monitoring equivalent ticket around testing redaction of failures is DRIVERS-2394. Ideally whatever we come up with here to test log message failure redaction would be mostly reusable there.

kmahar:
#1170 went in and added schema version 1.11 so I have updated accordingly:
* renamed my new schema file to schema-1.12.json and incorporated the change from 1.11
* updated all my new tests to use 1.12
* updated changelog and schema version in unified format spec file to say 1.12 instead.

source/command-logging-and-monitoring/tests/README.rst
```diff
@@ -0,0 +1,57 @@
...
+
+*Test 1: Default truncation limit*
+
+1. Configure logging with a minimum severity level of "debug" for the "command" component. Do not explicitly configure the max document length.
+2. Construct an array ``docs`` containing the document ``{"x" : "y"}`` repeated 100 times.
+3. Insert ``docs`` to a collection via ``insertMany``.
+4. Inspect the resulting "command started" log message and assert that the "command" value is a string of length 1000 + (length of trailing ellipsis).
+5. Inspect the resulting "command succeeded" log message and assert that the "reply" value is a string of length <= 1000.
+6. Run ``find()`` on the collection where the document was inserted.
+7. Inspect the resulting "command succeeded" log message and assert that the reply is a string of length 1000 + (length of trailing ellipsis).
```
BorisDog, line 39: minor: Is there a value in having the second test, as opposed to a parameterizing the first one?

source/unified-test-format/unified-test-format.rst
jmikola: Recapping the three assertions you presented above:

> 1. That an error code, error code name, and error labels are present/non-redacted.

Noted that this is already covered by existing syntax, as you said.

> 2. That no other properties are present/unredacted.

I agree that this does sound brittle. If you were to stick with a proprietary `hasFailure` on `expectedLogMessage`, does that avoid this concern at all?

> 3. That the attached server response only contains an error code, error code name, and error labels.

To my last question, I wonder if `expectedLogMessage.hasFailure` saying something to the effect of "the log message indicates a failure and MUST also assert that all fields other than error code, name, and labels are redacted."

Although even with that, asserting whether redaction have been made seems independent of whether the log indicates an error. I understand if there's really no other way to address that in test format syntax, though, without some additional field like `isFailureRedacted` that would always accompany `hasFailure` and be `true` (both verbose and redundant).

> Add isRedacted to expectedError, and have isRedacted make assertions 2 and 3 above.

Reading this, I'm not sure that allowing `expectedError` to be used within `expectedLogMessage` makes sense. Do you actually need to assert the logged error's specifics (e.g. code, name, labels) for this specification? Is that a nice-to-have? Or does the logging spec only care about whether there is an error and it's sufficiently redacted?

I'm wondering how other fields like `expectedError.expectResult` would work. I imagine the functionality isn't needed, but it'd be permitted by the schema and could complicate implementations. Likewise if other fields are later added to `expectedError` that don't make sense in the context of logging spec tests.

To that end, a basic `hasFailure` field with a loaded definition (and assertion logic) seems preferable, since it'd be fully contained to `expectedLogMessage` and less likely to conflict with anything else in a test runner.

----

> 2. Add isRedacted to expectedError, and have isRedacted assert that the error object has a code, codeName, and errorLabels, but no other properties. Have the expected server response being added in [DRIVERS-2385](https://jira.mongodb.org/browse/DRIVERS-2385) be matched as non-root by default, and use that for assertion 3.

My expectation (no pun intended) is that the server-side error document asserted for [DRIVERS-2385](https://jira.mongodb.org/browse/DRIVERS-2385) would be matched as a root document no different than the existing command/response assertions for APM events. I'll note that CommandFailedEvent doesn't have any assertion syntax for the `failure` property, but that's mainly because the Command Monitoring spec is very permissive about its type (i.e. string, exception, or document).

That agrees with your assumption from earlier in the comment:

> But, it does seem that in most cases we would probably want treat this expected server response as a root document, similar to how the reply in a CommandSucceededEvent is always matched as a root document.

----

> Aside: the command monitoring equivalent ticket around testing redaction of failures is [DRIVERS-2394](https://jira.mongodb.org/browse/DRIVERS-2394). Ideally whatever we come up with here to test log message failure redaction would be mostly reusable there.

I haven't given that ticket much thought before, but I imagine _that_ would be a case where a field like `isRedacted` would be necessary, since we don't have a reliable way to assert the possible types of CommandFailedEvent's attached error. In that case, `isRedacted` could be flexible enough to say "assert that whatever error type a driver uses to represented an error is redacted" and perhaps reference some common redaction rules already written for `expectedLogMessage.hasFailure`.

kmahar: So I went and tried to implement the "partial" (i.e. only redact certain properties) redaction discussed above in the Rust driver today. One thing I ran into is that, in our tracing events (essentially structured log messages) we can only attach primitive types to messages - strings plus a handful of numeric types. This means that errors will have to be represented as strings. Note that I already discuss that errors can be represented flexibly in log messages in the spec - see section "Representing Errors in Log Messages" - which is consistent with command monitoring.

In addition to the points you make above, I think this is another argument against my previous proposal to use `expectedError`, since we would not actually have an error object to assert against anyway. I also imagine that we are not the only drivers that might find ourselves in this situation. 

> Do you actually need to assert the logged error's specifics (e.g. code, name, labels) for this specification? Is that a nice-to-have? Or does the logging spec only care about whether there is an error and it's sufficiently redacted?

I would consider it more a nice-to-have than a necessity. We could assert on the error more thoroughly by expecting an error for the corresponding operation. Asserting presence and redaction/lack thereof seems likely good enough.

> Although even with that, asserting whether redaction have been made seems independent of whether the log indicates an error. I understand if there's really no other way to address that in test format syntax, though, without some additional field like isFailureRedacted that would always accompany hasFailure and be true (both verbose and redundant).

I think there is actually a use case for isFailureRedacted=false, which I mentioned above: non-command errors should _not_ be redacted. Maybe we should circle back around to Jeff's proposal at the beginning of this thread:

> Consider simplifying the boolean by making this one just about failure and a second one that is just about redaction.

We could use two booleans:
* `hasFailure`: this would probably only ever be true; when present we assert there is a failure attached to the message
* `isFailureRedacted`: this should only ever be specified when `hasFailure=true`. if false, assert failure is _not_ redacted (we would use this to test e.g. network errors are not redacted); if true assert failure _is_ redacted. 

This does still leave us with a couple problems: 
* brittle assertions around whether an error actually got redacted - a driver could forget to update the runner to assert that some new field is redacted once it is added. That said, I'm not sure drivers adding new fields to their `CommandError` types is a common occurrence at all (probably the last time was when error labels came to exist?) so this might not be worth worrying about.
* difficulty of asserting redaction for languages that represent errors as strings. In Rust for example an error string representation might look something like `CommandError { code: 1, code_name: "CodeNameHere", message: "Message here", labels: [] }` and so to assert redaction we'd have to do some kind of string parsing that verifies particular parts of that string are not there / replaced by some placeholder. That said, that doesn't seem like a big issue. We would only be using this on a single error type for now (command errors) and writing a regex to match only a command error and parse out its constituent parts would be straightforward. 

So I think I'm inclined to go something more like that route now. What do you think? @jyemin  / @BorisDog, how will the errors you include in log messages be represented (for Java, what is the intermediate representation before stringifying?) and would it be feasible to assert redaction/lack of redaction on that representation?

jmikola: I missed @jyemin's original suggestion when responding to your previous comment, but having two separate fields seems entirely reasonable. You should also be able to enforce that `isFailureRedacted` is only specified when `hasFailure: true` is specified -- I believe that would be similar to how we assert that an `operation` doesn't mix `expectError` with `expectResult`/`saveResultAsEntity`.

> brittle assertions around whether an error actually got redacted

I may be missing something else, but is there any proposal that avoids this concern? It seems like it'd always be an issue and there's no foolproof way to prevent a driver from making a CVE-worthy mistake.

> difficulty of asserting redaction for languages that represent errors as strings

This may be another misunderstanding on my part, but I also don't see how that can be avoided if a driver is limited to using a string type and the tests ultimately need to assert that redaction has taken place.

source/command-logging-and-monitoring/tests/README.rst
```diff
@@ -0,0 +1,57 @@
...
+=======
+
+Automated Tests
+^^^^^^^^^^^^^^^
+There are tests in the `Unified Test Format <../../unified-test-format/unified-test-format.rst>`__ for both logging and
+monitoring in `/logging </logging>`_ and `/monitoring </monitoring>`_, respectively.
+
+
+Prose Tests
+~~~~~~~~~~~
```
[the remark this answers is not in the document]
kmahar: yes, thanks - fixed.

source/command-logging-and-monitoring/tests/README.rst
```diff
@@ -0,0 +1,59 @@
...
+
+--------
+
+Testing
+=======
+
+Automated Tests
+^^^^^^^^^^^^^^^
+There are tests in the `Unified Test Format <../../unified-test-format/unified-test-format.rst>`__ for both logging and
+monitoring in `/logging </logging>`_ and `/monitoring </monitoring>`_, respectively. Drivers MUST run the logging
```
kmahar, line 18: fyi @BorisDog: added a sentence here about increasing the max document length when running these tests.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1341,6 +1364,65 @@ MUST assert that the field is not set, or, if the driver uses a nonpositive Int3
...
 
+expectedLogMessagesForClient
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+A list of log messages that are expected to be observed (in that order) for a 
+client while executing `operations <test_operations_>`_.
+
+The structure of each object is as follows:
+
+- ``client``: Required string. Client entity on which the events are expected
```
BorisDog, line 1375: `log messages` instead if `events`?

kmahar: fixed, thank you!

source/unified-test-format/unified-test-format.rst
kmahar: changes have been pushed to switch to the two-boolean approach (see my top-level reply below for some related details)

kmahar:
Just pushed a bunch of changes based on recent conversations. To summarize:

* Made clear in `logging.rst` that drivers should “naively” truncate JSON, i.e. they should not try to ensure the output is valid JSON. Also added rationale section justifying this decision. Both @BorisDog and @patrickfreed asked about this. https://github.com/mongodb/specifications/pull/1303/commits/2f41ac7b37695e3727a5ecdf7e47212c7b5656de
* Added a requirement in `logging.rst` that drivers use the exact names that are specified for values in structured log messages and do not e.g. use snake case instead of camel case. https://github.com/mongodb/specifications/pull/1303/commits/830c568944689356ebb5f54aa8683a17a3dc6b16
* Per @BorisDog’s question on Slack, made explicit in `unified-test-format.rst` that log messages produced by a client must be an exact match to the expected ones. (I haven’t found a need yet for drivers to ignore extra messages but we could add something akin to `ignoreExtraEvents`  if this comes up in the future. https://github.com/mongodb/specifications/pull/1303/commits/5bacb00da810fd2be8188a125976e11d6fa423b0
* Realized that all of my “invalid” tests for the format changes also needed a schema version bump that I missed when I rebased before. Updated those from 1.11 to 1.12. https://github.com/mongodb/specifications/pull/1303/commits/3a22c27ef14ed75d940c8b0b379c7594762bb126
* Switched to the discussed two-boolean approach for asserting that failures are present/not present and optionally are redacted. Updated relevant language in `expectedLogMessage` , updated JSON schema, and added new “invalid” tests to test the schema. Updated existing tests to use this approach. Also, added a number of new tests which check we do not redact client-side errors in the “command failed” log messages for sensitive commands. These are implemented by setting the `failCommand` failpoint with `closeConnection: true` for the relevant commands which should generate I/O exceptions. These are not intended to exhaustively cover all error types, and I also skipped one sensitive command (hello w/ speculative authenticate) because it seemed difficult to test in a way that would not be racy due to interference with hello commands sent for server monitoring. (https://github.com/mongodb/specifications/pull/1303/commits/5b0224bf211c55cc633b24913cb53f4173de4c0b, https://github.com/mongodb/specifications/pull/1303/commits/e2c7209ecc768f8c66b68e8cc6397b5163460615)

source/unified-test-format/schema-1.12.json
```diff
@@ -0,0 +1,692 @@
...
+        "additionalProperties": false,
+        "required": ["level", "component", "data"],
+        "properties": {
+            "level": { "$ref": "#/definitions/logSeverityLevel" },
+            "component": { "$ref": "#/definitions/logComponent" },
+            "data": { "type": "object" },
+            "hasFailure": {"type": "boolean" },
+            "failureIsRedacted": { "type": "boolean" }
+        },
+        "anyOf": [
```
kmahar, line 567: paging @jmikola. what I'm trying to express here is that any of the following are valid:
* specifying neither `hasFailure` nor `failureIsRedacted`
* specifying `hasFailure` (either true or false) and not specifying `failureIsRedacted`
* specifying `hasFailure=true` and `failureIsRedacted` (either true or false)

but that it is disallowed to specify `failureIsRedacted` without `hasFailure=true`. 

relatedly, I debated whether it made sense to allow `hasFailure` without `failureIsRedacted`, since it is redundant with specifying `failure: { $$exists: true }` in `data`. I went with allowing it for now, but if you think it would be better to only allow it along with `failureIsRedacted` I have no objections to being more restrictive. 

I also don't have a use case for `hasFailure: false` right now so we could consider prohibiting that too.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1336,6 +1363,76 @@ MUST assert that the field is not set, or, if the driver uses a nonpositive Int3
...
+
+The structure of each object is as follows:
+
+- ``client``: Required string. Client entity for which the messages are expected
+  to be observed. See `commonOptions_client`_.
+
+- ``messages``: Required array of `expectedLogMessage`_ objects. List of 
+  messages, which are expected to be observed (in this order) on the corresponding
+  client while executing `operations`_. If the array is empty, the test runner
+  MUST assert that no messages were observed on the client. The driver MUST assert
```
kmahar, line 1380: @BorisDog: added sentence here re: exact match of messages.

source/unified-test-format/schema-1.12.json
jmikola: If the three cases above are mutually exclusive, would it make sense to use `oneOf` instead? I trust these rules are all tested with "invalid" tests so I won't dig too much into the JSON schema syntax here.

Since `data` is a required field and you mentioned `failure: { $$exists: true }` is equivalent with `hasFailure: true`, what does `hasFailure` give us apart from the ability to enforce use of `failureIsRedacted` in the schema?

Noting the following asks from https://github.com/mongodb/specifications/pull/1303/files#r981488414:

> 1. That an error code, error code name, and error labels are present/non-redacted.
> 2. That no other properties are present/unredacted.
> 3. That the attached server response only contains an error code, error code name, and error labels.

If there isn't a use case for `hasFailure: false` and `failureIsRedacted` is only allowed when `hasFailure: true`, I wonder why we can't stick to single `failureIsRedacted` boolean and make it a runtime error if the assertion field is used and there isn't a failure present (IIUC that would mean `data` omits a `failure` field).

source/unified-test-format/unified-test-format.rst
jyemin: LGTM, but I won't resolve this thread myself since there are so many commenters.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1407,13 +1407,21 @@ The structure of each object is as follows:
 
 - ``hasFailure``: Optional boolean. When this field is set to ``true``, 
   the test runner MUST assert that the log message has a ``failure`` value
-  in its attached data, and that the failure has not been redacted in any
-  way. When this field is set to ``false``, the test runner MUST assert
-  that the log message either has no ``failure`` value at all or that it
-  has a redacted failure, e.g. an empty string or object. 
-  The exact form of those assertions will vary based on the driver's chosen
-  representations for failures and redacted failures in log messages.
+  in its attached data. Then this field is set to ``false``, the test runner
```
jyemin, line 1410: Then -> When

source/logging/logging.rst
```diff
@@ -348,6 +358,17 @@ Truncation of large documents
...
 
+3. Why do we implement naive truncation rather than truncating the JSON so it is still valid?
+    Designing and implementing a truncation algorithm for JSON that outputs valid JSON, but fits
+    in as much of the original JSON as possible, would be non-trivial. The server team wrote an
+    entire separate truncation design document when they implemented this for their log messages.
+    This is more of a necessity for the server where the entire log message is JSON, but we don't
+    know if parsing the documents included in log messages is something that users will actually
+    need to do. Furthermore, any users who want parseable documents have an escape hatch to do so:
+    they can set the max document length to a very large value. If we hear of use cases in the Future
+    for parsing the documents in log messages, we could make an additive change to this specification
```
jyemin, line 369: Future -> future

source/unified-test-format/schema-1.12.json
kmahar: you're right that `oneOf` would make more sense here. 

> what does `hasFailure` give us apart from the ability to enforce use of `failureIsRedacted` in the schema?

nothing I can think of besides that. 

> I wonder why we can't stick to single `failureIsRedacted` boolean and make it a runtime error if the assertion field is used and there isn't a failure present (IIUC that would mean `data` omits a `failure` field).

this is close to what I initially had in the first place (the single boolean `hasFailure`), though better named. I would be fine with that approach, but switched to this since Jeff had suggested two booleans and I thought you were advocating for it as well in the thread below for schema enforcement purposes, hence the switch. that said, there was a lot of back and forth down there and I think the fact that `failure: { $$exists: true/false }` already covered failure presence/absence likely got lost in the mix.

wdyt @jyemin?

source/logging/logging.rst
kmahar: fixed, thanks.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1336,6 +1363,76 @@ MUST assert that the field is not set, or, if the driver uses a nonpositive Int3
...
+  This value is often used in tandem with ``failureIsRedacted``.
+
+- ``failureIsRedacted``: Optional boolean. This field MUST only be specified
+  when ``hasFailure`` is present and its value is ``true``. When ``true``,
+  the test runner MUST assert that the failure has been redacted according to
+  the rules defined for error redaction in the `command logging and monitoring
+  specification <../command-logging-and-monitoring/command-logging-and-monitoring.rst#security>`__.
+  When ``false``, the test runner MUST assert that the failure has NOT been
+  redacted.
+  The exact form of these assertions and how thorough they are will vary based
```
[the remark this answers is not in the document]
kmahar: fixed, thanks.

source/unified-test-format/schema-1.12.json
kmahar: ok, I've switched to @jmikola's suggestion of a single `failureIsRedacted` boolean plus having drivers check at runtime that it is only used in combination with `failure: { $$exists: true }`

source/unified-test-format/unified-test-format.rst
kmahar: thanks; ended up deleting this per above thread about using a single boolean

kmahar:
I'm confused about the "Check if all JSON test files are up-to-date" failure. it seems to be complaining about a CMAP test I haven't touched here, and the "diff" is a line that looks identical before and after. it doesn't seem relevant
```
diff --git a/source/connection-monitoring-and-pooling/tests/pool-clear-interrupt-immediately.json b/source/connection-monitoring-and-pooling/tests/pool-clear-interrupt-immediately.json
index e61ca51..54e2566 100644
--- a/source/connection-monitoring-and-pooling/tests/pool-clear-interrupt-immediately.json
+++ b/source/connection-monitoring-and-pooling/tests/pool-clear-interrupt-immediately.json
@@ -24,7 +24,7 @@
       "name": "waitForEvent",
       "event": "ConnectionPoolCleared",
       "count": 1,
-      "timeout": 1000      
+      "timeout": 1000
     },
     {
       "name": "waitForEvent",
```

source/unified-test-format/unified-test-format.rst
```diff
@@ -1405,20 +1405,18 @@ The structure of each object is as follows:
...
-- ``hasFailure``: Optional boolean. When this field is set to ``true``, 
-  the test runner MUST assert that the log message has a ``failure`` value
-  in its attached data. Then this field is set to ``false``, the test runner
-  MUST assert has no ``failure`` value in its attached data.
-  Currently, this assertion is only relevant for "command failed" log messages.
-  This value is often used in tandem with ``failureIsRedacted``.
-
-- ``failureIsRedacted``: Optional boolean. This field MUST only be specified
-  when ``hasFailure`` is present and its value is ``true``. When ``true``,
+- ``failureIsRedacted``: Optional boolean. This field SHOULD only be specified
```
jyemin, line 1408: Nit: not sure how to interpret this combination of SHOULD and MUST.  Why aren't they both the same?

kmahar: I agree it feels funny; this comes back to the unified format's design rationale section on SHOULD vs MUST.

> This specification commonly uses "SHOULD" when providing guidance on writing test files. While this may appear contradictory to the driver mantra preferring "MUST", it is intentional. Some of this guidance addresses style (e.g. adding comments, using YAML anchors) and cannot be enforced with a JSON schema... The specification does prefer "MUST" in other contexts, such as discussing parts of the test file format that are enforceable by the JSON schema or the test runner implementation.

the "SHOULD" is used to describe something the schema allows but people shouldn't do (use `failureIsRedacted` without also asserting the failure `$$exists`) whereas the MUST describes something the test runner does.

jyemin: Gotcha

source/unified-test-format/unified-test-format.rst
```diff
@@ -1336,6 +1363,76 @@ MUST assert that the field is not set, or, if the driver uses a nonpositive Int3
...
+- ``component``: Required string. This MUST be one of the component names listed
+   in `components <../logging/logging.rst#components>`__. This specifies the
+   expected component for the log message. Note that since naming variations
+   are permitted for components, some drivers may need to map this to a
+   corresponding language-specific component name. Test runners MUST assert
+   that the actual component matches this value.
+
+- ``failureIsRedacted``: Optional boolean. This field SHOULD only be specified
+  when ``data`` contains ``{ failure: { $$exists: true }}``; the test runner
+  MUST report an error if this is not the case.
```
jmikola, line 1410: I didn't expect this would require a subsequent assertion on the _contents_ of `data` field in the test file. Rather, the mere presence of `failureIsRedacted` would require a driver to assert that there is some failure to inspect in the first place. In other words, a test could use `failureIsRedacted` on its own with no meaningful `data` assertion and trust the assertion would fail outright if the log data omitted a failure altogether.

I imagine there might also be cases where `data.failure` could be something other than `{ $$exists: true }` (perhaps a `$$type` assertion or something more specific.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1336,6 +1363,76 @@ MUST assert that the field is not set, or, if the driver uses a nonpositive Int3
...
+
+  When ``hasFailure`` is present and its value is ``true``,
```
jmikola, line 1412:
```suggestion
  When ``failureIsRedacted`` is present and its value is ``true``,
```

source/unified-test-format/unified-test-format.rst
```diff
@@ -1336,6 +1363,76 @@ MUST assert that the field is not set, or, if the driver uses a nonpositive Int3
...
+  the test runner MUST assert that the failure has been redacted according to
+  the rules defined for error redaction in the `command logging and monitoring
+  specification <../command-logging-and-monitoring/command-logging-and-monitoring.rst#security>`__.
+
+  When ``false``, the test runner MUST assert that the failure has NOT been
+  redacted.
```
jmikola, line 1418: To my earlier comment, you could also just change the behavior for handling `true` and `false`. For instance, this could be:

> When `failureIsRedacted` is specified and ``false``, the test runner MUST assert that a failure is present and has NOT been redacted.

And likewise:

> When `failureIsRedacted` is specified and `true`, the test runner MUST assert that a failure is present and...

source/unified-test-format/unified-test-format.rst
kmahar: ah, I misunderstood your comment above. and good point re: other assertions on `failure` being possible. 

I updated the language here to just say 

> This field SHOULD only be specified when the log message data is expected to contain a ``failure`` value.

source/unified-test-format/unified-test-format.rst
kmahar: done

jmikola (approved):

kmahar:
#1316 bumped the unified format schema version; just pushed another commit to reflect this so that the logging changes are added via schema version 1.13.

jyemin (approved):
LGTM!!!!

source/command-logging-and-monitoring/command-logging-and-monitoring.rst
```diff
@@ -43,70 +44,69 @@ Documentation
...
 
-The driver MUST guarantee that the ``requestId`` of the ``CommandStartedEvent`` and the corresponding ``CommandSucceededEvent`` or ``CommandFailedEvent`` is the same.
+The driver MUST guarantee that the ``requestId`` of the ``CommandStartedEvent`` and the corresponding ``CommandSucceededEvent`` or ``CommandFailedEvent`` is the same, and that the ``requestId`` of
+the "command started" log message and the corresponding "command succeeded" or "command failed" log message is the same.
 
 Unacknowledged/Acknowledged Writes
 ----------------------------------
 
- Unacknowledged writes must provide a ``CommandSucceededEvent`` with a ``{ ok: 1 }`` reply.
+ Unacknowledged writes must provide a ``CommandSucceededEvent`` and a "command succeeded" log message with a ``{ ok: 1 }`` reply .
```
BorisDog, line 80: nit: redundant space in `reply .`?

source/command-logging-and-monitoring/command-logging-and-monitoring.rst
```diff
@@ -340,42 +359,166 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
...
+     - Value
+
+   * - message
+     - String
+     - "Command started"
+
+   * - command
+     - String
+     - Relaxed extJSON representation of the command. This document MUST be truncated appropriately according to rules defined in the 
+       `logging specification <../logging/logging.rst>`_, and MUST be replaced with an empty document "{ }" if the command is considered sensitive.
```
BorisDog, line 446: For your consideration, you could add a link to the truncation section #truncation-of-large-documents.

BorisDog (approved):
LGTM (2 minor comments)

source/command-logging-and-monitoring/command-logging-and-monitoring.rst
kmahar: removed. thanks!

source/command-logging-and-monitoring/command-logging-and-monitoring.rst
kmahar: good thought to add a link. that section covers truncation rationale, but I am going to add one to the section on truncation rules: Configurable Max Document Length.

PR #1307, opened 2022-09-14
DRIVERS-2443 Add serverless: forbid to change stream clusterTime test

abr-egn:
- [ ] ~Bump spec version and last modified date.~
- [ ] ~Update changelog.~
- [ ] ~Test these changes against all server versions and topologies (including standalone, replica set, sharded clusters, and serverless).~

### DIFF
--- a/source/change-streams/tests/unified/change-streams-clusterTime.json
+++ b/source/change-streams/tests/unified/change-streams-clusterTime.json
@@ -1,6 +1,6 @@
 {
   "description": "change-streams-clusterTime",
-  "schemaVersion": "1.3",
+  "schemaVersion": "1.4",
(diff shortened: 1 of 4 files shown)


benjirewis (approved):
LGTM. The last test file that is missing `serverless: forbid` if I'm not mistaken is `showExpandedEvents`. Should we add it to those test files as well?

benjirewis (changes requested):
I think you'll have to bump to schema version 1.4 to get the `serverless: forbid` run on requirement.

abr-egn:
Thanks, bumped the schema, and agreed that we may as well pick up the other test as well.

benjirewis (approved):

PR #1308, opened 2022-09-26
DRIVERS-2449: Do not connect to mongocryptd if shared library is loaded

DmitryLukyanov:
- [ ] Update changelog.

### DIFF
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -1444,6 +1444,9 @@ is propagated to the user.
 Connecting to mongocryptd_
 --------------------------
 
+If the crypt_shared_ library is loaded, the driver MUST NOT attempt to connect
+to mongocryptd_. (Refer: `Detecting crypt_shared Availability`_).
+
 Single-threaded drivers MUST connect with `serverSelectionTryOnce=false <../server-selection/server-selection.rst#serverselectiontryonce>`_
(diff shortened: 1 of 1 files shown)


kevinAlbs (commented):
The changes have been implemented in Go.

Here is a [proposed prose test](https://github.com/mongodb/mongo-go-driver/pull/1082/files#diff-55617555c17bae6fadeb3e9a0c0e3a10837496998c75937eba67712b871f58eeR1994-R2036). It requires starting a connection listener in a separate goroutine. Does this test look OK for C#?

kevinAlbs:
Also, please "Bump spec version and last modified date." and "Update changelog".

DmitryLukyanov:
> The changes have been implemented in Go.
> 
> Here is a [proposed prose test](https://github.com/mongodb/mongo-go-driver/pull/1082/files#diff-55617555c17bae6fadeb3e9a0c0e3a10837496998c75937eba67712b871f58eeR1994-R2036). It requires starting a connection listener in a separate goroutine. Does this test look OK for C#?

looks interesting, I will check whether it will work.

DmitryLukyanov:
Yes, I think it should work, this is how it will look in c# https://github.com/DmitryLukyanov/mongo-csharp-driver/tree/drivers2449. I will update this PR on the next week

kevinAlbs:
Closing in favor of https://github.com/mongodb/specifications/pull/1330

PR #1311, opened 2022-10-05
DRIVERS-2253: Remove spec front matter and reformat changelogs

jmikola:
https://jira.mongodb.org/browse/DRIVERS-2253

This limits front matter fields to "Status" and "Minimum Server Version".

Spec versions were removed across the board, excluding the Unified Test Format, which retained a field for "Current Schema Version". The only other spec where versioning was somewhat relevant was Extended JSON, and the transition from 1.0 to 2.0 is clearly noted in the changelog (minor versions since 2.0 were not relevant).

Apart from the deprecated Bulk API spec, all specs have an "Accepted" status. Some were incorrectly listed as "Draft".

This also revises changelogs across all specs to use "Field List" formatting (like spec front matter). The CSFLE and Change Stream specs were left as-is, since they used CSV and grid tables, respectively. Some existing changelog entries were incorrectly ordered and fixed. Lastly, specs were inconsistent in whether changelogs entries used ascending or descending date order, so existing order was left in place when possible. For some very old specs without dated changelog entries, dates were inferred from git history.

- [x] Update changelogs

### DIFF
--- a/source/change-streams/change-streams.rst
+++ b/source/change-streams/change-streams.rst
@@ -2,15 +2,8 @@
 Change Streams
 ==============
 
-:Title: Change Streams
-:Author: Matt Broadstone
-:Advisory Group: David Golden, A. Jesse Jiryu Davis
-:Approvers: Jeff Yemin, A. Jesse Jiryu Davis, Bernie Hackett, David Golden, Eliot
 :Status: Accepted
-:Type: Standards
 :Minimum Server Version: 3.6
-:Last Modified: 2022-08-22
-:Version: 1.18
 
 .. contents::
 
@@ -1099,4 +1092,6 @@ Changelog
 | 2022-08-17 | Support `disambiguatedPaths` in `UpdateDescription`.       |
 +------------+------------------------------------------------------------+
 | 2022-08-22 | Added ``clusterTime`` to ``ChangeStreamDocument``.         |
++------------+------------------------------------------------------------+
+| 2022-10-05 | Remove spec front matter.                                  |
 +------------+------------------------------------------------------------+
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -535,8 +535,6 @@ accept arbitrary strings at runtime for forward-compatibility.
 Automatic Credentials
 `````````````````````
 
-.. versionadded:: 1.9.0 2022/06/22
-
 Certain values of KMSProviders_ indicate a request by the user that the
 associated KMS providers should be populated lazily on-demand. The driver MUST
 be able to populate the respective options object on-demand if-and-only-if such
@@ -599,8 +597,6 @@ __ ../auth/auth.html#obtaining-credentials
 Obtaining GCP Credentials
 `````````````````````````
 
-.. versionadded:: 1.11.0 2022/07/20
-
 Set ``HOST`` to ``metadata.google.internal``.
 
 Send an HTTP request to the URL
@@ -622,8 +618,6 @@ Return "access_token" as the credential.
 
 Obtaining an Access Token for Azure Key Vault
 `````````````````````````````````````````````
-
-.. versionadded:: 1.12.0 2022/08/30
 
 Virtual machines running on the Azure platform have one or more *Managed
 Identities* associated with them. From within the VM, an identity can be used by
@@ -2513,7 +2507,7 @@ Changelog
    :align: left
 
    Date, Description
-   22-10-05, Remove spec front matter.
+   22-10-05, Remove spec front matter and ``versionadded`` RST macros (since spec version was removed)
    22-09-26, Add behavior for automatic Azure KeyVault credentials for ``kmsProviders``.
    22-09-09, Prohibit ``rewrapManyDataKey`` with libmongocrypt <= 1.5.1.
    22-07-20, Add behavior for automatic GCP credential loading in ``kmsProviders``.
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -2,16 +2,8 @@
 Client Side Encryption
 ======================
 
-:Title: Client Side Encryption
-:Author: Kevin Albertson
-:Spec Lead: Jeff Yemin
-:Approvers: Bernie Hackett, David Golden, David Storch
-:Advisory Group: A\. Jesse Jiryu Davis, Kenn White, Scott L'Hommedieu, Mark Benvenuto, Bernie Hackett, Samantha Ritter, Matt Broadstone
 :Status: Accepted
-:Type: Standards
 :Minimum Server Version: 4.2 (CSFLE), 6.0 (Queryable Encryption)
-:Last Modified: 2022-09-26
-:Version: 1.12.0
 
 .. _lmc-c-api: https://github.com/mongodb/libmongocrypt/blob/master/src/mongocrypt.h.in
 
@@ -2521,6 +2513,7 @@ Changelog
    :align: left
 
    Date, Description
+   22-10-05, Remove spec front matter.
    22-09-26, Add behavior for automatic Azure KeyVault credentials for ``kmsProviders``.
    22-09-09, Prohibit ``rewrapManyDataKey`` with libmongocrypt <= 1.5.1.
    22-07-20, Add behavior for automatic GCP credential loading in ``kmsProviders``.
--- a/source/connection-string/connection-string-spec.rst
+++ b/source/connection-string/connection-string-spec.rst
@@ -5,14 +5,8 @@
 Connection String Spec
 ======================
 
-:Spec: 104
-:Title: Connection String Spec
-:Authors: Ross Lawley
-:Advisors: \A. Jesse Jiryu Davis, Jeremy Mikola, Anna Herlihy
-:Status: Approved
-:Type: Standards
-:Last Modified: 2019-04-26
-:Version: 1.5.0
+:Status: Accepted
+:Minimum Server Version: N/A
 
 .. contents::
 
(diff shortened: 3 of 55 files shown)


source/change-streams/change-streams.rst
```diff
@@ -1100,3 +1093,5 @@ Changelog
 +------------+------------------------------------------------------------+
 | 2022-08-22 | Added ``clusterTime`` to ``ChangeStreamDocument``.         |
 +------------+------------------------------------------------------------+
+| 2022-10-05 | Remove spec front matter.                                  |
```
alcaeus, line 1096: I'll note that the format of this changelog is different from all others.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -2521,6 +2513,7 @@ Changelog
    :align: left
 
    Date, Description
+   22-10-05, Remove spec front matter.
```
alcaeus, line 2516: Changelog format is different from all others.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-summary.rst
```diff
@@ -221,3 +216,8 @@ and single-threaded clients MUST request a scan before the next operation.
 The client MUST clear its connection pool for the server if the
 server is 4.0 or earlier, and MUST NOT clear its connection pool for the
 server if the server is 4.2 or later.
+
+Changelog
+---------
```
alcaeus, line 221: Formatting is inconsistent with other files

```suggestion
=========
```

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -2483,92 +2480,61 @@ Mathias Stearn's beautiful design for replica set monitoring in mongos 2.6
...
-
-2021-05-03: Require parsing "isWritablePrimary" field in responses.
-
-2021-06-09: Connection pools must be created and eventually marked ready for any server if a direct connection is used.
-
-2021-06-29: Updated to use modern terminology.
-
-2022-01-19: Add iscryptd and 90th percentile RTT fields to ServerDescription.
+Changelog
+---------
```
alcaeus, line 2484: Formatting is inconsistent with other files

```suggestion
=========
```

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -1143,29 +1139,22 @@ awaitable hello or legacy hello heartbeat in the new protocol.
 Changelog
 ---------
```
alcaeus, line 1140: Formatting is inconsistent with other files
```suggestion
=========
```

alcaeus (approved):
Changes LGTM. I noticed a few inconsistencies, but those are nits and can be ignored if you'd like.

I'll note that some changelogs are sorted ascending and others descending, but I'm happy to leave those in place or defer them to a later PR to get the bulk of these changes out of the way.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-summary.rst
jmikola: This is intentional, as the RST file uses `---` as its top-level heading syntax.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
jmikola: Also intentional, as the file uses `---` as its top-level heading syntax.

source/change-streams/change-streams.rst
jmikola: Indeed. I called this out in the commit message along with the CSFLE changelog syntax.

There was enough effort put into both files that I reasoned it wasn't worth changing them in this PR.

source/client-side-encryption/client-side-encryption.rst
jmikola: Same reasoning as https://github.com/mongodb/specifications/pull/1311/files#r987580269

This and change streams were intentionally left as-is.

source/connection-string/connection-string-spec.rst
```diff
@@ -375,20 +369,22 @@ Q: Why must the auth database be URL decoded by the parser?
...
-- 2018-01-09: Clarified that space characters should be encoded to ``%20``.
-- 2018-06-04: Revised Userinfo section to provide an explicit list of allowed
-  characters and clarify rules for exceptions.
-- 2019-02-04: In Repeated Keys section, clarified that the URI options spec may
-  override the repeated key behavior described here for certain options.
-- 2019-03-04: Require drivers to document option precedence rules
-- 2019-04-26: Database name in URI alone does not trigger authentication
-- 2020-01-21: Clarified how empty values in a connection string are parsed.
+Changelog
+=========
```
jmikola, line 373: This moves "Changelog" to a top-level heading (same as "Abstract"). The entire spec seems poorly formatted in this regard as everything falls under "Abstract" -- just seemed easy to fix this one case while I was revising the changelog.

source/driver-bulk-update.rst
```diff
@@ -2,43 +2,11 @@
 Bulk API Spec
 =============
 
-:Authors: Christian Kvalheim
 :Status: Deprecated
-:Type: Standards
-:Last Modified: May 18, 2021
+:Minimum Server Version: 2.4
```
jmikola, line 6: I inferred this from references to the server version elsewhere in the spec. IIRC, the bulk update spec was written before the actual write commands were implemented in 2.6 (which is what the related "server write commands" spec uses). That said, it probably doesn't matter too much since this spec is deprecated.

source/enumerate-databases.rst
```diff
@@ -2,14 +2,8 @@
 Enumerating Databases
 =====================
 
-:Spec Title: Enumerating Databases
-:Spec Ticket: SPEC-865
-:Spec Version: 1.0
-:Author: Jeremy Mikola
 :Status: Accepted
-:Type: Standards
 :Minimum Server Version: 3.8
```
jmikola, line 6: I don't recall why this server version was used, but I wasn't confident enough to change it. The most recent git blame on the file had it bumped from 3.7 to 3.8 in https://github.com/mongodb/specifications/commit/3d13d39aea4eeed590989cf6fd649d09c0b72cd2#diff-be10ea70ab383a03cf3f821429f4782517fcbbef5a82b9ff43a9e44a04fd3e79R11. It was originally 3.6 dating back to the original spec version, before being bumped to 3.7 by https://github.com/mongodb/specifications/commit/4d12e63038f14c961498002ddf808db4e5e2dbdf for the `authorizedDatabases` option.

That said, this probably doesn't matter too much.

source/extended-json.rst
```diff
@@ -2,16 +2,8 @@
...
-:Spec: 587
-:Spec-ticket: SPEC-587
-:Title: Extended JSON
-:Author: Luke Lovett, David Golden
-:Spec Lead: David Golden, Jeff Yemin
-:Advisory Group: Jeff Yemin, Christian Kvalheim, Hannes Magnusson, Matt Broadstone, Jesse Davis
-:Status: Proposed
-:Type: Standards
-:Last Modified: May 21, 2021
-:Version: 2.2.0
```
jmikola, line 14: I considered keeping this as a front matter field like the unified test format's schema version, but it doesn't seem as relevant since we don't refer to this version from outside the spec (AFAIK). I made sure to add references to the original 1.0 version and bump to 2.0 in the changelog, but removed further mention of bumps to 2.x minor versions as they didn't seem relevant.

source/retryable-reads/retryable-reads.rst
```diff
@@ -314,6 +306,9 @@ The above requirement can be fulfilled in one of two ways:
    there is no guarantee that the lower versioned server can support the
    original command.
 
+.. [1] The first and second commands will be identical unless variations in
```
jmikola, line 309: This is one of two or three instances where I relocated footnotes higher up to the end of the section from where they were referenced. The goal was the keep the content closer to its original context and avoid having it down below the changelog, which should ideally be the last section in a spec.

source/transactions/transactions.rst
```diff
@@ -1416,8 +1407,10 @@ durable, which achieves the primary objective of avoiding duplicate commits.
 **Changelog**
```
jmikola, line 1407: I have no idea why this spec adds bold formatting to section titles, but it seemed like too much to change and removing it for this one section would make it inconsistent with itself.

source/unified-test-format/unified-test-format.rst
```diff
@@ -3748,114 +3743,91 @@ better processes can be established for versioning other specs *and* collating
 spec changes developed in parallel or during the same release cycle.
 
 
-Change Log
-==========
+Changelog
+=========
 
+..
+  Please note schema version bumps in changelog entries where applicable.
```
jmikola, line 3750: This seemed like the kosher way to add an RST comment: https://stackoverflow.com/a/4783854/162228

patrickfreed (approved):
CMAP and Server Selection LGTM

source/driver-bulk-update.rst
kmahar: 2.4 sounds reasonable, but I agree it doesn't really matter both because this spec is deprecated and because no one is supporting 2.4 or 2.6 anymore so the distinction isn't meaningful. I'm hoping we will delete this spec altogether and move any info that is still relevant into the new bulk spec via DRIVERS-716.

source/enumerate-databases.rst
kmahar: it looks like the `authorizedDatabases` option was added in MongoDB 4.0.5 (and 4.0 was originally numbered 3.8 iirc): https://www.mongodb.com/docs/manual/reference/command/listDatabases/#command-fields 

so perhaps that is why it was chosen (though, that same commit also notes that MongoDB 4.0.5 added the option, not 3.8).

in any case I agree this seems inconsequential.

kmahar (approved):

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -2,16 +2,8 @@
...
-:Spec Lead: Jeff Yemin
-:Approvers: Bernie Hackett, David Golden, David Storch
-:Advisory Group: A\. Jesse Jiryu Davis, Kenn White, Scott L'Hommedieu, Mark Benvenuto, Bernie Hackett, Samantha Ritter, Matt Broadstone
 :Status: Accepted
-:Type: Standards
 :Minimum Server Version: 4.2 (CSFLE), 6.0 (Queryable Encryption)
-:Last Modified: 2022-09-26
-:Version: 1.12.0
 
 .. _lmc-c-api: https://github.com/mongodb/libmongocrypt/blob/master/src/mongocrypt.h.in
```
kevinAlbs, line 8: Suggest removing the version from the `versionadded` directives in this file, since the version is no longer defined.

kevinAlbs (approved):
Client Side Encryption LGTM

source/server-discovery-and-monitoring/server-monitoring.rst
jmikola: This is consistent with the header syntax for "Abstract".

source/client-side-encryption/client-side-encryption.rst
jmikola: Ah, I had no idea that was used (GitHub doesn't apply any special rendering, but I imagine @vector-of-bool had some nice formatting when building with Sphinx in https://github.com/mongodb/specifications/commit/52759c75816a89f8ae6430b22cec9571451e6c7d).

Would have been nice if we started using that in the unified test format spec from the beginning, but it's probably a bit late for that now. And we do have schema changes highlighted in the changelog there, so probably not really necessary.

jmikola: Removed those macros in https://github.com/mongodb/specifications/pull/1311/commits/cc51ecc2590755dfb0ad6c5e08e330cbf2e083e4

source/enumerate-databases.rst
jmikola: Reverted to 3.6 in https://github.com/mongodb/specifications/pull/1311/commits/4a476a635d706ca757c47b5d806fe2bca6a79e23.

PR #1313, opened 2022-10-06
DRIVERS-555 Number CSOT prose tests and bump some timeouts.

benjirewis:

- [ ] Update changelog. (n/a)
- [ ] Make sure there are generated JSON files from the YAML test files. (n/a)

Numbers the CSOT prose tests for organization within drivers. Bumps the timeout of the `Multi-batch writes` test, restricts it to single topologies and uses 50 1MB docs instead of 100001 docs.

### DIFF
--- a/source/client-side-operations-timeout/tests/README.rst
+++ b/source/client-side-operations-timeout/tests/README.rst
@@ -57,7 +57,7 @@ bulk encoding performance.
        }
 
 #. Create a new MongoClient (referred to as ``client``) with ``timeoutMS=2000``.
-#. Using ``client``, insert 100,001 empty documents in a single ``insertMany`` call.
+#. Using ``client``, insert 50 1-megabyte documents in a single ``insertMany`` call.
 
    - Expect this to fail with a timeout error.
 
@@ -68,7 +68,7 @@ bulk encoding performance.
 
 This test MUST only be run against enterprise server versions 4.2 and higher.
 
-#. Launch a mongoryptd process on 23000.
+#. Launch a mongocryptd process on 23000.
 #. Create a MongoClient (referred to as ``client``) using the URI ``mongodb://localhost:23000/?timeoutMS=1000``.
 #. Using ``client``, execute the ``{ ping: 1 }`` command against the ``admin`` database.
 #. Verify via command monitoring that the ``ping`` command sent did not contain a ``maxTimeMS`` field.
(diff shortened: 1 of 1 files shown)


source/client-side-operations-timeout/tests/README.rst
```diff
@@ -31,10 +31,12 @@ during a test MUST be unset using ``internalClient`` after the test has been exe
...
 
-Multi-batch writes
-~~~~~~~~~~~~~~~~~~
+1. Multi-batch writes
+~~~~~~~~~~~~~~~~~~~~~
 
-This test MUST only run against server versions 4.4 and higher.
+This test MUST only run against standalones on server versions 4.4 and higher.
+The ``insertMany`` call takes an exceedingly long time on replicasets and sharded
+clusters.
```
benjirewis, line 39: See the [Go driver PR](https://github.com/mongodb/mongo-go-driver/pull/1081) implementing this prose test. In order for the first `insert` of the `insertMany` call to succeed, the `timeoutMS` value needs to allow for a high operation execution time after the blocking failpoint.

>On a standalone topology, [the first `insert` of the `insertMany` call] takes 274ms; on a replicaset topology, this takes 654 ms; on a sharded topology, this takes 971ms.

These times are even longer on other OSes (I'm on a Mac M2). I'm also suggesting restricting this test to standalones. In order to consistently successfully test against sharded clusters, I had to bump the blocking failpoint to ~3 seconds and `timeoutMS` to 6 seconds. That made for a seriously long prose test. The top of this section does say drivers are allowed to run CSOT prose tests solely against single-node replsets, so maybe this is redundant? Let me know what you think.

ShaneHarvey: What if you use w=1 instead of the implicit w:majority default?

benjirewis: Using `w=1` seems to shave off about 50-100ms depending on the topology. I could also require that, but I think we should still bump the timeout.

benjirewis: > Should we add a note that drivers may need to adjust the timeouts depending on their performance of bulk encoding?

Done.

> Also if we did add the w=1 approach

I'm not sure we need to do suggest `w=1` given that it does not seem to significantly improve performance and drivers are allowed to adjust the timeouts.

ShaneHarvey: Another thought. It seems like most of the slowdown is due to having so many documents. Using 50 document 1MB each also results in 2 batches (with OP_MSG payloads 48MB can fit in one batch) but has much lower overhead.

benjirewis: Good idea. For the Go driver: 150ms on standalone, 550ms on replset, 1s on sharded replicaset. Modified the prose test. I've still left the requirement of only running against standalones, but let me know if you think I should allow replsets (sharded replicaset may require higher suggested timeouts).

ShaneHarvey: Thanks, let's leave it at only standalones.

PR #1314, opened 2022-10-06
[DRIVERS-2411] Prose Tests for IMDS Communication

vector-of-bool:
These test cases give a basis for testing various situations of an IMDS server without requiring testing within an Azure VM.

While writing this, I thought of possible additional IMDS behaviors that may be prudent to handle, but they would require updating the `fake_azure.py` application.

---

- [ ] ~~Update changelog.~~
- [ ] ~~Make sure there are generated JSON files from the YAML test files.~~
- [x] Test changes in at least one language driver. (Tests are running in mongo-c-driver CI)
- [ ] ~~Test these changes against all server versions and topologies (including standalone, replica set, sharded clusters, and serverless).~~

### DIFF
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -647,14 +647,16 @@ The below steps should be taken:
 5. Add HTTP headers ``Metadata: true`` and ``Accept: application/json`` to
    `Req`.
 6. Issue `Req` to the Azure IMDS server ``168.254.169.254:80``. Let `Resp` be
-   the response from the server.
-7.  If `Resp_{status} ≠ 200`, obtaining the access token has failed, and the
-    HTTP response body of `Resp` encodes information about the error that
-    occurred. Return an error instead of an access token.
-8.  Otherwise, let `J` be the JSON document encoded in the HTTP response body
-    of `Resp`.
-9.  The result access token `T` is given as the ``access_token`` string property
-    of `J`. Return `T` as the resulting access token.
+   the response from the server. If the HTTP response is not completely received
+   within three seconds, consider the request to have timed out, and return an
+   error instead of an access token.
+7. If `Resp_{status} ≠ 200`, obtaining the access token has failed, and the HTTP
+   response body of `Resp` encodes information about the error that occurred.
+   Return an error instead of an access token.
+8. Otherwise, let `J` be the JSON document encoded in the HTTP response body of
+   `Resp`.
+9. The result access token `T` is given as the ``access_token`` string property
+   of `J`. Return `T` as the resulting access token.
 10. The resulting "expires in" duration `d_{exp}` is a count of seconds given as an
     ASCII-encoded integer string ``expires_in`` property of `J`.
 
@@ -2513,6 +2515,7 @@ Changelog
    :align: left
 
    Date, Description
+   22-10-11, Specify a timeout on Azure IMDS HTTP requests and fix the resource URL
    22-10-05, Remove spec front matter and ``versionadded`` RST macros (since spec version was removed)
    22-09-26, Add behavior for automatic Azure KeyVault credentials for ``kmsProviders``.
    22-09-09, Prohibit ``rewrapManyDataKey`` with libmongocrypt <= 1.5.1.
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -624,6 +624,12 @@ Identities* associated with them. From within the VM, an identity can be used by
 obtaining an access token via HTTP from the *Azure Instance Metadata Service*
 (IMDS). `See this documentation for more information`__
 
+..note::
+
+   To optimize for testability, it is recommended to implement an isolated
+   abstraction for communication with IMDS. This will aide in the implementation
+   of the prose tests of the communication with an IMDS server.
+
 __ https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http
 
 .. default-role:: math
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -646,19 +646,19 @@ The below steps should be taken:
 
 5. Add HTTP headers ``Metadata: true`` and ``Accept: application/json`` to
    `Req`.
-6. Issue `Req` to the Azure IMDS server ``168.254.169.254:80``. Let `Resp` be
+6. Issue `Req` to the Azure IMDS server ``169.254.169.254:80``. Let `Resp` be
    the response from the server. If the HTTP response is not completely received
-   within three seconds, consider the request to have timed out, and return an
+   within ten seconds, consider the request to have timed out, and return an
    error instead of an access token.
 7. If `Resp_{status} ≠ 200`, obtaining the access token has failed, and the HTTP
    response body of `Resp` encodes information about the error that occurred.
-   Return an error instead of an access token.
+   Return an error including the HTTP response body instead of an access token.
 8. Otherwise, let `J` be the JSON document encoded in the HTTP response body of
    `Resp`.
 9. The result access token `T` is given as the ``access_token`` string property
    of `J`. Return `T` as the resulting access token.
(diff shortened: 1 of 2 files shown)


source/client-side-encryption/tests/README.rst
```diff
@@ -2188,3 +2188,142 @@ attached to the virtual machine.
...
+Case 6: Slow Response
+`````````````````````
+
+This case addresses an IMDS server responding very slowly. Drivers should not
+halt the application waiting on a peer to communicate.
+
+Set ``X-MongoDB-HTTP-TestParams`` to ``case=slow``.
+
+The HTTP response from the ``fake_azure`` server will take at least 1000 seconds
+to complete. The client should report a timeout within a few seconds.
```
kevinAlbs, line 2317: I suggest adding the timeout in the specification to favor consistency between driver implementations. Since C is [using 3 seconds](https://github.com/mongodb/mongo-c-driver/blob/d217ff83f852f14c9cb7b0e400b4ce983c2edf2e/src/libmongoc/src/mongoc/mcd-azure.c#L182), consider adding:

> 6. Issue `Req` to the Azure IMDS server ``168.254.169.254:80`` **with a timeout of 3 seconds**.

kevinAlbs (approved):
LGTM with addition of the timeout to the specification.

comandeo (approved):

source/client-side-encryption/tests/README.rst
DmitryLukyanov: can a timeout be 10 seconds? We already have this timeout in other places and based on c# guidelines, it's better to use a single http client for all http requests. Changing it on our side will be a BC

vector-of-bool: I do not feel strongly on any particular timeout duration. Three seconds is what I happened to write into the test, but it can be changed.

source/client-side-encryption/tests/README.rst
```diff
@@ -2188,3 +2188,142 @@ attached to the virtual machine.
...
+
+
+Case 7: Large Response
+``````````````````````
+
+This case addresses an IMDS server responding with an abnormally large message.
```
DmitryLukyanov, line 2323: after what size a response is considered giant? :) Currently this step fails for me because timeout on receiving server response

vector-of-bool: This is a somewhat arbitrary limit, and may depend on the environment, I suppose. In the C driver I have it 8MB before it errors.

DmitryLukyanov: does it matter to have this at all here? I suspect that Timeout exception handles this case? I doubt 8MB will be handled quicker than 3/8 seconds?

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -633,22 +639,24 @@ The below steps should be taken:
...
 2. Add a query parameter ``api-version=2018-02-01`` to `U`.
-3. Add a query parameter ``resource=http://vault.azure.com/`` to `U`.
+3. Add a query parameter ``resource=https://vault.azure.net/`` to `U`.
 4. Prepare an HTTP GET request `Req` based on `U`.
 
    .. note:: All query parameters on `U` should be appropriately percent-encoded
 
 5. Add HTTP headers ``Metadata: true`` and ``Accept: application/json`` to
    `Req`.
 6. Issue `Req` to the Azure IMDS server ``168.254.169.254:80``. Let `Resp` be
```
DmitryLukyanov, line 649: 16**9**.254.169.254:80

source/client-side-encryption/tests/README.rst
```diff
@@ -2188,3 +2188,142 @@ attached to the virtual machine.
...
+This case addresses an IMDS server responding very slowly. Drivers should not
+halt the application waiting on a peer to communicate.
+
+Set ``X-MongoDB-HTTP-TestParams`` to ``case=slow``.
+
+The HTTP response from the ``fake_azure`` server will take at least 1000 seconds
+to complete. The client should report a timeout within a few seconds.
+
+
+Case 7: Large Response
```
DmitryLukyanov, line 2320: should we have a case that tests caching?

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -633,22 +639,24 @@ The below steps should be taken:
...
-    HTTP response body of `Resp` encodes information about the error that
-    occurred. Return an error instead of an access token.
-8.  Otherwise, let `J` be the JSON document encoded in the HTTP response body
-    of `Resp`.
-9.  The result access token `T` is given as the ``access_token`` string property
-    of `J`. Return `T` as the resulting access token.
+   the response from the server. If the HTTP response is not completely received
+   within three seconds, consider the request to have timed out, and return an
+   error instead of an access token.
+7. If `Resp_{status} ≠ 200`, obtaining the access token has failed, and the HTTP
```
DmitryLukyanov, line 653: can we formulate this like it was done for gcp (lines 607-608 above) https://github.com/mongodb/specifications/pull/1314/files#diff-de01d398114f63480af23798e09ffca599aa758efd04f15a9beec89b4566041bR606-R607

and in particular include this part: 
>return an error including the body of the HTTP response in the error message.

DmitryLukyanov (commented):
I've implemented (mostly) this test in c# https://github.com/mongodb/mongo-csharp-driver/pull/898.
Some notes/questions:
1. The implemented test is unit, not integration as we usually do. Initially I wanted to provide a way to override imds base uri (and use only public API) and despite that it may work, I don't think that the same is true for assigning additional http headers :( So I ended up with testing only azure creds part without touching csfle logic itself. 
2. Then, can we consider not only mocking server but something with real env to be able to fully test this logic including csfle steps? cc: @kevinAlbs 
3. I've provided some minor comments/suggestion in the spec PR, @vector-of-bool please take a look
4. I think we don't need the case 7 from this PR (since as far as I can see it will go through timeout workflow (case 6), won't it?)
5. We probably need a new test case that will test caching

source/client-side-encryption/tests/README.rst
vector-of-bool: A pathologically broken server could send dozens of MBs to a client within a few seconds. Regardless, this is indeed a pathological test case, and such an error should be handled by the HTTP abstraction, so testing its correctness could be considered "beyond the scope" of IMDS testing, but the same could be said of invalid JSON or a slow server response.

@kevinAlbs what do you think should be considered "in scope" for these tests? I recognize that there's some novelty here that hasn't really been part of testing before.

kevinAlbs: I suggest removing `Case 7: Large Response` if there are difficulties implementing. I assume that is a very unlikely (if possible) failure scenario from an IMDS server.

`Case 6: Slow Response` validates the client-side timeout is correct. I suggest keeping it.

The invalid JSON tests seem like more plausible failure scenarios. If there are no implementation difficulties, I suggest keeping them.

kevinAlbs:
> Then, can we consider not only mocking server but something with real env to be able to fully test this logic including csfle steps? cc: @kevinAlbs

I support adding an integration test similar to the [tests for GCP](https://github.com/mongodb/specifications/blob/342b880d89c142990c20264d57fbeefdd8df5162/source/client-side-encryption/tests/README.rst#17--on-demand-gcp-credentials). The GCP integration tests are minimal and run every 2 weeks, but they provide some additional assurance. I am concerned with only relying on the mock server. A driver could goof the small amount of untested code (e.g. use an incorrect IP address for IMDS). The mock server may have subtle differences from the real IMDS. Or IMDS may change in the future.

Experiments suggest translating the [scripts to set up a GCP environment in drivers-evergreen-tools](https://github.com/mongodb-labs/drivers-evergreen-tools/tree/master/.evergreen/csfle/gcpkms) is not too much effort. I can do that in a separate PR. If that is OK, I propose leaving DRIVERS-2411 "In Progress" until integration tests are merged.

vector-of-bool:
Removed the pathological "large response" case. Additional test cases to be defined when we have a definite way to create Azure environments.

source/client-side-encryption/tests/README.rst
vector-of-bool: Additional cases are to be defined in a subsequent changeset, and I think cache handling may be better checked in the true environment.

PR #1315, opened 2022-10-06
fix SDAM code formatting to get syntax highlighting

nbbeeken:
Fixed our code formatting to get some highlighting and the SDAM subroutines are now entries in the table contents under `Actions`.

None of the usual checklist applies since this does not have functional changes.

### DIFF
--- a/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
+++ b/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
@@ -987,240 +987,247 @@ TopologyType ReplicaSetWithPrimary
     run `updateRSWithPrimaryFromMember`_.
 
 Actions
-~~~~~~~
-
-.. _updateUnknownWithStandalone:
+```````
 
 updateUnknownWithStandalone
-  This subroutine is executed
-  with the ServerDescription from Standalone
-  when the TopologyType is Unknown::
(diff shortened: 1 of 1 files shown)


source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -987,240 +987,247 @@ TopologyType ReplicaSetWithPrimary
...
+```````
 
 updateUnknownWithStandalone
-  This subroutine is executed
-  with the ServerDescription from Standalone
-  when the TopologyType is Unknown::
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This subroutine is executed with the ServerDescription from Standalone
+when the TopologyType is Unknown
```
ShaneHarvey, line 996: Here and for the other code blocks let's keep a single `:`, like:
```
... when the TopologyType is Unknown:

.. code-block:: python
```

ShaneHarvey (changes requested):
Nice stuff!

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
nbbeeken: Agreed, fixed!

ShaneHarvey (approved):
LGTM!

PR #1316, opened 2022-10-07
DRIVERS-2385: expectedError.errorResponse for asserting server-side error doc

jmikola:
https://jira.mongodb.org/browse/DRIVERS-2385

Prototype: https://github.com/mongodb/mongo-php-library/pull/989

----

- [x] Update changelog.

### DIFF
--- a/source/collection-management/tests/modifyCollection-errorResponse.yml
+++ b/source/collection-management/tests/modifyCollection-errorResponse.yml
@@ -1,0 +1,59 @@
+description: "modifyCollection-errorResponse"
+
+schemaVersion: "1.12"
+
+createEntities:
+  - client:
+      id: &client0 client0
+      observeEvents: [ commandStartedEvent ]
+  - database:
+      id: &database0 database0
+      client: *client0
+      databaseName: &database0Name collMod-tests
+  - collection:
+      id: &collection0 collection0
+      database: *database0
+      collectionName: &collection0Name test
+
+initialData: &initialData
+  - collectionName: *collection0Name
+    databaseName: *database0Name
+    documents:
+      - { _id: 1, x: 1 }
+      - { _id: 2, x: 1 }
+
+tests:
+  - description: "modifyCollection prepareUnique violations are accessible"
+    runOnRequirements:
+      - minServerVersion: "5.2" # SERVER-61158
+    operations:
+      - name: createIndex
+        object: *collection0
+        arguments:
+          keys: { x: 1 }
+      - name: modifyCollection
+        object: *database0
+        arguments:
+          collection: *collection0Name
+          index:
+            keyPattern: { x: 1 }
+            prepareUnique: true
+      - name: insertOne
+        object: *collection0
+        arguments:
+          document: { _id: 3, x: 1 }
+        expectError:
+          errorCode: 11000 # DuplicateKey
+      - name: modifyCollection
+        object: *database0
+        arguments:
+          collection: *collection0Name
+          index:
+            keyPattern: { x: 1 }
+            unique: true
+        expectError:
+          isClientError: false
+          errorCode: 359 # CannotConvertIndexToUnique
+          errorResponse:
+            violations:
+               - { ids: [ 1, 2 ] }
--- a/source/unified-test-format/tests/Makefile
+++ b/source/unified-test-format/tests/Makefile
@@ -1,4 +1,4 @@
-SCHEMA=../schema-1.9.json
+SCHEMA=../schema-1.12.json
 
 .PHONY: all invalid valid-fail valid-pass versioned-api load-balancers gridfs transactions crud collection-management sessions command-monitoring client-side-operations-timeout HAS_AJV
 
--- a/source/unified-test-format/unified-test-format.rst
+++ b/source/unified-test-format/unified-test-format.rst
@@ -1079,6 +1079,11 @@ The structure of this object is as follows:
   assert that the error does not contain any of the specified labels (e.g. using
   the ``hasErrorLabel`` method).
 
+- ``errorResponse``: Optional document. A value corresponding to the expected
+  server-generated error response. The test runner MUST assert that the error
+  includes a server-generated response that matches this value according to the
+  rules in `Evaluating Matches`_.
+
 .. _expectedError_expectResult:
 
 - ``expectResult``: Optional mixed type. This field follows the same rules as
@@ -3749,6 +3754,8 @@ Changelog
 ..
   Please note schema version bumps in changelog entries where applicable.
 
+:2022-10-07: **Schema version 1.12.**
+             Add ``errorResponse`` to ``expectedError``.
 :2022-10-05: Remove spec front matter, add "Current Schema Version" field, and
              reformat changelog. Add comment to remind editors to note schema
              version bumps in changelog updates (where applicable).
--- a/source/unified-test-format/unified-test-format.rst
+++ b/source/unified-test-format/unified-test-format.rst
@@ -1079,10 +1079,19 @@ The structure of this object is as follows:
   assert that the error does not contain any of the specified labels (e.g. using
   the ``hasErrorLabel`` method).
 
+.. _expectedError_errorResponse:
+
 - ``errorResponse``: Optional document. A value corresponding to the expected
-  server-generated error response. The test runner MUST assert that the error
-  includes a server-generated response that matches this value according to the
+  server response. The test runner MUST assert that the error includes a server
+  response that matches this value as a root-level document according to the
   rules in `Evaluating Matches`_.
+
+  Note that some drivers may not be able to evaluate ``errorResponse`` for write
+  commands (i.e. insert, update, delete) and bulk write operations. For example,
+  a BulkWriteException is derived from potentially multiple server responses and
+  may not provide direct access to a single response. Tests SHOULD avoid using
+  ``errorResponse`` for such operations if possible; otherwise, affected drivers
+  SHOULD skip such tests if necessary.
 
 .. _expectedError_expectResult:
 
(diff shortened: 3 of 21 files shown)


source/unified-test-format/tests/Makefile
```diff
@@ -1,4 +1,4 @@
-SCHEMA=../schema-1.9.json
+SCHEMA=../schema-1.12.json
```
jmikola, line 1: I noticed this file is quite outdated. Since [DRIVERS-1767](https://jira.mongodb.org/browse/DRIVERS-1767) was implemented and a CI workflow was added to test spec files against their exact `schemaVersion`, this file only serves two purposes:

 * Validate a wide number of specs against the latest schema (not sure how useful or necessary this is)
 * Test schema validation failures for `invalid` tests, which are ignored by the CI workflow

FWIW, the Makefile is still quite a bit faster than the [`check_schema_version.sh`](https://github.com/mongodb/specifications/blob/master/.github/workflows/check_schema_version.sh) script since it invokes `ajv` once for an entire directory of files. Going forward, I think it'd be best to eliminate this file and update `check_schema_version.sh` to also process `invalid` tests and possible optimize it to invoke fewer instances of `ajv` (e.g. collect specs per schema and run a single `ajv` instance per schema).

source/collection-management/tests/modifyCollection-errorResponse.yml
```diff
@@ -0,0 +1,59 @@
...
+
+initialData: &initialData
+  - collectionName: *collection0Name
+    databaseName: *database0Name
+    documents:
+      - { _id: 1, x: 1 }
+      - { _id: 2, x: 1 }
+
+tests:
+  - description: "modifyCollection prepareUnique violations are accessible"
```
jmikola, line 26: This test covers the original impetus for this functionality (see: [DRIVERS-2353](https://jira.mongodb.org/browse/DRIVERS-2353)). The other tests for `aggregate` and `findOneAndUpdate` are meant to provide some additional test coverage without being exhaustive.

source/unified-test-format/unified-test-format.rst
```diff
@@ -3749,6 +3754,8 @@ Changelog
 ..
   Please note schema version bumps in changelog entries where applicable.
 
+:2022-10-07: **Schema version 1.12.**
```
kmahar, line 3757: oh boy I get to do another rebase and schema version bump on #1303  🙂

source/unified-test-format/unified-test-format.rst
```diff
@@ -1079,6 +1079,11 @@ The structure of this object is as follows:
   assert that the error does not contain any of the specified labels (e.g. using
   the ``hasErrorLabel`` method).
 
+- ``errorResponse``: Optional document. A value corresponding to the expected
```
kmahar, line 1082: it seems we're treating this as a root document. should we call that out here or in Evaluating Matches?

kmahar (commented):
one comment, but LGTM otherwise. do you think we need to try these tests/changes out in any other drivers before merge?

source/unified-test-format/unified-test-format.rst
jmikola: In all seriousness, if your PR is nearly ready to merge I'm happy to hold off on this for a few more days and spare you the trouble.

source/unified-test-format/tests/Makefile
jmikola: Tracking in [DRIVERS-2466](https://jira.mongodb.org/browse/DRIVERS-2466).

jmikola (commented):
I just added some additional tests and language around an edge case I discovered with using `errorResponse` for basic write ops (e.g. insert, update, delete, and bulk writes). I spoke to @ShaneHarvey about this in the Python channel and decided it was worth incorporating into this spec, even if we ultimately discourage use of `errorResponse`.

> do you think we need to try these tests/changes out in any other drivers before merge?

Write operations in PHP all throw BulkWriteExceptions, which don't provide access to raw server responses. The main limitation is [`mongoc_bulk_operation_execute()`](http://mongoc.org/libmongoc/current/mongoc_bulk_operation_execute.html), which only returns a BSON reply document (the derived write result) and an error struct (message and error code). Therefore, I'll need to skip the affected CRUD tests in this PR and would request an additional driver (without that limitation) to POC.

benjirewis (commented):
Cool! I can POC this in the Go driver since we conveniently, recently added the raw server response to most of our errors with [GODRIVER-2325](https://jira.mongodb.org/browse/GODRIVER-2325).

source/unified-test-format/unified-test-format.rst
```diff
@@ -1079,11 +1079,20 @@ The structure of this object is as follows:
...
+
 - ``errorResponse``: Optional document. A value corresponding to the expected
-  server-generated error response. The test runner MUST assert that the error
-  includes a server-generated response that matches this value according to the
+  server response. The test runner MUST assert that the error includes a server
+  response that matches this value as a root-level document according to the
   rules in `Evaluating Matches`_.
 
+  Note that some drivers may not be able to evaluate ``errorResponse`` for write
+  commands (i.e. insert, update, delete) and bulk write operations. For example,
```
kmahar, line 1090: oh bulk is a very interesting case. I guess if the server starts adding arbitrary extra fields to write command responses we might have to do some work to figure out how to expose those via bulk results.

also, somewhat of an aside but interestingly, the scope for the new server bulk write command proposes that results are returned via a cursor (this is to handle situations where there are too many errors to fit in a single response, which can happen now sometimes for unordered bulk writes and causes an exception in the server). 
if a driver did choose to add a `rawResponses` property or something to their `BulkWriteFailure` they would have to consider how to handle that upcoming case as well.

kmahar (approved):
lgtm assuming the Go POC passes the tests

source/unified-test-format/unified-test-format.rst
kmahar: ha, I appreciate it, but it really isn't too big of a deal. I am really hoping my PR is going to get merged this week but I don't want to promise it. so let's just let things play out naturally timing wise

benjirewis (commented):
Tests [seem to pass](https://spruce.mongodb.com/version/6346dd2ed1fe07609ba4685a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) in the Go driver (apologies for the delay I was OOO), but I'm concerned about the ambiguity of this new field when multiple errors are returned from the server.

source/collection-management/tests/modifyCollection-errorResponse.yml
```diff
@@ -0,0 +1,59 @@
+description: "modifyCollection-errorResponse"
```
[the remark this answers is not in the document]
jmikola: For the benefit of test coverage, have you considered implementing the operation via `runCommand` in your test suite? PHP has a helper for it but we don't actually model all of the possible parameters, so it should be quite trivial to just build a command document from the `arguments` in the test operation if you wanted.

Maybe not so relevant for this case, since the purpose of the test is to ensure a _helper_ provides appropriate error access, but it could allow you to run some other tests that happen to use `modifyCollection`.

source/unified-test-format/unified-test-format.rst
jmikola: > I guess if the server starts adding arbitrary extra fields to write command responses we might have to do some work to figure out how to expose those via bulk results.

That thought just occurred to me this morning while I was revisiting the PHPLIB POC. I noticed that DuplicateKey(11000) returns extra fields in the write error document, which may be inaccessible in languages that model WriteError and WriteConcernError. See: [DRIVERS-2470](https://jira.mongodb.org/browse/DRIVERS-2470).

> Why are we advising against testing errorResponse with bulk writes [here](https://github.com/mongodb/specifications/pull/1316/files#diff-8b99e5ca9710d2ea778ecfe4e50cd1b3bbd51ef05b0f189f1f4f57614f343bba) and then adding a test for it here? Am I misunderstanding something? I went with a last-write-error-becomes-the-errorResponse approach (which works since there's only one error), but that was arbitrary.

My intention was to provide as much test coverage as possible for drivers that _can_ support this, and encourage incompatible drivers to skip them. I intentionally had a single operation in the `bulkWrite-errorResponse` test to avoid ambiguity, assuming that any exception would inherit the server-side error code from the `insert` command. If that seems problematic I'm happy to delete the test. Alternatively, I can add a comment explaining that things could be ambiguous with multiple operations in the bulk and note my intention for only performing a single insert.

Regarding WriteError and WriteConcernError, I don't think those come into play here since `errorResponse` is only asserting the _top-level_ command response. Technically, we should be able to assert the presence of gossiped cluster time if so inclined.

source/collection-management/tests/modifyCollection-errorResponse.yml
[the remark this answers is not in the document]
benjirewis: Yeah an interesting point; I imagine users are already running `collMod` through our `RunCommand` API, so would be good to test it. Filed [GODRIVER-2587](https://jira.mongodb.org/browse/GODRIVER-2587).

source/unified-test-format/unified-test-format.rst
benjirewis: >  Alternatively, I can add a comment explaining that things could be ambiguous with multiple operations in the bulk and note my intention for only performing a single insert.

That sounds good to me. I can see how these tests are really just to encourage drivers to expose the raw server response wherever they can.

> Regarding WriteError **[assume you mean extra write errors on non-bulk operations]** and WriteConcernError, I don't think those come into play here since errorResponse is only asserting the top-level command response.

Ah gotcha. Sounds good 👍

benjirewis (approved):
Thanks!

source/unified-test-format/unified-test-format.rst
jmikola: Added a comment to the bulk and non-bulk tests in https://github.com/mongodb/specifications/pull/1316/commits/b775b7afc1093d430b3886e82d76527551ca73ba, which explains that some drivers may need to skip them. The CRUD spec doesn't go into much detail about how to construct BulkWriteException or WriteException beyond collecting write and write concern errors.

> assume you mean extra write errors on non-bulk operations

I wasn't distinguishing between bulk and non-bulk writes, since the write commands all return `writeErrors` in an array. Rather, the fail point I'm using in these tests sets the top-level error code for the command (and forces an `ok: 0` response). That is all `errorResponse` is asserting here.

If we wanted, we could theoretically use `errorResponse` to peak inside of the `writeErrors` array (for example), but I didn't see the need for that.

jmikola (commented):
Waiting on the latest PHPLIB patch build https://spruce.mongodb.com/version/6348eecc850e6128847d3a21/tasks and will merge once that completes.

jmikola (commented):
Added a topology restriction to the aggregate test, which has different behavior on sharded clusters, and started a new patch build: https://spruce.mongodb.com/version/634902357742ae0d82027575/tasks

PR #1317, opened 2022-10-07
DRIVERS-1807 Keep existing retry behavior unless CSOT is enabled

ShaneHarvey:

- [x] Update changelog.

### DIFF
--- a/source/client-side-operations-timeout/client-side-operations-timeout.rst
+++ b/source/client-side-operations-timeout/client-side-operations-timeout.rst
@@ -48,7 +48,7 @@ timeoutMS
 
 This 64-bit integer option specifies the per-operation timeout value in
 milliseconds. The default value is unset which means this feature is not
-enabled, ie the existing timeout behavior is unchanged (including
+enabled, i.e. the existing timeout behavior is unchanged (including
 ``serverSelectionTimeoutMS``, ``connectTimeoutMS``, ``socketTimeoutMS`` etc..).
 An explicit value of 0 means infinite, though some client-side timeouts like
 ``serverSelectionTimeoutMS`` will still apply. Drivers MUST error if a
--- a/source/client-side-operations-timeout/client-side-operations-timeout.rst
+++ b/source/client-side-operations-timeout/client-side-operations-timeout.rst
@@ -47,8 +47,10 @@ timeoutMS
 ~~~~~~~~~
 
 This 64-bit integer option specifies the per-operation timeout value in
-milliseconds. The default value is unset. Both unset and an explicit value of
-0 mean infinite, though some client-side timeouts like
+milliseconds. The default value is unset which means this feature is not
+enabled, ie the existing timeout behavior is unchanged (including
+``serverSelectionTimeoutMS``, ``connectTimeoutMS``, ``socketTimeoutMS`` etc..).
+An explicit value of 0 means infinite, though some client-side timeouts like
 ``serverSelectionTimeoutMS`` will still apply. Drivers MUST error if a
 negative value is specified. This value MUST be configurable at the level of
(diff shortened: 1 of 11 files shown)


source/client-side-operations-timeout/client-side-operations-timeout.rst
```diff
@@ -47,8 +47,10 @@ timeoutMS
 ~~~~~~~~~
 
 This 64-bit integer option specifies the per-operation timeout value in
-milliseconds. The default value is unset. Both unset and an explicit value of
-0 mean infinite, though some client-side timeouts like
+milliseconds. The default value is unset which means this feature is not
+enabled, ie the existing timeout behavior is unchanged (including
```
benjirewis, line 51:
```suggestion
enabled, i.e. the existing timeout behavior is unchanged (including
```

source/retryable-writes/retryable-writes.rst
```diff
@@ -381,36 +382,35 @@ writes, drivers MUST add a transaction ID to the command. Drivers MUST only
...
-If the retry attempt also fails, drivers MUST update their topology according to
+If a retry attempt also fails, drivers MUST update their topology according to
 the SDAM spec (see: `Error Handling`_). If an error would not allow the caller
 to infer that an attempt was made (e.g. connection pool exception originating
-from the driver), the original error should be raised. If the retry failed due
-to another retryable error or some other error originating from the server, that
-error should be raised instead as the caller can infer that an attempt was made
-and the second error is likely more relevant (with respect to the current
-topology state).
+from the driver) or the error is labeled "NoWritesPerformed", the error from
```
benjirewis, line 408: Is there any other case besides `NoWritesPerformed` where we return the previous error?

source/retryable-writes/retryable-writes.rst
```diff
@@ -443,61 +443,65 @@ Consider the following pseudo-code:
...
-    try {
-      return executeCommand(server, retryableCommand);
-    } catch (DriverException ignoredError) {
-      throw originalError;
-    } catch (Exception secondError) {
-      handleError(secondError);
-      throw secondError;
+      /* CSOT is enabled and the operation has timed out. */
+      if (timeoutMS != null && is_expired(timeoutMS) {
+        throw previousError;
```
benjirewis, line 503: Do we have a spec test for this block?

benjirewis (commented):
Might want to add @prestonvasquez since some of this has to do with DRIVERS-2327

source/retryable-writes/retryable-writes.rst
ShaneHarvey: Yes, when a retry selects a server that does not support retryable writes or the retry fails with a server selection error.

source/retryable-writes/retryable-writes.rst
ShaneHarvey: Yes we do, it's part of the CSOT. The "timeoutMS applies to whole operation, not individual attempts" tests: https://github.com/mongodb/specifications/blob/b3e06831a09ae72d94da79f428d8941ec7221f7f/source/client-side-operations-timeout/etc/templates/retryability-timeoutMS.yml.template#L42-L47

benjirewis (approved):
Thanks!

prestonvasquez (approved):
LGTM!

isabelatkinson (approved):

ShaneHarvey:
I've updated this PR to include similar changes to the retryable reads spec. I've also adjusted some of the CSOT spec tests to be less flakey (mainly by increasing timeouts and using w:1 instead of w:majority). Ready for another look.

isabelatkinson (approved):
lgtm!

benjirewis (approved):
CSOT test changes LGTM and pass in the Go driver.

PR #1321, opened 2022-10-14
minor: fix RST list formatting in auth spec

kmahar:

- [ ] Update changelog.

### DIFF
--- a/source/auth/auth.rst
+++ b/source/auth/auth.rst
@@ -933,7 +933,7 @@ The order in which Drivers MUST search for credentials is:
 #. The URI
 #. Environment variables
 #. Using ``AssumeRoleWithWebIdentity`` if ``AWS_WEB_IDENTITY_TOKEN_FILE`` and
-``AWS_ROLE_SESSION_NAME``  are set.
+   ``AWS_ROLE_SESSION_NAME``  are set.
 #. The ECS endpoint if ``AWS_CONTAINER_CREDENTIALS_RELATIVE_URI`` is set. Otherwise, the EC2 endpoint.
 
 .. note::


source/auth/auth.rst
```diff
@@ -933,7 +933,7 @@ The order in which Drivers MUST search for credentials is:
 #. The URI
 #. Environment variables
 #. Using ``AssumeRoleWithWebIdentity`` if ``AWS_WEB_IDENTITY_TOKEN_FILE`` and
-``AWS_ROLE_SESSION_NAME``  are set.
+   ``AWS_ROLE_SESSION_NAME``  are set.
```
jmikola, line 936: Manually compared https://github.com/mongodb/specifications/blob/master/source/auth/auth.rst#id13 with the rendered view of this PR.

I assume this PR was prompted by [this Slack thread](https://mongodb.slack.com/archives/C0UD1CGUF/p1665783072792189) and wasn't caught by the [rstcheck](https://github.com/mongodb/specifications/blob/master/.github/workflows/rstcheck.yml) workflow. If so, is there anything we can do to fix that in the future? I wouldn't be surprised if there are a bunch of small RST formatting issues like this scattered throughout the repo.

jmikola (approved):

source/auth/auth.rst
kmahar: I'm not sure. This seems like a case where it might be hard for a linter to detect whether the formatting was intentional or done in error?
I have noticed that my RST checking tooling locally in VSCode (using [this](https://github.com/vscode-restructuredtext/vscode-restructuredtext)) catches some things that our workflow misses (e.g. I fixed some tiny issues in the unified format spec as part of #1303). It could be worth looking into whatever form that does rst checking in to at least get some improvement.

PR #1322, opened 2022-10-17
[DRIVERS-2312] Define a CreateEncryptedCollection helper

vector-of-bool:
This adds an additional convenience API for creating encrypted collections with the field datakeys automatically generated if they are omitted. This is based on the implementation for the C driver.

TODO: Spec tests, plus: `createEncryptedCollection` can likely become a new spec test op (as with `createCollection` and `dropCollection`).

- [ ] Update changelog.

DRIVERS-2312


source/client-side-encryption/client-side-encryption.rst
```diff
@@ -862,24 +901,75 @@ Drivers MUST support a BSON document option named ``encryptedFields`` for any
...
+  ``encryptedFields["ecocCollection"]`` is not set, use the collection name
+  ``enxcol_.<collectionName>.ecoc``. Creating this collection MUST NOT check if
+  the collection namespace is in the ``AutoEncryptionOpts.encryptedFieldsMap``.
+- Create the collection ``collectionName`` with ``collectionOptions`` and the
+  option ``encryptedFields`` set to the ``encryptedFields``.
+- Create the the index ``{"__safeContent__": 1}`` on collection
+  ``collectionName``.
+
+
+Create Encrypted Collection Helper
```
jmikola, line 938: I realize we don't have a [Collection Management spec](https://github.com/mongodb/specifications/tree/master/source/collection-management/tests) (just tests), but can a DRIVERS ticket be created and linked to DRIVERS-2312 to serve as a reminder to reference this helper when such a spec is ultimately created?

kevinAlbs (commented):
I like the new `GetEncryptedFields` section. Can you make a separate PR for the refactor? That may make it easier for other DBX engineers to refer to the behavior changes introduced by `CreateEncryptedCollection`.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -862,24 +901,75 @@ Drivers MUST support a BSON document option named ``encryptedFields`` for any
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+To support automatic generation of encryption data keys, a helper
+`CreateEncryptedCollection` is defined. This requires an associated
```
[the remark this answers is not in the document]
vector-of-bool: The distinction is syntactic, but I can clear the wording to make it more obvious that the API is "associated with" `ClientEncryption` API objects.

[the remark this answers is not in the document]
kevinAlbs: I suggest adding that clarification to ensure drivers associate `CreateEncryptedCollection` with the same `ClientEncryption`. Otherwise, a driver may associate it alongside the `CreateCollection` helper.

I suggest adding to the [`ClientEncryption` class declaration](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.rst#clientencryption).

source/client-side-encryption/client-side-encryption.rst
jmikola: https://jira.mongodb.org/browse/DRIVERS-2481

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -914,6 +994,11 @@ ClientEncryption
    class ClientEncryption {
       ClientEncryption(opts: ClientEncryptionOpts);
 
+      // The "Create Encrypted Collection" helper is a convenience function wrapping CreateCollection. It will
+      // create a collection with encrypted fields, automatically allocating and assigning new data encryption
+      // keys. Refer to "Create Encrypted Collection Helper"
+      createEncryptedCollection(collName: string, collOpts, kmsProvider: KMSProviderName, dkOpts: DataKeyOpts);
```
DmitryLukyanov, line 1000: as it was discussed offline:
>maybe another option may be passing not a clientEncription, but ClientEncryptionOptions  and create a client encryption internally each time?

vector-of-bool: What's the benefit over taking the ClientEncryption object directly?

DmitryLukyanov: it will work with Go since by this, as I understand, we avoid a cyclic physical dependency between components that prohibit using ClientEncryption directly as an argument. cc: @kevinAlbs

vector-of-bool: It doesn't create a dependency cycle in Go, but what is the benefit of overloading CreateCollection to accept the options rather than creating the ClientEncryption from those options and using that?

In some cases (e.g. the C driver) we would still need to add an additional API to accept the options object since it cannot be fit into the existing `mongoc_database_create_collection` interface.

kevinAlbs: > It doesn't create a dependency cycle in Go, but what is the benefit of overloading CreateCollection to accept the options rather than creating the ClientEncryption from those options and using that?

It may result in one less public method. Users wanting to create an encrypted collection only need to use CreateCollection. And CreateCollection already supports encryptedFields.

> In some cases (e.g. the C driver) we would still need to add an additional API to accept the options object since it cannot be fit into the existing mongoc_database_create_collection interface.

Good point. The current C options are not sufficient for accepting the keyVaultClient from ClientEncryptionOptions. C may be the only exception that would require additional API.

Is it OK to leave C as an exception, and have the specification specify in terms of CreateCollection?

DmitryLukyanov: >what is the benefit of overloading CreateCollection 

the benefit is to have a single API to create encrypted collection and don't spread this logic across driver (ie it's better to have a single place for creating collection instead ClientEncryption and regular collection based approaches).

But there is one more downside with my suggestion that I didn't consider, not sure how likely, but in theory `ClientEncryptionOptions` may contain many options and some of them may be fully unrelated to creating collection, so it will be unclear what value should be provided there if we just want to create encrypted collection. 

Given the above (and the fact we probably don't want deprecating already existed approach or created some options hierarchy), maybe we can stick to the simplest approach suggested initially?

vector-of-bool: If a user is intending to use an encrypted collection, they will likely need a `ClientEncryption` object anyway (at least for explicit encryption).

Even if it is an "overload" of the `CreateCollection` API, it will still be a distinct API that just happens to use the same name. Currently, the "CreateEncryptedCollection" is associated with the ClientEncryption object, but it could be migrated to be part of the Database APIs. It's mostly just a matter of parameter ordering.

kevinAlbs: Good points. I am in favor of keeping the `CreateEncryptedCollection` API as is currently proposed.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -881,6 +921,44 @@ If the collection namespace has an associated ``encryptedFields``, then do the f
...
+- If `Fields` is present and is an array value, then for each element `F` of
+  `Fields`:
+
+  - If `F` is not a document element, skip it.
+  - Otherwise, if `F` has a ``"keyId"`` named element `K` and `K` is a ``null``
+    value:
+
+    - Let `D` be the result of ``C.createDataKey(kmsProv, dkOpts)``.
+    - If generating `D` resulted in an error `E`, the entire
+      `CreateEncryptedCollection` must now fail with error `E`.
```
kevinAlbs, line 952: If an error occurs, it may be difficult to determine which keys were created before the error. A user may want to delete those keys.

Consider always returning `encryptedFields`. Even on error, created key ids can be included in the result.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -914,6 +994,12 @@ ClientEncryption
    class ClientEncryption {
       ClientEncryption(opts: ClientEncryptionOpts);
 
+      // The "Create Encrypted Collection" helper is a convenience function wrapping CreateCollection. It will
+      // create a collection with encrypted fields, automatically allocating and assigning new data encryption
+      // keys. It returns a handle to the  new collection, as well as a document consisting of the generated
+      // "create" options. Refer to "Create Encrypted Collection Helper"
```
kevinAlbs, line 1000:
```suggestion
      // "encryptedFields" options. Refer to "Create Encrypted Collection Helper"
```

I expect some drivers may not have a native way to convert CreateCollectionOptions to BSON. Only returning "encryptedFields" is needed to provide the datakeys to the caller.

kevinAlbs (approved):
LGTM with one change.

DmitryLukyanov (approved):
see minor comments and then LGTM
NOTE: I would say we need tests ready before moving this ticket to team implementing status

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -854,6 +858,42 @@ A collection supporting Queryable Encryption requires an index and three additio
...
+   element named by `QualName`, then `EF` is the value of that element.
+4. Otherwise, if `askDb` is `true`:
+
+   1. Issue a ``listCollections`` command against the database named by
+      `dbName`, filtered by ``{name: <collName>}``. Let the result be the
+      document `L`.
+   2. If `L` contains an ``options`` document element, and that element contains
+      an ``encryptedFields`` document element, `EF` is `L`\
+      ``["options"]["encryptedFields"]``.
+   3. Otherwise, `EF` is *not-found*
```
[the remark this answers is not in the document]
vector-of-bool: This pseudo-code function is exposition-only, so it the idea of "not-found" is conceptual. How that is realized in the actual code is up to the implementers.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -827,9 +827,13 @@ encryptedFieldsMap
 
 ``encryptedFieldsMap`` only applies to Queryable Encryption.
 
-If a collection is present on both the ``encryptedFieldsMap`` and ``schemaMap``, libmongocrypt_ will error on initialization. See :ref:`fle2-and-fle1-error`.
+If a collection is present on both the ``encryptedFieldsMap`` and ``schemaMap``,
+libmongocrypt_ will error on initialization. See :ref:`fle2-and-fle1-error`.
```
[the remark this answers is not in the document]
vector-of-bool: There are several `:ref:` roles around the spec docs. That's a Sphinx feature, and not part of rST itself, so GitHub doesn't render them. Changing those would is out-of-scope for this PR.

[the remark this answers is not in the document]
DmitryLukyanov: maybe changing other is out of scope, but not introducing a new one can be? As I know only csfle spec implements it which effectively makes hard reading this spec (which is the main goal?). I will rely on you and @kevinAlbs here

vector-of-bool:
@DmitryLukyanov @kevinAlbs I added a very primitive prose test for the new helper. More exhaustive testing is warranted, but it provides a starting point to verify the happy path.

source/client-side-encryption/client-side-encryption.rst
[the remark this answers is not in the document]
vector-of-bool: This paragraph is unchanged. The diff is just a reflowing of the paragraph.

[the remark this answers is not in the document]
DmitryLukyanov: good point, I missed this

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -881,6 +921,44 @@ If the collection namespace has an associated ``encryptedFields``, then do the f
...
+
+To support automatic generation of encryption data keys, a helper
+`CreateEncryptedCollection(CE, database, collName, collOpts, kmsProvider, dkOpts)`
+is defined, where `CE` is a ClientEncryption_ object, `kmsProvider` is a
+KMSProviderName_ and `dkOpts` is a DataKeyOpts_. It has the following behavior:
+
+- Let `dbName` be the name of `database`. Look up the encrypted fields `EF` for
+  the new collection as `GetEncryptedFields(collOpts, collName, dbName, false)`
+  (`See here <GetEncryptedFields_>`_).
+- If `EF` is *not-found*, report an error.
```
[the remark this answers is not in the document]
vector-of-bool: Not exactly (yet), but I just tweaked the wording to give more guidance.

PR #1324, opened 2022-10-18
DRIVERS-1675 Add log messages and tests to CMAP spec

kmahar:

- [X] Update changelog.

This PR includes the following:
* Add log messages to CMAP which match the events defined in CMAP. 
* Adds JSON/YAML tests for the log messages, utilizing unified format changes made in #1303.
    * As part of this, I rearranged the existing CMAP tests. Those in the proprietary CMAP format have been moved into a new `tests/cmap-format` directory, and the logging tests live in a `tests/logging` directory. Additionally, I moved all the content on the CMAP format tests into a new README in `tests/cmap-format` and edited the top-level README to discuss the different types of CMAP tests. I don't have strong feelings on the naming or particular organization here, so if anyone prefers something else, let me know. This just seemed like a logical enough way to do it. 
* Adds a description of a client `close` operation to the unified test format. This is required to be able to test pool close messages. Per a recent conversation with @jmikola I don't believe this warrants a new minor schema version: "Drivers are supposed to raise errors for unsupported operations so I'm not at all concerned about bumping a schema version for it. The reasoning behind schema changes is that drivers might silently ignore new syntax."

The new tests pass in the Rust driver, but there are a few of them I had to skip from `connection-pool-options.yml` due to options we don't support. So if another driver is able to test these and let me know if they work for you, please let me know:
* "maxConnecting should be included in connection pool created message when specified"
* "waitQueueTimeoutMS should be included in connection pool created message when specified"
* "waitQueueSize should be included in connection pool created message when specified"
* "waitQueueMultiple should be included in connection pool created message when specified"

For the latter two I wasn't sure if they were worth including given that the options are deprecated, but at least the .NET driver still supports those options so it seemed they would provide value to them. (cc @BorisDog)

I will leave some comments inline.


source/connection-monitoring-and-pooling/tests/logging/connection-logging.yml
```diff
@@ -0,0 +1,194 @@
...
+        arguments:
+          entities:
+            - client:
+                id: &client client
+                uriOptions:
+                  retryReads: false
+                  appname: &clientAppName clientAppName
+                  # use a high heartbeatFrequencyMS to avoid a successful monitor check marking the pool as
+                  # ready (and emitting another event) during the course of test execution.
+                  heartbeatFrequencyMS: 10000
```
kmahar, line 119: at least in Rust I was running into a failure on server versions that do not support streamable monitoring (< 4.4) like the following:
* failpoint is hit: connection pool is cleared, server is marked unknown, in progress check is canceled 
* the next check starts and succeeds quickly so pool is re-marked ready, log message is emitted
* we assert on emitted log messages and there is an extra "pool ready" message at the end 

I _think_ this wasn't happening with the streaming protocol because the server waits the full heartbeatFrequencyMS/maxAwaitTimeMS to send an update in response to the check following the failpoint rather than returning immediately, since from its perspective nothing actually changed, so in that case the test gets a little extra time to make the assertions before the pool is marked ready again and a new event goes out. (does this sound right @patrickfreed?) 
that said, it seems possible to hit this issue with the streaming protocol too with a low enough heartbeatFrequencyMS + sufficient delay in asserting on emitted log messages.

I did not have a strong reason for 10 seconds specifically here; I just chose a large value that seemed like it would very safely allow the test to complete before a second heartbeat after the initial one which discovers the server succeeds.

kmahar: > I think this wasn't happening with the streaming protocol because the server waits the full heartbeatFrequencyMS/maxAwaitTimeMS to send an update in response to the check following the failpoint rather than returning immediately

I'm actually not sure if this hypothesis makes sense because I think upon seeing the network error the server close the monitoring connection and open a new one on which it will re-negotiate the monitoring protocol. and so the first check following the failpoint would not be awaitable. I'll try to figure out what's actually happening tomorrow 🙂

kmahar: I figured out what was going on here. some key observations:
1) the Rust test runner coincidentally waits for exactly 500ms=default heartbeatFrequencyMS before making assertions on events/log messages, in order to give the events time to come in.
2) after an in-progress streaming monitoring check is canceled due to a network error during handshake, we wait heartbeatFrequencyMS post cancellation before starting a new one. 
3) however, if a network error during handshake occurs between heartbeats in the polling protocol, it has no impact on the timing of the next heartbeat. 

given those, what was happening in the polling protocol was something like:
1. t=0 heartbeat 1 starts
2. t=2 heartbeat 1 succeeds; server marked standalone 
3. t=10 failpoint hit; test runner begins 500ms sleep
4. t=15 connection pool is cleared and server is marked unknown
5. t=502 heartbeat 2 begins (per observation 3, 500ms after last heartbeat)
6. t=504 heartbeat 2 succeeds
7. t=508 server marked as standalone and connection pool marked ready
8. 7=510 test runner wakes up and starts asserting on events

vs in the streaming protocol:
1. t=0 heartbeat 1 starts (first one is non awaitable)
2. t=2 heartbeat 1 succeeds; server marked standalone 
3. t=2 heartbeat 2 begins (awaitable)
4. t=10 failpoint hit; test runner begins 500ms sleep
5. t=15 connection pool is cleared and server is marked unknown
6. t=15 in-progress heartbeat is canceled
7. t=510 test runner wakes up and asserts on events
8. t=515 heartbeat 3 begins (500ms after cancellation, due to observation 2 above) (non awaitable due to coming after the failpoint error) 

If the test runner sleep is increased to 510 seconds, the test also starts to fail in the streaming protocol, since it gives enough time for heartbeat 3 to happen. 

the Rust test runner's time spent sleeping could possibly be shortened, but even so and especially given that drivers tend to decrease heartbeatFrequencyMS to speed up tests it seems like others could run into this if their assertions aren't completed before the next heartbeat. given this I think it still makes sense to keep the high value here to remove any chance that a new heartbeat completes before the test completes.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -998,6 +998,337 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
...
+     - String
+     - "Connection created"
+
+   * - driverConnectionId
+     - Int
+     - The driver-generated ID for the connection as defined in `Connection <#connection>`_.
+
+The unstructured form SHOULD be as follows, using the values defined in the structured format above to fill in placeholders as appropriate:
+
+  Connection created with driver ID {{driverConnectionId}} to host with hostname {{serverHost}} on port {{serverPort}}
```
patrickfreed, line 1175: "driver ID" here sort of reads like its identifying the driver that created the connection rather than the connection itself. Do you think it would be too ambiguous to just say "ID" here?

source/connection-monitoring-and-pooling/tests/logging/connection-logging.yml
patrickfreed: Sounds good to me! great sleuthing to figure out the timing issue

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
kmahar: I'd like to make sure that it's clear this is a driver-generated ID, vs the server-generated IDs that they will see in server logs. I don't think it would necessarily be obvious to a user that at the time this message is emitted we don't have access to the server ID yet. maybe we could just say "driver-generated ID"? 

relatedly: I used "driver ID" in all the command log message suggested unstructured forms. but that could be revised to be consistent with these

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -998,6 +998,337 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
...
+     - Suggested Type
+     - Value
+
+   * - serverHost
+     - String
+     -  the hostname or IP address for the endpoint the pool is for.
+
+   * - serverPort
+     - Int
+     - The port for the endpoint the pool is for. Optional; only present if a port was specified.
```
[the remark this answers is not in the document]
kmahar: this seems reasonable in cases where there is a port. but do ports actually have any significance for unix domain sockets? from looking at the libmongoc code, [`mongoc_client_connect_unix`](https://github.com/mongodb/mongo-c-driver/blob/b6dad0ccb4ba2e1dfd3f4453a01ec440487af563/src/libmongoc/src/mongoc/mongoc-client.c#L743-L790) does nothing with the port (vs [`mongoc_client_connect_tcp`](https://github.com/mongodb/mongo-c-driver/blob/b6dad0ccb4ba2e1dfd3f4453a01ec440487af563/src/libmongoc/src/mongoc/mongoc-client.c#L645-L722) which uses `host->port`.)

jmikola (approved):
Unified changes LGTM although I have one question. Not going to review the CMAP stuff.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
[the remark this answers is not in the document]
patrickfreed: Yeah, in that case I don't think there would be one. I believe our representation of Unix domain sockets is just a path to a file on disk.

On a related note: would we want to combine these two host/port fields into a single "address" field? Might be a little bit cleaner since as you point out ports only apply to certain types of addresses.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1634,6 +1634,18 @@ specifications:
 Client operations that require special handling or are not documented by an
 existing specification are described below.
 
+close
```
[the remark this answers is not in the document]
kmahar: Yes, from a search through the specs repo the only other occurrences of the `close` operation I can find are for cursors and change streams. All of those occurrences are in the load balancer tests.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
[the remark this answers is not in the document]
kmahar: I'm fine with adding language saying drivers should put the default port here when it is being used. that said, I'm not sure we can do anything more than say drivers should do it, because I don't see how we can write a test for it 🤔 the connection strings we get from drivers-evergreen-tools always do have ports specified, and there's no way from a unified test to override the full connection string - only particular options on it. so without prose testing every log message you can't really confirm drivers do this right. a SHOULD might be good enough though. 

> On a related note: would we want to combine these two host/port fields into a single "address" field? Might be a little bit cleaner since as you point out ports only apply to certain types of addresses.

when there is no port (i.e. Unix domain socket) drivers who log strings can just emit the "with port X" portion and drivers doing structured logging can just not attach a port value at all, so I think we end up in a similar place. if you feel strongly about this I'm not deeply opposed to it, but if it's only a suggestion I think I'd rather stick with the precedent we set in the command log messages of keeping these fields separate. also, while I don't think we should let testability leak much into public API decisions, it might be worth noting that combining the fields would also prevent us from actually testing that drivers include both the host and port and that they are types we expect them to be (string and number) - all we could do is assert that an address string is present.

source/unified-test-format/unified-test-format.rst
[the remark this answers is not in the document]
jmikola: Indeed, that operation is already documented in [Cursor Operations](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#cursor-operations). Thanks for confirming.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -998,6 +998,337 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
...
+     - Suggested Type
+     - Value
+
+   * - message
+     - String
+     - "Connection pool ready"
+
+The unstructured form SHOULD be as follows, using the values defined in the structured format above to fill in placeholders as appropriate:
+
+  Connection pool ready for host with hostname {{serverHost}} on port {{serverPort}}
```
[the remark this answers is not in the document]
patrickfreed: Even if we do keep separate structured fields for the hostname and port portion, I do think this change is worth making, since otherwise, I'd imagine any consumers of these logs would need to write custom logic to parse "for host with hostname" and "on port", whereas "hostname:port" probably has many existing parsers. Similarly, grepping for "a:1234" is a bit quicker than grepping for the whole english phrase.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -998,6 +998,337 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
...
+components, handling of null values in log messages, and structured versus unstructured logging.
+
+Drivers MUST support logging of connection pool information via the following types of log messages. These messages MUST be logged at ``Debug`` level
+and use the ``connection`` log component. These messages MUST be emitted when specified in “Connection Pool Behaviors”.
+
+The log messages are intended to match the information contained in the events above. Drivers MAY implement connection logging support via an event
+subscriber if it is convenient to do so.
+
+The types used in the structured message definitions below are demonstrative, and drivers MAY use similar types instead so long as the information
+is present (e.g. a double instead of an integer, or a string instead of an integer if the structured logging framework does not support numeric types.)
```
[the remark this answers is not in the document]
kmahar: moved period out of the parentheses

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -998,6 +998,337 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
...
+   :header-rows: 1
+   :widths: 1 1 1
+
+   * - Key
+     - Suggested Type
+     - Value
+
+   * - serverHost
+     - String
+     -  the hostname or IP address for the endpoint the pool is for.
```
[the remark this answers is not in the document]
kmahar: done

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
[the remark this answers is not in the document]
kmahar: I updated the language here:
> The port for the endpoint the pool is for. Optional; not present for Unix domain sockets. When
       the user does not specify a port and the default (27017) is used, the driver SHOULD include it here. 

going to make a follow up PR to do similar for the command log messages

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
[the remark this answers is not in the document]
kmahar: I updated here and the other log messages to combine them. I'm going to make a follow-up PR to make the equivalent change in the command log message suggested string representations

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
kmahar: updated to "driver-generated ID" here and elsewhere. like above, will make similar change to command messages in follow up PR

source/connection-monitoring-and-pooling/tests/logging/connection-logging.yml
```diff
@@ -0,0 +1,194 @@
...
+              serverPort: { $$type: [int, long] }
+              reason: "An error occurred while using the connection"
+
+          - level: debug
+            component: connection
+            data:
+              message: "Connection checkout failed"
+              serverHost: { $$type: string }
+              serverPort: { $$type: [int, long] }
+              reason: "An error occurred while trying to establish a connection"
```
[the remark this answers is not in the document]
kmahar: yes good catch, I accidentally omitted the "new" here and in Rust impl. fixed.

source/unified-test-format/unified-test-format.rst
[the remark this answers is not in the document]
kmahar: FYI @jmikola I made a small addition to the paragraph here to clarify that entities descended from a closed client should also not be used after a `close`.

kmahar:
FYI: It came up in driver-devs that Ruby does not close the connection pool when a client is closed. I was thinking about how these tests would work (or not work) for Ruby and I noticed all of the connection-pool-options tests were unnecessarily closing the client and expecting a corresponding log message so I have removed those. I think Ruby will just have to skip the one test that does need the close and write some prose test to cover that log message.

BorisDog (approved):
LGTM (minor comment)

source/connection-monitoring-and-pooling/tests/logging/connection-logging.yml
```diff
@@ -0,0 +1,194 @@
...
+              reason: "Connection pool was closed"
+
+          - level: debug
+            component: connection
+            data:
+              message: "Connection pool closed"
+              serverHost: { $$type: string }
+              serverPort: { $$type: [int, long] }
+
+  # This test exists to probide coverage of checkout failed and pool cleared events.
```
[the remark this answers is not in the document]
kmahar: thanks, fixed.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -998,6 +998,338 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
...
+     - String
+     - "Connection checked out"
+
+   * - driverConnectionId
+     - Int
+     - The driver-generated ID for the connection as defined in `Connection <#connection>`_.
+
+The unstructured form SHOULD be as follows, using the values defined in the structured format above to fill in placeholders as appropriate:
+
+  Connection checked out with driver-generated ID {{driverConnectionId}} to {{serverHost}}:{{serverPort}}
```
patrickfreed, line 1307: This is kind of a grammar nit, but I think this message reads a bit easier if the "with id" and "to \<address\>" phrases came before the verb here, since they're describing "Connection". If you drop the "with" part, the issue is a bit more obvious (e.g. "Connection checked out to a:12345" doesn't quite make sense).

Something like "Connection to a:12345 with driver-generated ID checked out" reads a bit better to me. The drawback being "Connection" and "checked out" are now separate, though I'm not sure if that's an issue or not. A more dramatic shift that doesn't have either problem could be something like "Connection checked out (driver-generated ID=1, destination=a:12345)".

WDYT?

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -998,6 +998,338 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
...
+     - String
+     - "Connection checkout failed"
+
+   * - reason
+     - String
+     - A string describing the reason checkout. The following strings MUST be used for each possible reason
+       as defined in `Events <#events>`_ above:
+
+       - Timeout: "Wait queue timeout elapsed without a connection becoming available"
+       - ConnectionError: "An error occurred while trying to establish a new connection"
```
patrickfreed, line 1278: It would probably be very useful if the actual error that caused the failure to occur were included here. That way the user could discern between a network issue, auth issue, OSCP responder problem, etc. I know the existing CMAP events are also deficient in this regard, but would this be a good opportunity to fix that, at least for logging messages?

kmahar: I agree that would be useful and I was thinking about it as I made these spec changes actually that it was funny we don't include the actual error, both here and for `ConnectionClosedEvent`. 
I'm happy to add the errors to the log messages, but a caveat worth mentioning first: for any driver implementing logging via a CMAP event listener, this is going to implicitly require adding the error to the corresponding monitoring events. I don't think that's a bad thing, but it would mean drivers would potentially end up with API that is not directly tested (it would at least be implicitly tested via logging though). we could file a follow up ticket around that though. alternatively, we could formally add it the errors to the events here too, but I think that the work to formally write tests around that (including possibly adding to the CMAP test format) would be too much scope creep. 
LMK if you think just adding the errors to the log messages and filing a follow up ticket around formally adding them to the events is reasonable.

patrickfreed: Adding it to just the log messages and deferring the event changes to a follow up ticket sgtm. I can file the ticket for it.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
kmahar: yeah I was having a hard time figuring out how to write the english for some of these and it shows here 😅 

re-reading this I had a couple thoughts:
1) I feel like the host/port should come before the ID as it feels more important given that IDs are only unique per-pool/endpoint
2) maybe "for" would be a better word than to?

I do think keeping the "Connection checked out" part together and at the start of the message is useful so that if you are skimming you can easily see what the message is about from the start of it.

given these WDYT of either of the following:
1) Connection checked out for endpoint a:12345 with driver-generated ID 123
2) Connection checked out: endpoint=a:12345, driver-generated ID=123.

(borrowing "endpoint" terminology from elsewhere in CMAP)

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
kmahar: I pushed changes to the spec and the tests for this. they pass in Rust. FYI @BorisDog - can you review and see if adding the errors to .NET log messages is possible and if the tests pass for you?

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
patrickfreed: I kinda like 2. It's consistent and pretty easy to parse visually (and programatically in case a given language can't do structured logs).

"endpoint" is spec language-y though,  I'm not sure if it ends up in our API or docs that often. What about just using "address" like the events do?

kmahar: cool, I updated to the second one (but with address instead of endpoint) and updated some of the other log messages to make the format/phrasing more consistent

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
patrickfreed: Filed https://jira.mongodb.org/browse/DRIVERS-2495

patrickfreed (approved):
LGTM!

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
BorisDog: Sorry for the delay. Yes adding errors is straightforward and the tests pass.

BorisDog (approved):
LGTM

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
kmahar: woo! thanks @BorisDog

PR #1325, opened 2022-10-19
DRIVERS-2411 add integration tests for Azure KMS

kevinAlbs:
# Summary
- Add integration tests for Azure KMS

# Background & Motivation

Tests requiring an Azure VM depend on the scripts in drivers-evergreen-tools here: https://github.com/mongodb-labs/drivers-evergreen-tools/pull/239

- [ ] Update changelog. **N/A. Only tests are changed.**
- [ ] Make sure there are generated JSON files from the YAML test files. **N/A. Only prose tests are changed. **
- [x] Test changes in at least one language driver. **Tested in C**
- [ ] Test these changes against all server versions and topologies (including standalone, replica set, sharded clusters, and serverless). **N/A. Tests do not depend on server version or topology.**

### DIFF
--- a/source/client-side-encryption/tests/README.rst
+++ b/source/client-side-encryption/tests/README.rst
@@ -2315,3 +2315,67 @@ Set ``X-MongoDB-HTTP-TestParams`` to ``case=slow``.
 
 The HTTP response from the ``fake_azure`` server will take at least 1000 seconds
 to complete. The request should fail with a timeout.
+
+19. Azure IMDS Credentials Integration Test
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Refer: `Automatic Azure Credentials <auto-azure_>`_
+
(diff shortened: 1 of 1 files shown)


source/client-side-encryption/tests/README.rst
```diff
@@ -2315,3 +2315,67 @@ Set ``X-MongoDB-HTTP-TestParams`` to ``case=slow``.
...
+      keyVaultClient: <setupClient>,
+      keyVaultNamespace: "keyvault.datakeys",
+      kmsProviders: { "azure": {} },
+   }
+
+Case 1: Failure
+```````````````
+
+Do not run this test case in an Azure environment with an attached identity.
+This may be run in an AWS EC2 instance.
```
DmitryLukyanov, line 2341: can this be just a regular env without any configuration? If so, we can get rid of special EG configuration for this case

DmitryLukyanov (approved):
Checked with c# driver. Please take a look at my comment, other than this LGTM

vector-of-bool (approved):

comandeo (approved):

PR #1327, opened 2022-10-25
DRIVERS-2385 useMultipleMongoses false for errorResponse tests

benjirewis:
DRIVERS-2385

- [ ] Update changelog. (n/a)

Adds `useMultipleMongoses: false` to the new `errorResponse` spec tests added in DRIVERS-2385. Most of the tests use fail points that do not work correctly with multiple mongoses.

See GO POC [here](https://github.com/mongodb/mongo-go-driver/pull/1097).

### DIFF
--- a/source/collection-management/tests/modifyCollection-errorResponse.yml
+++ b/source/collection-management/tests/modifyCollection-errorResponse.yml
@@ -5,6 +5,7 @@ schemaVersion: "1.12"
 createEntities:
   - client:
       id: &client0 client0
+      useMultipleMongoses: false
       observeEvents: [ commandStartedEvent ]
   - database:
       id: &database0 database0
--- a/source/collection-management/tests/modifyCollection-errorResponse.yml
+++ b/source/collection-management/tests/modifyCollection-errorResponse.yml
@@ -5,7 +5,6 @@ schemaVersion: "1.12"
 createEntities:
   - client:
       id: &client0 client0
-      useMultipleMongoses: false
       observeEvents: [ commandStartedEvent ]
   - database:
       id: &database0 database0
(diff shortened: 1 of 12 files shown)


source/collection-management/tests/modifyCollection-errorResponse.yml
```diff
@@ -5,6 +5,7 @@ schemaVersion: "1.12"
 createEntities:
   - client:
       id: &client0 client0
+      useMultipleMongoses: false
```
jmikola, line 8: Was this necessary? This test never used `configureFailPoint`.

I notice you didn't make any changes to `aggregate-merge-errorResponse.yml`; however, that test was already skipping sharded clusters.

jmikola (approved):
One question about `modifyCollection-errorResponse.yml`. If you can verify that's not required feel free to remove the change and merge. If it is required, just let me know why and LGTM.

Either way, the trailing period in the PR title and commit message can be removed.

source/collection-management/tests/modifyCollection-errorResponse.yml
benjirewis: Ah it's not, great point. We don't actually run this test in the Go driver (see [GODRIVER-2587](https://jira.mongodb.org/browse/GODRIVER-2587)), so I just added it without thinking. I did _not_ add `useMultipleMongoses: false` to aggregate-merge-errorResponse` because it skips sharded clusters.

jmikola: @benjirewis: I think `useMultipleMongoses: false` was erroneously added to the `findOneAndUpdate` test, too, which doesn't use `configureFailPoint`. See: https://github.com/mongodb/specifications/pull/1329

PR #1328, opened 2022-10-26
add section to logging spec about timestamps

baileympearson:

- [ ] Update changelog.

### DIFF
--- a/source/logging/logging.rst
+++ b/source/logging/logging.rst
@@ -311,6 +311,10 @@ logging MUST use the exact names and casing specified for the names of fields in
 This will be easier for our support team since the names will be consistent across languages, and
 will simplify writing language-agnostic tooling to search through and parse structured logs.
 
+Including Timestamps in Log Messages
+------------------------------------
+Drivers MAY add timestamps to their log messages if one will not be added automatically by the logging framework(s) they use.
+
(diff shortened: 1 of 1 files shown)


source/logging/logging.rst
```diff
@@ -311,6 +311,10 @@ logging MUST use the exact names and casing specified for the names of fields in
 This will be easier for our support team since the names will be consistent across languages, and
 will simplify writing language-agnostic tooling to search through and parse structured logs.
 
+Including Timestamps in Log Messages
+------------------------------------
+Drivers MAY add timestamps to their log messages if one will not be added automatically by the logging framework(s) they use.
```
baileympearson, line 316: question: should be perscriptive about what type the timestamp is (i.e., BSON timestamp or something else) or leave it up to drivers?

kmahar: I think we can leave it as flexible. all of the values included in existing log messages generally have flexible types depending on what drivers can support, so it would be consistent to allow flexibility here. in Rust we can only attach a handful of primitive types to structured log messages, so if we didn't get this for free, we'd have to use a string. but if you could e.g. use a JS object for Node that would probably be nicer than a string.

kmahar (approved):

PR #1330, opened 2022-10-27
DRIVERS-2449 Omit mongocryptd client

DmitryLukyanov:

- [x] Update changelog.


source/client-side-encryption/tests/README.rst
```diff
@@ -2379,3 +2379,43 @@ This should successfully load and use the Azure credentials of the service accou
 attached to the virtual machine.
 
 Expect the key to be successfully created.
+
+20. Bypass Spawning mongocryptd client when shared library is loaded
```
kevinAlbs, line 2383:
```suggestion
20. Bypass creating mongocryptd client when shared library is loaded
```

"Spawning" is used to describe spawning the mongocryptd process. Suggesting using "creating" to refer to creating the client to mongocryptd.

kevinAlbs (approved):
LGTM with a small wording change.

Tested in [Go](https://github.com/kevinAlbs/mongo-go-driver/commits/D2449).

PR #1332, opened 2022-10-28
DRIVERS-2475 Consider Using AWS SDK for Auth

blink1073:

- [x] Update changelog.


JamesKovacs (commented):
Should we make this more generic and refer to a Cloud Provider's SDK?

blink1073:
> Should we make this more generic and refer to a Cloud Provider's SDK?

As I understand it, our use of other cloud providers' SDK is limited to a single request to a fixed endpoint.  I think it would make sense to hold off for now and re-evaluate if it gets more complicated.

JamesKovacs (approved):
LGTM

PR #1336, opened 2022-11-02
DRIVERS-2489 Improve test coverage for retryable handshake errors

eramongodb:

- [x] ~Update changelog.~ N/A (Test Files Only)

## Description

This PR resolves DRIVERS-2489.

DRIVERS-2489 is motivated by the possibility of a given Driver's implementation of DRIVERS-746 possibly failing to account for operations that may be implemented via alternate code paths due to the Driver's architecture. Increasing coverage of the operations tested reduces the possibility of such alternative code paths being unaccounted for by implementations of DRIVERS-746.

Test operations for `handshakeError.yml` unified spec tests were expanded to cover the following Retryable Writes operations ([per the spec](https://github.com/mongodb/specifications/blob/b0caa1683953cd85031a0a32fffd1ad660bd07f8/source/retryable-writes/retryable-writes.rst#supported-write-operations)):

- insertOne
- updateOne
- replaceOne
- deleteOne
- findOneAndDelete
- findOneAndReplace
- findOneAndUpdate
- insertMany
- bulkWrite (insertOne)

and the following Retryable reads operations ([per the spec](https://github.com/mongodb/specifications/blob/082306075eb2d125c47503e94fabc42f00a16784/source/retryable-reads/retryable-reads.rst#supported-read-operations)):

- find
- aggregate ($match)
- distinct
- estimatedDocumentCount
- countDocuments
- listDatabases
- listCollections
- listIndexes

The `Collection.count()` operation is omitted due to the API being deprecated.

~The read operation helpers in the change streams specification (`*.watch()`) are omitted due to the complexity of implementations of change streams (such as multiple potential connections to the server for a given operation) making deterministic cross-Driver behavior difficult to validate.~ This is now being tested after https://github.com/mongodb/specifications/pull/1336#issuecomment-1306260705.

Tests that validate a given server connection only attempts a retry _once_ per operation were not included due to the `failPoint` operation not supporting the setting of multiple fail points as required to setup an appropriate sequence of network errors (fail handshake -> successful retry -> fail operation -> no retry). Testing this scenario would currently require a prose test that registers an APM callback on detection of initial handshake failure that enables the next fail point to fail the subsequent operation. However, depending on the Driver's implementation of APM callbacks (which may be asynchronous), this may not be possible without a race condition between the registration of the 2nd fail point command and the execution of the operation that must be failed. Therefore, specification of these prose tests have been omitted for now.


ShaneHarvey (commented):
Nice work! A common pattern when adding spec tests for all operations like this is to autogenerate the tests via a script. For example:
- https://github.com/mongodb/specifications/blob/4e17d57/.github/workflows/json-regenerate-check.yml#L29-L32

https://github.com/mongodb/specifications/blob/d68e85c/source/client-side-operations-timeout/etc/generate-basic-tests.py

Could you add something similar for these tests? The benefit is that it's easier to review and maintain going forward, eg adding a new error case or operation.

source/retryable-reads/tests/unified/handshakeError.yml
```diff
@@ -126,4 +127,681 @@ tests:
...
+              databaseName: *databaseName
+          - commandStartedEvent:
+              command:
+                getMore: { $$type: [int, long] }
+                collection: *collectionName
+                batchSize: 2
+              commandName: getMore
+              databaseName: *databaseName
+
+  - description: "aggregate succeeds after retryable handshake network error (ShutdownInProgress)"
```
kevinAlbs, line 192:
```suggestion
  - description: "aggregate succeeds after retryable handshake server error (ShutdownInProgress)"
```

The `aggregate succeeds after retryable handshake network error` case tests an error due to the server closing the connection. This case tests an error reply from the server. Suggest updating all `ShutdownInProgress` cases.

source/retryable-reads/tests/unified/handshakeError.yml
```diff
@@ -126,4 +127,681 @@ tests:
...
+        object: *database0
+        arguments:
+          commandName: ping
+          command: { ping: 1 }
+        expectError:
+          isError: true
+
+      - name: listDatabases
+        object: *client0
+        ignoreResultAndError: true # Assume success.
```
kevinAlbs, line 564:
```suggestion
```

IIUC the absence of `ignoreResultAndError` and `expectError` will [assert no error occurred](https://github.com/mongodb/specifications/blob/c796a746327af27f9070bd1791ced6bce9176453/source/unified-test-format/unified-test-format.rst#executing-an-operation).

> If operation.expectError is specified, the test runner MUST assert that the operation yielded an error; otherwise, the test runner MUST assert that the operation did not yield an error.

Suggested change applies to other operations with the `Assume success` comment.

kevinAlbs (commented):
Thank you for contributing these tests back to the specification.

I support generating the tests in scripts. Though I think this PR is still a significant improvement to test coverage, and would not block merge if adding generation scripts is a significant effort.

I want another driver to run these tests before merging, especially since C cannot test the cmap events. I was unable to try these tests with Go. The Go driver has not implemented retryability for handshake errors ([GODRIVER-2191](https://jira.mongodb.org/browse/GODRIVER-2191)). When opening for review, one of the retryable spec owners may be able to run this in their own driver to verify.

source/retryable-reads/tests/unified/handshakeError.yml
```diff
@@ -126,4 +127,681 @@ tests:
...
+          commandName: ping
+          command: { ping: 1 }
+        expectError:
+          isError: true
+
+      - name: aggregate
+        object: *collection0
+        arguments:
+          pipeline: [{ $match: { _id: { $gt: 1 } } }]
+          batchSize: 2
```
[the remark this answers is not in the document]
eramongodb: I believe it can. Will remove.

eramongodb:
The test files were updated to be generated by new `generate-handshakeError-tests.py` scripts which were based on the `generate-basic-tests.py` file. Given the identical format of the Python scripts for retryable-reads and retryable-writes (just a different selection of operations), I can remove the duplication by moving them to a new `$ROOT/etc` directory if preferable.

The tests were refactored to utilize `commandFailedEvent` and `commandSucceededEvent` events to verify failure or success of operations instead of `expectResult` or `expectError` for uniformity across all operations being tested. As a result of this change, more operations are now covered by the test files, including `createChangeStream` which was being skipped prior.

The `count` and `findOne` operations are omitted due to being deprecated and optional to implement respectively.

source/retryable-reads/tests/etc/templates/handshakeError.yml.template
```diff
@@ -0,0 +1,155 @@
...
+          - commandStartedEvent:
+              command: { ping: 1 }
+              databaseName: *databaseName
+          - commandFailedEvent:
+              commandName: ping
+          - commandStartedEvent:
+              commandName: {{operation.command_name}}
+          - commandSucceededEvent:
+              commandName: {{operation.command_name}}
+          {%- if operation.operation_name == "createChangeStream" %}
```
ShaneHarvey, line 98: There shouldn't be a killCursors here (and same for the second test):
```
FAIL: test_createChangeStream_succeeds_after_retryable_handshake_network_error (__main__.TestUnifiedHandshakeError)
  File "/Users/shane/git/mongo-python-driver/test/unified_format.py", line 1647, in test_case
    self.run_scenario(spec)
  File "/Users/shane/git/mongo-python-driver/test/unified_format.py", line 1600, in run_scenario
    self._run_scenario(spec, uri)
  File "/Users/shane/git/mongo-python-driver/test/unified_format.py", line 1630, in _run_scenario
    self.check_events(expect_events)
  File "/Users/shane/git/mongo-python-driver/test/unified_format.py", line 1554, in check_events
    self.assertEqual(len(actual_events), len(events), actual_events)
AssertionError: 4 != 6 Actual commands: [
<CommandStartedEvent ('localhost', 27017) db: 'retryable-reads-handshake-tests', command: 'ping', operation_id: 681910962, service_id: None>,
<CommandFailedEvent ('localhost', 27017) command: 'ping', operation_id: 681910962, duration_micros: 276, failure: {'errmsg': 'connection closed', 'errtype': 'OSError'}, service_id: None>,
<CommandStartedEvent ('localhost', 27017) db: 'retryable-reads-handshake-tests', command: 'aggregate', operation_id: 1260973671, service_id: None>,
<CommandSucceededEvent ('localhost', 27017) command: 'aggregate', operation_id: 1260973671, duration_micros: 4148, service_id: None>]
```

Since the aggregate failed there's no cursor to kill.

source/retryable-writes/tests/etc/templates/handshakeError.yml.template
```diff
@@ -0,0 +1,155 @@
...
+  #   - Creates a connection in the connection pool that must be closed.
+  # - Ping operation.
+  #   - Triggers failpoint (first time).
+  #   - Closes the connection made by the fail point operation.
+  # - Test operation.
+  #   - New connection is created.
+  #   - Triggers failpoint (second time).
+  #   - Tests whether operation successfully retries the handshake and succeeds.
+{% for operation in operations %}
+  - description: "{{operation.operation_name}} succeeds after retryable handshake network error"
```
ShaneHarvey, line 55: Not 100% sure why but in Python this test fails for bulkWrite with a dup key error on `_id:1`:
```
ERROR: test_bulkWrite_succeeds_after_retryable_handshake_network_error (__main__.TestUnifiedHandshakeError)
  File "/Users/shane/git/mongo-python-driver/test/unified_format.py", line 1647, in test_case
    self.run_scenario(spec)
  File "/Users/shane/git/mongo-python-driver/test/unified_format.py", line 1600, in run_scenario
    self._run_scenario(spec, uri)
  File "/Users/shane/git/mongo-python-driver/test/unified_format.py", line 1624, in _run_scenario
    self.run_operations(spec["operations"])
  File "/Users/shane/git/mongo-python-driver/test/unified_format.py", line 1535, in run_operations
    self.run_entity_operation(op)
  File "/Users/shane/git/mongo-python-driver/test/unified_format.py", line 1276, in run_entity_operation
    result = cmd(**dict(arguments))
  File "/Users/shane/git/mongo-python-driver/pymongo/_csot.py", line 105, in csot_wrapper
    return func(self, *args, **kwargs)
  File "/Users/shane/git/mongo-python-driver/pymongo/collection.py", line 531, in bulk_write
    bulk_api_result = blk.execute(write_concern, session)
  File "/Users/shane/git/mongo-python-driver/pymongo/bulk.py", line 513, in execute
    return self.execute_command(generator, write_concern, session)
  File "/Users/shane/git/mongo-python-driver/pymongo/bulk.py", line 394, in execute_command
    _raise_bulk_write_error(full_result)
  File "/Users/shane/git/mongo-python-driver/pymongo/bulk.py", line 136, in _raise_bulk_write_error
    raise BulkWriteError(full_result)
pymongo.errors.BulkWriteError: batch op errors occurred, full error: {'writeErrors': [{'index': 0, 'code': 11000, 'keyPattern': {'_id': 1}, 'keyValue': {'_id': 1}, 'errmsg': 'E11000 duplicate key error collection: retryable-writes-handshake-tests.coll index: _id_ dup key: { _id: 1 }', 'op': {'_id': 1}}], 'writeConcernErrors': [], 'nInserted': 0, 'nUpserted': 0, 'nMatched': 0, 'nModified': 0, 'nRemoved': 0, 'upserted': []}
```

Perhaps we can change the `_id:1` doc to `x:1` to match the insert/insert_many tests?

ShaneHarvey (changes requested):

source/retryable-reads/tests/etc/templates/handshakeError.yml.template
ShaneHarvey: Oh I see. The killCursors could come from the driver closing the cursor because it's not saved with saveResultAsEntity. In Python we cleanup these unsaved cursors at the very end of the test so they don't show up in the command events. Perhaps we can ignore killCursors events in these tests to avoid the potential inconsistency across drivers?

source/retryable-reads/tests/etc/generate-handshakeError-tests.py
```diff
@@ -0,0 +1,118 @@
...
+COLLECTION_OPERATIONS = COLLECTION_READ_OPERATIONS + COLLECTION_WRITE_OPERATIONS
+
+# Session and GridFS operations are generally tested in other files, so they're not included in the list of all
+# operations. Individual generation functions can choose to include them if needed.
+OPERATIONS = CLIENT_OPERATIONS + DB_OPERATIONS + COLLECTION_OPERATIONS
+
+RETRYABLE_READ_OPERATIONS = [op for op in OPERATIONS if op.operation_name in
+                             ['find', 'findOne', 'aggregate', 'distinct', 'count', 'estimatedDocumentCount', 'countDocuments', 'createChangeStream', 'listDatabases',
+                              'listDatabaseNames', 'listCollections', 'listCollectionNames', 'listIndexes']
+                             ]
```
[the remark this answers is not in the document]
eramongodb: No; `RETRYABLE_READ_OPERATIONS` is a filter of `OPERATIONS`. There are more `OPEARTIONS` than there are `RETRYABLE_READ_OPERATIONS`.

source/retryable-reads/tests/etc/generate-handshakeError-tests.py
```diff
@@ -0,0 +1,118 @@
...
+    Operation('distinct', 'distinct', 'collection',
+              ['fieldName: x', 'filter: {}']),
+    Operation('find', 'find', 'collection', ['filter: {}']),
+    # Operation('findOne', 'find', 'collection', ['filter: {}']), # Optional.
+    Operation('listIndexes', 'listIndexes', 'collection', []),
+    Operation('listIndexNames', 'listIndexes', 'collection', []),
+    Operation('createChangeStream', 'aggregate',
+              'collection', ['pipeline: []']),
+]
+COLLECTION_WRITE_OPERATIONS = [
```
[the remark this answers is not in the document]
eramongodb: Would this be preferable to using a single `generate-handshakeError-tests.py` to generate the tests for both Reads and Writes?

source/retryable-reads/tests/etc/templates/handshakeError.yml.template
eramongodb: I think we can use a `saveResultAsEntity` to ensure consistent `killCursors` behavior across Drivers.

source/retryable-writes/tests/etc/templates/handshakeError.yml.template
eramongodb: I overlooked a copy-paste error when copying the reads template to writes. Will address.

source/retryable-reads/tests/etc/generate-handshakeError-tests.py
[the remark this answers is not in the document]
ShaneHarvey: Oh I missed that question. Yeah I think adding a single script for this makes since.

source/retryable-reads/tests/etc/generate-handshakeError-tests.py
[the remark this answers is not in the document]
ShaneHarvey: Can you name one such operation?

source/retryable-reads/tests/etc/templates/handshakeError.yml.template
ShaneHarvey: Good call.

source/retryable-reads/tests/etc/generate-handshakeError-tests.py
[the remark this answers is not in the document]
eramongodb: I was confusing the current script with the original template where `OPERATIONS` included both read and write operations. This will once again be the case with the use of a single generator script.

ShaneHarvey (approved):
Tests pass in Python. Thanks!

kevinAlbs (approved):
LGTM with addition of `listIndexNames`. Nicely done.

source/etc/generate-handshakeError-tests.py
```diff
@@ -0,0 +1,155 @@
...
+                              'distinct',
+                              'count',
+                              'estimatedDocumentCount',
+                              'countDocuments',
+                              'createChangeStream',
+                              'listDatabases',
+                              'listDatabaseNames',
+                              'listCollections',
+                              'listCollectionNames',
+                              'listIndexes',
```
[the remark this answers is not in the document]
kevinAlbs: Disregard. `listIndexNames` is [specified as optional](https://github.com/mongodb/specifications/blob/d268119c9d4354f60e59d2e798c3cb3ee5323a29/source/enumerate-indexes.rst#getting-index-names) for drivers to implement and is not implemented in the C driver.

> [Getting Index Names](https://github.com/mongodb/specifications/blob/master/source/enumerate-indexes.rst#id11)
> Drivers MAY implement a method to enumerate all indexes, and return only the index names.

isabelatkinson (approved):
lgtm! tests pass in Rust

eramongodb:
For consistency with other unified spec tests, optional operations such as `listCollectionNames` are _assumed_ to be implemented by Drivers to maximize coverage. Drivers may elect to skip these tests if they do not support the operations. The addition of these operations may fail tests that previously succeeded, so I recommend Drivers re-run tests to verify if the tests still pass with these latest changes.

ShaneHarvey (approved):
Tests still pass in python.

eramongodb:
After adding some skip conditions for unsupported helpers, tests pass in C Driver as well. Will merge this PR later today if no further objections are raised.

isabelatkinson:
👍 tests still pass in Rust

PR #1341, opened 2022-11-09
DRIVERS-1671 Add server selection log message and tests

kmahar:
Adds a new server selection log message along with some corresponding unified tests. These tests pass in the Rust driver.


source/server-selection/tests/logging/server-selection-logging.yml
```diff
@@ -0,0 +1,218 @@
...
+    runOnRequirements:
+      - topologies:
+          - single
+          - replicaset
+        minServerVersion: "4.0" # min version for failCommand on mongod
+      - topologies:
+          - sharded
+        minServerVersion: "4.1.5" # min version for failCommand on mongos
+    operations:
+      # sleep to ensure there is sufficient time for all hosts to be initially discovered.
```
kmahar, line 49: I don't love this solution, but I was running into issues where depending on timing, the driver would not have a complete view of the topology yet by the time the test started, e.g. it might think the servers were [primary, unknown, unknown] when server selection for the `updateOne` started, which would lead to extraneous log messages because the driver would re-log in the server selection loop each time it changed a secondary's types from unknown -> secondary. 
that said, I'm realizing that I might be able to write this test in topology-specific ways and have each one start with a `waitForEvent` for the server descriptions changing. I'm not sure whether events that could happen prior to the operations starting are guaranteed to be observable though 🤔 it's possible that the driver ends up discovering everything before we get to the operations. I asked about this in #driver-devs.

source/server-selection/server-selection.rst
```diff
@@ -849,7 +849,10 @@ For multi-threaded clients, the server selection algorithm is as follows:
 
 5. If there are any suitable servers, filter them according to `Filtering
    suitable servers based on the latency window`_ and continue to the next step;
-   otherwise, goto Step #9.
+   otherwise, log a `"Waiting for suitable server to become available" message`_
+   if one has not already been logged for this operation OR if the topology
```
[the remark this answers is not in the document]
kmahar: I was thinking it's the same definition of "changed" that causes us to emit a SDAM monitoring event saying that the topology changed, which I believe happens whenever a server is added, removed, or the new description is not equal to the old one per the equality definition in [SDAM](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#server-description-equality), which omits RTT and some other fields. 
if we want to keep this behavior then I will make that more explicit.

regarding how beneficial this is, @patrickfreed and I discussed this some and thought that it could be useful to be able to identify cases where an operation continued to fail to select a server as the topology changed. I was thinking it would be nice in that it means we automatically get a view into SDAM state changes that are happening during the time period we are waiting for a suitable server. 

on the flipside, if there are a lot of operations going at once, we might end up with a lot of redundant logs since each operation would be re-logging on every topology change. it might be better to just tell the user to turn on SDAM logging instead, so there is a single log message for each topology change.

source/server-selection/server-selection.rst
```diff
@@ -1180,11 +1185,85 @@ server without an explicit ``start_request`` (or comparable) method call.
...
+       <../sessions/driver-sessions.rst#how-to-check-whether-a-deployment-supports-sessions>`__.
+
+   * - operation
+     - String
+     - The name of the operation for which a server is being selected. When server selection is
+       being performed to select a server for a command, this MUST be the command name. For any
+       non-command server selection this MUST be a string describing the reason a server is being
+       selected, e.g. "Check for sessions support".
+
+   * - topology
```
patrickfreed, line 1251: Would we want to use `topologyDescription` as the name for this field here to signify that its the same thing as is logged in SDAM events?

Also, as an alternative, what do you think about logging the `topologyVersion` here instead? That might cut down on duplicated logging of the topology description, though it would require log parsers to cross reference with SDAM logs to see what exactly the topology contains.

source/server-selection/server-selection.rst
```diff
@@ -849,7 +849,10 @@ For multi-threaded clients, the server selection algorithm is as follows:
```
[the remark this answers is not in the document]
kmahar: adding start and succeeded messages seems reasonable. debug level? they seem equivalent verbosity-wise to command started and command succeeded.

regarding server selection failure, I've generally tried to avoid adding log messages for cases where the user will always see the error via API (and therefore likely in application-generated logs) anyway. I know there are some other "failure" messages e.g. "command failed" and "connection checkout failed", but retryability makes those failures not a 1:1 mapping with driver errors. I think(?) that isn't true of server selection though - are there any cases where server selection can fail but the driver doesn't already throw an error?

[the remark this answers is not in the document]
jyemin: Do we need a policy on whether to log failures?  In Java driver we took the position that we would not try to catch/log/rethrow exceptions, on the grounds that the application is likely to log exceptions.   This is generally true in Java applications but probably not in other ecosystems.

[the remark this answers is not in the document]
kmahar: I think it would be good to have a policy on this in the spec. at minimum, I think we should say drivers must not log anything that could (or will definitely) also be reported as an error via API at anything more severe than `debug` level to avoid the possibility of duplicate error-level messages if the app logs them too.

optionally, we could also say that anything which will _definitely_ be reported as an error (like server selection failure) should not be logged at all. for log messages that sometimes (but not always) correspond to an error we throw, it might be overly complicated to ask drivers to only log them if they are not also getting thrown, since the driver likely could not tell that at the place the error condition initially arises. 

I started a thread in driver-devs to ask about application error logging in other language ecosystems. for my languages (copied from the post for posterity): 
* in Swift, per the guidelines [here](https://www.swift.org/server/guides/libraries/log-levels.html) libraries should always prefer to report errors via API and let the application handle logging it.
* in Rust, I’ve seen a bit of a mix of opinions; I see some examples of logging errors but at debug level and also reporting them via API, but also see some people express the opinion that errors should be logged where they are handled, at the app level.

source/server-selection/server-selection.rst
```diff
@@ -1180,11 +1185,85 @@ server without an explicit ``start_request`` (or comparable) method call.
...
+   * - selector
+     - String
+     - String representation of the selector being used to select the server. This can be
+       a read preference or an application-provided custom selector. The exact content of
+       is flexible depending on what the driver is able to log. At minimum, when the selector
+       is a read preference this string MUST contain all components of the read preference,
+       and when it is an application-provided custom selector the string MUST somehow indicate
+       that it is a custom selector.
+
+   * - requestId
```
BorisDog, line 1237: .Net initiates the operation info only after the server was selected, `requestId` is generated only before the message goes on the wire. But we do have `operationId` at that point.

source/server-selection/server-selection.rst
```diff
@@ -1180,11 +1185,85 @@ server without an explicit ``start_request`` (or comparable) method call.
...
+     - Int
+     - The driver-generated ID for the request a server is being selected for. Optional; this
+       field will not be present in cases where server selection does not correspond to an operation,
+       e.g. when the driver performs server selection to check if a `deployment supports sessions
+       <../sessions/driver-sessions.rst#how-to-check-whether-a-deployment-supports-sessions>`__.
+
+   * - operation
```
BorisDog, line 1244: Currently in .Net operation information is not propagated into server selection. Do you think we'll need it for the rest of server selection logs?

kmahar: what I currently have proposed locally is that all the additional messages (started, succeeded, failed) would have the operation name and operation ID as well.

source/server-selection/server-selection.rst
BorisDog: Adding operation information looks feasible, but it would be a significant refactoring to generate `requestId` before the server selection. And it's even more complicated with retries.
Wondering what is the complexity of this in other languages, particularly in Java.

kmahar: cc @jyemin re complexity for Java?

if it ends up being too big an undertaking, we could make the requestId a SHOULD and not a MUST, and split out the assertions for its presence into separate tests so you could skip them. for those who can do it easily enough, I think it's worthwhile to keep.

jyemin: Java has neither operationId (at all) or requestId (until much further down the stack).  In practice, requestId seems problematic anyway due to operations like bulk writes that can be split into multiple commands.  It wouldn't be that hard to add operationId though.

kmahar: >  In practice, requestId seems problematic anyway due to operations like bulk writes that can be split into multiple commands

if a driver includes both the requestId and operationId for bulk writes, how would this be problematic? it's the same thing we do for command events/log messages

jyemin: A bulk write looks like this (ignoring many details)

```
     select server
          <log server selection events> 

     while moreSplitPointsAvailable
          execute insert/update/delete command
              <log/publish events for command started>
              <log/publish events for command succeeded>
```    
     
I don't see how a driver would have a single requestId available during server selection.  

Moreover, requestId is part of the wire protocol and not something that should be available at the operation layer of abstraction.

kmahar: Jeff and I talked about this offline. It turns out at least some drivers (Java, .NET) only perform server selection once for a bulk write, and the request ID is generated all the way down at the point of creating a wire protocol message.
For Rust however, (while we don't have bulk write we do have `insert_many` at least), we re-perform server selection for each batch. So for Rust, there is a 1:1 mapping of requestId to server selection attempt, but this won't necessarily hold up for other drivers.
I can't find anything in the specs on how drivers should handle server selection for bulk writes, so I've filed DRIVERS-2502 about addressing this in the new bulk write spec.

Given all this, I don't think it makes sense to put the requestId in these messages since some (and possibly most) drivers do not have a 1:1 mapping of server selection attempt to command. That said, I will still update the messages to include the operationId, for drivers that generate those.

source/server-selection/server-selection.rst
```diff
@@ -1180,11 +1185,85 @@ server without an explicit ``start_request`` (or comparable) method call.
...
+     - "Waiting for suitable server to become available"
+
+   * - selector
+     - String
+     - String representation of the selector being used to select the server. This can be
+       a read preference or an application-provided custom selector. The exact content of
+       is flexible depending on what the driver is able to log. At minimum, when the selector
+       is a read preference this string MUST contain all components of the read preference,
+       and when it is an application-provided custom selector the string MUST somehow indicate
+       that it is a custom selector.
```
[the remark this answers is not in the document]
patrickfreed: I think the separate top-level keys might end up hurting readability actually. Sticking with the language's toString implementation seems fine to me :+1:

kmahar:
I realized I forgot to add operationId tests for drivers that have operationIds. writing some of those now. but the rest of this should be ready for another look

source/server-selection/server-selection.rst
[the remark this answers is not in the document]
jyemin: Can you paste the logs from your test somewhere so we can take a look?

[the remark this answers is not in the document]
patrickfreed: > I think I would prefer to remove the re-logging aspect here in favor of having users use server selection + SDAM logging in tandem to get the desired effect.

SGTM

jyemin (commented):
Dropped a comment on an open thread requesting more information

source/server-selection/server-selection.rst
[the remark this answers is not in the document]
kmahar: see [here](https://gist.github.com/kmahar/f3cfdbeef09cf78c6e4e18564511a0e5) for some sample messages running with min severity level info during a failover. I had to re-generate the messages because I had restarted my computer and so lost my terminal history. this time I got slightly more interesting results: you can see in the diff for M1 vs M2 that in M2 the third server has changed its view of who the primary is and also gains an election ID. 
(note: at first glance it looks like the tag sets change, but this is just because we store them in a hashmap and so the ordering for printing out the values is not deterministic.) 

nonetheless, those changes would also be captured in a "topology description changed" log message.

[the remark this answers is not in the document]
kmahar: I didn't include any debug level messages there since it's quite noisy due to the way my program continually loops (it's hard to time starting the program with the failover starting), but if you want to see any of those let me know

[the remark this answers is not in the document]
jyemin: > I think I would prefer to remove the re-logging aspect here in favor of having users use server selection + SDAM logging in tandem to get the desired effect.

The examples you linked only contain server selection log messages.  Can you show what they look like in tandem?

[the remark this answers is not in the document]
kmahar: talked about on slack with Jeff, for posterity: Rust doesn't have SDAM logging yet so that's why those aren't included 
I went ahead and updated the spec so that the "waiting" messages only get logged once and updated tests accordingly

kmahar:
tests for operationId have been added. also, per conversation with @BorisDog on Slack, the `waitForEvent` expectations in the replica set tests were unnecessarily strict, so those have been removed and/or relaxed where necessary so that the tests can run on replica sets both with and without arbiters.

kmahar:
latest commit: I had to add back a `waitForEvent` to the test "server selection fails due to unsatisfiable read preference".  there was a race condition around whether the driver has determined that the server supports sessions by the time the `find` starts; in cases where it is not yet determined, an extra server selection is performed to check for sessions support, which led to extra log messages and the test failing. cc @BorisDog

patrickfreed (commented):
looks good! just a few minor things

source/server-selection/server-selection.rst
```diff
@@ -1180,11 +1187,200 @@ server without an explicit ``start_request`` (or comparable) method call.
...
+     - Int
+     - The driver-generated operation ID. Optional; only present if the driver generates
+       operation IDs and this command has one.
+
+   * - operation
+     - String
+     - The name of the operation for which a server is being selected. When server selection is
+       being performed to select a server for a command, this MUST be the command name. For any
+       non-command server selection this MUST be a string describing the reason a server is being
+       selected, e.g. "Check for sessions support".
```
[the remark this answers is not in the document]
kmahar: nope, given the decision to make this work a dependency of DRIVERS-2337 we can get rid of this. removed

kmahar:
Copying my slack message here for posterity: 

> as Boris has been implementing the log messages he found some of them were flaky/failing due to the way that drivers perform server selection to determine sessions support status; depending on timing there can be extra log messages generated since there are extra server selections performed for any operations that are sent when the driver is unsure if the topology supports sessions.
this issue didn’t turn up in rust because of a bug I found today where we don’t follow the spec: [RUST-1557](https://jira.mongodb.org/browse/RUST-1557).
Daria is planning to pick up [DRIVERS-2337](https://jira.mongodb.org/browse/DRIVERS-2337) in January, which would remove this “server selection to check for sessions support” logic from drivers altogether. rather than trying to work around this issue by e.g. adding a way to filter out certain log messages when matching, or by trying to encode all the messages that server selection support checks would generate, I think the best path forward to address the issue Boris encountered is to make drivers’ implementations of DRIVERS-1671 dependent on their implementing DRIVERS-2337. then the tests don’t need to be concerned with the extraneous messages.
as for the PR that is up now, I think for a driver that has done DRIVERS-2337 the tests would pass as-is, but I will wait to merge until we’ve been able to validate that in a couple drivers. (I will plan to do the Rust impl of DRIVERS-2337 as soon as it is ready for work to confirm there.)

source/server-selection/tests/logging/replica-set.yml
```diff
@@ -0,0 +1,223 @@
...
+          - level: debug
+            component: serverSelection
+            data:
+              message: "Server selection failed"
+              selector: { $$exists: true }
+              operation: find
+              topologyDescription: { $$exists: true }
+              failure: { $$exists: true }
+
+  - description: "Waiting message is logged when a suitable server is not immediately available"
```
BorisDog, line 130: This test is racy (server can be rediscovered before selection), so I think having "Failure due to unreachable server" instead is sufficient. cc @kmahar.

BorisDog: Sorry, looks like removing sessions check helped with the timings. Ran the tests for a while, not issues so far.

BorisDog: It's still racy after all, [example](https://evergreen.mongodb.com/test_log/dot_net_driver_unsecure_tests__version~5.0_os~windows_64_topology~sharded_cluster_auth~noauth_ssl~nossl_test_net472_patch_1f761191fb44d0c53dcbfe5d739a20f057be6c7e_6466c28ca4cf472554c992b6_23_05_19_00_28_26/0?test_name=05b3b956d9eec9c7ad72aa7f863e74f5&group_id=#L1).
Although it's quite rare 2% of cases, and only on [EG](https://spruce.mongodb.com/version/6466c28ca4cf472554c992b6/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC). 
I think` Failure due to unreachable server` is sufficient to test this code path.

kmahar: ah I see. I tried to do the high heartbeat frequency to prevent that but maybe it still isn't enough on occasion. I agree if we are covering the waiting message via the other test then it seems ok to remove this one.

kmahar:
made obselete by #1424

PR #1343, opened 2022-11-10
Clarify requirement of TLS options API

kevinAlbs:
No DRIVERS ticket since this is a clarification.

- [ ] ~Update changelog.~
- [ ] ~Make sure there are generated JSON files from the YAML test files.~
- [ ] ~Test changes in at least one language driver.~
- [ ] ~Test these changes against all server versions and topologies (including standalone, replica set, sharded clusters, and serverless).~


abr-egn (approved):
Thanks!

PR #1344, opened 2022-11-10
Clarify driver-specific null omission

andymina:
Allows driver's to decide whether null values should be omitted outside of common specification components.

- [x] Update changelog.


kmahar (approved):
lgtm! thanks @andymina

PR #1346, opened 2022-11-14
DRIVERS-2489 Properly skip createChangeStream tests on serverless

ShaneHarvey:
This PR fixes an issue with the tests added in https://github.com/mongodb/specifications/pull/1336

Change streams don't yet work on serverless so we have to skip createChangeStream tests there.

### DIFF
--- a/source/retryable-writes/tests/etc/templates/handshakeError.yml.template
+++ b/source/retryable-writes/tests/etc/templates/handshakeError.yml.template
@@ -50,7 +50,7 @@ tests:
   #   - Triggers failpoint (second time).
   #   - Tests whether operation successfully retries the handshake and succeeds.
 {% for operation in operations %}
-  - description: "{{operation.operation_name}} succeeds after retryable handshake network error"
+  - description: "{{operation.object}}.{{operation.operation_name}} succeeds after retryable handshake network error"
     operations:
       - name: failPoint
         object: testRunner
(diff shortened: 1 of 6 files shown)


ShaneHarvey:
Tests are passing in Python here: https://github.com/mongodb/mongo-python-driver/pull/1116

source/retryable-writes/tests/etc/templates/handshakeError.yml.template
```diff
@@ -51,6 +51,10 @@ tests:
   #   - Tests whether operation successfully retries the handshake and succeeds.
 {% for operation in operations %}
   - description: "{{operation.operation_name}} succeeds after retryable handshake network error"
+    {%- if operation.operation_name == 'createChangeStream' %}
+    runOnRequirements:
+      - serverless: forbid
+    {%- endif %}
```
jmikola, line 57: Was this just done for consistency? It seems unlikely that `createChangeStream` would ever be tested for retryable writes.

jmikola (approved):
The retryable write template changes may be unnecessary but I defer to you.

eramongodb (approved):

source/retryable-writes/tests/etc/templates/handshakeError.yml.template
eramongodb: > It seems unlikely

Note; this was also why the retryable writes template file does not contain other change stream specific behavior such as `saveResultAsEntity`.

ShaneHarvey: Oh I did this without thinking. Removed.

ShaneHarvey:
Okay I removed the unneeded retryable-writes template changes and fixed the duplciate descriptions problem (by adding operation.object).

eramongodb (approved):
Confirmed the changes pass on the C Driver after updating test descriptions for skipped (unsupported) operations.

PR #1348, opened 2022-11-16
DRIVERS-2490 Only run tests using getnonce on servers < 6.2; correct maxServerVersion for command/reply redaction tests

kmahar:
* The `getnonce` command is being removed in 6.2. This updates various driver tests which used getnonce to only run on lower server versions.
* While looking into this, I noticed that a lot of the other redaction tests used commands which have also been removed in recent server versions. These passed on newer server versions because they expected errors, so the server returned an error regardless: on newer versions, it would just be an error about the command not existing. This coverage didn't seem particularly valuable, and it seemed worthwhile to have each test annotated with the correct maxServerVersion so that as those versions get dropped from drivers we know when the tests can be removed. 

Tests pass in Rust, and since all this changes here is skipping tests in more places I wouldn't expect this to cause failures for anyone.

- [x] Update changelog.

### DIFF
--- a/source/unified-test-format/tests/valid-pass/observeSensitiveCommands.yml
+++ b/source/unified-test-format/tests/valid-pass/observeSensitiveCommands.yml
@@ -38,6 +38,8 @@ createEntities:
 
 tests:
   - description: "getnonce is observed with observeSensitiveCommands=true"
+    runOnRequirements:
+    - maxServerVersion: 6.1.99 # getnonce removed as of 6.2 via SERVER-71007
     operations:
       - name: runCommand
         object: *databaseObserveSensitiveCommands
@@ -57,6 +59,8 @@ tests:
                 nonce: { $$exists: false }
 
   - description: "getnonce is not observed with observeSensitiveCommands=false"
+    runOnRequirements:
+    - maxServerVersion: 6.1.99 # getnonce removed as of 6.2 via SERVER-71007
     operations:
       - name: runCommand
         object: *databaseDoNotObserveSensitiveCommands
@@ -68,6 +72,8 @@ tests:
         events: []
 
   - description: "getnonce is not observed by default"
+    runOnRequirements:
+    - maxServerVersion: 6.1.99 # getnonce removed as of 6.2 via SERVER-71007
     operations:
       - name: runCommand
         object: *databaseDoNotObserveSensitiveCommandsByDefault
--- a/source/unified-test-format/unified-test-format.rst
+++ b/source/unified-test-format/unified-test-format.rst
@@ -3898,8 +3898,6 @@ Changelog
 
 ..
   Please note schema version bumps in changelog entries where applicable.
-:2022-11-16: Update tests for ``observeSensitiveCommands`` that use ``getnonce`` to
-             only run on server versions < 6.2.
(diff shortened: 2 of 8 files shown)


source/unified-test-format/tests/valid-pass/observeSensitiveCommands.yml
```diff
@@ -38,6 +38,8 @@ createEntities:
 
 tests:
   - description: "getnonce is observed with observeSensitiveCommands=true"
+    runOnRequirements:
```
kmahar, line 41: I wasn't sure if it made sense to try to write new tests for `observeSensitiveCommands` that don't use a command which is limited in server version. We also have tests below that cover 4.9+ (hello with speculative authenticate) so we at least cover all server versions with some command. So for now I just skipped these on 6.2+ but let me know if you think we should switch these to some other command.

jmikola: > so we at least cover all server versions with some command.

That seems fine. I don't think we need to replace this for 6.2+.

source/unified-test-format/unified-test-format.rst
```diff
@@ -3898,6 +3898,8 @@ Changelog
 
 ..
   Please note schema version bumps in changelog entries where applicable.
+:2022-11-16: Update tests for ``observeSensitiveCommands`` that use ``getnonce`` to
+             only run on server versions < 6.2.
```
jmikola, line 3902:
```suggestion
```

Tests shouldn't need changelog entries in the main spec. I don't recall this ever being done before, although some test directories do have a README file with a changelog (mostly for prose tests).

jmikola (approved):
LGTM with suggested change.

I noted you have these passing on Rust already, but let me know if you need another language to test. Assuming `MONGODB_LATEST` in [download-mongodb.sh](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/download-mongodb.sh) is 6.2, then I suppose I could trivially test this in a PHPLIB patch build.

source/unified-test-format/unified-test-format.rst
kmahar: sgtm. there was some precedent for describing test changes in the CLAM spec changelog, but happy to remove it from the changelog here

kmahar:
> I noted you have these passing on Rust already, but let me know if you need another language to test. Assuming `MONGODB_LATEST` in [download-mongodb.sh](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/download-mongodb.sh) is 6.2, then I suppose I could trivially test this in a PHPLIB patch build.

You may as well run them if you don't mind doing so! It actually looks like, at least on Ubuntu 18.04, "latest" is already a 6.3 tag (`v6.3.0-alpha-14-gd608948`). But per Sara Golemon on Slack yesterday they are still hoping to get this change into 6.2.

jmikola:
> You may as well run them if you don't mind doing so

FYI: I created a patch build for https://github.com/mongodb/mongo-php-library/pull/1003 and everything passed against `6.3.0-alpha-14-gd608948`.

alcaeus (approved):
LGTM

PR #1349, opened 2022-11-17
DRIVERS-2501 Break NoWritesPerformed-Only Error Sequence

prestonvasquez:
DRIVERS-2501

- [x] Update changelog.

### DIFF
--- a/source/retryable-writes/retryable-writes.rst
+++ b/source/retryable-writes/retryable-writes.rst
@@ -455,8 +455,8 @@ The above rules are implemented in the following pseudo-code:
     while true {
       try {
         return executeCommand(server, retryableCommand);
-      } catch (Exception originalError) {
-        handleError(originalError);
+      } catch (Exception currentError) {
+        handleError(currentError);
 
         /* If the error has a RetryableWriteError label, remember the exception
          * and proceed with retrying the operation.
@@ -464,26 +464,31 @@ The above rules are implemented in the following pseudo-code:
          * IllegalOperation (code 20) with errmsg starting with "Transaction
          * numbers" MUST be re-raised with an actionable error message.
          */
-        if (!originalError.hasErrorLabel("RetryableWriteError")) {
-          if ( originalError.code == 20 && originalError.errmsg.startsWith("Transaction numbers") ) {
-            originalError.errmsg = "This MongoDB deployment does not support retryable...";
+        if (!currentError.hasErrorLabel("RetryableWriteError")) {
+          if ( currentError.code == 20 && originalError.errmsg.startsWith("Transaction numbers") ) {
+            currentError.errmsg = "This MongoDB deployment does not support retryable...";
           }
-          throw originalError
+          throw currentError;
+        }
+
+        /*
+         * If the "previousError" is "null", then the "currentError" is the
+         * first error encountered during the retry attempt cycle. We must
+         * persist the first error in the case where all succeeding errors are
+         * labeled "NoWritesPerformed", which would otherwise raise "null" as
+         * the error.
+         */
+        if (previousError == null) {
+          previousError = currentError;
         }
 
         /*
          * For exceptions that originate from the driver (e.g. no socket available
          * from the connection pool), we should raise the previous error if there
          * was one.
-         *
-         * In addition, if the previousError is null (i.e. this is the first
-         * attempt), we should persist the previousError as the originalError,
-         * in case we need to raise it later (e.g. if all errors are labeled
-         * "NoWritesPerformed").
          */
-        if (originalError is not DriverException && ! originalError.hasErrorLabel("NoWritesPerformed") ||
-                previousError == null) {
-          previousError = originalError;
+        if (currentError is not DriverException && ! originalError.hasErrorLabel("NoWritesPerformed")) {
+          previousError = currentError;
         }
       }
 
@@ -501,7 +506,7 @@ The above rules are implemented in the following pseudo-code:
        * is very rare, and likely means that the cluster is in the midst of a
        * downgrade. */
       if ( ! isRetryableWritesSupported(server)) {
-        throw originalError;
+        throw currentError;
       }
 
       /* CSOT is enabled and the operation has timed out. */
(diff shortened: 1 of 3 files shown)


ShaneHarvey:
@juliusgeo is going to test this in: https://github.com/mongodb/mongo-python-driver/pull/1117

source/retryable-writes/tests/unified/insertOne-noWritesPerformedError.json
```diff
@@ -0,0 +1,98 @@
...
+        "collectionName": "coll"
+      }
+    }
+  ],
+  "tests": [
+    {
+      "description": "InsertOne fails after NoWritesPerformed error",
+      "runOnRequirements": [
+        {
+          "minServerVersion": "4.0",
```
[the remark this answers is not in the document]
prestonvasquez: I've updated the minimum wire version to 6.0, based on the argument in PR #1304 .

source/retryable-writes/retryable-writes.rst
```diff
@@ -474,8 +475,14 @@ The above rules are implemented in the following pseudo-code:
...
          * was one.
+         *
+         * In addition, if the previousError is null (i.e. this is the first
+         * attempt), we should persist the previousError as the originalError,
+         * in case we need to raise it later (e.g. if all errors are labeled
+         * "NoWritesPerformed").
          */
-        if (originalError is not DriverException && ! originalError.hasErrorLabel("NoWritesPerformed")) {
+        if (originalError is not DriverException && ! originalError.hasErrorLabel("NoWritesPerformed") ||
+                previousError == null) {
```
jmikola, line 485: I imagine most languages give `&&` higher precedence than `||` but can we add parentheses here to remove ambiguity?

source/retryable-writes/retryable-writes.rst
```diff
@@ -474,8 +475,14 @@ The above rules are implemented in the following pseudo-code:
...
          * from the connection pool), we should raise the previous error if there
          * was one.
+         *
+         * In addition, if the previousError is null (i.e. this is the first
+         * attempt), we should persist the previousError as the originalError,
+         * in case we need to raise it later (e.g. if all errors are labeled
+         * "NoWritesPerformed").
          */
-        if (originalError is not DriverException && ! originalError.hasErrorLabel("NoWritesPerformed")) {
+        if (originalError is not DriverException && ! originalError.hasErrorLabel("NoWritesPerformed") ||
```
jmikola, line 484: I realize this predates this PR, but I'm finding it quite difficult to make sense of the current pseudo-code here. When I first wrote this for the two-attempt algorithm, the exceptions names were `originalError` and `secondError` (see: [this blame](https://github.com/mongodb/specifications/blame/6f21a767da943664d67bd967cb8ae97a35e40bba/source/retryable-writes/retryable-writes.rst#L450)).

It looks like things started to go downhill with changes for PoolCleared (see: [blame](https://github.com/mongodb/specifications/blame/88c583457c2481871ca6ab22be117a4da9d97135/source/retryable-writes/retryable-writes.rst#L461)), which introduced a syntax error (variable `e` was never defined). That stuck around in the first CSOT change (see: [blame](https://github.com/mongodb/specifications/blame/008d2f5aab41dabced03e0860bab614c0f15ded5/source/retryable-writes/retryable-writes.rst#L456)) until it was eventually removed in [343ff9a](https://github.com/mongodb/specifications/commit/343ff9a3864e7141c5a056fbf19e71ca57e65740#diff-01c94ffec48124f66d321265e57d6c892b1355813cf2bce099d0345ff222eabeR453) last month.

However, the variable name `originalError` stuck around. I think this should be renamed to `currentError` for clarity. That way, `currentError` is always the most recent exception and `previousError` gets set by this conditional.

source/retryable-writes/retryable-writes.rst
```diff
@@ -405,7 +405,8 @@ If a retry attempt also fails, drivers MUST update their topology according to
 the SDAM spec (see: `Error Handling`_). If an error would not allow the caller
 to infer that an attempt was made (e.g. connection pool exception originating
 from the driver) or the error is labeled "NoWritesPerformed", the error from
-the previous attempt should be raised.
+the previous attempt should be raised. If all server errors are labeled
+"NoWritesPerformed", then the first error should be raised.
```
jmikola, line 409: > If all server errors are labeled "NoWritesPerformed", then the first error should be raised.

Is `previousError` in the code below intended to be the _first_ error encountered? If not, and `previousError` could be updated multiple times, the behavior below seems to conflict with this statement as the _first_ error is never permanently saved.

prestonvasquez: Assuming that an error occurs, the idea is to always set "previousError" with the first error encountered, without condition. Otherwise, if the error is NOT labeled "NoWritesPerformed", then update the "previousError" to the "originalError":

```
if (originalError is not DriverException && ! originalError.hasErrorLabel("NoWritesPerformed") ||
        previousError == null) {
    previousError = originalError;
}
```

This behavior will "save" the first error only on the condition that the sequence of errors encountered are all labeled "NoWritesPerformed". Otherwise, a sequence of errors all labeled "NoWritesPerformed" will result in a "null" error.

jmikola (approved):

source/retryable-writes/retryable-writes.rst
```diff
@@ -494,7 +506,7 @@ The above rules are implemented in the following pseudo-code:
        * is very rare, and likely means that the cluster is in the midst of a
        * downgrade. */
       if ( ! isRetryableWritesSupported(server)) {
-        throw originalError;
+        throw currentError;
```
ShaneHarvey, line 509: I beleive this should be previousError too.

ShaneHarvey (approved):

source/retryable-writes/retryable-writes.rst
ShaneHarvey: (adding another comment because I accidentally approved.)

ShaneHarvey (changes requested):

source/retryable-writes/retryable-writes.rst
prestonvasquez: Originally added in [SPEC-972](https://jira.mongodb.org/browse/SPEC-972) , the comment for this block says 

> If the server selected for retrying is too old, throw the original error

What is the purpose for changing the logic here?

ShaneHarvey: Using currentError here is no longer correct when there are multiple retrie. Imagine:
- Attempt 1 fails with NoWritesPerformed
- Attempt 2 fails with ShutdownInProgress
- Attempt 3 fails with NoWritesPerformed
- Attempt 4 selects a server that doesn't support retryable writes

We have to raise the previous non-NoWritesPerformed error which would be the ShutdownInProgress from attempt 2, not the one from attempt 3.

ShaneHarvey (approved):
LGTM!

PR #1356, opened 2022-12-13
DRIVERS-2516 Collection of follow-up logging changes

kmahar:
This is a collection of minor changes that have been discussed on other spec PRs as well as in the drivers-1204-logging Slack channel; see DRIVERS-2516 for details.

### DIFF
--- a/source/logging/logging.rst
+++ b/source/logging/logging.rst
@@ -418,14 +418,15 @@ Invalid Values of Environment Variables
 ---------------------------------------
 For drivers supporting configuration via environment variables, the spec requires that if an environment
 variable is set to an invalid value the driver behaves as if the value were not specified at all, and
-optionally warns the user but does not throw an error.
-We considered the following alternatives:
-1. Drivers could be required to throw an exception if a value is invalid: This was rejected because of concerns
+optionally warns the user but does not throw an error. We considered the following alternatives:
+
+1. Drivers could be required to throw an exception if a value is invalid: This was rejected because of concerns 
    around the implications for environments/applications where multiple versions of the driver or multiple
    drivers may be present and where the validation logic may not match, meaning a value considered valid for
    one driver/version might not be by another. Additionally, there is no obvious place to throw an exception
    from; ``MongoClient`` constructors would be one possibility, but not all languages will support per-client
-  configuration so throwing there might be surprising to users.
+   configuration so throwing there might be surprising to users.
+
 2. Drivers could be required to log a warning if a value is invalid: While drivers MAY do this, requiring
    it was rejected because depending on the language/framework log messages may not be a viable way to
    communicate a warning: if a language's default behavior is to log nothing, or only log messages at a
@@ -436,10 +437,11 @@ We considered the following alternatives:
 Programmatic Configuration Taking Precedence
 --------------------------------------------
 We chose to have programmatic configuration win out over environment variables because:
+
 1. This allows applications built atop drivers (e.g. mongosh) to fully control the driver's logging behavior
-  by setting options for it programmatically.
+   by setting options for it programmatically.
 2. This is consistent with how many drivers treat options specified both in a connection string and programmatically:
-  programmatic options win out.
+   programmatic options win out.
 3. It is straightforward for users to override this behavior (by writing logic to read in environment variables and override
    programmatic defaults), but if we went with the opposite default, it would be more complicated for users to override:
    not all languages will necessarily have an easy way to override/unset an environment variable from within
--- a/source/logging/logging.rst
+++ b/source/logging/logging.rst
@@ -511,5 +511,3 @@ Changelog
              a straightforward way for users to consume the log messages programmatically and
              write to a file themselves.
              Require that programmatic configuration take precedence over environment variables.
-
-             
--- a/source/logging/logging.rst
+++ b/source/logging/logging.rst
@@ -424,10 +424,17 @@ optionally warns the user but does not throw an error. We considered the followi
    around the implications for environments/applications where multiple versions of the driver or multiple
    drivers may be present and where the validation logic may not match, meaning a value considered valid for
    one driver/version might not be by another. Additionally, there is no obvious place to throw an exception
-   from; ``MongoClient`` constructors would be one possibility, but not all languages will support per-client
-   configuration so throwing there might be surprising to users.
+   from about invalid environment variables; ``MongoClient`` constructors would be one possibility, but not all
+   languages will support per-client configuration so throwing there regarding an environment variable might be
+   surprising to users. 
+  
+   Note that these same concerns do not apply to logging options that are specified via driver API: there is no
+   risk of such options propagating to other drivers/driver versions present, and drivers can report exceptions at
+   the point the options are specified, either globally or per-client. Therefore, drivers MUST validate
+   programmatic logging options in a manner consistent with how they validate all other programmatic
+   options, and if possible SHOULD prefer to throw exceptions for invalid configuration.
 
-2. Drivers could be required to log a warning if a value is invalid: While drivers MAY do this, requiring
+1. Drivers could be required to log a warning if a value is invalid: While drivers MAY do this, requiring
    it was rejected because depending on the language/framework log messages may not be a viable way to
    communicate a warning: if a language's default behavior is to log nothing, or only log messages at a
    more severe level than ``warn``, the user will not actually receive the message unless it is logged at a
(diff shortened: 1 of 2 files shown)


kmahar:
I requested the original reviewers of the logging spec as well as everyone that had some part in conversations around these changes, but feel free to remove yourself or only review the parts that you have opinions on.

source/logging/logging.rst
```diff
@@ -392,6 +413,40 @@ Additionally, always including null values would increase the size of log messag
...
+---------------------------------------
+For drivers supporting configuration via environment variables, the spec requires that if an environment
+variable is set to an invalid value the driver behaves as if the value were not specified at all, and
+optionally warns the user but does not throw an error. We considered the following alternatives:
+
+1. Drivers could be required to throw an exception if a value is invalid: This was rejected because of concerns 
+   around the implications for environments/applications where multiple versions of the driver or multiple
+   drivers may be present and where the validation logic may not match, meaning a value considered valid for
+   one driver/version might not be by another. Additionally, there is no obvious place to throw an exception
+   from; ``MongoClient`` constructors would be one possibility, but not all languages will support per-client
```
dariakp, line 427: For the drivers where per-client configuration *is* implemented, is there a downside to supporting client option validation? It feels inconsistent to generally support URI/client option validation, but specifically forbid it in the case of logging. My understanding is that the concerns around validation are specific to handling environment variables.

It may be worth taking a step back to ask why we are allowing configuration via env variables at all. If it's to ease debugging by our team, then it's worth noting that with the client option precedence rule, this benefit may not work if users have already configured their own logging. It is arguable that 3rd party libraries should not be accessing environment variables at all on behalf of the applications in which they are used (I believe @addaleax has some thoughts on this point), and that we could accomplish a similar benefit by utilizing the URI string to specify options.

If we are intent on keeping env variable configuration, then I still think we should decide whether we'd rather align logging client option validation with environment variable parsing or with URI option parsing. In my opinion, the latter would be best from an application developer experience standpoint because helpful errors provide more immediate feedback than silent failures.

addaleax: > this benefit may not work if users have already configured their own logging

I guess one relevant question is how common we'd expect this case to be? For applications, I'd expect them to want to be in control of logging directly, but I could imagine that there are cases in which it's more convenient to use environment variables, e.g. a cloud function that's just a small wrapper around some DB queries?

> I believe @addaleax has some thoughts on this point

(yes, and you very succinctly summarized them :+1:)

jyemin (commented):
Requested a bit more info

source/logging/logging.rst
kmahar: > For the drivers where per-client configuration is implemented, is there a downside to supporting client option validation? It feels inconsistent to generally support URI/client option validation, but specifically forbid it in the case of logging. My understanding is that the concerns around validation are specific to handling environment variables.

Yes, the validation concerns are only around environment variables, and this section is intended to only cover rationale for how we handle those. For non-environment variable configuration, I don't think there is a downside to validation, and I think it's totally fine and I agree a better user experience for drivers to do it in the same way they'd do other client options.

> If it's to ease debugging by our team, then it's worth noting that with the client option precedence rule, this benefit may not work if users have already configured their own logging. 

The idea behind the environment variables in general was that it's a low-friction way to quickly turn on logging without having to actually change your code in any way or redeploy it. You're right that given the precedence rule this might not always work as a debugging tool, but to answer @addaleax's about how common this is, I think the people who mostly need this env var feature and that we designed in mind of are the people who are not using driver logging yet _at all_. Those who have taken the time to actually get it set up _before_ they have a problem are less of a concern here, and the requirement to document how to "have it both ways" (use programmatic config but allow an env var to override as needed) will hopefully encourage users to adopt that as a best practice.
Given that most if not all drivers seem to give precedence to programmatic configuration over URI options, I don't think URI options would help with the precedence problem either, so short of flipping the precedence rules (which there were good arguments against in our original discussion on Slack) I think this may be the best we can do.

More broadly, I recall considering URI options over environment variables early on in this project; a couple of the arguments in favor of environment variables are the following:
1. URI options are per-client, and so this would not work for drivers that only support a single global logger. 
2. As much as possible, connection strings are supposed to be "portable" between drivers. While this isn't 100% the case as there are some options only certain drivers support (e.g., only single-threaded drivers support `serverSelectionTryOnce`) introducing logging options to the connection string might be very confusing in that many drivers that _do_ support logging would not actually be able to support such options themselves - this is going to generally be true for any language that relies on a framework for logging, like Rust, Java, or .NET, where configuration is something that is handled between the user and the framework and the driver stays out of it. E.g. I think it might be confusing that e.g. `commandLogLevel=debug` works in Node, but not in Rust, even though Rust supports command logging and debug level. 

> If we are intent on keeping env variable configuration, then I still think we should decide whether we'd rather align logging client option validation with environment variable parsing or with URI option parsing. In my opinion, the latter would be best from an application developer experience standpoint because helpful errors provide more immediate feedback than silent failures.

I agree, for drivers who support client options around logging it makes sense to validate them. That said, the URI options spec actually requires just logging a warning, not erroring, on invalid URI options, though there has been some disagreement on that subject in the past and if it would be reasonable to allow errors (or even require them): https://jira.mongodb.org/browse/DRIVERS-2056 
Given the issues we've discussed with users even seeing log messages that warn them about logging configuration problems, and that you seem in favor of errors (vs. warnings) in this case - which I am too - what I'd propose instead of aligning with the URI options spec/validation is instead telling drivers to be consistent with how they would handle any other invalid programmatically specified client options. WDYT?

dariakp: @kmahar thanks for the added context - that's interesting regarding the URI options spec, it seems that my impression regarding allowing validation came from the definitions in the auth/tests and the connection-string/tests, both of which have the following guidance:

> Testing whether a URI is valid or not should simply be a matter of checking
whether URI parsing (or MongoClient construction) raises an error or exception.

and the node driver has always validated options, standardizing to throwing where the spec only requires a warning; it also throws on unknown keys, which seems to explicitly go against the spec as written, but also I wasn't able to find any spec tests that assert that invalid keys should be ignored - if that's a desired behavior, it seems like it should be codified in the tests. That said, I think silently ignoring invalid options is a very developer-unfriendly behavior - logging a warning isn't really a substitute for throwing an exception because warnings are easy to miss (and in a legacy environment could be completely drowned out by deprecation/vulnerability warnings).

The argument about logging being unsuitable for URI options in drivers that don't support per-client configuration is valid, and in that case it may be worth explicitly forbidding logging to be configurable as a URI option. 

> Given the issues we've discussed with users even seeing log messages that warn them about logging configuration problems, and that you seem in favor of errors (vs. warnings) in this case - which I am too - what I'd propose instead of aligning with the URI options spec/validation is instead telling drivers to be consistent with how they would handle any other invalid programmatically specified client options.

Yes, I think that would be a good compromise approach. And we can revisit the URI option validation itself in DRIVERS-2056 (I am not convinced that the benefit of being able to copy paste a complete URI string from one driver into another and it "just working" outweighs the risk of the "just working" behavior not actually working as expected due to the other driver ignoring unsupported options, but that's a discussion for another day).

JamesKovacs (approved):
LGTM

nbbeeken (approved):

prestonvasquez (approved):
LGTM

patrickfreed (approved):

baileympearson (approved):

source/logging/logging.rst
kmahar: > That said, I think silently ignoring invalid options is a very developer-unfriendly behavior - logging a warning isn't really a substitute for throwing an exception because warnings are easy to miss (and in a legacy environment could be completely drowned out by deprecation/vulnerability warnings).

> I am not convinced that the benefit of being able to copy paste a complete URI string from one driver into another and it "just working" outweighs the risk of the "just working" behavior not actually working as expected due to the other driver ignoring unsupported options, but that's a discussion for another day

Yeah, I agree, and in Rust in Swift we also throw errors rather than warnings. Given that it sounds like at least 4 drivers including the most-used driver already do what DRIVERS-2056 proposes (Node, Rust, Swift, C) a large portion of users are already not getting that "benefit" and also don't seem to be complaining about it. I left a comment on the ticket linking to this thread.

> in that case it may be worth explicitly forbidding logging to be configurable as a URI option

this SGTM, but I'm not 100% positive that no drivers have a precedent of supporting a bunch of custom URI options that this would clash with, so I just sent out a message in driver-devs to ask and also raise the possibility of more generally discouraging language-specific URI options. 

> Yes, I think that would be a good compromise approach.

Cool, I rephrased and added some language to this section explaining why the "ignore invalid values" approach isn't relevant for programmatic options and saying drivers must validate programmatic logging options in a consistent manner with their other programmatic options.

dariakp: The language change looks good to me

kmahar: @dariakp I think we resolved the one open point left here on slack; any other concerns or anything else you need to approve this?

dariakp: The updated language looks good to me, but there seems to be a merge conflict now

kmahar: ah yeah, in the changelog since Jeremy made a small PR fixing a typo earlier this week. fixed

dariakp (approved):

PR #1360, opened 2022-12-29
Fix typo in trace log level example

jmikola:

- [x] Update changelog.

### DIFF
--- a/source/logging/logging.rst
+++ b/source/logging/logging.rst
@@ -249,7 +249,7 @@ For example:
 
 - If an Informational level is not available and Debug is, messages defined as Informational in a
   specification MUST be emitted at Debug level.
-- If a Trace level is not available, Trace messages MUST be omitted at Debug level.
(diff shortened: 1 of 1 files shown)


source/logging/logging.rst
```diff
@@ -249,7 +249,7 @@ For example:
 
 - If an Informational level is not available and Debug is, messages defined as Informational in a
   specification MUST be emitted at Debug level.
-- If a Trace level is not available, Trace messages MUST be omitted at Debug level.
+- If a Trace level is not available, Trace messages MUST be emitted at Debug level.
```
jmikola, line 252: In my initial read of the spec, I thought trace was intentionally left as optional because it might be too noisy; however, I realized omitting it entirely would conflict with this earlier line:

> If an equivalent level is not available, drivers MUST emit messages for that level at the closest, less severe level if one is available, or the closest more severe level otherwise.

kmahar: nice catch, thank you!

kmahar (approved):

PR #1363, opened 2023-01-11
test(DRIVERS-2523): allow optional bulk write result inserted ids

durran:
The `BulkWriteResult` can contain an optional `insertedIds` property so we check for it not being present or empty.

Tests running in the Node driver suite from PR: https://github.com/mongodb/node-mongodb-native/pull/3516

- [ ] Update changelog.


comandeo (approved):
LGTM!
The new tests pass on Ruby driver

PR #1365, opened 2023-01-13
DRIVERS-2415 Implement OIDC SASL mechanism

blink1073:

- [ ] Update changelog.


source/auth/auth.rst
```diff
@@ -1159,6 +1159,199 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+    MUST be "MONGODB-OIDC"
+
+mechanism_properties
+    PRINCIPAL_NAME
+        Drivers MUST allow the user to specify a principal name, which
+        is required by the server when more than one Identity Provider
+        is configured.
+    DEVICE_NAME
+        Drivers MUST allow the user to specify a name for the device
+        workflow that is one of "aws", "azure", or "gcp".
```
DmitryLukyanov, line 1266: should it be case sensitive?

source/auth/auth.rst
```diff
@@ -1159,6 +1159,199 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+Drives MUST support device workflows for "aws", "azure", and "gcp", given
+by the DEVICE_NAME mechanism property.  In all cases the acquired token
+will be given as the ``jwt`` argument in the second client step of the
+SASL exchange.
+
+AWS
+___
+
+When the DEVICE_NAME mechanism property is set to "aws", the driver MUST
+attempt to read the value given by the ``AWS_WEB_IDENTITY_TOKEN_FILE`` and
```
DmitryLukyanov, line 1325: Can we use the full aws workflow as in aws kms (similar to below approach for gcp and azure)?

source/auth/auth.rst
```diff
@@ -1159,6 +1159,199 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+
+The optional token refresh callback will accept the IDP information as
+well as the cached OIDCRequestTokenResult and return a new OIDCRequestTokenResult.
+
+
+.. code:: typescript
+
+    function onOIDCRefreshToken(serverInfo: OIDCMechanismServerStep1, tokenResult: OIDCRequestTokenResult): OIDCRequestTokenResult
+
+If no callbacks are given, the driver MUST enforce that a DEVICE_NAME
```
DmitryLukyanov, line 1306: What should happen if no callback defined and no DEVICE_NAME? Or If device_name is set but no env varaibles set? I guess it should be exception, if so can we define wording?

source/auth/auth.rst
```diff
@@ -1159,6 +1159,199 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+        is configured.
+    DEVICE_NAME
+        Drivers MUST allow the user to specify a name for the device
+        workflow that is one of "aws", "azure", or "gcp".
+
+
+User Provided Callbacks
+```````````````````````
+
+Drivers MUST allow the user to provide callbacks for token request and
```
DmitryLukyanov, line 1272: what if a callback provided but returns null? I guess it should be exception?

source/auth/auth.rst
DmitryLukyanov: based on .net aws sdk, in this case we will need to look at `AWS_ROLE_ARN` too (otherwise web-identity workflow will fail) as I know we discsussed we don't want doing it, but if we put this aside, will accessToken provided by let's say ec2/ecs envs work here?

source/auth/auth.rst
blink1073: I used lowercase to be consistent with how it is specified in kms

source/auth/auth.rst
blink1073: Updated

source/auth/auth.rst
```diff
@@ -1159,6 +1159,205 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+    MUST NOT be specified.
+
+mechanism
+    MUST be "MONGODB-OIDC"
+
+mechanism_properties
+    PRINCIPAL_NAME
+        Drivers MUST allow the user to specify a principal name, which
+        is required by the server when more than one Identity Provider
+        is configured.
```
addaleax, line 1263: It’s a surprise that we expect `PRINCIPAL_NAME` as an auth mechanism property instead of passing it in the username. Is there a specific reason for this? As I understand it, it fulfills the same role as the username, and in particular also matches the role of the username for similar auth mechanisms like Kerberos or LDAP.

blink1073: I thought it might be misleading to use the `username` field, since the server is actually creating a derived username for this session.

source/auth/auth.rst
```diff
@@ -1159,6 +1159,205 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+    MUST be "$external". Defaults to ``$external``.
+
+password
+    MUST NOT be specified.
+
+mechanism
+    MUST be "MONGODB-OIDC"
+
+mechanism_properties
+    PRINCIPAL_NAME
```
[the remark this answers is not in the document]
blink1073: My understanding is that mechanism_properties should be specifiable in a url string, and can't be non-string objects.

source/auth/auth.rst
```diff
@@ -1159,6 +1159,199 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+  OIDCMechanismClientStep2:
+      description: "Client's request with signed token"
+      strict: false
+      fields:
+          jwt:
+              description: "Compact serialized JWT with signature"
+              cpp_name: JWT
+              type: string
+
+The driver MAY, if it has previously acquired a valid Access Token, or has cached the reply from serverStep1, begin their authentication by sending clientStep2 with their initial saslStart request and authenticate in a single round trip.  Similarly, as an authenticated client's token's expirationTime approaches, it should proactively acquire a new Access Token, and it MAY do so with cached IDP details, skipping directly to clientStep2.
```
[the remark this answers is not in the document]
blink1073: This paragraph comes from the server technical design, we need to determine how this applies to drivers as the "client".    I think we should add a stipulation that if the principal name does not change, we can skip to step 2.  That means we'd have to store the serverStep1 response as well to give to the callbacks.

source/auth/auth.rst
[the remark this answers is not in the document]
DmitryLukyanov: in c# impl mechanism properties are `<string, object>`, so we can put anything here. Yes, it won't work with connection string (for a clear reason), but it will work with ClientSettings.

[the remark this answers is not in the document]
blink1073: Updated

source/auth/auth.rst
addaleax: I guess the question then is about the semantics of the `username` field, whether it is “This is the identity I am attempting to authenticate as” or “This is the username string that the database uses and reports”. I would maybe still lean towards interpreting it as the former, although I see your point about this differing from the DB username.

source/auth/auth.rst
```diff
@@ -1159,6 +1159,218 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+
+
+Caching Credentials
+```````````````````
+
+Drivers MUST enable caching when callback(s) are provided to the mongo client.
+When an authorization request is made and there is a valid cached response,
+the driver MUST use the cached response if it has not expired.
+
+The cached responses are tied to the principal name if given, and an id of the callback function, if possible in the driver's language.
```
addaleax, line 1368: Maybe I’m misunderstanding this, but:

I assume that this cache is implemented on a per-MongoClient basis. Would it be possible to change the callback function after a client has been instantiated? Otherwise I’m not quite sure what the purpose of including the id of the callback is.

blink1073: No, the cache is meant to be global, where the keys are some combination of the principal name and the callback hash.  We could clarify that if a per-client cache is used, then the only key needs to be the principal name.

addaleax: I’d add a clarifying note about that here, yes :slightly_smiling_face:

blink1073: Done!

source/auth/auth.rst
spencerjackson: @sgolemon do you have any feelings on this? I'm vaguely reminded of how during LDAP authentication we map the presented authentication name into a name which the LDAP server will recognize through the UserToDNMapping.

I'm also leaning heavily toward the username in `mongodb://[username:password@]host1` is "This is the identity I am attempting to authenticate as" as Anna said

sgolemon: I do have feelings! My assumption was that the principal name hint would and should be the `username`, and not a custom property as that seems like "least surprise" from an end-developer point of view.  It's true that the derived principal name may differ (depending on IDP and MongoDB configurations), but from the application point of view the synthesized principal name is an implementation detail and all we get by placing the hint into mechanism properties and prohibiting the username field is a little confusion about why it's not in the usual spot.

blink1073: From my understanding of the server technical design and implementation, the principal name is by default the "sub" field of the jwt token, which if the IdP uses pairwise identifier is obfuscated by a [salt](https://openid.net/specs/openid-connect-core-1_0.html#PairwiseAlg).

blink1073: So in that case it feels even less like a user name.

source/auth/auth.rst
```diff
@@ -1159,6 +1159,230 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+When an authorization request is made and there is a valid cached response,
+the driver MUST use the cached response if it has not expired.
+
+The driver MAY, if it has previously acquired a valid Access Token, or has cached the reply from serverStep1, begin their authentication by sending clientStep2 with their initial saslStart request and authenticate in a single round trip.
+
+If the driver implements a global cache, the cache keys MUST include the
+principal name if given, and a hash of the callback function, if possible
+in the driver's language.  A global cache should be preferred, to prevent
+multiple browser interactions in the case of an authentication code
+workflow.
```
addaleax, line 1376: Two things:

- In the devtools products, we won’t really be able to make use of this feature since we’ll need to maintain a cache across MongoClient instances in different processes. I could imagine that other products, if there are any, that use browser flows, would run into similar issues at some point. Consequently, I don’t know if adding global mutable state is worth it here – definitely not a blocker though!
- The cache key should probably include some form of cluster identifier, since the principal name may refer to different things for different clusters. Right now, on our side we’re planning to use the `client_id` provided in the initial SASL response for this.

blink1073: If we rely on a value from server step 1, we cannot skip client step 1 on subsequent calls.

DmitryLukyanov: >The cache key should probably include some form of cluster identifier, since the principal name may refer to different things for different clusters. Right now, on our side we’re planning to use the client_id provided in the initial SASL response for this.

@addaleax , if we need to determine what cluster provided a cached value, is it easier to add a host name to a cache key instead tracking a client_id? It would be good to build a cache key based on initial client configuration and not on something dynamic. Also it will allow skipping step 1.

addaleax: Right, that’s a good point – I don’t mind using the host address as a key here if that’s easy enough. I’m not sure how common it would be to use the same client id for multiple clusters anyway.

blink1073: Updated

source/auth/tests/mongodb-oidc.rst
```diff
@@ -71,6 +72,18 @@ Note: Even thogh only one of the the principals is using the device workflow,
 if the user does not provide a principal name then they cannot be
 distinguished, which is why the second URL should fail.
 
+Valid Configuration
```
DmitryLukyanov, line 75: Invalid configuration?

blink1073: Done

source/auth/auth.rst
```diff
@@ -1159,6 +1159,249 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+DEVICE_NAME is set but credentials cannot be automatically obtained,
+the driver MUST raise an error.
+
+
+Supported Device Workflows
+``````````````````````````
+
+Drives MUST support device workflows for "aws", "azure", and "gcp", given
+by the DEVICE_NAME mechanism property.  In all cases the acquired token
+will be given as the ``jwt`` argument and the second client step of the
```
DmitryLukyanov, line 1338: as we discussed, the second client step in this wording means `saslStart`, not `saslContinue` (which is second step regularly). Can we describe this a bit more clear?

source/auth/auth.rst
```diff
@@ -1159,6 +1159,249 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+
+When the DEVICE_NAME mechanism property is set to "aws", the driver MUST
+attempt to read the value given by the ``AWS_WEB_IDENTITY_TOKEN_FILE`` and
+interpret it as a file path.  The contents of the file are read as the
+access token.
+
+Azure
+_____
+
+When the DEVICE_NAME mechanism property is set to "azure", the driver MUST
```
DmitryLukyanov, line 1354: I think we need to clear cache (inside azure) too in case of failure?

source/auth/auth.rst
blink1073: Done

source/auth/auth.rst
blink1073: Done

source/auth/auth.rst
```diff
@@ -1159,6 +1159,274 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+Before calling a callback, the driver MUST acquire a lock, and verify that
+there are no valid cache credentials before calling the callback.
+This is because request callbacks may involve human interaction, and refresh
+callbacks could use refresh tokens that can only be used once.
+
+The driver MUST provide a way for the callback to be either automatically
+cancelled, or to cancel itself.  This can be as a timeout argument to the
+callback, a cancellation context passed to the callback, or some other
+language-appropriate mechanism.  The timeout MUST respect the client side
+operation timeout if set, or 60 seconds.
```
addaleax, line 1318: Does this mean that there is a hard limit of 60 seconds or CSOT for completing authentication? That seems a bit restrictive for human flows (think 2FA codes taking a while to propagate etc.)

source/auth/auth.rst
```diff
@@ -1159,6 +1159,274 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+The cache is kept alive to preserve the serverStep1 response, as well as
+account for refresh tokens, which typically have an (unknown) lifetime that
+is longer than the access token lifetime.  The refresh token may either be
+stored by the user application, or as part of the request/refresh token
+response in the driver cache. To prevent a memory leak, the driver MUST clear
+invalid cache values at a regular interval, or during every authentication
+attempt.
+
+The cache keys MUST include the principal name (or empty string) and the
+resolved host name for the current server.  Using the resolved host name
```
addaleax, line 1397: This should clarify what “resolved host name” means specifically (or link to some definition if this is a common term in driver specs) … the text below makes it sounds like it’s the hostname of the primary for replica sets?

source/auth/auth.rst
blink1073: I agree.  The user might also have to manually retrieve a password from a vault, etc.  The question is then whether we want to impose any time limit on this.  Perhaps 5 minutes?

source/auth/auth.rst
blink1073: Updated to "DNS resolved name".

addaleax: It’s honestly still not really clear what this means… Does this mean the server’s IP address? Is the cache for a single endpoint, or for an entire replicaset? Does “DNS resolved” mean CNAME resolution? Does it include port information? etc.

source/auth/auth.rst
addaleax: I’d expect any time limit to be configured by the callback itself. 5 minutes seems like it should be enough for human flows and I’d definitely be okay with using that here, I’m just wondering if the driver is really the right place to enforce a limit.

DmitryLukyanov: I would say differently. On the driver side, are we ok with possibly infinite waiting under lock? Since user interaction can potentially take any time. I would say not relly

source/auth/tests/legacy/connection-string.json
```diff
@@ -80,7 +80,7 @@
     },
     {
       "description": "should accept generic mechanism property (GSSAPI)",
-      "uri": "mongodb://user%40DOMAIN.COM@localhost/?authMechanism=GSSAPI&authMechanismProperties=SERVICE_NAME:other,CANONICALIZE_HOST_NAME:forward,SERVICE_HOST:example.com",
+      "uri": "mongodb://user%40DOMAIN.COM@localhost/?authMechanism=GSSAPI&authMechanismProperties=SERVICE_NAME:other,CANONICALIZE_HOST_NAME:true",
```
DmitryLukyanov, line 83: is this correct change?

source/auth/tests/legacy/connection-string.json
```diff
@@ -89,46 +89,10 @@
...
         "mechanism_properties": {
           "SERVICE_NAME": "other",
-          "SERVICE_HOST": "example.com",
-          "CANONICALIZE_HOST_NAME": "forward"
+          "CANONICALIZE_HOST_NAME": true
         }
       }
     },
-    {
-      "description": "should accept forwardAndReverse hostname canonicalization (GSSAPI)",
```
DmitryLukyanov, line 98: why are these tests removed?

source/auth/tests/legacy/connection-string.json
durran: This is basically reverting DRIVERS-1803 which only the Node driver implemented because it was needed for mongosh. I don't know if any of the other drivers will actually implement that so not sure what to do here.

source/auth/tests/legacy/connection-string.json
durran: Same as other comment re DRIVERS-1803

source/auth/tests/legacy/connection-string.json
DmitryLukyanov: This change is unrelated to oidc, I think we can leave it as it is in the master

source/auth/tests/legacy/connection-string.json
blink1073: Ah, I see the problem.  I copied the file from pymongo where I had added the new tests, but we were not up to date with the spec.  I will revert these changes.

source/auth/tests/legacy/connection-string.json
blink1073: Fixed

source/auth/tests/legacy/connection-string.json
blink1073: Fixed

source/auth/auth.rst
blink1073: I agree, even if the driver is multi-threaded, we still do not want to have an orphaned thread.  I've updated this section.

source/auth/auth.rst
```diff
@@ -1159,6 +1170,259 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+attempt.
+
+The cache keys MUST include the principal name (or empty string) and the
+IP address and port for the current server's endpoint.  Using the IP address and port accounts for the case when two different servers use the same
+principal name but could be configurated differently.  There is an edge case
+where if the same principal name is used and two aliases to the same local
+host address are given, there will be duplicate user/device interactions,
+unless the driver can resolve the local host address as well.  Note that
+because we use the current host name, changing the primary host on a replica
+set will result in a cache miss.
```
addaleax, line 1400: Does this still apply? If the cache is based on the server IP address, then it will have different entries for all RS members.

blink1073: I'll rephrase it, that is what I was trying to convey.

source/auth/tests/legacy/connection-string.json
durran: Thanks!

source/auth/tests/legacy/connection-string.json
durran: Thanks!

source/auth/auth.rst
```diff
@@ -254,6 +254,17 @@ or "SCRAM-SHA-256").
 The cache entry value MUST be either the ``saltedPassword`` parameter or the
 combination of the ``clientKey`` and ``serverKey`` parameters.
 
+Reauthentication
+~~~~~~~~~~~~~~~~
+
+On any operation that requires authentication, the server may raise the
+error ``ReauthenticationRequired`` (391) if the user's credential has expired.
+Drivers MUST immediately attempt to force a reauthenication on the connection
```
durran, line 262: Should we also be more specific here on what reauthentication exactly entails? Is this the step1/step2 flow (assuming we remove the cached credentials potentially) or is it the full handshake steps with speculative auth? And will we have more prose tests added to test this? The current unified tests are only testing retryability but don't have anything to do with auth specifically.

durran (changes requested):

source/auth/auth.rst
blink1073: I updated the language.  I'm not sure of a way to test it that will not result in a race condition, because we need the initial token to succeed, but be "about to expire".

durran: Thanks for the language update and understand the test issue. I'll mark this comment resolved - great work so far!

source/auth/tests/mongodb-oidc.rst
```diff
@@ -0,0 +1,130 @@
...
+#. Ensure that the cache has been cleared.
+
+#. Clear the cache.
+#. Create a new client using the AWS device workflow.
+#. Ensure that a ``find`` operation does not add credentials to the cache.
+
+Reauthentication
+================
+
+If it is possible to force a reauthenictation on a connection, the driver
```
addaleax, line 119:
```suggestion
If it is possible to force a reauthentication on a connection, the driver
```

source/auth/auth.rst
addaleax: @blink1073 On today's initiative sync we briefly touched on this, and the sentiment was to use the username field like we would for LDAP and Kerberos. Unless there are concerns with it that haven't been brought up so far, this shouldn't be an auth mechanism property.

blink1073: Okay fair, I will update the spec and connection string tests.

blink1073: Done

source/auth/auth.rst
```diff
@@ -1159,6 +1175,264 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
 .. note::
     Five minutes was chosen based on the AWS documentation for `IAM roles for EC2 <https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html>`_ : "We make new credentials available at least five minutes before the expiration of the old credentials". The intent is to have some buffer between when the driver fetches the credentials and when the server verifies them.
 
+MONGODB-OIDC
+~~~~~~~~~~~~
+
+:since: 7.0 Enterprise
+
+MONGODB-OIDC authenticates using an `OIDC <https://openid.net/specs/openid-connect-core-1_0.html>`_ access tokens.  Drivers MUST support
+both the Authentication Code Flow and Device Workflow for AWS, Azure,
```
addaleax, line 1184: This sounds a bit like drivers would need to implement Authentication Code Flow out of the box, when in reality that’s left to user-provided callbacks (which may be populated with example code for ACF) – maybe this could be reworded to clarify this?

source/auth/auth.rst
```diff
@@ -1159,6 +1175,264 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+attempt.
+
+The cache keys MUST include the username (or empty string) and the
+actually used socket address and port for the current server.
+Using the socket address and port accounts for the case when two different
+servers use the same username but could be configurated differently.
+There is an edge case where if the same username is used and two aliases
+to the same local host address are given, there will be duplicate user/device
+interactions, unless the driver can resolve the local host address as well.
+Note thatbecause we use the server socket address, there will different cache
```
addaleax, line 1403:
```suggestion
Note that because we use the server socket address, there will different cache
```

source/auth/auth.rst
```diff
@@ -1159,6 +1175,264 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+The driver MUST cache the serverStep1 reponse as part of the cache value,
+to enable skipping serverStep1 on subsequent authentications of the same
+cache key.
+
+A global cache should be preferred, to prevent multiple browser interactions
+in the case of an authentication code workflow.  However, drivers or dev tools
+can choose to use their own caching scheme if appropriate for their language/
+environment.
+
+A cached value will expire within 5 minutes of the ``expiresInSeconds`` time.
```
addaleax, line 1415: I assume “within 5 minutes” means “5 minutes _before_ `expiresInSeconds` will be reached”, not “after”, but it might be good to make this wording unambiguous?

What happens if the callback returns an `expiresInSeconds` value that is less than 5 minutes? (If that means not caching the result, then that would be very nice for us in devtools as a way of disabling the cache, but I’m not sure if that’s really what would be most desirable in the big picture.)

addaleax (approved):

source/auth/auth.rst
blink1073: Good catch, updated

source/auth/auth.rst
blink1073: I clarified the paragraph.

addaleax (approved):

durran (approved):

source/auth/auth.rst
```diff
@@ -1159,6 +1175,266 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+              type: int
+              optional: true
+          refreshToken:
+              description: "The OIDC refresh token"
+              type: str
+              optional: true
+
+.. code:: typescript
+
+    function onOIDCRequestToken(serverInfo: OIDCMechanismServerStep1, timeoutSeconds: int): OIDCRequestTokenResult
```
addaleax, line 1316: Would it be possible to add the principal name as an argument to these (or at least the request callback here)? I think that’s actually a requirement on our side to understand whether an OIDC auth attempt for different `MongoClient`s happens for the same user or not.

blink1073: Done

addaleax (approved):
Just re-affirming my LGTM

DmitryLukyanov (approved):
LGTM

source/auth/tests/mongodb-oidc.rst
```diff
@@ -0,0 +1,167 @@
...
+AWS Device Auth
+===============
+
+Drivers MUST be able to authenticate using the "aws" device workflow simulating
+an EC2 instance with an enabled web identity token provider, generated by
+Drivers Evergreen Tools.
+
+#. Set the ``AWS_WEB_IDENTITY_TOKEN_FILE`` environment variable to the location
+of the ``test_user1`` generated token file.
+#. Create a client with the url parameters ``?authMechanism=MONGODB-OIDC&authMechanismProperties=DEVICE_NAME=aws``.
```
durran, line 46:
```suggestion
#. Create a client with the url parameters ``?authMechanism=MONGODB-OIDC&authMechanismProperties=DEVICE_NAME:aws``.
```

durran (changes requested):

durran (approved):

PR #1366, opened 2023-01-13
DRIVERS-2525 Add a command logging test covering unacknowledged write behavior

kmahar:
Adds a command logging test covering unacknowledged write behavior. Drivers are supposed to emit a "command succeeded" message with a reply of ok: 1 for unacknowledged writes but we were missing test coverage.

Rust doesn't have unacknowledged writes but @jyemin has confirmed this test passes for the Java driver.


jyemin (approved):
LGTM.  Thanks for the quick turnaround on this.

BorisDog (approved):
LGTM, works in .NET.

PR #1367, opened 2023-01-17
test(DRIVERS-2468): add unified test for ok 1

durran:
Adds a unified test in command monitoring tests that ensures a command succeeded event is emitted when a reply contains write concern errors and `{ ok: 1 }`.

Node implementation passing here: https://github.com/mongodb/node-mongodb-native/pull/3525

- [ ] Update changelog.


source/command-logging-and-monitoring/tests/monitoring/writeConcernError.yml
```diff
@@ -0,0 +1,80 @@
...
+          client: *client
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+              failCommands: [ insert ]
+              writeConcernError:
+                code: 91 # ShutdownInProgress
+                errorLabels: [RetryableWriteError]
+      - name: insertOne
```
kmahar, line 47: is the extra operation and started/succeeded event necessary?

kmahar (commented):
the test looks ok from a quick inspection. let me see tomorrow if I can quickly fix the bug in Rust and see if the test passes there too. I'd like to add an equivalent command logging test

source/command-logging-and-monitoring/tests/monitoring/writeConcernError.yml
durran: I wanted to do something similar to the [retryable writes prose test #3](https://github.com/mongodb/specifications/tree/master/source/retryable-writes/tests#prose-tests) where the first write concern error causes a retry. So in this case we get the extra started/succeeded event because it does retry the first insert.

source/command-logging-and-monitoring/tests/monitoring/writeConcernError.yml
```diff
@@ -0,0 +1,80 @@
...
+        events:
+          - commandStartedEvent:
+              command:
+                insert: *collectionName
+                documents:
+                  - { _id: 2, x: 22 }
+                ordered: true
+              commandName: insert
+              databaseName: *databaseName
+          - commandSucceededEvent:
```
kmahar, line 62: the test passes in Rust; for some reason I thought we had the bug mentioned in the drivers ticket but I guess not.

that said, I did have to update some Rust logic that expected a write concern error to always have an errmsg to get this test to pass. I checked with a friend on server and she said errmsg is technically an optional field so I filed a ticket about that for Rust: https://jira.mongodb.org/browse/RUST-1574

kmahar (approved):

PR #1369, opened 2023-01-20
DRIVERS-1003 Introduce CMAP error tests

patrickfreed:
DRIVERS-1003

This PR introduces test coverage for some basic CMAP error cases. The changes involve a few new CMAP format tests as well as a few new unified format tests. The CMAP format's version of runOnRequirements was updated to support the `auth` requirement from the Unified Test Format. Also, the existing logging tests were moved into a single `unified` test directory along with the new tests.

This PR also addresses DRIVERS-1784 and DRIVERS-1785, both of which were closely related.


source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -519,10 +527,6 @@ Populating the pool MUST NOT block any application threads. For example, it
 could be performed on a background thread or via the use of non-blocking/async
 I/O. Populating the pool MUST NOT be performed unless the pool is "ready".
 
-If an error is encountered while populating a connection, it MUST be handled via
```
patrickfreed, line 522: This notice and associated psuedocode was moved to the "Establishing a Connection" section above to avoid duplication with "Checking out a Connection". The requirement still remains.

source/connection-monitoring-and-pooling/tests/README.rst
```diff
@@ -26,11 +26,4 @@ The following tests have not yet been automated, but MUST still be tested:
...
 #. A user MUST be able to specify all ConnectionPoolOptions via a URI string
 #. A user MUST be able to subscribe to Connection Monitoring Events in a manner idiomatic to their language and driver
-#. When a check out attempt fails because connection set up throws an error,
-   assert that a ConnectionCheckOutFailedEvent with reason="connectionError" is emitted.
 
-Logging Tests
-=============
-
-Tests for connection pool logging can be found in the `/logging <./logging>`__ subdirectory and are written in the 
-`Unified Test Format <../../unified-test-format/unified-test-format.rst>`__.
```
ShaneHarvey, line 36: Was this intentionally removed?

ShaneHarvey (commented):
Great stuff! I will give this a more thorough review including POCing the new tests asap.

source/connection-monitoring-and-pooling/tests/README.rst
patrickfreed: Yep, the logging and new unified tests were combined into a single `unified` directory, rather than having a separate one for logging.

source/connection-monitoring-and-pooling/tests/unified/network-error.json
```diff
@@ -0,0 +1,464 @@
...
+                  "hello"
+                ],
+                "closeConnection": true,
+                "appName": "cmapNetworkErrorCheckOutTest"
+              }
+            }
+          }
+        },
+        {
+          "name": "find",
```
BorisDog, line 255: I think there might be a race here:
Pool is paused before Find happens, and then we don't get `connectionCreated`/`connectionClosed `event.

source/connection-monitoring-and-pooling/tests/unified/network-error.json
```diff
@@ -0,0 +1,464 @@
...
+                "client": {
+                  "id": "client",
+                  "useMultipleMongoses": false,
+                  "observeEvents": [
+                    "connectionCheckOutStartedEvent",
+                    "connectionCheckedOutEvent",
+                    "connectionCheckOutFailedEvent",
+                    "connectionCheckedInEvent",
+                    "connectionClosedEvent",
+                    "poolClearedEvent",
```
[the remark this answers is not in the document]
patrickfreed: good catch, fixed

source/connection-monitoring-and-pooling/tests/unified/network-error.json
patrickfreed: good catch, i added a really high heartbeatFrequency to this test so that the failpoint wont be triggered by the monitors.

source/connection-monitoring-and-pooling/tests/unified/network-error.json
```diff
@@ -0,0 +1,464 @@
...
+            },
+            "count": 1
+          }
+        }
+      ],
+      "expectEvents": [
+        {
+          "client": "client",
+          "eventType": "cmap",
+          "events": [
```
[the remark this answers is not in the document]
patrickfreed: done

source/connection-monitoring-and-pooling/tests/cmap-format/pool-checkout-error-auth.json
```diff
@@ -0,0 +1,69 @@
...
+    {
+      "type": "ConnectionClosed",
+      "connectionId": 1,
+      "address": 42,
+      "reason": "error"
+    },
+    {
+      "type": "ConnectionCheckOutFailed",
+      "connectionId": 1,
+      "address": 42,
```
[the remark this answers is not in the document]
patrickfreed: fixed

source/connection-monitoring-and-pooling/tests/cmap-format/pool-checkout-error-concurrent.yml
```diff
@@ -0,0 +1,47 @@
...
+  - name: checkOut
+    thread: thread2
+  - name: checkOut
+    thread: thread3
+  - name: waitForEvent
+    event: ConnectionCheckOutFailed
+    count: 3
+  - name: ready
+events:
+  - type: ConnectionPoolReady
```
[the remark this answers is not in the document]
patrickfreed: The cmap-format tests are unit tests of an individual pool, so the events in each one should only be associated with that pool. If a driver is testing this via an actual client which makes pools per server, I think it would probably be best if the test runner for that driver just filtered out any events observed from other pools.

BorisDog (dismissed):
LGTM! Tests work in .NET.

source/connection-monitoring-and-pooling/tests/cmap-format/pool-checkout-error-network.yml
```diff
@@ -0,0 +1,39 @@
...
+  - type: ConnectionClosed
+    address: 42
+    connectionId: 1
+    reason: error
+  - type: ConnectionCheckOutFailed
+    address: 42
+    reason: connectionError
+ignore:
+  - ConnectionPoolCreated
+  - ConnectionPoolReady
```
ShaneHarvey, line 39: This test fails in Python. For example the python cmap-format test runner mocks various pieces of the MongoClient and Topology to get a connection Pool suitable for the tests, but this test expects to observe the effects of the Topology's error handling rules. I think this kind of test is outside the scope of the cmap test format and has to be implemented in the unified runner. Any test for the interaction between SDAM and CMAP needs to be moved over to the unified runner IMO. Does that make sense?

ShaneHarvey (changes requested):

source/connection-monitoring-and-pooling/tests/cmap-format/pool-checkout-error-network.yml
ShaneHarvey: @BorisDog, how does this test pass in .NET?

BorisDog: This is an [integration tests format](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/tests/cmap-format/README.rst#integration-test-format), so it runs against real deployment, with real SDAM in .NET.

ShaneHarvey: I'm not sure using SDAM here is expected. If SDAM is running then the heartbeats will fail and clear the pool at any time, thus the events would be non-deterministic.

BorisDog: Yeah we hack it a bit: we set 10 mins heartbeat interval, and then force the pool to mark itself as ready.

ShaneHarvey: I suppose I can implement more hacks to get this to "work" but at that point it wouldn't even test the same code path that the driver takes for handling an error while checking out a connection for an operation. Is there any reason _not_ to move these tests over to the unified format?

BorisDog: In C# it's just the initial setup that is adjusted, so the rest of SDAM works as usual.

But I agree that if unified format fits better, than it's preferable.
BTW this PR already contains an unified test similar to this one: ["Pool properly handles network error during checkout"](https://github.com/mongodb/specifications/pull/1369/files/bec7969f068b7e1778bd6929d4ae9a7157e2b749#diff-f83bf8d2e7fc7772a5452ed362a7ddc1d4a888036ee8e527bd49b92a1a322104R100), not sure if they test the exact same thing...

ShaneHarvey: Good point. Yes it looks like that test is exactly what I'm asking for. Let's remove this test. Let me check which other cmap-format tests are problematic.

patrickfreed: Sorry for letting this one simmer for a while--I've gone ahead and updated all the newly introduced CMAP-format tests to the unified format, with the exception of "checking in a connection after checkout fails closes connection" as mentioned above.

PR #1372, opened 2023-01-26
relax logging to allow expected to subset actual

prestonvasquez:
When validating the logs, it could be the case that there are more "actual" logs being queued than "expected" logs. For example, the Go Driver closes sessions when the client is disconnected which triggers three extra checkout logs than are expected in the CMAP UST. In this case, the unified test will result in |actual| = |expected| + 3 and fail.

It seems realistic to assume that most drivers will have "noisy"/"extra" logs while running unified spec tests. To be more precise, a valid case could meet the condition such that  |actual| >= |expected| > 0. Perhaps we should relax the conditions on what it means for "expected" to match "actual". Instead of an exact match, we could require that expected messages be an "ordered subset" of the actual messages.

- [x] Update changelog.
- [ ] Make sure there are generated JSON files from the YAML test files. NA
- [x] Test changes in at least one language driver. Go

### DIFF
--- a/source/unified-test-format/unified-test-format.rst
+++ b/source/unified-test-format/unified-test-format.rst
@@ -1394,9 +1394,8 @@ The structure of each object is as follows:
   messages, which are expected to be observed (in this order) on the corresponding
   client while executing `operations`_. If the array is empty, the test runner
   MUST assert that no messages were observed on the client. The driver MUST assert
-  that the messages produced are an exact match, i.e. that the expected and actual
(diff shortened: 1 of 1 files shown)


prestonvasquez:
The team has decided that it would be best to keep strict expectations for what is logged. The "observeLogMessages"  on a client will probably ensure unification across all drivers. https://github.com/mongodb/specifications/pull/1374 is the follow-up PR to include the "end session" logs created by closing the client as part of the test runner's operations. This will resolve the issue that e to this PR.

PR #1374, opened 2023-01-30
DRIVERS-1675 Add the end session log expectations for CMAP

prestonvasquez:
DRIVERS-1675

Drivers need to account for the logs created by closing the client as an operation, including those produced by ending a session.

- [ ] Update changelog. NA
- [x] Test changes in at least one language driver. Go

### DIFF
--- a/source/connection-monitoring-and-pooling/tests/logging/connection-logging.yml
+++ b/source/connection-monitoring-and-pooling/tests/logging/connection-logging.yml
@@ -66,7 +66,31 @@ tests:
               driverConnectionId: { $$type: [int, long] }
               serverHost: { $$type: string }
               serverPort: { $$type: [int, long] }
-          
+
+          - level: debug
+            component: connection
+            data:
+              message: "Connection checked out"
+              driverConnectionId: { $$type: [int, long] }
+              serverHost: { $$type: string }
+              serverPort: { $$type: [int, long] }
+
+          - level: debug
+            component: connection
+            data:
+              message: "Connection checked in"
+              driverConnectionId: { $$type: [int, long] }
+              serverHost: { $$type: string }
+              serverPort: { $$type: [int, long] }
+
+          #  The next three expected logs are for ending a session.
+          - level: debug
+            component: connection
+            data:
+              message: "Connection checkout started"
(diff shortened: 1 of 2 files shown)


kmahar:
Thanks @prestonvasquez! This looks good to me based on what we discussed, but as I mentioned I can't prove that it works in Rust since we have a bug around not sending `endSession` which is non-trivial to fix and I won't be able to get to before my transfer. @BorisDog and I talked about him fixing the .NET bug and then using that to prove your change here works for another driver, so I'll defer to his review.

kmahar:
BTW, I think it's ok to not file a separate drivers ticket for this since only Rust, Go, and .NET have gotten to it yet, but once this gets merged please update the downstream changes section on DRIVERS-1675 to say there was a follow up commit and link to the commit, since I had linked to my original commit there that didn't have this and people might sync to the old commit and miss this change otherwise.

source/connection-monitoring-and-pooling/tests/logging/connection-logging.yml
```diff
@@ -192,5 +216,3 @@ tests:
               message: "Connection checkout failed"
               serverHost: { $$type: string }
               serverPort: { $$type: [int, long] }
-              reason: "An error occurred while trying to establish a new connection"
-              error: { $$exists: true }
```
BorisDog, line 196: +1

kmahar: Somehow I missed this change when I originally reviewed the PR. What was the reasoning?

kmahar: @BorisDog did you have some issue with this?

BorisDog: No issues.
I thought the reason was missing, and was added now.

BorisDog (approved):
LGTM.
After `endSessions` in .NET the tests pass as expected.

kmahar (approved):
lgtm. since @BorisDog also owns CMAP and @patrickfreed is now on loan to service arch for the quarter I think it's ok to merge without waiting for Patrick's review.

PR #1376, opened 2023-01-31
DRIVERS-2312: do not consider AutoEncryptionOptions in createEncryptedCollection

baileympearson:
Given that:

- createEncryptedCollection does not mutate the auto encryption options on the MongoClient if it finds them there
- libmongocrypt throws if the encryptedFieldsMap is incomplete

The following code always throws:

```typescript
const client = new MongoClient(uri, { encryptedFieldsMap })
const ce = new ClientEncryption(...);
await cd.createEncryptedCollection(client.db('foo'), <options>))

// assume createEncryptedCollection succeeds

await client.connect(); // libmongocrypt throws because `encryptedFieldsMap` is not complete, even though the collection has been created
```

This PR updates the helper to never look for the `encryptedFieldsMap` on AutoEncryptionOptions and ensures that drivers document this potentially confusing behavior.

Additional context available in slack [here](https://mongodb.slack.com/archives/CFPHLTRMK/p1674762265000629).

### DIFF
--- a/source/client-side-encryption/client-side-encryption.rst
+++ b/source/client-side-encryption/client-side-encryption.rst
@@ -4,7 +4,7 @@ Client Side Encryption
 
 :Status: Accepted
 :Minimum Server Version: 4.2 (CSFLE), 6.0 (Queryable Encryption)
-:Last Modified: 2023-01-30
+:Last Modified: 2023-01-31
 :Version: 1.12.0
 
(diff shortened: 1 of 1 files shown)


kevinAlbs (approved):
LGTM. Thank you for the analysis and fix.

Update the `Changelog` and `Last Modified` before merging.

PR #1377, opened 2023-02-01
DRIVERS-2539 Replace `DataKeyOpts` with `masterKey` in `CreateEncryptedCollection`

kevinAlbs:
# Summary

- Replace `DataKeyOpts` with `masterKey` in `CreateEncryptedCollection`
- Run test cases for `CreateEncryptedCollection` with `aws` KMS provider.

Updated API is implemented and tested in [the Go driver here](https://github.com/mongodb/mongo-go-driver/pull/1169).

# Background & Motivation

Running test cases with the ``aws`` KMS provider is intended to provide test coverage for handling ``masterKey`` option. The ``local`` KMS provider not requiring a ``masterKey`` option.

The motivation for removing `keyAltNames` from `CreateEncryptedCollection` is noted in DRIVERS-2539:
> The problem is that if keyAltNames is given and createEncryptedCollection creates >1 key, then the method will always fail because of a duplicate key error.
> MONGOCRYPT-432 would allow createEncryptedCollection to create keys with different keyAltNames through encryptedFieldsMap/encryptedFields.

The motivation for removing `keyMaterial` from `CreateEncryptedCollection` is due to having no known use-case.
As previously specified, calling `CreateEncryptedCollection` with a custom keyMaterial creates all data keys with the same custom keyMaterial.
Users can still create data keys with custom keyMaterial through the CreateDataKey API.
The rationale for adding custom keyMaterial to CreateDataKey is [documented in the Key Management API scope](https://docs.google.com/document/d/1NXPZwDY62bI81RmORnGpdXQRthzAb5_GwZrP10nDDRw/edit#heading=h.lfd5yrq9s8bm).

# Caveats

C# has released the CreateEncryptedCollection API. Removing these options is a backwards breaking change. For C#, the best option may be to validate and raise an error if keyMaterial or keyAltName is set in CreateEncryptedCollection.

- [x] Update changelog.
- [x] Test changes in at least one language driver. **Tested in Go**
- [ ] Test these changes against all server versions and topologies (including standalone, replica set, sharded clusters, and serverless). (Pending)

### DIFF
--- a/source/client-side-encryption/tests/README.rst
+++ b/source/client-side-encryption/tests/README.rst
@@ -2509,6 +2509,17 @@ options::
       },
    }
 
+Run each test case with each of these KMS providers: ``aws``, ``local``. The KMS provider name is referred to as ``kmsProvider``.
+When testing ``aws``, use the following as the ``masterKey`` option:
+
+.. code:: javascript
+
+   {
+      region: "us-east-1",
+      key: "arn:aws:kms:us-east-1:579766882180:key/89fcc2c4-08b0-4bd9-9f25-e30687b580d0"
+   }
+
+When testing ``local``, set ``masterKey`` to ``null``.
 
 Case 1: Simple Creation and Validation
 ``````````````````````````````````````
@@ -2535,7 +2546,7 @@ rejects an attempt to insert plaintext in an encrypted fields.
          }
       }
 
-2. Invoke `CreateEncryptedCollection(CE, DB, "testing1", Opts, "local", null)`
+2. Invoke `CreateEncryptedCollection(CE, DB, "testing1", Opts, kmsProvider, masterKey)`
    to obtain a new collection `Coll`. Expect success.
 3. Attempt to insert the following document into `Coll`::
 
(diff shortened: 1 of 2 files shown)


vector-of-bool (approved):

DmitryLukyanov (approved):
LGTM

baileympearson (approved):
tests passing in Node in https://github.com/mongodb/node-mongodb-native/pull/3544/files

source/client-side-encryption/tests/README.rst
```diff
@@ -2509,6 +2509,17 @@ options::
       },
```
baileympearson, line 2509: should the `clientEncryptionOptions` also include the `aws` kms provider?

baileympearson (changes requested):

source/client-side-encryption/tests/README.rst
kevinAlbs: Yes. Good eye. Updated.

baileympearson (approved):

ShaneHarvey (approved):

PR #1378, opened 2023-02-15
DRIVERS-2333 add steps to credential caching tests

kevinAlbs:
# Summary
- Create a new client before a `find` operation. 
- Clear the AWS environment variables after the test.
- Clear the cache before steps expecting to cache credentials.

# Background & Motivation
Create a new client before a `find` operation to ensure authentication is attempted.
Clear the AWS environment variables after the test to prevent the environment variable interacting with  other tests.

~~- [ ] Update changelog.~~
- [x] Test changes in at least one language driver. **Updated steps are tested in [C](https://github.com/mongodb/mongo-c-driver/pull/1207/files#diff-6b8b6c50d722af8c2029c7c665ebb7f8db62b13816f8b9bd65c8e077d367303aR278)**
~~- [ ] Test these changes against all server versions and topologies (including standalone, replica set, sharded clusters, and serverless).~~

### DIFF
--- a/source/auth/auth.rst
+++ b/source/auth/auth.rst
@@ -991,7 +991,7 @@ case where another part the application has refreshed the credentials.
 
 However, if environment variables are not present during initial authorization,
 credentials may be fetched from another source and cached.  Even if the
-environmnet variables are present in subsequent authorization attempts,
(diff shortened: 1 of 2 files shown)


blink1073 (approved):
Thanks!

PR #1379, opened 2023-02-24
DRIVERS-2530: Defer checking for session support until after connection checkout

dariakp:

- [x] Update changelog.
- [N/A] Make sure there are generated JSON files from the YAML test files.

### DIFF
--- a/source/sessions/driver-sessions.rst
+++ b/source/sessions/driver-sessions.rst
@@ -1152,4 +1152,4 @@ Changelog
 :2022-03-24: ServerSession Pooling is required and clarifies session acquisition bounding
 :2022-06-13: Move prose tests to test README and apply new ordering
 :2022-10-05: Remove spec front matter
-:2022-02-24: Defer checking for session support until after connection checkout
+:2023-02-24: Defer checking for session support until after connection checkout
--- a/source/sessions/driver-sessions.rst
+++ b/source/sessions/driver-sessions.rst
@@ -205,7 +205,7 @@ its lifetime.
 
 Drivers MUST NOT check for session support in `startSession`. Instead, if sessions 
 are not supported, the error MUST be reported the first time the session is used
-for an operation (See `How to Check Whether a Connection Supports Sessions`_).
+for an operation (See `How to Tell Whether a Connection Supports Sessions`_).
 
 Explicit vs implicit sessions
 -----------------------------
@@ -437,8 +437,8 @@ operation and ended immediately after this operation completes. The actual
 implementation will likely involve calling ``client.startSession``, but that is not
 required by this spec. Regardless, please consult the startSession section to
 replicate the required steps for creating a session.
-The driver MUST check for session support, but only after the connection is checked out
-(See `How to Check Whether a Connection Supports Sessions`_) and MUST NOT consume a server  
+The driver MUST NOT use the session if the checked out connection does not support sessions
+(see `How to Tell Whether a Connection Supports Sessions`_) and, in all cases, MUST NOT consume a server
 session id until after the connection is checked out and session support is confirmed.
 
 MongoCollection changes
@@ -479,8 +479,8 @@ operation and ended immediately after this operation completes. The actual
 implementation will likely involve calling ``client.startSession``, but that is not
 required by this spec. Regardless, please consult the startSession section to
 replicate the required steps for creating a session.
-The driver MUST check for session support, but only after the connection is checked out
-(See `How to Check Whether a Connection Supports Sessions`_) and MUST NOT consume a server  
+The driver MUST NOT use the session if the checked out connection does not support sessions
+(see `How to Tell Whether a Connection Supports Sessions`_) and, in all cases, MUST NOT consume a server
 session id until after the connection is checked out and session support is confirmed.
 
 Sessions and Cursors
@@ -502,7 +502,7 @@ A driver SHOULD NOT attempt to release the acquired session before connection ch
 
 Explicit sessions MAY be changed to allocate a server session similarly.
 
-How to Check Whether a Connection Supports Sessions
+How to Tell Whether a Connection Supports Sessions
 ===================================================
 
 A driver can determine whether a connection supports sessions by checking whether
--- a/source/sessions/driver-sessions.rst
+++ b/source/sessions/driver-sessions.rst
@@ -523,9 +523,8 @@ to the server:
   opened (and reported a value for logicalSessionTimeoutMinutes in the initial
   response to the `handshake <https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.rst>`_),
   but have subsequently been downgraded to not support sessions. The server does
-  not close the socket in this scenario, and the driver will forever conclude that
-  the server at the other end of this connection supports sessions. This scenario
-  will only be a problem until the next heartbeat against that server.
+  not close the socket in this scenario, so the driver will conclude that
+  the server at the other end of this connection supports sessions.
 
 There is nothing that the driver can do about this race condition, and the server 
 will just return an error in this scenario.
--- a/source/sessions/driver-sessions.rst
+++ b/source/sessions/driver-sessions.rst
@@ -203,10 +203,9 @@ instances at the same time (see the Server Session Pool section). Additionally,
 a ``ClientSession`` may only ever be associated with one ``ServerSession`` for
 its lifetime.
 
-Drivers MUST report an error if sessions are not supported by the deployment
-(see How to Check Whether a Deployment Supports Sessions). This error MUST either
-be reported by ``startSession``, or be reported the first time the session is used
-for an operation.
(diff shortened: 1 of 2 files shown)


source/sessions/driver-sessions.rst
```diff
@@ -553,17 +527,8 @@ to the server:
   the server at the other end of this connection supports sessions. This scenario
   will only be a problem until the next heartbeat against that server.
```
ShaneHarvey, line 528: "This scenario will only be a problem until the next heartbeat against that server." is no longer true and can be removed now.

source/sessions/tests/README.rst
```diff
@@ -17,26 +17,25 @@ Several prose tests, which are not easily expressed in YAML, are also presented
...
-Snapshot session tests
-======================
-Snapshot sessions tests require server of version 5.0 or higher and
-replica set or a sharded cluster deployment.
-Default snapshot history window on the server is 5 minutes. Running the test in debug mode, or in any other slow configuration
-may lead to `SnapshotTooOld` errors. Drivers can work around this issue by increasing the server's `minSnapshotHistoryWindowInSeconds` parameter, for example:
-
-.. code:: python
-
-    client.admin.command('setParameter', 1, minSnapshotHistoryWindowInSeconds=60)
```
ShaneHarvey, line 29: AFAICT minSnapshotHistoryWindowInSeconds is unrelated to the ``snapshot`` and ``causalConsistency`` test described below. Could you revert this change? 

Also if the default is 5 minutes, wouldn't 60 seconds be a _decrease_? Wouldn't decreasing the window from 5 minutes to minute make a SnapshotTooOld error more likely?

source/sessions/tests/README.rst
```diff
@@ -234,6 +233,39 @@ This test only applies to drivers that allow authentication to be changed on the
...
+
+This test uses a mongocryptd server as the test server; however, in the event that mongocryptd 
+starts supporting sessions or is not a viable option for this test, another server may be substituted
+in these tests as long as it does not return a non-null value for ``logicalSessionTimeoutMinutes``;
+in the event that no such server is readily available, a mock server may be used instead. 
+
+* Ensure a mongocryptd process is running
+* Create a new ``MongoClient`` pointed at the mongocryptd server with command monitoring enabled
+* Verify that the server does NOT define a value for ``logicalSessionTimeoutMinutes`` by sending a hello command to the server and checking the response
+* Send a read command to the server (e.g., ``findOne``), ignoring any errors from the server response
```
ShaneHarvey, line 247: I take it the error happens because mongocryptd doesn't actually support find or insert commands. Instead of ignoring the error can we assert that a particular error code is returned?

source/sessions/tests/README.rst
```diff
@@ -234,6 +233,39 @@ This test only applies to drivers that allow authentication to be changed on the
...
+in the event that no such server is readily available, a mock server may be used instead. 
+
+* Ensure a mongocryptd process is running
+* Create a new ``MongoClient`` pointed at the mongocryptd server
+* Verify that the server does NOT define a value for ``logicalSessionTimeoutMinutes`` by sending a hello command to the server and checking the response
+* Create a new explicit session by calling ``startSession`` (this MUST NOT error)
+* Attempt to send a read command to the server (e.g., ``findOne``) with the explicit session passed in
+* Assert that a client-side error is generated indicating that sessions are not supported
+* Attempt to send a write command to the server (e.g., ``insertOne``) with the explicit session passed in
+* Assert that a client-side error is generated indicating that sessions are not supported
```
ShaneHarvey, line 267: Should we combine these tests to avoid the duplication? I imagine when I implement this in python I would make it one test anyway:
```
...
* Send a read command to the server (e.g., ``findOne``), ignoring any errors from the server response
* Check the corresponding ``commandStarted`` event: verify that ``lsid`` is not set
* Send a write command to the server (e.g., ``insertOne``), ignoring any errors from the server response
* Check the corresponding ``commandStarted`` event: verify that lsid is not set
* Create a new explicit session by calling ``startSession`` (this MUST NOT error)
* Attempt to send a read command to the server (e.g., ``findOne``) with the explicit session passed in
* Assert that a client-side error is generated indicating that sessions are not supported
* Attempt to send a write command to the server (e.g., ``insertOne``) with the explicit session passed in
* Assert that a client-side error is generated indicating that sessions are not supported
```

ShaneHarvey (changes requested):

source/sessions/tests/README.rst
dariakp: This was under the heading for snapshot sessions, I didn't change the text, so I actually don't know what it is referencing. There was previously a "snapshot sessions" heading that all tests (incorrectly) got grouped under, even though only the first test in the list had to do with snapshot sessions. I think you were the one to add the original note, so I'm happy to defer to your judgment.

source/sessions/tests/README.rst
dariakp: Correct; and yes, we could, but in this case, we don't actually care whether the command succeeds or fails, or how it succeeds or fails, so there is no benefit to asserting on the particulars (especially since the test allows for testing against a different server) - we only care about asserting on what we send to the actual server. My preference is to keep prose tests focused on the aspects that are important to the validity of the scenario under test.

source/sessions/tests/README.rst
dariakp: Code reuse is an implementation detail for the individual driver. There are two distinct scenarios being tested, so from the perspective of clarity of intent, it makes sense to have two distinct test cases (even if some of the setup is shared). The implicit session scenario could fail while the explicit session scenario is passing and vice versa. If a driver ever has a bug which respect to one of these conditions, there is benefit to having one of the tests implemented and passing and one skipped versus having a combined test skipped or a combined test running where the description doesn't match the assertions because some of them must be skipped until a fix is implemented.

source/sessions/driver-sessions.rst
dariakp: Updated the wording, thanks for the catch!

source/sessions/tests/README.rst
BorisDog: It's a general remark for all snapshots tests in unified format. In fact it does not apply to the snapshot prose test, as it just configuration validation.
So I think the original version is correct.

This spec was developed before the default [increase to 5 mins](https://jira.mongodb.org/browse/SERVER-47855). Looks like it was also backported to [5.0.0, 5.1.0) versions (C# uses 5.0.14 in EG). In this case we might not longer need this setting, but that needs to be double checked.

dariakp: Ah, so it was intended for the unified tests and not for the prose tests. The heading made it seem like it was a note for the following snapshot prose tests. Thanks for clarifying! The official docs all say that the default is 5 minutes (`=300`) for both 5.0 and 6.0. It seems like perhaps the increase may have originally intended to be `=600` since that is exactly the example used in the [docs](https://www.mongodb.com/docs/v5.0/reference/parameters/#mongodb-parameter-param.minSnapshotHistoryWindowInSeconds). I can update the guidance accordingly and move it up to be with the unified test notes, or I can remove it entirely. Thoughts?

source/sessions/tests/README.rst
ShaneHarvey: I agree it's up to the implementor wether to combine or not but I would still prefer to avoid duplicating the test setup part in the description (with the aim of making these tests easier to read/maintain). If there's no immediately obvious way to do that then I'm happy as is.

source/sessions/tests/README.rst
ShaneHarvey: Fair enough. Usually I worry about blindly ignoring errors but, like you point out, we're asserting on the command events so it seems fine.

source/sessions/tests/README.rst
ShaneHarvey: Thanks for tracking this down. Keeping at 600 SGTM but it's worth pointing out that we've never needed to use this workaround in pymongo's test suite.

source/sessions/tests/README.rst
dariakp: I did think about combining the note about mongocryptd setup: if you think it would improve readability, I could factor this out as a note about "Testing against servers that do not support sessions" and then reference it in both test cases

source/sessions/tests/README.rst
BorisDog: I think it's worth keeping this note for debugging purposes.
600 SGTM.  Thanks for catching this.

source/sessions/tests/README.rst
dariakp: Made the corresponding update, let me know how that looks

source/sessions/tests/README.rst
dariakp: Updated

ShaneHarvey (approved):

BorisDog (approved):
LGTM.

PR #1381, opened 2023-02-28
DRIVERS-2415 OIDC SASL Clarifications

blink1073:

- [x] Update changelog.


blink1073:
cc @dima @durran

source/auth/auth.rst
```diff
@@ -1438,6 +1438,11 @@ since it is has been identified as expired by the server.  If a refresh
 callback is given, it will be called as usual.  Otherwise the request callback
 will be called.
 
+If the command fails and callbacks are in use, the driver
+MUST clear the cache value and attempt to authenticate more time from
```
durran, line 1442:
```suggestion
MUST clear the cache value and attempt to authenticate one more time from
```

durran (changes requested):

durran (approved):

alcaeus (approved):

source/auth/tests/mongodb-oidc.rst
```diff
@@ -130,6 +130,47 @@ credentials.
...
+
+#. Clear the cache.
+#. Create a client with a request callback that returns a valid token
+  that will expire soon.
+#. Set a fail point for ``saslStart`` commands of the form:
+
+.. code:: javascript
+
+    {
+        "mode": {"times": 2},
```
DmitryLukyanov, line 163: let's add appName to ensure that unused configuration won't affect rest of tests

DmitryLukyanov: also, let's sue the full form like in reauth example

blink1073: We have a different way of handling it it Python, updated.

source/auth/tests/mongodb-oidc.rst
```diff
@@ -130,6 +130,66 @@ credentials.
...
+        "errorCode": 18
+      }
+    }
+
+Note: the driver MUST either use a unique ``appName`` or explicitly
+remove the ``failCommand`` after the test to prevent leakage.
+
+#. Perform a ``find`` operation.
+#. Close the client.
+#. Create a new client with the same properties.
```
durran, line 161: Do we need to do this repeatedly? It seems like if the first find succeeds we know the speculative auth was in the initial handshake correct?

source/auth/tests/mongodb-oidc.rst
```diff
@@ -160,6 +220,9 @@ of listening for SASL commands.
       }
     }
 
+Note: the driver MUST either use a unique ``appName`` or explicitly
+remove the ``failCommand`` after the test to prevent leakage.
+
```
DmitryLukyanov, line 225: if we use speculative auth, the below steps regarding saslStart will be different

DmitryLukyanov (approved):
LGTM

source/auth/tests/mongodb-oidc.rst
```diff
@@ -160,8 +236,67 @@ of listening for SASL commands.
...
-#. Assert that a ``find`` operation failed once during the command execution.
+.. note::
+
+  the driver MUST either use a unique ``appName`` or explicitly
+  remove the ``failCommand`` after the test to prevent leakage.
+
+- Perform another find operation that succeeds.
+- Assert that the refresh callback has been called, if possible.
+- Assert that a ``find`` operation was started twice and a ``saslStart`` operation was started once during the command execution.
+- Assert that a ``find`` operation succeeeded once and the ``saslStart`` operation succeeded during the command execution.
```
baileympearson, line 247: Maybe it's worth clarifying the wording here.  Node's initial implementation counts the number of command started / succeeded events and asserts on event counts, but other scenarios would satisfy these same event counts. (a connection connection + a normal retryable find would produce the same event counts but in the wrong order)

```suggestion
- Assert that the ordering of command started events is `find`, `saslStart` , `find`.
- Assert that the ordering of command succeeded events is `saslStart`, `find`.
```

source/auth/tests/mongodb-oidc.rst
```diff
@@ -97,53 +112,110 @@ verify usage.  Unless otherwise specified, the tests MUST be performed with
...
+~~~~~~~~~~~~~~~~~~
+- Clear the cache.
+- Create a new client with a request callback and a refresh callback.  Both callbacks will read the contents of the ``AWS_WEB_IDENTITY_TOKEN_FILE`` location to obtain a valid access token.
+- Validate the request callback inputs, including the timeout parameter if
+  possible.
+- Give a callback response with a valid accessToken and an expiresInSeconds
+  that is within one minute.
+- Ensure that a ``find`` operation adds credentials to the cache.
+- Create a new client with the same request callback and a refresh callback.
+- Ensure that a ``find`` operation results in a call to the refresh callback.
```
katcharov, line 125: I believe that clients are not required and only lowercase-should advised to use a global cache (in Java, we are opting to omit it until customer need arises). Could conditional instructions be provided (if global cache ... if no global cache...)? Or if that is not possible, a general instruction (similar to "clear the cache") that states what must obtain, but omits the details?

source/auth/tests/mongodb-oidc.rst
```diff
@@ -97,53 +112,110 @@ verify usage.  Unless otherwise specified, the tests MUST be performed with
...
+~~~~~~~~~~~~~~~~~~~~~
+- Clear the cache.
+- Ensure there is a cache with credentials that will expire in less than 5 minutes, using a client with an appropriate request callback.
+- Create a new client with the a request callback but no refresh callback.
+- Ensure that a ``find`` operation results in a call to the request callback.
+- Close the client.
+
+Cache key includes callback
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+If the driver does not support using callback hashes as part of the cache key,
```
katcharov, line 140: Could the "callback hashes" term be clarified? It is not mentioned in the spec. Some drivers are choosing to compare callbacks by reference - is that variation addressed?

source/auth/tests/mongodb-oidc.rst
```diff
@@ -97,53 +112,110 @@ verify usage.  Unless otherwise specified, the tests MUST be performed with
...
+- Close the client.
+
+Error clears cache
+~~~~~~~~~~~~~~~~~~
+- Clear the cache.
+- Create a new client with a valid request callback that gives credentials that expire within 5 minutes and a
+  refresh callback that gives invalid credentials.
+- Ensure that a ``find`` operation adds a new entry to the cache.
+- Ensure that a subsequent ``find`` operation results in an error.
+- Ensure that the cache has been cleared.
```
katcharov, line 156: Could this (and perhaps elsewhere) be explicit about what items are stored in the cache, and what items are being cleared? (If I understand correctly, there are two credentials and one original response?)

source/auth/tests/mongodb-oidc.rst
```diff
@@ -97,53 +112,110 @@ verify usage.  Unless otherwise specified, the tests MUST be performed with
...
+- Validate the refresh callback inputs, including the timeout parameter if
+  possible.
+- Close the client.
+
+Cache with no refresh
+~~~~~~~~~~~~~~~~~~~~~
+- Clear the cache.
+- Ensure there is a cache with credentials that will expire in less than 5 minutes, using a client with an appropriate request callback.
+- Create a new client with the a request callback but no refresh callback.
+- Ensure that a ``find`` operation results in a call to the request callback.
```
katcharov, line 135:
```suggestion
- Perform a ``find`` operation that succeeds.
- Ensure that the request callback has been called once.
```

1. Might it be clearer to make the steps consistent with earlier sections (1, 2)?
2. It seems fine to say "a" find operation, but shouldn't we check that there is exactly "one" call to the callback? (Here and elsewhere.)

katcharov (changes requested):
Thanks for the changes in this PR, I find them helpful. Apologies for the number of comments - I didn't review the original PR, and this is also the first large spec I'm implementing (Java).

source/auth/tests/mongodb-oidc.rst
```diff
@@ -97,53 +112,110 @@ verify usage.  Unless otherwise specified, the tests MUST be performed with
 the authorization code workflow with and without a provided refresh callback.
 If desired, the caching tests can be done using mock server responses.
```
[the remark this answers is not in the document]
blink1073: 1. Done
2. I moved that sentence to the top level notes
3. I removed the note about overriding creds
4. That sentence was a mistake, the tests should provide both callbacks unless otherwise noted.  Updated
5. I added a clarifying sentence
6. I switched the terminology to use "callback-driven"
7. Done

source/auth/tests/mongodb-oidc.rst
```diff
@@ -4,89 +4,104 @@ MongoDB OIDC
...
 
+Drivers MUST be able to authenticate using the "aws" provider workflow
+simulating an EC2 instance with an enabled web identity token provider,
+generated by Drivers Evergreen Tools.
+
+Single Principal
+~~~~~~~~~~~~~~~~
+- Set the ``AWS_WEB_IDENTITY_TOKEN_FILE`` environment variable to the location
+  of the ``test_user1`` generated token file.
+- Perform a ``find`` operation that succeeds.
```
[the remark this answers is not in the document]
blink1073: Fixed

source/auth/tests/mongodb-oidc.rst
```diff
@@ -97,53 +112,110 @@ verify usage.  Unless otherwise specified, the tests MUST be performed with
...
+- Create a new client with the same request callback and a refresh callback.
+- Ensure that a ``find`` operation results in a call to the refresh callback.
+- Validate the refresh callback inputs, including the timeout parameter if
+  possible.
+- Close the client.
+
+Cache with no refresh
+~~~~~~~~~~~~~~~~~~~~~
+- Clear the cache.
+- Ensure there is a cache with credentials that will expire in less than 5 minutes, using a client with an appropriate request callback.
```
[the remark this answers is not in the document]
blink1073: Cleaned up to be self-contained.

source/auth/tests/mongodb-oidc.rst
blink1073: I added a note in the "cached credentials" intro paragraph

source/auth/tests/mongodb-oidc.rst
blink1073: I added that option

source/auth/tests/mongodb-oidc.rst
blink1073: Done

source/auth/tests/mongodb-oidc.rst
blink1073: This whole section was refactored, and there is no callback validation.

source/auth/tests/mongodb-oidc.rst
```diff
@@ -4,89 +4,104 @@ MongoDB OIDC
...
 =================
 
-Ensure the of the following conditions raise an error:
+Ensure the of the following conditions raise an error on a ``find`` operation:
 
-#. Any callback returns null
-#. Any callback returns unexpected result
+- Request callback returns null
+- Refresh callback returns null
+- Request callback returns unexpected result
```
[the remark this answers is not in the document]
katcharov: Is it possible that future versions might return additional fields into such a callback, which would cause older drivers to fail?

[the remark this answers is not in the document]
blink1073: Good point, I've relaxed this requirement.

katcharov (approved):
LGTM

source/auth/tests/mongodb-oidc.rst
```diff
@@ -4,164 +4,350 @@ MongoDB OIDC
...
+- Perform a ``find`` operation that succeeds.
+- Close the client.
+
+Multiple Principal User 1
+~~~~~~~~~~~~~~~~~~~~~~~~~
+- Create a client with a url of the form ``mongodb://localhost:27018/?authMechanism=MONGODB-OIDC&authMechanismProperties=PROVIDER_NAME:aws&directConnection=true&readPreference=secondaryPreferred``.
+- Perform a ``find`` operation that succeeds.
+- Close the client.
+
+Multiple Principal User 2
```
katcharov, line 97: The subsections "Multiple Principal User 1" and 2 are identical.

source/auth/auth.rst
```diff
@@ -1210,24 +1210,12 @@ username.
...
-          URL where the IDP may be contacted for authorization
-          code <=> ID/access token exchange.
-        type: string
-        optional: true # Req if deviceAuthorizeEndpoint not present
-      deviceAuthorizationEndpoint:
-        description: >-
-          URL where the IDP may be contacted for device
-          authentication and authorization code generation.
-        type: string
-        optional: true # Req if authorizeEndpoint not present
```
addaleax, line 1230: `clientSecret` below also needs to be dropped to match [SERVER-75122](https://jira.mongodb.org/browse/SERVER-75122), right?

source/auth/auth.rst
```diff
@@ -1210,24 +1210,12 @@ username.
     description: "Server's reply to clientStep1"
     strict: false
     fields:
```
addaleax, line 1212: Just a note – we know it’s likely that in the future, other information will need to be passed from the driver to callbacks, e.g. the SRV hostname for resource indicators in the sense of RFC8707. Does it make sense to prepare for that? For example, instead of passing `principalName` as a string parameter to the callbacks, it could be a BSON document like `OIDCMechanismServerStep1` currently is.

source/auth/auth.rst
blink1073: Good call out, thanks! Fixed

source/auth/auth.rst
```diff
@@ -1281,6 +1269,14 @@ mechanism_properties
         Drivers MUST allow the user to specify a callback of the form
         "onOIDCRefreshToken" (defined below), if the driver supports
         providing objects as mechanism property values.
+    ALLOWED_HOSTS
+        The list of allowed hosts for MongoDB connections, allowing Unix shell-style wildcards. This is a security feature and defaults to
+        ``[*.mongodb.net, *.mongodb-dev.net, *.mongodbgov.net, localhost]``.
+        When ``MONGODB-OIDC`` authentication is attempted against a hostname
+        that does not match any of list of allowed hosts, the driver MUST
+        raise a client-side error without invoking any user-provided
```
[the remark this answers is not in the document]
blink1073: It should be after. I added a note

source/auth/auth.rst
blink1073: I merged the `timeout` parameter into that object as well

durran (approved):

source/auth/tests/mongodb-oidc.rst
```diff
@@ -4,164 +4,386 @@ MongoDB OIDC
...
+- Clear the cache.
+- Create a request callback that reads in the generated ``test_user2`` token file.
+- Create a client with a url of the form ``mongodb://test_user2@localhost:27018/?authMechanism=MONGODB-OIDC&directConnection=true&readPreference=secondaryPreferred`` and a valid OIDC request callback.
+- Perform a ``find`` operation that succeeds.
+- Close the client.
+
+Multiple Principal No User
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+- Clear the cache.
+- Create a client with a url of the form ``mongodb://localhost:27018/?authMechanism=MONGODB-OIDC&directConnection=true&readPreference=secondaryPreferred``and a valid OIDC request callback.
```
katcharov, line 74:
```suggestion
- Create a client with a url of the form ``mongodb://localhost:27018/?authMechanism=MONGODB-OIDC&directConnection=true&readPreference=secondaryPreferred`` and a valid OIDC request callback.
```
Spacing breaks formatting

source/auth/tests/mongodb-oidc.rst
```diff
@@ -4,164 +4,386 @@ MongoDB OIDC
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+- Clear the cache.
+- Create a client with a url of the form ``mongodb://localhost:27018/?authMechanism=MONGODB-OIDC&directConnection=true&readPreference=secondaryPreferred`` and a valid OIDC request callback.
+- Assert that a ``find`` operation fails.
+- Close the client.
+
+OIDC Allowed Hosts Blocked
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+- Clear the cache.
+- Create a client that uses the OIDC url and the request callback, and an
```
DmitryLukyanov, line 81: should we be more specific here? At least to check cases where OIDC is configured with callbacks and deviceName?

DmitryLukyanov: Maybe with not only empty case, but with something like `localhost1` value in the allowed list?

source/auth/tests/mongodb-oidc.rst
```diff
@@ -4,164 +4,386 @@ MongoDB OIDC
...
+  ``authOIDCAllowHosts`` that is an empty list.
```
DmitryLukyanov, line 82: also, now it's called `ALLOWED_HOSTS` auth property

source/auth/tests/mongodb-oidc.rst
blink1073: Done

source/auth/tests/mongodb-oidc.rst
blink1073: Fixed

source/auth/auth.rst
```diff
@@ -1281,6 +1265,17 @@ mechanism_properties
         Drivers MUST allow the user to specify a callback of the form
         "onOIDCRefreshToken" (defined below), if the driver supports
         providing objects as mechanism property values.
+    ALLOWED_HOSTS
+        The list of allowed hostnames or ip-addresses (ignoring ports) for
+        MongoDB connections, allowing Unix shell-style wildcards. This is a
+        security feature and defaults to
+        ``["*.mongodb.net", '*.mongodb-dev.net", "*.mongodbgov.net", "localhost", "127.0.0.1"]``.
```
JamesKovacs, line 1272: Should `::1` also be included?

source/auth/auth.rst
```diff
@@ -1281,6 +1265,17 @@ mechanism_properties
         Drivers MUST allow the user to specify a callback of the form
         "onOIDCRefreshToken" (defined below), if the driver supports
         providing objects as mechanism property values.
+    ALLOWED_HOSTS
+        The list of allowed hostnames or ip-addresses (ignoring ports) for
+        MongoDB connections, allowing Unix shell-style wildcards. This is a
```
JamesKovacs, line 1270: We should specify what we mean by "Unix shell-style wildcards". Should the impl support `*` (yes), `?` (maybe), and `[]` (possibly)? What about `[a-d]`? How about `[[:upper:]]`? `[![:digit:]]`? I suspect that we want to support `*` and maybe `?`, but nothing else.

Furthermore when it comes to `*`, how greedy is the matching? If you specify:

```
*.example.com
```

which of the following will be successfully matched?

```
dev.example.com
dev.subdomain.example.com
evildomain.info?ignored=example.com
```

First one, yes. Second one, probably not, but maybe. Third one definitely not, but we should have some tests to verify.

JamesKovacs (changes requested):
Requesting clarification of glob pattern matching semantics for `ALLOWED_HOSTS`.

source/auth/auth.rst
blink1073: I based it on the semantics of `fnmatch` in [Python](https://docs.python.org/3/library/fnmatch.html#module-fnmatch), which closely matches the glob wiki [page](https://en.wikipedia.org/wiki/Glob_(programming)).  We could make it more explicit here:

image

source/auth/auth.rst
blink1073: Yes, that makes sense, updated

source/auth/auth.rst
DmitryLukyanov: NOTE: I'm not sure whether any language has easy support for this filtering. At this point, in c# impl I implement this logic manually. Maybe regex can be more portable since I would imagine it's more common functionality for all languages?

addaleax: Yeah, Spencer also initially mentioned that this could be a regex; the reason why I preferred not to do that is that it can be hard to get a regex “right” and not have it accidentally match more than intended.

My suggestion here would be to only allow `*` as a wildcard, with the semantics of `[a-zA-Z0-9:.]+` in regex language (i.e. match characters that are valid in domain names or IP addresses, ignoring things like IDNA).

source/auth/tests/mongodb-oidc.rst
```diff
@@ -4,164 +4,397 @@ MongoDB OIDC
...
+
+Error clears cache
+~~~~~~~~~~~~~~~~~~
+- Clear the cache.
+- Create a new client with a valid request callback that gives credentials
+  that expire within 5 minutes and a refresh callback that gives invalid
+  credentials.
+- Ensure that a ``find`` operation adds a new entry to the cache.
+- Ensure that a subsequent ``find`` operation results in an error.
+- Ensure that the cached token has been cleared.
```
katcharov, line 253: If a find causes a division by zero during query execution, is the intent to clear the cache?

blink1073: Not the full cache, but marking the access token as expired or removing it.

source/auth/auth.rst
blink1073: I agree with @addaleax on all points

source/auth/tests/mongodb-oidc.rst
```diff
@@ -4,164 +4,397 @@ MongoDB OIDC
...
+
+Multiple Principal User 2
+~~~~~~~~~~~~~~~~~~~~~~~~~
+- Set the ``AWS_WEB_IDENTITY_TOKEN_FILE`` environment variable
+  to the location of valid ``test_user2`` credentials.
+- Create a client with a url of the form ``mongodb://localhost:27018/?authMechanism=MONGODB-OIDC&authMechanismProperties=PROVIDER_NAME:aws&directConnection=true&readPreference=secondaryPreferred``.
+- Perform a ``find`` operation that succeeds.
+- Close the client.
+- Restore the ``AWS_WEB_IDENTITY_TOKEN_FILE`` environment variable
+  to the location of valid ``test_user2`` credentials.
```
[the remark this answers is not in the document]
blink1073: Can we move this comment to the WIP document so we're only working in one place at a time?

source/auth/auth.rst
blink1073: On second thought, @DmitryLukyanov and think we should go with regex: it is easier and less error-prone than implementing our own custom logic, and easier for the user to reason about and test using standard tools.

source/auth/tests/mongodb-oidc.rst
katcharov: (Could we use the term "Access Token" to refer to that specific token?)

Why would such an error indicate that the token is expired or otherwise invalid?

In cases where we don't have a refresh, the user would need to manually reauthenticate any time an error is encountered (even if it is otherwise ignored by the application). In cases with a refresh, it seems it would be called unnecessarily.

source/auth/auth.rst
katcharov: It might help if we considered some examples (as in @JamesKovacs 's original comment)

     "www.mongodbgov.net",

If we use regex, wouldn't this match `wwwxmongodbgov.net`? This seems too easy to get wrong.

DmitryLukyanov: From another side, it's easy to check whether particular host matches to provided expression :)

addaleax: If we do go with the regex approach (and I am not entirely convinced that that’s a good idea, but also practically speaking want to see this PR move forward of course), are we convinced that all drivers have access to regex libraries with reasonably similar semantics? Or are we okay with the same ALLOWED_HOSTS string resulting in slightly different validation rules with different drivers?

blink1073: What if we only allowed a leading greedy asterisk on the parsed host name, since that matches the concept of a subdomain.  In James' example, the first two would match and the last would not, because the parsed host name would be `evildomain.info`:

image

source/auth/tests/mongodb-oidc.rst
```diff
@@ -4,164 +4,397 @@ MongoDB OIDC
...
+~~~~~~~~~~~~~~~
+- Clear the cache.
+- Create request and refresh callback that validate their inputs and return
+  a valid token.  The request callback must return a token that expires in
+  one minute.
+- Create a client that uses the above callbacks.
+- Perform a ``find`` operation that succeeds.  Verify that the request
+  callback was called with the appropriate inputs, including the timeout
+  parameter if possible.  Ensure that there are no unexpected fields.
+- Perform another ``find`` operation that succeeds.  Verify that the refresh
```
durran, line 139: How is the refresh callback supposed to be called in this test? Is it expected that the second `find` happens on a separate connection in order for another initial handshake to occur? Should this be behaving as later tests that `find` first, then close the client before creating a new client and then issuing a new `find`?

source/auth/tests/mongodb-oidc.rst
```diff
@@ -4,164 +4,397 @@ MongoDB OIDC
...
+- Close the client.
+
+Request Callback Returns Null
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+- Clear the cache.
+- Create a client with a request callback that returns ``null``.
+- Perform a ``find`` operation that fails.
+- Close the client.
+
+Refresh Callback Returns Null
```
durran, line 151: Same comment on how this test expects the refresh callback to actually get called here.

durran (changes requested):

source/auth/tests/mongodb-oidc.rst
durran: Edit: I seemed to answer my question looking at the Python PR - so could we update these 2 areas to note the client must be closed and a new one created like the tests that do the same later in the prose tests?

source/auth/tests/mongodb-oidc.rst
```diff
@@ -4,164 +4,397 @@ MongoDB OIDC
...
+the cache key, skip this test.  This test ensures that the callback is
+considered as part of the cache key.
+
+- Clear the cache.
+- Create a new client with a request callback that does not give an
+  ```expiresInSeconds``` value.
+- Ensure that a ``find`` operation adds credentials to the cache.
+- Close the client.
+- Create a new client with a different request callback.
+- Ensure that a ``find`` operation adds a new entry to the cache.
```
durran, line 242:
```suggestion
- Ensure that a ``find`` operation replaces the one-time use entry and adds a new entry to the cache.
```

source/auth/auth.rst
```diff
@@ -1290,8 +1285,33 @@ User Provided Callbacks
...
+        principalName:
+            description: "The client identifier provided in the access token,
+            typically the Subject Identifier."
+            type: string
+        timeout_seconds:
+            description: "The number of seconds before a request should time out"
+            optional: true
+        timeout_context:
+            description: "A timeout context variable used to cancel the callback"
+            optional: true
```
addaleax, line 1312: I know it’s late, but here’s an idea: Let’s version this struct? That way we can prevent using a version of the callback implementation that doesn’t understand potential future requirements that the driver puts on it. (For example, when we drop the `ALLOWED_HOSTS` restriction, the callbacks may need to perform some type of validation of their own.)

```suggestion
            optional: true
        version:
            description: "A version identifying breaking changes in the callback protocol, currently 0"
            type: int
```

source/auth/tests/mongodb-oidc.rst
```diff
@@ -4,146 +4,430 @@ MongoDB OIDC
...
+- Assert that the ordering of list started events is [``find``],
+  , ``find``.  Note that if the listener stat could not be cleared then there
+  will and be extra ``find`` command.
+- Assert that the list of command succeeded events is [``find``].
+- Assert that a ``find`` operation failed once during the command execution.
+- Close the client.
 
-Cached Credentials
-==================
+Retries and Succeeds with Cache
```
durran, line 367: What is the purpose of this reauth test and the following one? (They appear to be doing the same thing but one has a success assertion and the other has a fail assertion). And the spec states that reauthentication must not be using cache values so I'm wondering if these 2 tests are actually needed?

durran (changes requested):

source/auth/tests/mongodb-oidc.rst
blink1073: These two are covering the case where the server info has changed, but we are using stale server info.  I agree "Separate Connections Avoid Extra Callback Calls" can be merged with "Succeeds", will do.

blink1073: No actually, that would complicate the "Succeeds" test.  This test is about ensuring that we do not call the callbacks more than is necessary, by adding guards/checks on whether a cache value has been updated since we last used it.

source/auth/tests/mongodb-oidc.rst
blink1073: Updated

source/auth/tests/mongodb-oidc.rst
blink1073: Updated

source/auth/auth.rst
```diff
@@ -259,16 +259,16 @@ Reauthentication
 
 On any operation that requires authentication, the server may raise the
 error ``ReauthenticationRequired`` (391), typically if the user's credential
-has expired.  Drivers MUST immediately attempt to force a reauthenication on
-the connection when this error is raised, and then re-attempt the operation.
+has expired.  Drivers MUST immediately attempt a reauthenication on
```
addaleax, line 262:
```suggestion
has expired.  Drivers MUST immediately attempt a reauthentication on
```

source/auth/auth.rst
```diff
@@ -1177,76 +1177,56 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
-        optional: true
       requestScopes:
         description: "Additional scopes to request from IDP"
         type: array<string>
         optional: true
 
-Server will use principalName (n) if provided in clientStep1 to select an appropriate IDP.  This IDP's configuration will be returned in serverStep1 to instruct the client on how to acquire an Access Token.  This Access Token will be used in clientStep2 to complete authentication.  The principalName identified within the Access Token MUST match any provided during clientStep1, or the authentication operation will fail.
+Server will use principalName (n) if provided in the driver’s PrincipalStepRequest to select an appropriate IDP.  This IDP's configuration will be returned in the server’s response that will be used by the end-user to acquire an Access Token.
+
+This Access Token will be used as the JWT in the driver’s JwtStepRequest to complete authentication. Users implementing callbacks must ensure that the principalName identified within the Access Token must match the principalName provided to the callback, otherwise the JwtStepRequest will fail on the server.
```
addaleax, line 1224: [non-blocking, can be clarified later] Given that the callbacks do not receive the principalName directly anymore, I'm not sure how much sense the clarification in the last sentence of this paragraph makes anymore

source/auth/auth.rst
```diff
@@ -1272,30 +1251,49 @@ mechanism
...
-        providing objects as mechanism property values.
-
-Drivers MUST skip client step 1 when using a service to obtain credentials
-or when the server step 1 response value is cached.  When skipping step 1,
-drivers will use ``saslStart`` and a payload with the ``jwt`` value.
+        "onRefresh" (defined below), if the driver supports
+        providing objects as mechanism property values.  Otherwise the driver MUST allow it as a MongoClientOption.
+    ALLOWED_HOSTS
+        The list of allowed hostnames or ip-addresses (ignoring ports) for
+        MongoDB connections. The hostnames may include a leading "*." wildcard, which allows for matching subdomains. ALLOWED_HOSTS is a
```
addaleax, line 1265: At least Github's RST processing seems to process the `*` as an italics marker

Also, it might be good to clarify:

```suggestion
        MongoDB connections. The hostnames may include a leading "*." wildcard, which allows for matching (potentially  nested) subdomains. ALLOWED_HOSTS is a
```

source/auth/auth.rst
```diff
@@ -1311,47 +1309,38 @@ the form:
...
-    function onOIDCRequestToken(principalName: str, serverInfo: OIDCMechanismServerStep1, timeoutSeconds: int): OIDCRequestTokenResult
+The token refresh callback must take the same arguments as the request callback, as well as the ``refreshToken`` string given by the ``IdpServerResponse``.
 
 Callbacks can be synchronous and/or asynchronous, depending on the driver
 and/or language.  Asynchronous callbacks should be preferred when other
 operations in the driver use asynchronous functions.
 
-Before calling a callback, the driver MUST acquire a lock, and verify that
-there are no valid cache credentials before calling the callback.
+Before calling a callback, the driver MUST acquire a lock, and, if there is a non-expired cached Access Token, attempt to authenticate via a JwtStepRequest before calling the callback.  The lock is released after all attempts to authenticate have failed, or a non-recoverable error was encountered.  The driver MUST ensure that credentials have not changed between when the lock was requested and when it was acquired.
```
addaleax, line 1318: Probably makes sense to clarify this later unless it's trivial, but since the cache is supposed to be global, is this lock also supposed to be global? That would not be ideal imo but I'm wondering if it's implied here.

addaleax (approved):

source/auth/auth.rst
katcharov, line 262: Because we are now tracking connection status (via generation id, Access Token comparison, etc.) we are able to check (in a non-blocking manner) whether the connection has become invalid before each operation (which would be guaranteed to throw a 391 if we try to send it). We can then preemptively re-authenticate, rather than only re-authenticating after seeing the 391.

If we don't do this pre-emptive check, we will be unnecessarily sending a command that we know will fail, per connection per Access Token expiry.

Should we mandate this check, and do the tests allow for this behaviour?

source/auth/tests/mongodb-oidc.rst
```diff
@@ -4,146 +4,437 @@ MongoDB OIDC
...
+
+Error clears cache
+~~~~~~~~~~~~~~~~~~
+- Clear the cache.
+- Create a new client with a valid request callback that gives credentials
+  that expire within 5 minutes and a refresh callback that gives invalid
+  credentials.
+- Ensure that a ``find`` operation adds a new entry to the cache.
+- Ensure that a subsequent ``find`` operation results in an error.
+- Ensure that the cached access token has been cleared.
```
katcharov, line 278: Following up on [earlier](https://github.com/mongodb/specifications/pull/1381#discussion_r1163048854). I believe the spec now says that Access Tokens are cleared only when the operation fails due to a 391. It seems that we should have a test that confirms that Access Tokens are preserved when a find fails for a reason that is not 391 (like division by 0 in the query).

source/auth/auth.rst
```diff
@@ -1272,30 +1251,49 @@ mechanism
...
+        ``["*.mongodb.net", '*.mongodb-dev.net", "*.mongodbgov.net", "localhost", "127.0.0.1", "::1"]``.
+        When ``MONGODB-OIDC`` authentication is attempted against a hostname
+        that does not match any of list of allowed hosts, the driver MUST
+        raise a client-side error without invoking any user-provided
+        callbacks.  This value MUST not be allowed in the URI connection
+        string.  The hostname check MUST be performed after SRV record
+        resolution, if applicable.
+
+Drivers MUST NOT send a PrincipalStepRequest when performing automatic authentication
+or when there is a cached IdPServerResponse .  Drivers must instead  use ``saslStart`` with a JwtStepRequest.
```
DmitryLukyanov, line 1276: nit: space between '.'

source/auth/auth.rst
DmitryLukyanov: Yeah, I also have a question about it. 
>attempt to authenticate via a JwtStepRequest before calling the callback

I guess this means that we need to do a server call under the lock which looks really wrong to me. Is it really meant here?

source/auth/auth.rst
blink1073: Removed

source/auth/auth.rst
blink1073: Added "a lock unique to the cache key"

source/auth/auth.rst
blink1073: Fixed

source/auth/tests/mongodb-oidc.rst
blink1073: The full cache is cleared during an error.  I updated the test.

source/auth/auth.rst
blink1073: This is an optimization that can be made later, it shouldn't hold up this PR.

addaleax (approved):
This is good to merge for me.

source/auth/auth.rst
```diff
@@ -1316,7 +1316,7 @@ Callbacks can be synchronous and/or asynchronous, depending on the driver
 and/or language.  Asynchronous callbacks should be preferred when other
 operations in the driver use asynchronous functions.
 
-Before calling a callback, the driver MUST acquire a lock unique to the cache key, and, if there is a non-expired cached Access Token, attempt to authenticate via a JwtStepRequest before calling the callback.  The lock is released after all attempts to authenticate have failed, or a non-recoverable error was encountered.  The driver MUST ensure that credentials have not changed between when the lock was requested and when it was acquired.
+Before calling a callback, the driver MUST acquire a lock unique to the cache key.  The driver MUST ensure that credentials have not changed between when the lock was requested and when it was acquired.
```
katcharov, line 1329: I believe this no longer specifies when the lock is released. If we specify when the lock should be acquired, should we not specify when it is released?

source/auth/tests/mongodb-oidc.rst
blink1073: I don't think we should hold up this PR for more prose tests.

source/auth/auth.rst
blink1073: Updated

source/auth/auth.rst
katcharov: Sounds good, created a ticket as you suggested here: https://jira.mongodb.org/browse/DRIVERS-2615

source/auth/tests/mongodb-oidc.rst
katcharov: Alright, added https://jira.mongodb.org/browse/DRIVERS-2616 .

We discussed, and only 391 errors should trigger a cache clear (the aforementioned division by zero should not). Seems fine to clarify this as part of 2616.

source/auth/auth.rst
```diff
@@ -1373,70 +1360,62 @@ ___
...
-arguments, and it will return a new OIDCRequestTokenResult response.
-
-In order to prevent a memory leak, the driver MUST loop over the existing cache
-at some interval, or per-authorization, to remove expired credentials from
-the cache.
+A cached Access Token will expire 5 minutes before the ``expiresInSeconds``
+time, if given.  If there is no ``expiresInSeconds``, the token must be considered expired as soon as the ensuing JwtStepRequest is started.  If a cached value is found but its
+Access Token is rejected by the server with a ``ReauthenticationRequired`` error, the Access Token must be marked expired and the Refresh callback MUST be called (if given) with the IdPServerInfo and Refresh Token, and it will return a new IdpServerResponse.  If the Refresh Callback fails, the error is raised to the user.  If the Refresh Callback succeeds, the new Access Token MUST be sent using a JwtStepRequest.  If the request fails with a ReauthenticationRequired error, the cache should be cleared, and a PrincipalStepRequest MUST be sent.  Next, the Request Callback should be called.  If the callback fails, the error is raised to the user.
```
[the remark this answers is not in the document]
katcharov: Moved to https://jira.mongodb.org/browse/DRIVERS-2616 .

katcharov (approved):
LGTM!

source/auth/tests/mongodb-oidc.rst
```diff
@@ -4,146 +4,433 @@ MongoDB OIDC
...
+- Ensure that a ``find`` operation replaces the one-time use entry and adds a new entry to the cache.
+- Close the client.
+
+Error clears cache
+~~~~~~~~~~~~~~~~~~
+- Clear the cache.
+- Create a new client with a valid request callback that gives credentials
+  that expire within 5 minutes and a refresh callback that gives invalid
+  credentials.
+- Ensure that a ``find`` operation adds a new entry to the cache.
```
durran, line 272: Client needs to be closed here and a new client with the same callbacks created. This is in line with the Python implementation of the prose tests.

durran (changes requested):

source/auth/tests/mongodb-oidc.rst
blink1073: Fixed

durran (approved):

source/auth/tests/mongodb-oidc.rst
```diff
@@ -4,146 +4,435 @@ MongoDB OIDC
...
+- Close the client.
+
+Lock Avoids Extra Callback Calls
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- Clear the cache.
+- Create a request callback that returns a token that will expire soon, and
+  a refresh callback.  Ensure that the request callback has a time delay, and
+  that we can record the number of times each callback is called.
+- Spawn two threads that do the following:
```
durran, line 98:
```suggestion
- Spawn two threads or async operations that do the following:
```

source/auth/tests/mongodb-oidc.rst
```diff
@@ -4,146 +4,435 @@ MongoDB OIDC
...
+  - Create a client with the callbacks.
+  - Run a find operation that succeeds.
+  - Close the client.
+  - Create a new client with the callbacks.
+  - Run a find operation that succeeds.
+  - Close the client.
+- Join the two threads.
```
durran, line 105:
```suggestion
- Join the two threads or simultaneously call the async operations
```

source/auth/tests/mongodb-oidc.rst
```diff
@@ -4,146 +4,435 @@ MongoDB OIDC
...
+- Ensure that the request callback has been called once, and the refresh
+  callback has been called twice.
```
durran, line 107:
```suggestion
- Ensure that the request callback has been called once, and the refresh
  callback has been called twice, or that no async function has been
  entered simultaneously.
```

durran (changes requested):

durran (approved):

source/auth/auth.rst
```diff
@@ -1390,76 +1361,62 @@ ___
...
 
-The cache keys MUST include the username (or empty string) and the
+If a global cache is used, the cache keys MUST include the username (or empty string) and the
 actually used socket address and port for the current server.  The cache key
-MUST also include a hashes of the callback function(s), if relevant and
-possible in the driver language.
-
-Using the socket address and port accounts for the case when two different
-servers use the same username but could be configurated differently.
+MUST also include a hashes of the callback function(s), if hash comparisons are possible in the driver language.  In the case of a global cache, Using the socket address and port accounts for the case when two different servers use the same username but could be configured differently.
```
JamesKovacs, line 1381: a hashes => hashes
Using => using

JamesKovacs (approved):
LGTM.

Noted some minor typos.

source/auth/auth.rst
blink1073: Fixed

PR #1383, opened 2023-03-02
fix(DRIVERS-2415): change device terminology to service provider

durran:
The device workflow terminology was a bit confusing since OIDC device flow generally refers to a user interaction on a device (such as a phone or computer) to perform an action to get a token. This changes to use `PROVIDER_NAME` instead of `DEVICE_NAME` as a mechanism property and changes language of device workflow to "as service to obtain credentials".

- [ ] Update changelog.

### DIFF
--- a/source/auth/auth.rst
+++ b/source/auth/auth.rst
@@ -1258,7 +1258,7 @@ Server will use principalName (n) if provided in clientStep1 to select an approp
 
 username
     MUST specified if more than one OIDC provider is configured and
-    DEVICE_NAME mechanism property is not specified.
+    SERVICE_NAME mechanism property is not specified.
 
 source
     MUST be "$external". Defaults to ``$external``.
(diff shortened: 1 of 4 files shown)


source/auth/auth.rst
```diff
@@ -1270,9 +1270,9 @@ mechanism
     MUST be "MONGODB-OIDC"
 
 mechanism_properties
-    DEVICE_NAME
-        Drivers MUST allow the user to specify a name for the device
-        workflow that is one of "aws", "azure", or "gcp".
+    SERVICE_NAME
+        Drivers MUST allow the user to specify a name for using a service
+        to obtain credentials that is one of "aws", "azure", or "gcp".
```
addaleax, line 1275: This has very different semantics than it does in the GSSAPI case – not to make your life harder, but maybe it’s a good idea to keep using a different name (suggestion off the top of my head: `PROVIDER_NAME`)?

durran: That was @blink1073 's original suggestion - I thought it was a bit too generic but since you do point out the Kerberos meaning I'm fine with changing it to PROVIDER_NAME. Thoughts @blink1073 ?

blink1073: `PROVIDER_NAME` works for me, it is generic enough for us to use a non-cloud provider in the future.

durran: I've updated to `PROVIDER_NAME`

addaleax (approved):

PR #1386, opened 2023-03-15
DRIVERS-2209 Collect FaaS information in handshake metadata

abr-egn:

- [*] Update changelog.
- [*] Test changes in at least one language driver.
- [*] Test these changes against all server versions and topologies (including standalone, replica set, sharded clusters, and serverless).

I've implemented this in a [branch](https://github.com/abr-egn/mongo-rust-driver/tree/RUST-1501/faas-poc) of the Rust driver, and tests pass across all configurations.

### DIFF
--- a/source/mongodb-handshake/handshake.rst
+++ b/source/mongodb-handshake/handshake.rst
@@ -540,6 +540,7 @@ command succeeds in the presence of the following sets of environment variables:
 +-----------------------+--------------------------+
 
 7. Invalid - wrong types
+
 +-------------------------------------+----------------------+
 | ``AWS_EXECUTION_ENV``               | ``AWS_Lambda_java8`` |
 +-------------------------------------+----------------------+
--- a/source/mongodb-handshake/handshake.rst
+++ b/source/mongodb-handshake/handshake.rst
@@ -355,8 +355,8 @@ be populated:
 Missing variables or variables with values not matching the expected type MUST cause the
 corresponding ``client.env`` field to be omitted and MUST NOT cause a user-visible error.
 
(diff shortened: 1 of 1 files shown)


source/mongodb-handshake/handshake.rst
```diff
@@ -418,11 +474,78 @@ the ``platform`` field before all others. Additionally, implementors are
 encouraged to place high priority information about the platform earlier in the
 string, in order to avoid possible truncating of those details.
 
+If the ``client.env`` field would be populated but would cause the metadata document
+to exceed the size limit, drivers MUST clear all fields of ``client.env`` except for
+``client.env.name``, and if that would still exceed the size limit drivers MUST omit
+``client.env`` entirely.
```
alcaeus, line 480: Regarding the last part:
> if that would still exceed the size limit drivers MUST omit `client.env` entirely.

Reading this along with the paragraph before I'm not entirely certain whether drivers should first truncate the `platform` field or omit `env`. This could be clarified here, as I believe it's better to truncate `platform` before removing `client.env.name`.
This would give us the following priority list (in descending order of importance):
* `application.name`
* `driver.*`
* `os.type`
* `env.name`
* `os.*` (except `type`)
* `env.*` (except `name`)
* `platform`

If the handshake metadata size exceeds 512 bytes, drivers can then use this list to decide what information to truncate to get to 512 bytes.

alcaeus (changes requested):

source/mongodb-handshake/handshake.rst
abr-egn: Ah, yup.  Updated the section to lay this out.

alcaeus (approved):

PR #1389, opened 2023-03-28
DRIVERS-2577: add a runCommand specification

nbbeeken:
[Link to github generated document](https://github.com/mongodb/specifications/blob/DRIVERS-2533-runCursorCommand/source/run-command/run-command.rst)

Process meta:
- [DRIVERS-2577](https://jira.mongodb.org/browse/DRIVERS-2577) is repurposed to both remove the RW concern options for the UTF and add the runCommand spec
- A following PR will use the driver epic number ([DRIVERS-2533](https://jira.mongodb.org/browse/DRIVERS-2533)) to track the new runCursorCommand spec

---------

Adds a specification for the existing runCommand API. Generally, it summarizes requirements across the specifications and links the reader to the more detailed specification on the subject. Adds language that encourages drivers to not inspect nor modify user input, drivers will have to assess their current implementation and determine how that guidance can best be
followed.

Test coverage is targeted at each subject in the spec inspecting command construction is performed as expected. For example, a new test asserts that a database configured with a RC will not attach read concern fields to the user's command document per the spec.

Removing Read Concern and Write Concern settings from the runCommand unified test operation because they intentionally are not supported and no tests have been written to use them.

- [x] Update changelog.
  - Node.js: https://github.com/mongodb/node-mongodb-native/pull/3643
  - Probably need to adjust runOnRequirements still


source/run-command/tests/README.rst
```diff
@@ -0,0 +1,19 @@
+=====================
+Retryable Reads Tests
+=====================
```
jmikola, line 3:
```suggestion
=================
Run Command Tests
=================
```

source/run-command/tests/unified/run-command.yml
```diff
@@ -0,0 +1,242 @@
...
+      client: stableApiClient
+      databaseName: stableApiDb
+
+initialData:
+- collectionName: collection
+  databaseName: plainDb
+  documents: [ { _id: 0 } ]
+
+tests:
+  - description: executes a ping with $db attached
```
dariakp, line 52: It would be better to be more specific with the test case names to communicate both what the expected behavior is and what condition prompts it, e.g., in this case, `always attaches the $db to the ping command` (although, should it be to "every" command?)

source/run-command/tests/unified/run-command.yml
```diff
@@ -0,0 +1,242 @@
...
+    expectEvents:
+      - client: client
+        events:
+        - commandStartedEvent:
+              command:
+                ping: 1
+                $db: plainDb
+              commandName: ping
+
+  - description: executes a ping with lsid attached
```
dariakp, line 71: -> "appends the provided session to \<type of\> command"

source/run-command/tests/unified/run-command.yml
```diff
@@ -0,0 +1,242 @@
...
+      - client: client
+        events:
+        - commandStartedEvent:
+              command:
+                ping: 1
+                lsid: { $$sessionLsid: session }
+                $db: plainDb
+              commandName: ping
+
+  - description: executes a ping with $readPreference attached
```
dariakp, line 92: -> "attaches the provided $readPreference ..."

source/run-command/tests/unified/run-command.yml
```diff
@@ -0,0 +1,242 @@
...
+      - client: client
+        events:
+        - commandStartedEvent:
+              command:
+                ping: 1
+                $readPreference: { mode: 'nearest' }
+                $db: plainDb
+              commandName: ping
+
+  - description: executes a ping without attaching readConcern
```
dariakp, line 113: -> "does not attach a readConcern specified at the db level"

source/run-command/tests/unified/run-command.yml
```diff
@@ -0,0 +1,242 @@
...
+      - client: client
+        events:
+        - commandStartedEvent:
+              command:
+                ping: 1
+                readConcern: { $$exists: false }
+                $db: dbWithRC
+              commandName: ping
+
+  - description: executes a ping without attaching writeConcern
```
dariakp, line 133: similar to readconcern case

source/run-command/tests/unified/run-command.yml
```diff
@@ -0,0 +1,242 @@
...
+      - client: client
+        events:
+        - commandStartedEvent:
+              command:
+                ping: 1
+                writeConcern: { $$exists: false }
+                $db: dbWithWC
+              commandName: ping
+
+  - description: executes a ping and does not retry retryable error
```
dariakp, line 153: -> "does not retry retryable errors on \<type of\> command"

source/run-command/tests/unified/run-command.yml
```diff
@@ -0,0 +1,242 @@
...
+      - name: runCommand
+        object: plainDb
+        arguments:
+          commandName: ping
+          command:
+            ping: 1
+        expectError:
+          isError: true
+
+  - description: executes a insert via runCommand inside a transaction
```
dariakp, line 174: "via runCommand" is redundant since this is the `runCommand` spec

source/run-command/tests/unified/run-command.yml
```diff
@@ -0,0 +1,242 @@
...
+                lsid: { $$sessionLsid: session }
+                txnNumber: 1
+                autocommit: false
+                # omitted fields
+                writeConcern: { $$exists: false }
+                readConcern: { $$exists: false }
+              commandName: commitTransaction
+              databaseName: admin
+
+  - description: executes a ping with stableApi fields attached
```
dariakp, line 222: this should be something like "attaches stableApi fields to type of command when configured on the client"

dariakp (changes requested):

source/run-command/tests/unified/run-command.yml
nbbeeken: > "always attaches $db to the given command"

Yes it will always be present on every command, even meta commands like hello target the `admin.$cmd` namespace.

source/run-command/run-command.rst
```diff
@@ -0,0 +1,180 @@
...
+The command sent MUST attach stable API fields as configured on the MongoClient.
+
+* See Stable API's section on `Generic Command Helper Behaviour <https://github.com/mongodb/specifications/blob/master/source/versioned-api/versioned-api.rst#generic-command-helper>`_
+
+Client Side Operations Timeout
+""""""""""""""""""""""""""""""
+
+A driver's runCommand API MUST provide an optional timeoutMS option to support client side operations timeout.
+Drivers MUST NOT attempt to check the command document for the presence of a ``maxTimeMS`` field.
+Drivers MUST document the behavior of runCommand if a ``maxTimeMS`` field  is already set on the command (ex. overwrites the field or not).
```
dariakp, line 172:
```suggestion
Drivers MUST document the behavior of runCommand if a ``maxTimeMS`` field  is already set on the command (such as overwriting the command field).
```

dariakp (changes requested):

source/run-command/tests/unified/run-command.yml
```diff
@@ -0,0 +1,242 @@
...
+      - client: client
+        events:
+        - commandStartedEvent:
+              command:
+                ping: 1
+                readConcern: { $$exists: false }
+                $db: dbWithRC
+              commandName: ping
+
+  - description:  does not inherit writeConcern specified at the db level
```
dariakp, line 133: looks like there's an extra space here

source/run-command/tests/unified/run-command.yml
```diff
@@ -0,0 +1,242 @@
...
+    expectEvents:
+      - client: client
+        events:
+        - commandStartedEvent:
+              command:
+                ping: 1
+                $db: plainDb
+              commandName: ping
+
+  - description: attaches the provided session lsid to ping command
```
dariakp, line 71: you say "given command" in the other test cases

dariakp (changes requested):

source/run-command/run-command.rst
```diff
@@ -0,0 +1,180 @@
...
+* Drivers MAY rename their API to match the language syntax.
+
+RunCommand implementation details
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+RunCommand provides a way to access MongoDB server commands directly without requiring a driver to implement a bespoke helper.
+The API is intended to take a document from a user, apply a number of common driver internal concerns before forwarding the command to a server.
+In general, a driver SHOULD avoid inspecting or modifying the user's command document in an effort to remain forward compatible with any potential server commands.
+
+Drivers that have historically modified user input SHOULD strive to instead clone so modifications do not impact reusing the same input command document.
```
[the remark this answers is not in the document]
nbbeeken: I think I was focusing on a use case, what matters is not to modify the user's input, I'm thinking that is a useful garuntee so that a command document could be reused by the user who made it.

source/run-command/run-command.rst
```diff
@@ -0,0 +1,180 @@
...
+* See Server Selection's section on `Use of read preferences with commands <https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#use-of-read-preferences-with-commands>`_
+
+If the provided Read Preference is NOT primary, the command sent MUST include the ``$readPreference`` global command argument.
+
+* See OP_MSG's section on `Global Command Arguments <https://github.com/mongodb/specifications/blob/master/source/message/OP_MSG.rst#global-command-arguments>`_
+
+Driver Sessions
+"""""""""""""""
+
+A driver's runCommand API MUST provide an optional session option to support explicit sessions and transactions.
```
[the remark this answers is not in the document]
nbbeeken: Good call, added a sentence

source/run-command/run-command.rst
```diff
@@ -0,0 +1,180 @@
...
+Every ClientSession has a corresponding Logical Session ID representing the server side session ID.
+The LSID MUST be included under ``lsid`` in the command sent to the server without modifying user input.
```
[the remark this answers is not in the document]
nbbeeken: fixed

source/run-command/run-command.rst
```diff
@@ -0,0 +1,180 @@
...
+Deviations
+^^^^^^^^^^
+
+* Drivers MAY rename their API to match the language syntax.
+
+RunCommand implementation details
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+RunCommand provides a way to access MongoDB server commands directly without requiring a driver to implement a bespoke helper.
+The API is intended to take a document from a user, apply a number of common driver internal concerns before forwarding the command to a server.
```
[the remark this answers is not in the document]
nbbeeken: fixed

source/run-command/run-command.rst
```diff
@@ -0,0 +1,180 @@
...
+       *
+       * @defaultValue primary
+       *
+       * @see https://www.mongodb.com/docs/manual/core/read-preference/
+       */
+      readPreference?: ReadPreference;
+
+      /**
+       * An optional explicit client session.
+       * The associated logical session id (`lsid`) will be applied to the command.
```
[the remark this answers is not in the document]
nbbeeken: Must, fixed

source/run-command/run-command.rst
```diff
@@ -0,0 +1,180 @@
...
+      /**
+       * Takes an argument representing an arbitrary BSON document and executes it against the server.
+       */
+      runCommand(command: BSONDocument, options: RunCommandOptions): BSONDocument;
+    }
+
+    interface RunCommandOptions {
+      /**
+       * An optional readPreference setting to apply to server selection logic.
+       * Also to apply to the command document as the $readPreference global command argument.
```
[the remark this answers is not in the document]
nbbeeken: Must, fixed

source/run-command/run-command.rst
```diff
@@ -0,0 +1,180 @@
...
+Every ClientSession has a corresponding Logical Session ID representing the server side session ID.
+The LSID MUST be included under ``lsid`` in the command sent to the server without modifying user input.
+
+* See Driver Sessions' section on `Sending the session ID to the server on all commands <https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst#sending-the-session-id-to-the-server-on-all-commands>`_
+
+Transactions
+""""""""""""
+
+If RunCommand is used within a Transaction the read preference MUST be sourced from the Transaction's options.
+The command sent to the server MUST include the transaction specific fields, the following is a summary:
```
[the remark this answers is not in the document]
nbbeeken: fixed

source/run-command/run-command.rst
```diff
@@ -0,0 +1,180 @@
...
+
+  * ``autocommit`` - The autocommit flag MUST be set to false.
+  * ``txnNumber`` - MUST be set.
+
+* If ``runCommand`` is the first operation of the transaction:
+
+  * ``startTransaction`` - MUST be set to true.
+  * ``readConcern`` - MUST be set to the Transaction's readConcern setting if it is NOT the default.
+
+* See Transaction's section on `Generic RunCommand helper within a transaction <https://github.com/mongodb/specifications/blob/DRIVERS-2533-runCursorCommand/source/transactions/transactions.rst#generic-runcommand-helper-within-a-transaction>`_
```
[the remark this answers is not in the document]
nbbeeken: fixed

source/run-command/run-command.rst
```diff
@@ -0,0 +1,180 @@
...
+
+ReadConcern and WriteConcern
+""""""""""""""""""""""""""""
+
+A RunCommand API MUST NOT support either Read Concern nor Write Concern options.
+Additionally, unless within a transaction the command sent MUST not have any Read Concern nor Write Concern fields applied that may be inherited from client, database, or collection options.
```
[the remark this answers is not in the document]
nbbeeken: fixed

source/run-command/run-command.rst
```diff
@@ -0,0 +1,180 @@
...
+
+  * ``startTransaction`` - MUST be set to true.
+  * ``readConcern`` - MUST be set to the Transaction's readConcern setting if it is NOT the default.
+
+* See Transaction's section on `Generic RunCommand helper within a transaction <https://github.com/mongodb/specifications/blob/DRIVERS-2533-runCursorCommand/source/transactions/transactions.rst#generic-runcommand-helper-within-a-transaction>`_
+
+ReadConcern and WriteConcern
+""""""""""""""""""""""""""""
+
+A RunCommand API MUST NOT support either Read Concern nor Write Concern options.
```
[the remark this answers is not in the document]
nbbeeken: fixed

source/run-command/run-command.rst
nbbeeken: fixed.

source/run-command/run-command.rst
```diff
@@ -0,0 +1,180 @@
...
+Additionally, unless within a transaction the command sent MUST not have any Read Concern nor Write Concern fields applied that may be inherited from client, database, or collection options.
+
+* See Read Concern's section on `Generic Command Method <https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#generic-command-method>`_
+* See Write Concern's section on `Generic Command Method <https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#generic-command-method-1>`_
+
+Retryability
+""""""""""""
+
+All commands passed to runCommand are considered non-retryable reads.
+Drivers MUST NOT inspect the command to determine if it is a write and MUST NOT attach a transaction ID.
```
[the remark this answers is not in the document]
nbbeeken: addressed

source/run-command/run-command.rst
```diff
@@ -0,0 +1,184 @@
...
+      /**
+       * Takes an argument representing an arbitrary BSON document and executes it against the server.
+       */
+      runCommand(command: BSONDocument, options: RunCommandOptions): BSONDocument;
+    }
+
+    interface RunCommandOptions {
+      /**
+       * An optional readPreference setting to apply to server selection logic.
+       * Also the value that MUST be applied to the command document as the $readPreference global command argument if not set to primary.
```
dariakp, line 60:
```suggestion
       * This value MUST be applied to the command document as the $readPreference global command argument if not set to primary.
```

source/run-command/tests/unified/run-command.yml
nbbeeken: fixed.

source/run-command/run-command.rst
```diff
@@ -0,0 +1,184 @@
...
+* Drivers MAY rename their API to match the language syntax.
+
+RunCommand implementation details
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+RunCommand provides a way to access MongoDB server commands directly without requiring a driver to implement a bespoke helper.
+The API is intended to take a document from a user and apply a number of common driver internal concerns before forwarding the command to a server.
+In general, a driver SHOULD avoid inspecting or modifying the user's command document in an effort to remain forward compatible with any potential server commands.
+
+Drivers that have historically modified user input SHOULD strive to instead clone the input such that appended fields do not effect the user's input.
```
dariakp, line 96:
```suggestion
Drivers that have historically modified user input SHOULD strive to instead clone the input such that appended fields do not affect the user's input.
```

source/run-command/tests/unified/run-command.yml
nbbeeken: fixed

dariakp (changes requested):

source/run-command/run-command.rst
nbbeeken: fixed

source/run-command/run-command.rst
nbbeeken: fixed

dariakp (approved):

source/run-command/run-command.rst
```diff
@@ -0,0 +1,185 @@
...
+       */
+      runCommand(command: BSONDocument, options: RunCommandOptions): BSONDocument;
+    }
+
+    interface RunCommandOptions {
+      /**
+       * An optional readPreference setting to apply to server selection logic.
+       * This value MUST be applied to the command document as the $readPreference global command argument if not set to primary.
+       *
+       * @defaultValue primary
```
jmikola, line 62:
```suggestion
       * @defaultValue ReadPreference(mode: primary)
```

This may be pedantic, but this is just the mode option. I think expressing the entire [ReadPreference structure](https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#read-preference) would be more clear.

I don't recall offhand if the `$readPrefrence` [global command argument]9https://github.com/mongodb/specifications/blob/master/source/message/OP_MSG.rst#command-arguments-as-payload) takes all of the same options as the structure defined in the Server Selection spec, but the example does show it expressed as an object with a `mode` property, so I think that would still better match the suggested replacement here.

source/run-command/run-command.rst
```diff
@@ -0,0 +1,185 @@
...
+       *
+       * @see https://www.mongodb.com/docs/manual/core/read-preference/
```
jmikola, line 64: Is https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#read-preference a better reference link? I think our own specs are more verbose than the server manual.

source/run-command/run-command.rst
```diff
@@ -0,0 +1,185 @@
...
+       */
+      readPreference?: ReadPreference;
+
+      /**
+       * An optional explicit client session.
+       * The associated logical session id (`lsid`) MUST be applied to the command.
+       *
+       * @see https://www.mongodb.com/docs/manual/core/read-isolation-consistency-recency/#std-label-sessions
```
jmikola, line 72: Likewise, should this refer to our own session spec?

source/run-command/run-command.rst
```diff
@@ -0,0 +1,185 @@
...
+       *
+       * @ see https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.rst
+       */
+      timeoutMS?: number;
+    }
+
+Deviations
+^^^^^^^^^^
+
+* Drivers MAY rename their API to match the language syntax.
```
jmikola, line 88: You can optionally refer to the deviations guidelines in the CRUD spec, since that's much more fleshed out. Feel free to disregard, though.

Separate note: "Deviations" is typically a top-level field for the spec itself, so I think this can be moved up and out of the API section.

source/run-command/run-command.rst
```diff
@@ -0,0 +1,185 @@
...
+
+The ``$db`` global command argument MUST be sent on the command sent to the server and it MUST correspond to the database name the runCommand API was invoked on.
+
+* See OP_MSG's section on `Global Command Arguments <https://github.com/mongodb/specifications/blob/master/source/message/OP_MSG.rst#global-command-arguments>`_
+
+ReadPreference
+""""""""""""""
+
+A driver's runCommand implementation MUST assume all commands are read operations.
+To facilitate selecting the server the command should be sent to a driver's runCommand MUST accept an optional Read Preference option.
```
jmikola, line 110:
```suggestion
To facilitate server selection the runCommand operation MUST accept an optional Read Preference option.
```

I'm not sure if this suggestion conveys the original intention, but the existing text definitely needs some correction.

source/run-command/run-command.rst
```diff
@@ -0,0 +1,185 @@
...
+
+* See Server Selection's section on `Use of read preferences with commands <https://github.com/mongodb/specifications/blob/master/source/server-selection/server-selection.rst#use-of-read-preferences-with-commands>`_
+
+If the provided Read Preference is NOT primary, the command sent MUST include the ``$readPreference`` global command argument.
```
jmikola, line 114:
```suggestion
If the provided ReadPreference is NOT ``{mode: primary}``, the command sent MUST include the ``$readPreference`` global command argument.
```

Small changes. Referring to the type itself here, and also denoting the entire structure (related to my earlier suggestion regarding default value).

source/run-command/run-command.rst
```diff
@@ -0,0 +1,185 @@
...
+
+* See OP_MSG's section on `Global Command Arguments <https://github.com/mongodb/specifications/blob/master/source/message/OP_MSG.rst#global-command-arguments>`_
+
+Driver Sessions
+"""""""""""""""
+
+A driver's runCommand API MUST provide an optional session option to support explicit sessions and transactions.
+If a session is not provided the driver MUST attach an implicit session if the selected server supports sessions.
```
jmikola, line 122: Is this still relevant if drivers now assume all servers support sessions? IIRC, min supported server version is now 3.6. I think this came up in another spec, but I could be mistaken.

source/run-command/run-command.rst
```diff
@@ -0,0 +1,185 @@
...
+
+Every ClientSession has a corresponding Logical Session ID representing the server side session ID.
```
jmikola, line 124:
```suggestion
Every ClientSession has a corresponding Logical Session ID representing the server-side session ID.
```

source/run-command/run-command.rst
```diff
@@ -0,0 +1,185 @@
...
+The Logical Session ID MUST be included under ``lsid`` in the command sent to the server without modifying user input.
+
+* See Driver Sessions' section on `Sending the session ID to the server on all commands <https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.rst#sending-the-session-id-to-the-server-on-all-commands>`_
+
+Transactions
+""""""""""""
+
+If RunCommand is used within a Transaction the read preference MUST be sourced from the Transaction's options.
```
jmikola, line 132:
```suggestion
If runCommand is used within a transaction the read preference MUST be sourced from the transaction's options.
```

"runCommand" seemed consistent with prior text in the spec. If you prefer "RunCommand", I'd suggest going through and changing everything to a single form.

Transaction isn't a proper noun, so we can lowercase that.

"read preference" is fine as-is since you're not referring to the special type.

Alternatively, "transaction's options" can refer to the special type: [TransactionOptions](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#transactionoptions).

source/run-command/run-command.rst
```diff
@@ -0,0 +1,185 @@
...
+
+* If ``runCommand`` is executing within a transaction:
+
+  * ``autocommit`` - The autocommit flag MUST be set to false.
+  * ``txnNumber`` - MUST be set.
+
+* If ``runCommand`` is the first operation of the transaction:
+
+  * ``startTransaction`` - MUST be set to true.
+  * ``readConcern`` - MUST be set to the Transaction's readConcern setting if it is NOT the default.
```
jmikola, line 143:
```suggestion
  * ``readConcern`` - MUST be set to the transaction's read concern if it is NOT the default.
```

source/run-command/run-command.rst
```diff
@@ -0,0 +1,185 @@
...
+
+* See Transactions' section on `Generic RunCommand helper within a transaction <https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#generic-runcommand-helper-within-a-transaction>`_
```
jmikola, line 145:
```suggestion
* See `Generic RunCommand helper within a transaction <https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#generic-runcommand-helper-within-a-transaction>`_ in the Transactions specification.
```

I assume this is what you meant, as there's no "Transactions" section under that heading.

source/run-command/run-command.rst
```diff
@@ -0,0 +1,185 @@
...
+
+ReadConcern and WriteConcern
+""""""""""""""""""""""""""""
+
+A RunCommand API MUST NOT support Read Concern and Write Concern options.
+Additionally, unless within a transaction, the command sent MUST NOT have any Read Concern or Write Concern fields applied that may be inherited from client, database, or collection options.
```
jmikola, line 151:
```suggestion
RunCommand MUST NOT support read concern and write concern options.

Additionally, unless executing within a transaction, the RunCommand operation MUST NOT set the ``readConcern`` or ``writeConcern`` fields in the command document. For example, default values MUST NOT be inherited from client, database, or collection options.

If the user-provided command document already includes ``readConcern`` or ``writeConcern`` fields, the values MUST be left as-is.
```

Added the last paragraph to be clear. I don't think there's anything stopping a user from supplying their own options.

Change "RunCommand" casing for consistency if needed.

source/run-command/run-command.rst
```diff
@@ -0,0 +1,185 @@
...
+""""""
+
+The ``$db`` global command argument MUST be sent on the command sent to the server and it MUST correspond to the database name the runCommand API was invoked on.
+
+* See OP_MSG's section on `Global Command Arguments <https://github.com/mongodb/specifications/blob/master/source/message/OP_MSG.rst#global-command-arguments>`_
+
+ReadPreference
+""""""""""""""
+
+A driver's runCommand implementation MUST assume all commands are read operations.
```
jmikola, line 109: Is "MUST assume all commands are read operations" just so we don't have a logical conflict with the operation accepting a ReadPreference option?

This makes it sound as if a user can't use this to execute write commands, although I see no reason to prohibit that (they'll just have to make sure a primary is selected).

source/run-command/run-command.rst
```diff
@@ -0,0 +1,185 @@
...
+A RunCommand API MUST NOT support Read Concern and Write Concern options.
+Additionally, unless within a transaction, the command sent MUST NOT have any Read Concern or Write Concern fields applied that may be inherited from client, database, or collection options.
+
+* See Read Concern's section on `Generic Command Method <https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#generic-command-method>`_
+* See Write Concern's section on `Generic Command Method <https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.rst#generic-command-method-1>`_
+
+Retryability
+""""""""""""
+
+All commands passed to runCommand are considered non-retryable reads.
```
jmikola, line 159: Can we just say "non-retryable" here? I don't think we need to enforce that runCommand is only for reads, if the important factor is saying it's _treated_ like a read operation so we can solicit an optional read preference (and doesn't get default routing to a primary like explicit write operations).

source/run-command/run-command.rst
```diff
@@ -0,0 +1,185 @@
...
+""""""""""
+
+The command sent MUST attach stable API fields as configured on the MongoClient.
+
+* See Stable API's section on `Generic Command Helper Behaviour <https://github.com/mongodb/specifications/blob/master/source/versioned-api/versioned-api.rst#generic-command-helper>`_
+
+Client Side Operations Timeout
+""""""""""""""""""""""""""""""
+
+A driver's runCommand API MUST provide an optional timeoutMS option to support client side operations timeout.
```
jmikola, line 175:
```suggestion
A driver's runCommand API MUST provide an optional ``timeoutMS`` option to support client side operations timeout.
```

If we're referring to actual field names, I think using code formatting is good for consistency.

source/run-command/run-command.rst
```diff
@@ -0,0 +1,185 @@
...
+Drivers MUST NOT attempt to check the command document for the presence of a ``maxTimeMS`` field.
```
jmikola, line 176: This language might be something we can use earlier when talking about not modifying existing ``readConcern`` or ``writeConcern`` fields on the command document.

Likewise, should we add similar text for the section talking about ``lsid`` setting? IIRC, drivers are supposed to append the field regardless of whether it exists on the user-supplied document or not. And depending on the driver and BSON representation I think that could result in duplicate fields or overwriting the existing value -- which we don't really care about since the user is probably doing something they shouldn't.

jmikola (commented):
Posting feedback for the spec document only. Will come back to the tests in a subsequent review.

source/run-command/tests/unified/run-command.yml
```diff
@@ -0,0 +1,242 @@
+description: runCommand spec requirements
+
+schemaVersion: "1.5"
+
+createEntities:
```
durran, line 5: General comment for the YAML test here. I *think* we need to continue following the pattern established in the other tests in both aliasing and naming conventions. The first I *believe* is going to be needed for drivers that consume the YAML directly and do not use the generated JSON. (I would like confirmation here from those drivers - Ruby I think of first - @DmitryLukyanov can you confirm)? The second is just for general consistency - the other specs follow the "entityName0, entityName1" pattern and I think it's just better to stay with that for now rather than have a unified test that breaks the pattern. If that pattern suits us is another discussion.

durran (changes requested):

source/run-command/tests/unified/run-command.yml
jyemin: I think you meant to tag @comandeo for Ruby

source/run-command/run-command.rst
nbbeeken: The node driver, at least, will send mode, tags, maxStalenessSeconds, and hedge

source/run-command/run-command.rst
nbbeeken: I think two other components of this assumption is for performing retryability, and applying writeConcern settings both of which are respectively prohibited. So it is perhaps redundant to "assume commands are reads". For the sake of this section maybe it is important to include mention of server selection:

> For the purposes of server selection a RunCommand MUST assume all commands are read operations.

source/run-command/run-command.rst
nbbeeken: Mongocryptd does not support sessions, the recent change to the spec related to this was that drivers no longer perform server selection to determine session support, but support for sessions is still checked before attaching an lsid

source/run-command/run-command.rst
nbbeeken: Yea, this keeps it succinct

source/run-command/run-command.rst
nbbeeken: Added notes to session, RW concern. And improved the detail in the implementation details section about modification to highlight append vs overwrite

source/run-command/run-command.rst
```diff
@@ -0,0 +1,185 @@
...
+* Drivers MAY rename their API to match the language syntax.
+
+RunCommand implementation details
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+RunCommand provides a way to access MongoDB server commands directly without requiring a driver to implement a bespoke helper.
+The API is intended to take a document from a user and apply a number of common driver internal concerns before forwarding the command to a server.
+In general, a driver SHOULD avoid inspecting or modifying the user's command document in an effort to remain forward compatible with any potential server commands.
+
+Drivers that have historically modified user input SHOULD strive to instead clone the input such that appended fields do not affect the user's input.
```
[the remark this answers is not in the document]
nbbeeken: Updated the language here to be more strict and discuss a bit of the impacts BSON implementation has here.

source/run-command/run-command.rst
```diff
@@ -0,0 +1,185 @@
...
+       *
+       * @defaultValue primary
+       *
+       * @see https://www.mongodb.com/docs/manual/core/read-preference/
+       */
+      readPreference?: ReadPreference;
+
+      /**
+       * An optional explicit client session.
+       * The associated logical session id (`lsid`) MUST be applied to the command.
```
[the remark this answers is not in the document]
nbbeeken: Updated to mention 'driver'

source/run-command/tests/unified/run-command.yml
nbbeeken: Put back the use of anchors, I would like to keep the human-ish names coll0/1 makes sense when you just need a collection but the three dbs I have here are declared differently to show the differences in behavior in each test, personally I find it easier to follow/write, but I don't have a strong attachment either.

source/run-command/run-command.rst
nbbeeken: Moved deviations up, and linked to the CRUD spec I think it has more detail worth referring to

source/run-command/run-command.rst
nbbeeken: updated!

source/run-command/run-command.rst
nbbeeken: updated!

durran (approved):

source/run-command/tests/unified/run-command.yml
```diff
@@ -0,0 +1,242 @@
...
+          commandName: ping
+          command:
+            ping: 1
+        expectError:
+          isError: true
+
+  - description: attaches transaction fields to given command
+    runOnRequirements:
+      - topologies: [ replicaset, sharded-replicaset, load-balanced, sharded ]
+        minServerVersion: '4.4'
```
jmikola, line 177:
```suggestion
      - minServerVersion: "4.0"
        topologies: [ replicaset ]
      - minServerVersion: "4.2"
        topologies: [ sharded-replicaset, load-balanced ]
```

Not sure why `minServerVersion: 4.4` was used, but other transaction tests allow 4.0+ replica sets and 4.2+ sharded clusters.

Specifying `sharded-replicaset` and `sharded` together also seemed redundant.

I don't think `sharded` should be used for any spec tests that require successfully executing a transaction since an oplog is needed. It looks like there are a few instances of tests requiring `sharded` instead of `sharded-replicaset`. I'll open a separate JIRA ticket about that.

source/run-command/tests/unified/run-command.yml
```diff
@@ -0,0 +1,242 @@
...
+        strict: true
+  - database:
+      id: &stableApiDb stableApiDb
+      client: *stableApiClient
+      databaseName: *stableApiDb
+
+initialData:
+- collectionName: *collection
+  databaseName: *plainDb
+  documents: [ { _id: 0 } ]
```
jmikola, line 49: Do any of the tests actually need this document to exist?

If you only need the collection to exist for the transaction test you can use `documents: []` here.

source/run-command/tests/unified/run-command.yml
```diff
@@ -0,0 +1,242 @@
...
+        arguments:
+          commandName: ping
+          command:
+            ping: 1
+        expectResult:
+          ok: 1
+    expectEvents:
+      - client: *client
+        events:
+        - commandStartedEvent:
```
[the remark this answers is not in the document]
nbbeeken: Either level of indentation seems to generate the same JSON 🤔 updated to be consistent though

source/run-command/tests/unified/run-command.yml
```diff
@@ -0,0 +1,242 @@
...
+
+  - description: attaches the provided $readPreference to given command
+    operations:
+      - name: runCommand
+        object: *plainDb
+        arguments:
+          commandName: ping
+          command:
+            ping: 1
+          readPreference: 'nearest'
```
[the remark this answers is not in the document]
nbbeeken: Yes it does, it will invoke our runCommand like so:
```ts
await db.commmand({ ping: 1 }, { readPreference: 'nearest' })
```
Nesting `'nearest'` inside of a `mode` property also works.

source/run-command/tests/unified/run-command.yml
```diff
@@ -0,0 +1,242 @@
...
+                $readPreference: { mode: 'nearest' }
+                $db: *plainDb
+              commandName: ping
+
+  - description: does not inherit readConcern specified at the db level
+    operations:
+      - name: runCommand
+        object: *dbWithRC
+        arguments:
+          commandName: ping
```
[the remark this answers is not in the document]
nbbeeken: Good call, went with an agg

source/run-command/tests/unified/run-command.yml
```diff
@@ -0,0 +1,242 @@
...
+                readConcern: { $$exists: false }
+                $db: *dbWithRC
+              commandName: ping
+
+  - description: does not inherit writeConcern specified at the db level
+    operations:
+      - name: runCommand
+        object: *dbWithWC
+        arguments:
+          commandName: ping
```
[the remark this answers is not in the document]
nbbeeken: Changed this to an insert

source/unified-test-format/unified-test-format.rst
```diff
@@ -3898,6 +3893,7 @@ Changelog
 
 ..
   Please note schema version bumps in changelog entries where applicable.
+:2023-04-13: Remove readConcern and writeConcern settings from runCommand operation
```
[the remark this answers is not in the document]
nbbeeken: fixed

source/run-command/run-command.rst
```diff
@@ -0,0 +1,194 @@
...
+Drivers MUST NOT attempt to check the command document for the presence of a ``maxTimeMS`` field.
+Drivers MUST document the behavior of RunCommand if a ``maxTimeMS`` field  is already set on the command (such as overwriting the command field).
+
+* See Client Side Operations Timeout's section on `runCommand <https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.rst#runcommand>`_
+* See Client Side Operations Timeout's section on `runCommand behavior <https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.rst#runcommand-behavior>`_
+
+Changelog
+=========
+
+:2023-03-28: Add run command specification.
```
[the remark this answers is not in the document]
nbbeeken: bumped this, and I'll keep in mind to update it before we merge

source/run-command/tests/unified/run-command.yml
```diff
@@ -0,0 +1,242 @@
...
+              commandName: ping
+
+  - description: does not retry retryable errors on given command
+    operations:
+      - name: failPoint
+        object: testRunner
+        arguments:
+          client: *client
+          failPoint:
+            configureFailPoint: failCommand
```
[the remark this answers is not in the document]
jmikola: Oy, typo on my part. Yes, I meant `useMultipleMongoses: false`.

source/run-command/tests/unified/run-command.yml
jmikola: Just to close the loop here, I created [DRIVERS-2609](https://jira.mongodb.org/browse/DRIVERS-2609) to clear up other spec tests.

jmikola:
FWIW, I also POC-ed the tests in PHPLIB: https://github.com/mongodb/mongo-php-library/pull/1066

jmikola (approved):
I also POC-ed the tests in PHPLIB: https://github.com/mongodb/mongo-php-library/pull/1066

PR #1390, opened 2023-03-29
DRIVERS-2056: mandate that drivers error on invalid connection string options

baileympearson:
Node POC: https://github.com/mongodb/node-mongodb-native/pull/3611

With the changes in this PR, drivers must error when parsing the connection string if:

- an invalid option key is encountered
- an invalid option value is encountered

- [x] Update changelog.

### DIFF
--- a/source/connection-string/connection-string-spec.rst
+++ b/source/connection-string/connection-string-spec.rst
@@ -140,13 +140,14 @@ A key value pair represents the option key and its associated value. The key is 
 - Key:
    The connection option's key string.  Keys should be normalised and
    character case should be ignored.
-- Value: (optional)
-   The value if provided otherwise it defaults to an empty string.
+- Value:
+   The value associated with a key.  A value MUST be provided for every key.  
 
 ---------------------------
 Defining connection options
 ---------------------------
-Connection option key values MUST be defined in the relevant specification that describes the usage of the key and value.  The value data type MUST also be defined there. The value's default value SHOULD also be defined if it is relevant.
+Connection option key values MUST be defined in the relevant specification that describes the usage of the key and value.  The value data type MUST also be defined there. 
+The value's default value SHOULD also be defined if it is relevant.
 
 Keys
 ----
@@ -154,9 +155,7 @@ Keys are strings and the character case must be normalized by lower casing the u
 
 When defining and documenting keys, specifications should follow the camelCase naming convention with the first letter in lowercase, snake\_case MUST not be used. Keys that aren't supported by a driver MUST be ignored.
 
-Keys that aren't supported by a driver MUST be ignored. A WARN level logging message MUST be issued for unsupported keys. For example::
-
-  Unsupported option 'connectMS'.
+Keys that aren't supported by a driver MUST NOT be ignored. A error MUST be raised for unsupported keys. 
   
 Keys should be descriptive and follow existing conventions:
 
@@ -173,9 +172,9 @@ The values in connection options MUST be URL decoded by the parser. The values c
 - Strings:
     The value
 - Integer:
-  The value parsed as a integer. If the value is the empty string, the key MUST be ignored.
+  The value parsed as a integer. If the value is the empty string, the key MUST NOT be ignored and an error MUST be reported.
 - Boolean:
-  "true" and "false" strings MUST be supported. If the value is the empty string, the key MUST be ignored.
+  "true" and "false" strings MUST be supported. If the value is the empty string, the key MUST NOT be ignored and an error MUST be reported.
 
   - For legacy reasons it is RECOMMENDED that alternative values for true and false be supported:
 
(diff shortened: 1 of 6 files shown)


source/connection-string/connection-string-spec.rst
```diff
@@ -237,7 +240,7 @@ Connection strings are a mechanism to configure a MongoClient outside the user's
 
 Keys MUST follow existing connection option naming conventions as defined above. Values MUST also follow the existing, specific data types.
 
-Any options that are not supported MUST raise a WARN log level as described in the keys section.
+Any options that are not supported MUST raise a error as described in the keys section.
```
alcaeus, line 243:
```suggestion
Any options that are not supported MUST raise an error as described in the keys section.
```

source/connection-string/connection-string-spec.rst
```diff
@@ -323,9 +326,6 @@ Q: Why is it recommended that Connection Options take precedence over applicatio
   2. MongoClient hosts and options
   3. Connection String hosts and options
 
-Q: Why WARN level warning on unknown options rather than throwing an exception?
- It is responsible to inform users of possible misconfigurations and both methods achieve that.  However, there are conflicting requirements of a  Connection String.  One goal is that any given driver should be configurable by a connection string but different drivers and languages have different feature sets.  Another goal is that Connection Strings should be portable and as such some options supported by language X might not be relevant to language Y. Any given driver does not know is an option is specific to a different driver or is misspelled or just not supported.  So the only way to stay portable and support configuration of all options is to not throw an exception but rather log a warning.
-
 Q: How long should deprecation options be supported?
```
alcaeus, line 329: Just noticed this typo, I think we can update that at the same time.

```suggestion
Q: How long should deprecated options be supported?
```

source/connection-string/tests/invalid-options.yml
```diff
@@ -0,0 +1,126 @@
...
+        valid: false
+        warning: false
+        hosts:
+            -
+                type: "hostname"
+                host: "example.com"
+                port: ~
+        auth: ~
+        options:
+            replicaset: "test"
```
alcaeus, line 52: Are the options even relevant when we expect an error? If so, this should never be `test` as the URI string only specifies `test1` and `test2`.

source/connection-string/tests/invalid-options.yml
```diff
@@ -0,0 +1,126 @@
...
```
alcaeus, line 65: Should be `null` as no `replicaSet` option is specified in the URI.

```suggestion
        options: ~
```

source/connection-string/tests/invalid-options.yml
```diff
@@ -0,0 +1,126 @@
...
```
alcaeus, line 78: See above.

```suggestion
        options: ~
```

source/connection-string/tests/invalid-options.yml
```diff
@@ -0,0 +1,126 @@
...
+        hosts:
+            -
+                type: "hostname"
+                host: "localhost"
+                port: ~
+        auth: ~
+        options: ~
+    -
+        description: "Error on empty non-integer value for integer option"
+        uri: "mongodb://localhost/?maxIdleTimeMS="
```
alcaeus, line 93: This is the same URI as above. Also, what's an "empty non-integer value"?

source/uri-options/uri-options.rst
```diff
@@ -50,7 +50,7 @@ occur:
    ``tlsDisableCertificateRevocationCheck`` appear in the URI options.
 7. Both ``tlsDisableOCSPEndpointCheck`` and
    ``tlsDisableCertificateRevocationCheck`` appear in the URI options.
-8. All instances of ``tls`` and ``ssl`` in the URI options do not have the
+8. Both ``tls`` and ``ssl`` appear in the URI options and they do not have the
```
alcaeus, line 53: Can't comment on contextual code in a PR review, but the paragraph for this section states:
> Per the Connection String spec, the behavior of duplicates of most URI options is undefined.

This is no longer true, and should be removed entirely, leaving the paragraph as:
> Due to the security implications of certain options, drivers MUST raise an error to the user during parsing if any of the following circumstances occur:

alcaeus (changes requested):

source/connection-string/tests/invalid-options.yml
baileympearson: I've clarified the wording to specify a "non-empty invalid value" instead

source/connection-string/tests/invalid-options.yml
baileympearson: The options are never relevant if we're expecting errors.  I cleaned up the tests so there are never any expected options, hosts or auth.

baileympearson:
Moving back into draft for now.  @alexbevi is going write a PD for this before we implement the spec changes.

PR #1394, opened 2023-04-03
DRIVERS-2209 Simplify metadata truncation rules

abr-egn:
DRIVERS-2209

The preservation order introduced in #1386 turned out to be both confusing and mildly broken - the server has constraints about the `driver` and `application` documents it wasn't respecting.  This PR simplifies the spec to only resolve the ambiguity that would otherwise have been introduced between `platform` and `env`; a larger scope set of rules, if needed, should be its own spec update.

### DIFF
--- a/source/mongodb-handshake/handshake.rst
+++ b/source/mongodb-handshake/handshake.rst
@@ -474,8 +474,8 @@ the following order until the document is under the size limit:
 
 1. Omit fields from ``env`` except ``env.name``.
 2. Omit fields from ``os`` except ``os.type``.
-2. Omit the ``env`` document entirely.
-3. Truncate ``platform``.
+3. Omit the ``env`` document entirely.
+4. Truncate ``platform``.
 
 Additionally, implementors are encouraged to place high priority information about the
 platform earlier in the string, in order to avoid possible truncating of those details.
--- a/source/mongodb-handshake/handshake.rst
+++ b/source/mongodb-handshake/handshake.rst
@@ -469,16 +469,12 @@ The entire metadata BSON document MUST NOT exceed 512 bytes. This includes all
 BSON overhead.  The ``client.application.name`` cannot exceed 128 bytes.  MongoDB
 will return an error if these limits are not adhered to, which will result in
 handshake failure. Drivers MUST validate these values and truncate or omit driver
-provided values if necessary.  Implementors SHOULD prioritize fields to preserve in
-this order:
+provided values if necessary.  Implementors SHOULD cumulatively update fields in
+the following order until the document is under the size limit:
 
-1. ``application.name``
-2. ``driver.*``
-3. ``os.type``
-4. ``env.name``
-5. ``os.*`` (except ``type``)
-6. ``env.*`` (except ``name``)
-7. ``platform``
+1. Truncate ``platform``.
+2. Omit fields from ``env`` except ``env.name``.
+3. Omit the ``env`` document entirely.
 
 Additionally, implementors are encouraged to place high priority information about the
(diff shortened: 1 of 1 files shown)


source/mongodb-handshake/handshake.rst
```diff
@@ -469,16 +469,12 @@ The entire metadata BSON document MUST NOT exceed 512 bytes. This includes all
...
-1. ``application.name``
-2. ``driver.*``
-3. ``os.type``
-4. ``env.name``
-5. ``os.*`` (except ``type``)
-6. ``env.*`` (except ``name``)
-7. ``platform``
+1. Truncate ``platform``.
+2. Omit fields from ``env`` except ``env.name``.
+3. Omit the ``env`` document entirely.
```
alexbevi, line 478:
```
/* REQUIRED */
        os: {
            type: "<string>",         /* REQUIRED */
            name: "<string>",         /* OPTIONAL */
            architecture: "<string>", /* OPTIONAL */
            version: "<string>"       /* OPTIONAL */
        },
```
Since the spec indicates the `os` field only requires `os.type` would it make sense to also include language such as: _Omit fields from `os` except `os.type`._?

alexbevi: Also due to the need for data in the `platform` field (for Node.js, and likely Java) I'd advocate for reorganizing this to be:

1. Omit fields from ``env`` except ``env.name``.
2. Omit fields from os except os.type
3. Omit the ``env`` document entirely.
4. Truncate platform

prestonvasquez:
I think it would help with readability to update the line "The entire metadata BSON document MUST NOT exceed 512 bytes":

> The entire client metadata BSON document MUST NOT exceed 512 bytes.

source/mongodb-handshake/handshake.rst
abr-egn: Sounds reasonable - I've updated to this.

abr-egn:
> It may help readability to update "The entire metadata BSON document MUST NOT exceed 512 bytes" to:
> 
> > The entire client metadata BSON document MUST NOT exceed 512 bytes.

Good thought, I've included this as well.

source/mongodb-handshake/handshake.rst
```diff
@@ -465,20 +465,17 @@ acceptable deviations are as follows:
...
-2. ``driver.*``
-3. ``os.type``
-4. ``env.name``
-5. ``os.*`` (except ``type``)
-6. ``env.*`` (except ``name``)
-7. ``platform``
+1. Omit fields from ``env`` except ``env.name``.
+2. Omit fields from ``os`` except ``os.type``.
+2. Omit the ``env`` document entirely.
+3. Truncate ``platform``.
```
alexbevi, line 478: Minor note: 2. 3. above should be 3. 4.

abr-egn: Whoops, fixed.

source/mongodb-handshake/handshake.rst
prestonvasquez: What happens if the client metadata exceeds the 512 byte limit with os.type but is under the limit without it? IIUC this is technically possible since the application name is configurable.

abr-egn: That's orthogonal to this PR, and is covered under the more general
> Drivers MUST validate these values and truncate or omit driver provided values if necessary.
The list here is giving relative priorities of some fields but it's neither mandatory nor exhaustive.

prestonvasquez (approved):

alexbevi (approved):
LGTM

alcaeus (approved):

PR #1396, opened 2023-04-05
DRIVERS-2524 do not create or drop `eccCollection`

kevinAlbs:
# Summary
- Do not create or drop `eccCollection`.
- Require `escCollection` and `ecocCollection` not be documented.
- Update spec tests to assert `eccCollection` is not created.
- Return error if attempting to create a QE collection with server version < 7.0.0 (maxWireVersion < 21)

## Additional Improvements
- Use YAML anchors for repeated collection names and `encryptedFields`.
- Remove no longer applicable test: `encryptedFieldsMap with cyclic entries does not loop`
- Add test `encryptedFields are consulted for metadata collection names`

Changes have been tested in the C driver here: https://github.com/mongodb/mongo-c-driver/pull/1232

Tests require libmongocrypt 1.8.0-alpha0 or newer. Binaries for 1.8.0-alpha0 are available here: https://spruce.mongodb.com/task/libmongocrypt_publish_upload_all_042603b1d72f49d9034e7059b82a65e843c7e38a_23_03_29_15_26_54/logs?execution=0

# Background & Motivation

## Terminology
Queryable Encryption is also referred to as QE or FLE2.
The versions of the QE protocols are noted as QEv1 and QEv2.
QEv1 was introduced in 6.0.0. QEv2 is introduced in 7.0.0.

## Removal of eccCollection
This is requested by DRIVERS-2524.
In QEv1, each QE collection required three additional metadata collections: `eccCollection`, `ecocCollection`, and `escCollection`.
In QEv2, the `eccCollection` is no longer required.

## Removal of documenting `escCollection` and `ecocCollection`
This is requested by DRIVERS-2586.

## Addition of Wire Version check

QEv2 is a backwards breaking change to QEv1. Drivers supporting QEv2 will not support QEv1. Drivers supporting QEv1 will not support QEv2. QEv2 is introduced in server `7.0.0`. I expect QEv2 will not be backported.

Sending an QEv1 payload to `mongod` `7.0.0-alpha-1139-gc9e3390` results in a server error. Example:
`(Location7292602) Encountered a Queryable Encryption find payload type that is no longer supported: 5`

Sending a QEv2 payload to `mongod` `6.1.0` results in a server error. Example:
`Enumeration value '11' for field 'subtype' is not a valid value.`

Creating a QE collection using `encryptedFields` without including `eccCollection` does not result in a server error from `6.1.0`. Instead, the `eccCollection` is implicitly created by the server. The wire version check is intended to give users a helpful error if attempting to use a QEv2 driver on a QEv1 server. Without the wire version check, the user may not discover the incompatibility until a QEv2 payload is sent. This may result in the QE collections being created but being unusable.

Creating a collection with `encryptedFields` with `eccCollection` does not result in a server error from `7.0.0-alpha-1139-gc9e3390`. SERVER-75683 requests returning a helpful error if attempting to create a QEv1 collection on a QEv2 server.

### Interaction with `createEncryptedCollection`

The wire version check is proposed in `createCollection`. `createEncryptedCollection` calls `createCollection` after creating the data keys. If the wire version check fails, this may result in data keys being created and returned. This behavior is expected to be improved by the proposal in DRIVERS-2540 to create data keys and collections in a transaction.

- [x] Update changelog.
- [x] Test changes in at least one language driver. **Tested in C driver [here](https://github.com/mongodb/mongo-c-driver/pull/1232)**
~~- [] Test these changes against all server versions and topologies (including standalone, replica set, sharded clusters, and serverless).~~ **QE is skipped on serverless pending DRIVERS-2589. C driver does not currently test QE on sharded**


markbenvenuto (approved):

source/client-side-encryption/tests/legacy/fle2v2-CreateCollection-OldServer.yml
```diff
@@ -0,0 +1,41 @@
+# Requires libmongocrypt 1.8.0.
+runOn:
+  - minServerVersion: "6.0.0"
+    maxServerVersion: "6.3.99"
+    # FLE 2 Encrypted collections are not supported on standalone.
+    topology: [ "replicaset", "sharded", "load-balanced" ]
+    
+database_name: &database_name "default"
+collection_name: &collection_name "default"
```
jmikola, line 9: I noticed that various tests hard-code `collection: "encryptedCollection"` in operations. Would it make more sense for them to reference `*collection_name`?

source/client-side-encryption/tests/legacy/fle2v2-CreateCollection.yml
```diff
@@ -45,7 +43,7 @@ tests:
         arguments:
           database: *database_name
           collection: "enxcol_.encryptedCollection.esc"
-      - name: assertCollectionExists
+      - name: assertCollectionNotExists
```
jmikola, line 46: I think this comment is helpful for the first occurrence of `assertCollectionNotExists` in this file.

```suggestion
      # ecc collection is no longer created for QEv2
      - name: assertCollectionNotExists
```

source/client-side-encryption/tests/legacy/fle2v2-CreateCollection.yml
```diff
@@ -19,7 +18,6 @@ tests:
         encryptedFieldsMap:
           default.encryptedCollection: &encrypted_fields0 {
                 "escCollection": "enxcol_.encryptedCollection.esc",
-                "eccCollection": "enxcol_.encryptedCollection.ecc",
```
jmikola, line 22: I noticed these `createCollection` tests never assert that `encryptedFieldsMap` is passed to the `create` command. That seems vital to the implementation of the `createCollection()` helper, so shouldn't we be doing that?

This also relates to my earlier comment in the spec file about removing the metadata collections from `encryptedFieldsMap` entirely, so we can then assert that the driver never sends values for these fields.

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -914,6 +912,17 @@ Drivers MUST support a BSON document option named ``encryptedFields`` for any
 will be interpreted by the helper method and MUST be passed to the `create`_
 command.
 
+.. note::
+   Users are not expected to set the ``escCollection`` and ``ecocCollection`` in
+   ``encryptedFields``. SERVER-74069 added server-side validation for those fields
+   and no longer allows names to deviate from the following:
```
jmikola, line 918: Does it make sense for the `encryptedFieldsMap` in the test files to still define `escCollection` and `ecocCollection`? I originally assumed [DRIVERS-2586](https://jira.mongodb.org/browse/DRIVERS-2586) would remove the options entirely. Drivers would never consult the options and just infer names for the metadata collections using the spec format. We have CommandStartedEvent expectations for creating the metadata collections so that's covered. For the encrypted collection's CommandStartedEvent, we could then assert that the metadata names are _not_ included in the `encryptedFields` option.

This relates to my later comment in `fle2v2-CreateCollection.yml` about adding `encryptedFields` to the `create` command's CommandStartedEvent and asserting that `escCollection` and `ecocCollection` are left unset.

----

As far as suggested spec language changes, I see two alternatives:

### Drivers consult the metadata collection options

Add language that drivers MUST consult the options if specified and use them to infer the metadata collections to create.

Add tests where metadata collections with invalid names are created (using CommandStartedEvent expectations and/or assertCollectionExists after the fact). The test should expect the `createCollection()` operation to fail due to a server-side error. The CommandStartedEvent for the encrypted collection's `create` command (which fails) should assert that the `encryptedFields` option includes the invalid metadata names.

### Drivers ignore the metadata collection options

Add language that drivers MUST ignore the options. They will not be used to derive names of metadata collections to create or drop.

Add a test where invalid metadata collection names are specified in `encryptedFields` (either the client-level map or operation-level option). Assert that metadata collections with the correct names are created (thus the options are ignored). Assert that `createCollection()` fails due to a server-side error since the driver leaves those options as-is

source/client-side-encryption/client-side-encryption.rst
```diff
@@ -914,6 +912,17 @@ Drivers MUST support a BSON document option named ``encryptedFields`` for any
...
+
+   - ``enxcol_.<collectionName>.esc``
+   - ``enxcol_.<collectionName>.ecoc`
+
+   Drivers SHOULD NOT document the ``escCollection`` and ``ecocCollection``
```
jmikola, line 923: Any reason not to use `MUST NOT` here? Once this spec change is implemented, the options will serve no purpose.

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: > Does it make sense for the `encryptedFieldsMap` in the test files to still define `escCollection` and `ecocCollection`?

Good question. I think "no" since users are not expected to set `escCollection` or `ecocCollection`. https://github.com/mongodb/specifications/pull/1396/commits/f7c028a86779ba2a073af9f0df72c5c2d1bc0f6c removes `escCollection` and `ecocCollection` from tests.

> I originally assumed [DRIVERS-2586](https://jira.mongodb.org/browse/DRIVERS-2586) would remove the options entirely. Drivers would never consult the options and just infer names for the metadata collections using the spec format. We have CommandStartedEvent expectations for creating the metadata collections so that's covered. For the encrypted collection's CommandStartedEvent, we could then assert that the metadata names are _not_ included in the `encryptedFields` option.

426533fc0bc373128c4d9527342ae25cbf2afec5 adds `{$$exists: false}` for metadata name fields in `create` command sent to server.


> As far as suggested spec language changes, I see two alternatives:

I chose to keep the existing behavior: first consult `encryptedFields` for the metadata collection names. Ignoring `encryptedFields` may result in a mismatch of the metadata collection names for the collections the driver creates, and the names referenced in `encryptedFields` sent to the server. If the server ever relaxes restrictions added in SERVER-74069, the `createCollection` may operation succeed.

> 
> ### Drivers consult the metadata collection options
> Add language that drivers MUST consult the options if specified and use them to infer the metadata collections to create.

`Create Collection Helper` already requires consulting `encryptedFields` first for the metadata collection names. No change was made.


> Add tests where metadata collections with invalid names are created (using CommandStartedEvent expectations and/or assertCollectionExists after the fact). The test should expect the `createCollection()` operation to fail due to a server-side error. The CommandStartedEvent for the encrypted collection's `create` command (which fails) should assert that the `encryptedFields` option includes the invalid metadata names.

I like the idea. fbc7b17e2da1282a430f7f18a8ef49db48540e1a adds a test.

source/client-side-encryption/client-side-encryption.rst
kevinAlbs: No reason. Updated to use `MUST NOT`.

source/client-side-encryption/tests/legacy/fle2v2-CreateCollection-OldServer.yml
kevinAlbs: Not using the root `database_name` and `collection_name` was intended to avoid interacting with the drop/create behavior specified in the test runner.

https://github.com/mongodb/specifications/pull/1396/commits/1996d637689bc3b4612c5ddf275abc12ab8a9439 updates to use YAML anchors for the collection names.

source/client-side-encryption/tests/legacy/fle2v2-CreateCollection.yml
kevinAlbs: Done in https://github.com/mongodb/specifications/pull/1396/commits/40317875d1cf421ced79487d75577b947abb731f

source/client-side-encryption/tests/legacy/fle2v2-CreateCollection.yml
kevinAlbs: The `state collections and index are created` test expects to get `encryptedFields` from the `encryptedFieldsMap`. The test asserts the `encryptedFields` is sent in the `create` command here: https://github.com/mongodb/specifications/blob/364464f97872d6894a80f07f43320adcbe8e29bf/source/client-side-encryption/tests/legacy/fle2v2-CreateCollection.yml#L117

jmikola: I definitely missed that. Thanks for pointing it out.

source/client-side-encryption/tests/README.rst
```diff
@@ -67,6 +67,8 @@ The following matches a command_started_event for an insert of a document where
 
 The values of `$$type` correspond to `these documented string representations of BSON types <https://www.mongodb.com/docs/manual/reference/bson-types/>`_.
 
+- Addition of `$$exists`. The semantics match the `Unified Test Format <https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#exists>`_.
```
jmikola, line 70: Can we do without this change? The [Transactions test format](https://github.com/mongodb/specifications/tree/master/source/transactions/tests#null-values) already supports this:

> Tests MUST assert that the actual command omits any field that has a `null` value in the expected command.

Side note: seeing this made me consider how this PR is effectively piling on technical debt for migrating CSFLE tests to the unified format ([DRIVERS-2295](https://jira.mongodb.org/browse/DRIVERS-2295)), assuming we ever get to that.

source/client-side-encryption/client-side-encryption.rst
jmikola: I followed up on the `$$exists` change in https://github.com/mongodb/specifications/pull/1396/files#r1161553002.

jmikola: > I chose to keep the existing behavior: first consult encryptedFields for the metadata collection names. Ignoring encryptedFields may result in a mismatch of the metadata collection names for the collections the driver creates, and the names referenced in encryptedFields sent to the server.

For drivers where encryptedFields is an arbitrary BSON document, I can see the logic in still consulting the state collection names if present.

For strongly typed languages that may be modeling encryptedFields, wouldn't this PR suggest they remove the options entirely from their public API entirely? Otherwise, having the options in a public API might as well be documentation (the setters could show up in code completion).

Just curious if that's something that still needs to be addressed here.

kevinAlbs: > For strongly typed languages that may be modeling encryptedFields

I expect no drivers are modeling encryptedFields. [Create Collection Helper](https://github.com/mongodb/specifications/blob/fcbd905969a1178ee979d426be4dc73437e96079/source/client-side-encryption/client-side-encryption.rst#create-collection-helper) notes:
> Drivers MUST support a BSON document option named `encryptedFields`

[This survey](https://docs.google.com/spreadsheets/d/10EGNGOR4b5vyRuVI3VryN3Fl18Xe660OEYgVTYFCDYY/edit#gid=0) suggests no drivers are modeling `encryptedFields`.

source/client-side-encryption/tests/README.rst
kevinAlbs: Nice. Reverted addition of `$$exists` and replaced assertions with `null`.

source/client-side-encryption/tests/legacy/fle2v2-CreateCollection.yml
```diff
@@ -17,10 +16,7 @@ tests:
         kmsProviders:
           aws: {} # Credentials filled in from environment.
         encryptedFieldsMap:
-          default.encryptedCollection: &encrypted_fields0 {
-                "escCollection": "enxcol_.encryptedCollection.esc",
-                "eccCollection": "enxcol_.encryptedCollection.ecc",
-                "ecocCollection": "enxcol_.encryptedCollection.ecoc",
+          default.encryptedCollection: &encrypted_fields {
```
jmikola, line 19: Likewise, no reason to use `{` here. Consider the more idiomatic YAML syntax I suggested above:

```
default.encryptedCollection: &encrypted_fields
  fields:
    - path: "firstName"
      bsonType: "string"
      keyId: { $binary: { base64: "AAAAAAAAAAAAAAAAAAAAAA==", subType: "04" }}
```

source/client-side-encryption/tests/legacy/fle2v2-CreateCollection.yml
```diff
@@ -35,92 +31,94 @@ tests:
...
+                "ecocCollection": null,
+                "eccCollection": null,
+                "fields": [
+                    {
+                        "path": "firstName",
+                        "bsonType": "string",
+                        "keyId": { "$binary": { "subType": "04", "base64": "AAAAAAAAAAAAAAAAAAAAAA==" }}
+                    }
+                ]
+            }
```
jmikola, line 115:
```suggestion
              encryptedFields: &encrypted_fields_expectation
                # Expect state collections are not included in the encryptedFields sent to the server.
                escCollection: null
                ecocCollection: null
                eccCollection: null
                fields:
                  - path: "firstName"
                    bsonType: "string"
                    keyId: { $binary: { base64: "AAAAAAAAAAAAAAAAAAAAAA==", subType: "04" }}
```

Removing the JSON-like syntax with trailing commas.

source/client-side-encryption/tests/legacy/fle2v2-CreateCollection.yml
```diff
@@ -1240,27 +890,53 @@ tests:
...
             command_name: drop
             database_name: *database_name
         # events from dropCollection ... end
+  - description: "encryptedFields are consulted for metadata collection names"
+    clientOptions:
+      autoEncryptOpts:
+        kmsProviders:
+          aws: {} # Credentials filled in from environment.
+        encryptedFieldsMap:
+          default.encryptedCollection: {
```
jmikola, line 918: Same criticism as the above regarding JSON-like syntax. I'll leave this to you to address. The `fields` array can be copied over from an earlier suggestion for consistency.

source/client-side-encryption/tests/legacy/fle2v2-EncryptedFields-vs-EncryptedFieldsMap.yml
```diff
@@ -18,9 +18,6 @@ tests:
           local: {'key': {'$binary': {'base64': 'Mng0NCt4ZHVUYUJCa1kxNkVyNUR1QURhZ2h2UzR2d2RrZzh0cFBwM3R6NmdWMDFBMUN3YkQ5aXRRMkhGRGdQV09wOGVNYUMxT2k3NjZKelhaQmRCZGJkTXVyZG9uSjFk', 'subType': '00'}}}
         encryptedFieldsMap: {
           "default.default": {
```
jmikola, line 20: I realize this is outside the scope of the PR, but appears to be more JSON-like syntax in YAML. This probably deserves another ticket to clean up in a separate PR.

jmikola (approved):
Some suggested YAML formatting changes that I missed in earlier reviews. I'll leave you to address these on your own and merge w/o another round of review.

Leaving an LGTM to you're not blocked.

source/client-side-encryption/tests/legacy/fle2v2-CreateCollection.yml
kevinAlbs: I do not think rewriting the JSON adds value. The JSON more closely matches the the `encryptedFields` shown in output of `listCollections` from the server. The JSON may be easier for developers to visually compare. I prefer not to change.

source/client-side-encryption/etc/data/encryptedFields-Range-Date.json
```diff
@@ -1,7 +1,4 @@
 {
-    "escCollection": "enxcol_.default.esc",
-    "eccCollection": "enxcol_.default.ecc",
-    "ecocCollection": "enxcol_.default.ecoc",
```
jmikola, line 4: @kevinAlbs: I just noticed this, but if you're removing the metadata collections here, doesn't that make this file identical to its `range-encryptedFields-<type>.json` counterpart used by [Prose Test 22](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/tests/README.rst#range-explicit-encryption)?

If so, would it make sense to just stick with a single file and update the `yamlfile()` calls in the test templates?

jmikola: Related PR where I encountered this: https://github.com/mongodb/specifications/pull/1399

kevinAlbs: > The only difference I see is that the range- files also specify contention: 0, which is the default and may be irrelevant.

Another difference: the field name in [encryptedFields-Range-Decimal.json is encryptedDecimal](https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/etc/data/encryptedFields-Range-Decimal.json#L13).

I suggest filing a DRIVERS ticket for the improvement. This change can be made in a separate PR to avoid adding unrelated changes to this PR.

katcharov (approved):
LGTM

source/client-side-encryption/etc/data/encryptedFields-Range-Date.json
jmikola: Opened https://jira.mongodb.org/browse/DRIVERS-2604.

PR #1398, opened 2023-04-10
DRIVERS-2209 Remove VERCEL_URL from collection and specify vercel>aws precedence

abr-egn:
DRIVERS-2209

VERCEL_URL could potentially contain sensitive information, so it shouldn't be collected.  Also, as vercel runs on top of aws, it needs to take precedence when detecting the host environment.


nbbeeken (approved):
LGTM

FYI @DmitryLukyanov 

Looks like CSHARP-4353 and RUST-1501 have been merged so there needs to be follow up.

PR #1400, opened 2023-04-11
DRIVERS-2604 deduplicate encrypted fields files

kevinAlbs:
# Summary

- Replace use of `encryptedFields-Range-*` files with `range-encryptedFields-*`. 
- Replace references to `encryptedDouble` and `encryptedDecimal` with `encryptedDoubleNoPrecision` and `encryptedDecimalNoPrecision` to match the field name in the `range-encryptedFields-*` files.

There is no expected test coverage or behavior change.

Changes tested with C driver in this patch build: https://spruce.mongodb.com/version/6436ce262a60edecd5b8f761

# Background & Motivation

Motivated by this comment: https://github.com/mongodb/specifications/pull/1396#discussion_r1162257008
The `range-encryptedFields-*.json` files and the `encryptedFields-Range-*.json` files are almost identical.

14c5099a65448f4397bf336c52f6ea048d855c91 adds `contention` to `encryptedFields-Range-*.json`. The server response to `listCollections` includes the default value chosen for `contention`. `encryptedFields-Range-*.json` is used both for collection creation and for the `command_started` event assertions. Without adding `contention` to the expected `command_started` event, the test assertion fails since the actual event contains the default `contention`.

9b2df75bc3c8139916dafbf3bd48ced852390179 uses `$numberLong` consistently for sparsity. The server accepts both `$numberLong` and `$numberInt`. The only motivation was for consistency.

- ~~[ ] Update changelog.~~ **N/A. Only test files changed**
- [x] Test changes in at least one language driver. **Tested in C driver**
- ~~[ ] Test these changes against all server versions and topologies (including standalone, replica set, sharded clusters, and serverless).~~ **Tested in C driver on replica set only.**


jmikola (approved):
I only glossed over the regenerated test files, but the template and data file changes LGTM. Thanks for cleaning this up!

PR #1402, opened 2023-04-12
DRIVERS-2232: remove usages of `currentOp` and `collStats` from tests

baileympearson:
Node POC: https://github.com/mongodb/node-mongodb-native/pull/3638

This PR removes usage of the `collStats` command and `currentOp` from tests.  

- [x] Update changelog.

### DIFF
--- a/source/client-side-encryption/etc/test-templates/bypassedCommand.yml.template
+++ b/source/client-side-encryption/etc/test-templates/bypassedCommand.yml.template
@@ -26,7 +26,7 @@ tests:
           command:
             ping: 1
           command_name: ping
-  - description: "current op is not bypassed"
+  - description: "kill op is not bypassed"
     clientOptions:
       autoEncryptOpts:
         kmsProviders:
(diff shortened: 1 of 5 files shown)


durran (approved):
Test changes LGTM and am fine with the reasoning for skipping the currentOp test.

dariakp (approved):

alcaeus (approved):
LGTM.

> An alternative solution would be to choose a max server version (say, 6.99.0) and prohibit it from running on any future server versions.

That's what we usually did, but feel free to do this in a separate PR. IMO, there's no advantage to wait until the first driver breaks, so we may as well do this already.

source/client-side-encryption/etc/test-templates/bypassedCommand.yml.template
```diff
@@ -26,17 +26,18 @@ tests:
           command:
             ping: 1
           command_name: ping
-  - description: "current op is not bypassed"
+  - description: "kill op is not bypassed"
```
baileympearson, line 29: The intent of this test seems to test that bypassed commands are passed (it tests with `ping` above) and then this test asserts that a command that isn't encrypted or bypassed errors.  

Replacing `currentOp` with `killOp` (another command that is neither bypassed or encrypted) tests the same scenario.

baileympearson:
@alcaeus I decided to address the `currentOp` test here but instead of adding a minimum server version, I changed the command to a `killOp`.

dariakp (approved):

alcaeus (approved):

PR #1404, opened 2023-04-17
Update CMAP logging test desctiption

W-A-James:
The test description for the connection-pool-options suite was the same as the connection-logging suite which caused problems when running the unified tests in the Node driver.

- [x] Update changelog.


durran:
@W-A-James Can you create a DRIVERS ticket for this so we can track that there was a change that would need a test sync?

patrickfreed (approved):
changes lgtm, but yeah I agree there probably should be a DRIVERS ticket for this.

durran (changes requested):
Need to open a DRIVERS ticket and mark this PR with it.

durran (approved):

dariakp (approved):

PR #1407, opened 2023-04-19
chore(DRIVERS-2384): update localhost test domain

durran:
Fixes `MONGODB_URI` typo, makes notes on how Docket can access a localhost server, increases assume role duration to 1 hour (from default 15 minutes)

- [x] Update changelog.

### DIFF
--- a/source/faas-automated-testing/faas-automated-testing.rst
+++ b/source/faas-automated-testing/faas-automated-testing.rst
@@ -369,3 +369,4 @@ Changelog
 :2023-08-17: Add docker container notes for localhost.
 :2023-06-22: Updated evergreen configuration to use task groups.
 :2023-04-14: Added list of supported variants, added additional template config.
+:2023-04-28: Fixed URI typo, added host note, increase assume role duration.
--- a/source/faas-automated-testing/faas-automated-testing.rst
+++ b/source/faas-automated-testing/faas-automated-testing.rst
@@ -223,7 +223,9 @@ Run the function locally from the same directory where the template.yaml resides
   sam local invoke --parameter-overrides "MongoDbUri=${MONGODB_URI}"
 
 *NOTE* "127.0.0.1" in the MONGODB_URI MUST be replaced with "host.docker.internal" to test
-a local MongoDB deployment.
+a local MongoDB deployment. If "host.docker.internal" does not work (can occur on M1
+machines), drivers MAY choose to use a [bridged docker container](https://docs.docker.com/network/bridge/)
+to test locally.
 
 
 Implementing the Function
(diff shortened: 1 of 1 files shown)


source/faas-automated-testing/faas-automated-testing.rst
```diff
@@ -213,6 +213,9 @@ Run the function locally from the same directory where the template.yaml resides
   sam build
   sam local invoke --parameter-overrides "MongoDbUri=${MONGODB_URI}"
 
+*NOTE* "127.0.0.1" in the MONGODB_URT MUST be replaced with "host.docker.internal" to test
```
jmikola, line 216:
```suggestion
*NOTE*: "127.0.0.1" in the MONGODB_URI MUST be replaced with "host.docker.internal" to test
```

durran: Fixed.

source/faas-automated-testing/faas-automated-testing.rst
```diff
@@ -213,6 +213,9 @@ Run the function locally from the same directory where the template.yaml resides
   sam build
   sam local invoke --parameter-overrides "MongoDbUri=${MONGODB_URI}"
 
+*NOTE* "127.0.0.1" in the MONGODB_URI MUST be replaced with "host.docker.internal" to test
```
prestonvasquez, line 216: The [go1.x runtime](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html) only supports x86_64 architectures, which is fine for the CI since RHEL 8.0 and Ubuntu 20.04 are amd64. However, the "host.docker.internal" solution does not appear work on M1. We could use a [provider.al2](https://aws.amazon.com/blogs/compute/migrating-aws-lambda-functions-to-al2/) solution, but this adds additional complexity in that we would have to maintain a custom runtime and maintain both a amd64 bin and arm64 bin.

It _could_ be beneficial to suggest that if the "host.docker.internal" solution does not work, then run a server in a workstation or in a [bridged docker container](https://docs.docker.com/network/bridge/). And use those IP as the MONGODB_URI.

durran: I've rebased and added the suggested note about the bridged container.

prestonvasquez (approved):

source/faas-automated-testing/faas-automated-testing.rst
```diff
@@ -346,3 +352,4 @@ Changelog
 =========
 
 :2023-04-14: Added list of supported variants, added additional template config.
+:2023-04-28: Fixed URI typo, added host note, increase assume role duration.
```
alcaeus, line 355: Reminder to update the date before merging.

alcaeus (approved):
LGTM besides the date change.

PR #1411, opened 2023-04-24
DRIVERS-2505: Clarifications for Range Explicit Encryption prose tests 

jmikola:
https://jira.mongodb.org/browse/DRIVERS-2505

Follow-up to https://github.com/mongodb/specifications/pull/1409.

This adds clarifying text and revises formatting inconsistencies throughout the test file. It does not change test behavior.

### DIFF
--- a/source/client-side-encryption/tests/README.rst
+++ b/source/client-side-encryption/tests/README.rst
@@ -2629,7 +2629,7 @@ when attempting to create a collection with such invalid settings.
 Case 4: Insert encrypted value
 ``````````````````````````````
 
-This test is continuation of the case 1 and provides a way to complete inserting 
+This test is continuation of the case 1 and provides a way to complete inserting
 with encrypted value.
 
 1. Create a new create-collection options `Opts` including the following::
@@ -2737,14 +2737,14 @@ Use ``encryptedClient`` to insert the following documents into ``db.explicit_enc
 
 Test Setup: RangeOpts
 `````````````````````
-This section lists the values to use for ``RangeOpts`` for each of the supported data types, since each data type requires a different ``RangeOpts``. 
+This section lists the values to use for ``RangeOpts`` for each of the supported data types, since each data type requires a different ``RangeOpts``.
 
-Each test listed in the cases below must pass for all supported data types unless it is stated the type should be skipped. 
+Each test listed in the cases below must pass for all supported data types unless it is stated the type should be skipped.
 
 #. DecimalNoPrecision
 
    .. code:: typescript
-   
+
       class RangeOpts {
          sparsity: 1,
       }
@@ -2752,7 +2752,7 @@ Each test listed in the cases below must pass for all supported data types unles
 #. DecimalPrecision
 
    .. code:: typescript
-   
+
       class RangeOpts {
          min: { "$numberDecimal": "0" },
          max: { "$numberDecimal": "200" },
@@ -2763,7 +2763,7 @@ Each test listed in the cases below must pass for all supported data types unles
 #. DoubleNoPrecision
 
    .. code:: typescript
-   
+
       class RangeOpts {
          sparsity: 1,
       }
@@ -2771,7 +2771,7 @@ Each test listed in the cases below must pass for all supported data types unles
 #. DoublePrecision
 
    .. code:: typescript
-   
+
       class RangeOpts {
          min: { "$numberDouble": "0" },
          max: { "$numberDouble": "200" },
@@ -2782,7 +2782,7 @@ Each test listed in the cases below must pass for all supported data types unles
 #. Date
 
    .. code:: typescript
-   
+
       class RangeOpts {
          min: {"$date": { "$numberLong": "0" } } ,
          max: {"$date": { "$numberLong": "200" } },
@@ -2792,7 +2792,7 @@ Each test listed in the cases below must pass for all supported data types unles
 #. Int
 
    .. code:: typescript
-   
+
       class RangeOpts {
          min: {"$numberInt": "0" } ,
          max: {"$numberInt": "200" },
@@ -2802,7 +2802,7 @@ Each test listed in the cases below must pass for all supported data types unles
 #. Long
 
    .. code:: typescript
-   
+
       class RangeOpts {
          min: {"$numberLong": "0" } ,
          max: {"$numberLong": "200" },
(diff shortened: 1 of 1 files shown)


source/client-side-encryption/tests/README.rst
```diff
@@ -2693,302 +2694,343 @@ Create a ClientEncryption object named ``clientEncryption`` with these options:
...
    }
 
-Ensure 201 matches the type of the encrypted field. The error should be raised because 201 is greater than the maximum value in ``RangeOpts``.
+Assert that an error was raised because 201 is greater than the maximum value in ``RangeOpts``.
 
-Assert that an error was raised.
 
-Case 7: encrypting a document of a different type errors 
-````````````````````````````````````````````````````````
+Case 7: encrypting a value of a different type errors
```
jmikola, line 2993: The original prose test was unclear if this was testing insertion (e.g. wrong type in the `encrypted<Type>` field) or direct encryption of a value.

I assumed the latter based on the [libmongoc test](https://github.com/mongodb/mongo-c-driver/commit/7a6b6369eb918bc2698fd6dc3aeab12845176a36#diff-d363a5539f2fbabfadf2d7cc6cdb96a203fb63f2caa31bae731c3e12d97b204cR4333).

Is there any concern other drivers got this wrong and are instead testing with document insertion? Should we have two separate prose tests?

source/client-side-encryption/tests/README.rst
```diff
@@ -2693,302 +2694,343 @@ Create a ClientEncryption object named ``clientEncryption`` with these options:
...
+      bypassQueryAnalysis: true,
    }
 
 The remaining tasks require setting ``RangeOpts``. `Test Setup: RangeOpts`_ lists the values to use for ``RangeOpts`` for each of the supported data types.
 
-Use ``clientEncryption`` to encrypt these values: 0, 6, 30, and 200. Ensure the type matches with the type of the encrypted field. For example, if the encrypted field is ``encryptedDoubleNoPrecision`` encrypt the value 6.0.
+Use ``clientEncryption`` to encrypt these values: 0, 6, 30, and 200. Ensure the type matches that of the encrypted field. For example, if the encrypted field is ``encryptedDoubleNoPrecision`` encrypt the value 6.0.
 
-Encrypt these values with the matching ``RangeOpts`` listed in `Test Setup: RangeOpts`_ and these ``EncryptOpts``:
+Encrypt using the following ``EncryptOpts``:
```
jmikola, line 2717: The following text appears in an earlier paragraph in this section:

> The remaining tasks require setting ``RangeOpts``. Test Setup: RangeOpts lists the values to use for ``RangeOpts`` for each of the supported data types.

I don't think we needed to repeat the link. In all of the `EncryptOpts` examples that follow, I added `rangeOpts: <RangeOpts for Type>` to remind readers. I think that's more concise and ensure that `rangeOpts` appears in the `EncryptOpts` itself and is less likely to be missed.

source/client-side-encryption/tests/README.rst
```diff
@@ -2693,302 +2694,343 @@ Create a ClientEncryption object named ``clientEncryption`` with these options:
...
 
-Assert these three documents ``{ "encrypted<Type>": 6 }, { "encrypted<Type>": 30 }, { "encrypted<Type>": 200}`` are returned.
+Assert the following three documents are returned:
+
+.. code:: javascript
 
+   // Convert 6, 30, and 200 to the encrypted field type
+   { "_id": 1, "encrypted<Type>": 6 }
+   { "_id": 2, "encrypted<Type>": 30 }
+   { "_id": 3, "encrypted<Type>": 200 }
```
jmikola, line 2869: The original documents did not include `_id` fields, but I think that's helpful for matching. Breaking this out to a code block also made it more readable and easier to remind drivers to convert values.

source/client-side-encryption/tests/README.rst
```diff
@@ -2693,302 +2694,343 @@ Create a ClientEncryption object named ``clientEncryption`` with these options:
...
-      contentionFactor: 0
-   }
-   
-   class RangeOpts {
-      min: 0,
-      max: 200,
-      sparsity: 1,
-      precision: 2,
+      contentionFactor: 0,
+      rangeOpts: <RangeOpts for Type with precision added>,
```
jmikola, line 3033: The previous text didn't mention anything about using RangeOpts for the specific type. I think this makes it more explicit and ensures the min/max values with have the correct type.

source/client-side-encryption/tests/README.rst
```diff
@@ -2693,302 +2694,343 @@ Create a ClientEncryption object named ``clientEncryption`` with these options:
...
 
-If the encrypted field is ``encryptedInt`` encrypt ``{ "encryptedInt": { "$numberDouble": "6" } }``.
-Otherwise, encrypt ``{ "encrypted<Type>": { "$numberInt": "6" }``.
-Assert an error was raised.
+Ensure that ``RangeOpts`` corresponds to the type of the encrypted field (i.e. expected type) and not that of the value being passed to ``clientEncryption.encrypt()``.
+
+Assert that an error was raised.
+
 
+Case 8: setting precision errors if the type is not a floating point
```
jmikola, line 3017: "floating point" seemed more accurate to describe a Decimal128 _or_ double type. I'm not sure if the original name was written before Decimal128 was added in a subsequent commit.

source/client-side-encryption/tests/README.rst
```diff
@@ -2693,302 +2694,343 @@ Create a ClientEncryption object named ``clientEncryption`` with these options:
...
-Ensure 201 matches the type of the encrypted field. The error should be raised because 201 is greater than the maximum value in ``RangeOpts``.
+Assert that an error was raised because 201 is greater than the maximum value in ``RangeOpts``.
 
-Assert that an error was raised.
 
-Case 7: encrypting a document of a different type errors 
-````````````````````````````````````````````````````````
+Case 7: encrypting a value of a different type errors
+`````````````````````````````````````````````````````
 This test case should be skipped if the encrypted field is ``encryptedDoubleNoPrecision`` or ``encryptedDecimalNoPrecision``.
```
jmikola, line 2995: Is there a particular reason this test is skipped for "NoPrecision"? If so, it might helpful to add that to this sentence.

Unlike test 8, the reason for this skip wasn't immediately clear to me.

source/client-side-encryption/tests/README.rst
jmikola: Following up: I realize now that this actually changes the test behavior.

I originally thought that the use of fixed RangeOpts options was a mistake. It see now that this is relying on libmongocrypt validating the `precision` field _before_ checking types based on `min` and `max`. That seems pretty fragile, but I'm OK to revert this in order to avoid any downstream changes.

Please let me know what you think.

kevinAlbs (approved):
The improvements and clarifications are appreciated. LGTM with one suggested change.

source/client-side-encryption/tests/README.rst
jmikola: Ok, I'll keep this in place. And it sounds like you don't care to notify drivers of this clarification. I suppose if anything breaks down the line (change in libmongocrypt's validation order), they can come back to this test and easily correct the tests.

PR #1419, opened 2023-05-12
DRIVERS-1670: Add log messages to SDAM spec

BorisDog:
[DRIVERS-1670](https://jira.mongodb.org/browse/DRIVERS-1670)

- [x] Update changelog.


source/unified-test-format/schema-1.14.json
```diff
@@ -0,0 +1,682 @@
...
+          "client": { "type": "string" },
+          "messages": {
+            "type": "array",
+            "items": { "$ref": "#/definitions/expectedLogMessage" } 
+            },
+          "ignoreExtraLogs": { "type": "boolean" },
+          "ignoreLogMessages": {
+            "type": "array",
+            "items": { "$ref": "#/definitions/expectedLogMessage" } 
+            }
```
jmikola, line 557: These warrant new tests under `unified-test-format/tests/invalid`. See https://github.com/mongodb/specifications/commit/d4f83d83a380a1863094c3f7bcb5d27b5963a2c9 for tests added with the original logging spec changes.

Tests for `ignoreLogMessages` will likely resemble existing tests for `messages`.

`ignoreExtraLogs` will be more straightforward and likely just need a test with an invalid type (e.g. "string" instead of boolean).

source/unified-test-format/unified-test-format.rst
```diff
@@ -3894,6 +3894,9 @@ Changelog
 ..
   Please note schema version bumps in changelog entries where applicable.
 
+:2022-05-10: **Schema version 1.14.**
+             Add ``ignoreLogMessages`` and ``ignoreExtraLogs`` fields
+             to ``expectLogMessages``section.
```
jmikola, line 3899:
```suggestion
             to ``expectedLogMessagesForClient``.
```

This is consistent with what you actually changed in `schema-1.14.json`.

Also, you forgot to add documentation for these fields in the spec itself. Please add that to the [`expectedLogMessagesForClient`](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#expectedlogmessagesforclient) section.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-logging-and-monitoring.rst
```diff
@@ -0,0 +1,713 @@
+.. role:: javascript(code)
+  :language: javascript
+
+=========================================
```
kmahar, line 4: I'm a little concerned about the way that it seems like GitHub is considering this a brand new file rather than a rename of the existing one since I think it is going to totally blow out the git history for this file.

it also makes this more difficult to review since you can't tell from looking at this file what is old and what is new.

can we hold off on the file rename for now and do that in a separate commit at the end? 

(I'll wait to review this file until it is renamed back since it will be easier then)

source/server-discovery-and-monitoring/tests/unified/logging-replicaset.yml
```diff
@@ -0,0 +1,269 @@
...
+
+runOnRequirements:
+  - topologies:
+      - replicaset
+    minServerVersion: "4.4" # awaitable hello
+
+createEntities:
+  - client:
+      id: &setupClient setupClient
+      useMultipleMongoses: false
```
kmahar, line 13: nit: I think we can omit this option since it's a replica set; it won't have any effect.

source/server-discovery-and-monitoring/tests/unified/logging-replicaset.yml
```diff
@@ -0,0 +1,269 @@
...
+                observeEvents:
+                  - topologyDescriptionChangedEvent
+      # ensure the topology has been fully discovered before closing the client.
+      - name: waitForEvent
+        object: testRunner
+        arguments:
+          client: *client
+          event:
+            topologyDescriptionChangedEvent: {}
+          count: 4
```
kmahar, line 35: it might be helpful to leave a comment stating what the events we'd expect to see here are. (same applies in similar test for other topologies)

source/server-discovery-and-monitoring/tests/unified/logging-replicaset.yml
```diff
@@ -0,0 +1,269 @@
...
+              message: "Server heartbeat succeeded"
+          - data:
+              message: "Server heartbeat failed"
+        messages:
+          - level: debug
+            component: topology
+            data:
+              message: "Starting topology monitoring"
+              topologyId: { $$exists: true }
+          - level: debug
```
kmahar, line 55: similarly, could we add comments for these "changed" events indicating what the expected changes are? I know we can't assert on them, but for anyone debugging failures e.g. missing change events it would be helpful for them to know what we were expecting to receive here. (same applies in similar test for other topologies)

source/server-discovery-and-monitoring/tests/unified/logging-replicaset.yml
```diff
@@ -0,0 +1,269 @@
...
+              message: "Topology description changed"
+              topologyId: { $$exists: true }
+              previousDescription:  { $$exists: true }
+              newDescription: { $$exists: true }
+          - level: debug
+            component: topology
+            data:
+              message: "Stopped topology monitoring"
+              topologyId: { $$exists: true }
+  - description: Succefull hearbeat
```
kmahar, line 116:
```suggestion
  - description: Successful hearbeat
```

source/server-discovery-and-monitoring/tests/unified/logging-replicaset.yml
```diff
@@ -0,0 +1,269 @@
...
+        arguments:
+          entities:
+          - client:
+              id: *client
+              observeLogMessages:
+                topology: debug
+              observeEvents:
+              - serverHeartbeatSucceededEvent
+              uriOptions:
+                heartbeatFrequencyMS: 500
```
kmahar, line 129: nit: comment explaining why we set this option? (same applies elsewhere that we set it)

source/server-discovery-and-monitoring/tests/unified/logging-replicaset.yml
```diff
@@ -0,0 +1,269 @@
...
+              serverHost: { $$type: string }
+              serverPort: { $$type: [int, long] }
+              driverConnectionId: { $$exists: true }
+              serverConnectionId: { $$exists: true }
+              durationMS: { $$type: [int, long] }
+              reply:
+                $$matchAsDocument:
+                  "$$matchAsRoot":
+                    ok: 1
+  - description: Failing hearbeat
```
kmahar, line 200:
```suggestion
  - description: Failing heartbeat
```

source/server-discovery-and-monitoring/tests/unified/logging-replicaset.yml
```diff
@@ -0,0 +1,269 @@
...
+            uriOptions:
+              heartbeatFrequencyMS: 500
+              appname: failingHeartbeatLoggingTest
+    - name: failPoint
+      object: testRunner
+      arguments:
+        failPoint:
+          configureFailPoint: failCommand
+          mode:
+            times: 2
```
kmahar, line 222: why 2x? (leave a comment? same for other similar tests)

source/server-discovery-and-monitoring/tests/unified/logging-sharded.yml
```diff
@@ -0,0 +1,256 @@
...
+              message: "Topology description changed"
+              topologyId: { $$exists: true }
+              previousDescription:  { $$exists: true }
+              newDescription: { $$exists: true }
+          - level: debug
+            component: topology
+            data:
+              message: "Stopped topology monitoring"
+              topologyId: { $$exists: true }
+  - description: Succefull hearbeat
```
kmahar, line 103:
```suggestion
  - description: Successful heartbeat
```

source/server-discovery-and-monitoring/tests/unified/logging-sharded.yml
```diff
@@ -0,0 +1,256 @@
...
+              serverHost: { $$type: string }
+              serverPort: { $$type: [int, long] }
+              driverConnectionId: { $$exists: true }
+              serverConnectionId: { $$exists: true }
+              durationMS: { $$type: [int, long] }
+              reply:
+                $$matchAsDocument:
+                  "$$matchAsRoot":
+                    ok: 1
+  - description: Failing hearbeat
```
kmahar, line 187:
```suggestion
  - description: Failing heartbeat
```

source/server-discovery-and-monitoring/tests/unified/logging-replicaset.yml
```diff
@@ -0,0 +1,269 @@
...
+      - level: debug
+        component: topology
+        data:
+          message: "Starting topology monitoring"
+          topologyId:
+            "$$exists": true
+      - level: debug
+        component: topology
+        data:
+          message: "Server heartbeat failed"
```
kmahar, line 261: do we only expect to see one heartbeat failure or should there be one for each host?

source/server-discovery-and-monitoring/tests/unified/logging-standalone.yml
```diff
@@ -0,0 +1,227 @@
...
+
+runOnRequirements:
+  - topologies:
+      - single
+    minServerVersion: "4.4" # awaitable hello
+
+createEntities:
+  - client:
+      id: &setupClient setupClient
+      useMultipleMongoses: false
```
kmahar, line 13: ditto re unneeded option

source/server-discovery-and-monitoring/tests/unified/logging-standalone.yml
```diff
@@ -0,0 +1,227 @@
...
+              message: "Topology description changed"
+              topologyId: { $$exists: true }
+              previousDescription:  { $$exists: true }
+              newDescription: { $$exists: true }
+          - level: debug
+            component: topology
+            data:
+              message: "Stopped topology monitoring"
+              topologyId: { $$exists: true }
+  - description: Succefull heartbeat
```
kmahar, line 93:
```suggestion
  - description: Successful heartbeat
```

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
```diff
@@ -101,7 +101,7 @@ Terms
 Server
 ``````
 
-A mongod or mongos process.
+A mongod or mongos process, or a load balancer.
```
kmahar, line 104: I think we should have load balanced mode tests as well and likely add something to this section of the load balancer spec: https://github.com/mongodb/specifications/blob/master/source/load-balancers/load-balancers.rst#monitoring

kmahar (changes requested):
some initial comments - so far, mostly a bunch of nits around adding comments to help readers understand the YAML tests. 
also, did we decide against adding any logs that don't correspond to existing SDAM events? I remember discussing with @jyemin that Java had some possibly useful SDAM logs we could consider including.

source/server-discovery-and-monitoring/tests/unified/logging-replicaset.yml
BorisDog: Accounting for RTT thread. Done.

source/server-discovery-and-monitoring/tests/unified/logging-replicaset.yml
BorisDog: I think just the primary heartbeat is supposed to fail.

source/server-discovery-and-monitoring/tests/unified/logging-sharded.yml
BorisDog: Done.

source/server-discovery-and-monitoring/tests/unified/logging-sharded.yml
BorisDog: Done, tnx.

source/server-discovery-and-monitoring/tests/unified/logging-standalone.yml
BorisDog: Done.

source/server-discovery-and-monitoring/tests/unified/logging-standalone.yml
BorisDog: Done.

source/unified-test-format/unified-test-format.rst
BorisDog: Thanks, done.

source/unified-test-format/schema-1.14.json
BorisDog: Thanks for the pointers, done.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1398,6 +1398,14 @@ The structure of each object is as follows:
   message counts are the same and that there are no extra messages emitted by the
   client during the test run.
 
+- ``ignoreLogMessages``: Optional array of `expectedLogMessage`_ objects. List of
+  messages, which should be ignored on the corresponding client while executing `operations`_.
```
jmikola, line 1402: `ignoreExtraLogs` is pretty clear, but I think `ignoreLogMessages` can still use some clarification.

I'm assuming the following:

 * The array is an unordered set of messages (unlike `messages` where order is significant)
 * When a log message is observed, the test runner will first check if it matches anything in `ignoreLogMessages` (following the same "exact match" rules as `messages`). If so, the log message is discarded _before_ being considered for `messages` evaluation.

If that's correct, can you capture this in the spec?

----

Also, the language in `messages` may need some revision to avoid conflict with the new fields being introduced. I'd suggest adding the following paragraph to the end of `messages`:

```
  Note: ``ignoreMessages`` and ``ignoreExtraMessages`` may exclude log messages
  from this evaluation.
```

And lastly, I'd suggest renaming the field names:

 * `ignoreLogMessages` -> `ignoreMessages`
 * `ignoreExtraLogs` -> `ignoreExtraMessages`

I think these are more consistent since the original field is just `messages`. "Log" is already included in the higher-level `expectLogMessages` field name, so we already have that context. `ignoreExtraMessages` would also be consistent with `ignoreExtraEvents` (where the main array field there is `events`, like we have with `messages` here).

BorisDog: Thanks for the good suggestions, done.

source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst
BorisDog: Added test and short description to load-balancer spec.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1397,6 +1397,18 @@ The structure of each object is as follows:
   that the messages produced are an exact match, i.e. that the expected and actual
   message counts are the same and that there are no extra messages emitted by the
   client during the test run.
+  Note: ``ignoreMessages`` and ``ignoreExtraMessages`` may exclude log messages from this evaluation.
+
+- ``ignoreMessages``: Optional array of `expectedLogMessage`_ objects. Unordered set of
+  messages, which MUST be ignored on the corresponding client while executing `operations`_.
+  Test runer MUST exclude all log messages from observed messages that match any of the messages
```
kmahar, line 1404:
```suggestion
  The test runner MUST exclude all log messages from observed messages that match any of the messages 
```

source/server-discovery-and-monitoring/tests/unified/logging-replicaset.yml
kmahar: I think this one is still misspelled

source/server-discovery-and-monitoring/tests/unified/logging-replicaset.yml
```diff
@@ -0,0 +1,292 @@
...
+              message: "Topology description changed"
+              topologyId: { $$exists: true }
+              previousDescription:  { $$exists: true } # ReplicaSet topology
+              newDescription: { $$exists: true } # unknown topology
+          - level: debug
+            component: topology
+            data:
+              message: "Stopped topology monitoring"
+              topologyId: { $$exists: true }
+  - description: Successful hearbeat
```
kmahar, line 124:
```suggestion
  - description: Successful heartbeat
```

source/server-discovery-and-monitoring/tests/unified/logging-sharded.yml
kmahar: i think this one still needs to be fixed as well

source/server-discovery-and-monitoring/tests/unified/logging-replicaset.yml
kmahar: is that the case? I'd think the failpoint would apply on any member of the topology. 

that said, I realize now we only set this to times: 2, so it's (in theory) hit once by the RTT monitor and once for a heartbeat, on the first server that we happen to reach out to. 

however, I wonder if it's possible a driver has an interleaving of RTT monitoring commands and server monitoring commands that would lead us to not hit this failpoint on a server monitoring command? i.e. if two RTT monitoring commands hit the failpoint and so it gets disabled before any server monitoring commands encounter it.

it might be more robust to just set the failpoint to always on?

source/load-balancers/load-balancers.rst
```diff
@@ -116,6 +116,16 @@ the following series of SDAM events:
 Drivers MUST also emit a :code:`ServerClosedEvent` and :code:`TopologyClosedEvent` when
 the topology is closed and MUST NOT emit any other events when operating in this mode.
 
+Log Messages
+^^^^^^^^^^^^
+
+Please refer to the `SDAM logging specification <../server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst#log-messages>`_
```
kmahar, line 122: may be helpful to also refer users to the monitoring section and point out the same considerations apply for the messages that correspond to the events

kmahar:
my remaining comments are fairly minor, so I don't need to re-review. 

my previous question still applies but I'll defer to all of you on that:
> also, did we decide against adding any logs that don't correspond to existing SDAM events? 

also, it might be good to have someone who owns SDAM review this too.

source/load-balancers/load-balancers.rst
BorisDog: Sorry I didn't fully understand. Current reference is to the logging section in monitoring.
Do you mean instead of listing the events here just refer to the relevant events?

jmikola (approved):
Changes within `unified-test-format/` LGTM. Not reviewing the other files.

Note that you'll need to rebase/merge master to fix the conflict in `unified-test-format.rst`. Please maintain a blank line after the "Please note schema version bumps" RST comment when doing so.

alcaeus (approved):

source/unified-test-format/unified-test-format.rst
```diff
@@ -1397,6 +1397,18 @@ The structure of each object is as follows:
   that the messages produced are an exact match, i.e. that the expected and actual
   message counts are the same and that there are no extra messages emitted by the
   client during the test run.
+  Note: ``ignoreMessages`` and ``ignoreExtraMessages`` may exclude log messages from this evaluation.
+
+- ``ignoreMessages``: Optional array of `expectedLogMessage`_ objects. Unordered set of
+  messages, which MUST be ignored on the corresponding client while executing `operations`_.
+  The test runner MUST exclude all log messages from observed messages that match any of the messages
+  in ``ignoreMessages`` array before ``messages`` evaluation.
+  Matching rules used to match messages in ``ignoreMessages`` are identical to match rules used for ``messages`` matching.
```
prestonvasquez, line 1406: Should we relax the "ignoreMessages" conditions?

>  Matching rules used to match messages in ``ignoreMessages`` are identical to match rules used for ``messages`` matching.

Something like this:

> Matching rules used to match messages in ``ignoreMessages`` are upto level, component, and data.message.

For example, in the logging-replicatset.yml unified spec test we have the following message in the ignore list:

```yml
- level: debug
  component: topology
  data:
    message: "Server heartbeat started"
```
However, according to the section ["Server Heartbeat Started" Log Message](https://github.com/mongodb/specifications/pull/1419/files?short_path=e2f3a7c&unchanged=expanded#id29) the unstructured log is this:

> Heartbeat started for {{serverHost}}:{{serverPort}} on connection with driver-generated ID {{driverConnectionId}} and server-generated ID {{serverConnectionId}} in topology with ID {{topologyId}}. Awaited: {{awaited}}

The above implies that serverHost, serverPort, serverConnectionId, topologyId and awaited fields should all be present in the "data" structure. Thus I would expect the structured analogue to always looks something like this:

```json
{"level": "debug", "component": "topology", "data": {"message": "Server heartbeat started" , "serverPort": "x", "serverHost": "y", "topologyId": "z", "awaited": true}}
```
And so the above yaml will never actually indicate ignoring anything (i.e. the data would always be a subset of the actual message's data).

source/server-discovery-and-monitoring/tests/unified/logging-sharded.yml
```diff
@@ -0,0 +1,249 @@
...
+              topologyId: { $$exists: true }
+  - description: Successful heartbeat
+    operations:
+      - name: createEntities
+        object: testRunner
+        arguments:
+          entities:
+          - client:
+              id: *client
+              observeLogMessages:
```
prestonvasquez, line 120:
```suggestion
              uriOptions:
                  heartbeatFrequencyMS: 5000
              observeLogMessages:
```

If comment (A) is accepted, this would need the above udpate.

source/server-discovery-and-monitoring/tests/unified/logging-standalone.yml
```diff
@@ -0,0 +1,259 @@
...
+              topologyId: { $$exists: true }
+  - description: Successful heartbeat
+    operations:
+      - name: createEntities
+        object: testRunner
+        arguments:
+          entities:
+            - client:
+                id: &client client
+                observeLogMessages:
```
prestonvasquez, line 107:
```suggestion
              uriOptions:
                  heartbeatFrequencyMS: 5000
              observeLogMessages:
```

If comment (A) is accepted, this would need the above udpate.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst
```diff
@@ -15,7 +15,8 @@ SDAM Monitoring Specification
 Abstract
 ========
 
-The SDAM monitoring specification defines a set of behaviour in the drivers for providing runtime information about server discovery and monitoring events (SDAM) to any 3rd party library as well internal driver use, such as logging.
+The SDAM logging and monitoring specification defines a set of behaviour in the drivers for providing runtime information about server discovery and monitoring (SDAM) in log
+messages as well as in events which users can consume programmatically, either directly or by integrating with third-party APM libraries.
```
prestonvasquez, line 19:
```suggestion
The SDAM logging and monitoring specification defines a set of behaviors in the driver for providing runtime information about server discovery and monitoring (SDAM) in log messages, as well as in events that users can consume programmatically, either directly or by integrating with third-party APM libraries.
```

source/server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst
```diff
@@ -39,107 +40,111 @@ Terms
...
+--------
 
-Events that MUST be published (with their conditions) are as follows.
+Documentation
+-------------
 
-.. list-table::
-   :header-rows: 1
-   :widths: 50 50
+The documentation provided in code below is merely for driver authors and SHOULD NOT be taken as required documentation for the driver.
```
prestonvasquez, line 57:
```suggestion
The documentation provided in the code below is merely for driver authors and SHOULD NOT be taken as required documentation for the driver.
```

prestonvasquez (changes requested):

source/server-discovery-and-monitoring/tests/unified/logging-replicaset.yml
```diff
@@ -0,0 +1,291 @@
...
+              topologyId: { $$exists: true }
+  - description: Successful heartbeat
+    operations:
+      - name: createEntities
+        object: testRunner
+        arguments:
+          entities:
+          - client:
+              id: *client
+              observeLogMessages:
```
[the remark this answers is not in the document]
ShaneHarvey: I believe the intention is that when a client is first created it should immediately start monitoring (not wait 10 seconds). Otherwise we could be adding extra latency when the client is first used.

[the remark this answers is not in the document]
BorisDog: Yes, C# starts monitoring immediately, looks like [SDAM monitoring](https://github.com/mongodb/specifications/blob/a10e270087bc8ec661561be8cffd287bc6982d91/source/server-discovery-and-monitoring/server-monitoring.rst#monitor-thread) code behaves the same.

source/unified-test-format/unified-test-format.rst
BorisDog: I think we are good here.
Matching the data field is done according to [root document evaluation rules](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst). Which permits extra fields. [expectedLogMessage](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#id40)

Also we have to permit extra fields as drivers may add their own parameters as needed.

source/server-discovery-and-monitoring/tests/unified/logging-replicaset.yml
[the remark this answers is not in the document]
prestonvasquez: The Go Driver will also begin monitoring immediately, but the server [updates](https://github.com/mongodb/mongo-go-driver/blame/master/x/mongo/driver/topology/server.go#L209) are done concurrently. This is why there is a conflict between waiting for the event and starting the connection -- i.e. it is possible that the test-runner code for awaiting events starts before monitoring. 

This is _essentially_ the command that the Go Driver will construct for non-legacy hello operations:

```js
db.adminCommand({
  "isMaster": 1,
  "maxAwaitTimeMS": 10000,
  "topologyVersion": {
    "processId": ObjectId("6488dd97c4627cd040311e50"),
    "counter": NumberLong("421")
  },
  "internalClient": {
    "minWireVersion": 17,
    "maxWireVersion": 17
  },
  "$db": "admin"
})
```

This command will always wait for 10 seconds before receiving a response from the server. So since we are awaiting the heartbeat succeeded event, this test will take minimally 10 seconds to complete. Does that sounds right?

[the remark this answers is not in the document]
ShaneHarvey: The first heartbeat is supposed to be the initial connection handshake. Does Go ignore that event?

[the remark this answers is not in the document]
prestonvasquez: Thank you both, this was due to a bug in the Go Driver.

source/server-discovery-and-monitoring/tests/unified/logging-standalone.json
```diff
@@ -0,0 +1,522 @@
...
+            },
+            {
+              "level": "debug",
+              "component": "topology",
+              "data": {
+                "message": "Server heartbeat succeeded"
+              }
+            }
+          ],
+          "messages": [
```
isabelatkinson, line 479: The expected messages assertion here is failing in Rust because we see multiple heartbeat failed events I think this is reasonable/expected behavior: the failpoint is always on, so as the monitor continues to send `hello`s/listen for `hello` replies and emit heartbeat failed events when they fail. I can make the test work either by adding `ignoreExtraMessages` here or changing the fail point mode above to `{ "times": 2 }` (2 instead of 1 to account for the retry) rather than always on. Does either of those sound like fine change here, or am I missing something?

source/server-discovery-and-monitoring/tests/unified/logging-loadbalanced.json
```diff
@@ -0,0 +1,150 @@
+{
+  "description": "standalone-logging",
```
isabelatkinson, line 2:
```suggestion
  "description": "loadbalanced-logging",
```

source/server-discovery-and-monitoring/tests/unified/logging-loadbalanced.yml
```diff
@@ -0,0 +1,74 @@
+description: "standalone-logging"
```
isabelatkinson, line 1:
```suggestion
description: "loadbalanced-logging"
```

isabelatkinson (commented):
Tests are passing in Rust except two issues in the standalone logging file; I've left suggestions in comments.

source/server-discovery-and-monitoring/tests/unified/logging-standalone.json
BorisDog: Good catch, thanks. Added `ignoreExtraMessages`.

source/server-discovery-and-monitoring/tests/unified/logging-standalone.json
```diff
@@ -0,0 +1,522 @@
...
+                    "int",
+                    "long"
+                  ]
+                },
+                "driverConnectionId": {
+                  "$$exists": true
+                },
+                "serverConnectionId": {
+                  "$$exists": true
+                }
```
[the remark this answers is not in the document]
BorisDog: Removed.

source/server-discovery-and-monitoring/tests/unified/logging-standalone.json
```diff
@@ -0,0 +1,520 @@
...
+                    "int",
+                    "long"
+                  ]
+                },
+                "driverConnectionId": {
+                  "$$exists": true
+                },
+                "serverConnectionId": {
+                  "$$exists": true
+                },
```
isabelatkinson, line 503: This check (and ditto in the "Failing heartbeat" tests for other topologies) is flaky for me in CI. I *think* it's due to a race between setting the failpoint for hellos and establishing a monitoring connection. If a monitor establishes a connection first, the sequence of events will look something like this:

1. Send a hello for initial connection establishment. Emit a **"Server heartbeat started"** log message with no serverConnectionId, as we haven't talked to the server yet.
2. The server replies with a successful hello response, including a serverConnectionId. Emit a **"Server heartbeat succeeded"** message containing the serverConnectionId from the response.
3. The failpoint for hellos is set.
4. Begin a new server heartbeat and emit a **"Server heartbeat started"** log message with the serverConnectionId stored from the previous successful response.
5. The heartbeat fails. Emit a **"Server heartbeat failed"** log message with the serverConnectionId stored from the previous successful response.

However, if the failpoint is set before the monitoring connection is established, the following will occur:
1. The failpoint for hellos is set.
2. Send a hello for initial connection establishment. Emit a **"Server heartbeat started"** log message with no serverConnectionId, as we haven't talked to the server yet.
3. The heartbeat fails. Emit a **"Server heartbeat failed"** log message. The serverConnectionId will not be present because we do not have a stored value from a previous successful heartbeat.

Let me know if this makes sense to you. If so, I think we'll need to remove this check from each "Failing heartbeat" test because of the previously discussed discrepancy between drivers regarding when heartbeat events begin to be emitted.

BorisDog: Yes this makes sense, thanks for catching this @isabelatkinson.
Could you please confirm that tests are stable after removing `serverConnectionId`?

isabelatkinson: Yes, I get a clean patch with the ID removed from all of the failing heartbeat tests.

isabelatkinson (approved):
lgtm!

source/server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst
```diff
@@ -388,26 +399,292 @@ the value to the default read preference, ``primary``, or treat the call as if `
...
+
+   * - serverConnectionId
+     - Heartbeat-related log messages
+     - Int
+     - The server's ID for the monitoring connection, if known. This value will be unknown and can be omitted in certain cases, e.g. the first
+       "heartbeat started" message for a monitoring connection. Only present on server versions 4.2+.
+
+"Starting Topology Monitoring" Log Message
+------------------------------------------
+This message MUST be published under the same circumstances as a ``TopologyOpeningEvent`` as detailed in `Events API <#events-api>_`.
```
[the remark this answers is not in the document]
BorisDog: All these syntactic changes done.

BorisDog:
> some initial comments - so far, mostly a bunch of nits around adding comments to help readers understand the YAML tests. also, did we decide against adding any logs that don't correspond to existing SDAM events? I remember discussing with @jyemin that Java had some possibly useful SDAM logs we could consider including.

Will be done in https://jira.mongodb.org/browse/DRIVERS-2676.

prestonvasquez (changes requested):

source/server-discovery-and-monitoring/tests/unified/logging-loadbalanced.json
```diff
@@ -0,0 +1,150 @@
+{
+  "description": "loadbalanced-logging",
+  "schemaVersion": "1.14",
```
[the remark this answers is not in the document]
BorisDog: Tnx, done.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-logging-and-monitoring.rst
ShaneHarvey: Was this issue addressed? It still looks like a new file to me in the PR diff

BorisDog: Yes it had the original naming for a while, and I renamed it back today.
LMK if I need to rename it temporary back.

prestonvasquez (approved):
LGTM!

source/unified-test-format/unified-test-format.rst
```diff
@@ -1177,7 +1177,7 @@ The structure of each object is as follows:
   `CMAP
   <../connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#events>`__
   events, and ``sdam`` for `SDAM
-  <../server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst#events>`__
+  <../server-discovery-and-monitoring/server-discovery-and-monitoring-logging-and-monitoring.rst#events>`__
```
ShaneHarvey, line 1180: @BorisDog should we remain all this file references too?

BorisDog: The intention is to rename `server-discovery-and-monitoring-monitoring.rst `to `server-discovery-and-monitoring-logging-and-monitoring.rst` as last step of the PR. So the references are updated to reflect that.

BorisDog: I have failed to force github to recognize the renaming within the PR (need to ensure it's recognized in git history).
The naming follows [command-logging-and-monitoring](https://github.com/mongodb/specifications/blob/master/source/command-logging-and-monitoring/command-logging-and-monitoring.rst) change where `logging` was added to `monitoring`. I see a value in having consistent naming across the spec, as long as it does not introduce unnecessary overhead.

Side note: For me `monitoring-monitoring` suffix has been a bit confusing, so longer name with `logging` is bit clearer :)

ShaneHarvey (approved):

source/unified-test-format/unified-test-format.rst
BorisDog: Reverted all renaming residue. Due to non-trivial commit history in this PR, renaming has to be considered in a separate ticket.

PR #1421, opened 2023-05-16
DRIVERS-2416 OIDC: Automatic token acquisition for Azure Identity Provider

blink1073:

- [ ] Update changelog.

### DIFF
--- a/source/auth/auth.rst
+++ b/source/auth/auth.rst
@@ -1252,7 +1252,10 @@ mechanism
 mechanism_properties
     PROVIDER_NAME
         Drivers MUST allow the user to specify a name for using a service
-        to obtain credentials that is one of ["aws"].
+        to obtain credentials that is one of ["aws", "azure"].
+    TOKEN_AUDIENCE
+        The authorization token audience claim, needed for the "azure"
+        provider.
     REQUEST_TOKEN_CALLBACK
         Drivers MUST allow the user to specify a callback of the form
         "onRequest" (defined below), if the driver supports
@@ -1349,8 +1352,8 @@ the driver MUST raise an error.
 Supported Service Providers
 ```````````````````````````
 
-Drivers MUST support obtaining credentials for a service for "aws", given
-by the PROVIDER_NAME mechanism property.  In all cases the acquired token
+Drivers MUST support obtaining credentials for a service for "aws" and "azure",
+given by the PROVIDER_NAME mechanism property.  In all cases the acquired token
 will be given as the ``jwt`` argument and the JwtStepRequest MUST be made immediately, as part of speculative authentication if appropriate, skipping the PrincipalStepRequest.
 Drivers MUST raise an error if both a PROVIDER_NAME and username are
 given, since using a service will not use the username.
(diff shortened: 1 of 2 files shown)


source/auth/auth.rst
```diff
@@ -1364,6 +1367,21 @@ interpret it as a file path.  The contents of the file are read as the
...
+When the PROVIDER_NAME mechanism property is set to "azure", the driver MUST
+attempt to retrieve a credential from the Azure Instance Metadata Service (IMDS).
+The driver MUST ensure that the TOKEN_AUDIENCE mechanism property is set when
+the PROVIDER_NAME is "azure".
+
+The driver MUST use the same procedure given in `Obtaining an Access Token for Azure Key Vault`_ to
+obtain an access token, with the one exception of using the TOKEN_AUDIENCE as
+the ``resource`` parameter.  The azure credentials MUST be cached by
+TOKEN_AUDIENCE, and expire within 5 minutes of the time given by the
+``expires_in`` parameter of the IMDS response.
```
durran, line 1381: I think we should call out that the `resource=https://vault.azure.net/` query parameter in the URI generated in the other link should be changed to `resource=api://${AZUREOIDC_CLIENTID}` and may in some cases need toURL encode the `://` characters.

durran (changes requested):

source/auth/tests/mongodb-oidc.rst
```diff
@@ -141,6 +142,83 @@ Allowed Hosts Ignored
 - Assert that a ``find`` operation succeeds.
 - Close the client.
 
+Azure Automatic Auth
+==================
+
+Drivers MUST be able to authenticate using the "azure" provider workflow, using
+an Azure VM provisioned using the helper scripts in Drivers Evergreen Tools.
+These tests will most likely need to be run in a separate test file from the
+rest of the tests, to avoid needing to skip multiple tests.
```
durran, line 151: Maybe it's worth noting here that the URI for all tests is the same, and must be equal to:

`mongodb://localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=PROVIDER_NAME:azure,TOKEN_AUDIENCE:api://${AZUREOIDC_CLIENTID}`

Then each individual test can reuse the same URI. Thoughts? I did the same as Python and just set it as the `MONGODB_URI` and had all the clients use that.

source/auth/auth.rst
```diff
@@ -1252,7 +1252,10 @@ mechanism
 mechanism_properties
     PROVIDER_NAME
         Drivers MUST allow the user to specify a name for using a service
-        to obtain credentials that is one of ["aws"].
+        to obtain credentials that is one of ["aws", "azure"].
+    TOKEN_AUDIENCE
```
durran, line 1256: Question here that would negate my other questions. Will the resource we pass always be `api://${AZUREOIDC_CLIENTID}`? I'm curious if this mech property can be removed and we just read `AZUREOIDC_CLIENTID` from the environment and construct the resource parameter we send to the Azure endpoint.

blink1073: I wrestled with whether this should be an env variable or a mech property.  The GCP integration needs a similar value.   The user would have to set the environment variable on their application, so we'd want to choose something that wouldn't conflict with other values, perhaps `MONGODB_OIDC_TOKEN_AUDIENCE`?

image

durran: Ah I guess if the user needs to manually set it then I'm less worried about it being in the URI vs an environment variable. So I think URI is fine then.

source/auth/auth.rst
blink1073: I think it would be more accurate to say that the audience itself is an encoded URI, most likely an Application ID URI of the form `app://<app id>`.

source/auth/tests/mongodb-oidc.rst
```diff
@@ -141,6 +142,83 @@ Allowed Hosts Ignored
...
+- Close the client.
+- Assert that the main OIDC cache is empty.
+
+Azure Cache is Used
+~~~~~~~~~~~~~~~~~~~
+- Clear the Azure OIDC cache.
+- Create a client with a url of the form ``mongodb://localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=PROVIDER_NAME:azure,TOKEN_AUDIENCE:<foo>``.
+- Assert that a ``find`` operation succeeds.
+- Close the client.
+- Assert that the Azure OIDC cache has one entry.
```
JamesKovacs, line 180: Why not combine the `Main Cache Not Used` and `Azure Cache is Used` tests? Clear both caches, perform the operation, and check that the main cache remains empty and the Azure one has a single entry.

JamesKovacs (approved):
LGTM

blink1073:
Closing since this will have to be re-worked to use the machine callback.

PR #1423, opened 2023-05-17
DRIVERS-2549: add search index management helpers

baileympearson:
This PR adds a new search management API to the index management spec.

Node POC: https://github.com/mongodb/node-mongodb-native/pull/3672

E2E testing of the new helpers against atlas will be implemented in [DRIVERS-2630](https://jira.mongodb.org/browse/DRIVERS-2630).

- [x] Update changelog.


source/index-management/index-management.rst
```diff
@@ -852,6 +852,181 @@ Example::
...
+
+    /**
+    * Drops the search index with the given name.
+    */
+    dropSearchIndex(name: String, options: Optional<DropSearchIndexOptions>): void;
+
+    /**
+    * Gets index information for all search indexes in the collection.
+    */
+    listSearchIndexes(name: Optional<String>, aggregationOptions: Optional<AggregationOptions>, listIndexOptions: Optional<ListSearchIndexOptions>): Cursor<{ name: String }>;
```
durran, line 961: Are we not allowing `id` to be specified?

source/index-management/index-management.rst
```diff
@@ -852,6 +852,181 @@ Example::
...
+     *   method. See below.
+     */
+    iterator(): Iterator<{ name: string }>;
+
+    /**
+     * For drivers that cannot make SearchIndexView iterable, they MUST implement this method to
+     * return a list of indexes. In the case of async drivers, this MAY return a Future<Cursor>
+     *  or language/implementation equivalent.
+     */
+    list(): Cursor<{ name: String }>;
```
durran, line 995: Pending the above comment on the return value from the agg pipeline.. Even though the original view api `list()` doesn't take arguments, I think it makes sense here to allow the same optional arguments as `listSearchIndexes()`, if it's confirmed the agg pipeline result can return the full definitions.

source/index-management/index-management.rst
```diff
@@ -852,6 +852,181 @@ Example::
...
+     * implemented as described in the :ref:`Enumerate Indexes` section, although the naming
+     * requirement is dropped in favor of the driver language standard for handling iteration
+     * over a sequence of objects.
+     *
+     * @see https://github.com/mongodb/specifications/blob/master/source/enumerate-indexes.rst
+     *
+     * @note For drivers that cannot make the IndexView iterable, they MUST implement a list
+     *   method. See below.
+     */
+    iterator(): Iterator<{ name: string }>;
```
durran, line 988: Is this going to only return index names or a list of the names and full definitions?

source/index-management/index-management.rst
```diff
@@ -852,6 +852,181 @@ Example::
...
+     *
+     * @note Drivers MAY opt to implement this method signature, the signature that
+     *   takes an IndexModel as a parameter, or for those languages with method
+     *   overloading MAY decide to implement both.
+     *
+     * @note Drivers MAY combine the two options types into a single one. If the options are
+     *   explicitly typed, the combined options type MUST be named CreateOneIndexOptions or an acceptable
+     *   variation.
+     */
+    createOne(definition: SearchIndexDefinition, options: CreateSearchIndexOptions): String;
```
durran, line 1010: I think you want to specify `name: string, definition: Document` and provide an optional overload with a `SearchIndexModel` just as the standard API allows. `SearchIndexDefinition` is not defined in the spec changes, and I think that's intentional for future-proofing. If not, we should define what those fields are and replace the usages of `Document` in the changes. The syntax doc looks like:

```js
{
  analyzer: "<analyzer for index>",
  searchAnalyzer: "<analyzer for query>",
  analyzers: [<custom analyzer>],
  storedSource: [<boolean | stored source definition>],
  synonyms: [<synonym mapping>],
  mappings: {
    dynamic: true
  }
}
```

source/index-management/index-management.rst
```diff
@@ -852,6 +852,181 @@ Example::
...
+
+    /**
+     * Creates multiple search indexes in the collection.
+     *
+     * @return The names of the created indexes.
+     */
+    createMany(models: Iterable<SearchIndexDefinition>, options: CreateSearchIndexOptions): Iterable<String>;
```
durran, line 1017: Same comment for `SearchIndexDefinition`.

source/index-management/index-management.rst
```diff
@@ -852,6 +852,181 @@ Example::
...
+  interface Collection {
+    /**
+    * Convenience method for creating a single search index.
+    * 
+    * @return The name of the created search index
+    * @note Drivers MAY opt to implement this method signature, the signature that
+    *   takes an SearchIndexModel as a parameter, or for those languages with method
+    *   overloading MAY decide to implement both.
+    */
+    createSearchIndex(name: String, definition: Document, options: Optional<CreateSearchIndexOptions>): String;
```
durran, line 927: Let's make this one optional and the other that takes a model required. Otherwise the unified tests, which use the method that takes a model would not align with drivers choosing to implement this one.

source/index-management/tests/createSearchIndex.yml
```diff
@@ -0,0 +1,62 @@
...
+
+tests:
+  - description: "no name provided for an index definition"
+    operations:
+      - name: createSearchIndex
+        object: *collection0
+        arguments:
+          definition: { definition: &definition { mappings: { dynamic: true } }  }
+        expectError:
+          isError: true
```
durran, line 31: I think it makes sense in all the new unified test yaml files to add comments around all the operations expecting errors. Just to note the reasoning - these command don't currently exist outside of Atlas, and we will be updating the unified tests in the future once we have an Atlas instance to test against. I'd also link the new drivers ticket for that work.

durran (changes requested):

source/index-management/index-management.rst
baileympearson: Correct, you can see this slack thread for context: https://mongodb.slack.com/archives/C04TMC263M3/p1683837096346969?thread_ts=1683836146.256049&cid=C04TMC263M3

source/index-management/index-management.rst
baileympearson: Does it matter?  I assumed that each driver would transform the unified test operation into the correct format for their driver.  In Node, we could accommodate either:

```typescript

// index model API
operations.set('createSearchIndex', async ({ entities, operation }) => {
  const collection: Collection<any> = entities.getEntity('collection', operation.object);
  const { definition } = operation.arguments ?? {};
  return collection.createSearchIndex(definition);
});

// name + definition api
operations.set('createSearchIndex', async ({ entities, operation }) => {
  const collection: Collection<any> = entities.getEntity('collection', operation.object);
  const { definition: arguments } = operation.arguments ?? {};
  const { definition, name } = arguments ?? {};
  return collection.createSearchIndex(name, definition);
});
```

If we do end up supporting both, then maybe we should adjust the unified tests to be a bit more clearly accommodating of either approach, something like

```yaml
- operation: createSearchIndex
  arguments: 
    - indexName: <string>
    - indexDefinition: <string>
```

source/index-management/index-management.rst
baileympearson: I updated the references to `SearchIndexDefinition` to be `Document` instead, that was a mistake.  

I did intentional leave the definition as a `Document` for future compatibility.

source/index-management/index-management.rst
baileympearson: It can return the full definitions, I'll make that update.

I added the options to the `searchIndexView` method instead of the list method, because not every driver is required to have an iterator or a list method, since the index view itself is iterable.

source/index-management/index-management.rst
durran: I think that's fine with having the unified test suite modify the parameters for their driver. But I'd say we call that out explicitly somewhere.

durran (changes requested):
Just last 2 open comments.

source/index-management/index-management.rst
```diff
@@ -852,6 +852,186 @@ Example::
...
+--------------------------------------
+
+There are currently no supported options for any of the search index management commands.  To future proof
+drivers implementations so that any options added in the future do not constitute a breaking change to drivers,
+empty options structs have been added as placeholders.  If a driver's language has a mechanism to add options 
+in a non-breaking manner (i.e., method overloading) drivers MAY omit the empty options structs from their
+search index management helpers.
+
+``listSearchIndexes`` is implemented using an aggregation pipeline.  The list helper MUST support a driver's aggregation
+options as outline in the `crud specification <https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#read>`_.  Drivers MAY combine the aggregation options with
```
jmikola, line 873:
```suggestion
options as outline in the `CRUD specification <https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#read>`_.  Drivers MAY combine the aggregation options with
```

Acronym is capitalized.

source/index-management/index-management.rst
```diff
@@ -910,3 +1090,4 @@ Changelog
 :2022-10-05: Remove spec front matter and reformat changelog.
 :2023-5-10:  Merge index enumeration and index management specs and get rid of references 
              to legacy server versions.
+:2023-5-18:  Add the search index management API.
```
jmikola, line 1093:
```suggestion
:2023-05-10:  Merge index enumeration and index management specs and get rid of references 
             to legacy server versions.
:2023-05-18:  Add the search index management API.
```

For consistency with previous changelog entries.

source/index-management/tests/createSearchIndex.yml
```diff
@@ -0,0 +1,62 @@
+description: "createSearchIndex convenience helper"
```
jmikola, line 1:
```suggestion
description: "createSearchIndex"
```

This isn't enforced, but most unified spec tests keep the base filename and top-level description consistent.

source/index-management/tests/createSearchIndex.yml
jmikola: If these commands don't exist outside of Atlas, do we need new runOnRequirement syntax to restrict these tests to an Atlas environment?

I don't see anything in the PR that says these tests should only be run on Atlas (f.e. some language in the [test README](https://github.com/mongodb/specifications/blob/master/source/index-management/tests/README.rst)).

source/index-management/tests/createSearchIndex.yml
```diff
@@ -0,0 +1,62 @@
...
+        events:
+          - commandStartedEvent:
+              command:
+                createSearchIndexes: *collection0
+                indexes: [ { definition: *definition, name: 'test index' } ]
+                $db: *database0
+
+    
+
+
```
jmikola, line 66:
```suggestion
```

source/index-management/tests/createSearchIndex.yml
```diff
@@ -0,0 +1,62 @@
...
+    topologies: [ replicaset, load-balanced, sharded ]
+    serverless: forbid
+
+tests:
+  - description: "no name provided for an index definition"
+    operations:
+      - name: createSearchIndex
+        object: *collection0
+        arguments:
+          definition: { definition: &definition { mappings: { dynamic: true } }  }
```
jmikola, line 29:
```suggestion
          definition: { definition: &definition { mappings: { dynamic: true } } }
```

source/index-management/tests/createSearchIndexes.yml
```diff
@@ -0,0 +1,82 @@
...
+        events:
+          - commandStartedEvent:
+              command:
+                createSearchIndexes: *collection0
+                indexes: [ { definition: *definition, name: 'test index' } ]
+                $db: *database0
+
+    
+
+
```
jmikola, line 88:
```suggestion
```

source/index-management/tests/createSearchIndex.yml
```diff
@@ -0,0 +1,62 @@
+description: "createSearchIndex convenience helper"
+schemaVersion: "1.10"
```
jmikola, line 2: What about these test files require 1.10? That was when the `thread` entity was introduced (see: [changelog](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#changelog)).

Per [schemaVersion](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#schemaversion):

This plays nicely with CI, which uses the test file's `schemaVersion` field for validation.

jmikola (commented):
Just leaving some drive-by feedback, which you can do with as you wish.

source/index-management/tests/createSearchIndex.yml
durran: Part 2 of this ticket, which will run the tests against Atlas, would most likely address that. We don't run on Atlas for the first set of tests, which is simply to ignore the errors. We broke it into 2 parts to unblock shell faster.

durran: See: https://jira.mongodb.org/browse/DRIVERS-2630

source/index-management/index-management.rst
```diff
@@ -852,6 +852,186 @@ Example::
...
+    // Returns the definition for this index.
+    definition: Document;
+
+    // Returns the name for this index, if present.
+    name: Optional<string>;
+  }
+
+  /**
+   * The following interfaces are empty but are provided as placeholders for drivers that cannot 
+   * add options in a non-breaking manner, if options are added in the future.
```
kevinAlbs, line 907: Is `Comment` supported on the new search commands? Has adding `Comment` to the options been considered?

source/index-management/index-management.rst
```diff
@@ -852,6 +852,186 @@ Example::
...
+    await setTimeout(1000);
+  }
+ 
+Common Interfaces
+-----------------
+
+.. code:: typescript
+
+  interface SearchIndexModel {
+    // Returns the definition for this index.
```
kevinAlbs, line 898:
```suggestion
    // The definition for this index.
```

`Returns` suggests `definition` is a method.

source/index-management/index-management.rst
baileympearson: Comment is not supported (just the first comment in [this thread](https://mongodb.slack.com/archives/C04TMC263M3/p1683836146256049) for context).

source/index-management/tests/dropSearchIndexes.yml
kevinAlbs, line 1: Suggest renaming file to `dropSearchIndex.yml` to match operation name.

kevinAlbs (approved):
LGTM with one suggested rename. Tests pass in C driver.

durran (approved):

source/index-management/tests/createSearchIndex.yml
jmikola: We do not. I'd totally be down for integrating a YAML linter into our CI pipeline, though.

PR #1424, opened 2023-05-19
DRIVERS-1671: Add server selection log message and tests

BorisDog:
Based on the [original PR](https://github.com/mongodb/specifications/pull/1341)

- [x] Update changelog.


isabelatkinson (approved):
lgtm! the updated tests all pass in Rust

source/server-selection/server-selection.rst
```diff
@@ -859,13 +861,14 @@ For multi-threaded clients, the server selection algorithm is as follows:
    ``operationCount``. If both servers have the same ``operationCount``, select
    arbitrarily between the two of them.
 
-8. Increment the ``operationCount`` of the selected server and return it. Do not
-   go onto later steps.
+8. Increment the ``operationCount`` of the selected server and return it. Log a
+   `"Server selection succeeded" message`_.  Do not go onto later steps.
 
-9. Request an immediate topology check, then block the server selection thread
+9.  Request an immediate topology check, then block the server selection thread
```
vector-of-bool, line 867:
```suggestion
9. Request an immediate topology check, then block the server selection thread
```

source/server-selection/server-selection.rst
```diff
@@ -859,13 +861,14 @@ For multi-threaded clients, the server selection algorithm is as follows:
...
    until the topology changes or until the server selection timeout has elapsed
 
-10. If server selection has timed out, raise a `server selection error`_
+10. If server selection has timed out, raise a `server selection error`_ and log
+   a `"Server selection failed" message`_.
```
vector-of-bool, line 871:
```suggestion
10. If server selection has timed out, raise a `server selection error`_ and log
    a `"Server selection failed" message`_. 
```

source/server-selection/server-selection.rst
```diff
@@ -917,17 +922,19 @@ as follows:
...
-   goto Step #4
+1. If `serverSelectionTryOnce`_ is true and the last scan time is newer than
+   the selection start time, raise a `server selection error`_ and log a
+   `"Server selection failed" message`_; otherwise, log a `"Waiting for suitable
+   server to become available" message`_ if one has not already been logged for
+   this operation, and goto Step #4
 
-10. If the current time exceeds the maximum time, raise a
-    `server selection error`_
+1.  If the current time exceeds the maximum time, raise a
```
vector-of-bool, line 934:
```suggestion
10. If the current time exceeds the maximum time, raise a
```

source/server-selection/server-selection.rst
```diff
@@ -917,17 +922,19 @@ as follows:
...
    suitable servers based on the latency window`_ and return one at random from
-   the filtered servers; otherwise, mark the topology stale and continue to step
-   #9.
+   the filtered servers, and log a `"Server selection succeeded" message`_.;
+   otherwise, mark the topology stale and continue to step #9.
 
-9. If `serverSelectionTryOnce`_ is true and the last scan time is newer than
-   the selection start time, raise a `server selection error`_; otherwise,
-   goto Step #4
+1. If `serverSelectionTryOnce`_ is true and the last scan time is newer than
```
vector-of-bool, line 928:
```suggestion
9. If `serverSelectionTryOnce`_ is true and the last scan time is newer than
```

source/server-selection/server-selection.rst
```diff
@@ -917,17 +922,19 @@ as follows:
...
+   server to become available" message`_ if one has not already been logged for
+   this operation, and goto Step #4
 
-10. If the current time exceeds the maximum time, raise a
-    `server selection error`_
+1.  If the current time exceeds the maximum time, raise a
+    `server selection error`_ and log a `"Server selection failed" message`_.
 
-11. Goto Step #4
+1.  Goto Step #4
```
vector-of-bool, line 937:
```suggestion
11. Goto Step #4
```

source/server-selection/server-selection.rst
```diff
@@ -1180,11 +1187,198 @@ server without an explicit ``start_request`` (or comparable) method call.
 Outside a legacy "request" API, drivers MUST use server selection for each
 individual read operation.
 
-Reference Implementation
-========================
+Logging
+-------
+Please refer to the `logging specification <../logging/logging.rst>`_ for
```
vector-of-bool, line 1192:
```suggestion
Please refer to the `logging specification <../logging/logging.rst>`__ for
```

source/server-selection/server-selection.rst
```diff
@@ -1180,11 +1187,198 @@ server without an explicit ``start_request`` (or comparable) method call.
...
+     - Value
+
+   * - message
+     - String
+     - "Server selection failed"
+
+   * - failure
+     - Flexible
+     - Representation of the error the driver will throw regarding server selection failing. The type and format of this
+       value is flexible; see the `logging specification <../logging/logging.rst#representing-errors-in-log-messages>`_
```
vector-of-bool, line 1330:
```suggestion
       value is flexible; see the `logging specification <../logging/logging.rst#representing-errors-in-log-messages>`__
```

vector-of-bool (approved):
LGTM!

My change suggestions are only rST syntax fixes.

source/unified-test-format/unified-test-format.rst
```diff
@@ -3987,7 +4002,8 @@ Changelog
 
 ..
   Please note schema version bumps in changelog entries where applicable.
-
+:2023-05-26: **Schema version 1.14.**
```
BorisDog, line 4005: @jmikola in case you'd like to review this schema change.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1377,6 +1379,19 @@ The structure of this object is as follows:
...
+- ``topologyDescriptionChangedEvent``: Optional object. Assertions for one or more
+  `TopologyDescriptionChangedEvent <../server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst#events>`__ fields.
+
+  The structure of this object is as follows:
+
+  - ``previousDescription``: Optional object. A value corresponding to the topology
+    description as it was before the change that triggered this event.
+
+  - ``newDescription``: Optional object. A value corresponding to the topology
+    description as it was after the change that triggered this event.
```
jmikola, line 1393: These fields are never used in the tests accompanying this change. Moreover, the spec doesn't address _how_ to compare these values. Contrast this with the existing rules for [serverDescriptionChangedEvent](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#expectedsdamevent), which defines a single `type` field that can be compared. There, `type` is also defined in the schema.

Using `object` types for these fields in the schema will make it difficult to distinguish tests that start asserting on these values, and without a schema bump we wouldn't necessarily know that new test runner rules were necessary. To that end, `object` should be used sparingly (we use that for operation arguments, but there is also guidance on raising errors for unsupported arguments).

With that in mind, I'd suggest removing the following from the schema:

```json
"properties": {
    "previousDescription": { "type": "object" },
    "newDescription": { "type": "object" }
}
```

When a test actually needs to start asserting on topology descriptions, we can start modeling fields like was done for ServerDescriptionChangedEvent.

It looks like you only need `topologyDescriptionChangedEvent` to exist for the purpose of [waitForEvent](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#waitforevent). If so, the prior art would be [expectedCmapEvents](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.rst#expectedcmapevent), most of which are modeled as empty objects.

BorisDog: Thanks, done.

source/unified-test-format/unified-test-format.rst
```diff
@@ -1377,6 +1379,11 @@ The structure of this object is as follows:
     <../server-discovery-and-monitoring/server-discovery-and-monitoring.rst#servertype>`__
     for a list of valid values.
 
+.. _expectedEvent_topologyDescriptionChangedEvent:
+
+- ``topologyDescriptionChangedEvent``: Optional object. If present, this object MUST be an
+  empty document as all fields in this event are non-deterministic.
```
jmikola, line 1385:
```suggestion
- ``topologyDescriptionChangedEvent``: Optional object. If present, this object
  MUST be an empty document as no assertions are currently supported for
  `TopologyDescriptionChangedEvent <../server-discovery-and-monitoring/server-discovery-and-monitoring-monitoring.rst#events>`__ fields.
```

"all fields in this event are non-deterministic" is specific to some of the CMAP events but it's not applicable to TopologyDescriptionChangedEvent. This text is more accurate.

jmikola (approved):
`unified-test-format.rst` LGTM with the suggestion for `topologyDescriptionChangedEvent` applied.

Also, don't forget about @vector-of-bool's RST suggestions.

PR #1425, opened 2023-05-22
DRIVERS-2617 Add prose test for change stream splitting

abr-egn:
- [✓] Update changelog.
- [✓] Test changes in at least one language driver.
- [✓] Test these changes against all server versions and topologies (including standalone, replica set, sharded clusters, and serverless).

DRIVERS-2617

This adds a simple test that validates that drivers handle the large event splitting properly.  PoC implementation for Rust can be seen [here](https://github.com/mongodb/mongo-rust-driver/pull/878).

### DIFF
--- a/source/change-streams/change-streams.rst
+++ b/source/change-streams/change-streams.rst
@@ -1000,6 +1000,7 @@ Reference Implementations
 Changelog
 =========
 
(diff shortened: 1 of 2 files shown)


ShaneHarvey (approved):
LGTM! Test works in locally Python. Running an evergreen patch here: https://github.com/mongodb/mongo-python-driver/pull/1232

PR #1427, opened 2023-05-31
DRIVERS-2641: Allow int32 or int64 for getMore cursor ID

jmikola:
This accommodate drivers that encode BSON integers according to their range and is consistent with what is done in most other spec tests.


nbbeeken (approved):
Node ✅: https://github.com/mongodb/node-mongodb-native/pull/3693 (I sifted through the unrelated red)

JamesKovacs (approved):
LGTM

PR #1435, opened 2023-06-16
DRIVERS-2612: Require 7.0+ for explain apiVersion bug fix

jmikola:
In previous server versions, apiVersion is not properly appended to the root of an explain command document.

When testing with requireApiVersion=1, that can cause test failures if the driver does not have additional logic to always append the field to a crypt_shared/mongocryptd respnse.

POC: https://github.com/mongodb/mongo-php-library/pull/1111

As noted in [DRIVERS-2612](https://jira.mongodb.org/browse/DRIVERS-2612?focusedCommentId=5502676&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-5502676), this is blocked on the release of 7.0.0-rc4 for [BACKPORT-16190](https://jira.mongodb.org/browse/BACKPORT-16190).

### DIFF
--- a/source/client-side-encryption/tests/legacy/explain.yml
+++ b/source/client-side-encryption/tests/legacy/explain.yml
@@ -1,5 +1,5 @@
 runOn:
-  - minServerVersion: "4.1.10"
+  - minServerVersion: "7.0.0"
 database_name: &database_name "default"
 collection_name: &collection_name "default"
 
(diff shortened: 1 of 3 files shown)


source/client-side-encryption/tests/legacy/explain.yml
```diff
@@ -1,5 +1,5 @@
 runOn:
-  - minServerVersion: "4.1.10"
+  - minServerVersion: "7.0.0"
```
kevinAlbs, line 2:
```suggestion
  # Require server version 7.0 to include fix of SERVER-77040
  - minServerVersion: "7.0.0"
```

kevinAlbs (approved):
Thank you. LGTM

jmikola:
Planning to come back to this Thursday, as the PHPLIB PR is currently blocked on an unrelated issue.

PR #1436, opened 2023-06-21
DRIVERS-2452 Add database name to the "command failed" and "command succeeded" log messages

qingyang-hu:

- [x] Update changelog.

[DRIVERS-2452](https://jira.mongodb.org/browse/DRIVERS-2452)
Add the database name to the "command failed" and "command succeeded" log messages.

### DIFF
--- a/source/command-logging-and-monitoring/tests/logging/command.yml
+++ b/source/command-logging-and-monitoring/tests/logging/command.yml
@@ -52,13 +52,14 @@ tests:
             component: command
             data:
               message: "Command succeeded"
+              databaseName: *databaseName
               commandName: *commandName
               reply: { $$type: string }
               requestId: { $$type: [int, long] }
               serverHost: { $$type: string }
               serverPort: { $$type: [int, long] }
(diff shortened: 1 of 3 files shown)


source/command-logging-and-monitoring/tests/logging/command.yml
```diff
@@ -52,13 +52,14 @@ tests:
             component: command
             data:
               message: "Command succeeded"
+              databaseName: *databaseName
```
jmikola, line 55: Noted that the schema defines `data` as an `object` ([here](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/schema-1.13.json#L563)), so a new field here has no impact on the schema version.



Check unified test format files against schema / build (pull_request)

jmikola (approved):
Take my review with a grain of salt as I haven't worked on the logging spec and don't have an implementation I can use to POC this.

BorisDog (approved):
LGTM. Verified that the test works in .NET.

PR #1437, opened 2023-06-22
Fix error condition for contentionFactor option

jmikola:
I'd consider this a small fix not worthy of a changelog entry but let me know if you disagree.


source/client-side-encryption/client-side-encryption.rst
```diff
@@ -1339,7 +1339,7 @@ NOTE: The Range algorithm is experimental only. It is not intended for public us
 contentionFactor
 ^^^^^^^^^^^^^^^^
 contentionFactor only applies when algorithm is "Indexed" or "RangePreview".
-It is an error to set contentionFactor when algorithm is not "Indexed".
+It is an error to set contentionFactor when algorithm is not "Indexed" or "RangePreview".
```
jmikola, line 1342: Other options were amended to say "or RangePreview", so I think it was just an oversight that `contentionFactor` was not. The prose test for range index certainly uses this option.

addaleax (approved):

PR #1442, opened 2023-07-14
DRIVERS-2630: add e2e testing for search index commands and clarifications to search index spec

baileympearson:
This PR does three things:

- This PR adds prose tests for the search index management helpers
- This PR strengthens assertions on the existing unified tests for the search index management helpers.
- This PR includes some clarifications in the search index management spec.

The third bullet can be separated into a different PR if necessary.

Node POC: https://github.com/mongodb/node-mongodb-native/pull/3736

#### Why have unified tests and prose tests? 

The initial plan was to enhance the unified test runner to support the new search index management commands.  There ended up being complications integrating support for the new commands into the UTR (the largest issue was that the commands are asynchronous, and return from the server before the changes have finished).  The unified tests were already written and expected errors when run against non-Atlas deployments, so they effectively serve as unit tests for drivers without writing another specialized test format.  The new prose tests add end-to-end functionality for drivers' implementations.

#### Why setup/teardown a cluster for each test?

- Drivers already spinup/teardown resources for existing tests (serverless, gcp/azure kms, FAAS testing), so additional CI time was not a concern.
- We have tooling in drivers-evergreen-tools to setup and teardown Atlas clusters so the setup was easy to reuse.
- Provisioning a cluster on-demand reduces the changes for collisions and the need to ensure driver use unique databases / collections / commit hashes to avoid CI runs interfering on the same cluster.

#### CSOT

These commands should work with CSOT as per the CSOT spec, but looking through the specs repo we don't seem to have tests for every driver helper that supports `timeoutMS`.  Consequently, I chose not to add tests for these helpers with `timeoutMS`.

- [x] Update changelog.

### DIFF
--- a/source/index-management/index-management.rst
+++ b/source/index-management/index-management.rst
@@ -414,6 +414,10 @@ Index View API
      * For drivers that cannot make IndexView iterable, they MUST implement this method to
      * return a list of indexes. In the case of async drivers, this MAY return a Future<Cursor>
      *  or language/implementation equivalent.
+     * 
+     *  If drivers are unable to make the IndexView iterable, they MAY opt to provide the options for 
+     *  listing search indexes via the `list` method instead of the `Collection.listSearchIndexes` method.
+
      */
     list(): Cursor;
 
@@ -873,8 +877,8 @@ search index management helpers.
 options as outline in the `CRUD specification <https://github.com/mongodb/specifications/blob/master/source/crud/crud.rst#read>`_.  Drivers MAY combine the aggregation options with
 any future ``listSearchIndexes`` stage options, if that is idiomatic for a driver's language.
 
-Notes
------
+Asynchronicity
+--------------
 
 The search index commands are asynchronous and return from the server before the index is successfully updated, created or dropped.
 In order to determine when an index has been created / updated, users are expected to run the ``listSearchIndexes`` repeatedly
@@ -888,7 +892,33 @@ An example, from Javascript:
   while (!(await collection.listSearchIndexes({ name }).hasNext())) {
     await setTimeout(1000);
   }
- 
+
+Where are read concern and write concern?
+-----------------------------------------
+
+Th
+
+Consistency with Existing APIs
+------------------------------
+
+Drivers SHOULD strive for a search index management API that is as consistent with their existing search index management API as much as possible.
+
+
+NamespaceNotFound Errors
+------------------------
+
+Some drivers suppress NamespaceNotFound errors for CRUD helpers.  Drivers MAY suppress NamespaceNotFound errors from 
+the search index management helpers.
+
+Drivers MUST suppress NamespaceNotFound errors for the ``dropSearchIndex`` helper.  Drop operations should be idempotent:
+
+.. code:: typescript
+  await collection.dropSearchIndex('my-test-index');
+  // subsequent calls should behave the same for the user as the first call
+  await collection.dropSearchIndex('my-test-index');
+  await collection.dropSearchIndex('my-test-index');
+
+
 Common Interfaces
 -----------------
 
@@ -990,6 +1020,9 @@ Index View API for Search Indexes
      * For drivers that cannot make SearchIndexView iterable, they MUST implement this method to
      * return a list of indexes. In the case of async drivers, this MAY return a Future<Cursor>
      *  or language/implementation equivalent.
+     *  
+     *  If drivers are unable to make the SearchIndexView iterable, they MAY opt to provide the options for 
+     *  listing search indexes via the `list` method instead of the `Collection.listSearchIndexes` method.
      */
     list(): Cursor<Document>;
 
--- a/source/index-management/index-management.rst
+++ b/source/index-management/index-management.rst
@@ -914,6 +914,7 @@ the search index management helpers.
 Drivers MUST suppress NamespaceNotFound errors for the ``dropSearchIndex`` helper.  Drop operations should be idempotent:
 
 .. code:: typescript
+
   await collection.dropSearchIndex('my-test-index');
   // subsequent calls should behave the same for the user as the first call
   await collection.dropSearchIndex('my-test-index');
@@ -1005,7 +1006,7 @@ Index View API for Search Indexes
     /**
      * Returns the search index view for this collection.
      */
-    searchIndexes(name: Optional<String>, aggregateOptions: Optional<AggregationOptions>, options: Optional<ListSearchIndexesOptions>): SearchIndexView;
+    searchIndexes(name: Optional<String>, aggregateOptions: Optional<AggregationOptions>, options: Optional<ListSearchIndexOptions>): SearchIndexView;
   }
 
   interface SearchIndexView extends Iterable<Document> {
@@ -1065,7 +1066,7 @@ Index View API for Search Indexes
     /**
      * Updates a single search index from the collection by the index name.
      */
-    updateOne(name: String, options: Optional<UpdateSearchIndexOptions>): Result;
+    updateOne(name: String, definition: Document, options: Optional<UpdateSearchIndexOptions>): Result;
   }
 
 ---------
--- a/source/index-management/index-management.rst
+++ b/source/index-management/index-management.rst
@@ -964,7 +964,7 @@ Standard API for Search Indexes
      *   
      * @note Drivers MAY combine the `indexOptions` with the `createSearchIndexOptions`, if that is idiomatic for their language.
      */
-    createSearchIndex(definition: Document, indexOptions: SearchIndexOptions, createSearchIndexOptions: Optional<CreateSearchIndexOptions>): String;
+    createSearchIndex(definition: Document, indexOptions: Optional<SearchIndexOptions>, createSearchIndexOptions: Optional<CreateSearchIndexOptions>): String;
 
     /**
      * Convenience method for creating a single index.
@@ -1046,7 +1046,7 @@ Index View API for Search Indexes
      *   
      * @note Drivers MAY combine the `indexOptions` with the `createSearchIndexOptions`, if that is idiomatic for their language.
      */
-    createOne(definition: Document, indexOptions: SearchIndexOptions, createSearchIndexOptions: Optional<CreateSearchIndexOptions>): String;
+    createOne(definition: Document, indexOptions: Optional<SearchIndexOptions>, createSearchIndexOptions: Optional<CreateSearchIndexOptions>): String;
 
     /**
      * This is a convenience method for creating a single index.
--- a/source/index-management/index-management.rst
+++ b/source/index-management/index-management.rst
@@ -902,7 +902,7 @@ write concern are not relevant for the search index management commands.
 Consistency with Existing APIs
 ------------------------------
 
-Drivers SHOULD strive for a search index management API that is as consistent with their existing search index management API as much as possible.
+Drivers SHOULD strive for a search index management API that is as consistent with their existing index management API as much as possible.
 
 
 NamespaceNotFound Errors
--- a/source/index-management/index-management.rst
+++ b/source/index-management/index-management.rst
@@ -957,7 +957,7 @@ Standard API for Search Indexes
      *   takes an SearchIndexModel as a parameter, or for those languages with method
      *   overloading MAY decide to implement both.
      */
-    createSearchIndex(name: String, definition: Document, options: Optional<CreateSearchIndexOptions>): String;
+    createSearchIndex(definition: Document, name: Optional<string>, options: Optional<CreateSearchIndexOptions>): String;
 
     /**
      * Convenience method for creating a single index.
@@ -1037,7 +1037,7 @@ Index View API for Search Indexes
      *   takes an SearchIndexModel as a parameter, or for those languages with method
      *   overloading MAY decide to implement both.
      */
-    createOne(name: String, definition: Document, options: Optional<CreateSearchIndexOptions>): String;
+    createOne(definition: Document, name: Optional<string>, options: Optional<CreateSearchIndexOptions>): String;
 
     /**
      * This is a convenience method for creating a single index.
--- a/source/index-management/index-management.rst
+++ b/source/index-management/index-management.rst
@@ -1136,4 +1136,4 @@ Changelog
 :2023-05-10:  Merge index enumeration and index management specs and get rid of references 
              to legacy server versions.
 :2023-05-18:  Add the search index management API.
-:2023-07-17:  Add search index management clarifications.
+:2023-07-27:  Add search index management clarifications.
--- a/source/index-management/index-management.rst
+++ b/source/index-management/index-management.rst
@@ -1128,3 +1128,4 @@ Changelog
 :2023-05-10:  Merge index enumeration and index management specs and get rid of references 
              to legacy server versions.
 :2023-05-18:  Add the search index management API.
+:2023-07-17:  Add search index management clarifications.
--- a/source/index-management/index-management.rst
+++ b/source/index-management/index-management.rst
@@ -896,7 +896,8 @@ An example, from Javascript:
 Where are read concern and write concern?
 -----------------------------------------
 
-Th
+These commands internally proxy the search index management commands to a separate process that runs alongside an Atlas cluster.  As such, read concern and 
+write concern are not relevant for the search index management commands.
 
 Consistency with Existing APIs
 ------------------------------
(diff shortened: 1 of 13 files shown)


GromNaN (approved):

source/index-management/tests/README.rst
```diff
@@ -45,4 +44,157 @@ Tests
...
+#. Assert that the command does not error and the server responds with a success.
+#. Run ``coll0.listSearchIndexes()`` repeatedly every 5 seconds until the following condition is satisfied:
+   1. An index with the ``name`` of ``test-search-index`` is present.  This index is referred to as ``index``.
+   2. The ``latestDefinition`` property of the search index has the field ``queryable`` with a value of ``true``.
+   3. The ``latestDefinition`` property of the search index has the field ``status`` with the value of ``READY``.
+  
+#. Assert that an index is present with the name ``test-search-index`` and the definition has a
+  property ``mappings``, whose value is ``{ dynamic: true }``
+
+Case 5: ``dropSearchIndex`` suppresses namespace not found errors
```
GromNaN, line 196: I didn't see this in the spec. I'm ignoring [error with code `26`](https://github.com/mongodb/mongo/blob/master/src/mongo/base/error_codes.yml)

GromNaN (commented):
Tests implemented with success in PHP: [tests/Collection/SearchIndexFunctionalTest.php](https://github.com/mongodb/mongo-php-library/pull/1097/files#diff-1aa076f39e1455d89beb8204087af66e52738865b00e098a839fb137586b3082)

To improve the coverage, I suggest adding this e2e tests:
- Call to `createSearchIndexes` with an empty list of indexes.
- Call `createSearchIndex` without `name`. It fallbacks to the "default" name.

source/index-management/tests/README.rst
```diff
@@ -45,4 +44,157 @@ Tests
...
+      }
+    }
+    {
+      name: 'test-search-index-2',
+      definition: {
+        mappings: { dynamic: false }
+      }
+    }
+
+#. Assert that the command does not error and the server responds with a success.
```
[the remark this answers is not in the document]
baileympearson: Done!

source/index-management/tests/README.rst
GromNaN: Nothing to change. It's just me who didn't read the spec as carefully as I should have.

It's [in the spec](https://github.com/mongodb/specifications/blob/78554de6a78fc996fa8d1ef9733b123a29ad8309/source/index-management/index-management.rst#namespacenotfound-errors).

> Drivers MUST suppress NamespaceNotFound errors for the dropSearchIndex helper.

source/index-management/tests/README.rst
```diff
@@ -45,4 +44,157 @@ Tests
...
+#. Run ``coll0.listSearchIndexes()`` repeatedly every 5 seconds until the following condition is satisfied.  Store
```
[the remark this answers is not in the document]
GromNaN: If the driver sends `{ name: null }`, the servers produces this error: `PlanExecutor error during aggregation :: caused by :: org.bson.BsonInvalidOperationException: Value expected to be of type STRING is of unexpected type NULL`

But I see this is already covered by the UST.

source/index-management/tests/README.rst
```diff
@@ -45,4 +44,157 @@ Tests
...
+Case 1: Driver can successfully create and list search indexes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+#. Create a collection with a randomly generated name (referred to as ``coll0``).
+#. Create a new search index on ``coll0`` with the following definition:
+
+  .. code:: typescript
+
+    {
+      name: 'test-search-index',
```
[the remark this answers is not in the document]
GromNaN: Correct, the server may decide to change the default name in a future version. We don't have to test it. And the UST test the command sent when no name is provided.

GromNaN (approved):
LGTM. I've updated the PHP implementation following our discussions.

source/index-management/index-management.rst
```diff
@@ -888,7 +892,35 @@ An example, from Javascript:
...
+Where are read concern and write concern?
+-----------------------------------------
+
+These commands internally proxy the search index management commands to a separate process that runs alongside an Atlas cluster.  As such, read concern and 
+write concern are not relevant for the search index management commands.
+
+Consistency with Existing APIs
+------------------------------
+
+Drivers SHOULD strive for a search index management API that is as consistent with their existing search index management API as much as possible.
```
dariakp, line 905: This sentence is confusing - which APIs is it trying to differentiate?

source/index-management/index-management.rst
```diff
@@ -1031,7 +1066,7 @@ Index View API for Search Indexes
     /**
      * Updates a single search index from the collection by the index name.
      */
-    updateOne(name: String, options: Optional<UpdateSearchIndexOptions>): Result;
+    updateOne(name: String, definition: Document, options: Optional<UpdateSearchIndexOptions>): Result;
```
dariakp, line 1069: it feels inconsistent to have name sometimes as the first parameter and sometimes as the second when both `name` and `definition` are potential parameters; if definition is always required and name is sometimes optional, it would indicate that we want to use `definition` always as the first parameter

dariakp (changes requested):

source/index-management/index-management.rst
baileympearson: I view the `create` APIs separate from the `list`/`update`/`drop` apis.  Whereas the create APIs are saying "create a search index with definition <definition> and optional name <name>", the other APIs (specifically `drop` and `update`) are saying "do something to the index with name <name>".  Specifically, in this case, we're updating index with name <name> to have the new definition - flipping the parameters makes less sense.

source/index-management/tests/updateSearchIndex.yml
```diff
@@ -28,10 +28,11 @@ tests:
         arguments:
           name: &indexName 'test index'
           definition: &definition {}
-        expectError:          
+        expectError:
           # Search indexes are only available on 7.0+ atlas clusters.  DRIVERS-2630 will add e2e testing
```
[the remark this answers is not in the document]
baileympearson: Done.

source/index-management/index-management.rst
baileympearson: That was a typo - good catch

source/index-management/index-management.rst
```diff
@@ -888,7 +892,35 @@ An example, from Javascript:
...
+Where are read concern and write concern?
+-----------------------------------------
+
+These commands internally proxy the search index management commands to a separate process that runs alongside an Atlas cluster.  As such, read concern and 
+write concern are not relevant for the search index management commands.
+
+Consistency with Existing APIs
+------------------------------
+
+Drivers SHOULD strive for a search index management API that is as consistent with their existing index management API as much as possible.
```
dariakp, line 905:
```suggestion
Drivers SHOULD strive for a search index management API that is as consistent as possible with their existing index management API.
```

source/index-management/tests/README.rst
```diff
@@ -45,4 +44,154 @@ Tests
...
+      definition: {
+        mappings: { dynamic: false }
+      }
+    }
+
+#. Assert that the command returns the name of the index: ``"test-search-index"``.
+#. Run ``coll0.listSearchIndexes()`` repeatedly every 5 seconds until the following condition is satisfied:
+   1. An index with the ``name`` of ``test-search-index`` is present and the index has a field ``queryable`` with a value of ``true``.
+
+#. Assert that ``index`` has a property ``mappings`` whose value is ``{ dynamic: false }``
```
joshbsiegel, line 98: Should we specify on line 95 that the result of `coll0.listSearchIndexes()` is stored in a variable called `index`? `index` is referenced on line 98 without any prior mention of initialization. Something similar is done on line 124 where it's mentioned to store the result of a list command into `indexes` (which is later referenced).

source/index-management/index-management.rst
```diff
@@ -414,6 +414,10 @@ Index View API
      * For drivers that cannot make IndexView iterable, they MUST implement this method to
      * return a list of indexes. In the case of async drivers, this MAY return a Future<Cursor>
      *  or language/implementation equivalent.
+     * 
+     *  If drivers are unable to make the IndexView iterable, they MAY opt to provide the options for 
+     *  listing search indexes via the `list` method instead of the `Collection.listSearchIndexes` method.
+
```
baileympearson, line 420: After discussion with @dariakp on slack, we decided that a more consistent API is to include the `name` in an options object.  There were two reasons for this change:

- It's odd to have a set of search index management methods that have a name as the first parameter and an optional second parameter in others.  It makes more sense to include an optional name in an "options" object.
- This is more closely aligned with the regular index management API.

dariakp (approved):

source/index-management/tests/README.rst
```diff
@@ -45,4 +44,154 @@ Tests
...
+    }
+
+#. Assert that the command returns an array containing the new indexes' names: ``["test-search-index-1", "test-search-index-2"]``.
+#. Run ``coll0.listSearchIndexes()`` repeatedly every 5 seconds until the following condition is satisfied.  Store
+   the result in ``indexes``.
+   1. An index with the ``name`` of ``test-search-index-1`` is present and index has a field ``queryable`` with the value of ``true``.
+   2. An index with the ``name`` of ``test-search-index-2`` is present and index has a field ``queryable`` with the value of ``true``.
+#. For each ``index`` in ``indexDefinitions``
+   1. Find the matching index definition in ``indexes`` by matching on ``index.name``.  If no index exists, raise an error.
+   2. Assert that the matching index ``mappings``, whose value is ``{ dynamic: false }``
```
kevinAlbs, line 129:
```suggestion
#. Run ``coll0.listSearchIndexes()`` repeatedly every 5 seconds until the following condition is satisfied.
   1. An index with the ``name`` of ``test-search-index-1`` is present and index has a field ``queryable`` with the value of ``true``. Store result in ``index1``.
   2. An index with the ``name`` of ``test-search-index-2`` is present and index has a field ``queryable`` with the value of ``true``. Store result in ``index2``.
#. Assert that ``index1`` and ``index2`` have the property ``latestDefinition`` whose value is ``{ "mappings" : { "dynamic" : false } }``
```

Suggest using separate `index1` and `index2` variables to simplify test.

Suggest updating assert to use `latestDefinition` to match returned document. Here is an example of the documents returned in the C driver testing:

```
  { "id" : "64c15fc9a199d3199e27ad70", "name" : "test-search-index-1", "status" : "READY", "queryable" : true, "latestDefinition" : { "mappings" : { "dynamic" : false } } }
  { "id" : "64c15fc9a199d3199e27ad71", "name" : "test-search-index-2", "status" : "READY", "queryable" : true, "latestDefinition" : { "mappings" : { "dynamic" : false } } }
```

source/index-management/tests/README.rst
```diff
@@ -45,4 +44,154 @@ Tests
...
+      definition: {
+        mappings: { dynamic: false }
+      }
+    }
+
+#. Assert that the command returns the name of the index: ``"test-search-index"``.
+#. Run ``coll0.listSearchIndexes()`` repeatedly every 5 seconds until the following condition is satisfied:
+   1. An index with the ``name`` of ``test-search-index`` is present and index has a field ``queryable`` with the value of ``true``.
+
+#. Run a ``dropSearchIndexes`` on ``coll0``, using ``test-search-index`` for the name.
```
kevinAlbs, line 150:
```suggestion
#. Run a ``dropSearchIndex`` on ``coll0``, using ``test-search-index`` for the name.
```

source/index-management/tests/README.rst
```diff
@@ -45,4 +44,154 @@ Tests
...
+      }
+    }
+
+#. Assert that the command does not error and the server responds with a success.
+#. Run ``coll0.listSearchIndexes()`` repeatedly every 5 seconds until the following condition is satisfied:
+   1. An index with the ``name`` of ``test-search-index`` is present.  This index is referred to as ``index``.
+   2. The index has a field ``queryable`` with a value of ``true`` and has a field ``status`` with the value of ``READY``.
+  
+#. Assert that an index is present with the name ``test-search-index`` and the definition has a
+  property ``mappings``, whose value is ``{ dynamic: true }``
```
kevinAlbs, line 191:
```suggestion
  property ``latestDefinition``, whose value is ``{ 'mappings' : { 'dynamic' : true } }``
```

Suggest updating assert to use `latestDefinition` to match returned document. Here is an example of the documents returned in the C driver testing:

```
{ "id" : "64c1651aa199d3199e27ae1a", "name" : "test-search-index", "status" : "PENDING", "queryable" : false, "latestDefinition" : { "mappings" : { "dynamic" : false } } }
```

source/index-management/index-management.rst
```diff
@@ -414,6 +414,10 @@ Index View API
      * For drivers that cannot make IndexView iterable, they MUST implement this method to
      * return a list of indexes. In the case of async drivers, this MAY return a Future<Cursor>
      *  or language/implementation equivalent.
+     * 
+     *  If drivers are unable to make the IndexView iterable, they MAY opt to provide the options for 
+     *  listing search indexes via the `list` method instead of the `Collection.listSearchIndexes` method.
```
kevinAlbs, line 419:
```suggestion
     *  listing search indexes via the `list` method instead of the `Collection.indexes` method.
```

source/index-management/index-management.rst
```diff
@@ -990,6 +1028,9 @@ Index View API for Search Indexes
      * For drivers that cannot make SearchIndexView iterable, they MUST implement this method to
      * return a list of indexes. In the case of async drivers, this MAY return a Future<Cursor>
      *  or language/implementation equivalent.
+     *  
+     *  If drivers are unable to make the SearchIndexView iterable, they MAY opt to provide the options for 
+     *  listing search indexes via the `list` method instead of the `Collection.listSearchIndexes` method.
```
kevinAlbs, line 1033:
```suggestion
     *  listing search indexes via the `list` method instead of the `Collection.searchIndexes` method.
```

source/index-management/index-management.rst
```diff
@@ -902,6 +933,11 @@ Common Interfaces
     name: Optional<string>;
   }
 
+  interface SearchIndexOptions {
+    // The name for this index, if present.
+    name: Optional<string>;
```
kevinAlbs, line 938: Since the optional `name` only applies to the create operations, I suggest placing `name` in `CreateSearchIndexOptions` to simplify the API.

source/index-management/index-management.rst
GromNaN: I understand this was already allowed, this is how [it is already implemented in PHP](https://github.com/mongodb/mongo-php-library/blob/9b6d5a16411c07a52bae634f37d40bf193f30097/src/Collection.php#L400).

source/index-management/tests/README.rst
```diff
@@ -45,4 +44,154 @@ Tests
...
+These tests must run against an Atlas cluster with a 7.0+ server.  Tools are available drivers-evergreen-tools which can setup and teardown
+Atlas clusters.  To ensure that the Atlas cluster is cleaned up after each CI run, drivers should configure evergreen to run these tests 
+as a part of a task group.  Be sure that the cluster gets torn down! 
+
+When working locally on these tests, the same Atlas setup and teardown scripts can be used locally to provision a cluster for development.
+
+Case 1: Driver can successfully create and list search indexes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+#. Create a collection with a randomly generated name (referred to as ``coll0``).
```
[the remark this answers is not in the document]
GromNaN: I confirm the collection must be created first.

source/index-management/index-management.rst
baileympearson: I don't feel strongly here, but I intended the new search indexes API to be as similar to the existing API as possible:

```typescript
    createIndex(keys: Document, indexOptions: Optional<IndexOptions>, options: Optional<CreateIndexOptions>): String;
```

Some drivers (like PHP) do combine the options types, which has been noted in a comment on the new createSearchIndex methods (same as is done for createIndex).

kevinAlbs (approved):

dariakp (approved):

source/index-management/index-management.rst
```diff
@@ -1093,3 +1136,4 @@ Changelog
 :2023-05-10:  Merge index enumeration and index management specs and get rid of references 
              to legacy server versions.
 :2023-05-18:  Add the search index management API.
+:2023-07-17:  Add search index management clarifications.
```
dariakp, line 1139: Should we bump the date to today? @baileympearson

baileympearson, line 1139:
```suggestion
:2023-07-27:  Add search index management clarifications.
```

PR #1446, opened 2023-08-01
DRIVERS-2687 Add BSON Binary Data subtype Sensitive

NoahStapp:
The new test validates only the correct subtype, number encoding/decoding, and binary data encoding/decoding. Any binary data itself is not validated. [DRIVERS-2687](https://jira.mongodb.org/browse/DRIVERS-2687) specifies that data using this subtype is meant to not be logged, so this should satisfy testing requirements.


blink1073 (approved):
LGTM!

PR #1447, opened 2023-08-02
DRIVERS-2679: make delimiting slash between hosts and connection options optional

W-A-James:
- Change spec language to make delimiting slash between hosts and connection options optional.
- Updated connection-string spec tests to allow optional delimiting slash

- [x] Update changelog.

### DIFF
--- a/source/connection-string/tests/valid-options.json
+++ b/source/connection-string/tests/valid-options.json
@@ -26,7 +26,13 @@
       "uri": "mongodb://example.com?tls=true",
       "valid": true,
       "warning": false,
-      "hosts": null,
+      "hosts": [
+        {
+          "type": "hostname",
+          "host": "example.com",
+          "port": null
+        }
+      ],
       "auth": null,
       "options": {
         "tls": true
--- a/source/connection-string/tests/valid-options.json
+++ b/source/connection-string/tests/valid-options.json
@@ -20,6 +20,15 @@
       "options": {
         "authmechanism": "MONGODB-CR"
       }
+    },
+    {
+      "description": "Missing delimiting slash between hosts and options",
+      "uri": "mongodb://example.com?w=1",
+      "valid": true,
+      "warning": null,
+      "hosts": null,
+      "auth": null,
+      "options": null
     }
   ]
 }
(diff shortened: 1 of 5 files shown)


source/connection-string/tests/valid-options.yml
```diff
@@ -15,3 +15,11 @@ tests:
...
         options:
             authmechanism: "MONGODB-CR"
+    -
+        description: "Missing delimiting slash between hosts and options"
+        uri: "mongodb://example.com?w=1"
+        valid: true
+        warning: ~
+        hosts: ~
+        auth: ~
+        options: ~
```
jyemin, line 25: Consider adding an assertion that options are respected when there is no delimiting slash (e.g. that the write concern is set here, or some other option if it's easier).

W-A-James: Changed this to check for tls being set to true.

source/connection-string/tests/valid-options.yml
```diff
@@ -15,3 +15,12 @@ tests:
             db: "admin"
         options:
             authmechanism: "MONGODB-CR"
+    -
+        description: "Missing delimiting slash between hosts and options"
+        uri: "mongodb://example.com?tls=true"
+        valid: true
+        warning: false
+        hosts: ~
```
alcaeus, line 23: This should also assert that the hostname is parsed correctly:

```suggestion
        hosts:
            -
                type: "hostname"
                host: "example.com"
                port: ~
```

W-A-James:
@jyemin tests have been verified to pass on C# by @BorisDog

source/connection-string/tests/valid-options.json
```diff
@@ -20,6 +20,17 @@
       "options": {
         "authmechanism": "MONGODB-CR"
       }
+    },
+    {
+      "description": "Missing delimiting slash between hosts and options",
+      "uri": "mongodb://example.com?tls=true",
+      "valid": true,
+      "warning": false,
+      "hosts": null,
```
BorisDog, line 29: Looks like `hosts` property is different from one yml.

BorisDog (commented):
The

source/connection-string/tests/valid-options.json
W-A-James: Good catch, forgot to regen the json files

BorisDog (approved):
Test passes in C#

PR #1448, opened 2023-08-04
DRIVERS-2469 Add durations to connection pool events

stIncMale:
This spec change is implemented in https://github.com/mongodb/mongo-java-driver/pull/1166 for the MongoDB Java Driver.

There are no unified tests for `CommandSucceededEvent`/`CommandFailedEvent.duration` (see https://github.com/mongodb/specifications/blob/master/source/command-logging-and-monitoring/command-logging-and-monitoring.rst), similarly, there are no unified tests in this PR except those for logging.

DRIVERS-2469

- [x] Update changelog.

### DIFF
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -921,8 +921,8 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
        * it is the time elapsed between the `ConnectionCreatedEvent`
        * emitted by the same checking out and this event.
        *
-       * Naturally, this duration does not include any part of
-       * `ConnectionCheckedOutEvent`/`ConnectionCheckOutFailedEvent.waited`.
+       * Naturally, this duration is not greater than
+       * `ConnectionCheckedOutEvent`/`ConnectionCheckOutFailedEvent.duration`.
        *
        * A driver that delivers events synchronously MUST NOT include in this duration
        * the time to deliver the `ConnectionCreatedEvent`.
@@ -995,9 +995,9 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
       reason: string|Enum;
 
       /**
-       * See `ConnectionCheckedOutEvent.waited`.
+       * See `ConnectionCheckedOutEvent.duration`.
        */
-      waited: Duration;
+      duration: Duration;
     }
 
     /**
@@ -1015,20 +1015,20 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
       connectionId: int64;
 
       /**
-       * The time elapsed waiting for conditions
-       * specified by `ConnectionPoolOptions.waitQueueTimeoutMS`.
-       * More specifically, the time elapsed between the `ConnectionCheckOutStartedEvent` and:
-       *  - the `ConnectionCreatedEvent` emitted by the same checking out,
-       *    if this checking out emitted `ConnectionCreatedEvent`;
-       *  - this event, otherwise.
+       * The time it took to check out the connection.
+       * More specifically, the time elapsed between the `ConnectionCheckOutStartedEvent` and
+       * the `ConnectionCheckedOutEvent`/`ConnectionCheckOutFailedEvent`
+       * emitted by the same checking out.
        *
-       * Naturally, this duration is usually
+       * Naturally, if a new connection was not created (`ConnectionCreatedEvent`)
+       * and established (`ConnectionReadyEvent`) as part of checking out,
+       * this duration is usually
        * smaller than or equal to `ConnectionPoolOptions.waitQueueTimeoutMS`,
        * but MAY occasionally be greater than that,
        * because a driver does not provide hard real-time guarantees.
        *
        * A driver that delivers events synchronously MUST NOT include in this duration
-       * the time to deliver the `ConnectionCheckOutStartedEvent`, `ConnectionCreatedEvent`.
+       * the time to deliver the `ConnectionCheckOutStartedEvent`.
        * Doing so eliminates a thing to worry about in support cases related to this duration,
        * because the time to deliver synchronously is affected by user code.
        * The driver MUST document this behavior
@@ -1036,9 +1036,9 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
        *
        * A driver MAY choose the type idiomatic to the driver.
        * If the type chosen does not convey units, e.g., `int64`,
-       * then the driver MAY include units in the name, e.g., `waitedMS`.
+       * then the driver MAY include units in the name, e.g., `durationMS`.
        */
-      waited: Duration;
+      duration: Duration;
     }
 
     /**
@@ -1350,13 +1350,13 @@ In addition to the common fields defined above, this message MUST contain the fo
      - If ``reason`` is ``ConnectionError``, the associated error. The type and format of this value is flexible; see the
        `logging specification <../logging/logging.rst#representing-errors-in-log-messages>`_  for details on representing errors in log messages.
 
-   * - waitedMS
+   * - durationMS
      - Int64
-     - ``ConnectionCheckOutFailedEvent.waited`` converted to milliseconds.
+     - ``ConnectionCheckOutFailedEvent.duration`` converted to milliseconds.
 
 The unstructured form SHOULD be as follows, using the values defined in the structured format above to fill in placeholders as appropriate:
 
-  Checkout failed for connection to {{serverHost}}:{{serverPort}}. Reason: {{reason}}. Error: {{error}}. Waited: {{waitedMS}} ms
+  Checkout failed for connection to {{serverHost}}:{{serverPort}}. Reason: {{reason}}. Error: {{error}}. Duration: {{durationMS}} ms
 
 Connection Checked Out
 -----------------------
@@ -1378,13 +1378,13 @@ In addition to the common fields defined above, this message MUST contain the fo
      - Int64
      - The driver-generated ID for the connection as defined in `Connection <#connection>`_.
 
-   * - waitedMS
+   * - durationMS
      - Int64
-     - ``ConnectionCheckedOutEvent.waited`` converted to milliseconds.
+     - ``ConnectionCheckedOutEvent.duration`` converted to milliseconds.
 
 The unstructured form SHOULD be as follows, using the values defined in the structured format above to fill in placeholders as appropriate:
 
-  Connection checked out: address={serverHost}}:{{serverPort}}, driver-generated ID={{driverConnectionId}}, waited={{waitedMS}} ms
+  Connection checked out: address={serverHost}}:{{serverPort}}, driver-generated ID={{driverConnectionId}}, duration={{durationMS}} ms
 
 Connection Checked In
 ---------------------
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -1016,9 +1016,8 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
 
       /**
        * The time it took to check out the connection.
-       * More specifically, the time elapsed between the `ConnectionCheckOutStartedEvent` and
-       * the `ConnectionCheckedOutEvent`/`ConnectionCheckOutFailedEvent`
-       * emitted by the same checking out.
+       * More specifically, the time elapsed between
+       * the `ConnectionCheckOutStartedEvent` and this event.
        *
        * Naturally, if a new connection was not created (`ConnectionCreatedEvent`)
        * and established (`ConnectionReadyEvent`) as part of checking out,
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
@@ -921,7 +921,8 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
        * it is the time elapsed between the `ConnectionCreatedEvent`
        * emitted by the same checking out and this event.
        *
-       * Naturally, this duration is not greater than
+       * Naturally, when establishing a connection is part of checking out,
+       * this duration is not greater than
        * `ConnectionCheckedOutEvent`/`ConnectionCheckOutFailedEvent.duration`.
        *
        * A driver that delivers events synchronously MUST NOT include in this duration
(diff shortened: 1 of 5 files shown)


source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -12,6 +12,8 @@ Abstract
 
 Drivers currently support a variety of options that allow users to configure connection pooling behavior. Users are confused by drivers supporting different subsets of these options. Additionally, drivers implement their connection pools differently, making it difficult to design cross-driver pool functionality. By unifying and codifying pooling options and behavior across all drivers, we will increase user comprehension and code base maintainability.
 
+This specification does not apply to drivers that do not support multitasking.
```
stIncMale, line 15: This fleshes out the current state of affairs.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -131,8 +133,8 @@ Additionally, Drivers that implement a Connection Pool MUST support the followin
       /**
        *  NOTE: This option has been deprecated in favor of timeoutMS.
        *
-       *  The maximum amount of time a thread can wait for a connection
-       *  to become available.
+       *  The maximum amount of time a thread can wait for either an available non-perished connection (limited by `maxPoolSize`),
+       *  or a pending connection (limited by `maxConnecting`).
```
stIncMale, line 137: This clarifies the definition to match the reality. Needs to be done as part of this PR because this PR relies on the definition of `waitQueueTimeoutMS`.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -663,6 +659,13 @@ Before a given `Connection <#connection>`_ is returned from checkOut, it must be
     else:
         decrement availableConnectionCount
     set connection state to "in use"
+
+    # If there is no background thread, the pool MUST ensure that
+    # there are at least minPoolSize total connections.
+    do asynchronously:
+      while totalConnectionCount < minPoolSize:
+        populate the pool with a connection
+
```
stIncMale, line 668: 1. Moved the part of the `checkOut` logic that is concerned with populating the pool at the end of the `checkOut` logic because the previous position was weird: between creating a new connection that is about to be checked out and establishing it.
2. Replaced `This MUST be done in a non-blocking manner` with the `do asynchronously` block, because it's the only interpretation of "non-blocking manner" I can see here, and I find it clearer. Expressing the requirement as pseudocode is also more explicit as opposed to being expressed in a comment.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -1288,9 +1338,13 @@ In addition to the common fields defined above, this message MUST contain the fo
      - If ``reason`` is ``ConnectionError``, the associated error. The type and format of this value is flexible; see the
        `logging specification <../logging/logging.rst#representing-errors-in-log-messages>`_  for details on representing errors in log messages.
 
+   * - waitedMS
+     - Int64
+     - ``ConnectionCheckOutFailedEvent.waited`` converted to milliseconds.
```
stIncMale, line 1343: I chose milliseconds because this unit is used when logging `CommandSucceededEvent/CommandFailedEvent.duration`. However, at least when logging `ConnectionCheckedOutEvent`/`ConnectionCheckOutFailedEvent.waited`, nanoseconds is definitely a more appropriate unit because those wait times can be quite small. Fortunately, an application can still access the value using a smaller unit by listening to `ConnectionCheckedOutEvent`/`ConnectionCheckOutFailedEvent`.

**For reviewers**: should nanoseconds be used when logging `ConnectionCheckedOutEvent`/`ConnectionCheckOutFailedEvent.waited`?

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -131,8 +133,8 @@ Additionally, Drivers that implement a Connection Pool MUST support the followin
       /**
        *  NOTE: This option has been deprecated in favor of timeoutMS.
        *
-       *  The maximum amount of time a thread can wait for a connection
-       *  to become available.
+       *  The maximum amount of time a thread can wait for either an available non-perished connection (limited by `maxPoolSize`),
```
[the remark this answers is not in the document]

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -909,6 +912,24 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
...
+       * Naturally, this duration does not include any part of `ConnectionCheckedOutEvent`/`ConnectionCheckOutFailedEvent.waited`.
+       *
+       * A driver that delivers events synchronously MUST NOT include in this duration
+       * the time to deliver the `ConnectionCreatedEvent`.
+       * Doing so eliminates a thing to worry about in support cases related to this duration,
+       * because the time to deliver synchronously is affected by user code.
+       * The driver MUST document this behavior as well as explicitly warn users that the behavior may change in the future.
+       *
+       * A driver MAY choose the type idiomatic to the driver.
+       * If the type chosen does not convey units, e.g., `int64`, then the driver MAY include units in the name, e.g., `durationMS`.
```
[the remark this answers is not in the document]

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -981,6 +1007,26 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
...
+       * but MAY occasionally be greater than that, because a driver does not provide hard real-time guarantees.
+       *
+       * A driver that delivers events synchronously MUST NOT include in this duration
+       * the time to deliver the `ConnectionCheckOutStartedEvent`, `ConnectionCreatedEvent`.
+       * Doing so eliminates a thing to worry about in support cases related to this duration,
+       * because the time to deliver synchronously is affected by user code.
+       * The driver MUST document this behavior as well as explicitly warn users that the behavior may change in the future.
+       *
+       * A driver MAY choose the type idiomatic to the driver.
+       * If the type chosen does not convey units, e.g., `int64`, then the driver MAY include units in the name, e.g., `waitedMS`.
```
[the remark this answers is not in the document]

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -981,6 +1007,26 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
        *  The ID of the Connection
        */
       connectionId: int64;
+
+      /**
+       * The time elapsed waiting for conditions specified by `ConnectionPoolOptions.waitQueueTimeoutMS`.
+       * More specifically, the time elapsed between the `ConnectionCheckOutStartedEvent` and:
+       *  - the `ConnectionCreatedEvent` emitted by the same checking out, if this checking out emitted `ConnectionCreatedEvent`;
+       *  - this event, otherwise.
```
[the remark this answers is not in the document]
stIncMale: > As a result, the average check-out duration actually goes down

The important thing to note here is that `ConnectionCheckedOutEvent`/`ConnectionCheckOutFailedEvent.waited` is not a "check-out duration", and it is not documented as such.

> Instead, we should always report the entire time between when a check-out starts and when it delivers a connection to the caller.

The description of https://jira.mongodb.org/browse/DRIVERS-2469 requests the waited time and time to establish, which is what I added.

If the checkout time appears to be also needed, we may

1. either replace the waited time with the checkout time,
2. or add the checkout time to be the third reported duration.

Is my understanding correct that you prefer the first option?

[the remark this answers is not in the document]
baileympearson: here's the phrasing from the ticket:
> - The amount of time a request to check out a connection waited for an idle or new connection.
> - The total amount of time a new connection took to establish (i.e. the time between requesting a new connection and the new connection being ready to send application operations on).

I agree with Matt here, for the reasons that Matt described.  But also because I interpret the request in the drivers ticket to deliver the following new metrics:

- the whole time it took for any checkout to succeed or fail (the time between `ConnectionCheckoutStarted` and `ConnectionCheckoutSucceeded` or `ConnectionCheckoutFailed`)
- the time it takes for any connection to establish (the time between `ConnectionCreated` and `ConnectionReady`)

As it currently stands, a user would not be able to interpret the `duration` property on the `ConnectionCheckoutSucceededEvent` without monitoring for connection establishment and correlating the events.  I'd argue that this makes the `duration` properties meaningless as its currently specified.

If we make the `duration` on the `ConnectionCheckoutSucceededEvent` the time between `ConnectionCheckoutStarted` and the successful connection checkout, users might see spikes in this value when new connections are created and they can confirm this corresponds to new connections being established if they wish, but the duration can be interpreted by itself without monitoring for other events.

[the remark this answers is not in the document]
matthewdale: That's correct, I think we should replace "waited" time with the full checkout duration. My intent in DRIVERS-2469 was not to distinguish separate time waited for an idle connection and time waited for a new connection, but to add two mostly orthogonal metrics (check-out duration and connection establishment duration).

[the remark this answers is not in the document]

matthewdale (approved):
Looks good 👍

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -909,6 +913,29 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
...
+
+      /**
+       * The time it took to establish the connection.
+       * In accordance with the definition of establishment of a connection
+       * specified by `ConnectionPoolOptions.maxConnecting`,
+       * it is the time elapsed between the `ConnectionCreatedEvent`
+       * emitted by the same checking out and this event.
+       *
+       * Naturally, this duration is not greater than
+       * `ConnectionCheckedOutEvent`/`ConnectionCheckOutFailedEvent.duration`.
```
baileympearson, line 925: This doesn't make sense for minPoolSize population, because connections created during minPoolSize population are not associated with any checkout events.  Can we update the phrasing here?

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -966,6 +993,11 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
        *   - "connectionError": The Connection check out attempt experienced an error while setting up a new Connection
        */
       reason: string|Enum;
+
+      /**
+       * See `ConnectionCheckedOutEvent.duration`.
+       */
```
baileympearson, line 999: I understand the desire to reduce duplication but can we include the full description here?  All other events have duplicate comments on duplicate fields, and it's annoying to have to jump around the file manually to see what this field contains (especially since it's not a link, just text).

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
```diff
@@ -909,6 +913,29 @@ See the `Load Balancer Specification <../load-balancers/load-balancers.rst#event
...
+       *
+       * Naturally, this duration is not greater than
+       * `ConnectionCheckedOutEvent`/`ConnectionCheckOutFailedEvent.duration`.
+       *
+       * A driver that delivers events synchronously MUST NOT include in this duration
+       * the time to deliver the `ConnectionCreatedEvent`.
+       * Doing so eliminates a thing to worry about in support cases related to this duration,
+       * because the time to deliver synchronously is affected by user code.
+       * The driver MUST document this behavior
+       * as well as explicitly warn users that the behavior may change in the future.
```
baileympearson, line 932: can you elaborate on this?  I'm not too familiar with how emitting events synchronously would be impacted by user's code.

baileympearson (changes requested):
I'm concerned about testing.  First, we're missing tests for `ConnectionCheckoutFailed.duration` with the existing tests.  Second, the tests currently provide no guarantees about what the fields actually mean.  Lastly, the testing strategy relies on logging having been implemented by a driver to test these changes.

Did you consider either trying to add tests to the CMAP unit/integration tests somehow or adding prose tests?

--

Also - can we update the title of the PR to include the driver's ticket?  That's the usual convention in the specs repo (look at other open PRs and commits).

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
stIncMale: Synchronous delivery with necessity means that checking out waits for some sort of a confirmation from user code that an emitted event was delivered. If the user code delays the confirmation for whatever reason, it will affect the duration reported by the driver despite the driver having nothing to do with the delay. In the most simplest form it may be invocation of a synchronous user function and passing an emitted event as an argument to it, as part of checking out.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
stIncMale: > it's annoying to have to jump around

I argue that it is more annoying to try and figure out whether `ConnectionCheckOutFailedEvent.duration` is the same thing as `ConnectionCheckedOutEvent.duration` or not by comparing two large pieces of text. It will also be annoying when those pieces need to be changed, especially when they diverge due to a human error because there is no mechanism keeping them consistent.

Given the above, do you still require the documentation to be duplicated?

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
stIncMale: Thank you! Done.

stIncMale:
A response to https://github.com/mongodb/specifications/pull/1448#pullrequestreview-1570052630 (@baileympearson)

> First, we're missing tests for ConnectionCheckoutFailed.duration with the existing tests.

It's here: https://github.com/stIncMale/specifications/blob/DRIVERS-2469/source/connection-monitoring-and-pooling/tests/logging/connection-logging.yml#L216-L224.

> Second, the tests currently provide no guarantees about what the fields actually mean. Lastly, the testing strategy relies on logging having been implemented by a driver to test these changes.

> Did you consider either trying to add tests to the CMAP unit/integration tests somehow or adding prose tests?

I did. I looked at `CommandSucceededEvent`/`CommandFailedEvent.duration`, observed that the only tests for that field is logging tests, and used the same approach, which I also mentioned in the description of the PR. Do you request prose tests for the fields added in this PR?

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst
baileympearson: Ah gotcha, a callback mechanism or something similar as opposed to event emitters or streams.  Thanks!

baileympearson (approved):

PR #1451, opened 2023-08-17
Add missing retryable write case to pseudo code

kkloberdanz:
The pseudo code example for executing retryable write commands was missing the case below:

- CSOT is not enabled and one retry was attempted.

This corrects the pseudo code to unify it with the prose description.

Related to: DRIVERS-1807

- [ ] Update changelog.


kkloberdanz:
The original pseudo code has some additional errata, which is documented in the ticket: https://jira.mongodb.org/browse/DRIVERS-2695

Closing for now.

PR #1452, opened 2023-08-21
DRIVERS-2578 Drivers use polling SDAM on AWS Lambda

ShaneHarvey:
[DRIVERS-2578](https://jira.mongodb.org/browse/DRIVERS-2578) Disable streaming SDAM by default on AWS Lambda and similar FaaS platforms. Introduce the sdamMode=stream/poll/auto URI option.

This PR also updates the Unified Test Format to version 1.17 by adding support for listening to SDAM ServerHeartbeat events and asserting on the "awaited" field.

- [X] Update changelog.
- [X] Test changes in at least one language driver. Python implementation: https://github.com/mongodb/mongo-python-driver/pull/1360


source/uri-options/uri-options.rst
```diff
@@ -282,17 +282,23 @@ pertaining to URI options apply here.
...
      - Enables retryable reads on server 3.6+
 
    * - retryWrites
-     - "true" or "false
+     - "true" or "false"
      - defined in `retryable writes spec <https://github.com/mongodb/specifications/blob/master/source/retryable-writes/retryable-writes.rst#retrywrites>`_
      - no
      - Enables retryable writes on server 3.6+
 
+   * - sdamMode
```
jyemin, line 296: While "SDAM" trips easily off the tongue for all of us, it's not commonly understood by others.  Let's consider other names, e.g. "serverMonitoringMode".

alcaeus: I agree. `serverMonitoringMode` is more descriptive than `sdamMode`.

ShaneHarvey: Agreed. I've renamed it "serverMonitoringMode".

alcaeus (approved):

source/server-discovery-and-monitoring/tests/unified/serverMonitoringMode.json
```diff
@@ -0,0 +1,129 @@
...
+                "client": {
+                  "id": "client0",
+                  "uriOptions": {
+                    "serverMonitoringMode": "auto"
+                  }
+                }
+              },
+              {
+                "database": {
+                  "id": "dbSdamModeAuto",
```
jyemin, line 23: Nit

dbSdamModeAuto => dbServerMonitoringModeAuto

in all places

jyemin (changes requested):

source/faas-automated-testing/faas-automated-testing.rst
```diff
@@ -243,6 +243,8 @@ the function implementation the driver MUST:
 - Drivers MUST record the durations and counts of the heartbeats, the durations of the
   commands, as well as keep track of the number of open connections, and report this information in
   the function response as JSON.
+- Drivers MUST assert no ServerHeartbeat events contain the ``awaited=True`` flag to
```
prestonvasquez, line 246: Do you think it's worth extending the test plan for handshakes (https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.rst#test-plan) to check that the the initial handshake uses OP_QUERY? As of this ticket, there is a possibility that polling could be used with a server versions >= 4.4 . And AFAIK this is novel and so potentially important to ensure that correct handshake logic is being implemented.

It's probably not very important, as the wire message between streaming / polling are not that different. But I do notice we don't check this anywhere and so it could be useful to do so as a part of this change.

source/uri-options/tests/sdam-options.yml
```diff
@@ -0,0 +1,35 @@
...
+      warning: false
+      hosts: ~
+      auth: ~
+      options:
+          serverMonitoringMode: "poll"
+
+    - description: "invalid serverMonitoringMode"
+      uri: "mongodb://example.com/?serverMonitoringMode=invalid"
+      valid: true
+      warning: true
```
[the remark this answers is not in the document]
ShaneHarvey: Can you elaborate here? My understanding is that invalid values still cause warnings [according to the connection string spec](https://github.com/mongodb/specifications/blob/master/source/connection-string/connection-string-spec.rst#values):

> Any invalid Values for a given key MUST be ignored and MUST log a WARN level message. For example:
> > Unsupported value for "fsync" : "ifPossible"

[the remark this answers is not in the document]
alcaeus: I agree. The spec requires warnings for unsupported values, and we shouldn't make an exception for this one configuration option.

source/server-discovery-and-monitoring/tests/unified/serverMonitoringMode.json
ShaneHarvey: Done.

source/faas-automated-testing/faas-automated-testing.rst
ShaneHarvey: I think this idea is worthwhile but as a new drivers ticket since this PR doesn't change the OP_QUERY vs OP_MSG behavior.

source/uri-options/tests/sdam-options.yml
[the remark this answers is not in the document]
jyemin: ok, i relent.

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -1140,6 +1185,32 @@ the "awaited" field on server heartbeat events so that applications can
...
+up spending a long time in a catch up phase processing outdated responses.
+This problem was discovered in `DRIVERS-2246`_.
+
+We decided to make polling the default behavior when running on FaaS platforms
+like AWS Lambda to improve scalability, performance, and reliability.
+
+Why introduce a knob for serverMonitoringMode?
+''''''''''''''''''''''''''''''''''''''''''''''
+
+The serverMonitoringMode knob provides an workaround in cases where the polling
```
jyemin, line 1210:
```suggestion
The serverMonitoringMode knob provides a workaround in cases where the polling
```

source/server-discovery-and-monitoring/tests/unified/serverMonitoringMode.yml
```diff
@@ -0,0 +1,67 @@
+description: serverMonitoringMode
+
+schemaVersion: "1.3"
+
+tests:
```
jyemin, line 5: I was expecting these tests to assert on the value of `ServerHeartbeatSucceedEvent.awaited`.  As it is I'm not clear what use these tests are.  Can you explain?

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -584,8 +622,8 @@ current monitoring connection. (See `Drivers cancel in-progress monitor checks`_
 Polling Protocol
 ''''''''''''''''
 
-The polling protocol is used to monitor MongoDB <= 4.4 servers. The client
-`checks`_ a server with a hello or legacy hello command and then sleeps for
+The polling protocol is used to monitor MongoDB <= 4.4 servers or when `streaming is disabled`_.
```
prestonvasquez, line 625: Should this be "<= 4.4" or "< 4.4" ? It looks like streaming could be enabled for 4.4 versions of the server, testing on 4.4.24 it looks like the hello command contains the topology version.

prestonvasquez (approved):

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -661,6 +699,12 @@ The event API here is assumed to be like the standard `Python Event
         heartbeatFrequencyMS = heartbeatFrequencyMS
         minHeartbeatFrequencyMS = 500
         stableApi = stableApi
+        if serverMonitoringMode == "stream":
+            streamingEnabled = True
+        elif serverMonitoringMode == "poll":
+            streamingEnabled = False
+        else:  # serverMonitoringMode == "auto"
```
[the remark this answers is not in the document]
ShaneHarvey: It's meant to be informative, as in 'this else handles the serverMonitoringMode == "auto" case'. Do you have an alternative suggestion?

source/server-discovery-and-monitoring/server-monitoring.rst
ShaneHarvey: Good catch. Fixed.

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -700,7 +742,10 @@ The event API here is assumed to be like the standard `Python Event
             serverSupportsStreaming = description.type != Unknown and description.topologyVersion != Null
             connectionIsStreaming = connection != Null and connection.moreToCome
             transitionedWithNetworkError = isNetworkError(description.error) and previousDescription.type != Unknown
-            if serverSupportsStreaming or connectionIsStreaming or transitionedWithNetworkError:
+            if streamingEnabled and serverSupportsStreaming and not rttMonitor.started:
+                # Start the RttMonitor.
+                rttMonitor.run()
```
matthewdale, line 747: Is there a reason we want to disable the RTT monitor for servers that don't support streaming heartbeat (i.e. server versions before 4.4)?

From reading the rationale "Why disable the streaming protocol on FaaS platforms like AWS Lambda?", it sounds like the goal of not starting the RTT monitor is to reduce open connections for FaaS environments. However, the suggested change would disable the RTT monitor for all databases earlier than 4.4, not just FaaS environments.

baileympearson: This isn't a new requirement - drivers shouldn't be creating RTT monitors when the polling protocol is used.  The rtt monitors are only necessary when using the awaitable hello protocol, because then a round trip to the server can take up to `maxAwaitTimeMS`.  https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-monitoring.rst#measuring-rtt

source/server-discovery-and-monitoring/server-monitoring.rst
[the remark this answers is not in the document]
prestonvasquez: No this makes sense, thank you

source/server-discovery-and-monitoring/server-monitoring.rst
matthewdale: The spec wording and pseudo-code don't match your assertion.

The [Measuring RTT](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-monitoring.rst#measuring-rtt) section specifically says drivers can still use a separate RTT monitor for any server version:
> For consistency, clients MAY use dedicated connections to measure RTT for all servers, even those that do not support awaitable hello or legacy hello.

The current pseudo-code from the [Monitor thread](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-monitoring.rst#monitor-thread) section also describes always starting the RTT monitor, independent of server version:
```python
class Monitor(Thread):
...

  def run():
      # Start the RttMonitor.
      rttMonitor.run()
...
```

The new spec wording in this PR only suggests disabling the RTT monitor when streaming is _disabled_, but nothing about whether or not it's _supported_:
> When the streaming protocol is disabled the client MUST use the polling protocol and MUST NOT start an extra thread or connection for Measuring RTT.

Disabling the RTT monitor when streaming isn't supported is allowed by the spec. However, putting it in the pseudo-code makes it seem like a new requirement that conflicts with the spec wording.

prestonvasquez: @ShaneHarvey Can you opine on this?

ShaneHarvey: Yes this PR changes the rules to disallow using an extra thread/socket for RTT on <4.4 servers. The rationale is that the extra connection limits scaling and puts unnecessary load on the server, especially in Lambda where MongoClients are not shared. Although the motivation is mostly Lambda this is still a beneficial change for all environments.

source/server-discovery-and-monitoring/tests/unified/serverMonitoringMode.yml
ShaneHarvey: Good call out. I've updated the unified test format to support capturing these SDAM events and added the awaited=True/False assertions to these tests.

ShaneHarvey:
I've updated the Unified Test Format to version 1.17 by adding support for listening to SDAM ServerHeartbeat events and asserting on the "awaited" field. This improves the SDAM integration test. Ready for another look.

source/server-discovery-and-monitoring/server-monitoring.rst
prestonvasquez: To @matthewdale 's point, should this section from  [Measuring RTT](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-monitoring.rst#measuring-rtt) be removed or updated to reflect these rule changes? :

> For consistency, clients MAY use dedicated connections to measure RTT for all servers, even those that do not support awaitable hello or legacy hello.

ShaneHarvey: Good point. I updated this section to say clients MUST NOT dedicated connections to measure RTT in this case.

source/server-discovery-and-monitoring/server-monitoring.rst

source/server-discovery-and-monitoring/server-monitoring.rst
matthewdale: Thanks for the clarification. Does that imply that the polling monitor should measure RTT (e.g. to support CSOT)? For the Go driver, that would significantly increases the scope of this change because we'd have to build RTT monitoring capability into the polling monitor.

ShaneHarvey: I'm confused by this question because the polling protocol always included RTT monitoring.

source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -91,6 +91,28 @@ Round trip time. The client's measurement of the duration of one hello or legacy
...
+
+A Function-as-a-Service (FaaS) environment like AWS Lambda.
+
+serverMonitoringMode
+````````````````````
+
+The serverMonitoringMode option configures which server monitoring protocol to use. Valid modes are
+"stream", "poll", or "auto". The default value MUST be "auto":
+
+- With "stream" mode, the client MUST use the streaming protocol when the server supports
```
baileympearson, line 105: Is there a reason we don't throw an error when a user explicitly requests the streaming protocol but is using an incompatible server?

source/server-discovery-and-monitoring/server-monitoring.rst
matthewdale: Looking closer, the actual change in the Go driver would probably be be minimal. It does add complexity because there is no longer a single source of RTT data for CSOT, but the change only impacts a few lines of code.

source/server-discovery-and-monitoring/server-monitoring.rst
ShaneHarvey: There are a few reasons:
- "stream" mode is the same as how driver's work today. Throwing an error would introduce even more new behavior.
- Better compatibility. It's theoretically possible for the server to dynamically disable streaming support in which case we'd want the drivers to keep working.
- If a user really wants to raise an error in this case they can easily check for streaming support themselves.

baileympearson: Compatibility is something I hadn't considered, thanks.

baileympearson (approved):

prestonvasquez (approved):

matthewdale (approved):
Looks good 👍

PR #1453, opened 2023-08-21
Sync retryable reads with SDAM

prestonvasquez:
This PR proposes a re-phrasing of the "[Why are read operations only retried once by default?](https://github.com/mongodb/specifications/blob/master/source/retryable-reads/retryable-reads.rst#id40)" section of retryable reads to fix inconsistencies between the retryable reads and SDAM specifications. From the retryable reads [Q&A](https://github.com/mongodb/specifications/blob/master/source/retryable-reads/retryable-reads.rst#why-are-read-operations-only-retried-once-by-default) on why read operations are only retried once:

> In the case of a transient error, the driver will mark the server as “unknown” per the [SDAM](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst) spec.

From the SDAM rationale for “[Other transient errors](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#other-transient-errors)” :
>There are other transient errors a server may return, e.g. retryable errors listed in the retryable writes spec. SDAM does not consider these because they do not imply the connected server should be marked as “Unknown”.

- [x] Update changelog.
~~- [ ] Test changes in at least one language driver.~~
~~- [ ] Test these changes against all server versions and topologies (including standalone, replica set, sharded clusters, and serverless).~~

### DIFF
--- a/source/retryable-reads/retryable-reads.rst
+++ b/source/retryable-reads/retryable-reads.rst
@@ -678,6 +678,7 @@ degraded performance can simply disable ``retryableReads``.
 Changelog
 =========
 
+:2023-08-21: Update Q&A that contradicts SDAM transient error logic
 :2022-11-09: CLAM must apply both events and log messages.
 :2022-10-18: When CSOT is enabled multiple retry attempts may occur.
 :2022-10-05: Remove spec front matter, move footnote, and reformat changelog.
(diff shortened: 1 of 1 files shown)


kkloberdanz (approved):

ShaneHarvey:
Thanks Preston!

PR #1454, opened 2023-08-25
DRIVERS-2570 Report docker or kubernetes environment in handshake

abr-egn:

- [✓] Update changelog.
- [✓] Test changes in at least one language driver.
- [✓] Test these changes against all server versions and topologies (including standalone, replica set, sharded clusters, and serverless).

https://github.com/mongodb/mongo-rust-driver/pull/940 implements this in the Rust driver.

### DIFF
--- a/source/mongodb-handshake/handshake.rst
+++ b/source/mongodb-handshake/handshake.rst
@@ -379,7 +379,7 @@ Container runtime information is captured in ``client.env.container``.
 ``client.env.container.runtime`` SHOULD be set to ``"docker"`` if the file ``.dockerenv``
 exists in the root directory.
 
-``client.env.container.orchestrator`` SHOULD be set to ``"kubernetes"`` if the environment
+``client.env.container.orchestrator`` MUST be set to ``"kubernetes"`` if the environment
 variable ``KUBERNETES_SERVICE_HOST`` is populated.
 
 If no fields of ``client.env.container`` would be populated, ``client.env.container`` MUST
--- a/source/mongodb-handshake/handshake.rst
+++ b/source/mongodb-handshake/handshake.rst
@@ -167,10 +167,15 @@ provided as a BSON object. This object has the following structure::
             platform: "<string>",
             /* OPTIONAL */
             env: {
-                name: "<string>",         /* REQUIRED */
+                name: "<string>",         /* OPTIONAL */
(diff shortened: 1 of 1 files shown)


source/mongodb-handshake/handshake.rst
```diff
@@ -355,8 +371,19 @@ be populated:
...
 corresponding ``client.env`` field to be omitted and MUST NOT cause a user-visible error.
 
-The contents of ``client.env`` MUST be adjusted to keep the handshake below the size limit;
-see `Limitations`_ for specifics.
+Container
+^^^^^^^^^
+
+Container runtime information is captured in ``client.env.container``.
+
+``client.env.container.runtime`` SHOULD be set to ``"docker"`` if the file ``.dockerenv``
```
alcaeus, line 379: What's the reason for using `SHOULD` here?  Usually we opt for `MUST` in specifications unless there are reasons to not mandate it. I don't see why these should be optional if we detect a container runtime or orchestrator.

```suggestion
``client.env.container.runtime`` MUST be set to ``"docker"`` if the file ``.dockerenv``
```

source/mongodb-handshake/handshake.rst
```diff
@@ -355,8 +371,19 @@ be populated:
...
+exists in the root directory.
+
+``client.env.container.orchestrator`` SHOULD be set to ``"kubernetes"`` if the environment
```
alcaeus, line 382:
```suggestion
``client.env.container.orchestrator`` MUST be set to ``"kubernetes"`` if the environment
```

source/mongodb-handshake/handshake.rst
```diff
@@ -167,10 +167,15 @@ provided as a BSON object. This object has the following structure::
             platform: "<string>",
             /* OPTIONAL */
             env: {
-                name: "<string>",         /* REQUIRED */
+                name: "<string>",         /* OPTIONAL */
```
alcaeus, line 170: I'm checking to see if `env.name` is expected to be reported, as changing this field to optional may mess up internal processes relying on this metadata.

alcaeus (changes requested):

source/mongodb-handshake/handshake.rst
alcaeus: Update: this change is fine, no problem with making `env.name` optional.

PR #1459, opened 2023-09-01
DRIVERS-2713: Run legacy transaction tests on LB topologies

jmikola:
https://jira.mongodb.org/browse/DRIVERS-2713

d2de77692b727b805168eeba02c3ab29e6a1d3e2 previously added "load-balanced" to other legacy test formats but omitted the transactions spec.

This change ensures legacy transaction tests can be run on LB topologies, including Atlas Serverless (where "serverless: forbid" is not specified).

- [x] Update changelog.


jyemin:
I ran these on Java driver and these tests failed for Serverless:

- retryable-commit.json: commitTransaction fails after two errors
- retryable-commit.json: commitTransaction applies majority write concern on retries
- retryable-commit.json: commitTransaction succeeds after connection error
- retryable-commit.json: commitTransaction succeeds after InterruptedAtShutdown
- retryable-commit.json: commitTransaction succeeds after ShutdownInProgress

I'm not sure if this a Serverless issue or a Java driver issue, so I'd like a couple of other drivers to test these changes before merging this.  We can always add `serverless: forbid` to the failing tests and open a ticket against Serverless to address the failures..

source/transactions/tests/legacy/mongos-pin-auto-tests.py
```diff
@@ -16,7 +16,7 @@
 runOn:
     -
         minServerVersion: "4.1.8"
-        topology: ["sharded"]
+        topology: ["sharded", "load-balanced"]
```
jmikola, line 19: Adding "load-balanced" here and in `pin-mongos.yml` may be problematic, as I observed some test failures in PHPLIB (see: [task failure](https://parsley.mongodb.com/test/mongo_php_library_test_loadBalanced__php_edge_versions~latest_stable_mongodb_versions~6.0_driver_versions~latest_stable_os~debian11_test_loadBalanced_patch_f6ce2119fad23641df100248c11f081f15ad6cb5_64f22ec72fbabe969b92bee1_23_09_01_18_35_31/0/4b0564721909a64a107218d00faa4e06?bookmarks=0,8&shareLine=1)).

This could be related to the LB spec defining separate rules for [connection pinning](https://github.com/mongodb/specifications/blob/master/source/load-balancers/load-balancers.rst#connection-pinning), which might not be compatible with the assumptions made in these tests for normal [mongos pinning](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#mongos-pinning).

@jyemin: Any insights here? Does Java have any problem passing these pinning tests on LB and/or serverless?

jmikola: I'm pretty sure the assertion failure on PHP here is by design, as we use [`mongoc_client_session_get_server_id()`](https://mongoc.org/libmongoc/current/mongoc_client_session_get_server_id.html) to check for a pinned session, and a LB topology only has a single server so it's unlikely that method would return anything. Furthermore, PHP uses libmongoc's single-threaded mode, so we are implicitly pinned to a single connection at all times (LB spec's rules for [connection pinning](https://github.com/mongodb/specifications/blob/master/source/load-balancers/load-balancers.rst#connection-pinning) wouldn't even come into play).

That said, I'm curious if these tests might still apply to other languages, in which case I should only skip them in PHP's test suite (and libmongoc can do the same if running in single-threaded mode).

jmikola: Additionally, several other pinning tests fail due to [`recoveryToken` fields](https://github.com/mongodb/specifications/blob/master/source/transactions/transactions.rst#recoverytoken-field) not being found in outgoing `commitTransaction` and `abortTransaction` commands (see: [task failure](https://parsley.mongodb.com/test/mongo_php_library_test_loadBalanced__php_edge_versions~latest_stable_mongodb_versions~6.0_driver_versions~latest_stable_os~debian11_test_loadBalanced_patch_f6ce2119fad23641df100248c11f081f15ad6cb5_64f22ec72fbabe969b92bee1_23_09_01_18_35_31/0/7589ac969195a0bb11acdccb96538e4a?bookmarks=0,7&shareLine=1)).

jmikola: @jyemin: I confirmed with @kevinAlbs that libmongoc intentionally does not pin LB connections, since a client would only have one connection to each server (libmongoc implements client pooling instead of CMAP-compliant connection pooling). Given that, I think I'll just need to skip these tests for PHP.

I'll defer to you to confirm that all of these pass for Java, though.

source/transactions/tests/legacy/retryable-commit-errorLabels.yml
```diff
@@ -1,7 +1,7 @@
 runOn:
     -
         minServerVersion: "4.3.1"
-        topology: ["replicaset", "sharded"]
+        topology: ["replicaset", "sharded", "load-balanced"]
```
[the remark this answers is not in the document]
jyemin: > I did not observe any failures in retryable-commit-errorLabels, but there were five failures from retryable-commit:

Those look like the same failures as in Java driver.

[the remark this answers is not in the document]
jyemin: Commented on the wrong file!  This should be on r`etryable-commit.yml`.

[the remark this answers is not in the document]
nbbeeken: I can't recall but going out on a limb wasn't there a period when serverless didn't support failPoints? From today's perspective if it targets LB then they should work on serverless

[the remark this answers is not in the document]
jmikola: I went over these tests with @ShaneHarvey this afternoon and we compared them with other legacy transaction spec tests that _are_ skipped on serverless. Those other tests include the following comments for `serverless: "forbid"`:

> serverless proxy doesn't append error labels to errors in transactions caused by failpoints (CLOUDP-88216)

This appears in the `error-labels`, `mongos-pin-auto`, and `pin-mongos` tests, which include assertions for labels populated by the server. That doesn't apply to the `retryable-commit-errorLabels` test we're looking at here, which only entails client-side labels due to `closeConnection: true` being used in the fail point.

> serverless proxy doesn't use recovery tokens

This is used in `mongos-recovery-token`, which asserts for a `recoveryToken` field in outgoing commands. This also doesn't apply to `retryable-commit-errorLabels`.

----

Based on the [task log from PHPLIB](https://parsley.mongodb.com/test/mongo_php_library_serverless__php_edge_versions~latest_stable_driver_versions~latest_stable_os~debian11_test_serverless_patch_f6ce2119fad23641df100248c11f081f15ad6cb5_64f77843a4cf47743fbc8ac9_23_09_05_18_50_42/0/7a3229a9e2a4c79e70b19ab8845be4fa?bookmarks=0,27&shareLine=1) (from https://github.com/mongodb/mongo-php-library/pull/1148), we think the "Transaction with { txnNumber: 1 } has been aborted" error originates from the third `commitTransaction` operation in the test. I'm going to spin up a serverless environment and test this locally to confirm and will follow up here once I have more information.

source/transactions/tests/legacy/retryable-commit-forbid_serverless.yml
```diff
@@ -0,0 +1,410 @@
...
+    -
+        minServerVersion: "4.1.8"
+        topology: ["sharded", "load-balanced"]
+        # This file contains tests originally found in retryable-commit.yml that
+        # are incompatible with serverless:
+        #
+        #  * Failing commitTransaction with closeConnection:true may abort the
+        #    transaction (CLOUDP-202309).
+        #  * Fail points using state change errors (e.g. InterruptedAtShutdown,
+        #    ShutdownInProgress) may remain enabled indefinitely (CLOUDP-84573).
```
jmikola, line 14: This comment addresses the discussion in https://github.com/mongodb/specifications/pull/1459#discussion_r1316000236

source/transactions/tests/legacy/retryable-commit-errorLabels.yml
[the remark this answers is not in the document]
jmikola: Cross-referencing this with https://github.com/mongodb/specifications/pull/1459/files#r1339312910

I split out the incompatible tests into a `retryable-commit-forbid_serverless.yml` file with the following comment:

> This file contains tests originally found in retryable-commit.yml that are incompatible with serverless:
>
> * Failing commitTransaction with closeConnection:true may abort the transaction ([CLOUDP-202309](https://jira.mongodb.org/browse/CLOUDP-202309)).
> * Fail points using state change errors (e.g. InterruptedAtShutdown, ShutdownInProgress) may remain enabled indefinitely ([CLOUDP-84573](https://jira.mongodb.org/browse/CLOUDP-88216)).

alcaeus (approved):

source/transactions/tests/legacy/mongos-pin-auto-tests.py
jyemin: I think I ran all these tests when I first reviewed the PR and the only failures were the five that failed on Serverless.

jyemin (approved):
LGTM

source/transactions/tests/README.rst
```diff
@@ -629,6 +629,7 @@ is the only command allowed in a sharded transaction that uses the
 Changelog
 =========
 
+:2023-09-01: Add ``load-balanced`` to test topology requirements.
```
jmikola, line 632:
```suggestion
:2023-09-28: Add ``load-balanced`` to test topology requirements.
```

PR #1462, opened 2023-09-06
DRIVERS-2698: clarify that the unified test format is only intended for enterprise servers

baileympearson:

- [ ] Update changelog.

### DIFF
--- a/source/unified-test-format/unified-test-format.rst
+++ b/source/unified-test-format/unified-test-format.rst
@@ -81,7 +81,8 @@ Iterable
 Server Compatibility
 --------------------
 
-The unified test format is intended to only run against enterprise servers.
+The unified test format is intended to only run against enterprise servers.  Additionally, the 
+unified test format makes use of the `FailCommand <enableTestCommands>`_ feature of mongodb.  The server MUST be launched with the ``enableTestCommands`` parameter enabled.
 
 Schema Version
 --------------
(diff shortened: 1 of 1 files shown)


source/unified-test-format/unified-test-format.rst
```diff
@@ -78,6 +78,10 @@ Iterable
   spec uses this as the return value for ``find`` and permit API flexibility
   rather than stipulate that a cursor object be returned directly.
 
+Server Compatibility
+--------------------
+
+The unified test format is intended to only run against enterprise servers.
```
jyemin, line 84: Consider adding: `with enableTestCommand=true` (or whatever the syntax is)

baileympearson: Is the suggestion to clarify which tests / syntax doesn't work against community servers?

jyemin: Yes, it seems similar to the use case presented in DRIVERS-2698: if you try to run the tests without this they will fail (and in great numbers!).

jyemin (approved):

source/unified-test-format/unified-test-format.rst
```diff
@@ -78,6 +78,11 @@ Iterable
   spec uses this as the return value for ``find`` and permit API flexibility
   rather than stipulate that a cursor object be returned directly.
 
+Server Compatibility
+--------------------
+
+The unified test format is intended to only run against enterprise servers.  Additionally, the 
+unified test format makes use of the `FailCommand <enableTestCommands>`_ feature of mongodb.  The server MUST be launched with the ``enableTestCommands`` parameter enabled.
```
jmikola, line 85:
```suggestion
The unified test format is intended to only run against enterprise servers. Additionally, many tests make use of `Server Fail Points`_, so servers MUST be launched with the ``enableTestCommands`` parameter enabled.
```

jmikola (changes requested):
See above.

jmikola (approved):

alcaeus (approved):

PR #1465, opened 2023-10-04
Hide generated JSON spec test files from displaying in diffs by default

jmikola:
See: https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

We recently started doing this in the PHP driver repositories for generated code (https://github.com/mongodb/mongo-php-driver/pull/1473#discussion_r1338192073).


jyemin:
I love this.  Thank you.

PR #1466, opened 2023-10-04
DRIVERS-2695 Add a case of CSOT not enabled for the retryable writes.

qingyang-hu:
- Add the case the pseudo-code for the retryable writes specification: CSOT is not enabled and one retry was attempted.
- Fix some syntax errors.

- [x] Update changelog.
- [ ] ~~Make sure there are generated JSON files from the YAML test files.~~
- [ ] ~~Test changes in at least one language driver.~~
- [ ] ~~Test these changes against all server versions and topologies (including standalone, replica set, sharded clusters, and serverless).~~

Driver changes are not needed.

### DIFF
--- a/source/retryable-writes/retryable-writes.rst
+++ b/source/retryable-writes/retryable-writes.rst
@@ -455,6 +455,7 @@ The above rules are implemented in the following pseudo-code:
     retryableCommand = addTransactionIdToCommand(command, session);
 
     Exception previousError = null;
+    retrying = false;
     while true {
       try {
         return executeCommand(server, retryableCommand);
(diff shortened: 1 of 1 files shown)


kkloberdanz (approved):
Looks good!

source/retryable-writes/retryable-writes.rst
```diff
@@ -516,8 +516,24 @@ The above rules are implemented in the following pseudo-code:
...
+         * propagate. */
+        try {
+          return executeCommand(server, retryableCommand);
+        } catch (DriverException ignoredError) {
+          throw previousError;
+        } catch (Exception secondError) {
+          handleError(secondError);
+          throw secondError;
+        }
+        break
```
ShaneHarvey, line 534: This doesn't account for NoWritesPerformed or error code 20. Is it possible to avoid duplicating the call to executeCommand and other logic here? Perhaps:
```
if ((timeoutMS == null && retrying) || (timeoutMS != null && isExpired(timeoutMS))) {
    throw previousError;
}
retrying = True;
```

ShaneHarvey (changes requested):

source/retryable-writes/retryable-writes.rst
```diff
@@ -519,23 +520,15 @@ The above rules are implemented in the following pseudo-code:
       if (timeoutMS == null) {
```
qingyang-hu, line 520: I have not merged the conditions to keep the reference code straightforward with the comments.

ShaneHarvey (approved):
LGTM!

kkloberdanz (approved):

PR #1460, opened 2023-09-05
DRIVERS-2616 OIDC-SASL Follow-Up

blink1073:
Please complete the following before merging:

- [ ] Update changelog.
- [ ] Make sure there are generated JSON files from the YAML test files.
- [ ] Test changes in at least one language driver.
- [ ] Test these changes against all server versions and topologies (including standalone, replica set, sharded clusters, and serverless).

PR #1471, opened 2023-10-22
DRIVERS-2672 Add OIDC machine workflow spec.

matthewdale:
[DRIVERS-2672](https://jira.mongodb.org/browse/DRIVERS-2672)

- [x] Update changelog.


source/auth/tests/mongodb-oidc.rst
```diff
@@ -5,31 +5,30 @@ MongoDB OIDC
...
-to the location of valid ``test_user1`` credentials at the beginning of each
-test, unless otherwise specified.
+If using local servers is preferred, using the `Local Testing`_ method,
+use ``mongodb://localhost/?authMechanism=MONGODB-OIDC`` for ``MONGODB_URI_SINGLE`` and
+``mongodb://localhost:27018/?authMechanism=MONGODB-OIDC&directConnection=true&readPreference=secondaryPreferred``
+for ``MONGODB_URI_MULTI`` because the other server is a secondary on a replica set, on port ``27018``.
 
-Unless otherwise specified, tests will use a URL
-of the form ``mongodb://localhost/?authMechanism=MONGODB-OIDC``.
+The default OIDC client used in the tests will be configured with ``MONGODB_URI_SINGLE`` and a valid request callback handler
```
durran, line 27: Should this be implemented with `CUSTOM_TOKEN_CALLBACK` and `PROVIDER_NAME=custom`?

source/auth/tests/mongodb-oidc.rst
```diff
@@ -5,31 +5,30 @@ MongoDB OIDC
...
-``directConnection=true`` and ``readPreference=secondaryPreferred`` are needed because the server is a secondary on a replica set, on port ``27018``.
+Drivers MUST be able to authenticate against a server configured with either one or two configured identity providers.
 
-Drivers will need to be able to internally query and clear the cached
-credentials to verify usage for testing purposes.  Clearing the cache
-means removing all data from the cache, including ``OIDCMechanismServerStep1``
-information.
+Note that typically the preconfigured Atlas Dev clusters are used for testing, in Evergreen and localy.  The URIs can be fetched
+from the ``drivers/oidc`` Secrets vault, see `vault instructions`_.  Use ``OIDC_ATLAS_URI_SINGLE`` for ``MONGODB_URI_SINGLE`` and
+``OIDC_ATLAS_URI_MULTI`` for ``OIDC_ATLAS_URI_MULTI``.
```
durran, line 20: Should this be `MONGODB_URI_MULTI` as the tests use below?

source/auth/tests/mongodb-oidc.rst
```diff
@@ -39,149 +38,65 @@ is one principal configured.
...
-  - Close the client.
-  - Create a new client with the callbacks.
-  - Run a find operation that succeeds.
-  - Close the client.
-- Join the two threads or simultaneously call the async operations
-- Ensure that the request callback has been called once, and the refresh
-  callback has been called twice, or that no async function has been
-  entered simultaneously.
-
-AWS Automatic Auth
```
durran, line 110: Shouldn't we be keeping these tests and changing the URIs from localhost to `MONGODB_URI_SINGLE` and `MONGODB_URI_MULTI` with a `PROVIDER_NAME` of aws?

durran (changes requested):
Initial comments.

source/auth/auth.rst
```diff
@@ -1251,172 +1212,234 @@ mechanism
...
+This value is used in the ``Reauthentication`` section below.
 
-Caching Credentials
-```````````````````
+Speculative Authentication
+``````````````````````````
+Drivers MUST implement speculative authentication for MONGODB-OIDC during the ``hello`` handshake.
+If there is a cached Access Token the ``JwtStepRequest`` SASL command will be used as the speculation command.
+If there is no cached Access Token, the ``PrincipalStepRequest`` will be used as the speculation command.
+The driver MUST NOT call the callback during speculative authentication.
```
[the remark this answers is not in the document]
blink1073: I think we want two different callbacks, but I still think we shouldn't call it, because for Azure and perhaps other "custom" types there is an http request.

source/auth/tests/legacy/connection-string.yml
```diff
@@ -401,44 +401,13 @@ tests:
     mechanism: MONGODB-OIDC
     mechanism_properties:
       REQUEST_TOKEN_CALLBACK: true
-- description: should recognise the mechanism with aws device (MONGODB-OIDC)
```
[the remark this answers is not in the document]
blink1073: Note that this PR also **contains** #1460

source/auth/tests/mongodb-oidc.rst
blink1073: No, it will depend on where it is being run.  We'll want to use the same unified tests on Azure, AWS, and GCP, hence the placeholder.

source/auth/tests/mongodb-oidc.rst
blink1073: `MONGODB_URI_MULTI` will only be used for human workflows, which will continue to be prose tests only.

source/auth/tests/legacy/connection-string.yml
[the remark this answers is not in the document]
durran: Ah ok. Should we just close that PR then to avoid confusion?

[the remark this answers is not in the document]
blink1073: Done

source/auth/auth.rst
```diff
@@ -1175,31 +1175,39 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+
+  TODO: Is there a 1:1 corresponding OIDC spec flow?
+
+- Workforce Identity Federation - Also called the "human authentication flow",
+  this flow is intended to authorize access for end users and typically requires
+  human interaction to complete (e.g. clicking a button in a browser).
 
-- Machine authentication workflow
-- Human authentication workflow
+  TODO: Is there a 1:1 corresponding OIDC spec flow?
```
addaleax, line 1198: No, all authentication flows described in the OIDC spec are human authentication flows. I’d argue that it’s a bit confusing to refer to workload/workforce identity federation as “machine authentication flow”/“human authentication flow”, neither of these are really single auth flows in the sense of OIDC.

addaleax, line 1198: No, all authentication flows described in the OIDC spec are human authentication flows. I’d argue that it’s a bit confusing to refer to workload/workforce identity federation as “machine authentication flow”/“human authentication flow”, neither of these are really single auth flows in the sense of OIDC.

source/auth/auth.rst
```diff
@@ -1213,14 +1221,113 @@ mechanism
 mechanism_properties
     PROVIDER_NAME
         Drivers MUST allow the user to specify the name of the OIDC service to
-        use to obtain credentials. MUST be one of ["aws", "custom"].
+        use to obtain credentials. MUST be one of ["aws"].
```
addaleax, line 1224: Is custom callback usage just the expected default here, if it can no longer be specified in the `PROVIDER_NAME` field?

source/auth/auth.rst
```diff
@@ -1213,14 +1221,113 @@ mechanism
...
+
+The signature of the callback is up to the driver's discretion, but the driver
+MUST ensure that additional optional parameters can be added to the callback
+signature in the future. An example might look like:
+
+.. code:: typescript
+
+  interface CallbackParameters {
+      // All key-value pairs from the AUTH_MECHANISM_PROPERTIES parameter.
+      mechanismProperties Object;
```
addaleax, line 1261:
```suggestion
      mechanismProperties: Object;
```

addaleax, line 1261:
```suggestion
      mechanismProperties: Object;
```

source/auth/auth.rst
```diff
@@ -1213,14 +1221,113 @@ mechanism
...
+
+- ``timeoutMS``: A timeout, in milliseconds, deadline, or ``timeoutContext``.
+- ``mechanismProperties``: All key-value pairs from the
+  AUTH_MECHANISM_PROPERTIES parameter.
+- ``invalidatedToken``: Optional. The previous token that was invalidated after
+  receiving a ``ReauthenticationRequired`` error.
+
+The callback MUST return an OIDC access token, formatted as a base64-encoded
+JWT.
+TODO: Is that correct?
```
addaleax, line 1251: Not sure whether this TODO refers to this specifically, but “base64-encoded JWT” might be confusing language; JWTs consist of period-separated base64url-encoded parts already by definition, so “base64-encoded JWT” sounds like the expectation is that there is an additional round of base64 encoding happening on the full JWT.

```suggestion
The callback MUST return an OIDC access token in JWT format.
```

addaleax, line 1251: Not sure whether this TODO refers to this specifically, but “base64-encoded JWT” might be confusing language; JWTs consist of period-separated base64url-encoded parts already by definition, so “base64-encoded JWT” sounds like the expectation is that there is an additional round of base64 encoding happening on the full JWT.

```suggestion
The callback MUST return an OIDC access token in JWT format.
```

addaleax: (sorry, not sure what was up with Github here)

source/auth/auth.rst
```diff
@@ -1175,70 +1175,148 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+connection establishment deadline (TODO: Does that make sense?).
+
+Callbacks can be synchronous and/or asynchronous, depending on the driver
+and/or language.  Asynchronous callbacks should be preferred when other
+operations in the driver use asynchronous functions.
+
+The driver MUST pass the following information to the callback:
+
+- ``timeoutMS``: A timeout, in milliseconds, deadline, or ``timeoutContext``.
+- ``mechanismProperties``: All key-value pairs from the
```
jyemin, line 1244: This seems odd.  The driver will pass `["PROVIDER_NAME": "aws"]` but will have no way of enforcing that "aws" is actually used by the callback. Also, isn't the callback itself specified as an auth mechanism property?

Seems like if a callback is provided, it shouldn't need any of the mechanism properties and therefore not should be required when a callback is provided.

source/auth/auth.rst
```diff
@@ -1175,70 +1175,146 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+(391), the driver MUST clear the access Token cached on the connection object
+and and reauthenticate the connection as described in ``Reauthentication``
+section below.
+
+Speculative Authentication
+``````````````````````````
+
+Drivers MUST implement speculative authentication for MONGODB-OIDC during the
+``hello`` handshake. If there is a cached Access Token on the ``MongoClient``,
+use it for authentication. Otherwise, call the configured workload callback to
```
blink1073, line 1286: I thought we were not calling any callbacks during speculative auth, since it could still be a lengthy process.

source/auth/auth.rst
matthewdale: Agreed, I was trying to relate the driver process to public OIDC documentation as much as possible, but I realized that was difficult because we're really satisfying a MongoDB auth mechanism that is based on OIDC, but is not an OIDC flow itself. I've updated the document to remove the explicit separation of "machine" and "human" flows, instead assuming the "machine" flow is the normal use case for "MONGODB-OIDC", and listing the "human" flow as a separate section that most drivers can ignore.

source/auth/auth.rst
matthewdale: It's not the expected default. However, having to specify `PROVIDER_NAME="custom"` as well as specifying a custom callback seems redundant. Instead, I've updated the document to say that `PROVIDER_NAME` and a custom callback cannot be used together and drivers should return an error if both are specified.

source/auth/auth.rst
matthewdale: Agreed, simplified the wording based on your suggestion.

source/auth/auth.rst
matthewdale: My understanding is that the _intended_ use case of the custom machine-to-machine callback is to allow users to integrate with OIDC providers that the driver doesn't have built-in support for (using a machine-to-machine auth flow). In that case, it _should_ behave similarly to a built-in provider (e.g. AWS or Azure) and should be significantly faster than the "human" callback. However, we don't control the implementation, so a custom callback could take several seconds (it's subject to CSOT, so technically could take up to 10-30s), which is longer than we'd expect to wait for most other auth mechanisms.

Should the "no callbacks during speculative auth" rule be extended to the custom machine-to-machine callback? Is there a case where a slow callback is a problem for creating the speculative auth payload (i.e. the initial `hello` command) but not a problem for post-`hello` auth commands (e.g. `saslStart`)?

source/auth/auth.rst
```diff
@@ -1175,70 +1175,146 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+
+The driver MUST pass the following information to the callback:
+
+- ``timeoutMS``: A timeout, in milliseconds, deadline, or ``timeoutContext``.
+- ``mechanismProperties``: All key-value pairs from the
+  AUTH_MECHANISM_PROPERTIES parameter.
+- ``invalidatedToken``: Optional. The previous token that was invalidated after
+  receiving a ``ReauthenticationRequired`` error.
+
+The callback MUST return an OIDC access token in JWT format.
```
katcharov, line 1241: Is this intended to be a documentation requirement? (Drivers could enforce this by introducing a type that validates user input, but I believe we were fine with returning strings.)

source/auth/auth.rst
```diff
@@ -1175,70 +1175,146 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+signature in the future. An example might look like:
+
+.. code:: typescript
+
+  interface CallbackParameters {
+      // All key-value pairs from the AUTH_MECHANISM_PROPERTIES parameter.
+      mechanismProperties: Object;
+
+      // The refresh token, if applicable, to be used by the callback to request a new token from the issuer.
+      invalidatedToken: Optional<string>;
```
katcharov, line 1254: I think this comment is incorrect. The invalidated token is the "previous" accessToken (that is, the accessToken associated with the connection that has received a 391-reauth).

The user, who is expected to handle caching and locking on their end, would use this to see if the cache is invalidated (that is, if the cache's token is the same as this invalid token, the cached refreshToken should then be used to communicate with the issuer).

We might consider renaming this to `invalidatedAccessToken`, since this might be confusing for users as well.

katcharov (changes requested):

source/auth/auth.rst
matthewdale: My initial intent was to define an API that would satisfy either a custom machine-to-machine callback or a built-in callback (e.g. the future built-in Azure OIDC logic, which will need extra configs to be provided by `authMechanismProperties`). On reflection, that seems overly prescriptive of the implementation rather than just the configuration API. I also intended `authMechanismProperties` to communicate additional configuration to the custom callback, but I realized that drivers validate the keys in `authMechanismProperties`, so passing arbitrary config values isn't possible.

As far as specifying the callback name in `authMechanismProperties`, I'm currently planning to drop that from the spec. I haven't been able to discover a use case for it and implementing that logic would be unconventional or nonsensical in many driver languages.

source/auth/auth.rst
katcharov: I think the original problem was that we were worried about delays for non-authenticated hellos (ping), but we decided that this was not an issue. Later, we agreed that a server response might be missing a `speculativeAuthenticate`, and that this might be either because spec auth failed, or because it was removed by (for example) a proxy. From some of my Java notes:

> When the process is initiated during the initial handshake, speculative auth will be used. It is possible to know that speculative auth succeeded, but no reliable way to know it failed. Therefore, callbacks should not be invoked during the initial handshake.

At the time, it must have been more obvious to me why being unable to determine failure implied that (apparently all) callbacks were unworkable during initial auth (though using a cached token was ok - though, now I think we cannot tell the difference?), but now I do not immediately see why. But perhaps someone else can.

source/auth/auth.rst
addaleax: Yeah, that clarifies things. Thank you!

source/auth/auth.rst
addaleax: Generally speaking, it’s strongly recommended for clients to not inspect access tokens themselves, only for resource servers. For example, at some theoretical point in the future, MongoDB servers could add support for opaque access tokens, which would be incompatible with added JWT validation in the driver.

source/auth/auth.rst
katcharov: If the initial auth succeeded, but `speculativeAuthenticate` was absent, then we would redundantly invoke the callback a second time (since the absence implies that it might have failed). We should not invoke callbacks spuriously, so callbacks should not be invoked during speculative auth.

source/auth/tests/mongodb-oidc.rst
```diff
@@ -39,149 +38,65 @@ is one principal configured.
...
-- Create a client with a url of the form ``mongodb://test_user2@localhost:27018/?authMechanism=MONGODB-OIDC&directConnection=true&readPreference=secondaryPreferred`` and a valid OIDC request callback.
+- Create a client with ``MONGODB_URI_MULTI``, a username of ``test_user2``, and the OIDC request callback.
 - Perform a ``find`` operation that succeeds.
 - Close the client.
 
 Multiple Principal No User
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
-- Clear the cache.
-- Create a client with a url of the form ``mongodb://localhost:27018/?authMechanism=MONGODB-OIDC&directConnection=true&readPreference=secondaryPreferred`` and a valid OIDC request callback.
+- Create a client with ``MONGODB_URI_MULTI``, no username, and the OIDC request callback.
```
durran, line 66:
```suggestion
- Create a client with ``MONGODB_URI_MULTI``, no username, `authMechanism=MONGODB-OIDC`, and the OIDC request callback.
```

source/auth/tests/mongodb-oidc.rst
```diff
@@ -39,149 +38,65 @@ is one principal configured.
...
+- Create a client with ``MONGODB_URI_MULTI``, a username of ``test_user1``, and the OIDC request callback.
 - Perform a ``find`` operation that succeeds.
 - Close the client.
 
 Multiple Principal User 2
 ~~~~~~~~~~~~~~~~~~~~~~~~~
-- Clear the cache.
 - Create a request callback that reads in the generated ``test_user2`` token file.
-- Create a client with a url of the form ``mongodb://test_user2@localhost:27018/?authMechanism=MONGODB-OIDC&directConnection=true&readPreference=secondaryPreferred`` and a valid OIDC request callback.
+- Create a client with ``MONGODB_URI_MULTI``, a username of ``test_user2``, and the OIDC request callback.
```
durran, line 60:
```suggestion
- Create a client with ``MONGODB_URI_MULTI``, a username of ``test_user2``, `authMechanism=MONGODB-OIDC`, and the OIDC request callback.
```

source/auth/tests/mongodb-oidc.rst
```diff
@@ -39,149 +38,65 @@ is one principal configured.
...
+- Create a client with ``MONGODB_URI_SINGLE``, a username of ``test_user1``, and the OIDC request callback.
 - Perform a ``find`` operation that succeeds.
 - Close the client.
 
 Multiple Principal User 1
 ~~~~~~~~~~~~~~~~~~~~~~~~~
-- Clear the cache.
-- Create a request callback that returns a valid token.
-- Create a client with a url of the form  ``mongodb://test_user1@localhost:27018/?authMechanism=MONGODB-OIDC&directConnection=true&readPreference=secondaryPreferred`` and a valid OIDC request callback.
+- Create a client with ``MONGODB_URI_MULTI``, a username of ``test_user1``, and the OIDC request callback.
```
durran, line 53:
```suggestion
- Create a client with ``MONGODB_URI_MULTI``, a username of ``test_user1``, `authMechanism=MONGODB-OIDC`, and the OIDC request callback.
```

source/auth/tests/mongodb-oidc.rst
```diff
@@ -39,149 +38,65 @@ is one principal configured.
...
+- Create default OIDC client.
 - Perform a ``find`` operation. that succeeds.
 - Close the client.
 
 Single Principal Explicit Username
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-- Clear the cache.
-- Create a request callback that returns a valid token.
-- Create a client with a url of the form  ``mongodb://test_user1@localhost/?authMechanism=MONGODB-OIDC`` and the OIDC request callback.
+- Create a client with ``MONGODB_URI_SINGLE``, a username of ``test_user1``, and the OIDC request callback.
```
durran, line 47:
```suggestion
- Create a client with ``MONGODB_URI_SINGLE``, a username of ``test_user1``, `authMechanism=MONGODB-OIDC`, and the OIDC request callback.
```

source/auth/tests/mongodb-oidc.rst
```diff
@@ -39,149 +38,65 @@ is one principal configured.
 
 Single Principal Implicit Username
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-- Clear the cache.
-- Create a request callback returns a valid token.
-- Create a client that uses the default OIDC url and the request callback.
+- Create default OIDC client.
```
durran, line 41:
```suggestion
- Create default OIDC client with `authMechanism=MONGODB-OIDC`.
```

durran (changes requested):

source/auth/auth.rst
```diff
@@ -1251,172 +1327,191 @@ mechanism
...
         providing objects as mechanism property values.  Otherwise the driver MUST allow it as a MongoClientOption.
+
     ALLOWED_HOSTS
         The list of allowed hostnames or ip-addresses (ignoring ports) for
-        MongoDB connections. The hostnames may include a leading "*." wildcard, which allows for matching (potentially  nested) subdomains. ALLOWED_HOSTS is a
+        MongoDB connections. The hostnames may include a leading "\*." wildcard, which allows for matching
+        (potentially  nested) subdomains. ALLOWED_HOSTS is a
         security feature and MUST default to
-        ``["*.mongodb.net", '*.mongodb-dev.net", "*.mongodbgov.net", "localhost", "127.0.0.1", "::1"]``.
+        ``["*.mongodb.net", '*.mongodb-qa.net", '*.mongodb-dev.net", "*.mongodbgov.net", "localhost", "127.0.0.1", "::1"]``.
```
durran, line 1341:
```suggestion
        ``["*.mongodb.net", "*.mongodb-qa.net", "*.mongodb-dev.net", "*.mongodbgov.net", "localhost", "127.0.0.1", "::1"]``.
```

durran (changes requested):

source/auth/auth.rst
matthewdale: This was intended to be a documentation requirement to inform users how to implement the callback. I will update it to clarify that.

source/auth/auth.rst
```diff
@@ -1175,70 +1175,146 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+      // All key-value pairs from the AUTH_MECHANISM_PROPERTIES parameter.
+      mechanismProperties: Object;
+
+      // The refresh token, if applicable, to be used by the callback to request a new token from the issuer.
+      invalidatedToken: Optional<string>;
+  }
+
+.. code:: typescript
+
+  function oidcToken(timeoutMS: int, params CallbackParameters): string
```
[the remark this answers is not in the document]
matthewdale: 1. That makes sense, I'll move the timeout parameter to the parameters object/struct.
2. That's a good idea and is already relevant because we want callbacks to be able to return a token expiry time, if available, to help caching work more correctly. I'll change the return type to an object/struct.

source/auth/auth.rst
matthewdale: I believe there's no longer a reason to pass the invalidated token into the callback because it was only needed to allow the callback author to manage their cache. Since we're handling caching in the driver, it shouldn't be necessary, so I'm planning to remove it.

source/auth/auth.rst
```diff
@@ -1251,172 +1327,191 @@ mechanism
...
+When reauthentication is requested by the server (as a 391 error code) and MONGODB-OIDC is in use, the driver MUST perform a reauthentication.
+The driver MUST account for the case of multiple connections hitting a reauthentication error at different times.
+The driver MUST also account for a reauthenication that results from an IdP configuration change on the server.
+To accomplish these goal, the following algorithm is used to handle a reauthenication error:
 
-The driver MUST account for the case of multiple connections hitting a reauthentication error at different times, to prevent unnecessary callback calls.   If another connection has already reauthenticated, then the Access Token should not be expired.  The driver can either cache a token generation id per connection as well as in the main cache, or some other equivalent method to track whether a reauthentication has already occurred.
+- First, see if the Access Token on the MongoClient is different than the Access Token on the connection.
+   - If they are different, optimisitically try to authenticate.  On error, continue.
```
durran, line 1510:
```suggestion
   - If they are different, optimisitically try to authenticate.  On error, continue.
```

Can we clear up exactly what this means? Does this mean to use the accessToken from the client and set it on the connection? Does it mean to use the refresh token? Do we need to cache the conversationId from the original authentication flow?

durran (changes requested):

source/auth/tests/mongodb-oidc.rst
```diff
@@ -39,149 +38,65 @@ is one principal configured.
...
+- Create a client with ``MONGODB_URI_MULTI``, no username, and the OIDC request callback.
 - Assert that a ``find`` operation fails.
 - Close the client.
 
 Allowed Hosts Blocked
 ~~~~~~~~~~~~~~~~~~~~~
-- Clear the cache.
-- Create a client that uses the OIDC url and a request callback, and an
-  ``ALLOWED_HOSTS`` that is an empty list.
+- Create a default OIDC client, with an ``ALLOWED_HOSTS`` that is an empty list.
```
katcharov, line 72: Could we either split this into two files, or into two distinct sections, one for machine workflow tests, and the other for human workflow tests (which can all be omitted, for drivers that do not implement them)?

source/auth/tests/mongodb-oidc.rst
```diff
@@ -5,31 +5,30 @@ MongoDB OIDC
...
 - ``Reauthentication``
+- ``Separate Connections Avoid Extra Callback Calls``
 
 
 .. sectnum::
 
-Drivers MUST be able to authenticate using either authentication or provider
-name if there are multiple principals configured on the server.  Note that
-``directConnection=true`` and ``readPreference=secondaryPreferred`` are needed because the server is a secondary on a replica set, on port ``27018``.
+Drivers MUST be able to authenticate against a server configured with either one or two configured identity providers.
```
katcharov, line 16: Is it enough to test with either one (pick one), or should it succeed with either (both)?

katcharov, line 16:
```suggestion
In the tests below, drivers will authenticate against a server configured with one identity provider (``MONGODB_URI_SINGLE``), and against a server configured with two identity providers (``MONGODB_URI_MULTI``).
```

source/auth/tests/legacy/connection-string.yml
```diff
@@ -401,44 +401,13 @@ tests:
     mechanism: MONGODB-OIDC
     mechanism_properties:
       REQUEST_TOKEN_CALLBACK: true
```
durran, line 403: Can we update these tests to use `OIDC_TOKEN_CALLBACK` instead and remove the tests with the refresh token?

source/auth/tests/mongodb-oidc.rst
```diff
@@ -61,37 +58,91 @@ Prose Tests
...
+        failCommands: [
+          "find"
+        ],
+        closeConnection: true
+      }
+    }
+
+- Perform a ``find`` operation that fails. This is to force the ``MongoClient``
+  to cache an access token.
 - Set a fail point for ``saslStart`` commands of the form:
```
blink1073, line 85: This doesn't raise a failure because it will successfully connect in speculative auth.

source/auth/tests/mongodb-oidc.rst
```diff
@@ -61,37 +58,91 @@ Prose Tests
...
+      },
+      data: {
+        failCommands: [
+          "saslStart"
+        ],
+        errorCode: 18
+      }
+    }
+
+- Perform a ``find`` operation that fails.
```
blink1073, line 144: Same, this will succeed on spec auth.

source/auth/auth.rst
```diff
@@ -1175,70 +1175,146 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+to the callback, or some other language-appropriate mechanism.  The timeout
+deadline MUST be the same as the connection establishment deadline.
+
+Callbacks can be synchronous and/or asynchronous, depending on the driver
+and/or language.  Asynchronous callbacks should be preferred when other
+operations in the driver use asynchronous functions.
+
+The driver MUST pass the following information to the callback:
+
+- ``timeoutMS``: A timeout, in milliseconds, deadline, or ``timeoutContext``.
```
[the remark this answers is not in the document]
matthewdale: Renamed.

source/auth/auth.rst
```diff
@@ -1175,70 +1175,146 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+- ``mechanismProperties``: All key-value pairs from the
+  AUTH_MECHANISM_PROPERTIES parameter.
+- ``invalidatedToken``: Optional. The previous token that was invalidated after
+  receiving a ``ReauthenticationRequired`` error.
```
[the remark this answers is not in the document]
matthewdale: We've decided to manage the token cache in the driver, so passing the invalidated token to the callback is no longer required.

source/auth/auth.rst
matthewdale: Rewrote this section to only describe what the callback API must support.

source/auth/auth.rst
matthewdale: Removed.

source/auth/auth.rst
matthewdale: Updated to not pass the mechanism properties to the callback and require that drivers enforce that `PROVIDER_NAME` and a custom callback cannot be specified at the same time. Also updated to support either specifying the callback via a `MongoClientConfig` property or a mechanism property, depending on which is supported or idiomatic for each driver.

source/auth/auth.rst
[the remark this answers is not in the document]
matthewdale: Moved the timeout param to the `OIDCCallbackParams` and renamed it to `callbackTimeoutMS`. Also made the return type an object/struct instead of a single value.

source/auth/auth.rst
```diff
@@ -1251,172 +1246,514 @@ mechanism
...
 
-Before calling a callback, the driver MUST acquire a lock unique to the cache key.  The driver MUST ensure that credentials have not changed between when the lock was requested and when it was acquired.  The lock MUST be released
-when the callback call as finished or errored.
-This is because request callbacks may involve human interaction, and refresh
-callbacks could use refresh tokens that can only be used once.
+  interface OIDCCallbackParams {
+      callbackTimeoutMS: int;
+      version: int;
+      idpInfo: Optional<IdpInfo>;
+      refreshToken: Optional<string>;
```
addaleax, line 1399: Do we enforce this being a string? From the driver's perspective, it can be any opaque value (whatever that looks like in a given language), right?

addaleax (approved):
Looks great! Just a few very minor comments, feel free to ignore if you prefer

source/auth/tests/mongodb-oidc.rst
matthewdale: Done!

alcaeus (commented):
Unified test format changes look good. Can you please add tests for `authMechanism` similar to the `invalid/runOnRequirement-auth-type` test?

source/auth/tests/mongodb-oidc.rst
matthewdale: Replaced this test with one that works with speculative auth.

source/auth/tests/mongodb-oidc.rst
matthewdale: Replaced this test with one that works with speculative auth.

source/auth/auth.rst
```diff
@@ -1367,7 +1373,8 @@ authentication:
 - Check if the the *Client Cache* has an access token.
     - If it does, cache the returned access token in *Connection Cache* and
       optimisitically try to authenticate using the access token. If the server
-      returns ``AuthenticationFailed`` (error code 18), continue.
+      returns ``AuthenticationFailed`` (error code 18), slep 100ms then
```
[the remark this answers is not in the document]
matthewdale: Fixed!

source/auth/auth.rst
matthewdale: No, it can be any value. I changed the type to `any`.

source/auth/auth.rst
```diff
@@ -1251,172 +1246,514 @@ mechanism
...
+        an `OIDC Callback`_ using a ``MongoClient`` configuration instead of a
+        mechanism property, depending on what is conventional for the driver.
+        Drivers MUST NOT support both the ``CALLBACK`` mechanism property and
+        the ``MongoClient`` configuration.
+
+    CALLBACK_TYPE
+        The type of `OIDC Callback`. The value MUST be one of ["machine",
+        "human"]. If an `OIDC Callback`_ is configured and ``CALLBACK_TYPE`` is
+        not specified, the driver MUST raise an error. This property is only
+        required for drivers that support the `Human Authentication Flow`_.
```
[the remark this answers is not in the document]
matthewdale: Yeah, that makes sense. It would be weird if they were specified in different places. I also added a note that drivers can use a `MongoClient` config to specify the callback type if they also use a `MongoClient` config to specify the callback.

source/auth/auth.rst
```diff
@@ -1175,70 +1182,58 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
-      fields:
-          jwt:
-              description: "Compact serialized JWT with signature"
-              cpp_name: JWT
-              type: string
+In addition to the **access token**, the human authentication flow introduces
+some additional concepts:
+
+- **Identity Provider (IdP)**: A service that manages user accounts and
+  authenticates users, such as Okta or OneLogin.
```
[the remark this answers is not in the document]
matthewdale: Good point, I've added that clarification and moved the definitions to the top-level OIDC section.

source/auth/auth.rst
matthewdale: Rewrote this whole section to be clearer.

source/auth/auth.rst
```diff
@@ -1251,172 +1251,527 @@ mechanism
...
+
+    CALLBACK_TYPE
+        The type of `OIDC Callback`_. The value MUST be one of ["machine",
+        "human"]. Drivers MUST NOT allow the user to specify ``CALLBACK_TYPE``
+        in the connection string. Drivers MAY allow the user to specify the
+        callback type using a ``MongoClient`` configuration instead of a
+        mechanism property to be consistent with how the `OIDC Callback`_ is
+        configured. If an `OIDC Callback`_ is configured and ``CALLBACK_TYPE``
+        is not specified, the driver MUST raise an error. This property is only
+        required for drivers that support the `Human Authentication Flow`_.
```
jyemin, line 1276: I think we have to say that the property is required in all cases.  Otherwise introduction of the human flow in a subsequent release will be a breaking change, because an existing app that has registered a machine callback will error on upgrade.

source/auth/auth.rst
```diff
@@ -1251,172 +1251,527 @@ mechanism
...
+
+    CALLBACK
+        An `OIDC Callback`_ that returns credentials for OIDC providers that do
+        not have a built-in integraiton. Drivers MAY allow the user to specify
+        an `OIDC Callback`_ using a ``MongoClient`` configuration instead of a
+        mechanism property, depending on what is conventional for the driver.
+        Drivers MUST NOT support both the ``CALLBACK`` mechanism property and
+        the ``MongoClient`` configuration.
+
+    CALLBACK_TYPE
```
jyemin, line 1268: This property seems odd. Consider a MongoClient configuration for the OIDC callback: what should its signature be? For statically typed languages, you would define a type for each callback.  In Java, you might have:

```
interface HumanOidcCallback {
   HumanCallbackReturnType callback(HumanCallbackParams);
}

interface MachineOidcCallback {
   MachineCallbackReturnType callback(MachineCallbackParams);
}

...
   void oidcCallbackType(String type) ...
   void registerOidcCallback(...  What type should this be? In Java the only choice would be Object
```

I think it would be better, at least in Java, to instead overload the second method:

```
   void registerOidcCallback(HumanOidcCallback callback) ...
   void registerOidcCallback(MachineOidcCallback callback) ...
```

and it would be a runtime error to register both, and then the callback type is not necessary.

Also, see my comment below on callback type not being required in all cases, as it's related.

durran (approved):

matthewdale:
@alcaeus Added an invalid test for the type of `authMechanism`.

source/auth/auth.rst
matthewdale: I believe the redundant callback invocation problem is resolved by the updated cache management strategy for the human auth flow. The new logic assumes a missing specauth response means nothing and continues with non-speculative authentication with the cached access token. If the non-speculative authentication fails with the cached access token, then call for a new access token, first using a refresh token, then getting new `IdpInfo` and calling the human callback.

I'll go back through the proposed algorithms and make sure they're consistent, but I think we can remove the requirement to skip the callback during speculative authentication.

source/auth/auth.rst
```diff
@@ -1175,70 +1175,146 @@ If AWS authentication fails for any reason, the cache MUST be cleared.
...
+
+Custom
+______
+
+Drivers MUST allow the user to integrate with other OIDC providers by
+implementing a custom callback that returns an OIDC token. Drivers MUST provide
+a way for the callback to be either automatically canceled, or to cancel itself.
+This can be as a timeout argument to the callback, a cancellation context passed
+to the callback, or some other language-appropriate mechanism.  The timeout
+deadline MUST be the same as the connection establishment deadline.
```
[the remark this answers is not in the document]
matthewdale: Updated to describe timeout concepts and values consistent with the CSOT spec.

alcaeus (approved):
Unified test format changes LGTM.

source/auth/auth.rst
matthewdale: Good point, that would create a breaking change for drivers that add human auth flow support later. Based on that and your other comments, I think we should not use `CALLBACK_TYPE`. Instead, we should replace it with either a property like `USE_HUMAN_CALLBACK:true` or a property like `HUMAN_CALLBACK` that expects a callback function and is mutually exclusive with `CALLBACK`.

source/auth/auth.rst
matthewdale: In the first example, both callbacks would use the same interface, so there would only be one `registerOidcCallback` method that accepted the `OidcCallback` type. The params and return values for the human auth flow are a superset of the params and return types for the machine auth flow. To support the human auth flow, `OidcCallback` would include all params and return values for both the machine and human auth flows, and `oidcCallbackType` would specify which flow to use.

However, I see your point about the more idiomatic pattern in Java being to use method overriding to set the appropriate OIDC callback. Seems like the pattern for passing callbacks can be very different in different drivers and languages, so I think the current language in the callback API section is probably too prescriptive. I will update that section to be more permissive so drivers can implement a callback API that is idiomatic for their driver and language.

source/auth/auth.rst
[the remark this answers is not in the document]
matthewdale: Based on Jeff's feedback, my plan is to actually remove `CALLBACK_TYPE` and replace it with either a separate callback argument specifically for the human callback or a boolean value to enable the human auth flow. In either case, I will still require that it not be specified in the connection string.

source/auth/tests/mongodb-oidc.rst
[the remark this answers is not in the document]
matthewdale: Fixed!

source/auth/tests/legacy/connection-string.yml
```diff
@@ -411,46 +382,23 @@ tests:
...
-  uri: mongodb://principalName@localhost/?authMechanism=MONGODB-OIDC&PROVIDER_NAME:gcp
+- description: should throw an exception if specified provider is not supported (MONGODB-OIDC)
+  uri: mongodb://localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=PROVIDER_NAME:invalid
   valid: false
   credential:
-- description: should throw an exception if specified deviceName is not supported
-    (MONGODB-OIDC)
-  uri: mongodb://localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=PROVIDER_NAME:unexisted
+- description: should throw an exception custom callback is chosen but no callback is provided (MONGODB-OIDC)
+  uri: mongodb://localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=PROVIDER_NAME:custom
```
sanych-sun, line 394: I suppose "custom" is not allowed value for PROVIDER_NAME. As far as I remember there is only "aws" allowed so far.

source/auth/tests/legacy/connection-string.yml
```diff
@@ -350,59 +350,30 @@ tests:
...
     password:
     source: "$external"
     mechanism: MONGODB-OIDC
     mechanism_properties:
-      REQUEST_TOKEN_CALLBACK: true
-- description: should recognise the mechanism with request and refresh callback (MONGODB-OIDC)
-  uri: mongodb://localhost/?authMechanism=MONGODB-OIDC
+      PROVIDER_NAME: aws
+- description: should ignore username and password if specified for aws provider (MONGODB-OIDC)
+  uri: mongodb://user:pass@localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=PROVIDER_NAME:aws
```
sanych-sun, line 374: PROVIDER_NAME and callbacks are mutual exclusive. Callback parameter probably has to be removed from this test.

source/auth/tests/legacy/connection-string.yml
matthewdale: You're right, that test is vestigial from an earlier version of the spec. I've removed this test case because it's no longer relevant.

source/auth/tests/legacy/connection-string.yml
matthewdale: You're correct. The `callback` legacy test case parameter was never documented and is not supported by most drivers, so I intended to remove it from all test cases here, but accidentally left that one in. I've updated that test case to assert that providing a password causes a validation error, which matches the spec.

source/auth/auth.rst
```diff
@@ -1250,45 +1250,44 @@ mechanism
...
-        against a hostname that does not match any of list of allowed hosts, the
-        driver MUST raise a client-side error without invoking any user-provided
-        callbacks. This value MUST NOT be allowed in the URI connection string.
-        The hostname check MUST be performed after SRV record resolution, if
-        applicable. This property is only required for drivers that support the
-        `Human Authentication Flow`_.
+  PROVIDER_NAME
+    Drivers MUST allow the user to specify the name of a built-in OIDC provider
+    integration to use to obtain credentials. If provided, the value MUST be one
+    of ["aws"]. If both ``PROVIDER_NAME`` and an `OIDC Callback`_ or `Human
```
jyemin, line 1256: Should we call `Human Callback` `OIDC Human Callback` to be consistent with the mechanism property name?

source/auth/auth.rst
jyemin: Resolved by adding OIDC_HUMAN_CALLBACK

jyemin (approved):
One suggestion for a wording change that I leave to your discretion.

LGTM!

source/auth/auth.rst
matthewdale: I simplified the speculative auth behavior to only apply to cached access tokens. The result is that no new access tokens are fetched by any means (callback, built-in, etc) during speculative auth.

source/auth/tests/legacy/connection-string.yml
matthewdale: Removed all tests that use callbacks from the connection string tests. They are all prose tests now.

blink1073 (approved):
LGTM!

source/auth/tests/mongodb-oidc.rst
matthewdale: Rewrote this section to clarify what drivers should test.

source/auth/tests/mongodb-oidc.rst
```diff
@@ -2,458 +2,435 @@
...
-Refresh Callback Returns Null
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-- Clear the cache.
-- Create request callback that returns a valid token that will expire in a
-  minute, and a refresh callback that returns ``null``.
+**2.4 Invalid Client Configuration with Callback**
+
+- Create a ``MongoClient`` configured with an OIDC callback and auth mechanism
+  property ``PROVIDER_NAME:aws``.
+- Assert it returns a client configuration error.
```
JamesKovacs, line 96: This prose test is already covered by the unified connection string tests above. Isn't this an unnecessary duplicate?

JamesKovacs (approved):
LGTM. Check to see if the noted prose test is truly a duplicate of the unified test or not.

source/auth/tests/mongodb-oidc.rst
matthewdale: The updated Auth connection string tests don't support passing an OIDC callback, so I believe this prose test covers a case that the Auth connection string tests can't. The previous Auth connection string OIDC tests did include a "callback" param, but it was not documented in the test format README, was only implemented in 1 or 2 drivers, and was confusing with the updated OIDC callback API, so I decided to replace it with a prose test instead.

PR #1473, opened 2023-11-07
DRIVERS-2668 Add SRV test with uppercase hostname

ShaneHarvey:
[DRIVERS-2668](https://jira.mongodb.org/browse/DRIVERS-2668). Unfortunately, as noted in [DEVPROD-174](https://jira.mongodb.org/browse/DEVPROD-174), it's not possible to add the exact test case I envisioned for DRIVERS-2668 (where DNS returns uppercase letters in the SRV result). However we can still add this related test case with uppercase letters in the initial SRV hostname.

- [X] Test changes in at least one language driver. Python implementation here: https://github.com/mongodb/mongo-python-driver/pull/1423

### DIFF
--- a/source/initial-dns-seedlist-discovery/tests/replica-set/uri-with-uppercase-hostname.json
+++ b/source/initial-dns-seedlist-discovery/tests/replica-set/uri-with-uppercase-hostname.json
@@ -1,0 +1,16 @@
+{
+  "uri": "mongodb+srv://TEST1.TEST.BUILD.10GEN.CC",
+  "seeds": [
+    "localhost.test.build.10gen.cc:27017",
+    "localhost.test.build.10gen.cc:27018"
+  ],
+  "hosts": [
+    "localhost:27017",
+    "localhost:27018",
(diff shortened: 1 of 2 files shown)


dariakp:
@ShaneHarvey I was out on PTO when this landed. It looks like the original ticket has not been updated based on the most recent information about what's possible to test - could you add the new information to the ticket summary so that the final outcome (what we will actually test and what sort of bug this is meant to catch) is more obvious to the teams that go to implement this? Also, is this more limited test still sufficient to prevent the kind of issue that python originally ran into?

ShaneHarvey:
I was planning to update the ticket description once this is merged. 

> Also, is this more limited test still sufficient to prevent the kind of issue that python originally ran into?

No. This test does not cover the bug in [PYTHON-3800](https://jira.mongodb.org/browse/PYTHON-3800) but it's still useful because it ensures drivers normalize the hostname before SRV+TXT lookup.

dariakp (approved):

PR #1474, opened 2023-11-08
DRIVERS-2761 do not apply `readConcern` or `writeConcern` to search index commands

kevinAlbs:
# Summary

- Require not applying read and write concern to search index commands.
- Add unit and integration tests to ensure `readConcern` and `writeConcern` is not applied to search index commands.

Tests were run in the C++ driver in this PR: https://github.com/mongodb/mongo-cxx-driver/pull/1055

# Background & Motivation

This PR proposes adding tests to ensure each search index command does not include `readConcern` or `writeConcern`.

Unit tests apply a non-default `readConcern` and `writeConcern` to the collection:

```yml
# Set a non-default read and write concern.
uriOptions:
    readConcernLevel: local
    w: 1
```

Unit tests verify the command started event is checked:

```yml
# Expect no writeConcern or readConcern to be sent.
writeConcern: { $$exists: false }
readConcern: { $$exists: false }
```

The integration test is as an extra precaution to test against a live Atlas instance.

The C++ driver failed unit and integration tests prior to fixes applied in https://github.com/mongodb/mongo-cxx-driver/pull/1055.

---

- [x] Update changelog.
- [x] Test changes in at least one language driver. **Tested in [C++](https://github.com/mongodb/mongo-cxx-driver/pull/1055)**


baileympearson (approved):

source/index-management/index-management.rst
```diff
@@ -1136,4 +1138,6 @@ Changelog
 :2023-05-10:  Merge index enumeration and index management specs and get rid of references 
              to legacy server versions.
 :2023-05-18:  Add the search index management API.
-:2023-07-27:  Add search index management clarifications.
\ No newline at end of file
+:2023-07-27:  Add search index management clarifications.
+:2023-11-08: Clarify that ``readConcern`` and ``writeConcern`` must not be
+             applied to search index managment commands.
```
GromNaN, line 1143: You may add an empty line at EOF to prevent diff on unrelated line.

While on it, can you remove the double spaces:

```suggestion
:2023-07-27: Add search index management clarifications.
:2023-11-08: Clarify that ``readConcern`` and ``writeConcern`` must not be
             applied to search index managment commands.

```

GromNaN (approved):

source/index-management/tests/README.rst
```diff
@@ -198,3 +198,26 @@ Case 5: ``dropSearchIndex`` suppresses namespace not found errors
 
 #. Create a driver-side collection object for a randomly generated collection name.  Do not create this collection on the server.
 #. Run a ``dropSearchIndex`` command and assert that no error is thrown.
+
+Case 6: Driver can successfully create and list search indexes with non-default readConcern and writeConcern
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+#. Create a collection with the "create" command using a randomly generated name (referred to as ``coll0``).
+#. Apply a write concern ``WriteConcern(w=1)`` and a read concern with ``ReadConcern(level="local")`` to ``coll0``.
```
vbabanin, line 206: During my testing, I noticed that the server rejects everything but 'local'` ReadConcern` for the `$listSearchIndexes `stage. Should we consider using any other `ReadConcern` to avoid potential false positives in the test?

kevinAlbs: Great catch. Updated the prose test to use `majority`. If `majority` is applied as the `readConcern`, the test fails with `$listSearchIndexes cannot run with a readConcern other than 'local'. Current readConcern: majority`.

PR #1475, opened 2023-11-15
DRIVERS-2584 Errors handling in Convenient Transactions API

comandeo:

- [x] Update changelog.
- [x] _Make sure there are generated JSON files from the YAML test files - not applicable._
- [x] _Test changes in at least one language driver - not applicable._
- [x] _Test these changes against all server versions and topologies (including standalone, replica set, sharded clusters, and serverless) - not applicable._

[Ruby Implementation](https://github.com/mongodb/mongo-ruby-driver/pull/2809)

### DIFF
--- a/source/transactions-convenient-api/transactions-convenient-api.rst
+++ b/source/transactions-convenient-api/transactions-convenient-api.rst
@@ -303,13 +303,13 @@ Errors handling inside the callback
 -----------------------------------
 
 Drivers MUST document that the callback MUST NOT silently swallow errors that
-do not have the "TransientTransactionError" label. Such errors abort the
-transaction on the server; trying to commit aborted transaction will result in
-an error.
+do not have the "TransientTransactionError" label. Such errors abort
+the transaction on the server; trying to commit an aborted transaction will
+result in an error.
 
-Drivers SHOULD recommend that the callback re-throw errors that do not have the
-"TransientTransactionError" label or otherwise using Core Transaction API if
-they want to handle errors in a custom way.
+Drivers SHOULD recommend that the callback re-throw errors that do not have
+the "TransientTransactionError" label. Drivers SHOULD also recommend using
+Core Transaction API if a user wants to handle errors in a custom way.
 
 Test Plan
 =========
@@ -506,6 +506,7 @@ client-side operation timeout, withTransaction can continue to use the
 Changes
 =======
 
+:2023-11-15: Document error handling inside the callback.
 :2022-10-05: Remove spec front matter and reformat changelog.
 :2022-01-19: withTransaction applies timeouts per the client-side operations
              timeout specification.
--- a/source/transactions-convenient-api/transactions-convenient-api.rst
+++ b/source/transactions-convenient-api/transactions-convenient-api.rst
@@ -299,6 +299,18 @@ MongoClient within the callback in order to execute operations within the
 transaction. Per the `Driver Session`_ specification, ClientSessions should
 already provide access to a client object.
 
+Errors handling inside the callback
+-----------------------------------
+
+Drivers MUST document that the callback MUST NOT silently swallow errors that
+do not have the "TransientTransactionError" label. Such errors abort the
+transaction on the server; trying to commit aborted transaction will result in
+an error.
+
+Drivers SHOULD recommend that the callback re-throw errors that do not have the
+"TransientTransactionError" label or otherwise using Core Transaction API if
+they want to handle errors in a custom way.
+
 Test Plan
 =========
 
--- a/source/transactions-convenient-api/transactions-convenient-api.rst
+++ b/source/transactions-convenient-api/transactions-convenient-api.rst
@@ -314,7 +314,6 @@ commit attempt with ``NoSuchTransaction`` error. This error has the
 "TransientTransactionError" label and the driver will retry the commit. This
 will result in an infinite loop.
 
-
 Drivers MUST recommend that the callback re-throw command errors if they
 need to be handled inside the callback. Drivers SHOULD also recommend using
 Core Transaction API if a user wants to handle errors in a custom way.
--- a/source/transactions-convenient-api/transactions-convenient-api.rst
+++ b/source/transactions-convenient-api/transactions-convenient-api.rst
@@ -302,14 +302,22 @@ already provide access to a client object.
 Handling errors inside the callback
 -----------------------------------
 
-Drivers MUST document that the callback MUST NOT silently swallow errors that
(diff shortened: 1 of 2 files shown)


source/transactions-convenient-api/transactions-convenient-api.rst
```diff
@@ -299,6 +299,18 @@ MongoClient within the callback in order to execute operations within the
 transaction. Per the `Driver Session`_ specification, ClientSessions should
 already provide access to a client object.
 
+Errors handling inside the callback
```
alcaeus, line 302: Either "Error handling" or "Handling errors":
```suggestion
Handling errors inside the callback
```

source/transactions-convenient-api/transactions-convenient-api.rst
```diff
@@ -299,6 +299,18 @@ MongoClient within the callback in order to execute operations within the
...
+-----------------------------------
+
+Drivers MUST document that the callback MUST NOT silently swallow errors that
```
alcaeus, line 305:
```suggestion
Drivers MUST document that the callback MUST NOT silently handle errors that
```

source/transactions-convenient-api/transactions-convenient-api.rst
```diff
@@ -299,6 +299,18 @@ MongoClient within the callback in order to execute operations within the
...
+do not have the "TransientTransactionError" label. Such errors abort
```
alcaeus, line 306: Is it guaranteed that the transaction will be aborted? Also, consider adding an example error. DuplicateKeyError would be the obvious choice since that's what prompted the clarification.

source/transactions-convenient-api/transactions-convenient-api.rst
comandeo: Changed the wording, thank you.

source/transactions-convenient-api/transactions-convenient-api.rst
comandeo: > Is it guaranteed that the transaction will be aborted?
Looks like this is not stated anywhere, so I changed the wording to "may abort the transaction". But I thinks we should play safe here.

> Also, consider adding an example error. 
Added an example with some additional clarification.

ShaneHarvey:
The underlying issue of suppressing command errors is also a pitfall in the core transaction API. Should we update that spec too?

comandeo:
> The underlying issue of suppressing command errors is also a pitfall in the core transaction API. Should we update that spec too?

Good point, thank you! Added some clarification to the core API.

source/transactions/transactions.rst
```diff
@@ -982,6 +982,15 @@ label. For example:
                 continue
             raise
 
+Handling command errors
+-----------------------
+
+Drivers MUST document that command errors that do not have the
```
kmahar, line 988: I think this should just be "command errors may abort the transaction on the server", regardless of the label. the server also aborts the transaction on many transient transaction errors (e.g. WriteConflict).

comandeo: Should we recommend re-raising all command errors inside `with_transaction` callback then? Since we do not have a reliable indicator which command aborts the transaction and with does not.

kmahar: I guess the distinction is perhaps relevant in that, if the user _doesn't_ re-raise an error but it's a TransientTransactionError, the driver still ends up doing the correct thing and retrying, since it gets NoSuchTransaction, a different TransientTransactionError.
whereas for a non-transient error we end up with the infinite loop issue. 

so I think it's less strictly necessary to recommend re-raising TransientTransactionErrors. however, the reason it happens to work in that case is a little funny. so for simplicity maybe it's best to not distinguish so users don't have to think about error labels and just say re-raise all errors?

comandeo: Good point, I agree. We should recommend re-raising all errors. If a user has some specific use case, core API will be a better choice.

source/transactions-convenient-api/transactions-convenient-api.rst
```diff
@@ -299,6 +299,26 @@ MongoClient within the callback in order to execute operations within the
...
+
+For example, ``DuplicateKeyError`` is an error that aborts a transaction on the
+server. If the callback catches ``DuplicateKeyError`` and does not re-throw it,
+the driver will attempt to commit the transaction. The server will reject the
+commit attempt with ``NoSuchTransaction`` error. This error has the
+"TransientTransactionError" label and the driver will retry the commit. This
+will result in an infinite loop.
+
+
+Drivers SHOULD recommend that the callback re-throw command errors if they
```
durran, line 318: Just out of curiosity, why are these a SHOULD and not a MUST? It seems to me like it would be required, 1) to lower the potential cases and 2) to have a quick point of reference for all drivers when users run into this.

comandeo: Sounds reasonable, changed to MUST.

source/transactions-convenient-api/transactions-convenient-api.rst
```diff
@@ -299,6 +299,26 @@ MongoClient within the callback in order to execute operations within the
...
+``NoSuchTransaction`` error.
+
+For example, ``DuplicateKeyError`` is an error that aborts a transaction on the
+server. If the callback catches ``DuplicateKeyError`` and does not re-throw it,
+the driver will attempt to commit the transaction. The server will reject the
+commit attempt with ``NoSuchTransaction`` error. This error has the
+"TransientTransactionError" label and the driver will retry the commit. This
+will result in an infinite loop.
+
+
```
jmikola, line 317:
```suggestion
```

source/transactions-convenient-api/transactions-convenient-api.rst
```diff
@@ -299,6 +299,25 @@ MongoClient within the callback in order to execute operations within the
...
+Drivers MUST document that the callback MUST NOT silently handle command errors
+without allowing such errors to propagate. Command errors may abort the transaction
+on the server, and an attempt to commit the transaction will be rejected with
+``NoSuchTransaction`` error.
+
+For example, ``DuplicateKeyError`` is an error that aborts a transaction on the
+server. If the callback catches ``DuplicateKeyError`` and does not re-throw it,
+the driver will attempt to commit the transaction. The server will reject the
+commit attempt with ``NoSuchTransaction`` error. This error has the
```
jmikola, line 313: @comandeo: @alcaeus will bring this up with you, but when talking about this earlier today we discussed whether "NoSuchTransaction" should trigger an early break from the `withTransaction` retry loop since it may indicate that the callback did something unexpected.

durran (approved):

PR #1477, opened 2023-11-22
DRIVERS-2770 Add connection string tests for uppercase UNIX socket names

qingyang-hu:
DRIVERS-2770 UNIX socket names should be case-sensitive.

- [x] Update changelog.


blink1073 (approved):
LGTM, thanks!  Tested with Python: https://github.com/mongodb/mongo-python-driver/pull/1453

PR #1481, opened 2023-12-06
DRIVERS-1150 Writes are not retried within transactions

comandeo:

- [x] Update changelog.
- [ ] _Make sure there are generated JSON files from the YAML test files - not applicable._
- [ ] _Test changes in at least one language driver - not applicable._ 
- [ ] _Test these changes against all server versions and topologies (including standalone, replica set, sharded clusters, and serverless) - not applicable._

### DIFF
--- a/source/retryable-writes/retryable-writes.rst
+++ b/source/retryable-writes/retryable-writes.rst
@@ -197,7 +197,7 @@ response.
 .. _findAndModify: https://www.mongodb.com/docs/manual/reference/command/findAndModify/
 .. _aggregate: https://www.mongodb.com/docs/manual/reference/command/aggregate/
 
-Retryable Writes Withing Transactions
+Retryable Writes Within Transactions
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 In MongoDB 4.0 the only supported retryable write commands within a transaction
--- a/source/retryable-writes/retryable-writes.rst
+++ b/source/retryable-writes/retryable-writes.rst
@@ -850,7 +850,7 @@ inconsistent with the server and potentially confusing to developers.
 Changelog
 =========
 
-:2023-12-06: Clarify that writes are not retried with transactions
+:2023-12-06: Clarify that writes are not retried with transactions.
 :2023-10-02: When CSOT is not enabled, one retry attempt occurs.
 :2023-08-26: Require that in a sharded cluster the server on which the
              operation failed MUST be provided to the server selection
--- a/source/retryable-writes/retryable-writes.rst
+++ b/source/retryable-writes/retryable-writes.rst
@@ -850,6 +850,7 @@ inconsistent with the server and potentially confusing to developers.
 Changelog
 =========
 
+:2023-12-06: Clarify that writes are not retried with transactions
 :2023-10-02: When CSOT is not enabled, one retry attempt occurs.
 :2023-08-26: Require that in a sharded cluster the server on which the
              operation failed MUST be provided to the server selection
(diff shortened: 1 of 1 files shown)


source/retryable-writes/retryable-writes.rst
```diff
@@ -197,6 +197,18 @@ response.
 .. _findAndModify: https://www.mongodb.com/docs/manual/reference/command/findAndModify/
 .. _aggregate: https://www.mongodb.com/docs/manual/reference/command/aggregate/
 
+Retryable Writes Withing Transactions
```
kkloberdanz, line 200:
```suggestion
Retryable Writes Within Transactions
```

kkloberdanz (changes requested):

kkloberdanz:
Hi Dmitry, is there any background information for this change that you would be able to share with us on this PR? Also, does it make sense to have a test for this behavior? Thank you.

source/retryable-writes/retryable-writes.rst
```diff
@@ -838,6 +850,7 @@ inconsistent with the server and potentially confusing to developers.
 Changelog
 =========
 
+:2023-12-06: Clarify that writes are not retried with transactions.
```
isabelatkinson, line 853:
```suggestion
:2023-12-06: Clarify that writes are not retried within transactions.
```

source/retryable-writes/retryable-writes.rst
```diff
@@ -197,6 +197,18 @@ response.
 .. _findAndModify: https://www.mongodb.com/docs/manual/reference/command/findAndModify/
 .. _aggregate: https://www.mongodb.com/docs/manual/reference/command/aggregate/
 
+Retryable Writes Withing Transactions
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In MongoDB 4.0 the only supported retryable write commands within a transaction
+are ``commitTransaction`` and ``abortTransaction``. Therefore drivers MUST NOT
+retry write commands within transactions even when ``retryWrites`` has been
+enabled on the ``MongoClient``. In addition, drivers MUST NOT add
```
isabelatkinson, line 206:
```suggestion
set to true on the ``MongoClient``. In addition, drivers MUST NOT add
```

source/retryable-writes/retryable-writes.rst
```diff
@@ -197,6 +197,18 @@ response.
...
+the ``RetryableWriteError`` label to any error that occurs during a write
+command within a transaction (excepting ``commitTransation``
+and ``abortTransaction``), even when ``retryWrites`` has been enabled on
```
isabelatkinson, line 209:
```suggestion
and ``abortTransaction``), even when ``retryWrites`` has been set to true on
```

source/retryable-writes/retryable-writes.rst
isabelatkinson: this sounds a little more clear to me but feel free to accept or reject, ditto below

isabelatkinson (approved):
LGTM with a few suggestions, I don't need to do another round of review

@kkloberdanz I see the following comment on DRIVERS-1150:
> This is a simple but important clarification for non-engineer audiences. It requires no driver changes or new tests.

kkloberdanz:
> > Hi Dmitry, is there any background information for this change that you would be able to share with us on this PR? Also, does it make sense to have a test for this behavior? Thank you.
> 
> Hi Kyle, as I understood [the ticket](https://jira.mongodb.org/browse/DRIVERS-1150), we just need to copy instructions from transactions spec to retryable writes spec. The behaviour is tested [here](https://github.com/mongodb/specifications/blob/master/source/transactions/tests/legacy/retryable-writes.yml#L160)

Sounds good to me, I'll approve.

kkloberdanz (approved):

PR #1486, opened 2024-01-02
DRIVERS-1641: pre-4.4 mongos writeConcernError does not determine retryability

jmikola:
https://jira.mongodb.org/browse/DRIVERS-1641

Also clarify that pre-4.4 writeErrors[].code should never be used to determine retryability. This is intentionally untested because failCommand does not provide a way to mock writeErrors.

POC: https://github.com/mongodb/mongo-c-driver/pull/1501

Patch build: https://spruce.mongodb.com/version/65a19aa461837db75d4d8efe/tasks (4.2-replica-auth and 4.2-sharded-auth are the relevant tasks).

----

- [x] Update changelog.

### DIFF
--- a/source/retryable-writes/retryable-writes.rst
+++ b/source/retryable-writes/retryable-writes.rst
@@ -861,7 +861,7 @@ inconsistent with the server and potentially confusing to developers.
 Changelog
 =========
 
-:2024-01-02: Do not use ``writeConcernError.code`` in pre-4.4 mongos response to
+:2024-01-16: Do not use ``writeConcernError.code`` in pre-4.4 mongos response to
              determine retryability. Do not use ``writeErrors[].code`` in
              pre-4.4 server responses to determine retryability.
 :2023-12-06: Clarify that writes are not retried within transactions.
--- a/source/retryable-writes/retryable-writes.rst
+++ b/source/retryable-writes/retryable-writes.rst
@@ -252,7 +252,8 @@ The RetryableWriteError label might be added to an error in a variety of ways:
   errors that meet the following criteria if the retryWrites option is set to
   true on the client performing the relevant operation:
 
-  - a server error response with any the following codes:
+  - a mongod or mongos response with any the following error codes in the
+    top-level ``code`` field:
 
     .. list-table::
       :header-rows: 1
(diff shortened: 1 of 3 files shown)


jmikola:
Reverting to draft until I can address [DRIVERS-2802](https://jira.mongodb.org/browse/DRIVERS-2802).

source/retryable-writes/tests/unified/insertOne-serverErrors.yml
```diff
@@ -75,3 +75,122 @@ tests:
...
+        topologies: [ replicaset, sharded ]
+    operations:
+      - name: failPoint
+        object: testRunner
+        arguments:
+          client: *client0
+          failPoint:
+            configureFailPoint: failCommand
+            # Trigger the fail point twice to allow asserting the error label in
+            # the retry attempt's response.
```
jmikola, line 92: This initially tripped me up while implementing the libmongoc POC and diagnosing some CI failures. When we assert that the RetryableWriteError label is added, those tests will implicitly have a retry attempt, so we need to fail the second attempt to ensure the label is present for our assertion.

This differs from the mongos test asserting that the label is _not_ added. There, the driver makes no retry attempt so we need only fail once and inspect the `insertOne` result for an error w/o the label.

source/retryable-writes/tests/unified/insertOne-serverErrors.yml
```diff
@@ -75,3 +75,123 @@ tests:
...
+                documents: [{ _id: 3, x: 33 }]
+              commandName: insert
+              databaseName: *databaseName
+          - commandStartedEvent: *insertCommandStartedEvent
+    outcome:
+      - collectionName: *collectionName
+        databaseName: *databaseName
+        documents:
+          - { _id: 1, x: 11 }
+          - { _id: 2, x: 22 }
```
jmikola, line 118: In contrast to the mocked writeConcernError tests below, failing `insert` with an `errorCode` does prevent the write from being applied.

kkloberdanz (approved):
Looks good!

source/retryable-writes/retryable-writes.rst
```diff
@@ -854,6 +861,9 @@ inconsistent with the server and potentially confusing to developers.
 Changelog
 =========
 
+:2024-01-02: Do not use ``writeConcernError.code`` in pre-4.4 mongos response to
```
jmikola, line 864:
```suggestion
:2024-01-16: Do not use ``writeConcernError.code`` in pre-4.4 mongos response to
```

PR #1491, opened 2024-01-09
DRIVERS-2790 Add codespell checker

blink1073:

- [ ] Update changelog.
  clusters, and serverless).


.pre-commit-config.yaml
```diff
@@ -55,8 +55,8 @@ repos:
...
-# - repo: https://github.com/codespell-project/codespell
-#   rev: "v2.2.6"
-#   hooks:
-#   - id: codespell
-#     exclude_types: [json,yaml]
+- repo: https://github.com/codespell-project/codespell
+  rev: "v2.2.6"
+  hooks:
+  - id: codespell
+    args: ["-L", "fle,re-use,merchantibility,synching,crate,nin,infinit,te"]
```
jmikola, line 62: "merchantability" is a misspelling of "merchantability", but I suppose most of this is required for `LICENSE.md`.

source/client-side-encryption/tests/legacy/fle2v2-MissingKey.json
```diff
@@ -54,7 +54,7 @@
   "key_vault_data": [],
   "tests": [
     {
-      "description": "FLE2 encrypt fails with mising key",
+      "description": "FLE2 encrypt fails with missing key",
```
jmikola, line 57: Please remember to make a note of any changed spec tests when entering downstream changes for DRIVERS-2780. It's possible that drivers depend on existing test names for custom skip logic, in which case some updates may be required (languages can always ignore syncing if they like).

jmikola (approved):

ShaneHarvey (approved):

source/client-side-encryption/tests/legacy/fle2v2-MissingKey.json
blink1073: The underlying ticket is closed, I don't see the word `mising` used in this org.  Do you think it is worth opening a new ticket?  Or perhaps ignoring `mising` in this PR and making a new ticket?

jmikola: If "mising" isn't used anywhere in the org I think we can just ignore this. Good idea running a search.

I also didn't consider the DRIVERS ticket would have already been triaged, and past the stage to denote driver changes.

PR #1495, opened 2024-01-11
DRIVERS-2711: Publish Topology Description Changed Event on Topology Close

W-A-James:
What's changing?

- load-balancer spec now explicitly requires that a `TopologyDescriptionChangedEvent` transitioning the topology from `LoadBalanced` to `Unknown` is emitted just before the `TopologyClosedEvent` is emitted.
- sdam monitoring spec new explicitly requires that a `TopologyDescriptionChangedEvent` transitioning the topology to `Unknown` is emitted just before the `TopologyClosedEvent` is emitted.
- load-balanced logging test asserts that `TopologyDescriptionChangedEvent` is emitted just before `TopologyClosedEvent`
- Add monitoring tests to check for new behaviour across all topology types

- [x] Update changelog.
  clusters, and serverless).


ShaneHarvey (dismissed):
Let's add SDAM monitoring tests for this change (that don't depend on logging).

source/server-discovery-and-monitoring/tests/unified/standalone-topology-lifecycle.json
```diff
@@ -0,0 +1,78 @@
+{
```
ShaneHarvey, line 1: Should this file be removed? I notice there's no yaml verison.

W-A-James: Good catch. Deleted.

source/server-discovery-and-monitoring/server-discovery-and-monitoring-logging-and-monitoring.rst
```diff
@@ -117,6 +117,12 @@ Initial Topology Description
 
 The first ``TopologyDescriptionChangedEvent`` to be emitted from a monitored Topology MUST set its ``previousDescription`` property to be a ``TopologyDescription`` object in the "unknown" state.
 
+Closing Topology Description
+----------------------------
+
+When a ``Topology`` object or equivalent is being shut-down or closed, the driver SHOULD change the
+``TopologyDescription`` to an "unknown" state.
```
alcaeus, line 124: Nit: this conflicts with the requirement in the load balancer spec that there must be an event that transitions the topology to `unknown`, whereas this leaves it optional. I suggest changing this to `MUST` as well.
The "replicaset-emit-topology-description-changed-before-close" test confirms this by explicitly checking for an `unknown` state on the topology description.

alcaeus (changes requested):

alcaeus (approved):

PR #1496, opened 2024-01-17
Clarify zero-case connectTimeoutMS in SDAM setupConnection PC

prestonvasquez:

- [x] Update changelog.
  clusters, and serverless).

We can't use a 0 connectTimeoutMS when creating a socket for monitoring, since that will result in a timeout error. IIUC we should not set a timeout, [described here](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-monitoring.rst#L555). This should be reflected in the psuedocode. This update is validated by this unified spec test: https://github.com/mongodb/specifications/blob/f1bbc245e68916ab76cb831f6372aa5d6a50ed25/source/server-discovery-and-monitoring/tests/unified/connectTimeoutMS.yml#L23


source/server-discovery-and-monitoring/server-monitoring.rst
```diff
@@ -756,7 +756,9 @@ The event API here is assumed to be like the standard `Python Event
             # Server API versioning implies that the server supports hello.
             helloOk = stableApi != Null
             connection = new Connection(serverAddress)
-            set connection timeout to connectTimeoutMS
+
+            if connectTimeoutMS != 0:
+              set connection timeout to connectTimeoutMS
```
ShaneHarvey, line 761: At first glance this check seems a little too specific for pseudocode. I imagine this case being handled by the "set connection timeout to connectTimeoutMS" pseudocode. What's the motivation for this change?

prestonvasquez: I was thrown off by the fact that we _are_ specific in the handling of connectTimeoutMS in the streaming case on [L782](https://github.com/mongodb/specifications/blob/f1bbc245e68916ab76cb831f6372aa5d6a50ed25/source/server-discovery-and-monitoring/server-monitoring.rst#L7822)

It took some time to determine what to do with the zero case when establishing a connection for server monitoring, as the Go Driver doesn't currently apply any timeout. I think [this socket timeout](https://github.com/mongodb/specifications/blob/f1bbc245e68916ab76cb831f6372aa5d6a50ed25/source/server-discovery-and-monitoring/server-monitoring.rst#L555) section of the spec covers my understanding of what to do:

> Clients MUST use connectTimeoutMS as the timeout for the connection handshake.
When connectTimeoutMS=0, the timeout is unlimited and MUST remain unlimited

---

Looking closer, I think we would want to apply the `if` block to this line as well: https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-monitoring.rst#L903

Or remove the block on L782, either way I think we should be consistent.

ShaneHarvey:
@prestonvasquez are you planning to update this PR?

ShaneHarvey (approved):

PR #1450, opened 2023-08-16
DRIVERS-1571 Retry on different mongos when possible

comandeo:
Please complete the following before merging:

- [x] Update changelog.
- [x] _Make sure there are generated JSON files from the YAML test files_ - not applicable, no spec tests
- [x] Test changes in at least one language driver.
- [x] Test these changes against all server versions and topologies (including standalone, replica set, sharded clusters, and serverless).

Ruby PoC - https://github.com/mongodb/mongo-ruby-driver/pull/2717
Go PoC     - https://github.com/mongodb/mongo-go-driver/pull/1358

PR #1504, opened 2024-02-06
DRIVERS-2828 Update mongos redirection prose tests to workaround SDAM behavior

eramongodb:

- [x] Update changelog.
- [x] Test changes in at least one language driver. (CDRIVER-4099)
  clusters, and serverless).

## Description

This PR proposes changes to the mongos redirection prose tests introduced in DRIVERS-2828 to workaround SDAM behavior preventing execution of server deprioritization code paths.

## SDAM Network Error Handling

SDAM error handling behavior labels a server as Unknown [when a post-handshake non-timeout network error occurs](https://github.com/mongodb/specifications/blob/3b8e7e4135f2615fedef1ea6cd2046f5be5a1725/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#L1443). This makes the server ineligible during server selection during the retry operation before the "on other mongos" prose tests have a chance to deprioritize the intended server (only one suitable server). This appears to be caused by the `closeConnection: true` field in the configured fail points. Therefore, this PR proposes their removal.

## Randomized Server Selection

Normal SDAM behavior randomly selects a server from the list of suitable servers. Consequently, the "on other mongos" prose tests cannot distinguish that a different mongos was used during the retry operation due to deprioritization (the behavior intended to be tested) from lucky randomized server selection. A note is added to the prose tests encouraging Drivers to manually inspect and ensure the deprioritization code paths are being executed by the prose tests as intended.

If Drivers have an alternative means of verifying that deprioritization code paths were executed, such as capturing debug/trace log output, they can do so, but this is not noted in the proposed PR due to falling outside the spec. This can be included in the note as an encouragement to Drivers if preferred, but I believe it unnecessary.

## Initial TopologyType

There is variation among Drivers w.r.t. how they handle the [directConnection](https://github.com/mongodb/specifications/blob/3b8e7e4135f2615fedef1ea6cd2046f5be5a1725/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#L400) URI option. Many Drivers appear to use `directConnection=true` by default, which means, given only a single host in the initial seed list, many Drivers will consider the topology of a single mongos connection to be Single rather than Sharded. Server deprioritization is only applicable for Sharded topologies. Therefore, the "on same mongos" prose test is likely not executing deprioritization code paths on many Drivers as currently written. Therefore, this PR adds the requirement that `directConnection=false` be included in the URI options when it not the default.

### DIFF
--- a/source/retryable-reads/tests/README.rst
+++ b/source/retryable-reads/tests/README.rst
@@ -238,71 +238,87 @@ Retrying Reads in a Sharded Cluster
 These tests will be used to ensure drivers properly retry reads on a different
 mongos.
 
-Retryable Reads Are Retried on a Different mongos if One is Available
----------------------------------------------------------------------
+Note: this test cannot reliably distinguish "retry on a different mongos due to
+server deprioritization" (the behavior intended to be tested) from "retry on a
+different mongos due to normal SDAM behavior of randomized suitable server
+selection". Verify relevant code paths are correctly executed by the tests using
+external means such as a logging, debugger, code coverage tool, etc.
+
+Retryable Reads Are Retried on a Different mongos When One is Available
+-----------------------------------------------------------------------
 
 This test MUST be executed against a sharded cluster that has at least two
-mongos instances.
+mongos instances, supports ``retryReads=true``, and has enabled the
+``configureFailPoint`` command (MongoDB 4.2+).
 
-1. Ensure that a test is run against a sharded cluster that has at least two
-   mongoses. If there are more than two mongoses in the cluster, pick two to
-   test against.
+1. Create two clients ``s0`` and ``s1`` that each connect to a single mongos
+   from the sharded cluster. They must not connect to the same mongos.
 
-2. Create a client per mongos using the direct connection, and configure the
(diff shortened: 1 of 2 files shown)


kevinAlbs (approved):
LGTM. The test improvements are much appreciated.

kkloberdanz (approved):
Looks good!

eramongodb:
Poke: @alexbevi @comandeo @prestonvasquez

Your reviews were requested since you reviewed the prior PR (https://github.com/mongodb/specifications/pull/1450). Given the expected impact of DRIVERS-1571, I think we should give some priority to this PR (DRIVERS-2828). I also think it is valuable to validate these changes with at least one more Driver (Ruby? Go?). If unwilling or unable to review, please remove yourself from the reviewers list to unblock this PR. Thank you!

alexbevi:
I'll defer to engineering to LGTM this, but from a Product POV it LGTM 😺

comandeo (approved):
👍

prestonvasquez (approved):
LGTM!

PR #1502, opened 2024-02-01
DRIVERS-1709: Convert legacy transactions spec tests to unified test format

jmikola:
https://jira.mongodb.org/browse/DRIVERS-1709
https://jira.mongodb.org/browse/DRIVERS-2834

Note: withTransaction tests will be ported in a separate PR: https://github.com/mongodb/specifications/pull/1505

Please complete the following before merging:

- [x] Update changelog.
- [x] Make sure there are generated JSON files from the YAML test files.
- [x] Test changes in at least one language driver.
- [x] Test these changes against all server versions and topologies (including standalone, replica set, sharded
  clusters, and serverless).

POC: https://github.com/mongodb/mongo-php-library/pull/1226

PR #1505, opened 2024-02-07
DRIVERS-1709: Convert withTransaction tests to unified format

jmikola:
https://jira.mongodb.org/browse/DRIVERS-1709

Note: this is being split off from https://github.com/mongodb/specifications/pull/1502

- [x] Update changelog.
  clusters, and serverless).

POC: https://github.com/mongodb/mongo-php-library/pull/1226


jmikola:
@jyemin: Just tagged you in case you'd like to add these to a Java patch build. Feel free to disregard and remove yourself.

Note: there was no use of `targetedFailPoint` in the `withTransaction` tests, so no further modifications were made to skip `load-balanced` topologies.

durran (approved):
LGTM. Node just had to skip 2 callback retry tests that I think is a bug in the driver.

jyemin:
Java driver is also failing the two callback-retry tests:

1. callback is not retried after non-transient error (DuplicateKeyError)
2. callback succeeds after multiple connection errors

So maybe not a bug in the Node driver, @durran.

There are four other tests failing as well, but those seem to be test runner issues.

Patch build: https://spruce.mongodb.com/version/65cb8c513e8e86ed75242d71/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

source/transactions-convenient-api/tests/unified/callback-retry.yml
```diff
@@ -0,0 +1,245 @@
...
+                writeConcern: { $$exists: false }
+              commandName: commitTransaction
+              databaseName: admin
+    outcome:
+      - collectionName: *collectionName
+        databaseName: *databaseName
+        documents:
+          - { _id: 1 }
+  -
+    description: callback is not retried after non-transient error (DuplicateKeyError)
```
jyemin, line 172: The legacy test has useMultipleMongoses: true for this one, and the Java driver's legacy test runner skip it when running against a replica set. I'm not sure whether that is correct behavior, but when running this test as a UTR it fails against a 7.0 replica set (it actually times out on Evergreen after about 40 minutes). Can we determine whether it's intended to be run against a replica set?

source/transactions-convenient-api/tests/unified/callback-retry.yml
```diff
@@ -0,0 +1,245 @@
...
+      client: *client1
+
+initialData:
+  - collectionName: *collectionName
+    databaseName: *databaseName
+    documents: []
+
+tests:
+  -
+    description: callback succeeds after multiple connection errors
```
jyemin, line 51: This test fails with the Java driver UTR against a 7.0 replica set:

```
[2024/02/13 11:23:17.424]     org.junit.ComparisonFailure: Command names must be equal
[2024/02/13 11:23:17.424]     Assertion Context:
[2024/02/13 11:23:17.424]     Event Matching Context
[2024/02/13 11:23:17.424]        event position: 1
[2024/02/13 11:23:17.424]        expected event: {"commandStartedEvent": {"command": {"abortTransaction": 1, "lsid": {"$$sessionLsid": "session0"}, "txnNumber": 1, "autocommit": false, "readConcern": {"$$exists": false}, "startTransaction": {"$$exists": false}, "writeConcern": {"$$exists": false}}, "commandName": "abortTransaction", "databaseName": "admin"}}
[2024/02/13 11:23:17.424]        actual event:   {"commandStartedEvent": {"command": {"commitTransaction": 1, "$db": "admin", "$clusterTime": {"clusterTime": {"$timestamp": {"t": 1707841397, "i": 7}}, "signature": {"hash": {"$binary": {"base64": "M02Xhbame6Ux6mCG3nsbWTV52i0=", "subType": "00"}}, "keyId": 7335118067787104262}}, "lsid": {"id": {"$binary": {"base64": "sFYW5mm/QLq+0OFJa9yqFA==", "subType": "04"}}}, "txnNumber": 1, "autocommit": false}, "databaseName": "admin"}}
```

The legacy test passes.  I haven't done an RCA, but lmk if you need it.

jyemin (commented):
Fixed locations of my comments.

source/transactions-convenient-api/tests/unified/callback-retry.yml
jyemin: It also times out against a sharded cluster, whereas the legacy test passes.

jmikola:
@jyemin: While diagnosing some PHPLIB build failures, I realized that several tests probably had to be skipped on serverless. The legacy `withTransaction` tests never included `load-balanced` so there always excluded on serverless.

source/transactions-convenient-api/tests/unified/callback-retry.yml
jyemin: If this test is passing for others than it's likely an issue with the Java driver UTR.

source/transactions-convenient-api/tests/unified/callback-retry.yml
```diff
@@ -0,0 +1,245 @@
...
+              # We do not assert the result here, as insertOne will fail for
+              # the first two executions of the callback before ultimately
+              # succeeding and returning a result. Asserting the state of the
+              # output collection after the test is sufficient.
+              name: insertOne
+              object: *collection0
+              arguments:
+                session: *session0
+                document: { _id: 1 }
+              ignoreResultAndError: true
```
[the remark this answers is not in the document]
jmikola: The `withTransaction` spec doesn't discuss callback return values ([The callback function has a flexible signature](https://github.com/mongodb/specifications/blob/master/source/transactions-convenient-api/transactions-convenient-api.rst#the-callback-function-has-a-flexible-signature) is probably the only commentary on the subject). I know some drivers do relay the callback's return value through `withTransaction()`; however, I wouldn't expect this to come into play with the unified test format.

In the UTF, the callback is expressed as a list of operations. I assume the Java implementation attempts to relay the return value of the last operation, but I'd argue there's no basis for doing so in the UTF spec. And there certainly isn't a spec test on the matter.

What _is_ important (and tested) is being able to relay exceptions from the callback. And that is precisely why I resorted to using `ignoreResultAndError` here, as it was the only way to preserve the original behavior in the legacy test format where the absence of a result _and_ error expectation allowed the test runner to forgo any assertion.

See: [`withTransaction` UTF operation in PHPLIB](https://github.com/mongodb/mongo-php-library/blob/1.17.0/tests/UnifiedSpecTests/Operation.php#L753) and the corresponding [`Operation::assert()`](https://github.com/mongodb/mongo-php-library/blob/1.17.0/tests/UnifiedSpecTests/Operation.php#L128) method for reference.

source/transactions-convenient-api/tests/unified/callback-retry.yml
jmikola: @jyemin: I think this may be related to the Java UTR's handling of `ignoreResultAndError: true` and deciding to not propagate exceptions from the callback's operations to `withTransaction`. `ignoreResultAndError` should _only_ apply to the UTR's assertion logic and should not get in the way of how `withTransaction()` operates (i.e. [sequence of actions](https://github.com/mongodb/specifications/blob/master/source/transactions-convenient-api/transactions-convenient-api.rst#sequence-of-actions) in the spec).

See my other response in https://github.com/mongodb/specifications/pull/1505#discussion_r1488840134.

source/transactions-convenient-api/tests/unified/callback-retry.yml
durran: I think what I'm running into here is https://jira.mongodb.org/browse/DRIVERS-2816

jmikola: @jyemin: In response to your [top-level comment](https://github.com/mongodb/specifications/pull/1505#issuecomment-1941807228) above:

> There are four other tests failing as well, but those seem to be test runner issues.
>
> Patch build: https://spruce.mongodb.com/version/65cb8c513e8e86ed75242d71/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

I see that two tasks timed out, but don't see any failures within them. Are you referring to something in a different patch build?

jyemin: The test that runs for 40 minutes on Evergreen before the Evergreen task times out fails when run locally after about a minute.  So I was able to see these four other test failures locally but not yet on Evergreen.

source/transactions-convenient-api/tests/unified/callback-retry.yml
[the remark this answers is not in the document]
jyemin: > The withTransaction spec doesn't discuss callback return values ([The callback function has a flexible signature](https://github.com/mongodb/specifications/blob/master/source/transactions-convenient-api/transactions-convenient-api.rst#the-callback-function-has-a-flexible-signature) is probably the only commentary on the subject). I know some drivers do relay the callback's return value through withTransaction(); however, I wouldn't expect this to come into play with the unified test format.

Agreed that the return value is irrelevant here.

> In the UTF, the callback is expressed as a list of operations. I assume the Java implementation attempts to relay the return value of the last operation, but I'd argue there's no basis for doing so in the UTF spec. And there certainly isn't a spec test on the matter.

It does not attempt to relay the return value.

> What is important (and tested) is being able to relay exceptions from the callback. And that is precisely why I resorted to using ignoreResultAndError here, as it was the only way to preserve the original behavior in the legacy test format where the absence of a result and error expectation allowed the test runner to forgo any assertion.

This is what's missing in the Java UTR.  It treats any exception thrown by the operation the same as it treats the return value.  It asserts on one or the other, depending on the test, and "swallows" both the return value and the exception. (It actually will throw an exception, but only a Junit AssertionException because that's how test failures are communicated in Java).

I think I can change the withTransaction operation to work a bit differently so that the exception is propagated, and get this test to pass.

I also realized that the test that is timing out (`callback is not retried after non-transient error (DuplicateKeyError)`) is related to this as well.  But unlike this test, the insertOne operation that is expected to fail in that test does not include `ignoreResultAndError: true`.  I'm not actually sure whether `ignoreResultAndError` should change the behavior of error propagation.  If in the Java driver UTR I just always propagate errors when inside a transaction callback, regardless of the value of `ignoreResultAndError`, both tests pass.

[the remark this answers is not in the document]
jmikola: > If in the Java driver UTR I just always propagate errors when inside a transaction callback, regardless of the value of ignoreResultAndError, both tests pass.

That sounds correct to me, and it's consistent with what I was already doing in PHPLIB. Regardless of `ignoreResultAndError`, I don't think there's any reason for the UTR's assertions to interfere with the propagation of exceptions from the callback to `withTransaction()`.

It's reasonable that this would come up now, as the original tests in [poc-transactions-convenient-api.yml](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/tests/valid-pass/poc-transactions-convenient-api.yml) never dealt with errors.

I made sure to note this in the downstream changes for DRIVERS-1709:

> Ensure that neither ignoreResultAndError nor expectError interfere with propagating exceptions from a callback operation to withTransaction.

jmikola:
> Unfortunately, there are now new test failures on sharded deployments from 3.6 through 4.2. The failing tests are not the new tests and not even all of them are unified tests. I am not sure what is causing those failures, but you can see them here: https://github.com/mongodb/mongo-java-driver/pull/1310.

@jyemin: Just followed up in: https://github.com/mongodb/mongo-java-driver/pull/1310#issuecomment-1946216488

For the failures outside of the transactions and `withTransaction` spec tests, it may be helpful to schedule more limited builds without these suites to see if there's something with these tests leaving the cluster in a state where other spec suites are failing down the line.

alcaeus (approved):

PR #1518, opened 2024-02-16
DRIVERS-2835 Increase interruptInUseConnections timeout to fix flaky test

ShaneHarvey:
DRIVERS-2835. Testing in python here: https://github.com/mongodb/mongo-python-driver/pull/1528


baileympearson (approved):

ShaneHarvey:
Note this test still runs just as fast since we don't actually wait for this failpoint to timeout.

PR #1522, opened 2024-02-21
DRIVERS-2789 Add Generated Markdown Index File

blink1073:
After this is merged I can open a ticket to add this repo to our internal docs.


scripts/generate_index.py
```diff
@@ -0,0 +1,29 @@
+import os
+from pathlib import Path
+source = Path(__file__).resolve().parent.parent / "source"
+source = source.resolve()
+info = {}
+for p in Path(source).rglob("*"):
+    if p.suffix not in ['.md']:
+        continue
+    relpath = os.path.relpath(p.parent, start=source)
```
Jibola, line 9: NIT: We could use `"*.md"` reduce the amount of paths to iterate. Thoughts?
```suggestion
for p in Path(source).rglob("*.md"):
    relpath = os.path.relpath(p.parent, start=source)
```

scripts/generate_index.py
```diff
@@ -0,0 +1,29 @@
...
+    if "tests" in relpath:
+        continue
+    if p.name in ['index.md']:
+        continue
+    fpath = relpath + '/' + p.name
+    name = None
+    with p.open() as fid:
+        for line in fid:
+            if line.startswith("# "):
+                name = line.replace('# ', '').strip()
```
Jibola, line 19: NIT: Is it guaranteed that all the `.md` files have headings that are hyphen delimited?

Jibola (approved):
LGTM

scripts/generate_index.py
blink1073: Done

scripts/generate_index.py
blink1073: Do you mean whether they start with `# `?  There is a guard for that.  For hyphens, I'm adding them explicitly below.

PR #1524, opened 2024-02-22
DRIVERS-2789 Restore RST spec files with pointers to MD files

blink1073:

- [ ] Update changelog.
  clusters, and serverless).


blink1073:
I've added a "Risks" section to the ticket with further information on the migration plan.

source/benchmarking/benchmarking.rst
```diff
@@ -0,0 +1,1163 @@
+
+.. note::
+  This specification has been converted to Markdown and renamed to
+  `benchmarking.md <benchmarking.md>`_.  
+
+  Use the link above to access the latest version of the specification as the
+  current reStructuredText file will no longer be updated.
```
jmikola, line 7: What is the value of restoring the original RST content? Since they're being introduced in a new commit, I don't think this would help with our ability to git-blame.

alexbevi: @jmikola this is called out in DRIVERS-2789 under _Risks_. The RST files were restored for the time being to ensure existing links in internal/external documentation and content don't result in 404s.

alexbevi (approved):
Header text LGTM, as does the plan laid out in DRIVERS-2789.

PR #266, opened 2018-01-26
Only send bypassDocumentValidation if it's false

ajdavis:
I don't think this change is worth adding CRUD Spect tests for.

PR #1534, opened 2024-03-01
DRIVERS-716 Improved Bulk Write API

isabelatkinson:
### Summary
Specifies the driver API for the new `bulkWrite` command. The API in the new specification was previously discussed and approved in [WRITING-13533](https://jira.mongodb.org/browse/WRITING-13533). I made a few changes to the syntax which are called out in inline comments.

An implementation in the Rust Driver of this spec is currently in review (https://github.com/mongodb/mongo-rust-driver/pull/1034). A C driver implementation is also in progress.

### Test Plan
I've added unified tests in the `crud` directory for bulk write options, errors, and results. There are also some prose tests for bulk write batching. This seemed preferable to trying to add functionality to the unified test runner to build very large documents based on `hello` response values.

Basic tests are also added for the following specs:
- Transactions: executing a bulk write within a transaction
- Retryable writes: executing bulk writes with/without multi: true operations
- Stable API: appending an API version to a bulk write command

- [x] Update changelog.
  clusters, and serverless).


source/crud/bulk-write.md
```diff
@@ -0,0 +1,724 @@
...
+defined in the
+[Client Side Encryption Specification](../client-side-encryption/client-side-encryption.rst).
+
+#### Unencrypted bulk writes with document sequences
+
+When `ops` and `nsInfo` are provided as document sequences, drivers MUST ensure that the total size
+of the `OP_MSG` built for the `bulkWrite` command does not exceed `maxMessageSizeBytes`. Some
+drivers may perform batch-splitting prior to constructing the full `OP_MSG` to be sent to the
+server. In this case, drivers MAY use `maxMessageSizeBytes - 16,384` as the upper bound for the
+combined number of bytes in `ops` and `nsInfo`. 16KiB is subtracted to accommodate for the bytes in
```
isabelatkinson, line 576: I'm not sure if other drivers will encounter a similar issue, but in Rust we build all of the operation-specific portions of a command prior to appending operation-agnostic fields and building an actual `OP_MSG`. This means that the total size of the message isn't known at the time of constructing `ops` and `nsInfo`. I lifted the 16 KiB number from the extra bytes the server allows in documents for overhead and it seems to be fine from my testing.

isabelatkinson:
(Removing spec owners as reviewers for now, going to gather initial feedback from Kevin & Jeremy)

source/crud/bulk-write.md
```diff
@@ -0,0 +1,724 @@
...
+the rest of the message.
```
kevinAlbs, line 577: Suggest adding a check for documents within the document sequence (not just inserts):

```md
Drivers MUST ensure each document within a document sequence does not exceed `maxBsonObjectSize` + 16384 if the write is unacknowledged. 16KiB is added to account for command overhead allowed in the server (refer: [SERVER-10643](https://jira.mongodb.org/browse/SERVER-10643)).
```

Notably: C and Python drivers have different existing size checks for OP_MSG document sequences:

[The C driver](https://github.com/mongodb/mongo-c-driver/blob/f6547a375c8d0b9ae7eff898487835ad451ab3d0/src/libmongoc/src/mongoc/mongoc-write-command.c#L643-L647) does not check if the write is unacknowledged, and includes the 16k allowance:

```c
if (slen > max_bson_obj_size + BSON_OBJECT_ALLOWANCE) {
```

[The Python driver](https://github.com/mongodb/mongo-python-driver/blob/b041ca5f7c201f1b0ecb97c517a1ece8d775c00a/pymongo/message.py#L1340) checks if the write is unacknowledged, and does not include the 16k allowance:
```python
unacked_doc_too_large = not ack and (doc_length > max_bson_size)
```

source/crud/bulk-write.md
```diff
@@ -0,0 +1,724 @@
...
+    "errorsOnly": <Boolean>,
+    "ordered": <Boolean>,
+    "bypassDocumentValidation": <Boolean>,
+    "comment": <Bson>,
+    "let": <Document>,
+    ...additional operation-agnostic fields
+}
+```
+
+Drivers SHOULD use document sequences ([`OP_MSG`](../message/OP_MSG.rst) payload type 1) for the
```
kevinAlbs, line 467:
```suggestion
Drivers MUST use document sequences ([`OP_MSG`](../message/OP_MSG.rst) payload type 1) for the
```

Suggest preferring "MUST" over "SHOULD" to reduce possible variations in driver implementations. I expect it is easier to relax the spec later if needed.

Preferring "MUST" is also recommended in https://github.com/mongodb/specifications?tab=readme-ov-file#when-in-doubt-use-must-not-should-in-specs

source/crud/bulk-write.md
```diff
@@ -0,0 +1,724 @@
...
+{
+    "insert": <Int32>,
+    "document": <Document>
+}
+```
+
+Drivers MUST add an `_id` field at the beginning of the insert document if one is not already
+present.
+
+When a user executes a bulk write with an unacknowledged write concern, drivers SHOULD check the
```
kevinAlbs, line 491:
```suggestion
When a user executes a bulk write with an unacknowledged write concern, drivers MUST check the
```

kevinAlbs (commented):
Reviewed `bulk-write.md`. Looks great so far. Posting feedback now, since I expect the C implementation will take a while. Enough was implemented to review the API.

source/crud/bulk-write.md
```diff
@@ -0,0 +1,724 @@
...
+The `ops` field is a list of write operation documents. The first entry in each document has
+the name of the operation (i.e. "insert", "update", or "delete") as its key and the index in the
+`nsInfo` array of the namespace on which the operation should be performed as its value. The
+documents have the following format:
+
+#### Insert
+
+```json
+{
+    "insert": <Int32>,
```
jmikola, line 483: Is the server flexible about accepting an Int32 or Int64 here, or is Int32 absolutely required? I do realize how absurd an Int64 would be in this context.

source/crud/bulk-write.md
```diff
@@ -0,0 +1,724 @@
...
+    "document": <Document>
+}
+```
+
+Drivers MUST add an `_id` field at the beginning of the insert document if one is not already
+present.
```
jmikola, line 489: To leave nothing to the imagination, clarify that drivers should generate an [ObjectId](https://github.com/mongodb/specifications/blob/master/source/objectid.rst). FWIW, the CRUD spec isn't entirely clear on that.

source/crud/bulk-write.md
```diff
@@ -0,0 +1,724 @@
...
+`maxBsonObjectSize`. The value for `maxBsonObjectSize` can be retrieved from the selected server's
+`hello` response.
+
+#### Update
+
+```json
+{
+    "update": <Int32>,
+    "filter": <Document>,
+    "updateMods": <Document | Array>,
```
jmikola, line 503: This is a bizarre name choice since this can also be an update document, but I realize `update` would have conflicted with the first field. Either way, not your problem -- I just felt compelled to rant.

source/crud/bulk-write.md
```diff
@@ -0,0 +1,724 @@
...
+    "multi": <boolean>,
+    "hint": Optional<String>,
+    "collation": Optional<Document>
+}
+```
+
+### Namespace Information
+
+The `nsInfo` field is an array containing the namespaces on which the write operations should be
+performed. Drivers MUST NOT include duplicate namespaces in this list. The documents in the
```
jmikola, line 526: Is the prohibition of duplicate namespaces enforced, or are we on the honor system here?

source/crud/bulk-write.md
```diff
@@ -0,0 +1,724 @@
...
+     * client with a different write concern configured than that of the user's standard URI
+     * options is nontrivial. Drivers MAY omit this option if they provide a way to retrieve a
+     * lightweight handle to a client with a custom write concern configured, e.g. a
+     * MongoClient.withWriteConcern() method.
+     */
+    writeConcern: Optional<WriteConcern>;
+
+    /**
+     * Whether detailed results for each successful operation should be included in the returned
+     * BulkWriteResult.
```
jmikola, line 284: It may be helpful to note that this corresponds inversely to the value of `errorsOnly` in the `bulkWrite` command. Not necessary for other options since those have a direct counterpart.

source/crud/bulk-write.md
```diff
@@ -0,0 +1,724 @@
...
+call to `MongoClient.bulkWrite` may require multiple `bulkWrite` commands to be sent to the server.
+Drivers MUST split bulk writes into separate commands when the user's list of operations exceeds
+one or more of these maximums: `maxWriteBatchSize`, `maxBsonObjectSize`, and `maxMessageSizeBytes`.
+Each of these values can be retrieved from the selected server's `hello` command response. Drivers
+MUST merge results from multiple batches into a single `BulkWriteResult` or `BulkWriteException`
+to return from `MongoClient.bulkWrite`.
+
+### Number of Writes
+
+The `ops` array MUST NOT include more than `maxWriteBatchSize` operations.
```
jmikola, line 559: Given the third prose test, I assume clients do not enforce this and instead rely on the server to do so. Should that be clarified here? I don't recall seeing it mentioned earlier.

source/crud/bulk-write.md
jmikola: Sounds reasonable. Should you clarify that 16KiB is an approximation?

source/crud/bulk-write.md
```diff
@@ -0,0 +1,724 @@
...
+```json
+{
+    "ok": <0 | 1>,
+    "code": Optional<Int32>,
+    "errmsg": Optional<String>,
+    "n": <Int32>,
+    "nMatched": <Int32>,
+    "nModified": <Int32>,
+    "upsertedId": Optional<ObjectId>
+}
```
jmikola, line 630: Are there additional fields in the response not shown here? For instance, the CRUD spec refers to an `errInfo` field for write and write concern errors, which might be used to relay document validation errors.

source/crud/bulk-write.md
```diff
@@ -0,0 +1,724 @@
...
+`getMore` call.
+
+The documents in the results cursor have the following format:
+
+```json
+{
+    "ok": <0 | 1>,
+    "code": Optional<Int32>,
+    "errmsg": Optional<String>,
+    "n": <Int32>,
```
jmikola, line 626: Can we clarify exactly what these fields refer to? That was an omission of the CRUD spec and one reason why the deprecated bulk write spec remained relevant (as it summarized server responses).

Would `n` here only be relevant for insert and delete operations? I forget if there's relevance for update operations.

source/crud/bulk-write.md
```diff
@@ -0,0 +1,724 @@
...
+    "nMatched": <Int32>,
+    "nModified": <Int32>,
+    "upsertedId": Optional<ObjectId>
+}
+```
+
+Note that the responses do not contain information about the type of operation that was performed.
+Drivers may need to maintain the user's list of write models to infer which type of result should
+be recorded.
```
jmikola, line 635: How do drivers correlate cursor documents with original operations? Particularly, in an errorsOnly scenario, I assume the cursor would only report errorred operations. I don't see any field in the response that indicates the index of a write operation.

source/crud/bulk-write.md
```diff
@@ -0,0 +1,724 @@
...
+```json
+{
+    "bulkWrite": 1,
+    "ops": <Array>,
+    "nsInfo": <Array>,
+    "errorsOnly": <Boolean>,
+    "ordered": <Boolean>,
+    "bypassDocumentValidation": <Boolean>,
+    "comment": <Bson>,
+    "let": <Document>,
```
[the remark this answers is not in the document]
jmikola: Elsewhere, the spec says that `errorsOnly` "MUST always be defined in the `bulkWrite` command". Is it actually optional on the server? If so, what does it default to? It'd be helpful to clarify that and other server-side defaults just so we have some accessible reference on that.

Speaking from experience with the other specs, this comes up from time to time and we can't always rely on the server manual to provide adequate command documentation.

source/crud/bulk-write.md
```diff
@@ -0,0 +1,724 @@
...
+     *
+     * NOT REQUIRED TO IMPLEMENT. Drivers may choose not to provide this property so long as it is
+     * always possible to discern whether an upsert took place.
+     */
+    upsertedCount: Int64;
+
+    /**
+     * The _id field of the upserted document if an upsert occurred.
+     */
+    upsertedId: Optional<Any>;
```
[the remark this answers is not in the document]
kevinAlbs: > what is the API impact? Are you referring to the spec document itself?

I prefer leaving `upsertedCount` out of the spec unless needed. I expect `upsertedCount` can only be 0 or 1. `upsertedCount` is motivated by supporting upserting an `_id` with BSON NULL (seems rare IMO).

I interpret "NOT REQUIRED TO IMPLEMENT" to mean drivers are free to implement regardless of whether it is needed.

> This would require an envelope object in PHP, which would be clunkier than using a native `null` value and providing an `upsertedCount` field

Given the need, here is a revised suggestion:

```typescript
/**
 * The _id field of the upserted document if an upsert occurred.
 * It MUST be possible for a caller to discern whether an upsert took place, even if the `upsertedId` is a BSON NULL value.
 * If necessary, drivers MAY add a boolean `didUpsert` field.
*/
upsertedId: Optional<Any>;
```

source/crud/bulk-write.md
```diff
@@ -0,0 +1,724 @@
...
+    "ordered": <Boolean>,
+    "bypassDocumentValidation": <Boolean>,
+    "comment": <Bson>,
+    "let": <Document>,
+    ...additional operation-agnostic fields
+}
+```
+
+Drivers SHOULD use document sequences ([`OP_MSG`](../message/OP_MSG.rst) payload type 1) for the
+`ops` and `nsInfo` fields. Drivers MUST NOT use document sequences when auto-encryption is enabled.
```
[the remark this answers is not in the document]
kevinAlbs: > Are there any plans for to change that down the line?
Filed DRIVERS-2859 to request support of document sequences. I expect there would be a performance benefit.

source/crud/bulk-write.md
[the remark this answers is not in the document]
jmikola: Fine to remove the explicit `upsertedCount` in exchange for some extra text under `upsertedId`.

I'd personally keep `upsertedCount` in PHP for consistency with existing APIs, but I'm not going to lobby for the suggested field name in this extra text.

source/crud/bulk-write.md
```diff
@@ -0,0 +1,724 @@
...
+provided as `true`. This field MUST always be defined in the `bulkWrite` command regardless of
+whether the user specified a value for `verboseResults`.
+
+## Command Batching
+
+Drivers MUST accept an arbitrary number of operations as input to the `MongoClient.bulkWrite` method.
+Because the server imposes restrictions on the size of write operations, this means that a single
+call to `MongoClient.bulkWrite` may require multiple `bulkWrite` commands to be sent to the server.
+Drivers MUST split bulk writes into separate commands when the user's list of operations exceeds
+one or more of these maximums: `maxWriteBatchSize`, `maxBsonObjectSize`, and `maxMessageSizeBytes`.
```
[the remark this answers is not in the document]
kevinAlbs: Retracting suggestion. `maxBsonObjectSize` and `maxMessageSizeBytes` apply to both payload types.

source/crud/bulk-write.md
```diff
@@ -0,0 +1,724 @@
...
+drivers may perform batch-splitting prior to constructing the full `OP_MSG` to be sent to the
+server. In this case, drivers MAY use `maxMessageSizeBytes - 16,384` as the upper bound for the
+combined number of bytes in `ops` and `nsInfo`. 16KiB is subtracted to accommodate for the bytes in
+the rest of the message.
+
+#### Unencrypted bulk writes without document sequences
+
+When `ops` and `nsInfo` are provided within the `bulkWrite` command document and auto-encryption is
+not enabled, drivers MUST ensure that the total size of the `bulkWrite` command document does not
+exceed `maxBsonObjectSize`.
```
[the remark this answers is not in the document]
kevinAlbs: Retracting suggestion. This section describes batch splitting. I expect the command overhead should not be included when batch splitting. Testing suggests sending a payload 0 sized to maxBsonObjectSize + 16384 results in an [error on mongos](https://github.com/kevinAlbs/c-bootstrap/blob/5cdd9262ccda033fcd3c0c3f5a737b73a4694529/tools/miniclient/test_opmsg_limits.py#L120-L128). I expect this is due to mongos internally adding fields to the sent document.

source/crud/bulk-write.md
```diff
@@ -0,0 +1,724 @@
...
+```
+
+Drivers MUST add an `_id` field at the beginning of the insert document if one is not already
+present.
+
+When a user executes a bulk write with an unacknowledged write concern, drivers SHOULD check the
+size of the insert document to verify that it does not exceed `maxBsonObjectSize`. For acknowledged
+bulk writes, drivers MAY rely on the server to return an error if the document exceeds
+`maxBsonObjectSize`. The value for `maxBsonObjectSize` can be retrieved from the selected server's
+`hello` response.
```
kevinAlbs, line 495: Suggest adding a check for all documents in `ops` (not just inserts). This may inform a user when there is no chance of a write succeeding. Consider adding a separate section under `### Operations`:

```md
#### Unacknowledged size checks

When a user executes a bulk write with an unacknowledged write concern, drivers MUST check the size of documents in `ops` and error if there is no expected chance of success:

Check that the size of each insert document does not exceed `maxBsonObjectSize`. The value for `maxBsonObjectSize` can be retrieved from the selected server's `hello` response.

Check that the size of other documents does not exceed `maxBsonObjectSize` + 16384 (16 KiB). 16KiB is added to account for command overhead allowed in the server (refer: [SERVER-10643](https://jira.mongodb.org/browse/SERVER-10643)).

For acknowledged bulk writes, drivers MUST NOT do these checks and instead rely on the server for size checks.
```

Notably: C and Python drivers have different existing size checks for OP_MSG document sequences:

[The C driver](https://github.com/mongodb/mongo-c-driver/blob/f6547a375c8d0b9ae7eff898487835ad451ab3d0/src/libmongoc/src/mongoc/mongoc-write-command.c#L643-L647) does not check if the write is unacknowledged, and includes the 16k allowance:

```c
if (slen > max_bson_obj_size + BSON_OBJECT_ALLOWANCE) {
```

[The Python driver](https://github.com/mongodb/mongo-python-driver/blob/b041ca5f7c201f1b0ecb97c517a1ece8d775c00a/pymongo/message.py#L1340) checks if the write is unacknowledged, and does not include the 16k allowance:
```python
unacked_doc_too_large = not ack and (doc_length > max_bson_size)
```

source/crud/bulk-write.md
kevinAlbs: Retracting suggestion. Suggestion does not seem relevant to this section. This section describes batch splitting. Added suggestion to: https://github.com/mongodb/specifications/pull/1534/files#r1518880882

source/crud/bulk-write.md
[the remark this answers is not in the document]
isabelatkinson: added a "Future Work" section with a link to that ticket

source/crud/bulk-write.md
```diff
@@ -15,12 +15,15 @@ defined in the CRUD specification and the
 
 ## Specification
 
-### `MongoClient.bulkWrite` Interface
+> \[!NOTE\]
```
jmikola, line 18: The escaping shouldn't be necessary here. See #1546

source/crud/bulk-write.md
```diff
@@ -0,0 +1,724 @@
...
+#### Individual Results
+
+The `InsertOneResult`, `UpdateResult`, and `DeleteResult` classes are the same as or similar to
+types of the same name defined in the [CRUD specification](crud.md). Drivers MUST redefine these
+classes if their existing result classes deviate from the definitions in this specification (e.g.
+if they contain acknowledgement information, which is not applicable for individual bulk write
+operations). Drivers MAY reuse their existing types for these classes if they match the ones
+defined here exactly.
+
+### Exception
```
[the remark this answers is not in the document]
jmikola: Oh, I missed that it referred to BulkWriteException already. Fine as-is.

source/crud/bulk-write.md
```diff
@@ -0,0 +1,724 @@
...
+`ops` and `nsInfo` fields. Drivers MUST NOT use document sequences when auto-encryption is enabled.
+
+The `bulkWrite` command is executed on the "admin" database.
+
+### Operations
+
+The `ops` field is a list of write operation documents. The first entry in each document has
+the name of the operation (i.e. "insert", "update", or "delete") as its key and the index in the
+`nsInfo` array of the namespace on which the operation should be performed as its value. The
+documents have the following format:
```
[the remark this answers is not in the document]
jmikola: Disregard my last comment. I wrote it before realizing it was an optimization and that an `ops` element referenced an index within `nsInfo`. I must have been thinking of a previous design from many months ago.

source/unified-test-format/unified-test-format.md
```diff
@@ -1416,6 +1428,53 @@ the resulting change stream might be saved with [operation.saveResultAsEntity](#
...
+        update: { $inc: { x: 1 } }
+    - deleteOne:
+        filter: { x: 2 }
+    - deleteMany:
+        filter: { x: { $gt: 2 } }
+  ordered: true
+```
+
+Because the `insertResults`, `updateResults`, and `deleteResults` may be absent or empty in the `BulkWriteResult` returned
+from a summary-only bulk write, the `clientBulkWrite` operation MUST use the [$$unsetOrMatches](#unsetormatches) operator
```
jmikola, line 1465: Note: this spec typically uses SHOULD instead of MUST, particular for things that cannot be enforced via the JSON schema. That's explained in the Design Rationale section.

This would be a candidate for SHOULD.

source/unified-test-format/unified-test-format.md
```diff
@@ -923,6 +923,18 @@ The structure of this object is as follows:
...
+  contains numeric keys representing the index of the write that failed and `writeError` object values. The test runner MUST
+  assert that the error contains a `writeError` for each index present in `writeErrors` and MUST assert that the `writeError`s
+  match as root-level documents according to the rules in [Evaluating Matches](#evaluating-matches). The test runner MUST
+  assert that the error does not contain any additional `writeError`s.
+
+- `writeConcernErrors`: Optional array of one or more objects. An ordered list of write concern errors expected to be
+  present in the error. The test runner MUST assert that each `writeConcernError` in this list matches the
+  `writeConcernError` present at the same index in the error's list of `writeConcernError`s as a root-level document according
+  to the rules in [Evaluating Matches](#evaluating-matches). The test runner MUST assert that the error does not contain any
+  additional `writeConcernErrors`s.
```
jmikola, line 936: Do you suppose these fields are suitable for use with the collection-level bulkWrite method defined in the CRUD spec? There, BulkWriteException still has a `writeErrors` field. Although `writeConcernError` is singular, the unified spec could always suggest wrapping it in an array so that `expectedError` assertions can be performed consistently.

I realize this is outside the scope of this PR, but I'm just thinking more generally about whether these new fields are limited to the client API or could get more general use.

source/crud/bulk-write.md
isabelatkinson: Update here: `errInfo` is the only other field expected to be in this response. Added it in with the other error fields.

source/unified-test-format/unified-test-format.md
isabelatkinson: makes sense, updated

source/unified-test-format/unified-test-format.md
isabelatkinson: I don't see why not. Your proposal for write concern errors seems reasonable. The collection-level `BulkWriteException` has a different format for its `writeErrors` field (a list of individual `BulkWriteError`s that store the index of the operation as a field) that doesn't fully align with the index-to-error mapping defined here, but drivers should still be able to make the right assertions.

source/crud/bulk-write.md
```diff
@@ -0,0 +1,724 @@
...
+     * Defaults to false.
+     */
+    ordered: Optional<Boolean>;
+
+    /**
+     * If true, allows the writes to opt out of document-level validation.
+     *
+     * Defaults to false.
+     *
+     * This option is only sent if the caller explicitly provides a value.
```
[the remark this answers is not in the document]
jmikola: FWIW, there doesn't appear to be any record of why the CRUD spec was modified to not sent `false` for `bypassDocumentValidation`. See: https://github.com/mongodb/specifications/pull/266#issuecomment-431491212

My last follow-up comment in [DRIVERS-450](https://jira.mongodb.org/browse/DRIVERS-450) (ticket created _after_ the spec change was merged) speculated that `bypassDocumentValidation: false` was unique in expressing a double-negative.

I could go either way here, but if we decide to keep the language as-is here, I'd suggest we also open a DRIVERS ticket to revise the CRUD spec and remove the ignorance of `false`. That can be a low priority change, as it's unlikely to affect an application and there's presently no test coverage anyway (see: [DRIVERS-1316](https://jira.mongodb.org/browse/DRIVERS-1316)).

source/crud/bulk-write.md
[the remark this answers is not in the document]
jmikola: I'm on board with having drivers explicitly send default values for `ordered` and `errorsOnly`. Note that the CRUD spec has no language about only sending `ordered` when specified by the client, so I expect it is always passed over the wire.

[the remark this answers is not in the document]
ShaneHarvey: PyMongo always sends "ordered" on all insert/update/delete commands.

source/unified-test-format/unified-test-format.md
jmikola: > The collection-level BulkWriteException has a different format for its writeErrors field

Ah, thanks for pointing that out. In that case, we'd need to sort out _both_ the `writeConcernError` (single to array) and `writeErrors` (object with `index` fields to a map) in order to share this structure.

In the meantime, can you clarify that both of these assertion fields are only intended to be used for the client API's BulkWriteException?

source/crud/bulk-write.md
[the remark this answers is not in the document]
kevinAlbs: > I think we should leave this as-is.

Agreed. I am convinced to send what is set, and rely on the server to interpret the value. Suggest documenting in the Q&A section since it is a notable change with existing CRUD spec:

```md
### Why is `bypassDocumentValidation: false` sent?

`BulkWriteOptions.bypassDocumentValidation` is only sent if set (to either true or false).
`InsertOneOptions.bypassDocumentValidation` is only sent if set to true (not sent if false).

The rationale for only sending `InsertOneOptions.bypassDocumentValidation` when
true is not clear. DRIVERS-450 notes possible rationale.

Checking the values that users explicitly provide for options and omitting them
if they match the server default seems like unnecessary additional work for
drivers. It could also lead to issues in the unlikely event that the server
changes the default value for an option.
```

> speculated that `bypassDocumentValidation: false` was unique in expressing a double-negative.

I can see the double-negative being confusing. Though in the event `bypassDocumentValidation=false` is set and the user's intent is to opt-out of document validation, the result is doing an extra document validation (not harmful IMO).

source/crud/bulk-write.md
kevinAlbs: Addressed by new `### Size Limits` section.

source/crud/bulk-write.md
```diff
@@ -0,0 +1,778 @@
...
+any checks related to these size limits and MUST rely on the server to raise an error if a limit is
+exceeded. However, when an unacknowledged write concern is used, drivers MUST raise an error if one
+of the following limits is exceeded:
+
+- The size of a document to be inserted MUST NOT exceed `maxBsonObjectSize`. This applies to the
+  `document` field of an `InsertOneModel` and the `replacement` field of a `ReplaceOneModel`.
+- The size of an entry in the `ops` array MUST NOT exceed `maxBsonObjectSize + 16KiB`.
+- The size of the `bulkWrite` command document MUST NOT exceed `maxBsonObjectSize`. The overhead
+  bytes are omitted here intentionally to accommodate additional fields that mongos appends to the
+  command document.
```
kevinAlbs, line 571:
```suggestion
- The size of the `bulkWrite` command document MUST NOT exceed `maxBsonObjectSize + 16KiB`.
```

I prefer the driver only error when the the operation is certain not to succeed. Otherwise rely on the server to error.

When using unacknowledged writes, I expect there will be many cases where a write is likely to error, but a driver cannot reasonably check (e.g. server document validation). When in doubt, I rather err towards having the driver send what the user sets.

source/crud/bulk-write.md
```diff
@@ -0,0 +1,778 @@
...
+
+The `bulkWrite` command returns top-level summary result counts and, optionally, individual results
+for each operation. Compiling the individual results server-side and consuming these results
+driver-side is less performant than only recording the summary counts. We expect that most users
+are not interested in the individual results of their operations and that most users will rely on
+defaults, so `verboseResults` defaults to `false` to improve performance in the common case.
+
+### Why are document sequences not permitted for encrypted bulk writes?
+
+
```
kevinAlbs, line 774:
```suggestion
```

Remove stray Q&A.

source/crud/bulk-write.md
```diff
@@ -0,0 +1,778 @@
...
+### Size Limits
+
+The server reports a `maxBsonObjectSize` in its `hello` response. This value defines the maximum
+size for documents that are inserted into the database. Documents that are sent to the server but
+are not intended to be inserted into the database (e.g. command documents) have a size limit of
+`maxBsonObjectSize + 16KiB`. When an acknowledged write concern is used, drivers MUST NOT perform
+any checks related to these size limits and MUST rely on the server to raise an error if a limit is
+exceeded. However, when an unacknowledged write concern is used, drivers MUST raise an error if one
+of the following limits is exceeded:
```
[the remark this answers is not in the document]
isabelatkinson: Arguments I have for not checking these sizes for acknowledged writes:

- Relying on the server to return an error for oversized documents aligns with [this](https://github.com/mongodb/specifications/?tab=readme-ov-file#where-possible-depend-on-server-to-return-errors) driver specification mantra.
- Not checking sizes driver-side means the server can still proceed with other valid writes in the bulk write.
- If we have a hard cap of `maxMessageSizeBytes` for the total message size, acknowledged or unacknowledged, users will still be limited to what the server deems an acceptable number of bytes to be sent in one message.

Would additional language here to clarify that drivers should never send a message larger than `maxMessageSizeBytes` address your concerns about potential malicious input to this method? I ask because it seems like users do value their other writes being performed when an oversized document is specified, e.g. [this recent forum post](https://www.mongodb.com/community/forums/t/pymongo-insert-many-raising-operationfailure-instead-of-bulkwriteerror-on-oversized-documents/270467).

source/crud/bulk-write.md
isabelatkinson: that's reasonable, updated

source/unified-test-format/unified-test-format.md
isabelatkinson: yep, updated

source/crud/bulk-write.md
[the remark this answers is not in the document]
kevinAlbs: > Not checking sizes driver-side means the server can still proceed with other valid writes in the bulk write.

Right, I forgot that point.

> Would additional language here to clarify

No need IMO. I think the batch splitting already addresses that messages will not exceed `maxMessageSizeBytes`.

source/unified-test-format/unified-test-format.md
```diff
@@ -1416,6 +1429,53 @@ the resulting change stream might be saved with [operation.saveResultAsEntity](#
 Test runners MUST NOT iterate the change stream when executing this operation and test files SHOULD NOT specify
 [operation.expectResult](#operation_expectResult) for this operation.
 
+#### clientBulkWrite
```
jmikola, line 1432: This should just be `bulkWrite`, as we're already under a "Client Operations" heading. If you'd like another link to it you can add something like the following:

```
<span id="client_bulkWrite" />
```

That's similar to what was done for `createChangeStream` above (see: https://github.com/mongodb/specifications/commit/68da5c5c9af9d87677ed90d30f8ad9ca3ea7e798).

Existing links to `#bulkwrite` can be changed to `#collection_bulkWrite` if you're willing to add a `<span id="collection_bulkWrite" />` above its heading as well.

source/crud/bulk-write.md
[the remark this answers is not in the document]
isabelatkinson: Added a Q&A entry and filed DRIVERS-2865

source/unified-test-format/unified-test-format.md
isabelatkinson: I intentionally named the UTR operation for the new bulk write "clientBulkWrite" to avoid conflicts between collection bulk write and client bulk write. This was useful in Rust, and I imagine it will be for other drivers that convert test JSON/YML into native types since the two APIs use different types. I used the "clientBulkWrite" name here for consistency with the unified tests.

Note: we do use the same name for collection- and database-level aggregate in the unified test runner, which was the only precedent I could think of for this situation. However, these operations have the same arguments, so the same deserialization concerns don't apply.

jmikola: I didn't realize `clientBulkWrite` was also being used in tests. Operation names have historically matched the defining spec, excluding custom operations introduced by the test runner (e.g. `createChangeStream`, `createFindCursor`).

If this is required, please make a note of why this naming deviation was done somewhere in the spec. I'm more concerned about this opening the door for future deviations.

isabelatkinson: Gotcha, I added a blurb to the "Entity Test Operations" section to clarify that a) test files should use the same names as those defined in specs and b) why we're deviating here.

source/crud/tests/README.md
```diff
@@ -201,3 +201,70 @@ CommandSucceededEvents. Then, insert an invalid document (e.g. `{x: 1}`) and ass
...
+
+Construct the following write model (referred to as `model`):
+
+```json
+InsertOne: {
+  "namespace": "db.coll",
+  "document": document
+}
+```
+
```
kevinAlbs, line 259: Suggest testing the maximum in one batch to ensure drivers are not splitting earlier than expected.


```suggestion

#### Test maximum in one batch

Use the following calculation to determine the number of inserts that should be provided to `MongoClient.bulkWrite`:
`maxMessageSizeBytes / maxBsonObjectSize` (referred to as `numModels`). This number is intended to be close
to the maximum before splitting.

Construct as list of write models (referred to as `models`) with `model` repeated `numModels` times. Then execute
`bulkWrite` on `client` with `models`. Assert that the bulk write succeeds and returns a `BulkWriteResult` with
an `insertedCount` value of `numModels`.

Assert that one CommandStartedEvents (referred to as `firstEvent`) was observed. Assert that the length of
`firstEvent.command.ops` is equal to `numModels`.

#### Test two batches
```

kevinAlbs (commented):
Tests pass in C driver locally. The minimum was implemented to get tests passing to identify possible additional tests. Suggest adding tests from [this branch](https://github.com/isabelatkinson/specifications/compare/bulk-write...kevinAlbs:specifications:bulk-write-additional-tests?expand=1):

- Test writeConcern option is applied: https://github.com/mongodb/specifications/commit/c62884841de208edb319e8ff92c115c3eb2862d2
- Test writeConcern is inherited from client, and can be overriden by the option: https://github.com/mongodb/specifications/commit/28bce6c56df48945519466dc9a78613cda51f8eb
- Test `bypassDocumentValidation=false` is sent: https://github.com/mongodb/specifications/commit/1537faadef3b0805a997f530d868106c90d0c15a
- Test an empty bulkWrite is a client-side error: https://github.com/mongodb/specifications/commit/a8fb40f04698469ddb5721755d3d7f9300df5504
- Test client bulkWrite with `retryWrites: false` does not retry: https://github.com/mongodb/specifications/commit/96e883511eaf12d98d16deeb029525b887836ab5
- Test unaknowledged writes check insert size: https://github.com/mongodb/specifications/commit/f26567269ee492f360dc3f2b518b4cbf3e719be0
- Test cursor iteration fails (with a failpoint), and cursor iteration succeeds. Example of getting a non-zero cursor ID: https://gist.github.com/kevinAlbs/e4f3046cbb94201c181d1d3341f28374

source/crud/bulk-write.md
```diff
@@ -0,0 +1,787 @@
...
+a write concern error is encountered, it should not terminate execution of the bulk write for
+either ordered or unordered bulk writes. However, drivers MUST throw an exception at the end of
+execution if any write concern errors were observed.
+
+### Individual Write Errors
+
+Individual write errors retrieved from the cursor are recorded in the `writeErrors` field on
+`BulkWriteException`. If an individual write error is encountered during an ordered bulk write,
+drivers MUST record the error in `writeErrors` and immediately throw the exception. Otherwise,
+drivers MUST continue to iterate the results cursor and execute any further `bulkWrite` batches.
```
kevinAlbs, line 714: Suggest testing `writeErrors` are collected across batches for unordered bulk writes: https://github.com/mongodb/specifications/commit/cf5743812843fdbaabee099cce85309033743c20

This revealed a bug in the C driver implementation (was not incrementing `txnNumber`).

source/crud/bulk-write.md
```diff
@@ -0,0 +1,787 @@
...
+Encountering a top-level error MUST halt execution of a bulk write for both ordered and unordered
+bulk writes. This means that drivers MUST NOT attempt to retrieve more responses from the cursor or
+execute any further `bulkWrite` batches and MUST immediately throw an exception.
+
+### Write Concern Errors
+
+Write concern errors are recorded in the `writeConcernErrors` field on `BulkWriteException`. When
+a write concern error is encountered, it should not terminate execution of the bulk write for
+either ordered or unordered bulk writes. However, drivers MUST throw an exception at the end of
+execution if any write concern errors were observed.
```
kevinAlbs, line 707: Suggest testing `writeConcernErrors` are collected across batches: https://github.com/mongodb/specifications/commit/44b7efe56099b8286fd80ebec7036cfff22af5f7

kevinAlbs (commented):
Suggested more tests after discovering bugs in the C implementation.

kevinAlbs (commented):
Suggested two additional tests related.

isabelatkinson (commented):
@kevinAlbs I pushed changes for your inline comments. All of the tests pass locally for Rust. I have not yet added the changes suggested in your overall comment.

source/crud/bulk-write.md
```diff
@@ -0,0 +1,810 @@
...
+When a top-level error is encountered and individual results and/or errors have already been
+observed, drivers MUST embed the top-level error within a `BulkWriteException` as the `error` field
+to retain this information. Otherwise, drivers MAY throw an exception containing only the top-level
+error.
+
+Encountering a top-level error MUST halt execution of a bulk write for both ordered and unordered
+bulk writes. This means that drivers MUST NOT attempt to retrieve more responses from the cursor or
+execute any further `bulkWrite` batches and MUST immediately throw an exception. If the results
+cursor has not been exhausted on the server when a top-level error occurs, drivers MUST send the
+`killCursors` command to attempt to close it.
```
kevinAlbs, line 723:
```suggestion
`killCursors` command to attempt to close it. The result from the `killCursors` command MAY be ignored.
```

I expect sending `killCursors` is a best-effort way to clean-up the cursor. If it fails, ignore the error. Suggested sentence is copied from [Find, getMore and killCursors commands](https://github.com/mongodb/specifications/blob/611b12ccbdd012dcd9ab2877a32200b3835c97af/source/find_getmore_killcursors_commands.rst#L446).

source/crud/tests/unified/client-bulkWrite-errors.yml
```diff
@@ -0,0 +1,228 @@
...
+runOnRequirements:
+  - minServerVersion: "8.0"
+
+createEntities:
+  - client:
+      id: &client0 client0
+      observeEvents: [ commandStartedEvent ]
+      uriOptions:
+        retryWrites: false
+      useMultipleMongoses: false
```
kevinAlbs, line 12:
```suggestion
      useMultipleMongoses: false # Target a single mongos with failpoint
```

source/crud/tests/README.md
```diff
@@ -201,3 +201,70 @@ CommandSucceededEvents. Then, insert an invalid document (e.g. `{x: 1}`) and ass
...
+CommandStartedEvents. Perform a `hello` command using `client` and record the `maxWriteBatchSize` value contained in the
+response. Then, construct the following write model (referred to as `writeModel`):
+
+```json
+InsertOne: {
+  "namespace": "db.coll",
+  "document": { "a": "b" }
+}
+```
+
```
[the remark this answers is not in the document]
kevinAlbs: Good point. Retracting suggestion.

source/crud/tests/README.md
kevinAlbs: Good point. Retracting suggestion.

source/crud/tests/README.md
```diff
@@ -201,3 +201,315 @@ CommandSucceededEvents. Then, insert an invalid document (e.g. `{x: 1}`) and ass
...
+```json
+InsertOne {
+  "namespace": namespace,
+  "document": document
+}
+```
+
+Construct a list of write models (referred to as `models`) with `model` repeated 2 times. Execute `bulkWrite` on
+`client` with `models` and `ordered` set to false. Assert that the bulk write fails and returns a `BulkWriteError`
```
kevinAlbs, line 406: Suggest also testing cursor iteration in a transaction: https://github.com/mongodb/specifications/commit/96656df8da7cc742b6d2ac8600faef1882c4fc4b

This may ensure drivers use the same session for the `bulkWrite` and `getMore`.

In the C implementation, not setting the same session on `getMore` results in an error:
```
Cannot run getMore on cursor 1983114598982261992, which was created in session 7ffcebae-ac99-40b4-b59f-830748d7d19f - 47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU= -  - , in session 9bf2e0e8-c903-460b-9f35-258c76777dbb - 47DEQpj8HBSa+/TImW+5JCeuQeRkm5NMpJWZG3hSuFU=
```

source/crud/bulk-write.md
isabelatkinson: updated

source/crud/tests/unified/client-bulkWrite-errors.yml
isabelatkinson: added

source/crud/tests/README.md
```diff
@@ -201,3 +201,264 @@ CommandSucceededEvents. Then, insert an invalid document (e.g. `{x: 1}`) and ass
...
+(referred to as `bulkWriteError`).
+
+Assert that the length of `bulkWriteError.writeErrors` is 1.
+
+Assert that `bulkWriteError.error` is populated with an error with error code 8.
+
+Assert that a CommandStartedEvent was observed for the `getMore` command.
```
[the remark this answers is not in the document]
isabelatkinson: added

source/crud/bulk-write.md
```diff
@@ -0,0 +1,810 @@
...
+
+## Handling Errors
+
+### Top-Level Errors
+
+A top-level error is any error that occurs that is not the result of a single write operation
+failing or a write concern error. Examples include network errors that occur when communicating
+with the server, command errors (`{ "ok": 0 }`) returned from the server, client-side errors, and
+errors that occur when attempting to perform a `getMore` to retrieve results from the server.
+
```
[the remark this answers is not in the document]
isabelatkinson: Added language and synced the test. We haven't done DRIVERS-2385 in Rust yet so I skipped it in my implementation, will assume it passes in C.

source/crud/tests/README.md
isabelatkinson: Added a test. The new test uses large upserted IDs rather than duplicate key errors to create a cursor with a nonzero ID because ordered:false is not respected within a transaction. I updated the other tests to use this strategy for consistency.

kevinAlbs (commented):
Finished reviewing tests. Surveying tests containing `bulkWrite` operations helped identify possible additions. Here are notable tests that may be worth adding a `client-bulkWrite` equivalent:

`client-side-encryption/etc/test-templates/bulk.yml.template`
Suggest ignoring. Tracked in DRIVERS-2888 for future work.

`client-side-operations-timeout/tests/bulkWrite.yml`.
Consider adding. C and Rust do not yet implement CSOT. A driver with CSOT implemented (Java or Python) can verify.

`command-logging-and-monitoring/tests/logging/operation-id.yml`
Consider adding. C does not yet implement logging from CLAM. Rust can verify.

`command-logging-and-monitoring/tests/monitoring/bulkWrite.yml`
Consider adding: https://github.com/mongodb/specifications/commit/97d3b7dcaf04348dab2286f71fb6bbf7f92df36d. Test may be low value (did not require C driver changes).

`command-logging-and-monitoring/tests/monitoring/unacknowledgedBulkWrite.yml`.
Suggest adding: https://github.com/mongodb/specifications/commit/5a0e0a0ca603ffcc13d9d2c606f81cb1ffe5e4e8. Test required C driver changes.

`retryable-writes/tests/unified/handshakeError.yml`
Consider adding: https://github.com/mongodb/specifications/commit/c32b8249ccfe99927f78f2c377ff607c4cbdc1bc. Test may be low value (did not require C driver changes).

`server-selection/tests/logging/operation-id.yml`
Consider adding. C does not implement logging from CLAM. Rust can verify.

`transactions/tests/unified/mongos-pin-auto.yml`
Consider adding: https://github.com/mongodb/specifications/commit/d883efdc89c26a09ee04b65c0b1c3728a21ab28a. Test may be low value (did not require C driver changes).

`transactions/tests/unified/transaction-options.yml`
Consider adding: https://github.com/mongodb/specifications/commit/a0fe8372d170f4669ca1fc7e1b01df0b5bfc7017. Test may be low value (did not require C driver changes).

source/crud/bulk-write.md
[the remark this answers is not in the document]
isabelatkinson: Added.

source/transactions/tests/unified/client-bulkWrite.yml
```diff
@@ -0,0 +1,163 @@
...
+    collectionName: *collection0Name
+    documents:
+      - { _id: 1, x: 11 }
+      - { _id: 2, x: 22 }
+      - { _id: 3, x: 33 }
+      - { _id: 5, x: 55 }
+      - { _id: 6, x: 66 }
+      - { _id: 7, x: 77 }
+
+tests:
```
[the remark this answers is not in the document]
isabelatkinson: Added.

isabelatkinson (commented):
> client-side-encryption/etc/test-templates/bulk.yml.template
Suggest ignoring. Tracked in DRIVERS-2888 for future work.

👍

> client-side-operations-timeout/tests/bulkWrite.yml.
Consider adding. C and Rust do not yet implement CSOT. A driver with CSOT implemented (Java or Python) can verify.

Added a prose test similar to the existing one for inserts.

> command-logging-and-monitoring/tests/logging/operation-id.yml
Consider adding. C does not yet implement logging from CLAM. Rust can verify.

I propose we skip this. The batching prose tests already have assertions on operation IDs in command events. Command events and logging events have the same definitions, so this should be sufficient coverage (assuming that drivers have unified implementations for monitoring/logging, Rust does). We can also make stronger assertions in the prose tests because a) we can trigger multiple bulkWrite commands and b) we can assert that the two commands have the same operation ID.

> command-logging-and-monitoring/tests/monitoring/bulkWrite.yml
Consider adding: 97d3b7d. Test may be low value (did not require C driver changes).

Propose we skip this one as well. Most of the CRUD tests use command monitoring, so we're not getting any coverage here that we don't have already.

> command-logging-and-monitoring/tests/monitoring/unacknowledgedBulkWrite.yml.
Suggest adding: 5a0e0a0. Test required C driver changes.

Added, I can't run this one in Rust but I trust it works in C.

> retryable-writes/tests/unified/handshakeError.yml
Consider adding: c32b824. Test may be low value (did not require C driver changes).

Added.

> server-selection/tests/logging/operation-id.yml
Consider adding. C does not implement logging from CLAM. Rust can verify.

Added, this seems valuable. We're not getting coverage of this in the prose tests. Rust does not implement the `operationId` field (it's optional in the spec), so we'll have to wait for a driver that does to implement this to verify.

> transactions/tests/unified/mongos-pin-auto.yml
Consider adding: d883efd. Test may be low value (did not require C driver changes).
Added.

> transactions/tests/unified/transaction-options.yml
> Consider adding: a0fe837. Test may be low value (did not require C driver changes).
Added.

source/crud/bulk-write.md
```diff
@@ -0,0 +1,814 @@
...
+    "ordered": Optional<Boolean>,
+    "bypassDocumentValidation": Optional<Boolean>,
+    "comment": Optional<BSON value>,
+    "let": Optional<Document>,
+    ...additional operation-agnostic fields
+}
+```
+
+Drivers MUST use document sequences ([`OP_MSG`](../message/OP_MSG.rst) payload type 1) for the
+`ops` and `nsInfo` fields. Drivers MUST NOT use document sequences when auto-encryption is enabled.
```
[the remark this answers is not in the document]
isabelatkinson: `nsInfo` could exceed `maxBsonObjectSize` (~16MB) if

1. a user is doing `maxWriteBatchSize` operations,
2. each operation is on a different namespace, and
3. each namespace is close to the maximum length defined [here](https://www.mongodb.com/docs/manual/reference/limits/#mongodb-limit-Restriction-on-Collection-Names).

Just the namespaces would be 25MB in this scenario, and then there would be additional bytes to wrap each namespace in a document and then put those documents into an array.

Large `nsInfo` payloads could cause earlier batch splitting when they're used within the command document rather than as a document sequence. Batch splitting logic may also become more complex if adding one model to a bulk write `OP_MSG` requires size checks for both the command document and the overall message.

How much of a lift would it be to support multiple document sequences in the C driver? I'm inclined to leave in this requirement and expect drivers to update their `OP_MSG` implementations to support multiple document sequences, which is in line with the `OP_MSG` spec. FWIW, I made this change in the Rust driver https://github.com/mongodb/mongo-rust-driver/pull/1009/and it wasn't too much work.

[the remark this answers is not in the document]
kevinAlbs: > How much of a lift would it be to support multiple document sequences in the C driver?

It was not significant. Change has been applied to the C implementation.

Retracting suggestion.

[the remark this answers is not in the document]
isabelatkinson: Sounds good, I'll add a Q&A entry for this.

source/crud/tests/README.md
```diff
@@ -201,3 +201,357 @@ CommandSucceededEvents. Then, insert an invalid document (e.g. `{x: 1}`) and ass
...
+Construct a list of write models (referred to as `models`) with `model` repeated `maxWriteBatchSize + 1` times.
+Execute `bulkWrite` on `client` with `models`. Assert that the bulk write succeeds and returns a `BulkWriteResult`
+with an `insertedCount` value of `maxWriteBatchSize + 1`.
+
+Assert that two CommandStartedEvents (referred to as `firstEvent` and `secondEvent`) were observed for the `bulkWrite` command.
+Assert that the length of `firstEvent.command.ops` is `maxWriteBatchSize`. Assert that the length of `secondEvent.command.ops`
+is 1. If the driver exposes `operationId`s in its CommandStartedEvents, assert that `firstEvent.operationId` is equal to
+`secondEvent.operationId`.
+
+### 4. `MongoClient.bulkWrite` batch splits when an `ops` payload exceeds `maxMessageSizeBytes`
```
jyemin, line 233: Consider adding a test for the following scenario: a write with an as-yet-unseen namespace is encoded into ops, and maxMessageSizeBytes is not yet exceeded, but the addition of one more element to nsInfo causes maxMessageSizeBytes to be exceeded. In that case, the driver has to remove the last element from both ops and nsinfo.

source/crud/bulk-write.md
```diff
@@ -0,0 +1,820 @@
...
+
+- The size of a document to be inserted MUST NOT exceed `maxBsonObjectSize`. This applies to the
+  `document` field of an `InsertOneModel` and the `replacement` field of a `ReplaceOneModel`.
+- The size of an entry in the `ops` array MUST NOT exceed `maxBsonObjectSize + 16KiB`.
+- The size of the `bulkWrite` command document MUST NOT exceed `maxBsonObjectSize + 16KiB`.
+
+See [SERVER-10643](https://jira.mongodb.org/browse/SERVER-10643) for more details on these size
+limits.
+
+## Command Batching
```
jyemin, line 595: Lets specify whether `nsInfo` contains only the namespaces actually used in the current batch, or all the namespaces passed to the `bulkWrite` method (as these can be different depending on the split points).  The latter may be simpler to implement, as it makes the logic for calculating the split points more straightforward (you can pre-calculate the size of `nsInfo` before you even execute the splitting logic).

jyemin (changes requested):

source/crud/tests/README.md
jyemin: It may be brittle, but the consequence of getting this wrong could be sending a message to the server that is too large and will be rejected.  Without a test for it, I fear that we will have a bunch of difficult-to-find bugs in drivers around this edge case.

source/crud/bulk-write.md
jyemin: @kevinAlbs I think C driver does the latter?

isabelatkinson: I pushed a commit with more specific batch splitting logic plus a Q&A entry explaining the reasoning behind the design. This change should make implementation and testing of splitting up `nsInfo` simpler. @kevinAlbs, let me know if this proposed new logic makes sense for your implementation. I'm specifically curious whether differentiating between command-specific fields and command-agnostic fields is sensible. Approximating in this way will let us avoid a pretty major refactor in Rust. I will also add some tests around batch splitting on nsInfo that can rely on the 1000 bytes of overhead (i.e. the test proposed [here](https://github.com/mongodb/specifications/pull/1534#discussion_r1572908056)).

source/crud/bulk-write.md
```diff
@@ -0,0 +1,899 @@
...
+#### Unencrypted bulk writes
+
+When `ops` and `nsInfo` are provided as document sequences, drivers MUST ensure that the total size
+of the `OP_MSG` built for each `bulkWrite` command does not exceed `maxMessageSizeBytes`.
+
+The upper bound for the size of an `OP_MSG` includes opcode-related bytes (e.g. the `OP_MSG`
+header) and operation-agnostic command field bytes (e.g. `txnNumber`, `lsid`). Drivers MUST limit
+the combined size of the `bulkWrite`-specific command fields, `ops` document sequence, and `nsInfo`
+document sequence to `maxMessageSizeBytes - 1000` to account for this overhead. The following
+pseudocode demonstrates how to apply this limit in batch-splitting logic:
```
isabelatkinson, line 634: Previous language for reference:

> Some drivers may perform batch-splitting prior to constructing the full `OP_MSG` to be sent to the server. In this case, drivers MAY use `maxMessageSizeBytes - 16,384` as the upper bound for the combined number of bytes in `ops` and `nsInfo`. 16KiB is subtracted as an approximate overhead allowance to accommodate for the bytes in the rest of the message.

The major difference here is that `bulkWrite` options are now excluded from the overhead allowance, which was problematic because _technically_ a `comment` or `let` document could be > 16,384 bytes and therefore cause the batch-splitting boundaries to be inaccurate.

source/crud/tests/README.md
isabelatkinson: Was able to make this work with some less-brittle numbers with the 1000 byte overhead approximation.

source/crud/bulk-write.md
```diff
@@ -0,0 +1,820 @@
...
+
+#### Unencrypted bulk writes
+
+When `ops` and `nsInfo` are provided as document sequences, drivers MUST ensure that the total size
+of the `OP_MSG` built for each `bulkWrite` command does not exceed `maxMessageSizeBytes`. Some
+drivers may perform batch-splitting prior to constructing the full `OP_MSG` to be sent to the
+server. In this case, drivers MAY use `maxMessageSizeBytes - 16,384` as the upper bound for the
+combined number of bytes in `ops` and `nsInfo`. 16KiB is subtracted as an approximate overhead
+allowance to accommodate for the bytes in the rest of the message.
+
```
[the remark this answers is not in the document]
isabelatkinson: good idea, added

source/crud/bulk-write.md
```diff
@@ -0,0 +1,901 @@
...
+### Number of Writes
+
+`maxWriteBatchSize` defines the total number of writes allowed in one command. Drivers MUST split a
+bulk write into multiple commands if the user provides more than `maxWriteBatchSize` operations in
+the argument for `models`.
+
+### Total Message Size
+
+#### Encrypted bulk writes
```
kevinAlbs, line 618: Out of caution, I suggest:

- Remove mention of supporting auto-encryption: https://github.com/mongodb/specifications/commit/ed483b5e9aee3f78d333cdbe9d9ea411b3d82770
- Return an error if auto-encryption is enabled: https://github.com/mongodb/specifications/commit/e39c82c4f68786d7dd29c0c1b4e1c342bf3cf069
- Add prose test to verify: https://github.com/mongodb/specifications/commit/62814a38b71d9edd796af9f5f0dacf487ead5e83

Rationale: This PR does not include tests for auto encryption. Tests are planned in DRIVERS-2888. As specified, a driver may implement `MongoClient.bulkWrite` from this PR and bypass use of libmongocrypt. That may result in a `MongoClient` configured with auto encryption sending a `bulkWrite` and ignoring auto encryption settings (may send plaintext when ciphertext was intended).

These changes may be soon-removed once DRIVERS-2888 is resolved. But I prefer to keep this PR self-contained. This may enable a driver to release `MongoClient.bulkWrite` without DRIVERS-2888.

source/retryable-writes/retryable-writes.rst
```diff
@@ -143,14 +143,15 @@ Supported single-statement write operations include ``insertOne()``,
 ``updateOne()``, ``replaceOne()``, ``deleteOne()``, ``findOneAndDelete()``,
 ``findOneAndReplace()``, and ``findOneAndUpdate()``.
 
+
```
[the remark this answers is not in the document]
jyemin: Fair enough. Thanks for giving it a try.

jmikola (approved):
Minor comments on spec files. I'm not reviewing the test files.

source/versioned-api/tests/crud-api-version-1.yml
```diff
@@ -155,6 +156,44 @@ tests:
                     multi: { $$unsetOrMatches: false }
                     upsert: true
                 <<: *expectedApiVersion
+  
+  - description: "client bulkWrite appends declared API version"
```
kevinAlbs, line 160:
```suggestion
  - description: "client bulkWrite appends declared API version"
    runOnRequirements:
      - minServerVersion: "8.0" # `bulkWrite` added to server 8.0
```

source/etc/generate-handshakeError-tests.py
```diff
@@ -6,10 +6,16 @@
...
+          - insertOne:
+              namespace: retryable-writes-handshake-tests.coll
+              document: { _id: 8, x: 88 }'''
+
 CLIENT_OPERATIONS = [
     Operation('listDatabases', 'listDatabases', 'client', ['filter: {}']),
     Operation('listDatabaseNames', 'listDatabases', 'client', []),
-    Operation('createChangeStream', 'aggregate', 'client', ['pipeline: []'])
+    Operation('createChangeStream', 'aggregate', 'client', ['pipeline: []']),
+    Operation('clientBulkWrite', 'bulkWrite', 'client', [CLIENT_BULK_WRITE_ARGUMENTS])
```
kevinAlbs, line 18: Skip `clientBulkWrite` handshake tests on pre-8.0 servers: https://github.com/mongodb/specifications/commit/b4a224440434175563180c8662558233fd208e31

source/transactions/tests/unified/mongos-pin-auto-tests.py
```diff
@@ -291,6 +291,11 @@
             insert: *collection_name
             documents:
               - { _id : 1 }'''),
+    # clientBulkWrite:
+    'clientBulkWrite': ('bulkWrite', '*client0', r'''models:
```
kevinAlbs, line 295: Skip test on pre-8.0 servers: https://github.com/mongodb/specifications/commit/afd52d1e1c411387b7a6b5ac0516e176d8e7a1bd

kevinAlbs (approved):
LGTM. Great work!

jyemin (approved):
LGTM!

PR #1536, opened 2024-03-05
DRIVERS-1710: Convert retryable reads spec tests to unified test format

jmikola:
https://jira.mongodb.org/browse/DRIVERS-1710

- [ ] Update changelog.
  clusters, and serverless).


source/retryable-reads/tests/unified/gridfs-download-serverErrors.yml
```diff
@@ -0,0 +1,421 @@
...
+                    $oid: '000000000000000000000001'
+              databaseName: *database_name
+          - *retryable_command_started_event
+          - &find_chunks_command_started_event
+            commandStartedEvent:
+              command:
+                find: *chunks_collection_name
+                # Avoid checking additional fields since the exact query may
+                # differ among drivers. expectResult is sufficient to assert
+                # correctness.
```
jmikola, line 107: Quoting the commit:

> Drivers may differ in how they query the chunks collection. For instance, [PHPLIB queries](https://github.com/mongodb/mongo-php-library/blob/1.17.0/src/GridFS/CollectionWrapper.php#L115) the chunk's n field with $gte in addition to files_id. Rather than use $$matchAsRoot (schema version 1.13+) to emulate the legacy test behavior of ignoring extra fields, the commandStartedEvent assertion was relaxed and expectResult was introduced to assert that the file is read successfully.

I know various drivers are behind on schema versions, so sticking with an older version seemed preferable. The chunk query also isn't vital to these tests, and `expectResult` with `$$matchesHexBytes` is overall an improvement IMO.

source/retryable-reads/tests/README.rst
```diff
@@ -9,192 +9,18 @@ Retryable Reads Tests
...
-
-Optional Enumeration Commands
-=============================
-
-A driver only needs to test the optional enumeration commands it has chosen to
-implement (e.g. ``Database.listCollectionNames()``).
-
-PoolClearedError Retryability Test
-==================================
+1. PoolClearedError Retryability Test
```
jmikola, line 22: Note: prose tests were originally not numbered.

source/retryable-reads/tests/unified/listCollectionObjects.yml
```diff
@@ -0,0 +1,136 @@
...
+    databaseName: *database_name
+    documents: []
+
+tests:
+  -
+    description: 'ListCollectionObjects succeeds on first attempt'
+    operations:
+      - &retryable_operation
+        object: *database0
+        name: listCollectionObjects
```
jyemin, line 46: We already have `listCollections` in other spec tests.  Is `listCollectionObjects` just an alias for that, and if so, is it necessary?

Same for `listDatabaseObjects`.

jyemin: I see.  If a driver doesn't offer those methods, I guess it can skip the tests of them, as they appear to be the same set of tests as for the non-object variants.

jmikola: > If a driver doesn't offer those methods, I guess it can skip the tests of them

Yes. AFAIK, that is the standard practice for any driver when dealing with tests for unsupported/unimplemented functionality.

jyemin (approved):

kkloberdanz (approved):

alcaeus (approved):

PR #1538, opened 2024-03-06
DRIVERS-2789 Convert DNS Seedlist Spec to Markdown

blink1073:

- [x] Update changelog.
  clusters, and serverless).


scripts/migrate_to_md.py
```diff
@@ -115,6 +116,7 @@
 with open(md_file, 'w') as fid:
     fid.write('\n'.join(new_lines))
 
+
```
NoahStapp, line 119: Unneeded space?

NoahStapp (changes requested):

scripts/migrate_to_md.py
blink1073: Fixed

NoahStapp (approved):

PR #1541, opened 2024-03-06
DRIVERS-2768 [Vector Search GA] Add support for types in search index creation

NoahStapp:
[DRIVERS-2768](https://jira.mongodb.org/browse/DRIVERS-2768)

- [x] Update changelog.
  clusters, and serverless).

### DIFF
--- a/source/index-management/index-management.md
+++ b/source/index-management/index-management.md
@@ -666,15 +666,6 @@ interface IndexOptions {
   name: String;
   
   /**
-   * Optionally specify a type for the index. Defaults to "search" if not provided.
-   * Either "search" for regular search indexes or "vectorSearch" for vector search indexes.
-   *
-   * Note that to create a vector search index using a helper method, the type "vectorSearch" must be provided.
-   *
-   */
-  type: String;
-
-  /**
    * Optionally tells the index to only reference documents with the specified field in
    * the index.
    */
@@ -912,6 +903,9 @@ interface SearchIndexModel {
 
   // The name for this index, if present.
   name: Optional<string>;
+     
+  // The type for this index, if present. Can be either "search" or "vectorSearch", defaulting to "search".  
+  type: Optional<string>;
 }
 
 interface SearchIndexOptions {
--- a/source/index-management/index-management.md
+++ b/source/index-management/index-management.md
@@ -664,6 +664,15 @@ interface IndexOptions {
    * @example For an index of name: 1, age: -1, the generated name would be "name_1_age_-1".
    */
   name: String;
+  
+  /**
+   * Optionally specify a type for the index. Defaults to "search" if not provided.
+   * Either "search" for regular search indexes or "vectorSearch" for vector search indexes.
+   *
+   * Note that to create a vector search index using a helper method, the type "vectorSearch" must be provided.
+   *
+   */
+  type: String;
 
   /**
    * Optionally tells the index to only reference documents with the specified field in
--- a/source/index-management/index-management.md
+++ b/source/index-management/index-management.md
@@ -904,7 +904,7 @@ interface SearchIndexModel {
   // The name for this index, if present.
   name: Optional<string>;
      
-  // The type for this index, if present. Can be either "search" or "vectorSearch", defaulting to "search".  
+  // The type for this index, if present. Can be either "search" or "vectorSearch".  
   type: Optional<string>;
 }
 
--- a/source/index-management/tests/README.md
+++ b/source/index-management/tests/README.md
@@ -283,23 +283,23 @@ This test fails if it times out waiting for the deletion to succeed.
 
 10. Create a new vector search index on `coll0` with the `createSearchIndex` helper. Use the following definition:
 
-```typescript
+    ```typescript
 
-  {
-    name: 'test-search-index-case7-vector',
-    type: 'vectorSearch',
-    definition: {
-      "fields": [
-         {
-             "type": "vector",
-             "path": "plot_embedding",
-             "numDimensions": 1536,
-             "similarity": "euclidean",
-         },
-      ]
-    }
-  }
-```
+      {
+        name: 'test-search-index-case7-vector',
+        type: 'vectorSearch',
+        definition: {
+          fields: [
+             {
+                 type: 'vector',
+                 path: 'plot_embedding',
+                 numDimensions: 1536,
+                 similarity: 'euclidean',
+             },
+          ]
+        }
+      }
+    ```
 
 11. Assert that the command returns the name of the index: `"test-search-index-case7-vector"`.
 
@@ -311,23 +311,28 @@ This test fails if it times out waiting for the deletion to succeed.
 
 13. Assert that `index3` has a property `type` whose value is `vectorSearch`.
 
-14. Create a new vector search index on `coll0` with the `createSearchIndex` helper. Use the following definition:
+#### Case 8: Driver requires explicit type to create a vector search index
 
-```typescript
+1. Create a collection with the "create" command using a randomly generated name (referred to as `coll0`).
 
-  {
-    name: 'test-search-index-case7-error',
-    definition: {
-      "fields": [
-         {
-             "type": "vector",
-             "path": "plot_embedding",
-             "numDimensions": 1536,
-             "similarity": "euclidean",
-         },
-      ]
-    }
-  }
-```
+2. Create a new vector search index on `coll0` with the `createSearchIndex` helper. Use the following definition:
 
-15. Assert that the command throws an exception due to the `mappings` field missing.
+   ```typescript
+
+     {
+       name: 'test-search-index-case8-error',
+       definition: {
+         fields: [
+            {
+                type: 'vector',
+                path: 'plot_embedding',
+                numDimensions: 1536,
+                similarity: 'euclidean',
+            },
+         ]
+       }
+     }
+   ```
+
+3. Assert that the command throws an exception containing the string "Attribute mappings missing" due to the `mappings`
+   field missing.
--- a/source/index-management/tests/README.md
+++ b/source/index-management/tests/README.md
@@ -234,94 +234,100 @@ This test fails if it times out waiting for the deletion to succeed.
 
 #### Case 7: Driver can successfully handle search index types when creating indexes
 
-1. Create a collection with the "create" command using a randomly generated name (referred to as ``coll0``).
+01. Create a collection with the "create" command using a randomly generated name (referred to as `coll0`).
 
-2. Create a new search index on ``coll0`` with the ``createSearchIndex`` helper. Use the following definition:
+02. Create a new search index on `coll0` with the `createSearchIndex` helper. Use the following definition:
 
-   ```typescript
+    ```typescript
 
-     {
-       name: 'test-search-index-case7-implicit',
-       definition: {
-         mappings: { dynamic: false }
-       }
-     }
-   ```
+      {
+        name: 'test-search-index-case7-implicit',
+        definition: {
+          mappings: { dynamic: false }
+        }
+      }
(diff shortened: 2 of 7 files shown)


blink1073:
Can you also add new spec tests that use `type:vector` to ensure they fail in a non-Atlas environment?

blink1073:
We should also either update one of the existing prose tests or add a new one that explicitly uses `type: search`.

NoahStapp:
> We should also either update one of the existing prose tests or add a new one that explicitly uses `type: search`.

The new prose test does explicitly create an index using `type: search`.

NoahStapp:
> Can you also add new spec tests that use `type:vector` to ensure they fail in a non-Atlas environment?

The existing spec tests aren't run against an Atlas cluster, they're used as unit tests to verify the driver sends the correct command document. They expect an error telling the user to use Atlas for search support. I'm not sure adding more of the same tests that have the `vectorSearch` type makes sense in that context.

blink1073:
> I'm not sure adding more of the same tests that have the vectorSearch type makes sense in that context.

Perhaps just one extra test to make sure the field is handled appropriately?

source/index-management/tests/README.md
```diff
@@ -231,3 +231,97 @@ This test fails if it times out waiting for the deletion to succeed.
...
+     }
+   ```
+
+3. Assert that the command returns the name of the index: ``"test-search-index-case7-implicit"``.
+
+4. Run ``coll0.listSearchIndexes('test-search-index-case7-implicit')`` repeatedly every 5 seconds until the following condition is satisfied and store the value in a variable ``index1``:
+
+   - An index with the ``name`` of ``test-search-index-case7-implicit`` is present and the index has a field ``queryable`` with a value of ``true``.
+
+5. Assert that ``index1`` has a property ``type`` whose value is ``search``
```
blink1073, line 257:
```suggestion
5. Assert that ``index1`` has a property ``type`` whose value is ``search``.
```

source/index-management/tests/README.md
```diff
@@ -231,3 +231,97 @@ This test fails if it times out waiting for the deletion to succeed.
...
+     }
+   ```
+
+7. Assert that the command returns the name of the index: ``"test-search-index-case7-explicit"``.
+
+8. Run ``coll0.listSearchIndexes('test-search-index-case7-explicit')`` repeatedly every 5 seconds until the following condition is satisfied and store the value in a variable ``index2``:
+
+   - An index with the ``name`` of ``test-search-index-case7-explicit`` is present and the index has a field ``queryable`` with a value of ``true``.
+
+9. Assert that ``index2`` has a property ``type`` whose value is ``search``
```
blink1073, line 278:
```suggestion
9. Assert that ``index2`` has a property ``type`` whose value is ``search``.
```

source/index-management/tests/README.md
```diff
@@ -231,3 +231,97 @@ This test fails if it times out waiting for the deletion to succeed.
...
+     }
+   ```
+
+11. Assert that the command returns the name of the index: ``"test-search-index-case7-vector"``.
+
+12. Run ``coll0.listSearchIndexes('test-search-index-case7-vector')`` repeatedly every 5 seconds until the following condition is satisfied and store the value in a variable ``index3``:
+     
+    - An index with the ``name`` of ``test-search-index-case7-vector`` is present and the index has a field ``queryable`` with a value of ``true``.
+
+13. Assert that ``index3`` has a property ``type`` whose value is ``vectorSearch``
```
blink1073, line 306:
```suggestion
13. Assert that ``index3`` has a property ``type`` whose value is ``vectorSearch``.
```

source/index-management/tests/createSearchIndexes.json
```diff
@@ -157,7 +160,70 @@
...
+                  "$db": "database0"
+                }
+              }
+            }
+          ]
+        }
+      ]
+    },
+        {
+      "description": "create a vector search index",
```
blink1073, line 176: This test is not in the yml file.

blink1073 (changes requested):

blink1073:
Can you please install and run `pre-commit run --all-files` locally?

source/index-management/tests/createSearchIndexes.json
NoahStapp: Good catch, fixed!

blink1073 (approved):
LGTM!  Let's hold off on merging until the first two implementations are done.

source/index-management/index-management.md
```diff
@@ -664,6 +664,15 @@ interface IndexOptions {
    * @example For an index of name: 1, age: -1, the generated name would be "name_1_age_-1".
    */
   name: String;
+  
+  /**
+   * Optionally specify a type for the index. Defaults to "search" if not provided.
```
ShaneHarvey, line 669: Many other driver options are implemented such that we only send a value if the user explicitly provides one. How do you feel about mandating that driver MUST NOT send this field if not provided?

NoahStapp: Makes sense to me if that's a consistent pattern among drivers. The server should default to `search` implicitly if `type` is not provided, so the behavior is consistent.

NoahStapp: `CreateSearchIndexOptions` is currently unused. If drivers support adding `type` to `IndexOptions`, that seems to be a better fit as then the top-level fields of`type` and `name` are present in the same interface.

source/index-management/tests/README.md
```diff
@@ -231,3 +231,103 @@ This test fails if it times out waiting for the deletion to succeed.
      of `true`.
 
 6. Assert that `index` has a property `latestDefinition` whose value is `{ 'mappings': { 'dynamic': false } }`
+
+#### Case 7: Driver can successfully handle search index types when creating indexes
+
+01. Create a collection with the "create" command using a randomly generated name (referred to as `coll0`).
+
+02. Create a new search index on `coll0` with the `createSearchIndex` helper. Use the following definition:
```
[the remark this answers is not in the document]
NoahStapp: The command is identical, yes. Prose test 1 does not verify that `type` is present or that its value is correct. We could modify prose test 1 to include the similar portion and remove this duplicated code if that's preferable.

source/index-management/tests/README.md
```diff
@@ -231,3 +231,103 @@ This test fails if it times out waiting for the deletion to succeed.
...
+
+12. Run `coll0.listSearchIndexes('test-search-index-case7-vector')` repeatedly every 5 seconds until the following
+    condition is satisfied and store the value in a variable `index3`:
+
+    - An index with the `name` of `test-search-index-case7-vector` is present and the index has a field `queryable` with
+      a value of `true`.
+
+13. Assert that `index3` has a property `type` whose value is `vectorSearch`.
+
+14. Create a new vector search index on `coll0` with the `createSearchIndex` helper. Use the following definition:
```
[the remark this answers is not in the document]
NoahStapp: Breaking these final steps into a separate test makes sense, agreed. The current error returned by the server is not particularly helpful: the Python implementation currently checks that the string `"Attribute mappings missing"` is present in the error message. Adding that to the spec is reasonable while the error message remains vague.

source/index-management/index-management.md
baileympearson: Actually, looking more closely it probably belongs in the `SearchIndexModel`?  

`IndexOptions` is used for regular indexes, not search indexes.

NoahStapp: Oh good catch! Yes, `SearchIndexModel` seems like the correct place for `type`.

source/index-management/tests/README.md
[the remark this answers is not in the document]
NoahStapp: Considering how rapidly the search APIs have developed recently, I think it's valuable to ensure we have a test that verifies the expected default behavior.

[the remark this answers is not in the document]
baileympearson: My preference is not to test server defaults.  This isn't really testing driver behavior, so it doesn't provide much value for drivers.  It also would only fail if the server changes its default, meaning that this test would potentially just fail if the server makes a change to its default.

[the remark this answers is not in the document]
NoahStapp: Normally I agree, but since we've had several occurrences recently where the server behavior has changed and broken the driver expectations, I'd prefer to have a test that ensures the driver and server are aligned here.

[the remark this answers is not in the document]
NoahStapp: The most recent breaking change was that creating a vector search index with `createSearchIndex` required specifying the `type=vectorSearch` field. Previously this was not required, which resulted in several customers asking why they were getting an error message when using their existing workflows.

If the server does change the default and this test fails, I would expect our implementations to change depending on the behavior the results from the new default. For example, now the server returns an error message when trying to create a vector search index without specifying the type, drivers could catch the server's error to surface a more useful message to end users. 

I'm not against removing this specific test, but by doing so I worry that we will end up in a situation where we aren't validating that the server behaves as we expect and are instead relying on assumption rather than an actual test.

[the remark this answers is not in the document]
baileympearson: > If the server does change the default and this test fails, I would expect our implementations to change depending on the behavior the results from the new default. For example, now the server returns an error message when trying to create a vector search index without specifying the type, drivers could catch the server's error to surface a more useful message to end users.

The only assertion unique to this test is step 5:

```
05. Assert that `index1` has a property `type` whose value is `search`.
```

For this test specifically to fail, the server would have to change the default so that `type` is no longer `search`.  right?

I'm not entirely opposed to testing server behavior, so if we want this test let's keep it.  I'm just reluctant to tie driver test assertions to server response values that aren't something controlled by _our_ semver.  For example, recently the server made changes to the structure of `explain` commands' responses, which repeatedly broke Node's tests until we finally stopped asserting on the structure of the response at all.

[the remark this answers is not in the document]
NoahStapp: For this test to fail, the server would have to change the expected schema so that `type` is no longer present or the expected default from `search`, correct. If either of those change, then the driver implementation would have to change as well.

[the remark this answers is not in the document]
baileympearson: I think Shane's comment got lost in the comments on this PR: https://github.com/mongodb/specifications/pull/1541#discussion_r1538228033

Generally, drivers rely on the server's default for optional fields in commands rather than providing a default.  If drivers don't provide a default, than there won't be changes if the server changes its default value.

source/index-management/tests/README.md
```diff
@@ -231,3 +231,108 @@ This test fails if it times out waiting for the deletion to succeed.
...
+                path: 'plot_embedding',
+                numDimensions: 1536,
+                similarity: 'euclidean',
+            },
+         ]
+       }
+     }
+   ```
+
+3. Assert that the command throws an exception containing the string "Attribute mappings missing" due to the `mappings`
```
baileympearson, line 337: Does this test actually confirm that `type` is required?  The error `Attribute mappings missing` seems to indicate that `mappings` must be present.  I added a `mappings` and confirmed this with a patch: https://spruce.mongodb.com/task/mongo_node_driver_next_rhel8_test_search_indexes_test_search_index_helpers_patch_3ae4ff5d3ebb3ebb251eea650e81b8dd4bf444b9_661d5383b25584000708ce66_24_04_15_16_19_48/logs?execution=0

NoahStapp: It confirms that `type` is required to create a vector search index. Vector search indexes use `fields` instead of `mappings`, but the error returned by the server doesn't provide this information.

baileympearson (approved):
POC passing in Node: https://github.com/mongodb/node-mongodb-native/pull/4060

blink1073 (approved):

PR #1544, opened 2024-03-08
DRIVERS-2856 Rename OIDC integration name and values.

blink1073:
(cherry picked from commit dde0632)

- [x] Update changelog.
  clusters, and serverless).

### DIFF
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -1252,7 +1252,7 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
 
 Drivers MUST support all of the following built-in OIDC application environment integrations.
 
-####### Test
+**Test**
 
 The test integration is enabled by setting auth mechanism property `ENVIRONMENT:test`. It is meant for driver testing
 purposes, and is not meant to be documented as a user-facing feature.
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -1254,13 +1254,13 @@ Drivers MUST support all of the following built-in OIDC application environment 
 
 ####### Test
 
-The test integration is enabled by setting auth mechanism property `OIDC_ENV:test`. It is not meant to be documented as
-a user-facing feature, but used for testing purposes, with the self-generated token created by the script in Drivers
-Evergreen Tools.
+The test integration is enabled by setting auth mechanism property `OIDC_ENV:test`. It is meant for driver testing
+purposes, and is not meant to be documented as a user-facing feature.
 
-If enabled, drivers MUST read the file path from environment variable `AWS_WEB_IDENTITY_TOKEN_FILE` and then read the
-OIDC access token from that file. The driver MUST use the contents of that file as value in the `jwt` field of the
-`saslStart` payload.
+If enabled, drivers MUST generate a token using the script in Drivers-Evergreen-Tools. The must then set the
+`AWS_WEB_IDENTITY_TOKEN_FILE` environment variable to the path to that file. At runtime, the driver MUST use the
+`AWS_WEB_IDENTITY_TOKEN_FILE` environment variable and read the OIDC access token from that path. The driver MUST use
+the contents of that file as value in the `jwt` field of the `saslStart` payload.
 
 Drivers MAY implement the "test" integration so that it conforms to the function signature of the
 [OIDC Callback](#oidc-callback) to prevent having to re-implement the "test" integration logic in the OIDC prose tests.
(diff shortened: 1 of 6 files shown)


source/auth/auth.md
```diff
@@ -1246,20 +1246,24 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
...
+<div id="built-in-provider-integrations">
 
-Drivers MUST support all of the following built-in OIDC providers.
+#### Built-in OIDC Environment Integrations
 
-####### AWS
+Drivers MUST support all of the following built-in OIDC application environment integrations.
 
-The AWS provider is enabled by setting auth mechanism property `PROVIDER_NAME:aws`.
+####### Test
```
katcharov, line 1255: The [doc](https://docs.google.com/document/d/1KfjHr-114ADX_2GRW6jTwYbRUsMcJZihS3F5xbOQKFc/edit) suggests that there will be a mechanism that uses:

> A file at a path specified by an environment variable

It seems that the "Test" env, and the current env var, will be redundant with this (future?) env. Could we not use that env? This would also make our tests more useful, because they would test an external API rather than internal.

source/auth/auth.md
```diff
@@ -1246,20 +1246,24 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
...
-If enabled, drivers MUST read the file path from environment variable `AWS_WEB_IDENTITY_TOKEN_FILE` and then read the
-OIDC access token from that file. The driver MUST use the contents of that file as value in the `jwt` field of the
-`saslStart` payload.
+The test integration is enabled by setting auth mechanism property `OIDC_ENV:test`. It is meant for driver testing
+purposes, and is not meant to be documented as a user-facing feature.
 
-Drivers MAY implement the AWS provider so that it conforms to the function signature of the
-[OIDC Callback](#oidc-callback) to prevent having to re-implement the AWS provider logic in the OIDC prose tests.
+If enabled, drivers MUST generate a token using the script in Drivers-Evergreen-Tools. The must then set the
+`AWS_WEB_IDENTITY_TOKEN_FILE` environment variable to the path to that file. At runtime, the driver MUST use the
```
katcharov, line 1261: Since this is a generalized "test" environment, should the var be renamed to exclude AWS, for example, `ACCESS_TOKEN_FILE`?

source/auth/auth.md
```diff
@@ -1246,20 +1246,24 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
...
 
-If enabled, drivers MUST read the file path from environment variable `AWS_WEB_IDENTITY_TOKEN_FILE` and then read the
-OIDC access token from that file. The driver MUST use the contents of that file as value in the `jwt` field of the
-`saslStart` payload.
+The test integration is enabled by setting auth mechanism property `OIDC_ENV:test`. It is meant for driver testing
+purposes, and is not meant to be documented as a user-facing feature.
 
-Drivers MAY implement the AWS provider so that it conforms to the function signature of the
-[OIDC Callback](#oidc-callback) to prevent having to re-implement the AWS provider logic in the OIDC prose tests.
+If enabled, drivers MUST generate a token using the script in Drivers-Evergreen-Tools. The must then set the
```
katcharov, line 1260: the -> they

Could we specify the script that should be used?

katcharov (changes requested):

source/auth/auth.md
blink1073: The kubernetes case will be the only other one that uses an env variable, this is meant to be run locally or on EVG without being on k8s.

source/auth/auth.md
blink1073: That makes sense to me, to differentiate this from EKS, which will use the environment variable.  The tokens we are minting are not the same as the ones that are provided in an EKS environment.  I'll make a PR to update drivers-tools.

source/auth/auth.md
blink1073: I didn't want to all out a specific script, because it would easily get out of date, what if we instead pointed to the auth_oidc README?

alcaeus (approved):
Reviewed unified test format changes, those LGTM.

blink1073:
@katcharov updated and awaiting your review.

source/auth/auth.md
katcharov: Yep any sort of reference works.

katcharov (approved):
LGTM!

PR #1547, opened 2024-03-12
DRIVERS-2855: Convert default write concern tests to unified format

jmikola:
https://jira.mongodb.org/browse/DRIVERS-2855

- [ ] Update changelog.

POC: https://github.com/mongodb/mongo-php-library/pull/1252

### DIFF
--- a/source/read-write-concern/tests/operation/default-write-concern-2.6.json
+++ b/source/read-write-concern/tests/operation/default-write-concern-2.6.json
@@ -1,19 +1,55 @@
 {
-  "data": [
-    {
-      "_id": 1,
-      "x": 11
-    },
-    {
-      "_id": 2,
-      "x": 22
-    }
-  ],
-  "collection_name": "default_write_concern_coll",
-  "database_name": "default_write_concern_db",
-  "runOn": [
+  "description": "default-write-concern-2.6",
+  "schemaVersion": "1.0",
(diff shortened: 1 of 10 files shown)


source/read-write-concern/tests/operation/default-write-concern-2.6.json
```diff
@@ -1,19 +1,55 @@
 {
```
alcaeus, line 1: Nothing major, just noted that for other tests you created a `unified` directory under tests, whereas this PR keeps original files.

alcaeus (approved):
LGTM with a minor note on test location.

source/read-write-concern/tests/operation/default-write-concern-2.6.json
jmikola: This was intentional, as the tests were already under an `operation/` directory. In other specs, legacy tests were either in `tests/` or, in the case of CRUD, a specific directory.

These tests are old and few enough that I'm not worried at all about people mixing up legacy and unified test files.

jmikola:
Note: link check will be fixed by https://github.com/mongodb/specifications/pull/1549

jmikola:
@jyemin: Do you want to POC this in Java before I merge?

jyemin:
Risk seems low on these so go ahead without me.

PR #1551, opened 2024-03-19
DRIVERS-2836 OIDC Spec Cleanup

blink1073:

- [x] Update changelog.
  clusters, and serverless).


blink1073:
https://github.com/mongodb/mongo-python-driver/pull/1556

blink1073:
I'm planning to add one additional prose test suggested by @pmeredit, "in human flow, if there is no cached access token but there is a refresh token, that the refresh token is passed to the callback."

pmeredit (approved):
LGTM. It might be out of scope for this ticket, but for human_4_3 should also say "create a default client with a callback that returns the test_user1 access token and a bad refresh token"

source/auth/auth.md
```diff
@@ -1647,8 +1647,8 @@ Use the following algorithm to authenticate a new connection:
 
 - Check if the the *Client Cache* has an access token.
   - If it does, cache the access token in the *Connection Cache* and perform a `One-Step` SASL conversation using the
-    access token in the *Client Cache*. If the server returns an error, invalidate that access token, sleep 100ms then
-    continue.
+    access token in the *Client Cache*. If the server returns a Authentication error (18), invalidate that access token,
+    sleep 100ms then continue.
```
katcharov, line 1651: In a [prior discussion](https://mongodb.slack.com/archives/C04PGGEFCNM/p1708624000650649?thread_ts=1708109719.923909&cid=C04PGGEFCNM), I believe the plan was to remove the 100ms sleep, and leave it up to the callback.

The problem was "The 20 threads would not hit the callback all at once, but one after the other without a 100ms delay, which only applies to the first thread." - I don't recall the logic, but I think this means that the throttling would not be effective in certain common cases. (Also, there is no need to throttle AWS file-reading.)

Perhaps this is not the right PR to address this, but I think the DRIVERS-2836 ticket was the only one mentioned (so, if not here, could we create a ticket?), and I noticed a comment on the ticket related to the timeout as well.

source/auth/auth.md
```diff
@@ -1660,18 +1660,21 @@ def auth(connection):
   access_token, is_cache = get_access_token()
 
   # If there is a cached access token, try to authenticate with it. If
-  # authentication fails, it's possible the cached access token is expired. In
+  # authentication fails with an Authentication error (18), 
+  # it's possible the cached access token is expired. In
   # that case, invalidate the access token, fetch a new access token, and try
```
katcharov, line 1665:
```suggestion
  # authentication fails with an Authentication error (18), 
  # invalidate the access token, fetch a new access token, and try
```
To clarify. I believe we always invalidate (whether expired, invalidated, or any other possibility), we do not check that it has expired and only then invalidate.

source/auth/auth.md
```diff
@@ -1682,14 +1685,15 @@ authenticate a new connection when a [OIDC Human Callback](#oidc-human-callback)
 
 - Check if the *Client Cache* has an access token.
   - If it does, cache the access token in the *Connection Cache* and perform a [One-Step](#one-step) SASL conversation
-    using the access token. If the server returns an error, invalidate the access token token from the *Client Cache*,
-    clear the *Connection Cache*, and continue.
+    using the access token. If the server returns an Authentication error (18), invalidate the access token token from
+    the *Client Cache*, clear the *Connection Cache*, and continue. If the server returns another error, continue.
```
katcharov, line 1689: See prior comment regarding discussion - this again might not be the right PR, but it seems we should clarify the use of "continues", and should also change the logic to use explicit looping (to avoid the example of 20 redundant invocations of a callback when a token expires across 21 connections).

source/auth/tests/mongodb-oidc.md
```diff
@@ -233,13 +259,38 @@ Drivers MUST be able to authenticate using OIDC callback(s) when there is one pr
...
+**2.2 Human Callback Returns Missing Data**
 
 - Create a `MongoClient` with a human callback that returns data not conforming to the `OIDCCredential` with missing
   fields.
 - Perform a `find` operation that fails.
 - Close the client.
 
+**2.3 Refresh Token Is Passed To The Callback**
+
+- Create a `MongoClien`t with a human callback that checks for the presence of a refresh token.
```
katcharov, line 271:
```suggestion
- Create a `MongoClient` with a human callback that checks for the presence of a refresh token.
```

source/auth/tests/mongodb-oidc.md
```diff
@@ -233,13 +259,38 @@ Drivers MUST be able to authenticate using OIDC callback(s) when there is one pr
...
+      "find"
+    ],
+    errorCode: 391
+  }
+}
+```
+
+- Perform a `find` operation that succeeds.
+- Assert that the callback has been called twice.
+- Assert that the refresh token was used once.
```
katcharov, line 292:
```suggestion
- Assert that the refresh token was provided to the callback once.
```
If that is correct?

katcharov (changes requested):
Just some minor clarifications, and questions about where best to address some remaining issues.

source/auth/tests/mongodb-oidc.md
blink1073: Comment is now out of date

source/auth/auth.md
blink1073: Updated

source/auth/auth.md
blink1073: Updated

source/auth/auth.md
```diff
@@ -1744,14 +1746,15 @@ authenticate a new connection when a [OIDC Human Callback](#oidc-human-callback)
 
 - Check if the *Client Cache* has an access token.
   - If it does, cache the access token in the *Connection Cache* and perform a [One-Step](#one-step) SASL conversation
-    using the access token. If the server returns an error, invalidate the access token token from the *Client Cache*,
-    clear the *Connection Cache*, and continue.
+    using the access token. If the server returns an Authentication error (18), invalidate the access token token from
+    the *Client Cache*, clear the *Connection Cache*, and restart the authentication flow. Raise any other errors to the
+    user.
```
katcharov, line 1751:
```suggestion
    user. On success, exit the algorithm.
```

source/auth/auth.md
```diff
@@ -1744,14 +1746,15 @@ authenticate a new connection when a [OIDC Human Callback](#oidc-human-callback)
...
 - Check if the *Client Cache* has a refresh token.
   - If it does, call the [OIDC Human Callback](#oidc-human-callback) with the cached refresh token and `IdpInfo` to get
     a new access token. Cache the new access token in the *Client Cache* and *Connection Cache*. Perform a
-    [One-Step](#one-step) SASL conversation using the new access token. If the
-    [OIDC Human Callback](#oidc-human-callback) or the server return an error, invalidate the access token from the
-    *Client Cache*, clear the *Connection Cache*, and continue.
+    [One-Step](#one-step) SASL conversation using the new access token. If the the server returns an Authentication
+    error (18), clear the refresh token, invalidate the access token from the *Client Cache*, clear the *Connection
+    Cache*, and restart the authentication flow. Raise any other errors to the user.
```
katcharov, line 1757:
```suggestion
    Cache*, and restart the authentication flow. Raise any other errors to the user.
    On success, exit the algorithm.
```

source/auth/auth.md
```diff
@@ -1709,9 +1709,9 @@ Use the following algorithm to authenticate a new connection:
 
 - Check if the the *Client Cache* has an access token.
   - If it does, cache the access token in the *Connection Cache* and perform a `One-Step` SASL conversation using the
-    access token in the *Client Cache*. If the server returns an error, invalidate that access token, sleep 100ms then
-    continue.
-- Call the configured built-in provider integration or the OIDC callback to retrieve a new access token.
+    access token in the *Client Cache*. If the server returns a Authentication error (18), invalidate that access token.
```
katcharov, line 1712:
```suggestion
    access token in the *Client Cache*. If the server returns a Authentication error (18), invalidate that access token.
    Raise any other errors to the user. On success, exit the algorithm.
```

source/auth/auth.md
```diff
@@ -1722,18 +1722,20 @@ def auth(connection):
...
       sasl_start(connection, payload={"jwt": access_token})
       return
-    except ServerError:
-      invalidate(access_token)
-      sleep(0.1)
-      access_token, _ = get_access_token()
+    except ServerError as e:
+      if e.code == 18:
+        invalidate(access_token)
+        sleep(0.1)
```
katcharov, line 1737:
```suggestion
```
The sleep is no longer present after invalidation. I think the new delay is outside the example pseudocode.

source/auth/auth.md
```diff
@@ -1709,9 +1709,9 @@ Use the following algorithm to authenticate a new connection:
 
 - Check if the the *Client Cache* has an access token.
```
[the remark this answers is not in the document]
blink1073: Fixed

source/auth/auth.md
```diff
@@ -1480,7 +1479,7 @@ An example human callback API might look like:
 ```typescript
 interface IdpInfo {
   issuer: string;
-  clientId: string;
```
pmeredit, line 1483: What is the intended behavior when `clientId` is `None`? The code I have written on client side assumes the `clientId` will always exist.

source/auth/tests/mongodb-oidc.md
```diff
@@ -209,13 +253,38 @@ Drivers MUST be able to authenticate using OIDC callback(s) when there is one pr
...
+**2.2 Human Callback Returns Missing Data**
 
 - Create an OIDC configured client with a human callback that returns data not conforming to the `OIDCCredential` with
   missing fields.
 - Perform a `find` operation that fails.
 - Close the client.
 
+**2.3 Refresh Token Is Passed To The Callback**
+
+- Create a `MongoClien`t with a human callback that checks for the presence of a refresh token.
```
pmeredit, line 265:
```suggestion
- Create a `MongoClient` with a human callback that checks for the presence of a refresh token.
```

pmeredit (approved):
LGTM other than the one question and a typo :)

source/auth/auth.md
blink1073: It would be None if the user used a Human Callback and the IdP is configured on the server with `supportsHumanFlows:False`.  They shouldn't really be doing that, but it is allowed.  See https://jira.mongodb.org/browse/DRIVERS-2773 for details.

pmeredit: We may need a test to cover this case

pmeredit (approved):
LGTM!

katcharov (approved):
LGTM, with the 2 suggestions

PR #1553, opened 2024-03-21
DRIVERS-2871 CSOT tests: Expect withTransaction to propagate errors

rozza:
DRIVERS-2871

- [ ] Update changelog.
  clusters, and serverless).

### DIFF
--- a/source/client-side-operations-timeout/tests/convenient-transactions.yml
+++ b/source/client-side-operations-timeout/tests/convenient-transactions.yml
@@ -49,6 +49,8 @@ tests:
                 timeoutMS: 100
               expectError:
                 isClientError: true
+        expectError:
+          isClientError: true
     expectEvents:
       # The only operation run fails with a client-side error, so there should be no events for the client.
       - client: *client
@@ -88,9 +90,6 @@ tests:
     expectEvents:
       - client: *client
         events:
-          # Because the second insert expects an error and gets an error, it technically succeeds, so withTransaction
-          # will try to run commitTransaction. This will fail client-side, though, because the timeout has already
-          # expired, so no command is sent.
           - commandStartedEvent:
               commandName: insert
               databaseName: *databaseName
@@ -103,3 +102,9 @@ tests:
               command:
                 insert: *collectionName
                 maxTimeMS: { $$type: ["int", "long"] }
+          - commandStartedEvent:
+              commandName: abortTransaction
+              databaseName: admin
+              command:
+                abortTransaction: 1
+                maxTimeMS: { $$type: [ "int", "long" ] }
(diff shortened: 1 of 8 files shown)


source/client-side-operations-timeout/tests/convenient-transactions.yml
```diff
@@ -103,3 +100,11 @@ tests:
               command:
                 insert: *collectionName
                 maxTimeMS: { $$type: ["int", "long"] }
+          - commandStartedEvent:
```
rozza, line 103: As `withTransaction` expects an error, then an `abortTransaction` event will be issued.

source/client-side-operations-timeout/tests/convenient-transactions.yml
```diff
@@ -49,6 +49,8 @@ tests:
                 timeoutMS: 100
               expectError:
                 isClientError: true
+        expectError:
+          isClientError: true
```
ShaneHarvey, line 53: [In PyMongo](https://github.com/mongodb/mongo-python-driver/blob/f7a7b5a3329981cf39edfa4246578960796da825/test/unified_format.py#L1448-L1449) ([and Go](https://github.com/mongodb/mongo-go-driver/blob/f192906/mongo/integration/unified/operation.go#L32-L43)) the "expectError" on the insertOne above causes the test runner to catch, validate, and the ignore the exception so it never reaches the withTransaction block. So this new test would fail. Is it a bug that Java propagates the exception?

ShaneHarvey (changes requested):

source/client-side-operations-timeout/tests/convenient-transactions.yml
ShaneHarvey: Oh I see. This was recently changed in https://jira.mongodb.org/browse/DRIVERS-1709 but wasn't documented in the UTR spec itself. The only reference is in the jira ticket:
> Ensure that neither ignoreResultAndError nor expectError interfere with propagating exceptions from a callback operation to withTransaction. This came up with the [Java driver's POC](https://github.com/mongodb/specifications/pull/1505#discussion_r1488378728).

ShaneHarvey: Ah I was looking at the old RST version. I is documented in the markdown version: https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.md#withtransaction

rozza: To clarify was this change correct?

ShaneHarvey: Yes this change is correct given the new rules for handling errors in the UTR.

ShaneHarvey (approved):

PR #1556, opened 2024-03-25
DRIVERS-2875: add support for TopologyDescriptionChangedEvent to expectEvents

W-A-James:
- Updated unified test format spec to detail how assertions on `topologyDescriptionChangedEvents` should be done
- Updated schema to support this spec change
- Added valid pass test for addition to schema
- Bumped schema version in makefile for unified test runner

- [x] Update changelog.
  clusters, and serverless).

### DIFF
--- a/source/server-discovery-and-monitoring/server-discovery-and-monitoring-logging-and-monitoring.rst
+++ b/source/server-discovery-and-monitoring/server-discovery-and-monitoring-logging-and-monitoring.rst
@@ -111,6 +111,11 @@ Initial Server Description
 
 ``ServerDescription`` objects MUST be initialized with a default description in an “unknown” state, guaranteeing that the previous description in the events and log messages will never be null.
 
+
+Initial Topology Description
+----------------------------
+The first ``TopologyDescriptionChangedEvent`` to be emitted from a monitored Topology MUST set its ``previousDescription`` property to be a ``TopologyDescription`` object in the "unknown" state.
+
 ----------
 Events API
 ----------
@@ -698,6 +703,8 @@ See the `README <https://github.com/mongodb/specifications/server-discovery-and-
 Changelog
 =========
 
+:2024-03-29: Updated to clarify expected initial value of TopologyDescriptionChangedEvent's
+             previousDescription field
 :2024-01-04: Updated to clarify when ServerHeartbeatStartedEvent should be emitted
 :2023-03-31: Renamed to include "logging" in the title. Reorganized contents and made consistent with CLAM spec, and added requirements
              for SDAM log messages. 
--- a/source/server-discovery-and-monitoring/server-discovery-and-monitoring-logging-and-monitoring.rst
+++ b/source/server-discovery-and-monitoring/server-discovery-and-monitoring-logging-and-monitoring.rst
@@ -114,6 +114,7 @@ Initial Server Description
 
 Initial Topology Description
 ----------------------------
+
 The first ``TopologyDescriptionChangedEvent`` to be emitted from a monitored Topology MUST set its ``previousDescription`` property to be a ``TopologyDescription`` object in the "unknown" state.
 
 ----------
--- a/source/unified-test-format/tests/valid-pass/expectedEventsForClient-topologyDescriptionChangedEvent.yml
+++ b/source/unified-test-format/tests/valid-pass/expectedEventsForClient-topologyDescriptionChangedEvent.yml
@@ -34,8 +34,8 @@ tests:
         events:
           - topologyOpeningEvent: {}
           - topologyDescriptionChangedEvent: 
-            previousDescription:
-              type: "Unknown"
-            newDescription:
-              type: "ReplicaSetWithPrimary"
+              previousDescription:
+                type: "Unknown"
+              newDescription:
+                type: "ReplicaSetWithPrimary"
 
--- a/source/unified-test-format/tests/valid-pass/expectedEventsForClient-topologyDescriptionChangedEvent.yml
+++ b/source/unified-test-format/tests/valid-pass/expectedEventsForClient-topologyDescriptionChangedEvent.yml
@@ -1,0 +1,48 @@
+description: "expectedEventsForClient-topologyDescriptionChangedEvent.yml"
+
+schemaVersion: "1.20"
+
+runOnRequirements:
+  - topologies:
+      - replicaset
+    minServerVersion: "4.4" # awaitable hello
+
+tests:
+  - description: "can assert on values of newDescription and previousDescription fields"
+    operations:
+      - name: createEntities
+        object: testRunner
+        arguments:
+          entities:
+            - client:
+                id: &client client
+                observeEvents:
+                  - topologyDescriptionChangedEvent
+      # ensure the topology has been fully discovered before closing the client.
+      # expected events are initial server discovery and 3 server connect events.
+      - name: waitForEvent
+        object: testRunner
+        arguments:
+          client: *client
+          event:
+            topologyDescriptionChangedEvent: {}
+          count: 4
+      - name: close
+        object: *client
+    expectEvents:
+      - client: *client
+        eventType: sdam
+        ignoreExtraEvents: false
+        events:
+          - topologyOpeningEvent: {}
+          - topologyDescriptionChangedEvent: {} # unknown -> replset no primary
+          - topologyDescriptionChangedEvent: {} # server connected
+          - topologyDescriptionChangedEvent: {} # server connected
+          - topologyDescriptionChangedEvent: {} # server connected
+          - topologyDescriptionChangedEvent:  # replicaset -> unknown
+              previousDescription:
+                type: "ReplicaSetWithPrimary"
+              newDescription:
+                type: "Unknown"
+          - topologyClosedEvent: {}
+
--- a/source/unified-test-format/tests/valid-pass/expectedEventsForClient-topologyDescriptionChangedEvent.yml
+++ b/source/unified-test-format/tests/valid-pass/expectedEventsForClient-topologyDescriptionChangedEvent.yml
@@ -26,23 +26,16 @@ tests:
           client: *client
           event:
             topologyDescriptionChangedEvent: {}
-          count: 4
-      - name: close
-        object: *client
+          count: 1
     expectEvents:
       - client: *client
         eventType: sdam
-        ignoreExtraEvents: false
+        ignoreExtraEvents: true 
         events:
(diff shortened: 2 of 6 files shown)


W-A-James:
See changes made to the test runner implementation in [this PR](https://github.com/mongodb/node-mongodb-native/pull/3968)

source/unified-test-format/tests/valid-pass/expectedEventsForClient-topologyDescriptionChangedEvent.yml
```diff
@@ -0,0 +1,48 @@
...
+          count: 4
+      - name: close
+        object: *client
+    expectEvents:
+      - client: *client
+        eventType: sdam
+        ignoreExtraEvents: false
+        events:
+          - topologyOpeningEvent: {}
+          - topologyDescriptionChangedEvent: {} # unknown -> replset no primary
```
jmikola, line 38: Why doesn't this assert that the description goes from "Unknown" to "ReplicaSetWithPrimary"? If that's possible to assert, then you can potentially stop the entire test after the `waitForEvent` operation. Add `ignoreExtraEvents: true` and avoid closing the client entity.

In that case, you could avoid any assumptions about replica set size and subsequent event assertions. That might also make the test more resilient to variations in SDAM monitoring (not every driver uses awaitable `hello`).

---

I was initially curious why the TopologyDescription starts as "Unknown", but verified that locally with the PHP driver. 

I think it'd also be helpful to call out why the TopologyDescription starts as "Unknown".

[Initial TopologyType](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#initial-topologytype) in the SDAM spec suggests that it would start as "ReplicaSetNoPrimary" (if we assume the URI includes `replicaSet`).

When I ran this locally with PHP, I did observe an initial "Unknown" state for the previous TopologyDescription; however, I don't see where that's mandated in the spec. 
[Initial ServerDescription](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring-logging-and-monitoring.rst#initial-server-description) in the SDAM Logging/Monitoring spec notes why ServerDescription is initialized to "Unknown" (to avoid a `null` value in the previous description). I assume that same applies to TopologyDescription. Is that specified anywhere? If not, I'd propose reporting a ticket to get that clarified -- or working it into this ticket and having an SDAM spec owner review.

W-A-James: Good catch here on the Initial Topology type. I think the wording of the spec here is pretty clear as it relates to the Initial Topology type, I just happened to miss it when working on this, so I don't think we should assume it always starts as `UNKNOWN`.
The part that's not entirely clear is whether or not the `replicaSet` option should be required in the URI when running unified tests against replica sets. It seems that it should be when doing these kinds of spec tests given that, to my knowledge, pretty much all drivers make use of drivers-evergreen-tools scripts for their cluster setups in CI, but that's not specified anywhere. Do you think that's a change I should make here?

W-A-James: That being said, I'm in favour of changing that spec test here to expect this replica set to start with `ReplicaSetNoPrimary`.

jmikola: I'm not convinced that [Initial TopologyType](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#initial-topologytype) applies to SDAM monitoring events. If it does, then there is evidently an issue with libmongoc.

What did you observe in your driver?

> ...but that's not specified anywhere. Do you think that's a change I should make here?

I don't see much benefit in doing so, as it'd likely be ignored. There's presently nothing in the unified test format spec that talks about a well-formed URI, and that's not really in the scope of the test format anyway since it'd mainly impact other spec tests (excluding the "valid" test you're cooking up here).

I think it's more reasonable to just document the assumption in the test file here, such that it'd be easier for someone to diagnose a potential failure down the line.

W-A-James: Our driver has the `previousDescription` hardcoded to `UNKNOWN` on the first `TopologyDescriptionChangedEvent` to be emitted.

W-A-James: Why wouldn't the initial TopologyType apply to SDAM monitoring events though? Aren't the monitoring events supposed to give a view of the internal view of the Topology? If so, then it would make sense for them to set the `previousDescription` field of the first `TopologyDescriptionChangedEvent` to the internal view of the topology that we get from the connection string.

W-A-James: @jmikola Talked to Shane about it and seems it's just something we left off of the sdam monitoring spec. Will add that clarification in this PR.

source/unified-test-format/tests/valid-pass/expectedEventsForClient-topologyDescriptionChangedEvent.yml
```diff
@@ -0,0 +1,41 @@
...
+      - client: *client
+        eventType: sdam
+        ignoreExtraEvents: true 
+        events:
+          - topologyOpeningEvent: {}
+          - topologyDescriptionChangedEvent: 
+              previousDescription:
+                type: "Unknown"
+              newDescription:
+                type: "ReplicaSetWithPrimary"
```
ShaneHarvey, line 40: When I run this test in Python, this first event is ReplicaSetNoPrimary, not ReplicaSetWithPrimary:
```python
<TopologyDescriptionChangedEvent topology_id: 66104eba924530e675e287ae changed from:
  <TopologyDescription id: 66104eba924530e675e287ae, topology_type: Unknown, servers: []>, to:
  <TopologyDescription id: 66104eba924530e675e287ae, topology_type: ReplicaSetNoPrimary, servers: [<ServerDescription ('localhost', 27019) server_type: Unknown, rtt: None>]>>
```

We do this when creating the client because in this case we know ahead of time we are connecting to a replica set (because our test suite adds the replicaSet option.

ShaneHarvey (changes requested):

source/unified-test-format/tests/valid-pass/expectedEventsForClient-topologyDescriptionChangedEvent.yml
ShaneHarvey: How about we add `"directConnection": true` to the client and use `type: Single`?

ShaneHarvey (approved):
LGTM! Python tests are passing here: https://spruce.mongodb.com/version/66107586f9da6e00077e3ff1/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

source/server-discovery-and-monitoring/server-discovery-and-monitoring-logging-and-monitoring.rst
```diff
@@ -111,6 +111,11 @@ Initial Server Description
 
 ``ServerDescription`` objects MUST be initialized with a default description in an “unknown” state, guaranteeing that the previous description in the events and log messages will never be null.
 
+
+Initial Topology Description
+----------------------------
```
jmikola, line 116:
```suggestion
----------------------------

```

Consistency with other sections, which add a line break. This probably isn't something checked by CI.

source/unified-test-format/tests/valid-pass/expectedEventsForClient-topologyDescriptionChangedEvent.yml
```diff
@@ -0,0 +1,40 @@
...
+  - description: "can assert on values of newDescription and previousDescription fields"
+    operations:
+      - name: createEntities
+        object: testRunner
+        arguments:
+          entities:
+            - client:
+                id: &client client
+                uriOptions:
+                  directConnection: true
```
jmikola, line 20: Assuming the cluster is launched with mongo-orchestration, wouldn't we expect the connection string to include multiple seeds and a `replicaSet` option?

I expect the multiple seeds would conflict with `directConnection`, as mentioned in [directConnection URI option with multiple seeds or SRV URI](https://github.com/mongodb/specifications/blob/master/source/uri-options/uri-options.rst#directconnection-uri-option-with-multiple-seeds-or-srv-uri).

@ShaneHarvey how did this pass in Python?

ShaneHarvey: The python test suite helper to create a MongoClient always connects with a single seed and only adds the replicaSet option if directConnection is not given (or False): 
https://github.com/mongodb/mongo-python-driver/blob/e37394d4029d61f72a02dc465fbfc9afe35a5701/test/utils.py#L570-L579

jmikola: > always connects with a single seed and only adds the replicaSet option if directConnection is not given (or False)

This sounds like something that needs to be clarified in the Unified Test Format spec, when we talk about feeding the test runner with a connection string. I imagine these may be the first unified tests with `directConnection`, so this hasn't come up before.

ShaneHarvey (approved):

source/unified-test-format/tests/valid-pass/expectedEventsForClient-topologyDescriptionChangedEvent.yml
jmikola: > Why wouldn't the initial TopologyType apply to SDAM monitoring events though?

Closing the loop here. My point was that [Initial TopologyType](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst#initial-topologytype) specifies what the initial topology should be but makes no mention of the "Unknown" state that is used for the `previousDescription` field of the first emitted `TopologyDescriptionChangedEvent`. The rules for `previousDescription` seems to exist entirely within the SDAM Monitoring spec.

I wasn't debating whether that's correct or not; merely pointing out that the main SDAM spec doesn't address this.

jmikola:
Feel free to merge with @ShaneHarvey's review. I'd just like to see the spec link text changed.

PR #1557, opened 2024-03-25
DRIVERS-2789 Fix Math Rendering in Client Side Encryption Spec

blink1073:

- [ ] Update changelog.
  clusters, and serverless).

### DIFF
--- a/source/client-side-encryption/tests/README.md
+++ b/source/client-side-encryption/tests/README.md
@@ -64,7 +64,7 @@ The following matches a command_started_event for an insert of a document where 
     command_name: insert
 ```
 
-The values of $$type\` correspond to
+The values of `$$type` correspond to
 [these documented string representations of BSON types](https://www.mongodb.com/docs/manual/reference/bson-types/).
 
 Each YAML file has the following keys:
--- a/source/client-side-encryption/tests/README.md
+++ b/source/client-side-encryption/tests/README.md
@@ -32,8 +32,8 @@ The spec tests format is an extension of the
 - An `encrypted_fields` to set on the collection used for operations.
 - A `key_vault_data` of data that should be inserted in the key vault collection before each test.
 - Introduction `autoEncryptOpts` to `clientOptions`
-- Addition of $db`to command in`command_started_event\`
-- Addition of $$type\` to command_started_event and outcome.
+- Addition of `$db`to command in`command_started_event`
+- Addition of `$$type` to `command_started_event` and outcome.
 
 The semantics of $$type\` is that any actual value matching one of the types indicated by either a BSON type string or
 an array of BSON type strings is considered a match.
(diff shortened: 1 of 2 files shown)


W-A-James (changes requested):
Running the following command, there are also instances of this pattern in `source/client-side-encryption/tests/README.md`. It does seem that these predate the conversion to markdown, but this seems like a good place to fix them.

```bash
grep -n '`\$.+\$`' **/*.md
```
Running the following should fix these instances and might be useful going forward as we continue to convert more specs to markdown

```bash
sed -i --regexp-extended 's/`(\$.+\$)`/\1/' **/*.md # Removes wrapping backticks in all markdown files and writes results in place
```

blink1073:
@W-A-James what platform are you using?

When I run those commands I get no output from `grep`, and `sed: 1: "s/`(\$.+\$)`/\1/": \1 not defined in the RE` from the second command.

W-A-James:
@blink1073 I'm running on Ubuntu 20.04 through WSL2.

blink1073:
Thanks, I got `grep -E -n '`\$.+\$`' **/*.md` to work on MacOS, but I couldn't find an equivalent `sed` invocation that worked.

source/client-side-encryption/tests/README.md
```diff
@@ -32,10 +32,10 @@ The spec tests format is an extension of the
 - An `encrypted_fields` to set on the collection used for operations.
 - A `key_vault_data` of data that should be inserted in the key vault collection before each test.
 - Introduction `autoEncryptOpts` to `clientOptions`
-- Addition of `$db` to command in `command_started_event`
-- Addition of `$$type` to command_started_event and outcome.
+- Addition of $db`to command in`command_started_event\`
+- Addition of $$type\` to command_started_event and outcome.
```
W-A-James, line 36:
```suggestion
- Addition of `$db`to command in`command_started_event`
- Addition of `$$type` to `command_started_event` and outcome.
```

source/client-side-encryption/tests/README.md
```diff
@@ -32,10 +32,10 @@ The spec tests format is an extension of the
...
 
-The semantics of `$$type` is that any actual value matching one of the types indicated by either a BSON type string or
+The semantics of $$type\` is that any actual value matching one of the types indicated by either a BSON type string or
```
W-A-James, line 38:
```suggestion
The semantics of `$$type` is that any actual value matching one of the types indicated by either a BSON type string or
```

source/client-side-encryption/tests/README.md
```diff
@@ -64,7 +64,7 @@ The following matches a command_started_event for an insert of a document where
     command_name: insert
 ```
 
-The values of `$$type` correspond to
+The values of $$type\` correspond to
```
W-A-James, line 67:
```suggestion
The values of `$$type` correspond to
```

W-A-James (changes requested):
Almost done, just a couple unmatched backticks

W-A-James (approved):

PR #1559, opened 2024-04-01
DRIVERS-2789 Add Section Stubs to Converted RST Files

blink1073:

- [ ] Update changelog.
  clusters, and serverless).


ShaneHarvey (commented):
I just tried a few of the links in auth.rst and they all redirect to the auth.md file but none of them redirect to the intended section (they all link to the top of the md file). My opinion is that we shouldn't attempt to include anchor links at all.

blink1073:
> I just tried a few of the links in auth.rst and they all redirect to the auth.md file but none of them redirect to the intended section (they all link to the top of the md file). My opinion is that we shouldn't attempt to include anchor links at all.

Hmm yeah I think it is because the anchors aren't actually there when the page first loads.

PR #1560, opened 2024-04-01
DRIVERS-2877: Follow w: 0 write with a find in spec tests

jyemin:
This will force completion of the w: 0 write, which addresses a TCP-based race condition in the tests

- [N/A] Update changelog.
- [Done] Make sure there are generated JSON files from the YAML test files.
- [Done] Test changes in at least one language driver.
- [Done] Test these changes against all server versions and topologies (including standalone, replica set, sharded
  clusters, and serverless).

Note: there are other w: 0 writes in the unified tests but none that appear to be causing issues so leaving them alone for now.

### DIFF
--- a/source/command-logging-and-monitoring/tests/logging/unacknowledged-write.json
+++ b/source/command-logging-and-monitoring/tests/logging/unacknowledged-write.json
@@ -53,11 +53,27 @@
               "_id": 2
             }
           }
+        },
+        {
+          "name": "find",
+          "object": "collection",
+          "arguments": {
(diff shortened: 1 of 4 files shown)


durran (approved):
This looks good to me. Node patch is actually failing on the logging section as we have not implemented support for `ignoreExtraMessages`. https://spruce.mongodb.com/version/660d47b916df3f000765d833/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

PR #1569, opened 2024-04-24
DRIVERS-2836 OIDC: Clarify that TOKEN_RESOURCE must be url-encoded

blink1073:

- [x] Update changelog.
  clusters, and serverless).

### DIFF
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -1224,8 +1224,7 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
   - TOKEN_RESOURCE\
     The URI of the target resource. If `TOKEN_RESOURCE` is provided and `ENVIRONMENT` is not one of
     `["azure", "gcp"]` or `TOKEN_RESOURCE` is not provided and `ENVIRONMENT` is one of `["azure", "gcp"]`, the driver
-    MUST raise an error. Drivers MUST ensure that `TOKEN_RESOURCE` is url-encoded, while ensuring that it is not
-    double-encoded.
+    MUST raise an error.
 
   - OIDC_CALLBACK\
     An [OIDC Callback](#oidc-callback) that returns OIDC credentials. Drivers MAY allow the user to
@@ -1295,7 +1294,8 @@ Metadata: true
 
 where `<resource>` is the value of the `TOKEN_RESOURCE` mechanism property and `<client_id>` is the `username` from the
 connection string. If a `username` is not provided, the `client_id` query parameter should be omitted. The timeout
-should equal the `callbackTimeoutMS` parameter given to the callback.
+should equal the `callbackTimeoutMS` parameter given to the callback. Drivers MUST ensure that `TOKEN_RESOURCE` is
+url-encoded, while ensuring that it is not double-encoded.
 
 Example code for the above using curl, where `$TOKEN_RESOURCE` is the value of the `TOKEN_RESOURCE` mechanism property.
 
@@ -1354,7 +1354,8 @@ Metadata-Flavor: Google
 ```
 
 where `<resource>` is the value of the `TOKEN_RESOURCE` mechanism property. The timeout should equal the
-`callbackTimeoutMS` parameter given to the callback.
+`callbackTimeoutMS` parameter given to the callback. Drivers MUST ensure that `TOKEN_RESOURCE` is url-encoded, while
+ensuring that it is not double-encoded.
 
 Example code for the above using curl, where `$TOKEN_RESOURCE` is the value of the `TOKEN_RESOURCE` mechanism property.
 
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -1202,7 +1202,7 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
 #### [MongoCredential](#mongocredential) Properties
 
 > [!NOTE]
-> Drivers MUST not url-decode the entire `authMechanismProperties` given in an connection string when the
+> Drivers MUST NOT url-decode the entire `authMechanismProperties` given in an connection string when the
 > `authMechanism` is `MONGODB-OIDC`. This is because the `TOKEN_RESOURCE` itself will typically be a URL containing a
 > `:` character. The values of the individual `authMechanismProperties` MUST still be url-decoded when given as part of
 > the connection string, and MUST NOT be url-decoded when given as part of a `MongoClient` property.
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -1224,7 +1224,8 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
   - TOKEN_RESOURCE\
     The URI of the target resource. If `TOKEN_RESOURCE` is provided and `ENVIRONMENT` is not one of
     `["azure", "gcp"]` or `TOKEN_RESOURCE` is not provided and `ENVIRONMENT` is one of `["azure", "gcp"]`, the driver
-    MUST raise an error.
+    MUST raise an error. Drivers MUST ensure that `TOKEN_RESOURCE` is url-encoded, such as by using a regex for special
+    characters or by checking if unencoding the value changes it, and then encoding if necessary.
 
   - OIDC_CALLBACK\
     An [OIDC Callback](#oidc-callback) that returns OIDC credentials. Drivers MAY allow the user to
@@ -2043,6 +2044,8 @@ to EC2 instance metadata in ECS, for security reasons, Amazon states it's best p
 [IAM Roles for Tasks](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html))
 
 ## Changelog
+
+- 2024-04-24: Clarify that TOKEN_RESOURCE for MONGODB-OIDC must be url-encoded.
 
 - 2024-04-22: Fix API description for GCP built-in OIDC provider.
 
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -1208,7 +1208,7 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
 > of the connection string, and MUST NOT be url-decoded when not given as part of the connection string, such as through
 > a `MongoClient` or `Credential` property. Drivers MUST parse the `TOKEN_RESOURCE` by splitting only on the first `:`
 > character. Drivers MUST document that users must url-encode `TOKEN_RESOURCE` when it is provided in the connection
-> string and it contains and of the special characters in \[`,`, `+`, `&`\].
+> string and it contains and of the special characters in \[`,`, `+`, `&`, `%`\].
 
 - username\
   MAY be specified. Its meaning varies depending on the OIDC provider integration used.
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -1203,9 +1203,12 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
 
 > [!NOTE]
 > Drivers MUST NOT url-decode the entire `authMechanismProperties` given in an connection string when the
-> `authMechanism` is `MONGODB-OIDC`. This is because the `TOKEN_RESOURCE` itself will typically be a URL containing a
-> `:` character. The values of the individual `authMechanismProperties` MUST still be url-decoded when given as part of
-> the connection string, and MUST NOT be url-decoded when not given as part of the connection string, such as through a `MongoClient` or `Credential` property.
+> `authMechanism` is `MONGODB-OIDC`. This is because the `TOKEN_RESOURCE` itself will typically be a URL and may contain
+> a `,` character. The values of the individual `authMechanismProperties` MUST still be url-decoded when given as part
+> of the connection string, and MUST NOT be url-decoded when not given as part of the connection string, such as through
+> a `MongoClient` or `Credential` property. Drivers MUST parse the `TOKEN_RESOURCE` by splitting only on the first `:`
+> character. Drivers MUST document that users must url-encode `TOKEN_RESOURCE` when it is provided in the connection
+> string and it contains and of the special characters in \[`,`, `+`, `&`\].
 
 - username\
   MAY be specified. Its meaning varies depending on the OIDC provider integration used.
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -1224,8 +1224,8 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
   - TOKEN_RESOURCE\
     The URI of the target resource. If `TOKEN_RESOURCE` is provided and `ENVIRONMENT` is not one of
     `["azure", "gcp"]` or `TOKEN_RESOURCE` is not provided and `ENVIRONMENT` is one of `["azure", "gcp"]`, the driver
-    MUST raise an error. Drivers MUST ensure that `TOKEN_RESOURCE` is url-encoded, such as by using a regex for special
-    characters or by checking if unencoding the value changes it, and then encoding if necessary.
+    MUST raise an error. Drivers MUST ensure that `TOKEN_RESOURCE` is url-encoded, while ensuring that it is not
+    double-encoded.
 
   - OIDC_CALLBACK\
     An [OIDC Callback](#oidc-callback) that returns OIDC credentials. Drivers MAY allow the user to
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -1205,7 +1205,7 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
 > Drivers MUST NOT url-decode the entire `authMechanismProperties` given in an connection string when the
 > `authMechanism` is `MONGODB-OIDC`. This is because the `TOKEN_RESOURCE` itself will typically be a URL containing a
 > `:` character. The values of the individual `authMechanismProperties` MUST still be url-decoded when given as part of
-> the connection string, and MUST NOT be url-decoded when given as part of a `MongoClient` property.
+> the connection string, and MUST NOT be url-decoded when not given as part of the connection string, such as through a `MongoClient` or `Credential` property.
 
 - username\
   MAY be specified. Its meaning varies depending on the OIDC provider integration used.
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -1201,6 +1201,12 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
 
 #### [MongoCredential](#mongocredential) Properties
 
+> [!NOTE]
+> Drivers MUST not url-decode the entire `authMechanismProperties` given in an connection string when the
+> `authMechanism` is `MONGODB-OIDC`. This is because the `TOKEN_RESOURCE` itself will typically be a URL containing a
+> `:` character. The values of the individual `authMechanismProperties` MUST still be url-decoded when given as part of
+> the connection string, and MUST NOT be url-decoded when given as part of a `MongoClient` property.
+
 - username\
   MAY be specified. Its meaning varies depending on the OIDC provider integration used.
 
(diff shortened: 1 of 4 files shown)


blink1073:
Tested with https://github.com/mongodb/mongo-python-driver/pull/1616

source/auth/auth.md
```diff
@@ -1224,7 +1224,8 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
   - TOKEN_RESOURCE\
     The URI of the target resource. If `TOKEN_RESOURCE` is provided and `ENVIRONMENT` is not one of
     `["azure", "gcp"]` or `TOKEN_RESOURCE` is not provided and `ENVIRONMENT` is one of `["azure", "gcp"]`, the driver
-    MUST raise an error.
+    MUST raise an error. Drivers MUST ensure that `TOKEN_RESOURCE` is url-encoded, such as by using a regex for special
+    characters or by checking if unencoding the value changes it, and then encoding if necessary.
```
katcharov, line 1228: I suspect the intent is to say "the value of TOKEN_RESOURCE must be url-DEcoded" (with no mention of checks; and we must then document that the value should be ENcoded by the user). 

If the intent is to tell drivers to, in tests, encode the environment variable before supplying it to tests, I think that the value should come pre-encoded out of any scripts that produce it.

A longer explanation to make sure I'm on the same page: supposing the following URLs start with `mongodb://localhost/?authMechanism=MONGODB-OIDC`, there should be no difference between these URLS:
```
&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE:foo:bar

&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE%3Afoo%3Abar
```
That is, both should decode to an equivalent `authMechanismProperties` value, with no error. In Java, there was a check against a second `:` that I [removed](https://github.com/mongodb/mongo-java-driver/pull/1371#discussion_r1578189423), because it did not appear to violate any spec - it is easy enough to split on only the first `:`.

However, if we need to support other characters in TOKEN_RESOURCE, (for example, we need to allow `,`, `&`, `%`...), then we need to specify that the value of TOKEN_RESOURCE must be decoded by the driver (independently of the url-decoding already being done at the url params level). For example, the value `foo:bar,comma` would need to be encoded, and the following two would be equivalent:

```
&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE:foo%3Abar%2Cbaz

&authMechanismProperties=ENVIRONMENT%3Aazure%2CTOKEN_RESOURCE%3Afoo%253Abar%252Cbaz
```
Note that there is a `:` in `TOKEN_RESOURCE:foo`, in contrast to the original two examples.

source/auth/tests/legacy/connection-string.json
```diff
@@ -569,6 +569,36 @@
...
+      "uri": "mongodb://user@localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE:mongodb%253A//test-cluster",
+      "valid": true,
+      "credential": {
+        "username": "user",
+        "password": null,
+        "source": "$external",
+        "mechanism": "MONGODB-OIDC",
+        "mechanism_properties": {
+          "ENVIRONMENT": "azure",
+          "TOKEN_RESOURCE": "mongodb%253A//test-cluster"
```
katcharov, line 583: This appears to be double-encoded - I do not think that `mongodb%3A//test-cluster` is the intended final value of the resource. It is fine that this is testing a partial encoding, where unlike in `mongodb%3A%2F%2Ftest-cluster`, the `//` is not encoded but the `:` is, but a single decode would be:
```suggestion
          "TOKEN_RESOURCE": "mongodb%3A//test-cluster"
```

source/auth/auth.md
```diff
@@ -2044,6 +2045,8 @@ to EC2 instance metadata in ECS, for security reasons, Amazon states it's best p
 
 ## Changelog
 
+- 2024-04-24: Clarify that TOKEN_RESOURCE for MONGODB-OIDC must be url-encoded.
```
katcharov, line 2048: This reminds me, should we url-encode the resource when sending to GCP? For example, I currently have,

```
    public static CredentialInfo fetchGcpCredentialInfo(final String resource) {
        String endpoint = "http://metadata/computeMetadata/v1/instance/service-accounts/default/identity?audience="
                + resource;
```
but I think we want to specify that we should wrap `resource` in `urlEncode(resource)`?

katcharov: It occurs to me that if we are not doing this, then users can just pass in url-encoded values to TOKEN_RESOURCE, which would then remain unprocessed, and appended here - so no spec changes would be needed?

katcharov (changes requested):

source/auth/tests/legacy/connection-string.json
katcharov: It seems users MUST encode property values, and then MUST encode at least the TOKEN_RESOURCE sub-value. This is unusual and unintuitive, because let’s say the user wants to use an already-url-encoded value, like `something?hello%20world` for the resource. Great - just pass it in as a url-encoded value `something%3Fhello%2520world` right after TOKEN_RESOURCE:
```
&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE:something%3Fhello%2520world
```
But then, it gets decoded once by the parameter decoder:
```
ENVIRONMENT:azure,TOKEN_RESOURCE:something?hello%20world
```
And then again by the TOKEN_RESOURCE value decoder:
```
something?hello world
```
The `%20` was lost. The only solutions seem to be to a) document that users MUST url-encode parameters when providing a url-encoded value, and we can’t just say “double encode the resource value”, and all of this is confusing, or b) to not encode TOKEN_RESOURCE using url-encoding, but some other encoding that won’t be decoded by the param decoder.

source/auth/auth.md
sanych-sun: The idea was to specify what value we are expecting there. If it should be plain value - then driver should urlEncode it while building the GCP or Azure callback Url. If we expect encoded value - then we have to validate if it was encoded properly and just concatenate it without url encoding. But @blink1073 decided to have encoding auto-detection functionality which might be useful, but looks confusing at least for me.

blink1073: My point was - users would most likely *have* to encode when using a connection string, but they won't have to when using MongoClient input.  I'm saying use whatever mechanism makes sense to ensure it is url-encoded without double-encoding.

sanych-sun: If we just let the value though it might be a problem, for example if user set `TOKEN_RESOURCE = "token&another-magic-parameter=42"`.

blink1073: Then that is a malformed connection string

source/auth/auth.md
```diff
@@ -1224,7 +1224,8 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
   - TOKEN_RESOURCE\
     The URI of the target resource. If `TOKEN_RESOURCE` is provided and `ENVIRONMENT` is not one of
     `["azure", "gcp"]` or `TOKEN_RESOURCE` is not provided and `ENVIRONMENT` is one of `["azure", "gcp"]`, the driver
-    MUST raise an error.
+    MUST raise an error. Drivers MUST ensure that `TOKEN_RESOURCE` is url-encoded, such as by using a regex for special
```
sanych-sun, line 1227: I would prefer to have plain value here, because if some driver decide to use Azure/GCP/other SDK to pass the value - most likely SDK will do encoding on it's end. The fact that the value is being used as a url query parameter - this is implementation details and might be vary (if some other provider will use POST and pass the value in request body).

sanych-sun: But then in the Azure and GCP section we have to require driver to encode the value as it being used as a query parameter.

sanych-sun (changes requested):

source/auth/auth.md
blink1073: I updated it to be more clear but less prescriptive on the how.

source/auth/auth.md
blink1073: Done

source/auth/auth.md
sanych-sun: If you creating mongo client from connection string - yes. But if you creating mongo client by constructing it from properties dictionary. Should the value be encoded or not?

blink1073: I think we should allow either, so it is easier to copy-paste between contexts (like from a connection string to a dictionary).

source/auth/tests/legacy/connection-string.json
```diff
@@ -569,6 +569,36 @@
...
+      "uri": "mongodb://user@localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE:a$b",
+      "valid": true,
+      "credential": {
+        "username": "user",
+        "password": null,
+        "source": "$external",
+        "mechanism": "MONGODB-OIDC",
+        "mechanism_properties": {
+          "ENVIRONMENT": "azure",
+          "TOKEN_RESOURCE": "a%24b"
```
sanych-sun, line 598: Should we expect `a$b` here?

blink1073: No, because it is `azure` and we would have encoded it.

source/auth/tests/legacy/connection-string.yml
```diff
@@ -412,6 +412,28 @@ tests:
     mechanism_properties:
       ENVIRONMENT: azure
       TOKEN_RESOURCE: foo
+- description: should accept a url-encoded TOKEN_RESOURCE (MONGODB-OIDC)
+  uri: mongodb://user@localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE:mongodb%3A//test-cluster
```
katcharov, line 416: Connection strings can have options, and options have values. The option `authMechanismProperties` has a "[value that represents one or more key and value pairs](https://github.com/mongodb/specifications/blob/44a7ecfc6782b247ad75dcfd17c6bb4f76214f24/source/connection-string/connection-string-spec.md?plain=1#L218C20-L218C76)". That spec says that if a key contains a `,` or a `:`, it must be url-encoded (so should `%` and `&`,  though these are omitted). Here is an example where I url-encode the `TOKEN_RESOURCE` value `abc,def:ghi`, as per the connection string spec, which I understand this PR mirrors:
```
ConnectionString cs = new ConnectionString(
        "mongodb://localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties="
                + "ENVIRONMENT:azure,TOKEN_RESOURCE:abc%2Cdef%3Aghi");
assertEquals("abc,def:ghi", cs.getCredential().getMechanismProperty("TOKEN_RESOURCE", null));
Expected :abc,def:ghi
Actual   :abc
```
The result is incorrect. When I place the same uri into a legacy `connection-string.json` test, it also fails, producing `abc`. As per "the values in connection options MUST be URL decoded by the parser" at the top of that section in the spec, Java decodes the entire value of the `authMechanismProperties` option, before parsing the name-value properties (the values are not url-decoded a second time). The spec lists Java as a reference implementation, along with Python, which I expect does the same.

This is an abstract problem in our connection string spec, but a concrete problem here because the token resource value appears to allow arbitrary characters. Documenting that the user must url-encode both the entire `authMechanismProperties` value, and the token resource (which ends up double-encoded) seems untenable. The only other (marginally less-bad) option I can think of is to specify that the `TOKEN_RESOURCE` value must always be base64 encoded - we would base64-decode, and then url-encode when sending to GCP.

```suggestion
  uri: mongodb://user@localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE:abc%2Cdef%3Aghi
```

blink1073: I think it is a mistake to decode the full `authMechanismProperties`.  It has a defined structure, split on `:` and `,`.  Then within that you have your keys and values.

sanych-sun: We have the same behavior as @katcharov described here on C# driver. Will check how complex would be to change it.

katcharov: I suspect the intent was to mirror how URL properties behave, and those can always be url-encoded (and it's a reasonable practice to do so), so users may have an expectation that our connection strings work the same way. This clearly gets in the way of using certain values. 

However, changing this behaviour would break backward compatibility. For example, users might be encoding `readPreferenceTags=dc%3Any%2Crack%3A1` as per our spec, with the expectation that this would decode to `dc:ny,rack:1`.

We discussed a bit in Java, and our proposed plan is to disable full-decoding of `authMechanismProperties` only when `authMechanism=MONGODB-OIDC`, and in just that case decode individual values. This is in line with your proposal, but avoids breaking backward compatibility. I believe this should be clarified in the spec.

The rejected alternatives were to A) forbid `,` in `TOKEN_RESOURCE` (no, because this can be a connection string), or B) base32 encode the value (contrary to the Python release), or C) break spec.

source/auth/tests/legacy/connection-string.yml
```diff
@@ -412,6 +412,28 @@ tests:
...
+  uri: mongodb://user@localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE:a$b
+  valid: true
+  credential:
+    username: user
+    password: null
+    source: $external
+    mechanism: MONGODB-OIDC
+    mechanism_properties:
+      ENVIRONMENT: azure
+      TOKEN_RESOURCE: a%24b
```
katcharov, line 436: This should represent the unencoded raw value. For example, when the user sets the property via:
```
credential = credential.withMechanismProperty("TOKEN_RESOURCE", 
    credential.getMechanismProperty("TOKEN_RESOURCE"));
```
The value must not change.

The user should only have to url-encode the TOKEN_RESOURCE value when it is in the connection string, and contains certain reserved characters. If "$" is one of those characters, then the above uri is invalid, and we should throw an error (replace it with `,`, `&`, etc. to see why).

Whatever raw value is given, it of course must be url-encoded (just once, of course) when building the GCP or Azure URIs.

source/auth/auth.md
katcharov: In other comments I argued that it is impossible to detect whether a string is encoded or not, so I believe the only viable option is to expect only raw unencoded values when constructing "in code" (via dictionary, etc.) rather than from the connection string.

(Others, feel free to resolve this convo, might be clearer to finish remaining discussions elsewhere.)

source/auth/tests/legacy/connection-string.yml
```diff
@@ -412,6 +412,28 @@ tests:
...
+  uri: mongodb://user@localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE:mongodb%3A//test-cluster
+  valid: true
+  credential:
+    username: user
+    password: null
+    source: $external
+    mechanism: MONGODB-OIDC
+    mechanism_properties:
+      ENVIRONMENT: azure
+      TOKEN_RESOURCE: mongodb%3A//test-cluster
```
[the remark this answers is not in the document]
katcharov: Additionally, perhaps a test somewhere that fails when ALLOWED_HOSTS is provided when ENVIRONMENT is set?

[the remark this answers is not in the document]
blink1073: We can't test ALLOWED_HOSTS here because it is only valid when a human callback is passed, which we can't handle with these tests.

blink1073:
Updated tests in https://github.com/mongodb/mongo-python-driver/pull/1620

source/auth/tests/legacy/connection-string.yml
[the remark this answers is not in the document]
katcharov: I mean, do we have a test (anywhere) that would check for a failure when env is azure (or gcp), but allowed_hosts is also provided?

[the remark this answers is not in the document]
blink1073: No, that will have to be a follow-on ticket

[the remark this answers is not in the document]
blink1073: (after GA)

sanych-sun (commented):
The latest revision looks good to me. Will implement the changes and verify if tests are working as expected for CSHARP Driver. If everything works well - will LGTM it later today.

source/auth/tests/legacy/connection-string.yml
```diff
@@ -412,6 +412,50 @@ tests:
...
+  credential:
+    username: user
+    password: null
+    source: $external
+    mechanism: MONGODB-OIDC
+    mechanism_properties:
+      ENVIRONMENT: azure
+      TOKEN_RESOURCE: 'mongodb://test-cluster'
+- description: should accept an un-encoded TOKEN_RESOURCE (MONGODB-OIDC)
+  uri: mongodb://user@localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE:mongodb://test-cluster
```
sanych-sun, line 427: Is this Uri valid? `TOKEN_RESOURCE:mongodb://test-cluster` Should we have `:` in the `mongodb://test-cluster` encoded as `%3A` (not sure about slash symbol though)?

katcharov: This usage is against the advice we give to users in the spec, but I think we should optimistically decode this (like [this in Java](https://github.com/mongodb/mongo-java-driver/pull/1371#discussion_r1572939075)), rather than failing. Although it is indeed ambiguous, I think we will never put a `:` in the name of any property (precisely because it is a separator).

sanych-sun: Yep, can do the same, just want to double check if we see this as expected behavior.

source/auth/auth.md
```diff
@@ -1201,6 +1201,12 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
 
 #### [MongoCredential](#mongocredential) Properties
 
+> [!NOTE]
+> Drivers MUST not url-decode the entire `authMechanismProperties` given in an connection string when the
```
alcaeus, line 1205:
```suggestion
> Drivers MUST NOT url-decode the entire `authMechanismProperties` given in an connection string when the
```

alcaeus (commented):
One minor note, LGTM other than that.

sanych-sun (approved):
LGTM

source/auth/tests/legacy/connection-string.json
```diff
@@ -569,6 +569,66 @@
...
+        "mechanism": "MONGODB-OIDC",
+        "mechanism_properties": {
+          "ENVIRONMENT": "azure",
+          "TOKEN_RESOURCE": "mongodb://test-cluster"
+        }
+      }
+    },
+    {
+      "description": "should accept an un-encoded TOKEN_RESOURCE (MONGODB-OIDC)",
+      "uri": "mongodb://user@localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE:mongodb://test-cluster",
```
katcharov, line 589: We should specify that drivers MUST ("optimistically"?) parse by splitting on only the first `:`.

Consider adding somewhere something like "Drivers MUST document that users must percent-encode the token resource when it is provided in the connection string". I am unsure if we usually specify what should be documented, versus behaviour, but in this case the two are different.

source/auth/auth.md
```diff
@@ -1201,6 +1201,12 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
 
 #### [MongoCredential](#mongocredential) Properties
 
+> [!NOTE]
+> Drivers MUST NOT url-decode the entire `authMechanismProperties` given in an connection string when the
+> `authMechanism` is `MONGODB-OIDC`. This is because the `TOKEN_RESOURCE` itself will typically be a URL containing a
+> `:` character. The values of the individual `authMechanismProperties` MUST still be url-decoded when given as part of
```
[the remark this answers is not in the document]
blink1073: Clarified

source/auth/tests/legacy/connection-string.json
blink1073: Updated

source/auth/auth.md
```diff
@@ -1203,9 +1203,12 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
...
 > Drivers MUST NOT url-decode the entire `authMechanismProperties` given in an connection string when the
-> `authMechanism` is `MONGODB-OIDC`. This is because the `TOKEN_RESOURCE` itself will typically be a URL containing a
-> `:` character. The values of the individual `authMechanismProperties` MUST still be url-decoded when given as part of
-> the connection string, and MUST NOT be url-decoded when given as part of a `MongoClient` property.
+> `authMechanism` is `MONGODB-OIDC`. This is because the `TOKEN_RESOURCE` itself will typically be a URL and may contain
+> a `,` character. The values of the individual `authMechanismProperties` MUST still be url-decoded when given as part
+> of the connection string, and MUST NOT be url-decoded when not given as part of the connection string, such as through
+> a `MongoClient` or `Credential` property. Drivers MUST parse the `TOKEN_RESOURCE` by splitting only on the first `:`
+> character. Drivers MUST document that users must url-encode `TOKEN_RESOURCE` when it is provided in the connection
+> string and it contains and of the special characters in \[`,`, `+`, `&`\].
```
katcharov, line 1211: I think this list must include `%` and might also include the chars `;?`(perhaps others?).
```suggestion
> string and it contains and of the special characters in \[`,`, `+`, `&`, `%`\].
```

(I read this requirement as providing a minimal list, so drivers can document "always encode", and it would meet this requirement.)

katcharov (approved):
LGTM, with one comment

PR #1572, opened 2024-04-30
DRIVERS-2789 Convert OP_MSG to Markdown

blink1073:

- [x] Update changelog.
  clusters, and serverless).


NoahStapp (approved):
Can we just remove the `.rst` files entirely? We've already had several run-ins with people modifying the now-unused versions of files.

blink1073:
We're leaving just this note in each, so that links won't go dead but at least you get a breadcrumb:

```
.. note::
  This specification has been converted to Markdown and renamed to
  `OP_MSG.md <OP_MSG.md>`_.  
```

PR #1574, opened 2024-05-01
DRIVERS-2511: Update maxWireVersion from 6 to 21

jyemin:
Once a driver removes support for MongoDB 3.6 and updates its minWireVersion from 6 to 7, these tests start to fail due to the wire version compatibility check required by the test runner.

Bumping from 6 to 21 (MongoDB 7.0 maxWireVersion) should hold us for a while.

Note: I don't think the intention of any of these tests were to check the minWireVersion of the driver, as they have been at 6 long before 6 was the minWireVersion.

- [ N/A] Update changelog.
- [ Done] Make sure there are generated JSON files from the YAML test files.
- [ Done (Java)] Test changes in at least one language driver.
- [ N/A] Test these changes against all server versions and topologies (including standalone, replica set, sharded
  clusters, and serverless).

### DIFF
--- a/source/resources
+++ b/source/resources
@@ -1,0 +1,1 @@
+/Users/jeff.yemin/git/mongo-java-driver/driver-core/src/test/resources
--- a/source/resources
+++ b/source/resources
@@ -1,1 +1,0 @@
-/Users/jeff.yemin/git/mongo-java-driver/driver-core/src/test/resources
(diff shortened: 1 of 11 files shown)


source/resources
W-A-James, line 1: Should this have been deleted?

jyemin: Ooh, I accidentally committed the symbolic link to my mongo-java-driver clone.  Let me see if I can revert that in a new commit or have to force push.

jyemin: Done

W-A-James (approved):

PR #1588, opened 2024-05-29
DRIVERS-2915 OIDC: Disallow comma character in authMechanismProperties connection string value

blink1073:

- [x] Update changelog.
  clusters, and serverless).

### DIFF
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -1201,11 +1201,6 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
 
 #### [MongoCredential](#mongocredential) Properties
 
-> [!NOTE]
-> Drivers MUST parse the `TOKEN_RESOURCE` by splitting only on the first `:` character. Drivers MUST document that users
-> must specify `TOKEN_RESOURCE` as part of a `MongoClient` option it contains a comma `,` character, since it would
-> interfere with the parsing rules for `authMechanismProperties`.
-
 - username\
   MAY be specified. Its meaning varies depending on the OIDC provider integration used.
 
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -1202,13 +1202,9 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
 #### [MongoCredential](#mongocredential) Properties
 
 > [!NOTE]
-> Drivers MUST NOT url-decode the entire `authMechanismProperties` given in an connection string when the
-> `authMechanism` is `MONGODB-OIDC`. This is because the `TOKEN_RESOURCE` itself will typically be a URL and may contain
-> a `,` character. The values of the individual `authMechanismProperties` MUST still be url-decoded when given as part
-> of the connection string, and MUST NOT be url-decoded when not given as part of the connection string, such as through
-> a `MongoClient` or `Credential` property. Drivers MUST parse the `TOKEN_RESOURCE` by splitting only on the first `:`
-> character. Drivers MUST document that users must url-encode `TOKEN_RESOURCE` when it is provided in the connection
-> string and it contains and of the special characters in \[`,`, `+`, `&`, `%`\].
+> Drivers MUST parse the `TOKEN_RESOURCE` by splitting only on the first `:` character. Drivers MUST document that users
+> must use `TOKEN_RESOURCE` as part of a `MongoClient` option it contains a comma `,` character, since it would
+> interfere with the parsing rules for `authMechanismProperties`.
 
 - username\
   MAY be specified. Its meaning varies depending on the OIDC provider integration used.
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -2045,6 +2045,8 @@ to EC2 instance metadata in ECS, for security reasons, Amazon states it's best p
 
 ## Changelog
 
+- 2024-05-29: Disallow comma character when `TOKEN_RESOURCE` is given in a connection string.
+
 - 2024-05-03: Clarify timeout behavior for OIDC machine callback. Add `serverless:forbid` to OIDC unified tests. Add an
   additional prose test for the behavior of `ALLOWED_HOSTS`.
 
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -1203,7 +1203,7 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
 
 > [!NOTE]
 > Drivers MUST parse the `TOKEN_RESOURCE` by splitting only on the first `:` character. Drivers MUST document that users
-> must use `TOKEN_RESOURCE` as part of a `MongoClient` option it contains a comma `,` character, since it would
+> must specify `TOKEN_RESOURCE` as part of a `MongoClient` option it contains a comma `,` character, since it would
 > interfere with the parsing rules for `authMechanismProperties`.
 
 - username\
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -1224,7 +1224,9 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
   - TOKEN_RESOURCE\
     The URI of the target resource. If `TOKEN_RESOURCE` is provided and `ENVIRONMENT` is not one of
     `["azure", "gcp"]` or `TOKEN_RESOURCE` is not provided and `ENVIRONMENT` is one of `["azure", "gcp"]`, the driver
-    MUST raise an error.
+    MUST raise an error. Note: because the `TOKEN_RESOURCE` is often itself a URL, drivers MUST document that a
+    `TOKEN_RESOURCE` with a comma `,` must be given as a `MongoClient` configuration and not as part of the connection
+    string, and that the `TOKEN_RESOURCE` value can contain a colon `:` character.
 
   - OIDC_CALLBACK\
     An [OIDC Callback](#oidc-callback) that returns OIDC credentials. Drivers MAY allow the user to
--- a/source/auth/tests/legacy/connection-string.json
+++ b/source/auth/tests/legacy/connection-string.json
@@ -602,7 +602,7 @@
     {
       "description": "should handle a complicated url-encoded TOKEN_RESOURCE (MONGODB-OIDC)",
       "uri": "mongodb://user@localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE:abcd%25ef%3Ag%26hi",
-      "valid": false,
+      "valid": true,
       "credential": {
         "username": "user",
         "password": null,
--- a/source/auth/tests/legacy/connection-string.json
+++ b/source/auth/tests/legacy/connection-string.json
@@ -600,21 +600,6 @@
       }
     },
     {
-      "description": "should handle a complicated url-encoded TOKEN_RESOURCE (MONGODB-OIDC)",
-      "uri": "mongodb://user@localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE:abc%2Cd%25ef%3Ag%26hi",
-      "valid": true,
-      "credential": {
-        "username": "user",
-        "password": null,
-        "source": "$external",
-        "mechanism": "MONGODB-OIDC",
-        "mechanism_properties": {
-          "ENVIRONMENT": "azure",
-          "TOKEN_RESOURCE": "abc,d%ef:g&hi"
-        }
-      }
-    },
-    {
       "description": "should url-encode a TOKEN_RESOURCE (MONGODB-OIDC)",
       "uri": "mongodb://user@localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE:a$b",
       "valid": true,
--- a/source/auth/tests/legacy/connection-string.json
+++ b/source/auth/tests/legacy/connection-string.json
@@ -600,6 +600,21 @@
       }
     },
     {
+      "description": "should handle a complicated url-encoded TOKEN_RESOURCE (MONGODB-OIDC)",
+      "uri": "mongodb://user@localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE:abcd%25ef%3Ag%26hi",
+      "valid": false,
+      "credential": {
+        "username": "user",
+        "password": null,
+        "source": "$external",
+        "mechanism": "MONGODB-OIDC",
+        "mechanism_properties": {
+          "ENVIRONMENT": "azure",
+          "TOKEN_RESOURCE": "abcd%ef:g&hi"
+        }
+      }
+    },
+    {
       "description": "should url-encode a TOKEN_RESOURCE (MONGODB-OIDC)",
       "uri": "mongodb://user@localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE:a$b",
       "valid": true,
--- a/source/auth/tests/legacy/connection-string.yml
+++ b/source/auth/tests/legacy/connection-string.yml
@@ -434,6 +434,17 @@ tests:
     mechanism_properties:
       ENVIRONMENT: azure
       TOKEN_RESOURCE: 'mongodb://test-cluster'
+- description: should handle a complicated url-encoded TOKEN_RESOURCE (MONGODB-OIDC)
+  uri: mongodb://user@localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE:abcd%25ef%3Ag%26hi
+  valid: true
+  credential:
+    username: user
+    password: null
+    source: $external
+    mechanism: MONGODB-OIDC
+    mechanism_properties:
+      ENVIRONMENT: azure
+      TOKEN_RESOURCE: 'abcd%ef:g&hi'
 - description: should url-encode a TOKEN_RESOURCE (MONGODB-OIDC)
   uri: mongodb://user@localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE:a$b
   valid: true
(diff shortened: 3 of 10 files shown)


source/auth/tests/legacy/connection-string.yml
```diff
@@ -434,17 +434,6 @@ tests:
...
-  uri: mongodb://user@localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE:abc%2Cd%25ef%3Ag%26hi
-  valid: true
-  credential:
-    username: user
-    password: null
-    source: $external
-    mechanism: MONGODB-OIDC
-    mechanism_properties:
-      ENVIRONMENT: azure
-      TOKEN_RESOURCE: 'abc,d%ef:g&hi'
```
addaleax, line 447: Do we want to keep this test? Making sure that percent-encoded characters are properly decoded and that `:` is kept in the value seems valuable in general

addaleax (approved):

source/auth/tests/legacy/connection-string.yml
blink1073: So just remove the comma then?

addaleax: Yeah, I think so?

source/auth/auth.md
```diff
@@ -1202,13 +1202,9 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
...
 > [!NOTE]
-> Drivers MUST NOT url-decode the entire `authMechanismProperties` given in an connection string when the
-> `authMechanism` is `MONGODB-OIDC`. This is because the `TOKEN_RESOURCE` itself will typically be a URL and may contain
-> a `,` character. The values of the individual `authMechanismProperties` MUST still be url-decoded when given as part
-> of the connection string, and MUST NOT be url-decoded when not given as part of the connection string, such as through
-> a `MongoClient` or `Credential` property. Drivers MUST parse the `TOKEN_RESOURCE` by splitting only on the first `:`
-> character. Drivers MUST document that users must url-encode `TOKEN_RESOURCE` when it is provided in the connection
-> string and it contains and of the special characters in \[`,`, `+`, `&`, `%`\].
+> Drivers MUST parse the `TOKEN_RESOURCE` by splitting only on the first `:` character. Drivers MUST document that users
+> must use `TOKEN_RESOURCE` as part of a `MongoClient` option it contains a comma `,` character, since it would
```
jyemin, line 1206:
```suggestion
> must specify `TOKEN_RESOURCE` as part of a `MongoClient` option it contains a comma `,` character, since it would
```

source/auth/auth.md
```diff
@@ -1202,13 +1202,9 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
...
 
 > [!NOTE]
-> Drivers MUST NOT url-decode the entire `authMechanismProperties` given in an connection string when the
-> `authMechanism` is `MONGODB-OIDC`. This is because the `TOKEN_RESOURCE` itself will typically be a URL and may contain
-> a `,` character. The values of the individual `authMechanismProperties` MUST still be url-decoded when given as part
-> of the connection string, and MUST NOT be url-decoded when not given as part of the connection string, such as through
-> a `MongoClient` or `Credential` property. Drivers MUST parse the `TOKEN_RESOURCE` by splitting only on the first `:`
-> character. Drivers MUST document that users must url-encode `TOKEN_RESOURCE` when it is provided in the connection
-> string and it contains and of the special characters in \[`,`, `+`, `&`, `%`\].
+> Drivers MUST parse the `TOKEN_RESOURCE` by splitting only on the first `:` character. Drivers MUST document that users
```
jyemin, line 1205: > Drivers MUST parse the `TOKEN_RESOURCE` by splitting only on the first `:` character

I'm unclear why TOKEN_RESOURCE gets special mention.  Does this require logic in parsers that is different for TOKEN_RESOURCE than any other auth mechanism property?

jyemin (changes requested):

source/auth/auth.md
katcharov: This should be part of the connection string spec (I believe it's ambiguous there).

source/auth/tests/legacy/connection-string.json
```diff
@@ -601,16 +601,16 @@
     },
     {
       "description": "should handle a complicated url-encoded TOKEN_RESOURCE (MONGODB-OIDC)",
-      "uri": "mongodb://user@localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE:abc%2Cd%25ef%3Ag%26hi",
-      "valid": true,
+      "uri": "mongodb://user@localhost/?authMechanism=MONGODB-OIDC&authMechanismProperties=ENVIRONMENT:azure,TOKEN_RESOURCE:abcd%25ef%3Ag%26hi",
+      "valid": false,
```
durran, line 605:
```suggestion
      "valid": true,
```

durran (changes requested):
Just the one JSON change.

source/auth/auth.md
blink1073: I've moved the explanation to the connection string spec.

addaleax (approved):

source/connection-string/connection-string-spec.md
```diff
@@ -216,8 +216,11 @@ The values in connection options MUST be URL decoded by the parser. The values c
   ```
 
 - Key value pairs: A value that represents one or more key and value pairs. Multiple key value pairs are delimited by a
-  comma (","). The key is everything up to the first colon sign (":") and the value is everything afterwards. If any
-  keys or values containing a comma (",") or a colon (":") they must be URL encoded. For example:
+  comma (","). The key is everything up to the first colon sign (":") and the value is everything afterwards. Drivers
+  MUST handle subsequent colon signs (":") within the value, unless otherwise specified in this document.\
+  If any keys
+  or values contain a comma (",") they MUST not be provided as part of the connection string, since it would interfere
+  with parsing. For example:
```
katcharov, line 223:
```suggestion
  with parsing. Key value pair example:
```
Since the example is not related to the preceding discussion of rules regarding `:` and `,`. (Alternatively, make the example demonstrate the rules.)

durran (approved):

source/connection-string/connection-string-spec.md
```diff
@@ -216,8 +216,11 @@ The values in connection options MUST be URL decoded by the parser. The values c
   ```
 
 - Key value pairs: A value that represents one or more key and value pairs. Multiple key value pairs are delimited by a
-  comma (","). The key is everything up to the first colon sign (":") and the value is everything afterwards. If any
-  keys or values containing a comma (",") or a colon (":") they must be URL encoded. For example:
+  comma (","). The key is everything up to the first colon sign (":") and the value is everything afterwards. Drivers
+  MUST handle subsequent colon signs (":") within the value, unless otherwise specified in this document.\
```
jyemin, line 220: > unless otherwise specified in this document

Is it otherwise specified anywhere?  If not, can we remove this part of the sentence?

source/connection-string/connection-string-spec.md
jyemin, line 223: With the addition of this sentence, the following example becomes confusing because the reader expects to see an example with a subsequent colon.  How about we provide two examples:

1. Keep this example, but put it directly after the sentence: `The key is everything up to the first colon sign (":") and the value is everything afterwards`
2. Add another example with a subsequent colon, e.g.

```
?authMechanismProperties=TOKEN_RESOURCE:mongodb://...
```

source/connection-string/connection-string-spec.md
```diff
@@ -216,8 +216,11 @@ The values in connection options MUST be URL decoded by the parser. The values c
   ```
 
 - Key value pairs: A value that represents one or more key and value pairs. Multiple key value pairs are delimited by a
-  comma (","). The key is everything up to the first colon sign (":") and the value is everything afterwards. If any
-  keys or values containing a comma (",") or a colon (":") they must be URL encoded. For example:
+  comma (","). The key is everything up to the first colon sign (":") and the value is everything afterwards. Drivers
```
jyemin, line 219: >  Drivers MUST handle subsequent colon signs (":") within the value

Can we add a spec test for this?

jyemin (changes requested):

source/auth/auth.md
```diff
@@ -1201,15 +1201,6 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
...
 #### [MongoCredential](#mongocredential) Properties
 
-> [!NOTE]
-> Drivers MUST NOT url-decode the entire `authMechanismProperties` given in an connection string when the
-> `authMechanism` is `MONGODB-OIDC`. This is because the `TOKEN_RESOURCE` itself will typically be a URL and may contain
-> a `,` character. The values of the individual `authMechanismProperties` MUST still be url-decoded when given as part
-> of the connection string, and MUST NOT be url-decoded when not given as part of the connection string, such as through
-> a `MongoClient` or `Credential` property. Drivers MUST parse the `TOKEN_RESOURCE` by splitting only on the first `:`
-> character. Drivers MUST document that users must url-encode `TOKEN_RESOURCE` when it is provided in the connection
-> string and it contains and of the special characters in \[`,`, `+`, `&`, `%`\].
```
katcharov, line 1211: I think we should still specify that this should be documented. Note that certain other characters still affect parsing of TOKEN_RESOURCE. It would be helpful to write the doc requirement in a way that can be copy-pasted into individual drivers:

> Drivers MUST document that: If the value of TOKEN_RESOURCE contains the characters \[`,`, `+`, `&`, `%`\], then it must not be specified in the connection string, but must instead be specified via a `MongoClient` option.

We can also say the following to allow more characters in the connection string, but it seems more convoluted:

> Drivers MUST document that: If the value of TOKEN_RESOURCE contains the characters \[`+`, `&`, `%`\], then it must be percent-encoded when specified in the connection string. If the value of TOKEN_RESOURCE contains a comma \[`,`\], then it must not be specified in the connection string, but must instead be specified via a `MongoClient` option.

katcharov (changes requested):

source/auth/auth.md
blink1073: I don't think we need to call out TOKEN_RESOURCE as special, but keep the logic in the connection string file.

source/connection-string/connection-string-spec.md
blink1073: Removed

source/connection-string/connection-string-spec.md
blink1073: Done

source/connection-string/connection-string-spec.md
blink1073: Updated

source/auth/auth.md
katcharov: I do not mean the parsing logic, referred to in an [earlier comment](https://github.com/mongodb/specifications/pull/1588#discussion_r1619009956), which I agree need not be specified individually.

I mean that we previously specified that drivers must document this non-obvious behaviour in user-facing documentation, in at least one location. We required this on TOKEN_RESOURCE because it was apparently the first case where it became relevant. Drivers should be as careful in how they document behaviour contracts, as they are in the implementation, since breaking the contract is what will be perceived by users as breaking compatibility.

I do think that this should be documented individually on each key-value-pair affected by this rule (especially since I believe TOKEN_RESOURCE is the only one affected).

source/connection-string/connection-string-spec.md
```diff
@@ -216,13 +216,24 @@ The values in connection options MUST be URL decoded by the parser. The values c
   ```
 
 - Key value pairs: A value that represents one or more key and value pairs. Multiple key value pairs are delimited by a
-  comma (","). The key is everything up to the first colon sign (":") and the value is everything afterwards. If any
-  keys or values containing a comma (",") or a colon (":") they must be URL encoded. For example:
+  comma (","). The key is everything up to the first colon sign (":") and the value is everything afterwards. Reserved
+  characters such as ':' must be escaped according to [RFC 2396](https://www.rfc-editor.org/rfc/rfc2396)
```
katcharov, line 220: I'm not sure that we need to change anything in this spec. The second (or third...) colon must already be treated as part of the value, since the spec already says:

> up to the first colon sign and the value is everything afterwards

The original wording is problematic because it confuses restrictions on users (doc requirements) with driver requirements, and does not specify enough restricted characters, but I don't think we should correct this in this PR, but as part of future work to clarify this spec. 

Likewise, I don't think we should perform other clarifications in this PR, since the formulation is not obvious. For example, I don't think we should refer to the URI RFC, and should instead just say "percent-decoded" (we do not have URI components since connection strings are not URIs, and even if we did, we would still need to, according to the RFC, define the "set of characters actually reserved" within each particular component; note also that `%` is not in the list in section 2.2, though obviously it needs to be encoded by users).

source/connection-string/connection-string-spec.md
```diff
@@ -216,13 +216,24 @@ The values in connection options MUST be URL decoded by the parser. The values c
...
+  For example:
 
   ```
   ?readPreferenceTags=dc:ny,rack:1
   ```
 
+  Drivers MUST handle unencoded colon signs (":") within the value. For example:
+
+  ```
+  ?authMechanismProperties=TOKEN_RESOURCE:mongodb%3A%2F%2F...
```
katcharov, line 231: Wouldn't "unencoded" be:
```suggestion
  ?authMechanismProperties=TOKEN_RESOURCE:mongodb://...
```
But, I think this should just be a test. In general, drivers should handle all unreserved characters that are unencoded, in any context, just because the spec does not reserve them, and because percent-decoding does not alter them. This appears to be an existing consequence of how the spec is written.

katcharov (changes requested):

source/connection-string/tests/valid-options.yml
```diff
@@ -28,3 +28,17 @@ tests:
         auth: ~
         options:
               tls: true
+    -
+        description: Colon in a key value pair
+        uri: mongodb://example.com?authMechanismProperties=TOKEN_RESOURCE:mongodb%3A%2F%2Ftest-cluster
```
katcharov, line 33:
```suggestion
        uri: mongodb://example.com?authMechanismProperties=TOKEN_RESOURCE:mongodb://test-cluster
```

blink1073: Ah, this was failing in the python driver, but I see it is because we have a custom check there for `:`, let me try without.

source/connection-string/connection-string-spec.md
```diff
@@ -216,13 +216,24 @@ The values in connection options MUST be URL decoded by the parser. The values c
...
+  ```
+
+  If any keys or values contain a comma (",") they MUST not be provided as part of the connection string, since it would
+  interfere with parsing.
```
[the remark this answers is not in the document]
blink1073: Updated

source/connection-string/connection-string-spec.md
blink1073: Updated

source/connection-string/connection-string-spec.md
blink1073: Clarified

source/connection-string/connection-string-spec.md
```diff
@@ -216,13 +216,25 @@ The values in connection options MUST be URL decoded by the parser. The values c
...
+  Drivers MUST handle unencoded colon signs (":") within the value. For example, given the encoded input:
+
+  ```
+  ?authMechanismProperties=TOKEN_RESOURCE%3Amongodb%3A%2F%2Ffoo
+  ```
+
+  the driver MUST interpret the key as `TOKEN_RESOURCE` and the value as `mongodb://foo`.
+
+  Drivers MUST document that if keys or values contain a comma (",") they MUST not be provided as part of the connection
+  string, since it would interfere with parsing.
```
katcharov, line 236:
```suggestion
  For any option key-value pair that may contain a comma (such as `TOKEN_RESOURCE`), drivers MUST document that: a value containing a comma (",") MUST NOT be provided as part of the connection
  string. This prevents use of values that would interfere with parsing.
```
1. Clarifies that we do not want the "since" clause to be part of the user-facing docs.
2. Are users allowed to specify arbitrary keys? I thought we controlled them all.
3. A general comment seems like it would be confusing. We use the term "value" to refer (at least) to both: option values, and option key-value pair values. Option values can have commas, but I'm not sure how to express the distinction in user docs without using a convoluted phrase like "option key-value pair values".

It also seems fine (preferable?) to not add this new paragraph/requirement to the connection string spec at this time, since it is already quite nicely phrased in the auth spec for `TOKEN_RESOURCE` in particular.

source/connection-string/connection-string-spec.md
```diff
@@ -216,13 +216,25 @@ The values in connection options MUST be URL decoded by the parser. The values c
   ```
 
 - Key value pairs: A value that represents one or more key and value pairs. Multiple key value pairs are delimited by a
-  comma (","). The key is everything up to the first colon sign (":") and the value is everything afterwards. If any
-  keys or values containing a comma (",") or a colon (":") they must be URL encoded. For example:
+  comma (","). The key is everything up to the first colon sign (":") and the value is everything afterwards.
+
+  For example:
```
katcharov, line 221: This is a nice clarification of the example 👍

source/connection-string/tests/valid-warnings.yml
```diff
@@ -73,3 +73,15 @@ tests:
                 port: ~
         auth: ~
         options: ~
+    -
+        description: Comma in a key value pair causes a warning
+        uri: mongodb://example.com?authMechanismProperties=TOKEN_RESOURCE:mongodb://host1%2Chost2
```
katcharov, line 78: Is this really a warning? If we are decoding the entire option value, which is the present (though perhaps undocumented) behaviour for at least some drivers, then we get:

`TOKEN_RESOURCE:mongodb://host1,host2`

Which would contain two properties: `TOKEN_RESOURCE`, and `host2`, with the latter not having a `:`, or value assigned, which seems like it would be an error.

katcharov (changes requested):

source/connection-string/tests/valid-warnings.yml
blink1073: I based this on "Any invalid Values for a given key MUST be ignored and MUST log a WARN level message"

katcharov (approved):
LGTM!

jyemin (approved):

JamesKovacs (approved):
LGTM

PR #1589, opened 2024-05-29
DRIVERS-2789 Convert Extended JSON spec to markdown

blink1073:

- [x] Update changelog.
  clusters, and serverless).


jyemin (approved):
Quick look at formatting, I don't see any problems.

Relying on automation to ensure that the contents are semantically equivalent.

LGTM

PR #1590, opened 2024-05-30
DRIVERS-2870 Fix content of retryable-writes-test readme

blink1073:

- [x] Update changelog.
  clusters, and serverless).

### DIFF
--- a/source/retryable-writes/tests/README.md
+++ b/source/retryable-writes/tests/README.md
@@ -167,9 +167,9 @@ This test MUST:
 - only run against replica sets as mongos does not propagate the NoWritesPerformed label to the drivers.
 - be run against server versions 6.0 and above.
 
-Additionally, this test requires drivers to set a fail point after an `insertOne` operation but before the
-subsequent retry. Drivers that are unable to set a failCommand after the CommandSucceededEvent SHOULD use mocking or
-write a unit test to cover the same sequence of events.
+Additionally, this test requires drivers to set a fail point after an `insertOne` operation but before the subsequent
+retry. Drivers that are unable to set a failCommand after the CommandSucceededEvent SHOULD use mocking or write a unit
+test to cover the same sequence of events.
 
 1. Create a client with `retryWrites=true`.
 
@@ -255,6 +255,37 @@ This test MUST be executed against a sharded cluster that has at least two mongo
 6. Assert that two failed command events occurred. Assert that the failed command events occurred on different mongoses.
 
 7. Disable the fail points on both `s0` and `s1`.
+
+### 5. Test that in a sharded cluster on the same mongos if no other is available
+
+This test MUST be executed against a sharded cluster
+
+1. Ensure that a test is run against a sharded cluster. If there are multiple mongoses in the cluster, pick one to test
+   against.
+
+2. Create a client that connects to the mongos using the direct connection, and configure the following fail point on
+   the mongos::
+
+```javascript
+{
+    configureFailPoint: "failCommand",
+    mode: { times: 1 },
+    data: {
+        failCommands: ["insert"],
+        errorCode: 6,
+        errorLabels: ["RetryableWriteError"],
+        closeConnection: true
+    }
+}
+```
+
+3. Create a client with `retryWrites=true` that connects to the cluster, providing the selected mongos as the seed.
+
+4. Enable command monitoring, and execute a write command that is supposed to fail.
+
+5. Asserts that there was a failed command and a successful command event.
+
+6. Disable the fail point.
 
 ## Changelog
 
--- a/source/retryable-writes/tests/README.md
+++ b/source/retryable-writes/tests/README.md
@@ -256,15 +256,17 @@ This test MUST be executed against a sharded cluster that has at least two mongo
 
 7. Disable the fail points on both `s0` and `s1`.
 
-### 5. Test that in a sharded cluster on the same mongos if no other is available
+### 5. Test that in a sharded cluster writes are retried on the same mongos when no others are available.
 
-This test MUST be executed against a sharded cluster
+This test MUST be executed against a sharded cluster that supports `retryWrites=true`, has enabled the
+`configureFailPoint` command, and supports the `errorLabels` field (MongoDB 4.3.1+).
 
-1. Ensure that a test is run against a sharded cluster. If there are multiple mongoses in the cluster, pick one to test
-   against.
+Note: this test cannot reliably distinguish "retry on a different mongos due to server deprioritization" (the behavior
+intended to be tested) from "retry on a different mongos due to normal SDAM behavior of randomized suitable server
+selection". Verify relevant code paths are correctly executed by the tests using external means such as a logging,
+debugger, code coverage tool, etc.
 
-2. Create a client that connects to the mongos using the direct connection, and configure the following fail point on
-   the mongos:
+1. Create a client `s0` that connects to a single mongos from the cluster.
 
    ```javascript
    {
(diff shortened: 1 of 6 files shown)


source/retryable-writes/tests/README.md
```diff
@@ -71,7 +71,7 @@ Drivers should also assert that command documents are properly constructed with
 on whether the write operation is supported.
 [Command Logging and Monitoring](../../command-logging-and-monitoring/command-logging-and-monitoring.rst) may be used to
 check for the presence of a `txnNumber` field in the command document. Note that command documents may always include an
-`lsid` field per the [Driver Session](../../sessions/driver-sessions.md) specification.
+`lsid` field per the [Driver Session](../../sessions/driver-sessions.rst) specification.
```
jmikola, line 74: The session spec was converted in https://github.com/mongodb/specifications/commit/c6349d199588fa6cd9ea3e304366f90268ea3091. Shouldn't this remain as `.md`?

source/retryable-writes/tests/README.md
```diff
@@ -106,17 +106,197 @@ Drivers should test that transactions IDs are always included in commands for su
 
 The following tests ensure that retryable writes work properly with replica sets and sharded clusters.
 
-1. Test that retryable writes raise an exception when using the MMAPv1 storage engine. For this test, execute a write
-   operation, such as `insertOne`, which should generate an exception. Assert that the error message is the replacement
-   error message:
+- Test that retryable writes raise an exception when using the MMAPv1 storage engine. For this test, execute a write
```
jmikola, line 109: This should be an ordered list so the prose tests remain numbered.

Alternatively, it may be more readable to use header tags with manual numbering:

```
### 1. Retryable writes error when using MMAPv1 storage engine
```

That approach is more common in test READMEs with more than just a couple of prose tests, and this file now has five. If you go this route I'd just suggest shortening the name a bit (or leaving the original text as-is and adding your own header above).

Bonus: if you go this route, the `[!NOTE]` syntax below will start working because it will no longer be nested within a list item.

source/retryable-writes/tests/README.md
```diff
@@ -106,17 +106,197 @@ Drivers should test that transactions IDs are always included in commands for su
...
 
-   and the error code is 20.
+  and the error code is 20.
 
-   [!NOTE]
-   storage engine in use MAY skip this test for sharded clusters, since `mongos` does not report this information in its
-   `serverStatus` response.
+  > [!NOTE]
+  > Drivers that rely on `serverStatus` to determine the storage engine in use MAY skip this test for sharded clusters,
+  > since `mongos` does not report this information in its `serverStatus` response.
```
jmikola, line 122: These admonitions don't render within other elements, so I think this would be better written as a basic paragraph with no "Note" lead-in. Rendering issues aside, I think that makes more sense since the paragraph is providing direction for the test implementation by use of "MAY".

See [this discussion](https://github.com/orgs/community/discussions/16925#discussioncomment-9156825) and/or [this test file](https://github.com/sinsukehlab/NOTE-test/blob/main/NOTE.md) for more info on rendering issues.

source/retryable-writes/tests/README.md
```diff
@@ -106,17 +106,197 @@ Drivers should test that transactions IDs are always included in commands for su
...
+
+- Test that drivers properly retry after encountering PoolClearedErrors. This test MUST be implemented by any driver
+  that implements the CMAP specification.
+
+  This test requires MongoDB 4.3.4+ for both the `errorLabels` and `blockConnection` fail point options.
+
+  - Create a client with maxPoolSize=1 and retryWrites=true. If testing against a sharded deployment, be sure to connect
```
jmikola, line 129: These steps were also numeric in the original content. I think it's worth preserving those, as some drivers may refer back to numbered steps from comments within their prose tests.

source/retryable-writes/tests/README.md
blink1073: Fixed

source/retryable-writes/tests/README.md
blink1073: Done

source/retryable-writes/tests/README.md
blink1073: They're left-aligned now.

source/retryable-writes/tests/README.md
blink1073: I had been fighting with the formatter, fixed now.

blink1073:
Argh, there is something about this file that mdformat doesn't like, I'll keep trying...

```
Error: Could not format
"/home/runner/work/specifications/specifications/source/retryable-writes/tests/README.md".

The formatted Markdown renders to different HTML than the input Markdown. This
is likely a bug in mdformat. Please create an issue report here, including the
input Markdown: [https://github.com/executablebooks/mdformat/issues](https://github.com/executablebooks/mdformat/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc)
```

blink1073:
Fixed!

blink1073:
The JSON failure is unrelated (and incorrect).  The problem was that the admonition needed to be alone on its line...

stIncMale:
I am no longer an SDAM spec maintainer. I have removed myself from https://github.com/orgs/mongodb/teams/dbx-spec-owners-sdam, it wasn't updated when the maintainers changed.

source/retryable-writes/tests/README.md
```diff
@@ -106,17 +104,199 @@ Drivers should test that transactions IDs are always included in commands for su
...
+This test MUST be implemented by any driver that implements the CMAP specification.
+
+This test requires MongoDB 4.3.4+ for both the `errorLabels` and `blockConnection` fail point options.
+
+1. Create a client with maxPoolSize=1 and retryWrites=true. If testing against a sharded deployment, be sure to connect
+   to only a single mongos.
+
+2. Enable the following failpoint:
+
+   ```javascript
```
jmikola, line 134: The content within list elements was not properly indented. This applied both to code blocks and subsequent paragraphs.

source/retryable-writes/tests/README.md
```diff
@@ -106,17 +104,199 @@ Drivers should test that transactions IDs are always included in commands for su
...
+
+9. Disable the failpoint.
+
+### 3. Test that drivers return the original error after encountering a WriteConcernError with a RetryableWriteError label.
+
+This test MUST:
+
+- be implemented by any driver that implements the Command Monitoring specification,
+- only run against replica sets as mongos does not propagate the NoWritesPerformed label to the drivers.
+- be run against server versions 6.0 and above.
```
jmikola, line 168: There was no need for these to be numbered. Also, the previous numbering continued in the steps that followed, which was incorrect.

I changed this to an unordered list and restarted the numbering for steps at one.

Also removed the code block fence around the subsequent paragraph.

source/retryable-writes/tests/README.md
```diff
@@ -106,17 +104,199 @@ Drivers should test that transactions IDs are always included in commands for su
...
+   `s1`.
+
+4. Enable failed command event monitoring for `client`.
+
+5. Execute an `insert` command with `client`. Assert that the command failed.
+
+6. Assert that two failed command events occurred. Assert that the failed command events occurred on different mongoses.
+
+7. Disable the fail points on both `s0` and `s1`.
+
```
jmikola, line 258: You're missing the fifth prose test here:

> 5. Test that in a sharded cluster writes are retried on the same mongos when no others are available.

Note that its second step includes a code block, so please make sure that gets indented.

blink1073: Fixed

source/retryable-writes/tests/README.md
```diff
@@ -106,17 +104,230 @@ Drivers should test that transactions IDs are always included in commands for su
...
+       configureFailPoint: "failCommand",
+       mode: { times: 1 },
+       data: {
+           failCommands: ["insert"],
+           errorCode: 6,
+           errorLabels: ["RetryableWriteError"],
+           closeConnection: true
+       }
+   }
+   ```
```
jmikola, line 280: I just fixed the indentation here but realized that the content doesn't match the last version of the RST file that was removed in https://github.com/mongodb/specifications/commit/d3a355c299d9840bf9a758910a1739fcbc33ec64#diff-974028a6b5a7e6c54bd76884a6212e54ecc1d5c325b7a2186d5847310a97657cL311.

blink1073: updated

jmikola (approved):
LGTM with my last commit to add a missing step to the fifth prose test.

blink1073:
Thanks for the help!

PR #1595, opened 2024-06-05
DRIVERS-2921: Use single mongos for unacknowledged write tests

jmikola:
https://jira.mongodb.org/browse/DRIVERS-2921

cd08b7248eb4e3ab901a8f7905874da41b908f7c added find operations after a w:0 write to ensure it applies, but the client also needs useMultipleMongoses:false to ensure this happens on the same host.

- [ ] Update changelog.
  clusters, and serverless).

----

Patch build: https://spruce.mongodb.com/version/666093b08fe04300077b53b7/tasks

### DIFF
--- a/source/command-logging-and-monitoring/tests/logging/unacknowledged-write.yml
+++ b/source/command-logging-and-monitoring/tests/logging/unacknowledged-write.yml
@@ -5,6 +5,7 @@ schemaVersion: "1.16"
 createEntities:
   - client:
       id: &client client
+      useMultipleMongoses: false
       observeLogMessages:
         command: debug
   - database:
--- a/source/command-logging-and-monitoring/tests/monitoring/unacknowledgedBulkWrite.yml
+++ b/source/command-logging-and-monitoring/tests/monitoring/unacknowledgedBulkWrite.yml
@@ -5,6 +5,7 @@ schemaVersion: "1.7"
 createEntities:
   - client:
       id: &client client
+      useMultipleMongoses: false
       observeEvents:
         - commandStartedEvent
         - commandSucceededEvent
--- a/source/command-logging-and-monitoring/tests/logging/unacknowledged-write.json
+++ b/source/command-logging-and-monitoring/tests/logging/unacknowledged-write.json
@@ -5,6 +5,7 @@
     {
       "client": {
         "id": "client",
+        "useMultipleMongoses": false,
         "observeLogMessages": {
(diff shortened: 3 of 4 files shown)


source/command-logging-and-monitoring/tests/monitoring/unacknowledgedBulkWrite.yml
```diff
@@ -5,6 +5,7 @@ schemaVersion: "1.7"
 createEntities:
   - client:
       id: &client client
+      useMultipleMongoses: false
```
jmikola, line 8: Without this option, we cannot assume whether the client is connected to one or more mongoses. There is also no guarantee for the internal client, per [Initializing the Test Runner]https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.md#initializing-the-test-runner). The spec permits creating a single internal client for all hosts (assuming the driver can still target each host independently), or multiple single-client hosts.

source/command-logging-and-monitoring/tests/logging/unacknowledged-write.yml
```diff
@@ -5,6 +5,7 @@ schemaVersion: "1.16"
 createEntities:
   - client:
       id: &client client
+      useMultipleMongoses: false
```
jmikola, line 8: Note: PHP doesn't yet run logging tests, so I'd defer to another driver to test this.

jyemin (approved):
Thanks for catching this.

I missed it because the Java driver connects to a single mongos when `useMultipleMongoses` is unspecified.  But the unified test runner spec says that a driver shouldn't do that, so this change LGTM.

jmikola:
Note: the [unacknowledged-client-bulkWrite.yml](https://github.com/mongodb/specifications/blob/master/source/command-logging-and-monitoring/tests/monitoring/unacknowledged-client-bulkWrite.yml) file from https://github.com/mongodb/specifications/commit/10919c92cb784e9730551841e57541bee2e1af93 ([DRIVERS-716](https://jira.mongodb.org/browse/DRIVERS-716)) is likely affected by the same issue, but never implemented the original work-around from [DRIVERS-2877](https://jira.mongodb.org/browse/DRIVERS-2877). I reported that to @isabelatkinson and she'll likely address it in a follow-up PR and apply it directly to DRIVERS-716 since only Rust and libmongoc have started work on that.

PR #1597, opened 2024-06-05
fix uri in username with test environment oidc test

durran:

When implementing OIDC in the Ruby driver I noticed that the username with `ENVIRONMENT:test` spec test was not failing because of the username being present but rather the URI was actually invalid, since `ENVIRONMENT:test` was not in the `authMechanismProperties`. This fixes that test.

Tests running on this PR: https://github.com/mongodb/mongo-ruby-driver/pull/2873

- [ ] Update changelog.
  clusters, and serverless).


blink1073 (approved):
Thanks!

comandeo-mongo (approved):

PR #1587, opened 2024-05-22
DRIVERS-2767 QE Range Protocol v2

adriandole:
Range preview was replaced with range in server 8.0. This updates JSON tests, removes references to "rangePreview", and documents the new `trimFactor` option for range queries.

Verification with C driver: https://spruce.mongodb.com/version/665e1d1d4e6c750007215866

PR #1601, opened 2024-06-13
DRIVERS-2928 document "range" as unstable

kevinAlbs:
# Summary

Document "range" as unstable.

# Background & Motivation
This is a follow-up to https://github.com/mongodb/specifications/pull/1587. Some drivers have implemented "range" support for DRIVERS-2767. There is pending work that may impact drivers (DRIVERS-2927). This is intended to unblock driver releases in the short term. Driver releases may be needed soon to comply with SSDLC.

- [x] Update changelog.
- ~~[ ] Make sure there are generated JSON files from the YAML test files.~~ N/A.
- ~~[ ] Test changes in at least one language driver.~~ N/A.
  clusters, and serverless).~~ N/A.

### DIFF
--- a/source/client-side-encryption/client-side-encryption.md
+++ b/source/client-side-encryption/client-side-encryption.md
@@ -983,6 +983,7 @@ class ClientEncryption {
    //   {$and: [{$gt: [<fieldpath>, <value1>]}, {$lt: [<fieldpath>, <value2>]}]
    // $gt may also be $gte. $lt may also be $lte.
    // Only supported when queryType is "range" and algorithm is "Range".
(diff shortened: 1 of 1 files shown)


ShaneHarvey:
Is this accurate? I thought the whole point was that we were replacing the unstable "rangePreview" with the stable "range".

kevinAlbs:
> Is this accurate? I thought the whole point was that we were replacing the unstable "rangePreview" with the stable "range".

"range" is intended to be stable, but is not yet. There is pending work that may impact drivers (DRIVERS-2927) and libmongocrypt (SERVER-91077). This is intended to unblock driver releases in the short term that have already implemented "range". Driver releases may be needed soon to comply with SSDLC.

ShaneHarvey (approved):
Thanks, LGTM!

PR #1604, opened 2024-06-27
Remove result expectation in command monitoring tests

alcaeus:
- [ ] ~Update changelog.~
  clusters, and serverless).

This removes a command expectation for a find command in the command monitoring tests. We've observed frequent failures on sharded clusters, which we assume are related to replication lag. Since the goal of the test is to test command monitoring for writes with `w: 0` and not the successful execution of such writes, I assume it's safe to remove the result expectation to reduce flakiness.


blink1073 (approved):
LGTM!

PR #1594, opened 2024-06-05
DRIVERS-2870 Clean up rendering

blink1073:
Please complete the following before merging:

- [ ] Update changelog.
- [ ] Make sure there are generated JSON files from the YAML test files.
- [ ] Test changes in at least one language driver.
- [ ] Test these changes against all server versions and topologies (including standalone, replica set, sharded
  clusters, and serverless).

PR #1608, opened 2024-07-03
fix: changelog rendering in server-discovery-and-monitoring-logging-and-monitoring.md

nbbeeken:

- [ ] Update changelog.
  clusters, and serverless).

Fix some git diff left in the markdown


W-A-James (changes requested):
Lint failure.

Run
```bash
pre-commit run --show-diff-on-failure --color=always --all-files
```

and commit the changes

ShaneHarvey:
This is already fixed in https://github.com/mongodb/specifications/pull/1594 which looks like it stalled. Let's see if #1594 can be merged instead.

PR #1628, opened 2024-08-08
DRIVERS-2922: Allow valid SRV hostnames with fewer than 3 parts

aditi-khare-mongoDB:
### Description
SRVs are no longer required to have a given number of parts, URI terminology is more clearly defined, and stricter parent match requirements are introduced.

#### What is changing?

1. Update SRV URI Validation Requirement such that needing 3 `.` separated parts in the hostname is no longer required. 
2. Clarify that `'hostname'` refers to the entire `{hostname}` and that the `{subdomain}` is only part of the hostname. Previously, in the document, `{hostname}` referred to the `{subdomain}` and `'hostname'` referred to the `{subdomain}.{domainname}` which was confusing
3. Update the parent domain match requirement for the returned address such that SRVS with <3 parts MUST end with all parts of the SRV and have an additional domain level
4. Update the change log to reflect all of the above changes. 

#### Proof of Concept
See [here](https://github.com/mongodb/node-mongodb-native/pull/4197) for Node's PoC/implementation of downstream changes. 

- [x] Update changelog.
  clusters, and serverless).


source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
```diff
@@ -200,11 +198,28 @@ mongodb://mongodb1.mongodb.com:27317,mongodb2.mongodb.com:27107/?ssl=true&replic
 
 ## Test Plan
 
+### Spec Tests 
 See README.md in the accompanying [test directory](tests).
 
 Additionally, see the `mongodb+srv` test `invalid-uris.yml` in the
 [Connection String Spec tests](../connection-string/tests).
 
+### Prose Tests
```
nbbeeken, line 207: I think most prose tests go in the local tests/README.md file so this change should be in `source/initial-dns-seedlist-discovery/tests/README.md`

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
```diff
@@ -200,11 +198,28 @@ mongodb://mongodb1.mongodb.com:27317,mongodb2.mongodb.com:27107/?ssl=true&replic
...
+* Test #1: The driver should not throw an error when given a valid SRV record that only contains the name of the domain and the TLD.
```
nbbeeken, line 208: I think you can merge this into one test with two cases, the subject is sort of like 
- "when given an SRV" 
  - "with one domain level" 
  - "with two domain levels"
Parse the connection string without throwing a parse related error.

We have to stub DNS but other driver will be able to just call their client constructor, so I think leave out the details about setup since it will be different for each language, maybe some drivers can look for a failure to connect specifically checking that it isn't a parse related error.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
```diff
@@ -81,9 +81,7 @@ parse error and MUST NOT do DNS resolution or contact hosts.
 It is an error to specify more than one host name in a connection string with the `mongodb+srv` protocol, and the driver
 MUST raise a parse error and MUST NOT do DNS resolution or contact hosts.
 
-A driver MUST verify that in addition to the `{hostname}`, the `{domainname}` consists of at least two parts: the domain
-name, and a TLD. Drivers MUST raise an error and MUST NOT contact the DNS server to obtain SRV (or TXT records) if the
-full URI does not consist of at least three parts.
+In the case that an hostname does not have a `{hostname}` and a `{domainname}` (consisting of a domain name and a TLD) prior to DNS resolution, drivers MUST NOT throw an error.
```
nbbeeken, line 84: Can you update the change log at the end of the file? 

Saying we "must not" do X feels natural since we're implementing this change now, but reading this in the future where we don't necessarily have the historical context it is something that was never asked of the implementer so they weren't planning on it. Maybe it would be better to frame this in the positive of what should be allowed?

nbbeeken (changes requested):
Good start, there's a markdown format linter error and see comments:

nbbeeken (changes requested):
> A driver MUST verify that the host names returned through SRV records have the same parent `{domainname}`. Drivers MUST
raise an error and MUST NOT initiate a connection to any returned host name which does not share the same
`{domainname}`.

This section needs adjusting to include what validation is done when there is only one level in the domain.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
aditi-khare-mongoDB: Yep, updated the change-log! I changed it to use different wording also.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
aditi-khare-mongoDB: Good idea, I changed it to this format

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
aditi-khare-mongoDB: done!

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
```diff
@@ -81,20 +81,26 @@ parse error and MUST NOT do DNS resolution or contact hosts.
...
 If `mongodb+srv` is used, a driver MUST implicitly also enable TLS. Clients can turn this off by passing `tls=false` in
 either the Connection String, or options passed in as parameters in code to the MongoClient constructor (or equivalent
 API for each driver), but not through a TXT record (discussed in a later section).
 
 #### Querying DNS
 
-In this preprocessing step, the driver will query the DNS server for SRV records on `{hostname}.{domainname}`, prefixed
-with the SRV service name and protocol. The SRV service name is provided in the `srvServiceName` URI option and defaults
-to `mongodb`. The protocol is always `tcp`. After prefixing, the URI should look like:
-`_{srvServiceName}._tcp.{hostname}.{domainname}`. This DNS query is expected to respond with one or more SRV records.
```
aditi-khare-mongoDB, line 97: this looks like a large change, but its really just a wording change for the code snippets on 101 and 103.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
```diff
@@ -118,12 +124,12 @@ randomization.
 
 ### Default Connection String Options
 
-As a second preprocessing step, a Client MUST also query the DNS server for TXT records on `{hostname}.{domainname}`. If
```
aditi-khare-mongoDB, line 121: same here.

source/initial-dns-seedlist-discovery/tests/README.md
```diff
@@ -3,6 +3,13 @@
 This directory contains platform-independent tests that drivers can use to prove their conformance to the Initial DNS
 Seedlist Discovery spec.
 
+### Prose Tests
+
+When given a valid SRV with
+
+1. Only one domain level (ex: `{localhost}`)
+2. Only two domain levels (ex: `{mongo.local}`) Parse the connection string without throwing a parse related error.
```
aditi-khare-mongoDB, line 11: EOD, will be OOO for aug 26-30,  but note for self when I return: need to add testing for stricter parent match as well.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
```diff
@@ -81,31 +81,51 @@ parse error and MUST NOT do DNS resolution or contact hosts.
...
 
-A driver MUST verify that the host names returned through SRV records have the same parent `{domainname}`. Drivers MUST
-raise an error and MUST NOT initiate a connection to any returned host name which does not share the same
+A driver MUST verify that the host names returned through SRV records end with the original SRV's `{domainname}`.
+Drivers MUST raise an error and MUST NOT initiate a connection to any returned host name which does not share the same
 `{domainname}`.
 
+In the case that the SRV record has less than three `.` separated parts, the returned address MUST NOT be identical to
+the original `{hostname}`. The next major version should the host names returned through all SRVs require to end with
+the entire `{hostname}`.
```
nbbeeken, line 127: > In a subsequent major release, the default UUID representation can be changed to UNSPECIFIED (along with appropriate documentation indicating the backwards-breaking change). Drivers MUST document this in a prior minor release.

Here's the section from the UUID spec that indicates drivers should make a change in the next major. Can you align with this? Important point to hit: "must document"

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
```diff
@@ -81,31 +81,51 @@ parse error and MUST NOT do DNS resolution or contact hosts.
...
 found.
 
-A driver MUST verify that the host names returned through SRV records have the same parent `{domainname}`. Drivers MUST
-raise an error and MUST NOT initiate a connection to any returned host name which does not share the same
+A driver MUST verify that the host names returned through SRV records end with the original SRV's `{domainname}`.
+Drivers MUST raise an error and MUST NOT initiate a connection to any returned host name which does not share the same
 `{domainname}`.
 
+In the case that the SRV record has less than three `.` separated parts, the returned address MUST NOT be identical to
+the original `{hostname}`. The next major version should the host names returned through all SRVs require to end with
```
nbbeeken, line 126: Can you make "all SRVs" more specific about what we want to disallow? "drivers must no longer permit __ for any number of parts" something along those lines,

When it becomes the new normal for drivers to assert this we will want to understand what the old way was through the prohibition of it.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
```diff
@@ -81,31 +81,51 @@ parse error and MUST NOT do DNS resolution or contact hosts.
...
+For the purposes of this document, `{hostname}` will be divided using the following terminology.
+
+If an SRV `{hostname}` has:
+
+- 3 or more `.` separated parts:
+  - the left-most part is the `{subdomain}`
+  - the remaining portion is the `{domain}`
+  - examples:
+    - `{cluster_1.tests.mongodb.co.uk}`
+    - `{hosts_34.example.com}`
```
nbbeeken, line 101: These are examples of SRVs with 3 parts or more but they don't help the reader tell which part is which. Even though the prose says left most is X and rest is Y, there should be bullets that break it down clearly: `For example 1, "domain=" and "subdomain="`

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
```diff
@@ -81,31 +81,51 @@ parse error and MUST NOT do DNS resolution or contact hosts.
...
 
+#### Terminology
+
+For the purposes of this document, `{hostname}` will be divided using the following terminology.
+
+If an SRV `{hostname}` has:
+
+- 3 or more `.` separated parts:
+  - the left-most part is the `{subdomain}`
+  - the remaining portion is the `{domain}`
```
nbbeeken, line 98: We say "domain" here but use "domainname" in other parts of the document, we should pick one and make it consistent everywhere.

source/initial-dns-seedlist-discovery/tests/README.md
```diff
@@ -3,6 +3,31 @@
...
 
+### Prose Tests
+
+1. When given the following valid SRVs, parse the connection string without throwing a parse related error.
+
+   a. Only one domain level (ex: `mongodb+srv://localhost`).
+
+   b. Only two domain levels (ex: `mongodb+srv://mongo.local`).
+
+<br />
```
nbbeeken, line 14: Do you need `<br />`? a blank line should do the trick in markdown

source/initial-dns-seedlist-discovery/tests/README.md
```diff
@@ -3,6 +3,31 @@
 This directory contains platform-independent tests that drivers can use to prove their conformance to the Initial DNS
 Seedlist Discovery spec.
 
+### Prose Tests
+
+1. When given the following valid SRVs, parse the connection string without throwing a parse related error.
```
nbbeeken, line 8:
```suggestion
1. When given the following valid SRVs, parse the connection string

2. assert the driver does not throw a parse related error.
```

I think the steps should be clear about what to check for, "without throwing" is a bit ambiguous, should the author squash that error and move on? etc.

source/initial-dns-seedlist-discovery/tests/README.md
```diff
@@ -3,6 +3,31 @@
...
+
+   b. Only two domain levels (ex: `mongodb+srv://mongo.local`).
+
+<br />
+
+2. When given an SRV that does NOT end with the original SRV's `{domainname}`, throw a runtime error.
+
+   For example, the SRV `mongodb+srv://blogs.mongodb.com` resolving to `blogs.evil.com` should prompt an error.
+
+   Run this test for the following test cases: SRVs with one, two, and three `.` separated parts.
```
nbbeeken, line 20: This should probably be the start of the test, "for the following steps run them for each of these cases" followed by what should be checked for each, it implies the table structure we usually employ but more importantly helps the author deduplicate assertions and steps they have to perform

source/initial-dns-seedlist-discovery/tests/README.md
```diff
@@ -3,6 +3,31 @@
 This directory contains platform-independent tests that drivers can use to prove their conformance to the Initial DNS
 Seedlist Discovery spec.
 
+### Prose Tests
```
nbbeeken, line 6: The implication of the following prose tests are that drivers can stub their DNS results right? We should mention that here if so.

nbbeeken (changes requested):

source/initial-dns-seedlist-discovery/tests/README.md
aditi-khare-mongoDB: A blank line doesn't show the distinction between the tests imo since there's blank lines between them, too. The `<br />` adds another line on top of the blank line. Happy to change it if you feel its clear either way.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
aditi-khare-mongoDB: I added more detail to the examples, let me know how it looks!

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
aditi-khare-mongoDB: Let's use `domainname`! Changed the other files to reflect this.

source/initial-dns-seedlist-discovery/tests/README.md
aditi-khare-mongoDB: Sounds good, just captured this.

source/initial-dns-seedlist-discovery/tests/README.md
aditi-khare-mongoDB: Okay sounds good, changed the wording.

source/initial-dns-seedlist-discovery/tests/README.md
aditi-khare-mongoDB: Yep! I'm thinking I'll change it to talk about the pre DNS validation step not causing any errors. Messaged you on slack to align with this.

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "cSpell.words": [
+        "domainname"
+    ]
```
nbbeeken, line 4: probably meant to keep this local, right?

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
```diff
@@ -122,9 +128,10 @@ A driver MUST verify that the host names returned through SRV records end with t
 Drivers MUST raise an error and MUST NOT initiate a connection to any returned host name which does not share the same
 `{domainname}`.
 
-In the case that the SRV record has less than three `.` separated parts, the returned address MUST NOT be identical to
-the original `{hostname}`. The next major version should the host names returned through all SRVs require to end with
-the entire `{hostname}`.
+In the case that the SRV record has less than three `.` separated parts, the returned address MUST must end with the
```
nbbeeken, line 131: > the returned address MUST must end with

`must--;`

source/initial-dns-seedlist-discovery/tests/README.md
nbbeeken: Odd, let's just match formatting from other prose markdown files, I think putting html into markdown is generally undesirable and I don't see `<br />` used in other prose test files.

nbbeeken (changes requested):

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
aditi-khare-mongoDB: removed!

ShaneHarvey:
Can we get review from someone on security?

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
```diff
@@ -9,9 +9,9 @@ ______________________________________________________________________
 
 Presently, seeding a driver with an initial list of ReplicaSet or MongoS addresses is somewhat cumbersome, requiring a
 comma-delimited list of host names to attempt connections to. A standardized answer to this problem exists in the form
-of SRV records, which allow administrators to configure a single domain to return a list of host names. Supporting this
-feature would assist our users by decreasing maintenance load, primarily by removing the need to maintain seed lists at
-an application level.
+of SRV records, which allow administrators to configure a single hostname to return a list of host names. Supporting
```
spencerjackson, line 12: It's not really a hostname either... Uhhh "... a single record.." or "single resource record"?

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
```diff
@@ -31,12 +31,36 @@ step before it considers the connection string and SDAM specifications. In this
...
 
 ```
-mongodb+srv://{hostname}.{domainname}/{options}
+mongodb+srv://{subdomain}.{domainname}/{options}
 ```
 
 `{options}` refers to the optional elements from the [Connection String](../connection-string/connection-string-spec.md)
 specification following the `Host Information`. This includes the `Auth database` and `Connection Options`.
 
+For the purposes of this document, `{hostname}` will be divided using the following terminology. If an SRV `{hostname}`
```
spencerjackson, line 40: `{hostname}`... isn't defined above anymore? Do you now mean `{subdomain}`? Or is the whole FQDN now the `{hostname}`? Can this be clarified?

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
```diff
@@ -81,31 +105,34 @@ parse error and MUST NOT do DNS resolution or contact hosts.
 It is an error to specify more than one host name in a connection string with the `mongodb+srv` protocol, and the driver
 MUST raise a parse error and MUST NOT do DNS resolution or contact hosts.
 
-A driver MUST verify that in addition to the `{hostname}`, the `{domainname}` consists of at least two parts: the domain
-name, and a TLD. Drivers MUST raise an error and MUST NOT contact the DNS server to obtain SRV (or TXT records) if the
-full URI does not consist of at least three parts.
+Node:
```
spencerjackson, line 108: What does 'Node:' mean?

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
```diff
@@ -81,31 +105,34 @@ parse error and MUST NOT do DNS resolution or contact hosts.
...
 either the Connection String, or options passed in as parameters in code to the MongoClient constructor (or equivalent
 API for each driver), but not through a TXT record (discussed in a later section).
 
 #### Querying DNS
 
-In this preprocessing step, the driver will query the DNS server for SRV records on `{hostname}.{domainname}`, prefixed
-with the SRV service name and protocol. The SRV service name is provided in the `srvServiceName` URI option and defaults
-to `mongodb`. The protocol is always `tcp`. After prefixing, the URI should look like:
-`_{srvServiceName}._tcp.{hostname}.{domainname}`. This DNS query is expected to respond with one or more SRV records.
+In this preprocessing step, the driver will query the DNS server for SRV records on the hostname, prefixed with the SRV
```
spencerjackson, line 116: The hostname is the `{hostname}` above, right?

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
```diff
@@ -81,31 +105,34 @@ parse error and MUST NOT do DNS resolution or contact hosts.
...
 
 The priority and weight fields in returned SRV records MUST be ignored.
 
 If the DNS result returns no SRV records, or no records at all, or a DNS error happens, an error MUST be raised
 indicating that the URI could not be used to find hostnames. The error SHALL include the reason why they could not be
 found.
 
-A driver MUST verify that the host names returned through SRV records have the same parent `{domainname}`. Drivers MUST
-raise an error and MUST NOT initiate a connection to any returned host name which does not share the same
+A driver MUST verify that the host names returned through SRV records end with the original SRV's `{domainname}`.
```
spencerjackson, line 127: Sorry to be obnoxious, but... I think "end with" is potentially ambiguous? Consider the example above:
```
     - `{hostname}` = `cluster_1.tests.mongodb.co.uk`

       - `{subdomain}` = `cluster_1`
       - `{domainname}` = `tests.mongodb.co.uk`
 ```
In this case, the `{domainname}` is `tests.mongodb.co.uk`.

Suppose an SRV lookup for the `{hostname}` returned `cluster_1.eviltests.mongodb.co.uk`.

Technically speaking, doesn't `cluster_1.eviltests.mongodb.co.uk` "end with" `tests.mongodb.co.uk`?

File: source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md [hunk 22]
```diff
@@ -81,31 +105,34 @@ parse error and MUST NOT do DNS resolution or contact hosts.
...
+Drivers MUST raise an error and MUST NOT initiate a connection to any returned host name which does not share the same
 `{domainname}`.
 
+In the case that the SRV record has less than three `.` separated parts, the returned address MUST end with the SRV's
```
spencerjackson, line 131: To confirm, when you say "the SRV record", do you mean the DNS name we're looking up?

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
```diff
@@ -81,31 +105,34 @@ parse error and MUST NOT do DNS resolution or contact hosts.
...
+entire `{hostname}` and MUST NOT be identical to the original `{hostname}`. The next major version MUST no longer allow
```
spencerjackson, line 132: "MUST end" kind of implies string comparison. This would probably not be appropriate, because you're going to be (99.99% of the time) looking up relative DNS names. But, you'll always be getting back Fully Qualified Domain Names, which will end with a `.` character.

source/initial-dns-seedlist-discovery/tests/README.md
```diff
@@ -3,6 +3,41 @@
...
+For the following prose tests, it is assumed drivers are be able to stub DNS results to easily test invalid DNS
+resolution results.
+
+### 1. Allow SRVs with less than 3 `.` separated parts
+
+For the following test, run each of these cases: SRVs with one and two `.` separated parts.
+
+When running validation on an SRV string before DNS resolution, do not throw a error due to number of SRV parts.
+
+For example, `mongodb+srv://localhost` is a valid SRV string with one part/one domain level, and
```
spencerjackson, line 17: This would mean... you could get back node1.localhost as a valid target, right?

spencerjackson (changes requested):

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
aditi-khare-mongoDB: I changed it to 'single SRV record', I think thats what's supposed to be conveyed.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
aditi-khare-mongoDB: For clarification, I've changed it to say: `mongodb+srv://{hostname}/{options} `.  `{hostname}` is the string portions between the protocol and options.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
aditi-khare-mongoDB: My mistake, must've ended up here on accident, good catch! Just removed it.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
aditi-khare-mongoDB: Yes!

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
aditi-khare-mongoDB: Yes! The reason I was trying to clarify hostname was defined as that it was referred to more uniformly in this spec.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
aditi-khare-mongoDB: That's true, I hadn't considered it. I'm not sure what to replace this working with. I think using 'parent domain' was also confusing since we're allowing 'sibling domains' for SRVs with >3 parts (ex: `mongodb.co.uk` and `other.co.uk` are sibling domains currently allowed). What do you suggest?

Do you think it's implied that there must be a `.` separation if the returned host contains an extra domain level? It might be a bit clunky to repeat throughout the document, but I'm happy to put it in if it is a clarification!

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
aditi-khare-mongoDB: That's true, I'm wondering the same as my above question, what would you suggest instead?

We can also include a section that clarifies we're requiring a leading period in our 'ends with' checks (if an additional domain level is present) and that trailing periods in the returned hostname will be disregarded, too.

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
spencerjackson: How about "the SRV record name"?

spencerjackson (changes requested):

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
```diff
@@ -81,31 +105,32 @@ parse error and MUST NOT do DNS resolution or contact hosts.
...
 
-A driver MUST verify that the host names returned through SRV records have the same parent `{domainname}`. Drivers MUST
-raise an error and MUST NOT initiate a connection to any returned host name which does not share the same
+A driver MUST verify that the host names returned through SRV records share the original SRV's `{domainname}`. Drivers
+MUST raise an error and MUST NOT initiate a connection to any returned host name which does not share the same
 `{domainname}`.
 
+In the case that the SRV record name has less than three `.` separated parts, the returned address' `{domainname}` MUST
+be the SRV's entire `{hostname}` and MUST have at least one more domain level than the `{hostname}`. The next major
+version MUST no longer allow an SRV record, with any number of parts, to return an address that doesn't contain the
```
spencerjackson: @nbbeeken and @aditi-khare-mongoDB, regarding "The next major version MUST no longer allow an SRV record [...] to return an address that doesn't contain the SRVs' entire `{hostname}`." I just checked some SRV URIs I could see in JIRA, and I think I have some bad news on this.

It seems that Atlas hands out URIs of the form <human readable name>.<unique ID>.mongodb.net, which resolve to hostnames of the form <human readable name>-shard00-node00.<unique ID>.mongodb.net.

This suggests we're actively handing out URIs which rely on SRVs resolving to siblings. The behavior is *safe*, because different clusters' nodes are not siblings.

However, deprecation of our current behavior is probably not possible right now. I would suggest striking this text.

aditi-khare-mongoDB: Removed the extra check!

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
```diff
@@ -169,7 +194,7 @@ _mongodb._tcp.server.mongodb.com. 86400 IN SRV   0        5      27317 mongodb1.
 _mongodb._tcp.server.mongodb.com. 86400 IN SRV   0        5      27017 mongodb2.mongodb.com.
 ```
 
-The returned host names (`mongodb1.mongodb.com` and `mongodb2.mongodb.com`) must share the same parent domain name
+The returned host names (`mongodb1.mongodb.com` and `mongodb2.mongodb.com`) must end with the same domain name
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Resolved on slack, going to use 'share the same domain'

spencerjackson (approved):

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
```diff
@@ -31,12 +31,36 @@ step before it considers the connection string and SDAM specifications. In this
...
+mongodb+srv://{hostname}/{options} 
 ```
 
 `{options}` refers to the optional elements from the [Connection String](../connection-string/connection-string-spec.md)
 specification following the `Host Information`. This includes the `Auth database` and `Connection Options`.
 
+For the purposes of this document, `{hostname}` will be divided using the following terminology. If an SRV `{hostname}`
+has:
+
+1. Three or more `.` separated parts, then the left-most part is the `{subdomain}` and the remaining portion is the
```
ShaneHarvey, line 43: `subdomain` is defined here but I don't see it ever used in the rest of the spec. What is its purpose?

ShaneHarvey (commented):
pre-commit is failing: https://github.com/mongodb/specifications/actions/runs/10885366203/job/30202807636

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
aditi-khare-mongoDB: `{subdomain}` is necessary distinction for SRVs with 3+ parts since with these SRVs, we want a less strict domain match.

For example,  the hostname returned by the SRV record `blogs.mongodb.com`  should share the `{domainname}='mongodb.com`.  The SRV record `mongo.local` has only 2 parts and no hostname, so the hostname returned by the SRV should match the entire domain (ex: `cluster_1.mongo.local`).

ShaneHarvey: > mongo.local has only 2 parts and no hostname

Now I'm more confused. Isn't "hostname" defined as the whole thing: `mongodb+srv://{hostname}/{options}`? 

Also, I get that `{subdomain}` is defined here but I still don't see it ever referenced in the rest of the spec which is confusing to me. Is this accurate to say:

> Only `{hostname} is used during SRV record verification and ``{subdomain}` is ignored.

source/initial-dns-seedlist-discovery/tests/README.md
```diff
@@ -3,6 +3,41 @@
...
+`mongodb+srv://mongo.local` is a valid SRV string with two parts/two domain levels.
+
+### 2. Throw when return address does not end with SRV domain
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Good test case, I'll add it in!

ShaneHarvey (approved):

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
```diff
@@ -254,6 +283,10 @@ In the future we could consider using the priority and weight fields of the SRV
 
 ## ChangeLog
 
+- 2024-08-20: Removed requirement for URI to have three '.' separated parts; these SRVs have stricter parent domain
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Sounds good, done!

source/initial-dns-seedlist-discovery/tests/README.md
```diff
@@ -3,6 +3,55 @@
...
 Seedlist Discovery spec.
 
+## Prose Tests
+
+For the following prose tests, it is assumed drivers are be able to stub DNS results to easily test invalid DNS
+resolution results.
+
+### 1. Allow SRVs with less than 3 `.` separated parts
+
+For the following test, run each of these cases: SRVs with one and two `.` separated parts.
```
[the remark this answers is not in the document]
aditi-khare-mongoDB: Followed up with you on Slack!

dariakp (approved):

source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.md
aditi-khare-mongoDB: Bookkeeping: Resolved over slack, added in line clarifying subdomain's role / lack thereof in the verification.

aclark4life:
Unlike this [commit which adds initial seedlist tests](https://github.com/mongodb/specifications/commit/847e1bd327e12b0a90df875c6bb70db888f02731), this PR does not include any JSON test data for new domains introduced (e.g. blogs.mongodb.com). Was this intentional or oversight?

nbbeeken:
Unfortunately, there are no new actual domains to resolve

> it is assumed drivers are able to stub DNS results

Hopefully, it is possible to either stub the actual DNS resolution or unit test the parent domain match assertion function

PR #1629, opened 2024-08-09
test(DRIVERS-2511): update wire versions for 4.0+

durran:
DRIVERS-2511 Updates the wire versions in the max staleness and SDAM test suites to support 4.0+ since 3.6 servers are no longer supported.

Node PR with spec tests passing: https://github.com/mongodb/node-mongodb-native/pull/4182

- [x] Update changelog.
  clusters, and serverless).

### DIFF
--- a/source/max-staleness/tests/ReplicaSetNoPrimary/DefaultNoMaxStaleness.yml
+++ b/source/max-staleness/tests/ReplicaSetNoPrimary/DefaultNoMaxStaleness.yml
@@ -8,14 +8,14 @@ topology_description:
     type: RSSecondary
     avg_rtt_ms: 50  # Too far.
     lastUpdateTime: 0
-    maxWireVersion: 6
+    maxWireVersion: 21
     lastWrite: {lastWriteDate: {$numberLong: "1000001"}}
   - &2
     address: b:27017
     type: RSSecondary
     avg_rtt_ms: 5
     lastUpdateTime: 0
-    maxWireVersion: 6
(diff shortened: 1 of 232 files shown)


source/max-staleness/tests/ReplicaSetNoPrimary/DefaultNoMaxStaleness.yml
```diff
@@ -8,14 +8,14 @@ topology_description:
     type: RSSecondary
     avg_rtt_ms: 50  # Too far.
     lastUpdateTime: 0
-    maxWireVersion: 6
+    maxWireVersion: 21
```
ShaneHarvey, line 11: 21 is MongoDB 7.0. 7 is MongoDB 4.0. Is it worth changing these to the min server version supported? Or is is better to use 7.0?

durran: I bumped it way up to 21 so we don't have to do big PRs like this for quite a long time. Changing it to the current min server version supported would require us to do a PR like this every time we drop support for a server. If 3.6 EOL was 2021 and we're dropping driver support in 2024 then I'd assume we'd drop support for 7.0 in 2029.

ShaneHarvey: Makes sense to me.

ShaneHarvey (approved):
Tests pass in python

PR #1640, opened 2024-08-22
DRIVERS-1256: specify how drivers set timeouts on explain commands

baileympearson:
This PR specifies that drivers with explain helpers must provide a mechanism to set timeouts on the explain command.

Node POC: https://github.com/mongodb/node-mongodb-native/pull/4207

- [x] Update changelog.
- [N/A] Make sure there are generated JSON files from the YAML test files.
  clusters, and serverless).

### DIFF
--- a/source/client-side-operations-timeout/client-side-operations-timeout.md
+++ b/source/client-side-operations-timeout/client-side-operations-timeout.md
@@ -430,8 +430,8 @@ See [runCommand behavior](#runcommand-behavior).
 
 ### Explain Helpers
 
-If a driver provides an explain helper, drivers MUST take care to ensure that timeoutMS is correctly applied to the top-level
-explain command, when specified.  Care should be taken by drivers with a fluent API - the following example
+If a driver provides an explain helper, drivers MUST take care to ensure that timeoutMS is correctly applied to the
+top-level explain command, when specified. Care should be taken by drivers with a fluent API - the following example
 should apply a timeoutMS of 1000 to the `explain` command:
 
 ```typescript
--- a/source/client-side-operations-timeout/client-side-operations-timeout.md
+++ b/source/client-side-operations-timeout/client-side-operations-timeout.md
@@ -665,7 +665,7 @@ timeout for each database operation. This would mimic using `timeoutMode=ITERATI
 
 ## Changelog
 
-- 2024-08-23: Specify that explain helpers support support timeoutMS.
+- 2024-09-12: Specify that explain helpers support support timeoutMS.
 - 2023-12-07: Migrated from reStructuredText to Markdown.
 - 2022-11-17: Use minimum RTT for maxTimeMS calculation instead of 90th percentile RTT.
 - 2022-10-05: Remove spec front matter.
--- a/source/client-side-operations-timeout/client-side-operations-timeout.md
+++ b/source/client-side-operations-timeout/client-side-operations-timeout.md
@@ -428,6 +428,16 @@ check the command document for the presence of a `maxTimeMS` field.
 
 See [runCommand behavior](#runcommand-behavior).
 
+### Explain Helpers
+
+If a driver provides an explain helper, drivers MUST take care to ensure that timeoutMS is correctly applied to the top-level
+explain command, when specified.  Care should be taken by drivers with a fluent API - the following example
+should apply a timeoutMS of 1000 to the `explain` command:
+
+```typescript
+collection.find({}, { timeoutMS: 1000 }).explain();
+```
+
 ## Test Plan
 
 See the [README.rst](tests/README.md) in the tests directory.
@@ -651,6 +661,7 @@ timeout for each database operation. This would mimic using `timeoutMode=ITERATI
 
 ## Changelog
 
+- 2024-08-23: Specify that explain helpers support support timeoutMS.
 - 2023-12-07: Migrated from reStructuredText to Markdown.
 - 2022-11-17: Use minimum RTT for maxTimeMS calculation instead of 90th percentile RTT.
 - 2022-10-05: Remove spec front matter.
--- a/source/client-side-operations-timeout/client-side-operations-timeout.md
+++ b/source/client-side-operations-timeout/client-side-operations-timeout.md
@@ -433,12 +433,13 @@ See [runCommand behavior](#runcommand-behavior).
 > [!NOTE]
 > This portion of the specification is only relevant for drivers that provide `explain` helpers.
 
-When `timeoutMS` is specified, drivers MUST take care to ensure that timeoutMS is correctly applied to the top-level
-explain command. Care should be taken by drivers with a fluent API - the following example should apply a timeoutMS of
-1000 to the `explain` command:
+When `timeoutMS` is specified, drivers MUST provide a way to specify timeoutMS that results in maxTimeMS being set on
+the `explain` command. For example, Node's implementation might look like:
 
 ```typescript
-collection.find({}, { timeoutMS: 1000 }).explain();
+collection.find({}).explain({ timeoutMS: 1000 });
+// sends:
+{ explain: { find: ... }, maxTimeMS: <remaining timeoutMS - min rtt>}
 ```
 
 ## Test Plan
--- a/source/client-side-operations-timeout/client-side-operations-timeout.md
+++ b/source/client-side-operations-timeout/client-side-operations-timeout.md
@@ -428,11 +428,14 @@ check the command document for the presence of a `maxTimeMS` field.
 
 See [runCommand behavior](#runcommand-behavior).
 
-### Explain Helpers
+### Explain
 
-If a driver provides an explain helper, drivers MUST take care to ensure that timeoutMS is correctly applied to the
-top-level explain command, when specified. Care should be taken by drivers with a fluent API - the following example
-should apply a timeoutMS of 1000 to the `explain` command:
+> [!NOTE]
+> This portion of the specification is only relevant for drivers that provide `explain` helpers.
+
+When `timeoutMS` is specified, drivers MUST take care to ensure that timeoutMS is correctly applied to the top-level
+explain command. Care should be taken by drivers with a fluent API - the following example should apply a timeoutMS of
+1000 to the `explain` command:
 
 ```typescript
 collection.find({}, { timeoutMS: 1000 }).explain();
--- a/source/crud/crud.md
+++ b/source/crud/crud.md
@@ -2217,8 +2217,8 @@ interface ExplainOptions {
 }
 ```
 
-If a driver does provide explain helpers, the driver MUST ensure that its helper permits users to specify a timeout
-(maxTimeMS or timeoutMS) for the explain command specifically. An example, using Node, might look like:
+Drivers MUST ensure that its helper permits users to specify a timeout (maxTimeMS or timeoutMS) for the explain command
+specifically. An example, using Node, might look like:
 
 ```typescript
 collection.find({ name: 'john doe' }).explain({ maxTimeMS: 1000 });
--- a/source/crud/crud.md
+++ b/source/crud/crud.md
@@ -2201,8 +2201,24 @@ the [$readPreference global command argument](../message/OP_MSG.md#global-comman
 
 ### Explain
 
-Drivers MAY provide explain helpers. If a driver does provide explain helpers, the driver MUST ensure that its helper
-permits users to specify maxTimeMS for the explain command specifically. An example, using Node, might look like:
+> [!NOTE]
+> Explain helpers are optional. Drivers that do not provide explain helpers may ignore this section.
+
+```typescript
+interface ExplainOptions {
+  /**
+   * The maximum amount of time to allow the explain to run.
+   *
+   * This option is sent only if the caller explicitly provides a value. The default is to not send a value.
+   *
+   * NOTE: This option is deprecated in favor of timeoutMS.
+   */
+  maxTimeMS: Optional<Int64>;
+}
+```
+
+If a driver does provide explain helpers, the driver MUST ensure that its helper permits users to specify a timeout
+(maxTimeMS or timeoutMS) for the explain command specifically. An example, using Node, might look like:
 
 ```typescript
 collection.find({ name: 'john doe' }).explain({ maxTimeMS: 1000 });
@@ -2212,7 +2228,18 @@ collection.find({ name: 'john doe' }).explain({ maxTimeMS: 1000 });
   explain: { find: <collection>, query: { name: 'john doe' } },
   maxTimeMS: 1000
 }
+
+collection.find({ name: 'john doe' }).explain({ timeoutMS: 1000 });
+
+// sends:
+{ 
+  explain: { find: <collection>, query: { name: 'john doe' } },
+  maxTimeMS: <1000 - average rtt>
+}
 ```
+
+This ensures that users have a mechanism to apply timeouts to `explain` specifically. Drivers MUST take care to ensure
+that
 
 Drivers MUST document how users can specify options on their explain helpers.
 
--- a/source/crud/crud.md
+++ b/source/crud/crud.md
@@ -105,7 +105,7 @@ Drivers MUST enforce timeouts for all operations per the
 operations that return cursors MUST support the timeout options documented in the
 [Cursors](../client-side-operations-timeout/client-side-operations-timeout.md#cursors) section of that specification.
 All explain helpers MUST support the timeout options documented in the
-[Explain Helpers](../client-side-operations-timeout/client-side-operations-timeout.md#explain_helpers) section of that
+[Explain Helpers](../client-side-operations-timeout/client-side-operations-timeout.md#explain) section of that
 specification.
 
 ### API
--- a/source/crud/crud.md
+++ b/source/crud/crud.md
@@ -104,6 +104,9 @@ Drivers MUST enforce timeouts for all operations per the
 [Client Side Operations Timeout](../client-side-operations-timeout/client-side-operations-timeout.md) specification. All
 operations that return cursors MUST support the timeout options documented in the
 [Cursors](../client-side-operations-timeout/client-side-operations-timeout.md#cursors) section of that specification.
+All explain helpers MUST support the timeout options documented in the
+[Explain Helpers](../client-side-operations-timeout/client-side-operations-timeout.md#explain_helpers) section of that
+specification.
 
 ### API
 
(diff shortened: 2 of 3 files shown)


source/client-side-operations-timeout/client-side-operations-timeout.md
```diff
@@ -427,6 +427,16 @@ check the command document for the presence of a `maxTimeMS` field.
 
 See [runCommand behavior](#runcommand-behavior).
 
+### Explain Helpers
+
+If a driver provides an explain helper, drivers MUST take care to ensure that timeoutMS is correctly applied to the
+top-level explain command, when specified. Care should be taken by drivers with a fluent API - the following example
```
nbbeeken, line 433: "top-level explain command" - Is that referencing how the "find" is nested under an `explain` field? there is no non top-level explain right? 

Maybe remove top-level? something along the lines of "when timeoutMS is used with explain API helpers, timeoutMS must apply to the explain command." Because IIUC the concern here is that it may be ignored accidentally if the implementation is set up in such a way to only apply it to the command _being_ explained right?

source/crud/crud.md
```diff
@@ -2198,6 +2198,23 @@ the [$readPreference global command argument](../message/OP_MSG.md#global-comman
 [passing read preference to mongos and load balancers](../server-selection/server-selection.md#passing-read-preference-to-mongos-and-load-balancers)
 (if applicable).
 
+### Explain
+
+Drivers MAY provide explain helpers. If a driver does provide explain helpers, the driver MUST ensure that its helper
+permits users to specify maxTimeMS for the explain command specifically. An example, using Node, might look like:
```
nbbeeken, line 2204:
```suggestion
permits users to specify maxTimeMS for the explain command. An example, using Node.js, might look like:
```

source/crud/crud.md
```diff
@@ -2305,11 +2322,15 @@ unfortunate, we felt it better to have the bulk api be consistent with the rest
...
-command that can be run using a runCommand method.
+Q: Should drivers offer explain helpers?\
+Originally, it was determined that explain should not be exposed via
+specialized APIs in drivers (runCommand was always an option after server 3.0.). However, some drivers historically have
+offered explain APIs and continue to do so.
+
+Explain helpers are not required because it has been determined to be not a normal use-case for a driver. We'd like
+users to use the shell for this purpose. However, explain is still possible from a driver. For find, it can be passed as
+a modifier. Aggregate can be run using a runCommand method passing the explain option. In addition, server 3.0 offers an
+explain command that can be run using a runCommand method.
```
nbbeeken, line 2333: > However, explain is still possible from a driver. For find, it can be passed as
a modifier. Aggregate can be run using a runCommand method passing the explain option. In addition, server 3.0 offers an
explain command that can be run using a runCommand method.

I find this section interesting it makes it sound like explain is an option that can be sent inside of a find/agg rather than wrapping it, but is that just a misinterpretation?

source/crud/tests/README.md
```diff
@@ -676,3 +676,15 @@ InsertOne {
...
 `error` is a client error containing the message: "bulkWrite does not currently support automatic encryption".
+
+### 14. `explain` helpers allow users to specify `maxTimeMS`
+
+Create a MongoClient with command monitoring enabled (referred to as `client`).
+
+Create a collection, referred to as `collection`, with the namespace `explain-test.collection`.
+
+Run an explained find on `collection`. The find will have the query predicate `{ name: 'john doe' }`. Specify a
+maxTimeMS value of 2000ms for the `explain`.
```
nbbeeken, line 687: Maybe mention "repeat this test for every explain helper API". Drivers should make sure that this builds the correct command for their `cursor.explain({maxTimeMS})` and `find({}, {explain: {maxTimeMS}}` etc. if applicable. 

In that vein maybe you don't need to give an example command to be explained, but say check your outgoing command for all helpers?

nbbeeken (changes requested):
Great work! mostly small clarifications

source/crud/tests/README.md
baileympearson: I'm happy to leave this vague and not specify the exact API drivers should use if you'd like.  However, prose tests do usually specify the exact commands and assertions, so my preference is to leave in the exact driver API and expect command.

Also, I don't believe exhaustiveness testing is something we should specify (and frankly, I think it's unnecessary - I didn't exhaustively test every API that supports explain in the Node driver).

source/crud/crud.md
baileympearson: I didn't write this section so I don't know for certain.  I think that prior to the explain command being added, explain was done via an option, not a separate command, and this is just a legacy explanation.

source/client-side-operations-timeout/client-side-operations-timeout.md
baileympearson, line 433:
```suggestion
explain command, when specified. Care should be taken by drivers with a fluent API - the following example
```

baileympearson (commented):
batchable suggestions

source/crud/tests/README.md
nbbeeken: That sounds fine by me too, I called this out because the spec mentions both but the test does not, seems like an easy place to leave a test gap if we do not call it out here too.

source/crud/crud.md
nbbeeken: I notice it mentions server 3.0. If this doesn't apply in any supported server versions is it worth removing?

source/client-side-operations-timeout/client-side-operations-timeout.md
```diff
@@ -427,6 +427,16 @@ check the command document for the presence of a `maxTimeMS` field.
 
 See [runCommand behavior](#runcommand-behavior).
 
+### Explain Helpers
+
+If a driver provides an explain helper, drivers MUST take care to ensure that timeoutMS is correctly applied to the
+explain command, when specified. Care should be taken by drivers with a fluent API - the following example
+should apply a timeoutMS of 1000 to the `explain` command:
```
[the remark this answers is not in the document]
baileympearson: maxTimeMS needs special handling for explain because maxTimeMS takes two separate meanings: it can mean the value of maxTimeMS applied to the explained command, or it can be the value applied to the explain command.  because maxTimeMS impacts the command being built by the driver, it leads to ambiguity in a fluent-api when used with explain.  Does `collection.find(.., { maxTimeMS: 1000 }).explain()` build `{ explain: { find: ..., maxTimeMS: 1000 } }` or ` { explain: { find: .. }, maxTimeMS: 1000 }`?  That's why it's preferable to make users explicitly state the maxTimeMS they want to use for the explain.  In Node, that looks like `collection.find({}, { explain: { maxTimeMS: 1000, .. })` or `collection.find({}).explain({ maxTimeMS: ... })`.

On the other hand, timeoutMS is only ever applied to the actual operation that is executed and doesn't impact the command the driver constructs.  The outcome of `collection.find({}, { timeoutMS: 1000 }).explain()` should be the same as `collection.find({}).explain({ timeoutMS: 1000 })` because 1. the command being constructed and executed doesn't change based on timeoutMS in either example and 2. there's a single command being executed, which should be governed by the timeoutMS 1000 setting.

It seems like a separate question as to whether fluent-api helpers should accept a timeoutMS option.  I think yes, and I think that's already covered in the [Operations](https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.md#operations) section of the spec - fluent APIs fall under CRUD or other specs mentioned.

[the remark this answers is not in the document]
dariakp: > The outcome of `collection.find({}, { timeoutMS: 1000 }).explain()` should be the same as `collection.find({}).explain({ timeoutMS: 1000 })`

That's the part I would like a second opinion on. 

If I didn't know anything about the driver's implementation, I _personally_ would expect `collection.find({}, { timeoutMS: 1000 }).explain()` to ignore the `timeoutMS` set in the options of the `find` stage (because if I am writing this code, then I probably copy-pasted my `.find` from production code in its entirety with the expectation that when I run explain that it will attach any applicable options to the explained command the same way as in production code and ignore ones that aren't relevant). 

If I understand correctly, it sounds like you are thinking of the command builder chain as `.find(<find predicate>, <chain options>).explain(<chain options>)`; whereas I am thinking of it as `.find(<find predicate>, <find execution options>).explain(<explain execution options>)`. Is that correct? Or are you thinking of it as `.find(<find predicate>, <find command or chain execution options>).explain(<explain command or chain execution options>)`, where `timeoutMS` is always considered a "chain execution" option.

source/crud/crud.md
```diff
@@ -2305,11 +2322,15 @@ unfortunate, we felt it better to have the bulk api be consistent with the rest
...
 which implemented the fluent bulk API should deprecate it and drivers that have not built it should not do so.
 
-Q: What about explain?\
-Explain has been determined to be not a normal use-case for a driver. We'd like users to use the
-shell for this purpose. However, explain is still possible from a driver. For find, it can be passed as a modifier.
-Aggregate can be run using a runCommand method passing the explain option. In addition, server 3.0 offers an explain
-command that can be run using a runCommand method.
+Q: Should drivers offer explain helpers?\
+Originally, it was determined that explain should not be exposed via
+specialized APIs in drivers (runCommand was always an option after server 3.0.). However, some drivers historically have
```
jmikola, line 2327:
```suggestion
specialized APIs in drivers (runCommand was always an option after server 3.0). However, some drivers historically have
```

Superfluous period.

source/crud/crud.md
jmikola: To clarify: server 3.0 introduced the [explain](https://www.mongodb.com/docs/manual/reference/command/explain/) command. Prior to that, queries were explained using the [$explain modifier](https://www.mongodb.com/docs/v5.0/reference/operator/meta/explain/). This all pre-dates the introduction of the `find` command.

[aggregate](https://www.mongodb.com/docs/manual/reference/command/aggregate/) has historically supported an `explain` option since 2.6, but its output is limited compared to the `explain` command. This is discussed in a bit more detail in [this old Google doc](https://docs.google.com/document/d/1kHIQLLZv7ZMlW9aktZ6CIQdC1L8j3_PJxBdUx_q3FcQ/edit#heading=h.ktd5lcont67g) on aggregation performance improvements.

I agree with removing any references to these old APIs from the CRUD spec.

source/client-side-operations-timeout/client-side-operations-timeout.md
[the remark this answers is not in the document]
baileympearson: Yes, that was I was thinking.

source/crud/crud.md
```diff
@@ -2199,6 +2199,23 @@ the [$readPreference global command argument](../message/OP_MSG.md#global-comman
 [passing read preference to mongos and load balancers](../server-selection/server-selection.md#passing-read-preference-to-mongos-and-load-balancers)
 (if applicable).
 
+### Explain
+
+Drivers MAY provide explain helpers. If a driver does provide explain helpers, the driver MUST ensure that its helper
+permits users to specify maxTimeMS for the explain command specifically. An example, using Node, might look like:
```
dariakp, line 2205: Based on the discussion in the ticket, we should loosen the language here. My suggestion is to say something like "permits users to specify a timeout using the maxTimeMS or timeoutMS option ..."

source/crud/tests/README.md
```diff
@@ -677,3 +677,19 @@ InsertOne {
 
 Execute `bulkWrite` on `client` with `model`. Assert that an error (referred to as `error`) is returned. Assert that
 `error` is a client error containing the message: "bulkWrite does not currently support automatic encryption".
+
+### 14. `explain` helpers allow users to specify `maxTimeMS`
+
+Drivers that provide a multiple APIs to specify explain should ensure this test is run at least once with each distinct
```
dariakp, line 683:
```suggestion
Drivers that provide multiple APIs to specify explain should ensure this test is run at least once with each distinct
```

source/crud/tests/README.md
```diff
@@ -677,3 +677,19 @@ InsertOne {
...
+API. For example, the Node driver runs this test with option API (`collection.find({}, { explain: ... })`)git and the
```
dariakp, line 684:
```suggestion
API. For example, the Node driver runs this test with option API (`collection.find({}, { explain: ... })`) and the
```

source/crud/tests/README.md
dariakp: Good to resolve on my end

dariakp (changes requested):

source/client-side-operations-timeout/client-side-operations-timeout.md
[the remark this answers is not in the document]
matthewdale: There are a some cases where the proposed behavior is difficult to implement or leads to unexpected results:

## Non-fluent APIs
The proposed behavior is suited to fluent APIs, but could be impractically difficult to implement for drivers that do not use fluent APIs for `explain` helpers. For example, consider a possible Go `explain` helper:

```go
client.Explain(context.Background(), func(ctx context.Context) error {
	ctx, cancel := context.WithTimeout(ctx, 10 * time.Second)
	defer cancel()
	return collection.FindOne(ctx, bson.D{}).Err()
})
```
It's not clear if it's possible to detect and apply the `FindOne` timeout to `Explain`.

## Calculated `maxTimeMS`
Consider the following operation:
```typescript
collection.find({ name: 'john doe' }, { timeoutMS: 1000 });
```

The CSOT spec [requires](https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.md#command-execution) drivers to append a calculated `maxTimeMS` value to commands:
> If `timeoutMS` is set, drivers MUST append a `maxTimeMS` field to commands executed against a MongoDB server using the `minRoundTripTime` field of the selected server.

So the driver should send a command like:
```typescript
{ find: <collection>, query: { name: 'john doe' }, maxTimeMS: <1000 - min RTT> }
```

Now consider an `explain` for the same operation:
```typescript
collection.find({ name: 'john doe' }, { timeoutMS: 1000 }).explain();
```

If we apply the timeout to the `explain` operation, the driver would change the command to be explained:
```typescript
{ 
  explain: { find: <collection>, query: { name: 'john doe' } },
  maxTimeMS: <1000 - min RTT>
}
```

Instead, the driver should send the same command, but as the body of an `explain` command:
```typescript
{ 
  explain: { find: <collection>, query: { name: 'john doe' }, maxTimeMS: <1000 - min RTT> },
}
```

## Low timeout
If a user is running explain on an operation with a very low timeout, applying the operation timeout to the explain command may cause the explain command to always time-out.

For example, consider the explain operation:
```typescript
collection.find({}, { timeoutMS: 10 }).explain();
```

A 10ms timeout may be too short to get a reply, especially if the user is trying to run the `explain` from a different host than the base operation. The user would have to change the operation to be explained and hope the result is still valid.

-----

Considering those cases, I think we should not consider the timeout on the base operation in `explain` helpers. Users should specify a separate timeout if that's the behavior they want.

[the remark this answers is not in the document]
baileympearson: Thanks for the detailed response, Matt.

> Non-fluent APIs
Yeah, this is a good point and something I hadn't considered.  I'm also not familiar enough with Go to even really understand how this code snippet would work (would the explain context somehow override command execution logic in the driver to intercept any commands being sent on it and wrap them in explain?).  

> Calculated maxTimeMS
So, you've actually outlined exactly what I understood as the expected behavior.  

First, the `find` is not being executed.  The `explain` is the command being executed.  I wouldn't expect CSOT logic to apply to apply to the decorated command.

Second, this may be related to differences in explain implementations, but it's infeasible to attach maxTimeMS to the internal command (find in this case).  When the Node driver runs a `find().explain()`, we first construct the find command, and then wrap it in explain.  This happens before connection checkout.  `maxTimeMS` is calculated and attached before writing to the socket - we'd need special logic to determine if the command being decorated was an explain command, and if we should decorate the internal explain command.

> Low timeout

Conversely, it may be a better UX to default to applying a timeout than to default to _not_ applying a timeout.  Suppose a user copy-pasted code directly from their prod environment into a test environment and explained it (this is more-or-less an actual use case for the Node driver via mongosh).  The user might be surprised to run an explain and have their timeout ignored, and need to resort to killOp to cancel it if they need to.  Defaulting to a timeoutMS could result in timeouts if timeoutMS is set too low, but that at least immediately informs the user of the issue and they can rectify it and try again.

nbbeeken (approved):

source/crud/crud.md
```diff
@@ -2199,6 +2199,47 @@ the [$readPreference global command argument](../message/OP_MSG.md#global-comman
...
+  explain: { find: <collection>, query: { name: 'john doe' } },
+  maxTimeMS: 1000
+}
+
+collection.find({ name: 'john doe' }).explain({ timeoutMS: 1000 });
+
+// sends:
+{ 
+  explain: { find: <collection>, query: { name: 'john doe' } },
+  maxTimeMS: <1000 - average rtt>
```
matthewdale, line 2237: The calculation of `maxTimeMS` based on `timeoutMS` uses minimum round-trip time (see [here](https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.md#drivers-use-minimum-rtt-to-short-circuit-operations)).

```suggestion
  maxTimeMS: <1000 - min rtt>
```

source/client-side-operations-timeout/client-side-operations-timeout.md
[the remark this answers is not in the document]
matthewdale: > I wouldn't expect CSOT logic to apply to apply to the decorated command.

I'm more concerned about the command document sent to `explain` being different than what's sent when the driver runs the op, which could make the explain output invalid. Can the presence or value of `maxTimeMS` on a command ever change the execution plan?

> I'm also not familiar enough with Go to even really understand how this code snippet would work

That API is purely speculative, but is based on the pattern the Go Driver uses for `WithTransaction` and could be implemented in a similar way (i.e. by modifying op behavior by checking for special values on `ctx`). Implementing the proposed timeout behavior in that API would be difficult and nonintuitive.

> Conversely, it may be a better UX to default to applying a timeout than to default to not applying a timeout.

I think whether or not the UX is intuitive depends a lot on the driver/language. IMO requiring all drivers to "pass-forward" the operation timeout to `explain` is too prescriptive, but I can see where it makes sense for some drivers. Should we change this requirement to "MAY" instead of "MUST"?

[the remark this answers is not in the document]
matthewdale: The updated description sounds great! Also, thanks for starting the conversation about `maxTimeMS` effect on `explain`!

matthewdale (approved):
Looks good! 👍

dariakp (approved):

dariakp (approved):

PR #1641, opened 2024-08-23
DRIVERS-2789 Read Write Concern Spec to Markdown

blink1073:

- [x] Update changelog.
  clusters, and serverless).


source/uri-options/uri-options.md
```diff
@@ -92,7 +92,7 @@ to URI options apply here.
...
 | proxyPassword                        | any string                                                                                                                                                                                | defined in the [SOCKS5 support spec](../socks5-support/socks5.rst#mongoclient-configuration)                                                                                                                                 | no                                                                                                                                                                                                          | The password for username/password authentication to the SOCKS5 proxy server specified in `proxyHost`.                                                                                                                                                                                           |
```
NoahStapp, line 94: Are we tracking these links that we'll have to come back and fix later?

NoahStapp (changes requested):

source/uri-options/uri-options.md
blink1073: It will be done automatically when the socks5 spec is migrated.

NoahStapp (approved):

PR #1644, opened 2024-08-30
DRIVERS-2822 Add sort option to `updateOne`

qingyang-hu:
# Summary

Add a sort option to `updateOne()` so if it matches more than one candidate document the first one matched by the sort order will be updated.

As `UpdateOptions` is used by both `UpdateOne` and `UpdateMany`, this PR proposes raising an error if the sort option is provided explicitly in `UpdateMany`.

# Background & Motivation

The new syntax allows a sort option to be supplied to an update command. The upstream server changes are at https://jira.mongodb.org/browse/SERVER-78140.

Some example commands can be found at: https://github.com/mongodb/mongo/commit/e443421240f5a111e19cf0870ff4f545deecea8c#diff-c9ff788fbdeb3594c45d85c449a[…]21f5ef32ede8f9c858e6e3072e0c8R27

The corresponding mongosh docs:
https://www.mongodb.com/docs/upcoming/reference/method/db.collection.updateOne/
Though the [command docs](https://www.mongodb.com/docs/upcoming/reference/command/update/) have not been updated yet.

---

- [x] Update changelog.
  clusters, and serverless).


source/crud/crud.md
```diff
@@ -1070,6 +1069,18 @@ class UpdateOptions {
...
    */
   comment: Optional<any>;
+
+  /**
+   * Specify which document the operation updates if the query matches multiple
+   * documents. The first document matched by the sort order will be updated.
+   *
+   * This option is sent only if the caller explicitly provides a value. The default is to not send a value.
+   * This option is only supported by servers >= 8.0. Older servers will report an error for using this option.
+   * The driver MUST raise an error if the caller explicitly provides a value with updateMany().
```
jmikola, line 1079: This seems like it'd be an annoyance for drivers that share a common, typed UpdateOptions structure for both `updateOne()` and updateMany()`. I have a vague recollection of similar situations coming up in the past, perhaps most recently during the new `bulkWrite` spec design.

@jyemin: Can you chime in with a Java perspective?

source/crud/crud.md
```diff
@@ -2369,6 +2391,8 @@ aforementioned allowance in the SemVer spec.
 
 ## Changelog
 
+- 2024-08-30: Add sort option to `updateOne`.
```
jmikola, line 2394:
```suggestion
- 2024-08-30: Add sort option to `replaceOne` and `updateOne`.
```

Based on the changes above, I assume this applies to both methods.

On a separate note, does this change also need to be introduced to the new `bulkWrite` spec?

source/crud/tests/unified/bulkWrite-updateOne-sort.yml
```diff
@@ -0,0 +1,125 @@
...
+      databaseName: crud-tests
+  - collection:
+      id: collection0
+      database: database0
+      collectionName: coll0
+initialData:
+  - collectionName: coll0
+    databaseName: crud-tests
+    documents:
+      - _id: 1
```
jmikola, line 21: If this YAML was hand-written, feel free to inline small documents like this in the interset of human readability (e.g. `- { _id: 1, x: 11 }`). Likewise for some of the request operation options below (e.g. `filter: { _id: { $gt: 1 } }`).

source/crud/tests/unified/bulkWrite-updateOne-sort.yml
```diff
@@ -0,0 +1,125 @@
...
+                filter:
+                  _id:
+                    $gt: 1
+                sort:
+                  _id: -1
+                update:
+                  - $set:
+                      x: 1
+        expectError:
+          errorContains: "'update.updates.sort' is an unknown field"
```
jmikola, line 95: Spec tests generally don't assert the contents of server-side error messages since they can be inconsistent across server versions or otherwise subject to change without notice. `errorContains` is more commonly used to assert client-side messages, where a spec enforces a particular format.

Consider using an error code or just relying on `isClientError: false`.

source/crud/tests/unified/updateMany-sort.yml
```diff
@@ -0,0 +1,117 @@
...
+          filter:
+            _id:
+              $gt: 1
+          sort:
+            _id: -1
+          update:
+            $inc:
+              x: 1
+        expectError:
+          isError: true
```
jmikola, line 43: Per the spec, this should be a client-side error.

```suggestion
          isClientError: true
```

source/crud/tests/unified/updateMany-sort.yml
```diff
@@ -0,0 +1,117 @@
...
+    expectEvents:
+      - client: client0
+        events:
+          - commandStartedEvent:
```
jmikola, line 47: Wouldn't a client-side error preclude sending a command to the server? Consider asserting that `events` is empty here. That may be done in other spec tests for client-side errors.

Likewise, I don't think there's any need for an `outcome` assertion if we know we're not sending anything to the server.

source/crud/tests/unified/updateMany-sort.yml
```diff
@@ -0,0 +1,117 @@
...
+      - collectionName: coll0
+        databaseName: crud-tests
+        documents:
+          - _id: 1
+            x: 11
+          - _id: 2
+            x: 22
+          - _id: 3
+            x: 33
+  - description: updateMany with sort option unsupported (server-side error)
```
jmikola, line 72: Client-side validation to not allow `sort` with `updateMany` suggests that we'd never communicate with the server and thus could never test for a server-side error using the CRUD API.

I suppose you could test for a server-side error by manually executing the `update` command with `runCommand`, but that'd be more of a server test and not relevant to the CRUD spec.

source/crud/tests/unified/updateOne-sort.yml
```diff
@@ -0,0 +1,125 @@
+description: updateOne-sort
```
jmikola, line 1: I assume this file mirrors what already exists in `bulkWrite-updateOne-sort.yml`, so consider my comments there.

source/crud/crud.md
ShaneHarvey: Can we instead rely on the server to error? IE the "depend on server to return errors" mantra: https://github.com/mongodb/specifications/blob/master/source/driver-mantras.md#where-possible-depend-on-server-to-return-errors

ShaneHarvey (dismissed):

source/crud/tests/unified/updateOne-sort.yml
```diff
@@ -0,0 +1,96 @@
...
+        arguments:
+          filter: { _id: { $gt: 1 } }
+          sort: { _id: -1 }
+          update: { $inc: { x: 1 } }
+        expectError:
+          isClientError: false
+    expectEvents:
+      - client: client0
+        events:
+          - commandStartedEvent:
```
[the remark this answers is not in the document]
jmikola: I don't recall that being done in other tests. Seems like it'd be more a test of APM than this functionality, since `expectError.isClientError: false` covers the check for a server-side error already.

source/crud/tests/unified/updateMany-sort.yml
```diff
@@ -0,0 +1,56 @@
...
+      - { _id: 3, x: 33 }
+
+tests:
+  - description: UpdateMany with sort option unsupported
+    operations:
+      - name: updateMany
+        object: collection0
+        arguments:
+          filter: { _id: { $gt: 1 } }
+          sort: { _id: -1 }
```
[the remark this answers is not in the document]
jmikola: I'd be OK with removing this test entirely since it's not testing client-side behavior at all.

We should be able to trust that the server is testing that `multi: true` and `sort` conflict.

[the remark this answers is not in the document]
qingyang-hu: I will remove this test if the Go driver is the only one who shares the `UpdateOptions` for both update_one and update_many. I will also separate `UpdateOneOptions` and `UpdateManyOptions` because the Go driver will introduce a new options interface anyway in v2.0.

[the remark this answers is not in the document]
jmikola: In the interest of pragmatism, I'm completely OK with keeping the singular UpdateOptions and leaving a note about `sort` not applying to `updateMany`.

Also, I think this was previously overlooked in [one of my previous comments](https://github.com/mongodb/specifications/pull/1644#discussion_r1739499229), but shouldn't this PR entail a change to [Write Models](https://github.com/mongodb/specifications/blob/master/source/crud/bulk-write.md#write-models) in the bulk write spec? That has a separate model class for UpdateOne, so I assume `sort` can just go there (no concern with a common options interface as we have in the CRUD spec).

ShaneHarvey:
Removing myself, I'll defer to Jeremy + the bulk write spec authors.

jmikola:
@qingyang-hu: Please ping me for another review once the Bulk Write spec/tests are updated.

qingyang-hu:
@jmikola, can you please take a look again?  I've added the sort to the client-level bulk write.

blink1073 (approved):
LGTM, tested in https://github.com/mongodb/mongo-python-driver/pull/1881

source/crud/bulk-write.md
```diff
@@ -90,6 +90,14 @@ class UpdateOneModel implements WriteModel {
      * value is false.
      */
     upsert: Optional<Boolean>;
+
+    /**
+     * Specify which document the operation updates if the query matches multiple
+     * documents. The first document matched by the sort order will be updated.
+     *
+     * This option is only sent if the caller explicitly provides a value.
```
jmikola, line 98: The corresponding model in `crud.md` also includes:

> This option is only supported by servers >= 8.0. Older servers will report an error for using this option.

Should that be included here?

The other line about a server-side error when used with `updateMany()` is rightfully omitted, since the UpdateOneModel and UpdateManyModel are distinguished here.

source/crud/tests/unified/bulkWrite-replaceOne-sort.yml
```diff
@@ -0,0 +1,94 @@
+description: BulkWrite replaceOne-sort
+
+schemaVersion: "1.0"
+
+createEntities:
```
jmikola, line 5: Unified tests typically use YAML anchors/aliases for entity IDs and names. This avoids duplication of the string literals later in the test, and ensures we get validation when converting YAML to JSON instead of possible runtime errors in test runners should an entity not be found.

Consider: https://github.com/mongodb/specifications/blob/c5dcae52a77d1ac90ca8c0cea7f371b73982062e/source/crud/tests/unified/aggregate-allowdiskuse.yml#L5-L21

The anchors defined in `createEntities` are referenced in subsequent entity definitions, as well as `initialData` and command expectations further down the test file.

source/crud/tests/unified/bulkWrite-replaceOne-sort.yml
```diff
@@ -0,0 +1,94 @@
...
+  - description: BulkWrite replaceOne with sort option
+    runOnRequirements:
+      - minServerVersion: "8.0"
+    operations:
+      - object: collection0
+        name: bulkWrite
+        arguments:
+          requests:
+            - replaceOne:
+                filter: { _id: {$gt: 1 } }
```
jmikola, line 36:
```suggestion
                filter: { _id: { $gt: 1 } }
```

Consistent whitespace.

source/crud/tests/unified/client-bulkWrite-replaceOne-sort.yml
```diff
@@ -0,0 +1,73 @@
...
+
+tests:
+  - description: client bulkWrite replaceOne with sort option
+    operations:
+      - object: client0
+        name: clientBulkWrite
+        arguments:
+          models:
+            - replaceOne:
+                namespace: crud-tests.coll0
```
jmikola, line 39: I'll note that this namespace string will still end up hard-coded in the test file since YAML provides no way to concatenate aliases. However, you can define it on first use here and then reference it later in `expectEvents[0].events[0].commandStartedEvent.command.nsInfo[0].ns`.

jmikola (commented):
Suggestions to use YAML anchors/aliases for consistency, and a question about doc comments in the spec.

source/crud/bulk-write.md
qingyang-hu: This bulk write doc is for 8.0+ as described at: https://github.com/mongodb/specifications/blob/9479fa1c520aee2fb5a175c9f0973aabaa93b089/source/crud/bulk-write.md?plain=1#L4
So we don't need to include it for the sort field.

qingyang-hu:
@jmikola, can you please review the updates?

source/crud/tests/unified/client-bulkWrite-replaceOne-sort.yml
```diff
@@ -0,0 +1,76 @@
...
+initialData:
+  - collectionName: *collection0Name
+    databaseName: *database0Name
+    documents:
+      - { _id: 1, x: 11 }
+      - { _id: 2, x: 22 }
+      - { _id: 3, x: 33 }
+
+_yamlAnchors:
+  namespace: &namespace "crud-tests.coll0"
```
jmikola, line 32: Thanks for using the convention from [Top-level Fields](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.md#top-level-fields) here.

jmikola (approved):

PR #1650, opened 2024-09-05
DRIVERS-2768: Restore missing "type" option docs for search indexes

jmikola:
https://jira.mongodb.org/browse/DRIVERS-2768

b746fcc71ad7c1b376482852c3c9cbba4eed7d26 incorrectly added the "type" option to IndexOptions instead of SearchIndexOptions. Moreover, that change was made to the RST file after it was converted to Markdown (https://github.com/mongodb/specifications/commit/5c131192922823d034570b13b839df90c51b9ad2) and shortly before it was replaced with a stub pointing to the Markdown file (https://github.com/mongodb/specifications/commit/47c97fb2e72a248a521e00f6d2dc4db8acb29e4a). As a result, the change was entirely lost.

This adds back documentation for the "type" option and also restructures the SearchIndexModel and SearchIndexOptions definitions to be more consistent with IndexModel and IndexOptions. The "name" option is no longer duplicated between both interfaces.

- [x] Update changelog.

### DIFF
--- a/source/index-management/index-management.md
+++ b/source/index-management/index-management.md
@@ -902,19 +902,39 @@ await collection.dropSearchIndex('my-test-index');
 
 ```typescript
 interface SearchIndexModel {
-  // The definition for this index.
+  /**
+   * Document describing the index to create.
+   *
+   * The definition syntax depends on whether you create a standard search index
+   * or a vector search index.
+   *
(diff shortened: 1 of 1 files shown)


NoahStapp (approved):
Good catch, Jeremy! Thanks for the fix.

PR #1658, opened 2024-09-13
DRIVERS-2926 BSON Binary Vector Subtype Support

caseyclements:
Adds support for new subtype of Binary BSON type to be used for densely-packed 1d arrays (vectors).

Corresponding changes for the PyMongo as reference implementation are [here](https://github.com/mongodb/mongo-python-driver/pull/1813).

DRIVERS-2926


caseyclements:
Prose descriptions of tests that we suggested added to each driver coming soon.

caseyclements:
Rough draft of prose tests added.
Guidance would be great!

jyemin (commented):
This would benefit from a more precise set of test cases, and all the better if they are represented in YAML format like other unit tests in the specifications repository.  I know that those tests have been implemented in pymongo, but they should all be specified here (i.e. an implementer shouldn't have to go spelunking in the pymongo source code to find all the necessary test cases).  

There is no existing test format that does what we need, but I imagine a new set of  tests could look something like this:

```
{
    "description: <some description>,
    "dtype" : "int8",
    "array" : [{ "$numberInt" : 42}, ...]
    "binaryAsBase64": "JwAAAP5CAADgQA==\"
}
```

source/bson-corpus/binary-vector-subtype.md
```diff
@@ -0,0 +1,94 @@
+# BSON Binary Subtype 9 - Vector
```
ShaneHarvey, line 1: This file doesn't belong in the bson corpus. Instead let's move it to its own folder like bson-decimal128 called `bson-vector/bson-vector.md`

source/bson-corpus/binary-vector-subtype.md
```diff
@@ -0,0 +1,94 @@
...
+This document describes the addition of a new subtype to the Binary BSON type. This subtype is used for efficient
+storage and retrieval of vectors. Vectors here refer to densely packed arrays of numbers, all of the same type.
+
+## Motivation
+
+These representations correspond to the numeric types supported by popular numerical libraries for vector processing,
+such as NumPy, PyTorch, TensorFlow and Apache Arrow. Storing and retrieving vector data using the same densely packed
+format used by these libraries can result in up to 8x memory savings and orders of magnitude improvement in processing
+efficiency. Without this support, MongoDB will be at a competitive disadvantage to databases that do and our users will
+bear the additional cost of storing and processing vector data.
```
ShaneHarvey, line 19: This last sentence doesn't belong in a spec.

source/bson-corpus/binary-vector-subtype.md
```diff
@@ -0,0 +1,94 @@
...
+
+`*` The early addition of the "Packed Bit" representation was to facilitate partnerships in the expanding market of
```
ShaneHarvey, line 21: What is `*` supposed to be referring to?

source/bson-corpus/binary-vector-subtype.md
```diff
@@ -0,0 +1,94 @@
...
+The following tests have not yet been automated, but MUST still be tested.
+
+### 1. Standard encoding / decoding from a list of numbers
+
+For each data type, the API must provide an idiomatic way to consume a list of that type that encodes to BSON and
+decodes back to its original form.
+
+### 2. JSON functionality
+
+For each data type, the API must provide an idiomatic way to consume a list of that type that dumps to JSON and loads
```
ShaneHarvey, line 76: For JSON, the representation must be the same as bson Binary.

source/bson-corpus/binary-vector-subtype.md
```diff
@@ -0,0 +1,94 @@
...
+be shorthand for the 16 bit vector `[1,1,1,1,1,1,1,1,0,0,0,0,0,0,0,0]`.\
+The authors are well-aware of the inherent
+ambiguity here. This is a market-standard, unfortunately. Change is inevitable.
+
+## META
+
+The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and
+"OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+## Specification
```
ShaneHarvey, line 33: This spec would be benefited by including a diagram or model of the format itself. For example the decimal128 spec defines "structs" to represent the data format.

It would also be great to include code examples in Python that show the expected use cases (python list -> bson vector and vice versa).

source/bson-corpus/binary-vector-subtype.md
```diff
@@ -0,0 +1,94 @@
...
+- The second byte (padding) prescribes the number of bits to ignore in the final byte of the value.Ω
+
+- The remainder contains the actual vector elements packed according to dtype.
+
+All values use the little-endian format.
+
+## Reference Implementation
+
+Please consult the Python driver's `pymongo.binary` module. Prose tests described below can be found in
+`test.test_bson.TestBSON.test_vector`.
```
ShaneHarvey, line 63: Please add a link to the PYTHON ticket.

source/bson-corpus/binary-vector-subtype.md
```diff
@@ -0,0 +1,94 @@
...
+| ---------------- | ---------- | ----------------------- | ------------------------------------------------------------------------------------------- |
+| `0x03`           | INT8       | 8                       | INT8                                                                                        |
+| `0x27`           | FLOAT32    | 32                      | FLOAT                                                                                       |
+| `0x10`           | PACKED_BIT | 1     `*`               | BOOL                                                                                        |
+
+As not all data types have a bit length equal to a multiple of 8, and hence do not fit squarely into a certain number of
+bytes, a second piece of metadata, the "padding" is included. This instructs the driver of the number of bits in the
+final byte that are to be ignored.
+
+The binary structure the vector subtype's value is this. Following the binary subtype `0x09` is a two-element byte
```
ShaneHarvey, line 48: "The binary structure the vector subtype's value is this."

Incomplete sentence?

ShaneHarvey (changes requested):
+1 to Jeff's suggestion for spec tests.

source/bson-corpus/binary-vector-subtype.md
caseyclements: Yeah. It wasn't clear where to put it so I just put it somewhere.

source/bson-corpus/binary-vector-subtype.md
caseyclements: Typo. I couldn't come up with a description that satisfied me. 

@jyemin , @ShaneHarvey Thanks for the comments. This is the first time I've done anything in the specs, so I need to calibrate. I'll make a new draft tomorrow.

source/bson-corpus/binary-vector-subtype.md
caseyclements: Changed.

source/bson-corpus/binary-vector-subtype.md
caseyclements: Changed

source/bson-corpus/binary-vector-subtype.md
caseyclements: I've moved the paragraph with the asterisk directly below the table so it should be easier to follow the reference.

source/bson-binary-vector/bson-binary-vector.md
```diff
@@ -0,0 +1,73 @@
...
+
+This specification introduces a new BSON binary subtype, the vector, with value `"\x09"`.
+
+Drivers SHOULD provide idiomatic APIs to translate between arrays of numbers and this BSON Binary specification.
+
+#### Data Types
+
+Each vector can take one of multiple data types (dtypes). The following table lists the first dtypes implemented.
+
+| Vector data type | Alias      | Bits per vector element | [PyArrow Data Type](https://arrow.apache.org/docs/cpp/api/datatype.html) (for illustration) |
```
blink1073, line 34: I would say "Arrow Data Type"

source/bson-binary-vector/bson-binary-vector.md
```diff
@@ -0,0 +1,73 @@
...
+
+- The second byte (padding) prescribes the number of bits to ignore in the final byte of the value.
+
+- The remainder contains the actual vector elements packed according to dtype.
+
+All values use the little-endian format.
+
+### Reference Implementation
+
+Please consult the Python driver's `pymongo.binary` module.
```
blink1073, line 69: We usually call out a specific JIRA ticket for the reference implementation.

source/bson-corpus/binary-vector-subtype.md
caseyclements: I didn't see an example of a ticket, but I did add a link to the github module.

source/bson-binary-vector/bson-binary-vector.md
caseyclements: I could find an example so I turned the above into a hyperlink. decimal128.md does that.

source/bson-binary-vector/bson-binary-vector.md
caseyclements: Changed

source/bson-corpus/binary-vector-subtype.md
caseyclements: Happy to oblige, but I'm unsure what you have in mind for the model.

As to code examples, it was suggested to me not to include python code.  There's now a link in the reference implementation. Sufficient?

source/bson-binary-vector/bson-binary-vector.md
blink1073: Here's an example: https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.md#reference-implementations

source/bson-corpus/binary-vector-subtype.md
blink1073: We can either add an image, e.g. https://github.com/mongodb/specifications/blob/b5cbae0211d3d5049680cfc0803e4589474464bc/source/transactions/transactions.md?plain=1#L261, or ascii art (https://github.com/mongodb/specifications/blob/b5cbae0211d3d5049680cfc0803e4589474464bc/source/connection-string/connection-string-spec.md?plain=1#L32).  Here's an example of some implementation suggestion code: https://github.com/mongodb/specifications/blob/b5cbae0211d3d5049680cfc0803e4589474464bc/source/auth/auth.md#hostname-canonicalization

caseyclements: This is over the top. How about an example:

For example, a vector [6, 7] of dtype PACKED_BIT (\x10) with a padding of 3 would like like this:
`b"\x10\x03\x06\x07'`: 1 byte for dtype, 1 for padding, and 1 for each uint8.

source/bson-binary-vector/bson-binary-vector.md
caseyclements: Added.

source/bson-corpus/binary-vector-subtype.md
caseyclements: done.

source/bson-corpus/binary-vector-subtype.md
caseyclements: @ShaneHarvey @blink1073 Please take a look now. I've added an html table.

source/bson-binary-vector/tests/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+Each JSON file contains three top-level keys.
+
+- `description`: human-readable description of what is in the file
+- `test_key`: Field name used when decoding/encoding a BSON document containing the single BSON Binary for the test
+  case. Applies to *every* case.
+- `tests`: array of test case objects, each of which have the following keys. Valid cases will also contain additional
+  binary and json encoding values.
+
+#### Keys of tests objects
```
jyemin, line 30: This file should describe all the assertions that the test runner should make.  Many are obvious, but one that might be missed, for example, is that you can round trip from vector to bson and then back to vector.

IIRC the bson_corpus README does this pretty well.

jyemin (changes requested):
Looks good.  Just gave a first pass to the test runner README.

source/bson-binary-vector/bson-binary-vector.md
```diff
@@ -0,0 +1,103 @@
+# BSON Binary Subtype 9 - Vector
+
+- Status: Pending
+- Minimum Server Version: N/A
+
+______________________________________________________________________
+
+## Abstract
+
+This document describes the addition of a new subtype to the Binary BSON type. This subtype is used for efficient
```
addaleax, line 10: "new" is always relative, I imagine this document should be fairly static in time, i.e. still be relevant as-is in 5 years if we don't find a need to make any changes to it?

source/bson-binary-vector/bson-binary-vector.md
```diff
@@ -0,0 +1,103 @@
...
+Each vector can take one of multiple data types (dtypes). The following table lists the dtypes implemented.
+
+| Vector data type | Alias      | Bits per vector element | [Arrow Data Type](https://arrow.apache.org/docs/cpp/api/datatype.html) (for illustration) |
+| ---------------- | ---------- | ----------------------- | ----------------------------------------------------------------------------------------- |
+| `0x03`           | INT8       | 8                       | INT8                                                                                      |
+| `0x27`           | FLOAT32    | 32                      | FLOAT                                                                                     |
+| `0x10`           | PACKED_BIT | 1     `*`               | BOOL                                                                                      |
+
+`*` A Binary Quantized (PACKED_BIT) Vector is a vector of 0s and 1s (bits), but it is represented in memory as a list of
+integers in \[0, 255\]. So, for example, the vector `[0, 255]` would be shorthand for the 16 bit vector
```
addaleax, line 41:
```suggestion
integers in \[0, 255\]. So, for example, the vector `[0, 255]` would be shorthand for the 16-bit vector
```

source/bson-binary-vector/bson-binary-vector.md
caseyclements: That's a great call. I'll update thinking not of the first users, but of those in 5 years.

source/bson-binary-vector/bson-binary-vector.md
caseyclements: Updated

source/bson-binary-vector/bson-binary-vector.md
caseyclements: Updated

source/bson-binary-vector/tests/README.md
caseyclements: I'm not sure if this is outdated now. If you need more, would you please be specific? I am also happy to revisit once another driver takes a stab at it.

source/bson-binary-vector/bson-binary-vector.md
```diff
@@ -0,0 +1,103 @@
...
+## Abstract
+
+This document describes the subtype of the Binary BSON type used for efficient storage and retrieval of vectors. Vectors
+here refer to densely packed arrays of numbers, all of the same type.
+
+## Motivation
+
+These representations correspond to the numeric types supported by popular numerical libraries for vector processing,
+such as NumPy, PyTorch, TensorFlow and Apache Arrow. Storing and retrieving vector data using the same densely packed
+format used by these libraries can result in up to significant memory savings and processing efficiency.
```
ShaneHarvey, line 17: "result in up to significant" -> "result in significant"

source/bson-binary-vector/bson-binary-vector.md
```diff
@@ -0,0 +1,103 @@
...
+
+### META
+
+The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and
+"OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+## Specification
+
+This specification introduces a new BSON binary subtype, the vector, with value `"\x09"`.
```
ShaneHarvey, line 26: `"\x09"` -> 9.

source/bson-binary-vector/bson-binary-vector.md
```diff
@@ -0,0 +1,103 @@
...
+| `0x27`           | FLOAT32    | 32                      | FLOAT                                                                                     |
+| `0x10`           | PACKED_BIT | 1     `*`               | BOOL                                                                                      |
+
+`*` A Binary Quantized (PACKED_BIT) Vector is a vector of 0s and 1s (bits), but it is represented in memory as a list of
+integers in \[0, 255\]. So, for example, the vector `[0, 255]` would be shorthand for the 16-bit vector
+`[0,0,0,0,0,0,0,0,1,1,1,1,1,1,1,1]`. The idea is that each number (a uint8) can be stored as a single byte. Of course,
+some languages, Python for one, do not have an uint8 type, so must be represented as an int in memory, but not on disk.
+
+The authors are well-aware of the inherent ambiguity here, and alternatives. This is a market-standard, unfortunately.
+Change is inevitable.
```
ShaneHarvey, line 46: This paragraph is out of place. We should move this to a "rationale" or "Q&A" section like we do for other specs.

source/bson-binary-vector/bson-binary-vector.md
```diff
@@ -0,0 +1,103 @@
...
+
+#### Byte padding
+
+As not all data types have a bit length equal to a multiple of 8, and hence do not fit squarely into a certain number of
+bytes, a second piece of metadata, the "padding" is included. This instructs the driver of the number of bits in the
+final byte that are to be ignored.
+
+#### Binary structure
+
+Following the binary subtype `0x09` a two-element byte array of metadata precedes the packed numbers.
```
ShaneHarvey, line 56: `0x09` -> 9. We usually just refer to the subtype as a regular number. For example in https://github.com/mongodb/specifications/blob/master/source/bson-binary-encrypted/binary-encrypted.md

source/bson-binary-vector/bson-binary-vector.md
```diff
@@ -0,0 +1,103 @@
...
+### META
+
+The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and
+"OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+## Specification
+
+This specification introduces a new BSON binary subtype, the vector, with value `"\x09"`.
+
+Drivers SHOULD provide idiomatic APIs to translate between arrays of numbers and this BSON Binary specification.
```
ShaneHarvey, line 28: This sentence seems to be hiding a lot of complexity. What should the driver APIs look like? What happens if the "padding" field is non-zero but the dtype is a multiple of 8? How does the padding field change the output? Are we planning to add a NONPACKED_BIT type which represents the data as unit1 (or bool) eg the user would actually give [1, 0, 0, 1] for a 4-bit vector?

source/bson-binary-vector/bson-binary-vector.md
```diff
@@ -0,0 +1,103 @@
...
+
+#### Data Types
+
+Each vector can take one of multiple data types (dtypes). The following table lists the dtypes implemented.
+
+| Vector data type | Alias      | Bits per vector element | [Arrow Data Type](https://arrow.apache.org/docs/cpp/api/datatype.html) (for illustration) |
+| ---------------- | ---------- | ----------------------- | ----------------------------------------------------------------------------------------- |
+| `0x03`           | INT8       | 8                       | INT8                                                                                      |
+| `0x27`           | FLOAT32    | 32                      | FLOAT                                                                                     |
+| `0x10`           | PACKED_BIT | 1     `*`               | BOOL                                                                                      |
```
ShaneHarvey, line 38: The previous documents I saw had other data types too, why is this spec limited to these 3?

ShaneHarvey (changes requested):

source/bson-binary-vector/bson-binary-vector.md
caseyclements: This is a can of worms. I'm going to leave it as it is.

source/bson-binary-vector/bson-binary-vector.md
caseyclements: This whole bit was placed here to make the `*` more easy to understand. I would prefer it to stay as-is. Going into detail about the brand new field of Quantization in LLMs is not something I want to do.

source/bson-binary-vector/bson-binary-vector.md
caseyclements: I admit that there is inconsistency.  It is coming from the Tech Spec document.  For our purposes, the drivers team, I am happy to make it consistent. When we refer to "subtype" we use integers. When we refer to "dtype", we use hex. Cool? @ShaneHarvey

source/bson-binary-vector/bson-binary-vector.md
caseyclements: It is my understanding that it is up to the driver to implement the API as they please. Is that not true?

If the padding is non-zero for a dtype where it does not make sense, then the tests will fail.

source/bson-binary-vector/bson-binary-vector.md
caseyclements: This was the defined work. The other document is not up-to-date or correct.

source/bson-binary-vector/bson-binary-vector.md
```diff
@@ -0,0 +1,103 @@
...
+`b"\x10\x03\x06\x07'`: 1 byte for dtype, 1 for padding, and 1 for each uint8.
+
+<table border="1" cellspacing="0" cellpadding="5">
+  <tr>
+    <td colspan="8">1st byte: dtype (from list in previous table) </td>
+    <td colspan="8">2nd byte: padding (values in [0,7])</td>
+    <td colspan="1">binary numbers packed according to dtype</td>
+  </tr>
+  <tr>
+    <td>0</td>
```
jyemin, line 75: This table is incorrect.  The bit pattern for the first byte should be `0001 0000`, not `0001 0100`

jyemin (changes requested):

source/bson-binary-vector/bson-binary-vector.md
jyemin: I agree with Shane here.  We generally avoid editorializing like this in our RFC-like specifications.

source/bson-binary-vector/tests/README.md
jyemin: We discussed this in person, so I think we're in alignment now.  The idea is to just look at all the assertions being done in the Python driver test runner, and write them down in prose here.

source/bson-binary-vector/bson-binary-vector.md
caseyclements: I take this back. I will use simple integers when describing subtype.

source/bson-binary-vector/bson-binary-vector.md
caseyclements: Ok.

source/bson-binary-vector/bson-binary-vector.md
caseyclements: Yesterday I gained some clarity on the initial requirements. They were for implementation in Python and MongoT. This specification is under the same time pressure, so perhaps we discuss the full design and whether this should be expanded to include those that are not yet implemented.

source/bson-binary-vector/bson-binary-vector.md
```diff
@@ -0,0 +1,103 @@
...
+some languages, Python for one, do not have an uint8 type, so must be represented as an int in memory, but not on disk.
+
+The authors are well-aware of the inherent ambiguity here, and alternatives. This is a market-standard, unfortunately.
+Change is inevitable.
+
+#### Byte padding
+
+As not all data types have a bit length equal to a multiple of 8, and hence do not fit squarely into a certain number of
+bytes, a second piece of metadata, the "padding" is included. This instructs the driver of the number of bits in the
+final byte that are to be ignored.
```
nbbeeken, line 52: Is it byte padding or bit padding?

I am not fully grokking what it means to ignore, from the example below:

```
// For example, a vector `[6, 7]` ... with a padding of `3`
0x10          0x03         0x06         0x07
00001 0000    0000 0011    0000 0110    0000 0111
```

Wouldn't the last 3 bits of the final byte here be "ignored" making the vector a total of 13-bits long ignoring the last three "111"?

source/bson-binary-vector/bson-binary-vector.md
caseyclements: * f50677b6 - Moved editorial comments about PACKED_BIT ambiguity to an FAQ

source/bson-binary-vector/tests/README.md
caseyclements: 2d4ea724 - Added Required Tests section to README. Removed JSON from tests.

source/bson-binary-vector/bson-binary-vector.md
caseyclements: I've chosen a bad example, as it will thrown away the only values with 1s. I'll change it.

Bit padding or byte padding. Could be either. It's padding to make sureI hope that the description itself explains it. 

* f50b1cce -  Further improvements to PACKED_BIT with padding example.

source/bson-binary-vector/bson-binary-vector.md
```diff
@@ -0,0 +1,103 @@
...
+- The second byte (padding) prescribes the number of bits to ignore in the final byte of the value.
+
+- The remainder contains the actual vector elements packed according to dtype.
+
+For example, a vector `[6, 7]` of dtype PACKED_BIT (`\x10`) with a padding of `3` would look like this:
+`b"\x10\x03\x06\x07'`: 1 byte for dtype, 1 for padding, and 1 for each uint8.
+
+<table border="1" cellspacing="0" cellpadding="5">
+  <tr>
+    <td colspan="8">1st byte: dtype (from list in previous table) </td>
```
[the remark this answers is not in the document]
caseyclements: unsigned. It's there somewhere.

source/bson-binary-vector/bson-binary-vector.md
```diff
@@ -0,0 +1,103 @@
...
+
+- The first byte (dtype) describes its data type. The table above shows those that MUST be implemented. This table may
+  increase.
+
+- The second byte (padding) prescribes the number of bits to ignore in the final byte of the value.
+
+- The remainder contains the actual vector elements packed according to dtype.
+
+For example, a vector `[6, 7]` of dtype PACKED_BIT (`\x10`) with a padding of `3` would look like this:
+`b"\x10\x03\x06\x07'`: 1 byte for dtype, 1 for padding, and 1 for each uint8.
```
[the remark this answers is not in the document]
caseyclements: done

source/bson-binary-vector/bson-binary-vector.md
nbbeeken: I would change it to bit padding because it is the bits that are doing the padding, it is the bytes that are being padded, the verb tense kinda has an implication. 

But I agree the prose are more important, should we add here that the bits to be ignored are the least significant ones? The example demonstrates it but probably should be spelled out here.

source/bson-binary-vector/bson-binary-vector.md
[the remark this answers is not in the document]
nbbeeken: If ya could point me to it that'd be great! 

I am grabbing the terms used on the bsonspec site here, it is worth clarifying because BSON type indicators are signed but binary subtypes are unsigned, it has implications for how these values are compared against in signed contexts. For example, `$type: 255` does not look up minKey only `$type: -1` does, so when the singular byte inevitably is expanded to a larger size (prob int32) should it be sign extended or not?

source/bson-binary-vector/bson-binary-vector.md
caseyclements: RE: bit or byte padding - we use bits to pack the byte. Either word makes sense and requires explanation, so I've left as-is.

source/bson-binary-vector/tests/packed_bit.json
```diff
@@ -0,0 +1,50 @@
...
+      "padding": 0
+    },
+    {
+      "description": "Underflow Vector PACKED_BIT",
+      "valid": false,
+      "vector": [-1],
+      "dtype_hex": "0x10",
+      "dtype_alias": "PACKED_BIT",
+      "padding": 0
+    }
```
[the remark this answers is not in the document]
caseyclements: Great suggestions. Added all of them.

source/bson-binary-vector/bson-binary-vector.md
blink1073: This still needs updating

source/bson-binary-vector/bson-binary-vector.md
```diff
@@ -0,0 +1,231 @@
...
+
+### Byte padding
+
+As not all data types have a bit length equal to a multiple of 8, and hence do not fit squarely into a certain number of
+bytes, a second piece of metadata, the "padding" is included. This instructs the driver of the number of bits in the
+final byte that are to be ignored. The least-significant bits are ignored.
+
+### Binary structure
+
+Following the binary subtype `\x09` a two-element byte array of metadata precedes the packed numbers.
```
blink1073, line 53: "9"

source/bson-binary-vector/bson-binary-vector.md
```diff
@@ -0,0 +1,231 @@
...
+
+    data     # Sequence of float or int
+    dtype    # Type: DtypeEnum
+    padding  # Integer: Number of padding bits
+ End Struct
+```
+
+## Reference Implementation
+
+- PYTHON (PYTHON-4577) [pymongo.binary](https://github.com/mongodb/mongo-python-driver/blob/master/bson/binary.py)
```
blink1073, line 216: This file path might change, let's just use the full JIRA link to the ticket.

source/bson-binary-vector/bson-binary-vector.md
caseyclements: Done

source/bson-binary-vector/bson-binary-vector.md
caseyclements: Done.

source/bson-binary-vector/bson-binary-vector.md
caseyclements: Done.

source/bson-binary-vector/tests/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Required tests
+
+To prove correct in a valid case (`valid: true`), one MUST
+
+- decode the canonical_bson into its binary form, and then assert that the numeric values, dtype, and padding all match
+  those provided in the JSON.
+- encode a document from the numeric values, dtype, and padding, along with the "test_key", and assert this matches the
+  canonical_bson string.
+- For floating point number types, numerical values need not match exactly.
```
vbabanin, line 46: When converting a float32 which is already stored as an approximation, like 127.699..... for 127.7, in Java for example, to a 4-byte array and back, no additional approximation error should occur since it's a direct encoding and decoding of the binary representation. So precise matching in tests seems important for drivers that support float32 to catch any mistakes in the `float to bytes & bytes to float` algorithm itself.  Should we consider mentioning this in the test guidance for drivers that natively support float32?

Should we also specify a small margin of error for comparing floating-point values for drivers supporting only float64 where precision loss could happen?

source/bson-binary-vector/bson-binary-vector.md
```diff
@@ -0,0 +1,231 @@
...
+Finally, after we remove the last 4 bits of padding, the actual bit vector has a length of 12 and looks like this!
+
+| 1   | 1   | 1   | 0   | 1   | 1   | 1   | 0   | 1   | 1   | 1   | 0   |
+| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
+
+## API Guidance
+
+Drivers MUST implement methods for explicit encoding and decoding that adhere to the pattern described below while
+following idioms of the language of the driver.
+
```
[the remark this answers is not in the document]
caseyclements: @vbabanin This is a good idea.  Would you like to propose some prose for this? How about the following?

Implementing drivers MUST validate metadata. Padding must be 0 for all dtypes where padding does not apply. When padding is applicable, it must be in [0, 7]. If implementing `INT4` valid values would only be {0, 4}

source/bson-binary-vector/tests/README.md
caseyclements: Something like this?

"For drivers that natively support types like float32, they SHOULD implement additional testing. For example,  precise matching  for drivers that support float32 to catch any mistakes in the float to bytes & bytes to float algorithm itself."

source/bson-binary-vector/bson-binary-vector.md
```diff
@@ -0,0 +1,231 @@
...
+### Encoding
+
+```
+Function from_vector(vector: Iterable<Number>, dtype: DtypeEnum, padding: Integer = 0) -> Binary
```
vbabanin, line 129: `from_vector` currently only accepts raw vector parameters, while `as_vector` returns a `Vector` structure. To keep the API consistent, it would make sense for `from_vector` to accept a `Vector` structure directly, similar to how `as_vector` returns one. This would also reduce the need for manually extracting individual components when encoding a `Vector.`

Should `from_vector` also accept a `Vector` type as input? Perhaps this could be offered as an alternative method for drivers to implement.

source/bson-binary-vector/tests/README.md
caseyclements: I like it. I made an update. Please take a look.

nbbeeken (approved):
LGTM!

jyemin (approved):
LGTM

source/bson-binary-vector/bson-binary-vector.md
caseyclements: What about this?

The above is suggestive only. If a driver chooses to implement a Vector type (or numerous)
they MAY decide that from_vector that has a single argument, a Vector.

vbabanin (approved):

blink1073 (approved):
LGTM!

source/bson-binary-vector/bson-binary-vector.md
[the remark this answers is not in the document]
vbabanin: Validation section was added in this commit: [0ccc39944](https://github.com/mongodb/specifications/pull/1658/commits/0ccc39944d76c0fd0de7cc205641aebd49234015)

PR #1660, opened 2024-09-17
DRIVERS-2914: Don't create output collection in aggregate tests

alcaeus:
This PR fixes errors when running the aggregate-write-readPreference tests on MongoDB 8.0. The error can be avoided by not creating the output collection during test setup. In addition, I removed the outcome assertions as the tests were flaky on sharded clusters; sometimes the query run to validate the outcome would yield no results, while they were available a split-second later. Tests were run using PHP, and I'd like @comandeo-mongo to confirm the changes fix the failures that Ruby is experiencing.

- [ ] Update changelog.
  clusters, and serverless).

### DIFF
--- a/source/crud/tests/unified/aggregate-write-readPreference.yml
+++ b/source/crud/tests/unified/aggregate-write-readPreference.yml
@@ -51,9 +51,6 @@ initialData:
       - { _id: 1, x: 11 }
       - { _id: 2, x: 22 }
       - { _id: 3, x: 33 }
-  - collectionName: *collection1Name
-    databaseName: *database0Name
-    documents: []
 
 tests:
   - description: "Aggregate with $out includes read preference for 5.0+ server"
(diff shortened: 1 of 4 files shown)


comandeo-mongo (approved):
LGTM, with the Ruby driver updated tests pass on 8.0 without errors.

source/crud/tests/unified/aggregate-write-readPreference.yml
```diff
@@ -78,12 +75,6 @@ tests:
                 $readPreference: *readPreference
                 readConcern: *readConcern
                 writeConcern: *writeConcern
-    outcome: &outcome
```
jmikola, line 81: Was it also necessary to remove the `outcome` assertions? I didn't see that mentioned in [DRIVERS-2914](https://jira.mongodb.org/browse/DRIVERS-2914) (only `initialData`), and you didn't remove this assertion from `db-aggregate-write-readPreference.yml` below.

I'll concede that this isn't essential to the test. The `commandStartedEvent` assertions for `$readPreference` is the most relevant bit.

alcaeus: I've ran into issues where the outcome assertion failed with no documents being found, but they were there when I checked. Step-debugging and introducing a tiny delay fixed the issue, so I decided to remove that element of flakiness from the test, as the real test is the assertion for `$readPreference`, and we're testing server behaviour beyond that.

Not removing the assertion from `db-aggregate-write-readPreference` was an oversight and I've now removed it from that test as well for consistency.

PR #1661, opened 2024-09-18
DRIVERS-2968 Relax index type requirements in bulk write spec

isabelatkinson:

- [x] Update changelog.
  clusters, and serverless).

### DIFF
--- a/source/crud/bulk-write.md
+++ b/source/crud/bulk-write.md
@@ -860,6 +860,8 @@ batch-splitting to standardize implementations across drivers and simplify batch
 
 ## **Changelog**
 
+- 2024-09-18: Relax numeric type requirements for indexes.
+
 - 2024-05-17: Update specification status to "Accepted".
 
 - 2024-05-10: Improve rendered format for JSON-like code blocks.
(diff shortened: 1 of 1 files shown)


kevinAlbs (approved):

stIncMale (approved):

source/crud/bulk-write.md
```diff
@@ -461,6 +461,13 @@ class BulkWriteException {
 }
 ```
 
+### Index Types
+
+The `insertResults`, `updateResults`, and `deleteResults` maps in `BulkWriteResult` and the `writeErrors` map in
+`BulkWriteException` specify `Index` as their key type. This value corresponds to the index of the operation in the
+`writeModels` list that was provided to `MongoClient.bulkWrite`. Drivers SHOULD use their language's standard numeric
+type for indexes for this type (e.g. `usize` in Rust). If no standard index type exists, drivers MUST use `Int64`.
```
jmikola, line 469: Why did this only come up in the Bulk Write spec? Would it make sense to amend the CRUD spec as well?

IIRC, Rust never implemented the old bulk write API. Did you not come across this with InsertManyResult?

isabelatkinson: See this thread: https://mongodb.slack.com/archives/C035ZJL6CQN/p1724787735885729

It probably makes sense to update CRUD as well, although drivers probably won't make any changes to those types at this point. Rust already deviates from the spec and uses `usize` in `InsertManyResult`. Filed DRIVERS-2983.

PR #1663, opened 2024-09-23
DRIVERS-2789 Clean up source file structure

blink1073:
- Update folder structure for markdown files so everything is in a folder
- Rename bson-related specs to group them


blink1073:
image

source/bson-dbref/dbref.md
```diff
@@ -1,4 +1,4 @@
-# Handling of DBRefs
+# BSON DBRef
```
alcaeus, line 1: Nit: Looking at other "BSON \<type\>" specs, this suggests that `DBRef` is a BSON type, which it isn't. Any objections to just calling this the "DBRef spec"?

source/index.md
```diff
@@ -2,9 +2,12 @@
 
 - [Atlas Serverless Tests](serverless-testing/README.md)
 - [Authentication](auth/auth.md)
-- [BSON Binary Subtype 6](client-side-encryption/subtype6.md)
+- [BSON Binary Encrypted](bson-binary-encrypted/binary-encrypted.md)
```
alcaeus, line 5: Similar to how we have multiple documents in the SDAM spec folder, how about moving all of the various BSON specs into a bson folder with an index document that links to the main BSON spec (https://bsonspec.org) and our own specifications?

source/bson-dbref/dbref.md
blink1073: Ah, right, I was confusing it with the deprecated `DBPointer`.

source/index.md
alcaeus: Works for me 👍

alcaeus (approved):

source/index.md
blink1073: Haha, I meant that as a question.  Should I make a `source/bson` folder and put the other `bson-` specs in it now?

alcaeus: Oh, sorry. Yes, having a `bson` folder containing all of the various BSON specs makes sense, I'm also fine leaving the `bson-corpus` out for the time being until we know all of the references.

There could also be a single `tests` folder that contains the various tests (potentially in subfolders similar to how it's done for the CSFLE spec). Looking at #1658, the question is why those tests aren't added to the BSON corpus in the first place - was it a matter of grouping tests with the spec document?

blink1073: The tests are different from the standard corpus tests, because there is additional metadata in the BSON binary value.  So I think we should keep these folders separate.

PR #1664, opened 2024-09-23
DRIVERS-2914: Regenerate aggregate-write-readPreference.json (#1660)

joshbsiegel:
This PR regenerates a json file that didn't get updated from these yml changes: https://github.com/mongodb/specifications/pull/1660.

### DIFF
--- a/source/crud/tests/unified/aggregate-write-readPreference.json
+++ b/source/crud/tests/unified/aggregate-write-readPreference.json
@@ -154,22 +154,6 @@
             }
           ]
         }
-      ],
-      "outcome": [
-        {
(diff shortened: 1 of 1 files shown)


nbbeeken (approved):
LGTM, after this merges let's make sure to update the downstream changes field of the DRIVERS ticket to include this commit as well so folks don't just grab the original one. 

Still a question of how this made it past CI 🤔 

Node wasn't impacted by these tests (we may have a bug) so let's wait for a driver that also uses the JSON to double check the change before we are more commits deep on this work. FYI @jmikola @comandeo-mongo

alcaeus (approved):

PR #1668, opened 2024-09-24
Regenerate all JSON tests

alcaeus:
This PR aims to fix an issue introduced in #1660. By using `make -B` we can unconditionally make all targets, meaning that file modification times don't come into effect. This should catch issues as the one in #1660, as re-running `make` without modifying the original YAML file did not yield an updated JSON file.

This also resulted in two additional JSON tests being updated with changes introduced a while back.


blink1073 (approved):
LGTM!

joshbsiegel (approved):
LGTM. It might be worth considering getting rid of the `make` that's ran at the end of `generate-error-tests.py` as the `make -B` change should take care of that.

alcaeus:
> It might be worth considering getting rid of the `make` that's ran at the end of `generate-error-tests.py` as the `make -B` change should take care of that.

I think the idea of that script is to ensure that tests are updated after running locally. I'd like to take a look at that separately.

PR #1670, opened 2024-10-07
DRIVERS-2993 specify behavior of `w:0` with `MongoClient.bulkWrite`

kevinAlbs:
# Summary

Specify behavior with unacknowledged writes for `MongoClient.bulkWrite`:
- Return error if verbose results are requested.
- Return error if ordered writes are requested.
- Apply `w:0` to each batch.

Existing spec and prose tests using `MongoClient.bulkWrite` with unacknowledged write concern have been updated to use `ordered: false`. The default (`ordered: true`) is now expected to result in a client-side error.

# Background & Motivation

See DRIVERS-2993 for motivation.

This PR expects the OP_MSG moreToCome bit is always set for all batches. This agrees with [OP_MSG](https://github.com/mongodb/specifications/blob/1de6a19410f9f03899676176f53a3c77de234a34/source/message/OP_MSG.md#moretocome-on-requests):

> Clients doing unacknowledged writes MUST set the `moreToCome` flag, and MUST set the writeConcern to `w=0`.

## Prose Test

A CRUD prose test is added to insert 100,001 documents with `w:0` write concern to test multi-batch unacknowledged write.

The resulting collection count is checked to ensure the unacknowledged writes completes and avoid interfering with later tests. This was similarly done in DRIVERS-2877.

The prose test includes a step to create the collection with the `create` command as a workaround for [SERVER-95537](https://jira.mongodb.org/browse/SERVER-95537).

---

- [x] Update changelog.
  clusters, and serverless).

Tests verified with the C driver in https://github.com/mongodb/mongo-c-driver/pull/1752

### DIFF
--- a/source/crud/bulk-write.md
+++ b/source/crud/bulk-write.md
@@ -614,6 +614,11 @@ returned. This field is optional and defaults to false on the server.
 value for `verboseResults`, drivers MUST define `errorsOnly` as the opposite of `verboseResults`. If the user did not
 specify a value for `verboseResults`, drivers MUST define `errorsOnly` as `true`.
 
+Drivers MUST return a client-side error if `verboseResults` is true with an unacknowledged write concern containing the
+following message:
+
+> Cannot request unacknowledged write concern and verbose results
+
 ### `ordered`
 
 The `ordered` field defines whether writes should be executed in the order in which they were specified, and, if an
@@ -621,6 +626,11 @@ error occurs, whether the server should halt execution of further writes. It is 
 server. Drivers MUST explicitly define `ordered` as `true` in the `bulkWrite` command if a value is not specified in
 `BulkWriteOptions`. This is required to avoid inconsistencies between server and driver behavior if the server default
 changes in the future.
+
+Drivers MUST return a client-side error if `ordered` is true (including when default is applied) with an unacknowledged
+write concern containing the following message:
+
+> Cannot request unacknowledged write concern and ordered writes
 
 ### Size Limits
 
--- a/source/crud/bulk-write.md
+++ b/source/crud/bulk-write.md
@@ -935,6 +935,8 @@ batch-splitting to standardize implementations across drivers and simplify batch
 
 ## **Changelog**
 
+- 2024-10-07: Error if `w:0` is used with `ordered=true` or `verboseResults=true`.
+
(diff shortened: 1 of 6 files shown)


source/crud/tests/README.md
```diff
@@ -693,3 +695,43 @@ maxTimeMS value of 2000ms for the `explain`.
...
+Construct the following write model (referred to as `model`):
+
+```javascript
+InsertOne: {
+  "namespace": "db.coll",
+  "document": { "a": "b" }
+}
+```
+
+Construct a list of write models (referred to as `models`) with `model` repeated `maxWriteBatchSize + 1` times.
```
jyemin, line 725: Would this test execute faster if it used 5 16MB documents, rather than 100,001 small ones?

kevinAlbs: Good idea. Testing the C driver with a sharded 8.0 cluster locally shows the median of 3 runs:

```
With many small docs to exceed `maxWriteBatchSize`  : 0.65s
With few large docs to exceed `maxMessageSizeBytes` : 0.35s
```

Updated prose test to use few large docs.

isabelatkinson (approved):

source/crud/bulk-write.md
```diff
@@ -622,6 +627,11 @@ server. Drivers MUST explicitly define `ordered` as `true` in the `bulkWrite` co
 `BulkWriteOptions`. This is required to avoid inconsistencies between server and driver behavior if the server default
 changes in the future.
 
+Drivers MUST return a client-side error if `ordered` is true (including when default is applied) with an unacknowledged
```
ShaneHarvey, line 630: What should drivers that already released this api do? Should we continue supporting it until the next major version? Drop it in a minor version? We usually try to never do the latter but I'd be willing to make an expection here since unack'd writes are so niche and the fix is simple, just add ordered=False.

ShaneHarvey: Just to clarify, Python does not have that bug. We send w:1 so we can stop if a batch fails.

jmikola: @ShaneHarvey: what is the bug you're referring to? I'm a bit confused how PyMongo sending `w:1` relates to the user specifying `w:0` with `ordered:true`. Does that mean PyMongo is already disregarding the user's original write concern in order to catch failed batches?

stIncMale (approved):

PR #1673, opened 2024-10-09
DRIVERS-2830 - Record both FaaS and container metadata when both are present

NoahStapp:
DRIVERS-2830 Summary of changes:

- If both a FaaS provider and a container are present, both metadatas must be recorded. 

Python implementation: https://github.com/mongodb/mongo-python-driver/pull/1908.

### DIFF
--- a/source/mongodb-handshake/handshake.md
+++ b/source/mongodb-handshake/handshake.md
@@ -513,7 +513,7 @@ the following sets of environment variables:
 | -------------------- | ----- |
 | `AWS_EXECUTION_ENV`  | `EC2` |
 
-9. Valid container and FaaS provider
+9. Valid container and FaaS provider. This test MUST verify that only the container metadata is present in `client.env`.
 
 | Environment Variable              | Value              |
 | --------------------------------- | ------------------ |
--- a/source/mongodb-handshake/handshake.md
+++ b/source/mongodb-handshake/handshake.md
@@ -604,7 +604,7 @@ support the `hello` command, the `helloOk: true` argument is ignored and the leg
 
 ## Changelog
 
-- 2024-010-09: Clarify that FaaS metadata must not be populated when a container is also present.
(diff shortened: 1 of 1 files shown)


source/mongodb-handshake/handshake.md
```diff
@@ -327,6 +327,8 @@ Depending on which `client.env.name` has been selected, other FaaS fields in `cl
 Missing variables or variables with values not matching the expected type MUST cause the corresponding `client.env`
 field to be omitted and MUST NOT cause a user-visible error.
 
+If any fields of `client.env.container` are populated, all FaaS values MUST be entirely omitted.
```
ShaneHarvey, line 330: What about cases where a container is deployed on AWS Lambda? https://docs.aws.amazon.com/lambda/latest/dg/images-create.html

Wouldn't we want to include both in that case? 

IIRC the issue that motivated this drivers ticket was a customer was not using Lambda but lambda still showed up in the client metadata. It looks like they were using pymongo 4.6.0. We added container detection in pymongo 4.7: PYTHON-3837.

NoahStapp:
https://web.archive.org/web/20191212194243/https://www.nginx.com/blog/nginx-power-of-two-choices-load-balancing-algorithm/ appears to be a dead link.

blink1073:
Perhaps update that link to https://en.wikipedia.org/wiki/Balls_into_bins_problem?

ShaneHarvey (approved):

dariakp:
Sorry, I was OOO last week when the review was requested. Changes look ok although it looks like there are some linting failures.

NoahStapp:
> Sorry, I was OOO last week when the review was requested. Changes look ok although it looks like there are some linting failures.

Temporarily dead link, resolved itself.

ShaneHarvey:
Please update the PR title and Jira ticket title to reflect the actual change before merging.

PR #1675, opened 2024-10-14
DRIVERS-2884 Avoid connection churn when operations timeout

prestonvasquez:

- [ ] Update changelog.
  clusters, and serverless).

### DIFF
--- a/source/client-side-operations-timeout/tests/connection-churn.yml
+++ b/source/client-side-operations-timeout/tests/connection-churn.yml
@@ -1,0 +1,401 @@
+description: "operation timeouts do not cause connection churn"
+
+schemaVersion: "1.9"
+
+runOnRequirements:
+  - minServerVersion: "4.4"
+    topologies: ["replicaset", "sharded"]
+
+createEntities:
+  - client:
+      id: &failPointClient failPointClient
+      useMultipleMongoses: false
+
+initialData:
+  - collectionName: &collectionName coll
+    databaseName: &databaseName test
+    documents: []
+
+tests:
+  - description: "write op with successful pending read"
+    operations:
+      - name: createEntities
+        object: testRunner
+        arguments:
+          entities:
+            - client:
+                id: &client client
+                uriOptions:  
+                  maxPoolSize: 1
+                useMultipleMongoses: false
+                observeEvents:
+                  - commandFailedEvent
+                  - connectionClosedEvent
+            - database:
+                id: &database database
+                client: *client
+                databaseName: *databaseName
+            - collection:
+                id: &collection collection
+                database: *database
+                collectionName: *collectionName
+
+      # Create a failpoint to block first op
+      - name: failPoint
+        object: testRunner
+        arguments:
+          client: *failPointClient
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+              failCommands: ["insert"]
+              blockConnection: true
+              blockTimeMS: 750
+
+      # Execute op with timeout < block time
+      - name: insertOne
+        object: *collection
+        arguments:
+          timeoutMS: 500
+          document: { _id: 3, x: 1 }
+        expectError:
+          isTimeoutError: true
+
+      # Execute a subsequent operation to complete the read when checking out 
+      # the single available connection.
+      - name: findOne
+        object: *collection
+        arguments:
+          filter: { _id: 1 }
+
+    expectEvents:
+      - client: *client
+        events:
+          - commandFailedEvent:
+              commandName: insert
+      - client: *client
+        eventType: cmap
+        events: [] # Expect no connection closure.
+
+  - description: "write op with unsuccessful pending read"
+    operations:
+      - name: createEntities
+        object: testRunner
+        arguments:
+          entities:
+            - client:
+                id: &client client
+                uriOptions:  
+                  maxPoolSize: 1
+                useMultipleMongoses: false
+                observeEvents:
+                  - commandFailedEvent
+                  - connectionClosedEvent
+            - database:
+                id: &database database
+                client: *client
+                databaseName: *databaseName
+            - collection:
+                id: &collection collection
+                database: *database
+                collectionName: *collectionName
+
+      # Create a failpoint to block first op
+      - name: failPoint
+        object: testRunner
+        arguments:
+          client: *failPointClient
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 2 }
+            data:
+              failCommands: ["insert"]
+              blockConnection: true
+              blockTimeMS: 750
+
+      # Execute op with timeout < block time
+      - name: insertOne
+        object: *collection
+        arguments:
+          timeoutMS: 50
+          document: { _id: 3, x: 1 }
+        expectError:
+          isTimeoutError: true
+
+      # The pending read should fail.
+      - name: insertOne
+        object: *collection
+        arguments:
+          timeoutMS: 50
+          document: { _id: 3, x: 1 }
+        expectError:
+          isTimeoutError: true
+
+    expectEvents:
+      - client: *client
+        events:
+          - commandFailedEvent:
+              commandName: insert
+      - client: *client
+        eventType: cmap
+        events: 
+          - connectionClosedEvent:
+              reason: error
+
+  - description: "read op with successful pending read"
+    operations:
+      - name: createEntities
+        object: testRunner
+        arguments:
+          entities:
+            - client:
+                id: &client client
+                uriOptions:  
+                  maxPoolSize: 1
+                useMultipleMongoses: false
+                observeEvents:
+                  - commandFailedEvent
+                  - connectionClosedEvent
+            - database:
+                id: &database database
+                client: *client
+                databaseName: *databaseName
+            - collection:
+                id: &collection collection
+                database: *database
+                collectionName: *collectionName
+
+      # Create a failpoint to block first op 
+      - name: failPoint
+        object: testRunner
+        arguments:
+          client: *failPointClient
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+              failCommands: ["insert"]
+              blockConnection: true
+              blockTimeMS: 750
+
+      # Execute op with timeout < block time 
+      - name: insertOne
+        object: *collection
+        arguments:
+          timeoutMS: 500
+          document: { _id: 3, x: 1 }
+        expectError:
+          isTimeoutError: true
+
+      # Execute a subsequent operation to complete the read when checking out 
+      # the single available connection.
+      - name: findOne
+        object: *collection
+        arguments:
+          filter: { _id: 1 }
+
+    expectEvents:
+      - client: *client
+        events:
+          - commandFailedEvent:
+              commandName: insert
+      - client: *client
+        eventType: cmap
+        events: [] # Expect no connection closure.
+
+  - description: "write op with unsuccessful pending read"
+    operations:
+      - name: createEntities
+        object: testRunner
+        arguments:
+          entities:
+            - client:
+                id: &client client
+                uriOptions:  
+                  # For single-threaded drivers, ensure the operating connection
+                  # is checked out to complete the read.
+                  maxPoolSize: 1
+                useMultipleMongoses: false
+                observeEvents:
+                  - commandFailedEvent
+                  - connectionClosedEvent
+            - database:
+                id: &database database
+                client: *client
+                databaseName: *databaseName
+            - collection:
+                id: &collection collection
+                database: *database
+                collectionName: *collectionName
+
+      # Create a failpoint to block first op 
+      - name: failPoint
+        object: testRunner
+        arguments:
+          client: *failPointClient
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 2 }
+            data:
+              failCommands: ["insert"]
+              blockConnection: true
+              blockTimeMS: 750
+
+      # Execute op with timeout < block time 
+      - name: insertOne
+        object: *collection
+        arguments:
+          timeoutMS: 50
+          document: { _id: 3, x: 1 }
+        expectError:
+          isTimeoutError: true
+
+      # The pending read should fail.
+      - name: insertOne
+        object: *collection
+        arguments:
+          timeoutMS: 50
+          document: { _id: 3, x: 1 }
+        expectError:
+          isTimeoutError: true
+
+    expectEvents:
+      - client: *client
+        events:
+          - commandFailedEvent:
+              commandName: insert
+      - client: *client
+        eventType: cmap
+        events: 
+          - connectionClosedEvent:
+              reason: error
+
+  - description: "read op with successful pending read"
+    operations:
+      - name: createEntities
+        object: testRunner
+        arguments:
+          entities:
+            - client:
+                id: &client client
+                uriOptions:  
+                  maxPoolSize: 1
+                useMultipleMongoses: false
+                observeEvents:
+                  - commandFailedEvent
+                  - connectionClosedEvent
+            - database:
+                id: &database database
+                client: *client
+                databaseName: *databaseName
+            - collection:
+                id: &collection collection
+                database: *database
+                collectionName: *collectionName
+
+      # Create a failpoint to block first op 
+      - name: failPoint
+        object: testRunner
+        arguments:
+          client: *failPointClient
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 1 }
+            data:
+              failCommands: ["insert"]
+              blockConnection: true
+              blockTimeMS: 750
+
+      # Execute op with timeout < block time 
+      - name: insertOne
+        object: *collection
+        arguments:
+          timeoutMS: 500
+          document: { _id: 3, x: 1 }
+        expectError:
+          isTimeoutError: true
+
+      # Execute a subsequent operation to complete the read when checking out 
+      # the single available connection.
+      - name: findOne
+        object: *collection
+        arguments:
+          filter: { _id: 1 }
+
+    expectEvents:
+      - client: *client
+        events:
+          - commandFailedEvent:
+              commandName: insert
+      - client: *client
+        eventType: cmap
+        events: [] # Expect no connection closure.
+
+  - description: "write op with unsuccessful pending read"
+    operations:
+      - name: createEntities
+        object: testRunner
+        arguments:
+          entities:
+            - client:
+                id: &client client
+                uriOptions:  
+                  # For single-threaded drivers, ensure the operating connection
+                  # is checked out to complete the read.
+                  maxPoolSize: 1
+                useMultipleMongoses: false
+                observeEvents:
+                  - commandFailedEvent
+                  - connectionClosedEvent
+            - database:
+                id: &database database
+                client: *client
+                databaseName: *databaseName
+            - collection:
+                id: &collection collection
+                database: *database
+                collectionName: *collectionName
+
+      # Create a failpoint to block first op 
+      - name: failPoint
+        object: testRunner
+        arguments:
+          client: *failPointClient
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 2 }
+            data:
+              failCommands: ["insert"]
+              blockConnection: true
+              blockTimeMS: 750
+
+      # Execute op with timeout < block time 
+      - name: insertOne
+        object: *collection
+        arguments:
+          timeoutMS: 50
+          document: { _id: 3, x: 1 }
+        expectError:
+          isTimeoutError: true
+
+      # The pending read should fail.
+      - name: insertOne
+        object: *collection
+        arguments:
+          timeoutMS: 50
+          document: { _id: 3, x: 1 }
+        expectError:
+          isTimeoutError: true
+
+    expectEvents:
+      - client: *client
+        events:
+          - commandFailedEvent:
+              commandName: insert
+      - client: *client
+        eventType: cmap
+        events: 
+          - connectionClosedEvent:
+              reason: error
+
--- a/source/client-side-operations-timeout/tests/connection-churn.json
+++ b/source/client-side-operations-timeout/tests/connection-churn.json
@@ -1,0 +1,663 @@
+{
+  "description": "operation timeouts do not cause connection churn",
+  "schemaVersion": "1.9",
+  "runOnRequirements": [
+    {
+      "minServerVersion": "4.4",
+      "topologies": [
+        "replicaset",
+        "sharded"
+      ]
+    }
+  ],
+  "createEntities": [
+    {
+      "client": {
+        "id": "failPointClient",
+        "useMultipleMongoses": false
+      }
+    }
+  ],
+  "initialData": [
+    {
+      "collectionName": "coll",
+      "databaseName": "test",
+      "documents": []
+    }
+  ],
+  "tests": [
+    {
+      "description": "write op with successful pending read",
+      "operations": [
+        {
+          "name": "createEntities",
+          "object": "testRunner",
+          "arguments": {
+            "entities": [
+              {
+                "client": {
+                  "id": "client",
+                  "uriOptions": {
+                    "maxPoolSize": 1
+                  },
+                  "useMultipleMongoses": false,
+                  "observeEvents": [
+                    "commandFailedEvent",
+                    "connectionClosedEvent"
+                  ]
+                }
+              },
+              {
+                "database": {
+                  "id": "database",
+                  "client": "client",
+                  "databaseName": "test"
+                }
+              },
+              {
+                "collection": {
+                  "id": "collection",
+                  "database": "database",
+                  "collectionName": "coll"
+                }
+              }
+            ]
+          }
+        },
+        {
+          "name": "failPoint",
+          "object": "testRunner",
+          "arguments": {
+            "client": "failPointClient",
+            "failPoint": {
+              "configureFailPoint": "failCommand",
+              "mode": {
+                "times": 1
+              },
+              "data": {
+                "failCommands": [
+                  "insert"
+                ],
+                "blockConnection": true,
+                "blockTimeMS": 750
+              }
+            }
+          }
+        },
+        {
+          "name": "insertOne",
+          "object": "collection",
+          "arguments": {
+            "timeoutMS": 500,
+            "document": {
+              "_id": 3,
+              "x": 1
+            }
+          },
+          "expectError": {
+            "isTimeoutError": true
+          }
+        },
+        {
+          "name": "findOne",
+          "object": "collection",
+          "arguments": {
+            "filter": {
+              "_id": 1
+            }
+          }
+        }
+      ],
+      "expectEvents": [
+        {
+          "client": "client",
+          "events": [
+            {
+              "commandFailedEvent": {
+              }
+                "commandName": "insert"
+            }
+          ]
+        },
+        {
+          "client": "client",
+          "eventType": "cmap",
+          "events": []
+        }
+      ]
+    },
+    {
+      "description": "write op with unsuccessful pending read",
+      "operations": [
+        {
+          "name": "createEntities",
+          "object": "testRunner",
+          "arguments": {
+            "entities": [
+              {
+                "client": {
+                  "id": "client",
+                  "uriOptions": {
+                    "maxPoolSize": 1
+                  },
+                  "useMultipleMongoses": false,
+                  "observeEvents": [
+                    "commandFailedEvent",
+                    "connectionClosedEvent"
+                  ]
+                }
+              },
+              {
+                "database": {
+                  "id": "database",
+                  "client": "client",
+                  "databaseName": "test"
+                }
+              },
+              {
+                "collection": {
+                  "id": "collection",
+                  "database": "database",
+                  "collectionName": "coll"
+                }
+              }
+            ]
+          }
+        },
+        {
+          "name": "failPoint",
+          "object": "testRunner",
+          "arguments": {
+            "client": "failPointClient",
+            "failPoint": {
+              "configureFailPoint": "failCommand",
+              "mode": {
+                "times": 2
+              },
+              "data": {
+                "failCommands": [
+                  "insert"
+                ],
+                "blockConnection": true,
+                "blockTimeMS": 750
+              }
+            }
+          }
+        },
+        {
+          "name": "insertOne",
+          "object": "collection",
+          "arguments": {
+            "timeoutMS": 50,
+            "document": {
+              "_id": 3,
+              "x": 1
+            }
+          },
+          "expectError": {
+            "isTimeoutError": true
+          }
+        },
+        {
+          "name": "insertOne",
+          "object": "collection",
+          "arguments": {
+            "timeoutMS": 50,
+            "document": {
+              "_id": 3,
+              "x": 1
+            }
+          },
+          "expectError": {
+            "isTimeoutError": true
+          }
+        }
+      ],
+      "expectEvents": [
+        {
+          "client": "client",
+          "events": [
+            {
+              "commandFailedEvent": {
+                "commandName": "insert"
+              }
+            }
+          ]
+        },
+        {
+          "client": "client",
+          "eventType": "cmap",
+          "events": [
+            {
+              "connectionClosedEvent": {
+                "reason": "error"
+              }
+            }
+          ]
+        }
+      ]
+    },
+    {
+      "description": "read op with successful pending read",
+      "operations": [
+        {
+          "name": "createEntities",
+          "object": "testRunner",
+          "arguments": {
+            "entities": [
+              {
+                "client": {
+                  "id": "client",
+                  "uriOptions": {
+                    "maxPoolSize": 1
+                  },
+                  "useMultipleMongoses": false,
+                  "observeEvents": [
+                    "commandFailedEvent",
+                    "connectionClosedEvent"
+                  ]
+                }
+              },
+              {
+                "database": {
+                  "id": "database",
+                  "client": "client",
+                  "databaseName": "test"
+                }
+              },
+              {
+                "collection": {
+                  "id": "collection",
+                  "database": "database",
+                  "collectionName": "coll"
+                }
+              }
+            ]
+          }
+        },
+        {
+          "name": "failPoint",
+          "object": "testRunner",
+          "arguments": {
+            "client": "failPointClient",
+            "failPoint": {
+              "configureFailPoint": "failCommand",
+              "mode": {
+                "times": 1
+              },
+              "data": {
+                "failCommands": [
+                  "insert"
+                ],
+                "blockConnection": true,
+                "blockTimeMS": 750
+              }
+            }
+          }
+        },
+        {
+          "name": "insertOne",
+          "object": "collection",
+          "arguments": {
+            "timeoutMS": 500,
+            "document": {
+              "_id": 3,
+              "x": 1
+            }
+          },
+          "expectError": {
+            "isTimeoutError": true
+          }
+        },
+        {
+          "name": "findOne",
+          "object": "collection",
+          "arguments": {
+            "filter": {
+              "_id": 1
+            }
+          }
+        }
+      ],
+      "expectEvents": [
+        {
+          "client": "client",
+          "events": [
+            {
+              "commandFailedEvent": {
+                "commandName": "insert"
+              }
+            }
+          ]
+        },
+        {
+          "client": "client",
+          "eventType": "cmap",
+          "events": []
+        }
+      ]
+    },
+    {
+      "description": "write op with unsuccessful pending read",
+      "operations": [
+        {
+          "name": "createEntities",
+          "object": "testRunner",
+          "arguments": {
+            "entities": [
+              {
+                "client": {
+                  "id": "client",
+                  "uriOptions": {
+                    "maxPoolSize": 1
+                  },
+                  "useMultipleMongoses": false,
+                  "observeEvents": [
+                    "commandFailedEvent",
+                    "connectionClosedEvent"
+                  ]
+                }
+              },
+              {
+                "database": {
+                  "id": "database",
+                  "client": "client",
+                  "databaseName": "test"
+                }
+              },
+              {
+                "collection": {
+                  "id": "collection",
+                  "database": "database",
+                  "collectionName": "coll"
+                }
+              }
+            ]
+          }
+        },
+        {
+          "name": "failPoint",
+          "object": "testRunner",
+          "arguments": {
+            "client": "failPointClient",
+            "failPoint": {
+              "configureFailPoint": "failCommand",
+              "mode": {
+                "times": 2
+              },
+              "data": {
+                "failCommands": [
+                  "insert"
+                ],
+                "blockConnection": true,
+                "blockTimeMS": 750
+              }
+            }
+          }
+        },
+        {
+          "name": "insertOne",
+          "object": "collection",
+          "arguments": {
+            "timeoutMS": 50,
+            "document": {
+              "_id": 3,
+              "x": 1
+            }
+          },
+          "expectError": {
+            "isTimeoutError": true
+          }
+        },
+        {
+          "name": "insertOne",
+          "object": "collection",
+          "arguments": {
+            "timeoutMS": 50,
+            "document": {
+              "_id": 3,
+              "x": 1
+            }
+          },
+          "expectError": {
+            "isTimeoutError": true
+          }
+        }
+      ],
+      "expectEvents": [
+        {
+          "client": "client",
+          "events": [
+            {
+              "commandFailedEvent": {
+                "commandName": "insert"
+              }
+            }
+          ]
+        },
+        {
+          "client": "client",
+          "eventType": "cmap",
+          "events": [
+            {
+              "connectionClosedEvent": {
+                "reason": "error"
+              }
+            }
+          ]
+        }
+      ]
+    },
+    {
+      "description": "read op with successful pending read",
+      "operations": [
+        {
+          "name": "createEntities",
+          "object": "testRunner",
+          "arguments": {
+            "entities": [
+              {
+                "client": {
+                  "id": "client",
+                  "uriOptions": {
+                    "maxPoolSize": 1
+                  },
+                  "useMultipleMongoses": false,
+                  "observeEvents": [
+                    "commandFailedEvent",
+                    "connectionClosedEvent"
+                  ]
+                }
+              },
+              {
+                "database": {
+                  "id": "database",
+                  "client": "client",
+                  "databaseName": "test"
+                }
+              },
+              {
+                "collection": {
+                  "id": "collection",
+                  "database": "database",
+                  "collectionName": "coll"
+                }
+              }
+            ]
+          }
+        },
+        {
+          "name": "failPoint",
+          "object": "testRunner",
+          "arguments": {
+            "client": "failPointClient",
+            "failPoint": {
+              "configureFailPoint": "failCommand",
+              "mode": {
+                "times": 1
+              },
+              "data": {
+                "failCommands": [
+                  "insert"
+                ],
+                "blockConnection": true,
+                "blockTimeMS": 750
+              }
+            }
+          }
+        },
+        {
+          "name": "insertOne",
+          "object": "collection",
+          "arguments": {
+            "timeoutMS": 500,
+            "document": {
+              "_id": 3,
+              "x": 1
+            }
+          },
+          "expectError": {
+            "isTimeoutError": true
+          }
+        },
+        {
+          "name": "findOne",
+          "object": "collection",
+          "arguments": {
+            "filter": {
+              "_id": 1
+            }
+          }
+        }
+      ],
+      "expectEvents": [
+        {
+          "client": "client",
+          "events": [
+            {
+              "commandFailedEvent": {
+                "commandName": "insert"
+              }
+            }
+          ]
+        },
+        {
+          "client": "client",
+          "eventType": "cmap",
+          "events": []
+        }
+      ]
+    },
+    {
+      "description": "write op with unsuccessful pending read",
+      "operations": [
+        {
+          "name": "createEntities",
+          "object": "testRunner",
+          "arguments": {
+            "entities": [
+              {
+                "client": {
+                  "id": "client",
+                  "uriOptions": {
+                    "maxPoolSize": 1
+                  },
+                  "useMultipleMongoses": false,
+                  "observeEvents": [
+                    "commandFailedEvent",
+                    "connectionClosedEvent"
+                  ]
+                }
+              },
+              {
+                "database": {
+                  "id": "database",
+                  "client": "client",
+                  "databaseName": "test"
+                }
+              },
+              {
+                "collection": {
+                  "id": "collection",
+                  "database": "database",
+                  "collectionName": "coll"
+                }
+              }
+            ]
+          }
+        },
+        {
+          "name": "failPoint",
+          "object": "testRunner",
+          "arguments": {
+            "client": "failPointClient",
+            "failPoint": {
+              "configureFailPoint": "failCommand",
+              "mode": {
+                "times": 2
+              },
+              "data": {
+                "failCommands": [
+                  "insert"
+                ],
+                "blockConnection": true,
+                "blockTimeMS": 750
+              }
+            }
+          }
+        },
+        {
+          "name": "insertOne",
+          "object": "collection",
+          "arguments": {
+            "timeoutMS": 50,
+            "document": {
+              "_id": 3,
+              "x": 1
+            }
+          },
+          "expectError": {
+            "isTimeoutError": true
+          }
+        },
+        {
+          "name": "insertOne",
+          "object": "collection",
+          "arguments": {
+            "timeoutMS": 50,
+            "document": {
+              "_id": 3,
+              "x": 1
+            }
+          },
+          "expectError": {
+            "isTimeoutError": true
+          }
+        }
+      ],
+      "expectEvents": [
+        {
+          "client": "client",
+          "events": [
+            {
+              "commandFailedEvent": {
+                "commandName": "insert"
+              }
+            }
+          ]
+        },
+        {
+          "client": "client",
+          "eventType": "cmap",
+          "events": [
+            {
+              "connectionClosedEvent": {
+                "reason": "error"
+              }
+            }
+          ]
+        }
+      ]
+    }
+  ]
+}


source/client-side-operations-timeout/tests/connection-churn.yml
```diff
@@ -0,0 +1,401 @@
...
+        object: testRunner
+        arguments:
+          client: *failPointClient
+          failPoint:
+            configureFailPoint: failCommand
+            mode: { times: 2 }
+            data:
+              failCommands: ["insert"]
+              blockConnection: true
+              blockTimeMS: 750
```
ShaneHarvey, line 115: I worry that these tests rely on the undocumented behavior of failCommand+blockTimeMS with maxTimeMS. What if the server decides to enforce maxTimeMS even inside failCommand?

source/client-side-operations-timeout/tests/connection-churn.yml
```diff
@@ -0,0 +1,401 @@
...
+    expectEvents:
+      - client: *client
+        events:
+          - commandFailedEvent:
+              commandName: insert
+      - client: *client
+        eventType: cmap
+        events: 
+          - connectionClosedEvent:
+              reason: error
```
ShaneHarvey, line 144: Should this case result in the connection being closed? I propose the connection remain pending for as many operations as it takes to read the response up to some reasonably small limit like 1-5 seconds.

source/client-side-operations-timeout/tests/connection-churn.yml
prestonvasquez: This test anticipates the server enforcing maxTimeMS and this behavior seems to be honored:  https://gist.github.com/prestonvasquez/0a5577593742c315537ceabfe6552b0f

source/client-side-operations-timeout/tests/connection-churn.yml
prestonvasquez: Are you concerned that an operation could take a very long time, say 50s, and we have a connection being checked in and out ~50 times to complete the read? I assume this would also require drivers to keep track of the remaining bytes for each CSOT failure while checking out:

```go
n, err := io.CopyN(io.Discard, conn.nc, int64(size))
if err != nil {
	// If the read times out, record the bytes left to read before exiting.
	nerr := net.Error(nil)
	if l := int32(n); l == 0 && errors.As(err, &nerr) && nerr.Timeout() {
		conn.awaitRemainingBytes = ptrutil.Ptr(l + *conn.awaitRemainingBytes)
	}

	return fmt.Errorf("error discarding %d byte message: %w", size, err)
}
```

source/client-side-operations-timeout/tests/connection-churn.yml
```diff
@@ -0,0 +1,312 @@
...
+            data:
+              failCommands: ["find"]
+              blockConnection: true
+              blockTimeMS: 750
+
+      # Execute operation with timeout less than block time
+      - name: findOne
+        object: *collection
+        arguments:
+          timeoutMS: 50
```
ShaneHarvey, line 240: In python this timeout is too small and causes this find to fail before sending anything to the server. The same problem exists in the other tests too. Perhaps all of theses tests should run a setup command (eg ping) to ensure a connection is created and available in the pool, then run the finds. What do you think?

ShaneHarvey (changes requested):

alcaeus (commented):
I'd like to wait until #1792 is merged to review the schema changes. From what I can see in the UTF specification, those changes look good.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -997,6 +1101,79 @@ interface ConnectionCheckedInEvent {
...
+   */
+  address: string;
+
+  /**
+   *  The ID of the Connection.
+   */
+  connectionId: int64;
+
+  /**
+   *  The time it took to complete the pending read.
```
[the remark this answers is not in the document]
baileympearson: (same comment for logging events)

source/connection-monitoring-and-pooling/tests/README.md
```diff
@@ -21,6 +21,51 @@ The following tests have not yet been automated, but MUST still be tested:
...
+#### Connection Aliveness Check Fails
+
+1. Initialize a mock TCP listener to simulate the server-side behavior. The listener should write at least 5 bytes to
+    the connection to prevent size-related errors (e.g. `0x01, 0x02, 0x03, 0x04, 0x05, 0xFF`). The response should be
+    delayed by 2x the size of the socket timeout.
+2. Implement a monitoring mechanism to capture the `ConnectionPendingResponseStarted` and
+    `ConnectionPendingResponseFailed` events.
+3. Instantiate a connection pool using the mock listener’s address, ensuring readiness without error. Attach the event
+    monitor to observe the connection’s state.
+4. Check out a connection from the pool and initiate a read operation with an appropriate socket timeout (e.g, 10ms)
```
baileympearson, line 39: This comment is related to my other comment about a less socket-api specific implementation.  Is it possible to write these tests in a way that doesn't require explicit use of a `read` API?  Node's connection layer doesn't expose a `read` method method, we only expose `command()` which performs write+read on the underlying socket.

source/connection-monitoring-and-pooling/tests/README.md
```diff
@@ -21,6 +21,51 @@ The following tests have not yet been automated, but MUST still be tested:
...
+    that will trigger a timeout due to the artificial delay of 2x the socket timeout (step 1). Ensure that the read
+    operation returns a timeout error.
+5. Check the connection back into the pool and sleep for 3 seconds so that the pending response state timestamp exceeds
+    the pending response timeout, forcing an aliveness check.
+6. Check the connection out. The aliveness check should fail since no additional bytes were added after the delay in
+    step 1.
+7. Verify that one event for each `ConnectionPendingResponseStarted` and `ConnectionPendingResponseFailed` was emitted.
+    Also verify that the fields were correctly set for each event.
+
+#### Connection Aliveness Check Succeeds
```
baileympearson, line 49: Can we add a test that demonstrates that multiple aliveness checks might be required to fully read the response from the socket?  I'm imagining the mock server emitting chunks of data every second for longer than 6s (2 * the static timeout).  Each checkout should fail, but we'll continue to read

source/connection-monitoring-and-pooling/tests/logging/connection-logging.yml
```diff
@@ -222,3 +222,236 @@ tests:
               reason: "An error occurred while trying to establish a new connection"
               error: { $$exists: true }
               durationMS: { $$type: [double, int, long] }
+
+  - description: "force a pending response read, fail first try, succeed second try"
+    operations:
+      - name: createEntities
```
baileympearson, line 228: If possible, can we add a test that demonstrates that when the pending read checkout has no timeoutMS set, we use socket_timeout_ms (if it is <3s)?

baileympearson (changes requested):

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -576,7 +576,106 @@ other threads from checking out [Connections](#connection) while establishing a
 Before a given [Connection](#connection) is returned from checkOut, it must be marked as "in use", and the pool's
```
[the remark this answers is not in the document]
prestonvasquez: > Can we choose a clear name for this new state?

Absolutely, great point!

> Should we allow drivers with background threads to close timed-out pending connections with the background thread? 

IIUC this would require polling connections and then performing aliveness checks. This would not be more performant than awaiting a pending read on check-out.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -576,7 +576,106 @@ other threads from checking out [Connections](#connection) while establishing a
...
 availableConnectionCount MUST be decremented.
 
-```text
+If an operation times out the socket while awaiting a server response and CSOT is enabled and `maxTimeMS` was added to
+the command, the driver MUST mark the connection as "pending" and record the current time in a way that can be updated.
+The next time the connection is checked out, the driver MUST attempt to read and discard the remaining response from the
+socket. The workflow for this is as follows:
+
+- The connection MUST persist the current time recorded immediately after the original socket timeout, and this
```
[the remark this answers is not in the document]
prestonvasquez: I have two thoughts on this. First, some drivers will likely add an object to the connection to maintain state for a pending response, for example:

```go
type pendingResponseState struct {
	remainingBytes int32
	requestID      int32
	start          time.Time
}

type connection struct {
	// pendingResponseState contains information required to attempt a pending read
	// in the event of a socket timeout for an operation that has appended
	// maxTimeMS to the wire message.
	pendingResponseState   *pendingResponseState
}
```

It would be more pragmatic to update the current time where remainginBytes and requestID are assigned (which is when the socket times out). 

Additionally, we want to start this “countdown” ASAP in case the connection is “dead”: in such cases the “aliveness check” will be a non-blocking failure while awaiting a pending response. Delaying when we set the current time reduces the likelihood (albeit small) that the cull 3 second pending response timeout has been exceeded while the connection remains idle in the pool.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -576,7 +576,106 @@ other threads from checking out [Connections](#connection) while establishing a
 Before a given [Connection](#connection) is returned from checkOut, it must be marked as "in use", and the pool's
 availableConnectionCount MUST be decremented.
 
-```text
+If an operation times out the socket while awaiting a server response and CSOT is enabled and `maxTimeMS` was added to
+the command, the driver MUST mark the connection as "pending" and record the current time in a way that can be updated.
```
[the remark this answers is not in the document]
prestonvasquez: IIUC in the Node case, bytes that have not been consumed will still sit on the socket after a timeout. So you would still need to read and discard any buffered data when checking out a connection that has been pinned to a pending response. The three bullet points you note do not conflict with the algorithm, AFAICT. I would be happy to troubleshoot this offline with you, though.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -576,7 +576,106 @@ other threads from checking out [Connections](#connection) while establishing a
...
+The next time the connection is checked out, the driver MUST attempt to read and discard the remaining response from the
+socket. The workflow for this is as follows:
+
+- The connection MUST persist the current time recorded immediately after the original socket timeout, and this
+    timestamp MUST be updated to the current time whenever any data is successfully read from the socket during a
```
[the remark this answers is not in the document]
prestonvasquez: Wow, good question. It’s notable that for a connection to be in the pending state there must have (1) been a socket timeout on the connection, (2) the operation has to enable CSOT and maxTimeMS has to be set. I would expect a server-side timeout to cancel further data transmission. Are we concerned specifically about a case where the socket timeout beats maxTimeMS and the operation succeeds server-side but the connection is still put into a "pending response” state? 

CC: @ShaneHarvey

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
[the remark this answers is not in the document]
prestonvasquez: I would anticipate this duration to be within the context of ConnectionPendingResponseStarted, i.e. 1 call to await_pending_response.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -997,6 +1101,79 @@ interface ConnectionCheckedInEvent {
...
+   *  The ServerAddress of the Endpoint the pool is attempting to connect to.
+   */
+  address: string;
+
+  /**
+   *  The ID of the Connection.
+   */
+  connectionId: int64;
+
+  /**
```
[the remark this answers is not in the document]
prestonvasquez: I do think this is a good idea, this could highlight the case where you are trickling 1 byte aliveness checks but the response continues to time out while attempting to discard the input TCP buffer. Good call.

source/connection-monitoring-and-pooling/tests/README.md
prestonvasquez: So Node only has access to a round-trip API? Even in the non-public API? Could you not just discard the write half in the mock listener?

source/connection-monitoring-and-pooling/tests/README.md
prestonvasquez: It’s unclear to me what the goal of this would be. Are you just wanting to make sure the driver does not pre-maturely close the connection if the aliveness check succeeds? We could use event monitoring for that instead.

alcaeus (approved):
Changes to the unified test format LGTM.

@prestonvasquez as per our conversation around where to add the missing event names in #1782, this schema version would be an ideal candidate as it already adds new events to the list.

source/connection-monitoring-and-pooling/tests/README.md
baileympearson: I interpreted "mock listener" as "mock server" but reading over this test more closely, that doesn't seem to be what you're asking for.  Are you asking for some way of stubbing a socket and pushing chunks into the sockets buffer, instead of a mock server that the driver opens real TCP sockets to?  That's possible in Node but I'd check with other drivers (I don't know if something like this exists yet in the specs, it may not be possible in other drivers).  

Regardless, I'm not sure if these tests would work for the Node implementation I mentioned [here](https://github.com/mongodb/specifications/pull/1675#discussion_r2071786887)

source/connection-monitoring-and-pooling/tests/README.md
baileympearson: Yes - I want a test somewhere that demonstrates that drivers might need to span multiple checkout + pending responses, and that that process can take longer than the static timeout, before reading the full response from the connection.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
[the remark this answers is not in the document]
baileympearson: > Are we concerned specifically about a case where the socket timeout beats maxTimeMS and the operation succeeds server-side but the connection is still put into a "pending response” state?

Yeah - like a network delay that slows down the response from the server initially.  There wouldn't be anything that ends the operation server side in that scenario

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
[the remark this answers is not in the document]
prestonvasquez: This change was made in 9500fd5b516920c2236f837cee4622d69a2d017f

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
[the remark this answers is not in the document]
ShaneHarvey: Yes this is a good call out. If moreToCome is present on the response, we should keep reading the connection until we get a response with moreToCome:false or the timeout expires.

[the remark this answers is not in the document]
prestonvasquez: Suggest we prohibit drivers from transitioning a connection to a "pending response" state if the command uses an OP_MSG with the `exhuastAllowed` bit flag. Instead, for socket timeouts we should close the connection per existing behavior.

This is introduced as a prose test in this PR but I've opened DRIVERS-3189 to expected `cursorType` with `exhaust` in the unified test format.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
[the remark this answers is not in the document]
prestonvasquez: Sounds good, I've made updates to the phrasing to account for both push and pull socket i/o patterns.

source/connection-monitoring-and-pooling/tests/README.md
prestonvasquez: Yeah, the word "mock" is redundant here. We just need a way to simulate the response portion of the round trip using a TCP listener. Do you have suggestions on how we can accommodate theses tests for Node? No other drivers indicated this would be an issue: https://mongodb.slack.com/archives/C72LB5RPV/p1746558337106509

source/connection-monitoring-and-pooling/tests/README.md
prestonvasquez: I suggest we defer this to the unified spec test and not further complicate the prose aliveness checks. Validating that a successful aliveness check does not result in a closed connection would imply that a subsequent check out would either perform an aliveness check or continue attempting to drain the buffer. LMK what you think.

source/client-side-operations-timeout/tests/pending-response.yml
```diff
@@ -0,0 +1,344 @@
...
+          - connectionPendingResponseStarted: {} # First find fails
+          - connectionPendingResponseFailed:
+              reason: timeout
+          - connectionPendingResponseStarted: {} # Second find drains the buffer.
+          - connectionPendingResponseSucceeded: {}
+          - connectionCheckedOutEvent: {}
+          - connectionCheckedInEvent: {} # Second find succeeds.  
+    # If the connection is closed server-side while draining the response, the 
+    # driver must close the connection.
+  - description: "connection closed server-side while draining response"
```
prestonvasquez, line 284: This seems to be a sufficient check that if the `awaitPendingResponse` function fails with a non-timeout error the connection should be closed.

source/client-side-operations-timeout/tests/pending-response.yml
```diff
@@ -0,0 +1,344 @@
...
+        arguments:
+          ms: 500
+      # Execute a subsequent operation with a timeout less than the block time.
+      - name: findOne
+        object: *collection
+        arguments:
+          timeoutMS: 50
+          filter: {_id: 1}
+        expectError:
+          isTimeoutError: false
```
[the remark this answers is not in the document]
prestonvasquez: CMAP only makes the [pool-cleared error ](https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md#clearing-a-non-load-balanced-pool) retryable at check-out. Since retryable reads occur at the operation layer and this particular network error happens at the connection pool layer (before a read command goes on the wire), I think we would have to extend the CMAP spec to say that network errors while checking out qualify as retryable.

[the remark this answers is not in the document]
ShaneHarvey: Checkout errors should already be retryable. For example, a network error when establishing a new connection will cause an automatic retry.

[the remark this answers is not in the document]
prestonvasquez: The only error type we tag as retryable during checkOut is `PoolClearedError`. Nothing else in that layer is marked as retryable, which is why the test in question passes in the Go Driver. Am I missing something in the CMAP spec?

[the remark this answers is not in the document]
prestonvasquez: This isn't a network error that occurs during a handshake, it's a network error encountered when trying to drain data from an established connection.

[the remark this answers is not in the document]
ShaneHarvey: Correct but my point is that it is the same case in spirit. We can't introduce this new error mode without making is retrtyable.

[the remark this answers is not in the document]
prestonvasquez: I've updated the retryable reads and writes specifications to retry for network errors when checking out a connection.

source/retryable-writes/retryable-writes.md
```diff
@@ -176,6 +176,9 @@ The RetryableWriteError label might be added to an error in a variety of ways:
     RetryableWriteError label to that error if the MongoClient performing the operation has the retryWrites
     configuration option set to true.
 
+- When the driver encounters a network error checking out a connection, it MUST add a RetryableWriteError label to that 
+    error if the MongoClient performing the operation has the retryWrites configuration option set to true.
```
ShaneHarvey: We should mention a specific case of failing to read a pending response here.

source/retryable-reads/retryable-reads.md
```diff
@@ -53,8 +53,9 @@ An error is considered retryable if it meets any of the following criteria:
 | SocketException                    | 9001           |
 
 - a [PoolClearedError](../connection-monitoring-and-pooling/connection-monitoring-and-pooling.md#connection-pool-errors)
-- Any of the above retryable errors that occur during a connection handshake (including the authentication step). For
-    example, a network error or ShutdownInProgress error encountered when running the hello or saslContinue commands.
+- Any of the above retryable errors that occur during a connection handshake (including the authentication step) or
+    connection checkout. For example, a network error or ShutdownInProgress error encountered when running the hello or
```
ShaneHarvey: Same here.

source/client-side-operations-timeout/tests/pending-response.yml
```diff
@@ -0,0 +1,411 @@
...
+      database: *database
+      collectionName: *collection
+initialData:
+  - collectionName: *collection
+    databaseName: *database
+    documents: []
+tests:
+  # Attempting a pending response read on a non-timeout operation that can  
+  # immediately read from the TCP buffer should complete the pending read and  
+  # the connection should be checked out.
```
baileympearson: > Attempting a pending response read on a non-timeout operation that can immediately read from the TCP buffer

Are we missing a `sleep(25ms)` in this test to ensure that the `find` operation runs after the insert has finished blocking on the server and responded to the client?  Or is there a mechanism I'm not noticing in this test that ensures that the checkout request for `find` doesn't need to wait for the read?

source/client-side-operations-timeout/tests/pending-response.yml
```diff
@@ -0,0 +1,411 @@
...
+          - connectionCheckedInEvent: {} # Ping finishes.
+          - connectionCheckedOutEvent: {}
+          - connectionCheckedInEvent: {} # Insert fails.
+          - connectionPendingResponseStarted: {} # Pending read fails on first find
+          - connectionPendingResponseFailed:
+              reason: error
+          - connectionClosedEvent:
+              reason: error
+          - connectionCheckedOutEvent: {}
+          - connectionCheckedInEvent: {} # Find finishes.
```
baileympearson: Could we add server selection events to this list (maybe we could use logging tests for this, even though not all drivers have implemented the CLAM spec?)?  It would be nice to clarify that the retry happens because the error returned is retryable and we use the existing retry mechanism, not that we retry checkout directly.  Basically:

```
- connectionPendingResponseStarted: {} # Pending read fails on first find
          - connectionPendingResponseFailed:
              reason: error
          - connectionClosedEvent:
              reason: error
          - serverSelectionStarted
          - serverSelectionFinished
          - connectionCheckedOutEvent: {}
          - connectionCheckedInEvent: {} # Find finishes.
```

source/connection-monitoring-and-pooling/tests/README.md
```diff
@@ -21,6 +21,51 @@ The following tests have not yet been automated, but MUST still be tested:
...
 
+### Pending Response: Connection Aliveness Tests
+
+If a connection with a pending response is idle for > 3 seconds, then drivers are expected to perform an aliveness check
+by attempting a non-blocking read of 1 byte from the inbound TCP buffer. The following two cases test both a successful
+read and a failed one.
+
+#### Connection Aliveness Check Fails
+
+1. Initialize a mock TCP listener to simulate the server-side behavior. The listener should write at least 5 bytes to
```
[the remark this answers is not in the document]
prestonvasquez: CC: @baileympearson could you POC what it would take to create a TCP listener to perform a round trip that holds 1 byte on the write side.

source/retryable-writes/retryable-writes.md
```diff
@@ -176,6 +176,11 @@ The RetryableWriteError label might be added to an error in a variety of ways:
     RetryableWriteError label to that error if the MongoClient performing the operation has the retryWrites
     configuration option set to true.
 
+- When the driver encounters a network error checking out a connection, it MUST add a RetryableWriteError label to that
+    error if the MongoClient performing the operation has the retryWrites configuration option set to true. For example,
+    a network error encountered when checking out a connection that must attempt to discard a pending response from the
```
ShaneHarvey: > a network error encountered when checking out a connection that must attempt to discard a pending response

Is this sentence correct? I'm getting tripped up by the "that must attempt". Should it be
> a network error encountered when reading a pending response during connection checkout.

baileympearson (changes requested):

source/client-side-operations-timeout/tests/pending-response.yml
```diff
@@ -0,0 +1,411 @@
...
+          document: {_id: 3, x: 1}
+        expectError:
+          isTimeoutError: true
+      # Execute a subsequent operation to complete the read.  
+      - name: findOne
+        object: *collection
+        arguments:
+          filter: {_id: 1}
+        expectError:
+          isTimeoutError: true
```
baileympearson: This test fails in Node because `isTimeoutError` only applies when `timeoutMS` is configured:

```
isTimeoutError: Optional boolean. If true, the test runner MUST assert that the error represents a timeout due to use of the timeoutMS option. If false, the test runner MUST assert that the error does not represent a timeout.
```

Maybe splitting hairs here but in Node, we only throw a CSOT timeout error when timeoutMS is enabled for the operation.  Because the `findOne` doesn't have timeoutMS configured, even though the timeout is caused by a CSOT-enabled command, the `findOne` itself doesn't throw a CSOT timeout error.

Zooming out further from just the test runner - what do you think the best solution would be for this situation?  Should Node just throw a CSOT timeout error in this circumstance?

source/client-side-operations-timeout/tests/pending-response.yml
prestonvasquez: Since this will require using logging instead of event monitoring, suggest we defer to DRIVERS-3216 to avoid adding additional complexity in this PR.

source/client-side-operations-timeout/tests/pending-response.yml
prestonvasquez: This test is actually incorrectly designed and the findOne should retry. I've updated the the yml, see comments for additional context.

source/retryable-reads/retryable-reads.md
```diff
@@ -53,8 +53,9 @@ An error is considered retryable if it meets any of the following criteria:
 | SocketException                    | 9001           |
 
 - a [PoolClearedError](../connection-monitoring-and-pooling/connection-monitoring-and-pooling.md#connection-pool-errors)
-- Any of the above retryable errors that occur during a connection handshake (including the authentication step). For
-    example, a network error or ShutdownInProgress error encountered when running the hello or saslContinue commands.
+- Any of the above retryable errors that occur during a connection handshake (including the authentication step) or
```
isabelatkinson: Are drivers without CSOT (currently Rust and C) expected to implement these changes in their retryability behavior as part of this ticket? If so, I think we should add retryability-only tests in this PR or prioritize DRIVERS-3216.

prestonvasquez: @isabelatkinson I think this update would only be relevant to driver that implement CSOT. Otherwise, the only I/O that occurs during checkout is the connection handshake which have tests prescribed in DRIVERS-746.

For context, this ticket introduces I/O to CMAP to both "drain" the TCP buffer and perform aliveness checks for connections pending response data. The pathological case is the aliveness check which will return an error but preserve the connection if successful and so should retry at the operation layer.

isabelatkinson (approved):
retryability spec changes LGTM. Rust does not have CSOT, so I'm not able to POC/test

source/client-side-operations-timeout/tests/pending-response.yml
```diff
@@ -0,0 +1,415 @@
...
+      # Execute operation with timeout less than block time.
+      - name: insertOne
+        object: *collection
+        arguments:
+          timeoutMS: 75
+          document: {_id: 3, x: 1}
+        expectError:
+          isTimeoutError: true
+      # Execute a subsequent operation to complete the read.
+      - name: findOne
```
sanych-sun: `findOne` operation is optional and not all drivers implement it (we do not support it in CSharp Driver for example). Can we replace its usage with `find`?

sanych-sun (changes requested):

source/connection-monitoring-and-pooling/tests/logging/connection-logging-csot.yml
```diff
@@ -0,0 +1,246 @@
...
+        object: *collection
+        arguments:
+          timeoutMS: 50
+          document: { _id: 3, x: 1 }
+        expectError:
+          isTimeoutError: true
+
+      # Execute a subsequent operation which should time out during the
+      # pending response read attempt.
+      - name: findOne
```
sanych-sun: Please use `find` here instead of `findOne`.

source/connection-monitoring-and-pooling/tests/logging/connection-logging-csot.yml
```diff
@@ -0,0 +1,246 @@
...
+
+          - level: debug
+            component: connection
+            data:
+              message: "Pending response failed"
+              serverHost: { $$type: string }
+              serverPort: { $$type: [int, long] }
+              driverConnectionId: { $$type: [int, long] }
+              requestId: { $$type: [int, long] }
+              reason: "timeout"
```
sanych-sun: Shouldn't we expect `Connection checkout failed` event here? To signal that previous checkout attempt failed.

prestonvasquez: @ShaneHarvey Can you opine on this? In the original review it was decided to mute `ConnectionCheckOutFailed` when draining a connection during the check out process. However, this wasn't mentioned in the scope or in the documentation updates for CMAP on this PR. Should the `ConnectionCheckOutFailed` event be propagated when we return an error to the operation layer after attempting to drain a pending response?

CC: @stIncMale

sanych-sun: From the logs reader prospective: if one need to investigate why some operations takes longer then usual, the logical path to investigate the problem would be go from operation level to more detailed levels IN BULK. So first thingy to check would be operation logs, then server selection (if it takes longer then usual, was there any errors), then connection checkout (the duration of checkout, was there some errors)... and if we DO NOT raise checkout failed event - then the fact of failed checkouts because of pending reads failure might be missed until one decide to investigate step by step a particular operation and will see the fact that checkout was started, but never completed (as per logs). Such behavior will be really confusing.
More over: if we decide to "reduce a noise in logs", then why do we reporting success twice: `Pending response succeeded` and then `Connection checked out`.

From my understanding we should either report 2 sets of "started" and "succeeded"/"failed" events, or report the only event for pending reads "started" and imply that depending on pending reads results we will report "Connection checked out" or "Connection checkout failed".

ShaneHarvey: Yeah a CheckoutFailed is required here since a CheckoutStarted always needs a corresponding CheckoutFailed or CheckoutSucceeded event.

prestonvasquez: Sounds good, will try to get an update in ASAP. Thanks @ShaneHarvey

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -997,6 +1120,96 @@ interface ConnectionCheckedInEvent {
...
   connectionId: int64;
 }
+
+/**
+ *  Emitted when the connection being checked out is attempting to read and
+ *  discard a pending server response.
+ */
+interface PendingResponseStarted {
+  /**
+   *  The ServerAddress of the Endpoint the pool is attempting to connect to.
```
stIncMale: Looks like this description was copied from another event, but it can't be the correct description of what `PendingResponseStarted.address` is: when a pool is attempting to read a pending server response, it does so using a connection that has already been established.

The same applies to `PendingResponseSucceeded`, `PendingResponseFailed`.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -997,6 +1120,96 @@ interface ConnectionCheckedInEvent {
...
+
+  /**
+   *  The ID of the Connection.
+   */
+  connectionId: int64;
+
+  /**
+   *  The driver-generated request ID associated with the network timeout.
+   */
+  requestId: int64;
```
stIncMale: We already have `CommandStartedEvent`/`CommandSucceededEvent`/`CommandFailedEvent.requestId` specified in https://github.com/mongodb/specifications/blob/master/source/command-logging-and-monitoring/command-logging-and-monitoring.md#events-api.

1. I am guessing that `PendingResponseStarted`/`PendingResponseSucceeded`/`PendingResponseFailed.requestId` represent the same thing - the ID of a command. If this is the case, the specification should explicitly point that out, instead of expecting that a reader will somehow guess this (to guess this correctly it is required but not sufficient to know and remember that there is `CommandStartedEvent`/`CommandSucceededEvent`/`CommandFailedEvent.requestId`).
2. Assuming that the above guess is correct, which command must `PendingResponseStarted`/`PendingResponseSucceeded`/`PendingResponseFailed.requestId` correspond to: the one that failed with a network timeout and left a response not completely read from the connection, or the one for which the connection is being checked out? In a way, they are both "associated" with that timeout, and the specification should make it clear which one it talks about.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -997,6 +1120,96 @@ interface ConnectionCheckedInEvent {
...
+}
+
+/**
+ *  Emitted when the connection successfully read the pending read and is ready
```
stIncMale: I suspect, "response" was meant to be used here:

```suggestion
 *  Emitted when the connection successfully read the pending response and is ready
```

But if not, then what does it mean to "read the pending read"?

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -576,7 +579,122 @@ other threads from checking out [Connections](#connection) while establishing a
...
+    connection to a "pending response" state and MUST instead immediately close the connection.
+
+The next time connection in the "pending response" state is checked out, the driver MUST ensure that any remaining
+response data is drained and discarded either by explicit reads or, in push-based I/O implementations (e.g. Node.JS), by
+consuming buffered data.
+
+1. **Persist and update timestamp**: The connection must record the current time immediately after the original socket
+    timeout. This timestamp MUST be updated to the current time whenever any bytes are successfully read, received, or
+    consumed while explicitly awaiting the pending response as part of checking out the connection.
+2. **Aliveness check**: If the undrained connection remains idle (i.e. no data is read or received) for more than 3
```
stIncMale: Is "undrained connection" different from a connection in the "pending response" state? If yes, then the specification should explain what "undrained connection" is. If there is no difference, then the specification should use only one term consistently.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -217,17 +217,20 @@ interface Connection {
...
    *
    * Possible values are the following:
-   *   - "pending":       The Connection has been created but has not yet been established. Contributes to
-   *                      totalConnectionCount and pendingConnectionCount.
+   *   - "pending":           The Connection has been created but has not yet been established. Contributes to
+   *                          totalConnectionCount and pendingConnectionCount.
    *
-   *   - "available":     The Connection has been established and is waiting in the pool to be checked
-   *                      out. Contributes to both totalConnectionCount and availableConnectionCount.
+   *   - "pending response":  The Connection is attempting to discard a response for an operation where the socket timed
```
stIncMale: Here, the state is described as _"Connection is attempting to **discard** a response"_. Not "read" (which is what I would have expected to be mentioned), not "drain", not "consume", not "receive", not "execute" but only "discard". In other places, however, all of those words are used, and even combinations of them:

- _read/drain operation_
- _drained and discarded_ vs _drained and successfully discarded_ vs _read and discard_
- _"pending response" drain_
- _bytes are successfully read, received, or consumed_
- _reading from the socket (or draining buffered data)_
- _data is drained and discarded either by explicit reads or, in push-based I/O implementations (e.g. Node.JS), by
consuming buffered data._ - The specification does not rely on this distinction, nor can I actually see a meaningful distinction, as any practical TCP implementation has to buffer inbound data one way or another. If a single term, like "reading", is deemed unclear by maintainers of different drivers, let's specify in one place that for the purpose of the specification the term "[pick one term]" is going to be used to denote [explain the meaning taking into account all the implementations quirks that are deemed necessary], and then use the single picked term consistently.
- _draining buffered data_ vs _consuming buffered data_
- _execute_pending_response_
- `source/connection-monitoring-and-pooling/tests/README.md` also uses different words in different places, and it is unclear whether they mean the same or not
  - _drain the rest of the response_
  - _discard bytes from the TCP stream_


Are all these terms meaningfully different? If yes, they should be defined clearly, and used strictly according to the definition. If not, then the specification should use a single word to refer to the same thing.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -576,7 +579,122 @@ other threads from checking out [Connections](#connection) while establishing a
...
+- **Exhaust commands**: If the command was sent with the `exhaustAllowed` bit, the driver MUST NOT transition the
+    connection to a "pending response" state and MUST instead immediately close the connection.
+
+The next time connection in the "pending response" state is checked out, the driver MUST ensure that any remaining
+response data is drained and discarded either by explicit reads or, in push-based I/O implementations (e.g. Node.JS), by
+consuming buffered data.
+
+1. **Persist and update timestamp**: The connection must record the current time immediately after the original socket
+    timeout. This timestamp MUST be updated to the current time whenever any bytes are successfully read, received, or
+    consumed while explicitly awaiting the pending response as part of checking out the connection.
```
stIncMale: What is the difference between _"awaiting a pending response"_ used previously and "_**explicitly** awaiting the pending response_" used here?

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -576,7 +579,122 @@ other threads from checking out [Connections](#connection) while establishing a
 Before a given [Connection](#connection) is returned from checkOut, it must be marked as "in use", and the pool's
 availableConnectionCount MUST be decremented.
 
-```text
+##### Awaiting Pending Read (drivers that support CSOT)
```
stIncMale: The specification introduces a new "pending response" connection state. But then there is this "Awaiting Pending Read" section, which uses the "pending read" term exactly once, and mostly uses "pending response" instead. The "Events" section, on the other hand, uses "pending read".

What is the difference between the meaning of "pending response" and "pending read"?

stIncMale (changes requested):
I started reading the PR to prepare for triaging https://jira.mongodb.org/browse/DRIVERS-3276, and left some comments. These comments do not represent a result of a full review, nor do I know if I will do such a review.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -633,6 +751,11 @@ if connection state is "pending":
     decrement pendingConnectionCount
 else:
     decrement availableConnectionCount
+
+error = await_pending_response(pool, connection)
```
stIncMale: `await_pending_response` accepts `timeout` and `conn`, based on its pseudocode. But here `pool` is passed instead of `timeout`. I don't think this is correct, especially given that `timeout` cannot be extracted from `pool`.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -997,6 +1120,96 @@ interface ConnectionCheckedInEvent {
    */
   connectionId: int64;
 }
+
+/**
+ *  Emitted when the connection being checked out is attempting to read and
+ *  discard a pending server response.
+ */
+interface PendingResponseStarted {
```
stIncMale: We have pseudocode showing when most other pool events should be emitted. Furthermore, we have pseudocode showing how `duration` for `ConnectionReadyEvent`, `ConnectionCheckOutFailedEvent`, `ConnectionCheckedOutEvent` should be computed.

Let's do the same for the new `PendingResponseStarted`, `PendingResponseSucceeded`, `PendingResponseFailed` events.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -576,7 +579,122 @@ other threads from checking out [Connections](#connection) while establishing a
...
+    second "pending response" window and (b) the `socketTimeoutMS` (if supported by the driver) as the effective
+    timeout for the read/drain operation.
+5. **Error or over-age**: If reading from the socket (or draining buffered data) results in an error that is not a
+    timeout, or if the connection exceeds the 3 second pending-response window, the driver MUST close the connection.
+6. **Clear pending state on success**: If the pending response is fully drained and successfully discarded, and the
+    connection remains healthy, the pending state may be cleared and the connection MAY be returned to the pool for
+    reuse.
+
+```mermaid
+sequenceDiagram
```
stIncMale: The diagram suggests that a connection in the pending response state can be checked out from a pool, and that reading a response which wasn't read in full can be done after the connection having been checked out. Both of these pieces of behavior contradict the pseudocode and the [design](https://docs.google.com/document/d/1XopiRlzjxw23_ZUoyC7mthiepa8BUkBRMYN6Xgu6dng/edit?tab=t.0).

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
stIncMale: There seem to be issues with this pseudocode and the way the "pending response" state is defined:

1. `await_pending_response` must be called if and only if `connection is "pending response"`, which is not expressed in the pseudocode.
2. The only state the `connection` can be at this point is "available", so `await_pending_response` can never be called.
3. The specification instructs to maintain `availableConnectionCount`, `pendingConnectionCount`, `totalConnectionCount` with the invariant being `totalConnectionCount = pendingConnectionCount + availableConnectionCount + <in use connection count, which is not explicitly maintained>` (search for `"pending" + "available" + "in use"`). While I don't know why `availableConnectionCount` is maintained (maybe we should figure this out), the introduction of the new "pending response" state affects the aforementioned, and has to be properly dealt with.
4. I haven't checked everything else, but given that currently it looks like the new state was slapped on top of the spec without much regard to the rest of the spec, I would not be surprised if there were more places that have to be adjusted to take into account the new "pending response" state. **Update**: I realized that there is at least one more place (please do look for more). "Checking In a Connection" currently either closes a connection or moves it to the "available" state. It should move the connection in either "available" or "pending response" state depending on what transpired before it being checked in. So this section needs a change.

I suspect that handling of connections that are in the the "pending response" state should be done at the same point where perished connections are handled. Note that being perished is not considered to be a state of a connection by the spec, but merely a value of the perishable property of a connection (both "value" and "property" are not used here in the same sense they are used in programming; that's my best interpretation of the spec as it is now).

We also may consider making "pending response" not a new state of a connection, but rather a value of another property, similarly to how it is done with the "perishable" property and its "perished"/"non-perished" values.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
stIncMale: Something is wrong with the structure of the spec here:

1. Instead of updating the "Checking Out a Connection" section to integrate the new logic into the checking out logic, the new "Awaiting Pending Read (drivers that support CSOT)" subsection was added without any integration. One can understand what this subsection means and how it is supposed to be integrated in the checking out logic only after looking at the pseudocode. But pseudocode is supposed to supplement the prose of the spec, not be a replacement for it. It seems to me that the new logic should be properly described as part of the checking out logic.
2. The "Awaiting Pending Read (drivers that support CSOT)" subsection also specifies how an "in use" connection transitions into the "pending response" state. This part has no relation to "Checking Out a Connection", and should not be inside that section.

See also https://github.com/mongodb/specifications/pull/1675#discussion_r2354273640.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -576,7 +579,122 @@ other threads from checking out [Connections](#connection) while establishing a
...
+    1. Put the connection into a "pending response" state.
+    2. Record the current time in a way that can be updated while awaiting a pending response.
+- **Exhaust commands**: If the command was sent with the `exhaustAllowed` bit, the driver MUST NOT transition the
+    connection to a "pending response" state and MUST instead immediately close the connection.
+
+The next time connection in the "pending response" state is checked out, the driver MUST ensure that any remaining
+response data is drained and discarded either by explicit reads or, in push-based I/O implementations (e.g. Node.JS), by
+consuming buffered data.
+
+1. **Persist and update timestamp**: The connection must record the current time immediately after the original socket
```
stIncMale: This enumerated list does not seem to be introduced in any way, or related to anything. It just exists out there, specifying six instructions. This concern is essentially part of the concern [expressed above](https://github.com/mongodb/specifications/pull/1675#discussion_r2354268167), and should be addressed as part of that one.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -576,7 +579,122 @@ other threads from checking out [Connections](#connection) while establishing a
...
+    timeout. This timestamp MUST be updated to the current time whenever any bytes are successfully read, received, or
+    consumed while explicitly awaiting the pending response as part of checking out the connection.
+2. **Aliveness check**: If the undrained connection remains idle (i.e. no data is read or received) for more than 3
+    seconds since the start of the "pending response" state or since the last successful read/receive, the driver MUST
+    attempt to verify the connection’s health by either performing a non-blocking read or using the minimal possible
+    timeout to check if at least one byte can be read/received. If at least one byte can be read the connection should
+    be returned to the pool for reuse and a retryable error should be propagated to the operation layer. If no bytes
```
stIncMale: This item and also item 6 below cannot say _"returned to the pool"_ when talking about a connection that is being checked out, because such a connection is still in the pool, and the `ConnectionCheckedOutEvent` hasn't been emitted for it.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -576,7 +579,122 @@ other threads from checking out [Connections](#connection) while establishing a
...
+3. **User-provided timeout**: If a user-provided timeout is specified for the "pending response" drain, the driver MUST
+    use the minimum of (a) the remaining time before the 3 second "pending response" window elapses and (b) the
+    user-provided timeout as the effective timeout for the read/drain operation.
+4. **Default timeout**: If no user-provided timeout is specified, the driver MUST use the minimum of (a) the remaining 3
+    second "pending response" window and (b) the `socketTimeoutMS` (if supported by the driver) as the effective
+    timeout for the read/drain operation.
+5. **Error or over-age**: If reading from the socket (or draining buffered data) results in an error that is not a
+    timeout, or if the connection exceeds the 3 second pending-response window, the driver MUST close the connection.
+6. **Clear pending state on success**: If the pending response is fully drained and successfully discarded, and the
+    connection remains healthy, the pending state may be cleared and the connection MAY be returned to the pool for
```
stIncMale: According to the pseudocode, if `await_pending_response` completes without an error, the checking out succeeds for the connection: the connection transitions into the "in use" state, the `ConnectionCheckedOutEvent` is emitted for it. This happens always. I fail to see how _"MAY be returned to the pool"_ makes sense both because of what I have just described, and because of https://github.com/mongodb/specifications/pull/1675#discussion_r2354283136.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -576,7 +579,122 @@ other threads from checking out [Connections](#connection) while establishing a
...
+    if len(data) > 0:
+        # Refresh the remaining time upon a successful read
+        conn.pending_start = end_time
+
+    # Check if the remaining time has been exceeded
+    if end_time - conn.pending_start >= PENDING_RESPONSE_TIMEOUT_MS:
+        close_connection(conn)
+
+    if error is not None:
+        raise error
```
stIncMale: The pseudocode that existed before the current PR used `throw`, for example, `throw PoolClosedError`. The new pseudocode should continue using that, instead of coming up with new syntax.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -633,6 +751,11 @@ if connection state is "pending":
     decrement pendingConnectionCount
 else:
     decrement availableConnectionCount
+
+error = await_pending_response(pool, connection)
+if error:
+  return error
```
stIncMale: The pseudocode that existed before the current PR used `throw`, for example, `throw PoolClosedError`. The new pseudocode should continue using that, instead of coming up with new syntax.

What is worse, is that the new pseudocode uses `return` here, but `raise` is two other places. That is, it is not even consistent within itself.

stIncMale: We should fix the pseudocode by emitting the `ConnectionCheckOutFailedEvent` before throwing this `error`. This issue was identified in https://github.com/mongodb/specifications/pull/1675#discussion_r2337951642, I am just pointing out to at least one place where the change has to be done.

stIncMale (changes requested):

stIncMale:
@baileympearson and I triaged https://jira.mongodb.org/browse/DRIVERS-3276 today. We disagree on what to do with that ticket. I think, the work on that ticket should be done as part of the work done in this PR. I expressed more of my thoughts in this [Jira comment](https://jira.mongodb.org/browse/DRIVERS-3276?focusedCommentId=7681911&focusedId=7681911&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-7681911).

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -576,7 +579,122 @@ other threads from checking out [Connections](#connection) while establishing a
...
+    attempt to verify the connection’s health by either performing a non-blocking read or using the minimal possible
+    timeout to check if at least one byte can be read/received. If at least one byte can be read the connection should
+    be returned to the pool for reuse and a retryable error should be propagated to the operation layer. If no bytes
+    can be read, the connection MUST be closed.
+3. **User-provided timeout**: If a user-provided timeout is specified for the "pending response" drain, the driver MUST
+    use the minimum of (a) the remaining time before the 3 second "pending response" window elapses and (b) the
+    user-provided timeout as the effective timeout for the read/drain operation.
+4. **Default timeout**: If no user-provided timeout is specified, the driver MUST use the minimum of (a) the remaining 3
+    second "pending response" window and (b) the `socketTimeoutMS` (if supported by the driver) as the effective
+    timeout for the read/drain operation.
```
stIncMale: @sanych-sun, thank you for [pointing out](https://jira.mongodb.org/browse/DRIVERS-3276?focusedCommentId=7690144&focusedId=7690144&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-7690144) that the CMAP spec has `ConnectionPoolOptions.waitQueueTimeoutMS`. If we don't take it into account here (and in the pseudocode), then we are changing its meaning, which may result in surprising behavior from the perspective of users: currently, checking out for a non-CSOT operation is expected to potentially go over `waitQueueTimeoutMS` only if<sup>(*)</sup> a new connection is created and established as part of checking out; otherwise, the duration of checking out is expected to be within `waitQueueTimeoutMS` (see the documentation for `ConnectionCheckedOutEvent.duration` and `waitQueueTimeoutMS`).

Thus, the timeout for draining should not exceed what is left of `waitQueueTimeoutMS`, and should not exceed the _"remaining 3 second "pending response" window"_. I am unsure if `socketTimeoutMS` needs to be involved at all. I remember proposing `socketTimeoutMS` when I was reviewing the design, because `waitQueueTimeoutMS` did not even come to my mind at that time, but I am not certain about my recollection.

---

<sup>(*)</sup> A driver not providing hard real-time guarantees is irrelevant for the purpose of the current comment, which is why I said "only if".

PR #1678, opened 2024-10-22
DRIVERS-3010 reformat specs for improved mkdocs rendering

mdbmes:
This patch intends to make specs more readable when rendered by the included mkdocs build system, such as when they're displayed on https://specifications.readthedocs.io

The main issue behind DRIVERS-3010 is that various Markdown implementations disagree about how much indentation is necessary for items within a list. The original Markdown spec requires 4 spaces, and mkdocs is strict about this. Other implementations, like GitHub flavored markdown, allow 2 or 3 spaces.

The main fix is to begin using the `mdformat-mkdocs` plugin for mdformat, which automatically adjusts this indentation.

Existing documents have been reformatted with an override that allows the generated HTML to change (because it was incorrect before) but future automatic invocations of mdformat will check that any markdown changes don't affect rendered output.

Along the way, I've been finding semi-related formatting problems that were also causing the mkdocs output to be incorrect or hard to read, this fixes those too:

- Some incorrect indentations were causing lists to be renumbered unexpectedly. Note that prior to this change, list renumbering would only be visible in the rendered HTML, but the `mdformat-mkdocs` plugin renumbers lists according to how they are parsed. List numbering problems will now be visible in the Markdown source. 
- The client side encryption spec relies on math mode, which was unsupported. Now it's rendered in browser using MathJax.
- Other specs used the $ or [] characters in ways which could unintentionally enter math mode.
- Indented code blocks could not be rendered by mkdocs without an extension.
- Escaped angle brackets, generated automatically by mdformat, would not render properly without an extension.
- Lists and tables were sometimes hard to read due to CSS style.

This change adds two new external dependencies at build time, imported via pip:
- https://github.com/KyleKing/mdformat-mkdocs
- https://github.com/facelessuser/pymdown-extensions

And one new JavaScript library at runtime, via jsdelivr CDN:
- https://github.com/mathjax/MathJax-src


blink1073 (approved):
Structural changes LGTM!  I spot-checked the markdown file changes.  😅

kevinAlbs (approved):
LGTM with a minor suggestion. Spot checked rendering of files under `source/client-side-encryption`.

source/client-side-encryption/tests/README.md
```diff
@@ -644,108 +651,109 @@ binary subtype 4 (or standard UUID), which MUST be decoded and encoded as subtyp
...
-   The method of passing TLS options for KMIP TLS connections is driver dependent.
+    The method of passing TLS options for KMIP TLS connections is driver dependent.
 
-   Where LOCAL_MASTERKEY is the following base64:
+    Where LOCAL_MASTERKEY is the following base64:
 
-   ```javascript
-   Mng0NCt4ZHVUYUJCa1kxNkVyNUR1QURhZ2h2UzR2d2RrZzh0cFBwM3R6NmdWMDFBMUN3YkQ5aXRRMkhGRGdQV09wOGVNYUMxT2k3NjZKelhaQmRCZGJkTXVyZG9uSjFk
-   ```
+    ```javascript
```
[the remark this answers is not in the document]
mdbmes: Thanks for the review, and for spotting the additional javascript-tagged base64 code.

alcaeus (approved):
Reviewed the changes, did not spot any changes I wouldn't expect to see.

Note that I disabled whitespace-only changes in the diff, but instead focused on changes that also changed contents (e.g. changing line wrapping due to the indentation).

source/server-discovery-and-monitoring/server-discovery-and-monitoring.md
W-A-James, line 1: superscript cross reference (`[^1]`) in Round Trip Time section not rendering correctly

W-A-James, line 1: Superscript cross reference ([^2]) in TopologyType table not rendering correctly

W-A-James (approved):
SDAM spec docs LGTM

mdbmes:
The remaining pre-commit failure is an mdformat difference due to changes in master since this PR's branch point.
I'll either need to rebase this PR or fix that in a followup change.

mdbmes:
> The remaining pre-commit failure is an mdformat difference due to changes in master since this PR's branch point. I'll either need to rebase this PR or fix that in a followup change.

Never mind, I didn't realize we could just merge from main back to this branch and let squash-and-merge eat the extra commit.

PR #1687, opened 2024-10-28
DRIVERS-1368: Use null when options shouldn't be asserted against

alcaeus:

- [ ] Update changelog.
- [x] Test changes in at least one language driver (tested in libmongoc)
- [ ] ~Test these changes against all server versions and topologies (including standalone, replica set, sharded
    clusters, and serverless).~

This PR updates URI options tests where the URI is invalid or warnings are expected, and the options are unpredictable. In this case, we use `null` to indicate that the options should not be asserted against.

### DIFF
--- a/source/uri-options/tests/compression-options.json
+++ b/source/uri-options/tests/compression-options.json
@@ -35,7 +35,7 @@
       "warning": true,
       "hosts": null,
       "auth": null,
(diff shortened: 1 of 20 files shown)


JamesKovacs (approved):
LGTM

PR #1688, opened 2024-10-28
DRIVERS-1408 - Add guidance on adding _id fields to documents to CRUD spec

NoahStapp:
Python implementation: https://github.com/mongodb/mongo-python-driver/pull/1976

The new prose test only verifies that drivers correctly implement prepending `_id` for `insert_one`. Is it worthwhile to verify every other insert operation (`insert_many`, `bulk_write`, `client_bulk_write`) as well?


source/crud/tests/README.md
```diff
@@ -737,3 +737,10 @@ that `firstEvent.operationId` is equal to `secondEvent.operationId`. Assert both
 To force completion of the `w:0` writes, execute `coll.countDocuments` and expect the returned count is
 `maxMessageSizeBytes / maxBsonObjectSize + 1`. This is intended to avoid incomplete writes interfering with other tests
 that may use this collection.
+
+### 16. Generated document identifiers are the first field in their document
+
+Construct a `MongoClient` (referred to as `client`) with
+[command monitoring](../../command-logging-and-monitoring/command-logging-and-monitoring.md) enabled to observe
+CommandStartedEvents. Perform an `insert` command using `client` and assert that one CommandStartedEvent (referred to as
+`event`) was observed, and that the first field of `event.command.documents[0]` is `_id`.
```
nbbeeken, line 746: Yea I think we should list out the other APIs to check this for as well.

Should be: insertOne, insertMany, updateOne, updateMany, replaceOne, legacy bulkWrite and client bulkWrite.

nbbeeken (changes requested):

source/crud/crud.md
```diff
@@ -815,6 +815,12 @@ database-level aggregation will allow users to receive a cursor from these colle
 
 ##### Insert, Update, Replace, Delete, and Bulk Writes
 
+###### Generated identifiers
+
+The insert and bulk insert operations described below MUST generate identifiers for all documents that do not already
+have them. These identifiers SHOULD be prepended to the document so they are the first field, in order to prevent the
+server from spending time re-ordering the document.
```
nbbeeken, line 822: It is implied, but may be we should make it abundantly clear, a driver should _not_ re-order an _id if it is already present right?

I would say yes, but also not sure if it matters and it may be good to do. If the goal is performance then we should weigh re-ordering vs the cost of leaving the _id out of place 🤔

source/crud/tests/README.md
nbbeeken: See my follow up, because some of these APIs need the user to supply the _id the driver does not generate it (update)

source/crud/crud.md
NoahStapp: The cost of re-ordering on the client-side is language-dependent. I'm not well-versed in how performant the server's re-ordering operation is, but it very likely could be faster than many drivers.

source/crud/tests/README.md
NoahStapp: If an API requires the user to supply `_id`, it should be excluded from this test. Maybe we should have a second test to verify that existing `_id` fields behave as expected, depending on if we decide to re-order or not?

source/crud/crud.md
jmikola: With respect to DRIVERS-1408, I don't see any reason to add re-ordering logic for user-supplied documents that already contain an `_id`. Just ensure that if the driver is generating an ID, the field gets prepended.

source/crud/crud.md
```diff
@@ -815,6 +815,12 @@ database-level aggregation will allow users to receive a cursor from these colle
 
 ##### Insert, Update, Replace, Delete, and Bulk Writes
 
+###### Generated identifiers
+
+The insert and bulk insert operations described below MUST generate identifiers for all documents that do not already
```
jmikola, line 820: Noted that this is already discussed in the [client bulk write spec](https://github.com/mongodb/specifications/blob/master/source/crud/bulk-write.md#insert).

source/crud/tests/README.md
```diff
@@ -737,3 +737,10 @@ that `firstEvent.operationId` is equal to `secondEvent.operationId`. Assert both
 To force completion of the `w:0` writes, execute `coll.countDocuments` and expect the returned count is
 `maxMessageSizeBytes / maxBsonObjectSize + 1`. This is intended to avoid incomplete writes interfering with other tests
 that may use this collection.
+
+### 16. Generated document identifiers are the first field in their document
+
+Construct a `MongoClient` (referred to as `client`) with
+[command monitoring](../../command-logging-and-monitoring/command-logging-and-monitoring.md) enabled to observe
+CommandStartedEvents. Perform an `insert` command using `client` and assert that one CommandStartedEvent (referred to as
```
jmikola, line 745: > Perform an `insert` command

This is ambiguous, as the prose test could be implementing using the generic command runner, in which case I'd expect no ID generation. I'd suggest changing this to explicitly suggesting running an `insertOne` operation. You can extend that with `insertMany` and a `bulkWrite` if you want full test coverage.

source/crud/tests/README.md
jmikola: @nbbeeken: I don't think any update operations are relevant here, as any generated ID would be client-side for an `upsert` operation. And in that case, we already have several spec tests that specify `_id` in a filter with `upsert: true` and assert that the client-supplied `_id` is relayed properly in the `upsertedIds` response field from the server.

See: "BulkWrite with replaceOne operations" in `bulkWrite.yml` some other tests in `client-bulkWrite-results.yml`.

source/crud/crud.md
NoahStapp: Having this behavior discussed here for general CRUD operations makes sense to me, as the client bulk write spec follows from this broader context.

jmikola: The client bulk write spec stands on its own, except for a reference back to CRUD for [modeling unacknowledged write results](https://github.com/mongodb/specifications/blob/master/source/crud/bulk-write.md#unacknowledged-results).

With my last comment, I just meant to acknowledge that no changes were needed to the client bulk write spec since it already addressed `_id` ordering. This was in the vein of previous PRs like https://github.com/mongodb/specifications/pull/1644 that required updates to _both_ specs.

source/crud/tests/README.md
```diff
@@ -737,3 +737,15 @@ that `firstEvent.operationId` is equal to `secondEvent.operationId`. Assert both
...
 that may use this collection.
+
+### 16. Generated document identifiers are the first field in their document
+
+Construct a `MongoClient` (referred to as `client`) with
+[command monitoring](../../command-logging-and-monitoring/command-logging-and-monitoring.md) enabled to observe
+CommandStartedEvents. For each of `insertOne`, client `bulkWrite`, and collection `bulkWrite`, do the following:
+
+- Execute the command with a document that does not contain an `_id` field.
+- If possible, capture the wire protocol message (referred to as `request`) of the command.
```
jmikola, line 748: Is this only necessary in languages where the ordering is still not deterministic in the CommandStartedEvent's `command` document?

source/crud/tests/README.md
```diff
@@ -737,3 +737,15 @@ that `firstEvent.operationId` is equal to `secondEvent.operationId`. Assert both
 To force completion of the `w:0` writes, execute `coll.countDocuments` and expect the returned count is
 `maxMessageSizeBytes / maxBsonObjectSize + 1`. This is intended to avoid incomplete writes interfering with other tests
 that may use this collection.
+
+### 16. Generated document identifiers are the first field in their document
```
alcaeus, line 741: Looking at this test, aside from asserting that `_id` is the first field in the document, I don't see a reason why this should be a prose test and can't be a unified test. Keep in mind that prose test cause significantly more downstream implementation effort than unified spec tests.

NoahStapp: The explicit purpose of the test is to assert that `_id` is the first field in the document. Unless we add the capability to check mapping order to the unified spec runner, a prose test is the only way to do this.

source/crud/crud.md
nbbeeken: I agree. I would take it further and say that drivers SHOULD NOT re-order given documents if the user has given a structure that has order. The choice to place the _id in the sub/optimal position along with any other positioning choices should be preserved. (TBH, mostly because it is useful for testing)

Maybe it could be a MUST NOT too 🤔 as long as you've been given an ordered object, no intervention should be done.

In JS the `_id` is always appended because JS objects are ordered by key insertion order so there is no way to put the _id in front. Naively we could special-case _id in our BSON library but that would indiscriminately reorder that key no matter where it is used. I am thinking ahead for my own driver here, we want a solution that _only_ fixes this for driver generated `_id` fields.

source/crud/tests/README.md
nbbeeken: I agree with Noah here, as a driver that could benefit from a key order solution in the unified spec runner, I don't think adding support for something like that is reasonable to finish within spec-fest, but paying the cost of prose for generally improving write performance wherever this advice is not followed is worth paying, in my opinion.

source/crud/tests/README.md
nbbeeken: I think it could potentially be rephrased if wire protocol parsing is difficult for drivers to achieve, but I agree it would be easy to assert `_id` is the first key in command monitoring and that not actually be the case when the wire message is produced. 

I think perhaps for this test; we can encourage drivers to just use command monitoring as the "main path" set of assertions to implement. But we should encourage drivers to check that their wire message key order either matches their command events or that reordering is done when the wire message is built. 

So for Node, I would write a test that asserts the JS object that is inspectable on the event (for any command) matches key order in BSON. Python may check that their wire messages are ordered correctly.

nbbeeken (changes requested):

source/crud/tests/README.md
nbbeeken: Agreed, learned that after I dove into writing tests for this 😅. I also wrote tests to check that my driver _does not_ reorder `_id`s provided by the user. For that second case update/replace does apply, if we feel we want that coverage as well.

source/crud/crud.md
NoahStapp: I'd leave it up to each driver on how to handle already-existing `_id` fields. The Python driver reorders user-supplied `_id` keys during the BSON conversion, but other drivers may be unable or unwilling to do the same.

source/crud/tests/README.md
NoahStapp: If drivers are able to check wire messages directly, they should take that path. Otherwise, they should fall back to using command monitoring, ideally verifying that they don't modify field ordering before sending data over the wire.

source/crud/crud.md
nbbeeken: That is fair and I wouldn't want to imply python should change that, nor do I want us to imply to drivers that do not re-order user `_id`s should do so. So I think it is fair to leave open to each language but we should capture this conversation in some form. 

- Generated _ids SHOULD be put first. 
- User _ids MAY be left as is. (hopefully that is enough to imply, "MAY be re-ordered as well"?)

source/crud/tests/README.md
nbbeeken: I agree, after all the point here is that the bytes are in an order that the server benefits from, if your command monitoring is a source of truth for such a thing you could/should use it but the real goal is in the wire message.

source/crud/tests/README.md
alcaeus: I had missed the part where this was specifically testing that `_id` is the first field in the document, god confused with all the `_id` pull requests. I agree that the prose test makes sense, especially since this assertion is not possible in the unified test format.

nbbeeken (approved):

source/crud/tests/README.md
```diff
@@ -737,3 +737,15 @@ that `firstEvent.operationId` is equal to `secondEvent.operationId`. Assert both
...
+
+Construct a `MongoClient` (referred to as `client`) with
+[command monitoring](../../command-logging-and-monitoring/command-logging-and-monitoring.md) enabled to observe
+CommandStartedEvents. For each of `insertOne`, client `bulkWrite`, and collection `bulkWrite`, do the following:
+
+- Execute the command with a document that does not contain an `_id` field.
+- If possible, capture the wire protocol message (referred to as `request`) of the command.
+- Assert that the first field of `request.documents[0]` is `_id`
+- Otherwise, capture the CommandStartedEvent (referred to as `event`) emitted by the command.
+- Assert that the first field of `event.command.documents[0]` is `_id`.
```
jmikola, line 751:
```suggestion
- If possible, capture the wire protocol message (referred to as `request`) of the command and assert that the first field of `request.documents[0]` is `_id`.
- Otherwise, capture the CommandStartedEvent (referred to as `event`) emitted by the command and assert that the first field of `event.command.documents[0]` is `_id`.
```

Optional suggestion to consolidate these bullets, since the "Assert" lines don't really stand on their own.

jmikola (approved):
LGTM w/ or w/o my suggestion.

PR #1690, opened 2024-10-28
DRIVERS-1785: clarify ordering of CMAP events when pool is cleared

baileympearson:
Node POC: https://github.com/mongodb/node-mongodb-native/pull/4296

This PR adds tests ensuring that drivers always clear the pool before checking a connection back in / emitting close events, regardless of if the pool cleared error occurs on establishment during minPoolSize population, connection checkout or during operation execution on a checked out connection.

---

This PR originally intended to just add a test that demonstrated that, when we encounter a pool clear error on an application connection, we clear the pool before checking the connection back into the pool.  This avoids a small race condition between SDAM and connection checkout for subsequent requests in the pool - the subsequent requests might check out the errored connection before the pool is cleared.

As I was looking into this, I did realize we do have this coverage already.  But as Matt pointed out on the drivers ticket, the ordering of events seemed to be specified differently in one CMAP minPoolSize test than in other CMAP and SDAM tests.  The race condition isn't present for minPoolSize population, but Matt pointed out that if we expect a different ordering of events for minPoolSize population and connection checkout, we can't share connection establishment logic in both code paths.  So I also updated the ordering of events for the incorrect CMAP test (pool clear must happen before connectionClosed events).

I also realized some tests only tested that handshakes clear the pool, while others test that authentication clears the pool.  I added tests for both errors during the handshake and auth during minPoolSize population and connection checkout, and assert on the expected ordering of events.

- [x] Update changelog.
    clusters, and serverless).

### DIFF
--- a/source/server-discovery-and-monitoring/server-discovery-and-monitoring.md
+++ b/source/server-discovery-and-monitoring/server-discovery-and-monitoring.md
@@ -1241,10 +1241,11 @@ if and only if the error is "node is shutting down" or the error originated from
 and [other transient errors](#other-transient-errors) and
 [Why close connections when a node is shutting down?](#why-close-connections-when-a-node-is-shutting-down).)
 
-##### Authentication errors
+##### Authentication and Handshake errors
 
-If the authentication handshake fails for a connection, drivers MUST mark the server Unknown and clear the server's
-connection pool if the TopologyType is not LoadBalanced. (See
+If the driver encounters errors when establishing application connections (this includes the initial handshake and
+authentication), the driver MUST drivers MUST mark the server Unknown and clear the server's connection pool if the
+TopologyType is not LoadBalanced. (See
 [Why mark a server Unknown after an auth error?](#why-mark-a-server-unknown-after-an-auth-error))
 
 ### Monitoring SDAM events
--- a/source/server-discovery-and-monitoring/server-discovery-and-monitoring.md
+++ b/source/server-discovery-and-monitoring/server-discovery-and-monitoring.md
@@ -1244,9 +1244,8 @@ and [other transient errors](#other-transient-errors) and
 ##### Authentication and Handshake errors
 
 If the driver encounters errors when establishing application connections (this includes the initial handshake and
-authentication), the driver MUST drivers MUST mark the server Unknown and clear the server's connection pool if the
-TopologyType is not LoadBalanced. (See
-[Why mark a server Unknown after an auth error?](#why-mark-a-server-unknown-after-an-auth-error))
+authentication), the driver MUST mark the server Unknown and clear the server's connection pool if the TopologyType is
+not LoadBalanced. (See [Why mark a server Unknown after an auth error?](#why-mark-a-server-unknown-after-an-auth-error))
 
 ### Monitoring SDAM events
 
--- a/source/server-discovery-and-monitoring/tests/unified/pool-clear-checkout-error.json
+++ b/source/server-discovery-and-monitoring/tests/unified/pool-clear-checkout-error.json
@@ -61,8 +61,7 @@
                   "observeEvents": [
                     "connectionCheckOutStartedEvent",
                     "poolClearedEvent",
-                    "connectionClosedEvent",
-                    "connectionCheckOutFailedEvent"
+                    "connectionClosedEvent"
                   ],
                   "uriOptions": {
                     "retryWrites": false,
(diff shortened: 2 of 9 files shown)


source/server-discovery-and-monitoring/server-discovery-and-monitoring.md
```diff
@@ -1241,10 +1241,11 @@ if and only if the error is "node is shutting down" or the error originated from
 and [other transient errors](#other-transient-errors) and
 [Why close connections when a node is shutting down?](#why-close-connections-when-a-node-is-shutting-down).)
 
-##### Authentication errors
+##### Authentication and Handshake errors
 
-If the authentication handshake fails for a connection, drivers MUST mark the server Unknown and clear the server's
-connection pool if the TopologyType is not LoadBalanced. (See
+If the driver encounters errors when establishing application connections (this includes the initial handshake and
+authentication), the driver MUST drivers MUST mark the server Unknown and clear the server's connection pool if the
```
dariakp, line 1247:
```suggestion
authentication), the driver MUST mark the server Unknown and clear the server's connection pool if the
```

source/server-discovery-and-monitoring/tests/unified/pool-clear-checkout-error.yml
```diff
@@ -0,0 +1,180 @@
...
+    expectEvents:
+      - client: *client
+        eventType: cmap
+        events:
+          - connectionCheckOutStartedEvent: {}
+          - poolClearedEvent: {}
+          - connectionClosedEvent: {}
+          - connectionCheckOutFailedEvent: {}
+
+  - description: Pool is cleared connection is closed (handshake error)
```
W-A-James, line 90:
```suggestion
  - description: Pool is cleared before connection is closed (handshake error)
```

matthewdale (approved):
Looks good! 👍

dariakp (approved):

source/server-discovery-and-monitoring/tests/unified/pool-clear-checkout-error.json
```diff
@@ -0,0 +1,304 @@
...
+              "connectionCheckOutStartedEvent": {}
+            },
+            {
+              "poolClearedEvent": {}
+            },
+            {
+              "connectionClosedEvent": {}
+            },
+            {
+              "connectionCheckOutFailedEvent": {}
```
stIncMale, line 145: We only wait for `connectionClosedEvent`, and don't wait for `connectionCheckOutFailedEvent`, yet `connectionCheckOutFailedEvent` is expected to be delivered the last. This means we have a race condition in this test: `connectionCheckOutFailedEvent` is not guaranteed to be observed. Let's wait for `connectionCheckOutFailedEvent`?

The same applies to the other test in this file, with the description `"Pool is cleared before connection is closed (handshake error)"`.

source/server-discovery-and-monitoring/tests/unified/pool-clear-min-pool-size-error.json
```diff
@@ -0,0 +1,232 @@
...
+  ],
+  "tests": [
+    {
+      "description": "Pool is cleared on authentication error during minPoolSize population",
+      "runOnRequirements": [
+        {
+          "minServerVersion": "4.4",
+          "serverless": "forbid",
+          "auth": true
+        }
```
stIncMale, line 29: Of all the 3 requirements here, only `auth: true` adds something to the top-level `runOnRequirements`. Let's remove the rest, as they achieve nothing.

stIncMale (changes requested):

source/server-discovery-and-monitoring/tests/unified/pool-clear-checkout-error.json
baileympearson: Omission of checkout failed is intentional - these tests are not testing the checkout event emission, but instead testing specifically that the pool cleared error preceeds a connection closed event.

stIncMale: If that's the case, then why are these tests required to observe `connectionCheckOutFailedEvent`, and why do they expect `connectionCheckOutFailedEvent`? It seems to me that we should either stop observing and expecting `connectionCheckOutFailedEvent`, or should wait for `connectionCheckOutFailedEvent` instead of / in addition to waiting for `connectionClosedEvent`.

W-A-James (approved):

stIncMale (approved):

PR #1691, opened 2024-10-29
DRIVERS-3009, DRIVERS-1447: Updates to find operations

alcaeus:

- [x] Update changelog.
- [x] Test changes in at least one language driver (PHP)
    clusters, and serverless).

This PR introduces changes to the CRUD spec, covering multiple tickets:
- DRIVERS-3009: Add a formal definition for `Collection::findOne`, especially in order to specify the `limit` behaviour and define available options
- DRIVERS-1447: Clarify how to handle `limit` and `batchSize` options when they are set to the same value to avoid an unnecessary cursor being created on the server.


source/crud/crud.md
```diff
@@ -702,6 +714,194 @@ class FindOptions {
    */
   let: Optional<Document>;
 }
+
+class FindOneOptions {
```
alcaeus, line 718: This duplicates a significant amount of options from `FindOptions`. There are two alternatives:
- Have `FindOptions` extend `FindOneOptions` and add options not supported by `findOne`
- Introduce `CommonFindOptions`, have `FindOptions` and `FindOneOptions` extend that class and add their own.

If we want to avoid duplication, I'd prefer option one.

source/crud/crud.md
```diff
@@ -2474,6 +2677,8 @@ aforementioned allowance in the SemVer spec.
 
 ## Changelog
 
+- 2024-10-29: Defined `findOne` operation as optional.
```
alcaeus, line 2680: "as optional" is tentative here pending an outcome of a poll across drivers. I'm considering making the operation required, but letting drivers choose their timeline for implementing this operation, as `findOne` is a common use case.

source/crud/tests/unified/findOne.yml
```diff
@@ -0,0 +1,60 @@
+description: "findOne"
```
alcaeus, line 1: These tests heavily borrow from the `find` tests. I've decided to only explicitly check for `limit` and the absence of `batchSize` in the outgoing command. I don't think there are any other special cases we need to test.

source/crud/crud.md
```diff
@@ -768,7 +968,24 @@ filter, skip, and limit are added as options to the `aggregate` command.
...
+- `cursorType`: `findOne` only supports non-tailable cursors
+- `limit`: drivers MUST set `limit` to 1 in the `find` command created for a `findOne` operation
+- `noCursorTimeout`: with a `limit` of 1 and no `batchSize`, there will not be an open cursor on the server
+
+##### Setting limit and batchSize options for find commands
+
+When users specify both `limit` and `batchSize` options with the same value, the server returns all results in the
+first batch, but still leaves an open cursor that needs to be closed using the `killCursors` command. To avoid this,
+drivers MUST send a value of `limit + 1` for `batchSize` in the resulting `find` command. This eliminates the open
+cursor issue.
```
alcaeus, line 986: Unlike suggested in the ticket, we can't just omit the `batchSize` option when `limit <= batchSize`. Omitting the batch size makes us fall back to the default batch size, which may be less than the limit requested by the user, defeating the expectation that all results are returned in a single batch. The only workaround for this is to set a batch size higher than limit to get the server to close the cursor immediately before returning results.

source/crud/crud.md
ShaneHarvey: I really dislike the duplication here. Here's what pymongo's docs say for find_one:

> All arguments to :meth:`find` are also valid arguments for :meth:`find_one`, although any `limit` argument will be ignored.

Can't we just say something similar here instead of duplicating?

source/crud/tests/unified/findOne.yml
```diff
@@ -0,0 +1,60 @@
...
+          - commandStartedEvent:
+              command:
+                find: *collection0Name
+                filter: { _id: 1 }
+                batchSize: { $$exists: false }
+                limit: 1
+              commandName: find
+              databaseName: *database0Name
+  -
+    description: 'FindOne with filter, sort, and skip'
```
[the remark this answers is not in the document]
alcaeus: Added

source/crud/tests/unified/find.yml
```diff
@@ -105,3 +105,30 @@ tests:
...
+          - { _id: 5, x: 55 }
+    expectEvents:
+      - client: *client0
+        events:
+          - commandStartedEvent:
+              command:
+                find: *collection0Name
+                filter: { _id: { $gt: 1 } }
+                limit: 4
+                batchSize: 5
```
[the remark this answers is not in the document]
alcaeus: Added

source/crud/crud.md
```diff
@@ -188,6 +188,18 @@ interface Collection {
    */
   find(filter: Document, options: Optional<FindOptions>): Iterable<Document>;
 
+  /**
+   * Find a documents matching the model.
+   *
+   * Note: The filter parameter below equates to the $query meta operator. It cannot
+   * contain other meta operators like $maxScan. However, do not validate this document
+   * as it would be impossible to be forwards and backwards compatible. Let the server
+   * handle the validation.
```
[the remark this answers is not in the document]
alcaeus: Removed this paragraph

source/crud/crud.md
```diff
@@ -188,6 +188,18 @@ interface Collection {
...
+   *
+   * @see https://www.mongodb.com/docs/manual/core/read-operations-introduction/
```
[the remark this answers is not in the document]
alcaeus: Changed the link to this one - I believe this works fine as an introduction to querying

source/crud/crud.md
```diff
@@ -176,18 +176,20 @@ interface Collection {
...
+   * @see https://www.mongodb.com/docs/manual/tutorial/query-documents/
    */
   find(filter: Document, options: Optional<FindOptions>): Iterable<Document>;
 
+  /**
+   * Find a documents matching the model.
+   *
+   * @see https://www.mongodb.com/docs/manual/tutorial/query-documents/
+   */
+  findOne(filter: Document, options: Optional<FindOneOptions>): Document;
```
nbbeeken, line 191:
```suggestion
  findOne(filter: Document, options: Optional<FindOneOptions>): Optional<Document>;
```

If there are no matches right?

source/crud/crud.md
nbbeeken: Typescript has syntax for this that we use in our driver:

For example:
```ts
type FindOneOptions = Omit<FindOptions, 'timeoutMode'>
```

Not suggesting we use that language specific syntax here, but I think it would be fine to psuedo declare FindOneOptions are equal to FindOptions _except_ for X Y Z

source/crud/crud.md
```diff
@@ -768,7 +958,23 @@ filter, skip, and limit are added as options to the `aggregate` command.
...
 
-##### Combining Limit and Batch Size for the Wire Protocol
+##### findOne API details
+
+The `findOne` operation is implemented using a `find` operation, but only returns the first document returned in the
+cursor. Due to the special case, `findOne` does not support the following options supported in `find`:
+
+- `batchSize`: drivers MUST NOT set a `batchSize` in the `find` command created for a `findOne` operation
+- `cursorType`: `findOne` only supports non-tailable cursors
+- `limit`: drivers MUST set `limit` to 1 in the `find` command created for a `findOne` operation
```
nbbeeken, line 968: Seems like a good place to mention `singleBatch` should be set right?

This should be in the tests too I think

nbbeeken (changes requested):

source/crud/crud.md
alcaeus: Ah, yes, will update

source/crud/crud.md
alcaeus: Wouldn't `singleBatch` be unnecessary if we're using `limit` 1?

nbbeeken: The node driver has always set that field for find one, maybe it is redundant but then again why does setting that field not close the cursor with the single batch it is requested to return? 

I would assume that `singleBatch == true` leads to `cursorId: 0` every time

alcaeus: To summarise the discussion we had about this, `singleBatch: true` does indeed lead to `cursorId: 0` every time, including in those cases when not all requested results would fit into the result document due to the BSON size limit. In those cases, our expectation would be to receive a result cursor with the rest of the documents, but `singleBatch: true` would lead to the result omitting those documents.
For that reason, we decided to only require `singleBatch` for `findOne`, where we don't expect to receive a cursor in any case. The case of `limit == batchSize` is solved by increasing the `batchSize` by one to ensure the server encounters an end-of-stream from the PlanExecutor while trying to fill the requested batch. The server is considering a change to this behaviour as part of [SERVER-96643](https://jira.mongodb.org/browse/SERVER-96643).

source/crud/crud.md
nbbeeken: @alcaeus What is your thinking here? Seems like we have a preference for not duplicating the options, but as for exactly how to write it, there are a couple of choices here.

nbbeeken (changes requested):

source/crud/crud.md
alcaeus: After thinking about a good solution, I figured that solving this for `find`/`findOne` is only half the story, as there's lots of other duplication (e.g. `UpdateOptions` and `ReplaceOptions`). I filed DRIVERS-3038 to solve the duplication issue and also remove the code blocks: they are unnecessarily prescriptive and prevent us from properly linking to various options in the document. As such, I'd defer making changes to this for the time being.

nbbeeken: Something for next spec-fest perhaps :) fine by me

nbbeeken (approved):

source/crud/crud.md
ShaneHarvey: Can't we just go with one of the proposed simple solutions for now? We can fix the find/findOne duplication in this PR and then use DRIVERS-3038 to fix the remaining issues.

alcaeus: Changed to the `Omit` syntax suggested by @nbbeeken.

source/crud/crud.md
```diff
@@ -176,18 +176,20 @@ interface Collection {
...
-   * Note: The filter parameter below equates to the $query meta operator. It cannot
-   * contain other meta operators like $maxScan. However, do not validate this document
-   * as it would be impossible to be forwards and backwards compatible. Let the server
-   * handle the validation.
-   *
    * Note: result iteration should be backed by a cursor. Depending on the implementation,
    * the cursor may back the returned Iterable instance or an iterator that it produces.
    *
-   * @see https://www.mongodb.com/docs/manual/core/read-operations-introduction/
+   * @see https://www.mongodb.com/docs/manual/tutorial/query-documents/
```
jmikola, line 182: This is documentation for constructing a query filter. The more relevant docs for a driver developer would be the `find` command itself.

```suggestion
   * @see https://www.mongodb.com/docs/manual/reference/command/find/
```

source/crud/crud.md
```diff
@@ -176,18 +176,20 @@ interface Collection {
...
    */
   find(filter: Document, options: Optional<FindOptions>): Iterable<Document>;
 
+  /**
+   * Find a document matching the model.
+   *
+   * @see https://www.mongodb.com/docs/manual/tutorial/query-documents/
```
jmikola, line 189:
```suggestion
   * @see https://www.mongodb.com/docs/manual/reference/command/find/
```

source/crud/crud.md
```diff
@@ -712,6 +714,8 @@ class FindOptions {
    */
   let: Optional<Document>;
 }
+
+type FindOneOptions = Omit<FindOptions, 'batchSize' | 'cursorType' | 'limit' | 'noCursorTimeout'>;
```
jmikola, line 718: In "findOne API details" below, you talk about how `singleBatch` must be set to `true`. Shouldn't that option also be omitted here?

Looking at PHPLIB, we don't expose that in the `Collection::findOne()` API.

source/crud/crud.md
```diff
@@ -2496,6 +2513,8 @@ aforementioned allowance in the SemVer spec.
 
 ## Changelog
 
+- 2024-11-05: Define `findOne` operation as optional.
```
jmikola, line 2516: You can also note that you're adding guidance on `limit` and `batchSize` for `find` operations.

ShaneHarvey (approved):

source/crud/crud.md
jmikola: Oh, I forgot that `singleBatch` isn't even in FindOptions (and there's a Q&A entry addressing that). Carry on.

jmikola (approved):
LGTM with `@see` and changelog updates.

PR #1693, opened 2024-10-29
DRIVERS-1400: Add test to ensure read preference tags are case sensitive

alcaeus:

- [x] Update changelog.
- ~[ ] Test these changes against all server versions and topologies (including standalone, replica set, sharded
    clusters, and serverless).~

This adds a single new test with read preference tags that are mixed case, to ensure that drivers don't inadvertently lowercase values.


JamesKovacs (approved):
LGTM

PR #1695, opened 2024-10-29
DRIVERS-1016 clarify decryption does not need the key ID or algorithm.

kevinAlbs:
# Summary

Clarify decryption does not need the key ID or algorithm.

As a drive-by fix, the terms "In-Use Encryption" (the umbrella term) "CSFLE" (the first version) and "QE" (the second version) are applied.

# Background & Motivation

Using "In-Use Encryption", "CSFLE", and "QE" is intended to more closely agree with public documentation. See [Naming](https://github.com/mongodb/specifications/blob/e26b8cdff55ec33c28cc0fbb2b4eeba004093c1e/source/client-side-encryption/client-side-encryption.md#naming).

- ~~[ ] Update changelog.~~
- ~~[ ] Test changes in at least one language driver.~~
    clusters, and serverless).~~


katcharov (approved):
LGTM

PR #1696, opened 2024-10-29
DRIVERS-2124: test that inserts and upserts respect all _id values

nbbeeken:

- [ ] Update changelog.
    clusters, and serverless).

### DIFF
--- a/.github/workflows/json-regenerate-check.yml
+++ b/.github/workflows/json-regenerate-check.yml
@@ -11,17 +11,18 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
       - name: Set up Python
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v5
         with:
-          python-version: 3.7
+          python-version: '3.x'
       - name: Set up npm
-        uses: actions/setup-node@v1
+        uses: actions/setup-node@v4
         with:
-          node-version: 12
+          node-version: lts/*
       - name: Install dependencies
         run: |
+          npm init -y && npm install js-yaml
           npm install -g js-yaml
           pip3 install pymongo pyyaml jinja2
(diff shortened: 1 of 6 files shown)


source/crud/tests/unified/create-id-types.yml
```diff
@@ -0,0 +1,903 @@
...
+      client: client
+      databaseName: crud_id
+  - collection:
+      id: collection
+      database: database
+      collectionName: type_tests
+tests:
+  - description: inserting _id with type double via insertOne
+    operations:
+      - {name: dropCollection, object: database, arguments: {collection: type_tests}}
```
ShaneHarvey, line 18: Drop database and collection are expensive, slow commands. If we're adding so many tests it would be good to avoid it.

Also do we really need to test every combination of operation and BSON type?

ShaneHarvey: Perhaps it would be faster to put all the operations in one test. There's a lot of overhead to setting up a unified test which really adds up in runtime cost when adding 150 tests.

.github/workflows/json-regenerate-check.yml
```diff
@@ -31,8 +32,8 @@ jobs:
           python3 ./source/client-side-encryption/etc/generate-test.py ./source/client-side-encryption/etc/test-templates/*.template ./source/client-side-encryption/tests/legacy
           python3 ./source/client-side-operations-timeout/etc/generate-basic-tests.py ./source/client-side-operations-timeout/etc/templates ./source/client-side-operations-timeout/tests
           python3 ./source/etc/generate-handshakeError-tests.py
+          node ./source/etc/generate-crud-id-type-tests.mjs
```
alcaeus, line 35: Nit: since all other scripts to generate test files are written in Python, should this one really be written in a different language?

nbbeeken: I wouldn't say no to php scripts :)

PR #1697, opened 2024-10-29
DRIVERS-943 Convert retryable writes command construction tests to unified format

isabelatkinson:
DRIVERS-943

This PR removes the retryable writes command construction tests that assert on the presence/absence of `txnNumber` and adds equivalent assertions to the unified tests. Most of the existing retryable writes unified tests don't currently use command monitoring, and I only added what was necessary to match the coverage of the removed construction tests. DRIVERS-2849 tracks adding more comprehensive command monitoring assertions to the retryable writes tests.

Draft PR for Rust: https://github.com/mongodb/mongo-rust-driver/pull/1234

The unacknowledged writes and collection bulk write tests are not run in the Rust driver, so another driver will need to verify that these tests work. 

- [x] Update changelog.
    clusters, and serverless).


source/retryable-writes/tests/README.md
```diff
@@ -65,40 +65,8 @@ allow the fail point to trigger on the second command.
...
-    - `deleteOne()`
-    - `findOneAndDelete()`
-    - `findOneAndReplace()`
-    - `findOneAndUpdate()`
-- Supported multi-statement write operations
-    - `insertMany()` with `ordered=true`
-    - `insertMany()` with `ordered=false`
-    - `bulkWrite()` with `ordered=true` (no `UpdateMany` or `DeleteMany`)
-    - `bulkWrite()` with `ordered=false` (no `UpdateMany` or `DeleteMany`)
+Command construction tests asserting on the absence or presence of the `txnNumber` field have been replaced by command
```
aditi-khare-mongoDB, line 68: I feel this wording is a little confusing - originally it made me think we are no longer asserting on `txnNumber`.

I think it'd be better to remove this section, and  if we want to include this information somewhere, we can expand in the changelog instead. Something like the following:

>  Convert command construction tests to unified format. In the retryable writes unified tests, the unified spec runner now asserts on the absence or presence of the `txnNumber` field using 'Command Monitoring'.

What do you think?

isabelatkinson: The specs README states:

>Outdated tests must not be removed completely, but may be marked as such (e.g. by striking through or replacing the entire test with a note (e.g. Removed).

I think it's useful to maintain this section as a reference for drivers that have not yet removed these tests. WDYT about adding a strikethrough to the section title and rephrasing this as:

>These prose tests have been removed in favor of unified format tests.

isabelatkinson (commented):
@aditi-khare-mongoDB do these tests pass in the Node driver?

source/retryable-writes/tests/README.md
aditi-khare-mongoDB: Sure, we can also rephrase it to something like.

> In the retryable writes unified tests, the unified spec runner now asserts on the absence or presence of the txnNumber field using 'Command Monitoring'.

This or the strikethrough works!

aditi-khare-mongoDB (approved):
See comment! LGTM with changes

PR #1699, opened 2024-10-29
DRIVERS-2124: test that inserts and upserts respect null _id values

nbbeeken:

- ~[ ] Update changelog.~ N/A
    clusters, and serverless).

## Description

Adds tests for null and undefined as `_id` values to check that drivers do not interpret these types as "unset" and generate an ObjectId and overwrite the value.

### DIFF
--- a/source/crud/tests/unified/create-null-ids.yml
+++ b/source/crud/tests/unified/create-null-ids.yml
@@ -1,0 +1,288 @@
+description: CRUD ID Type Tests
+schemaVersion: '1.0'
+createEntities:
+  - client:
+      id: client
+      observeEvents:
+        - commandStartedEvent
+  - database:
+      id: database
+      client: client
+      databaseName: crud_id
+  - collection:
+      id: collection
+      database: database
+      collectionName: type_tests
+tests:
+  - description: inserting _id with type undefined via insertOne
+    operations:
+      - name: dropCollection
+        object: database
+        arguments:
+          collection: type_tests
+      - name: insertOne
+        object: collection
+        arguments:
+          document:
+            _id:
+              $undefined: true
+    outcome:
+      - databaseName: crud_id
+        collectionName: type_tests
+        documents:
+          - _id:
+              $undefined: true
+  - description: inserting _id with type undefined via insertMany
+    operations:
+      - name: dropCollection
+        object: database
+        arguments:
+          collection: type_tests
+      - name: insertMany
+        object: collection
+        arguments:
+          documents:
+            - _id:
+                $undefined: true
+    outcome:
+      - databaseName: crud_id
+        collectionName: type_tests
+        documents:
+          - _id:
+              $undefined: true
+  - description: inserting _id with type undefined via updateOne
+    operations:
+      - name: dropCollection
+        object: database
+        arguments:
+          collection: type_tests
+      - name: updateOne
+        object: collection
+        arguments:
+          filter:
+            _id:
+              $undefined: true
+          update:
+            $unset:
+              a: ''
+          upsert: true
+    outcome:
+      - databaseName: crud_id
+        collectionName: type_tests
+        documents:
+          - _id:
+              $undefined: true
+  - description: inserting _id with type undefined via updateMany
+    operations:
+      - name: dropCollection
+        object: database
+        arguments:
+          collection: type_tests
+      - name: updateMany
+        object: collection
+        arguments:
+          filter:
+            _id:
+              $undefined: true
+          update:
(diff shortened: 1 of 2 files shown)


source/crud/tests/unified/create-null-ids.yml
```diff
@@ -0,0 +1,288 @@
...
+      client: client
+      databaseName: crud_id
+  - collection:
+      id: collection
+      database: database
+      collectionName: type_tests
+tests:
+  - description: inserting _id with type undefined via insertOne
+    operations:
+      - name: dropCollection
```
jmikola, line 19: Why is an explicit `dropCollection` operation necessary in each test? Wouldn't `initialData` with an empty collection suffice?

If we keep this, it could be refactored as a YAML anchor.

source/crud/tests/unified/create-null-ids.yml
```diff
@@ -0,0 +1,288 @@
...
+      observeEvents:
+        - commandStartedEvent
+  - database:
+      id: database
+      client: client
+      databaseName: crud_id
+  - collection:
+      id: collection
+      database: database
+      collectionName: type_tests
```
jmikola, line 15: I realize this file was generated, but you can still use YAML anchors for the database and collection names for consistency with other files. Hopefully it's just a quick search/replace operation.

I understand that won't handle the full namespace string for client bulkWrite, but it should remove repetition elsewhere.

source/crud/tests/unified/create-null-ids.yml
```diff
@@ -0,0 +1,288 @@
+description: CRUD ID Type Tests
```
jmikola, line 1:
```suggestion
description: create-null-ids
```

This should match the base name of the test file for consistency.

source/crud/tests/unified/create-null-ids.yml
nbbeeken: Added anchors, and also used some for the input data

source/crud/tests/unified/create-null-ids.yml
```diff
@@ -0,0 +1,288 @@
...
+      - name: dropCollection
+        object: database
+        arguments:
+          collection: type_tests
+      - name: insertOne
+        object: collection
+        arguments:
+          document:
+            _id:
+              $undefined: true
```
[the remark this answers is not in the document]
nbbeeken: Updated to flow better and anchored some of the input as well.

source/crud/tests/unified/create-null-ids.yml
nbbeeken: fixed!

source/crud/tests/unified/create-null-ids.yml
nbbeeken: Migrated to initialData, that's better practice than including drop in the test.

source/crud/tests/unified/create-null-ids.yml
```diff
@@ -0,0 +1,169 @@
+description: create-null-ids
+schemaVersion: '1.0'
+
+createEntities:
+  - client:
+      id: client
+      observeEvents:
+        - commandStartedEvent
```
jmikola, line 8:
```suggestion
```

Looks like you got rid of APM assertions.

jmikola (approved):
LGTM with suggestion applied.

source/crud/tests/unified/create-null-ids.yml
nbbeeken: I figured the result of count is the source of truth now since we are able to use a BSON type query avoiding any comparison looseness surrounding null/undefined. But happy to add it back, just saying it was intentional to remove.

jmikola: `countDocuments` seems sufficient. APM assertions don't help if we have to worry about how some drivers represent null/undefined values.

source/crud/tests/unified/create-null-ids.yml
```diff
@@ -0,0 +1,169 @@
...
+    operations:
+      - name: clientBulkWrite
+        object: client
+        arguments: {models: [{insertOne: {namespace: crud_id.type_tests, document: *null_id}}]}
+      - name: countDocuments
+        object: *collection
+        arguments: {filter: *null_id_filter}
+        expectResult: 1
+
+  - description: inserting _id with type undefined via insertOne
```
jmikola, line 99: Note: I'm removing these tests in https://github.com/mongodb/specifications/pull/1713 as the server prohibits undefined values for `_id`.

PR #1702, opened 2024-10-30
DRIVERS-2807 Update GridFS spec to add `delete_by_name` and `rename_by_name`

GromNaN:
Fix [DRIVERS-2808](https://jira.mongodb.org/browse/DRIVERS-2808) and [DRIVERS-2807](https://jira.mongodb.org/browse/DRIVERS-2807)

- [x] Update changelog.
- [x] Test changes in at least one language driver https://github.com/mongodb/mongo-php-library/pull/1504

### DIFF
--- a/source/gridfs/gridfs-spec.md
+++ b/source/gridfs/gridfs-spec.md
@@ -671,7 +671,7 @@ This method is an optimisation over deleting each revision of a stored file indi
 
 Drivers MUST first find the `_id` field of all files collection documents with the given filename. Drivers MUST then
 delete all files collection documents with the found ids. Drivers MUST then delete all chunks with `files_id` in the
-found ids that was just deleted.
+found ids that were just deleted.
 
 If there are no files collection documents with the given filename, drivers MUST raise an error.
 
--- a/source/gridfs/gridfs-spec.md
+++ b/source/gridfs/gridfs-spec.md
@@ -649,6 +649,32 @@ orphaned chunks with files_id equal to id before raising the error.
 
 If a networking or server error occurs, drivers MUST raise an error.
 
+### File deletion by filename
+
+To rename all revisions of a stored file with the specified filename, drivers SHOULD provide the method
+`delete_by_name`:
+
+```javascript
+class GridFSBucket {
+
+  /**
+   * Deletes all stored files with the specified @filename from a GridFS bucket.
+   */
+  void delete_by_name(string filename);
+
+}
+```
+
+This method is an optimisation over deleting each revision of a stored file individually.
+
+**Implementation details:**
+
+Drivers MUST first find the `_id` field of all files collection documents with the given filename. Drivers MUST then
+delete all chunks with `files_id` in the found ids that was just deleted. Drivers MUST then delete all files collection
+documents with the found ids.
+
+If there are no files collection documents with the given filename, drivers MUST raise an error.
+
 ### Generic Find on Files Collection
 
 ```javascript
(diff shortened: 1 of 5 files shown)


source/gridfs/gridfs-spec.md
```diff
@@ -649,6 +649,32 @@ orphaned chunks with files_id equal to id before raising the error.
...
+
+}
+```
+
+This method is an optimisation over deleting each revision of a stored file individually.
+
+**Implementation details:**
+
+Drivers MUST first find the `_id` field of all files collection documents with the given filename. Drivers MUST then
+delete all chunks with `files_id` in the found ids that was just deleted. Drivers MUST then delete all files collection
```
ShaneHarvey, line 673: Wouldn't we want to delete the files documents first and then the chunks? Doing the opposite increases the chance of reading a phantom corrupt file. This is the same reason we always upload the chunks before saving the files document.

GromNaN: You are correct. Files should be deleted before chunks to ensure there is no partial files. But if the operation that deletes chunks fails, then there is orphan chunks.

Spec updated.

source/gridfs/gridfs-spec.md
```diff
@@ -814,18 +840,41 @@ class GridFSBucket {
...
+Sets the filename field in the stored file's files collection document to the new filename.
 
 **Implementation details:**
 
 Drivers construct and execute an update_one command on the files collection using `{ _id: @id }` as the filter and
 `{ $set : { filename : "new_filename" } }` as the update parameter.
 
-To rename multiple revisions of the same filename, users must retrieve the full list of files collection documents for a
-given filename and execute "rename" on each corresponding `_id`.
+If `renameByName` is not implemented to rename multiple revisions of the same filename, users must retrieve the full
```
GromNaN, line 850:
```suggestion
If `rename_by_name` is not implemented to rename multiple revisions of the same filename, users must retrieve the full
```

alcaeus (approved):

source/gridfs/gridfs-spec.md
```diff
@@ -649,6 +649,32 @@ orphaned chunks with files_id equal to id before raising the error.
...
+}
+```
+
+This method is an optimisation over deleting each revision of a stored file individually.
+
+**Implementation details:**
+
+Drivers MUST first find the `_id` field of all files collection documents with the given filename. Drivers MUST then
+delete all files collection documents with the found ids. Drivers MUST then delete all chunks with `files_id` in the
+found ids that was just deleted.
```
ShaneHarvey, line 674: `ids that was just deleted.` -> `ids that were just deleted.`

Or use client.bulk_write() to delete both in one command :). We can leave that as a future optimization.

ShaneHarvey (approved):
LGTM after the small grammar fix.

source/gridfs/gridfs-spec.md
```diff
@@ -821,11 +847,34 @@ Sets the filename field in the stored file's files collection document to the ne
...
-given filename and execute "rename" on each corresponding `_id`.
+If `renameByName` is not implemented to rename multiple revisions of the same filename, users must retrieve the full
+list of files collection documents for a given filename and execute "rename" on each corresponding `_id`.
 
 If there is no file with the given id, drivers MUST raise an error.
 
+### Renaming stored files by filename
+
+To rename all revisions of a stored file with the specified filename, drivers SHOULD provide the method
+`rename_by_name`:
```
GromNaN, line 858: One thing I did not think about is if there is already a newer file with the new filename, then it will stay the last revision for this filename.

We can't change the date of the renamed files, as that would change the order of the revisions.

source/gridfs/gridfs-spec.md
GromNaN: > Or use client.bulk_write() to delete both in one command :). We can leave that as a future optimization.

This would require server version 8.0 with the new client bulk write API to perform operations on both collections.

source/gridfs/gridfs-spec.md
ShaneHarvey: Correct that seems like expected behavior to me.

PR #1703, opened 2024-10-30
DRIVERS-2865: Require drivers to always send out bypassDocumentValidation

alcaeus:

- [x] Update changelog.
- [ ] ~Test these changes against all server versions and topologies (including standalone, replica set, sharded
    clusters, and serverless).~

This effectively reverts #266 and adds tests for the behaviour. I've also decided to remove the rationale for sending `bypassDocumentValidation: true` in the bulk write specification as this difference no longer exists.

The new tests are grouped in a single file to make it easier for drivers to skip this particular test until they've had a chance to make the changes. No other test (aside client bulkWrite which already implements the new behaviour) use `bypassDocumentValidation: false`, so this change should not have any side effects in drivers.

### DIFF
--- a/source/crud/bulk-write.md
+++ b/source/crud/bulk-write.md
@@ -897,19 +897,6 @@ recording the summary counts. We expect that most users are not interested in th
 and that most users will rely on defaults, so `verboseResults` defaults to `false` to improve performance in the common
 case.
 
-### Why should drivers send `bypassDocumentValidation: false` for `bulkWrite`?
-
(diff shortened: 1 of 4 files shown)


jmikola (approved):
Test changes LGTM but I can't verify "Test changes in at least one language driver." Add a link to the related PR if possible.

alcaeus:
> Test changes LGTM but I can't verify "Test changes in at least one language driver." Add a link to the related PR if possible.

I tested this locally, but since we rely on libmongoc's bulk write implementation for all inserts, I was only testing the negative case. I was able to confirm that `aggregate` with `$out` works as expected.

PR #1708, opened 2024-10-31
DRIVERS-2926 Updated BSON Binary Subtype 9 - Vector Status to Accepted

caseyclements:
Updating status, which wasn't done before merge.


blink1073 (approved):

mdbmes:
It looks like feedback from our slack conversation hasn't made it anywhere durable yet, so I filed DRIVERS-3031.

PR #1712, opened 2024-11-01
chore(DRIVERS-2134): clarify write concern options in txn

durran:
This is to note that collection options have no bearing on the txn options if the txn was already stared.

- [ ] Update changelog.
    clusters, and serverless).


durran:
@comandeo-mongo Do you think we should add a prose test here? There's no way in the unified format to do it currently as the collection is created before the operations are run.

comandeo-mongo:
Yes, I think we should add a prose test for this case. Looks like the test may just reproduce the example you added to the specification, so it won't be long and complicated.

PR #1713, opened 2024-11-01
DRIVERS-2124: Remove tests for undefined _id values

jmikola:
https://jira.mongodb.org/browse/DRIVERS-2124

The server does not allow _id values to be undefined, nor does it allow comparisons to undefined:

I observed test failures in PHPLIB (see: [Evergreen task](https://spruce.mongodb.com/task/mongo_php_library_test_debian12_php_8.3_phpc_next_stable_test_mongodb_8.0_replicaset_noauth_nossl_patch_7781e7562754c88efd59b32ad177edf4c8486d95_6723d73f2dea2d00075703a6_24_10_31_19_15_15?execution=0&sortBy=STATUS&sortDir=ASC), which correlate with the following errors in the server:

 * https://github.com/mongodb/mongo/blob/fecaab0a15e5f41a326ec74576f2d27be3dd004e/src/mongo/db/update/storage_validation.cpp#L139
 * https://github.com/mongodb/mongo/blob/fecaab0a15e5f41a326ec74576f2d27be3dd004e/src/mongo/db/matcher/expression_leaf.cpp#L139

Follow-up to https://github.com/mongodb/specifications/pull/1699


source/crud/tests/unified/create-null-ids.yml
```diff
@@ -95,75 +95,3 @@ tests:
         object: *collection
         arguments: {filter: *null_id_filter}
         expectResult: 1
-
-  - description: inserting _id with type undefined via insertOne
```
jmikola, line 99: @nbbeeken: I assume these failures were missed in the original PR because Node silently converts undefined to null. In any event, undefined was never in the original scope of [DRIVERS-2124](https://jira.mongodb.org/browse/DRIVERS-2124) so let's just forget about this.

nbbeeken (approved):

source/crud/tests/unified/create-null-ids.yml
nbbeeken: Yep, exactly right, good catch thanks for fixing this so quick!

PR #1714, opened 2024-11-01
Move handshake prose tests from spec into prose test file

baileympearson:
The handshake spec had prose tests inside the spec instead of inside the prose test file.  This PR moves the handshake tests into the prose test file.

- [x] Update changelog.
- [N/A] Test changes in at least one language driver.
- [N/A] Test these changes against all server versions and topologies (including standalone, replica set, sharded
    clusters, and serverless).


dariakp (approved):

dariakp (changes requested):
I think we still want to update the change log in the files

dariakp (approved):

PR #1716, opened 2024-11-01
Audit less-than symbols followed by letters

ghost:
I noticed that the "KMS Provider" section of the Client Side Encryption spec was missing some text due to unescaped less-than characters, which are interpreted as HTML tags in markdown. I audited all markdown files for occurrences of a less-than symbol followed by a letter.

These less-than characters are problematic in markdown because there's disagreement on how they should be escaped. Our mdformat tooling will enforce an escape with backslash, but this usage isn't actually supported in the original markdown spec. We use an extension to enable additional backslash escapes in our mkdocs setup, but for the best compatibility all less than symbols should be in a code block or should be both escaped and followed by a non-letter. It's typical in this repo to use less/greater symbols as angle brackets for placeholder text, and this works best within a code block where no escaping is needed.

Changes:
* Escape and space non-code placeholder text in auth.md table
* use code rather than links for mongodb:// URIs
* Use code blocks for placeholder text in client-side-encryption/tests
* Use code blocks for placeholder text in connection-monitoring-and-pooling
* Fix code quoting in socks5-support/tests

### DIFF
--- a/scripts/check_md_html.py
+++ b/scripts/check_md_html.py
@@ -33,7 +33,7 @@ disallowed_re = re.compile(
         td[\s>] |
         tr> |
         !-- )
-""",
+    """,
     re.VERBOSE,
 )
 


dariakp (commented):
LGTM, but do we maybe want to add a lint check to prevent inadvertently reintroducing these?

mdbmes:
> LGTM, but do we maybe want to add a lint check to prevent inadvertently reintroducing these?

sure

stIncMale (approved):

PR #1717, opened 2024-11-01
DRIVERS-2223 Fix race condition in pool-checkout-returned-connection-maxConnecting.yml

matthewdale:
[DRIVERS-2223](https://jira.mongodb.org/browse/DRIVERS-2223)

Don't wait on the remaining checkOut events to resolve a race condition in the "threads blocked by maxConnecting check out returned connections" test caused by different connection pool implementations.

Based on a helpful [comment](https://github.com/mongodb/specifications/pull/1152#issuecomment-1057088946) by @stIncMale from https://github.com/mongodb/specifications/pull/1152.

- [x] Update changelog.
    clusters, and serverless).

### DIFF
--- a/source/connection-monitoring-and-pooling/tests/cmap-format/pool-checkout-returned-connection-maxConnecting.json
+++ b/source/connection-monitoring-and-pooling/tests/cmap-format/pool-checkout-returned-connection-maxConnecting.json
@@ -23,6 +23,7 @@
     }
   },
   "poolOptions": {
+    "maxConnecting": 2,
     "maxPoolSize": 10,
     "waitQueueTimeoutMS": 5000
   },
@@ -70,11 +71,6 @@
     {
       "name": "checkIn",
       "connection": "conn0"
-    },
-    {
-      "name": "waitForEvent",
-      "event": "ConnectionCheckedOut",
-      "count": 4
     }
   ],
   "events": [
@@ -103,14 +99,6 @@
     {
       "type": "ConnectionCheckedOut",
       "connectionId": 1,
-      "address": 42
-    },
-    {
-      "type": "ConnectionCheckedOut",
-      "address": 42
-    },
-    {
-      "type": "ConnectionCheckedOut",
       "address": 42
     }
   ],
--- a/source/connection-monitoring-and-pooling/tests/cmap-format/pool-checkout-returned-connection-maxConnecting.json
+++ b/source/connection-monitoring-and-pooling/tests/cmap-format/pool-checkout-returned-connection-maxConnecting.json
@@ -71,6 +71,10 @@
     {
       "name": "checkIn",
       "connection": "conn0"
+    },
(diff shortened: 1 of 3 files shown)


source/connection-monitoring-and-pooling/tests/cmap-format/pool-checkout-returned-connection-maxConnecting.json
```diff
@@ -70,11 +71,6 @@
     {
       "name": "checkIn",
       "connection": "conn0"
-    },
-    {
-      "name": "waitForEvent",
-      "event": "ConnectionCheckedOut",
-      "count": 4
     }
```
stIncMale, line 74: I think, we need another `wait` after `checkIn`. The `wait` before `checkIn` ensures (not really, but good enough) that all activities are blocked (waiting for 4 `ConnectionCheckOutStarted` events is not enough for that): two threads are blocked for 750 ms by the failpoint, and one is blocked on `maxConnecting`. The proposed `wait` after `checkIn` ensures that there is enough time for one of `thread1`, `thread2`, `thread3` to checkout `conn0` that has just been checked in, and deliver the corresponding `ConnectionCheckedOut` event.

When I run the test as it currently is in the PR, it fails for me right away, because the last `ConnectionCheckedOut` fails to be delivered quickly enough. The proposed `wait` fixes that.

stIncMale (changes requested):

source/connection-monitoring-and-pooling/tests/cmap-format/pool-checkout-returned-connection-maxConnecting.json
matthewdale: Would waiting for the 2nd `ConnectionCheckedOut` events work also? That might be more reliable than adding a sleep.

E.g. add this 
```json
{
  "name": "waitForEvent",
  "event": "ConnectionCheckedOut",
  "count": 2
}
```

stIncMale: The main idea of the fix applied in this PR, is that we to do all the work and checks while the two threads blocked by the failpoint keep holding all the `maxConnecting` permits. We know that they are blocked for at least 750 ms, so us waiting twice for at least 100 ms, should be OK. But if we start waiting for any events instead, we can no longer be sure that the above will hold. Thus, I don't think that is a viable approach.

stIncMale: Ok, will you make the change?

matthewdale: Yes, updated!

stIncMale (approved):

PR #1718, opened 2024-11-02
fix missing space before inline math

mdbmes:
This fixes a Markdown rendering error in the client-side-encryption spec due to a typo. The problem was much more noticeable on GitHub than on mkdocs (readthedocs):
* on GitHub, the word was no longer interpreted as math, and it rendered in plain text as "joining$dbName$"
* on mkdocs the math markup rendered normally still, so it appeared as "joiningdbName" with "dbName" in math italics, making the missing space somewhat hard to see


blink1073:
Yay, readthedocs preview in PRs is back!

kevinAlbs (approved):

PR #1723, opened 2024-11-04
DRIVERS-3033 SDAM logging tests should allow durationMS as a float

ShaneHarvey:
DRIVERS-3033 SDAM logging tests should allow durationMS as a float

See details in the ticket. 

- [X] Tested in python here: https://github.com/mongodb/mongo-python-driver/pull/1988


ShaneHarvey:
@jmikola do you mind reviewing this one?

jmikola (approved):

PR #1724, opened 2024-11-05
DEVPROD-11594 Document BSON Text Format

jbalint:

- [ ] Update changelog.
    clusters, and serverless).

### DIFF
--- a/source/extended-json/extended-json-query.md
+++ b/source/extended-json/extended-json-query.md
@@ -1,0 +1,54 @@
+# Extended JSON Query
+
+## Abstract
+
+Extended JSON query variation is a superset of Extended JSON. It is
+intended to be used for serializing queries that can be specified in
+the mongo shell. It supplements the Extended JSON syntax with support
+for the following features:
+
+- Trailing commas after the last value in an object
+
+- NumberLong values in quotation marks
+
+- Dates constructed with `new Date()` including an `ISODate` format
+
+- Regex pattern enclosed in slashes
+
+## Examples
+
+### Trailing Comma
+    The following example shows an example of a JSON object with a trailing comma after the `b: 2` element. The trailing comma is supported only in objects, not in arrays. This is convenient when manually editing the text representation.
+```javascript
+{$match:
+ {a: 1,
+  b: 2,}
+}
+```
+
+### Quoted NumberLong Value
+    The EJSON syntax supports a `NumberLong()` constructor but the argument must not be quoted. The following example shows the quoted version of the number.
+```javascript
+{$match:
+ {num:
+  {$gt: NumberLong("123456789")}
+ }}
+```
+
+### Date Constructor
+    The following example shows construction of a Date object with the date string represented in ISODate form.
+```javascript
+{$match:
+ {dateField:
+  {$lt: new Date("2019-10-23T08:58:17.868Z")}
+ }}
+```
+
+### Regex Pattern in Slashes
+    The following example shows a regex value enclosed in slashes. This is consistent with the MQL syntax.
+```javascript
+{$regexFindAll:
+ {input: \"$obj.obj.obj.obj.str\",
+  regex: /transparent|Forward/}
+}
+```
--- a/source/extended-json/extended-json-query.md
+++ b/source/extended-json/extended-json-query.md
@@ -9,9 +9,11 @@ for the following features:
 
 - Trailing commas after the last value in an object
 
-- NumberLong values in quotation marks
+- NumberLong values in quotation marks (aka "double quote")
+  - Apostrophe (aka "single quote") is not supported
 
 - Dates constructed with `new Date()` including an `ISODate` format
+  - Including support for pre-epoch dates
 
 - Regex pattern enclosed in slashes
 


ShaneHarvey (commented):
I'm not sure the specs repo is the appropriate place for this document since drivers are not going to implement this format.

source/extended-json/extended-json-query.md
```diff
@@ -0,0 +1,56 @@
...
+
+- Dates constructed with `new Date()` including an `ISODate` format
+  - Including support for pre-epoch dates
+
+- Regex pattern enclosed in slashes
+
+## Examples
+
+### Trailing Comma
+    The following example shows an example of a JSON object with a trailing comma after the `b: 2` element. The trailing comma is supported only in objects, not in arrays. This is convenient when manually editing the text representation.
```
jbalint, line 23: How can I reformat this to satisfy the linter? The text should not be considered a code block.

The linter suggests:
```diff
 ### Trailing Comma
-    The following example shows an example of a JSON object with a trailing comma after the `b: 2` element. The trailing comma is supported only in objects, not in arrays. This is convenient when manually editing the text representation.
+
+```
+The following example shows an example of a JSON object with a trailing comma after the `b: 2` element. The trailing comma is supported only in objects, not in arrays. This is convenient when manually editing the text representation.
+```
+
```

source/extended-json/extended-json-query.md
```diff
@@ -0,0 +1,56 @@
+# Extended JSON Query
+
+## Abstract
+
+Extended JSON query variation is a superset of Extended JSON. It is
```
addaleax, line 5: This text seems to be describing Shell BSON syntax, which is neither a JSON format nor a superset of Extended JSON

jbalint: @samontea can you help shed some light on this?

Is the Shell BSON syntax documented @addaleax?

dstorch: Yeah, I don't describing it as a superset of Extended JSON is accurate. The proposed format wouldn't accept, for example, `{$numberInt: <number as string>}` would it? See this table in the extended JSON spec: https://github.com/mongodb/specifications/blob/master/source/extended-json/extended-json.md#conversion-table.

Also calling it "Extended JSON Query" seems like an inaccuate name. I thought we settled on a name like "BSON Text Format", since the folks from [SPM-3714](https://jira.mongodb.org/browse/SPM-3714) argue (reasonably, I think) that a non-JSON format is more readable.

source/extended-json/extended-json-query.md
```diff
@@ -0,0 +1,56 @@
+# Extended JSON Query
+
+## Abstract
```
dstorch, line 3: I think the abstract should be rewritten. Here is my attempt to write an abstract as I understand it:

This document describes a standard format which can be used for representing BSON documents textually. The format is designed with two principles in mind:
- The ability to represent any legal BSON document in a string format uniquely. That is, any legal BSON document should have exactly one representation in the BSON Text Format. Conversely, a valid BSON Text Format string must be uniquely convertible to a particular byte string which is a valid BSON document.
- Human readability. A primary use case of the format is for describing MQL correctness tests declaratively, and we expect those tests to often by inspected by server developers.

I actually think bullet number 1 is pretty interesting. I am confident that we want a given BSON Text Format string to always uniquely represent a BSON document (where uniqueness is defined in terms of binary equality). I'm less sure of whether we require a single unique BSON Text Format string for a BSON document -- trailing commas for the last element being one example of where we could choose to relax this.

source/extended-json/extended-json-query.md
```diff
@@ -0,0 +1,56 @@
...
+
+- NumberLong values in quotation marks (aka "double quote")
+  - Apostrophe (aka "single quote") is not supported
+
+- Dates constructed with `new Date()` including an `ISODate` format
+  - Including support for pre-epoch dates
+
+- Regex pattern enclosed in slashes
+
+## Examples
```
dstorch, line 20: I expected this to be much more specific about the format. Like, can we describe the exact representation for each BSON type? How is DBPointer represented? How is BinData and our various BinData subtypes represented? I guess this question is closely related to whether it's true that this format is indeed supposed to be an extension of extended JSON, which was not my understanding when we first discussed this.

dstorch (changes requested):
Question to drivers folks: Is it true that this repository is only intended to be used for specs that drivers are expected to implement? If so, I agree it's in the wrong place. I'm just not aware of any other repository of well-maintained and well-reviewed specifications. Let me know if you have a specific proposal for where this should live instead.

ShaneHarvey:
> Is it true that this repository is only intended to be used for specs that drivers are expected to implement? 

Yes. I suggest this new doc live in whatever project is supposed to implement it (mongosh?).

source/extended-json/extended-json-query.md
addaleax: > Is the Shell BSON syntax documented @addaleax?

https://www.mongodb.com/docs/mongodb-shell/reference/data-types/ documents some of the constructors provided by the shell, but it's not exhaustive.

> I thought we settled on a name like "BSON Text Format", since the folks from [SPM-3714](https://jira.mongodb.org/browse/SPM-3714) argue (reasonably, I think) that a non-JSON format is more readable.

You can call it whatever you want, but if you want a name that's been "settled on", we've put in a lot of effort on our team to standardize on using "Shell BSON syntax" internally (to distinguish it from things like Extended JSON), and that long predates SPM-3714 _and_ is being used by the team actually owning mongosh.

addaleax:
@jbalint I'd suggest we set up a call or something like that to discuss what the goals of writing this spec actually are, and where the description of the format should live.

jbalint:
> @jbalint I'd suggest we set up a call or something like that to discuss what the goals of writing this spec actually are, and where the description of the format should live.

We're using `shell.tojsononeline()` to export queries from the fuzzer. We use the JParse/json.h parser as a base for this but added support for the syntax extensions mentioned in this document (in jparse_util.cpp). The consumer is the QueryTester tool (currently in it's own repo, will be merged into `mongo` soon). It was suggested that we document this syntax. I was directed to structure the documentation as a modification to EJSON. However, it seems this is not EJSON because we're not using `EJSON.stringify()`?

* What is the format output by `shell.tojsononeline()`? Furthermore, is it documented?
* Is there no existing parser for `shell.tojsononeline()` syntax which required us to modify the existing JSON parser? (@lynne-wang @samontea)

If we want to document this syntax, we need to understand what it is before deciding on a home for it.

source/extended-json/extended-json-query.md
jbalint: @dstorch the proposed format does support `{$numberInt: <number as string>}`. i tested that one explicitly but a quick pass over the code confirms the rest are supported. the method which parses this is called `fromFuzzerJson()`. but i don't think "fuzzer JSON" should be a syntax.

afaiu, conveniently reusing the fuzzer generated queries was prefered both to reduce effort and have queries with the BSON syntax (incl date constructor and native regex). I found [this documentation of tojsononeline()](https://www.mongodb.com/docs/mongodb-shell/reference/ejson/stringify/#legacy-tojsononeline--) but it's not very helpful. If tojsononeline() is documented, we can use that to specify the input syntax to QueryTester (even if the implementation accepts a more general grammar).

jbalint:
> @jbalint I'd suggest we set up a call or something like that to discuss what the goals of writing this spec actually are, and where the description of the format should live.

That would be fine, but let's continue the discussion async for the moment to gather some of the background/context first.

addaleax:
> > @jbalint I'd suggest we set up a call or something like that to discuss what the goals of writing this spec actually are, and where the description of the format should live.
> 
> We're using `shell.tojsononeline()` to export queries from the fuzzer. We use the JParse/json.h parser as a base for this but added support for the syntax extensions mentioned in this document (in jparse_util.cpp). The consumer is the QueryTester tool (currently in it's own repo, will be merged into `mongo` soon).

@jbalint Yeah, so ... these are all server internals – if this is a new format and if there is value in documenting this particular syntax aside from it being (close to?) the general user-facing shell syntax, maybe the server's `mongo/docs/` directory is a decent choice? Or is the goal explicitly to create a public-facing spec?

> It was suggested that we document this syntax. I was directed to structure the documentation as a modification to EJSON. However, it seems this is not EJSON because we're not using `EJSON.stringify()`?

EJSON is expressing BSON using JSON, but syntax like `{ _id: ObjectId('...') }` just isn't JSON, so yeah, this isn't EJSON.

>     * What is the format output by `shell.tojsononeline()`? Furthermore, is it documented?

I think `tojsononeline()` generally aims to print values in a way where, if the result is evaluated in the shell, you'd get back a 1:1 equivalent object. Yes, the naming of `tojsononeline()` is _horrible_, because it leads users to assume that it returns JSON, which it _not_ what it does.  

>     * Is there no existing parser for `shell.tojsononeline()` syntax which required us to modify the existing JSON parser? (@lynne-wang @samontea)

So, for shell syntax, you can of course use `eval()` in the legacy shell/mongosh if you're working with trusted input. Our team also additionally maintains the [`@mongodb-js/shell-bson-parser`](https://github.com/mongodb-js/devtools-shared/tree/main/packages/shell-bson-parser) package, which we use for handling unsanitized input in environments where we do not want to expose full JS evaluation capabilities.

> If we want to document this syntax, we need to understand what it is before deciding on a home for it.

So ... you mentioned above that

> the proposed format does support {$numberInt: <number as string>}.

Is this correct? So this would be an EJSON-shell-syntax hybrid that supports input like `{ int: {$numberInt: '32'}, long: NumberLong('32') }`? IIRC the goal was to be able to represent any BSON object as text, so this would prevent representing a genuine `$numberInt` key in the BSON object, no? 

If this is really a hybrid syntax, I'd strongly consider whether it's possible to change this format so that it aligns with an existing standard (like pure EJSON or pure shell syntax).

jbalint:
@addaleax thanks so much for the context and details here. much appreciated!

I can summarize the situation as follows:
* fuzzer generated queries are output using `shell.tojsononeline()`. the output is not JSON, or EJSON, but Javascript (code?) which can be evaluated in the shell
** this is a desirable syntax for working with (eg, writing, debugging, etc) correctness tests
** this syntax is not fully documented but a parser is implemented TypeScript
* in order to read the queries written using this syntax, the query tester extended the EJSON (`json.cpp`) parser to support the grammar elements mentioned in the PR. the implementation doesn't admit all possible syntax the shell would accept and additionally admits some constructs (eg. `{$numberInt: }`) the shell wouldn't treat in the same way.

@dstorch as i understood, this documentation request came from you. how would you like to proceed here?

> > > @jbalint I'd suggest we set up a call or something like that to discuss what the goals of writing this spec actually are, and where the description of the format should live.
> > 
> > 
> > We're using `shell.tojsononeline()` to export queries from the fuzzer. We use the JParse/json.h parser as a base for this but added support for the syntax extensions mentioned in this document (in jparse_util.cpp). The consumer is the QueryTester tool (currently in it's own repo, will be merged into `mongo` soon).
> 
> @jbalint Yeah, so ... these are all server internals – if this is a new format and if there is value in documenting this particular syntax aside from it being (close to?) the general user-facing shell syntax, maybe the server's `mongo/docs/` directory is a decent choice? Or is the goal explicitly to create a public-facing spec?
> 
> > It was suggested that we document this syntax. I was directed to structure the documentation as a modification to EJSON. However, it seems this is not EJSON because we're not using `EJSON.stringify()`?
> 
> EJSON is expressing BSON using JSON, but syntax like `{ _id: ObjectId('...') }` just isn't JSON, so yeah, this isn't EJSON.
> 
> > * What is the format output by `shell.tojsononeline()`? Furthermore, is it documented?
> 
> I think `tojsononeline()` generally aims to print values in a way where, if the result is evaluated in the shell, you'd get back a 1:1 equivalent object. Yes, the naming of `tojsononeline()` is _horrible_, because it leads users to assume that it returns JSON, which it _not_ what it does.
> 
> > * Is there no existing parser for `shell.tojsononeline()` syntax which required us to modify the existing JSON parser? (@lynne-wang @samontea)
> 
> So, for shell syntax, you can of course use `eval()` in the legacy shell/mongosh if you're working with trusted input. Our team also additionally maintains the [`@mongodb-js/shell-bson-parser`](https://github.com/mongodb-js/devtools-shared/tree/main/packages/shell-bson-parser) package, which we use for handling unsanitized input in environments where we do not want to expose full JS evaluation capabilities.
> 
> > If we want to document this syntax, we need to understand what it is before deciding on a home for it.
> 
> So ... you mentioned above that
> 
> > the proposed format does support {$numberInt: }.
> 
> Is this correct? So this would be an EJSON-shell-syntax hybrid that supports input like `{ int: {$numberInt: '32'}, long: NumberLong('32') }`? IIRC the goal was to be able to represent any BSON object as text, so this would prevent representing a genuine `$numberInt` key in the BSON object, no?
> 
> If this is really a hybrid syntax, I'd strongly consider whether it's possible to change this format so that it aligns with an existing standard (like pure EJSON or pure shell syntax).

dstorch:
For our generated testing, we need a text format which can represent any BSON document unambiguously. Extended JSON does not quite serve this purpose, and also can be a bit verbose so we went with a shell-like text format. It seems valuable to document this BSON text format as a standard, so that in the future test authors (that is, the internal audience of server developers) know how to write down the BSON document that they want to test, including trickier cases like NumberDecimal or various BinData subtypes.

From the previous discussion on this PR, it seems like the mongodb/specifications repo is only for specs relevant to driver authors. If we write a useful standard, one could imagine drivers adopting the BSON text format in the future, but there is no expectation that drivers actually do so anytime soon. Therefore, it sounds like we should move this PR from mongodb/specifications to some other place, like the `docs/` folder in 10gen/mongo as Anna suggested.

> fuzzer generated queries are output using shell.tojsononeline(). the output is not JSON, or EJSON, but Javascript (code?) which can be evaluated in the shell

Is this format documented elsewhere already, like in mongosh documentation? As far as I know there is no such artifact which we can start from, so we need to write the specification from scratch.

> in order to read the queries written using this syntax, the query tester extended the EJSON (json.cpp) parser to support the grammar elements mentioned in the PR. the implementation doesn't admit all possible syntax the shell would accept and additionally admits some constructs (eg. {$numberInt: }) the shell wouldn't treat in the same way.

It actually seems bad to me that the current format uses some constructs from extended JSON. Extended JSON has a goal of representing BSON as JSON which we do not have, so looking like extended JSON could get confusing. Also, that complicates the format because it means there are several quite different ways to represent the same BSON data. How would our BSON text format represent a BSON document that has a key name `$numberInt`, for example? Before this discussion, I assumed that QueryTester always used something a string like `{a: NumberInt(42)}` rather than `{a: {$numberInt: "42"}}`.

PR #1725, opened 2024-11-05
DRIVERS-3036 Update the requirements regarding the size validation

stIncMale:
[Preview](https://github.com/stIncMale/specifications/blob/bulkWriteUpdateSizeValidationRequirements/source/crud/bulk-write.md#size-limits)

- [x] Update changelog.
    clusters, and serverless).

DRIVERS-3036

### DIFF
--- a/source/crud/bulk-write.md
+++ b/source/crud/bulk-write.md
@@ -912,8 +912,8 @@ In accordance with the
 ["_Where possible, depend on server to return errors_"](https://github.com/mongodb/specifications/blob/f8dbd2469f18d093f917efa1f758024bca5d3aaa/source/driver-mantras.md#where-possible-depend-on-server-to-return-errors)
 mantra, drivers should rely on the server validating the size. Such reliance is not possible when a server
 acknowledgement is not expected, but for simplicity and given that unacknowledged writes does not provide a way to know
-whether a write succeeded on the server, it does not seem necessary to help an application with that info
-in the specific situation when the BSON document size limit is exceeded.  
+whether a write succeeded on the server, it does not seem necessary to help an application with that info in the
+specific situation when the BSON document size limit is exceeded.
 
 ## **Changelog**
 
--- a/source/crud/bulk-write.md
+++ b/source/crud/bulk-write.md
@@ -634,15 +634,6 @@ write concern containing the following message:
 
 > Cannot request unacknowledged write concern and ordered writes
 
-### BSON Document Size Limit
-
-The server limits the size of a BSON document. Clients MUST NOT validate the size
-of a BSON document, and MUST rely on server validation, following the
-["Where possible, depend on server to return errors"](https://github.com/mongodb/specifications/blob/f8dbd2469f18d093f917efa1f758024bca5d3aaa/source/driver-mantras.md#where-possible-depend-on-server-to-return-errors)
-mantra. For simplicity, this requirement remains in force even when a server acknowledgement is not expected.
-
-See [SERVER-10643](https://jira.mongodb.org/browse/SERVER-10643) for more details on these size limits.
-
 ## Auto-Encryption
 
 If `MongoClient.bulkWrite` is called on a `MongoClient` configured with `AutoEncryptionOpts`, drivers MUST return an
@@ -914,6 +905,15 @@ number was determined by constructing `OP_MSG` messages with various fields atta
 
 Drivers are required to use this value even if they are capable of determining the exact size of the message prior to
 batch-splitting to standardize implementations across drivers and simplify batch-splitting testing.
+
+### Why is there no requirement to validate the size of a BSON document?
+
+In accordance with the
+["_Where possible, depend on server to return errors_"](https://github.com/mongodb/specifications/blob/f8dbd2469f18d093f917efa1f758024bca5d3aaa/source/driver-mantras.md#where-possible-depend-on-server-to-return-errors)
+mantra, drivers should rely on the server validating the size. Such reliance is not possible when a server
+acknowledgement is not expected, but for simplicity and given that unacknowledged writes does not provide a way to know
+whether a write succeeded on the server, it does not seem necessary to help an application with that info
+in the specific situation when the BSON document size limit is exceeded.  
 
 ## **Changelog**
 
--- a/source/crud/bulk-write.md
+++ b/source/crud/bulk-write.md
@@ -908,12 +908,12 @@ batch-splitting to standardize implementations across drivers and simplify batch
 
 ### Why is there no requirement to validate the size of a BSON document?
 
-In accordance with the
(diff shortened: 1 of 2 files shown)


stIncMale:
@jyemin I dislike how the change turned out - the rules became complex for no good reason. I think the spec should stay unchanged, even if its requirement does not match Java's behavior for the rest of commands, especially given that the current (before the PR) spec requirement follows the [_"Where possible, depend on server to return errors"_](https://github.com/mongodb/specifications/blob/db8d8d181c183c3574d0eb993f069e66739e4f90/source/driver-mantras.md#where-possible-depend-on-server-to-return-errors).

source/crud/bulk-write.md
```diff
@@ -638,15 +638,10 @@ write concern containing the following message:
...
-database (e.g. command documents) have a size limit of `maxBsonObjectSize + 16KiB`. When an acknowledged write concern
-is used, drivers MUST NOT perform any checks related to these size limits and MUST rely on the server to raise an error
-if a limit is exceeded. However, when an unacknowledged write concern is used, drivers MUST raise an error if one of the
-following limits is exceeded:
-
-- The size of a document to be inserted MUST NOT exceed `maxBsonObjectSize`. This applies to the `document` field of an
-    `InsertOneModel` and the `replacement` field of a `ReplaceOneModel`.
-- The size of an entry in the `ops` array MUST NOT exceed `maxBsonObjectSize + 16KiB`.
-- The size of the `bulkWrite` command document MUST NOT exceed `maxBsonObjectSize + 16KiB`.
+database (e.g. command documents) have a size limit of `maxBsonObjectSize + 16KiB`. Clients MUST NOT validate the size
```
jyemin, line 641: Consider just removing the entire section, as it seems unnecessary to mention the size limit only to say drivers must not include any code that enforces it.

stIncMale: I simplified the section, but I think the section should stay for the specification to be explicit regarding the requirement to rely on the server validating the document size.

stIncMale:
Does anyone understand why the pre-commit failed? https://github.com/mongodb/specifications/actions/runs/11696304981/job/32573170376?pr=1725

The only error-related info I see there is

> Error: Process completed with exit code 1.

and it's not helpful.

alcaeus:
> Does anyone understand why the pre-commit failed?

Opening up the details of the pre-commit task shows the diff it produced: https://github.com/mongodb/specifications/actions/runs/11713782255/job/32627198204?pr=1725#step:4:57
```diff
diff --git a/source/crud/bulk-write.md b/source/crud/bulk-write.md
index 569e9a2..8724232 100644
--- a/source/crud/bulk-write.md
+++ b/source/crud/bulk-write.md
@@ -912,8 +912,8 @@ In accordance with the
 ["_Where possible, depend on server to return errors_"](https://github.com/mongodb/specifications/blob/f8dbd2469f18d093f917efa1f758024bca5d3aaa/source/driver-mantras.md#where-possible-depend-on-server-to-return-errors)
 mantra, drivers should rely on the server validating the size. Such reliance is not possible when a server
 acknowledgement is not expected, but for simplicity and given that unacknowledged writes does not provide a way to know
-whether a write succeeded on the server, it does not seem necessary to help an application with that info
-in the specific situation when the BSON document size limit is exceeded.  
+whether a write succeeded on the server, it does not seem necessary to help an application with that info in the
+specific situation when the BSON document size limit is exceeded.
 
 ## **Changelog**
```

@GromNaN since you already added logic to create a commit for updated JSON test files, do you think we can somehow also apply suggestions from the linter? I suppose it's not as easy if both jobs need to make changes, as the commits may happen near-simultaneously and conflict when pushed.

stIncMale:
Thanks, @alcaeus, for pointing out what exactly was wrong. I initially failed to interpret the logs correctly.

source/crud/tests/README.md
```diff
@@ -370,56 +370,15 @@ Assert that a CommandStartedEvent was observed for the `killCursors` command.
 
 ### 10. `MongoClient.bulkWrite` returns error for unacknowledged too-large insert
```
kevinAlbs, line 371: Suggest removing the test name and simplifying to:

```
### 10. (Removed)
```

source/crud/bulk-write.md
```diff
@@ -922,8 +906,19 @@ number was determined by constructing `OP_MSG` messages with various fields atta
 Drivers are required to use this value even if they are capable of determining the exact size of the message prior to
 batch-splitting to standardize implementations across drivers and simplify batch-splitting testing.
 
+### Why is there no requirement to validate the size of a BSON document?
```
kevinAlbs, line 909: Suggest noting a previous version of the spec included size checks. That may help describe why this rationale is here:

```md
Following ["_Where possible, depend on server to return
errors_"](https://github.com/mongodb/specifications/blob/f8dbd2469f18d093f917efa1f758024bca5d3aaa/source/driver-mantras.md#where-possible-depend-on-server-to-return-errors),
drivers should rely on the server to return errors. Such reliance is not possible for unacknowledged writes.

This specification previously required drivers to check size limits for unacknowledged writes. This requirement has
since been removed. Checking size limits complicates some driver implementations. Returning a driver error in this
specific situation does not seem helpful enough to require size checks.
```

kevinAlbs (approved):
Thank you very much for the PR. LGTM with minor suggestions.

--

> Does anyone understand why the pre-commit failed?

[README > Linting](https://github.com/mongodb/specifications/blob/fc7996db26d0ea92091a5034c6acb287ef7282fe/README.md#linting) describes how to run locally. Running the following works for me:

```
pre-commit run --all-files --hook-stage manual mdformat
```

source/crud/bulk-write.md
stIncMale: Removed the section in 125d1921cda0acefb96f8160b42499037515e674.

source/crud/tests/README.md
stIncMale: If someone linked to the test directly, https://github.com/mongodb/specifications/blob/master/source/crud/tests/README.md#10-mongoclientbulkwrite-returns-error-for-unacknowledged-too-large-insert, the link won't work after such a change.

We could have started to use explicit [heading IDs, like in kramdown](https://kramdown.gettalong.org/syntax.html#specifying-a-header-id) to prevent similar issues in the future, but [GitHub Flavored Markdown](https://github.github.com/gfm/) supports them neither for [ATX headings](https://github.github.com/gfm/#atx-headings) nor for [setext headings](https://github.github.com/gfm/#setext-heading).

source/crud/bulk-write.md
stIncMale: Done in f90c80762b3a9a9ea4b98b36275b5d8f7f29147e.

source/crud/tests/README.md
kevinAlbs: Retracting suggestions.

kevinAlbs (approved):

isabelatkinson (approved):

jyemin (approved):

PR #1731, opened 2024-11-14
MONGOSH-1906 Add OIDC workforce auth requirements docs 

addaleax:
Thank you @alexbevi for the support here 🙂


docs/workforce-human-oidc-auth.md
```diff
@@ -0,0 +1,172 @@
...
+
+Using a well-tested and standards-compliant third-party library for core OIDC logic for the respective ecosystem is highly recommended. If this is not possible, implementers need to pay special attention to the specifications referenced in this document.
+
+## Endpoint restrictions
+
+Currently, users who connect to a host other than localhost or an Atlas hostname need to explicitly opt-in into being able to do so by setting the `ALLOWED_HOSTS` flag (specified in the drivers auth spec). In the future, MongoDB is hoping to support Demonstrating Proof of Possession (DPoP, [RFC9449](https://datatracker.ietf.org/doc/html/rfc9449)) that will allow lifting this restriction. The goal here is to prevent users from connecting to untrusted endpoints that will advertise attacker-controlled IdP metadata and and intercept tokens intended for other clusters (or even other OIDC endpoints in general).
+
+We would also like to generally adopt [RFC8707](https://datatracker.ietf.org/doc/html/rfc8707), but have not decided on a specific format for expressing the MongoDB endpoints as resources.
```
addaleax, line 59: This section is pretty much entirely internal-facing, there isn't really anything secret or confidential in here, but it doesn't really apply to outside users

docs/workforce-human-oidc-auth.md
```diff
@@ -0,0 +1,172 @@
...
+   1. To preserve privacy, pages served from this local HTTP server MUST NOT reference external resources directly. They MAY provide external links that require user interaction.
+   2. The HTTP server SHOULD set default headers of `Content-Security-Policy: default-src 'self'` and `Referrer-Policy: no-referrer`, or similarly restrictive defaults.
+3. Compose the URL for Auth Code Flow, using the code challenge generated earlier and the redirect URL from the previous step, with the port replaced with the actual port if it was specified as `0`.
+   1. The application MUST add a `state` parameter containing cryptographically random data ([RFC6749](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.1) specifies this as optional but recommended).
+   2. The application MAY allow the user to indicate that it should add a `nonce` parameter containing cryptographically random data to the authentication request, as defined in the OpenID connect specification, which is then later embedded in the ID token itself.
```
addaleax, line 84: This item is currently under active investigation, [MONGOSH-1905](https://jira.mongodb.org/browse/MONGOSH-1905) is essentially a user request for this, on the other hand we've had tickets like [MONGOSH-1669](https://jira.mongodb.org/browse/MONGOSH-1669) where we've been taking a stance where we don't strictly require OIDC features and are happy with OAuth-2.0-only features. 

The text of the OIDC spec ("If present in the ID Token, Clients MUST verify that the nonce Claim Value is equal to the value of the nonce parameter sent in the Authentication Request.") does imply to me that the IdP cannot, on its own, add a `nonce` value that the client did not specify to begin with and still be considered spec-compliant. 

Now ... _maybe_ adding `nonce`s is something that all relevant identity providers support anyway, in which case we're fine to turn this line into a SHOULD/MUST and remove the user interaction, but we'd likely need to verify this first. So that's why, for now, this is the wording I'm going with here.

pmeredit (approved):
LGTM

.github/CODEOWNERS
```diff
@@ -46,5 +46,8 @@
 # URI Options
 /source/uri-options/ @mongodb/dbx-spec-owners-uri-options
 
+# Workforce OIDC
+/docs/workforce-human-oidc-auth.md @mongodb/dbx-spec-owners-workforce-oidc
```
nirinchev, line 50: reminder to create the team

docs/workforce-human-oidc-auth.md
```diff
@@ -0,0 +1,267 @@
+# Requirements for Client Applications implementing Workforce (Human) OIDC Auth For MongoDB
+
+# Overview
+
+## Abstract
+
+MongoDB offers OpenID Connect (OIDC) authentication and authorization for database users. OIDC auth in clients generally
+falls into one of two categories; either Workflow OIDC targeting programmatic users, which is
```
nirinchev, line 8:
```suggestion
falls into one of two categories: either Workflow OIDC targeting programmatic users, which is
```

Extreme nit and not really a grammar expert, but felt this should be a colon, shouldn't it?

docs/workforce-human-oidc-auth.md
```diff
@@ -0,0 +1,267 @@
...
+highly recommended. If this is not possible, implementers need to pay special attention to the specifications referenced
+in this document.
+
+## Endpoint restrictions
+
+Currently, users who connect to a host other than localhost or an Atlas hostname need to explicitly opt-in into being
+able to do so by setting the `ALLOWED_HOSTS` flag (specified in the drivers auth spec). In the future, MongoDB is hoping
+to support Demonstrating Proof of Possession (DPoP, [RFC9449](https://datatracker.ietf.org/doc/html/rfc9449)) that will
+allow lifting this restriction. The goal here is to prevent users from connecting to untrusted endpoints that will
+advertise attacker-controlled IdP metadata and and intercept tokens intended for other clusters (or even other OIDC
```
nirinchev, line 83:
```suggestion
advertise attacker-controlled IdP metadata and intercept tokens intended for other clusters (or even other OIDC
```

nirinchev (approved):

.github/CODEOWNERS
addaleax: It does, and I'm not sure why GH is complaining: https://github.com/orgs/mongodb/teams/dbx-spec-owners-workforce-oidc

addaleax: Ah, looks like the parent team needs to be set to dbx-spec-reviewers, I've requested that + the request is pending rn

docs/workforce-human-oidc-auth.md
```diff
@@ -0,0 +1,267 @@
...
+- IdP metadata discovery MUST be implemented as described in [RFC8414](https://datatracker.ietf.org/doc/html/rfc8414).
+
+Using a well-tested and standards-compliant third-party library for core OIDC logic for the respective ecosystem is
+highly recommended. If this is not possible, implementers need to pay special attention to the specifications referenced
+in this document.
+
+## Endpoint restrictions
+
+Currently, users who connect to a host other than localhost or an Atlas hostname need to explicitly opt-in into being
+able to do so by setting the `ALLOWED_HOSTS` flag (specified in the drivers auth spec). In the future, MongoDB is hoping
```
JamesKovacs, line 80: Awkward wording:
>explicitly opt-in into being able to do so by setting the `ALLOWED_HOST` flag

Maybe something like this instead:
>explicitly specify this host via `ALLOWED_HOSTS`

docs/workforce-human-oidc-auth.md
```diff
@@ -0,0 +1,267 @@
...
+to support Demonstrating Proof of Possession (DPoP, [RFC9449](https://datatracker.ietf.org/doc/html/rfc9449)) that will
+allow lifting this restriction. The goal here is to prevent users from connecting to untrusted endpoints that will
```
JamesKovacs, line 82: I suggest omitting this forward-looking statement about implementing DPoP.

docs/workforce-human-oidc-auth.md
```diff
@@ -0,0 +1,267 @@
...
+advertise attacker-controlled IdP metadata and intercept tokens intended for other clusters (or even other OIDC
+endpoints in general).
+
+We would also like to generally adopt [RFC8707](https://datatracker.ietf.org/doc/html/rfc8707), but have not decided on
+a specific format for expressing the MongoDB endpoints as resources.
```
JamesKovacs, line 87: Same here as we may never implement this feature.

docs/workforce-human-oidc-auth.md
```diff
@@ -0,0 +1,267 @@
...
+than the access token, in its place.[^2]
+
+## OIDC Scopes
+
+The initial request to the IdP involves specifying the `scope` parameter. This is a space-separated unordered list of
+strings which contains:
+
+- The scopes listed in the `requestScopes` field of the IdP response from the MongoDB endpoint, and
+- The `openid` and `offline_access` scopes[^3]. If the IdP metadata document contains a `scopes_supported` field, and
+    this field does not list one of these scopes, then the respective scope SHOULD NOT be added to the `scope`
```
JamesKovacs, line 107: Is this a SHOULD NOT or a MUST NOT? Does auth fail if an unsupported scope is provided?

docs/workforce-human-oidc-auth.md
```diff
@@ -0,0 +1,267 @@
...
+    parameter.
+
+# Authorization Code Flow
+
+In this flow, the client application uses a local HTTP server to receive a response from the Identity Provider, as
+described in [RFC8252](https://datatracker.ietf.org/doc/html/rfc8252).[^4] The application follows these steps:
+
+1. Generate a code challenge for PKCE using cryptographically random data, as described in
+    [RFC7636](https://datatracker.ietf.org/doc/html/rfc7636).
+2. Spin up a local HTTP server. The default (incoming) redirect URL for MongoDB applications is
```
JamesKovacs, line 117: "Spin up" is rather colloquial. "Launch a local HTTP server."

JamesKovacs (changes requested):
Overall fantastic. Requesting some minor changes.

docs/workforce-human-oidc-auth.md
addaleax: Thanks, absolutely right – done in b66f9609e7e608953241aa43a1711865f1f6c1ab 👍

docs/workforce-human-oidc-auth.md
addaleax: I've moved these into an appendix section at the end of the file and cleaned them up a bit. I _do_ think that these are items that implementers should be aware of, especially considering that "accidentally" implementing these may be counterproductive – depending on the OIDC library used, it may be quite easy to do so, but there would be a risk of doing it in a way that is incompatible with what we end up standardizing on in the future.

It's a good reminder, though. In particular RFC8707 adoption is essentially just a client-side thing that we need to figure out and shouldn't require server support, so I've opened https://jira.mongodb.org/browse/MONGOSH-1937 to make sure we do end up investigating this option.

docs/workforce-human-oidc-auth.md
addaleax: See above :)

docs/workforce-human-oidc-auth.md
addaleax: Hm, yeah, it's a good question, and I've gone a bit back and forth on this in response to you bringing it up 🙂 

I think the RFC2119 definition of "SHOULD NOT" is ultimately the right one here. The implementer would need to accept that their application's authentication attempts could be rejected by an IdP that doesn't provide refresh tokens (as that is what `offline_access` asks for). That seems like a sensible behavior for an application that strictly requires ongoing access to the cluster while it is active – e.g. mongosync might want to make this choice (Not saying they do, but that it would be understandable if they did).

That being said, if you feel strongly here, I'm still happy to make this change.

> Does auth fail if an unsupported scope is provided?

It's up to the individual IdP (this sentence referring to the IdP/server side of OIDC):

> Scope values used that are not understood by an implementation SHOULD be ignored.

addaleax:
@JamesKovacs gentle bump on this one 🙂

PR #1733, opened 2024-11-19
DRIVERS-2862: Benchmark Collection and Client BulkWrite

adelinowona:
The PR adds benchmarks for `Collection::BulkWrite` and `Client::BulkWrite` with insert-only operations and mixed operations. We already have a `Small doc Bulk Insert` and `Large doc Bulk Insert` benchmark in the benchmarking spec which may be sufficient for benchmarking `Collection::BulkWrite` with insert-only operations. However, the `Small doc Bulk Insert` and `Large doc Bulk Insert` benchmarks are implemented using `insertMany` and not all the drivers use `Collection::BulkWrite` in their implementation of `insertMany`. With that in mind, I have added explicit `Collection::BulkWrite` benchmarks to be implemented by all drivers who implement `Collection::BulkWrite`. This ensures we still maintain comprehensive performance testing for our batch-write performance.

Here are results of the new benchmarks as implemented on the C# driver:
<markdown-accessiblity-table data-catalyst="" style="box-sizing: border-box; display: block; color: rgb(240, 246, 252); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, &quot;Noto Sans&quot;, Helvetica, Arial, sans-serif, &quot;Apple Color Emoji&quot;, &quot;Segoe UI Emoji&quot;; font-size: 14px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; white-space: normal; background-color: rgb(13, 17, 23); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">
Name | MB/sec
-- | --
SmallDocClientBulkWriteMixedOps | 1.9320055323
SmallDocCollectionBulkWriteMixedOps | 0.7543254574
LargeDocClientBulkWriteInsert | 94.103463269
LargeDocCollectionBulkWriteInsert | 96.6259091508
SmallDocClientBulkWriteInsert | 36.7585335938
SmallDocCollectionBulkWriteInsert | 40.8752903883

</markdown-accessiblity-table><br class="Apple-interchange-newline">

- [ ] Update changelog.
    clusters, and serverless).


source/benchmarking/benchmarking.md
```diff
@@ -410,36 +416,144 @@ Dataset size: For score purposes, the dataset size for a task is the size of the
...
+times 20,000 operations (insert + replace), which equals 5,500,000 bytes or 5.5 MB.
+
+| Phase       | Description                                                                                                                                                                                                                                                                                                       |
+| ----------- |-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| Setup       | Construct a MongoClient object. Drop the `perftest` database. Load the SMALL_DOC dataset into memory as a language-appropriate document type (or JSON string for C). Make 10,000 copies of the document. DO NOT manually add an `_id` field; leave it to the driver or database. Construct a list of write models with insert, replace and delete operations for each copy of the document. |
```
BorisDog, line 546: I think we need more details here: ordering and explicit count for each operation kind.

BorisDog (changes requested):
partial review

ShaneHarvey (commented):
I have two questions:

1. What is the benefit of adding the mixed op benchmarks? I don't see much value in adding them.
1. Do we need to add the "collection bulkWrite insert" benchmarks?  It seems like they little value since we already have the existing insertMany ones.

I'm wary of introducing extra benchmarks because each once has a cost.

source/benchmarking/benchmarking.md
adelinowona: Yep will do! In the C# benchmarks, my proposed sequence involves performing an insert operation, followed by a replace, and then a delete operation for each document copy. Both the replace and delete operations use empty filters as well.

adelinowona:
@ShaneHarvey, we're adding these benchmarks as part of the Multi-Doc suite to evaluate batch-write efficiency. The Multi-Doc benchmarks should reflect real-world use cases, including mixed operation benchmarks. Without these, we risk leaving a gap in our performance analysis. Including them now gives us a baseline to track future improvements or regressions in the implementations that support mixed operations in `bulkWrite`.

Regarding the 'collection bulkWrite insert' benchmarks, I understand that drivers using `Collection::BulkWrite` for `insertMany`, or sharing code between these implementations, might question the need for these benchmarks. However, for thorough performance coverage, it's important to benchmark both APIs for batch inserts. Even if their implementations are similar now, they could diverge with future driver updates. Having these benchmarks allows us to identify any current performance differences and provides a baseline for comparison. Additionally, they offer a direct contrast to the 'client bulkWrite insert' benchmarks, facilitating easier performance comparisons between these APIs.

ShaneHarvey:
Fair points. I'm still skeptical on the mixed benchmarks. To me it seems like all we're measuring is the cost of extra roundtrips to the server but I suppose it could find a regression at some point.

source/benchmarking/benchmarking.md
```diff
@@ -410,36 +416,144 @@ Dataset size: For score purposes, the dataset size for a task is the size of the
...
+| Setup       | Construct a MongoClient object. Drop the `perftest` database. Load the SMALL_DOC dataset into memory as a language-appropriate document type (or JSON string for C). Create a list of insert models for 10 copies of the document, allowing the driver or database to auto-assign the `_id` field. |
+| Before task | Drop the `corpus` collection. Create an empty `corpus` collection with the `create` command. Construct a Collection object for the `corpus` collection to use for insertion.                                                                                                                       |
+| Do task     | Do an ordered `Collection::bulkWrite` in the same namespace ("perftest.corpus") with the list of insert models.                                                                                                                                                                                    |
```
BorisDog, line 458: Is namespace relevant here?

source/benchmarking/benchmarking.md
```diff
@@ -410,36 +416,144 @@ Dataset size: For score purposes, the dataset size for a task is the size of the
...
+| Setup       | Construct a MongoClient object. Drop the `perftest` database. Load the LARGE_DOC dataset into memory as a language-appropriate document type (or JSON string for C). Create a list of insert models for 10 copies of the document, allowing the driver or database to auto-assign the `_id` field. |
+| Before task | Drop the `corpus` collection. Create an empty `corpus` collection with the `create` command. Construct a Collection object for the `corpus` collection to use for insertion.<br/>                                                                                                                  |
+| Do task     | Do an ordered `Client::bulkWrite` in the same namespace ("perftest.corpus") with the list of insert models.                                                                                                                                                                                        |
```
BorisDog, line 512: I think the namespace is part of the model construction/setup phase.

BorisDog (changes requested):

source/benchmarking/benchmarking.md
adelinowona: Yep namespace isn't relevant here for `Collection:bulkWrite` since you can't have operations on different collections.

source/benchmarking/benchmarking.md
adelinowona: Since the model construction is very much linked with namespace selection, I thought implementers will understand that if even I mention the namespace requirements in the Do Task, it will be done in the setup phase. But I'll move the namespace requirement to the setup phase to be more direct.

BorisDog (approved):
LGTM!

source/benchmarking/benchmarking.md
```diff
@@ -611,12 +725,15 @@ TBD: spec system to hold scores over time
 TBD: generated datasets should be park in S3 or somewhere for retrieval by URL
 
 ## Changelog
+- 2024-11-19:
+    - Add Client and Collection BulkWrite benchmarks
+    - Use inline code syntax for all text in single quotes.
```
ShaneHarvey, line 730: This entry is causing the lint check to fail: https://github.com/mongodb/specifications/actions/runs/12436624595/job/34724832915?pr=1733

We can resolve it by removing the nested list (no need to mention the single quote change):
```
- 2024-12-XX: Add Client and Collection BulkWrite benchmarks

```

adelinowona: Thanks, I've updated it but the lint check is still failing. I am not sure as to what causes the error

ShaneHarvey: Do you have pre-commit set up locally? https://github.com/mongodb/specifications?tab=readme-ov-file#linting

adelinowona: Ah thanks! I have that sorted out now!

ShaneHarvey (approved):

PR #1735, opened 2024-11-22
DRIVERS-2798 Drivers do not gossip $clusterTime on SDAM commands

ShaneHarvey:

- [X] Update changelog.
    clusters, and serverless).

### DIFF
--- a/source/sessions/driver-sessions.md
+++ b/source/sessions/driver-sessions.md
@@ -934,7 +934,7 @@ has risks that do not justify the potential guaranteed `ServerSession` allocatio
 
 ## Changelog
 
-- 2024-11-21: Drivers MUST NOT gossip $clusterTime on SDAM commands.
+- 2025-02-24: Drivers MUST NOT gossip $clusterTime on SDAM commands.
 - 2024-05-08: Migrated from reStructuredText to Markdown.
 - 2017-09-13: If causalConsistency option is omitted assume true
 - 2017-09-16: Omit session ID when opening and authenticating a connection
--- a/source/sessions/driver-sessions.md
+++ b/source/sessions/driver-sessions.md
@@ -710,7 +710,8 @@ expiration.
 
 ## Gossipping the cluster time
 
-Drivers MUST gossip the cluster time when connected to a deployment that uses cluster times.
+Drivers MUST gossip the cluster time when connected to a deployment that uses cluster times. However, drivers MUST NOT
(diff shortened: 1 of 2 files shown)


source/sessions/driver-sessions.md
```diff
@@ -710,7 +710,8 @@ expiration.
 
 ## Gossipping the cluster time
 
-Drivers MUST gossip the cluster time when connected to a deployment that uses cluster times.
+Drivers MUST gossip the cluster time when connected to a deployment that uses cluster times. However, drivers MUST NOT
+gossip cluster time on SDAM commands.
```
jyemin, line 714: Should this link to the SDAM spec in some way, so that it's clear what SDAM commands refers to?

source/sessions/tests/README.md
```diff
@@ -238,6 +238,20 @@ and configure a `MongoClient` with default options.
 - Attempt to send a write command to the server (e.g., `insertOne`) with the explicit session passed in
 - Assert that a client-side error is generated indicating that sessions are not supported
 
+### 20. Drivers do not gossip `$clusterTime` on SDAM commands.
```
jyemin, line 241: I'm unsure about this test.  Previously we didn't have a test that asserts that cluster time _is_ gossiped by SDAM commands.  And now we have a test that it is _not_.  When we fixed this issue in the java driver we didn't end up adding an integration test for it, because it was fixed by design: in the Java driver in order to gossip cluster time you need access to the client-wide instance of `ClusterClock`.  In the fix we just removed the `ClusterClock` from the server monitor, so there in way it could unintentionally gossip.

That said, for other drivers I guess this test might be a useful one, so if you want to leave it I'm ok with that.

source/sessions/driver-sessions.md
ShaneHarvey: I followed the pattern in the load balancer spec and added "SDAM" to the terms section.

source/sessions/tests/README.md
ShaneHarvey: I'd prefer to keep the test. It does provide some value and it at least saves implementors' time trying to answer "how should I test this? Can I even test it?".

sanych-sun (approved):
LGTM

qingyang-hu (approved):
👍

jyemin (approved):

jyemin (approved):

PR #1738, opened 2024-11-27
Relax the WaitQueue fairness requirement in the CMAP spec

stIncMale:
The relevant Slack discussion: https://mongodb.slack.com/archives/C72LB5RPV/p1732060999166209.

DRIVERS-3056

- [x] Update changelog.
    clusters, and serverless).

### DIFF
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
@@ -1262,9 +1262,9 @@ from occurring.
 
 Additionally, these options have the effect of prioritizing newer requests over older requests, which is not necessarily
 the behavior that users want. For example, consider a situation when a waitQueue is full, and a request for
-[Connection](#connection) gets rejected. Then a spot opens in the waitQueue, and a newer request gets accepted.
-One may say that the newer request was prioritized over the older one, which violates the fairness recommendation that
-the waitQueue normally adheres to.
+[Connection](#connection) gets rejected. Then a spot opens in the waitQueue, and a newer request gets accepted. One may
+say that the newer request was prioritized over the older one, which violates the fairness recommendation that the
+waitQueue normally adheres to.
 
 Because of these issues, it does not make sense to
 [go against driver mantras and provide an additional knob](../driver-mantras.md#). We may eventually pursue an
--- a/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
+++ b/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
@@ -247,7 +247,7 @@ either receives a [Connection](#connection) or times out. A WaitQueue has the fo
 - **Ordered/fair**: When [Connections](#connection) are made available, they SHOULD be issued out to threads in the
     order that the threads entered the WaitQueue. If this is behavior poses too much of an implementation burden, then
     at the very least threads that have entered the queue more recently MUST NOT be intentionally prioritized over those
-    that entered it earlier. 
+    that entered it earlier.
 - **Timeout aggressively:** Members of a WaitQueue MUST timeout if they are enqueued for longer than the computed
     timeout and MUST leave the WaitQueue immediately in this case.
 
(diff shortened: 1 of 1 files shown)


source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -1263,7 +1265,7 @@ the behavior that users want. They can also result in cases where queue access o
 and not full. If a driver has a full waitQueue, then all requests for [Connections](#connection) will be rejected. If
 the client is continually spammed with requests, you could wind up with a scenario where as soon as the waitQueue is no
 longer full, it is immediately filled. It is not a favorable situation to be in, partially b/c it violates the fairness
-guarantee that the waitQueue normally provides.
+recommendation that the waitQueue normally adheres to.
```
stIncMale, line 1268: Does anyone understand what's written here?

> prioritizing older requests over newer requests, which is not necessarily the behavior that users want

That's literally fairness, which was required (and now will be recommended) by the spec. Yet here it is referred to as something bad.

> It is not a favorable situation to be in, partially b/c it violates the fairness
recommendation that the waitQueue normally adheres to.

I fail to understand how fairness is violated because of `waitQueueSize`. Not arguing for `waitQueueSize`, just want to understand what is written in this section. Could someone please help me?

ShaneHarvey: I believe what this is trying to say is that if you look at all requests in the "oscillates and forth between full
and not full" scenario, then we can say that those requests were not handled in a fair manner. For example, here's a specific scenario:

1. wait queue is full
1. new request comes in at time T1, it is rejected immediately because the wait queue is full
1. wait queue decreases in size
1. new request comes in at time T2, it is accepted into the wait queue.

A newer request (T2) is prioritized over an older request (T1) which violates fairness. With this in mind, I believe this is a typo and it's supposed to say "prioritizing newer requests over older requests".

ShaneHarvey: If you agree, could you add this example to help future readers?

stIncMale: Thank you, @ShaneHarvey! Do you think the change made in https://github.com/mongodb/specifications/pull/1738/commits/9becee2deb1d07c9c8c42306506f3dfe72ebd136 clarifies this adequately?

baileympearson (approved):

ShaneHarvey (approved):
LGTM!

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
ShaneHarvey: Thanks, that's great.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -1258,12 +1260,11 @@ These options were originally only implemented in three drivers (Java, C#, and P
...
-the behavior that users want. They can also result in cases where queue access oscillates back and forth between full
-and not full. If a driver has a full waitQueue, then all requests for [Connections](#connection) will be rejected. If
-the client is continually spammed with requests, you could wind up with a scenario where as soon as the waitQueue is no
-longer full, it is immediately filled. It is not a favorable situation to be in, partially b/c it violates the fairness
-guarantee that the waitQueue normally provides.
+Additionally, these options have the effect of prioritizing newer requests over older requests, which is not necessarily
+the behavior that users want. For example, consider a situation when a waitQueue is full, and a request for
+[Connection](#connection) gets rejected. Then a spot opens in the waitQueue, and a newer request gets accepted. One may
+say that the newer request was prioritized over the older one, which violates the fairness recommendation that the
+waitQueue normally adheres to.
```
matthewdale, line 1267: Optional: Consider making the capitalization of WaitQueue in this paragraph consistent with (most of) the rest of the doc.

```suggestion
Additionally, these options have the effect of prioritizing newer requests over older requests, which is not necessarily
the behavior that users want. For example, consider a situation when a WaitQueue is full, and a request for
[Connection](#connection) gets rejected. Then a spot opens in the WaitQueue, and a newer request gets accepted. One may
say that the newer request was prioritized over the older one, which violates the fairness recommendation that the
WaitQueue normally adheres to.
```

matthewdale (approved):
Looks good, with a capitalization suggestion! 👍

PR #1742, opened 2025-01-09
DRIVERS-3081: Relax requirement for optional fields and introducing appName for failPoints for transactions unified tests

sanych-sun:
While migrating to unified spec format for transactions tests in CSHARP Driver I've run into 2 minor problems which I suggest to fix by this PR (please let me know if the changes makes sense and I'll create an appropriate ticket):
1) CSharp Driver does not write default values for optional fields, that's why we should make 'new' field of `findOneAndModify` command optional (`{ $$unsetOrMatches: false }`)
2) When there is a failPoint set on `hello` command, I suggest to use appName to let other connection work (we use additional client to setup and tire down the failpoints).

### DIFF
--- a/source/transactions/tests/unified/pin-mongos.yml
+++ b/source/transactions/tests/unified/pin-mongos.yml
@@ -527,6 +527,7 @@ tests:
                   # flight heartbeat refreshes the first mongoes' SDAM state in between
                   # the insert connection error and the single commit attempt.
                   heartbeatFrequencyMS: 30000
+                  appName: &appName transactionsClient
                 observeEvents:
                   - commandStartedEvent
             - database:
@@ -572,6 +573,7 @@ tests:
                 - isMaster
                 - hello
               closeConnection: true
+              appName: *appName
       -
         object: *collection1
         name: insertOne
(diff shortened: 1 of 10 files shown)


ShaneHarvey:
Could you open a DRIVERS ticket for this change and link it here?

source/transactions/tests/unified/pin-mongos.yml
```diff
@@ -572,6 +573,7 @@ tests:
                 - isMaster
                 - hello
               closeConnection: true
+              appName: *appName
```
ShaneHarvey, line 576: I think these transaction test changes look good. I believe couldn't use this feature originally because of this server bug: SERVER-49336. Can you confirm that these tests pass on current 4.2 and up servers?

sanych-sun: Yes, the tests are green.

sanych-sun: There are 2 possible fixes for now: use appName with additional connection for failPoints OR simply wait for all failPoints to be processed and then remove fialPoint (the second approach makes us waste around of 3-4 seconds waiting)

ShaneHarvey (approved):
Tests pass in Python, LGTM!

PR #1743, opened 2025-01-13
feat(DRIVERS-2903): use custom aws configuration

durran:
Adds section to auth spec and CSFLE spec on allowing user provided custom credential providers for AWS and notes on testing.

Node implementation: https://github.com/mongodb/node-mongodb-native/pull/4373

- [x] Update changelog.
    clusters, and serverless).

### DIFF
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -1325,7 +1325,9 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
 
     - AWS_CREDENTIAL_PROVIDER
 
-        A function or object from the AWS SDK that can be used to return AWS credentials.
+        A function or object from the AWS SDK that can be used to return AWS credentials. Drivers MAY allow the user to
+        specify the callback using a `MongoClient` configuration instead of a mechanism property, depending on what is
+        idiomatic for the driver.
 
 <span id="built-in-provider-integrations"/>
 
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -995,7 +995,7 @@ and drivers MUST consult AWS SDK documentation to determine that format when imp
 be `AWS_CREDENTIAL_PROVIDER` and be part of the authentication mechanism properties options that can be provided to the
 client.
 
-Drivers that implement this MAY choose to implement the following scenarios when applicable in their labguage's SDK:
+Drivers that implement this MAY choose to implement the following scenarios when applicable in their language's SDK:
 
 1. The default SDK credential provider.
 2. A custom credential provider chain.
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -1007,6 +1007,7 @@ The order in which Drivers MUST search for credentials is:
 
 1. The URI
 2. Environment variables
+3. A custom AWS credential provider if the driver supports it.
 3. Using `AssumeRoleWithWebIdentity` if `AWS_WEB_IDENTITY_TOKEN_FILE` and `AWS_ROLE_ARN` are set.
 4. The ECS endpoint if `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` is set. Otherwise, the EC2 endpoint.
 
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -1008,8 +1008,8 @@ The order in which Drivers MUST search for credentials is:
 1. The URI
 2. Environment variables
 3. A custom AWS credential provider if the driver supports it.
-3. Using `AssumeRoleWithWebIdentity` if `AWS_WEB_IDENTITY_TOKEN_FILE` and `AWS_ROLE_ARN` are set.
-4. The ECS endpoint if `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` is set. Otherwise, the EC2 endpoint.
+4. Using `AssumeRoleWithWebIdentity` if `AWS_WEB_IDENTITY_TOKEN_FILE` and `AWS_ROLE_ARN` are set.
+5. The ECS endpoint if `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` is set. Otherwise, the EC2 endpoint.
 
 > [!NOTE]
 > See *Should drivers support accessing Amazon EC2 instance metadata in Amazon ECS* in [Q & A](#q-and-a)
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -989,10 +989,10 @@ application. Alternatively, you can create an AWS profile specifically for your 
 
 ##### Custom Credential Providers
 
-Drivers that choose to use the AWS SDK to fetch credentials MAY also allow users to provide a custom credential
-provider as an option to the `MongoClient`. The interface for the option provided depends on the individual language SDK
-and drivers MUST consult AWS SDK documentation to determine that format when implementing. The name of the option MUST
-be `AWS_CREDENTIAL_PROVIDER` and be part of the authentication mechanism properties options that can be provided to the
+Drivers that choose to use the AWS SDK to fetch credentials MAY also allow users to provide a custom credential provider
+as an option to the `MongoClient`. The interface for the option provided depends on the individual language SDK and
+drivers MUST consult AWS SDK documentation to determine that format when implementing. The name of the option MUST be
+`AWS_CREDENTIAL_PROVIDER` and be part of the authentication mechanism properties options that can be provided to the
 client.
 
 Drivers that implement this MAY choose to implement the following scenarios when applicable in their language's SDK:
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -987,6 +987,22 @@ those credentials will be used by default if AWS auth environment variables are 
 application. Alternatively, you can create an AWS profile specifically for your MongoDB credentials and set the
 `AWS_PROFILE` environment variable to that profile name."
 
+##### Custom Credential Providers
+
+Drivers that choose you use the AWS SDK to fetch credentials MAY also allow users to provide a custom credential
+provider as an option to the `MongoClient`. The interface for the option provided depends on the individual language SDK
+and drivers MUST consult AWS SDK documentation to determine that format when implementing. The name of the option MUST
+be `AWS_CREDENTIAL_PROVIDER` and be part of the authentication mechanism properties options that can be provided to the
+client.
+
+Drivers that implement this MAY choose to implement the following scenarios when applicable in their labguage's SDK:
+
+1. The default SDK credential provider.
+2. A custom credential provider chain.
+3. A single credential provider of any available SDK options provided by the SDK.
+
+##### Credential Fetching Order
+
 The order in which Drivers MUST search for credentials is:
 
 1. The URI
@@ -1305,6 +1321,10 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
         invoking any user-provided callbacks. This value MUST NOT be allowed in the URI connection string. The hostname
         check MUST be performed after SRV record resolution, if applicable. This property is only required for drivers
         that support the [Human Authentication Flow](#human-authentication-flow).
+
+    - AWS_CREDENTIAL_PROVIDER
+
+        A function or object from the AWS SDK that can be used to return AWS credentials.
 
 <span id="built-in-provider-integrations"/>
 
@@ -2133,6 +2153,8 @@ practice to avoid this. (See
 [IAM Roles for Tasks](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html))
 
 ## Changelog
+
+- 2025-01-29: Add support for custom AWS credential providers.
 
 - 2024-10-02: Add Kubernetes built-in OIDC provider integration.
 
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -995,7 +995,7 @@ drivers MUST consult AWS SDK documentation to determine that format when impleme
 `AWS_CREDENTIAL_PROVIDER` and be part of the authentication mechanism properties options that can be provided to the
 client.
 
-Drivers that implement this MAY choose to implement the following scenarios when applicable in their language's SDK:
+Drivers MAY expose API for default providers for the following scenarios when applicable in their language's SDK:
 
 1. The default SDK credential provider.
 2. A custom credential provider chain.
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -989,7 +989,7 @@ application. Alternatively, you can create an AWS profile specifically for your 
 
 ##### Custom Credential Providers
 
-Drivers that choose you use the AWS SDK to fetch credentials MAY also allow users to provide a custom credential
+Drivers that choose to use the AWS SDK to fetch credentials MAY also allow users to provide a custom credential
 provider as an option to the `MongoClient`. The interface for the option provided depends on the individual language SDK
 and drivers MUST consult AWS SDK documentation to determine that format when implementing. The name of the option MUST
 be `AWS_CREDENTIAL_PROVIDER` and be part of the authentication mechanism properties options that can be provided to the
--- a/source/client-side-encryption/client-side-encryption.md
+++ b/source/client-side-encryption/client-side-encryption.md
@@ -480,8 +480,8 @@ See
 
 The `credentialProviders` property may be specified on [ClientEncryptionOpts](#ClientEncryptionOpts) or
 [AutoEncryptionOpts](#AutoEncryptionOpts). Current support is for AWS only, but is designed to be able to accomodate
-additional providers in the future. If a custom credential provider is present, it MUST be used instead of the
-default flow for fetching automatic credentials.
+additional providers in the future. If a custom credential provider is present, it MUST be used instead of the default
+flow for fetching automatic credentials.
 
 ```typescript
 interface CredentialProviders {
@@ -498,10 +498,10 @@ The following shows an example object of `CredentialProviders` for Node.js:
 ```typescript
 import { fromNodeProviderChain } from '@aws-sdk/credential-providers';
 
-{
-   // Acquire credentials for AWS:
-   aws: fromNodeProviderChain()
-}
+const creentialProviders = {
+  // Acquire credentials for AWS:
+  aws: fromNodeProviderChain()
+};
 ```
 
 #### kmsProviders
--- a/source/client-side-encryption/client-side-encryption.md
+++ b/source/client-side-encryption/client-side-encryption.md
@@ -481,7 +481,8 @@ See
 The `credentialProviders` property may be specified on [ClientEncryptionOpts](#ClientEncryptionOpts) or
 [AutoEncryptionOpts](#AutoEncryptionOpts). Current support is for AWS only, but is designed to be able to accommodate
 additional providers in the future. If a custom credential provider is present, it MUST be used instead of the default
-flow for fetching automatic credentials.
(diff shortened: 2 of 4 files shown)


source/auth/auth.md
```diff
@@ -987,6 +987,22 @@ those credentials will be used by default if AWS auth environment variables are
 application. Alternatively, you can create an AWS profile specifically for your MongoDB credentials and set the
 `AWS_PROFILE` environment variable to that profile name."
 
+##### Custom Credential Providers
+
+Drivers that choose you use the AWS SDK to fetch credentials MAY also allow users to provide a custom credential
```
jyemin, line 992:
```suggestion
Drivers that choose to use the AWS SDK to fetch credentials MAY also allow users to provide a custom credential
```

source/auth/auth.md
```diff
@@ -987,6 +987,22 @@ those credentials will be used by default if AWS auth environment variables are
...
+provider as an option to the `MongoClient`. The interface for the option provided depends on the individual language SDK
+and drivers MUST consult AWS SDK documentation to determine that format when implementing. The name of the option MUST
+be `AWS_CREDENTIAL_PROVIDER` and be part of the authentication mechanism properties options that can be provided to the
+client.
+
+Drivers that implement this MAY choose to implement the following scenarios when applicable in their labguage's SDK:
```
jyemin, line 998:
```suggestion
Drivers that implement this MAY choose to implement the following scenarios when applicable in their language's SDK:
```

jyemin (changes requested):
Just a few quick edits while I perused this.

jyemin (changes requested):
Since this needs to be supported in FLE as well, I think we need to update the FLE spec as well and specify how to configure the "aws" kms provider. Consider adding a similar property with same semantics to `AWSKMSOptions` in https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.md#kmsproviders.

(While in some drivers you might be able to apply an auth mechanism property to FLE, in Java that isn't possible as auth mechanism properties are associated with a credential, not the MongoClient, and the credential of the MongoClient might not even be an AWS credential (or could be a different one).

durran:
> Since this needs to be supported in FLE as well, I think we need to update the FLE spec as well and specify how to configure the "aws" kms provider. Consider adding a similar property with same semantics to `AWSKMSOptions` in https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.md#kmsproviders.
> 
> (While in some drivers you might be able to apply an auth mechanism property to FLE, in Java that isn't possible as auth mechanism properties are associated with a credential, not the MongoClient, and the credential of the MongoClient might not even be an AWS credential (or could be a different one).

@kevinAlbs Do you have thoughts on this? Would there be a case where auto-encryption could want a different AWS credential provider than the associated MongoClient. (With ClientEncryption this is already possible since it has its own MongoClient)

baileympearson:
@durran What about users who aren't using AWS authentication for their clients but want to use AWS automatic KMS credential fetching with a custom provider?  I don't think Node would support that without a provider option for KMS specifically.

baileympearson:
@durran What about users who aren't using AWS authentication for their clients but want to use AWS automatic KMS credential fetching with a custom provider?  I don't think Node would support that without a provider option for KMS specifically.

durran:
> @durran What about users who aren't using AWS authentication for their clients but want to use AWS automatic KMS credential fetching with a custom provider? I don't think Node would support that without a provider option for KMS specifically.

That's a good point and I think a valid use case. I'll update everything accordingly.

source/client-side-encryption/client-side-encryption.md
```diff
@@ -1277,6 +1283,11 @@ non-applicable queryType.
 rangeOpts only applies when algorithm is "Range". libmongocrypt returns an error if rangeOpts is set for a
 non-applicable algorithm.
 
+### Custom AWS Credential Provider
+
+When using a `ClientEncryption` object directly and a custom AWS credential provider is specified as an auth mechanism
```
baileympearson, line 1288: It seems slightly odd to pull auth mechanism information from the keyvault client to use on the client encryption for KMS requests.  KMS requests are separate from client auth.  What about users who are using non-AWS auth for their keyvault client (i.e., SCRAM auth) but want to use a custom kms credential provider for KMS requests on their client encryption?

Would it make more sense instead to completely ignore the auth settings on the keyvault client and instead allow ClientEncryption objects to take custom AWS credential providers as an option?

durran: Yeah that makes sense, and I think it simplifies this even more. I'll update.

durran: I've updated and changed the Node implementation as well. I followed a similar pattern to `KMSProviders` and created a new `CredentialProviders` option that can be provided both to the client encryption options and the auto encryption options. Currently only AWS is supported but I did it in a manner that we could expand to GCP and Azure in the future.

blink1073:
cc @kevinAlbs since this affects Client Encryption options

source/client-side-encryption/client-side-encryption.md
```diff
@@ -472,6 +473,34 @@ See
...
+additional providers in the future. If a custom credential provider is present, it MUST be used instead of the default
+flow for fetching automatic credentials.
+
+```typescript
+interface CredentialProviders {
+   aws? AWSCredentialProvider
+}
+
+// The type of the AWS credential provider is dictated by the AWS SDK's credential provider for the specific
+// lamguage.
```
baileympearson, line 489:
```suggestion
// language.
```

source/client-side-encryption/client-side-encryption.md
```diff
@@ -472,6 +473,34 @@ See
...
+type AWSCredentialProvider = Function | Object;
+```
+
+The following shows an example object of `CredentialProviders` for Node.js:
+
+```typescript
+import { fromNodeProviderChain } from '@aws-sdk/credential-providers';
+
+const creentialProviders = {
```
baileympearson, line 498:
```suggestion
const credentialProviders = {
```

source/client-side-encryption/client-side-encryption.md
```diff
@@ -472,6 +473,34 @@ See
 <span id="GCPKMSOptions"></span> <span id="AWSKMSOptions"></span> <span id="KMSProvider"></span>
 <span id="KMSProviders"></span> <span id="AzureAccessToken"></span> <span id="kmsproviders"></span>
 
+#### credentialProviders
+
+The `credentialProviders` property may be specified on [ClientEncryptionOpts](#ClientEncryptionOpts) or
```
baileympearson, line 478: Do you think it would be helpful to throw errors if the user supplies a credentialProvider configuration but their kmsProviders are not configured for automatic credential fetching?

```typescript
new ClientEncryption(client, { 
  kmsProviders: { aws: { accessKeyId: .. } },
  credentialProviders: { aws: .. }
});

// MongoCryptError: credentialProviders can only be used with automatic KMS credential fetching.
```

baileympearson (changes requested):

source/client-side-encryption/client-side-encryption.md
durran: Fixed.

source/client-side-encryption/client-side-encryption.md
durran: Good idea. I've updated to indicate an error must be thrown in this situation.

source/client-side-encryption/client-side-encryption.md
```diff
@@ -472,6 +473,35 @@ See
 <span id="GCPKMSOptions"></span> <span id="AWSKMSOptions"></span> <span id="KMSProvider"></span>
 <span id="KMSProviders"></span> <span id="AzureAccessToken"></span> <span id="kmsproviders"></span>
 
+#### credentialProviders
```
baileympearson, line 476: Can we add some tests for FLE?  I can think of 3 tests that might be worth adding:

1. A CE / auto encryption configured with explicit KMS credentials and custom KMS providers throws an error.
2. Tests that demonstrate the explicit encryption / auto encryption fetch credentials using a custom credential provider.
3. A test that demonstrates that, given a client configured for auto AWS credential fetching with a custom credential provider, auto encryption / explicit encryption can be configured with a separate AWS credential provider. 

I'm not actually sure that 3. is a valuable feature (is there a use case for this?) but it should be supported with the current spec, so maybe we should add a test for it to make it explicit.

baileympearson (changes requested):

source/client-side-encryption/client-side-encryption.md
durran: Just a note I've just added 1 & 2 as 3 became difficult to set up with our existing tooling and would require a significant amount to additions in tools for an edge case. Perhaps we can revisit later if we see a need.

source/client-side-encryption/client-side-encryption.md
```diff
@@ -475,6 +476,35 @@ See
...
+
+The `credentialProviders` property may be specified on [ClientEncryptionOpts](#ClientEncryptionOpts) or
+[AutoEncryptionOpts](#AutoEncryptionOpts). Current support is for AWS only, but is designed to be able to accommodate
+additional providers in the future. If a custom credential provider is present, it MUST be used instead of the default
+flow for fetching automatic credentials and if the `kmsProviders` are not configured for automatic credential fetching
+an error MUST be thrown.
+
+```typescript
+interface CredentialProviders {
+   aws? AWSCredentialProvider
```
baileympearson, line 489:
```suggestion
   aws?: AWSCredentialProvider
```

source/client-side-encryption/tests/README.md
```diff
@@ -3683,3 +3683,43 @@ Run an aggregate operation on `db.csfle` with the following pipeline:
...
+Create a MongoClient named `setupClient`.
+
+Create a [ClientEncryption](../client-side-encryption.md#clientencryption) object with the following options:
+
+```typescript
+class ClientEncryptionOpts {
+   keyVaultClient: <setupClient>,
+   keyVaultNamespace: "keyvault.datakeys",
+   kmsProviders: { "aws": {} },
+   credentialProviders: { "aws": <object/function that returns valid credentials from the environment> }
```
baileympearson, line 3720: Is it sufficient to test that the data key is successfully created?  Credentials can already be fetched from the environment, so even if a driver didn't use the custom credential provider I'd expect this test to pass.

Is there a way we could make this test fail if the credential provider isn't used?  In Node, we can always spy on the provider but I don't think every language can do this.  Maybe we could manually implement spying by having the provider increment a call count, and then assert that callCount = 1.  Or we could use environment variables that don't match what the AWS sdk expects:
```
# these have the wrong names, aws sdk looks for ACCESS_KEY_ID and SECRET_ACCESS_KEY
process.env = {
  KEY: <key id>,
  SECRET: <secret>
}

const provider: AWSCredentialProvider = async () => {
  return {
    accessKeyId: process.env.KEY,
    secretAccessKey: process.env.SECRET
  }
}
```
(While this would work in Node, where we can re-write process.env as we wish, but I'm not sure this would be possible in other languages without a separate evergreen task that didn't load the AWS credentials into the environment.).

source/client-side-encryption/tests/README.md
```diff
@@ -3683,3 +3683,43 @@ Run an aggregate operation on `db.csfle` with the following pipeline:
 ```
 
 Expect an exception to be thrown with a message containing the substring `Upgrade`.
+
+### 26. Custom AWS Credentials
+
+These tests require valid AWS credentials. Refer:
+[Automatic AWS Credentials](../client-side-encryption.md#automatic-credentials).
+
+#### Case 1: Explicit encryption with credentials and custom credential provider
```
baileympearson, line 3692: Do you think we should test this case for auto encryption as well?  Most (all?) drivers have separate code paths for instantiating a CE object and instantiating a client configured for auto encryption.  This test only covers explicit encryption.

source/client-side-encryption/tests/README.md
baileympearson: I have the same concern for the regular auth tests too.

baileympearson (changes requested):

source/client-side-encryption/client-side-encryption.md
durran: Fixed

source/client-side-encryption/tests/README.md
durran: I've added a case 3 that checks the same for auto encryption.

source/client-side-encryption/tests/README.md
durran: This is a really good observation and it found some issues in the Node tests. I've reworded the client encryption tests to say they must not be run in an AWS environment and use the keys from the secrets manager instead for the remote AWS KMS provider. I've also reworded the auth tests to do what you've described here - we check the call count to the custom provider, which should be possible to all drivers.

source/client-side-encryption/tests/README.md
```diff
@@ -3683,3 +3683,60 @@ Run an aggregate operation on `db.csfle` with the following pipeline:
 ```
 
 Expect an exception to be thrown with a message containing the substring `Upgrade`.
+
+### 26. Custom AWS Credentials
+
+These tests require valid AWS credentials for the remote KMS provider via the secrets manager. These tests MUST NOT run
+inside an AWS environment that has the same credentials set in order to properly ensure the tests would fail using
+on-demand credentials.
```
baileympearson, line 3691: Suggest mentioning which keys we expect people to use:
```suggestion
These tests require valid AWS credentials for the remote KMS provider via the secrets manager (FLE_AWS_KEY and FLE_AWS_SECRET). These tests MUST NOT run
inside an AWS environment that has the same credentials set in order to properly ensure the tests would fail using
on-demand credentials.
```

source/client-side-encryption/tests/README.md
```diff
@@ -3683,3 +3683,60 @@ Run an aggregate operation on `db.csfle` with the following pipeline:
...
+```typescript
+class ClientEncryptionOpts {
+  keyVaultClient: <setupClient>,
+  keyVaultNamespace: "keyvault.datakeys",
+  kmsProviders: { "aws": {} },
+  credentialProviders: { "aws": <object/function that returns valid credentials from the secrets manager> }
+}
+```
+
+Use the client encryption to create a datakey using the "aws" KMS provider. This should successfully load and use the
```
baileympearson, line 3725: Suggest providing a concrete example in Node of what the provider looks like for this test, for clarity

baileympearson (approved):
LGTM, two minor clarifying comments.  I'm not an auth or FLE spec owner though.

source/client-side-encryption/tests/README.md
durran: Updated.

source/client-side-encryption/tests/README.md
durran: Added example in Node below this paragraph.

durran:
Pinging @jyemin as Java has an interest here. @kevinAlbs as I've added FLE options for the custom credential providers.

blink1073 (approved):
LGTM!

jyemin (approved):
I'm not sure how we will define the types in the Java driver, as applications are allowed to use either AWS SDK v1 or v2, and those dependencies are optional so the types can't be part of the API.  But we will figure out a way to make it work.

source/client-side-encryption/tests/README.md
```diff
@@ -3683,3 +3683,91 @@ Run an aggregate operation on `db.csfle` with the following pipeline:
...
 
 Expect an exception to be thrown with a message containing the substring `Upgrade`.
+
+### 26. Custom AWS Credentials
+
+These tests require valid AWS credentials for the remote KMS provider via the secrets manager (FLE_AWS_KEY and
+FLE_AWS_SECRET). These tests MUST NOT run inside an AWS environment that has the same credentials set in order to
+properly ensure the tests would fail using on-demand credentials.
+
+#### Case 1: Explicit encryption with credentials and custom credential provider
```
kevinAlbs, line 3693:
```suggestion
#### Case 1: ClientEncryption with `credentialProviders` and incorrect `kmsProviders`
```

Suggest avoiding "Explicit encryption" since the case only creates a data key (does not call `ClientEncryption.encrypt`). Suggest noting this case uses incorrect `kmsProviders.

source/client-side-encryption/tests/README.md
```diff
@@ -3683,3 +3683,91 @@ Run an aggregate operation on `db.csfle` with the following pipeline:
...
+  keyVaultClient: <setupClient>,
+  keyVaultNamespace: "keyvault.datakeys",
+  kmsProviders: { "aws": { "accessKeyId": <set from secrets manager>, "secretAccessKey": <set from secrets manager> } },
+  credentialProviders: { "aws": <default provider from AWS SDK> }
+}
+```
+
+Assert that an error is thrown.
+
+#### Case 2: Explicit encryption with custom credential provider
```
kevinAlbs, line 3710:
```suggestion
#### Case 2: ClientEncryption with `credentialProviders` works
```

source/client-side-encryption/tests/README.md
```diff
@@ -3683,3 +3683,91 @@ Run an aggregate operation on `db.csfle` with the following pipeline:
...
+    }
+  }
+};
+const clientEncryption = new ClientEncryption(keyVaultClient, options);
+const dk = await clientEncryption.createDataKey('aws', { masterKey });
+expect(dk).to.be.a(Binary);
+expect(calledCount).to.be.greaterThan(0);
+```
+
+#### Case 3: Auto encryption with credentials and custom credential provider
```
kevinAlbs, line 3759:
```suggestion
#### Case 3: AutoEncryptionOpts with `credentialProviders` and incorrect `kmsProviders`
```

source/auth/auth.md
```diff
@@ -1306,6 +1323,10 @@ in the MONGODB-OIDC specification, including sections or blocks that specificall
         check MUST be performed after SRV record resolution, if applicable. This property is only required for drivers
         that support the [Human Authentication Flow](#human-authentication-flow).
 
+    - AWS_CREDENTIAL_PROVIDER
+
+        A function or object from the AWS SDK that can be used to return AWS credentials.
```
kevinAlbs, line 1328: Consider similarly allowing the property to be defined on the `MongoClient` (as is done for `OIDC_HUMAN_CALLBACK`):


```suggestion
        A function or object from the AWS SDK that can be used to return AWS credentials. Drivers MAY allow the user to
        specify the callback using a `MongoClient` configuration instead of a mechanism property, depending on what is
        idiomatic for the driver.
```

I expect this may be preferred over specifying in the URI string.

source/auth/auth.md
```diff
@@ -987,12 +987,29 @@ those credentials will be used by default if AWS auth environment variables are
...
 
+##### Custom Credential Providers
+
+Drivers that choose to use the AWS SDK to fetch credentials MAY also allow users to provide a custom credential provider
+as an option to the `MongoClient`. The interface for the option provided depends on the individual language SDK and
+drivers MUST consult AWS SDK documentation to determine that format when implementing. The name of the option MUST be
+`AWS_CREDENTIAL_PROVIDER` and be part of the authentication mechanism properties options that can be provided to the
+client.
+
+Drivers that implement this MAY choose to implement the following scenarios when applicable in their language's SDK:
```
kevinAlbs, line 998:
```suggestion
Drivers MAY expose API for default providers for the following scenarios when applicable in their language's SDK:
```

Suggest clarifying the suggestions are API for default providers.

kevinAlbs (approved):
LGTM! Suggested wording changes, but nothing behavior changing.

PR #1749, opened 2025-01-29
DRIVERS-2869 Adjust getMore maxTimeMS Calculation for tailable awaitData Cursors

prestonvasquez:

- [x] Update changelog.
- [x] Test changes in at least one language driver. https://github.com/mongodb/mongo-go-driver/pull/1925
    clusters, and serverless).

### DIFF
--- a/source/client-side-operations-timeout/client-side-operations-timeout.md
+++ b/source/client-side-operations-timeout/client-side-operations-timeout.md
@@ -670,6 +670,7 @@ timeout for each database operation. This would mimic using `timeoutMode=ITERATI
 
 ## Changelog
 
+- 2024-01-29: Adjust getMore maxTimeMS Calculation for tailable awaitData Cursors
 - 2024-09-12: Specify that explain helpers support support timeoutMS.
 - 2023-12-07: Migrated from reStructuredText to Markdown.
 - 2022-11-17: Use minimum RTT for maxTimeMS calculation instead of 90th percentile RTT.
--- a/source/client-side-operations-timeout/client-side-operations-timeout.md
+++ b/source/client-side-operations-timeout/client-side-operations-timeout.md
@@ -334,10 +334,10 @@ resulting cursor but MUST NOT append a `maxTimeMS` field to any commands.
 ##### Tailable awaitData Cursors
 
 If `timeoutMS` is set, drivers MUST apply it to the original operation. Drivers MUST also apply the original `timeoutMS`
-value to each `next` call on the resulting cursor but MUST NOT use it to derive a `maxTimeMS` value for `getMore`
-commands. Helpers for operations that create tailable awaitData cursors MUST also support the `maxAwaitTimeMS` option.
-Drivers MUST error if this option is set, `timeoutMS` is set to a non-zero value, and `maxAwaitTimeMS` is greater than
-or equal to `timeoutMS`. If this option is set, drivers MUST use it as the `maxTimeMS` field on `getMore` commands.
+value to each `next` call on the resulting cursor. Helpers for operations that create tailable awaitData cursors MUST
+also support the `maxAwaitTimeMS` option. Drivers MUST error if this option is set, `timeoutMS` is set to a non-zero
+value, and `maxAwaitTimeMS` is greater than or equal to `timeoutMS`. If this option is set, drivers MUST use
+`min(maxAwaitTimeMS, remaining timeoutMS)` as the `maxTimeMS`` field on `getMore` commands.
 
 See [Tailable cursor behavior](#tailable-cursor-behavior) for rationale regarding both non-awaitData and awaitData
 cursors.
@@ -599,6 +599,11 @@ timeouts, while the `maxAwaitTimeMS` option is used as the `maxTimeMS` field for
 distinct meanings, so supporting both yields a more robust, albeit verbose, API. Drivers error if `maxAwaitTimeMS` is
 greater than or equal to `timeoutMS` because in that case, `getMore` requests would not succeed if the batch was empty:
 the server would wait for `maxAwaitTimeMS`, but the driver would close the socket after `timeoutMS`.
+
+For tailable awaitData cursors we use the `min(maxAwaitTimeMS, remaining timeoutMS - minRoundTripTime)` to allow the
+server more opportunities to respond with an empty batch before a client-side timeout. Additionally, this change is
+required to prevent an unnecessary client-side timeout during a pending read when checking out a connection. For
+example, maxAwaitTimeMS=1000 and remaining timeoutMS=100 will cause a pending read to hang for 900ms.
 
 ### Change stream behavior
 
--- a/source/client-side-operations-timeout/client-side-operations-timeout.md
+++ b/source/client-side-operations-timeout/client-side-operations-timeout.md
@@ -337,7 +337,7 @@ If `timeoutMS` is set, drivers MUST apply it to the original operation. Drivers 
 value to each `next` call on the resulting cursor. Helpers for operations that create tailable awaitData cursors MUST
 also support the `maxAwaitTimeMS` option. Drivers MUST error if this option is set, `timeoutMS` is set to a non-zero
 value, and `maxAwaitTimeMS` is greater than or equal to `timeoutMS`. If this option is set, drivers MUST use
-`min(maxAwaitTimeMS, remaining timeoutMS)` as the `maxTimeMS`` field on `getMore` commands.
+`min(maxAwaitTimeMS, remaining timeoutMS)` as the `maxTimeMS` field on `getMore` commands.
 
 See [Tailable cursor behavior](#tailable-cursor-behavior) for rationale regarding both non-awaitData and awaitData
 cursors.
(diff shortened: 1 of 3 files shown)


source/client-side-operations-timeout/tests/tailable-awaitData.json
```diff
@@ -417,6 +418,114 @@
...
+            "batchSize": 1,
+            "maxAwaitTimeMS": 100
+          },
+          "saveResultAsEntity": "tailableCursor"
+        },
+        {
+          "name": "iterateOnce",
+          "object": "tailableCursor",
+          "arguments": {
+            "timeoutMS": 50
```
vbabanin, line 440: In `timeoutMS is refreshed for getMore if maxAwaitTimeMS is set` test case, `timeoutMS` is set when the cursor is created. I suggest we follow the same approach here and the other test since not all drivers allow setting a timeout for each `next()` call as an argument - it gets refreshed under the hood.

For example, we could set timeoutMS to 150ms and maxAwaitTimeMS to 100ms like this:
```
{
          "name": "createFindCursor",
          "object": "collection",
          "arguments": {
            "filter": {},
            "cursorType": "tailableAwait",
            "batchSize": 1,
            "timeoutMS": 150
            "maxAwaitTimeMS": 100
          },
          "saveResultAsEntity": "tailableCursor"
        },
```

Also, we configure a failpoint to block the find command for `60ms` and call `iterateOnce()`. At this point, `maxTimeMS` on the command should be ≤ 90ms, which is ≤ `maxAwaitTimeMS`."

prestonvasquez: @vbabanin It's not idiomatic for the Go Driver to use the timeoutMS applied to constructing a cursor to each `next()` call. This would violate user expectations of [context.Context](https://pkg.go.dev/context#Context). For example, 

```go
findCtx, findCancel := context.WithTimeout(context.Background(), 1*time.Second)
defer findCancel()

cur, _ := coll.Find(findCtx, bson.D{}) // User is applying a 1s timeout

// User does not apply a timeoutMS to the context passed to Next, internally 
// applying a timeout here would violate expectations.
for cursor.Next(context.Background()) {
	// ...
}
```

Tests such as `timeoutMS is refreshed for getMore if maxAwaitTimeMS is set` are skipped in the Go Driver specifically because timeoutMS is not being applied at either the client or operation level. More than likely we will create a DRIVERS ticket to address this, see [GODRIVER-3480](https://jira.mongodb.org/browse/GODRIVER-3480) for more info.  IIUC, we are likely to suggest putting the same timeout on both constructor and next. Would this work for Java?

```yml
    operations:
      - name: createFindCursor
        object: *collection
        arguments:
          filter: {}
          cursorType: tailableAwait
          batchSize: 1
          maxAwaitTimeMS: 50
          timeoutMS: 100
        saveResultAsEntity: &tailableCursor tailableCursor
      # Iterate twice to force a getMore.
      - name: iterateOnce
        object: *tailableCursor
        arguments: 
          timeoutMS: 100
      - name: iterateOnce
        object: *tailableCursor
        arguments: 
          timeoutMS: 100
```

vbabanin: summary of discussion with @prestonvasquez:

Currently, the spec mandates that drivers reuse the same `timeoutMS` specified during cursor creation:

> Change stream: Drivers MUST also apply the original timeoutMS value to each next call on the change stream."
> Cursors: After the operation has executed, the original timeoutMS value MUST also be applied to each next call on the created cursor."

However, since Python and Go allow timeoutMS overrides per call, we should update the spec to clarify that overriding `timeoutMS` per next call is permitted, as it is idiomatic to their language. We should create a separate DRIVERS ticket to address that and update other unified tests for consistency to cover a use-case with override per `getMore`.

Some drivers might not currently expect `timeoutMS: 100` argument in their runners, that could lead to failed tests - this is the case for Java.

prestonvasquez: @vbabanin For now, I think we should update the unified spec tests to pass per the specifications (i.e. not applying a timeoutMS iteratively). Can you verify that the updated tests pass for Java?

vbabanin: I agree. I ran the tests and one of them `applying the remaining timeoutMS if it's less than maxAwaitTimeMS` fails with `state should be: maxAwaitTimeMS must be less than timeoutMS`. I’ve posted suggestion for handling this in a separate comment: https://github.com/mongodb/specifications/pull/1749#discussion_r1975975478

source/client-side-operations-timeout/tests/tailable-awaitData.json
```diff
@@ -417,6 +418,104 @@
...
+      "operations": [
+        {
+          "name": "createFindCursor",
+          "object": "collection",
+          "arguments": {
+            "filter": {},
+            "cursorType": "tailableAwait",
+            "batchSize": 1,
+            "maxAwaitTimeMS": 100,
+            "timeoutMS": 50
```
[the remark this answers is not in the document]
prestonvasquez: @vbabanin  Have you tested this solution? Isn't timeoutMS fully refreshed for each call to `Next`? It's unclear to me how blocking on the failpoint will result in a passing test.

```py
# Each next call starts a fresh timeout_ms.
def next(cursor):
	timeout_ms = cursor['timeout_ms']
	# ...

    	while get_more(remaining_timeout_ms):
        	pass # ...
```

Here is a gist with the updated JSON: https://gist.github.com/prestonvasquez/498ccc00c59a72b367de88272f0a777b

Does this pass in the Java driver?

vbabanin (approved):
LGTM!

ShaneHarvey:
@vbabanin can you confirm this test passes in .net on both linux and windows?

vbabanin:
@ShaneHarvey 

@ShaneHarvey with the changes introduced by this spec update and the added unified tests, all tests pass in Java across all topologies. However, I observed some flakiness over multiple runs in `apply remaining timeoutMS if less than maxAwaitTimeMS`, likely because the first `getMore` sometimes completes too quickly, causing `lte: 99 ms` to fail. In my earlier testing, this issue did not appear, but after running more iterations, I observed a failure couple of times.

I tested using a `failPoint` to delay the first `getMore`, which made the test more reliable.

I suggest adding a `failPoint` here to reduce raciness. Here’s a reference comment in the Java driver: https://github.com/mongodb/mongo-java-driver/pull/1650/files#r1994380696.

@prestonvasquez  what do you think?

source/client-side-operations-timeout/tests/tailable-awaitData.yml
```diff
@@ -245,3 +246,66 @@ tests:
...
+        object: *collection
+        arguments:
+          filter: { _id:  1 }
+          cursorType: tailableAwait
+          batchSize: 1
+          maxAwaitTimeMS: 100
+          timeoutMS: 200
+        saveResultAsEntity: &tailableCursor tailableCursor
+      - name: iterateOnce
+        object: *tailableCursor
```
baileympearson, line 262: Won't a tailable await data cursor return immediately when there are documents in the collection and batchSize = 1?  I'd expect that the find + first getMores return immediately (there are two documents in the collection), and then the second getMore's maxTimeMS may or may not be <= 99, simply depending on how much time the first find + getMore took.

ShaneHarvey (approved):

source/client-side-operations-timeout/tests/tailable-awaitData.yml
```diff
@@ -245,3 +246,77 @@ tests:
...
+                maxTimeMS: { $$lte: 100 }
+          - commandStartedEvent:
+              commandName: getMore 
+              databaseName: *databaseName
+              command:
+                maxTimeMS: { $$lte: 99 }
+
+  - description: "apply maxAwaitTimeMS if less than remaining timeout"
+    operations:
+      - name: failPoint
```
ShaneHarvey, line 287: According to @vbabanin this failpoint should be moved to the "apply remaining timeoutMS if less than maxAwaitTimeMS" test.

ShaneHarvey (changes requested):

ShaneHarvey (approved):
@vbabanin could you run the new tests and confirm the flakes are gone?

vbabanin:
> @vbabanin could you run the new tests and confirm the flakes are gone?

I re-ran the tests in this Java PR: https://github.com/mongodb/mongo-java-driver/pull/1650 that includes [tailable-awaitData.json](https://github.com/mongodb/specifications/pull/1749/files#diff-b217ebefb787a3b41b647944576741707b8dfa935bb7bc1155aad15af9b03681). All tests passed with no flakiness observed for these tests.

Here are the relevant details:
Evergreen patch changes: : https://evergreen.mongodb.com/filediff/67d351fec9bb0a0007cd5b53/?file_name=driver-core%2Fsrc%2Ftest%2Fresources%2Funified-test-format%2Fclient-side-operation-timeout%2Ftailable-awaitData.json&patch_number=0&commit_number=1
Successful patch execution: https://spruce.mongodb.com/version/67d351fec9bb0a0007cd5b53/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

PR #1750, opened 2025-02-03
DRIVERS-3097 Update test cases for BSON Binary Vector

qingyang-hu:
DRIVERS-3097

- Remove duplicate cases in "packed_bit.json"
- Add "canonical_bson" for appropriate invalid cases
- Add a case in "float32.json" containing corrupted "canonical_bson" with insufficient bytes for 4-byte float32

- [x] Update changelog.
    clusters, and serverless).


source/bson-binary-vector/tests/float32.json
```diff
@@ -44,8 +44,15 @@
       "vector": [127.0, 7.0],
       "dtype_hex": "0x27",
       "dtype_alias": "FLOAT32",
-      "padding": 3
+      "padding": 3,
+      "canonical_bson": "1C00000005766563746F72000A0000000927030000FE420000E04000"
+    },
+    {
+      "description": "Insufficient vector data FLOAT32",
```
vbabanin, line 51: It looks like this test introduces a use case where drivers should check for the absence of the array, as if a `null` reference is passed.

If that's the case, we should extend this test to cover INT8 and PackedBit, as well as other relevant fields which could be absent. Additionally, I think, explicitly defining `"vector": null` or `"dtype": null` would make the test case more explicit.

qingyang-hu: The "canonical_bson" in "Insufficient vector data FLOAT32" is a corrupted BSON with insufficient bytes for 4-byte float32, so it lacks a corresponding float32 vector. It makes sense to use `"vector": null` to make the case more explicit and consistent with other cases.

nbbeeken (changes requested):
Thanks for the test improvements! Is there a drivers ticket for this change?

source/bson-binary-vector/tests/float32.json
```diff
@@ -44,8 +44,15 @@
...
+      "valid": false,
+      "dtype_hex": "0x27",
+      "dtype_alias": "FLOAT32",
+      "canonical_bson": "1700000005766563746F7200050000000927002A2A2A00"
```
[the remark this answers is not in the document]
qingyang-hu: Will add a 5-byte case

source/bson-binary-vector/tests/README.md
```diff
@@ -29,7 +29,7 @@ Each JSON file contains three top-level keys.
 
 - `description`: string describing the test.
 - `valid`: boolean indicating if the vector, dtype, and padding should be considered a valid input.
-- `vector`: list of numbers
+- `vector`: (required if valid is true) list of numbers
```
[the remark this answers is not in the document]
vbabanin: When the `vector` field is absent in an invalid case, is `canonical_bson` now a required field?

qingyang-hu:
Created https://jira.mongodb.org/browse/DRIVERS-3097

source/bson-binary-vector/tests/README.md
```diff
@@ -50,7 +50,8 @@ MUST assert that the input float array is the same after encoding and decoding.
 
 #### To prove correct in an invalid case (`valid:false`), one MUST
 
-- raise an exception when attempting to encode a document from the numeric values, dtype, and padding.
+- when the vector field exists, raise an exception for encoding a document from the numeric values, dtype, and padding.
+- when the canonical_bson field exists, raise an exception for decoding it, as the field contains corrupted data that can't be decoded into a vector.
```
nbbeeken: It may be a breaking change for drivers to fail to read Vectors that have already been inserted without this requirement. 

How about drivers "fix" their serializers to stop making float32 vectors that have been created from the `canonical_bson` bytes?

source/bson-binary-vector/bson-binary-vector.md
```diff
@@ -242,3 +244,7 @@ See the [README](tests/README.md) for tests.
         you want to store or transmit binary data more efficiently by grouping 8 bits into a single byte (uint8). For an
         example in Python, see
         [numpy.unpackbits](https://numpy.org/doc/2.0/reference/generated/numpy.unpackbits.html#numpy.unpackbits).
+
+## Changelog
+
+- 2025-02-04: Update validation for decoding into a FLOAT32 vector.
```
caseyclements, line 250: Please add the date of the originally accepted spec:
- 2024-11-01: BSON Binary Subtype 9 accepted [DRIVERS-2926](https://jira.mongodb.org/browse/DRIVERS-2926) [(#1708)](https://github.com/mongodb/specifications/pull/1708[)

source/bson-binary-vector/tests/README.md
qingyang-hu: It should be fine to just "read" the Vector (as a BSON Binary). However, the specs need to clarify how to "decode/unserialize" to a sequence of float32 when the bytes for the vector elements are corrupted to store the sequence as the [description](https://github.com/mongodb/specifications/blob/master/source/bson-binary-vector/bson-binary-vector.md#data-structures) as well as expose the sequence to the driver users.

To prevent incompatibility issues, the exception doesn't have to be raised when reading the Binary of Vector from the server. It can be raised during the actual decoding/conversion (e.g. in the getter method of the float32 sequence).

nbbeeken: >  It can be raised during the actual decoding/conversion (e.g. in the getter method of the float32 sequence).

Perfect, and yeah that's how it works in node, our `toFloat32Array()` helper throws and `serialize`/`stringify` throw

qingyang-hu: Will clarify the specs with an update.

nbbeeken (approved):
LGTM

source/bson-binary-vector/tests/packed_bit.json
```diff
@@ -84,15 +78,6 @@
...
-      "description": "Vector with float values PACKED_BIT",
-      "valid": false,
-      "vector": [127.5],
-      "dtype_hex": "0x10",
-      "dtype_alias": "PACKED_BIT",
-      "padding": 0
     }
   ]
 }
-
```
caseyclements, line 98: And this one? Passing a float when packed bit requires an int (or unit) in [0,127]?

caseyclements (changes requested):

source/bson-binary-vector/tests/packed_bit.json
```diff
@@ -61,21 +62,14 @@
...
       "padding": 0
     },
-    {
-      "description": "Padding specified with no vector data PACKED_BIT",
-      "valid": false,
-      "vector": [],
-      "dtype_hex": "0x10",
-      "dtype_alias": "PACKED_BIT",
-      "padding": 1
-    },
```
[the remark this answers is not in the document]
qingyang-hu: Just a duplicate of L5-L12

source/bson-binary-vector/tests/packed_bit.json
qingyang-hu: Another duplicate of L57-L63

source/bson-binary-vector/bson-binary-vector.md
qingyang-hu: BTW, do we have the convention of appending the ticket numbers? IMO, they are trackable from the git change log.

source/bson-binary-vector/tests/packed_bit.json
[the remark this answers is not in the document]
caseyclements: Nice spot! Thank you!

source/bson-binary-vector/bson-binary-vector.md
caseyclements: The git change log contains all commits, not just those related to bson binary vectors. If we add just the PR, our convention is to include the jira ticket though.

caseyclements (approved):
LGTM

PR #1754, opened 2025-02-07
DRIVERS-1872 Add sessions spec tests to verify explicit sessions were created on the correct client

qingyang-hu:
DRIVERS-1872

- [ ] Update changelog.
    clusters, and serverless).

### DIFF
--- a/source/sessions/tests/driver-sessions-argument-for-correct-client.yml
+++ b/source/sessions/tests/driver-sessions-argument-for-correct-client.yml
@@ -1,0 +1,313 @@
+description: driver-sessions-argument-for-correct-client
+
+schemaVersion: "1.0"
+
+createEntities:
+  - client:
+      id: &client0 client0
+  - client:
+      id: &client1 client1
+  - database:
+      id: &database0 database0
+      client: *client0
+      databaseName: &database0Name session-tests
+  - collection:
+      id: &collection0 collection0
+      database: *database0
+      collectionName: &collection0Name test
+  - session:
+      id: &session1 session1
+      client: *client1
+
+tests:
+  - description: Session argument is for the right client (insertOne)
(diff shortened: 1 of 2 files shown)


source/sessions/tests/driver-sessions-argument-for-correct-client.yml
```diff
@@ -0,0 +1,313 @@
...
+  - collection:
+      id: &collection0 collection0
+      database: *database0
+      collectionName: &collection0Name test
+  - session:
+      id: &session1 session1
+      client: *client1
+
+tests:
+  - description: Session argument is for the right client (insertOne)
```
ShaneHarvey, line 23: It's expensive to setup and teardown each of these tests. Could we combine all of these tests into one to avoid that problem?

source/sessions/tests/driver-sessions-argument-for-correct-client.yml
```diff
@@ -0,0 +1,253 @@
...
+    runOnRequirements:
+      - minServerVersion: "8.0"
+    operations:
+      - name: clientBulkWrite
+        object: *client0
+        arguments:
+          models: []
+          session: *session1
+        expectError:
+          isClientError: true
```
sanych-sun, line 253: Where from we get the list of operations? Should we also test the following commands (I've reviewed C# Driver public API that accepts session as a parameter)?
- MongoClient.DropDatabase
- MongoClient.ListDatabaseNames
- MongoClient.Watch
- MongoDatabase.CreateCollection
- MongoDatabase.CreateView
- MongoDatabase.RenameCollection
- MongoDatabase.RunCommand
- MongoDatabase.Watch
- Collection.AggregateToCollection
- Collection.Watch

PR #1757, opened 2025-02-18
DRIVERS-3082 add prose tests for `$lookup`

kevinAlbs:
# Summary

Add prose test for `$lookup` support in QE/CSFLE.

Tests were run in the C driver in [this patch](https://spruce.mongodb.com/version/67b1feac78623900074b9ef8). Note: `latest` tasks were updated to use `latest-build` to use unreleased 8.1 builds (see [slack](https://mongodb.slack.com/archives/C72LB5RPV/p1739884710850709) for background).

Testing requires a unreleased version of libmongocrypt to include changes from [MONGOCRYPT-723](https://jira.mongodb.org/browse/MONGOCRYPT-723). Binaries are available on [this patch build](https://spruce.mongodb.com/task/libmongocrypt_publish_upload_all_patch_c3dff011c9c0dc326c31414471a4b613e0d91365_67af62d2eff4dc0007490c8f_25_02_14_15_35_47/files?execution=0) or by building from source at https://github.com/mongodb/libmongocrypt/commit/33fdf65cce5a0c0cdd293c64ed40e4a8205c3ce0.

The failed `Markdown Link Check` on the [GitHub action](https://github.com/mongodb/specifications/actions/runs/13551263630/job/37875175396?pr=1757) appears unrelated (and is failing on master).

# libmongocrypt protocol change

libmongocrypt's protocol previously assumed auto encrypting a command only required at most one schema. With `$lookup` there can be multiple schemas on an `aggregate` command. As a result, a small driver change is needed: feed all listCollections results to libmongocrypt in the `MONGOCRYPT_CTX_NEED_MONGO_COLLINFO(_WITH_DB)` states. The previous protocol only required returning the first. Call `mongocrypt_setopt_enable_multiple_collinfo` to indicate the new behavior is implemented (otherwise libmongocrypt returns an error if a command requires multiple schemas).

The libmongocrypt documentation includes a change to the protocol [in integrating.md](https://github.com/mongodb/libmongocrypt/commit/33fdf65cce5a0c0cdd293c64ed40e4a8205c3ce0#diff-831b7c30649e24ea657915e29deba4aec0b05937c9d96a76cf22e89dca32d042R143).

See the small protocol change implemented in the C driver [here](https://github.com/kevinAlbs/mongo-c-driver/commit/eb18fc82e822424a4a84c7754debd707ba3b1e9a).

- ~~[ ] Update changelog.~~ **Test changes only**
    clusters, and serverless).

### DIFF
--- a/source/client-side-encryption/tests/README.md
+++ b/source/client-side-encryption/tests/README.md
@@ -3659,9 +3659,9 @@ Run an aggregate operation on `db.csfle` with the following pipeline:
 
 Expect an exception to be thrown with a message containing the substring `not supported`.
 
-#### Case 9: test error with pre-8.1
+#### Case 9: test error with \<8.1
 
-This case requires mongocryptd/crypt_shared pre-8.1.
+This case requires mongocryptd/crypt_shared \<8.1.
 
 Recreate `encryptedClient` with the same `AutoEncryptionOpts` as the setup. (Recreating prevents schema caching from
 impacting the test).
--- a/source/client-side-encryption/tests/README.md
+++ b/source/client-side-encryption/tests/README.md
@@ -3416,7 +3416,8 @@ Repeat this test with the `azure` and `gcp` masterKeys.
 
 ### 25. Test $lookup
 
-Unless otherwise noted, these tests require: libmongocrypt 1.13.0, mongocryptd/crypt_shared 8.1+, and server 8.1+. Skip on standalone.
+Unless otherwise noted, these tests require: libmongocrypt 1.13.0, mongocryptd/crypt_shared 8.1+, and server 8.1+. Skip
+on standalone.
 
 The syntax `<filename.json>` is used to refer to the content of the corresponding file in `../etc/data/lookup`.
 
--- a/source/client-side-encryption/tests/README.md
+++ b/source/client-side-encryption/tests/README.md
@@ -3420,8 +3420,7 @@ This test requires libmongocrypt 1.13.0. Unless otherwise noted, tests require m
 
 Tests require server support of QE: Require MongoDB server 7.0+. Skip on standalone.
 
-The syntax `<filename.json>` is used to refer to files in
-[source/client-side-encryption/etc/data/lookup](../etc/data/lookup/).
+The syntax `<filename.json>` is used to refer to files in `../etc/data/lookup`.
 
 #### Setup
 
--- a/source/client-side-encryption/tests/README.md
+++ b/source/client-side-encryption/tests/README.md
@@ -3424,19 +3424,6 @@ The syntax `<filename.json>` is used to refer to the content of the correspondin
 
 #### Setup
 
-Create an unencrypted MongoClient named `unencryptedClient`. Drop database `db`.
-
-Insert `<key-doc.json>` into `db.keyvault` with majority write concern.
-
-Create the following collections:
-
-- `db.csfle` with options: `{ "validator": { "$jsonSchema": "<schema-csfle.json>"}}`.
-- `db.csfle2` with options: `{ "validator": { "$jsonSchema": "<schema-csfle2.json>"}}`.
-- `db.qe` with options: `{ "encryptedFields": "<schema-qe.json>"}`.
-- `db.qe2` with options: `{ "encryptedFields": "<schema-qe2.json>"}`.
-- `db.no_schema` with no options.
-- `db.no_schema2` with no options.
-
 Create an encrypted MongoClient named `encryptedClient` configured with:
 
 ```python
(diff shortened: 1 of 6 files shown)


nbbeeken (approved):
LGTM, passing in node, fairly straightforward implementation

katcharov (commented):
Some minor comments

source/client-side-encryption/tests/README.md
```diff
@@ -3413,3 +3413,259 @@ Repeat this test with the `azure` and `gcp` masterKeys.
...
+
+Tests require server support of QE: Require MongoDB server 7.0+. Skip on standalone.
+
+The syntax `<filename.json>` is used to refer to files in `../etc/data/lookup`.
+
+#### Setup
+
+Create an unencrypted MongoClient. Drop database `db`.
+
+Insert `<key-doc.json>` into `db.keyvault`.
```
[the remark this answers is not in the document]
kevinAlbs: > It is less error-prone to drop the whole db, but I have seen this cause permission errors on Atlas. Why does this test-set drop the db, vs the others which drop each individual item?

No reason. Updated to drop/create individual collections. QE collections also need to drop [state collections](https://github.com/mongodb/specifications/blob/366afd6b0972a4210184a7bd2d81c91f68ee01c2/source/client-side-encryption/client-side-encryption.md#drop-collection-helper). Using `encryptedClient` to drop collections drops the state collections.

> Let's name each client (`unencryptedClient`, `encryptedClient`, `client`)

I like that idea. Done.

> In some other test sets, collections are dropped and keys are inserted with majority. Why not here?

Good catch. Updated to insert key with majority write concern in setup.
Drivers [use majority read concern](https://github.com/mongodb/specifications/blob/366afd6b0972a4210184a7bd2d81c91f68ee01c2/source/client-side-encryption/client-side-encryption.md#key-vault-collection) to read key documents. I expect if majority write concern is not used for the setup, the test may fail to find the not-yet-majority-written key document.

source/client-side-encryption/tests/README.md
```diff
@@ -3413,3 +3413,259 @@ Repeat this test with the `azure` and `gcp` masterKeys.
 2. Call `client_encryption.createDataKey()` with "aws" as the provider. Expect this to fail.
 
 Repeat this test with the `azure` and `gcp` masterKeys.
+
+### 25. Test $lookup
+
+Unless otherwise noted, these tests require: libmongocrypt 1.13.0, mongocryptd/crypt_shared 8.1+, and server 8.1+.
+
+Tests require server support of QE: Require MongoDB server 7.0+. Skip on standalone.
```
katcharov, line 3421:
```suggestion
Unless otherwise noted, these tests require: libmongocrypt 1.13.0, mongocryptd/crypt_shared 8.1+, and server 8.1+. Skip on standalone.
```

katcharov (changes requested):

source/client-side-encryption/tests/README.md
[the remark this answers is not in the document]
katcharov: Thanks for the explanation, and looks good.

source/client-side-encryption/tests/README.md
```diff
@@ -3413,3 +3413,258 @@ Repeat this test with the `azure` and `gcp` masterKeys.
 2. Call `client_encryption.createDataKey()` with "aws" as the provider. Expect this to fail.
 
 Repeat this test with the `azure` and `gcp` masterKeys.
+
+### 25. Test $lookup
+
+Unless otherwise noted, these tests require: libmongocrypt 1.13.0, mongocryptd/crypt_shared 8.1+, and server 8.1+. Skip
+on standalone.
```
katcharov, line 3420: Sorry to leave yet another comment on this - bumping up/removing the 7.0 check causes [failures](https://spruce.mongodb.com/version/67be345226bcd40007b9b7fb/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) , such as [this one](https://evergreen.mongodb.com/test_log/mongo_java_driver_tests_jdk8_unsecure__version~4.0_os~linux_topology~replicaset_auth~noauth_ssl~nossl_jdk~jdk8_test_sync_patch_fb3f30b79aecfa5b17404c33a29fb1f5fb6f4ffb_67be345226bcd40007b9b7fb_25_02_25_21_21_26/0?test_name=79efb9ac7d173a68b39e05e56f4b1b45#L0) on 4.0. It looks like the setup section itself requires 7.0 (implies a global 7.0), then cases 1-8 require >=8.1, and then case 9 requires <8.1. I'm unsure how to best phrase, but maybe the following:

```suggestion
Unless otherwise noted, these tests require: libmongocrypt 1.13.0, and mongocryptd/crypt_shared 8.1+. Skip on standalone.

All tests require server 7.0+, which must be checked before setup. Unless otherwise noted, each test requires server 8.1+.
```

katcharov (approved):
LGTM, though note above comment.

All relevant tests are now passing in the Java implementation https://github.com/mongodb/mongo-java-driver/pull/1638

source/client-side-encryption/tests/README.md
kevinAlbs: Updated this sentence to only list global requirements. Listed additional requirements in the test cases.

PR #1760, opened 2025-02-25
DRIVERS-2915 Add ENVIRONMENT auth mechanism property to test URIs

isabelatkinson:
Tested in https://github.com/mongodb/mongo-rust-driver/pull/1315.

I didn't file a separate ticket for this as I don't expect other drivers will need to sync these tests. Added an "ENVIRONMENT" entry to a few tests that were failing in the Rust driver due to enforcement of the rule listed [here](https://github.com/mongodb/specifications/blob/master/source/auth/auth.md#mongocredential-properties-6) regarding the combination of "ENVIRONMENT" and "TOKEN_RESOURCE".

- [ ] Update changelog.
    clusters, and serverless).

### DIFF
--- a/source/connection-string/tests/valid-warnings.yml
+++ b/source/connection-string/tests/valid-warnings.yml
@@ -86,6 +86,3 @@ tests:
         auth: ~
         options:
             authMechanism: MONGODB-OIDC
-            authmechanismProperties:
-                TOKEN_RESOURCE: 'mongodb://test-cluster'
-                ENVIRONMENT: azure
--- a/source/connection-string/tests/valid-warnings.yml
+++ b/source/connection-string/tests/valid-warnings.yml
@@ -75,7 +75,7 @@ tests:
         options: ~
     -
         description: Comma in a key value pair causes a warning
-        uri: mongodb://localhost?authMechanism=MONGODB-OIDC&authMechanismProperties=TOKEN_RESOURCE:mongodb://host1%2Chost2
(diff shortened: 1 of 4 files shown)


source/connection-string/tests/valid-warnings.yml
```diff
@@ -86,3 +86,6 @@ tests:
         auth: ~
         options:
             authMechanism: MONGODB-OIDC
+            authmechanismProperties:
```
blink1073, line 89: This seems incorrect to me.  There should be no returned authMechanismProperties, since it fails to parse correctly.  Also, this TOKEN_RESOURCE doesn't match the one the uri.    Tested with https://spruce.mongodb.com/task/mongo_python_driver_test_rhel8_python3.10_test_4.4_standalone_noauth_nossl_sync_async_patch_eaae22c63ba6673460d18c7782b65716f9e18ea4_67bf16a8a0f2a800076ea1e0_25_02_26_13_27_22/logs?execution=0&sortBy=STATUS&sortDir=ASC

isabelatkinson: ah this was a bad copy/paste, removed. (we treat warnings in Rust as failures so the test runner wasn't doing any comparison here)

blink1073:
https://openid.net/specs/openid-connect-core-1_0.html was giving a 503, I kicked the build

blink1073:
...and it is still giving a 503.  I'm running the updated tests: https://spruce.mongodb.com/version/67bf37407f2f8d00082b8558/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

blink1073 (approved):
Thank you!

PR #1761, opened 2025-02-26
DRIVERS-3007: Rename WriteConcernFailed to WriteConcernTimeout

adelinowona:
Renames the WriteConcernError WriteConcernFailed to WriteConcernTimeout. WriteConcernFailed was mainly specified when setting up failpoints but we found out that there is no need to specify the codeName when setting up the failpoint with a writeConcernError, ideally we should be working the error codes and not the codeNames which are more likely to change than the error codes. 

- [x] Update changelog.
    clusters, and serverless).

### DIFF
--- a/source/read-write-concern/read-write-concern.md
+++ b/source/read-write-concern/read-write-concern.md
@@ -427,8 +427,8 @@ writeConcernError but may not be exhaustive. Note that some errors have been abb
 - `{ok:1, writeConcernError: {code: 11600, codeName: "InterruptedAtShutdown"}}`
 - `{ok:1, writeConcernError: {code: 11601, codeName: "Interrupted"}}`
 - `{ok:1, writeConcernError: {code: 11602, codeName: "InterruptedDueToReplStateChange"}}`
-- `{ok:1, writeConcernError: {code: 64, codeName: "WriteConcernFailed", errmsg: "waiting for replication timed out", errInfo: {wtimeout: True}}}`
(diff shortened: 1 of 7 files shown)


isabelatkinson (changes requested):
Rust patch: https://spruce.mongodb.com/version/67bf4349ed1ad10007bc3e99/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&variant=%5Erhel-8%24

I suggest changing "WriteConcernFailed" to "WriteConcernTimeout" in this note: https://github.com/mongodb/specifications/blob/master/source/read-write-concern/read-write-concern.md#user-content-fn-1-3740150915c0b53eaad0286809f994be

Looks like there's a formatting-related lint error as well - you should be able to fix this by running `pre-commit run --all-files`.

adelinowona:
I fixed the lint issue I had with the markdown file I modified. However, the lint check still fails but it is related to a broken link (https://openid.net/specs/openid-connect-core-1_0.html). I checked the site and it seems it is currently down. They might be performing maintenance or something but we could wait to see if it's resolved in the next few hours otherwise I'll just remove that link from our files.

isabelatkinson (approved):
lgtm!

> However, the lint check still fails but it is related to a broken link (https://openid.net/specs/openid-connect-core-1_0.html).

I saw the same in a PR I merged this morning - I'm fine with ignoring it for now and addressing the issue separately if it persists. The rest of the lint looks good.

BorisDog (approved):
LGTM!

PR #1769, opened 2025-03-13
DRIVERS-3134: increase timeouts in flaky CSOT runCommandCursor test

baileympearson:
The "Non-tailable cursor lifetime remaining timeoutMS applied to getMore if timeoutMode is unset" flakes fairly regularly in Node with the following error:

     AssertionError: Expected event count mismatch, expected [ 'commandStartedEvent', 'commandStartedEvent' ] but got [ 'CommandStartedEvent' ]
Turns out connection establishment is taking roughly 40ms worth of time. This causes that the `find` to time out, not the `getMore`. As a result, we still get a timeout error in the test but we have too few CommandStartedEvents and the test fails.

I modified the test locally to pre-populate the pool with a connection, and cannot reproduce the behavior anymore. I've also fixed the test by increasing the timeout to a value such that an unusually slow connection establishment doesn't cause the find to fail (timeoutMS = 200, blockTimeMS=120 seems to work fine for me locally).

- [ ] Update changelog.
    clusters, and serverless).


ShaneHarvey:
@baileympearson do you plan to pick this back up?

baileympearson:
@ShaneHarvey This PR fell off my radar (Daria reminded me about it).  We haven't seen this test flake for a while, although we do see other CSOT tests flake.  I think I'll just close this for now, and if it starts flaking again I can re-open it.

ShaneHarvey:
Thanks, should we also close https://jira.mongodb.org/browse/DRIVERS-3134?

PR #1777, opened 2025-03-27
Clarify Transaction prose test version requirements and collection creation

nbbeeken:

- [x] Update changelog.
    clusters, and serverless).

The changes don't require drivers to change their tests they just would've found these details out themselves when they went to implment the test.


baileympearson:
@dmitry-lukyanov - I added you as a reviewer because Durran is OOO.  This PR includes fixes a few issues we noticed with this prose test when we implemented it during spec fest.

comandeo-mongo (approved):

PR #1781, opened 2025-03-31
DRIVERS-3153 Fix incorrect batchSize assertion in ADL getMore test.

matthewdale:
[DRIVERS-3153](https://jira.mongodb.org/browse/DRIVERS-3153)

The ADL test case "A successful find event with getMore" in [getMore.json](https://github.com/mongodb/specifications/blob/f3549601e6bdfe4f18568985dfe706ca500dc679/source/atlas-data-lake-testing/tests/unified/getMore.json#L84) sets a "batchSize" of 3, then asserts that the `getMore` sends a "batchSize" of 1 (see [here](https://github.com/mongodb/specifications/blob/f3549601e6bdfe4f18568985dfe706ca500dc679/source/atlas-data-lake-testing/tests/unified/getMore.json#L84)). It seems like it's based on an [existing bug](https://github.com/mongodb/specifications/blob/68da5c5c9af9d87677ed90d30f8ad9ca3ea7e798/source/atlas-data-lake-testing/tests/getMore.yml#L32) that was previously ignored by the spec test runner, but then was persisted when the spec test was converted to a unified spec test.

- [x] Update changelog.
    clusters, and serverless).

### DIFF
--- a/source/atlas-data-lake-testing/tests/README.md
+++ b/source/atlas-data-lake-testing/tests/README.md
@@ -62,6 +62,8 @@ configuration and execute a ping command. Repeat this test using SCRAM-SHA-256.
 
 ## Changelog
 
+- 2025-03-31: Fix incorrect batchSize assertion in "A successful find event with getMore" test.
+
 - 2024-03-08: Convert legacy ADL tests to unified format. Convert test README from reStructuredText to Markdown.
 
 - 2022-10-05: Add spec front matter
--- a/source/atlas-data-lake-testing/tests/unified/getMore.json
+++ b/source/atlas-data-lake-testing/tests/unified/getMore.json
@@ -81,7 +81,7 @@
                   "collection": {
                     "$$type": "string"
                   },
-                  "batchSize": 1
+                  "batchSize": 3
                 },
                 "commandName": "getMore",
                 "databaseName": "cursors"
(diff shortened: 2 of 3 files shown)


dariakp:
@matthewdale Given DRIVERS-3144, should we pause any further work on ADL-related tickets?

dariakp:
Are we sure this is actually incorrect for DRIVERS that have the behavior to be removed in DRIVERS-1448? It's actually somewhat unclear to me why Java would set the 3 considering the existence of JAVA-3887 (and the current spec wording for limit and batch size in the CRUD spec [here](https://github.com/mongodb/specifications/blob/master/source/crud/crud.md#combining-limit-and-batch-size-for-op_query)) (credit to Bailey for digging all these up)

ShaneHarvey:
Python will fail this test change because we actually send batchSize=1 on the getMore (DRIVERS-1448).

matthewdale:
@dariakp @ShaneHarvey Considering [DRIVERS-3144](https://jira.mongodb.org/browse/DRIVERS-3144) will remove this test and there's no obvious way to resolve the conflict between drivers, we should abandon this PR. We can revisit the behavior with [DRIVERS-1448](https://jira.mongodb.org/browse/DRIVERS-1448).

PR #1782, opened 2025-04-02
DRIVERS-3155 Add topologyClosedEvent and topologyOpeningEvent

prestonvasquez:
DRIVERS-3155

DRIVERS-2711 added test to extend topologyClosedEvent and topologyOpeningEvent to observeEvents and SDAM expected events. This should be reflected in the unified spec tests and the subsequent tests should be updated to include the new schema version.

- [x] Update changelog.
- [x] Test changes in at least one language driver. (https://github.com/mongodb/mongo-go-driver/pull/2002)
    clusters, and serverless).

### DIFF
--- a/source/unified-test-format/schema-1.24.json
+++ b/source/unified-test-format/schema-1.24.json
@@ -1,0 +1,1171 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Unified Test Format",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "description",
+    "schemaVersion",
+    "tests"
+  ],
+  "properties": {
+    "description": {
+      "type": "string"
+    },
+    "schemaVersion": {
+      "$ref": "#/definitions/version"
+    },
+    "runOnRequirements": {
+      "type": "array",
+      "minItems": 1,
+      "items": {
+        "$ref": "#/definitions/runOnRequirement"
+      }
+    },
+    "createEntities": {
+      "type": "array",
+      "minItems": 1,
+      "items": {
+        "$ref": "#/definitions/entity"
+      }
+    },
+    "initialData": {
+      "type": "array",
+      "minItems": 1,
+      "items": {
+        "$ref": "#/definitions/collectionData"
+      }
+    },
+    "tests": {
+      "type": "array",
+      "minItems": 1,
+      "items": {
+        "$ref": "#/definitions/test"
+      }
+    },
+    "_yamlAnchors": {
+      "type": "object",
+      "additionalProperties": true
+    }
+  },
+  "definitions": {
+    "version": {
+      "type": "string",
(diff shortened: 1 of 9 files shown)


source/unified-test-format/schema-1.24.json
```diff
@@ -0,0 +1,1161 @@
+{
```
ShaneHarvey, line 1: I don't want to scope creep here but it's a bit of a pain to review UTF changes because with each version we create a entirely new file. What if we add a file called "schema-latest.json"? Then we could actually review the changes to schema-latest.json.

ShaneHarvey (approved):

source/unified-test-format/schema-1.24.json
prestonvasquez: @ShaneHarvey Good idea: https://github.com/mongodb/specifications/pull/1792

alcaeus:
In general I agree with this change. While wondering how this didn't trigger schema validation, I found multiple issues. For one, the SDAM tests weren't validated at all, and upon further investigation I found this affected more specs. I created #1794 to change the makefile and validate all test files.

Even if the SDAM tests were checked, the files would still be valid as we generally don't validate operations. This was to prevent each additional operation from generating a new schema version, which would've made the schema version go even higher than it already is. Had we created a client in a test-level `createEntities` directive, we could've caught this error.

As for the change itself, upon second thought I don't think this should be a new schema version. I'd rather see this as an errata for version 1.20, as drivers were generally expected to implement these events, and the only issue with them is if they are used in a test-level `createEntities` directive. This brings me to the last problem: some CMAP events, like `serverHeartbeatFailedEvent` first appear in schema version 1.17, but they haven't been added to that event name enum either. I would recommend adding them to the 1.17 (and all subsequent schema files) and add an errata to the changelog for this as well.

I have a rough idea for how to validate that event names are added wherever necessary and should have a pull request up for that shortly. That pull request would likely have to add these missing events already, and ideally I'd wait until #1794 is merged so all test files are properly validated. At the same time, I'll add validation or the `createEntities` operation since the arguments of that one are properly defined already.

prestonvasquez:
@alcaeus Treating this as a bug, updated existing schema to include closed and opening events.

source/unified-test-format/unified-test-format.md
```diff
@@ -3555,6 +3569,8 @@ other specs *and* collating spec changes developed in parallel or during the sam
 
 ## Changelog
 
+- 2025-04-07: Add `topologyOpeningEvent` and `topologyClosedEvent` to the unified test format.
```
alcaeus, line 3572: It may be worth calling out that this was added to the 1.20+ schema versions as it was omitted in error; that would provide context for people as to why this didn't introduce a new schema version.

alcaeus (approved):
One suggestion to call out the previous omission as a mistake that was now fixed, but changes LGTM.

alcaeus (approved):

PR #1784, opened 2025-04-08
DRIVERS-2295 Convert CSFLE spec tests to unified test format

mdb-ad:
Note: some legacy tests were skipped by the C driver. The unified version is also untested.
- badQueries (skipped in C driver due to due to test containing `$type` and C driver failing to parse: CDRIVER-3387).
- unsupportedCommand (skipped in C driver due to not having mapReduce helper)

[C driver Evergreen](https://spruce.mongodb.com/version/68472427ad20a0000771b8dd)

New tests depend on unified test runner changes from DRIVERS-3106.

### DIFF
--- a/source/client-side-encryption/limits/limits-encryptedFields.json
+++ b/source/client-side-encryption/limits/limits-encryptedFields.json
@@ -1,0 +1,14 @@
+{
+    "fields": [
+        {
+            "keyId": {
+                "$binary": {
+                    "base64": "LOCALAAAAAAAAAAAAAAAAA==",
+                    "subType": "04"
+                }
+            },
+            "path": "foo",
+            "bsonType": "string"
+        }
+    ]
+}
--- a/source/client-side-encryption/limits/limits-qe-doc.json
+++ b/source/client-side-encryption/limits/limits-qe-doc.json
@@ -1,0 +1,3 @@
+{
+    "foo": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+}
--- a/source/client-side-encryption/tests/README.md
+++ b/source/client-side-encryption/tests/README.md
@@ -563,10 +563,13 @@ First, perform the setup.
 2. Using `client`, drop and create the collection `db.coll` configured with the included JSON schema
     [limits/limits-schema.json](../limits/limits-schema.json).
 
-3. Using `client`, drop the collection `keyvault.datakeys`. Insert the document
+3. If using MongoDB 8.0+, use `client` to drop and create the collection `db.coll2` configured with the included
+    encryptedFields [limits/limits-encryptedFields.json](../limits/limits-encryptedFields.json).
+
+4. Using `client`, drop the collection `keyvault.datakeys`. Insert the document
     [limits/limits-key.json](../limits/limits-key.json)
 
-4. Create a MongoClient configured with auto encryption (referred to as `client_encrypted`)
+5. Create a MongoClient configured with auto encryption (referred to as `client_encrypted`)
 
     Configure with the `local` KMS provider as follows:
 
@@ -578,19 +581,19 @@ First, perform the setup.
 
 Using `client_encrypted` perform the following operations:
 
-1. Insert `{ "_id": "over_2mib_under_16mib", "unencrypted": <the string "a" repeated 2097152 times> }`.
+1. Insert `{ "_id": "over_2mib_under_16mib", "unencrypted": <the string "a" repeated 2097152 times> }` into `coll`.
 
     Expect this to succeed since this is still under the `maxBsonObjectSize` limit.
 
 2. Insert the document [limits/limits-doc.json](../limits/limits-doc.json) concatenated with
-    `{ "_id": "encryption_exceeds_2mib", "unencrypted": < the string "a" repeated (2097152 - 2000) times > }` Note:
-    limits-doc.json is a 1005 byte BSON document that encrypts to a ~10,000 byte document.
+    `{ "_id": "encryption_exceeds_2mib", "unencrypted": < the string "a" repeated (2097152 - 2000) times > }` into
+    `coll`. Note: limits-doc.json is a 1005 byte BSON document that encrypts to a ~10,000 byte document.
 
     Expect this to succeed since after encryption this still is below the normal maximum BSON document size. Note, before
     auto encryption this document is under the 2 MiB limit. After encryption it exceeds the 2 MiB limit, but does NOT
     exceed the 16 MiB limit.
 
-3. Bulk insert the following:
+3. Use MongoCollection.bulkWrite to insert the following into `coll`:
 
     - `{ "_id": "over_2mib_1", "unencrypted": <the string "a" repeated (2097152) times> }`
     - `{ "_id": "over_2mib_2", "unencrypted": <the string "a" repeated (2097152) times> }`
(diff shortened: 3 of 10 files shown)


kevinAlbs (commented):
Exciting. Initial comments.

source/client-side-encryption/tests/legacy/aggregate.json
kevinAlbs, line 1: Some drivers now use a git submodule to sync spec tests (example: [PHP](https://github.com/mongodb/mongo-php-library/tree/a4ccccd9011bad978e6f0d52efd1a512ef64d35a/tests)). As a result, when they update the submodule, all tests get synced. I request restoring the .json/.yml files (not templates) for the legacy tests. Once all drivers have synced the new tests, the old tests can be removed.

That may prevent a loss of test coverage (e.g. driver upgrades submodule to get non-CSFLE test updates, but does not-yet have the test runner changes implemented to run the new CSFLE tests)

ShaneHarvey: This seems like a big flaw in the git submodule approach.

jyemin: IMO it's not so much a flaw as a trade-off.  We get the benefit of auto-updates of specification changes but it comes with responsibility to treat specification tests as effectively read-only until everyone has migrated.  

A note about this has been added to the [README](https://github.com/mongodb/specifications?tab=readme-ov-file#automated-test-best-practices):

source/client-side-encryption/tests/unified/fle2v2-Update.yml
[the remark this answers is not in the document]
kevinAlbs: Resolving. Note added in https://github.com/mongodb/specifications/pull/1789.

source/unified-test-format/schema-1.23.json
```diff
@@ -427,6 +427,14 @@
...
           }
         },
+        "awsTemporary": {
+          "type": "object",
+          "properties": {}
+        },
+        "awsTemporaryNoSessionToken": {
+          "type": "object",
+          "properties": {}
+        },
```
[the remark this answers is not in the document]
kevinAlbs: Suggestion applied in https://github.com/mongodb/specifications/pull/1784/commits/101e50b1f7eb0a7da05af8746493d62822960208.

Revert the no-longer-needed changes in `schema-1.23.json`.

source/client-side-encryption/tests/unified/keyCache.yml
```diff
@@ -1,85 +1,105 @@
-description: keyCache-explicit
```
kevinAlbs: Revert changes to this file? This test was already in the unified format .

source/client-side-encryption/tests/unified/types.yml
```diff
@@ -17,12 +17,318 @@ createEntities:
       observeEvents:
         - commandStartedEvent
   - database:
-      id: "db"
+      id: "db0"
```
kevinAlbs, line 35: Revert this commit? I expect each test case can reuse the same client (each test is expected to drop the collection from `initialData` before running).

mdb-ad: It's not because of `initialData`. Every client in the types test has a different `schemaMap`.

source/client-side-encryption/tests/legacy/fle2v2-BypassQueryAnalysis.json
kevinAlbs, line 1: Suggest running `git merge master` to update PR diff. This file was restored in https://github.com/mongodb/specifications/pull/1788.

source/client-side-encryption/tests/unified/badQueries.yml
```diff
@@ -0,0 +1,548 @@
...
+          filter: {unencrypted: {$size: 2}}
+        object: "coll"
+        expectResult: []
+      - name: find
+        arguments:
+          filter: {encrypted_string: {$size: 2}}
+        object: "coll"
+        expectError:
+          errorContains: "Invalid match expression operator on encrypted field"
+      - name: find
```
kevinAlbs, line 355: The `badQueries` test can be run in the C driver by removing these two tests containing `$type`.

I would rather remove the `$type` tests to enable the C driver (and C++ and PHP driver) to run this test. This test does not seem to provide any significantly different coverage from the test above with `$size`.

source/client-side-encryption/tests/unified/badQueries.yml
```diff
@@ -0,0 +1,548 @@
...
+          # After it is:
+          # "Cannot encrypt element of type: array"
+          # Only check for the common prefix.
+          errorContains: "Cannot encrypt element of type"
+  - description: "an insert with encrypted field on _id errors"
+    operations:
+      - name: insertOne
+        arguments:
+          document: {_id: 1}
+        object: "coll"
```
kevinAlbs:
```suggestion
        object: "coll_with_encrypted_id"
```

The legacy form of this test set a schema with an encrypted `_id` field to get the expected error. Suggest creating a new collection entity:

```yml
  - collection:
      id: coll_with_encrypted_id
      database: db
      collectionName: coll_with_encrypted_id
```

And create the collection with a schema with an encrypted `_id` in `initialData`:
```yml
  - databaseName: default
    collectionName: coll_with_encrypted_id
    documents: []
    createOptions:
      validator:
        $jsonSchema: {'properties': {'_id': {'encrypt': {'keyId': [{'$binary': {'base64': 'AAAAAAAAAAAAAAAAAAAAAA==', 'subType': '04'}}], 'bsonType': 'string', 'algorithm': 'AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic'}}}}
```

source/client-side-encryption/tests/unified/types.yml
```diff
@@ -0,0 +1,786 @@
+description: types
+schemaVersion: '1.23'
+runOnRequirements:
+  - minServerVersion: "4.1.10"
+createEntities:
```
kevinAlbs, line 8:
```suggestion
# Create separate client entities for each tested type to isolate encryption schemas. mongocryptd/crypt_shared returns an error on schemas for unsupported types (e.g. double).
createEntities:
```

Suggest documenting why separate clients are needed.

source/client-side-encryption/tests/unified/aggregate.yml
```diff
@@ -0,0 +1,147 @@
...
+      collectionName: default
+initialData:
+  - databaseName: default
+    collectionName: default
+    documents:
+      - &doc0_encrypted {_id: 1, encrypted_string: {'$binary': {'base64': 'AQAAAAAAAAAAAAAAAAAAAAACwj+3zkv2VM+aTfk60RqhXq6a/77WlLwu/BxXFkL7EppGsju/m8f0x5kBDD3EZTtGALGXlym5jnpZAoSIkswHoA==',
+            'subType': '06'}}}
+    createOptions:
+      validator:
+        $jsonSchema: {'properties': {'encrypted_w_altname': {'encrypt': {'keyId': '/altname',
```
kevinAlbs: Consider running the YAML files through a formatter. Running through https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml seems to work OK (at the cost of a longer file).

source/client-side-encryption/tests/unified/bulk.yml
kevinAlbs, line 1: Restore bulk.yml (CSFLE with `MongoCollection.bulkWrite`). Remove client-bulkWrite-qe.yml (QE with `MongoClient.bulkWrite` will be added as part of DRIVERS-2888.)

rozza (changes requested):
Looks good one request:

Could `timeoutMS.yml` tests increase the `timeoutMS` and `blockTimeMS` by a factor of 10 as these timeout on evergreen. 

It appears the java legacy versions were skipped due to this.

Other than that - LGTM

baileympearson (approved):
I reviewed the libmongocrypt version requirement specifically.  All old FLE continue to pass on older versions of our bindings and the new tests are skipped, meaning these won't break Node's CI if we test an old bindings version against the latest driver.

rozza (approved):
LGTM!

PR #1785, opened 2025-04-09
DRIVERS-3184 assert unset `partialResult` in CRUD prose tests

kevinAlbs:
# Summary

Assert unset `BulkWriteException.partialResult` in CRUD prose tests.

# Background & Motivation

[CDRIVER-5969](https://jira.mongodb.org/browse/CDRIVER-5969) identified an implementation bug. Empty (rather than unset) results were sometimes returned on client errors.

The assert is optional. These cases do not require throwing a `BulkWriteException`. Quoting [spec](https://github.com/mongodb/specifications/blob/master/source/crud/bulk-write.md#top-level-errors):

> When a top-level error is encountered and individual results and/or errors have already been observed, drivers MUST embed the top-level error within a `BulkWriteException` as the error field to retain this information. Otherwise, drivers MAY throw an exception containing only the top-level error.

If acceptable, will file a DRIVERS ticket for the test tweak (will likely be a `spec-fest` change).

---

- ~~[ ] Update changelog.~~ (Tests only)
- [x] Test changes in at least one language driver. (Tested in [C driver](https://spruce.mongodb.com/version/67f6b1155653750007a15357))
    clusters, and serverless).~~ (Tested on all but serverless. C does not test serverless).


source/crud/tests/README.md
```diff
@@ -583,7 +583,7 @@ InsertOne {
 ```
 
 Execute `bulkWrite` on `client` with `largeDocumentModel`. Assert that an error (referred to as `error`) is returned.
-Assert that `error` is a client error.
+Assert that `error` is a client error. Assert the returned `BulkWriteException.partialResult` is unset.
```
isabelatkinson, line 586: The Rust driver returns an [`InvalidArgument`](https://docs.rs/mongodb/latest/mongodb/error/enum.ErrorKind.html#variant.InvalidArgument) error rather than a `BulkWriteError` for these cases in line with the language from [this section](https://github.com/mongodb/specifications/blob/master/source/crud/bulk-write.md#top-level-errors) of the spec:

> When a top-level error is encountered and individual results and/or errors have already been observed, drivers MUST embed the top-level error within a `BulkWriteException` as the error field to retain this information. Otherwise, drivers MAY throw an exception containing only the top-level error.

i.e., the invalid argument error is the "top-level" error, and no other errors or results have been observed, so we don't construct a wrapper `BulkWriteError`.

I'm still in favor of adding this assertion, but I think we should relax the language to be something like: "If a `BulkWriteException` was thrown, assert that `partialResult` is unset." I'll likely add assertions to the Rust tests that the error kind is indeed `InvalidArgument`.

Ditto for the other additions in this PR.

isabelatkinson (commented):
Rust PR: https://github.com/mongodb/mongo-rust-driver/pull/1354

source/crud/tests/README.md
kevinAlbs: Great catch. Updated.

isabelatkinson (approved):
lgtm!

GromNaN (approved):

PR #1790, opened 2025-04-28
Fix mdformat check

blink1073:

- [ ] Update changelog.
    clusters, and serverless).


blink1073:
We had to bump the versions of the formatters used for upstream compatibility, which caused some docs to be reformatted slightly.

comandeo-mongo (approved):

PR #1792, opened 2025-04-28
chore: add schema-latest pattern

prestonvasquez:
PR #1782 suggests adding a `schema-latest.json` file that will allow reviewers to grok changes to schema updates. Implementers would still be required to add a schema-1.n.json file when updating a schema but will now also be required to run `make update-schema-latest` to sync `schema-latest.json`. The diff should be verified in the GitHub workflow. 

- [ ] Update changelog.
    clusters, and serverless).


alcaeus (approved):
Great idea - not being able to directly see schema changes when a new version is added has bothered me for a while. Thank you for adding this!

dariakp (approved):

PR #1793, opened 2025-04-29
DRIVERS-3023 rename test operation to avoid clash with existing op

abr-egn:
- [✅] Update changelog.
- [✅] Test changes in at least one language driver. (https://github.com/mongodb/mongo-rust-driver/pull/1366)
- [✅] Test these changes against all server versions and topologies (including standalone, replica set, sharded
    clusters, and serverless).

This renames the gridfs rename operation from "rename" to "renameById" (consistent with the existing "renameByName") to avoid conflicting with the existing change streams "rename" test operation.  AFAICT the only driver that picked up the tests with the original name is PHP so churn should be minimal.

### DIFF
--- a/source/gridfs/tests/rename.json
+++ b/source/gridfs/tests/rename.json
@@ -93,7 +93,7 @@
       "description": "rename by id",
       "operations": [
         {
-          "name": "rename",
+          "name": "renameById",
           "object": "bucket0",
           "arguments": {
             "id": {
@@ -161,7 +161,7 @@
       "description": "rename when file id does not exist",
       "operations": [
         {
-          "name": "rename",
+          "name": "renameById",
           "object": "bucket0",
           "arguments": {
             "id": {
--- a/source/gridfs/tests/rename.yml
+++ b/source/gridfs/tests/rename.yml
@@ -51,7 +51,7 @@ initialData:
 tests:
   - description: "rename by id"
     operations:
-      - name: rename
+      - name: renameById
         object: *bucket0
         arguments:
           id: { "$oid": "000000000000000000000001" }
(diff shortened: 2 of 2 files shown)


GromNaN (changes requested):
> AFAICT the only driver that picked up the tests with the original name is PHP so churn should be minimal

We implement the method `GridFS::rename()` [since 2017](https://github.com/mongodb/mongo-php-library/blob/ce591998fccd7a1b64d4354542f33b14beb0c089/src/GridFS/Bucket.php#L431); as it [is specified](https://github.com/mongodb/specifications/blob/master/source/gridfs/gridfs-spec.md#renaming-stored-files).
You can't change this test without changing the spec.

abr-egn:
Sorry, I misspoke; I did mean the `collection` entity `rename` operation (currently used in change streams tests, hence the confusion).

To be honest I didn't anticipate that keeping the name of the method as defined in the API spec and the test operation the same would be as important as it clearly is - I saw keeping the API spec name as `rename` and updating the test operation name to `renameById` to be a quick and easy way to _save_ work across drivers and since that's no longer at all clear there's no need for this change.

For context, the Rust unified test runner parses the test files ahead of time into a vector of strongly-typed test operations, including argument structs for each operation.  The addition of the test files in DRIVERS-3023 introduced an ambiguity with `rename` where the shape of the expected arguments is not statically determined by the name of the operation (in contrast with `aggregate`, for example, which always has the same argument shape no matter which entity type it's targeting).  I saw that and had the impulse that other drivers might do the same ahead-of-time parsing as Rust, and so it might be kind to save aggregate work by renaming a few strings rather than requiring extra test runner logic across multiple drivers.

Anyway, this isn't a blocker for Rust; I've implemented indirection to deal with the ambiguity in https://github.com/mongodb/mongo-rust-driver/pull/1366 and I expect if any other drivers run into this they can do similar.

PR #1797, opened 2025-05-06
fix: markdown link check

durran:
Updates markdown link check to 3.13.7 to fix dead link failures.

- [ ] Update changelog.
    clusters, and serverless).


baileympearson (approved):
There isn't anything obvious in the markdown-link-check changelog that would explain why this consistently fails on v3.12.2 and passes on v3.13.7.  But it does - Durran and I have run the lint task a bunch of times each and the behavior is consistent: it passes on 3.13.7 and fails on 3.12.2.

PR #1798, opened 2025-05-07
DRIVERS-2985: Allow on-demand client metadata updates after MongoClient initialization.

vbabanin:
## Description
MongoDB’s handshake specification allows libraries and frameworks that wrap MongoDB drivers (e.g., ODMs, integrations) to append their own metadata.

However, this mechanism assumes the wrapping layer creates the `MongoClient`. When a wrapping environment instead receives a pre-created `MongoClient` (e.g., via dependency injection), there was no way to append additional metadata.

## Changes
This PR updates the handshake specification with the following additions:

-  Specifies that drivers MUST provide an API to allow updating client metadata (name, version, platform) after initialization.
- Appended values must be joined to existing `client.driver` fields using a `|` delimiter.
- Only new connections will include the updated metadata; existing connections MUST remain unaffected, as metadata propagation depends on the creation of new connections over time.
- Adds suggested prose tests to validate the above behavior. As some drivers may not support a testing backdoor to intercept handshake command events, or may face technical limitations, the exact test implementation is left to the discretion of the driver team.

- [x] Update changelog.
_Java Driver implementation:_ https://github.com/mongodb/mongo-java-driver/pull/1708
_Secondary Python Driver implementation:_ https://github.com/mongodb/mongo-python-driver/pull/2358
    clusters, and serverless).  _Evergreen path of aforementioned Java driver implementation:_ [Link](https://spruce.mongodb.com/version/681e4e305108bb0007b9d579/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)


source/mongodb-handshake/handshake.md
```diff
@@ -405,15 +405,25 @@ class DriverInfoOptions {
...
+
+If `DriverInfoOptions` are provided during or after MongoClient initialization, these options MUST NOT replace any
+existing metadata values, including driver-generated metadata and previously provided options. The provided options MUST
+be appended to their respective fields, and be delimited by a `|` character. For example, when
 [Motor](https://www.mongodb.com/docs/drivers/motor/) wraps PyMongo, the following fields are updated to include Motor's
 "driver info":
 
 ```typescript
-{
+{
```
sleepyStick, line 426: is this white space change intentional?

source/mongodb-handshake/handshake.md
```diff
@@ -405,15 +405,25 @@ class DriverInfoOptions {
...
 ```
 
-Note that how these options are provided to a driver is left up to the implementer.
+Note that how these options are provided to a driver during `MongoClient` initialization is left up to the implementer.
 
-If provided, these options MUST NOT replace the values used for metadata generation. The provided options MUST be
-appended to their respective fields, and be delimited by a `|` character. For example, when
+### Metadata updates after MongoClient initialization
+
+Drivers MUST provide an API that allows to append `DriverInfoOptions` to a MongoClient instance after initialization.
```
sleepyStick, line 412: "...that allows to append..." reads weirdly to me? not sure if its just me or not though. Its possible my brain just isn't parsing it correctly? I think my brain wants it to be "...that allows <a noun> to append..."

sleepyStick (commented):
Just a couple minor comments, everything else looks good! great work!

source/mongodb-handshake/handshake.md
vbabanin: Yeah, you're right -` that allows to append` is not correct. Originally, it was supposed to be `that allows wrapping libraries to append`. However,  I changed it to `that allows appending` to keep it more general and avoid specifying "users" or "wrapping libraries". Let me know what you think!

source/mongodb-handshake/handshake.md
vbabanin: Removed in https://github.com/mongodb/specifications/pull/1798/commits/5ce8b92fceb7363c8ed10f142e857c474fef6dd9, thanks!

source/mongodb-handshake/handshake.md
```diff
@@ -405,10 +405,21 @@ class DriverInfoOptions {
 }
 ```
 
-Note that how these options are provided to a driver is left up to the implementer.
+Note that how these options are provided to a driver during `MongoClient` initialization is left up to the implementer.
```
jyemin, line 408: Is this any less true for the new API?  I'm not sure it is since we don't specify a method signature below. If you agree that it holds true for the new API, then this change could be reverted.

source/mongodb-handshake/handshake.md
```diff
@@ -405,10 +405,21 @@ class DriverInfoOptions {
...
 
-If provided, these options MUST NOT replace the values used for metadata generation. The provided options MUST be
-appended to their respective fields, and be delimited by a `|` character. For example, when
+### Metadata updates after MongoClient initialization
+
+Drivers MUST provide an API that allows appending `DriverInfoOptions` to a MongoClient instance after initialization.
+
+After client metadata update, drivers MUST apply updated metadata to newly created connections. Drivers MUST NOT apply
+updated metadata to already established connections, nor create new connections solely for the purpose of transferring
```
jyemin, line 415:
```suggestion
updated metadata to already established connections, create new connections, or close existing connections solely for the purpose of transferring
```

source/mongodb-handshake/tests/README.md
```diff
@@ -82,3 +82,128 @@ the following sets of environment variables:
 2. Create and connect a `Connection` object that connects to the server that returns the mocked response.
 
 3. Assert that no error is raised.
+
+## Client Metadata Update Prose Tests
+
+The driver **MAY** implement the following tests. Drivers that do not emit events for commands issued as part of the
```
jyemin, line 88: Why is this a MAY?

jyemin (changes requested):

source/mongodb-handshake/handshake.md
vbabanin: Agreed, this still applies to the new API interface. I’ve reverted the change.

source/mongodb-handshake/tests/README.md
vbabanin: The prose tests assume interaction via the public API. However, when the public API is just a thin wrapper around internal components, and those components may already expose functionality to intercept commands (i.e., without requiring a backdoor), testing against internal layers may cover most functionality, assuming the public API’s interaction with the internals remains stable over time. The “MAY” reflects that the implementer can choose the trade-off between the complexity of introducing a backdoor for public API testing vs internal component testing. I’ve removed “MAY” to avoid over-assuming.

source/mongodb-handshake/handshake.md
sleepyStick: Yeah, this is much better. (upon re-reading, realized my suggestion at the end makes no sense because i wrote the same thing HAHA sorry for the confusion) thanks!

vbabanin:
Lint errors are unrelated to this PR, see: https://github.com/mongodb/specifications/pull/1808.

sleepyStick (approved):

source/mongodb-handshake/handshake.md
```diff
@@ -407,8 +407,19 @@ class DriverInfoOptions {
 
 Note that how these options are provided to a driver is left up to the implementer.
 
-If provided, these options MUST NOT replace the values used for metadata generation. The provided options MUST be
-appended to their respective fields, and be delimited by a `|` character. For example, when
+### Metadata updates after MongoClient initialization
+
+Drivers MUST provide an API that allows appending `DriverInfoOptions` to a `MongoClient` instance after initialization.
```
ShaneHarvey: We usually give an example API (or at least function signature) that implementors can reference. Could we add that here?

For example, see how startSession is documented: https://github.com/mongodb/specifications/blob/master/source/sessions/driver-sessions.md#mongoclient-changes

ShaneHarvey (changes requested):

source/mongodb-handshake/handshake.md
vbabanin: Thank you for the reference! I have added explicit API guidance for clarity. 

Previously, the specification left the mechanism for providing metadata during initialization up to the implementer. 

> (“Note that how these options are provided to a driver is left up to the implementer.”). 

I have refined the language to clarify that this flexibility applies only during initialization. For subsequent metadata updates, the API guidance is now clearly defined, requiring the update method to reside on `MongoClient`.

Updated in: https://github.com/mongodb/specifications/pull/1798/commits/4be80aa3686b8c9471ce60fa42044ae807b43725

source/mongodb-handshake/handshake.md
vbabanin: @jyemin I reverted the change back to:

> Note that how these options are provided to a driver during MongoClient initialization is left up to the implementer.

as we now provide explicit API guidance (method signature and location). Comment for context: https://github.com/mongodb/specifications/pull/1798#discussion_r2132728103

sleepyStick (approved):
Great work!

source/mongodb-handshake/tests/README.md
```diff
@@ -82,3 +82,160 @@ the following sets of environment variables:
...
+    | Field    | Value            |
+    | -------- | ---------------- |
+    | name     | library          |
+    | version  | 1.2              |
+    | platform | Library Platform |
+
+2. Send a `ping` command to the server and verify:
+
+    - The command succeeds.
+    - The wrapping library metadata is appended to the respective `client.driver` fields of the `hello` command.
```
ShaneHarvey: Could we simplify this test by removing the "library metadata" in step 1 and removing this  assert from step 2?

source/mongodb-handshake/tests/README.md
ShaneHarvey: If I understand correctly, we previously did not have any spec tests to verify a driver actually sends DriverInfo because (I assume) it was deemed not worth the effort. 

I'm fine with making these required for now. We could always change it back to "MAY" or "SHOULD" if we discover it's still not worth the effort in some languages.

source/mongodb-handshake/tests/README.md
vbabanin: Good catch - I missed removing the assertions in test 3. Removed assertions and library metadata in https://github.com/mongodb/specifications/pull/1798/commits/f0315f4b48fd849209639956698b3f99168340a6, thanks!

ShaneHarvey: Actually, couldn't this particular test be a unified spec test? It doesn't rely on any mocking.

vbabanin: Yes, good point - it makes sense to convert this to a unified test. I’ve updated it accordingly.

I also added an assertion to verify no new connection is created. Since unified tests don’t assume a test backdoor, we can’t directly assert on handshake commands. However, verifying that no new connection is established indirectly ensures no hello was sent during the handshake. Any subsequent hello's over an existing connection would appear as regular commands and cause the test to fail.

source/unified-test-format/unified-test-format.md
```diff
@@ -1491,6 +1491,19 @@ driver behavior when an operation is attempted on a closed client or one of its
...
+Appends client metadata to the client.
+
+The following arguments are supported:
+
+- `driverInfoOptions`: Required object with the following fields:
+    - `name`: Required string. The name of the wrapping library or framework.
+    - `version`: Optional string. The version of the wrapping library or framework.
+    - `platform`: Optional string. The platform of the wrapping library or framework.
+
+See [handshake](../mongodb-handshake/handshake.md#metadata-updates-after-mongoclient-initialization).
```
alcaeus:
```suggestion
See the [handshake spec](../mongodb-handshake/handshake.md#metadata-updates-after-mongoclient-initialization) for more details.
```

alcaeus (approved):
Unified test format changes look good with a nitpick.

source/unified-test-format/unified-test-format.md
vbabanin: Commited, thank you!

ShaneHarvey (approved):
LGTM

ShaneHarvey:
@sleepyStick could you verify the new unified tests in Python?

sleepyStick (approved):

PR #1804, opened 2025-05-26
DRIVERS-1815 Clarify PoolClearedError labels

comandeo-mongo:

- [ ] Update changelog.
    clusters, and serverless).


dariakp:
Just a note - once this merges, we'll want to either open a separate DRIVERS ticket, reopen the already completed individual language tickets, or file new individual language tickets in drivers that completed this work to ensure those drivers implement the updated prose tests.

PR #1809, opened 2025-06-11
DRIVERS-3114 Remove unified spec test behaviors only used by Astrolabe.

matthewdale:
[DRIVERS-3114](https://jira.mongodb.org/browse/DRIVERS-3114)

We're sunsetting the Drivers Atlas Testing (Astrolabe) project. Some unified spec test behaviors are only used by Astrolabe and can be removed:
* `storeEventsAsEntities` option for client entities
* `loop` operation

Changes:
* Remove `storeEventsAsEntities` option for client entities and `loop` operation from the Unified Test Format spec.
* Add unified spec test schema 1.24 with the above values removed.
* Update `schema-latest.json`.
* Remove all Unified Test Format spec tests that require the removed values.

- [x] Update changelog.
    clusters).


alcaeus (commented):
I believe this is the first time we're trying to remove functionality from the unified test format (I previously only deprecated the `serverless` option in `runOnRequirements`). While we never specified whether the test format follows something like semantic versioning, so far the assumption has held true that a test file that validates in schema version 1.x will validate in any schema version 1.y where `y > x`. This change would break that previous assumption. On the other hand, IIRC there are no tests other than astrolabe test files that use either `loop` or `storeEventsAsEntities` functionality, so removing the functionality from the schema would allow drivers to remove the logic from their test runners without fear of other tests re-introducing them.

baileympearson:
@alcaeus My preference would be to remove any unused code because:

1. less code to maintain = happier me
2. less confusion for new engineers, who need the full context about why these were added in the first place and then removed but left in the spec

Could your concerns be resolved by bumping to schema version 2.0?  Seems like the right use of a major version 🙂

matthewdale:
Another option is to remove the requirement to implement those behaviors from the Unified Test Format spec, but leave the properties in the JSON schema (with a note about why they're not in the spec). That would maintain the existing compatibility but allow teams to remove (or ignore) those behaviors from their unified test runners.

Additionally, JSON Schema draft 2019-09 adds a new ["deprecated"](https://json-schema.org/draft/2019-09/draft-handrews-json-schema-validation-02#rfc.section.9.3) property. We could consider using the newer JSON Schema and mark the unused properties as `"deprecated": true`.

matthewdale:
Updated the PR to bump JSON Schema to draft 2019-09 for schema 1.24, and deprecated the `loop` and `storeEventsAsEntities` instead of removing them.

alcaeus (approved):

PR #1812, opened 2025-06-23
DRIVERS-3123 Relaxes requirement on reads of the ignored bits of PACKED_BIT vectors.

caseyclements:
We have chosen not to require drivers to throw an error if data coming out of the server has non-zero ignored bits.  This is for backwards-compatibility. The changes are the result of conversations with Dev Tools.  The situation we want to avoid is that a non-technical user cannot get their data out.  Imagine the situations of a user of Compass, who tries to look at her data, and gets nothing but exceptions.  Instead, we require new writes to zero our data on writes, but allow freedom for drivers to migrate to the end goal of always requiring ignored bits to be zero according to their particular versioning and whether or not they've already implemented this feature.

Completed:

- [X] Update changelog.
- [X] Test changes in at least one language driver (python)
- [X] Test these changes against all server versions and topologies (including standalone, replica set, and sharded
    clusters).


source/bson-binary-vector/bson-binary-vector.md
```diff
@@ -256,6 +261,8 @@ See the [README](tests/README.md) for tests.
```
kevinAlbs, line 261: Also update the FAQ entry. Suggest rewording to:
> In PACKED_BIT, why are ignored bits **recommended** to be zero?

source/bson-binary-vector/bson-binary-vector.md
```diff
@@ -182,11 +182,16 @@ End Function
...
 
-- Padding MUST be 0 for all dtypes where padding doesn’t apply, and MUST be within [0, 7] for PACKED_BIT.
-- A PACKED_BIT vector MUST NOT be empty if padding is in the range [1, 7].
-- For a PACKED_BIT vector, ignored bits must be zero.
 - When unpacking binary data into a FLOAT32 Vector structure, the length of the binary data following the dtype and
     padding MUST be a multiple of 4 bytes.
+- Padding MUST be 0 for all dtypes where padding doesn’t apply, and MUST be within [0, 7] for PACKED_BIT.
+- A PACKED_BIT vector MUST NOT be empty if padding is in the range [1, 7].
+    - For a PACKED_BIT vector with non-zero padding, ignored bits SHOULD be zero.
+        - When encoding, if ignored bits aren't zero, drivers SHOULD zero out or raise an exception, but drivers MAY leave
```
kevinAlbs:
```suggestion
        - When encoding, if ignored bits aren't zero, drivers SHOULD raise an exception, but drivers MAY leave
```

On further thought, I recommend raising an exception. Even in a major release, I am concerned zeroing out would be a quiet breaking change:

```python
# Server has vector stored with all ignored bits set to 1.
v = Binary.from_vector([0b11111111], BinaryVectorDtype.PACKED_BIT, padding=7) # Ignored bits are 1.
got = db.coll.find_one({"v": v}) # Might match depending on driver implementation.
```

source/bson-binary-vector/tests/README.md
```diff
@@ -55,6 +55,40 @@ MUST assert that the input float array is the same after encoding and decoding.
...
+
+### Treatment of non-zero ignored bits
+
+Drivers SHOULD test this behavior, whether an error is thrown or bits are preserved, according to their design and
+version. For drivers that haven't been completed, raise an exception. For those that have, update to this behavior
+according to semantic versioning rules.
+
+A vector of length 1 with non-zero ignored bits provides a good example to use.
+
+#### 1. Encoding
```
nbbeeken, line 69: Can you make these a bit more prose-y, in particular for Decoding I'm not sure what python is returning from find_one that then needs to be passed to "as_vector". Just some 1,2,3 steps of what to create and what to assert would be great

source/bson-binary-vector/tests/README.md
```diff
@@ -55,6 +55,40 @@ MUST assert that the input float array is the same after encoding and decoding.
...
+
+```python
+    clxn.insert_one({"_id": "1", "v": Binary(b'\x10\x07\xff', subtype=9)})
+    Binary.as_vector(clxn.find_one({"_id": 1}["v"]))
+```
+
+### 3. Comparison
+
+Here is an example that shows that in pymongo < 5.0, one can decode non-zero ignored bits, but though these vectors are
+equivalent after ignoring the last 7 bits, we cannot compare their binary representations directly.
```
nbbeeken: I'd add a note that this matches current server behavior, to explain why it's a little surprising that the ignore section isn't ignored.

source/bson-binary-vector/tests/README.md
caseyclements: Ok. I haven't seen many prose tests. Will do!

source/bson-binary-vector/bson-binary-vector.md
caseyclements: > On further thought, I recommend raising an exception. Even in a major release, I am concerned zeroing out would be a quiet breaking change:
> 
> ```python
> # Server has vector stored with all ignored bits set to 1.
> v = Binary.from_vector([0b11111111], BinaryVectorDtype.PACKED_BIT, padding=7) # Ignored bits are 1.
> got = db.coll.find_one({"v": v}) # Might match depending on driver implementation.
> ```

I prefer to error, too. :)

source/bson-binary-vector/bson-binary-vector.md
```diff
@@ -182,11 +182,16 @@ End Function
...
-- Padding MUST be 0 for all dtypes where padding doesn’t apply, and MUST be within [0, 7] for PACKED_BIT.
-- A PACKED_BIT vector MUST NOT be empty if padding is in the range [1, 7].
-- For a PACKED_BIT vector, ignored bits must be zero.
 - When unpacking binary data into a FLOAT32 Vector structure, the length of the binary data following the dtype and
     padding MUST be a multiple of 4 bytes.
+- Padding MUST be 0 for all dtypes where padding doesn’t apply, and MUST be within [0, 7] for PACKED_BIT.
+- A PACKED_BIT vector MUST NOT be empty if padding is in the range [1, 7].
+    - For a PACKED_BIT vector with non-zero padding, ignored bits SHOULD be zero.
+        - When encoding, if ignored bits aren't zero, drivers SHOULD raise an exception, but drivers MAY leave them as-is if
+            backwards-compatibility is a concern.
```
kevinAlbs, line 191:
```suggestion
        - When encoding, if ignored bits aren't zero, drivers SHOULD raise an error, but drivers MAY leave them as-is for
            backwards-compatibility.
```

kevinAlbs, line 191:
```suggestion
        - When encoding, if ignored bits aren't zero, drivers SHOULD raise an error, but drivers MAY leave them as-is for backwards compatibility.
```

For consistent wording ("error" vs "exception") with the next bullet point.

source/bson-binary-vector/tests/README.md
```diff
@@ -1,3 +1,5 @@
+from xmlrpc.client import Binary
+
```
kevinAlbs:
```suggestion
```

source/bson-binary-vector/tests/README.md
```diff
@@ -55,6 +57,65 @@ MUST assert that the input float array is the same after encoding and decoding.
...
+All drivers MUST test encoding and decoding behavior according to their design and version. For drivers that haven't
+been completed, raise exceptions in both cases. For those that have, update to this behavior according to semantic
+versioning rules, and update tests accordingly.
+
+In both cases, [255], a single byte PACKED_BIT vector of length 1 (hence padding of 7) provides a good example to use,
+as all of its bits are ones.
+
+#### 1. Encoding
+
```
kevinAlbs, line 70:
```suggestion
Test encoding with non-zero ignored bits. Use the driver API that validates vector metadata.
```

Suggest clarifying to test the driver API with validation. I expect encoding can refer to many types: EJSON, BSON, `bson.binary.Binary`, `bson.binary.BinaryVector`:

> Drivers MUST perform this validation when a numeric vector and padding are provided through the API, and when unpacking binary data (BSON or similar) into a Vector structure.

source/bson-binary-vector/tests/README.md
```diff
@@ -55,6 +57,65 @@ MUST assert that the input float array is the same after encoding and decoding.
...
+- Test that an exception is raised when one attempts to encode a vector with non-zero ignored bits.
```
kevinAlbs: Since drivers may have differing behavior (to preserve backwards compat), suggest noting expectation may differ.
```suggestion
If the driver validates ignored bits are zero (preferred), expect an error. Otherwise expect the ignored bits are preserved.
```

source/bson-binary-vector/tests/README.md
```diff
@@ -55,6 +57,65 @@ MUST assert that the input float array is the same after encoding and decoding.
...
+
+```python
+with pytest.raises(ValueError):
+    Binary.from_vector([0b11111111], BinaryVectorDtype.PACKED_BIT, padding=7)
+```
+
+### 2. Decoding
+
+- Test that one can manually create a BSON Binary object following the vector structure (dtype + padding + data).
+- Test that this can be inserted and found
```
kevinAlbs:
```suggestion
```

Consider removing steps to insert+find. I expect this test is intended to only test decoding.

source/bson-binary-vector/tests/README.md
```diff
@@ -55,6 +57,65 @@ MUST assert that the input float array is the same after encoding and decoding.
...
+- Test the behaviour of your driver when one attempts to decode from binary to vector.
```
kevinAlbs, line 82: Consider adding a statement to "Skip this test if your driver does not provide a `Vector` type". I expect a vector type is not required:

> If a driver chooses to implement a `Vector` type

Similar suggestion applies to the "Comparison" test.

source/bson-binary-vector/tests/README.md
```diff
@@ -55,6 +57,65 @@ MUST assert that the input float array is the same after encoding and decoding.
...
+b1 = Binary.from_vector([0b10000000], BinaryVectorDtype.PACKED_BIT, padding=7)
+assert b1 == Binary(b'\x10\x07\x80', subtype=9) # This is effectively a roundtrip.
+v1 = Binary.as_vector(b1)
+
+b2 = Binary.from_vector([0b11111111], BinaryVectorDtype.PACKED_BIT, padding=7)
+assert b2 == Binary(b'\x10\x07\xff', subtype=9)
+v2 = Binary.as_vector(b2)
+
+assert b1 != b2  # Unequal at naive Binary level 
+assert v2 !=v1  # Also chosen to be unequal at BinaryVector level as [255] != [128]
```
kevinAlbs:
```suggestion
assert v2 != v1  # Also chosen to be unequal at BinaryVector level as [255] != [128]
```

source/bson-binary-vector/tests/README.md
```diff
@@ -55,6 +57,65 @@ MUST assert that the input float array is the same after encoding and decoding.
...
+Once we can guarantee that all ignored bits are non-zero, then equality can be tested on the binary subtype. Until then,
+equality is ambiguous, and depends on whether one compares by bits (uint1), or uint8. Drivers SHOULD test equality
+behavior according to their design and version.
+
+For example, in `pymongo < 5.0`, we define equality of a BinaryVector by matching padding, dtype, and integer. This
+means that two single bit vectors in which 7 bits are ignored do not match unless all bits match. This mirrors what the
+server does.
+
+```python
+b1 = Binary.from_vector([0b10000000], BinaryVectorDtype.PACKED_BIT, padding=7)
```
kevinAlbs, line 104: IIUC this example is showing comparison between two `bson.binary.Binary` types. Suggest changing to compare two `bson.binary.BinaryVector` types?

source/bson-binary-vector/tests/README.md
caseyclements: @kevinAlbs Have a look again.  I've added your suggestions, but not specified the many types that you rightly point out encoding could refer to. I was always assuming the bson binary with subtype 9.

source/bson-binary-vector/tests/README.md
kevinAlbs: Disregard. I see this is testing both comparing `Binary` and `BinaryVector`.

source/bson-binary-vector/tests/README.md
```diff
@@ -55,6 +55,66 @@ MUST assert that the input float array is the same after encoding and decoding.
...
+
+b2 = Binary.from_vector([0b11111111], BinaryVectorDtype.PACKED_BIT, padding=7)
+assert b2 == Binary(b'\x10\x07\xff', subtype=9)
+v2 = Binary.as_vector(b2)
+
+assert b1 != b2  # Unequal at naive Binary level 
+assert v2 != v1  # Also chosen to be unequal at BinaryVector level as [255] != [128]
+```
+
+Drivers MAY skip this test if they choose not to implement a `Vector` type.
```
kevinAlbs, line 116:
```suggestion
Drivers MAY skip this test if they choose not to implement a `Vector` type, or the type does not support comparison.
```

Suggest permitting drivers skip this test if the `Vector` type does not support comparison. The C driver `bson_vector_packed_bit_view_t` (which I expect is the closest analog to the `Vector` type) does not support comparison.

source/bson-binary-vector/tests/README.md
```diff
@@ -55,6 +55,66 @@ MUST assert that the input float array is the same after encoding and decoding.
...
+For example, in `pymongo < 5.0`, we define equality of a BinaryVector by matching padding, dtype, and integer. This
+means that two single bit vectors in which 7 bits are ignored do not match unless all bits match. This mirrors what the
+server does.
+
+```python
+b1 = Binary.from_vector([0b10000000], BinaryVectorDtype.PACKED_BIT, padding=7)
+assert b1 == Binary(b'\x10\x07\x80', subtype=9) # This is effectively a roundtrip.
+v1 = Binary.as_vector(b1)
+
+b2 = Binary.from_vector([0b11111111], BinaryVectorDtype.PACKED_BIT, padding=7)
```
kevinAlbs, line 108: I expect `Binary.from_vector` would raise an exception for non-zero ignored bits. Suggest instead directly doing:
```python
b1 = Binary(b'\x10\x07\x80', subtype=9)
b2 = Binary(b'\x10\x07\xff', subtype=9)
```

source/bson-binary-vector/tests/README.md
kevinAlbs: Latest changes LGTM.

kevinAlbs (approved):
LGTM with minor comments addressed.

Prose tests were run in the C driver [here](https://github.com/kevinAlbs/mongo-c-driver/commit/adfe35fa79994281ae8dbf114ad4006d681db68d) (excluding comparison, since the `bson_vector_packed_bit_view_t` does not support comparison).

nbbeeken (approved):
LGTM, however possibly another driver maintainer is worth looping in (up to you @caseyclements)

source/bson-binary-vector/tests/README.md
kevinAlbs: Tracked in https://github.com/mongodb/specifications/pull/1819.

source/bson-binary-vector/tests/README.md
kevinAlbs: Tracked in https://github.com/mongodb/specifications/pull/1819.

PR #1814, opened 2025-07-02
DRIVERS-3064 Add support for the rawData command option

qingyang-hu:
Add support for the new rawData command option for the following CRUD operations:
- aggregate
- bulkWrite (collection and client)
- count
- countDocuments
- deleteMany
- deleteOne
- distinct
- estimatedDocumentCount
- find
- findOne
- findAndModify
- insertMany
- insertOne
- replaceOne
- updateMany
- updateOne
- listIndexes
- dropIndexes
- createIndexes
- listCollections

- [x] Update changelog.
    clusters).


source/collection-management/tests/listCollections-rawdata.yml
```diff
@@ -0,0 +1,49 @@
...
+  - database:
+      id: &database0 database0
+      client: *client0
+      databaseName: &database0Name database0
+
+tests:
+  - description: "listCollections with rawData option"
+    runOnRequirements:
+      - minServerVersion: "8.2.0"
+        auth: false
```
alcaeus, line 18: Why do these tests require authentication to be disabled? My impression is that they should work just fine with or without auth. If so, let's remove the requirement, or alternatively document why this is necessary in a comment for future reference.

source/crud/bulk-write.md
```diff
@@ -309,6 +309,15 @@ class BulkWriteOptions {
      */
     comment: Optional<BSON value>;
 
+    /**
+     * This option allows operations to function on the bucket-level data while accessing the primary namespace (not system.buckets).
+     *
+     * @note This option will not be sent when connected to pre-9.0 servers.
+     *
+     * @since MongoDB 8.2
```
alcaeus: IIRC we typically don't do wire version checks for options that have an explicit value set - consider `comment` as an example, where the tests assert that a server error occurs on versions that don't support the `comment` field on a particular command.

In case we explicitly want this behaviour (e.g. because the cloud teams asked for it), we should change the wording that this option "MUST NOT be sent when connected to pre-8.2 servers". "will not be sent" is too ambiguous considering we're defining behaviour here.

source/crud/bulk-write.md
```diff
@@ -309,6 +309,15 @@ class BulkWriteOptions {
      */
     comment: Optional<BSON value>;
 
+    /**
+     * This option allows operations to function on the bucket-level data while accessing the primary namespace (not system.buckets).
```
alcaeus: This comment should contain a note that not all drivers need to add support for this field. Consider:
> This option MAY be implemented by drivers that need to grant access to underlying namespaces for time-series collections. Drivers SHOULD NOT implement this option unless asked to do so.

source/crud/crud.md
```diff
@@ -319,6 +319,15 @@ class AggregateOptions {
    * @see https://www.mongodb.com/docs/manual/reference/command/aggregate/
    */
   let: Optional<Document>;
+
+  /**
+   * This option allows operations to function on the bucket-level data while accessing the primary namespace (not system.buckets).
```
alcaeus: The comments from the corresponding bulk write option apply here as well.

source/collection-management/tests/listCollections-rawdata.yml
```diff
@@ -0,0 +1,49 @@
...
+          rawData: &rawdata true
+    expectEvents:
+      - client: *client0
+        events:
+          - commandStartedEvent:
+              command:
+                listCollections: 1
+                filter: {}
+                rawData: *rawdata
+  - description: "listCollections with rawData option on less than 8.2.0 - ignore argument"
```
alcaeus, line 32: Consider removing this test if wire version checks are not required.

source/index-management/index-management.md
```diff
@@ -214,6 +214,15 @@ interface CreateIndexOptions {
    */
   maxTimeMS: Optional<Int64>;
 
+  /**
+   * This option allows operations to function on the bucket-level data while accessing the primary namespace (not system.buckets).
```
alcaeus: Same comments apply as in other files.

alcaeus (changes requested):

source/crud/bulk-write.md
alcaeus: Noted that this is required as per the drivers ticket:
> Even in the initial release in a driver targetting 9.0, the addition of rawData to a command document should be guarded by a wire version check. That will allow tools like mongodump/restore to always enable rawData at the client level, and the driver will just ignore it when connected to pre-9.0 servers.

Resolving this comment

source/crud/bulk-write.md
```diff
@@ -309,6 +309,15 @@ class BulkWriteOptions {
...
+     *
+     * @note This option will not be sent when connected to pre-9.0 servers.
```
alcaeus: We should change the wording that this option "MUST NOT be sent when connected to pre-8.2 servers". "will not be sent" is too ambiguous considering we're defining behaviour here.

source/collection-management/tests/listCollections-rawdata.yml
qingyang-hu: 8.2 does not support the `rawDate` with `listCollections`:

> (Unauthorized) not authorized on database0 to execute command

alcaeus (approved):
Changes LGTM.

durran:
Node impl. LGTM.

durran (approved):

source/collection-management/tests/listCollections-rawdata.yml
```diff
@@ -0,0 +1,47 @@
...
+        object: *database0
+        arguments:
+          rawData: &rawdata true
+    expectEvents:
+      - client: *client0
+        events:
+          - commandStartedEvent:
+              command:
+                listCollections: 1
+                filter: {}
```
alcaeus: While testing for a potential auth failure with PHP, I noticed that this test initially failed because PHP does not send an empty `filter` field in the command. Since we're not testing for any filters, consider removing this line here and in the test below. Alternatively, use `$unsetOrMatches` as both are equivalent.

jmikola: I'm in support of the above change. Although our CRUD spec denotes that `find()` takes a required `filter` parameter, it says nothing about the outgoing command. And if we consult the [find command docs](https://www.mongodb.com/docs/manual/reference/command/find/#command-fields) we'll see:

> Optional. The query predicate. If unspecified, then all documents in the collection will match the predicate.

I think `$$unsetOrMatches: {}` makes sense in this case.

alcaeus (approved):
LGTM with the `filter` changes

PR #1819, opened 2025-07-17
DRIVERS-3123 add test skip and clarification

kevinAlbs:
Follow-up to https://github.com/mongodb/specifications/pull/1812. Applies pending suggestions from [review](https://github.com/mongodb/specifications/pull/1812#pullrequestreview-2971930645):

- Allow drivers to skip the comparison test if the `Vector` type does not support comparison. The C driver's [bson_vector_packed_bit_view_t](https://mongoc.org/libbson/current/bson_vector_packed_bit_view_t.html) does not support comparison.

- Do not use `Binary.from_vector` in test example with non-zero ignored bits. PyMongo 4.14.0-dev raises an exception:
```
>>> b2 = Binary.from_vector([0b11111111], BinaryVectorDtype.PACKED_BIT, padding=7)
  File "<stdin>", line 1, in <module>
  File "/Users/kevin.albertson/code/mongo-python-driver/bson/binary.py", line 482, in from_vector
    raise ValueError(
ValueError: Vector has a padding P, but bits in the final byte lower than P are non-zero. They must be zero.
```

- ~~[ ] Update changelog.~~
- ~~[ ] Test changes in at least one language driver.~~
- ~~[ ] Test these changes against all server versions and topologies (including standalone, replica set, and sharded
    clusters).~~

### DIFF
--- a/source/bson-binary-vector/tests/README.md
+++ b/source/bson-binary-vector/tests/README.md
@@ -103,7 +103,7 @@ server does.
 ```python
 b1 = Binary(b'\x10\x07\x80', subtype=9) # 1-bit vector with all 0 ignored bits.
 b2 = Binary(b'\x10\x07\xff', subtype=9) # 1-bit vector with all 1 ignored bits.
-b3 = Binary(b'\x10\x07\x80', subtype=9) # Same data as b1.
+b3 = Binary.from_vector([0b10000000], BinaryVectorDtype.PACKED_BIT, padding=7) # Same data as b1.
 
 v1 = Binary.as_vector(b1)
 v2 = Binary.as_vector(b2)
--- a/source/bson-binary-vector/tests/README.md
+++ b/source/bson-binary-vector/tests/README.md
@@ -101,8 +101,8 @@ means that two single bit vectors in which 7 bits are ignored do not match unles
 server does.
 
 ```python
-b1 = Binary(b'\x10\x07\x80', subtype=9)
-b2 = Binary(b'\x10\x07\xff', subtype=9)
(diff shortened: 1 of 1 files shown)


source/bson-binary-vector/tests/README.md
```diff
@@ -101,19 +101,17 @@ means that two single bit vectors in which 7 bits are ignored do not match unles
 server does.
 
 ```python
-b1 = Binary.from_vector([0b10000000], BinaryVectorDtype.PACKED_BIT, padding=7)
-assert b1 == Binary(b'\x10\x07\x80', subtype=9) # This is effectively a roundtrip.
-v1 = Binary.as_vector(b1)
+b1 = Binary(b'\x10\x07\x80', subtype=9)
+b2 = Binary(b'\x10\x07\xff', subtype=9)
```
caseyclements: I like this. Thank you for catching the oversight! I'd still like the example of equality though.

`assert b1 == Binary.from_vector([0b10000000], BinaryVectorDtype.PACKED_BIT, padding=7)`

caseyclements (changes requested):

source/bson-binary-vector/tests/README.md
kevinAlbs: I added another value with the same data representation as `b1` to add equality comparisons.

While here, I also realized it may not be reasonable to assume every driver may support constructing a `Vector` type with non-zero ignored bits. I added another condition to skip this test.

kevinAlbs: Discussed offline. Used the form `b3 = Binary.from_vector([0b10000000], BinaryVectorDtype.PACKED_BIT, padding=7)` to express the data in the base-literal notation.

JamesKovacs (approved):
LGTM

PR #1826, opened 2025-08-08
OpenTelemetry 

nhachicha:
DRIVERS-719
- [x] Update schema to 1.26
- [x] Added invalid test to unified test runner
- [x] Added nominal valid tests for the 3 tracing grouping (Operation, Cursor and Transaction)
- [ ] Update description of the new field `observeTracingMessages` and `expectTracingMessages`
- [ ] Merge back the draft [spec](https://github.com/mongodb/specifications/pull/1821)  into this

### DIFF
--- a/source/unified-test-format/tests/invalid/entity-client-observeTracingMessages-additionalProperties.yml
+++ b/source/unified-test-format/tests/invalid/entity-client-observeTracingMessages-additionalProperties.yml
@@ -1,0 +1,14 @@
+description: "entity-client-observeTracingMessages-additionalProperties"
+
+schemaVersion: "1.26"
+
+createEntities:
+  - client:
+      id: &client0 "client0"
+      observeTracingMessages: {
+        foo: "bar"
+      }
+
+tests:
+  - description: "observeTracingMessages must not have additional properties'"
+    operations: [ ]
--- a/source/unified-test-format/tests/invalid/entity-client-observeTracingMessages-additionalPropertyType.yml
+++ b/source/unified-test-format/tests/invalid/entity-client-observeTracingMessages-additionalPropertyType.yml
@@ -1,0 +1,14 @@
+description: "entity-client-observeTracingMessages-additionalPropertyType"
+
+schemaVersion: "1.26"
+
+createEntities:
+  - client:
+      id: &client0 "client0"
+      observeTracingMessages: {
+        enableCommandPayload: 0
+      }
+
+tests:
+  - description: "observeTracingMessages enableCommandPayload must be boolean"
+    operations: [ ]
--- a/source/unified-test-format/tests/invalid/expectedTracingSpans-additionalProperties.yml
+++ b/source/unified-test-format/tests/invalid/expectedTracingSpans-additionalProperties.yml
@@ -1,0 +1,16 @@
+description: "expectedTracingSpans-additionalProperties"
+
+schemaVersion: "1.26"
+
+createEntities:
+  - client:
+      id: &client0 "client0"
+
+tests:
+  - description: "additional property foo not allowed in expectTracingMessages"
+    operations: []
+    expectTracingMessages:
+      client: *client0
+      ignoreExtraSpans: false
+      spans: [{}]
+      foo: 0
--- a/source/unified-test-format/tests/invalid/expectedTracingSpans-clientType.yml
+++ b/source/unified-test-format/tests/invalid/expectedTracingSpans-clientType.yml
@@ -1,0 +1,14 @@
+description: "expectedTracingSpans-clientType"
+
+schemaVersion: "1.26"
+
+createEntities:
+  - client:
+      id: &client0 "client0"
+
+tests:
+  - description: "client type must be string"
+    operations: []
+    expectTracingMessages:
+      client: 0
+      spans: [{}]
--- a/source/unified-test-format/tests/invalid/expectedTracingSpans-missingPropertyClient.yml
+++ b/source/unified-test-format/tests/invalid/expectedTracingSpans-missingPropertyClient.yml
@@ -1,0 +1,13 @@
+description: "expectedTracingSpans-missingPropertyClient"
+
+schemaVersion: "1.26"
+
+createEntities:
+  - client:
+      id: &client0 "client0"
+
+tests:
+  - description: "missing required property client"
+    operations: []
+    expectTracingMessages:
+      spans: [{}]
--- a/source/unified-test-format/tests/invalid/expectedTracingSpans-missingPropertySpans.yml
+++ b/source/unified-test-format/tests/invalid/expectedTracingSpans-missingPropertySpans.yml
@@ -1,0 +1,13 @@
+description: "expectedTracingSpans-missingPropertySpans"
+
+schemaVersion: "1.26"
+
+createEntities:
+  - client:
+      id: &client0 "client0"
+
+tests:
+  - description: "missing required property spans"
+    operations: []
+    expectTracingMessages:
+      client: client0
--- a/source/unified-test-format/tests/invalid/expectedTracingSpans-spanMalformedAdditionalProperties.yml
+++ b/source/unified-test-format/tests/invalid/expectedTracingSpans-spanMalformedAdditionalProperties.yml
@@ -1,0 +1,19 @@
+description: "expectedTracingSpans-spanMalformedAdditionalProperties"
+
+schemaVersion: "1.26"
+
+createEntities:
+  - client:
+      id: &client0 "client0"
+
+tests:
+  - description: "Span must not have additional properties"
+    operations: []
+    expectTracingMessages:
+      client: client0
+      spans:
+        - name: "foo"
+          tags: {}
+          nested: []
+          foo: "bar"
+
--- a/source/unified-test-format/tests/invalid/expectedTracingSpans-spanMalformedMissingName.yml
+++ b/source/unified-test-format/tests/invalid/expectedTracingSpans-spanMalformedMissingName.yml
@@ -1,0 +1,16 @@
+description: "expectedTracingSpans-spanMalformedMissingName"
+
+schemaVersion: "1.26"
+
+createEntities:
+  - client:
+      id: &client0 "client0"
+
+tests:
+  - description: "missing required span name"
+    operations: []
+    expectTracingMessages:
+      client: client0
+      spans:
+        - tags:
+            db.system: mongodb
--- a/source/unified-test-format/tests/invalid/expectedTracingSpans-spanMalformedMissingTags.yml
+++ b/source/unified-test-format/tests/invalid/expectedTracingSpans-spanMalformedMissingTags.yml
@@ -1,0 +1,15 @@
+description: "expectedTracingSpans-spanMalformedMissingTags"
+
+schemaVersion: "1.26"
+
+createEntities:
+  - client:
+      id: &client0 "client0"
+
+tests:
+  - description: "missing required span tags"
+    operations: []
+    expectTracingMessages:
+      client: client0
+      spans:
+        - name: "foo"
--- a/source/unified-test-format/tests/invalid/expectedTracingSpans-spanMalformedNestedMustBeArray.yml
+++ b/source/unified-test-format/tests/invalid/expectedTracingSpans-spanMalformedNestedMustBeArray.yml
@@ -1,0 +1,18 @@
+description: "expectedTracingSpans-spanMalformedNestedMustBeArray"
+
+schemaVersion: "1.26"
+
+createEntities:
+  - client:
+      id: &client0 "client0"
+
+tests:
+  - description: "nested spans must be an array"
+    operations: []
+    expectTracingMessages:
+      client: client0
+      spans:
+        - name: "foo"
+          tags: {}
+          nested: {}
+
--- a/source/unified-test-format/tests/invalid/expectedTracingSpans-spanMalformedTagsMustBeObject.yml
+++ b/source/unified-test-format/tests/invalid/expectedTracingSpans-spanMalformedTagsMustBeObject.yml
@@ -1,0 +1,18 @@
+description: "expectedTracingSpans-spanMalformedNestedMustBeObject"
+
+schemaVersion: "1.26"
+
+createEntities:
+  - client:
+      id: &client0 "client0"
+
+tests:
+  - description: "span tags must be an object"
+    operations: []
+    expectTracingMessages:
+      client: client0
+      spans:
+        - name: "foo"
+          tags: []
+          nested: []
+
--- a/source/tracing/tests/cursor/cursor.json
+++ b/source/tracing/tests/cursor/cursor.json
@@ -1,0 +1,321 @@
+{
+  "description": "cursor retrieval",
+  "schemaVersion": "1.26",
+  "createEntities": [
+    {
+      "client": {
+        "id": "client0",
+        "useMultipleMongoses": false,
+        "observeTracingMessages": {
+          "enableCommandPayload": true
+        }
+      }
+    },
+    {
+      "database": {
+        "id": "database0",
+        "client": "client0",
+        "databaseName": "cursor"
+      }
+    },
+    {
+      "collection": {
+        "id": "collection0",
+        "database": "database0",
+        "collectionName": "test"
+      }
+    }
+  ],
+  "initialData": [
+    {
+      "collectionName": "test",
+      "databaseName": "cursor",
+      "documents": [
+        {
+          "_id": 1
+        },
+        {
+          "_id": 2
+        },
+        {
+          "_id": 3
+        },
+        {
+          "_id": 4
+        },
+        {
+          "_id": 5
+        },
+        {
+          "_id": 6
+        }
+      ]
+    }
+  ],
+  "tests": [
+    {
+      "description": "find with a cursor",
+      "operations": [
+        {
+          "name": "find",
+          "object": "collection0",
+          "arguments": {
+            "filter": {
+              "_id": {
+                "$gt": 1
+              }
+            },
+            "batchSize": 2
+          },
+          "expectResult": [
+            {
+              "_id": 2
+            },
+            {
+              "_id": 3
+            },
+            {
+              "_id": 4
+            },
+            {
+              "_id": 5
+            },
+            {
+              "_id": 6
+            }
+          ]
+        }
+      ],
+      "expectTracingMessages": {
+        "client": "client0",
+        "ignoreExtraSpans": false,
+        "spans": [
+          {
+            "name": "find cursor.test",
+            "tags": {
+              "db.system": "mongodb",
+              "db.namespace": "cursor",
+              "db.collection.name": "test",
+              "db.operation.name": "find",
+              "db.operation.summary": "find cursor.test"
+            },
+            "nested": [
+              {
+                "name": "command find",
+                "tags": {
+                  "db.system": "mongodb",
+                  "db.namespace": "cursor",
+                  "db.collection.name": "cursor.$cmd",
+                  "db.command.name": "find",
+                  "network.transport": "tcp",
+                  "db.mongodb.cursor_id": {
+                    "$$exists": false
+                  },
+                  "db.response.status_code": {
+                    "$$exists": false
+                  },
+                  "error.message": {
+                    "$$exists": false
+                  },
+                  "error.type": {
+                    "$$exists": false
+                  },
+                  "error.stacktrace": {
+                    "$$exists": false
+                  },
+                  "server.address": {
+                    "$$type": "string"
+                  },
+                  "server.port": {
+                    "$$type": [
+                      "long",
+                      "string"
+                    ]
+                  },
+                  "server.type": {
+                    "$$type": "string"
+                  },
+                  "db.query.summary": "find",
+                  "db.query.text": {
+                    "$$matchAsDocument": {
+                      "$$matchAsRoot": {
+                        "find": "test",
+                        "filter": {
+                          "_id": {
+                            "$gt": 1
+                          }
+                        },
+                        "batchSize": 2,
+                        "$db": "cursor"
+                      }
+                    }
+                  },
+                  "db.mongodb.server_connection_id": {
+                    "$$type": [
+                      "long",
+                      "string"
+                    ]
+                  },
+                  "db.mongodb.driver_connection_id": {
+                    "$$type": [
+                      "string"
+                    ]
+                  },
+                  "exception.message": {
+                    "$$exists": false
+                  },
+                  "exception.type": {
+                    "$$exists": false
+                  },
+                  "exception.stacktrace": {
+                    "$$exists": false
+                  }
+                }
+              },
+              {
+                "name": "command getMore",
+                "tags": {
+                  "db.system": "mongodb",
+                  "db.namespace": "cursor",
+                  "db.collection.name": "cursor.$cmd",
+                  "db.command.name": "getMore",
+                  "network.transport": "tcp",
+                  "db.mongodb.cursor_id": {
+                    "$$type": "string"
+                  },
+                  "db.response.status_code": {
+                    "$$exists": false
+                  },
+                  "error.message": {
+                    "$$exists": false
+                  },
+                  "error.type": {
+                    "$$exists": false
+                  },
+                  "error.stacktrace": {
+                    "$$exists": false
+                  },
+                  "server.address": {
+                    "$$type": "string"
+                  },
+                  "server.port": {
+                    "$$type": [
+                      "long",
+                      "string"
+                    ]
+                  },
+                  "server.type": {
+                    "$$type": "string"
+                  },
+                  "db.query.summary": "getMore",
+                  "db.query.text": {
+                    "$$matchAsDocument": {
+                      "$$matchAsRoot": {
+                        "getMore": {
+                          "$$type": "long"
+                        },
+                        "collection": "test",
+                        "batchSize": 2,
+                        "$db": "cursor"
+                      }
+                    }
+                  },
+                  "db.mongodb.server_connection_id": {
+                    "$$type": [
+                      "long",
+                      "string"
+                    ]
+                  },
+                  "db.mongodb.driver_connection_id": {
+                    "$$type": [
+                      "string"
+                    ]
+                  },
+                  "exception.message": {
+                    "$$exists": false
+                  },
+                  "exception.type": {
+                    "$$exists": false
+                  },
+                  "exception.stacktrace": {
+                    "$$exists": false
+                  }
+                }
+              },
+              {
+                "name": "command getMore",
+                "tags": {
+                  "db.system": "mongodb",
+                  "db.namespace": "cursor",
+                  "db.collection.name": "cursor.$cmd",
+                  "db.command.name": "getMore",
+                  "network.transport": "tcp",
+                  "db.mongodb.cursor_id": {
+                    "$$type": "string"
+                  },
+                  "db.response.status_code": {
+                    "$$exists": false
+                  },
+                  "error.message": {
+                    "$$exists": false
+                  },
+                  "error.type": {
+                    "$$exists": false
+                  },
+                  "error.stacktrace": {
+                    "$$exists": false
+                  },
+                  "server.address": {
+                    "$$type": "string"
+                  },
+                  "server.port": {
+                    "$$type": [
+                      "long",
+                      "string"
+                    ]
+                  },
+                  "server.type": {
+                    "$$type": "string"
+                  },
+                  "db.query.summary": "getMore",
+                  "db.query.text": {
+                    "$$matchAsDocument": {
+                      "$$matchAsRoot": {
+                        "getMore": {
+                          "$$type": "long"
+                        },
+                        "collection": "test",
+                        "batchSize": 2,
+                        "$db": "cursor"
+                      }
+                    }
+                  },
+                  "db.mongodb.server_connection_id": {
+                    "$$type": [
+                      "long",
+                      "string"
+                    ]
+                  },
+                  "db.mongodb.driver_connection_id": {
+                    "$$type": [
+                      "string"
+                    ]
+                  },
+                  "exception.message": {
+                    "$$exists": false
+                  },
+                  "exception.type": {
+                    "$$exists": false
+                  },
+                  "exception.stacktrace": {
+                    "$$exists": false
+                  }
+                }
+              }
+            ]
+          }
+        ]
+      }
+    }
+  ]
+}
--- a/source/tracing/tests/cursor/cursor.yml
+++ b/source/tracing/tests/cursor/cursor.yml
@@ -1,0 +1,144 @@
+description: cursor retrieval
+schemaVersion: '1.26'
+createEntities:
+  - client:
+      id: client0
+      useMultipleMongoses: false
+      observeTracingMessages:
+        enableCommandPayload: true
+  - database:
+      id: database0
+      client: client0
+      databaseName: cursor
+  - collection:
+      id: collection0
+      database: database0
+      collectionName: test
+initialData:
+  - collectionName: test
+    databaseName: cursor
+    documents:
+      - { _id: 1 }
+      - { _id: 2 }
+      - { _id: 3 }
+      - { _id: 4 }
+      - { _id: 5 }
+      - { _id: 6 }
+tests:
+  - description: find with a cursor
+    operations:
+      - name: find
+        object: collection0
+        arguments:
+          filter: { _id: { $gt: 1 } }
+          batchSize: 2
+        expectResult:
+          - { _id: 2 }
+          - { _id: 3 }
+          - { _id: 4 }
+          - { _id: 5 }
+          - { _id: 6 }
+    expectTracingMessages:
+      client: client0
+      ignoreExtraSpans: false
+      spans:
+        - name: find cursor.test
+          tags:
+            db.system: mongodb
+            db.namespace: cursor
+            db.collection.name: test
+            db.operation.name: find
+            db.operation.summary: find cursor.test
+          nested:
+            - name: command find
+              tags:
+                db.system: mongodb
+                db.namespace: cursor
+                db.collection.name: cursor.$cmd
+                db.command.name: find
+                network.transport: tcp
+                db.mongodb.cursor_id: { $$exists: false }
+                db.response.status_code: { $$exists: false }
+                error.message: { $$exists: false }
+                error.type: { $$exists: false }
+                error.stacktrace: { $$exists: false }
+                server.address: { $$type: string }
+                server.port: { $$type: [ 'long', 'string' ] }
+                server.type: { $$type: string }
+                db.query.summary: find
+                db.query.text:
+                  $$matchAsDocument:
+                    $$matchAsRoot:
+                      find: test
+                      filter: { _id: { $gt: 1 } }
+                      batchSize: 2
+                      $db: cursor
+                db.mongodb.server_connection_id:
+                  $$type: [ 'long', 'string' ]
+                db.mongodb.driver_connection_id:
+                  $$type: [ 'string' ]
+                exception.message: { $$exists: false }
+                exception.type: { $$exists: false }
+                exception.stacktrace: { $$exists: false }
+
+            - name: command getMore
+              tags:
+                db.system: mongodb
+                db.namespace: cursor
+                db.collection.name: cursor.$cmd
+                db.command.name: getMore
+                network.transport: tcp
+                db.mongodb.cursor_id: { $$type: string }
+                db.response.status_code: { $$exists: false }
+                error.message: { $$exists: false }
+                error.type: { $$exists: false }
+                error.stacktrace: { $$exists: false }
+                server.address: { $$type: string }
+                server.port: { $$type: [ 'long', 'string' ] }
+                server.type: { $$type: string }
+                db.query.summary: getMore
+                db.query.text:
+                  $$matchAsDocument:
+                    $$matchAsRoot:
+                      getMore: { $$type: long }
+                      collection: test
+                      batchSize: 2
+                      $db: cursor
+                db.mongodb.server_connection_id:
+                  $$type: [ 'long', 'string' ]
+                db.mongodb.driver_connection_id:
+                  $$type: [ 'string' ]
+                exception.message: { $$exists: false }
+                exception.type: { $$exists: false }
+                exception.stacktrace: { $$exists: false }
+            - name: command getMore
+              tags:
+                db.system: mongodb
+                db.namespace: cursor
+                db.collection.name: cursor.$cmd
+                db.command.name: getMore
+                network.transport: tcp
+                db.mongodb.cursor_id: { $$type: string }
+                db.response.status_code: { $$exists: false }
+                error.message: { $$exists: false }
+                error.type: { $$exists: false }
+                error.stacktrace: { $$exists: false }
+                server.address: { $$type: string }
+                server.port: { $$type: [ 'long', 'string' ] }
+                server.type: { $$type: string }
+                db.query.summary: getMore
+                db.query.text:
+                  $$matchAsDocument:
+                    $$matchAsRoot:
+                      getMore: { $$type: long }
+                      collection: test
+                      batchSize: 2
+                      $db: cursor
+                db.mongodb.server_connection_id:
+                  $$type: [ 'long', 'string' ]
+                db.mongodb.driver_connection_id:
+                  $$type: [ 'string' ]
+                exception.message: { $$exists: false }
+                exception.type: { $$exists: false }
+                exception.stacktrace: { $$exists: false }
+
--- a/source/tracing/tests/operation/find.json
+++ b/source/tracing/tests/operation/find.json
@@ -1,0 +1,136 @@
+{
+  "description" : "operation find",
+  "schemaVersion" : "1.26",
+  "createEntities" : [
+    {
+      "client" : {
+        "id" : "client0",
+        "useMultipleMongoses" : false,
+        "observeTracingMessages" : {
+          "enableCommandPayload" : true
+        }
+      }
+    },
+    {
+      "database" : {
+        "id" : "database0",
+        "client" : "client0",
+        "databaseName" : "operation-find"
+      }
+    },
+    {
+      "collection" : {
+        "id" : "collection0",
+        "database" : "database0",
+        "collectionName" : "test"
+      }
+    }
+  ],
+  "initialData" : [
+    {
+      "collectionName" : "test",
+      "databaseName" : "operation-find",
+      "documents" : []
+    }
+  ],
+  "tests" : [
+    {
+      "description" : "find an element",
+      "operations" : [
+        {
+          "name" : "find",
+          "object" : "collection0",
+          "arguments" : {
+            "filter" : {
+              "x" : 1
+            }
+          }
+        }
+      ],
+      "expectTracingMessages" : {
+        "client" : "client0",
+        "ignoreExtraSpans" : false,
+        "spans" : [
+          {
+            "name" : "find operation-find.test",
+            "tags" : {
+              "db.system" : "mongodb",
+              "db.namespace": "operation-find",
+              "db.collection.name": "test",
+              "db.operation.name": "find",
+              "db.operation.summary": "find operation-find.test"
+            },
+            "nested" : [
+              {
+                "name" : "command find",
+                "tags" : {
+                  "db.system" : "mongodb",
+                  "db.namespace" : "operation-find",
+                  "db.collection.name" : "operation-find.$cmd",
+                  "db.command.name": "find",
+                  "network.transport": "tcp",
+                  "db.mongodb.cursor_id": { "$$exists": false },
+                  "db.response.status_code": { "$$exists": false },
+                  "error.message": { "$$exists": false },
+                  "error.type": { "$$exists": false },
+                  "error.stacktrace": { "$$exists": false },
+                  "server.address" : {
+                    "$$type" : "string"
+                  },
+                  "server.port" : {
+                    "$$type" : [
+                      "long",
+                      "string"
+                    ]
+                  },
+                  "server.type" : {
+                    "$$type" : "string"
+                  },
+                  "db.query.summary" : "find",
+                  "db.query.text" : {
+                    "$$matchAsDocument" : {
+                      "$$matchAsRoot" : {
+                        "find" : "test",
+                        "filter" : {
+                          "x" : 1
+                        },
+                        "$db" : "operation-find"
+                      }
+                    }
+                  },
+                  "db.mongodb.server_connection_id" : {
+                    "$$type" : [
+                      "long",
+                      "string"
+                    ]
+                  },
+                  "db.mongodb.driver_connection_id" : {
+                    "$$type" : [
+                      "string"
+                    ]
+                  },
+                  "exception.message" : {
+                    "$$exists" : false
+                  },
+                  "exception.type" : {
+                    "$$exists" : false
+                  },
+                  "exception.stacktrace": {
+                    "$$exists" : false
+                  }
+                }
+              }
+            ]
+          }
+        ]
+      },
+      "outcome" : [
+        {
+          "collectionName" : "test",
+          "databaseName" : "operation-find",
+          "documents" : []
+        }
+      ]
+    }
+  ]
+}
--- a/source/tracing/tests/operation/find.yml
+++ b/source/tracing/tests/operation/find.yml
@@ -1,0 +1,74 @@
+description: operation find
+schemaVersion: '1.9'
+createEntities:
+  - client:
+      id: client0
+      useMultipleMongoses: false
+      observeTracingMessages:
+        enableCommandPayload: true
+  - database:
+      id: database0
+      client: client0
+      databaseName: operation-find
+  - collection:
+      id: collection0
+      database: database0
+      collectionName: test
+initialData:
+  - collectionName: test
+    databaseName: operation-find
+    documents: []
+tests:
+  - description: find an element
+    operations:
+      - name: find
+        object: collection0
+        arguments:
+          filter:
+            x: 1
+    expectTracingMessages:
+      client: client0
+      ignoreExtraSpans: false
+      spans:
+        - name: find operation-find.test
+          tags:
+            db.system: mongodb
+            db.namespace: operation-find
+            db.collection.name: test
+            db.operation.name: find
+            db.operation.summary: find operation-find.test
+          nested:
+            - name: command find
+              tags:
+                db.system: mongodb
+                db.namespace: operation-find
+                db.collection.name: operation-find.$cmd
+                db.command.name: find
+                network.transport: tcp
+                db.mongodb.cursor_id: { $$exists: false }
+                db.response.status_code: { $$exists: false }
+                error.message: { $$exists: false }
+                error.type: { $$exists: false }
+                error.stacktrace: { $$exists: false }
+                server.address: { $$type: string }
+                server.port: { $$type: ['long', 'string'] }
+                server.type: { $$type: string }
+                db.query.summary: find
+                db.query.text:
+                  $$matchAsDocument:
+                    $$matchAsRoot:
+                      find: test
+                      filter:
+                        x: 1
+                      $db: operation-find
+                db.mongodb.server_connection_id: 
+                  $$type: ['long', 'string']
+                db.mongodb.driver_connection_id: 
+                  $$type: ['string']
+                exception.message: { $$exists: false }
+                exception.type: { $$exists: false }
+                exception.stacktrace: { $$exists: false }
+    outcome:
+      - collectionName: test
+        databaseName: operation-find
+        documents: []
--- a/source/tracing/tests/operation/insert.json
+++ b/source/tracing/tests/operation/insert.json
@@ -1,0 +1,118 @@
+{
+  "description": "operation insert",
+  "schemaVersion": "1.26",
+  "createEntities": [
+    {
+      "client": {
+        "id": "client0",
+        "useMultipleMongoses": false,
+        "observeTracingMessages": {
+          "enableCommandPayload": true
+        }
+      }
+    },
+    {
+      "database": {
+        "id": "database0",
+        "client": "client0",
+        "databaseName": "operation-insert"
+      }
+    },
+    {
+      "collection": {
+        "id": "collection0",
+        "database": "database0",
+        "collectionName": "test"
+      }
+    }
+  ],
+  "initialData": [
+    {
+      "collectionName": "test",
+      "databaseName": "operation-insert",
+      "documents": []
+    }
+  ],
+  "tests": [
+    {
+      "description": "insert one element",
+      "operations": [
+        {
+          "object": "collection0",
+          "name": "insertOne",
+          "arguments": {
+            "document": {
+              "_id": 1
+            }
+          }
+        }
+      ],
+      "expectTracingMessages": {
+        "client": "client0",
+        "ignoreExtraSpans": false,
+        "spans": [
+          {
+            "name": "insert operation-insert.test",
+            "tags": {
+              "db.system": "mongodb",
+              "db.namespace": "operation-insert",
+              "db.collection.name": "test",
+              "db.operation.name": "insert",
+              "db.operation.summary": "insert operation-insert.test"
+            },
+            "nested": [
+              {
+                "name": "command insert",
+                "tags": {
+                  "db.system": "mongodb",
+                  "db.namespace": "operation-insert",
+                  "server.address": {
+                    "$$type": "string"
+                  },
+                  "server.port": {
+                    "$$type": [
+                      "long",
+                      "string"
+                    ]
+                  },
+                  "server.type": {
+                    "$$type": "string"
+                  },
+                  "db.query.summary": "insert",
+                  "db.query.text": {
+                    "$$matchAsDocument": {
+                      "$$matchAsRoot": {
+                        "insert": "test",
+                        "ordered": true,
+                        "txnNumber": {
+                          "$$unsetOrMatches": 1
+                        },
+                        "$db": "operation-insert",
+                        "documents": [
+                          {
+                            "_id": 1
+                          }
+                        ]
+                      }
+                    }
+                  }
+                }
+              }
+            ]
+          }
+        ]
+      },
+      "outcome": [
+        {
+          "collectionName": "test",
+          "databaseName": "operation-insert",
+          "documents": [
+            {
+              "_id": 1
+            }
+          ]
+        }
+      ]
+    }
+  ]
+}
--- a/source/tracing/tests/operation/insert.yml
+++ b/source/tracing/tests/operation/insert.yml
@@ -1,0 +1,63 @@
+description: operation insert
+schemaVersion: '1.26'
+createEntities:
+  - client:
+      id: client0
+      useMultipleMongoses: false
+      observeTracingMessages:
+        enableCommandPayload: true
+  - database:
+      id: database0
+      client: client0
+      databaseName: operation-insert
+  - collection:
+      id: collection0
+      database: database0
+      collectionName: test
+initialData:
+  - collectionName: test
+    databaseName: operation-insert
+    documents: []
+tests:
+  - description: insert one element
+    operations:
+      - object: collection0
+        name: insertOne
+        arguments:
+          document:
+            _id: 1
+    expectTracingMessages:
+      client: client0
+      ignoreExtraSpans: false
+      spans:
+        - name: insert operation-insert.test
+          tags:
+            db.system: mongodb
+            db.namespace: operation-insert
+            db.collection.name: test
+            db.operation.name: insert
+            db.operation.summary: insert operation-insert.test
+          nested:
+            - name: command insert
+              tags:
+                db.system: mongodb
+                db.namespace: operation-insert
+                server.address: { $$type: string }
+                server.port: { $$type: [ 'long', 'string' ] }
+                server.type: { $$type: string }
+                db.query.summary: insert
+                db.query.text:
+                  $$matchAsDocument:
+                    $$matchAsRoot:
+                      insert: test
+                      ordered: true
+                      txnNumber: { $$unsetOrMatches: 1 }
+                      $db: operation-insert
+                      documents:
+                        - _id: 1
+
+    outcome:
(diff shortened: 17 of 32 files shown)


source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,251 @@
...
+of `withTransaction`. In this case, spans for operations that are executed inside the callbacks SHOULD be nested into
+the `withTransaction` span.
+
+##### Span Name
+
+The span name SHOULD be:
+
+- `driver_operation_name db.collection_name` if the command is executed on a collection (e.g.,
+    `findOneAndDelete warehouse.users`).
+- `db.driver_operation_name` if there is no specific collection for the command (e.g., `warehouse.runCommand`).
```
GromNaN: This looks inconsistent. Why the `driver_operation_name` is at the beginning for a collection operation and at the end for a database operation?

One of this would be more consistent:
- `warehouse.users.findOneAndDelete` and `warehouse.runCommand`
- `findOneAndDelete warehouse.users` and `runCommand warehouse`

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,251 @@
...
+
+##### Span Kind
+
+Span kind MUST be "client".
+
+##### Span Attributes
+
+Spans SHOULD have the following attributes:
+
+| Attribute              | Type     | Description                                                                | Requirement Level     |
```
GromNaN, line 159: [`db.client.connection.pool.name`](https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/#db-client-connection-pool-name) is not listed. Is it because it doesn't have the "stable" flag? It would be easy to fill when the connection is created by an ODM, they usually have a connection name.

source/open-telemetry/open-telemetry.md
nhachicha: The naming follows the design document. We can revisit though 👍 https://docs.google.com/document/d/1182STFDVkQ6ixy1iCWZROC5K-Ef4M3aV__5--WXB7DI/edit?tab=t.0#heading=h.2s9zdv6a57gn

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,251 @@
...
+
+#### Tracer
+
+A Tracer is responsible for creating spans, and using a tracer is the only way to create a span. A Tracer is not
+responsible for configuration; this should be the responsibility of the TracerProvider instead.
+
+**OpenTelemetry API and SDK**
+
+OpenTelemetry offers two components for implementing instrumentation – API and SDK. The OpenTelemetry API provides all
+the necessary types and method signatures. If there is no OpenTelemetry SDK available at runtime, API methods are no-op.
```
jmikola:
```suggestion
the necessary types and method signatures. If there is no OpenTelemetry SDK available at runtime, API methods are no-ops.
```

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,251 @@
...
+Spans encapsulate:
+
+- The span name
+- An immutable SpanContext that uniquely identifies the Span
+- A parent span in the form of a Span, SpanContext, or null
+- A SpanKind
+- A start timestamp
+- An end timestamp
+- Attributes
+- A list of Links to other Spans
```
jmikola:
```suggestion
- A list of links to other Spans
```

Disregard if "Link" is intended to be a capitalized term (like "Span").

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,251 @@
...
+OpenTelemetry SHOULD be disabled by default.
+
+Drivers SHOULD support configuring OpenTelemetry on multiple levels.
+
+- **MongoClient Level**: Drivers SHOULD provide a configuration option for `MongoClient`'s Configuration/Settings that
+    enables or disables tracing for operations and commands executed with this client. This option MUST override
+    settings on higher levels. This configuration can be implemented with a `MongoClient` option, for example,
+    `tracing.enabled`.
+- **Driver Level**: Drivers SHOULD provide a global setting that enables or disables OpenTelemetry for all `MongoClient`
+    instances (excluding those that explicitly override the setting). This configuration can be implemented with an
```
jmikola:
```suggestion
    instances (excluding those that explicitly override the setting). This configuration SHOULD be implemented with an
```

If this is intended to be more prescriptive.

source/open-telemetry/open-telemetry.md
jmikola: @GromNaN: Note the same inconsistency exists in the _Instrumenting Server Commands > Span Name_ section below. I'd be in favor of using the operation/command name as the initial prefix and then appending the database or full namespace as appropriate.

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,251 @@
...
+| `db.mongodb.cursor_id`            | `int64`  | If a cursor is created or used in the operation                                                                                                                                                    | Required if available        |
+
+##### db.response.status_code and error.type
+
+These attributes should be added only if the command was not successful. The content of `error.type` is language
+specific; a driver decides what best describes the error.
+
+##### db.query.text
+
+This attribute contains the full database command executed serialized to JSON. If not truncated, the content of this
```
jmikola:
```suggestion
This attribute contains the full database command executed serialized to extended JSON. If not truncated, the content of this
```

Optionally decide if this should be canonical or relaxed.

source/open-telemetry/tests/cursor/cursor.yml
```diff
@@ -0,0 +1,135 @@
+description: cursor retrieval
+schemaVersion: '1.26'
+createEntities:
+  - client:
+      id: client0
```
jmikola: The entities IDs should utilize YAML anchors and references. See existing tests for prior art. This is also typically done for database and collection names.

The benefit is that it gives us validation at the time YAML is converted to JSON (error due to unresolved references), instead of at runtime when the test is executed.

source/open-telemetry/tests/cursor/cursor.yml
```diff
@@ -0,0 +1,135 @@
...
+            db.namespace: cursor
+            db.collection.name: test
+            db.operation.name: find
+            db.operation.summary: find cursor.test
+          nested:
+            - name: command find
+              tags:
+                db.system: mongodb
+                db.namespace: cursor
+                db.collection.name: cursor.$cmd
```
jmikola: If `db.collection.name` is always going to be `<db name>.$cmd` for a database command, is it even worth logging that?

source/open-telemetry/tests/cursor/cursor.yml
```diff
@@ -0,0 +1,135 @@
...
+                  $$type: [ 'string' ]
+
+            - name: command getMore
+              tags:
+                db.system: mongodb
+                db.namespace: cursor
+                db.collection.name: cursor.$cmd
+                db.command.name: getMore
+                network.transport: tcp
+                db.mongodb.cursor_id: { $$type: string }
```
jmikola: The spec says this is an int64 type. Shouldn't you use `$$type: [ int, long ]`? The flexibility there would account for drivers that may prefer using a smaller integer type is the value permits. I know that comes up in BSON, but I'm not sure if it will be relevant in this context.

source/open-telemetry/tests/cursor/cursor.yml
```diff
@@ -0,0 +1,135 @@
...
+                db.collection.name: cursor.$cmd
+                db.command.name: find
+                network.transport: tcp
+                db.mongodb.cursor_id: { $$exists: false }
+                db.response.status_code: { $$exists: false }
+                exception.message: { $$exists: false }
+                exception.type: { $$exists: false }
+                exception.stacktrace: { $$exists: false }
+                server.address: { $$type: string }
+                server.port: { $$type: [ 'long', 'string' ] }
```
jmikola: This is also int64 in the spec. Why would a string be possible here?

source/open-telemetry/tests/cursor/cursor.yml
```diff
@@ -0,0 +1,135 @@
...
+                server.type: { $$type: string }
+                db.query.summary: find
+                db.query.text:
+                  $$matchAsDocument:
+                    $$matchAsRoot:
+                      find: test
+                      filter: { _id: { $gt: 1 } }
+                      batchSize: 2
+                      $db: cursor
```
jmikola: According to the spec, `db.query.text`:

> SHOULD be equivalent to the `document` field of the CommandStartedEvent of the command monitoring excluding the following fields: `lsid`, `$db`, `$clusterTime`, `signature`.

source/open-telemetry/tests/cursor/cursor.yml
```diff
@@ -0,0 +1,135 @@
...
+                db.mongodb.server_connection_id:
+                  $$type: [ 'long', 'string' ]
+                db.mongodb.driver_connection_id:
+                  $$type: [ 'string' ]
```
jmikola: The spec defines both `server_connection_id` and `driver_connection_id` as int64 fields. That seems fine for the server ID, as it matches what's in the [CLAM spec](https://github.com/mongodb/specifications/blob/master/source/command-logging-and-monitoring/command-logging-and-monitoring.md). I reckon `server_connection_id` should use `$$type: [int, long]` for consistency with existing spec tests.

The driver ID seems more flexible, and CLAM doesn't prescribe a type for that. So I don't think the spec should mandate int64 for that, and this expectation should probably be changed to `{ $$exists: true }` to accept any value.

source/open-telemetry/tests/operation/find.yml
```diff
@@ -0,0 +1,71 @@
...
+                    $$matchAsRoot:
+                      find: test
+                      filter:
+                        x: 1
+                      $db: operation-find
+                db.mongodb.server_connection_id: 
+                  $$type: ['long', 'string']
+                db.mongodb.driver_connection_id: 
+                  $$type: ['string']
+    outcome:
```
jmikola: There is seemingly no need for this `outcome` assertion, as the only test is a read command.

source/open-telemetry/tests/operation/find_retries.yml
```diff
@@ -0,0 +1,113 @@
+description: operation find retrying failed command
```
jmikola: Consider including this in the existing `operation/find.yml` file so you can collect tests for like commands. It doesn't look like anything in this file differs outside of the `tests` element.

source/open-telemetry/tests/operation/find_retries.yml
```diff
@@ -0,0 +1,113 @@
...
+  - description: find an element
+    operations:
+    - name: failPoint
+      object: testRunner
+      arguments:
+        client: client0
+        failPoint:
+          configureFailPoint: failCommand
+          mode:
+            times: 1
```
jmikola: Feel free to be more concise in YAML. For instance, this can be `mode: { times: 1 }` to make the file more readable. Likewise for inlining small arrays like `failCommands`, or the `filter` argument to a `find` operation.

source/open-telemetry/tests/operation/insert.yml
```diff
@@ -0,0 +1,63 @@
...
+                server.address: { $$type: string }
+                server.port: { $$type: [ 'long', 'string' ] }
+                server.type: { $$type: string }
+                db.query.summary: insert
+                db.query.text:
+                  $$matchAsDocument:
+                    $$matchAsRoot:
+                      insert: test
+                      ordered: true
+                      txnNumber: { $$unsetOrMatches: 1 }
```
jmikola: Why would this use `$$unsetOrMatches`? I don't see that in any existing spec tests. Most assert that the value is `1` (or use `$numberLong`, which is redundant due to numeric comparison rules in the test runner).

source/open-telemetry/tests/transaction/transaction.yml
```diff
@@ -0,0 +1,117 @@
...
+              nested:
+                - name: command commitTransaction
+                  tags:
+                    db.system: mongodb
+                    db.query.text:
+                      $$matchAsDocument:
+                        $$matchAsRoot:
+                            commitTransaction: 1
+                            $db: admin
+                            lsid: { $$sessionLsid: session0 }
```
jmikola: Again, the spec talks about omitting the `$db` and `lsid` fields from logged `db.query.text` values. I can understand why `$db` would be irrelevant, but this is certainly a case where `lsid` is relevant. I can see how `lsid` might be relevant for _all_ commands.

source/open-telemetry/tests/transaction/transaction.yml
```diff
@@ -0,0 +1,117 @@
...
+                            txnNumber: 1
+                            autocommit: false
+        - name: find transaction-tests.test
+          tags: {}
```
jmikola: Why is `tags` empty here? If you don't mean to assert any tags would you not just omit the field entirely?

This might be a sensible place to leave a YAML comment explaining why you're not bothering to assert something (even if you remove this line entirely).

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,251 @@
...
+
+Drivers MAY add a dependency to the corresponding OpenTelemetry API. This is the recommended way for implementing
+OpenTelemetry in libraries. Alternatively, drivers can implement OpenTelemetry support using any suitable tools within
+the driver ecosystem. Drivers MUST NOT add a dependency to OpenTelemetry SDK.
+
+#### Enabling and Disabling OpenTelemetry
+
+OpenTelemetry SHOULD be disabled by default.
+
+Drivers SHOULD support configuring OpenTelemetry on multiple levels.
```
jmikola, line 64: Do I understand correctly that the spec tests will only exercise enabling OpenTelemetry at the MongoClient level? Would you intend to have prose tests to cover the other two mechanisms (and just assert functionality with a single operation)?

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,251 @@
...
+
+- **MongoClient Level**: Drivers SHOULD provide a configuration option for `MongoClient`'s Configuration/Settings that
+    enables or disables tracing for operations and commands executed with this client. This option MUST override
+    settings on higher levels. This configuration can be implemented with a `MongoClient` option, for example,
+    `tracing.enabled`.
```
jmikola: Should this section clarify that this is _not_ a URI option, or is that redundant? I'm not familiar with the language in other specs for non-URI MongoClient options, but we should follow prior art here. Disregard if there's no concern.

source/unified-test-format/schema-latest.json
```diff
@@ -369,6 +378,28 @@
         "connection"
       ]
     },
+    "spanComponent": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": [
+        "name",
+        "tags"
```
jmikola: If there are some cases where you do not intend to assert tags (thinking of the transaction test), should this field _not_ be required?

source/unified-test-format/tests/invalid/entity-client-observeTracingMessages-additionalProperties.yml
```diff
@@ -0,0 +1,14 @@
+description: "entity-client-observeTracingMessages-additionalProperties"
+
+schemaVersion: "1.26"
+
+createEntities:
+  - client:
+      id: &client0 "client0"
+      observeTracingMessages: {
+        foo: "bar"
+      }
```
jmikola:
```suggestion
      observeTracingMessages:
        foo: "bar"
```

The braces are redundant in YAML if you're already using multi-line syntax.

source/unified-test-format/tests/invalid/entity-client-observeTracingMessages-additionalPropertyType.yml
```diff
@@ -0,0 +1,14 @@
+description: "entity-client-observeTracingMessages-additionalPropertyType"
+
+schemaVersion: "1.26"
+
+createEntities:
+  - client:
+      id: &client0 "client0"
+      observeTracingMessages: {
+        enableCommandPayload: 0
+      }
```
jmikola:
```suggestion
      observeTracingMessages:
        enableCommandPayload: 0
```

source/unified-test-format/tests/invalid/expectedTracingSpans-additionalProperties.yml
```diff
@@ -0,0 +1,16 @@
...
+  - client:
+      id: &client0 "client0"
+
+tests:
+  - description: "additional property foo not allowed in expectTracingMessages"
+    operations: []
+    expectTracingMessages:
+      client: *client0
+      ignoreExtraSpans: false
+      spans: [{}]
```
jmikola: `spanComponent` requires `name` and `tags` fields, so this file might actually fail validation due to _that_ instead of the extra `foo` field below. For any invalid syntax files, make sure that there is exactly one syntax variation that causes a validation failure.

source/unified-test-format/tests/invalid/expectedTracingSpans-clientType.yml
```diff
@@ -0,0 +1,14 @@
...
+createEntities:
+  - client:
+      id: &client0 "client0"
+
+tests:
+  - description: "client type must be string"
+    operations: []
+    expectTracingMessages:
+      client: 0
+      spans: [{}]
```
jmikola: Same as above. This should have a valid span object to ensure `client` is actually what causes a validation error.

source/unified-test-format/tests/invalid/expectedTracingSpans-missingPropertyClient.yml
```diff
@@ -0,0 +1,13 @@
...
+
+createEntities:
+  - client:
+      id: &client0 "client0"
+
+tests:
+  - description: "missing required property client"
+    operations: []
+    expectTracingMessages:
+      spans: [{}]
```
jmikola: Same as above. This should have a valid span object to ensure a missing `client` is actually what causes a validation error.

source/unified-test-format/tests/invalid/expectedTracingSpans-missingPropertySpans.yml
```diff
@@ -0,0 +1,13 @@
...
+
+createEntities:
+  - client:
+      id: &client0 "client0"
+
+tests:
+  - description: "missing required property spans"
+    operations: []
+    expectTracingMessages:
+      client: client0
```
jmikola:
```suggestion
      client: *client0
```

Per an earlier comment, this is where you'd use YAML references. Otherwise, there's no point in defining a `&client0` anchor above (but we should for consistency).

source/unified-test-format/tests/invalid/expectedTracingSpans-spanMalformedAdditionalProperties.yml
```diff
@@ -0,0 +1,19 @@
...
+
+createEntities:
+  - client:
+      id: &client0 "client0"
+
+tests:
+  - description: "Span must not have additional properties"
+    operations: []
+    expectTracingMessages:
+      client: client0
```
jmikola:
```suggestion
      client: *client0
```

source/unified-test-format/schema-latest.json
```diff
@@ -369,6 +378,28 @@
...
+      "properties": {
+        "name": {
+          "type": "string"
+        },
+        "tags": {
+          "type": "object"
+        },
+        "nested": {
+          "type": "array",
+          "items": {
```
jmikola, line 401: Do you intend for the `nested` array to require a minimum of one item (similar to `expectTracingMessages.spans`), or is there a use case for using an empty array here?

I'll note that `minItems` introduces extra overhead for "invalid" tests, so any change here might warrant some additional test files.

source/unified-test-format/tests/invalid/expectedTracingSpans-spanMalformedNestedMustBeArray.yml
```diff
@@ -0,0 +1,18 @@
...
+
+createEntities:
+  - client:
+      id: &client0 "client0"
+
+tests:
+  - description: "nested spans must be an array"
+    operations: []
+    expectTracingMessages:
+      client: client0
```
jmikola:
```suggestion
      client: *client0
```

source/unified-test-format/tests/invalid/expectedTracingSpans-spanMalformedMissingTags.yml
```diff
@@ -0,0 +1,15 @@
...
+
+createEntities:
+  - client:
+      id: &client0 "client0"
+
+tests:
+  - description: "missing required span tags"
+    operations: []
+    expectTracingMessages:
+      client: client0
```
jmikola:
```suggestion
      client: *client0
```

source/unified-test-format/tests/invalid/expectedTracingSpans-spanMalformedMissingName.yml
```diff
@@ -0,0 +1,16 @@
...
+
+createEntities:
+  - client:
+      id: &client0 "client0"
+
+tests:
+  - description: "missing required span name"
+    operations: []
+    expectTracingMessages:
+      client: client0
```
jmikola:
```suggestion
      client: *client0
```

source/unified-test-format/tests/invalid/expectedTracingSpans-spanMalformedMissingName.yml
```diff
@@ -0,0 +1,16 @@
...
+      spans:
+        - tags:
```
jmikola:
```suggestion
        - tags: {}
```

If you'd like to simplify this.

source/unified-test-format/tests/invalid/expectedTracingSpans-spanMalformedTagsMustBeObject.yml
```diff
@@ -0,0 +1,18 @@
...
+
+createEntities:
+  - client:
+      id: &client0 "client0"
+
+tests:
+  - description: "span tags must be an object"
+    operations: []
+    expectTracingMessages:
+      client: client0
```
jmikola:
```suggestion
      client: *client0
```

source/unified-test-format/tests/invalid/expectedTracingSpans-spanMalformedTagsMustBeObject.yml
```diff
@@ -0,0 +1,18 @@
...
+      spans:
+        - name: "foo"
+          tags: []
+          nested: []
```
jmikola:
```suggestion
```

If you'd like to simplify this test.

source/unified-test-format/unified-test-format.md
```diff
@@ -523,6 +523,15 @@ The structure of this object is as follows:
...
         client's log collector(s) and SHOULD NOT be included in [test.expectLogMessages](#test_expectLogMessages) for this
         client.
+
+    - `observeTracingMessages`: Optional object that configures tracing behavior for the client. The structure of this
+        object is as follows:
+
+        - `enableCommandPayload`: Optional boolean. When set to `true`, enables capturing of command payload details in
+            tracing spans.
+            - If `true`, the test runner SHOULD capture detailed command payload information in tracing spans.
+            - If `false` or omitted, the test runner SHOULD exclude command payload details.
```
jmikola, line 533: I realize this impacts the test runner, but should it also refer to configuring the MongoClient to enable tracing?

source/open-telemetry/tests/transaction/transaction.yml
nhachicha: Good point we can remove it (alongside `$db`) there's another tag `db.mongodb.lsid` that checks the session id 
```YAML
                    db.query.text:
                      $$matchAsDocument:
                        $$matchAsRoot:
                          insert: test
                          ordered: true
                          txnNumber: 1
                          startTransaction: true
                          autocommit: false
                          documents:
                            - _id: 1
                    db.mongodb.lsid: { $$sessionLsid: *session0 }

```

source/unified-test-format/schema-latest.json
nhachicha: Good catch, `nested` is either not set or should have at least 1 span if specified.
I'll update the invalid tests 👍

source/unified-test-format/schema-latest.json
```diff
@@ -369,6 +378,29 @@
         "connection"
       ]
     },
+    "spanComponent": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": [
+        "name",
+        "tags"
```
comandeo-mongo: Why we want to call it `tags`? OpenTelemetry spans have `attributes`, not tags.

source/unified-test-format/schema-latest.json
```diff
@@ -1164,6 +1196,29 @@
             "$ref": "#/definitions/expectedLogMessagesForClient"
           }
         },
+        "expectTracingMessages": {
+          "additionalProperties" : false,
+          "type": "object",
```
comandeo-mongo: A similar attribute `expectedEvents` in [an array](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/schema-latest.json#L1154). Are we sure that we would never want to assert traces for different clients?

source/open-telemetry/open-telemetry.md
GromNaN: I see that the format of the span name has been updated for more consistency. Thank you 🙏🏻

source/open-telemetry/open-telemetry.md
comandeo-mongo: We decided to list here the minimal set of attributes that must be traced by every driver. If a driver or ODM decides to add more attributes, this is can be done of course.

GromNaN: Can this comment be added to the document, with a link to the full reference?
We could also decide that values outside the reference list are not allowed.

GromNaN: Thanks for the update.

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,312 @@
...
+OpenTelemetry offers two components for implementing instrumentation – API and SDK. The OpenTelemetry API provides all
+the necessary types and method signatures. If there is no OpenTelemetry SDK available at runtime, API methods are
+no-ops. OpenTelemetry SDK is an actual implementation of the API. If the SDK is available, API methods do work.
+
+### Implementation Requirements
+
+#### External Dependencies
+
+Drivers MAY add a dependency to the corresponding OpenTelemetry API. This is the recommended way for implementing
+OpenTelemetry in libraries. Alternatively, drivers can implement OpenTelemetry support using any suitable tools within
```
ShaneHarvey, line 57: > Alternatively, drivers can implement OpenTelemetry support using any suitable tools within the driver ecosystem.

Does this mean existing 3rd party driver OpenTelemetry integrations are suitable enough to say this spec is required? Or does it mean the driver has to implement 1st party support but can do so using the OpenTelemetry API or any alternative?

source/open-telemetry/open-telemetry.md
ShaneHarvey: Agreed. We usually call out if an option should be URI, MongoClient settings, or both. Like for versioned api:
https://github.com/mongodb/specifications/blob/master/source/versioned-api/versioned-api.md#no-uri-options

Also can you clarify how `tracing.enabled` is supposed to be translated into code? Is it `MongoClient(tracing=True)`, `MongoClient(tracing="enabled")`, or something else?

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,312 @@
...
+    disable OpenTelemetry that are more suitable for their language ecosystem. This option MUST override settings on the
+    higher level.
+- **Host Application Level**: If the host application enables OpenTelemetry for all available instrumentations (e.g.,
+    Ruby), and a driver can detect this, OpenTelemetry SHOULD be enabled in the driver.
+
+Drivers MUST NOT try to detect whether the OpenTelemetry SDK library is available, and enable tracing based on this.
+
+#### Tracer Attributes
+
+If a driver creates a Tracer using OpenTelemetry API, drivers MUST use the following attributes:
```
ShaneHarvey, line 106: Should this say "When a driver"? Is creating a Tracer optional?

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,312 @@
...
+the span as soon as possible so that the span’s duration reflects all activities made by the driver, such as server
+selection and serialization/deserialization.
+
+The span for the operation MUST be created within the current span of the host application, with the exceptions listed
+below.
+
+##### Transactions
+
+When a user starts a transaction with `startTransaction`, the driver SHOULD create a span for the pseudo operation
+`transaction`. This span MUST have only one attribute `db.system` with the value `mongodb`. All operations executed
```
ShaneHarvey, line 127: Should it also have an attribute to identify the session like lsid?

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,312 @@
...
+within the transaction SHOULD be nested to the `transaction` span.
+
+When a user commits or aborts a transaction with `commitTransaction` or `abortTransaction`, the driver SHOULD finish the
+`transaction` span.
+
+##### `withTransaction`
+
+In case of `withTransaction` operation spans for operations that are executed inside the callbacks SHOULD be nested into
+the `withTransaction` span.
```
ShaneHarvey, line 136: Should the withTransaction span have any attributes?

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,312 @@
...
+
+##### Operation Span Name
+
+The span name SHOULD be:
+
+- `driver_operation_name db.collection_name` if the operation is executed on a collection (e.g.,
+    `findOneAndDelete warehouse.users`).
+- `driver_operation_name db` if there is no specific collection for the operation (e.g., `runCommand warehouse`).
```
ShaneHarvey, line 149: Should we mention that driver_operation_name needs to be standardized to the camelCase spec-name?

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,312 @@
...
+Span kind MUST be "client".
+
+##### Operation Span Attributes
+
+Spans SHOULD have the following attributes:
+
+| Attribute              | Type     | Description                                                                | Requirement Level     |
+| :--------------------- | :------- | :------------------------------------------------------------------------- | :-------------------- |
+| `db.system`            | `string` | MUST be 'mongodb'                                                          | Required              |
+| `db.namespace`         | `string` | The database name                                                          | Required if available |
```
ShaneHarvey, line 162: I suggest we rename this attribute "db.database.name" or "db.name". In MongoDB "namespace" refers to the full "database.collection" string, not just the database name: https://www.mongodb.com/docs/manual/reference/glossary/

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,312 @@
...
+| `db.operation.name`    | `string` | The name of the driver operation being executed                            | Required              |
+| `db.operation.summary` | `string` | Equivalent to span name                                                    | Required              |
+
+Not all attributes are available at the moment of span creation. Drivers need to add attributes at later stages, which
+requires an operation span to be available throughout the complete operation lifecycle.
+
+###### db.namespace
+
+This attribute SHOULD be set to current database name except for operations executing against admin db. This field is
+omitted for transaction (abort|commit), and client `bulkWrite` operations.
```
ShaneHarvey: Why omit "admin" db operations? What happens when the app manually runs a command on "admin"?

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,312 @@
...
+
+###### db.collection.name
+
+This attribute should be set to the user's collection if the operation is executing against a collection, this field is
+omitted for commands running against `admin` database or commands that do not target a specific collection.
+
+##### Exceptions
+
+If the driver operation fails with an exception, drivers MUST record an exception to the current operation span. When
+recording an exception, drivers SHOULD add the following attributes to the span, when the content for the attribute if
```
ShaneHarvey: `for the attribute if` -> `for the attribute is`

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,312 @@
...
+| `db.command.name`                 | `string` | The name of the server command being executed                                                                                                            | Required                     |
+| `db.response.status_code`         | `string` | MongoDB error code represented as a string. This attribute should be added only if an error happens.                                                     | Required if an error happens |
+| `server.port`                     | `int64`  | Server port number                                                                                                                                       | Required                     |
+| `server.address`                  | `string` | Name of the database host, or IP address if name is not known                                                                                            | Required                     |
+| `network.transport`               | `string` | MUST be 'tcp' or 'unix' depending on the protocol                                                                                                        | Required                     |
```
ShaneHarvey, line 244: Should TLS also be tracked in "network"?

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,312 @@
...
+added and truncated to the provided value (similar to the Logging specification).
+
+On the `MongoClient` level this configuration can be implemented with a `MongoClient` option, for example,
+`tracing.query_text_max_length`.
+
+###### db.mongodb.cursor_id
+
+If the command returns a cursor, or uses a cursor, the `cursor_id` attribute SHOULD be added.
+
+##### Exceptions
```
ShaneHarvey, line 315: How is this `Exceptions` different from the one above? "If the driver operation fails with an exception, ..."

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,312 @@
...
+
+- `exception.message`
+- `exception.type`
+- `exception.stacktrace`
+
+## Motivation for Change
+
+A common complaint from our support team is that they don't know how to easily get debugging information from drivers.
+Some drivers provide debug logging, but others do not. For drivers that do provide it, the log messages produced and the
+mechanisms for enabling debug logging are inconsistent.
```
ShaneHarvey, line 329: Wasn't this the motivation for standardized logging? Could this section reference Standardized Logging and explain why OTel is a better (or necessary) alternative?

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,312 @@
...
+provide TSEs, CEs, and MongoDB users an easier way to get debugging information out of our drivers, facilitate support
+of drivers for our internal teams, and improve our documentation around troubleshooting.
+
+## Test Plan
+
+See [OpenTelemetry Tests](tests/README.md) for the test plan.
+
+## Covered operations
+
+The OpenTelemetry specification covers the following operations:
```
ShaneHarvey: Should this say something like "covers all driver operations including but not limited to"? My understanding is that every driver api should produce OTel info, even non-spec driver specific helpers.

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,312 @@
...
+| Operation                | Test                                                                                 |
+| :----------------------- | :----------------------------------------------------------------------------------- |
+| `aggregate`              | [tests/transaction/aggregate.yml](tests/operation/aggregate.yml)                     |
+| `findAndModify`          | [tests/transaction/find_one_and_update.yml](tests/operation/find_one_and_update.yml) |
+| `bulkWrite`              | [tests/transaction/bulk_write.yml](tests/operation/bulk_write.yml)                   |
+| `commitTransaction`      | [tests/transaction/transaction.yml](tests/transaction/transaction.yml)               |
+| `abortTransaction`       | [tests/transaction/transaction.yml](tests/transaction/transaction.yml)               |
+| `createCollection`       | [tests/transaction/create_collection.yml](tests/operation/create_collection.yml)     |
+| `createIndexes`          | [tests/transaction/create_indexes.yml](tests/operation/create_indexes.yml)           |
+| `createView`             | [tests/transaction/create_view.yml](tests/operation/create_view.yml)                 |
```
ShaneHarvey: createView is not a spec method. I don't think it should be included here or in the unified tests.

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,312 @@
...
+| `listIndexes`            | [tests/transaction/list_indexes.yml](tests/operation/list_indexes.yml)               |
+| `mapReduce`              | [tests/transaction/map_reduce.yml](tests/operation/map_reduce.yml)                   |
+| `estimatedDocumentCount` | [tests/transaction/count.yml](tests/operation/count.yml)                             |
+| `insert`                 | [tests/transaction/insert.yml](tests/operation/insert.yml)                           |
+| `delete`                 | [tests/transaction/delete.yml](tests/operation/delete.yml)                           |
+| `update`                 | [tests/transaction/update.yml](tests/operation/update.yml)                           |
+| `createSearchIndexes`    | [tests/transaction/atlas_search.yml](tests/operation/atlas_search.yml)               |
+| `dropSearchIndex`        | [tests/transaction/atlas_search.yml](tests/operation/atlas_search.yml)               |
+| `updateSearchIndex`      | [tests/transaction/delete.yml](tests/operation/delete.yml)                           |
+| `delete`                 | [tests/transaction/atlas_search.yml](tests/operation/atlas_search.yml)               |
```
ShaneHarvey: Is `watch()` an operation, or does it just get logged as an aggregate() cursor?

nhachicha: This will be reported as `aggregate` see more details in https://github.com/mongodb/specifications/blob/DRIVERS-719/source/open-telemetry/open-telemetry.md#operation-span-name

source/open-telemetry/open-telemetry.md
comandeo-mongo: This means drivers are not required to implement this specification using the OpenTelemetry API. Some drivers have expressed a desire not to add any additional dependencies, and this is reflected in this sentence.

source/open-telemetry/open-telemetry.md
comandeo-mongo: 1. In the Ruby driver I implemented `tracing` attribute as a `Hash`, so the code should be like `Mongo::Client.new('mongodb://...', tracing: { enabled: true })`. Do you think we should specify the syntax here?
2. As mentioned above, some driver may decide not to use OpenTelemetry API. In this case they do not have a `Tracer`.

source/open-telemetry/open-telemetry.md
comandeo-mongo: I am not sure whether this is necessary. We try to have as less attributes as possible in the first iteration of this spec (because it is hard for us to guess what is really needed).

source/open-telemetry/open-telemetry.md
comandeo-mongo: The `withTransaction` span should follow the operation span specification, and add attributes that are available.

source/open-telemetry/open-telemetry.md
comandeo-mongo: They are basically identical, we just repeated it for both operation and command spans. Maybe it worth extracting these instruction to some top level section.

source/open-telemetry/open-telemetry.md
comandeo-mongo: I think we could get the motivation part from the scope - https://docs.google.com/document/d/1h6qNCY0UCcfbGHCL4qzTCeYHf6W_IINJlLRQaTmhUZc/edit?tab=t.0

source/open-telemetry/open-telemetry.md
nhachicha: For instance, the Java driver uses Micrometre to implement this spec https://github.com/mongodb/mongo-java-driver/pull/1695

source/open-telemetry/open-telemetry.md
nhachicha: The nested commands grouped under this synthetic operation have `db.mongodb.lsid` set (see an [example](https://github.com/mongodb/specifications/pull/1826/files#diff-a3df424c5db7898786221c0a5f81c48a056a4c3c690a5a5b24fe475187eef56fR53-R84)).

source/open-telemetry/open-telemetry.md
nhachicha: `withTransaction` uses the core api, and reports similarly, i.e (setting `db.system` only see [example](https://github.com/mongodb/specifications/pull/1826/files#diff-f2f1b76eb5780c8a1658be4bea0743dc6bf762059dd132ec0a06dedb13c1bb4aR53-R56))

source/open-telemetry/open-telemetry.md
nhachicha: Yes, the naming should follow the table below https://github.com/mongodb/specifications/blob/DRIVERS-719/source/open-telemetry/open-telemetry.md#covered-operations (uses the command name)

source/open-telemetry/open-telemetry.md
nhachicha: WDYT @comandeo-mongo ? This is also the FQN in the OTel convention https://opentelemetry.io/docs/specs/semconv/registry/attributes/db/#db-namespace

source/open-telemetry/open-telemetry.md
nhachicha: This was updated, please see examples [here](https://github.com/mongodb/specifications/blob/DRIVERS-719/source/open-telemetry/open-telemetry.md#dbnamespace).

source/open-telemetry/open-telemetry.md
nhachicha: Drivers are free to add more attributes, we try to stick to what was agreed upon during the design phase. For example: Java also report `server.type` which tracks the topology (_standalone_,_replica set primary|secondary etc_.) These extra  attributes are ignored when running the unified test.

source/open-telemetry/open-telemetry.md
nhachicha: @comandeo-mongo exceptions attributes are reported only at the command level currently (see [example](https://github.com/mongodb/specifications/blob/DRIVERS-719/source/open-telemetry/tests/operation/find_retries.yml#L61-L63)). The spans collector will hoist the error state to the parent span to mark it as such, no need to set it manually

source/open-telemetry/open-telemetry.md
nhachicha: > Should this say something like "including but not limited to

Agree, in this iteration, we tried to include all the operations supported by the unified test runner. Example AFAIK there's no `dropDatabase` test in the UTR but the operation is reported in OTel.

source/open-telemetry/open-telemetry.md
nhachicha: It's a fair point, my understanding is that some operations predate the spec ex _createView_ defined in https://jira.mongodb.org/browse/DRIVERS-309 or `createCollection` only defined in unified tests. To your point in https://github.com/mongodb/specifications/pull/1826#discussion_r2337428753 We should aim to cover the majority of operations that will help the developer trace and troubleshoot their application

source/open-telemetry/open-telemetry.md
comandeo-mongo: Exceptions on the command level are reported on the command level; however, we should also report the exceptions that happens before sending the operation (for example, during the server selection process).

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,351 @@
...
+- The span name
+- An immutable SpanContext that uniquely identifies the Span
+- A parent span in the form of a Span, SpanContext, or null
+- A SpanKind
+- A start timestamp
+- An end timestamp
+- Attributes
+- A list of links to other Spans
+- A list of timestamped Events
+- A Status.
```
jmikola:
```suggestion
- A Status
```

Removing period for consistency.

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,351 @@
...
+
+#### Tracer
+
+A Tracer is responsible for creating spans, and using a tracer is the only way to create a span. A Tracer is not
+responsible for configuration; this should be the responsibility of the TracerProvider instead.
+
+**OpenTelemetry API and SDK**
```
jmikola, line 50: Is this section intended to be under `### Terms`? It definitely doesn't look like it belongs under `#### Tracer`.

It seems relevant to the paragraph under `#### External Dependencies` below, so perhaps drop the title here and just move the following paragraph to that section.

source/open-telemetry/open-telemetry.md
jmikola: @nhachicha @comandeo-mongo: Feel free to resolve this if the question is no longer relevant. Otherwise, respond and I'll resolve after reading.

source/open-telemetry/open-telemetry.md
jmikola: @ShaneHarvey: thanks for bringing up the typing. If this really is just a boolean type, I think `tracing` or `enableTracing` (more active language than `tracingEnabled`) with a default value of `false` would be most consistent with other driver option names in PHP. But since these aren't URI options there should be more naming flexibility available for drivers.

It looks like @comandeo-mongo is talking about this in a comment below, too: https://github.com/mongodb/specifications/pull/1826/files#r2339648026

source/open-telemetry/open-telemetry.md
jmikola: If you're using a `Hash`, do you anticipate there will be a need for additional options beyond `enabled`? My first impression is that such a design conflicts with the [Defy Augury](https://github.com/mongodb/specifications/blob/master/source/driver-mantras.md#defy-augury) mantra for our specs.

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,351 @@
...
+Drivers MUST NOT try to detect whether the OpenTelemetry SDK library is available, and enable tracing based on this.
+
+#### Tracer Attributes
+
+If a driver creates a Tracer using OpenTelemetry API, drivers MUST use the following attributes:
+
+- `name`: A string that identifies the driver. It can be the name of a driver's component (e.g., "mongo", "PyMongo") or
+    a package name (e.g., "com.mongo.Driver"). Drivers SHOULD select a name that is idiomatic for their language and
+    ecosystem. Drivers SHOULD follow the Instrumentation Scope guidance.
+- `version`: The version of the driver.
```
jmikola: Is this an arbitrary `string` field? If so, please clarify.

In the case of PHP, we may want to rely version information for multiple libraries (libmongoc, PHP extension, _and_ PHP library) since they can all be distinct. This is similar to what we do for the client metadata handshake.

If you'd like to borrow language from that spec you can refer to the [client.driver.version](https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.md#clientdriverversion) section, as it already has provisions for "drivers written on-top of other core drivers".

source/open-telemetry/open-telemetry.md
ShaneHarvey: > e.g., collection.findOneAndDelete(filter) will report findAndModify warehouse_db.users_coll

Thanks that's what I was looking for.

source/open-telemetry/open-telemetry.md
ShaneHarvey: Thanks, I'm okay to resolve if you feel there's nothing to clarify here.

source/open-telemetry/open-telemetry.md
ShaneHarvey: Thanks.

source/open-telemetry/open-telemetry.md
ShaneHarvey: Thanks, makes sense.

source/open-telemetry/open-telemetry.md
ShaneHarvey: The "Command Span Name" says:
> The span name SHOULD be the command name. For example, find, insert, update, etc.

I suggest we provide more clarity around the "Operation Span Name". Currently it seems ambiguous if Operation Span Name in Python is allowed to be "create_collection" (i.e our method name) or if it needs to be "createCollection".

source/open-telemetry/open-telemetry.md
ShaneHarvey: Okay I did not realize this name was predefined by OTel. Since that's the case perhaps we could add a footnote here explaining the difference between MongoDB "namespace" and OTel "namespace".

source/open-telemetry/open-telemetry.md
ShaneHarvey: Thanks, I wasn't aware of the distinction between the API vs SDK.

source/open-telemetry/open-telemetry.md
ShaneHarvey: SGTM

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,351 @@
...
+should be negligible), customers should be informed of this feature and how to disable it completely.
+
+If a driver is used in an application that has OpenTelemetry enabled, customers will see traces from the driver in their
+OpenTelemetry backends. This may be unexpected and MAY cause negative effects in some cases (e.g., the OpenTelemetry
+backend MAY not have enough capacity to process new traces). Customers should be informed of this feature and how to
+disable it completely.
+
+## Security Implication
+
+Drivers MUST take care to avoid exposing sensitive information (e.g. authentication credentials) in traces.
```
jmikola: Should this section have more guidance? Consider linking to the [Security](https://github.com/mongodb/specifications/blob/master/source/command-logging-and-monitoring/command-logging-and-monitoring.md#security) section of the CLAM spec to avoid repetition.

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,351 @@
...
+- `command_name db.collection_name` if the command is executed on a collection.
+- `command_name db` if there is no specific collection for the command.
+- `command_name admin` in other cases (e.g., commands executed against `admin` database, transaction or client
+    `bulkWrite`).
+
+###### db.query.text
+
+This attribute contains the full database command executed serialized to extended JSON. If not truncated, the content of
+this attribute SHOULD be equivalent to the `document` field of the CommandStartedEvent of the command monitoring
+excluding the following fields: `lsid`, `$db`, `$clusterTime`, `signature`.
```
jmikola, line 301: The PHP team was recently discussing [query shapes](https://www.mongodb.com/docs/manual/core/query-shapes/) and how one might remove dynamic values (e.g. `5` in `{ x: { $gt: 5 }}`) to yield a parameterized query for purposes of logging or query aggregation. In the case of CSFLE, the query might already have BSON binary values (with the encrypted subtype) by the time the query is sent along for logging, which can also be considered a form of parameterization. In that case, a driver could easily replace those binary values with placeholders (assuming the encrypted blobs are irrelevant for logging).

There's may also be some overlap here with making security redactions for other, sensitive commands.

With regard to this spec, I'd suggest mentioning query parameterization as a potential todo item in a "Future Work" section (as is found in other specs).

cc: @GromNaN @paulinevos @alcaeus

source/open-telemetry/tests/cursor/cursor.yml
```diff
@@ -0,0 +1,129 @@
+description: cursor retrieval
```
jmikola: If it's not too much work, I'd advocate for making these descriptions consistent with the base filename (sans `.yml` extension). So if "cursor retrieval" is the logical name to group tests in this file, I'd suggest using `cursor-retrieval` here and naming the file `cursor-retrieval.yml`.

source/open-telemetry/tests/cursor/cursor.yml
```diff
@@ -0,0 +1,129 @@
...
+                  db.collection.name: *collectionName0
+                  db.command.name: find
+                  network.transport: tcp
+                  db.mongodb.cursor_id: { $$exists: false }
+                  db.response.status_code: { $$exists: false }
+                  exception.message: { $$exists: false }
+                  exception.type: { $$exists: false }
+                  exception.stacktrace: { $$exists: false }
+                  server.address: { $$type: string }
+                  server.port: { $$type: ['int', 'long'] }
```
jmikola: I noticed you don't quote single values for `$$type` above, so consider using `[int, long]` here in the interest of consistency and keeping the tests as concise as possible (without sacrificing readability and introducing ambiguity in YAML parsing).

source/open-telemetry/tests/operation/aggregate.yml
```diff
@@ -0,0 +1,62 @@
...
+      database: *database0
+      collectionName: &collectionName0 test
+
+tests:
+  - description: aggregation
+    operations:
+      - name: aggregate
+        object: *collection0
+        arguments:
+          pipeline: &pipeline1
```
jmikola: Nit: was this test originally taken from a larger file? I would have expected `&pipeline0` here.

source/open-telemetry/tests/operation/atlas_search.yml
```diff
@@ -0,0 +1,139 @@
...
+
+tests:
+  - description: atlas search indexes
+    operations:
+      - name: createSearchIndex
+        object: *collection0
+        arguments:
+          model: { definition: { mappings: { dynamic: true } } , type: 'search' }
+        expectError:
+          # This test always errors in a non-Atlas environment.  The test functions as a unit test  by asserting
```
jmikola:
```suggestion
          # This test always errors in a non-Atlas environment.  The test functions as a unit test by asserting
```

Consider a search/replace on 2+ spaces within comments to reduce unnecessary whitespace.

source/open-telemetry/open-telemetry.md
ShaneHarvey: So are we going to remove createView here? Or mention that it's not standard?

source/open-telemetry/open-telemetry.md
comandeo-mongo: @nhachicha Can you please confirm that  spec tests will only exercise enabling OpenTelemetry at the MongoClient level?

nhachicha: @comandeo-mongo yes. 
The command payload, on the other hand, could be controlled via an environment variable as well.

We can add a prose test to cover enabling tracing via an env variable 👍

source/unified-test-format/schema-latest.json
nhachicha: Updated: this field is now named `attributes` and should have at least `db.system: mongodb` reported, so keeping it required.

source/unified-test-format/unified-test-format.md
nhachicha: This point is similar to https://github.com/mongodb/specifications/pull/1826#discussion_r2344133832 ? 
This option is both configured via the test runner and should be available to toggle on/off via MongoClient. This should be covered by a prose test 👍

source/open-telemetry/open-telemetry.md
nhachicha: Java uses [MongoClientSettings](https://github.com/mongodb/mongo-java-driver/pull/1695/files#diff-474bfe2587b6d1949dffa122bc49138da841f8da31e57bc9a310a30c4ff718f6R739) builder to pass in the tracer instance ([interface](https://github.com/mongodb/mongo-java-driver/pull/1695/files#diff-79f01f4f31661bc11d8c731e78eb5c8219ad41877c89415b564534dbec8d3541R40) abstracting the underlying collector) . Drivers should be free to define this in their idiomatic way.

source/open-telemetry/open-telemetry.md
nhachicha: @ShaneHarvey I believe I updated the [table](https://github.com/mongodb/specifications/blob/DRIVERS-719/source/open-telemetry/open-telemetry.md#covered-operations) while you were reviwing the PR :) 
In your case, it should be `createCollection` as this is the underlying command name (which is standardised across drivers)

source/open-telemetry/open-telemetry.md
nhachicha: It's included (with its corresponding test) https://github.com/mongodb/specifications/blob/DRIVERS-719/source/open-telemetry/open-telemetry.md#covered-operations 👍

source/open-telemetry/open-telemetry.md
comandeo-mongo: I added some clarification that we should not trace command level exception on the operation level.

source/open-telemetry/open-telemetry.md
comandeo-mongo: @ShaneHarvey I updated the motivation based on the scope, please let me know if you think it looks better now.

source/open-telemetry/open-telemetry.md
comandeo-mongo: Good point, thanks. Updated accordingly.

source/open-telemetry/open-telemetry.md
comandeo-mongo: I completely removed the Tracer section, and moved the API and SDK to the External Dependencies.

source/unified-test-format/unified-test-format.md
nhachicha: Prose test added

source/open-telemetry/tests/operation/find_retries.yml
nhachicha: done

source/open-telemetry/open-telemetry.md
comandeo-mongo: I added the section `MongoClient Options for configuring OpenTelemetry on the client level` and described the options we would like to have in the `tracing` dictionary -- https://github.com/mongodb/specifications/blob/80ae25fa9099b12b643bf8ebe0c0fd8d3643235a/source/open-telemetry/open-telemetry.md#mongoclient-options-for-configuring-opentelemetry-on-the-client-level

source/open-telemetry/open-telemetry.md
comandeo-mongo: Good point, thank you! Added the `Future Work` section.

source/open-telemetry/open-telemetry.md
jmikola: I see there's now a prose test for "Tracing Enable/Disable via Environment Variable". If that's what you had in mind feel free to resolve this thread.

source/open-telemetry/open-telemetry.md
jmikola: @nhachicha: Do you want to chime in here?

source/open-telemetry/open-telemetry.md
jmikola: Noted that `query_text_max_length` is a second option, so that addresses my previous question. I'll defer to @ShaneHarvey to resolve this thread.

source/open-telemetry/open-telemetry.md
jmikola: @nhachicha: This is still outstanding.

source/open-telemetry/open-telemetry.md
jmikola: @nhachicha @comandeo-mongo: this is outstanding.

source/open-telemetry/open-telemetry.md
```diff
@@ -0,0 +1,390 @@
...
+If a driver is used in an application that has OpenTelemetry enabled, customers will see traces from the driver in their
+OpenTelemetry backends. This may be unexpected and MAY cause negative effects in some cases (e.g., the OpenTelemetry
+backend MAY not have enough capacity to process new traces). Customers should be informed of this feature and how to
+disable it completely.
+
+## Security Implication
+
+Drivers MUST take care to avoid exposing sensitive information (e.g. authentication credentials) in traces.
+
+## Future Work
```
jmikola, line 391: @comandeo-mongo: Did you want to add a concise mention of "query parameterization" under this section? Doesn't need to be explained in detail, but just something to capture our original exchange in the [earlier comment thread](https://github.com/mongodb/specifications/pull/1826/files#r2349609313), which you can resolve.

source/open-telemetry/open-telemetry.md
```diff
@@ -192,33 +233,52 @@ Spans SHOULD have the following attributes:
...
 | `db.mongodb.driver_connection_id` | `int64`  | Local connection id                                                                                                                                      | Required if available        |
 | `db.query.text`                   | `string` | Database command that was sent to the server. Content should be equivalent to the `document` field of the CommandStartedEvent of the command monitoring. | Conditional                  |
 | `db.mongodb.cursor_id`            | `int64`  | If a cursor is created or used in the operation                                                                                                          | Required if available        |
+| `db.mongodb.lsid`                 | `string` | Logical session id                                                                                                                                       | Required is available        |
+| `db.mongodb.txn_number`           | `int64`  | Transaction number                                                                                                                                       | Required is available        |
```
ShaneHarvey: for lsid and txn_number: is -> if

source/open-telemetry/open-telemetry.md
ShaneHarvey: This typo still exists.

source/open-telemetry/open-telemetry.md
ShaneHarvey: The problem is that createView doesn't exist in any specs. I don't think this PR should introduce it as if it's a spec method that exists in all drivers.

ShaneHarvey (changes requested):

source/open-telemetry/open-telemetry.md
comandeo-mongo: The footnote added, thank you for the suggestion.

source/open-telemetry/open-telemetry.md
comandeo-mongo: I removed tests for the `createView`, and the row in this table.

source/open-telemetry/open-telemetry.md
comandeo-mongo: Added the clarification of the version format.  Resolving upon agreement with @jmikola

source/open-telemetry/open-telemetry.md
comandeo-mongo: Great suggestion, thank you. Added the link, resolving as discussed.

source/open-telemetry/open-telemetry.md
comandeo-mongo: Fixed, thanks!

source/open-telemetry/open-telemetry.md
comandeo-mongo: Addede the rationale why this is not a URI option.

source/open-telemetry/open-telemetry.md
comandeo-mongo: Added some clarification, I hope I understand correctly what you suggested.

jmikola: That LGTM. Thanks!

PR #1827, opened 2025-08-12
DRIVERS-3261 Remove IndexNotFound error checks for dropIndex

prestonvasquez:
DRIVERS-3261

> [SERVER-90152](https://jira.mongodb.org/browse/SERVER-90152) changed the behavior of dropIndex() to return ok:1 instead of an IndexNotFound error when the index does not exist.

This update removes error that would result from `IndexNotFound`. It also removed empty spaces from the yaml files via yaml language server.

- [ ] Update changelog.
    clusters).


prestonvasquez:
@sanych-sun Note that the Go Driver skips all tests in deprecated-options.yml, since this functionality was removed in Go Driver V2.

source/client-side-operations-timeout/tests/deprecated-options.yml
```diff
@@ -3748,11 +3748,6 @@ tests:
         arguments:
           timeoutMS: 100000
           name: "x_1"
-          
-        expectError:
-          isClientError: false
-          isTimeoutError: false
-
```
ShaneHarvey, line 3755: By removing expectError these tests will now fail on <8.3 servers. Is there a way to make these tests pass regardless of server version? Perhaps we should create the index first?

source/client-side-operations-timeout/tests/deprecated-options.yml
```diff
@@ -3743,16 +3743,17 @@ tests:
               failCommands: ["dropIndexes"]
               blockConnection: true
               blockTimeMS: 5
+      - name: createIndex
```
ShaneHarvey, line 3748: Could you add a comment for future readers explaining why we create the index first?

prestonvasquez:
@ShaneHarvey, @sanych-sun Suggest we remove the following templates since we need to extend the dropIndex operations to include createIndex and generate-basic-tests.py was designed to be [single-operation oriented](https://github.com/prestonvasquez/specifications/blob/0c86f31322773f2cd0ce3e9b84e3952472bf3c4c/source/client-side-operations-timeout/etc/generate-basic-tests.py#L6): 

- [deprecated-options.yml.template](https://github.com/prestonvasquez/specifications/blob/task/drivers-3261-dropIndex-doesnt-return-IndexNotFound/source/client-side-operations-timeout/etc/templates/deprecated-options.yml.template)
- [global-timeoutMS.yml.template](https://github.com/prestonvasquez/specifications/blob/task/drivers-3261-dropIndex-doesnt-return-IndexNotFound/source/client-side-operations-timeout/etc/templates/global-timeoutMS.yml.template)
- [override-collection-timeoutMS.yml.template](https://github.com/prestonvasquez/specifications/blob/task/drivers-3261-dropIndex-doesnt-return-IndexNotFound/source/client-side-operations-timeout/etc/templates/override-collection-timeoutMS.yml.template)
- [override-database-timeoutMS.yml.template](https://github.com/prestonvasquez/specifications/blob/task/drivers-3261-dropIndex-doesnt-return-IndexNotFound/source/client-side-operations-timeout/etc/templates/override-database-timeoutMS.yml.template)
- [override-operation-timeoutMS.yml.template](https://github.com/prestonvasquez/specifications/blob/task/drivers-3261-dropIndex-doesnt-return-IndexNotFound/source/client-side-operations-timeout/etc/templates/override-operation-timeoutMS.yml.template)

For example, this would be required to support the difference between creating an index for `dropIndex` v `dropIndexes`:

```yaml.template
{%- if operation.operation_name == "dropIndex" %}
# Create the index first so dropIndex doesn't return IndexNotFound on
# servers prior to 8.3.
- name: createIndex
  object: *collection
  arguments:
    keys: { x: 1 }
    timeoutMS: 100000
    {% for arg in operation.arguments -%}
    {{arg}}
    {%- endfor %}
{%- endif %}
{%- if operation.operation_name == "dropIndexes" %}
# Create the index first so dropIndex doesn't return IndexNotFound on
# servers prior to 8.3.
- name: createIndex
  object: *collection
  arguments:
    keys: { x: 1 }
    name: "x_1"
{%- endif %}
```

Additionally, each `Operation` takes arguments _specifically_ for 1 operation. That is, `generate-basic-tests.py` would have to be refactored to avoid hardcoding `keys: {x: 1}`.  We would likely have to create a new tuple / object for "Operations".

prestonvasquez:
@jyemin This will cause the [Unified Tests / Check if all JSON test files are up-to-date (pull_request)](https://github.com/mongodb/specifications/actions/runs/16977966395/job/48313292303?pr=1827) check to fail for all subsequent PRs, which will block merging. Alternatively, we could:

1. Temporarily implement the template updates required for this ticket. [jinja](https://jinja.palletsprojects.com/en/stable/) will likely lead to additional deletions via formatting.
2. Update [generate-basic-tests.py](https://github.com/mongodb/specifications/blob/master/source/client-side-operations-timeout/etc/generate-basic-tests.py) to skip the above templates.
3. Remove this check from the github workflow: https://github.com/mongodb/specifications/blob/a93fa90f78038dcb83b0b6dff0adb11e3e0386e1/.github/workflows/unified-tests.yml#L59

prestonvasquez:
@jyemin I think the second option is best. It's a very small code change and gives us time to re-assess if we've actually outgrown the template pattern for these cases.

source/client-side-operations-timeout/tests/deprecated-options.yml
```diff
@@ -3884,6 +3902,13 @@ tests:
               failCommands: ["dropIndexes"]
               blockConnection: true
               blockTimeMS: 5
+      # Create the index first so dropIndex doesn't return IndexNotFound on
+      # servers prior to 8.3.
+      - name: createIndex 
+        object: *collection
+        arguments:
+          keys: { x: 1 }
+          name: "x_1"
```
baileympearson, line 3911:
```suggestion
          keys: { x: 1 }
          name: "x_1"
          timeoutMS: 100000
```

Other socketTimeoutMS tests in this file use timeoutMS for the createIndex command, I think it was just missed here 🙂 

The test fails in Node without this change

baileympearson: Alternatively - we could use the utilClient to create the index instead of the test client

baileympearson: We'll need to regenerate the json file too, right?  @prestonvasquez

PR #1828, opened 2025-08-18
DRIVERS-2917 - Standardized Performance Testing of ODMs and Integrations

NoahStapp:

- [ ] Update changelog.
    clusters).

Python Django implementation: https://github.com/mongodb/django-mongodb-backend/pull/366.


source/benchmarking/odm-data/flat-models/small_doc.json
```diff
@@ -0,0 +1 @@
+{"field1":"miNVpaKW","field2":"CS5VwrwN","field3":"Oq5Csk1w","field4":"ZPm57dhu","field5":"gxUpzIjg","field6":"Smo9whci","field7":"TW34kfzq","field8":55336395,"field9":41992681,"field10":72188733,"field11":46660880,"field12":3527055,"field13":74094448}
```
Jibola: format

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,377 @@
...
+## Benchmark platform, configuration and environments
+
+### Benchmark Client
+
+The benchmarks should be run with the most recent stable version of the ODM and the newest version of the driver it
+supports.
+
+### Benchmark Server
+
+The MongoDB ODM Performance Benchmark must be run against a standalone MongoDB server running the latest stable database
```
Jibola: I think we can open up this to be a standalone or a replica set with a size of 1. (This is because some ODMs leverage transactions)

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,377 @@
...
+version without authentication or SSL enabled.
+
+### Benchmark placement and scheduling
+
+The MongoDB ODM Performance Benchmark should be placed within the ODM's test directory as an independent test suite. Due
```
Jibola: I still think we should leave an option for folks to create their own benchmarking repo if that helps out. I'm open to others take on this one seeing as I worry about maintainers not wanting a benchmark repo.

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,377 @@
...
+to the relatively long runtime of the benchmarks, including them as part of an automated suite that runs against every
+PR is not recommended. Instead, scheduling benchmark runs on a regular cadence is the recommended method of automating
+this suite of tests.
+
```
Jibola, line 394: Per your suggestion earlier, we should include some new information about testing mainline usecases.

source/benchmarking/odm-benchmarking.md
NoahStapp: Using a replica set of size 1 makes more sense here, agreed.

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,384 @@
...
+## ODM-specific benchmarking
+
+As discussed earlier in this document, ODM feature sets vary significantly across libraries. Many ODMs have features
+unique to them or their niche in the wider ecosystem, which makes specifying concrete benchmark test cases for every
+possible API unfeasible. Instead, ODM authors should determine what mainline use cases of their library are not covered
+by the benchmarks specified above and expand this testing suite with additional benchmarks to cover those areas.
```
NoahStapp, line 393: This section is attempting to specify that ODMs should implement additional benchmark tests to cover mainline use cases that do not fall into those included in this specification. One example would be the use of Django's `in` filter operator: `Model.objects.filter(field__in=["some_val"])`.

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,384 @@
...
+
+### Benchmark Client
+
+The benchmarks should be run with the most recent stable version of the ODM and the newest version of the driver it
+supports.
+
+### Benchmark Server
+
+The MongoDB ODM Performance Benchmark must be run against a MongoDB replica set of size 1 running the latest stable
+database version without authentication or SSL enabled.
```
dariakp: Are we concerned at all about accounting for performance variation due to server performance differences? In the drivers, we keep the server version patch-pinned and upgrade rarely and intentionally via independent commits in order to ensure that our performance testing results are meaningful and are only reflective of the changes in the system under test (the driver, or, in this case, the ODM). If the goal is only to track the performance of ODMs relative to each other and relative to the corresponding drivers, is the intention to have the drivers also implement these tests against the latest server so that we could get that apples-to-apples comparison?

aclark4life: > Are we concerned at all about accounting for performance variation due to server performance differences? 

From the [Django implementation](https://github.com/mongodb/django-mongodb-backend/pull/366):

>> This is NOT intended to be a comprehensive test suite for every operation, only the most common and widely applicable

@NoahStapp and @Jibola are working on this project for DBX Python (although I am reviewing the implementation PR), so this is just a drive by comment from me, but my impression is that the spec is at least initially focused on getting all the ODMs to agree on what to test.

> In the drivers, we keep the server version patch-pinned and upgrade rarely and intentionally via independent commits in order to ensure that our performance testing results are meaningful and are only reflective of the changes in the system under test (the driver, or, in this case, the ODM). If the goal is only to track the performance of ODMs relative to each other and relative to the corresponding drivers, is the intention to have the drivers also implement these tests against the latest server so that we could get that apples-to-apples comparison?

One more drive by comment: I'd expect each ODM to "perform well" under similar server circumstances (testing the driver is a good call out!) but I'm not sure apples-to-apples is the goal. If other ODMs test their performance using the spec and can demonstrate "good performance" and/or catch performance issues they would otherwise have missed, that would indicate some measure of success to me in the spec design.

NoahStapp: I chose `latest stable` server version here for the following reason: we've made server performance an explicit company-wide goal. When users experience performance issues on older server versions, one of the first things we recommend is that they upgrade to a newer version. At least in the Python driver, we only run performance tests against 8.0. Using the latest stable version ensures that our performance tests always take advantage of any server improvements and isolate performance issues in the ODM or underlying driver.

Implementing these same tests in the driver for a direct apples-to-apples comparison is a significant amount of work. Several of the tests here use similar datasets as the driver tests for easier comparison, so using the same version of the server as the driver tests to reduce differences could be useful.

dariakp: > Using the latest stable version ensures that our performance tests always take advantage of any server improvements and isolate performance issues in the ODM or underlying driver.

I think we should be careful about our goals here: if it is to take advantage of any server improvements and track performance explicitly relative to the most current server performance, then this approach is fine. However, this approach will *not* isolate performance issues in the ODM or driver because: 1) server performance is not guaranteed to always improve in every release for every feature: the overall trends of the server performance for most features will hopefully keep moving up, but between releases there may be "acceptable" regressions to certain features that are considered a tradeoff to an improvement in another area, and 2) server performance improvements could mask ODM regressions that happen concurrently with the server upgrade. We should be explicit about accepting both of these risks if we are going to move forward with this approach (i.e., note this somewhere in the spec text).

NoahStapp: Good callouts. What if we test the benchmarks against both the latest stable version as well as the latest major release? Currently that would be 8.1 and 8.0, for example. That would give us a yearly cadence of upgrading that should allow us to catch server regressions without blindly masking ODM regressions.

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,384 @@
...
+This balances measurement stability with a timing cap to ensure all tasks can complete in a reasonable time.
+
+For each task, the 10th, 25th, 50th, 75th, 90th, 95th, 98th and 99th percentiles will be recorded using the following
+algorithm:
+
+- Given a 0-indexed array A of N iteration wall clock times
+- Sort the array into ascending order (i.e. shortest time first)
+- Let the index i for percentile p in the range [1,100] be defined as: `i = int(N * p / 100) - 1`
+
+*N.B. This is the [Nearest Rank](https://en.wikipedia.org/wiki/Percentile#The_Nearest_Rank_method) algorithm, chosen for
```
JamesKovacs: `#The_Nearest_Rank_method` anchor should be `#The_nearest-rank_method`.

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,384 @@
...
+- iterations should stop after 10 iterations or 1 minute cumulative execution time, whichever is shorter
+
+This balances measurement stability with a timing cap to ensure all tasks can complete in a reasonable time.
+
+For each task, the 10th, 25th, 50th, 75th, 90th, 95th, 98th and 99th percentiles will be recorded using the following
+algorithm:
+
+- Given a 0-indexed array A of N iteration wall clock times
+- Sort the array into ascending order (i.e. shortest time first)
+- Let the index i for percentile p in the range [1,100] be defined as: `i = int(N * p / 100) - 1`
```
JamesKovacs: Given that the maximum iteration count is 10 (see line 109 above), the 90th, 95th, 98th, and 99th percentiles will all be A[8] since `int(float)` truncates the float.

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,384 @@
...
+    database
+
+The wall-clock execution time of each "do operation" phase will be recorded. We use wall clock time to model user
+experience and as a lowest-common denominator across ODMs. Iteration timing should be done with a high-resolution
+monotonic timer (or best language approximation).
+
+Unless otherwise specified, the number of iterations to measure per task is variable:
+
+- iterations should loop for at least 30 seconds cumulative execution time
+- iterations should stop after 10 iterations or 1 minute cumulative execution time, whichever is shorter
```
JamesKovacs: Those two conditions seem to be working at cross purposes. The measurement should loop for at least 30 seconds but not more than 60, but stop after 10 iterations. This caps the number of iterations at 10, possibly fewer if each iteration takes longer than 6 seconds.

source/benchmarking/odm-benchmarking.md
JamesKovacs: As noted in the Wikipedia article, fewer than 100 measurements will result in the same value being reported for multiple percentiles.

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,384 @@
...
+
+Datasets are in the `flat_models` tarball.
+
+Flat model tests focus on flatly-structured model reads and writes across data sizes. They are designed to give insights
+into the efficiency of the ODM's implementation of basic data operations.
+
+The data will be stored as strict JSON with no extended types. These JSON representations must be converted into
+equivalent models as part of each benchmark task.
+
+Flat model benchmark tasks include:s
```
JamesKovacs: Extraneous `s` at the end of the line.

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,384 @@
...
+
+Dataset: The dataset (SMALL_DOC) is contained within `small_doc.json` and consists of a sample document stored as strict
+JSON with an encoded length of approximately 250 bytes.
+
+Dataset size: For scoring purposes, the dataset size is the size of the `small_doc` source file (250 bytes) times 10,000
+operations, which equals 2,250,000 bytes or 2.5 MB.
+
+| Phase       | Description                                                                                                                                                                |
+| ----------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| Setup       | Load the SMALL_DOC dataset into memory as an ODM-appropriate model object. Insert 10,000 instances into the database, saving the inserted `id` field for each into a list. |
```
JamesKovacs: Is this the `_id`?

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,384 @@
...
+| After task  | n/a.                                                                                                                                                                                                                                                                                              |
+| Teardown    | Drop the collections associated with the SMALL_DOC and FOREIGN_KEY models.                                                                                                                                                                                                                        |
+
+#### Large model creation
+
+Summary: This benchmark tests ODM performance creating a single large model.
+
+Dataset: The dataset (LARGE_DOC) is contained within `large_doc.json` and consists of a sample document stored as strict
+JSON with an encoded length of approximately 8,000 bytes.
```
JamesKovacs: 8,000 bytes is still relatively small. Do we want to have perf tests for huge documents close to the 16MB limit? While we may not recommend large models, customers will run into these scenarios especially if their models contain large arrays of subdocuments.

source/benchmarking/odm-benchmarking.md
JamesKovacs: Good point about the stability. If we see a perf regression (or improvement), we then have to consider whether we actually made things worse (or better) or if we happened to run on a newer server version that had different perf characteristics. We have correctness tests against different server versions. I don't think there is value in testing the server's performance in our ODM tests. Thus I would suggest we choose 8.0.13 (latest stable as of today) and make an explicit choice to update it on an annual cadence.

source/benchmarking/odm-benchmarking.md
NoahStapp: Good points--this whole section is copied from the existing driver benchmark spec for consistency, which raises the question of should we (as a separate ticket) update that spec as well? I would say yes to keep both benchmarking specs as consistent in behavior and design as possible.

The data size, shape, and specific operation of a benchmark are the limiting factors for how many iterations are ultimately run. We expect most of the tests to run more than 100 iterations in the allotted time, but the more expensive ones don't. Have we historically actually used these percentiles or plan to in the future? From my experience, at least the Python team primarily uses the MB/s metric to identify regressions. If this is a consistent pattern across teams and continues to be, recording this additional data doesn't seem useful.

source/benchmarking/odm-benchmarking.md
NoahStapp: This is confusing on my part (also taken from the driver benchmarking spec).

The intent is to have a 30 second minimum execution time with a 120 second execution time cap. Once the minimum time is reached, we stop the benchmark being executed once it reaches 120 seconds of execution time or once at least 10 iterations have completed.

source/benchmarking/odm-benchmarking.md
NoahStapp: Yes. I'll update the wording to clarify since ODM naming conventions for the document `_id` will vary.

source/benchmarking/odm-benchmarking.md
NoahStapp: Close to the 16MB limit seems excessive and would increase execution time significantly. Increasing the size here to be a few MB, similar to what the driver benchmarks use for their large document tests, would likely result in similar performance characteristics without as large of a latency hit. The downside to increasing the size of documents here is that we need to define the data's structure carefully to not significantly complicate the process of model creation for implementing ODMs, which is not a concern for the driver benchmarks.

source/benchmarking/odm-benchmarking.md
NoahStapp: The main advantage of testing against rapid server versions is query performance improvements. Since ODMs necessarily construct database queries for the user, they don't have any control over what's actually sent to the server barring a feature like `raw_aggregate` that allows them to specify the actual query itself. With the server improving query performance and optimization (for example, $in inside $expr using indexes starting in 8.1: [SERVER-32549](https://jira.mongodb.org/browse/SERVER-32549)), it's possible we run into situations where the best way to improve performance is for a user to upgrade their server version. Some of these, such as using $expr where it's not necessary, can be fixed with ODM code improvements, but that isn't a guarantee. Being able to tell users that upgrading to the latest rapid release will improve performance for their use case could be helpful, but I can see the downside of testing an additional server version besides latest stable.

source/benchmarking/odm-benchmarking.md
JamesKovacs: A few MB makes sense to me for the reasons stated. If the intent is to test large payloads, we could have a binary field that contains a 1MB JPG of a leaf. (e.g. Something incompressible.)

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,385 @@
...
+### Task Hierarchy
+
+The benchmark suite consists of two groups of small, independent benchmarks. This allows us to better isolate areas
+within ODMs that are faster or slower.
+
+- Flat models -- reading and writing flat models of various sizes, to explore basic operation efficiency
+- Nested models -- reading and writing nested models of various sizes, to explore basic operation efficiency for complex
+    data
+
+The suite is intentionally kept small for several reasons. One, ODM feature sets vary significantly across libraries.
```
aclark4life: May prefer bulleted list here e.g. 

---
The suite is intentionally kept small for the following reasons:
- ODM feature sets vary …
- Several popular MongoDB ODMs are maintained by third-parties …

source/benchmarking/odm-benchmarking.md
aclark4life: We [don't agree](https://github.com/mongodb/django-mongodb-backend/pull/366#issuecomment-3198723449) that they should be in the `tests` directory but haven't ruled out [including them in the ODM](https://github.com/mongodb/django-mongodb-backend/pull/366#). For the purposes of getting the spec done, I wonder if requiring the ODM to document the location of the test suite is enough. If not, I would definitely remove the "test directory" requirement and make it "should be placed within the ODM". I think that is enough to make it clear that the goal is to have the perf tests included in the ODM.

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,385 @@
...
+
+### Name and purpose
+
+ODM performance will be measured by the MongoDB ODM Performance Benchmark. It will provide both "horizontal" insights
+into how individual ODM performance evolves over time and two types of "vertical" insights: the relative performance of
+different ODMs, and the relative performance of ODMs and their associated language drivers.
+
+We expect substantial performance differences between ODMs based on both their language families (e.g. static vs.
+dynamic or compiled vs. virtual-machine-based) as well as their inherent design (e.g. web frameworks such as Django vs.
+application-agnostic such as Mongoose). However we still expect "vertical" comparison within families of ODMs to expose
```
ajcvickers: I don't think it is worthwhile to compare different ODMs to each other. The performance of ODMs doing different types of things varies widely based on the approach taken by the ODM, as opposed to anything the provider/adapter for Mongo is doing.

I do think it could be valuable to compare a given ODM with Mongo to that same ODM but with a similar (e.g. Cosmos) database, and a different (e.g. PostgreSQL) database. Whether or not this will show differences in the client is dependent on many things. For example, in .NET making the PostgreSQL provider faster is measurable because the data transfer and server can keep up. On the other hand, making the SQL Server provider faster makes no difference, because the wire protocol and server blocking is already the limiting factor.

It may also be useful to test raw driver to ODM perf, especially since customers often ask about this. However, in most cases the performance overhead will come from the core ODM code, rather than anything we are doing, so I doubt that there will be much actionable to come out of this.

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,385 @@
...
+The MongoDB ODM Performance Benchmark must be run against a MongoDB replica set of size 1 running the latest stable
+database version without authentication or SSL enabled.
+
+### Benchmark placement and scheduling
+
+The MongoDB ODM Performance Benchmark should be placed within the ODM's test directory as an independent test suite. Due
+to the relatively long runtime of the benchmarks, including them as part of an automated suite that runs against every
+PR is not recommended. Instead, scheduling benchmark runs on a regular cadence is the recommended method of automating
+this suite of tests.
+
```
ajcvickers, line 387: Do we have a dedicated, isolated perf lab, with machines that won't get changes unless we know about it? My experience with perf testing over many years is that unless you have such a system, then the noise makes it very difficult to see when things really change. For example, OS updates, platform/language changes, virus checking, auto-updates kicking in mid run, and so on, all make the data hard to interpret.

How do you currently handle driver perf test machines? Can you point me to charts, or even raw data I guess, that should variation/noise over time? Also, how often do they run? Is there only a single change between each run so that it's feasible to trace back a perf difference to a single change, be that external or a code change?

source/benchmarking/odm-benchmarking.md
NoahStapp: I don't think a separate benchmark repo is a good choice here. We could reach out to existing maintainers and see if they want to weigh in, but I imagine having a separate repo for benchmarking is more trouble than it's worth.

source/benchmarking/odm-benchmarking.md
NoahStapp: Comparing ODMs to each other could be useful in identifying potential design or implementation issues. For example, if one ODM implements embedded document querying in an inefficient way, comparing its performance on a benchmark to a similar ODM with much better performance could unlock improvements that would be difficult to identify otherwise. Outside of that specific case, I agree that ODM comparisons are not very useful.

Comparing performance across databases is an interesting idea. Django did apples-to-apples comparisons with benchmarks against both MongoDB and Postgres and got a lot of useful data out of that. ODMs make doing so relatively easy as only the backend configuration and models (for differences like embedded documents and relational links) need to change. We'd need to be careful to isolate performance differences to the database alone as much as possible, due to all the factors you state.

Comparing raw driver to ODM perf is part of the stated goals of this project. Determining exactly which benchmarks should be directly compared is still under consideration, for both maintainability and overhead concerns.

source/benchmarking/odm-benchmarking.md
baileympearson: I worry that ODMs might not be receptive to a large addition of performance tests to their repository.  The ticket makes it sound like we (DBX) planned to run these tests ourselves, probably in a CI we maintain:

> The testing criteria would be documented in a human readable form (such as either a docs page or a markdown file), and once benchmarks have been developed we would run these against each new notable release of the client library. Providing well documented benchmarks will hopefully also encourage the developer community to contribute additional tests to further improve coverage.

And from the comments: https://jira.mongodb.org/browse/DRIVERS-2917#:~:text=Shane%20Harvey%20FYI%20that%20on%20https%3A//www.mongodb.com/services/support/mongoose%2Dodm%2Dsupport%20we%20specifically%20call%20out%20that%20%22MongoDB%E2%80%99s%20team%20of%20experts%20rigorously%20tests%20new%20Mongoose%20releases%20to%20ensure%20they%20are%20compatible%20with%20MongoDB%20and%20meet%20appropriate%20performance%20benchmarks.%22.

I don't see any mention of where these tests will live in the scope, either.

Why do we plan on contributing spec tests to ODM repos, instead of creating a pipeline similar to the ai testing pipeline?  Or just integrating perf testing within drivers' existing performance testing?  We already have the test runners and infrastructure to run these ourselves.  And to @ajcvickers 's [point](https://github.com/mongodb/specifications/pull/1828#discussion_r2336260180), we already have dedicated performance test hosts in evergreen that are stable and isolated from other hosts in CI.

source/benchmarking/odm-benchmarking.md
NoahStapp: Here's [an example](https://performance-monitoring-and-analysis.server-tig.prod.corp.mongodb.com/task/mongo_python_driver_performance_benchmarks_perf_8.0_standalone_patch_b84e1a7ce4f53faf3b4898fdf8b5b89251f8c047_68bedf2ba0787a000775077c_25_09_08_13_50_44/performanceData?time_series_filters=%5B%7B%22active%22%3Atrue%2C%22name%22%3A%22test%22%2C%22operator%22%3A%22contains%22%2C%22type%22%3A%22string%22%7D%2C%7B%22active%22%3Atrue%2C%22name%22%3A%22measurement%22%2C%22operator%22%3A%22contains%22%2C%22type%22%3A%22string%22%7D%2C%7B%22active%22%3Atrue%2C%22name%22%3A%22argString%22%2C%22operator%22%3A%22contains%22%2C%22type%22%3A%22string%22%7D%2C%7B%22active%22%3Atrue%2C%22name%22%3A%22patch_result%22%2C%22operator%22%3A%22eq%22%2C%22type%22%3A%22number%22%7D%2C%7B%22active%22%3Atrue%2C%22name%22%3A%22contexts%22%2C%22operator%22%3A%22inlist%22%2C%22type%22%3A%22listSelectFromList%22%7D%2C%7B%22active%22%3Atrue%2C%22name%22%3A%22mainline_commit%22%2C%22operator%22%3A%22contains%22%2C%22type%22%3A%22string%22%7D%2C%7B%22active%22%3Atrue%2C%22name%22%3A%22mainline_result%22%2C%22operator%22%3A%22eq%22%2C%22type%22%3A%22number%22%7D%2C%7B%22active%22%3Atrue%2C%22name%22%3A%22mainline_std%22%2C%22operator%22%3A%22eq%22%2C%22type%22%3A%22number%22%7D%2C%7B%22active%22%3Atrue%2C%22name%22%3A%22percent%22%2C%22operator%22%3A%22eq%22%2C%22type%22%3A%22number%22%7D%2C%7B%22active%22%3Atrue%2C%22name%22%3A%22execution%22%2C%22operator%22%3A%22eq%22%2C%22type%22%3A%22number%22%7D%2C%7B%22active%22%3Atrue%2C%22name%22%3A%22z%22%2C%22operator%22%3A%22eq%22%2C%22type%22%3A%22number%22%7D%5D) of what the Python driver perf tests output. The driver perf tests have experienced all of the issues you've stated, but still provide useful metrics that let us catch regressions and identify places for improvement. Running on Evergreen doesn't allow us (AFAIK) to have our own dedicated set of machines.

The Python driver perf tests run weekly.

source/benchmarking/odm-benchmarking.md
NoahStapp: I don't believe there was any concrete plan one way or the other at the time the ticket and scope were created.

In my view, there are a few fundamental differences between the libraries being tested here versus for AI integrations.
- Many ODMs are or are planned to be first-party libraries rather than contributions to third-party AI frameworks.
- The AI space moves extremely rapidly and broken CI/CD or testing suites are extremely common. Both factors were significant motivators in the creation of our AI testing pipeline. Those motivations don't seem to exist here.
- AI frameworks tend to have several to dozens of integrations all housed within a single repo, each with their own dependencies and tests. Third-party ODMs are more often standalone repos with far less complexity in this manner, so adding a single additional test file for performance testing is much less significant.

What would integrating perf testing within the existing drivers perf testing look like? Would all of the ODM benchmarks live in a separate repo, with each driver cloning and using the specific subdirectory that contains the ODMs they want to test? 

Using the same skeleton of test runners and infrastructure for the ODM testing makes it very easy to get these tests up and running without polluting the existing drivers tests.

aclark4life: > I don't think a separate benchmark repo is a good choice here. We could reach out to existing maintainers and see if they want to weigh in, but I imagine having a separate repo for benchmarking is more trouble than it's worth.

I'm suggesting resolving two issues:

- In the same repo or not
- In the repo's test dir or just in the repo, period.

NoahStapp: I would advocate for aligning with the structure we use for drivers: in the same repo as the ODM, in the repo's test directory.

baileympearson: What are we hoping to accomplish with the benchmarks?  Are we (MongoDB) trying to get insight into the performance of ODMs? Or do we want maintainers to take more ownership of performance of their ODM in a standardized way?

If the former - I think our repos make the most sense.  If the latter, I agree that it makes sense to put them in ODM's repos, but I think we need buy-in from ODM authors.  We've had issues contributing features to ODMs and AI integrations in the past, and I worry contributing large perf testing suites with specific requirements and a custom runner will face pushback in each repo we attempt to add it to.

NoahStapp: I don't think trying to get maintainers to take ownership of performance is a sensible goal here. 

How many third-party versus first-party ODMs do we have? This testing suite is designed to be relatively small and concise for exactly the reason of not wanting to add a large suite with requirements and complex runners.

baileympearson: The only first-party integrations I know of are symfony and mongoid.  Maybe there are more?

NoahStapp: Django, Entity Framework, and Hibernate are all first-party ODMs either in development or recently released.

We also have Spring Data MongoDB, Laravel, and Doctrine.

baileympearson: First party meaning "we own the repo" or "we contributed the code"?  I was going with the former

NoahStapp: Spring Data MongoDB and Doctrine we don't own the repo, true. Still, I expect the number of first-party ODMs to continue to grow.

source/benchmarking/odm-benchmarking.md
baileympearson: Drivers do have access to a special host in evergreen to run dedicated performance tasks on to ensure stability and consistency (`rhel90-dbx-perf-large`).

source/benchmarking/odm-benchmarking.md
baileympearson: Okay, that's more evenly split between "we control the changes that go in" and "others control the changes that go in" than I originally thought.

I still am not sure that putting the tests in ODM repos makes sense, unless the goal is to integrate testing into the ODM's processes (even for repos we do not control).  My primary concern here is that I anticipate difficulties getting buy-in from external stakeholders for these tests (speaking from experience contributing to integrations, both AI and ODMs, in the JS ecosystem).  Also, I'd like to understand the goal here, because if its for our own understanding of performance of ODMs, putting the tests in a repo we do not maintain or one that doesn't use evergreen raises a lot of unnecessary questions:

What do we do if a maintainer doesn't want / pushes back on these tests?  What happens if they break?  When should the tests be run? Who handles regressions? What is the triage process for potential regressions and how is flakiness/false positives handled?  How do maintainers / us ensure builds are stable without dedicated perf testing hosts in CI?  How do we expect each repository to setup the dedicated testing cluster (Mongoose pushed back against drivers-evergreen-tools)? etc.

NoahStapp: Here's the scope doc, which covers the motivations of this work: https://docs.google.com/document/d/1GCle2vTQLdoSaDJJXyXeXYqtcAfymr8pM5oyV4gSI4A/edit?tab=t.0#heading=h.b1os3ai9s8t3. 

Integrating testing into ODM processes is preferable for both visibility and maintenance reasons. Users will likely be more comfortable using a library with very public and integrated performance tests, and having all testing for an ODM live within a single repo streamlines maintenance work. Having the performance tests be integrated also shows a higher level of accountability and transparency, especially if we end up adding performance tests that directly compare against Postgres or other SQL databases. 

That said, I agree that maintainers refusing to allow us adding the perf test suite to a third-party repo puts us in a difficult spot. One option would be a split approach: first-party ODMs have performance tests within their own repos, third-party ODMs have theirs in an `odm-testing-pipeline` repo explicitly for that purpose. Then if maintainers tell us that they'd actually prefer to have the performance tests inside the ODM repo directly, we can migrate that suite out of the `odm-testing-pipeline` repo. 

Do any third-party ODMs already have robust performance testing that we would be competing with? What are the most common reasons we've gotten for pushback against similar work being contributed in the past?

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,387 @@
...
+| Before task | n/a.                                                                                                                                                                                                                                                         |
+| Do task     | For each of the 10,000 `unique_id` values, perform a filter operation to search for the parent LARGE_DOC_NESTED model.                                                                                                                                       |
+| After task  | n/a.                                                                                                                                                                                                                                                         |
+| Teardown    | Drop the collection associated with the LARGE_DOC_NESTED model.                                                                                                                                                                                              |
+
+## Benchmark platform, configuration and environments
```
dariakp, line 363: We call out client and server, but not the underlying platform configuration. Given how much the underlying hardware can impact performance, if we want meaningfully comparable perf results across different ODMs (or between ODMs and drivers), we need to keep as much of the surrounding environment fixed as possible.

source/benchmarking/odm-benchmarking.md
dariakp: If we don't intend to leverage existing in-house performance tooling, how are we intending to track any actual performance regressions? (To address the scope goal: "Identify performance regressions and provide an ongoing, automated mechanism for doing so.") Likewise, how will we track apples to oranges or apples to apples comparison goals? Internally, we have, e.g., dedicated distros (like Bailey mentioned), perf release dashboards that can easily pull the relevant data into one place, etc. It's fine to have tests living in the ODMs own repos so that there is transparency and so that they can be used to inform local development, but I don't see how we can realistically meet the scope goals I mentioned if we don't run them in our own pipeline.

source/benchmarking/odm-benchmarking.md
ajcvickers: I talked to @JamesKovacs and then @NoahStapp offline about my concerns. In a nutshell, there is a lot here, and perf testing is hard and resource (human and otherwise) intensive. It would be my preference to start with one goal, such as testing for ODM perf regressions, and get that working well end-to-end. In terms of the details here (iterations, payloads, and the scenarios themselves) they look fine as starting points, and then we can  modify them as we start getting data based on what we are seeing.

I'm going to say LGTM here, because this feels like a plan we can get started with and then iterate on.

source/benchmarking/odm-benchmarking.md
ajcvickers: With regard to including perf tests in externally owned ODM repos, I think it's good form to let community know what we're doing, and to have them be involved and even host if they want to--this gives us visibility in that community which is great. That being said, I would be very surprised if any community wants to do this. Perf testing is controversial, especially when people are designing and implementing tests for the perf of their own code. So, realistically, I think this all going to end up internal.

NoahStapp: For ODMs where we own the repo, it's easy to run the performance suite on our existing test infrastructure. For external ODMs, unless the maintainers are fine with us adding config to run evergreen tasks directly into the repo, we'll have to set up a separate ODM testing pipeline repo similar to the AI/ML one that already exists.

source/benchmarking/odm-benchmarking.md
NoahStapp: Is calling out a specific evergreen host/distro configuration sufficient for that purpose? The drivers benchmarking spec doesn't call out specifics, so we should choose a standard for both and ensure it's present.

aclark4life (approved):
Agree with @ajcvickers ! I think this is a reasonable framework to get folks started doing _something_ collaboratively across ODMs. We'll see how it goes and update the spec accordingly.

source/benchmarking/odm-benchmarking.md
dariakp: I agree that we should be consistent. For drivers benchmarks, we call it out in our internal "Drivers Performance Testing Infrastructure Guidelines" doc, and it was a requirement added in DRIVERS-2666. We don't capture that in the spec itself because the precise distro names are specific to our internal tooling. We also don't capture driver-specific platform pinning requirements (e.g., the Node.js runtime version for the node driver) because these can vary a lot by driver and also depend on the precise CI setup these drivers use to run the perf benchmarks. But I don't think it's unreasonable to have a general requirement to keep invariant as much of the system in which the perf tests execute as can be reasonably achieved.

source/benchmarking/odm-benchmarking.md
dariakp: Perhaps it's worth adjusting the wording here accordingly, then

Jibola: Circling back here, it looks like the agreement solely for _test placement_ is this:

**First Party ODMs** (libraries we **maintain** e.g. Django, Hibernate, EFCore ...)
1. Are **required** to include performance tests directly to the repository. They need to be in an explicit test directory, but do not need to be run as part of the standard set of CI/CD to confirm functionality.

**External ODMs** (libraries we **contribute to** e.g. Mongoose, Spring Data, Doctrine ...)
1. Make **best effort** attempts are including tests directly to the repository. If not possible, create a separate repository that clones the target repository and then executes the performance benchmark. These tests should still be an clear `tests` directory.

Have I understood that correctly?

cc: @dariakp , @baileympearson , @NoahStapp

source/benchmarking/odm-benchmarking.md
alcaeus: I agree -- let's start by using performance tests to get a first indicator of performance and prevent regressions, then focus on improving performance before starting to compare different ODMs against one another. FWIW, even though we've been testing drivers performance for a long time, we haven't really compared their performance, nor explored how language differences (e.g. compiled vs. interpreted) may favour one driver over another in the performance tests. I don't think it's good to include such a goal in the initial spec.

source/benchmarking/odm-benchmarking.md
NoahStapp: What is meant by "the standard set of CI/CD to confirm functionality"? I would expect the performance tests to be run on the same standard infrastructure we maintain for the ODM.

For external ODMs, there should be only one separate repository for all ODMs, similar to how the AI/ML testing pipeline operates.

source/benchmarking/odm-benchmarking.md
NoahStapp: Agreed that the initial focus of this spec is to prevent regressions and track performance for individual ODMs. However, I don't think that not having compared performance across drivers in the past should prevent us from aiming to compare performance across ODMs in the future. The driver benchmarking spec includes a similar paragraph about comparing across drivers that we may not have ever fully implemented, but for the reasons stated above I feel it's a worthwhile goal.

source/benchmarking/odm-benchmarking.md
Jibola: Sorry, to clarify. I meant they do not need to be run at the same time as the unit and integration tests.

source/benchmarking/odm-benchmarking.md
JamesKovacs: Fair enough.

source/benchmarking/odm-benchmarking.md
```diff
@@ -368,14 +368,19 @@ supports.
 ### Benchmark Server
 
 The MongoDB ODM Performance Benchmark must be run against a MongoDB replica set of size 1 running the latest stable
-database version without authentication or SSL enabled.
+database version without authentication or SSL enabled. The Benchmark should be run on the established internal
```
dariakp: Can we just add a sentence saying that the "latest stable" MongoDB version should be patch-pinned and updated in standalone commits as needed in order to be able to differentiate between performance profile changes due to server vs the ODM code?

source/benchmarking/odm-benchmarking.md
dariakp: I'm good to resolve this thread in favor of a small wording adjustment suggested in [the standalone comment below](https://github.com/mongodb/specifications/pull/1828#discussion_r2519369141) (ETA: addressed)

dariakp (approved):

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,393 @@
...
+
+To avoid various types of measurement skew, tasks will be measured over several iterations. Each iteration will have a
+number of operations performed per iteration that depends on the task being benchmarked. The final score for a task will
+be the median score of the iterations. A range of percentiles will also be recorded for diagnostic analysis.
+
+### Data sets
+
+Data sets will vary by task. In most cases, data sets will be synthetic line-delimited JSON files to be constructed by
+the ODM being benchmarked into the appropriate model. Some tasks will require additional modifications to these
+constructed models, such as adding generated ObjectIds.
```
jmikola: Is this referring to the contents of the `tgz` files in the PR? If so, are these files a single file with one JSON document per line?

Edit: I think the "Benchmark task definitions" section below answers this question. Maybe link to that section from here for context?

It was initially unclear to me if the individual `.json` files in the test directory were meant to be examples, or are actual test fixtures. I don't see any discussion of the `.tgz` files in this document.

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,393 @@
...
+
+### Versioning
+
+The MongoDB ODM Performance Benchmark will have vX.Y versioning. Minor updates and clarifications will increment "Y" and
+should have little impact on score comparison. Major changes, such as task modifications, MongoDB version tested
+against, or hardware used, will increment "X" to indicate that older version scores are unlikely to be comparable.
```
jmikola: Where is this version recorded? Should it be presented in the header of this document (similar to the schema version for the unified test runner)?

I think it'd make sense to record version bumps in the Changelog section as well. You can even start with the first if it'd help demonstrate formatting for future updates (as we do in the test runner spec).

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,393 @@
...
+- before operation -- (ONCE PER ITERATION) something to do before every task iteration, e.g. drop a collection, or
+    reload test data (if the test run modifies it), etc.
+- do operation -- (ONCE PER ITERATION) smallest amount of code necessary to execute the task; e.g. insert 10,000 models
+    one by one into the database, or retrieve 10,000 models of test data from the database, etc.
+- after operation -- (ONCE PER ITERATION) something to do after every task iteration (if necessary)
+- teardown -- (ONCE PER TASK) something done once after all benchmarking is complete (if necessary); e.g. drop the test
+    database
+
+The wall-clock execution time of each "do operation" phase will be recorded. We use wall clock time to model user
+experience and as a lowest-common denominator across ODMs. Iteration timing should be done with a high-resolution
```
jmikola: If this is a spec doc, we may want to use consistent formatting for admonitions. For example, this might be "SHOULD". Elsewhere you MAY want to use "MUST" 😄

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,393 @@
...
+- iterations should loop for at least 30 seconds cumulative execution time
+- once this 30 second minimum execution time is reached, iterations should stop after at least 10 iterations or 1 minute
+    cumulative execution time, whichever is shorter
+
+This balances measurement stability with a timing cap to ensure all tasks can complete in a reasonable time.
+
+For each task, the 10th, 25th, 50th, 75th, 90th, 95th, 98th and 99th percentiles will be recorded using the following
+algorithm:
+
+- Given a 0-indexed array A of N iteration wall clock times
```
jmikola: Some formatting for "A" and "N" would be helpful here for readability. You can simply use monospace formatting or something fancier like we have in the CSFLE spec: https://github.com/mongodb/specifications/blob/master/source/client-side-encryption/client-side-encryption.md#create-encrypted-collection-helper

Likewise for other special symbols below, such as "i" and "p".

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,393 @@
...
+
+### Flat models
+
+Datasets are in the `flat_models` tarball.
+
+Flat model tests focus on flatly-structured model reads and writes across data sizes. They are designed to give insights
+into the efficiency of the ODM's implementation of basic data operations.
+
+The data will be stored as strict JSON with no extended types. These JSON representations must be converted into
+equivalent models as part of each benchmark task.
```
jmikola, line 151: I assume it is the responsibility of each ODM to create a corresponding model for the flat model schema. Should the document shapes (e.g. fields and types) be communicated here, or do you expect folks to infer that by examining the JSON file contents?

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,393 @@
...
+
+Datasets are available in the `odm-data` directory adjacent to this spec.
+
+Note: The term "LDJSON" means "line-delimited JSON", which should be understood to mean a collection of UTF-8 encoded
+JSON documents (without embedded CR or LF characters), separated by a single LF character. (Some Internet definition of
+line-delimited JSON use CRLF delimiters, but this benchmark uses only LF.)
+
+### Flat models
+
+Datasets are in the `flat_models` tarball.
```
jmikola: This refers to the `flat_models.tgz` file, but the sub-sections then refer to specific files (e.g. `small_doc.json`). Do the benchmarks in this section use _both_?

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,393 @@
...
+| ----------- | ------------------------------------------------------------------------------------------------------------------- |
+| Setup       | Load the LARGE_DOC dataset into memory as an ODM-appropriate model object. Save 10,000 instances into the database. |
+| Before task | n/a.                                                                                                                |
+| Do task     | Update the `field1` field for each instance of the model to `updated_value` in an ODM-appropriate manner.           |
+| After task  | Drop the collection associated with the LARGE_DOC model.                                                            |
+| Teardown    | n/a.                                                                                                                |
+
+### Nested models
+
+Datasets are in the `nested_models` tarball.
```
jmikola: Nit: I think it's be clearer to use the complete filename here: `nested_models.tgz`. Likewise for `flat_models.tgz` above.

source/benchmarking/odm-benchmarking.md
```diff
@@ -0,0 +1,393 @@
...
+
+The benchmarks should be run with the most recent stable version of the ODM and the newest version of the driver it
+supports.
+
+### Benchmark Server
+
+The MongoDB ODM Performance Benchmark must be run against a MongoDB replica set of size 1 patch-pinned to the latest
+stable database version without authentication or SSL enabled. This database version should be updated as needed to
+better differentiate between ODM performance changes and server performance changes. The Benchmark should be run on the
+established internal performance distro for the sake of consistency.
```
jmikola, line 375: Should "established internal performance distro" reference another document. Based on @dariakp's resolved comment above, I assume this is referring to the basic driver benchmarks.

source/benchmarking/odm-benchmarking.md
jmikola: Should the dedicated Evergreen hosts be discussed in this document? Alternatively, if it's already discussed in the general driver benchmarks spec we can reference that.

source/benchmarking/odm-benchmarking.md
NoahStapp: Sure, I'll add a reference to the "Benchmark task definitions" section here.

source/benchmarking/odm-benchmarking.md
NoahStapp: Good catch: putting it in the document header + changelog makes sense to me.

source/benchmarking/odm-benchmarking.md
NoahStapp: None of the data types here are complex enough to meaningfully differ between ODMs, so I think expecting they be inferred from the JSON files is reasonable.

source/benchmarking/odm-benchmarking.md
NoahStapp: The `flat_models.tgz` file contains multiple flat json documents referred to by name (such as `small_doc.json`).

source/benchmarking/odm-benchmarking.md
NoahStapp: This is an internal document specifying infrastructure-specific config and details. I would support linking it here, but that is against our current policy for specs.

source/benchmarking/odm-data/flat-models/small_doc.json
jmikola: @Jibola are you asking for a trailing newline here?


```suggestion
{"field1":"miNVpaKW","field2":"CS5VwrwN","field3":"Oq5Csk1w","field4":"ZPm57dhu","field5":"gxUpzIjg","field6":"Smo9whci","field7":"TW34kfzq","field8":55336395,"field9":41992681,"field10":72188733,"field11":46660880,"field12":3527055,"field13":74094448}

```

source/benchmarking/odm-benchmarking.md
jmikola: @NoahStapp let me know if this is also an internal document or something we can easily reference. Feel free to resolve accordingly.

source/benchmarking/odm-data/flat-models/small_doc.json
Jibola: Yep

source/benchmarking/odm-benchmarking.md
JamesKovacs: This anchor needs to be corrected.

source/benchmarking/odm-benchmarking.md
JamesKovacs: Not addressed yet.

source/benchmarking/odm-benchmarking.md
NoahStapp: Missed this, sorry. My main concern with a several MB document is modeling it in a way that is representative of real data manipulated using an ODM. The existing driver benchmark simply uses a document with a massive amount of top-level fields, which is not the kind of structure you would expect for data wrapped by an ODM. Something closer to a very large array of embedded documents could be a better representation, or your suggestion of a large binary field.

I'll try adding a large binary field and see how practical that is.

source/benchmarking/odm-benchmarking.md
NoahStapp: This is in an internal document titled "Drivers Performance Testing Infrastructure Guidelines".

source/benchmarking/odm-benchmarking.md
NoahStapp: I've updated the flat `large_doc.json` to be 3MB using a binary image field.

PR #1831, opened 2025-08-26
DRIVERS-3106: Replace empty UUIDs in localSchema.json spec test

adelinowona:
The current localSchema.json spec test uses empty UUIDs which causes test failures for the C# driver. This PR updates the test data to use non-empty UUIDs.

In C#, UUIDs are represented as GUIDs (value types/structs) with a default value of an empty UUID (00000000-0000-0000-0000-000000000000). Following standard C# conventions, an empty UUID indicates an uninitialized GUID. Therefore, the C# driver correctly treats empty UUID IDs as "not-set" and automatically generates new IDs during insert operations, which is the expected behavior for the language.

After evaluation, we've decided to maintain the C# driver's current behavior as it aligns with C# language conventions. This behavior will be documented in our driver documentation to ensure users are aware of this UUID handling.

We could workaround the issue in our test runner but updating the spec test data is a lot easier. 

Tested in this [patch](https://spruce.mongodb.com/version/68ae263f10c35300077bad9f/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

- [ ] Update changelog.
    clusters).

### DIFF
--- a/source/client-side-encryption/tests/unified/localSchema.json
+++ b/source/client-side-encryption/tests/unified/localSchema.json
@@ -27,7 +27,7 @@
                     "keyId": [
                       {
                         "$binary": {
-                          "base64": "AAAAAAAAAAAAAAAAAAAAAA==",
+                          "base64": "OyQRAeK7QlWMr0E2xWapYg==",
                           "subType": "04"
                         }
                       }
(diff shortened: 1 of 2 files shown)


kevinAlbs (approved):
LGTM. Updated test passes in C driver locally.

PR #1833, opened 2025-09-03
DRIVERS-3262: `appendMetadata()` does not append duplicate metdata

baileympearson:
This PR updates the handshake specification to clarify that drivers MUST NOT append the same metadata multiple times to a MongoClient.

- [x] Update changelog.
    clusters).

### DIFF
--- a/source/mongodb-handshake/handshake.md
+++ b/source/mongodb-handshake/handshake.md
@@ -449,7 +449,7 @@ be appended to their respective fields, and be delimited by a `|` character. For
 Some client libraries provide APIs that accept a pre-initialized MongoClient as an argument. In these circumstances, it
 is possible for multiple library objects to be associated with the same MongoClient, which could result in the same
 metadata being appended multiple times. Drivers MUST ensure that any duplicate `DriverInfoOptions` objects provided to a
-MongoClient or appended to a MongoClient do not result in additional metadata being appended (i.e., a no-op). See
+MongoClient or appended to a MongoClient do not result in additional metadata being appended. See
 [Supporting Wrapping Libraries](#supporting-wrapping-libraries).
 
 **NOTE:** All strings provided as part of the driver info MUST NOT contain the delimiter used for metadata concatention.


source/mongodb-handshake/handshake.md
```diff
@@ -442,6 +445,11 @@ be appended to their respective fields, and be delimited by a `|` character. For
 }
 ```
 
+Some client libraries provide APIs that accept a pre-initialized MongoClient as an argument. In these circumstances, it
+is possible for multiple library objects to be associated with the same MongoClient, which could result in the same
+metadata being appended multiple times. Drivers MUST ensure that a `DriverInfo` object can be appended to a MongoClient
```
dariakp: Since the wording added above is explicitly defining "identical" DriverInfoOptions objects, would it make sense to adjust this sentence to refer to that definition? Perhaps something along the lines of "Drivers MUST ensure that all `DriverInfoObjects` are distinct (see <link to definition>)"?

source/mongodb-handshake/tests/README.md
```diff
@@ -202,8 +202,102 @@ Before each test case, perform the setup.
...
             - If test case's version is non-null: `1.2|<version>`
             - Otherwise, the field remains unchanged: `1.2`
-        - `client.driver.platform`:
+        - `client.platform`:
             - If test case's platform is non-null: `Library Platform|<platform>`
             - Otherwise, the field remains unchanged: `Library Platform`
 
     - All other subfields in the `client` document remain unchanged from `updatedClientMetadata`.
+
+## Test 3: Multiple Successive Metadata Updates with Identical/Partially Identical `DriverInfo`
```
dariakp:
```suggestion
## Test 3: Multiple Successive Metadata Updates with Duplicate Data
```

source/mongodb-handshake/tests/README.md
```diff
@@ -202,8 +202,102 @@ Before each test case, perform the setup.
...
+            - Assert metadata.version is equal to `1.2`
+            - Assert metadata.platform is equal to `LibraryPlatform`
+        - Otherwise:
+            - Assert metadata.name is equal to `library|<name>`
+            - Assert metadata.version is equal to `1.2|<version>`
+            - Assert metadata.platform is equal to `LibraryPlatform|<platform>`
+
+    - All other subfields in the `client` document remain unchanged from `updatedClientMetadata`.
+
+## Test 4: Metadata is not appended if identical to initial metadata
```
dariakp: I think for both this test and full duplication test in the preceding group, it may be worthwhile to add another test where the updates are separated by a non-identical update (to guard against the scenario where the implementation only checks the most recently appended DriverInfoOptions)

dariakp (changes requested):

source/mongodb-handshake/tests/README.md
baileympearson: I've added tests for both of these scenarios

source/mongodb-handshake/handshake.md
```diff
@@ -442,6 +445,11 @@ be appended to their respective fields, and be delimited by a `|` character. For
 }
 ```
 
+Some client libraries provide APIs that accept a pre-initialized MongoClient as an argument. In these circumstances, it
+is possible for multiple library objects to be associated with the same MongoClient, which could result in the same
+metadata being appended multiple times. Drivers MUST ensure that a `DriverInfo` objects provided to a MongoClient or
```
dariakp:
```suggestion
metadata being appended multiple times. Drivers MUST ensure that all `DriverInfoOptions` objects provided to a MongoClient or
```

source/mongodb-handshake/tests/README.md
```diff
@@ -202,8 +202,193 @@ Before each test case, perform the setup.
         - `client.driver.version`:
             - If test case's version is non-null: `1.2|<version>`
             - Otherwise, the field remains unchanged: `1.2`
-        - `client.driver.platform`:
```
dariakp, line 205: from async conversation - we want to update this everywhere

dariakp (changes requested):
Thanks! Just a couple of small things here; will approve once cross-referenced with Node implementation

source/mongodb-handshake/handshake.md
```diff
@@ -405,6 +405,9 @@ class DriverInfoOptions {
 }
 ```
 
+Two `DriverInfoOptions` objects are considered identical if all fields are strictly equal with case sensitive string
```
dariakp: It just occurred to me that we may want to adjust this wording further to account for the null cases: perhaps, "if all non-empty fields". But the testing will likely want to be deferred to DRIVERS-3251 - thoughts?

dariakp (changes requested):

source/mongodb-handshake/handshake.md
dariakp: From async conversation, expanded wording to: "Two DriverInfoOptions objects are considered equal if they would result in the same metadata in the client handshake.  In practice, this means if a field is the empty string (""), treat it as unset.  Assert that each field is strictly equal with case sensitive string comparison." Planning to add test cases, but only assert on the resulting metadata for the non-empty fields.

source/mongodb-handshake/handshake.md
```diff
@@ -405,6 +405,10 @@ class DriverInfoOptions {
 }
 ```
 
+Two `DriverInfoOptions` objects are considered equal if they would result in the same metadata in the client handshake.\
```
addaleax:
```suggestion
Two `DriverInfoOptions` objects are considered equal if they would result in the same metadata in the client handshake.
```

Rogue backslash?

source/mongodb-handshake/tests/README.md
```diff
@@ -202,8 +202,288 @@ Before each test case, perform the setup.
...
+        - `client.platform`:
             - If test case's platform is non-null: `Library Platform|<platform>`
             - Otherwise, the field remains unchanged: `Library Platform`
 
     - All other subfields in the `client` document remain unchanged from `updatedClientMetadata`.
+
+## Test 3: Multiple Successive Metadata Updates with Duplicate Data
```
prestonvasquez: The headers for tests 2-7 are misaligned

Screenshot 2025-09-09 at 10 13 05 PM

source/mongodb-handshake/handshake.md
baileympearson: Done

source/mongodb-handshake/handshake.md
baileympearson: 🤔 hmm.  Fixed

source/mongodb-handshake/tests/README.md
baileympearson: fixed

source/mongodb-handshake/tests/README.md
baileympearson: Done

source/mongodb-handshake/handshake.md
baileympearson: Fixed

source/mongodb-handshake/tests/README.md
```diff
@@ -202,8 +202,288 @@ Before each test case, perform the setup.
...
+| Case | Name      | Version | Platform           |
+| ---- | --------- | ------- | ------------------ |
+| 1    | library   | 1.2     | Library Platform   |
+| 2    | framework | 1.2     | Library Platform   |
+| 3    | library   | 2.0     | Library Platform   |
+| 4    | library   | 1.2     | Framework Platform |
+| 5    | framework | 2.0     | Library Platform   |
+| 6    | framework | 1.2     | Framework Platform |
+| 7    | library   | 2.0     | Framework Platform |
+| 8    | framework | 2.0     | Framework Platform |
```
prestonvasquez: Isn't test 3.8 the same as 2.1?

source/mongodb-handshake/handshake.md
```diff
@@ -405,6 +405,10 @@ class DriverInfoOptions {
 }
 ```
 
+Two `DriverInfoOptions` objects are considered equal if they would result in the same metadata in the client handshake.
+In practice, this means if a field is the empty string (`""`), treat it as unset. Assert that each field is strictly
+equal with case sensitive string comparison.
```
prestonvasquez, line 410: This seems to contradict a non-goal of the original [scope](https://docs.google.com/document/d/1G0B4ygLIwWEUbjt4wPD_fWxv-LJtjDBDJKJYeSIFb4Q/edit?tab=t.0), presumably because it's difficult: 

> Each client metadata update will be treated as a new entry in telemetry, meaning that driver and framework usage may be counted multiple times

Consider this example:

```go
{ // Starting point for our metadata
	Name:     "X|Y|Z",
	Version:  "A|B",
	Platform: "C",
}
```

Would the following be an exhaustive list of duplicates?

```go
{Name: "X", Version: "A", Platform: "C"}
{Name: "Y", Version: "B", Platform: ""}
{Name: "Z", Version: "", Platform: ""}
```

Or would these also be duplicates?

```go
{Name: "Y", Version: "B", Platform: "C"}
{Name: "Z", Version: "B", Platform: "C"}
{Name: "Z", Version: "B", Platform: ""}
{Name: "Z", Version: "", Platform: "C"}
```

And if we appended the following: `{Name: "X", Version: "B", Platform: "C"}`, would the metadata be updated to:

```go
{
	Name:     "X|Y|Z|X",
	Version:  "A|B|B",
	Platform: "C|C",
}
```

To avoid all the confusion, could this:

>  Two `DriverInfoOptions` objects are considered equal if they would result in the same metadata in the client handshake.
In practice, this means if a field is the empty string (`""`), treat it as unset. Assert that each field is strictly
equal with case sensitive string comparison.

Be simplified with this:

> Successive calls to `appendMetadata` with the same `DriverInfoOptions` MUST be a no-op, regardless of order.

prestonvasquez: Also is this what is meant by treating empty strings as unset?

```go
{Name: "X", Version: "A", Platform: "C"} // Wrapper 1
{Name: "Y", Version: "B", Platform: ""}  // Wrapper 2
{Name: "Z", Version: "", Platform: ""}   // Wrapper 3
```
You Get: 
```go
{
	Name:     "X|Y|Z",
	Version:  "A|B|",
	Platform: "C||",
}
```

dariakp: @prestonvasquez [DRIVERS-3251](https://jira.mongodb.org/browse/DRIVERS-3251) addresses the version to name mapping for the metadata. I think your wording works, but we wanted to clarify that unset and empty string should be treated the same for the purposes of determining whether two driver info options objects are considered equal.

dariakp: Perhaps, "Successive calls to appendMetadata with equivalent DriverInfoOptions MUST be a no-op, even when nonconsecutive."

prestonvasquez: @dariakp Sounds good to me. I prefer the "successive" phrasing because it makes it more clear that a driver will have to maintain some state regarding previous input to appendMetadata. Especially since without DRIVERS-3251 it would be impossible to parse driverInfo.* for duplication.

baileympearson: @prestonvasquez The current phrasing is:

> Drivers MUST ensure that all `DriverInfo` objects provided to a MongoClient or
appended to a MongoClient are distinct (see [Supporting Wrapping Libraries](#supporting-wrapping-libraries)).

The phrasing "all `DriverInfo` objects provided to a MongoClient or appended to a MongoClient are distinct" implies that drivers must do something to keep track of the driverInfo objects appended to a MongoClient.

I can change the phrasing to use "successive" if you would prefer though.

source/mongodb-handshake/tests/README.md
baileympearson: Yeah.  I wasn't sure if it would be preferable to _not_ have the duplicate test (in which case, people could ask "why isn't the last test case included in this matrix?") or to have the duplicate test (in which case, people could ask "why do we have a duplicate test?").

I decided to add the duplicate test because I found it clearer, and because table-tests are cheap and easy in Node so the overhead was minimal.  But I'm happy to revisit if you'd prefer

source/mongodb-handshake/handshake.md
```diff
@@ -442,6 +446,11 @@ be appended to their respective fields, and be delimited by a `|` character. For
 }
 ```
 
+Some client libraries provide APIs that accept a pre-initialized MongoClient as an argument. In these circumstances, it
+is possible for multiple library objects to be associated with the same MongoClient, which could result in the same
+metadata being appended multiple times. Drivers MUST ensure that all `DriverInfo` objects provided to a MongoClient or
```
dariakp:
```suggestion
metadata being appended multiple times. Drivers MUST ensure that all `DriverInfoOptions` objects provided to a MongoClient or
```

source/mongodb-handshake/tests/README.md
dariakp: I think Preston has a point - the title of this set of cases is "Multiple Successive Metadata Updates with *Duplicate* Data"... Having an entry that has no duplicate data in it is/could be confusing.

dariakp (changes requested):
LGTM, pending negotiated wording in the spec text and the removal of the redundant test case from the duplicate metadata set in Test 3

source/mongodb-handshake/handshake.md
prestonvasquez: @baileympearson @dariakp I agree that it's fine either way. I prefer the no-op wording because it specifies concrete, observable behavior for duplicates and keeps the API idempotent:

> Repeated calls to appendMetadata with equivalent DriverInfoOptions MUST be a no-op.

```
if seen[input] {
	return
}
```

(I don't think we need the "even when nonconsecutive" part)

I also think the qualifier for empty strings is unnecessary. Happy to defer.

baileympearson: Okay, I've adjusted the phrasing to use no-op.  Lmk what you think

I want to keep the empty string callout because, while unlikely to happen in practice, it is an edge case and I'd prefer to codify it in the spec than rely on drivers catching it and testing it themselves.

source/mongodb-handshake/tests/README.md
baileympearson: Done

source/mongodb-handshake/handshake.md
dariakp: The non-consecutive callout was to remove the ambiguity of "successive", since one of the common meanings of that word is synonymous with "consecutive". @prestonvasquez (Bailey's current wording doesn't have it, just addressing your earlier point)

source/mongodb-handshake/handshake.md
```diff
@@ -442,6 +446,12 @@ be appended to their respective fields, and be delimited by a `|` character. For
 }
 ```
 
+Some client libraries provide APIs that accept a pre-initialized MongoClient as an argument. In these circumstances, it
+is possible for multiple library objects to be associated with the same MongoClient, which could result in the same
+metadata being appended multiple times. Drivers MUST ensure that any duplicate `DriverInfoOptions` objects provided to a
```
dariakp, line 451: If we are using no-op language, it should be applied to the `appendMetadata`, so I recommend rephrasing the `Drivers MUST ensure that any duplicate ...` sentence to read: `Drivers MUST ensure that any repeated calls to appendMetadata with equivalent DriverInfoOptions are a no-op` (and, optionally, "resulting in no additional metadata being appended")

dariakp (changes requested):

source/mongodb-handshake/handshake.md
prestonvasquez: @baileympearson , @dariakp Has node considered returning a sentinel error or a boolean for `appendMetadata`. In the no-op case, should the user be notified that what they did was not honored by the driver?

baileympearson: All an error or value would indicate is whether or not that value had been appended previously, it doesn't indicate a problem or error.  Do we expect users to care about this information?  I don't think so.

But additionally, what would we expect users to do with this information?  If the driver indicates the metadata has already been appended, there isn't an action for a user to take.  So imo there isn't a reason to provide this information to users.

prestonvasquez: @baileympearson The appendMetadata method in Go will be a wrapper for [atomic.Pointer](https://pkg.go.dev/sync/atomic#Pointer) and [sync.Map.LoadOrStore](https://pkg.go.dev/sync#Map.LoadOrStore). The latter returns status about wether the object will be appended or not:

```
func (m *Map) LoadOrStore(key, value any) (actual any, loaded bool) 
```

Not returning a bool is "lossy" / asymmetric in this context:

```go
func (c *Client) AppendDriverInfo(info options.DriverInfo) {
	if _, loaded := c.seenDriverInfo.LoadOrStore(info, struct{}{}); loaded {
		// This is information that we have that the user doesn't 
		return
	}

	// ...
}
```

Here's a possible use case:

```go
if wasAdded := client.AppendMetadata(info); wasAdded {
	log.Printf("Registered %s %s with MongoDB Go Driver\n", info.Name, info.Version)
}
```

Without this update, Go Driver will try to opt to put the `appendMetadata` logic in an experimental API and defer deciding on the stable signature for now. I can make a follow up ticket if that helps move things along here.

CC: @matthewdale

baileympearson: I would suggest a follow up ticket, if you'd like to pursue this, I'm not going to change the API signature of this method in this PR unless there is large consensus that we want to move in that direction because:

1. That is likely breaking in compiled drivers which have already implemented this feature
2. I still do not think communicating information about whether or not metadata was appended is valuable.  Unlike atomic map operations, where the user likely would want to act differently depending on whether the update appended or not, there's nothing actionable for users to do in this scenario.  Also - even if we did want to communicate this to users, it feels sort of moot because client metadata is already a "best effort" feature.  Even successfully appending driver metadata does not guarantee that the appended metadata actually is included in the handshake because of the handshake size limits.

fwiw I still consider your logging example to be "correct" even if nothing is appended because the observable outcome of the operation for the user is that their DriverInfoOptions are included in the handshake metadata.  That is true regardless of whether or not `appendMetadata()` is a no-op because they're appending a duplicate object or not:

```go
client.AppendMetadata(info);
log.Printf("Registered %s %s with MongoDB Go Driver\n", info.Name, info.Version)
```

prestonvasquez: > Even successfully appending driver metadata does not guarantee that the appended metadata actually is included in the handshake because of the handshake size limits.

This method should also return false if we can't append because of size limitations, IMO.

dariakp: I would definitely recommend a follow up ticket if we want to pursue this further, appending partial metadata is a larger discussion: a boolean value would not communicate the full extent of the relevant information to the user, since the metadata could be appended fully, partially, or not at all. And, to echo Bailey's point, it's unclear what the user/framework developer would do with this information if they did max out the appendable data: the prioritization of metadata is at the discretion of the driver, conforming to the common spec; and there is no mechanism for removing metadata from an existing set.

PR #1834, opened 2025-09-05
DRIVERS-3282: Introduce optin setting to await for MinPoolSize population

sanych-sun:
Some CSOT tests are require pre-heat connection pool to have a predictable test outcome. The suggestion is to use `minPoolSize` together with new setting `awaitMinPoolSize` that makes unified test runner to await until pool will be populated before run the actual test case.

- [x] Update changelog.
    clusters).


source/client-side-operations-timeout/tests/global-timeoutMS.yml
```diff
@@ -2,7 +2,7 @@
 
 description: "timeoutMS can be configured on a MongoClient"
 
-schemaVersion: "1.9"
+schemaVersion: "1.26"
```
sanych-sun, line 5: @prestonvasquez I've updated this file only with the new approach to pre-heat the pool. Should we apply similar settings to more tests?

prestonvasquez: Suggest updating the runCursorCommand "Non-tailable cursor lifetime remaining timeoutMS applied to getMore if timeoutMode is unset" test, which would close DRIVERS-3134 (PR https://github.com/mongodb/specifications/pull/1769/files). 

Otherwise, I will collect a list of candidates during Go Driver's second implementation.

sanych-sun: I'll update the test and list of candidates would be much appreciated.

source/unified-test-format/unified-test-format.md
```diff
@@ -542,6 +542,12 @@ The structure of this object is as follows:
         - `bypassQueryAnalysis`: Optional. Disables analysis of outgoing commands. Defaults to `false`.
         - `keyExpirationMS`: The same as in [`clientEncryption`](#entity_clientEncryption).
 
+    <span id="entity_client_awaitMinPoolSize"></span>
+
+    - `awaitMinPoolSize`: Optional boolean. If `true`, the unified spec runner must wait for the connection pool to be
+        populated for all servers according to the `minPoolSize` option. If `false`, not specified, or if minPoolSize
+        equals 0, there is no need to wait for any specific pool state.
```
prestonvasquez: Drivers that asynchronously construct the pool size with minPoolSize after client construction could make this check block indefinitely. We should prescribe an appropriate amount of time to await warming the connection, perhaps 500ms? From experimentation, actual connection time seems to be around 50-200ms on local networks.

source/client-side-operations-timeout/tests/global-timeoutMS.yml
prestonvasquez: I think we should add this functionality to any CSOT test that (1) doesn't require threads,(2) relies on a failpoint, and (3) uses `timeoutMS` either as an operation argument or on the client. The following seem relevant:

- retryability-timeoutMS.yml 
- command-execution.jyml
- non-tailable-cursors.yml
- error-transformations.yml
- convenient-transactions.yml
- sessions-inherit-timeoutMS.yml

prestonvasquez (changes requested):

source/unified-test-format/unified-test-format.md
sanych-sun: Let's put limit to 1 second, we can tune it in future if needed. In case if this is not enough, should we throw and report test as failed?

prestonvasquez: Sounds good to me, thank you

baileympearson: Is there a downside to setting a higher limit?  Worst case, the tests takes a bit longer to fail.  Best case, we avoid some flakiness and modifications of these tests if 1s proves to be too short.

prestonvasquez: We could give a recommended timeoutMS based on the 20-500ms observation for 8.1 and let drivers update as needed. Does that seem reasonable? The important distinction is that we don't block indefinitely listening to the event monitor.

sanych-sun: I suppose there is no downsides, it simply should be some reasonable limit. In my proof of concept I've used 10 seconds limit, and never hit it. 1 seconds sounds as more then enough to establish 1 connection, but we might want to tune this timeout if there will be tests that need to prepopulate pool with 10 or more connections.

baileympearson: I don't feel strongly but I do think that if we're debating "is the timeout high enough" and there aren't downsides of setting it higher, we should just set it higher now to avoid test updates and test syncs in the future if the timeout value doesn't end up being high enough.

sanych-sun: To make it resilient we might want to use formula like: `waitLimit = minPoolSize * 10s`. This way we will automatically extent the limit if there should be more connection to establish.
@prestonvasquez @baileympearson WDYT? Should I add this to spec?

source/client-side-operations-timeout/tests/global-timeoutMS.yml
sanych-sun: Done.

source/unified-test-format/unified-test-format.md
sanych-sun: @baileympearson I've updated text to wait for `minPoolSize * 10s`. I'm positive if connection could not be established during 10seconds - then we really have a problem, and there is no sense to wait any more. Does it sounds as reliable solution?

baileympearson: Works for me.  10s seems plenty.

matthewdale: We could specify the value in the spec test. For example, instead of an `awaitMinPoolSize` boolean, we could add an `awaitMinPoolSizeMS` integer.

E.g. to wait for 10 seconds:
```yml
awaitMinPoolSizeMS: 10000
```

source/unified-test-format/unified-test-format.md
```diff
@@ -542,6 +542,15 @@ The structure of this object is as follows:
...
         - `keyExpirationMS`: The same as in [`clientEncryption`](#entity_clientEncryption).
 
+    <span id="entity_client_awaitMinPoolSizeMS"></span>
+
+    - `awaitMinPoolSizeMS`: Optional, integer. When specified, this parameter defines the maximum duration (in
+        milliseconds) that the Unified Spec Runner must wait for each connection pool to be populated with `minPoolSize`
+        connections. Only connection pools for data-bearing nodes should be considered. If any connection pool is not
+        populated within the specified timeframe, the Unified Spec Runner must raise an error and mark the test as failed.
+        If the parameter is omitted or if `minPoolSize` is set to 0, no waiting is required for a specific pool state
+        prior to test execution.
```
ShaneHarvey: Do we need to mention the behavior of CMAP and SDAM events/logs that occur before the pool is populated? Is the intention that those events will show up or will they be ignored?

sanych-sun: Ignoring of any events mentioned in the changes.
Thanks

source/unified-test-format/unified-test-format.md
```diff
@@ -542,6 +542,16 @@ The structure of this object is as follows:
         - `bypassQueryAnalysis`: Optional. Disables analysis of outgoing commands. Defaults to `false`.
         - `keyExpirationMS`: The same as in [`clientEncryption`](#entity_clientEncryption).
 
+    <span id="entity_client_awaitMinPoolSizeMS"></span>
+
+    - `awaitMinPoolSizeMS`: Optional, integer. When specified, this parameter defines the maximum duration (in
+        milliseconds) that the Unified Spec Runner must wait for each connection pool to be populated with `minPoolSize`
```
jmikola: I don't recall ever using this term as a proper noun. I think you can just write "test runner" for consistency with other text.

Likewise on the line below.

source/unified-test-format/unified-test-format.md
```diff
@@ -542,6 +542,16 @@ The structure of this object is as follows:
...
+        connections. Only connection pools for data-bearing nodes should be considered. If any connection pool is not
+        populated within the specified timeframe, the Unified Spec Runner must raise an error and mark the test as failed.
+        Any CMAP and SDAM events/logs listeners configured on the client should ignore any events that occur before the
+        pool is being populated. If the parameter is omitted or if `minPoolSize` is set to 0, no waiting is required for a
```
jmikola: This can clarify that you're referring to the `minPoolSize` URI option; lest readers think it's just another client entity option.

PR #1835, opened 2025-09-08
clarify change stream resume behavior

abr-egn:
This updates the description of resume logic to clarify the scenario where consecutive resume `aggregate`s succeed but `getMore`s fail.

- [✔️] Update changelog.
- [n/a] Test changes in at least one language driver.
- [n/a] Test these changes against all server versions and topologies (including standalone, replica set, and sharded
    clusters).


tadjik1 (approved):
Thanks for clarification, @abr-egn!

PR #1836, opened 2025-09-08
DRIVERS-2295 fix typo in `commandName`

kevinAlbs:
Follow-up to https://github.com/mongodb/specifications/issues/1784. Replace `collMod` with `createIndexes` so `commandName` and `command` match.

- [ ] Update changelog.
- [x] Test changes in at least one language driver. (Tested locally in C driver)
    clusters).


blink1073:
I've tested all of these changes with the Python driver

kevinAlbs (commented):
Thanks for the added fixes. Tested updateOne.json in the C driver, but not timeoutMS.json (since C does not-yet implement CSOT).

PR #1838, opened 2025-09-09
DRIVERS-3207: Custom AWS credential providers execute earlier

durran:
Updates the AWS auth spec to require drivers that implement custom AWS credential providers to use the custom provider before any other method.

Updates test wording to include the extra scenario for testing a custom provider when explicit credentials are also provided.

Node implementation: https://github.com/mongodb/node-mongodb-native/pull/4656

- [x] Update changelog.
    clusters).

### DIFF
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -2161,6 +2161,8 @@ practice to avoid this. (See
 
 ## Changelog
 
+- 2025-09-10: Update precedence of MONGODB-AWS credential fetching behaviour.
+
 - 2025-01-29: Add support for custom AWS credential providers.
 
 - 2024-10-02: Add Kubernetes built-in OIDC provider integration.
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -961,8 +961,9 @@ Examples are provided below.
 
     - AWS_CREDENTIAL_PROVIDER
 
-        Drivers MAY allow the user to specify a custom credential provider object or function. See
-        [Custom Credential Providers](#custom-credential-providers)
+        An AWS [Custom Credential Provider](#custom-credential-providers) that returns AWS credentials. Drivers MAY
+        allow the user to specify an object or function, depending on what is idiomatic for the driver. This property
+        MUST follow the same API as the driver language's AWS SDK credential provider.
 
 #### Obtaining Credentials
 
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -1010,8 +1010,8 @@ Drivers MAY expose API for default providers for the following scenarios when ap
 
 The order in which Drivers MUST search for credentials is:
 
-1. A custom AWS credential provider if the driver supports it.
-2. The URI
+1. The URI
+2. A custom AWS credential provider if the driver supports it.
 3. Environment variables
 4. Using `AssumeRoleWithWebIdentity` if `AWS_WEB_IDENTITY_TOKEN_FILE` and `AWS_ROLE_ARN` are set.
 5. The ECS endpoint if `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` is set. Otherwise, the EC2 endpoint.
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -959,6 +959,10 @@ Examples are provided below.
 
         Drivers MUST allow the user to specify an AWS session token for authentication with temporary credentials.
 
+    - AWS_CREDENTIAL_PROVIDER
+
+        Drivers MAY allow the user to specify a custom credential provider object or function.
+
 #### Obtaining Credentials
 
 Drivers will need AWS IAM credentials (an access key, a secret access key and optionally a session token) to complete
@@ -1005,9 +1009,9 @@ Drivers MAY expose API for default providers for the following scenarios when ap
 
 The order in which Drivers MUST search for credentials is:
 
-1. The URI
-2. Environment variables
-3. A custom AWS credential provider if the driver supports it.
+1. A custom AWS credential provider if the driver supports it.
+2. The URI
+3. Environment variables
 4. Using `AssumeRoleWithWebIdentity` if `AWS_WEB_IDENTITY_TOKEN_FILE` and `AWS_ROLE_ARN` are set.
 5. The ECS endpoint if `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` is set. Otherwise, the EC2 endpoint.
 
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -962,6 +962,7 @@ Examples are provided below.
     - AWS_CREDENTIAL_PROVIDER
 
         Drivers MAY allow the user to specify a custom credential provider object or function.
+        See [Custom Credential Providers](https://github.com/mongodb/specifications/blob/master/source/auth/auth.md#custom-credential-providers)
 
 #### Obtaining Credentials
 
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -961,8 +961,8 @@ Examples are provided below.
 
     - AWS_CREDENTIAL_PROVIDER
 
-        Drivers MAY allow the user to specify a custom credential provider object or function.
-        See [Custom Credential Providers](https://github.com/mongodb/specifications/blob/master/source/auth/auth.md#custom-credential-providers)
+        Drivers MAY allow the user to specify a custom credential provider object or function. See
+        [Custom Credential Providers](https://github.com/mongodb/specifications/blob/master/source/auth/auth.md#custom-credential-providers)
 
 #### Obtaining Credentials
 
--- a/source/auth/auth.md
+++ b/source/auth/auth.md
@@ -961,9 +961,9 @@ Examples are provided below.
 
     - AWS_CREDENTIAL_PROVIDER
 
-        An AWS [Custom Credential Provider](#custom-credential-providers) that returns AWS credentials. Drivers MAY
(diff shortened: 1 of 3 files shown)


source/auth/tests/mongodb-aws.md
```diff
@@ -21,15 +21,17 @@ SecretAccessKey=wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY
...
-If the driver supports user provided custom AWS credential providers, then the driver MUST also test the above scenarios
-2-6 with a user provided `AWS_CREDENTIAL_PROVIDER` auth mechanism property. This value MUST be the default credential
-provider from the AWS SDK. If the default provider does not cover all scenarios above, those not covered MAY be skipped.
-In these tests the driver MUST also assert that the user provided credential provider was called at least once in each
-test.
-
-If the driver supports a custom AWS credential provider, it MUST verify the custom provider was used when testing. This
-may be via a custom function or object that wraps the calls to the custom provider and asserts that it was called at
-least once.
+## Testing custom credential providers
```
baileympearson, line 24: Do we have any tests that test the precedence that drivers use when fetching credentials?a

source/auth/auth.md
```diff
@@ -959,6 +959,10 @@ Examples are provided below.
 
         Drivers MUST allow the user to specify an AWS session token for authentication with temporary credentials.
 
+    - AWS_CREDENTIAL_PROVIDER
```
baileympearson, line 962: Was this just a miss from the original PR?

source/auth/auth.md
```diff
@@ -1005,9 +1009,9 @@ Drivers MAY expose API for default providers for the following scenarios when ap
 
 The order in which Drivers MUST search for credentials is:
 
-1. The URI
-2. Environment variables
-3. A custom AWS credential provider if the driver supports it.
+1. A custom AWS credential provider if the driver supports it.
+2. The URI
+3. Environment variables
```
baileympearson: I would expect that the URI takes precedence over any dynamic credential loading mechanism and that the custom credential provider should be the first place we look for credentials, if they're not explicitly provided.

Thoughts?

baileympearson (dismissed):

source/auth/auth.md
dariakp: It looks like the original custom credentials spec wording was intending to specify it in a separate section; I think it's useful to have it listed here, too, but perhaps it's also worth linking to the `Custom Credential Providers` header below.

source/auth/auth.md
dariakp: I think that is the fundamental question posed in this spec change (the DRIVERS ticket AC leaves this open ended) - it may be helpful to have this conversation in slack with the auth spec owners;

notably, for encryption, the URI isn't relevant, but this spec change as a whole would affect automatic credential fetching - should we add a test there, too? (e.g., see Case 15 and 26)

source/auth/auth.md
```diff
@@ -1005,9 +1009,9 @@ Drivers MAY expose API for default providers for the following scenarios when ap
...
 4. Using `AssumeRoleWithWebIdentity` if `AWS_WEB_IDENTITY_TOKEN_FILE` and `AWS_ROLE_ARN` are set.
 5. The ECS endpoint if `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` is set. Otherwise, the EC2 endpoint.
```
dariakp, line 1019: Let's make sure to remember to update the changelog

source/auth/auth.md
durran: Yes it was just missing.

source/auth/tests/mongodb-aws.md
durran: We do not, as our drivers tools scripts only set up one type of credential per test, not multiples.

source/auth/auth.md
durran: I'm not too concerned about it here, since we won't even look at the URI after DRIVERS-3131 is implemented. This is all assuming in this case that the auth mechanism is MONGODB-AWS - I still think even before DRIVERS-3131 if a custom provider is supplied, that it should take preference over all, since it's provided as an option to the client and not a URI option, and we've set the standard that programmatic options always take preference over URI options.

source/auth/auth.md
durran: Updated.

source/auth/auth.md
baileympearson: We do have a standard that equivalent MongoClient options take precedence over URI options, but we don't have precedent for a MongoClient option to take precedence over a related but different URI option.  The current behavior is such that if credentials are provided in the URI, we do not attempt to fetch credentials.  While it is true that Node will drop support for explicitly specified credentials, other drivers might not (they need to wait for a major release) and so we do need to consider explicitly provided credentials here.

My thought is that we should preserve the existing behavior: explicitly provided credentials first, then attempt to fetch credentials.  If fetching credentials, use the custom provider first.

durran: Ok I've moved URI options back to #1 now.

source/auth/auth.md
```diff
@@ -959,6 +959,11 @@ Examples are provided below.
 
         Drivers MUST allow the user to specify an AWS session token for authentication with temporary credentials.
 
+    - AWS_CREDENTIAL_PROVIDER
+
+        Drivers MAY allow the user to specify a custom credential provider object or function. See
+        [Custom Credential Providers](https://github.com/mongodb/specifications/blob/master/source/auth/auth.md#custom-credential-providers)
```
dariakp: Can we make this a relative anchor tag link similar to `[MONGODB-OIDC](#mongodb-oidc)` in the Reauthentication section?

source/auth/auth.md
dariakp: @durran Good point about test 15 - thanks for adding a case to test 26!

dariakp (changes requested):

blink1073 (approved):

source/auth/auth.md
durran: Updated.

source/auth/auth.md
```diff
@@ -959,6 +959,11 @@ Examples are provided below.
 
         Drivers MUST allow the user to specify an AWS session token for authentication with temporary credentials.
 
+    - AWS_CREDENTIAL_PROVIDER
+
+        Drivers MAY allow the user to specify a custom credential provider object or function. See
+        [Custom Credential Providers](#custom-credential-providers)
```
katcharov: Would this be similar to https://github.com/mongodb/specifications/blob/3dc84d1558bf7ff247fb775a64a47e09dd4e657f/source/auth/auth.md?plain=1#L1298-L1303 ? If so, consider using similar wording.

durran: I've updated to use similar wording to the OIDC section.

blink1073 (approved):

source/auth/tests/mongodb-aws.md
```diff
@@ -21,15 +21,17 @@ SecretAccessKey=wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY
...
+
+If the driver supports custom AWS credential providers, the driver MUST test the following:
+
+Scenarios 1-6 from the previous section with a user provided `AWS_CREDENTIAL_PROVIDER` auth mechanism property. This
+value MAY be the default credential provider from the AWS SDK. If the default provider does not cover all scenarios
```
kevinAlbs:
```suggestion
credentials MAY be obtained from the default credential provider from the AWS SDK. If the default provider does not cover all scenarios
```

I expect a test would need to wrap the default credential provider to track if the provider was called.

source/auth/tests/mongodb-aws.md
```diff
@@ -21,15 +21,17 @@ SecretAccessKey=wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY
...
+above, those not covered MAY be skipped. In these tests the driver MUST also assert that the user provided credential
+provider was called in each test. This may be via a custom function or object that wraps the calls to the custom
+provider and asserts that it was called at least once. For test scenarios where the drivers tools scripts put the
+credentials in the MONGODB_URI, drivers MAY extract the credentials from the URI and return the AWS credentials directly
+from the custom provider instead of using the AWS SDK default provider.
```
kevinAlbs, line 59: Suggest adding test cases that would fail without this change: test the custom provider is chosen after URI and before environment variables: 

```
### Case 1: Works
(... happy-path tests of scenarios 1-6 above ...)

### Case 2: Chosen after URI

Create a `MongoClient` configured with AWS auth and credentials in the URI. Example: `mongodb://<AccessKeyId>:<SecretAccessKey>@localhost:27017/?authMechanism=MONGODB-AWS`

Configure a custom credential provider to pass valid AWS credentials. The provider must track if it was called.

Expect authentication to succeed and the custom credential provider was *not* called.

### Case 2: Chosen before environment

Run this test in an environment with AWS credentials configured as environment variables (e.g. `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_SESSION_TOKEN`)

Create a `MongoClient` configured to use AWS auth. Example: `mongodb://localhost:27017/?authMechanism=MONGODB-AWS`. 

Configure a custom credential provider to pass valid AWS credentials. The provider must track if it was called.

Expect authentication to succeed and the custom credential provider was called.

```

durran: Updated per suggestion.

katcharov (approved):

PR #1843, opened 2025-09-23
DRIVERS-2782: Expose snapshotTime

papafe:
C# driver implementation POC: https://github.com/mongodb/mongo-csharp-driver/pull/1777

TODO:
- [x] Update changelog.
    clusters).
- [x] Add unified tests.
- [x] Add prose test for validation of input. 
- [x] Add information about the new `getSnapshotTime` operation (for unified tests).
- [x] Add information about the new `snapshotTime` parameter of `sessionOptions` and that it must be retrieved from the entity map (for unified tests). 
- [x] Check that re-doing the query operation is best course of action for unified tests to verify that the snapshotTime does not get modified again.

### DIFF
--- a/source/unified-test-format/unified-test-format.md
+++ b/source/unified-test-format/unified-test-format.md
@@ -3467,6 +3467,8 @@ other specs *and* collating spec changes developed in parallel or during the sam
 
 ## Changelog
 
+- 2025-10-22: Added `getSnapshotTime` operation on sessions.
+
 - 2025-09-17: **Schema version 1.27.**
 
     Add `observeTracingMessages` configuration for clients and `expectTracingMessages` for test expectations. This allows


source/sessions/snapshot-sessions.md
```diff
@@ -123,10 +150,12 @@ MUST raise an error.
...
 There are no new server commands related to snapshot reads. Instead, snapshot reads are implemented by:
 
-1. Saving the `atClusterTime` returned by 5.0+ servers for the first find/aggregate/distinct operation in a private
-    `snapshotTime` property of the `ClientSession` object. Drivers MUST save `atClusterTime` in the `ClientSession`
-    object.
-2. Passing that `snapshotTime` in the `atClusterTime` field of the `readConcern` field for subsequent snapshot read
+1. If `snapshotTime` is specified in `SessionOptions`, saving the value in a `snapshotTime` property of the
+    `ClientSession`.
+2. If `snapshotTime` is not specified in `SessionOptions`, saving the `atClusterTime` returned by 5.0+ servers for the
+    first find/aggregate/distinct operation in a private `snapshotTime` property of the `ClientSession` object. Drivers
```
sanych-sun: It's not a private property anymore, right?

source/sessions/snapshot-sessions.md
```diff
@@ -89,13 +93,15 @@ MongoCollection API changes.
...
     Optional<bool> snapshot;
+    Optional<BsonTimestamp> snapshotTime;
 
     // other options defined by other specs
 }
 ```
 
-In order to support snapshot reads a new property named `snapshot` is added to `SessionOptions`. Applications set
-`snapshot` when starting a client session to indicate whether they want snapshot reads. All read operations performed
+In order to support snapshot reads two new properties called `snapshot` and `snapshotTime` are added to
```
sanych-sun: I would avoid term "new", as that properties are new just for a little bit, but we are not going to change the spec just to mark them as "not new" :blush: .

source/sessions/snapshot-sessions.md
papafe: Definitely.

source/sessions/snapshot-sessions.md
papafe: I agree with that 😁 
I was trying to be consistent with the previous phrasing, but I'll fix it.

source/sessions/tests/snapshot-sessions.yml
```diff
@@ -309,6 +309,7 @@ tests:
             atClusterTime:
               "$$exists": true
 
+## This test seems to be wrong
```
papafe: TODO create ticket

source/sessions/tests/snapshot-sessions.yml
```diff
@@ -480,3 +481,410 @@ tests:
...
+                snapshotTime: *savedSnapshotTime
+    - name: find
+      object: collection0
+      arguments:
+        session: session2
+        filter: {}
+      expectResult:
+        - { _id: 1, x: 11 }
+        - { _id: 2, x: 11 }
+    ## Calling find again to verify that atClusterTime/snapshotTime has not been modified after the first query
```
papafe: Would this be an ok approach? It seems we do not have a way of doing entity matching on an entity itself, but only on fields of it.

source/sessions/snapshot-sessions.md
```diff
@@ -110,11 +116,36 @@ Snapshot reads and causal consistency are mutually exclusive. Therefore if `snap
...
+    // other options defined by other specs
+}
+```
+
+Getting the value of `snapshotTime` on a non-snapshot session MUST raise an error.
+
 Transactions are not allowed with snapshot sessions. Calling `session.startTransaction(options)` on a snapshot session
 MUST raise an error.
 
+Note that a new operation on session called `getSnapshotTime` must be supported for [unified tests](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.md). This operation returns the value of `snapshotTime` on the session, so that it can be used in following operations.
```
papafe: Is this a good place for this note?

source/sessions/snapshot-sessions.md
```diff
@@ -110,11 +116,36 @@ Snapshot reads and causal consistency are mutually exclusive. Therefore if `snap
...
 
+### snapshotTime
+
+Applications set `snapshotTime` when starting a snapshot session to specify the desired snapshot time.
+
+Note that the `snapshotTime` property is optional. The default value of this property is null.
+
+Client MUST throw an error if `snapshotTime` is set and `snapshot` is not set to true.
+
+Note that when parsing `snapshotTime` for [unified tests](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.md), the parsed string is the name of the key for the actual value of `snapshotTime` to be found in the [entity map](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.md#entity-map).
```
papafe: Is this a good place for this note?

source/sessions/snapshot-sessions.md
```diff
@@ -110,11 +116,41 @@ Snapshot reads and causal consistency are mutually exclusive. Therefore if `snap
...
 
+### snapshotTime
+
+Applications set `snapshotTime` when starting a snapshot session to specify the desired snapshot time.
+
+Note that the `snapshotTime` property is optional. The default value of this property is null.
+
+Client MUST throw an error if `snapshotTime` is set and `snapshot` is not set to true.
+
+Note that when parsing `snapshotTime` from `sessionOptions` for
```
papafe: Is this a good place to describe this behaviour?

source/sessions/snapshot-sessions.md
```diff
@@ -110,11 +116,41 @@ Snapshot reads and causal consistency are mutually exclusive. Therefore if `snap
...
+[unified tests](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.md),
+the parsed string is the name of the key for the actual value of `snapshotTime` to be found in the
+[entity map](https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.md#entity-map).
+
 ## ClientSession changes
 
+A new readonly property called `snapshotTime` will be added to `ClientSession` that allows applications to retrieve the
```
papafe:
```suggestion
A readonly property called `snapshotTime` will be added to `ClientSession` that allows applications to retrieve the
```

source/sessions/tests/README.md
```diff
@@ -250,8 +250,24 @@ and configure a `MongoClient` with default options.
...
+
+- `client.startSession(snapshot = false, snapshotTime = new Timestamp(1))`
+- Assert that an error was raised by driver
+
+### 22. Retrieving `snapshotTime` on a non-snapshot session raises an error
+
+Snapshot sessions tests require server of version 5.0 or higher and replica set or a sharded cluster deployment.
+
+- `client.startSession(snapshot = false)`
+- `client.snapshotTime`
```
sanych-sun: it suppose to be `session`, right?

source/sessions/tests/README.md
```diff
@@ -250,8 +250,24 @@ and configure a `MongoClient` with default options.
...
+Snapshot sessions tests require server of version 5.0 or higher and replica set or a sharded cluster deployment.
+
+- `client.startSession(snapshot = false, snapshotTime = new Timestamp(1))`
+- Assert that an error was raised by driver
+
+### 22. Retrieving `snapshotTime` on a non-snapshot session raises an error
+
+Snapshot sessions tests require server of version 5.0 or higher and replica set or a sharded cluster deployment.
+
+- `client.startSession(snapshot = false)`
```
sanych-sun: I would prefer to have more verbal steps:
```
- Start a session by calling `startSession` with `snapshot = false`.
- Try to access the session's snapshot time.
- Assert that client side error was raised.
```

source/sessions/snapshot-sessions.md
sanych-sun: I would say we have to mention this in `tests/README.md` file.

source/sessions/snapshot-sessions.md
```diff
@@ -110,11 +116,41 @@ Snapshot reads and causal consistency are mutually exclusive. Therefore if `snap
...
+    // other options defined by other specs
+}
+```
+
+Getting the value of `snapshotTime` on a non-snapshot session MUST raise an error.
+
 Transactions are not allowed with snapshot sessions. Calling `session.startTransaction(options)` on a snapshot session
 MUST raise an error.
 
+Note that a new operation on session called `getSnapshotTime` must be supported for
```
sanych-sun: Same here, move top `tests/README.md`

source/sessions/tests/snapshot-sessions.yml
```diff
@@ -480,3 +481,404 @@ tests:
...
+                snapshotTime: *savedSnapshotTime
+    - name: find
+      object: collection0
+      arguments:
+        session: session2
+        filter: {}
+      expectResult:
+        - { _id: 1, x: 11 }
+        - { _id: 2, x: 11 }
+    - name: find
```
sanych-sun, line 520: Why do we making `find` with the `session2` twice, or have I missed some differences in the operations? If this is because we want to make sure `snapshot` time will not be changes after the first call - then it might makes sense to add comments here (same with other tests below).

source/sessions/tests/snapshot-sessions.yml
```diff
@@ -480,3 +481,404 @@ tests:
...
+      arguments:
+        filter: {}
+      expectResult:
+        - { _id: 1, x: 11 }
+        - { _id: 2, x: 11 }
+        - { _id: 3, x: 33 }
+  expectEvents:
+    - client: client0
+      events:
+        - commandStartedEvent:
```
sanych-sun, line 539: Why don't we expect event for `insertOne` operation (here and in other tests).

source/sessions/tests/README.md
papafe: Sure.

source/sessions/tests/snapshot-sessions.yml
papafe: Yes, that's exactly the reason. Otherwise we'll have to resolve to creating more prose tests. 
I'll add comments back in all the places where I'm doing this.

source/sessions/snapshot-sessions.md
papafe: Makes sense

source/sessions/tests/snapshot-sessions.yml
papafe: They are ignored in `ignoreCommandMonitoringEvents` inside `createEntities` for the client used in this whole test file. I suppose those events were ignored as they were not the focus of these tests.

source/sessions/snapshot-sessions.md
papafe: Done

source/sessions/tests/snapshot-sessions.yml
sanych-sun: Let's create the ticket and remove this comment as non-informative.

source/sessions/tests/README.md
```diff
@@ -29,6 +29,15 @@ As part of the test setup for these cases, create a `MongoClient` pointed at the
 in the test case and verify that the test server does NOT define a value for `logicalSessionTimeoutMinutes` by sending a
 hello command and checking the response.
 
+## Specific operations and behaviour for unified tests
+
+An operation on sessions called `getSnapshotTime` must be supported for unified tests. This operation returns the value
```
sanych-sun, line 34: Sorry for returning with this again, but I've just realized that we should mirror this changes into unified test format spec:
here: https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.md#session-operations
and here: https://github.com/mongodb/specifications/blob/master/source/unified-test-format/unified-test-format.md#entity_session

papafe: I've mirrored this in the Session Operations section in unified test format spec. I did not modify the Entity Session section, as it already refers to the single spec documents, where I've already added the necessary info.

alcaeus (approved):
Changes to the unified test format LGTM. Since we don't add operations to the schema, the unified test format does not get a version bump. Since this effectively prevents targeting a schema version in tests, please make sure to call out the requirement to implement `getSnapshotTime` in test runners in the ticket's downstream instructions.

PR #1845, opened 2025-09-26
DRIVERS-2884: CSOT avoid connection churn when operations timeout

sanych-sun:
This PR is based on the Preston's work in another PR that was closed by mistake: https://github.com/mongodb/specifications/pull/1675

This PR implements the design for connection pooling improvements described in [DRIVERS-2884](https://jira.mongodb.org/browse/DRIVERS-2884), based on the CSOT (Client-Side Operation Timeout) spec. It addresses connection churn caused by network timeouts during operations, especially in environments with low client-side timeouts and high latency.

When a connection is checked out after a network timeout, the driver now attempts to resume and complete reading any pending server response (instead of closing and discarding the connection). This may require multiple connection to be attempted during the connection check out from the pool.
Each pending response draining is subject to a cumulative 3-second static timeout. The timeout is refreshed after each successful read, acknowledging that progress is being made. If no data is read and the timeout is exceeded, the connection is closed.

This update introduces new CMAP events and logging messages (PendingResponseStarted, PendingResponseSucceeded, PendingResponseFailed) to improve observability of this path.

- [x] Update changelog.
- [x] Test changes in at least one language driver ([CSharp Driver implementation PR](https://github.com/mongodb/mongo-csharp-driver/pull/1792)).
    clusters).


source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -997,6 +1093,96 @@ interface ConnectionCheckedInEvent {
...
+
+  /**
+   *  The ID of the Connection.
+   */
+  connectionId: int64;
+
+  /**
+   *  The driver-generated request ID associated with the network timeout.
+   */
+  requestId: int64;
```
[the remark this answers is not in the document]
sanych-sun: Current approach from the log reader prospective will looks like: there are many operations, there are many checkout attempts, some of them faster, some slower, and also there are draining of connections happening. No way to build a relation between connection checkout duration and connections draining. Slower checkouts might be caused by the draining or simply waiting. If we have `requestId` and `originalRequestId` (need better name sure) - then reader can see, we have a longer checkout for THIS `requestId`, because there were some connection draining for that `originalRequestId`.

jmikola (approved):
I reviewed the unified test schema changes alone and those LGTM. I'll defer to CSOT spec folks for the other files.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -592,6 +675,18 @@ try:
         if connection is perished:
           close connection
           connection = Null
+        if connection is in "pending response" state:
+          drain the pending response
```
baileympearson: Do we have pseudocode for "drain the pending response" somewhere?  I don't see it in this PR.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -592,6 +675,18 @@ try:
         if connection is perished:
           close connection
           connection = Null
+        if connection is in "pending response" state:
```
baileympearson: Connection pools are FIFO.  Considering the case where we are successfully reading chunks of data off of the socket but we're timing out before we finish the read (the final else block in your pseudocode), won't checking the connection back into the pool and looping just cause this request to immediately checkout the same connection again (the behavior https://jira.mongodb.org/browse/DRIVERS-3276 is intending to avoid)? 

And regarding https://jira.mongodb.org/browse/DRIVERS-3276: are we convinced that its better to potentially block a number of operations on the slow connection than to block just one?

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -592,6 +675,18 @@ try:
...
+          drain the pending response
+          if error:
+            if non-timeout error:
```
baileympearson: Pseudocode needs to include new event and logging emissions.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -997,6 +1093,93 @@ interface ConnectionCheckedInEvent {
...
+
+  /**
+   *  The ID of the Connection.
+   */
+  connectionId: int64;
+
+  /**
+   *  The driver-generated request ID of the operation that caused the pending response state.
+   */
+  requestId: int64;
```
baileympearson, line 1130: Seems odd that the events emitted for an operation include a requestId that corresponds to the request that made the connection `pending response`.  What is the value in this datapoint?

If we decide to keep it - do you think a more precise name would be beneficial?  `requestId` is ambiguous - users could easily assume it refers to the request that is reading the pending response, not the operation that made the connection pending.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
baileympearson, line 1681: Please update the changelog.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -576,6 +578,87 @@ other threads from checking out [Connections](#connection) while establishing a
...
+2. **Aliveness check**: If driver uses synchronous read from the socket and the timestamp is older than 3 seconds, the
+    driver MUST perform aliveness check either performing a non-blocking read or using the minimal possible timeout to
+    check if at least one byte can be read/received. If at least one byte can be read the timestamp must be updated and
+    connection SHOULD be drained. If no bytes can be read, the connection MUST be closed.
+3. **User-provided timeout**: If a user-provided timeout is specified for the operation, the driver MUST use the minimum
+    of (a) the remaining time before the 3 second "pending response" window elapses and (b) the remaining of
+    user-provided timeout as the effective timeout for the draining.
+4. **Default timeout**: If no user-provided timeout is specified, the driver MUST use the minimum of (a) the remaining 3
+    second "pending response" window and (b) the `waitQueueTimeoutMS` (if not supported by the driver remaining of
+    `serverSelectionTimeoutMS` should be used instead) as the effective timeout for the draining.
```
baileympearson:
```suggestion
4. **Default timeout**: If no user-provided timeout is specified, the driver MUST use the minimum of (a) the remaining 3
    second "pending response" window and (b) remaining timeout for connection checkout.
```

regardless of whether CSOT is enabled or not or waitQueueTimeoutMS is supported, the timeout should always be `min(remaining connection checkout timeout, remaining static timeout)`.  With the existing phrasing, drivers would apply a full `waitQueueTimeoutMS` to every pending read attempt.

source/unified-test-format/unified-test-format.md
```diff
@@ -1133,6 +1136,27 @@ The structure of this object is as follows:
 - `connectionCheckedInEvent`: Optional object. If present, this object MUST be an empty document as all fields in this
     event are non-deterministic.
 
+<span id="expectedEvent_connectionPendingResponseStarted"></span>
+
+- `connectionPendingResponseStarted`: Optional object. If present, this object MUST be an empty document as all fields
+    in this event are non-deterministic.
+
```
baileympearson: The values are non-deterministic, but we can still assert on the structure of the new CMAP events. Right?

It might be worth instructing drivers to assert that properties exist on the events and that they're of the expected type but keep the phrasing that prohibits drivers from asserting on the values.

source/client-side-operations-timeout/tests/pending-response.yml
```diff
@@ -0,0 +1,349 @@
...
+        events:
+          - connectionCheckedOutEvent: {}
+          - connectionCheckedInEvent: {} # Insert fails.
+          - connectionPendingResponseStarted: {}
+          - connectionPendingResponseFailed:
+              reason: timeout
+          - connectionClosedEvent:
+              reason: error
+          - connectionCheckedOutEvent: {} # Find's retry starts and succeeds
+          - connectionCheckedInEvent: {}
```
baileympearson, line 148: There's minimal overhead to including more events than just these events, so I suggest adding other CMAP events, just to be explicit.  I'd expect it looks something like:

```
          - connectionCheckedOutEvent: {}
          - connectionCheckedInEvent: {} # Insert fails.
          - connectionCheckOutStarted: {} # NEW
          - connectionPendingResponseStarted: {}
          - connectionPendingResponseFailed:
              reason: timeout
          - connectionClosedEvent:
              reason: error
          - connectionCreatedEvent: {} # NEW
          - connectionReadyEvent: {} # NEW
          - connectionCheckedOutEvent: {} # Find's retry starts and succeeds
          - connectionCheckedInEvent: {}
```

Preston's original PR relied on throwing a retryable error after failing to read a pending response.  If we're explicitly _not_ allowing that approach, including extra events enforce that the retry happens inside a single checkout request.

baileympearson: (I don't see a reason not to apply this to all tests in the file too)

source/client-side-operations-timeout/tests/pending-response.yml
```diff
@@ -0,0 +1,349 @@
...
+        arguments:
+          timeoutMS: 50
+          filter: {_id: 1}
+        expectError:
+          isTimeoutError: true
+      # Execute a final operation to drain the buffer.
+      - name: find
+        object: *collection
+        arguments:
+          filter: {_id: 1}
```
baileympearson, line 235: Are we missing an explicit `timeoutMS` value here?

source/client-side-operations-timeout/tests/pending-response.yml
```diff
@@ -0,0 +1,349 @@
...
+      - name: wait
+        object: testRunner
+        arguments:
+          ms: 500
+      # Execute a subsequent operation with a timeout less than the block time.
+      - name: find
+        object: *collection
+        arguments:
+          timeoutMS: 50
+          filter: {_id: 1}
```
baileympearson: We've seen flakiness with establishing connections and this test will require a new connection to be established.  Should we up the timeout on the `find` just in case?

source/client-side-operations-timeout/tests/pending-response.yml
```diff
@@ -0,0 +1,349 @@
...
+          - connectionCheckedInEvent: {} # Insert fails.
+          - connectionPendingResponseStarted: {} # Pending read fails on first find
+          - connectionPendingResponseFailed:
+              reason: error
+          - connectionClosedEvent:
+              reason: error
+          - connectionCheckedOutEvent: {}
+          - connectionCheckedInEvent: {} # Find finishes.
+    # If the connection is closed server-side while draining the response, the
+    # driver must retry with a different connection.
```
baileympearson:
```suggestion
    # If the connection is closed server-side while draining the response, the
    # driver must retry with a different connection (write operation)
```

ditto for the test above, but with `read`.  Just reading the comments I thought these were duplicate tests.

source/client-side-operations-timeout/tests/pending-response.yml
```diff
@@ -0,0 +1,349 @@
...
+      - name: wait
+        object: testRunner
+        arguments:
+          ms: 500
+      # Execute a subsequent operation with a timeout less than the block time.
+      - name: insertOne
+        object: *collection
+        arguments:
+          timeoutMS: 50
+          document: {_id: 3, x: 1}
```
baileympearson: same comment as on above test.

source/connection-monitoring-and-pooling/tests/README.md
```diff
@@ -21,6 +21,74 @@ The following tests have not yet been automated, but MUST still be tested:
...
+1. Connect to the proxy server with `maxPoolSize=1` and `direct=true`, subscribing to the following CMAP events:
+    - `PendingResponseStarted`
+    - `PendingResponseSucceeded`
+    - `PendingResponseFailed`
+    - `ConnectionClosed`
+2. Send a command (e.g. an insert) with a 200 millisecond timeout and the following `proxyTest` actions:
+    - `sendBytes`: We have 3 possible states here:
+        1. Message size was partially read: random value between 1 and 3 inclusive
+        2. Message size was read, body was not read at all: use 4
+        3. Message size was read, body read partially: random value between 5 and 100 inclusive
```
baileympearson: The `insert` is the first command on the connection, so where are the values for `sendBytes` coming from?

source/connection-monitoring-and-pooling/tests/README.md
```diff
@@ -21,6 +21,74 @@ The following tests have not yet been automated, but MUST still be tested:
...
 
+### Pending Response
+
+Due to the complexity of managing a proxy layer, the following qualifying tests should only be run for non-ssl,
+non-auth, and non-compression connections.
+
+#### Recover Partially Read Response
+
+This test verifies that if only part of a response was read before the timeout, the driver can drain the rest of the
+response and reuse the connection for the next operation.
```
baileympearson: This doesn't match the actual contents of the test, so either the description or the test needs to be updated.

But if its the description that is inaccurate - with `sendAll: true` and the events (only one pending read started + finished pair), the full response will be read in the first operation.  Don't we have coverage for this scenario from our unified tests?

source/connection-monitoring-and-pooling/tests/README.md
```diff
@@ -21,6 +21,74 @@ The following tests have not yet been automated, but MUST still be tested:
...
+    - `ConnectionPendingResponseStarted`
+    - `ConnectionPendingResponseSucceeded`
+6. Verify that NONE the following events was observed:
+    - `ConnectionPendingResponseFailed`
+    - `ConnectionClosed`
+
+#### Non-destructive Aliveness Check
+
+This test verifies that if a connection idles past the pending response window (3 seconds) after a partial header, the
+aliveness check does not attempt to discard bytes from the TCP stream.
```
baileympearson, line 94: Aren't we supposed to close the connection if we receive no bytes in 3s?

baileympearson (changes requested):
@sanych-sun Have we implemented these changes in any driver?  Which driver is supposed to be the second implementer?  iirc it was originally Python

source/client-side-operations-timeout/tests/pending-response.yml
sanych-sun: We should not establish any new connection in this test. We have `minPoolSize` set to 1 together with `awaitMinPoolSizeMS` parameter, which means the connection should be established before the test start. Also we have `maxPoolSize` set to 1 too. So the only connection should be marked as "pending response" after the timeout, then we wait "long enough" for the server to reply... and then in scope of the last `find` operation it should be simple discarding of already buffered data.

source/connection-monitoring-and-pooling/tests/README.md
sanych-sun: This is the parameter for proxy, which specifies how many bytes should be streamed, before sleeping.

source/connection-monitoring-and-pooling/tests/README.md
sanych-sun: We are using proxy server here to control when/how server response arrived to the client side. So in step 2 in addition to the regular insert command payload, we have to add additional `proxyTest` property, which instruct proxy to emulate timeout. It works as following:
1) it send request to the server
2) based on the `sendBytes` parameter it stream requested amount of bytes instantly
3) it sleeps for `delayMS`
4) it streams rest of the response.


I think it makes sense to add this explanation to the test summary.

source/connection-monitoring-and-pooling/tests/README.md
sanych-sun: In scenario when connection is being sitting in the pool for some time longer then 3 seconds we do not want just to close the connection, because the response might be already buffered on the client side. We want to check if there are any new data arrived, if so we refresh the last read timestamp and performing draining as regular.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
sanych-sun: We can check in the connection back only if the current check out is timed out (either because of CSOT or waitQueueTimeout has elapsed), but "pending response sliding window" has not expired yet. But the "current" operation is timed out anyway, so it should raise an error instead of attempting another connection.

sanych-sun: If we have "slow to drain" connection in the pool - it will cause the same kind of problem regardless we do retry on operation level or if we attempting to drain several connections inside the checkout. Because such connection will return back to pool and affect next "unlucky" operation anyway.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
sanych-sun: The idea of reporting the original requestId is to keep tracking how long/how many draining attempt were made before success or failure. Honestly I would prefer to have BOTH, current request Id and "original timed out requestId", but it could make sense only if other check out event had "current requestId" field, which is not there =(

source/unified-test-format/unified-test-format.md
sanych-sun: We simply follow other check out events here, like `connectionCheckedOutEvent`. I can add simple validation for properties to exist.

source/connection-monitoring-and-pooling/tests/README.md
baileympearson: The value of `sendBytes` is determined based on the state of the connection.  But this is the first operation on the connection, so the connection will never have anything in the buffer and it isn't clear what the value of sendBytes should be.

Unless I misunderstand what sub-bullets 1,2, and 3 mean?

source/connection-monitoring-and-pooling/tests/README.md
baileympearson: I missed the sleep below.  thanks

source/connection-monitoring-and-pooling/tests/README.md
baileympearson: Okay, maybe my confusion stems from the interaction between `sendBytes` and `sendAll`.  Does `sendAll` being enabled not mean that the proxy will forward the full response back to the client?

source/client-side-operations-timeout/tests/pending-response.yml
baileympearson: Maybe I'm confused by how `closeConnection: true` fits into the picture here.  I imagined that:

- failCommand is triggered, write blocks for 500ms 
- client-side, insert times out
- failCommand expires, and connection is closed
- the second insert is executed, checks the connection in the pool, attempts to read the bytes from the socket only to find that the socket has been closed.  the connection is discarded, the checkout request immediately then attempts to open a new connection, which succeeds and the second insert completes successfully

Am I misunderstanding how this test works?

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
ShaneHarvey: This might depend on driver internals but in Python the new request ID will not be available at this point because the command has not been serialized (we need to checkout the connection first).

As for including the old requestId, I'm not sure how useful it is but it seems harmless to add. The driver needs to validate requestId == responseTo field on the server reply using the old requestId anyway to make sure the wire protocol isn't violated so that value will be available.

source/client-side-operations-timeout/tests/pending-response.yml
sanych-sun: Oh, yes! You are right! I've overlooked `closeConnection: true`. Thank you for spotting the problem!

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
sanych-sun: Done

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
sanych-sun: Thank you Shane! I suppose we want to keep the field. @baileympearson is this OK with you?

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
sanych-sun: Done.

source/client-side-operations-timeout/tests/pending-response.yml
sanych-sun: Tests we updated to include the following events:
- connectionCheckOutStarted
- connectionReadyEvent
- connectionCreatedEvent
- connectionReadyEvent

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
sanych-sun: As we have events emission in the current pseudocode, not sure if we need to have pseudocode for "drain the pending response", as it basically "consume the bytes from underling stream/socket and ignore them".

source/client-side-operations-timeout/tests/pending-response.yml
sanych-sun: If we do not have timeout we fallback to sockettimeout, but you are right, we better to have the explicit timeout here.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -592,6 +674,27 @@ try:
         if connection is perished:
           close connection
           connection = Null
+        if connection is in "pending response" state:
+          tConnectionDrainingStarted = current instant (use a monotonic clock if possible)
+          emit PendingResponseStartedEvent and equivalent log message
+          drain the pending response
+          if error:
```
matthewdale: The mix of exception/error paradigms in this pseudocode block is a bit confusing. We should rewrite this new pseudocode to use try/catch like the existing code.

sanych-sun: Rewritten, thank you!

source/connection-monitoring-and-pooling/tests/README.md
sanych-sun: Yes. I've updated steps with sample of payload, it might help to understand the idea.

source/client-side-operations-timeout/tests/pending-response.yml
sanych-sun: I've moved the test into the separate file, because I've run into a problem: on MongoDB version 4.4 connection was closed server side without waiting for `blockTimeMS`. Also as far as we have the retry on the checkout level, we do not need separate test for read and write operation.

source/connection-monitoring-and-pooling/tests/README.md
sanych-sun: That 3 sub-bullets - it's different test cases for the same test scenario. This is parameter for proxy, that define how many bytes of the server's response should be streamed instantly, before delay.
I rephrased a little, hope it's more readable now.

source/connection-monitoring-and-pooling/tests/README.md
sanych-sun: Here is the example of proxyTest parameter:
```
 proxyTest: {
          actions: [
             { sendBytes : 2 },
             { delayMs : 400 },
             { sendAll : true },
          ]
       }
```

Which can be read as:
Hey proxy, here are steps for you, do it one-by-one:
action 1: stream 2 bytes from the server response
action 2: wait for 400ms
action 3: stream rest of the response to the client.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -1339,6 +1581,47 @@ some equivalent configuration, but this configuration will also require target f
 5.0. The advantage of using Background Thread to manage perished connections is that it will work regardless of
 environment setup.
```
ShaneHarvey, line 1591: Could we add a rationale question to cover the motivation for this feature? Like "Why introduce the draining pending responses?" where the answer is to reduce connection churn in cases when the configured maxTimeMS does not allow enough time for the driver to read the MaxTimeMSExpired error and cases where the server or network delays the response.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
baileympearson: > We can check in the connection back only if the current check out is timed out (either because of CSOT or waitQueueTimeout has elapsed), but "pending response sliding window" has not expired yet. But the "current" operation is timed out anyway, so it should raise an error instead of attempting another connection.

Ah.  Good point

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
baileympearson: I'd like the pseudocode, it doesn't seem terrible involved to add and it would be nice to codify the calculation of the timeout for the pending read in pseudocode:

```
read_timeout =  timeoutMS set 
    ? csotMin(timeoutMS, remaining static timeout) 
    : waitQueueTimeoutMS set 
        ? csotMin(waitQueueTimeoutMS, remaining static timeout) 
        : remaining static timeout
```

source/client-side-operations-timeout/tests/pending-response-close-connection.yml
```diff
@@ -0,0 +1,98 @@
...
+              closeConnection: true
+      # Execute operation with timeout less than block time.
+      - name: insertOne
+        object: *collection
+        arguments:
+          timeoutMS: 50
+          document: {_id: 3, x: 1}
+        expectError:
+          isTimeoutError: true
+      # Draining pending response should failure because of closed connection,
```
baileympearson:
```suggestion
      # Draining pending response should fail because of closed connection,
```

source/client-side-operations-timeout/tests/pending-response-close-connection.yml
```diff
@@ -0,0 +1,98 @@
...
+      database: *database
+      collectionName: *collection
+initialData:
+  - collectionName: *collection
+    databaseName: *database
+    documents: []
+tests:
+    # If the connection is closed server-side while draining the response, the
+    # driver must retry with a different connection.
+  - description: "write op retries when connection closes server-side while draining response"
```
baileympearson:
```suggestion
  - description: "op retries when connection closes server-side while draining response"
```

No need to specify read/write operations now, right?

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
baileympearson: Yeah, fine to keep it - thoughts on workshopping the name to make it clearer?

source/connection-monitoring-and-pooling/tests/README.md
```diff
@@ -21,6 +21,111 @@ The following tests have not yet been automated, but MUST still be tested:
...
+2. Connect to the proxy server with `maxPoolSize=1`, `direct=true`, disable read and write retryability, subscribing to
+    the following CMAP events:
+    - `PendingResponseStartedEvent`
+    - `PendingResponseSucceededEvent`
+    - `PendingResponseFailedEvent`
+    - `ConnectionCreatedEvent`
+    - `ConnectionClosedEvent`
+3. Execute `ping` command to populate the connection pool.
+4. Send a command (e.g. an insert) with a 200 millisecond timeout and the following `proxyTest` actions:
+    - `sendBytes`: random value between 1 and 3 inclusive
```
baileympearson:
```suggestion
    - `sendBytes`: 2
```

No reason for random value here - suggest we just choose a number.  I chose 2 because that matches your example payload below

source/connection-monitoring-and-pooling/tests/README.md
```diff
@@ -21,6 +21,111 @@ The following tests have not yet been automated, but MUST still be tested:
...
+    - `delayMs`: 400 ( to exceed the 200 ms timeout)
+    - `sendAll`: `true` Example of run command payload:
+    ```
```
baileympearson: This is really minor but GH seems to render the last bullet and the codeblock without any spacing.  
```suggestion
    - `sendAll`: `true` Example of run command payload:
    
    ```
```

baileympearson (changes requested):
Nothing major from me!  Still waiting on a second implementation.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -592,6 +674,27 @@ try:
...
           close connection
           connection = Null
+        if connection is in "pending response" state:
+          tConnectionDrainingStarted = current instant (use a monotonic clock if possible)
+          emit PendingResponseStartedEvent and equivalent log message
+          try:
+            drain the pending response
+          except timeout:
+            tConnectionDrainingFailed = current instant (use a monotonic clock if possible)
+            emit PendingResponseFailedEvent(reason="timeout", duration = tConnectionDrainingFailed - tConnectionDrainingStarted) and quivalent log message
```
matthewdale: Typo.

```suggestion
            emit PendingResponseFailedEvent(reason="timeout", duration = tConnectionDrainingFailed - tConnectionDrainingStarted) and equivalent log message
```

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -592,6 +674,27 @@ try:
         if connection is perished:
           close connection
           connection = Null
+        if connection is in "pending response" state:
+          tConnectionDrainingStarted = current instant (use a monotonic clock if possible)
+          emit PendingResponseStartedEvent and equivalent log message
+          try:
+            drain the pending response
```
matthewdale: The pseudocode notes above that drivers should only allow one thread to run the check-out process at a time:
> Note that in a lock-based implementation of the wait queue would only allow one thread in the following block at a time

Reading connections in "pending response" state for 3 seconds while holding an exclusive lock could introduce a huge bottleneck in multi-threaded drivers. Is there a way to move that pending response read out of the locked block?

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -592,6 +674,27 @@ try:
...
+          except timeout:
+            tConnectionDrainingFailed = current instant (use a monotonic clock if possible)
+            emit PendingResponseFailedEvent(reason="timeout", duration = tConnectionDrainingFailed - tConnectionDrainingStarted) and quivalent log message
+            if last read timestamp on connection > 3 seconds old
+              close connection
+            else
+              check in the connection
+            connection = Null
```
matthewdale: If an operation with no timeout picks up a "pending response" connection, are there cases where that operation can wait an indefinitely long time to get a connection?

Consider the scenario:
There is one connection in a connection pool. The one connection is in the  "pending response" state, where the connection is consistently but very slowly receiving data from the server (at least 1 byte every 3 seconds).

1. An operation with no timeout tries to check out a connection.
2. The first connection in the stack is the above "pending response" connection. The driver tries to drain the pending response.
3. Draining the pending response takes longer than 3 seconds, so the connection is checked back into the pool.
4. The driver selects the next connection, which is the same connection. Go to step 2 and repeat.

Questions:
A. Is that scenario realistic?
B. For drivers that support the deprecated "waitQueueTimeoutMS", should they apply that timeout to the above check-out process?
C. For drivers that no longer support the deprecated "waitQueueTimeoutMS", should they apply "serverSelectionTimeoutMS" to the above check-out process?

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
sanych-sun: Don't we have the same issue for `create connection` as well? How we fix the issue with `create connection` in lock based implementation? I suppose we have to apply the same fix/workaround to pending response draining as well.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
ShaneHarvey: I don't think we need to design a solution to handle the "very slowly receiving data" problem under the assumption that it's extremely unlikely to occur. However that question has come up often enough that we should add justification/rationale.

That scenario could be manufactured by an attacker that controls the network but in that case they can already DOS the system by blocking all traffic.

sanych-sun: The proposed changes DO NOT change any requirements regarding the timeouts for the whole checkout process. For Drivers that supports `waitQueueTimeoutMS` - we should use it to limit checkout time, otherwise it's remaining of `serverSelectionTimeoutMS`. Regarding the draining of the pending response timeout, there is requirement to use **Default timeout** if no timeout was provided:
> 4. **Default timeout**: If no user-provided timeout is specified, the driver MUST use the minimum of (a) the remaining 3 second "pending response" window and (b) remaining timeout for connection checkout.

What means there is NO unlimited amount of retries, even if the connection will be checked in back into the connection pool, we whole process cannot take longer then timeout prescribed for connection checkout.

Answers to the questions:
A. As Shane stated - even though this sounds like a potential problem, we are assuming this is unrealistic scenario for now. In order to mitigate the problem we could introduce some additional timeout to limit time from the "first-byte" to the "whole response read", but not sure if we should do that just now.
B./C. I believe it was mentioned that we should respect remaining waitQueueTimeoutMS or serverSelectionTimeoutMS for draining pending response.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
matthewdale: My understanding is that `create connection` is a non-blocking operation that creates a connection in the "pending" state. The `establish connection` in the block below is where the blocking work is done. That block specifically notes:
> This MUST NOT block other threads from acquiring connections.

ShaneHarvey: How about we add note like we do in the maxConnecting case?:
```
        # this waiting MUST NOT prevent other threads from checking Connections
        # back in to the pool.
        wait until pendingConnectionCount < maxConnecting or a connection is available
        continue
```
Like
```
            # this I/O MUST NOT block other threads performing connection checkout
            drain the pending response
```

matthewdale: I'm concerned adding that note glosses over the complexity of implementing such a locking strategy. However, I also realize that the pseudocode is primarily for communicating the process and may not resemble actual driver code for checkOut.

Would a change like the following work?
```
...

try:
  enter WaitQueue
  wait until at top of wait queue

  while connection is Null:
    # Note that in a lock-based implementation of the wait queue would
    # only allow one thread in the following block at a time
    if a connection is available:
      while connection is Null and a connection is available:
        connection = next available connection
        if connection is perished:
          close connection
          connection = Null
    else if totalConnectionCount < maxPoolSize:
      if pendingConnectionCount < maxConnecting:
        connection = create connection
      else:
        # this waiting MUST NOT prevent other threads from checking Connections
        # back in to the pool.
        wait until pendingConnectionCount < maxConnecting or a connection is available
        continue

    if connection is in "pending response" state:
      tConnectionDrainingStarted = current instant (use a monotonic clock if possible)
      emit PendingResponseStartedEvent and equivalent log message
      # this I/O MUST NOT block other threads performing connection checkout
      drain the pending response

...
```

sanych-sun: @matthewdale Bailey and I agreed to adjust the code very close to what you are suggesting here. With the only difference: `enter WaitQueue` will be under the outer loop. I'll push new version of the pseudocode soon.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
matthewdale: Sounds good. @sanych-sun can you add an explicit note that `waitQueueTimeoutMS` or `serverSelectionTimeoutMS` must bound connection checkout? Maybe add a link to the relevant [section](https://github.com/sanych-sun/specifications/blob/master/source/client-side-operations-timeout/client-side-operations-timeout.md#server-selection) about using `serverSelectionTimeoutMS` for connection checkout from the CSOT spec?

sanych-sun: Timeout calculations added to the pseudocode.
Thank you!

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
sanych-sun: Done. Outer loop was added to support draining of the pending response outside of wait queue .
@matthewdale @baileympearson Could you please re-review the checkout  pseudocode?
Thanks!

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
sanych-sun: Timeout calculation was added. I've introduced calculation for entire check out accordingly to CSOT spec (please find `checkout_timeout` variable) together with `draining_timeout` which is calculated as min (remaining of checkout_timeout, remaining of expiration window).

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
sanych-sun: Open for suggestions on better name.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
sanych-sun: Done.

source/unified-test-format/unified-test-format.md
```diff
@@ -3459,6 +3483,11 @@ other specs *and* collating spec changes developed in parallel or during the sam
 
 ## Changelog
 
+- 2025-09-25: **Schema version 1.28**.
```
alcaeus, line 3486: Reminder to change the date before merging.

alcaeus (dismissed):
Changes to the unified test files LGTM. Remember to change the changelog date before merging this PR.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
matthewdale: Looks good! The pseudocode makes it clearer that a checkOut draining a pending response doesn't get to keep its place in the wait queue and must re-enter the wait queue if the draining times out.

matthewdale (dismissed):
Looks good! 👍

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
matthewdale: Thanks for adding the note about the timeout! I checked and confirmed that the Go Driver uses `serverSelectionTimeoutMS` to bound conn checkout currently, but it's easy to miss and more important with this change.

source/connection-monitoring-and-pooling/tests/README.md
```diff
@@ -21,6 +21,142 @@ The following tests have not yet been automated, but MUST still be tested:
...
+
+We use mongoproxy to emulate the network delays. See
+[mongoproxy/README.md](https://github.com/mongodb-labs/drivers-evergreen-tools/blob/master/.evergreen/mongoproxy/README.md)
+for more details. Due to the complexity of managing a proxy layer, the following qualifying tests should only be run for
+non-ssl, non-auth, and non-compression connections.
+
+#### Recover Partially Read Response
+
+This test verifies that if only part of a response was read before the timeout, the driver can drain the rest of the
+response and reuse the connection for the next operation. This test has 3 variation, see step 4, `sendBytes` parameter.
```
prestonvasquez: [nit]

```suggestion
response and reuse the connection for the next operation. This test has 3 variations, see step 4, `sendBytes` parameter.
```

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -576,31 +578,140 @@ other threads from checking out [Connections](#connection) while establishing a
...
+The next time connection in the "pending response" state is checked out, the driver MUST ensure that any remaining
+response data is drained and the connection is ready to send the next command.
+
+1. **Persist and update timestamp**: The connection must record the current time immediately after the original timeout.
+    This timestamp MUST be updated to the current time whenever any bytes are successfully read, received, or consumed
+    while draining of the pending response.
+2. **Aliveness check**: If driver uses synchronous read from the socket and the timestamp is older than 3 seconds, the
+    driver MUST perform aliveness check either performing a non-blocking read or using the minimal possible timeout to
+    check if at least one byte can be read/received. If at least one byte can be read the timestamp must be updated and
+    connection SHOULD be drained. If no bytes can be read, the connection MUST be closed.
```
prestonvasquez: [nit] 

```suggestion
2. **Aliveness check**: If driver uses synchronous read from the socket and the timestamp is older than 3 seconds, then the
    driver MUST do an aliveness check by either performing a non-blocking read or using the minimal possible timeout to
    check if at least one byte can be read/received. If at least one byte can be read, then the timestamp must be updated and the
    connection SHOULD be drained. If no bytes can be read, the connection MUST be closed.
```

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -576,31 +578,140 @@ other threads from checking out [Connections](#connection) while establishing a
...
+3. **User-provided timeout**: If a user-provided timeout is specified for the operation, the driver MUST use the minimum
+    of (a) the remaining time before the 3 second "pending response" window elapses and (b) the remaining of
+    user-provided timeout as the effective timeout for the draining.
```
prestonvasquez: [nit]

```suggestion
3. **User-provided timeout**: If a user-provided timeout is specified for the operation, the driver MUST use the minimum
    of (a) the remaining time before the 3 second "pending response" window elapses and (b) the remaining
    user-provided timeout as the effective timeout for the draining.
```

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -576,31 +578,140 @@ other threads from checking out [Connections](#connection) while establishing a
...
+
+##### Pseudocode
+
+This subsection outlines the pseudocode steps for acquiring a connection from the pool.
+
 ```text
 connection = Null
 tConnectionCheckOutStarted = current instant (use a monotonic clock if possible)
 emit ConnectionCheckOutStartedEvent and equivalent log message
+checkout_timeout = timeoutMS is set ? remaining computedServerSelectionTimeout : waitQueueTimeoutMS
```
prestonvasquez: [blocking] IIUC this line is the interpretation of part 4: i.e., "remaining timeout for connection checkout". But `waitQueueTimeoutMS` is a [deprecated option](https://github.com/mongodb/specifications/blob/909b4d6ddbefc3e2ecbf6e75002ae695f4860bc4/source/uri-options/uri-options.md#list-of-specified-options) and may not be available in major releases of drivers post-deprecation. For example, the Go Driver removed support in v2.  Suggest removing this line for simplicity and updating `draining_timeout` to be calculated in the following way:

```
if waitQueueTimeoutMS is supported and configured and timeoutMS is not set:
	# waitQueueTimeoutMS is a deprecated option that may not be supported by the latest 
	# major version of all drivers.
	draining_timeout = min(remaining waitQueueTimeoutMS, remaining expiration window)
else:
	draining_timeout = min(remaining computedServerSelectionTimeout, remaining expiration window)
```

Edit: fix typo in first consequent, should use remaining waitQueueTimeoutMS.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
```diff
@@ -1339,6 +1586,54 @@ some equivalent configuration, but this configuration will also require target f
 5.0. The advantage of using Background Thread to manage perished connections is that it will work regardless of
 environment setup.
```
prestonvasquez, line 1591: [docs] Could we add a rationale addressing the trickling issue mentioned here: https://github.com/mongodb/specifications/pull/1845#discussion_r2434216712

prestonvasquez (changes requested):

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
sanych-sun: `min(waitQueueTimeoutMS, remaining expiration window)` - is wrong formula for `draining_timeout`, it should be `remaining of checkout_timeout` to emphasize that waitQueueTimeout SHOULD NOT refresh in case of multiple draining attempts.

sanych-sun: BTW, for Go Driver what if user does not specified timeoutMs, does it mean connection checkout could run infinite time?

prestonvasquez: > min(waitQueueTimeoutMS, remaining expiration window) - is wrong formula for draining_timeout, it should be remaining of checkout_timeout to emphasize that waitQueueTimeout SHOULD NOT refresh in case of multiple draining attempts.

There's a typo in the suggestion, the consequent of the first branch should be this:

```go
draining_timeout = min(remaining waitQueueTimeoutMS, remaining expiration window)
```

If you think it reads better at the top-level:

```go
if waitQueueTimeoutMS is supported and configured and timeoutMS is not set:
    checkout_timeout = waitQueueTimeoutMS
else:
    checkout_timeout = remaining computedServerSelectionTimeout
```

> for Go Driver what if user does not specified timeoutMs, does it mean connection checkout could run infinite time?

In the context of the changes describe in this PR and the [pseudo-code above](https://github.com/mongodb/specifications/pull/1845#discussion_r2500825232), the Go Driver will use the minimum of the remaining computed server selection timeout and the remaining expiration window. I expect serverSelectionTimeoutMS to always be realistically finite and bounded by the same pathological cases that would affect waitQueueTimeoutMS (e.g., a user explicitly setting it to an extremely large value).

For the current Go Driver behavior, we are compliant with the CSOT specifications which prescribe the following: 	

> If timeoutMS is set, drivers MUST use min(serverSelectionTimeoutMS, remaining timeoutMS), referred to as computedServerSelectionTimeout as the timeout for server selection and connection checkout.
>
> After a server has been selected, drivers MUST use the remaining computedServerSelectionTimeout value as the timeout for connection checkout.

sanych-sun: Pseudocode updated accordingly to the last suggestion.

source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.md
sanych-sun: Added as a part of "Why do we refresh the pending response timeout each time we successfully drain bytes from the TCP stream?" question.

PR #1924, opened 2026-04-14
DRIVERS-3446 - Add prose tests to verify correct retry behavior when a mix of overload and non-overload errors are encountered

NoahStapp:
DRIVERS-3446

- [X] Is the relevant DRIVERS ticket in the PR title?

- [X] Update changelog.
- [X] Test changes in at least one language driver. [Python](https://github.com/mongodb/mongo-python-driver/pull/2755).
- [X] Test these changes against all server versions and topologies (including standalone, replica set, and sharded
    clusters).

Ruby PoC: https://github.com/mongodb/mongo-ruby-driver/pull/3021

### DIFF
--- a/source/client-backpressure/client-backpressure.md
+++ b/source/client-backpressure/client-backpressure.md
@@ -159,9 +159,9 @@ The retry policy in this specification is separate from the other retry policies
 specifications. Drivers MUST ensure:
 
 - When a failed attempt is retried, backoff MUST be applied if and only if the error is an overload error.
-- If an overload error is encountered:
+- If an overload error is encountered at any point during an operation's retry loop:
     - Regardless of whether CSOT is enabled or not, the maximum number of retries for any retry policy becomes
-        `MAX_RETRIES`.
+        `MAX_RETRIES`. This includes retryable non-overload errors following the encountered overload error.
     - If CSOT is enabled and a command has been retried at least `MAX_RETRIES` times, it MUST NOT be retried further.
 
 #### Pseudocode
@@ -307,7 +307,7 @@ The Node and Python drivers will provide the reference implementations. See
 The client backpressure retry loop is primarily concerned with spreading out retries to avoid retry storms. The exact
 sleep duration is not critical to the intended behavior, so long as we sleep at least as long as we say we will.
 
-### Why override existing maximum number of retry attempt defaults for retryable reads and writes if an overload error is received?
+### Why override existing maximum number of retry attempt defaults for retryable reads and writes if an overload error is received at any point during an operation's retry loop?
 
 Load-shedded errors indicate that the request was rejected by the server to minimize load, not that the command failed
 for logical reasons. So, when determining the number of retries an operation should attempt:
@@ -325,8 +325,8 @@ specifications with load-shedding behavior:
     [retryable writes](../retryable-writes/retryable-writes.md) and
     [CSOT](../client-side-operations-timeout/client-side-operations-timeout.md) specifications when no overload errors are
     encountered.
-- Adjusting the maximum number of retry attempts to 5 if an overload error is returned from the server gives requests
-    more opportunities to succeed and helps reduce application errors.
+- Adjusting the maximum number of retry attempts to MAX_RETRIES if an overload error is returned from the server gives
+    requests more opportunities to succeed and helps reduce application errors.
 - An alternative approach would be to retry once if we don't receive an overload error, in which case we'd retry 5
     times. The approach chosen allows for additional retries in scenarios where a non-overload error fails on a retry
     with an overload error.
--- a/source/client-backpressure/client-backpressure.md
+++ b/source/client-backpressure/client-backpressure.md
@@ -159,9 +159,10 @@ The retry policy in this specification is separate from the other retry policies
 specifications. Drivers MUST ensure:
 
 - When a failed attempt is retried, backoff MUST be applied if and only if the error is an overload error.
-- If an overload error is encountered at any point during an operation's retry loop:
+- If an overload error is encountered:
     - Regardless of whether CSOT is enabled or not, the maximum number of retries for any retry policy becomes
-        `MAX_RETRIES`. This includes retryable non-overload errors following or preceding the encountered overload error.
+        `MAX_RETRIES`. This limiting applies to all retry attempts, including retries for errors that are not overload
+        errors.
     - If CSOT is enabled and a command has been retried at least `MAX_RETRIES` times, it MUST NOT be retried further.
 
 #### Pseudocode
--- a/source/client-backpressure/client-backpressure.md
+++ b/source/client-backpressure/client-backpressure.md
@@ -161,7 +161,7 @@ specifications. Drivers MUST ensure:
 - When a failed attempt is retried, backoff MUST be applied if and only if the error is an overload error.
 - If an overload error is encountered at any point during an operation's retry loop:
     - Regardless of whether CSOT is enabled or not, the maximum number of retries for any retry policy becomes
-        `MAX_RETRIES`. This includes retryable non-overload errors following the encountered overload error.
+        `MAX_RETRIES`. This includes retryable non-overload errors following or preceding the encountered overload error.
     - If CSOT is enabled and a command has been retried at least `MAX_RETRIES` times, it MUST NOT be retried further.
 
 #### Pseudocode
@@ -431,6 +431,8 @@ another dimension to read preference selection that users need to reason about. 
 to understand and configure.
 
 ## Changelog
+
+- 2026-04-14: Clarify correct retry behavior when a mix of overload and non-overload errors are encountered.
 
 - 2026-03-30: Introduce phase 1 support without token buckets.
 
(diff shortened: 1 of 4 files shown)


source/client-backpressure/tests/README.md
```diff
@@ -59,7 +59,7 @@ executed against a MongoDB 4.4+ server that has enabled the `configureFailPoint`
         The sum of 2 backoffs is 0.3 seconds. There is a 0.3-second window to account for potential variance between the
         two runs.
 
-#### Test 3: Overload Errors are Retried a Maximum of MAX_RETRIES times
+#### Test 2: Overload Errors are Retried a Maximum of MAX_RETRIES times
```
stIncMale: We should not change the test numbers. They were, I believe, on purpose left out unchanged when [one of the tests was previously removed](https://github.com/mongodb/specifications/commit/1b15cd299011172f941893cbe3c9d4737cb2ee5d) - drivers may refer to the prose tests either by their numbers, or by linking to a specific test.

Changing the number or the test name breaks those references. It is especially bad when a test is referred to by its number - the old test 3 is now 2, but there is still a test with the number 3, and if a driver refers to the 3rd test, it is not obvious that the reference is broken.

The same applies to the test below.

source/client-backpressure/client-backpressure.md
```diff
@@ -159,9 +159,9 @@ The retry policy in this specification is separate from the other retry policies
 specifications. Drivers MUST ensure:
 
 - When a failed attempt is retried, backoff MUST be applied if and only if the error is an overload error.
-- If an overload error is encountered:
+- If an overload error is encountered at any point during an operation's retry loop:
```
stIncMale: It's not operations but commands that the read/write retry policies, or the new overload retry policy prescribes the drivers to retry. Previously, the effort was made to make sure that `client-backpressure.md` correctly refers to either command or operation, given that they are not the same thing.

Let's replace "operation" with "command" here, because that's the correct term here and below in _"Why override existing maximum number of retry attempt defaults for retryable reads and writes if an overload error is received at any point during an operation's retry loop?"_.

stIncMale: What is the "_at any point during an operation's retry loop_" part intended to clarify? Nowhere else in `client-backpressure.md` the "at any point of the command retry loop" qualification is made, because it is implied and seems obvious. What is specific about this item that requires the qualification?

Hopefully, another reviewer may see this question as useful, get to the answer, and then act on it however it seems appropriate.

source/client-backpressure/client-backpressure.md
```diff
@@ -159,9 +159,9 @@ The retry policy in this specification is separate from the other retry policies
...
     - Regardless of whether CSOT is enabled or not, the maximum number of retries for any retry policy becomes
-        `MAX_RETRIES`.
+        `MAX_RETRIES`. This includes retryable non-overload errors following or preceding the encountered overload error.
```
stIncMale: `MAX_RETRIES` limits `attempt`s (_"the execution attempt number (starting with 0)"_). `attempt` is currently documented in the "Overload retry policy" section as

> includes retries for errors that are not overload errors (this might include attempts under other retry policies, see [Interactions with Other Retry Policies](https://github.com/mongodb/specifications/blob/master/source/client-backpressure/client-backpressure.md#interaction-with-other-retry-policies)).

It appears that the addition made here tries to duplicate the above information, but the way it is done does not seem like correct English to me (though I am not a native speaker): the subject in the new sentence is "this", but I fail to find anything in the previous sentence that could act as the thing referred to by "this". Maybe the proposed sentence could be reformulated as

> Note that `MAX_RETRIES` limits `attempts`, and `attempts` includes retries for errors that are not overload errors.

stIncMale (dismissed):
I won't be able to come back to this PR review, which means, my comments/suggestions should not block the PR. But I hope both the author and the other reviewers will take them into account.

source/client-backpressure/client-backpressure.md
stIncMale: Another idea:

When reviewing the PR that introduced the requirements under the _"If an overload error is encountered"_ condition, I suggested to formulate the two them by explicitly saying something like "extends the retry attempt limit..." and "forbids retry attempts after...". The current wording was chosen because it was much shorter than what I envisioned, and was still correct. However, given that it turned out to be practically difficult to understand, we may consider reformulating these requirements using the aforementioned wordings (if we do that, great care should be taken to formulate them correctly, which, historically, is difficult to do).

source/client-backpressure/tests/README.md
comandeo-mongo: If so, we should probably add a stub, something like "Test 2: REMOVED" to avoid confusion in the future.

source/client-backpressure/tests/README.md
```diff
@@ -117,3 +117,30 @@ option.
 5. Assert that the raised error contains both the `RetryableError` and `SystemOverloadedError` error labels.
 
 6. Assert that the total number of started commands is `maxAdaptiveRetries` + 1 (2).
+
+#### Test 4: Backoff is not applied for non-overload retryable errors
```
comandeo-mongo: The suggested test covers a non-backpressure path; we can test a more complex situation - Non-overload error inside the overload loop. No backoff should be applied to the non-overload error:

1. Let client be a MongoClient.

2. Let coll be a collection.

3. Configure the following failpoint:
```javascript
      {
          configureFailPoint: 'failCommand',
          mode: { times: 1 },
          data: {
              failCommands: ['find'],
              errorCode: 91,
              errorLabels: ['RetryableError', 'SystemOverloadedError']
          }
      }
```

4. Using the CommandFailedEvent from the command monitoring API, configure a callback that, when the first find error is observed, configures the following failpoint:
```javascript
      {
          configureFailPoint: 'failCommand',
          mode: 'alwaysOn',
          data: {
              failCommands: ['find'],
              errorCode: 91,
              errorLabels: ['RetryableError']
          }
      }
```

5. Perform a findOne operation with coll. Expect the operation to fail.

6. Assert that backoff was applied only once — for the initial overload error — and not for the subsequent non-overload retryable errors. Drivers MAY implement this assertion by overriding BASE_BACKOFF to a large value (e.g., 5 seconds) before running the test and verifying that the total elapsed time of the findOne operation is less than BASE_BACKOFF plus a small margin for network overhead. With correct behavior, only one backoff is applied (bounded by BASE_BACKOFF), so the elapsed time stays under that bound. If backoff were incorrectly applied for non-overload errors as well, the elapsed time would significantly exceed BASE_BACKOFF.

comandeo-mongo (commented):
I believe that the existing wording is good enough, we can leave the text as it is.

Two new tests add the necessary clarification and help to catch incorrect implementations.

source/client-backpressure/client-backpressure.md
NoahStapp: It was intended to clarify that a single overload error at any point will set the maximum number of retries for all errors encountered by the command. With the clarification added below, this addition can be removed.

source/client-backpressure/tests/README.md
NoahStapp: Good suggestion. This test is also more consistent with our existing retryable reads tests for backpressure, I'll replace the original with your suggestion and move it there.

source/client-backpressure/client-backpressure.md
NoahStapp: Updated the wording for clarity here.

comandeo-mongo (approved):

source/retryable-writes/tests/README.md
```diff
@@ -419,8 +419,58 @@ to cover the same sequence of events.
...
+
+5. Disable the fail point:
+
+    ```javascript
+    {
+        configureFailPoint: "failCommand",
+        mode: "off"
+    }
+    ```
+
```
papafe, line 469: Should we have an equivalent test 5 also for retryable writes?

source/retryable-writes/tests/README.md
```diff
@@ -419,8 +419,58 @@ to cover the same sequence of events.
...
+            errorLabels: ["RetryableError", "RetryableWriteError"],
+            errorCode: 91
+        }
+    }
+    ```
+
+    Configure the second fail point command only if the failed event is for the first error configured in step 2.
+
+4. Attempt an `insertOne` operation on any record for any database and collection. Expect the `insertOne` to fail with a
+    server error. Assert that `MAX_ADAPTIVE_RETRIES + 1` attempts were made.
```
papafe: Should we call this `MAX_RETRIES` to be consistent wit the spec files?

papafe (changes requested):
Overall it looks good, just two minor comments.

source/retryable-writes/tests/README.md
NoahStapp: Since this test is in `retryable-writes`, the intent was to specify `MAX_ADAPTIVE_RETRIES` as we do in the `retryable-reads` tests to remove ambiguity.

source/retryable-writes/tests/README.md
NoahStapp: Yes, good catch.

source/retryable-writes/tests/README.md
papafe: I forgot to specify that this comment should be for all occurrences of this parameter in both retryable reads and writes tests. I think we just introduced it, didn't we?

papafe (approved):
LGTM!

PR #1925, opened 2026-04-24
DRIVERS-3321 Support Substring/Suffix/Prefix indexes as GA

mdb-ad:
Pending completion of SPM-4539


rozza:
@mdb-ad do we have to update the minServerVersion for these tests?

kevinAlbs:
Closing. I expect to open a new PR for DRIVERS-3321.

PR #1928, opened 2026-04-29
DRIVERS-3461 skip QE "prefixPreview" and "suffixPreview" tests on server 9.0.0+

kevinAlbs:
# Summary

Skip Queryable Encryption (QE) tests of "prefixPreview" and "suffixPreview" on server 9.0.0+.

# Background & Motivation
Following SERVER-123416, creating a collection with "suffixPreview" and "prefixPreview" is rejected by 9.0.0-rc0 servers, resulting in an error:

```
Cannot create a collection with an encrypted field with query type prefixPreview, as it is deprecated
```

This PR skips tests of "suffixPreview" and "prefixPreview" tests on 9.0.0+ servers. "substringPreview" is still supported and tested on 9.0.0+ servers.

As a drive-by fix, a typo is fixed in a test description of QE-Text-substringPreview: "suffixPreview" => "substringPreview".

This PR is intended as a short-term test fix. Upcoming spec changes of DRIVERS-3321 will update to use and test "suffix" and "prefix".

---

- ~~[ ] Update changelog.~~ (N/A. Test changes only)
- [x] Test changes in at least one language driver. (Tested in C driver: https://github.com/mongodb/mongo-c-driver/pull/2284)
    clusters).

### DIFF
--- a/source/client-side-encryption/tests/README.md
+++ b/source/client-side-encryption/tests/README.md
@@ -3884,7 +3884,8 @@ Using [QE CreateCollection() and Collection.Drop()](../client-side-encryption.md
 create the following collections with majority write concern:
 
 - `db.prefix-suffix` using the `encryptedFields` option set to the contents of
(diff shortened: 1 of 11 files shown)


rozza (approved):
LGTM

PR #1930, opened 2026-04-29
DRIVERS-3274 Send afterClusterTime on writes in causally-consistent sessions.

matthewdale:
[DRIVERS-3274](https://jira.mongodb.org/browse/DRIVERS-3274)

Update the Causal Consistency spec to require that `afterClusterTime` is sent on write commands in causally-consistent sessions. Write commands that should send `afterClusterTime` are:
- `insert`
- `update`
- `findAndModify`
- `delete`
- `bulkWrite`
- `create`
- `createIndexes`
- `drop`
- `dropDatabase`
- `dropIndexes`

Previously, `afterClusterTime` was only sent on read commands and the first command in a transaction.

Changes:
*  Update the spec prose
* Add two new unified test files for the Causal Consistency spec
* Update the Read Write Concern spec to list write commands that accept `readConcern`
* Update Transactions and Transactions Convenient API spec tests that explicitly required write commands to omit `readConcern`.

---

- [x] Update changelog.
    clusters).

### DIFF
--- a/source/causal-consistency/causal-consistency.md
+++ b/source/causal-consistency/causal-consistency.md
@@ -200,7 +200,7 @@ There are no new server commands related to causal consistency. Instead, causal 
     The server reports the `operationTime` whether the operation succeeded or not and drivers MUST save the
     `operationTime` in the `ClientSession` whether the operation succeeded or not.
 2. Passing that `operationTime` in the `afterClusterTime` field of the `readConcern` field for subsequent causally
-    consistent read operations (for all commands that support a `readConcern`)
+    consistent read and write operations (for all commands that support a `readConcern`)
 3. Gossiping clusterTime (described in the Driver Session Specification)
 
 ## Server Command Responses
@@ -230,10 +230,10 @@ standalone node are causally consistent automatically because there is only one 
 When connected to a deployment that supports cluster times the command response also includes a field called
 `$clusterTime` that drivers MUST use to gossip the cluster time. See the Sessions Specification for details.
 
-## Causally consistent read commands
+## Causally consistent read and write commands
 
 For causal consistency the driver MUST send the `operationTime` saved in the `ClientSession` as the value of the
-`afterClusterTime` field of the `readConcern` field:
+`afterClusterTime` field of the `readConcern` field for read and write commands:
 
 ```typescript
 {
--- a/source/causal-consistency/causal-consistency.md
+++ b/source/causal-consistency/causal-consistency.md
@@ -248,9 +248,7 @@ For causal consistency the driver MUST send the `operationTime` saved in the `Cl
 ```
 
 For the list of commands that support causally consistent reads, see the
-[ReadConcern](../read-write-concern/read-write-concern.md#read-concern) spec. The write commands `insert`, `update`,
-`delete`, `findAndModify`, and `bulkWrite` also accept `readConcern.afterClusterTime` when used in causally consistent
-sessions.
+[ReadConcern](../read-write-concern/read-write-concern.md#read-concern) spec.
 
 The driver MUST merge the `ReadConcern` specified for the operation with the `operationTime` from the `ClientSession`
 (which goes in the `afterClusterTime` field) to generate the combined `readConcern` to send to the server. If the level
@@ -406,8 +404,8 @@ resolving many discussions of spec details. A final reference implementation mus
 
 ## Changelog
 
-- 2026-05-04: Require `afterClusterTime` on write commands (`insert`, `update`, `delete`, `findAndModify`, `bulkWrite`)
-    in causally consistent sessions, not only on read commands. Added prose test 13.
+- 2026-05-04: Require `afterClusterTime` on all write commands in causally-consistent sessions, not only on read
+    commands.
 
 - 2024-02-08: Migrated from reStructuredText to Markdown.
 
--- a/source/causal-consistency/causal-consistency.md
+++ b/source/causal-consistency/causal-consistency.md
@@ -101,7 +101,7 @@ options = new SessionOptions(causalConsistency = true);
 session = client.startSession(options);
 ```
 
-All read operations performed using this session will now be causally consistent.
+All read and write operations performed using this session will now be causally consistent.
 
 If no value is provided for `causalConsistency` and snapshot reads are not requested a value of true is implied. See the
 `causalConsistency` section.
@@ -125,7 +125,7 @@ class SessionOptions {
 
 In order to support causal consistency a new property named `causalConsistency` is added to `SessionOptions`.
 Applications set `causalConsistency` when starting a client session to indicate whether they want causal consistency.
-All read operations performed using that client session are then causally consistent.
+All read and write operations performed using that client session are then causally consistent.
 
 Each new member is documented below.
 
@@ -218,8 +218,8 @@ and write operations).
 ```
 
 The `operationTime` MUST be stored in the `ClientSession` to later be passed as the `afterClusterTime` field of the
-`readConcern` field in subsequent read operations. The `operationTime` is returned whether the command succeeded or not
-and MUST be stored in either case.
+`readConcern` field in subsequent causally consistent read and write operations. The `operationTime` is returned whether
+the command succeeded or not and MUST be stored in either case.
 
 Drivers MUST examine all responses from the server for the presence of an `operationTime` field and store the value in
 the `ClientSession`.
@@ -237,7 +237,7 @@ For causal consistency the driver MUST send the `operationTime` saved in the `Cl
 
 ```typescript
 {
-    find : <string>, // or other read command
+    find : <string>, // or other read or write command
     ... // the rest of the command parameters
     readConcern :
     {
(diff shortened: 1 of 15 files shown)


source/causal-consistency/tests/causal-consistency-write-commands.yml
```diff
@@ -0,0 +1,320 @@
+description: "causal consistency write commands include afterClusterTime"
+
+schemaVersion: "1.3"
+
+runOnRequirements:
+  - minServerVersion: "3.6"
```
sanych-sun: Do we really need to specify such an ancient min version? We have 4.2 as the lowest supported by Drivers version as for now.

source/causal-consistency/tests/causal-consistency-write-commands.yml
```diff
@@ -0,0 +1,320 @@
...
+              command:
+                findAndModify: *collectionName
+                lsid: { $$sessionLsid: *session0 }
+                readConcern:
+                  afterClusterTime: { $$exists: true }
+
+  - description: "bulkWrite includes afterClusterTime in causally consistent session"
+    operations:
+      - *findOne
+      - name: bulkWrite
```
sanych-sun, line 295: What about `client.bulkWrite`?

source/causal-consistency/causal-consistency.md
```diff
@@ -200,7 +200,7 @@ There are no new server commands related to causal consistency. Instead, causal
     The server reports the `operationTime` whether the operation succeeded or not and drivers MUST save the
     `operationTime` in the `ClientSession` whether the operation succeeded or not.
 2. Passing that `operationTime` in the `afterClusterTime` field of the `readConcern` field for subsequent causally
-    consistent read operations (for all commands that support a `readConcern`)
+    consistent read and write operations (for all commands that support a `readConcern`)
```
sanych-sun, line 203: Should we specify list of commands here to avoid any confusion?

source/causal-consistency/tests/causal-consistency-write-commands.yml
matthewdale: I'll add a test for `clientBulkWrite`.

source/causal-consistency/tests/causal-consistency-write-commands.yml
matthewdale: Will remove.

source/causal-consistency/causal-consistency.md
matthewdale: Yes, I'll update the list in "Read Write Concern", which "Causal Consistency" links to.

sanych-sun (approved):
LGTM

dariakp:
@sanych-sun Could you link the C# implementation draft / CI for tests?

source/causal-consistency/tests/causal-consistency-clientBulkWrite.json
```diff
@@ -0,0 +1,146 @@
...
+        }
+      ]
+    }
+  ],
+  "tests": [
+    {
+      "description": "clientBulkWrite includes afterClusterTime in causally consistent session",
+      "operations": [
+        {
+          "name": "findOne",
```
sanych-sun: Could you please replace `findOne` with something else, because `findOne` is optional and not all drivers support it? `find` should be OK.

sanych-sun (changes requested):

source/causal-consistency/tests/causal-consistency-clientBulkWrite.json
matthewdale: Updated.

matthewdale:
Here's the Go Driver proof-of-concept implementation: https://github.com/mongodb/mongo-go-driver/pull/2380

source/read-write-concern/read-write-concern.md
```diff
@@ -68,7 +68,7 @@ class ReadConcern {
 }
 ```
 
-The read concern option is available for the following operations:
+The read concern option is available for most operations, including the following:
```
rozza: This comment seems a little loose and confusing. Is there a definitive list and for operations?  Can we clarify the scenario where insert will have a ReadConcern? Eg:

```
- `insert` command via causally-consistent sessions

File: source/read-write-concern/read-write-concern.md [hunk 6]
```diff
@@ -78,6 +78,16 @@ The read concern option is available for the following operations:
...
+- `insert` command
+- `update` command
+- `findAndModify` command
+- `delete` command
+- `bulkWrite` command
+- `create` command
+- `createIndexes` command
+- `drop` command
+- `dropDatabase` command
+- `dropIndexes` command
```
rozza: Does this change warrant the Changelog being updated?

rozza (changes requested):
Looking good - just a couple of questions.

Also should there be a `dropDatabase` test in `causal-consistency-write-commands.yml`?

sanych-sun (approved):
LGTM + [CSharp PR](https://github.com/mongodb/mongo-csharp-driver/pull/2005) + [custom patch](https://spruce.corp.mongodb.com/version/6a0ca8052dd0d200085c39d7/tasks) with all variants.

source/read-write-concern/read-write-concern.md
matthewdale: Unfortunately there is not a definitive list AFAIK. The public documentation for [Operations That Support Read Concern](https://www.mongodb.com/docs/manual/reference/read-concern/#operations-that-support-read-concern) provides a subset of supported operations, but the list is incomplete. In a previous thread, Max H. suggested using [this query for `allowsAfterClusterTime`](https://github.com/search?q=repo%3A10gen%2Fmongo%20allowsAfterClusterTime&type=code) in server code to determine which operations _don't_ support read concern with `afterClusterTime`, but I don't think it's practical to use that as the source of truth. I think the best we can do is test the server behavior with the unified spec tests added in this PR.

matthewdale:
@rozza There is no "dropDatabase" command supported by the unified test spec. That does means there's a coverage gap for that command in the Causal Consistency spec tests. I'll check if it's worth adding to the unified test spec.

source/read-write-concern/read-write-concern.md
matthewdale: Yes, it does. I'll update the changelog!

source/read-write-concern/read-write-concern.md
matthewdale: To clarify how/when read concern is sent for writes, I will separate the section for write commands and add a clarifying note.

source/read-write-concern/read-write-concern.md
```diff
@@ -533,6 +556,8 @@ instance, we send it.
 
 ## Changelog
 
+- 2025-05-04: Describe how `readConcern.afterClusterTime` is used in causally-consistent sessions on reads and writes.
```
rozza:
```suggestion
- 2026-05-04: Describe how `readConcern.afterClusterTime` is used in causally-consistent sessions on reads and writes.
```

rozza (approved):
LGTM!

I pushed a date nit fix, rather than dragging out time on a re-review.

The dropDatabase case is unusual with the unified specs - eg it could be added without bumping spec versions, but that could cause issues downstream. So I'm inclined to leave it up to you. The test plan already guides to cover all write scenarios.

source/causal-consistency/causal-consistency.md
qingyang-hu, line 292:
```suggestion
2. The first read or write in a causally consistent session must not send `afterClusterTime` to the server (because the
```
"write" seems to apply here as well. Also, do we need unified test cases to cover this scenario?

matthewdale: Yep, "write" seems to apply there. As far as a unified test case, I'll check with other driver teams if we can ever assume that "operationTime" is not set in unified test runners. As far as I can tell, no other tests currently assume that an operation in a test case is the very first operation for that `MongoClient` (i.e. unified test runners may reuse the same `MongoClient` across test cases).

matthewdale: After further investigation, I believe we can assume that the `MongoClient` has been reset at the beginning of a test case. I've added a unified spec test that covers the same condition as prose test 2, but for a write. That is preferable over a prose test change because it shouldn't require any additional code changes.

matthewdale:
@rozza I've added a unified spec test for "dropDatabase" and confirmed it works in the Go Driver proof-of-concept, after adding code to handle the "dropDatabase" test operation. I suspect most drivers will have to implement support for "dropDatabase".

dariakp:
@sanych-sun Could you pull in the new test into the C# implementation to make sure it works as expected?

sanych-sun:
@dariakp @matthewdale I've re-synced spec tests, all green after adding support for `dropDatabase` unified operation. Here is the [CSharp PR](https://github.com/mongodb/mongo-csharp-driver/pull/2005) + [EG patch](https://spruce.corp.mongodb.com/version/6a19be10de6f3900076286c4/tasks).

dariakp (approved):

qingyang-hu (approved):

PR #1933, opened 2026-05-08
DRIVERS-2944 Update handshake spec to allow using OP_MSG exclusively

sleepyStick:

- [x] Update changelog.
        Python driver PR: https://github.com/mongodb/mongo-python-driver/pull/2799
    clusters).

### DIFF
--- a/source/message/OP_MSG.md
+++ b/source/message/OP_MSG.md
@@ -409,6 +409,7 @@ In the near future, this opcode is expected to be extended and include support f
 
 ### Changelog
 
+- 2026-05-07: Allow OP_MSG for all handshakes.
 - 2024-04-30: Convert from RestructuredText to Markdown.
 - 2022-10-05: Remove spec front matter.
 - 2022-01-13: Clarify that `OP_MSG` must be used when using stable API
--- a/source/message/OP_MSG.md
+++ b/source/message/OP_MSG.md
@@ -409,7 +409,7 @@ In the near future, this opcode is expected to be extended and include support f
 
 ### Changelog
 
-- 2026-05-07: Allow OP_MSG for all handshakes.
+- 2026-05-08: Allow OP_MSG for all handshakes.
 - 2024-04-30: Convert from RestructuredText to Markdown.
 - 2022-10-05: Remove spec front matter.
 - 2022-01-13: Clarify that `OP_MSG` must be used when using stable API
--- a/source/message/OP_MSG.md
+++ b/source/message/OP_MSG.md
@@ -20,7 +20,7 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SH
 `OP_MSG` is only available in MongoDB 3.6 (`maxWireVersion >= 6`) and later. MongoDB drivers MUST perform the MongoDB
 handshake using `OP_MSG` if an API version was declared on the client.
 
-If no API version was declared, drivers that have historically supported MongoDB 3.4 and earlier MUST perform the
+If no API version was declared, drivers that have historically supported MongoDB 3.4 and earlier MAY perform the
 handshake using `OP_QUERY` to determine if the node supports `OP_MSG`. Drivers that have only ever supported MongoDB 3.6
 and newer MAY default to using `OP_MSG`.
 
--- a/source/message/OP_MSG.md
+++ b/source/message/OP_MSG.md
@@ -17,15 +17,11 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SH
 
 #### Usage
 
-`OP_MSG` is only available in MongoDB 3.6 (`maxWireVersion >= 6`) and later. MongoDB drivers MUST perform the MongoDB
-handshake using `OP_MSG` if an API version was declared on the client.
-
-Refer to the
+All messages, including authentication messages, MUST use `OP_MSG`. Refer to the
 [handshake specification](https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.md)
-for the appropriate use of `OP_MSG`.
-
-If the node supports `OP_MSG`, any and all messages MUST use `OP_MSG`, optionally compressed with `OP_COMPRESSED`.
-Authentication messages MUST also use `OP_MSG` when it is supported, but MUST NOT use `OP_COMPRESSED`.
+for the appropriate use of `OP_MSG`. See also
+[OP_COMPRESSED specification](https://github.com/mongodb/specifications/blob/master/source/compression/OP_COMPRESSED.md)
+for more details.
 
 #### OP_MSG
 
@@ -409,7 +405,7 @@ In the near future, this opcode is expected to be extended and include support f
 
 ### Changelog
 
-- 2026-06-03: Allow OP_MSG for all handshakes.
+- 2026-06-05: Use OP_MSG for all messages.
 - 2024-04-30: Convert from RestructuredText to Markdown.
 - 2022-10-05: Remove spec front matter.
 - 2022-01-13: Clarify that `OP_MSG` must be used when using stable API
--- a/source/message/OP_MSG.md
+++ b/source/message/OP_MSG.md
@@ -409,7 +409,7 @@ In the near future, this opcode is expected to be extended and include support f
 
 ### Changelog
 
-- 2026-05-08: Allow OP_MSG for all handshakes.
+- 2026-06-03: Allow OP_MSG for all handshakes.
 - 2024-04-30: Convert from RestructuredText to Markdown.
 - 2022-10-05: Remove spec front matter.
 - 2022-01-13: Clarify that `OP_MSG` must be used when using stable API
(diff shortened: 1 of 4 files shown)


sleepyStick:
removed @blink1073 (OOO currently) and requested @vbabanin instead :)

source/mongodb-handshake/handshake.md
```diff
@@ -49,11 +49,12 @@ MongoDB uses the `hello` or `isMaster` commands for handshakes and topology moni
...
-If a [server API version](../versioned-api/versioned-api.md) is requested or `loadBalanced: True`, drivers MUST use the
-`hello` command for the initial handshake and use the `OP_MSG` protocol. If server API version is not requested and
-`loadBalanced: False`, drivers MUST use legacy hello for the first message of the initial handshake with the `OP_QUERY`
-protocol (before switching to `OP_MSG` if the `maxWireVersion` indicates compatibility), and include `helloOk:true` in
-the handshake request.
+Drivers SHOULD use the `OP_MSG` protocol for all handshakes if their minWireVersion is 6 (MongoDB 3.6) or higher. If a
+[server API version](../versioned-api/versioned-api.md) is requested or `loadBalanced: True`, drivers MUST also use the
+`hello` command for the initial handshake. If server API version is not requested and `loadBalanced: False`, drivers
+MUST use legacy hello for the first message of the initial handshake, and include `helloOk:true` in the handshake
+request. If the server does not understand `OP_MSG`, drivers MUST show the same error message as when wire version
```
jyemin: The driver's only visibility into whether the server understands `OP_MSG` is, I believe:

1. Socket connect succeeds
2. OP_MSG send
3. Socket is closed by the server 

#3 can happen for multiple reasons.  The most likely is that the server does not understand `OP_MSG`, but it's also possible that the server just happened to get shut down right after connect succeeds, or some other network error occurs. Either way it looks like a socket error to the driver, and I think the error message needs to account for that ambiguity.  It should include the original socket exception/error, and then add a message on top about the likely cause (an incompatible server).

source/mongodb-handshake/handshake.md
```diff
@@ -99,9 +99,9 @@ if creds:
 
 reply = conn.send_command("admin", cmd)
 
-if reply["maxWireVersion"] >= 6:
-    # Use OP_MSG for all future commands, including authentication.
-    conn.supports_op_msg = True
+if reply["maxWireVersion"] < 6:
+    # Server is reporting that it doesn't support OpMSG
```
jyemin: See above.  The server "reports" it doesn't support OpMSG by hanging up the socket.

source/message/OP_MSG.md
```diff
@@ -20,7 +20,7 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SH
 `OP_MSG` is only available in MongoDB 3.6 (`maxWireVersion >= 6`) and later. MongoDB drivers MUST perform the MongoDB
 handshake using `OP_MSG` if an API version was declared on the client.
 
-If no API version was declared, drivers that have historically supported MongoDB 3.4 and earlier MUST perform the
+If no API version was declared, drivers that have historically supported MongoDB 3.4 and earlier MAY perform the
```
jyemin: I feel like this historical distinction is now irrelevant.  Should we instead just say that drivers SHOULD use OP_MSG for the handshake, and MUST if load balanced mode is on or API version is specified?

Alternatively, we can just link to the handshake spec since this is duplicating what is already says there.

source/mongodb-handshake/handshake.md
```diff
@@ -49,11 +49,12 @@ MongoDB uses the `hello` or `isMaster` commands for handshakes and topology moni
 preferred command. `hello` must always be sent using the `OP_MSG` protocol. `isMaster` is referred to as "legacy hello"
 and is maintained for backwards compatibility with servers that do not support the `hello` command.
 
-If a [server API version](../versioned-api/versioned-api.md) is requested or `loadBalanced: True`, drivers MUST use the
-`hello` command for the initial handshake and use the `OP_MSG` protocol. If server API version is not requested and
-`loadBalanced: False`, drivers MUST use legacy hello for the first message of the initial handshake with the `OP_QUERY`
-protocol (before switching to `OP_MSG` if the `maxWireVersion` indicates compatibility), and include `helloOk:true` in
-the handshake request.
+Drivers SHOULD use the `OP_MSG` protocol for all handshakes if their minWireVersion is 6 (MongoDB 3.6) or higher. If a
```
jyemin: What's the justification for SHOULD rather than MUST?  Is there a reason for any driver not to do this?

jyemin (changes requested):

source/mongodb-handshake/handshake.md
sleepyStick: tbh i wrote SHOULD because that's what the driver ticket said https://jira.mongodb.org/browse/DRIVERS-2944
(And I justified it to myself by saying that if a driver doesn't do this they're just doing the current behaviour which isn't inherently breaking anything?)

source/message/OP_MSG.md
sleepyStick: I like the idea of linking to the handshake spec!

nbbeeken (approved):
🎉 its happeninggggg!!!! exciting times 😁

source/mongodb-handshake/handshake.md
sleepyStick: Thank you for pointing this out (sorry i didn't say it earlier!) per our conversations, I've updated it to reflect that the existing socket error should be propagated as it seems unlikely for uses to run into said socket error due to outdated server version and they should update their server if so. 
(also added a unified test that utilizes the fail point command like we briefly discussed) 
If i've misunderstood any of our conversations, please let me know! 🙇

source/message/OP_MSG.md
```diff
@@ -20,9 +20,9 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SH
 `OP_MSG` is only available in MongoDB 3.6 (`maxWireVersion >= 6`) and later. MongoDB drivers MUST perform the MongoDB
```
jyemin: The second sentence is now misleading, since drivers should now use OP_MSG in all situations.  Do you think we should just remove the whole section, except for the link to the handshake spec? None of it seems relevant anymore.

source/message/OP_MSG.md
```diff
@@ -20,9 +20,9 @@ The keywords "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SH
...
 handshake using `OP_MSG` if an API version was declared on the client.
 
-If no API version was declared, drivers that have historically supported MongoDB 3.4 and earlier MUST perform the
-handshake using `OP_QUERY` to determine if the node supports `OP_MSG`. Drivers that have only ever supported MongoDB 3.6
-and newer MAY default to using `OP_MSG`.
+Refer to the
+[handshake specification](https://github.com/mongodb/specifications/blob/master/source/mongodb-handshake/handshake.md)
+for the appropriate use of `OP_MSG` and `OP_QUERY`.
```
jyemin:
```suggestion
for the appropriate use of `OP_MSG`.
```

source/mongodb-handshake/handshake.md
```diff
@@ -78,12 +79,11 @@ Consider the following pseudo-code for establishing a new connection:
 ```python
 conn = Connection()
 conn.connect()  # Connect via TCP / TLS
+conn.supports_op_msg = True  # Always send the initial command via OP_MSG.
```
jyemin: Consider removing this since it can never be False (there is no `OP_QUERY` fallback)?

source/mongodb-handshake/handshake.md
```diff
@@ -49,11 +49,12 @@ MongoDB uses the `hello` or `isMaster` commands for handshakes and topology moni
...
-If a [server API version](../versioned-api/versioned-api.md) is requested or `loadBalanced: True`, drivers MUST use the
-`hello` command for the initial handshake and use the `OP_MSG` protocol. If server API version is not requested and
-`loadBalanced: False`, drivers MUST use legacy hello for the first message of the initial handshake with the `OP_QUERY`
-protocol (before switching to `OP_MSG` if the `maxWireVersion` indicates compatibility), and include `helloOk:true` in
-the handshake request.
+Drivers MUST use the `OP_MSG` protocol for all handshakes if their minWireVersion is 6 (MongoDB 3.6) or higher. If a
+[server API version](../versioned-api/versioned-api.md) is requested or `loadBalanced: True`, drivers MUST also use the
+`hello` command for the initial handshake. If server API version is not requested and `loadBalanced: False`, drivers
+MUST use legacy hello for the first message of the initial handshake, and include `helloOk:true` in the handshake
+request. If the server does not understand `OP_MSG`, the server will close the socket and drivers MUST include the
```
jyemin: The reference to `original error` is confusing.  It's just the error.  And it's not clear what it means by `include`.  What code path is this referring to?

From what I understand, there is no special handling in the driver at all for this, so the spec doesn't have to say anything in particular in terms of MUST/SHOULD.  

Noting the server behavior seems like a good aside though, since it's not obvious what will happen in that scenario.

jyemin (changes requested):

source/mongodb-handshake/handshake.md
sleepyStick: good point -- keeping it feels silly now HAHA

jyemin (approved):
OP_QUERY is dead!  Long live OP_MSG!

Oh also, LGTM.

Jibola (approved):

Jibola:
Bypassing and merging because @jyemin is one of the `mongodb/dbx-spec-maintainters-connection-establishment` members (per the Client Libraries Common Spec Maintainers document)

PR #1942, opened 2026-05-13
DRIVERS-3475 Add DBRef test fixtures for encoding and decoding

GromNaN:
Jira: https://jira.mongodb.org/browse/DRIVERS-3475

Converts the prose test plan from `dbref.md` into machine-readable YAML fixtures; covering the 4 behavioral areas described in the spec:

- `valid-documents` — 9 valid documents that MUST decode to a DBRef model (`$id` as ObjectId/integer/null, with `$db`, with extra fields including dollar-prefixed and dot-notation names)
- `invalid-documents` — 5 documents that MUST NOT decode (missing `$ref`, missing `$id`, wrong types for `$ref`/`$db`)
- `out-of-order` — 5 documents with non-canonical field order that MUST decode correctly and MUST re-encode in canonical order (`$ref`, `$id`, `$db`, then extra fields)
- `encoding` — 9 DBRef models that MUST encode with fields in canonical order, preserving relative order of extra fields

## Relationship with bson-corpus

`source/bson-corpus/tests/dbref.json` already tests DBRef at the **BSON/Extended JSON layer** (byte sequences ↔ Extended JSON round-trips). Documents with missing `$id`, non-string `$ref`, etc. appear there as *valid BSON* — which they are; bson-corpus does not prescribe driver model behavior.

These tests are **complementary**, not redundant:

| Concern | bson-corpus | These tests |
|---------|-------------|-------------|
| BSON bytes ↔ Extended JSON round-trip | ✓ | — |
| Driver decodes valid DBRef into a model object | — | ✓ |
| Driver exposes `$ref`, `$id`, `$db`, extra fields | — | ✓ |
| Driver rejects invalid documents (missing `$id`, wrong types) | — | ✓ |
| Out-of-order fields decoded and re-encoded in canonical order | — | ✓ |
| Encoding produces canonical field order | — | ✓ |

### DIFF
--- a/source/dbref/tests/README.md
+++ b/source/dbref/tests/README.md
@@ -1,0 +1,75 @@
+# DBRef Tests
+
+These tests cover driver-level DBRef **model** encoding and decoding behavior. They do not require a running MongoDB
+server.
+
+## Relationship with bson-corpus
+
+`source/bson-corpus/tests/dbref.json` already tests DBRef at the **BSON/Extended JSON layer**: it verifies that specific
+byte sequences round-trip correctly through a driver's BSON codec. Documents with missing `$id`, non-string `$ref`, etc.
+appear there as *valid BSON documents* — because they are valid BSON; the bson-corpus does not prescribe driver model
+behavior.
+
+The tests in this directory complement bson-corpus by testing the **driver model layer**:
+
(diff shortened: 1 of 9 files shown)


GromNaN:
Closing as this is adding a new spec test format. It's not worth the effort since we already have prose tests.

PR #1949, opened 2026-06-11
DRIVERS-2944 followup: remove test and clarify spec

sleepyStick:

- [x] Update changelog.
    clusters).

followup PR to https://github.com/mongodb/specifications/pull/1933
the test was broken / didn't correctly test what we wanted to test. Now we defer to SDAM testing.


source/mongodb-handshake/handshake.md
```diff
@@ -57,10 +57,12 @@ request.
 
 ASIDE: If the legacy handshake response includes `helloOk: true`, then subsequent topology monitoring commands MUST use
```
jyemin:
```suggestion
If the legacy handshake response includes `helloOk: true`, then subsequent topology monitoring commands MUST use
```

jyemin: Seems strange now to have a MUST in a paragraph starting with ASIDE.

jyemin (changes requested):
One small suggestion.

source/mongodb-handshake/handshake.md
sleepyStick: okay fair haha

jyemin (approved):

Jibola (approved):

